PR #1, opened 2018-11-22
Adapt code to the new user-schemas

jmelis:


pbergene:
This is starting to take off :grinning:

PR #6, opened 2018-12-05
Add openshift rolebinding integration

jmelis:


riuvshin:
it looks good, but I haven't tested it.
It would be great if we will have some tests to track regressions on it (we have already discussed that and tests can be added later)

reconcile/openshift_rolebinding.py
```diff
@@ -0,0 +1,226 @@
...
+    cluster_store = ClusterStore()
+
+    current_state = fetch_current_state(cluster_store)
+    desired_state = fetch_desired_state()
+
+    # calculate diff
+    diff = current_state.diff(desired_state)
+
+    # ensure all namespace/roles are well known
+    if diff['insert'] != []:
```
jzelinskie: You should probably compare the length here to zero rather for sanity sake.

reconcile/openshift_rolebinding.py
```diff
@@ -0,0 +1,226 @@
...
+        keys = ['cluster', 'namespace', 'role']
+
+        unknown_combinations = ", ".join([
+            "/".join([item['params'][k] for k in keys])
+            for item in diff['insert']
+        ])
```
jzelinskie: This seems a bit dense.

reconcile/openshift_rolebinding.py
```diff
@@ -0,0 +1,226 @@
...
+
+        raise RunnerException((
+            "Unkown cluster/namespace/combinations found:\n"
```
jzelinskie: unknown

reconcile/openshift_rolebinding.py
```diff
@@ -0,0 +1,229 @@
...
+
+class ClusterStore(object):
+    _clusters = {}
+
+    def __init__(self):
+        gqlapi = gql.get_api()
+        result_json = gqlapi.query(CLUSTER_CATALOG_QUERY)
+        result = json.loads(result_json)
+
+        for cluster_info in result['data']['cluster']:
```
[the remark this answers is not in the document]
jmelis: Thanks, that was a very big overlook. It turns out whenever there is a GraphQL error, the result is 200 OK, but it comes with another key called `errors`. I improved the logic so the client aborts if there are errors. It also raises an error if `data` is not present.

reconcile/openshift_rolebinding.py
jmelis: agreed

reconcile/openshift_rolebinding.py
jmelis: thanks, improving this logic

reconcile/openshift_rolebinding.py
jmelis: fixed

reconcile/openshift_rolebinding.py
```diff
@@ -0,0 +1,229 @@
...
+    def clusters(self):
+        return self._clusters.keys()
+
+    def api(self, cluster):
+        return self._clusters[cluster]['api']
+
+    def namespaces(self, cluster):
+        return [
+            r['namespace']
+            for r in self._clusters[cluster]['managed_roles']
```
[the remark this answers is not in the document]
jmelis: thanks, totally agree

reconcile/openshift_rolebinding.py
```diff
@@ -0,0 +1,229 @@
...
+ROLEBINDINGS_QUERY = """
+{
+  role {
+    name
+    members {
+      ...on Bot_v1 {
+        schema
+        github_username_optional: github_username
+      }
+      ... on User_v1 {
```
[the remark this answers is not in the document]
jmelis: fixed

reconcile/openshift_api.py
```diff
@@ -0,0 +1,360 @@
+import requests
```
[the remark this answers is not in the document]
jmelis: Very good idea, can you please review that the newly uploaded structure makes sense in your opinion?

reconcile/vault_client.py
```diff
@@ -0,0 +1,31 @@
...
+from reconcile.config import get_config
+
+_client = None
+
+
+def init(server, role_id, secret_id):
+    global _client
+
+    if _client is None:
+        _client = hvac.Client(url=server)
```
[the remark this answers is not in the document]
jmelis: for the moment going to merge as is, will revisit this later on

PR #7, opened 2019-01-08
Quay org

jmelis:
This PR adds`qontract-reconcile quay-org` which manages the membership of users into quay orgs and teams.


.gitignore
```diff
@@ -1,6 +1,6 @@
 *.pyc
 *.egg-info
-*.toml
+/*.toml
```
jmelis: `*.toml` is ignored because we don't want to commit secret tokens accidentally. But inside `reconcile/test/fixtures` there are some toml files that need to be commited, that's the reason for this change.

reconcile/quay_org.py
```diff
@@ -0,0 +1,184 @@
...
+
+import reconcile.gql as gql
+import utils.vault_client as vault_client
+from utils.quay_api import QuayApi
+
+from reconcile.aggregated_list import (AggregatedList,
+                                       AggregatedDiffRunner,
+                                       RunnerException)
+
+QUAY_ORG_CATALOG_QUERY = """
```
jmelis: The app-interface schemas contain the list of managed quay orgs and teams as specified in this schema:
https://github.com/app-sre/qontract-server/pull/11/files#diff-a3f753bf6d00d416297a369b20e319e7

PR #11, opened 2019-02-10
WIP: LDAP integration

maorfr:


jmelis:
I'll do a more in-depth review once you remove the `WIP` tag, but it's looking really good!! Was surprised to see that you are using AggregateList, hope it wasn't too confusing! It's probably a bit of an overkill, but using the RunnerAction is quite cool.

One little nitpick, if I'm not mistaken in the rest of integrations `desired_state` is response of the GraphQL query, and current state is the third-party service (in this case LDAP), I think you have it the other way around here. In any case, it was quite clever to pass the result of the gql query as an argument :)

maorfr:
Thanks for your feedback :)

AggregateList is indeed an overkill. I started out without it, but I found myself duplicating code from `diff`, which is something I would rather avoid. RunnerAction is another incentive to use it, indeed :)

You are also right about `desired_state`, but this doesn't feel like a classic current-desired use case. Maybe more similar to desired-valid? As the gql response represents the desired state, and the ldap response determines if this desire is at all valid.

I can change the functions accordingly if that makes more sense.

jmelis:
On second thought, since this is really simple, I think I would do something
like:

```python
import logging

import reconcile.gql as gql
import utils.ldap_client as ldap_client
# TODO: load gitlab API

USERS_QUERY = """
{
  user {
    path
    redhat_username
  }
}
"""

def remove_user_mr(path):
    # send merge request to remove user from app-interface repo
    # 
    # we should add the the path to the repo as a conf option in config.toml.
    # the MR should follow a naming convention. The first would be to check 
    # if the MR already exists before opening a new one... this has to be 
    # idempotent

    pass

def run(dry_run=False):
    gqlapi = gql.get_api() result =
    gqlapi.query(USERS_QUERY)

    for user in result['user']:
        username = user['redhat_username']
        path = user['path']

        if not ldap_client.user_exists(username):
            logging.info(['delete_user', username, path])

            if not dry_run:
                remove_user_mr(path)
```

maorfr:
Looks better. I'll do as you suggested :)

maorfr:

PR #12, opened 2019-02-11
LDAP and GitLab integration

maorfr:


jmelis:
This looks great!

before I do a more in-depth review, there's something that I'd like to discuss -

If you look at the [vault_client](https://github.com/app-sre/qontract-reconcile/blob/master/utils/vault_client.py) you'll see that it doesn't use classes, it is a module based functionality, which makes sense because we only have one Vault, so no need to create a class around this.

However, if you look at the [quay_api](https://github.com/app-sre/qontract-reconcile/blob/master/utils/quay_api.py) for instance, you'll see that it indeed uses a class, and that it doesn't have the `init_from_config`, etc methods.

I somehow feel that the `gitlab_api` should be a class, much like `quay_api`, perhaps at some point we want to manage more than one gitlab repo? For LDAP I also thing it'd better be a class, but I feel less strongly about this... I don't know. What do you think? I using a class makes sense if at some point we will interact with more than just one instance.

maorfr:
class it is!

utils/gitlab_api.py
```diff
@@ -0,0 +1,97 @@
...
+        prefix = 'qcontract-reconcile'
+        target_branch = 'master'
+        branch_name = '{}-delete-{}-{}'.format(
+            prefix,
+            username,
+            str(uuid.uuid4())[0:6]
+        )
+        title = '[{}] delete user {} ({})'.format(prefix, username, path)
+
+        if (self.delete_user_mr_exists(title)):
```
jmelis: unnecessary outer parens

config.toml.example
```diff
@@ -25,3 +25,20 @@ token = "<token>"
...
+
+# This section is required by `qcontract-reconcile ldap-users`
+[gitlab]
+# These parameters can be overridden at project level
+server = "https://gitlab.example.com"
+token = "..."
+
+# One sub-section per managed GitLab project
+[gitlab.<prject_name>]
+project_id = "12345"
```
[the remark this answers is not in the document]
jmelis: OK, instead of `[gitlab]` and `[gitlab.<project_name>]` I suggest we have a single:

```
[app-interface-repo]
server = ...
token = ...
project_id = ...
```

utils/gitlab_api.py
```diff
@@ -0,0 +1,97 @@
...
+
+    server = init_key_from_config('server', gitlab_config, project_config)
+    token = init_key_from_config('token', gitlab_config, project_config)
+    project_id = project_config['project_id']
+
+    gl = gitlab.Gitlab(server, private_token=token, ssl_verify=ssl_verify)
+    return gl.projects.get(project_id)
+
+
+def init_key_from_config(key, *configs):
```
jmelis: we probably don't need this if we move from `[gitlab]` to `[app-interface-repo]`

utils/ldap_client.py
```diff
@@ -0,0 +1,38 @@
...
+    config = get_config()
+
+    serverUrl = config['ldap']['server']
+    _base_dn = config['ldap']['base_dn']
+
+    return init(serverUrl)
+
+
+def user_exists(username):
+
```
jmelis: perhaps remove this empty line?

utils/gitlab_api.py
```diff
@@ -0,0 +1,97 @@
...
+
+        if (self.delete_user_mr_exists(title)):
+            return
+
+        self.create_branch(branch_name, target_branch)
+        self.delete_file(branch_name, path, title)
+        self.create_mr(branch_name, target_branch, title)
+
+
+def init_project_from_config(project_name, ssl_verify):
```
jmelis: instead of this, I think I would use a method like:

`def get_app_interface_gitlab_api`

which returns a new instance of the `GitLabApi`

using the config file

utils/gitlab_api.py
```diff
@@ -0,0 +1,97 @@
+import gitlab
+import urllib3
+import uuid
+from reconcile.config import get_config
+
+
+class GitLabApi(object):
+    def __init__(self, project_name, ssl_verify=False):
```
jmelis: I would change this to be:

`def __init__(self, server, token, project_id, ssl_verify=True):`

and create the gl api internally here:

```
self.gl = gitlab.Gitlab(server, private_token=token, ssl_verify=ssl_verify)
self.project = self.gl.projects.get(project_id)
```

jmelis: note that I would default to ssl_verify=True

utils/gitlab_api.py
```diff
@@ -0,0 +1,97 @@
...
+            'remove_source_branch': str(remove_source_branch)
+        }
+        self.project.mergerequests.create(data)
+
+    def delete_user_mr_exists(self, title):
+        mrs = self.project.mergerequests.list(state='opened')
+        for mr in mrs:
+            # since we are using a naming convention for these MRs
+            # we can determine if a pending MR exists based on the title
+            if (mr.attributes['title'] != title):
```
jmelis: unnecessary outer paens

jmelis: also, perhaps you should do mr.attributes.get('title') instead, who knows if GLApi always returns that...

utils/ldap_client.py
jmelis: how about we follow the same approach for the ldap_client than for the gitlab_client? I think it's better so we can reuse  this logic for other ldap servers in the future.

This would mean:
- no `init` methods
- define class that applies contains the functionality
- new method that returns the instantiated class using the toml config `def get_ldap_client_from_config`

jmelis: after discussing with @maorfr , we believe this is not necessary, and we can leave it as is

utils/gitlab_api.py
```diff
@@ -0,0 +1,73 @@
...
+                  remove_source_branch=True):
+        data = {
+            'source_branch': source_branch,
+            'target_branch': target_branch,
+            'title': title,
+            'remove_source_branch': str(remove_source_branch)
+        }
+        self.project.mergerequests.create(data)
+
+    def delete_user_mr_exists(self, title):
```
jmelis: better rename this to `mr_exists` (this is not related to `delete_user`)

utils/gitlab_api.py
```diff
@@ -0,0 +1,73 @@
+import gitlab
+import urllib3
+import uuid
+
+
+class GitLabApi(object):
+    def __init__(self, server, token, project_id, ssl_verify=True):
+        # The following line will supress
+        # `InsecureRequestWarning: Unverified HTTPS request is being made`
+        urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
jmelis: can you put this on top of the `class` statement?

maorfr:

PR #13, opened 2019-02-12
Handle ldap race condition gracefully

maorfr:


utils/gitlab_api.py
```diff
@@ -70,5 +73,16 @@ def create_delete_user_mr(self, username, path):
...
         self.create_branch(branch_name, target_branch)
-        self.delete_file(branch_name, path, title)
+
+        try:
+            self.delete_file(branch_name, path, title)
+        except gitlab.exceptions.GitlabCreateError as e:
+            self.delete_branch(branch_name)
+            if str(e) != "400: A file with this name doesn't exist":
+                raise e
+            print(
```
jmelis: use `logging.info`

utils/gitlab_api.py
```diff
@@ -70,5 +73,16 @@ def create_delete_user_mr(self, username, path):
             return
 
         self.create_branch(branch_name, target_branch)
-        self.delete_file(branch_name, path, title)
+
+        try:
+            self.delete_file(branch_name, path, title)
+        except gitlab.exceptions.GitlabCreateError as e:
+            self.delete_branch(branch_name)
```
jmelis: This is very clever! I would've forgotten to do this...

utils/gitlab_api.py
maorfr: Doh!

maorfr:

PR #15, opened 2019-02-19
Apply k8s

jmelis:


reconcile/openshift_resources.py
```diff
@@ -0,0 +1,271 @@
...
+
+        # Obtain `oc` client
+        cluster_info = namespace_info['cluster']
+        cluster = cluster_info['name']
+        if oc_map.get(cluster) is None:
+            at = cluster_info.get('automationToken')
+
+            # Skip if cluster has no automationToken
+            if at is None:
+                continue
```
riuvshin: maybe it worth to rise WARN or something here because it might be tricky to debug for new cluster why it does nothing. WDYT?

jmelis: ah, I see what you mean. If it has managedRoles or managedResourceTypes but no automationToken, it should definitely fail, yeah. I think we should also fail the PR.

Making the change.

reconcile/openshift_resources.py
```diff
@@ -0,0 +1,277 @@
...
+    for c, n, rt, data in ri.iterate():
+        # desired items
+        for name, d_item in data['desired'].items():
+            c_item = data['current'].get(name)
+
+            if c_item is not None:
+                if c_item.has_qontract_annotations():
+                    if c_item.sha256sum() == d_item.sha256sum():
+                        # don't apply if sha256sum hashes match
+                        continue
```
riuvshin: would it make sense to add DEBUG log here which can help to trouble shoot things?

jmelis: :+1:

reconcile/openshift_resources.py
```diff
@@ -0,0 +1,285 @@
...
+        e_msg = "Invalid data (TypeError). Skipping resource: {}"
+        raise FetchResourceError(e_msg.format(path))
+    except KeyError:
+        e_msg = "Invalid data (KeyError). Skipping resource: {}"
+        raise FetchResourceError(e_msg.format(path))
+
+    return openshift_resource
+
+
+def fetch_data(namespaces_query):
```
jzelinskie: like you said, this function is huuuuge

reconcile/openshift_resources.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def add_desired(self, cluster, namespace, resource_type, name, value):
+        desired = self._dict[cluster][namespace][resource_type]['desired']
+        desired[name] = value
+
+    def add_current(self, cluster, namespace, resource_type, name, value):
+        current = self._dict[cluster][namespace][resource_type]['current']
+        current[name] = value
+
+    def iterate(self):
```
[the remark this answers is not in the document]
jmelis: brilliant, thanks

jmelis:
@jzelinskie I have applied all your proposed changes, except the refactoring of `fetch_data` which is a bit clunky after refactoring.

reconcile/openshift_resources.py
jmelis: leaving this as is for the moment

PR #19, opened 2019-02-24
general refactor, add secrets integration

maorfr:


reconcile/openshift_resources.py
```diff
@@ -111,14 +120,49 @@ def fetch_provider_resource(path):
...
+def fetch_provider_vault_secret(name, path):
+    content = """apiVersion: v1
+kind: Secret
+metadata:
+  name: {}
+type: Opaque
+data:""".format(name)
+
+    try:
+        body = anymarkup.parse(
```
jmelis: instead of calling `anymarkup` here, could we just build a python dict?

reconcile/openshift_resources.py
```diff
@@ -111,14 +120,49 @@ def fetch_provider_resource(path):
 
     try:
         openshift_resource.verify_valid_k8s_object()
-    except (KeyError, TypeError):
+    except (KeyError, TypeError) as e:
```
jmelis: THANKS!

reconcile/openshift_resources.py
```diff
@@ -274,8 +329,11 @@ def run(dry_run=False):
             logging.info(['apply', cluster, namespace,
                           resource_type, d_item.name])
 
-            logging.debug("CURRENT: " +
-                          OR.serialize(OR.canonicalize(c_item.body)))
+            if c_item is None:
+                logging.debug("CURRENT: None")
+            else:
+                logging.debug("CURRENT: " +
+                              OR.serialize(OR.canonicalize(c_item.body)))
```
jmelis: thanks for catching this bug!

utils/vault_client.py
```diff
@@ -48,3 +48,15 @@ def read(path, field):
...
+
+
+def read_all(path):
+    global _client
+    init_from_config()
+
+    secret = _client.read(path)
+
+    if secret is None or 'data' not in secret:
+        raise SecretNotFound(path)
```
jmelis: minor pet-peeve, but I kinda prefer try/except KeyError

reconcile/openshift_resources.py
jmelis: can you submit this as a separate PR, so I can merge, and you can rebase?
I need the fix for this ASAP in the master branch, as running integrations would fail now.

utils/vault_client.py
maorfr: Plus, I was going for compliance:

https://github.com/app-sre/qontract-reconcile/blob/2237e286aef2ba6cc3cf746d3b204b8646292472/utils/vault_client.py#L42

would you suggest we change the `read` function as well?

maorfr: Does this make more sense?

```
secret = _client.read(path)

    if secret is None:
        raise SecretNotFound(path)

    try:
        return secret['data']
    except KeyError:
        raise SecretHasNoData(path)
```

reconcile/openshift_resources.py
maorfr: Done!

maorfr: added as part of #20

reconcile/openshift_resources.py
```diff
@@ -89,6 +104,47 @@ def __iter__(self):
                     yield (cluster, namespace, resource_type, data)
 
 
+_error_occured = False
+_error_prefix = ''
+
+
+def error(msg):
+    global _error_occured
```
jmelis: cc @jzelinskie wdyt?

reconcile/openshift_resources.py
```diff
@@ -119,107 +175,134 @@ def fetch_provider_resource(path):
...
+        for item in oc.get_items(resource_type, namespace=namespace):
+            openshift_resource = OR(item)
+            ri.add_current(
+                cluster,
+                namespace,
+                resource_type,
+                openshift_resource.name,
+                openshift_resource
+            )
+    return ri
```
jmelis: unnecessary

reconcile/openshift_resources.py
```diff
@@ -119,107 +175,134 @@ def fetch_provider_resource(path):
...
+
+    return openshift_resource
+
+
+def fetch_openshift_resource(resource):
+    provider = resource['provider']
+    path = resource['path']
+
+    if provider == 'resource':
+        try:
```
jmelis: remove the try/except blocks from this function and put them in `fetch_desired_state`

reconcile/openshift_resources.py
```diff
@@ -119,107 +175,134 @@ def fetch_provider_resource(path):
...
+            error("Unknown kind: {}.".format(openshift_resource.kind))
+            continue
+        except ResourceKeyExistsError:
+            # This is failing because an attempt to add
+            # a desired resource with the same name and
+            # the same type was already added previously
+            error("Desired item already exists: {}/{}.".format(
+                openshift_resource.kind, openshift_resource.name))
+            continue
+    return ri
```
jmelis: unnecessary

reconcile/openshift_resources.py
```diff
@@ -267,8 +350,7 @@ def run(dry_run=False):
                         "Skipping resource '{}/{}' in '{}/{}'. "
                         "Present w/o annotations."
                     ).format(resource_type, name, cluster, namespace)
-                    logging.error(e_msg)
-                    errors.append(e_msg)
+                    error(e_msg)
```
jmelis: I think this error would display the last error_prefix that was updated inside fetch_data

reconcile/openshift_resources.py
```diff
@@ -119,107 +175,134 @@ def fetch_provider_resource(path):
...
-                error("Cluster {} has no automationToken.")
-                oc_map[cluster] = False
-            else:
-                token = vault_client.read(at['path'], at['field'])
-                oc_map[cluster] = OC(cluster_info['serverUrl'], token)
-
-        oc = oc_map[cluster]
-
+        cluster_info = namespace_info['cluster']
+        oc = obtain_oc_client(oc_map, cluster_info)
```
jmelis: very nice, I like this idea

reconcile/openshift_resources.py
```diff
@@ -119,107 +175,134 @@ def fetch_provider_resource(path):
...
+        "metadata": {
+            "name": name
+        },
+        "data": {}
+    }
+
+    # get the fields from vault
+    raw_data = vault_client.read_all(path)
+    for k, v in raw_data.items():
+        body['data'][k] = base64.b64encode(v)
```
jmelis: I checked and this doesn't wrap the text with newlines, so that's good :)

reconcile/openshift_resources.py
```diff
@@ -119,107 +175,134 @@ def fetch_provider_resource(path):
...
+            # combination was not initialized, meaning that it shouldn't be
+            # managed. But someone is trying to add it via app-interface
+            error("Unknown kind: {}.".format(openshift_resource.kind))
+            continue
+        except ResourceKeyExistsError:
+            # This is failing because an attempt to add
+            # a desired resource with the same name and
+            # the same type was already added previously
+            error("Desired item already exists: {}/{}.".format(
+                openshift_resource.kind, openshift_resource.name))
```
jmelis: very nice!

reconcile/openshift_resources.py
maorfr: indeed good, i checked that as well ;)

reconcile/openshift_resources.py
maorfr: thanks!

reconcile/openshift_resources.py
maorfr: thanks!
i usually find that comments are the best indication that this should be a function

reconcile/openshift_resources.py
maorfr: fixed

reconcile/openshift_resources.py
maorfr: fixed

reconcile/openshift_resources.py
maorfr: fixed

reconcile/openshift_resources.py
maorfr: fixed by resetting the prefix

PR #24, opened 2019-02-28
don't add empty labels, add desired debug message, fix secrets null value handle

maorfr:


maorfr:
@jmelis this is ready, all issues are resolved! :)

PR #29, opened 2019-03-04
Use semver for openshift resources integration

jmelis:
If major components of the version do not match, then don't manage them.


jmelis:
@maorfr can you please review it? note that I haven't tried it out...

cc @aditya-konarde

utils/openshift_resource.py
```diff
@@ -30,12 +32,28 @@ def has_qontract_annotations(self):
             annotations = self.body['metadata']['annotations']
 
             assert annotations['qontract.integration'] == self.integration
-            assert annotations['qontract.integration_version'] is not None
+
+            integration_version = annotations['qontract.integration_version']
+
+            # handle old version that were not using semver
```
aditya-konarde: Can we instead re-annotate all existing objects and add support only for semver instead of hardcoding this? I am not against this approach, just trying to understand in what use cases we'd need to run this code once the objects are using a new version.

jmelis: well, this saves up from re-annotate all existing resources, but it's indeed an option. I believe this should work, but your proposal also makes sense. wdyt @maorfr ?

maorfr: This is only for resources that were migrated before this PR is merged.

maorfr: I think we should re-annotate all resources now, rather then having some old ones floating around.

setup.py
```diff
@@ -26,6 +26,7 @@
         "ldap3==2.5.2",
         "anymarkup==0.7.0",
         "python-gitlab==1.7.0",
+        "semver-2.8.1",
```
maorfr: semver==2.8.1

jmelis:
LGTM

PR #32, opened 2019-03-10
add route implementation to openshift-resources

maorfr:


reconcile/openshift_resources.py
```diff
@@ -172,6 +177,22 @@ def fetch_provider_vault_secret(path, version, name, labels, annotations):
     return openshift_resource
 
 
+def fetch_provider_route(path, tls_path, tls_version):
+    openshift_resource = fetch_provider_resource(path)
+
+    if tls_path is None or tls_version is None:
+        return openshift_resource
+
+    tls = {}
```
jmelis: ah, so you are replacing whatever is in tls. would it make sense to merge instead?

maorfr: i expect nothing to be present in `tls`, as it is secret data

reconcile/openshift_resources.py
```diff
@@ -172,6 +177,22 @@ def fetch_provider_vault_secret(path, version, name, labels, annotations):
     return openshift_resource
 
 
+def fetch_provider_route(path, tls_path, tls_version):
+    openshift_resource = fetch_provider_resource(path)
+
+    if tls_path is None or tls_version is None:
```
jmelis: can we log this, so this can be tracked? let's say someone specifies tls_path but not tls_version, then the tls section would not be replaced, and nothing would fail. I'd like some logging for this

reconcile/openshift_resources.py
jmelis: ok, this should be documented

jmelis: (I don't mean in the code)

maorfr: it will :)

reconcile/openshift_resources.py
maorfr: as discussed, this will be handled as a schema validation

PR #33, opened 2019-03-11
handle acme-tls routes, override tls fields from vault, improve Route canonicalize

maorfr:


utils/openshift_resource.py
```diff
@@ -113,6 +113,20 @@ def canonicalize(body):
                 body.get('type') == 'Opaque':
             body.pop('type')
 
+        if body['kind'] == 'Route':
+            if 'status' in body:
+                body.pop('status')
+            if body['spec'].get('wildcardPolicy') == 'None':
```
jmelis: This means that if wildcardPolicy is not explicitely set, it will be be set to `None`, right?

jmelis: I mean, that's what k8s will do, no?

utils/openshift_resource.py
```diff
@@ -113,6 +113,20 @@ def canonicalize(body):
...
+                body['spec'].pop('wildcardPolicy')
+            # remove tls-acme specific params from Route
+            if 'kubernetes.io/tls-acme' in annotations:
+                annotations.pop(
+                    'kubernetes.io/tls-acme-awaiting-authorization-owner',
+                    None)
+                tls = body['spec']['tls']
```
jmelis: what if there is no `spec/tls` section? perhaps wrap this in a try except?

jmelis: I can imagine an unfortunate case where someone puts this in the resources folder, without the tls section and with that annotation.

utils/openshift_resource.py
maorfr: right. same as ConfigMap's type `Opaque`.

utils/openshift_resource.py
maorfr: for this kind of route (`kubernetes.io/tls-acme`) there is always a `tls` section:

https://github.com/app-sre/qontract-reconcile/pull/33/files#diff-c09ef4e4196ae9f15e156d3448ed7727R122

maorfr: added a test if `tls` is in `spec`

PR #36, opened 2019-03-14
improve parallelism in openshift-resource

maorfr:
The previous parallelism was running only in the `fetch_desired_state` function, and was running in parallel per namespace.

The current state runs both `fetch_current_state` and `fetch_desired_state` in parallel, for all resources in parallel (from all clusters, from all namespaces).

This improves the execution time to be well under a minute!

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -65,9 +65,13 @@ def openshift_rolebinding(ctx):
 
 
 @integration.command()
+@click.option('--thread-pool-size',
+              help='number of threads to run in parallel',
+              default=10)
 @click.pass_context
-def openshift_resources(ctx):
-    run_integration(reconcile.openshift_resources.run, ctx.obj['dry_run'])
+def openshift_resources(ctx, thread_pool_size):
+    run_integration(reconcile.openshift_resources.run,
+                    ctx.obj['dry_run'], thread_pool_size)
 
 
 @integration.command()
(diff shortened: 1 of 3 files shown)


maorfr:
this last commit improves the execution time to be in the lower 30s!

maorfr:
ready for review

jmelis:
@maorfr: I like this A LOT, it's really impressive... hats off :)

I particularly like the idea of doing both current and desired in the same pool!

Only one comment. Discriminating whether or not it's desired or current based on `oc is None` took me a bit of time to understand. It may make sense to be more explicit. For example, maybe put an variable in StateSpec that is kind, that explicitely says: `current` or `desired` (preferably with enums). wdyt?

Let's discuss this and I will approve right away

maorfr:
thought the same when i was doing it :)
added `type`, as `kind` in this context may be confusing ;)
and also added that `realize_data` is parallel!

jmelis:
It's one thing to do `oc get` in parallel, but `oc apply / delete` in parallel in the same cluster, I'm not so sure... how about we group them by cluster, and execute them in parallel for all the clusters but in sequence for a given cluster?

maorfr:
since every time the integration runs it only applies/deletes a small number of resources, parallelism is not so crucial. my thinking was to improve execution time in cases where we on-board a new service, and want the clusters reconciled quickly.

in addition, since the pool size was set to 10, i thought the clusters can take it.

again, this is not super important at this point, so i reverted the last change and tagged it for possible future use.

jmelis:
Well, I agree with what you said. It's not like this integration is going to continuously apply resources. And even if it fails, we have idempotency for that.

+1 from me if you want to merge with the realize data in parallel commit too

jmelis:
lgtm

PR #37, opened 2019-03-15
add parallelism to ldap-users

maorfr:
this change does not improve execution time by much, but if we need to in the future, we can increase the number of threads.


reconcile/cli.py
```diff
@@ -83,9 +83,13 @@ def quay_repos(ctx):
 
 
 @integration.command()
+@click.option('--thread-pool-size',
```
jmelis: the args should be refactored as they are used twice

jmelis:
LGTM!

PR #38, opened 2019-03-17
enable field type for Secret

maorfr:


jmelis:
LGTM (but don't merge yet)

PR #39, opened 2019-03-19
detect manual changes

maorfr:
Currently, if a manual change is done on an object in the cluster, the sha256sum will not change, and the integration will not detect that change. The current result is that manual changes persist through a run of the integration.

The expected result is that manual changes are detected and reconciled back to their previous state.

This PR makes it so that the sha256sum is calculated even for existing objects which have the sha256sum annotation, which will detect manual changes.

### DIFF
--- a/reconcile/openshift_resources.py
+++ b/reconcile/openshift_resources.py
@@ -418,21 +418,20 @@ def realize_data(dry_run, oc_map, ri):
                     continue
 
                 # don't apply if sha256sum hashes match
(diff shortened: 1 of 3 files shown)


maorfr:
Only one thing here - if a manual change occurs, and we reconcile in app-interface, the old incorrect sha will be kept. Is this something we want to address now? Its an edge case, but still a case.

maorfr:
the last commit handles the last comment. waiting for a second review before merging.

PR #40, opened 2019-03-20
fix metadata.resourceVersion: Invalid value issue

maorfr:
Some resources require that when applying them one must specify the current resourceVersion. if not defined (or defined with the wrong version):
```
The <resourceType> <resourceName> is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
```
I have encountered that when migrating Secrets, and @aditya-konarde is encountering this now when migrating ServiceMonitors.

The resourceVersion field is automatically generated when you apply a resource, and in some cases, if we want to apply the same resource, the resourceVersion field must be equal to the one in the cluster.

This PR handles all the above by doing:
* before canonicalizing the resource, the resourceVersion will be kept, and re-added after the sha calculation. this will allow for backwards compatibility with existing sha256sums.
* in case this is a generated resource (desired item) nothing changes.
* we remove the `--export=true` because that outputs the resourceVersion field as "". imo this is fine as we canonicalize the resources anyway.

### DIFF
--- a/reconcile/openshift_resources.py
+++ b/reconcile/openshift_resources.py
@@ -71,7 +71,7 @@ NAMESPACES_QUERY = """
 """
 
 QONTRACT_INTEGRATION = 'openshift_resources'
-QONTRACT_INTEGRATION_VERSION = semver.format_version(1, 3, 1)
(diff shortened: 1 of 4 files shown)


maorfr:
closing this PR for now as we worked around the issue.

PR #51, opened 2019-04-07
fix rrr in managedTerrraformResources

maorfr:
cc @mmclanerh


maorfr:
Thanks! much appreciated!

PR #55, opened 2019-04-08
[terraform-resources] small fixes

maorfr:


utils/terraform_client.py
```diff
@@ -149,7 +149,11 @@ def construct_oc_resource(self, name, data):
         for k, v in data.items():
             if self.integration_prefix in k:
                 continue
-            body['data'][k] = base64.b64encode(v)
+            if v == "":
+                v = None
```
jmelis: what does the secret need to look like when the key is "" ?

maorfr: `db.name: null`

maorfr: works the same in openshift resources:
https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/openshift_resources.py#L178

PR #56, opened 2019-04-10
[terraform-resources] introduce parallelism

maorfr:


utils/openshift_resource.py
```diff
@@ -209,19 +209,19 @@ def initialize_resource_type(self, cluster, namespace, resource_type):
...
-        if name in desired:
-            self._lock.release()
-            raise ResourceKeyExistsError(name)
-        desired[name] = value
-        self._lock.release()
+        with self._lock:
+            desired = \
+                self._clusters[cluster][namespace][resource_type]['desired']
+            if name in desired:
+                self._lock.release()
```
jmelis: I wonder if this is necessary. I would assume that once the raise is called, the lock will be automatically released as it exits the `with` context. I may be wrong though...

utils/terraform_client.py
```diff
@@ -23,81 +26,144 @@ def __init__(self, body, integration, integration_version):
...
             tfs[name] = tf
         self.tfs = tfs
 
+    class initSpec(object):
+        def __init__(self, name, wd):
+            self.name = name
+            self.wd = wd
+
+    def init_init_specs(self, working_dirs):
+        init_specs = []
```
jmelis: can you use a comprehension list here?

utils/terraform_client.py
```diff
@@ -23,81 +26,144 @@ def __init__(self, body, integration, integration_version):
...
+
+    # terraform plan
     def plan(self, enable_deletion):
         errors = False
         deletions_detected = False
-        for name, tf in self.tfs.items():
-            return_code, stdout, stderr = tf.plan(detailed_exitcode=False)
-            error = self.check_output(name, return_code, stdout, stderr)
+
+        pool = ThreadPool(self.thread_pool_size)
```
jmelis: we could probably define the pool as an instance variable and reuse it everytime: `self.pool`

utils/terraform_client.py
```diff
@@ -23,81 +26,144 @@ def __init__(self, body, integration, integration_version):
...
-            tf = Terraform(working_dir=wd)
-            return_code, stdout, stderr = tf.init()
-            error = self.check_output(name, return_code, stdout, stderr)
-            if error:
-                return None
+        for name, tf in results:
             tfs[name] = tf
         self.tfs = tfs
 
+    class initSpec(object):
```
jmelis: why class as opposed to a simple dict?

utils/terraform_client.py
```diff
@@ -23,81 +26,144 @@ def __init__(self, body, integration, integration_version):
...
         return deletions_detected
 
+    # terraform apply
     def apply(self):
         errors = False
-        for name, tf in self.tfs.items():
-            return_code, stdout, stderr = tf.apply(auto_approve=True)
-            error = self.check_output(name, return_code, stdout, stderr)
+
+        pool = ThreadPool(1)  # TODO: replace 1 with self.thread_pool_size
```
jmelis: TODO

utils/terrascript_client.py
```diff
@@ -56,37 +59,80 @@ def __init__(self, integration, integration_prefix, oc_map):
...
 
     def populate(self, tf_query):
+        populate_specs = self.init_populate_specs(tf_query)
+
+        pool = ThreadPool(self.thread_pool_size)
+        pool.map(self.populate_tf_resources, populate_specs)
+
+        self.validate()
+
+    class populateSpec(object):
```
jmelis: there are several cases of snakeCase in this PR, do we have more of these? I don't like it very much

utils/terraform_client.py
maorfr: to remain TODO until we gain more experience :)

utils/openshift_resource.py
maorfr: you are right, i just missed it. thanks!

utils/terraform_client.py
maorfr: done

utils/terraform_client.py
maorfr: done

utils/terraform_client.py
maorfr: we need a tuple-like iterable list

utils/terrascript_client.py
maorfr: changed them to CamelCase

utils/terraform_client.py
maorfr: removed all relevant classes

utils/terrascript_client.py
```diff
@@ -56,37 +59,72 @@ def __init__(self, integration, integration_prefix, oc_map):
...
 
     def populate(self, tf_query):
+        populate_specs = self.init_populate_specs(tf_query)
+
+        pool = ThreadPool(self.thread_pool_size)
+        pool.map(self.populate_tf_resources, populate_specs)
+
+        self.validate()
+
+    def init_populate_specs(self, tf_query):
```
jmelis: all these functions are better defined as list comprehensions

utils/terrascript_client.py
```diff
@@ -56,37 +59,72 @@ def __init__(self, integration, integration_prefix, oc_map):
...
 
-    def get_vault_tf_secrets(self, type):
-        secrets = {}
-        for name, data in self.accounts:
-            secrets_path = data['secrets_path']
-            secret = vault_client.read_all(secrets_path + '/' + type)
-            secrets[name] = secret
-        return secrets
+    def init_vault_tf_secret_specs(self):
+        vault_specs = []
```
jmelis: list comprehension

utils/terrascript_client.py
jmelis: on second thought these are complex functions. It's clear what they do, maybe a list comprehension is not necessary here.

PR #57, opened 2019-04-10
fix account_name

maorfr:


jmelis:
Why is this needed?

maorfr:
https://github.com/app-sre/qontract-reconcile/blob/da3da8536a9fbc8fa5f52a63bde0224ea8295ffd/utils/terrascript_client.py#L96

maorfr:
```
KeyError: 'account'
```

PR #58, opened 2019-04-11
[terraform-users] introduce new integration

maorfr:


utils/smtp_client.py
```diff
@@ -0,0 +1,82 @@
...
+    _client.sendmail(_username, to, msg.as_string())
+
+
+def send_mails(mails):
+    global _client
+
+    init_from_config()
+    for name, subject, body in mails:
+        send_mail(name, subject, body)
+    teardown()
```
jmelis: maybe try/finalize if we really need to run the teardown

reconcile/terraform_users.py
```diff
@@ -0,0 +1,126 @@
...
+
+TF_QUERY = """
+{
+  roles: roles_v1 {
+    users {
+      redhat_username
+      public_gpg_key
+    }
+    aws_groups {
+      ...on AWSGroup_v1 {
```
jmelis: this is not needed (the `... on ` )

utils/smtp_client.py
maorfr: fixed

reconcile/terraform_users.py
maorfr: fixed

PR #60, opened 2019-04-11
[openshift-rolebinding] add jump host

maorfr:


utils/openshift_api.py
```diff
@@ -10,25 +10,35 @@ class OpenshiftRestApi(object):
...
-                 verify_ssl=True):
+                 verify_ssl=True,
+                 jump_host=None):
 
         self.api_host = host
         self.headers = headers
         self.verify_ssl = verify_ssl
 
+        ssh_server = DummySSHServer()
+        if jump_host is not None:
```
jmelis: I prefer if/else here

utils/jump_host.py
```diff
@@ -24,23 +71,66 @@ def get_identity_from_vault(self, jh):
...
-            'ssh', '-o', 'StrictHostKeyChecking=no',
-            '-o', 'UserKnownHostsFile=/dev/null', '-o', 'LogLevel=ERROR',
-            '-i', identity_file, '-p', self.port, user_host])
+        known_hosts_file = self._identity_dir + '/known_hosts'
+        with open(known_hosts_file, 'w') as f:
+            f.write(self.known_hosts)
+        os.chmod(known_hosts_file, 0o600)
+        self.known_hosts_file = known_hosts_file
 
     def get_ssh_base_cmd(self):
```
jmelis: maybe we can put here:

```
        user_host = '{}@{}'.format(self.user, self.hostname)

         return [
            'ssh',
            '-o', 'StrictHostKeyChecking=yes',
            '-o', 'UserKnownHostsFile={}'.format(self.known_hosts_file),
            '-i', self.identity_file, '-p', str(self.port), user_host]
```
and remove the init and the setter

utils/jump_host.py
```diff
@@ -24,23 +71,66 @@ def get_identity_from_vault(self, jh):
...
+            '-i', self.identity_file, '-p', str(self.port), user_host])
+
+    def init_ssh_session(self):
+        session = jumpssh.SSHSession(
+            self.hostname,
+            self.user,
+            private_key_file=self.identity_file,
+            port=self.port,
+        )
+        self.set_ssh_session(session)
```
jmelis: I wouldn't use a setter here, it make it more complicated. init is implicitely a setter...

utils/jump_host.py
```diff
@@ -24,23 +71,66 @@ def get_identity_from_vault(self, jh):
...
     def cleanup(self):
         shutil.rmtree(self._identity_dir)
+
+
+class DummySSHServer(object):
+    def __init__(self, dummy_resource=None):
+        self.dummy_resource = dummy_resource
+
+    def __enter__(self):
+        return requests
```
jmelis: very clever :D

maorfr: thanks!

utils/jump_host.py
maorfr: fixed

utils/jump_host.py
maorfr: fixed

utils/openshift_api.py
maorfr: fixed

maorfr:
the reason behind JumpHost in the first place is to re-use it between the 2 integrations :)
it is essentially the same functionality, but the representation is different.

plus, openshift-resources already works with this class, so this is actually a lower risk ;)

maorfr:
the only actual changes for `openshift-resources` in this PR is the addition of known_hosts file and the port turned from string to int, hence the version update for the resources integration.

maorfr:
> Do you think it would make sense to split it into two separate classes (maybe inheriting from the same place)? JumpHostSSH and JumpHostSSHRestApi ?

So the answer is yes, and its now done :)

PR #72, opened 2019-05-06
allow inline policies through app-interface

maorfr:


utils/terrascript_client.py
```diff
@@ -486,15 +508,18 @@ def init_common_outputs(self, tf_resources, namespace_info, name):
         output_value = 'Secret'
         tf_resources.append(output(output_name, value=output_value))
 
-    def get_values(self, path):
+    def get_values(self, path, **repalce_args):
```
jmelis: typo

utils/terrascript_client.py
```diff
@@ -486,15 +508,18 @@ def init_common_outputs(self, tf_resources, namespace_info, name):
...
         gqlapi = gql.get_api()
         try:
             raw_values = gqlapi.get_resource(path)
+            content = raw_values['content']
         except gql.GqlApiError as e:
             raise FetchResourceError(e.message)
+        for old, new in repalce_args.items():
+            content = content.replace(old, new)
```
jmelis: is this supposed to be at the same leve of identation?

utils/terrascript_client.py
maorfr: thanks!

jmelis:
LGTM

PR #74, opened 2019-05-07
[openshift-namespaces] introduce new integration

maorfr:
this integration:

1. gets the desired state
2. finds namespaces which do not exists
3. creates those namespaces

this integration does not:
1. delete namespaces
2. annotate namespaces

### DIFF
--- a/reconcile/openshift_namespaces.py
+++ b/reconcile/openshift_namespaces.py
@@ -42,8 +42,12 @@ def get_desired_state():
     query = gqlapi.query(QUERY)['namespaces']
     ri = ResourceInventory()
     oc_map = {}
-    openshift_resources.init_specs_to_fetch(ri, oc_map, query,
-        override_managed_types=['Namespace'])
+    openshift_resources.init_specs_to_fetch(
+        ri,
+        oc_map,
+        query,
+        override_managed_types=['Namespace']
+    )
     desired_state = [{"cluster": cluster, "namespace": namespace}
                      for cluster, namespace, _, _ in ri]
 
--- a/reconcile/openshift_namespaces.py
+++ b/reconcile/openshift_namespaces.py
@@ -1,14 +1,12 @@
 import logging
-
-import reconcile.openshift_resources as openshift_resources
+from multiprocessing.dummy import Pool as ThreadPool
+from functools import partial
 
 import utils.gql as gql
-
+import reconcile.openshift_resources as openshift_resources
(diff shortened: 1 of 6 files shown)


riuvshin:
nice! I wanted to work on it once back :) GJ!
+1 to 
```
this integration does not:

delete namespaces
annotate namespaces```

**Comment by maorfr** [comment]
Anything you see that needs work @riuvshin?

**Comment by riuvshin** [comment]
@maorfr it looks good so far but I'm more interested to see how it will be structured in app-interface :)

File: reconcile/openshift_namespaces.py [hunk 1]
```diff
@@ -0,0 +1,90 @@
...
+
+import reconcile.openshift_resources as openshift_resources
+
+import utils.gql as gql
+
+from utils.openshift_resource import ResourceInventory
+from utils.oc import StatusCodeError
+
+from multiprocessing.dummy import Pool as ThreadPool
+from functools import partial
```
jmelis: I just realized we are not following the correct order in other files...

```
import logging
from multiprocessing.dummy import Pool as ThreadPool
from functools import partial

import reconcile.openshift_resources as openshift_resources
import utils.gql as gql
from utils.openshift_resource import ResourceInventory
from utils.oc import StatusCodeError
```

first the native ones, then python packages, and then our libraries. Also I prefer first import and then from.

reconcile/openshift_namespaces.py
```diff
@@ -0,0 +1,90 @@
...
+}
+"""
+
+
+def get_desired_state():
+    gqlapi = gql.get_api()
+    query = gqlapi.query(QUERY)['namespaces']
+    for namespace_info in query:
+        # adjust to match openshift_resources functions
+        namespace_info['managedResourceTypes'] = ['Namespace']
```
jmelis: how about instead of doing this, which is a bit ugly, we make it a kwarg for init_specs_to_fetch (defaulting to `["managedResourcesTypes"]`)

jmelis: also by the way, the proper way to create the default value is inside the method, not in the method's signature

reconcile/openshift_namespaces.py
maorfr: fixed

reconcile/openshift_namespaces.py
maorfr: fixed

PR #79, opened 2019-05-19
[openshift-groups] introduce new integration

maorfr:
this integration does:
* add users to groups
* delete users from groups

this integration does not:
* manage all groups (only ones specified in a cluster file under `managedGroups`)
* creates groups
* deletes groups

### DIFF
--- a/reconcile/openshift_groups.py
+++ b/reconcile/openshift_groups.py
@@ -1,14 +1,7 @@
 import logging
-import sys
-import copy
 
 import utils.gql as gql
-import utils.vault_client as vault_client
-
-from utils.openshift_api import Openshift
-from utils.aggregated_list import (AggregatedList,
-                                   AggregatedDiffRunner,
-                                   RunnerException)
+import reconcile.openshift_resources as openshift_resources
 
 CLUSTERS_QUERY = """
 {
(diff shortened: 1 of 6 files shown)


reconcile/openshift_groups.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def fetch_desired_state():
+    gqlapi = gql.get_api()
+    roles = gqlapi.query(ROLES_QUERY)['roles']
+    desired_state = []
+
+    for r in roles:
+        for p in r['permissions']:
+            if 'service' not in p:
```
jmelis: a bit redundant?

jmelis: Since we already have the one below...

reconcile/openshift_groups.py
```diff
@@ -0,0 +1,160 @@
...
+def subtract_states(from_state, subtract_state, action):
+    result = []
+
+    for f_user in from_state:
+        found = False
+        for s_user in subtract_state:
+            if f_user != s_user:
+                continue
+            found = True
+            break
```
jmelis: maybe this is clearer?

```
        for s_user in subtract_state:
            if f_user == s_user:
                found = True
                break
```

jmelis:
Excellent stuff @maorfr !!

reconcile/openshift_groups.py
maorfr: right!

reconcile/openshift_groups.py
maorfr: Flat is better than nested? ;)

maorfr:
for future reference -
https://wiki.jenkins.io/display/JENKINS/Role+Strategy+Plugin

PR #80, opened 2019-05-20
[jenkins-roles] introduce new integration

maorfr:
this integration does:
* assigns roles to users
* unassignes roles from users

this integration does not:
* creates users
* deletes users
* creates roles
* deletes roles


reconcile/jenkins_roles.py
```diff
@@ -0,0 +1,177 @@
...
+    }
+    bots {
+      redhat_username
+    }
+    permissions {
+      service
+      ...on PermissionJenkinsRole_v1 {
+        instance
+        role
+        token {
```
jmelis: you are not using token, right?

maorfr: true

PR #82, opened 2019-05-27
[jenkins-plugins] introduce new integration

maorfr:
this integration does:
* installs the latest version of the listed plugins

this integration does not:
* uninstalls plugins
* updates plugins

### DIFF
--- a/utils/jenkins_api.py
+++ b/utils/jenkins_api.py
@@ -17,6 +17,7 @@ class JenkinsApi(object):
         self.user = config['jenkins']['user']
         self.password = config['jenkins']['password']
         self.ssl_verify = ssl_verify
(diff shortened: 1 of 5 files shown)


utils/jenkins_api.py
```diff
@@ -60,3 +61,44 @@ def unassign_role_from_user(self, role, user):
...
+            url,
+            verify=self.ssl_verify,
+            data=data,
+            headers=header,
+            auth=(self.user, self.password)
+        )
+
+        res.raise_for_status()
+
+    def safe_restart(self, force_restart=False):
```
jmelis: can we add a logging.debug in this method?

maorfr: done

PR #86, opened 2019-05-30
WIP: [openshift-resources] enable jobs

maorfr:


jmelis:
Marking as WIP because we are not sure we want to follow this path.

PR #87, opened 2019-05-30
[aws-iam-keys] introduce new integration

maorfr:
this integration does:
* deletes IAM keys from AWS accounts

this integration does not:
* create IAM keys
* activate/deactivate IAM keys

depends on #84.


jmelis:
LGTM

PR #92, opened 2019-06-03
[terraform-users] validate gpg key

maorfr:
public_gpg_key should be a base64 encoded binary gpg key. this PR adds a test that this is a valid gpg key using the `gpg` binary.


utils/gpg.py
```diff
@@ -0,0 +1,18 @@
+import os
+import base64
+
+from subprocess import PIPE, Popen, STDOUT
+
+
+def gpg_key_valid(public_gpg_key):
+    DENNULL = open(os.devnull, 'w')
```
jfchevrette: not a big deal, but `DENNULL` vs `DEVNULL`

utils/gpg.py
```diff
@@ -0,0 +1,18 @@
...
+    try:
+        public_gpg_key_dec = base64.b64decode(public_gpg_key)
+    except Exception:
+        return False
+    proc = Popen(['gpg'], stdin=PIPE, stdout=DENNULL, stderr=STDOUT)
```
jfchevrette: should we blindly assume gpg is installed?

maorfr: I checked it manually. Do you think we should check it programmatically?

maorfr: I checked in the docker image that is

utils/gpg.py
maorfr: Thanks!

utils/gpg.py
jfchevrette: I think it's fine for our usage but if it is ever run outside of that docker image or anywhere where gpg is missing it will report a bunch of "user xyz has an invalid public gpg key" which would be confusing

maorfr: We have had this problem in the past and is good to verify.
I created an issue to do this for all binary usages in all integrations.

maorfr:
Thanks @jfchevrette!

PR #99, opened 2019-06-11
[slack-usergroups] introduce new integration

maorfr:
this integration does:

* reconciles channels for user groups
* reconciles users for user groups

this integration does not:

* create user groups/channels/users
* add/remove users from channels

by using the a user group, members will be added automatically to newly added channels, so the order will be first to update channels and only then to update users.

### DIFF
--- a/reconcile/slack_usergroups.py
+++ b/reconcile/slack_usergroups.py
@@ -63,7 +63,7 @@ def get_slack_map():
 
         workspace_spec = {
             "slack": SlackApi(workspace['token']),
-            "manged_usergroups": workspace['managedUsergroups']
+            "managed_usergroups": workspace['managedUsergroups']
         }
         slack_map[workspace_name] = workspace_spec
 
@@ -75,7 +75,7 @@ def get_current_state(slack_map):
 
     for workspace, spec in slack_map.items():
         slack = spec['slack']
-        managed_usergroups = spec['manged_usergroups']
+        managed_usergroups = spec['managed_usergroups']
         for ug in managed_usergroups:
             users, channels = slack.describe_usergroup(ug)
             current_state.append({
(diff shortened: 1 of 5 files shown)


reconcile/slack_usergroups.py
```diff
@@ -0,0 +1,193 @@
...
+
+    return slack_map
+
+
+def get_current_state(slack_map):
+    current_state = []
+
+    for workspace, spec in slack_map.items():
+        slack = spec['slack']
+        managed_usergroups = spec['manged_usergroups']
```
jmelis: typo

reconcile/slack_usergroups.py
```diff
@@ -0,0 +1,193 @@
...
+    f = from_state[type]
+    s = subtract_state[type]
+    return [v for k, v in f.items() if k not in s.keys()]
+
+
+def act(desired_state, slack_map):
+    for state in desired_state:
+        workspace = state['workspace']
+        ugid = state['usergroup_id']
+        users = ','.join(state['users'].keys())
```
jmelis: can you move this to inside the update_usergroup_users method? I think it's cleaner... passing an array of users sounds much better than a preformatted string.

utils/slack_api.py
```diff
@@ -0,0 +1,108 @@
...
+        return {k: v for k, v in self.get('users').items()
+                if k in users_ids}
+
+    def get(self, type):
+        result_key = 'members' if type == 'users' else type
+        results = {}
+        cursor = ''
+
+        if type in self.results:
+            return self.results[type]
```
jmelis: :heart:

maorfr: :heart_decoration:

reconcile/slack_usergroups.py
maorfr: thanks!

reconcile/slack_usergroups.py
maorfr: makes sense

PR #101, opened 2019-06-11
WIP: git_metrics module

jmelis:


jmelis:
not putting this in qontract-reconcile on second thought

Issue #112, opened 2019-06-18
Have a single way to interact with OpenShift API


aditya-konarde:
Currently, the `openshift-resources` subcommand talks to openshift API with a wrapper to the binary, while the `openshift-rolebindings` integration uses `requests` to talk directly to the OpenShift api. 

This is causing confusion debugging, where we are seeing one of the integration fail on an SSL cert verification check, while the others interacting with OpenShift work just fine.

Proposal: We should have a uniform way of talking to the OpenShift API across integrations.

CC @jmelis @maorfr

jmelis:
Totally! @maorfr is planning on rewriting openshift-rolebindings to use `oc` instead of the API. :100: !

maorfr:
created https://jira.coreos.com/browse/APPSRE-897

PR #113, opened 2019-06-18
Add trusted certs bundle to OpenShift client constructor call

aditya-konarde:
Mitigate SSL trust errors mentioned in : https://github.com/app-sre/qontract-reconcile/issues/112

This is required as the Python requests library with the certifi module has its own ca certs bundle which it trusts inside the virtualenv. This was resulting in errors where we have added custom certs to the host, but they don't show up in the integration run.

### DIFF
--- a/reconcile/openshift_rolebinding.py
+++ b/reconcile/openshift_rolebinding.py
@@ -87,7 +87,7 @@ class ClusterStore(object):
                 )
 
                 api = Openshift(cluster_info['serverUrl'], token,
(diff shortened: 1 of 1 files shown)


aditya-konarde:
[test]

aditya-konarde:
Closing this as we've fixed this elsewhere, specifically in the build_deploy script where the integrations are called, by adding a 
`-e REQUESTS_CA_BUNDLE=/etc/pki/tls/cert.pem \`

PR #114, opened 2019-06-19
speed up AWS related integrations

maorfr:
the point of this PR is to speed things up in the AWS related integrations.

to do this, things changed on several aspects:

* terraform-resources and terraform-users:
  * increase `terraform plan` parallelism to 20 (default is 10)
  * decrease number of calls to `terraform output`

* aws-garbage-collector:
  * each `map` function now runs in parallel for all resources (the `map` functions still run in order)

in addition, all 3 integrations use a `self.specs` property, which gets updated with data throughout the run of the integration (instead of creating separate properties). this helps us especially in the `terraform-*` integrations since it saves us the need to initialize these specs for each terraform action.

it is recommended to review this PR in 3 cycles:
1. review the cli
2. review terraform-* integrations (terraform_client.py)
3. review aws-gc integration (aws_garbage_collector.py)

In addition, i have added a change that attempts to handle the problem we encountered here:
https://ci-int-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/service-app-interface-gl-build-master/515/artifact/reports/reconcile_reports_success/reconcile-terraform-resources.txt

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -29,7 +29,7 @@ from utils.aggregated_list import RunnerException
 def threaded(function):
     function = click.option('--thread-pool-size',
                             help='number of threads to run in parallel',
-                            default=10)(function)
+                            default=20)(function)
 
     return function
 
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -26,12 +26,15 @@ import reconcile.aws_iam_keys
 from utils.aggregated_list import RunnerException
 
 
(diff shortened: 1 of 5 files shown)


maorfr:
closing in favor of:
https://github.com/app-sre/qontract-reconcile/pull/116
https://github.com/app-sre/qontract-reconcile/pull/117

this one was too much for a single review, will apply it incrementally.

PR #120, opened 2019-06-23
[slack-usergroups] enable multiple pager duty instances per usergroup

maorfr:


maorfr:
thanks @mmclanerh! i totally agree.

PR #122, opened 2019-06-25
[terraform-*] terraform validate without init

maorfr:
the `ts.validate()` function in terrascript was calling `terraform init` for no reason

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -1,4 +1,5 @@
 import sys
+import shutil
 import semver
 
 import utils.gql as gql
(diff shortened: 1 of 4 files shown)


reconcile/terraform_resources.py
```diff
@@ -107,21 +108,28 @@ def setup(print_only, thread_pool_size):
...
+    if error:
+        cleanup_and_exit(status=error, working_dirs=working_dirs)
 
     return ri, oc_map, tf
 
 
-def cleanup_and_exit(tf=None, status=False):
-    if tf is not None:
+def cleanup_and_exit(tf=None, status=False, working_dirs={}):
+    if tf is None:
```
aditya-konarde: Better if we log the `status` here if its not set to `False`?

aditya-konarde: I missed that its a bool, might be able to change it so that we can log the error message. But non-blocking on this PR

PR #124, opened 2019-06-27
[gitlab-housekeeping] introduce new integration

maorfr:
this integration does:

iterates over housekeeping and adds `stale` label to stale issues (no activity over 30d) + adds a comment.
after 'stale' is added and 30d more passed - the integration closes the issue.

if after adding 'stale' someone added a note of '/stale cancel' - the integration will remove the label.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -17,10 +17,9 @@ def get_housekeeping_gitlab_api():
     return GitLabApi(server, token, project_id=project_id, ssl_verify=False)
 
 
-def handle_stale_issues(dry_run, gl):
+def handle_stale_issues(dry_run, gl, days_interval):
     DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
     LABEL = 'stale'
-    DAYS = 30
 
     issues = gl.get_issues(state='opened')
     now = datetime.utcnow()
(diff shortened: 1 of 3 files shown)


reconcile/gitlab_housekeeping.py
```diff
@@ -0,0 +1,73 @@
...
+
+def get_housekeeping_gitlab_api():
+    config = get_config()
+
+    gitlab_config = config['gitlab']
+    server = gitlab_config['server']
+    token = gitlab_config['token']
+    project_id = gitlab_config['housekeeping']['project_id']
+
+    return GitLabApi(server, token, project_id=project_id, ssl_verify=False)
```
aditya-konarde: We can perhaps have `ssl_verify` set to true now that we have internal certs trusted :) 

But okay to leave it at

reconcile/gitlab_housekeeping.py
```diff
@@ -0,0 +1,73 @@
...
+
+    issues = gl.get_issues(state='opened')
+    now = datetime.utcnow()
+    for issue in issues:
+        issue_iid = issue.attributes.get('iid')
+        issue_labels = issue.attributes.get('labels')
+        updated_at = issue.attributes.get('updated_at')
+        update_date = datetime.strptime(updated_at, DATE_FORMAT)
+
+        # if issue is over 30d old
```
aditya-konarde: I would suggest we mark as stale at 15 days of no activity, and then close after 15 days of having a stale label

jfchevrette: my vote goes to 15 days as well :+1:

reconcile/gitlab_housekeeping.py
maorfr: lets keep this PR self contained. wdyt?

reconcile/gitlab_housekeeping.py
maorfr: updated to 15!

aditya-konarde:
We should also bump the Readme to add info around the integrations :)

maorfr:
oh, missed your comment.
i'll add it in the next PR

PR #130, opened 2019-07-09
[jenkins-job-builder] introduce new integration

maorfr:
this integration is running jenkins-jobs to apply job configurations to jenkins.

when this integration is working in dry-run mode, it is executing `jenkins-jobs test`.
when it is not in dry run mode, it is executing `jenkins-jobs update`.

to be able to show diffs between the current state and the desired state, this integration needs to be executed twice:

1. execute with a config file pointing to prod (current state) and add the `--no-compare` flag, to indicate that this should only collect data
2. execute with a config file pointing to a local server (in a PR) (desired state). this run will fetch the desired state and compare it to the current state.

1 + 2 are not required when running this integration in a non dry-run mode, but only a single execution.

the ability to run integrations twice, first querying production and then querying the local server is already enabled[[1](https://gitlab.cee.redhat.com/service/app-interface/merge_requests/965/diffs#a5c3b3b8337ffd291fedf7a9fa32e2edead57400_149_136)]

### DIFF
--- a/utils/jjb_client.py
+++ b/utils/jjb_client.py
@@ -1,0 +1,171 @@
+import os
+import shutil
+import yaml
+import tempfile
+import logging
+import filecmp
+import xml.etree.ElementTree as et
+import utils.vault_client as vault_client
+
+import utils.gql as gql
+
+from os import path
+
+from jenkins_jobs.cli.entry import JenkinsJobs
+
+
+class FetchResourceError(Exception):
+    def __init__(self, msg):
+        super(FetchResourceError, self).__init__(
+            "error fetching resource: " + str(msg)
+        )
+
+
+class JJB(object):
+    """Wrapper around Jenkins Jobs"""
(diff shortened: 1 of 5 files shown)


utils/jjb_client.py
```diff
@@ -0,0 +1,184 @@
...
+            jjb.execute()
+
+    def modify_logger(self):
+        formatter = logging.Formatter('%(levelname)s: %(message)s')
+        logger = logging.getLogger()
+        logger.handlers[0].setFormatter(formatter)
+        self.default_logging = logger.level
+
+    @contextmanager
+    def toggle_logger(self):
```
jmelis: :heart:

jmelis: very clever way to solve the JJB logging problem!

maorfr: :heartbeat:

PR #131, opened 2019-07-09
[gitlab-members] new integration

SydneyDlh:
new integration that checks app-interface for which gitlab groups we track and what users have permission for that group and matches that with gitlab.

still needs a few fixes so that it works for when a user or group is deleted

### DIFF
--- a/reconcile/gitlab_members.py
+++ b/reconcile/gitlab_members.py
@@ -1,0 +1,106 @@
+import sys
+import logging
+
+import utils.gql as gql
+from utils.config import get_config
+from utils.gitlab_api import GitLabApi
+
+GROUPS_QUERY = """
+{
+  gitlabgroups_v1{
+    managedGroups
+  }
+}
+"""
+
+USERS_QUERY = """
+{
+  users_v1{
+    redhat_username
+    roles{
+      permissions{
+      	... on PermissionGitlabGroups_v1{
+          name
+          group
+        }
+      }
+    }
+  }
+}
+  """
+
+def get_groups():
+  groups = gqlapi.query(GROUPS_QUERY)
+  group_dict = {}
+  for g in groups:
+    group_dict[g] = []
+  return group_dict
+
+def test():
+    groups = gqlapi.query(GROUPS_QUERY)
+    users = gqlapi.query(USERS_QUERY)
+    print(groups)
+    print(users)
+
+def get_desired_state():
+    groups = gqlapi.query(GROUPS_QUERY)
+    users = gqlapi.query(USERS_QUERY)
+    desired_group_members = get_groups()
+    for g in groups:
+      for u in users:
+        for r in u['roles']:
+          for p in r['permissions']:
+            if p['group'] == g:
+              desired_group_members[g].append(u['redhat_username']) #redhat_username == gitlab username ??? 
+  return desired_group_members
+
+
+def get_current_state():
+    groups = gqlapi.query(GROUPS_QUERY)
+    current_group_members = get_groups()
+    for g in groups:
+      current_group_members[g]=[get_gitlab_group_members(g)]
+    return current_group_members
+
+def calculate_diff(current_state, desired_state):
+    diff = []
+    users_to_add = \
+        subtract_states(desired_state, current_state,
+                        "add_user_to_group")
+    diff.extend(users_to_add)
+    users_to_remove = \
+        subtract_states(current_state, desired_state,
+                        "remove_user_from_group")
+    diff.extend(users_to_remove)
+
+    return diff
+
+
+def subtract_states(from_state, subtract_state, action):
+    result = []
+    for f_group in from_state:
+      s_group = subtract_state[f_group] #assumming groups are the same in both states which is a bad assumption
+      for f_user in f_group:
+        found = False
+        for s_user in s_group:
+          if f_user != s_user:
+            continue
+          found = True
+          break
+        if not found:
+          result.append({
+            "action": action,
+            "group": f_group,
+            "user": f_user,
+          })
+    return result
+
+
+
+def run(dry_run=False):
+  desired_state = get_desired_state()
+  current_state = get_current_state()
+  diffs = calculate_diff(current_state,desired_state)
+
+  for diff in diffs:
+    logging.info(diff.values())
--- a/reconcile/gitlab_members.py
+++ b/reconcile/gitlab_members.py
@@ -7,59 +7,67 @@ from utils.gitlab_api import GitLabApi
 
 GROUPS_QUERY = """
 {
-  gitlabgroups_v1{
-    managedGroups
-  }
+    groups: gitlabgroups_v1{
+        managedGroups
+    }
 }
 """
 
 USERS_QUERY = """
 {
-  users_v1{
-    redhat_username
-    roles{
-      permissions{
-      	... on PermissionGitlabGroups_v1{
-          name
-          group
+    users: users_v1{
+        redhat_username
+        roles{
+            permissions{
+                ... on PermissionGitlabGroups_v1{
+                    name
+                    group
+                }
+            }
         }
-      }
     }
-  }
 }
   """
 
-def get_groups():
-  groups = gqlapi.query(GROUPS_QUERY)
-  group_dict = {}
-  for g in groups:
-    group_dict[g] = []
-  return group_dict
+def get_gitlab_api():
+    config = get_config()
+    gitlab_config = config['gitlab']
+    server = gitlab_config['server']
+    token = gitlab_config['token']
+    return GitLabApi(server, token, ssl_verify=False)
 
-def test():
-    groups = gqlapi.query(GROUPS_QUERY)
-    users = gqlapi.query(USERS_QUERY)
-    print(groups)
-    print(users)
+
+def get_groups():
+    gqlapi = gql.get_api()
+    groups = gqlapi.query(GROUPS_QUERY)['groups']
+    group_dict = {}
+    for g in groups[0]['managedGroups']:
+        group_dict[g] = []
+    return group_dict
 
 def get_desired_state():
-    groups = gqlapi.query(GROUPS_QUERY)
-    users = gqlapi.query(USERS_QUERY)
+    gqlapi = gql.get_api()
+    gl = get_gitlab_api()
+    groups = gqlapi.query(GROUPS_QUERY)['groups']
+    users = gqlapi.query(USERS_QUERY)['users']
     desired_group_members = get_groups()
-    for g in groups:
-      for u in users:
-        for r in u['roles']:
-          for p in r['permissions']:
-            if p['group'] == g:
-              desired_group_members[g].append(u['redhat_username']) #redhat_username == gitlab username ??? 
-  return desired_group_members
+    for g in groups[0]['managedGroups']:
+        for u in users:
+            for r in u['roles']:
+                for p in r['permissions']:
+                    if 'group' in p and p['group'] == g:
+                        desired_group_members[g].append(gl.get_user(u['redhat_username'])) #does redhat_username == gitlab username? this should also account for if user was deleted
+        desired_group_members[g].append(gl.get_user('devtools-bot'))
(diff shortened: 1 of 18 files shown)


reconcile/gitlab_members.py
```diff
@@ -0,0 +1,126 @@
...
+    server = gitlab_config['server']
+    token = gitlab_config['token']
+    return GitLabApi(server, token, ssl_verify=False)
+
+
+def get_groups():
+    gqlapi = gql.get_api()
+    groups = gqlapi.query(GROUPS_QUERY)['groups']
+    group_dict = {}
+    for g in groups[0]['managedGroups']:
```
maorfr: is this assuming we have only one gitlab instance?
i know this will probably never change, but i'd still prefer we support `n` gitlab instances.
wdyt?

utils/gitlab_api.py
```diff
@@ -103,6 +103,10 @@ def get_app_sre_group_users(self):
         app_sre_group = self.gl.groups.get('app-sre')
         return [m for m in app_sre_group.members.list()]
 
+    def get_gitlab_group_members(self, group_name):
```
maorfr: i think we can rename this to `get_group_members`

reconcile/gitlab_members.py
```diff
@@ -0,0 +1,126 @@
...
+    gqlapi = gql.get_api()
+    gl = get_gitlab_api()
+    groups = gqlapi.query(GROUPS_QUERY)['groups']
+    users = gqlapi.query(USERS_QUERY)['users']
+    desired_group_members = get_groups()
+    for g in groups[0]['managedGroups']:
+        for u in users:
+            for r in u['roles']:
+                for p in r['permissions']:
+                    if 'group' in p and p['group'] == g:
```
maorfr: it is strange to me that `service == 'gitlab-group'` is not here...

reconcile/gitlab_members.py
```diff
@@ -0,0 +1,126 @@
...
+
+def get_gitlab_api():
+    config = get_config()
+    gitlab_config = config['gitlab']
+    server = gitlab_config['server']
+    token = gitlab_config['token']
+    return GitLabApi(server, token, ssl_verify=False)
+
+
+def get_groups():
```
maorfr: this function seem to do a bit more then just get the groups. the name of the function should reflect what it does as accurately as possible imo.

reconcile/gitlab_members.py
```diff
@@ -0,0 +1,126 @@
...
+                        desired_group_members[g].append(gl.get_user(u['redhat_username'])) #does redhat_username == gitlab username? this should also account for if user was deleted
+        desired_group_members[g].append(gl.get_user('devtools-bot'))
+    return desired_group_members
+
+
+def get_current_state():
+    gqlapi = gql.get_api()
+    gl = get_gitlab_api()
+    groups = gqlapi.query(GROUPS_QUERY)['groups']
+    current_group_members = get_groups()
```
maorfr: this is an example how `get_groups` is not indicative of what it does (does it get the groups or the groups members?

reconcile/gitlab_members.py
```diff
@@ -0,0 +1,126 @@
...
+        subtract_states(current_state, desired_state,
+                        "remove_user_from_group")
+    diff.extend(users_to_remove)
+
+    return diff
+
+
+def subtract_states(from_state, subtract_state, action):
+    result = []
+    for f_group in from_state:
```
maorfr: you can unpack a dictionary and its values in a single `for` loop:
```
for key, value in dict.items()
```

reconcile/gitlab_members.py
```diff
@@ -0,0 +1,126 @@
...
+                "group": f_group,
+                "user": f_user,
+                })
+    return result
+
+def act(diff):
+    group = diff['group']
+    username = diff['user']
+    action = diff['action']
+    gl = get_gitlab_api()
```
maorfr: we seem to be calling `get_gitlab_api` a lot. i'd expect to initiate the api only once and pass it around. another option is that `get_gitlab_api` will always return the same api instance if its already initiated, but i'd suggest to go with the first option.

reconcile/gitlab_members.py
```diff
@@ -0,0 +1,126 @@
...
+    if action == "remove_user_from_group":
+        gl.remove_group_member(group,username)
+    if action == "add_user_to_group":
+        gl.add_group_member(group,username)
+
+
+def run(dry_run=False):
+    desired_state = get_desired_state()
+    current_state = get_current_state()
```
maorfr: just a minor minor minor minor thing - can we get the current state before we get the desired state? just to conform to other integrations...

reconcile/gitlab_members.py
```diff
@@ -0,0 +1,126 @@
...
+
+GROUPS_QUERY = """
+{
+    groups: gitlabgroups_v1{
+        managedGroups
+    }
+}
+"""
+
+USERS_QUERY = """
```
maorfr: i really like it that you attack this from the users direction and not from the roles direction like other integrations. just be sure to filter out irrelevant users as early as possible so we save some run time.

reconcile/gitlab_members.py
```diff
@@ -0,0 +1,126 @@
+import sys
+import logging
+
+import utils.gql as gql
+from utils.config import get_config
+from utils.gitlab_api import GitLabApi
+
+GROUPS_QUERY = """
+{
+    groups: gitlabgroups_v1{
```
maorfr: putting this here, but i think we already talked about it -
i think that `gitlabgroups` is not exactly the right name for this...
maybe its actually the representation of a gitlab instance, and we can add that to the schema as well?

we can probably even re-use such an entity in other integrations.
if this object will include the token, we may even be able to remove some configuration from the toml file in a future PR.

if this was too abstract, here is an example:
https://gitlab.cee.redhat.com/service/app-interface/blob/master/data/dependencies/ci-int/ci-int.yml
(ignore the `plugins` stuff)
the `path` in this example is a path to a secret in vault which holds the token.

reconcile/gitlab_members.py
SydneyDlh: current_group_members is set to get_groups() which is a dict with all the groups as keys and empty lists as values, later the members are added to those empty lists.

reconcile/gitlab_members.py
```diff
@@ -0,0 +1,144 @@
...
+                    name
+                    group
+                }
+            }
+        }
+    }
+}
+  """
+
+BOTS_QUERY = """
```
maorfr: now that we have both users and bots, would it make more sense to have a single query?
example: https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/jenkins_roles.py#L26

maorfr: and also if you can make the queries look more like that example visually?

reconcile/gitlab_members.py
maorfr: so, in the query you use, those permissions are indeed the only ones returning (with data). the rest of the permissions also return on the query, but they return empty. so we need to filter the empty ones at least.

utils/gitlab_api.py
```diff
@@ -102,6 +102,16 @@ def get_project_maintainers(self, repo_url):
     def get_app_sre_group_users(self):
         app_sre_group = self.gl.groups.get('app-sre')
         return [m for m in app_sre_group.members.list()]
+    
+
+    def get_group_members(self, group_name):
+        groups = self.gl.groups.list()
+        group_names = list(map(lambda x: x.name,groups))
+        if group_name not in group_names:
+            logging.info(group_name + " group not found")
```
maorfr: if the group is not found we just log and continue?

utils/gitlab_api.py
```diff
@@ -114,6 +124,21 @@ def add_project_member(self, repo_url, user):
             member = project.members.get(user.id)
             member.access_level = gitlab.MAINTAINER_ACCESS
 
+    def add_group_member(self, group_name, user):
+        group = self.gl.groups.get(group_name)
+        try:
+            group.members.create({
+                'user_id': user.id,
+                'access_level': gitlab.MAINTAINER_ACCESS
```
maorfr: maybe this should be a part of the schema? the group membership would also define the access level. wdyt?

utils/gitlab_api.py
```diff
@@ -155,6 +180,15 @@ def close_issue(self, issue):
         issue.state_event = 'close'
         issue.save()
 
+        
+    def get_user(self, username):
+        user = self.gl.users.list(search=username)
+        if len(user) == 0:
```
maorfr: same here. if a user is not found we should be less forgiving imo

maorfr:
in addition to the comments, the code should pass linting. you can run `make test` locally to see the errors.

utils/gitlab_api.py
SydneyDlh: I think that we should create the group when its not found, but I thought for this version there wasn't going to be anything involving creating and deleting groups or users, I just added this so that there wouldn't be an error if a group was not found.

utils/gitlab_api.py
SydneyDlh: hmm. Which schema would it go in? Would each user have access level in their schemas that applies for all groups? Or would the access level be in the gitlab instance schema, so each managed group has an access level? Or I guess the best way to do it would be creating permissions for different access levels for each group. 

Also this could be something the integration could reconcile, for example in app-interface if a user has a permission with owner access  but on gitlab they are just a maintainer the integration could change their access to owner. Not sure how important access levels are to the way you guys manage groups if this is worthwhile or not.

utils/gitlab_api.py
SydneyDlh: I am not sure the solution for this one. If a user is in app-interface but does not have a gitlab account I don't really think this is the integration that should be making those users. I don't think that is in the scope of the integration because it is more about groups.

Do you have a suggestion for a less forgiving way to deal with this if it happens? Possibly sending an email to the redhat user telling them to make a gitlab account similar to the github_users integration. Or if you think the integration should be making accounts it should probably notify the user that a account was created for them and what their username and pass is.

Could be a new integration that checks what users should have gitlab accounts and creates/emails them if it doesn't fit with this one.

reconcile/gitlab_members.py
SydneyDlh: Does doing a single query change anything other than style? Or do you want me to filter by roles instead of by users and bots separately like in the example.

maorfr: 1 query is usually faster then 2, depending on the payload of course. in this case, it probably doesn't make much difference. so i would say that this is more about style, and maybe a bit about avoiding code duplication. what is more likely to change in the future wrt queries is the query for these specific roles. we are less likely to need anything other then redhat_username. these were my thoughts. i'm leaving this for you to decide, either way is fine :)

utils/gitlab_api.py
maorfr: that is great thinking. these are the kinds of things that can be added as comments to explain your thought path. i'm with you on not failing, but i would suggest to use `logging.error` instead just so it stands out (it will not fail the integration).

utils/gitlab_api.py
maorfr: its probably not super important, but rather easy to add, and would be really good to have for future use. i'd say it should go in the `gitlab-group-membership` permission schema object. wdyt?

utils/gitlab_api.py
maorfr: i just meant that we should do more then just log. maybe exit with an error in this case, or to keep consistency with previous discussion, log an error and continue?

reconcile/gitlab_members.py
```diff
@@ -0,0 +1,177 @@
...
+def act(diff, gl):
+    group = diff['group']
+    user = diff['user']
+    action = diff['action']
+    access = diff['access']
+    if action == "remove_user_from_group":
+        gl.remove_group_member(group, user)
+    if action == "add_user_to_group":
+        gl.add_group_member(group, user, access)
+    if action == "change_access":
```
maorfr: :heart:

PR #141, opened 2019-07-18
[slack-usergroups] add github as source

maorfr:
like we can have slack usergroup members be based on pagerduty (schedules and escalation policies) - this PR adds the ability to base the members on github OWNERS files!

### DIFF
--- a/reconcile/slack_usergroups.py
+++ b/reconcile/slack_usergroups.py
@@ -1,4 +1,6 @@
 import logging
+import anymarkup
+import requests
 
 import utils.gql as gql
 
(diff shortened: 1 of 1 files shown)


mmclanerh:
code looks fine to me! however: this can add a very wide broadcast for notifications and can lead to all of the pitfalls of notification fatigue. i don't think it's a good idea to blindly spam all the owners of a github repo across slack; of which slack is ephemeral and asynchronous by nature in the space of notification-accountability. this is (probably) ok if it's explicitly opt-in by all owners of a repo so there is explicit understanding what they are signing up for. note: this type of notification will not take into account timezones, vacations, etc which are often critical to keeping notifications from becoming noise. good luck!

maorfr:
i totally agree. i will add an option for a user to opt out of slack user groups in a future PR. this is a really good idea!

PR #144, opened 2019-07-23
[gitlab-members] integration cleanup

SydneyDlh:
- change access level from integer to string with an enum
- update qontract-reconcile README with the new integration
- fix integration logging to be more like `INFO: ['add_user_to_group', u'app-sre', u'owner', u'jzelinsk']`

### DIFF
--- a/utils/gitlab_api.py
+++ b/utils/gitlab_api.py
@@ -127,7 +127,9 @@ class GitLabApi(object):
             logging.error(group_name + " group not found")
             return []
         group = self.gl.groups.get(group_name)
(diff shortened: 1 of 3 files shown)


utils/gitlab_api.py
```diff
@@ -110,12 +110,26 @@ def check_group_exists(self, group_name):
             return False
         return True
 
+    def get_string_access_level(self, access):
```
jmelis: can you refactor the code that translates from access to number? I see this is repeated

SydneyDlh: one of the times it is string -> number the other it is number -> string

PR #149, opened 2019-07-28
[github-orgs] allow team membership in unmanaged orgs

maorfr:
this integration does:
1. migrate the github configuration from config.toml to app-interface (1st commit)
2. adds a github-org schema object, which may define a `managedTeams` field.
  * if this field is defined - we only manage the specified teams
  * if this field is not defined - we manage the entire org

### DIFF
--- a/reconcile/github_org.py
+++ b/reconcile/github_org.py
@@ -151,10 +151,7 @@ class GHApiStore(object):
     def __init__(self, config):
         for org_name, org_config in config['github'].items():
             token = org_config['token']
-            if 'managed_teams' in org_config:
-                managed_teams = org_config['managed_teams']
-            else:
-                managed_teams = None
+            managed_teams = org_config.get('managed_teams', None)
             self._orgs[org_name] = \
                 (Github(token), RawGithubApi(token), managed_teams)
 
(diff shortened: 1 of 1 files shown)


reconcile/github_org.py
```diff
@@ -33,29 +46,47 @@
...
 def fetch_current_state(gh_api_store):
     state = AggregatedList()
 
     for org_name in gh_api_store.orgs():
         g = gh_api_store.github(org_name)
         raw_gh_api = gh_api_store.raw_github_api(org_name)
+        managed_teams = gh_api_store.managed_teams(org_name)
+        # if 'managedTeams' is not specified
+        # we manage all teams
+        is_managed = managed_teams is None
```
jmelis: maybe it makes sense to do:

`is_managed = managed_teams is None or managed teams == []`

jmelis: (in case someone defines an empty `managedTeams` )

jmelis: actually, better:
`is_managed = managed_teams is None or len(managed teams) == 0`

maorfr: thanks!

PR #152, opened 2019-07-31
[openshift-groups] add threading

SydneyDlh:
add threading to the openshift groups integration to make it run faster

### DIFF
--- a/reconcile/openshift_groups.py
+++ b/reconcile/openshift_groups.py
@@ -1,5 +1,8 @@
 import sys
 import logging
+import time
+from multiprocessing.dummy import Pool as ThreadPool
+from functools import partial
 
 import utils.gql as gql
 import reconcile.openshift_resources as openshift_resources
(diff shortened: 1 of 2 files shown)


reconcile/openshift_groups.py
```diff
@@ -58,32 +61,59 @@
...
+
+def create_groups_list(clusters):
+    groups_list = []
     for cluster_info in clusters:
         groups = cluster_info['managedGroups']
         if groups is None:
             continue
+        for group_name in groups:
+            groups_list.append({
+                    "cluster_info": cluster_info,
```
maorfr: can we append only the cluster name instead of the cluster_info?
just to simplify it a bit

reconcile/openshift_groups.py
```diff
@@ -58,32 +61,59 @@
...
-                    "user": user
-                })
 
+def fetch_current_state():
+    gqlapi = gql.get_api()
+    clusters = gqlapi.query(CLUSTERS_QUERY)['clusters']
+    current_state = []
+    oc_map = create_oc_map(clusters)
+
+    pool = ThreadPool(10)
```
maorfr: can we have the `10` be passed in as a variable like in other `threaded` (hint, hint) integrations?

maorfr:
feel free to rebase, merge and ship to app-interface.
well done!

p.s. its usually good to reply and resolve the conversations ;)

PR #154, opened 2019-08-01
[openshift-*] add retry mechanism

maorfr:


maorfr:
as discussed, this will all be in a future PR!

skryzhny:
My vote is using randomized delays (for future PR)

PR #155, opened 2019-08-02
Add jinja2 templating for resources

jfchevrette:
This PR adds support for Jinja2 template processing on resource files prior to processing them.

I've tried to design the feature so that it is easy to add more kinds of (template) processing. A potential future candidate for this may be processing `jsonnet` templates to render OpenShift resources and then reconcile them

As part of the PR, I've also added a custom `vault` function in Jinja2 which allows retrieving secrets from vault

This can be useful to retrieve information from vault to apply on one or more resources

Example
```yaml
apiVersion: v1
kind: Deployment
metadata:
  name: myapp
spec:
  ...
  template:
    spec:
      containers:
      - image: '{{ vault('some/vault/path', 'image', 1) }}'
        ...
```

### DIFF
--- a/reconcile/openshift_resources.py
+++ b/reconcile/openshift_resources.py
@@ -324,7 +324,7 @@ def fetch_openshift_resource(resource, parent):
         if resource['variables']:
             tv = anymarkup.parse(resource['variables'], force_types=None)
         tv['resource'] = resource
-        tv['resource']['parent'] = parent
+        tv['resource']['namespace'] = parent
         tt = resource['type']
         tt = 'jinja2' if tt is None else tt
         if tt == 'jinja2':
--- a/reconcile/openshift_resources.py
+++ b/reconcile/openshift_resources.py
@@ -4,6 +4,7 @@ import base64
 import json
 
 import anymarkup
+import jinja2
 import semver
 
 import utils.gql as gql
@@ -42,6 +43,10 @@ NAMESPACES_QUERY = """
       provider
       ... on NamespaceOpenshiftResourceResource_v1 {
         path
+      }
+      ... on NamespaceOpenshiftResourceResourceTemplate_v1 {
+        path
+        type
       }
       ... on NamespaceOpenshiftResourceVaultSecret_v1 {
         path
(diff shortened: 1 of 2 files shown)


maorfr:
hey @jfchevrette,

first of all, thanks a lot for this PR!
this is something that we needed for quite a long time. thanks for picking it up!

i'll start with - this looks good to me. i particularly liked the exceptions flow (throw one, catch it, throw another, etc). does this have a formal name? ;)
i am fine with merging this PR as is. however, i do have a couple of thoughts that i would like to share with you, and let you decide if this should go into this PR or (if at all) to a future PR.

this PR allows us to have jinja2 templates in resources in app-interface, which is really awesome, and it is for sure something we could use right away.

when thinking of the top objective of templating in my mind, i immediately think of having the ability to define a resource (say, a `ServiceMonitor`) once, and use it across multiple environments.

consider this example:
```yaml
kind: ServiceMonitor
...
spec:
  namespaceSelector:
    matchNames:
    - {{ namespace }}
  ...
```

if i understand correctly, this PR only allows templating using values from vault, and not values defined directly in app-interface.

would it make sense to allow something similar to this in app-interface and have the integration handle it?
```yaml
- provider: resource-template
  path: /path/to/resource
  type: jinja2
  variables:
    namespace: my-namespace
```

in addition, would it make sense to apply all templating options to every resource?
i mean, instead of having to define `provider: resource-template`, could we (implicitly) template every resource that goes through `fetch_provider_resource`? and apply all templating options (jinja2, extracurlyjinja2, and the direct method i described)?

everything that i mentioned above will help us reduce duplications across our resources significantly.
again, i'm totally fine with merging as is and doing all these suggestions future in PRs!
what do you think?

jfchevrette:
Thanks for the detailed review!

> would it make sense to allow something similar to this in app-interface and have the integration handle it?
> 
> ```yaml
> - provider: resource-template
>   path: /path/to/resource
>   type: jinja2
>   variables:
>     namespace: my-namespace
> ```

I've added basic variables support in 163588d and 3268c15

I've also added a change in 765d6bc that passes the resource spec and it's parent spec to the template engine(s). This allow things like:

```yaml
apiVersion: v1
kind: SomeResource
metadata:
  labels:
    cluster: {{resource.namespace.cluster.name}}
    description: {{resource.namespace.description}}
    app: {{resource.namespace.app.name}}
    appdesc: {{resource.namespace.app.description}}
  name: my-cool-resource
spec:
  ...
```

Though the above directly depends on what is returned by the GraphQL query. Additional fields would need to be added to be able to access them in the template.

> in addition, would it make sense to apply all templating options to every resource?
> i mean, instead of having to define `provider: resource-template`, could we (implicitly) template every resource that goes through `fetch_provider_resource`? and apply all templating options (jinja2, extracurlyjinja2, and the direct method i described)?

I would not do that because it may cause results which users may not be expecting. Having a distinct resource type for templates makes it explicit that the file will be processed some way or another. One gotcha I hit during development is that some files already contains delimiters such as `{{` which we do not want to process.

Another thing to keep in mind is that both go templates and jinja2 use `{{` and `}}` as delimiters by default so I think it's better if users make a concious decision when using templates.

PR #162, opened 2019-08-07
Retry decorator

SydneyDlh:
created a retry decorator that retries the function if it throws an error as requested in this [PR](https://github.com/app-sre/qontract-reconcile/pull/154#pullrequestreview-269575846)

### DIFF
--- a/utils/retry.py
+++ b/utils/retry.py
@@ -1,0 +1,30 @@
+
+from functools import wraps
+import time
+
+class StatusCodeError(Exception):
+    pass
+
+
+class NoOutputError(Exception):
+    pass
+
+def retry():
+
+    def deco_retry(f):
+
+        @wraps(f)
+        def f_retry(*args, **kwargs):
+            attempt = 0
+            attempts = 3
+            while True:
+                try:
+                    return f(*args, **kwargs)
+                except Exception as e:
+                    attempt += 1
+                    if attempt == attempts:
+                        raise e
+                    else:
+                        time.sleep(attempt)
+        return f_retry
+    return deco_retry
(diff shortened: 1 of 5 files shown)


utils/retry.py
```diff
@@ -0,0 +1,31 @@
...
+
+        @wraps(f)
+        def f_retry(*args, **kwargs):
+            attempt = 0
+            attempts = 3
+            while True:
+                try:
+                    return f(*args, **kwargs)
+                except Exception as e:
+                    attempt += 1
```
jmelis: we can drop this line if we change the `while` above

utils/retry.py
```diff
@@ -0,0 +1,31 @@
...
+                    if attempt == attempts:
```
jmelis: and this can we change to:

`if attempt > max_attempts - 1`

utils/retry.py
```diff
@@ -0,0 +1,31 @@
...
+                        raise e
+                    else:
```
jmelis: no need for this `else`, just the `time.sleep`  after the `if` statement above

utils/retry.py
```diff
@@ -0,0 +1,32 @@
...
+# source: https://www.calazan.com/retry-decorator-for-python-3/
+
+
+def retry(max_attempts=3):
+
+    def deco_retry(f):
+
+        @wraps(f)
+        def f_retry(*args, **kwargs):
+            attempt = 0
```
jmelis: ah, sorry, I missed this... this line is not needed anymore

maorfr:
thanks for this as well @SydneyDlh!

PR #165, opened 2019-08-11
[openshift-users] introduce new integration

maorfr:
this integration does:

for each cluster - get the list of users, compare with list of desired users (have any RoleBinding in openshift-rolebindings or members of any group in openshift-groups), and deletes users who are not found in the desired state.

this integration only deletes users with `github` identities.

this integration does not create users.


jmelis:
[test]

jmelis:
[test]

PR #185, opened 2019-08-29
consolidate usage of oc_map

maorfr:
as many integrations were added in a short period of time, a lot of code was duplicated.
this PR consolidates the usage of oc_map by creating a reusable `OC_Map` class.

this will also lay the ground work for #112.

covers https://jira.coreos.com/browse/APPSRE-922
contributes to https://jira.coreos.com/browse/APPSRE-897

PR #186, opened 2019-08-29
[openshift-rolebinding] use oc instead of api

### LINKED ISSUE #112: Have a single way to interact with OpenShift API

maorfr:
depends on #185 
closes #112 
covers https://jira.coreos.com/browse/APPSRE-897


jfchevrette:
I went through the changes and I see no glaring issues.

Thanks for this huge and much needed refactor :)

jfchevrette:
LGTM! 👍 🚀

jfchevrette:
I think #185 should be merged first so we have a better rollback plan in case something goes wrong with the changes from this PR

PR #187, opened 2019-09-01
[openshift-*] handle cleanup of oc_map (jump host)

maorfr:
when initializing oc_map, temporary directories are created for jump host configurations.
this PR adds a cleanup of those directories.
covers https://jira.coreos.com/browse/APPSRE-605

### DIFF
--- a/e2e_tests/create_namespace.py
+++ b/e2e_tests/create_namespace.py
@@ -4,9 +4,13 @@ import logging
 import e2e_tests.test_base as tb
 import e2e_tests.dedicated_admin_test_base as dat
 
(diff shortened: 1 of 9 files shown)


maorfr:
> Perhaps at some point we'll want the deferred function to be executed immediately after the integration finishes instead of at the very end, but we can leave that for a future refactor.
can you elaborate?

> I also seem to notice none of the current integrations calls were changed to use this, is this expected?
all openshift related integrations now use this. i added the `@defer` decorator to `run` functions, as this is the function that holds `oc_map` which should be deferred to do `cleanup`.

PR #190, opened 2019-09-01
[openshift-resources] Routes - validate TLS certificate FQDN matches route FQDN

maorfr:
covers https://jira.coreos.com/browse/APPSRE-637
depends on #188

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -235,7 +235,7 @@ def aws_iam_keys(ctx, thread_pool_size):
 
 @integration.command()
 @threaded(default=20)
-@binary(['oc', 'ssh'])
+@binary(['oc', 'ssh', 'openssl'])
 @click.pass_context
 def openshift_resources(ctx, thread_pool_size):
     run_integration(reconcile.openshift_resources.run,
--- a/reconcile/openshift_resources.py
+++ b/reconcile/openshift_resources.py
@@ -9,6 +9,7 @@ import semver
 
 import utils.gql as gql
 import utils.vault_client as vault_client
+import utils.openssl as openssl
 
 from utils.oc import OC_Map, StatusCodeError
 from utils.defer import defer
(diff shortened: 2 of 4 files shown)


reconcile/cli.py
```diff
@@ -231,6 +231,8 @@ def aws_iam_keys(ctx, thread_pool_size):
 
 @integration.command()
 @threaded(default=20)
+# merge after https://github.com/app-sre/qontract-reconcile/pull/188
+# binary(['oc', 'ssh', 'openssl'])
```
jfchevrette: To be changed after merging #188

reconcile/openshift_resources.py
```diff
@@ -292,6 +300,14 @@ def fetch_provider_route(path, tls_path, tls_version):
         logging.info(msg)
         _log_lock.release()
 
+    host = openshift_resource.body['spec'].get('host')
+    certificate = openshift_resource.body['spec']['tls'].get('certificate')
+    if host and certificate:
+        match = openssl.certificate_matches_host(certificate, host)
```
jfchevrette: Will this work for wildcards?

utils/openssl.py
```diff
@@ -0,0 +1,18 @@
...
+def get_certificate_common_name(certificate):
+    proc = Popen(
+        ['openssl', 'x509', '-noout', '-subject'],
+        stdin=PIPE,
+        stdout=PIPE,
+        stderr=PIPE
+    )
+    out, err = proc.communicate(certificate)
+
+    return out.split('/CN=')[1].strip()
```
jfchevrette: Somewhere in there we should probably extracts SANs and check those too. 

https://stackoverflow.com/questions/20983217/how-to-display-the-subject-alternative-name-of-a-certificate

Some certificates we (may) have would have something like `domain.com` as the CN and then a list of alternative domains in the SAN extension (ex: www.domain.com, foo.domain.com, bar.domain, ...)

Or we can wait until we hit this and refactor at that point in time

reconcile/cli.py
maorfr: done

reconcile/openshift_resources.py
maorfr: it will, thanks to:
https://github.com/app-sre/qontract-reconcile/pull/190/files#diff-b1b248c88c2f9d9e7923df0911315418R6

utils/openssl.py
maorfr: that makes sense, and i'd prefer to solve when it hits us :)

maorfr:
thanks a lot for the review!

PR #192, opened 2019-09-02
[quay-membership] Hurry up

maorfr:
instead of testing each user processed, only check users about to be added (as anyone already in the team obviously exists).


maorfr:
reverts most of the changes done in #134

PR #201, opened 2019-09-10
[openshift-*] add option to disable specific clusters

maorfr:
This PR adds the ability to specify integrations and e2e tests that should not be executed for specific clusters.

### DIFF
--- a/reconcile/openshift_resources.py
+++ b/reconcile/openshift_resources.py
@@ -84,6 +84,9 @@ NAMESPACES_QUERY = """
         field
         format
       }
+      disable {
+        integrations
+      }
     }
   }
 }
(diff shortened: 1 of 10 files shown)


reconcile/openshift_resources.py
```diff
@@ -494,7 +504,8 @@ def init_specs_to_fetch(ri, oc_map, namespaces,
 
 def fetch_data(namespaces, thread_pool_size):
     ri = ResourceInventory()
-    oc_map = OC_Map(namespaces=namespaces)
+    oc_map = OC_Map(namespaces=namespaces,
+                    integration=QONTRACT_INTEGRATION.replace('_', '-'))
```
jmelis: can we move the replace to inside the function? I would also add a regex to the schema to ensure that no disabled integrations contain `_` in the app-interface.

maorfr: fixed the replace.
regarding the regex - i planned to reference the integration objects in app-interface at a later iteration. for now, i've created an enum.

PR #230, opened 2019-10-02
[jira-watcher] introduce new integration

maorfr:
this integration detects changes in Jira issues and sends updates on Slack

### DIFF
--- a/reconcile/jira_watcher.py
+++ b/reconcile/jira_watcher.py
@@ -69,21 +69,21 @@ def format_message(server, key, event,
 def calculate_diff(server, current_state, previous_state):
     messages = []
     new_issues = [format_message(server, key, 'created')
-                    for key in current_state
-                    if key not in previous_state]
+                  for key in current_state
+                  if key not in previous_state]
     messages.extend(new_issues)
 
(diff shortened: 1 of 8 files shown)


reconcile/jira_watcher.py
```diff
@@ -0,0 +1,70 @@
...
+    for jira_board in jira_boards:
+        jira = JiraClient(jira_board)
+        issues = jira.get_issues()
+        current_state = {issue.key: issue.fields.status.name
+                         for issue in issues}
+        project_file_path = get_project_file_path(io_dir, jira.project)
+
+        try:
+            with open(project_file_path, 'r') as f:
+                previous_state = json.load(f)
```
jmelis: tbh this struck me a bit odd at first, but after thinking about this for 5 minutes I agree this is the best solution, I mean keeping the previous state in the file, and overwriting it each cycle. My first impulse would have been to iterate through all issues and report if any changes have been in the last X minutes, where X minutes is the cadence of the job periodicity. But this makes much more sense.

awesome work @maorfr

maorfr: thanks for the cool feedback!

PR #246, opened 2019-10-16
Switch openssl utils to pyOpenSSL

jfchevrette:
Using the `openssl` binary and parsing the output to extract certificate info is error prone.

This PR changes this to use pyOpenSSL instead.

This fixes an issue where on my system running OpenSSL 1.1.1d the output format is different from version 1.0.2k in CentOS 7

OpenSSL 1.1.1d
```
# openssl x509 -noout -subject < crt
subject=C = US, ST = North Carolina, L = Raleigh, O = "Red Hat, Inc.", OU = Developer Tools, CN = *.api.prod-preview.openshift.io
```

OpenSSL 1.0.2k
```
# openssl x509 -noout -subject < crt
subject= /C=US/ST=North Carolina/L=Raleigh/O=Red Hat, Inc./OU=Developer Tools/CN=*.api.prod-preview.openshift.io
```

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -254,7 +254,7 @@ def aws_iam_keys(ctx, thread_pool_size):
 
 @integration.command()
 @threaded(default=20)
-@binary(['oc', 'ssh', 'openssl'])
+@binary(['oc', 'ssh'])
 @click.pass_context
 def openshift_resources(ctx, thread_pool_size):
     run_integration(reconcile.openshift_resources.run,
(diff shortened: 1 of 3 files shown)


jfchevrette:
@maorfr PTAL

jfchevrette:
@maorfr PTAL

maorfr:
ah, we should probably remove `openssl` from the binary check:
https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/cli.py#L257

PR #252, opened 2019-10-19
[aws-support-cases-sos] introduce new integration

maorfr:
This PR introduces a new integration called aws-support-cases-sos.

This integration iterates over support cases in AWS support center to find support cases reporting of leaked AWS access keys. For each found key, the integration will submit a PR to app-interface to remove that key from the account (will be handled by aws-iam-keys integration upon merge).

Future work will include searching the page where the leak was detected for leaked GitHub tokens, and if those exist - the PR to app-interface will be one that removes the user completely.

**NOTE** This PR removes python 3.4 from tests due to incompatibility with ruamel.yaml.

### DIFF
--- a/reconcile/aws_support_cases_sos.py
+++ b/reconcile/aws_support_cases_sos.py
@@ -1,7 +1,11 @@
+import logging
+
 import utils.gql as gql
 
 from reconcile.queries import AWS_ACCOUNTS_QUERY
+from reconcile.queries import GITLAB_INSTANCES_QUERY
 from utils.aws_api import AWSApi
+from utils.gitlab_api import GitLabApi
 
 
 def get_deleted_keys(accounts):
@@ -25,7 +29,7 @@ def get_keys_to_delete(aws_support_cases):
     return keys
 
 
-def run(dry_run=False, thread_pool_size=10, enable_deletion=False):
+def run(gitlab_project_id, dry_run=False, thread_pool_size=10, enable_deletion=False):
     gqlapi = gql.get_api()
     accounts = gqlapi.query(AWS_ACCOUNTS_QUERY)['accounts']
     aws = AWSApi(thread_pool_size, accounts)
@@ -36,6 +40,15 @@ def run(dry_run=False, thread_pool_size=10, enable_deletion=False):
     keys_to_delete = [ktd for ktd in keys_to_delete_from_cases
                       if ktd['key'] not in deleted_keys[ktd['account']]
                       and ktd['key'] in existing_keys[ktd['account']]]
+
+    if not dry_run and keys_to_delete:
+        # assuming a single GitLab instance for now
+        instance = gqlapi.query(GITLAB_INSTANCES_QUERY)['instances'][0]
+        gl = GitLabApi(instance, project_id=gitlab_project_id)
+
     for k in keys_to_delete:
-        print(k['account'])
-        print(k['key'])
+        logging.info(['delete_aws_access_key', k['account'], k['key']])
+        if not dry_run:
+            path = 'data' + \
+                [a['path'] for a in accounts if a['name'] == k['account']][0]
+            gl.create_delete_aws_access_key_mr(account, path, k['key'])
--- a/reconcile/aws_support_cases_sos.py
+++ b/reconcile/aws_support_cases_sos.py
@@ -29,7 +29,8 @@ def get_keys_to_delete(aws_support_cases):
     return keys
 
 
-def run(gitlab_project_id, dry_run=False, thread_pool_size=10, enable_deletion=False):
+def run(gitlab_project_id, dry_run=False, thread_pool_size=10,
+        enable_deletion=False):
     gqlapi = gql.get_api()
     accounts = gqlapi.query(AWS_ACCOUNTS_QUERY)['accounts']
     aws = AWSApi(thread_pool_size, accounts)
--- a/reconcile/aws_support_cases_sos.py
+++ b/reconcile/aws_support_cases_sos.py
@@ -47,8 +47,10 @@ def run(gitlab_project_id, dry_run=False, thread_pool_size=10, enable_deletion=F
         gl = GitLabApi(instance, project_id=gitlab_project_id)
 
     for k in keys_to_delete:
-        logging.info(['delete_aws_access_key', k['account'], k['key']])
+        account = k['account']
+        key = k['key']
+        logging.info(['delete_aws_access_key', account, key])
         if not dry_run:
             path = 'data' + \
-                [a['path'] for a in accounts if a['name'] == k['account']][0]
-            gl.create_delete_aws_access_key_mr(account, path, k['key'])
+                [a['path'] for a in accounts if a['name'] == account][0]
+            gl.create_delete_aws_access_key_mr(account, path, key)
--- a/reconcile/aws_support_cases_sos.py
+++ b/reconcile/aws_support_cases_sos.py
@@ -17,6 +17,9 @@ def get_deleted_keys(accounts):
 def get_keys_to_delete(aws_support_cases):
     search_pattern = 'We have become aware that the AWS Access Key '
     keys = []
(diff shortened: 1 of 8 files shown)


maorfr:
current status: https://gitlab.cee.redhat.com/service/app-interface/merge_requests/1871

maorfr:
ready for review.
example: https://gitlab.cee.redhat.com/service/app-interface/merge_requests/1873/diffs

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = flake8, py27, py34, py36
+envlist = flake8, py27, py36
```
jmelis: remove this too: https://github.com/app-sre/qontract-reconcile/blob/master/setup.py#L50

utils/aws_api.py
```diff
@@ -424,3 +424,19 @@ def get_users_keys(self):
     def get_user_keys(self, iam, user):
         key_list = iam.list_access_keys(UserName=user)['AccessKeyMetadata']
         return [uk['AccessKeyId'] for uk in key_list]
+
+    def get_support_cases(self):
+        self.support_cases = {}
```
jmelis: why `self`?

utils/aws_api.py
```diff
@@ -424,3 +424,19 @@ def get_users_keys(self):
...
+            try:
+                support = s.client('support')
+                support_cases = support.describe_cases(
+                    includeResolvedCases=True,
+                    includeCommunications=True
+                )['cases']
+                self.support_cases[account] = support_cases
+            except Exception:
+                msg = '[{}] error getting support cases, check support level'
+                logging.error(msg.format(account))
```
jmelis: don't we want to log the actual error returned?

utils/gitlab_api.py
```diff
@@ -1,8 +1,8 @@
 import logging
-
 import gitlab
 import urllib3
 import uuid
+import ruamel.yaml as yaml
```
jmelis: :heart:

utils/gitlab_api.py
```diff
@@ -44,7 +44,21 @@ def delete_file(self, branch_name, file_path, commit_message):
                 {
                     'action': 'delete',
                     'file_path': file_path
-                    }
+                }
```
jmelis: had this escaped flake8???

utils/gitlab_api.py
```diff
@@ -44,7 +44,21 @@ def delete_file(self, branch_name, file_path, commit_message):
...
+            ]
+        }
+        self.project.commits.create(data)
+
+    def update_file(self, branch_name, file_path, commit_message, content):
```
jmelis: perhaps wortth supporting a list of files to update?

`def update_file(self, branch_name, commit_message, file_contents):`

where `file_contents` is a dictionary where the key is the path and the value is the content? or a list of tuples, where the first element is the path, and the second the content?

It's a minor change and increases reusability.

utils/gitlab_api.py
```diff
@@ -102,6 +116,41 @@ def create_delete_user_mr(self, username, paths):
...
+        except gitlab.exceptions.GitlabCreateError as e:
+            self.delete_branch(branch_name)
+            if str(e) != "400: A file with this name doesn't exist":
+                raise e
+            logging.info(
+                "File {} does not exist, not opening MR".format(path)
+            )
+            return
+
+        self.create_mr(branch_name, target_branch, title)
```
jmelis: no exception handling here?

reconcile/aws_support_cases_sos.py
```diff
@@ -0,0 +1,57 @@
...
+from reconcile.queries import AWS_ACCOUNTS_QUERY
+from reconcile.queries import GITLAB_INSTANCES_QUERY
+from utils.aws_api import AWSApi
+from utils.gitlab_api import GitLabApi
+
+
+def get_deleted_keys(accounts):
+    return {account['name']: account['deleteKeys']
+            for account in accounts
+            if account['deleteKeys'] not in (None, [])}
```
jmelis: good practice, I like that `not in`

reconcile/aws_support_cases_sos.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def get_keys_to_delete(aws_support_cases):
+    search_pattern = 'We have become aware that the AWS Access Key '
+    keys = []
+    for account, cases in aws_support_cases.items():
+        for case in cases:
+            comms = case['recentCommunications']['communications']
```
jmelis: can you leave a comment here with the expected content of the communication? if they change it in the future it would be nice to have it handy

jmelis: also, pehaps we could come up with a regexp that validates that the response has the format we are expecting, and handle the exception otherwise

reconcile/aws_support_cases_sos.py
```diff
@@ -0,0 +1,57 @@
...
+            for comm in comms:
+                body = comm['body']
+                split = body.split(search_pattern, 1)
+                if len(split) == 2:  # sentence is found, get the key
+                    key = split[1].split(' ')[0]
+                    keys.append({'account': account, 'key': key})
```
jmelis: I prefer using tuples for this.

utils/aws_api.py
maorfr: for future reuse

utils/gitlab_api.py
maorfr: :rocket:

utils/gitlab_api.py
maorfr: yup!

utils/aws_api.py
jmelis: I would initialize it to `None` then, in the class

utils/gitlab_api.py
maorfr: each time `update_file` is called, it creates a commit. we are later creating a single MR containing all commits.
i think its better this way. wdyt?

utils/gitlab_api.py
maorfr: not at the moment. we have never seen this one fail before, so i'd like to just let a potential exception propogate at this point.

we can add exception handling to `create_mr` when we meet one. wdyt?

reconcile/aws_support_cases_sos.py
maorfr: yay!

reconcile/aws_support_cases_sos.py
maorfr: i'm leaving room for future work.
in addition, this is how we do it in other integrations as well (using dictionaries). wdyt?

tox.ini
maorfr: done, thanks!

utils/aws_api.py
maorfr: remove it from self.

utils/aws_api.py
maorfr: done, thanks!

reconcile/aws_support_cases_sos.py
maorfr: added a comment with the reference.
we are currently not using regex, but splitting by `search_pattern`.
from what i've seen until now, the `search_pattern` is consistent across AWS support cases.

utils/gitlab_api.py
jmelis: if we do it like I proposed we leave it up to the user of this library

utils/gitlab_api.py
jmelis: sgtm

reconcile/aws_support_cases_sos.py
jmelis: true, it's just that in this case it's a nice fit, but no need to do this now, I agree

utils/gitlab_api.py
maorfr: agreed to be done in future iteration.

PR #258, opened 2019-10-23
New integration: openshift-acme

jfchevrette:
This integration can be used to deployment [openshift-acme](https://github.com/tnozicka/openshift-acme) to namespaces.


jfchevrette:
@maorfr PTAL

PR #259, opened 2019-10-23
[WIP] New integration: openshift-clusterrolebinding

jfchevrette:
This is a first stab at a ClusterRoleBinding integration

This allows setting ClusterRoleBindings under a cluster definition as such:
```yaml
clusterRoleBindings:
- name: testjfc
  roleRef:
    name: testjfc
  subjects:
  - kind: ServiceAccount
    name: my-sa
    namespace: my-ns
  userNames:
  - system:serviceaccount:my-ns:my-sa
- name: my-cool-crb
  roleRef:
    name: my-cool-cr
  subjects:
  - kind: User
    name: foo
  - kind: User
    name: bar
  userNames:
  - foo
  - bar
```

It uses the usual openshift_resources functions so qontract annotations are set. It'll happily start to manage existing ClusterRoleBindings if told to.

Notes for reviewers:
- I did some not-so-greatcode gymnastics in `utils/oc.py` and `utils/openshift_resource.py` where I check or set the namespace name to `_` (underscore) for global/cluster-scoped resources to avoid having to do a larger refactor.
- I had to refactor `fetch_current_state` and `init_specs_to_fetch` to teach them how to handle cluster-scoped resources. I think I've updated all code that used these functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,10 @@ In addition, e2e tests are available to detect potential problems reconciling se
 
 - `aws-garbage-collector`: Delete orphan AWS resources.
 - `aws-iam-keys`: Delete IAM access keys by access key ID.
+- `aws-support-cases-sos`: Scan AWS support cases for reports of leaked keys and remove them (only submits PR)
 - `github`: Configures the teams and members in a GitHub org.
 - `github-repo-invites`: Accept GitHub repository invitations for known repositories.
+- `github-scanner`: Scan GitHub repositories for leaked keys and remove them (only submits PR).
 - `github-users`: Validate compliance of GitHub user profiles.
 - `gitlab-housekeeping`: Manage issues and merge requests on GitLab projects.
 - `gitlab-members` : Manage GitLab group members.
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -7,6 +7,7 @@ import utils.gql as gql
 import reconcile.github_org
 import reconcile.github_users
 import reconcile.github_scanner
+import reconcile.openshift_clusterrolebindings
 import reconcile.openshift_rolebindings
 import reconcile.openshift_groups
 import reconcile.openshift_users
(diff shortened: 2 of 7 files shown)


maorfr:
the change in openshift_base to handle both namespaces and clusters looks good.
setting the namespace to `_` for now seems reasonable, we can refactor namespace to be a kv arg or anything else later. for sure outside the scope of this PR.

i love all the reusing you did!

i think the approach you selected to define the ClusterRoleBindings in the schema is solid and isolated (as in, nothing else has to be defined in app-interface for this to work), i like that.

i just wanted to comment that it is very different from how we handle RoleBindings, which is probably fine, but i suppose we may want them to be more similar? we just did a major refactor with RoleBindings, here is an example:
https://gitlab.cee.redhat.com/service/app-interface/blob/master/data/teams/app-sre/roles/app-sre-managed-namespaces-admin.yml#L9

wdyt?

overall this is very nice work!

jfchevrette:
@maorfr I originally though we could refactor the RoleBinding integration in a similar (almost identical) way to the ClusterRoleBinding one. Both are using the same `roleRef` schema in the k8s API so effectively both could be managed the same, with the exception that one is namespace-scoped and the other is cluster-scoped.

For now I'm going to merge this and we can revisit this at a later time

jfchevrette:
@maorfr I've pushed a change in 4d0258c that will prevent overwriting existing ClusterRoleBinding objects that don't already have the qontract-reconcile annotations. A message will be shown:

```
ERROR: [app-sre] resource 'ClusterRoleBinding/cluster-readers' present w/o annotations. Refusing to overwrite an unmanaged ClusterRoleBinding
ERROR: One or more desired ClusterRoleBinding object already exist on the cluster without qontract-reconcile annotations. If you are sure you want to override it you can add the qontract-reconcile annotation manually and run the integration again.
```

maorfr:
We've had that initially with openshift-resources and ended up automatically taking over. I think we should keep consistency around that. Wdyt?

maorfr:
And LGTM, merge at will and we can iterate!

maorfr:
we should also consider the case when a user is deleted from app-interface (how will we remove his ClusterRoleBindings)

jfchevrette:
I will re-do this change and mimic what we're doing for regular RoleBindings.

PR #268, opened 2019-10-27
pull request gateway

maorfr:
covers part of https://jira.coreos.com/browse/APPSRE-1148

this PR introduces a PR gateway - in case the repo to submit PRs to is not reachable, define `pullRequestGateway` to be `sqs` and the requests will be routed to a queue.

another integration will be created to read messages and submit PRs to app-interface.

### DIFF
--- a/utils/sqs_gateway.py
+++ b/utils/sqs_gateway.py
@@ -1,4 +1,10 @@
 import os
+import boto3
+
+
+class SQSGatewayInitError(Exception):
+    pass
+
 
 class SQSGateway(object):
     """Wrapper around SQS AWS SDK"""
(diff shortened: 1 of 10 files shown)


utils/sqs_gateway.py
```diff
@@ -0,0 +1,47 @@
...
+
+class SQSGatewayInitError(Exception):
+    pass
+
+
+class SQSGateway(object):
+    """Wrapper around SQS AWS SDK"""
+
+    def __init__(self):
+        access_key = os.environ['aws_access_key_id']
```
jmelis: add decorator

utils/sqs_gateway.py
```diff
@@ -0,0 +1,47 @@
...
+            aws_secret_access_key=secret_key,
+            region_name=region_name,
+        )
+        self.sqs = session.client('sqs')
+        self.queue_url = queue_url
+
+    def send_message(self, body):
+        self.sqs.send_message(
+            QueueUrl=self.queue_url,
+            MessageBody=str(body)
```
jmelis: change to json.dumps

utils/sqs_gateway.py
maorfr: will do

utils/sqs_gateway.py
maorfr: done

PR #312, opened 2019-11-06
Introducing Secret reader

maorfr:
covers https://jira.coreos.com/browse/APPSRE-1102

This PR introduces a new abstraction to read secrets called secret reader.

Secret reader reads secrets from Vault or from a configuration file (currently not implemented), to allow the use of a Vault-less app-interface.

To specify Vault as the secret backend, the app-interface [settings](https://gitlab.cee.redhat.com/service/app-interface/blob/master/data/app-interface/app-interface-settings.yml#L9-11) should include `vault: true`.


utils/secret_reader.py
```diff
@@ -0,0 +1,48 @@
...
+      * Note: if this is Vault secret and a v2 KV engine
+
+    The input settings is an optional app-interface-settings object
+    queried from app-interface.
+
+    Default vault setting is false, to allow using a config file
+    without creating app-interface-settings.
+    """
+
+    if settings and settings.get('vault'):
```
rrati: Does this key need to be documented in the function description?  I don't see "vault" listed in the description of the fields in settings.

utils/secret_reader.py
```diff
@@ -0,0 +1,48 @@
...
+        return vault_client.read_all(secret)
```
rrati: Same comment here

README.md
```diff
@@ -30,7 +30,6 @@ Additional tools that use the libraries created by the reconciliations are also
 - `openshift-groups`: Manages OpenShift Groups.
 - `openshift-namespaces`: Manages OpenShift Namespaces.
 - `openshift-network-policies`: Manages OpenShift NetworkPolicies.
-- `openshift-resources-annotate`: Annotates OpenShift Resources so they can be used by the `openshift-resources` integration.
```
jfchevrette: Any special reason why is this part of this PR?

utils/secret_reader.py
maorfr: The input settings is an optional app-interface-settings object queried from app-interface.

One of the keys is vault.

README.md
maorfr: Since this pr would have required changes in this "integration", and it is not used already, i figured a cleanup would be best...

utils/secret_reader.py
maorfr: added vault to the input settings description!

PR #314, opened 2019-11-06
Add openshift-limitranges integration

jfchevrette:


reconcile/openshift_limitranges.py
```diff
@@ -0,0 +1,109 @@
...
+        for l in limitranges['limits']:
+            for ltype in SUPPORTED_LIMITRANGE_TYPES:
+                if ltype in l and l[ltype] is not None:
+                    speclimit[ltype] = l[ltype]
+            body['spec']['limits'].append(speclimit)
+
+        resource = OR(body, QONTRACT_INTEGRATION, QONTRACT_INTEGRATION_VERSION)
+
+        # Create the resources and append them to the namespace
+        namespace["resources"] = [resource]
```
maorfr: i really like it that you modify the existing namespace instead of creating a new object. that makes a lot of sense and i will start doing it as well!
its also good that you are creating a list, following this comment:
https://gitlab.cee.redhat.com/service/app-interface/merge_requests/2039#note_967595

does that make sense?

reconcile/openshift_limitranges.py
```diff
@@ -0,0 +1,109 @@
...
+@defer
+def run(dry_run=False, thread_pool_size=10, defer=None):
+    gqlapi = gql.get_api()
+    namespaces = [namespace_info for namespace_info
+                  in gqlapi.query(NAMESPACES_QUERY)['namespaces']
+                  if namespace_info.get('limitRanges')]
+
+    namespaces = construct_resources(namespaces)
+
+    if not namespaces:
```
maorfr: since in `contruct_resources` you are modifying the existing namespaces, can namespaces actually return empty?

reconcile/openshift_limitranges.py
```diff
@@ -0,0 +1,109 @@
...
+                    # re-add to inventory
+                    ri.add_current(cluster, namespace,
+                                   resource_type, name, c_item)
+
+
+@defer
+def run(dry_run=False, thread_pool_size=10, defer=None):
+    gqlapi = gql.get_api()
+    namespaces = [namespace_info for namespace_info
+                  in gqlapi.query(NAMESPACES_QUERY)['namespaces']
```
maorfr: +1 on using NAMESPACES_QUERY!
may i suggest to add a `get_namespaces` function to queries.py and use it here?

reconcile/openshift_limitranges.py
```diff
@@ -0,0 +1,109 @@
...
+
+        # Create the resources and append them to the namespace
+        namespace["resources"] = [resource]
+
+    return namespaces
+
+
+def add_desired_state(namespaces, ri):
+    for namespace in namespaces:
+        for resource in namespace["resources"]:
```
maorfr: since we are modifying the existing namespaces, may this be uninitialized?
sorry if i'm repeating this too much and missing something obvious...

reconcile/openshift_limitranges.py
```diff
@@ -0,0 +1,109 @@
...
+            ri.add_desired(
+                namespace['cluster']['name'],
+                namespace['name'],
+                resource.kind,
+                resource.name,
+                resource
+            )
+
+
+def set_delete_state(namespaces, ri):
```
maorfr: is this function intended to allow us to "take over" all existing LimitRanges and delete ones that we do not manage?
if so, i absolutely love it. i must admit that it was very hard to understand it though.

i would say that this pattern will be very reusable in the future, and even right now. for example, when we write `openshift-resourcequotas` (right?) and even for `openshift-network-policies` to not allow manual NetworkPolicy resources in a namespace.

it absolutely doesn't have to be in this PR, but let me know if you think its a good idea!

reconcile/openshift_limitranges.py
```diff
@@ -0,0 +1,109 @@
...
+            },
+            'spec': {
+                'limits': []
+            }
+        }
+
+        # Build each limit item ignoring null ones
+        speclimit = {}
+        for l in limitranges['limits']:
+            for ltype in SUPPORTED_LIMITRANGE_TYPES:
```
maorfr: :heart:

reconcile/openshift_limitranges.py
```diff
@@ -0,0 +1,109 @@
...
+
+    ri, oc_map = \
+        ob.fetch_current_state(namespaces, thread_pool_size,
+                               QONTRACT_INTEGRATION,
+                               QONTRACT_INTEGRATION_VERSION,
+                               override_managed_types=['LimitRange'])
+    add_desired_state(namespaces, ri)
+    set_delete_state(namespaces, ri)
+
+    defer(lambda: oc_map.cleanup())
```
maorfr: can we move this to be right after the initialization of oc_map? that way we get the cleanup even if something fails in `add_desired_state` and `set_delete_state`.

reconcile/openshift_limitranges.py
```diff
@@ -0,0 +1,109 @@
...
+def construct_resources(namespaces):
+    for namespace in namespaces:
+        # Get the linked limitRanges schema settings
+        limitranges = namespace.get("limitRanges", {})
+
+        body = {
+            'apiVersion': 'v1',
+            'kind': 'LimitRange',
+            'metadata': {
+                'name': limitranges['name'],
```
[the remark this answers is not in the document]
jfchevrette: I've added a check for the 'limitRanges' key and it will print a warning if it is missing.

It should not be missing though as I filter it when doing the graphql query. Check doesn't hurt though.

reconcile/openshift_limitranges.py
```diff
@@ -0,0 +1,109 @@
...
+                    ri.add_current(cluster, namespace,
+                                   resource_type, name, c_item)
+
+
+@defer
+def run(dry_run=False, thread_pool_size=10, defer=None):
+    gqlapi = gql.get_api()
+    namespaces = [namespace_info for namespace_info
+                  in gqlapi.query(NAMESPACES_QUERY)['namespaces']
+                  if namespace_info.get('limitRanges')]
```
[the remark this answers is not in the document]
jfchevrette: see response to previous comment

reconcile/openshift_limitranges.py
jfchevrette: We want to take over the limit ranges on namespaces we control. I thought about allowing more than one to be created (ie. a list instead of a single reference) but I could not think of a use case.

I'll leave it as is for now and we can revisit if needed in the future

reconcile/openshift_limitranges.py
jfchevrette: I'd prefer that we do this in a different PR which would also change integrations that uses this query to use the new function instead

reconcile/openshift_limitranges.py
jfchevrette: You're right. I've added a key check

reconcile/openshift_limitranges.py
jfchevrette: It can if we have no namespaces that has `limitRanges` defined. It's unlikely to ever happen but it did happen initially when I developped the integration and app-interface has not been updated to have at least one namespace with that key definied.

The result was `ob.fetch_current_state` crashed somewhere down in `openshift_base` because it expects either a list of clusters or a list of namespaces, and an empty list is neither.

reconcile/openshift_limitranges.py
jfchevrette: The intent here is indeed to take over all LimitRanges and only apply ours. I'll admit this is a bit hack-ish and the preferable solution would probably be to support this in `ResourceInventory` with a function such as `ResourceInventory.delete(cluster, namespace, kind, name)`

I feel that was out of scope here and there is urgency in getting this integration deployed by next week

reconcile/openshift_limitranges.py
jfchevrette: Done!

reconcile/openshift_limitranges.py
maorfr: sgtm

reconcile/openshift_limitranges.py
maorfr: sgtm

reconcile/openshift_limitranges.py
maorfr: sgtm!

PR #323, opened 2019-11-08
[openshift-*] allow taking over resources

maorfr:
This PR introduces a new argument to `realize_data` - `take_over`.

If `take_over` is set to `True`, all existing un-managed resources (of the types managed by the running integration) will be taken over and removed if they do not exist in app-interface.

This makes an integration the sole owner of a resource type in a namespace.

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -81,6 +81,16 @@ def threaded(**kwargs):
     return f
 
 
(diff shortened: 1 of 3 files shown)


maorfr:
@jfchevrette can you ptal?

maorfr:
just copied your idea @jfchevrette

PR #328, opened 2019-11-11
[app-interface-reporter] promotions and merge activity

maorfr:
covers https://jira.coreos.com/browse/APPSRE-1190

This PR adds the following sections to an app's report:
* production promotions - number of builds for saas-deploy jobs
* merge activity - number of builds for build-master jobs

the jobs are extracted from the jjb definitions, and their history is queried from jenkins.
we are currently blocked from getting the same reports from any jobs managed in openshiftio-cico-jobs. that is handled in another story: https://jira.coreos.com/browse/APPSRE-1213


maorfr:
agreed with @jmelis to merge this as the first version and iterate over it in https://jira.coreos.com/browse/APPSRE-1298

PR #347, opened 2019-11-19
Add internal/external/all option for OpenShift related integrations

maorfr:
This PR adds a new capability for OpenShift related integrations - running them on all resources (default), running only for internal clusters (`--internal`) or running only for external clusters (`--external`).

This new feature is required so we can run our integrations on a cluster without access to internal clusters and ignore any internal clusters, or to run only for internal clusters (integration should run internally as well).

Done as part of https://jira.coreos.com/browse/APPSRE-1206

Currently we can move all our OpenShift related integrations to our cluster, except for `openshift-resources` and `terraform-resources` which should ADDITIONALLY be running internally with the `--internal` flag.

app-interface PR: https://gitlab.cee.redhat.com/service/app-interface/merge_requests/2137

important logic is here: https://github.com/app-sre/qontract-reconcile/pull/347/files#diff-5fbe62e4ee1dc72875e4c91e78c4ea25R289

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -361,10 +361,11 @@ def openshift_namespaces(ctx, thread_pool_size, internal):
 @integration.command()
 @threaded()
 @binary(['oc', 'ssh'])
+@internal()
 @click.pass_context
-def openshift_network_policies(ctx, thread_pool_size):
+def openshift_network_policies(ctx, thread_pool_size, internal):
     run_integration(reconcile.openshift_network_policies.run,
-                    ctx.obj['dry_run'], thread_pool_size)
+                    ctx.obj['dry_run'], thread_pool_size, internal)
 
 
 @integration.command()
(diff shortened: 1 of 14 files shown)


jfchevrette:
If we're going to productize app-interface/qontract-*, I think the concept of `internal`/`external` is very use-case specific. 

Could we perhaps adapt this to run (openshift) integrations based on a label selector? To my knowledge, almost all schemas support labels, so in this case we would change it to have an option in the openshift resources integrations such as `--label-selector` which would allow doing something such as `--label-selector type=internal`

I think this would lay the foundation for many other use cases

jmelis:
@jfchevrette I've discussed your suggestion with Maor, but my take is that I prefer it to be a field. The problem with labels is that they are not schema enforced, so there can be typos, or people can forget about them.

I have asked Maor to make the internal field mandatory so we can regulate the exact values

maorfr:
i will merge this as is, but i do think that the label thing is something we need to start working towards.

thank you both for the review!

PR #354, opened 2019-11-22
labels metdata doesn't always exist.. why are we even checking this?

mmclanerh:
assert project['metadata']['labels']['name'] == \
KeyError: 'labels'


mmclanerh:
@maorfr  this maybe isn't the best solution here, but i need to unblock e2e from breaking. there are some namespaces that aren't intended to be monitored.

mmclanerh:
actually maybe i have another solution..

mmclanerh:
fwiw: if someone creates a project for a random test, the e2e check will bail out here. probably needs a better solution than my if statement..

PR #371, opened 2019-11-27
[terraform-resource] add ecr support

maorfr:
covers https://jira.coreos.com/browse/APPSRE-1282

This PR adds a new terraform provider: `ecr`.

### DIFF
--- a/utils/terrascript_client.py
+++ b/utils/terrascript_client.py
@@ -813,6 +813,7 @@ class TerrascriptClient(object):
         # iam user policy for bucket
         values = {}
         values['user'] = identifier
+        values['name'] = identifier
         policy = {
             "Version": "2012-10-17",
             "Statement": [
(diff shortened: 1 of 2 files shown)


utils/terrascript_client.py
```diff
@@ -765,6 +768,95 @@ def populate_tf_resource_dynamodb(self, resource, namespace_info):
...
+                    "Sid": "ListImagesInRepository",
+                    "Effect": "Allow",
+                    "Action": ["ecr:ListImages"],
+                    "Resource": "${" + ecr_tf_resource.fullname + ".arn}"
+                },
+                {
+                    "Sid": "GetAuthorizationToken",
+                    "Effect": "Allow",
+                    "Action": ["ecr:GetAuthorizationToken"],
+                    "Resource": "*"
```
tparikh: should the resource really be `all` here?

maorfr: iiuc, you get a token for the registry (one per region), and not per repository. so i think that `all` makes sense...

PR #380, opened 2019-12-05
[terraform-users] add more gpg validations

maorfr:
covers the following cases that were uncaught but were invalid:

https://gitlab.cee.redhat.com/service/app-interface/merge_requests/2279
https://gitlab.cee.redhat.com/service/app-interface/merge_requests/2283

@apahim @mmclanerh please review

### DIFF
--- a/utils/gpg.py
+++ b/utils/gpg.py
@@ -4,6 +4,13 @@ from subprocess import PIPE, Popen, STDOUT
 
 
 def gpg_key_valid(public_gpg_key):
+    if ' ' in public_gpg_key.rstrip():
+        return False
+
+    equal_sign_count = public_gpg_key.count('=')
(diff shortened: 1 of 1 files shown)


utils/gpg.py
```diff
@@ -4,6 +4,13 @@
 
 
 def gpg_key_valid(public_gpg_key):
+    if ' ' in public_gpg_key.rstrip():
+        return False
+
+    equal_sign_count = public_gpg_key.count('=')
+    if not public_gpg_key.rstrip().endswith('=' * equal_sign_count):
```
apahim: `rstrip` is not that expensive, but I'd avoid calling it twice.

maorfr: fixed!

maorfr:
thanks @apahim!

PR #383, opened 2019-12-08
handle gql conflict gracefully

maorfr:
intercept the `409: Conflict` returned from https://github.com/app-sre/qontract-server/blob/master/src/server.ts#L56 and exit gracefully with exit code `3`, which will indicate this conflict.

this will make the output much clearer to our users (think of fluentd).

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -36,6 +36,7 @@ import reconcile.aws_garbage_collector
 import reconcile.aws_iam_keys
 import reconcile.aws_support_cases_sos
 
+from utils.gql import GqlApiError
 from utils.aggregated_list import RunnerException
 from utils.binary import binary
 from utils.environ import environ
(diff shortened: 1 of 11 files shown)


jmelis:
the fluentd sidecar I'm working on will have conflicts with this change. Can we hold this after I submit the fluentd sidecar PR?

maorfr:
sure. i did it so the fluentd setup would be simpler, but as you wish.

reconcile/cli.py
```diff
@@ -164,6 +165,14 @@ def run_integration(func, *args):
     except RunnerException as e:
         sys.stderr.write(str(e) + "\n")
         sys.exit(1)
+    except GqlApiError as e:
+        if '409' in str(e):
```
jmelis: a bit fragile, can we detect the exit code a bit more robustly?

utils/gql.py
```diff
@@ -30,7 +32,11 @@ def __init__(self, url, token=None):
 
     def query(self, query, variables=None):
         try:
-            result_json = self.client.execute(query, variables)
+            # supress print on HTTP error
```
jmelis: why do we want to suppress the message?

reconcile/cli.py
maorfr: agreed to improve the GqlApiError in the future

jmelis: add a `TODO` and let's leave this for the future.

utils/gql.py
maorfr: agreed to keep as is

PR #384, opened 2019-12-08
change log message to include integration name

maorfr:
part of https://jira.coreos.com/browse/APPSRE-1274

Since we intend to ship our logs using fluentd, it would be helpful to know which integration produced which output.

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -166,9 +166,10 @@ def run_integration(func, *args):
         sys.exit(1)
 
 
-def init_log_level(log_level):
+def init_log_level(log_level, invoked_subcommand):
     level = getattr(logging, log_level) if log_level else logging.INFO
-    logging.basicConfig(format='%(levelname)s: %(message)s', level=level)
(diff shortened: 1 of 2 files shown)


reconcile/cli.py
```diff
@@ -166,9 +166,10 @@ def run_integration(func, *args):
         sys.exit(1)
 
 
-def init_log_level(log_level):
+def init_log_level(log_level, invoked_subcommand):
     level = getattr(logging, log_level) if log_level else logging.INFO
-    logging.basicConfig(format='%(levelname)s: %(message)s', level=level)
+    log_format = f'[{invoked_subcommand}] %(levelname)s: %(message)s'
```
apahim: f-strings are Python >= 3.6. Not sure how big of a problem that is for this project, but according to the `setup.py`, we do support previous version of Python.

maorfr: thanks for spotting that! removed the py3 line.

jmelis:
@maorfr I prefer to manage this in fluentd itself, we can change the records and prepend whatever we want. This is an interesting functionality, but can you make it toggleable so we can switch it on or off?

PR #385, opened 2019-12-09
use script to run integration

jmelis:
Instead of embedding the command in the container spec, we are now
calling a specific script that will use the environment variables to run
the integration.

Note that there is some logic borrowed from:
https://github.com/app-sre/qontract-reconcile/pull/383

meaning that that PR needs to be rebased


jmelis:
@maorfr can you please review?

PR #400, opened 2019-12-18
[gitlab-projects] introduce new integration

maorfr:
covers https://issues.redhat.com/browse/APPSRE-1371

completes https://gitlab.cee.redhat.com/service/app-interface/merge_requests/2485

This PR adds a new integration that creates projects in gitlab on on request.

### DIFF
--- a/utils/gitlab_api.py
+++ b/utils/gitlab_api.py
@@ -200,6 +200,8 @@ class GitLabApi(object):
 
     def get_project_maintainers(self, repo_url):
         project = self.get_project(repo_url)
+        if project is None:
+            return None
         members = project.members.all(all=True)
         return [m.username for m in members if m.access_level >= 40]
 
@@ -226,6 +228,8 @@ class GitLabApi(object):
 
     def add_project_member(self, repo_url, user, access="maintainer"):
         project = self.get_project(repo_url)
+        if project is None:
+            return
         access_level = self.get_access_level(access)
         try:
             project.members.create({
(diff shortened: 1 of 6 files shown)


utils/gitlab_api.py
```diff
@@ -293,22 +297,42 @@ def get_access_level(access):
         elif access == "guest":
             return gitlab.GUEST_ACCESS
 
+    def get_group_id_and_projects(self, group_name):
+        groups = self.gl.groups.list()
+        group = [g for g in groups if g.path == group_name]
```
apahim: Do you ever expect more than one group here? If not, you should break the loop when you find your group.

utils/gitlab_api.py
```diff
@@ -293,22 +297,42 @@ def get_access_level(access):
...
+        if len(group) != 1:
```
apahim: Then `if not len(group):` should be enough.

utils/gitlab_api.py
maorfr: i do not. in a single gitlab instance there can only be one group with a given name.

utils/gitlab_api.py
```diff
@@ -293,22 +297,42 @@ def get_access_level(access):
...
+            logging.error(group_name + " group not found")
```
apahim: Otherwise (if more than one group can be in the `group` variable), this message is misleading. But I don't think you will even have more then one group there anyway, so don't bother.

utils/gitlab_api.py
maorfr: see previous comment

utils/gitlab_api.py
maorfr: same here :)

utils/gitlab_api.py
```diff
@@ -293,22 +297,42 @@ def get_access_level(access):
...
+            return []
+        [group] = group
```
apahim: Now you're just removing the list... maybe it's not worth it to use the comprehension in the first place...

maorfr: is there a better/cleaner way to get a single element from a list based on a condition?

apahim: My point is that maybe you should not have a list in the first place. Why to create a list of one group?

apahim: It seems like the fact you hace a list is just a consequence of using the list comprehension on line 302. A regular for would avoid that.

maorfr: as discussed, we'll keep the current implementation for now

PR #401, opened 2019-12-18
annotate secrets with path and version

tparikh:

### DIFF
--- a/reconcile/openshift_resources.py
+++ b/reconcile/openshift_resources.py
@@ -354,6 +354,8 @@ def fetch_openshift_resource(resource, parent):
         labels = {} if rl is None else json.loads(rl)
         ra = resource['annotations']
         annotations = {} if ra is None else json.loads(ra)
+        annotations['vault.secret.path'] = path
+        annotations['vault.secret.version'] = version
         rt = resource['type']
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -5,7 +5,7 @@
 .tox
 .pytest_cache
 .eggs
-
+.vscode/*
```
maorfr: this should go in your global git ignore so you don't have to repeat it in every project

reconcile/openshift_resources.py
```diff
@@ -354,6 +354,8 @@ def fetch_openshift_resource(resource, parent):
         labels = {} if rl is None else json.loads(rl)
         ra = resource['annotations']
         annotations = {} if ra is None else json.loads(ra)
+        annotations['vault.secret.path'] = '' if path is None else path
+        annotations['vault.secret.version'] = '' if version is None else version
```
maorfr: line too long

tparikh:
@maorfr I did run it locally but this can wait until you have verified as this will update all existing secrets with these annotations. Please merge after you have verified it tomorrow.

maorfr:
i ran it locally, and it seems that these annotations were not actually applied.

PR #402, opened 2019-12-19
Quay mirroring

apahim:


apahim:
@maorfr @jmelis @aditya-konarde

reconcile/quay_mirror.py
```diff
@@ -0,0 +1,107 @@
...
+
+        state = {}
+
+        for app in result['apps']:
+            quay_repos = app.get('quayRepos')
+
+            if quay_repos is None:
+                continue
+
+            for quay_repo in quay_repos:
```
maorfr: i usually use `for quay_repo in quay_repos or []` to avoid checking for None explicitly

reconcile/quay_mirror.py
```diff
@@ -0,0 +1,107 @@
...
+                org = quay_repo['org']['name']
+                for item in quay_repo['items']:
+                    mirror = item['mirror']
+                    if mirror is None:
+                        continue
+
+                    name = item['name']
+                    state[org] = (f'{mirror}',
+                                  f'quay.io/{org}/{name}')
```
maorfr: will this work as intended if there is more then 1 repo to mirror?

reconcile/quay_mirror.py
```diff
@@ -0,0 +1,107 @@
...
+        return state
+
+    def _get_push_creds(self):
```
maorfr: very nice! good re-use of VaultSecret

maorfr:
all in all, this looks very nice @apahim.
i like the fact that you created a class for the integration, that would make it very reusable.

aside for the small comments, i'd like this integration to log what is about to happen before it happens, and i'd like those prints to be the same in dry run and not in dry run.

example expected output:
```
INFO: ['copy_tag', '<source_repo>', '<destination_repo>', '<tag>']
```
very nice work and a very quick turnaround. thanks!

reconcile/quay_mirror.py
apahim: Why to avoid checking for `None`? I mean, your suggestion is way more expensive (like ~50% more expesive) and, for my eyes, less readable. Mind if I keep it as is?

reconcile/quay_mirror.py
apahim: Oops... good catch. Fixing.

apahim:
@maorfr thank you! I will rework the logs as requested.

maorfr:
@apahim you will also need to add the new integration to the helm chart and `make generate` to add it to the openshift template. This will deploy the integration to stage after we merge.

maorfr:
Memory reqs will be interesting...

apahim:
It sits right below 60MB the during whole copy process. A bit of overhead to keep it safe would give us around 80MB requests, 100MB limits? 
quay_mirror_mem

utils/skopeo.py
```diff
@@ -0,0 +1,239 @@
...
+        :param image: The image to be inspected.
+        :type image: str
+        :return: The JSON-loaded skopeo inspection information.
+        :rtype: dict
+        """
+        inspect_result = self._run_skopeo('inspect',
+                                          self._get_image_url(image))
+        try:
+            return json.loads(inspect_result)
+        except json.decoder.JSONDecodeError:
```
jmelis: if the json is invalid do we really want to be silent about it? I would let it raise an unhandled exception I think?

utils/skopeo.py
```diff
@@ -0,0 +1,239 @@
...
+                safe_cmd.append(':'.join([item.split(':')[0], '******']))
+            else:
+                safe_cmd.append(item)
+        self.log.info('Running "%s"', ' '.join(safe_cmd))
+
+        if self.dry_run:
+            self.log.info(' DRY-RUN: Nothing really executed.')
+            return ''
+
+        result = subprocess.run(cmd, capture_output=True)
```
jmelis: I didn't know about capture_output, it's really convenient!

utils/skopeo.py
```diff
@@ -0,0 +1,239 @@
...
+class Skopeo:
+    """Wrapper around the skopeo utility."""
+
+    default_scheme = 'docker://'
+    default_repo = ''
+    default_tag = 'latest'
+
+    def __init__(self, dry_run=False):
+        self.dry_run = dry_run
+        self.log = logging.getLogger(__name__)
```
jmelis: why? we just use the global logger elsewhere in the project, should we change this?

utils/skopeo.py
```diff
@@ -0,0 +1,239 @@
...
+             'tag': 'latest'}
+
+        :param image: The image url to be parsed.
+        :type image: str
+        :return: A data structure with all the parsed components of
+                 the image URL, already filled with the defaults for
+                 those not provided.
+        :rtype: dict
+        """
+        image_url_struct = re.search(
```
jmelis: would it be possible to leverage https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse here?

selftests/utils/test_skopeo.py
```diff
@@ -0,0 +1,49 @@
+import pytest
```
jmelis: I :heart: that you wrote tests, so needed in this project!

utils/skopeo.py
apahim: The way this is implemented is:

- We have the quay destination, e.g. quay.io/app-sre/fedora
- We have the source image, e.g. docker.io/fedora:28
- We inspect docker.io/fedora:28, retrieving all the tags, copiyng them all.

The reason for this inspect to silently fail is that, when that happens, I still try to copy from docker.io/fedora:28 to quay.io/app-sre/fedora directly, without handling tags. Does that make sense?

utils/skopeo.py
apahim: I'd to avoid that. Using the root logger, `%(name)s` will always be `root`, regerdless the module you log the message from.

utils/skopeo.py
jmelis: ok, sounds good

utils/skopeo.py
jmelis: makes sense

utils/skopeo.py
apahim: There are some issues with using `urlparse` for this use case that I would have to work around:
- It requires the scheme to be provided, so I'd have to pre-process the url:
```
>>> urlparse('registry.example.com/apahim/fedora')
ParseResult(scheme='', netloc='', path='registry.example.com/apahim/fedora', params='', query='', fragment='')
```
- It does not parse the tag, when it exists, requiring some post-processing:
```
>>> urlparse('socker://registry.example.com/apahim/fedora:31')
ParseResult(scheme='socker', netloc='registry.example.com', path='/apahim/fedora:31', params='', query='', fragment='')
```
- On top of that, after parsed, you are not able to overwrite the attributes, so you have to create another data structure out of it so you can manipulate the values.

With all that, I think the regex gives us a very straight forward implementation.

jmelis: :+1: thanks for looking into it

apahim:
@maorfr, @jmelis, how does these logs look:
```
$ qontract-reconcile --config config.toml --dry-run quay-mirror
INFO: ['inspect', 'docker://docker.io/tnozicka/openshift-acme:latest']
INFO: ['copy', 'docker://docker.io/tnozicka/openshift-acme:latest', 'docker://quay.io/app-sre/openshift-acme:latest']
INFO: ['copy', 'docker://docker.io/tnozicka/openshift-acme:v0.8.0-pre-alpha', 'docker://quay.io/app-sre/openshift-acme:v0.8.0-pre-alpha']
INFO: ['copy', 'docker://docker.io/tnozicka/openshift-acme:v0.8.0', 'docker://quay.io/app-sre/openshift-acme:v0.8.0']
INFO: ['inspect', 'docker://docker.io/vault:latest']
INFO: ['copy', 'docker://docker.io/vault:0.10.0-beta1', 'docker://quay.io/app-sre/vault:0.10.0-beta1']
INFO: ['copy', 'docker://docker.io/vault:0.10.0-rc1', 'docker://quay.io/app-sre/vault:0.10.0-rc1']
INFO: ['copy', 'docker://docker.io/vault:0.10.0', 'docker://quay.io/app-sre/vault:0.10.0']
INFO: ['copy', 'docker://docker.io/vault:0.10.1', 'docker://quay.io/app-sre/vault:0.10.1']
...
```

reconcile/quay_mirror.py
maorfr: don't mind at all! just consulting on things we've done in the past.

maorfr:
this output looks good to me. i'll remove the `inspect` log though, or make its level `debug` (assuming that all relevant tags were copied already and this integration is running again, we'd want the output to be empty).

reconcile/cli.py
```diff
@@ -427,6 +428,13 @@ def quay_membership(ctx):
     run_integration(reconcile.quay_membership.run, ctx.obj['dry_run'])
 
 
+@integration.command()
+@click.pass_context
+@binary(['skopeo'])
```
maorfr: :heart:

PR #403, opened 2019-12-19
Added sentry integration

rrati:


utils/sentry_client.py
```diff
@@ -0,0 +1,155 @@
...
+class SentryClient:
+  ORGANIZATION = "sentry"
+
+  def __init__(self, host, token):
+    self.host = host
+    self.auth_token = token
+
+  def do_sentry_api_call(self, method, path, slugs, payload=None):
+    success_codes = {"get": 200, "post": 201, "delete": 204, "put": 200}
+    url = "%s/api/0/%s/" % (self.host, path)
```
jmelis: we tend to prefer the f-strings, or at `format` over the `%` formatting, which is very py2. I realize this is not in the pystyle guide by @apahim. Do you agree?

utils/sentry_client.py
```diff
@@ -0,0 +1,155 @@
...
+import json
+
+class SentryClient:
+  ORGANIZATION = "sentry"
+
+  def __init__(self, host, token):
+    self.host = host
+    self.auth_token = token
+
+  def do_sentry_api_call(self, method, path, slugs, payload=None):
```
jmelis: I like to prepend private methods with `_`

utils/sentry_client.py
```diff
@@ -0,0 +1,155 @@
...
+    success_codes = {"get": 200, "post": 201, "delete": 204, "put": 200}
+    url = "%s/api/0/%s/" % (self.host, path)
+
+    if len(slugs) > 0:
+      url = "%s%s/" % (url, "/".join(slugs))
+
+    headers = {"Authorization": "Bearer " + self.auth_token}
+    call = getattr(requests, method, None)
+    if call == None:
+      print ("invalid http method %s" % method)
```
jmelis: perhaps we should raise an exception here? In any case we try to never do `print`, I would suggest `logging.error` in this case, but I personally prefer raising an exception.

utils/sentry_client.py
```diff
@@ -0,0 +1,155 @@
...
+      return None
+
+    response = call(url, headers=headers, json=payload)
+    if response.status_code != success_codes[method]:
```
jmelis: there's a `response.raise_for_status()` call that we usually use here, to catch errors

reconcile/sentry_config.py
```diff
@@ -0,0 +1,317 @@
...
+  secret["path"] = path
+  secret["field"] = field
+  skip_users = secret_reader.read(secret, settings)
+
+  sentry_client = SentryClient(host, token)
+  current_state = fetch_current_state(sentry_client, [skip_users])
+  desired_state = fetch_desired_state()
+
+  reconciler = SentryReconciler(sentry_client, dry_run)
+  reconciler.reconcile(current_state, desired_state)
```
jmelis: missing end of line

utils/sentry_client.py
```diff
@@ -0,0 +1,155 @@
...
+    request = self.do_sentry_api_call("get", "projects", [])
+    return request
+
+  def get_project(self, slug):
+    request = self.do_sentry_api_call("get", "projects", [self.ORGANIZATION, slug])
+    return request
+
+  def create_project(self, team_slug, name, slug=None):
+    params = {"name": name}
+    if slug != None:
```
jmelis: I think it's better to do: `if slug is not None`

utils/sentry_client.py
```diff
@@ -0,0 +1,155 @@
...
+      print ("error: http %s returned %d trying to access %s" % (method, response.status_code, url))
+      return None
+
+    if response.status_code != 204:
+      return response.json()
+
+
+  # Organization functions
+  def get_organizations(self):
+    request = self.do_sentry_api_call("get", "organizations", [])
```
jmelis: nitpick: `request` is deceiving here, it would be better to use `response` or `response_json`

reconcile/sentry_config.py
```diff
@@ -0,0 +1,317 @@
...
+            config[field] = project_config[field]
+        team_projects.append(config)
+      projects[team] = team_projects
+  state.init_projects_from_desired_state(projects)
+  return state
+
+def run(dry_run=False):
+  settings = queries.get_app_interface_settings()
+  secret = {
+    "path": path,
```
jmelis: hmm, what is this `path` variable? I don't see it defined anywhere.

reconcile/sentry_config.py
```diff
@@ -0,0 +1,317 @@
...
+    "field": field,
+  }
+  host = secret_reader.read(secret, settings)
+  secret["field"] = field
+  token = secret_reader.read(secret, settings)
+
+  secret["path"] = path
+  secret["field"] = field
+  skip_users = secret_reader.read(secret, settings)
```
jmelis: I don't understand this block very much, from L300-L310. Can you add comments to explain what is it that you are doing here?

reconcile/sentry_config.py
```diff
@@ -0,0 +1,317 @@
...
+class SentryState:
+  def __init__(self):
+    # Map of user:teams[]
+    self.users = {}
+    # List of team names
+    self.teams = []
+    # Map of team:projects_config[]
+    self.projects = {}
+  
+  def init_users_from_current_state(self, users):
```
jmelis: I would rename this `init_users`

reconcile/sentry_config.py
```diff
@@ -0,0 +1,317 @@
...
+        if "resolveAge" in p.keys():
+          pdata["auto_resolve_age"] = p["resolveAge"]
+        if "allowedDomains" in p.keys():
+          pdata["allowed_domains"] = p["allowedDomains"]
+        if team not in self.projects.keys():
+          self.projects[team] = [pdata]
+        else:
+          self.projects[team].append(pdata)
+
+  def init_projects_from_desired_state(self, projects):
```
jmelis: I would rename this `init_projects`.

reconcile/sentry_config.py
```diff
@@ -0,0 +1,317 @@
...
+          "platform": p["platform"]
+        }
+        if "sensitiveFields" in p.keys():
+          pdata["sensitive_fields"] = p["sensitiveFields"]
+        if "safeFields" in p.keys():
+          pdata["safe_fields"] = p["safeFields"]
+        if "resolveAge" in p.keys():
+          pdata["auto_resolve_age"] = p["resolveAge"]
+        if "allowedDomains" in p.keys():
+          pdata["allowed_domains"] = p["allowedDomains"]
```
jmelis: super minor comment, but there's an opportunity here to restructure and refactor this with a for loop and a single if

reconcile/sentry_config.py
```diff
@@ -0,0 +1,317 @@
...
+  return False
+
+def fetch_current_state(client, ignore_users):
+  state = SentryState()
+
+  # Retrieve all the teams
+  sentry_teams = client.get_teams()
+  teams = []
+  for team in sentry_teams:
+    teams.append(team["slug"])
```
jmelis: this is good opportunity for a list comprehension:

```python
teams = [team['slug'] for team in sentry_teams]
```

reconcile/sentry_config.py
```diff
@@ -0,0 +1,317 @@
...
+  state.init_teams(teams)
+
+  # Retrieve the projects and the teams associated with them
+  sentry_projects = client.get_projects()
+  projects = {}
+  for sentry_project in sentry_projects:
+    project_slug = sentry_project["slug"]
+    if project_slug == "internal":
```
jmelis: nice catch

reconcile/sentry_config.py
```diff
@@ -0,0 +1,317 @@
...
+{
+  teams: sentry_teams_v1 {
+    name
+  }
+}
+"""
+
+class SentryState:
+  def __init__(self):
+    # Map of user:teams[]
```
jmelis: I *really* like that you give the spec of the content of the data. I wish we would do this always! cc @maorfr @apahim

reconcile/sentry_config.py
```diff
@@ -0,0 +1,317 @@
...
+  def init_teams(self, teams):
+    self.teams = teams
+
+class SentryReconciler:
+  def __init__(self, client, dry_run):
+    self.client = client
+    self.dry_run = dry_run
+
+  def reconcile(self, current, desired):
+    # Reconcile the teams frist
```
jmelis: typo :nerd_face:

reconcile/sentry_config.py
```diff
@@ -0,0 +1,317 @@
...
+    for team in current.teams:
+      if team not in desired.teams:
+        logging.info("deleting sentry team %s" % team)
```
jmelis: we generally log actions in the form of lists like this:

```python
logging.info(["delete_team", team])
```

That way we could parse it in the future if we ever want to.

This is a related comment of a review @maorfr did for @apahim:
https://github.com/app-sre/qontract-reconcile/pull/402#issuecomment-567616372

utils/sentry_client.py
apahim: That should go into the style guide for sure :)

rrati: No problems using f-strings.  Style guide should probably call out using upper or lower case f as well as single or double quoted strings. :)

utils/sentry_client.py
rrati: Yep.  Client still needs a bit of work for better error handing and style changes like this.  Done.

utils/sentry_client.py
rrati: Done

reconcile/sentry_config.py
rrati: Done

reconcile/sentry_config.py
rrati: Fixed

reconcile/sentry_config.py
rrati: Changed all logging to be in this format

utils/sentry_client.py
rrati: Agreed, exception is really the path to take here

reconcile/sentry_config.py
rrati: It doesn't appear all the files end with a blank line.  I don't care either way, just want consistency. :)

reconcile/sentry_config.py
rrati: Done

reconcile/sentry_config.py
rrati: Done

reconcile/sentry_config.py
rrati: Reworked

utils/sentry_client.py
rrati: Renamed

utils/sentry_client.py
rrati: Done

reconcile/sentry_config.py
```diff
@@ -0,0 +1,356 @@
...
+      field
+      format
+      version
+    }
+  }
+}
+"""
+
+class SentryState:
+  def __init__(self):
```
apahim: Indentation has to be a multiple of 4.

apahim: There are many others flake8 errors...

rrati: Done

utils/sentry_client.py
```diff
@@ -0,0 +1,151 @@
...
+        user_teams.remove(t)
+    params = {"teams": user_teams}
+    response = self._do_sentry_api_call_("put", "organizations", [self.ORGANIZATION, "members", user["id"]], payload=params)
+    return response
+
+  def change_user_role(self, email, role):
+    user = self.get_user(email)
+    params = {"role": role}
+    response = self._do_sentry_api_call_("put", "organizations", [self.ORGANIZATION, "members", user["id"]], payload=params)
+    return response
```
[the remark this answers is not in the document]
rrati: Done

utils/sentry_client.py
```diff
@@ -0,0 +1,151 @@
...
+import json
+
+class SentryClient:
+  ORGANIZATION = "sentry"
+
+  def __init__(self, host, token):
+    self.host = host
+    self.auth_token = token
+
+  def _do_sentry_api_call_(self, method, path, slugs, payload=None):
```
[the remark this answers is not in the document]
rrati: Just makes it symmetrically clear that the function is intended for internal only.  I prefer the symmetric approach as it's more similar to other languages, but not sure if there is a python equivalent style

reconcile/sentry_config.py
rrati: Not a fan of this requirement though. :)

apahim: That's PEP8: https://www.python.org/dev/peps/pep-0008/#indentation
We don't have to follow, but we would need a reason for that... and to change it everywhere else :)

utils/sentry_client.py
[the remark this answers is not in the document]
apahim: The only case where a trailing underscore is reccomended is to avoid collisions with reserved keywords. And, in python, it looks off in any case, but that's due to the dunders, I guess.

PR #418, opened 2019-12-26
[email-sender] introduce new integration

maorfr:
This PR introduces a new integration that allows us to send emails through app-interface.
Matching app-interface PR: https://gitlab.cee.redhat.com/service/app-interface/merge_requests/2531

The biggest challenge was: how do we only send each email only once.
The proposed solution: we use S3 as a simple state for app-interface.

This approach will allow us to implement new integrations that do not have a source of truth that we can fetch current state from. For example, this will allow running SQL queries only once (RDS self service).

### DIFF
--- a/reconcile/email_sender.py
+++ b/reconcile/email_sender.py
@@ -1,7 +1,6 @@
 import sys
 import logging
 
-# import utils.gql as gql
 import utils.smtp_client as smtp_client
 import reconcile.queries as queries
 
@@ -26,8 +25,8 @@ def collect_to(to):
 
     aliases = to.get('aliases')
     if aliases:
+        # TODO: implement this
         for alias in aliases:
-            # gql = gql.get_api()
             if alias == 'all-users':
                 pass
             elif alias == 'all-service-owners':
@@ -45,22 +44,22 @@ def collect_to(to):
             for service_owner in service_owners:
                 audience.add(service_owner['email'])
 
-    clusters - to.get('clusters')
+    clusters = to.get('clusters')
     if clusters:
+        # TODO: implement this
         for cluster in clusters:
-            # TODO: implement this
             pass
 
     namespaces = to.get('namespaces')
     if namespaces:
+        # TODO: implement this
         for namespace in namespaces:
-            # TODO: implement this
             pass
 
     aws_accounts = to.get('aws_accounts')
     if aws_accounts:
+        # TODO: implement this
         for account in aws_accounts:
-            # TODO: implement this
             pass
 
     roles = to.get('roles')
--- a/reconcile/email_sender.py
+++ b/reconcile/email_sender.py
@@ -110,9 +110,8 @@ def run(dry_run=False):
         logging.info(['send_email', email['name'], email['subject']])
 
         if not dry_run:
-            state.add(email['name'])
-
             names = collect_to(email['to'])
             subject = email['subject']
             body = email['body']
             smtp_client.send_mail(names, subject, body, settings=settings)
+            state.add(email['name'])
--- a/reconcile/email_sender.py
+++ b/reconcile/email_sender.py
@@ -12,13 +12,13 @@ QONTRACT_INTEGRATION = 'email-sender'
 
 def collect_to(to):
     """Collect audience to send email to from to object
-    
+
     Arguments:
         to {dict} -- AppInterfaceEmailAudience_v1 object
-    
+
     Raises:
         AttributeError: Unknown alias
-    
+
     Returns:
         set -- Audience to send email to
     """
--- a/reconcile/email_sender.py
+++ b/reconcile/email_sender.py
@@ -24,37 +24,59 @@ def collect_to(to):
     """
     audience = set()
 
-    for alias in to.get('aliases') or []:
-        # gql = gql.get_api()
-        if alias == 'all-users':
+    aliases = to.get('aliases')
+    if aliases:
+        for alias in aliases:
+            # gql = gql.get_api()
+            if alias == 'all-users':
+                pass
+            elif alias == 'all-service-owners':
+                pass
+            else:
+                raise AttributeError(f"unknown alias: {alias}")
+
+    services = to.get('services')
+    if services:
+        for service in services:
+            service_owners = service.get('serviceOwners')
+            if not service_owners:
+                continue
+
+            for service_owner in service_owners:
+                audience.add(service_owner['email'])
+
+    clusters - to.get('clusters')
+    if clusters:
+        for cluster in clusters:
+            # TODO: implement this
             pass
-        elif alias == 'all-service-owners':
+
+    namespaces = to.get('namespaces')
+    if namespaces:
+        for namespace in namespaces:
+            # TODO: implement this
             pass
-        else:
-            raise AttributeError(f"unknown alias: {alias}")
 
-    for service in to.get('services') or []:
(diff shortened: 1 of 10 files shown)


utils/state.py
```diff
@@ -0,0 +1,64 @@
...
+        aws_api = AWSApi(1, accounts, settings=settings)
+        session = aws_api.get_session(account)
+
+        self.client = session.client('s3')
+
+    def exists(self, key):
+        """checks if a key exists in the state
+
+        Arguments:
+            key {string} -- key to check
```
apahim: Please use the same docstring format everywhere. Ideally, the format you're using in the class docstring, which is the one described in our style guide.

utils/state.py
```diff
@@ -0,0 +1,64 @@
...
+                Bucket=self.bucket, Key=f"{self.state_path}/{key}")
+            return True
+        except ClientError:
+            return False
+
+    def add(self, key):
+        """adds a key to the state and fails if the key already exists
+
+        Arguments:
+            key {string} -- key to add
```
apahim: Ditto

utils/state.py
```diff
@@ -0,0 +1,64 @@
...
+
+        Raises:
+            KeyError: key already exists in the state
+        """
+        if self.exists(key):
+            raise KeyError(
+                f"[state] key {key} already exists in {self.state_path}")
+        self.client.put_object(
+            Bucket=self.bucket, Key=f"{self.state_path}/{key}")
```
apahim: I like this class. It's clean and direct to the point. I thinking... how about using `__setitem__` and `__getitem__` to implement the functionality here? Anyway, it looks good as is.

utils/smtp_client.py
```diff
@@ -78,9 +78,14 @@ def send_mail(name, subject, body, settings=None):
...
-    to = '{}@{}'.format(name, _mail_address)
     msg['From'] = formataddr((from_name, _username))
-    msg['To'] = to
+    to = set()
+    for name in names:
+        if '@' in name:
+            to.add(name)
+        else:
+            to.add(f"{name}@{_mail_address}")
+    msg['To'] = ', '.join(to)
```
apahim: ok

apahim: What's the differentece now between `send_mail` and `send_mails`? Do we still need `send_mails` after this?

reconcile/queries.py
```diff
@@ -22,6 +22,48 @@ def get_app_interface_settings():
...
+    body
+  }
+}
+"""
+
+
+def get_app_interface_emails():
+    """ Returns Email resources defined in app-interface """
+    gqlapi = gql.get_api()
+    return gqlapi.query(APP_INTERFACE_EMAILS_QUERY)['emails']
```
apahim: It's probably good... app-interface counterpart still not ready, so I'm not able to verify this.

reconcile/email_sender.py
```diff
@@ -0,0 +1,96 @@
...
+
+    Raises:
+        AttributeError: Unknown alias
+
+    Returns:
+        set -- Audience to send email to
+    """
+    audience = set()
+
+    for alias in to.get('aliases') or []:
```
apahim: that `or []` hurts

reconcile/email_sender.py
```diff
@@ -0,0 +1,96 @@
...
+    for email in emails_to_send:
+        logging.info(['send_email', email['name'], email['subject']])
+
+        if not dry_run:
+            state.add(email['name'])
+
+            names = collect_to(email['to'])
+            subject = email['subject']
+            body = email['body']
+            smtp_client.send_mail(names, subject, body, settings=settings)
```
apahim: bad things might happen here... what about calling the state.add(email['name']) only after this line?

apahim:
Very nice. A couple of comments, but overwall I like it a lot.

utils/state.py
apahim: So, this is what I have in mind:
```python

class Status:
    def __init__(self):
        """
        Status
        """

    def __getitem__(self, item):
        """
        Returns the value of an object:
            client.get_object()['Body'].read()

        If the object body is a JSON serialized content (which I think it's
        a good idea), this returns:
            json.loads(client.get_object()['Body'].read())

        Raises KeyError when the object does not exist.

        Usage:

            foo = status['foo']
        """

    def __setitem__(self, key, value):
        """
        Stores an object. I'd use:
            put_object(..., Body=json.dumps(value))

        Usage:

            status['foo'] = None
        """

    def __delitem__(self, key):
        """
        Deletes an object.

        Raises KeyError when the object does not exist.

        Usage:

            del(status['foo'])
        """

    def __contains__(self, item):
        """
        Checks if an object exists (client.head_object()).

        Usage:

            if 'foo' in status:
                # do something

        Returns True/False.
        """

    def __len__(self):
        """
        Returns the number of objects. Probably the best approach
        is to use boto3.resource.Bucket.objects.all()

        Usage:

            len(status)
        """

    def __iter__(self):
        """
        Creates a generator that yields all the keys. Probably the
        best approach is to use boto3.resource.Bucket.objects.all()

        Usage:

            for key in status:
                print(key)
        """

    def keys(self):
        """
        Returns list(self), which takes advantage of __iter__.

        Usage:

            for key in status.keys():
                print(key)

        """

    def values(self):
        """
        Returns [self[key] for key in self.keys()], which takes advantage
        of __getitem__ and self.keys().

        Usage:

            for value in status.values():
                print(value)
        """

    def items(self):
        """
        For key in self.keys(), this yields (key, self[key]).

        Usage:

            for key, value in status.items():
                print(key, value)
        """
```

utils/state.py
maorfr: done for ocm.py as well!

utils/state.py
maorfr: done!

utils/state.py
maorfr: I totally agree with this suggestion and the proposed solution.
that said, we currently only maintain a state of happened/did not happen. this will be the first feature we do when we have an integration that needs state with content. i've added an issue for it: #419.

I'll keep this class as is, as your first comment says that as well :)

utils/smtp_client.py
maorfr: `send_mail` sends a single mail to (now) multiple recipients.
`send_mails` sends multiple mails to (now) multiple recipients.

so still both needed. The reason we want to send multiple mails is because the `send_mails` function is used by the `terraform-users` integration, which sends email invitations to our AWS accounts, and those mails contain a GPG encrypted password for that specific user. so these mails must be sent separately. hope that makes sense!

reconcile/queries.py
maorfr: It's good! :)

reconcile/email_sender.py
maorfr: fixed!

reconcile/email_sender.py
maorfr: that makes sense. i did that intentionally to avoid sending the same email over and over in case we get an error but the mail is sent anyway (not sure if that can happen, but still).

i moved the state.add to the correct place as you suggested.

utils/state.py
apahim: Sure, thank you! I didn't mean to implement all the interfaces... just those needed at this point, which are the `add()` and the `__contains__`. Thak you for creating the issue.

PR #422, opened 2019-12-29
[app-interface-reporter] add email notification

maorfr:
covers https://issues.redhat.com/browse/APPSRE-1081

This PR adds an email notification of App SRE monthly reports.
It creates an email to be picked up by email-sender (#418), which will be sent to `all-service-owners`.
Tthis PR also adds the `all-service-owners` alias implementation.

Everything in this PR is negotiable, as this is an implementation suggestion.

### DIFF
--- a/tools/app_interface_reporter.py
+++ b/tools/app_interface_reporter.py
@@ -442,7 +442,22 @@ def main(configfile, dry_run, log_level, gitlab_project_id, reports_path):
                 f.write(report['content'])
 
     if not dry_run:
+        email_schema = '/app-interface/app-interface-email-1.yml'
+        email_body = """Hello,
+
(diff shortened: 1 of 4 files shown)


tools/app_interface_reporter.py
```diff
@@ -442,7 +442,22 @@ def main(configfile, dry_run, log_level, gitlab_project_id, reports_path):
                 f.write(report['content'])
 
     if not dry_run:
+        email_schema = '/app-interface/app-interface-email-1.yml'
+        email_body = """Hello,
```
jmelis: nitpick, perhaps use `textwrap.dedent` here?

maorfr: done

PR #423, opened 2019-12-29
Introducing qontract-cli

maorfr:
This PR adds a new tool called `qontract-cli`. As it grows on us, a `q` alias is suggested.

This cli tool is currently really good at getting information, but doesn't do anything else.

Update: the tool now has 2 subcommands -get and state.

Try it:
```
$ q get users
$ q get -o json cluster app-sre
$ q state ls
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,6 +39,7 @@ setup(
         "pyOpenSSL>=19.0.0,<20.0.0",
         "ruamel.yaml>=0.16.5,<0.17.0",
         "terrascript>=0.6.1,<0.7.0",
+        "tabulate>=0.8.6,<0.9.0"
     ],
 
     test_suite="tests",
--- a/setup.py
+++ b/setup.py
@@ -54,6 +54,7 @@ setup(
             'qontract-reconcile = reconcile.cli:integration',
             'e2e-tests = e2e_tests.cli:test',
             'app-interface-reporter = tools.app_interface_reporter:main',
+            'qontract-cli = tools.qontract_cli:root',
         ],
     },
 )
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -8,6 +8,8 @@ import reconcile.queries as queries
 
 from tabulate import tabulate
 
+from utils.state import State
+from utils.environ import environ
 from reconcile.cli import config_file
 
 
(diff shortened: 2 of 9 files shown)


setup.py
```diff
@@ -54,6 +54,7 @@
             'qontract-reconcile = reconcile.cli:integration',
             'e2e-tests = e2e_tests.cli:test',
             'app-interface-reporter = tools.app_interface_reporter:main',
+            'qontract-cli = tools.qontract_cli:root',
```
apahim: Have you considered making this a sub-command of `qontract-reconcile`? The naming `qontract-cli` seems quite generic. One can think it has functionalities from all qcontract-* projects. I mean, I'm totally in for having a generic 
`qontract-cli` with functionalities from all qcontract-* projects, but this that you're proposing is just exposing some queries from the reconcile...

tools/qontract_cli.py
```diff
@@ -0,0 +1,72 @@
...
+@click.pass_context
+def get(ctx):
+    pass
+
+
+@get.command()
+@click.pass_context
+def settings(ctx):
+    settings = queries.get_app_interface_settings()
+    print(json.dumps(settings))
```
apahim: Can we use `logging` instead? For now, `print` is enough, but I see this growing and by then we will want `logging` being used everywhere.

tools/qontract_cli.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@get.command()
+@click.pass_context
+def aws_accounts(ctx):
+    accounts = queries.get_aws_accounts()
+    print(json.dumps(accounts))
```
apahim: Why json? Will the output be used as input somewhere?

apahim: By the way, the `| jq` can be avoided by using `json.dumps(..., indent=2)`. Without colors nonetheless.

maorfr: `-o table|json|yaml|`! :)

tools/qontract_cli.py
maorfr: logging will not be picked up by the next tool, whatever that may be

setup.py
maorfr: as discussed, we don't know where this initiative is headed, so we'll keep it simple and hidden for the time being.

tools/qontract_cli.py
apahim: how so? logging can spit to stdout, just like print, as long as you set the streamhandler. Anyway, as we discussed, let's keep the effort to the minimum for now.

utils/state.py
```diff
@@ -48,6 +48,14 @@ def exists(self, key):
         except ClientError:
             return False
 
+    def ls(self):
```
apahim: `__iter__`? so one can do `for key in state:`

utils/state.py
```diff
@@ -61,3 +69,17 @@ def add(self, key):
                 f"[state] key {key} already exists in {self.state_path}")
         self.client.put_object(
             Bucket=self.bucket, Key=f"{self.state_path}/{key}")
+
+    def rm(self, key):
```
apahim: `__delitem__`? so on can do `del(state[key])`

utils/state.py
maorfr: the state object represents a remote state. i'd like this class to emphasize that the state is remote in every way possible.

`state.ls()` to me shows that some actions will be carried out here. `for key in state` feels so in-memory. maybe it's just me.

utils/state.py
maorfr: i chose not to use `__setitem__` because (currently) the state only stores keys, without values. that's why using `state.add()` makes more sense to me, rather then `state[key] = 'whatever'`.

for this reason for me it feels less intuitive to use `state[key]` at this point. the dictionary feeling does not represent reality as of yet.

utils/state.py
apahim: I see. Never thought of `__iter__` as an interface for in-memory content. Actualy I see it more as the opposite... it typically creates a generator so the items are retrieved one by one, from wherever they are. Eg `f = open('foo.txt')`, the `for line in f` will give you one line at a time from the file. Anyway, aways fun to think about interfaces :)

utils/state.py
apahim: Ok, got it.

utils/state.py
maorfr: __iter__ itself, yes.
but when using state without looking in the class, `for key in state`?

apahim: yep... I see just a loop over the state elements, wherever they are coming from. But I get it, we see it differently. No worries. Too bad that those interfaces got worn out by python's own data structures, now it's hard to think of them differently.
To be clear, I'm ok with the `ls()`.

PR #432, opened 2020-01-04
[quay-mirror] Avoiding constant deep (slow) image sync

apahim:
The deep sync - where we compare each image's tag manifest - is expensive
and time consuming. Also, there's no need to run it all the time.

For new tags in the mirror images, the tag will still be sync'ed right away,
but checking tag's manifests will now happen every ~8h.

That means that tags that are constantly updated, like 'latest', 'master',
'build', ... will face a delay to be freshly available in quay.io

### DIFF
--- a/reconcile/quay_mirror.py
+++ b/reconcile/quay_mirror.py
@@ -1,4 +1,7 @@
 import logging
+import os
+import tempfile
+import time
 
 from collections import defaultdict
 
@@ -80,6 +83,9 @@ class QuayMirror:
         return summary
 
     def process_sync_tasks(self):
+        eight_hours = 28800  # 60 * 60 * 8
+        is_deep_sync = self._is_deep_sync(interval=eight_hours)
+
         summary = self.process_repos_query()
 
         sync_tasks = defaultdict(list)
@@ -104,6 +110,12 @@ class QuayMirror:
                                    downstream, upstream)
                         continue
 
+                    # Deep (slow) check only from time to time
+                    if not is_deep_sync:
+                        _LOG.debug('Image %s and mirror %s are in sync',
+                                   downstream, upstream)
+                        continue
+
                     if downstream == upstream:
                         _LOG.debug('Image %s and mirror %s are in sync',
                                    downstream, upstream)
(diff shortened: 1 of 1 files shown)


maorfr:
thanks for taking the time to do this!

I believe that the reconcile/ directory should hold high level instructions, and not get to the depths of opening/writing to files. The only integration where this "rule" is broken is in jira-watcher, which was meant to be a temporary integration from the minute it was created (and is no longer active).

I think that this idea is good, but maybe this is a place to expand the State class to be able to manage state on local file system or in s3.

to be brutally honest, i'm not sure if the approach suggested in this PR would work. assuming a full deep sync, even if it happens once every 8 hours - we will get rate limited as far as i understand. i think that a deep sync should be spread across time and across iterations of the integration.

Here is an alternate suggestion that requires no state and will produce similar results:
```
# Deep (slow) check only 1 per 100 iterations in average
if int(random() * 100) != 42:
    _LOG.debug(
        'Image %s and mirror %s are probably in sync',
        downstream, upstream)
    continue
```

wdyt?

apahim:
No, at the moment, we are not being rate limited. Current state is:

- Fast check takes 40s. New tags are copied after the fast check. Dry-run runs only the fast check.
- Slow check takes 50min (now that all imaages are synced). Updated tags are copied after the slow check. No issues with connections rate limit.

Yes, I'm not happy myself about having this file read/write here. I like the idea of using the Status object for this, if you agree we can use it to store context (key and value).

But the baseline for this approach is: I see value in being deterministic. Using the random approach, there's no way to say when or how often a given tag will be synced again. That lack of determinism might be challenging when it comes to managing users expectations. It will probably lead us to manually sync images from time to time, under users requests.

On the other hand, the deterministic approach gives us a minimum and maximun time users are garanteed to have the tag synced, so all they have to do is wait a known maximum period.

Even though we have no issues so far with running all the deep syncs in the same interation, we can still spread the sync over different iterations with some level of randomness, but I would then use a radom sync interval within a limit, per image tag basis. The state class could also be used for that.

Well, those are just my thougts and they are surely up for challenge. If you don't believe determinism to be a feature, we can go for the random implementation. Just let me know.

maorfr:
totally agree with everything you said. determinism is important. although if we want to set expectations, we can probably say "new tags are copied almost immediately, updated tags are updated after ~24 hours".

i have no objection enhancing the State to include content, but i'm not sure it will be good to use it in this integration, as we have the state in Quay.

PR #437, opened 2020-01-08
Validate project resolve age is valid

rrati:


utils/sentry_client.py
```diff
@@ -89,6 +97,20 @@ def optional_project_fields(self):
...
+        valid_resolve_age = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 15, 18, 21,
+                             24, 30, 36, 48, 72, 96, 120, 144, 168, 192, 216,
+                             240, 288, 312, 336, 360, 384, 408, 432, 456, 480,
+                             504, 528, 552, 576, 600, 624, 648, 672, 696, 720]
+
+        optional_fields = self.optional_project_fields()
+        resolve_age_field = optional_fields["resolveAge"]
+        if resolve_age_field in options.keys() and \
+           options[resolve_age_field] not in valid_resolve_age:
+            raise Exception(
```
apahim: Could you please pick a more specific exception (https://docs.python.org/3/library/exceptions.html) or a custom one?

utils/sentry_client.py
```diff
@@ -71,7 +71,15 @@ def update_project(self, slug, options):
             "subjectPrefix": options["email_prefix"]
         }
 
-        for k, v in self.optional_project_fields().items():
+        try:
+            self.validate_project_options(options)
+        except Exception as e:
```
apahim: this try/except block doesn't do anything.

utils/sentry_client.py
```diff
@@ -89,6 +97,20 @@ def optional_project_fields(self):
...
+    def validate_project_options(self, options):
+        # If the resolve age is not one of these then sentry will set to 0
+        valid_resolve_age = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 15, 18, 21,
+                             24, 30, 36, 48, 72, 96, 120, 144, 168, 192, 216,
+                             240, 288, 312, 336, 360, 384, 408, 432, 456, 480,
+                             504, 528, 552, 576, 600, 624, 648, 672, 696, 720]
+
+        optional_fields = self.optional_project_fields()
+        resolve_age_field = optional_fields["resolveAge"]
+        if resolve_age_field in options.keys() and \
```
apahim: nitpick: you can remove the `.keys()` call... `if resolve_age_field in options` is half the price.

utils/sentry_client.py
```diff
@@ -89,6 +97,20 @@ def optional_project_fields(self):
         }
         return optional_fields
 
+    def validate_project_options(self, options):
+        # If the resolve age is not one of these then sentry will set to 0
+        valid_resolve_age = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 15, 18, 21,
+                             24, 30, 36, 48, 72, 96, 120, 144, 168, 192, 216,
+                             240, 288, 312, 336, 360, 384, 408, 432, 456, 480,
+                             504, 528, 552, 576, 600, 624, 648, 672, 696, 720]
```
apahim: Where did you get these numbers from? Maybe a comment on that to make the maintenance easier in the future?

PR #454, opened 2020-01-21
Parametrize integrations container resources

jfchevrette:
This adds template parameters for all integration containers.

The goal is to then be able to tweak as needed in the saas repo as we grow and as we need more resources

### DIFF
--- a/openshift/qontract-reconcile.yaml
+++ b/openshift/qontract-reconcile.yaml
@@ -37,11 +37,11 @@ objects:
             value: ${SLEEP_DURATION_SECS}
           resources:
             limits:
-              cpu: 50m
-              memory: 400Mi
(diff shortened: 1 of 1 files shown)


openshift/qontract-reconcile.yaml
```diff
@@ -37,11 +37,11 @@ objects:
             value: ${SLEEP_DURATION_SECS}
           resources:
             limits:
-              cpu: 50m
-              memory: 400Mi
+              cpu: ${QR_AWS_GARBAGE_COLLECTOR_CPU_LIMIT}
```
apahim: Those numbers are coming from helm... https://github.com/app-sre/qontract-reconcile/blob/master/helm/qontract-reconcile/values.yaml

jfchevrette: /facepalm
closing this then...

PR #460, opened 2020-01-24
[sentry-config] use org email instead of github email

maorfr:
This PR introduces a breaking change to the sentry-config integration - using a potentially different email for all existing users.

The org email is probably better to use.
The only possible limitation is that the used email has to be identical to the one specified in GitHub, which I believe is not the case.

This PR also removes `bots` from the query, so bot users are no longer supported.

Another optimization is that we check that the team is a part of the same sentry instance before calculating the users.

### DIFF
--- a/reconcile/sentry_config.py
+++ b/reconcile/sentry_config.py
@@ -3,7 +3,7 @@ import reconcile.queries as queries
 import utils.gql as gql
 import utils.secret_reader as secret_reader
 
-from reconcile.github_users import init_github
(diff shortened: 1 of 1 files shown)


rrati:
#459 makes a lot of changes in this area, including introducing a cache for users.  The email for the user that is created with must match the primary e-mail in github.  I don't think this will work.

maorfr:
Can we add users without emails at all?

PR #470, opened 2020-01-28
Adding support for SQL Queries

apahim:


reconcile/queries.py
```diff
@@ -338,3 +340,73 @@ def get_users():
...
+    }
+}
+"""
+
+
+def get_resource(path):
+    gqlapi = gql.get_api()
+    raw_values = gqlapi.query(RESOURCE_QUERY,
+                              {'path': path})['resources'][0]
+    return anymarkup.parse(raw_values['content'], force_types=None)
```
maorfr: why is this needed here?
https://github.com/app-sre/qontract-reconcile/blob/master/utils/gql.py#L56-L79

reconcile/sql_query.py
```diff
@@ -0,0 +1,228 @@
...
+    for tf_resource in tf_resources:
+        if tf_resource['identifier'] != identifier:
+            continue
+        if tf_resource['provider'] != 'rds':
+            continue
+
+        defaults = queries.get_resource(tf_resource['defaults'])
+        return {
+            'cluster': namespace['cluster']['name'],
+            'output_resource_name': tf_resource['output_resource_name'],
```
maorfr: `output_resource_name` is not always defined.
we should either copy the logic here, it's fairly simple:
https://github.com/app-sre/qontract-reconcile/blob/master/utils/terrascript_client.py#L1059-L1062

reconcile/sql_query.py
```diff
@@ -0,0 +1,228 @@
...
+            # building up the final query dictionary
+            {
+                'name': name,
+                'namespace': namespace,
+                'identifier': sql_query['identifier'],
+                'db_conn': db_conn,
+                'query': sql_query['query'],
+                **tf_resource_info,
+            }
+        )
```
maorfr: beautiful!

reconcile/sql_query.py
```diff
@@ -0,0 +1,228 @@
...
+
+    queries_list = collect_queries()
+    remove_candidates = []
+    for query in queries_list:
+        query_name = query['name']
+
+        # Checking the sql-query state:
+        # - No state: up for execution.
+        # - State is not None: executed and up for removal after the JOB_TTL.
+        # - State is None: executed and removed.
```
maorfr: how about "DONE"?

reconcile/sql_query.py
```diff
@@ -0,0 +1,228 @@
...
+            if query_state is not None:
+                remove_candidates.append({'name': query_name,
+                                          'timestamp': query_state})
+            continue
+        except KeyError:
+            pass
+
+        job_yaml = process_template(query)
+        job = yaml.safe_load(job_yaml)
+        job_resource = OpenshiftResource(job, QONTRACT_INTEGRATION, '0.0.1')
```
maorfr: can we extract the `0.0.1`?

reconcile/queries.py
apahim: Fixed!

reconcile/sql_query.py
apahim: oh, ok, thanks! fixed.

reconcile/sql_query.py
apahim: :)

reconcile/sql_query.py
apahim: sure

reconcile/sql_query.py
apahim: yes, right

PR #483, opened 2020-02-03
[openshift-serviceaccount-tokens] introduce new integration

maorfr:
covers https://issues.redhat.com/browse/APPSRE-929

depends on https://gitlab.cee.redhat.com/service/app-interface/merge_requests/3006

this PR adds a new integration that takes ServiceAccount tokens from a namespace and places them in a Secret in a different namespace (could be a different cluster as well) to be consumed by other applications.

the output can also go to Vault to be consumed in templating.

### DIFF
--- a/openshift/qontract-reconcile.yaml
+++ b/openshift/qontract-reconcile.yaml
@@ -1748,6 +1748,50 @@ objects:
   metadata:
     labels:
       app: qontract-reconcile
+    name: qontract-reconcile-openshift-serviceaccount-tokens
+  spec:
+    replicas: 1
+    selector:
+      matchLabels:
+        app: qontract-reconcile
+    template:
+      metadata:
+        labels:
+          app: qontract-reconcile
+      spec:
+        securityContext:
+          runAsUser: ${{USER_ID}}
+        containers:
+        - name: int
+          image: ${IMAGE}:${IMAGE_TAG}
+          env:
+          - name: DRY_RUN
+            value: ${DRY_RUN}
+          - name: INTEGRATION_NAME
+            value: openshift-serviceaccount-tokens
+          - name: INTEGRATION_EXTRA_ARGS
+            value: "--vault-output-path app-sre/integrations-output"
+          - name: SLEEP_DURATION_SECS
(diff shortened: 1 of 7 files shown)


openshift/qontract-reconcile.yaml
```diff
@@ -1743,6 +1743,50 @@ objects:
...
+    selector:
+      matchLabels:
+        app: qontract-reconcile
+    template:
+      metadata:
+        labels:
+          app: qontract-reconcile
+      spec:
+        securityContext:
+          runAsUser: ${{USER_ID}}
```
aditya-konarde: Why do we need to template this? i.e. why don't we just let it run with a random UID?

maorfr: the openshift integrations are (sometimes) using ssh, which requires a valid user.
this is not a part of this PR though, it's the default templating.

i know it sucks.

reconcile/openshift_serviceaccount_tokens.py
```diff
@@ -0,0 +1,91 @@
...
+              error_details=name)
+
+
+def fetch_desired_state(namespaces, ri, oc_map):
+    for namespace_info in namespaces:
+        if not namespace_info.get('openshiftServiceAccountTokens'):
+            continue
+        namespace_name = namespace_info['name']
+        cluster_name = namespace_info['cluster']['name']
+        for sat in namespace_info['openshiftServiceAccountTokens']:
```
aditya-konarde: nit: can we please call this `token` for better readability?

aditya-konarde: Instead of `sat`

aditya-konarde: Nvm, I see that its a rather filled up object and not just the token

maorfr: this is not the `token` per se:
https://gitlab.cee.redhat.com/service/app-interface/merge_requests/3006/diffs#ea2a513d0ff90c2171e9c823a58e7f634be9beaf_81_83

PR #487, opened 2020-02-04
[openshift-resources] enable pods recycle

maorfr:
This PR will enable pods recycle in the openshift-resources integration.

If openshift-resources applies a Secret used by a pod, that pod will be recycled (deleted + validated to have come back to life).

WDYT?

### DIFF
--- a/reconcile/openshift_resources.py
+++ b/reconcile/openshift_resources.py
@@ -497,7 +497,7 @@ def run(dry_run=False, thread_pool_size=10, internal=None, defer=None):
     ob.check_unused_resource_types(ri)
 
     ob.realize_data(dry_run, oc_map, ri,
-                    recycle_pods=False)
+                    recycle_pods=True)
 
     if ri.has_error_registered():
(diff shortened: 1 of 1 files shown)


aditya-konarde:
I'm overall +1 for this, just have a few implementation concerns.

- Does the secret get mounted as a volume, or is the application using it as an env? Secrets mounted as volumes are mostly okay, because the application has a watch on the file and just reloads as needed.
- How does this work with applications that are using AWS creds? Do the pods get recycled twice?
- In terms of service resiliency, are we at a point where we can randomly delete a pod with a config change and expect it to come back up fine? In case of a rollout, we should make sure this rollout is happening with an n+1 replica to avoid causing service degradation. 

There's a bunch of unanswered questions that we'll have to figure out as we go. If you want to give this a try, I'd suggest that we somehow only enable it for stage and let it "soak" for a while before we push it to prod.

maorfr:
ok, taking a different approach here. thanks!

PR #488, opened 2020-02-04
Secret recycle annotation

maorfr:
This PR adds an ability to specify an annotation on a Secret - `qontract.recycle`.
When this annotation is set to `true`, pods which use this Secret will be recycled.

The annotation will be added to all terraform-resources Secrets to keep recycling pods as before, when `recycle_pods` was passed as `True`.
In order to avoid recycling all pods just by adding this annotation, the recycle_pods call is currently commented out and will be uncommented after all Secrets are updated.

The annotation can be defined on app-interface resources to specify that pods using that Secret should be recycled.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -172,8 +172,7 @@ def fetch_current_state(namespaces=None,
     return ri, oc_map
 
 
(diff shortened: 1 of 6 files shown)


apahim:
LGTM. Very nice approach!

maorfr:
thank you both!

PR #490, opened 2020-02-04
Remove threading when populating resources in terraform-resources

rrati:


rrati:
@maorfr PTAL.  We could probably refine the level of locking if needed, but this allows to run --print-only and threaded

maorfr:
since `populate_tf_resources` is not very time consuming (only doing internal calculations) - we can just not run it threaded at all, there is no real benefit there IMO.

wdyt?

apahim:
+1 for no threads. If you lock the entire method that runs on a thread, you don't need threads.

rrati:
I don't care either way.  I'm not sure threading is needed at all, but this is really just proof of concept.  If we wanted threading we could work to fine tune the locking more.  My earlier attempts showed it might be difficult to do.  I'm also fine nuking threads

PR #496, opened 2020-02-09
[openshift-saas-deploy] introduce new integration

maorfr:
covers https://issues.redhat.com/browse/APPSRE-1276

this PR adds a new (WIP) integration called openshift-saas-deploy, which replaces:
- saasherder
- saas repos
- jenkins jobs (except for image build)

this integration already includes:
- filter resources by kind (similar to saasherder's `--filter`)
- resources apply & delete (replaces the `saasherder label` mechanism)

this integration intends to include (perhaps not at the first iteration):
- manifest-bouncer checks

still very much WIP.
example app-interface MR: https://gitlab.cee.redhat.com/service/app-interface/merge_requests/3111


utils/saasherder.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class SaasHerder():
+    """Wrapper around SaaS deployment actions."""
+
+    def __init__(self, saas_files,
+                 github=None,
+                 gitlab=None,
+                 integration=None,
+                 integration_version=None):
```
apahim: Those parameters are not really optional since the defaults will break the code

utils/saasherder.py
```diff
@@ -0,0 +1,120 @@
...
+        self.saas_files = saas_files
+        self.github = github
+        self.gitlab = gitlab
+        self.integration = integration
+        self.integration_version = integration_version
+        self._collect_namespaces()
```
apahim: mind initializing everything here? I mean,
```
        self.namespaces = self._collect_namespaces()
```

utils/saasherder.py
maorfr: github and gitlab may be optional as we move forward, since we will divide this integration into internal and external executions.

regarding the rest, i agree, but would like to postpone the fix for a future iteration if you are ok with that.

utils/saasherder.py
maorfr: fixed!

utils/saasherder.py
apahim: sure thing

PR #502, opened 2020-02-11
[openshift-*] compare resources recursively by desired item

maorfr:
based on a suggestion by the glorious @jmelis:
https://gist.github.com/jmelis/594ca4fc7566159c4efbce8e80c4bc3f

compare resources only by the desired resource and the attributes it has. in case default values are added by k8s, they are ignored in this comparison.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -232,6 +232,15 @@ def realize_data(dry_run, oc_map, ri,
                     ).format(cluster, namespace, resource_type, name)
                     logging.info(msg)
 
+                # don't apply if resources match
+                elif d_item == c_item:
+                    msg = (
+                        "[{}/{}] resource '{}/{}' present "
+                        "and matches desired, skipping."
+                    ).format(cluster, namespace, resource_type, name)
+                    logging.debug(msg)
+                    continue
+
                 # don't apply if sha256sum hashes match
                 elif c_item.sha256sum() == d_item.sha256sum():
                     if c_item.has_valid_sha256sum():
--- a/utils/openshift_resource.py
+++ b/utils/openshift_resource.py
@@ -28,6 +28,48 @@ class OpenshiftResource(object):
         self.error_details = error_details
         self.verify_valid_k8s_object()
 
+    def __eq__(self, other):
+        return self.obj_intersect_equal(self.body, other.body)
+
+    def obj_intersect_equal(self, obj1, obj2):
+        if obj1.__class__ != obj2.__class__:
+            return False
+        elif isinstance(obj1, dict):
+            for obj1_k, obj1_v in obj1.items():
+                obj2_v = obj2.get(obj1_k, None)
+                if not obj2_v:
+                    return False
+                if obj1_k == 'cpu':
+                    equal = self.cpu_equal(obj1_v, obj2_v)
+                    if not equal:
+                        return False
+                elif not self.obj_intersect_equal(obj1_v, obj2_v):
+                    return False
+        elif isinstance(obj1, list):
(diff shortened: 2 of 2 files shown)


reconcile/openshift_base.py
```diff
@@ -232,6 +232,15 @@ def realize_data(dry_run, oc_map, ri,
                     ).format(cluster, namespace, resource_type, name)
                     logging.info(msg)
 
+                # don't apply if resources match
+                elif d_item == c_item:
```
apahim: great start!

utils/openshift_resource.py
```diff
@@ -28,6 +28,48 @@ def __init__(self, body, integration, integration_version,
...
+    def __eq__(self, other):
+        return self.obj_intersect_equal(self.body, other.body)
+
+    def obj_intersect_equal(self, obj1, obj2):
+        if obj1.__class__ != obj2.__class__:
+            return False
+        elif isinstance(obj1, dict):
+            for obj1_k, obj1_v in obj1.items():
+                obj2_v = obj2.get(obj1_k, None)
+                if not obj2_v:
```
apahim: `if obj2_v is not None` (so you don't enter the `if` when `obj2` is `0` or `''` )

utils/openshift_resource.py
```diff
@@ -28,6 +28,48 @@ def __init__(self, body, integration, integration_version,
         self.error_details = error_details
         self.verify_valid_k8s_object()
 
+    def __eq__(self, other):
+        return self.obj_intersect_equal(self.body, other.body)
+
+    def obj_intersect_equal(self, obj1, obj2):
+        if obj1.__class__ != obj2.__class__:
+            return False
+        elif isinstance(obj1, dict):
```
apahim: just `if` her... no need for elif.

utils/openshift_resource.py
```diff
@@ -28,6 +28,48 @@ def __init__(self, body, integration, integration_version,
...
+                obj2_v = obj2.get(obj1_k, None)
+                if not obj2_v:
+                    return False
+                if obj1_k == 'cpu':
+                    equal = self.cpu_equal(obj1_v, obj2_v)
+                    if not equal:
+                        return False
+                elif not self.obj_intersect_equal(obj1_v, obj2_v):
+                    return False
+        elif isinstance(obj1, list):
```
apahim: same here... just `if`

utils/openshift_resource.py
```diff
@@ -28,6 +28,48 @@ def __init__(self, body, integration, integration_version,
...
+            if len(obj1) != len(obj2):
+                return False
+            for i in range(len(obj1)):
+                if not self.obj_intersect_equal(obj1[i], obj2[i]):
+                    return False
+        else:
```
apahim: it's safe to remove the else

reconcile/openshift_base.py
maorfr: thanks!

utils/openshift_resource.py
maorfr: thanks!
look at what i did. does it makes sense?

utils/openshift_resource.py
maorfr: thanks!

utils/openshift_resource.py
maorfr: thanks!

utils/openshift_resource.py
maorfr: thanks!

utils/openshift_resource.py
apahim: yep, looks great. very easy to read.

PR #506, opened 2020-02-12
github-users tweaks

apahim:


maorfr:
-1 for logs, as this integration will repeat the same output endlessly

apahim:
Oh, ok. I thought it had just missed the train while disabled.

PR #507, opened 2020-02-14
Catch exception in populate_oc_resources

tparikh:
Reconciling openshift-resources breaks if one of the cluster's token has expired.
Also he `-config` flag for OC client has been deprecated and `kubeconfig` is the new flag to use.

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -138,8 +138,8 @@ def populate_oc_resources(spec, ri):
     except StatusCodeError as e:
         msg = 'cluster: {},'
         msg += 'namespace: {},'
-        msg += 'resource: {}'
-        msg += ' exception: {}'
+        msg += 'resource: {},'
+        msg += 'exception: {}'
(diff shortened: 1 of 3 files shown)


reconcile/terraform_resources.py
```diff
@@ -115,18 +117,31 @@
...
+                                  QONTRACT_INTEGRATION_VERSION)
+          ri.add_current(
+              spec.cluster,
+              spec.namespace,
+              spec.resource,
+              openshift_resource.name,
+              openshift_resource
+          )
+    except StatusCodeError as e:
+      logging.error("[populate_oc_resources] :"+str(e))
```
rrati: I would say we don't care about what the error is.  If there's an error, just print it and the cluster it is associated with.  Looking fo specific string in an error message is inherently fragile, and in this case I really don't think we care.  We just want to know what cluster had an error, right?

tparikh: We may want to handle errors depending on what they are. Some may require us to stop processing and other may allow us to continue

PR #511, opened 2020-02-16
[openshift-saas-deploy] check images

maorfr:
Part of https://issues.redhat.com/browse/APPSRE-1276

This PR adds image existence check in saas-deploy before deploying them.

Currently very raw, but as we move forward, a failure in this check for one image should not fail the process and should allow the deployment of other resources.

### DIFF
--- a/utils/saasherder.py
+++ b/utils/saasherder.py
@@ -2,7 +2,7 @@ import yaml
 import json
 
 from utils.oc import OC
-from utils.skopeo import Skopeo
+from utils.container import Image
 from utils.openshift_resource import OpenshiftResource as OR
 
 
@@ -98,10 +98,9 @@ class SaasHerder():
         return images
 
     def _check_images(self, resources):
-        skopeo = Skopeo()
         images = self._collect_images(resources)
         for image in images:
-            skopeo.inspect(image)
+            bool(Image(image))
 
     def populate_desired_state(self, ri):
         for saas_file in self.saas_files:
--- a/utils/saasherder.py
+++ b/utils/saasherder.py
@@ -2,6 +2,7 @@ import yaml
 import json
 
 from utils.oc import OC
+from utils.skopeo import Skopeo
 from utils.openshift_resource import OpenshiftResource as OR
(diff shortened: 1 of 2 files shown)


utils/saasherder.py
```diff
@@ -85,6 +86,23 @@ def _process_template(self, url, path, hash_length, target, parameters):
...
+            except KeyError:
+                pass
+
+        return images
+
+    def _check_images(self, resources):
+        skopeo = Skopeo()
+        images = self._collect_images(resources)
+        for image in images:
+            skopeo.inspect(image)
```
apahim: `bool(container.Image(image))` is way cheaper...

apahim: Also it's more resilient.

maorfr: thanks! and it supports authentication, perfect.

maorfr:
@apahim i left the skopeo changes included, let me know if you prefer that i remove them,

utils/saasherder.py
```diff
@@ -85,6 +86,22 @@ def _process_template(self, url, path, hash_length, target, parameters):
...
+                    images.add(c["image"])
+            except KeyError:
+                pass
+
+        return images
+
+    def _check_images(self, resources):
+        images = self._collect_images(resources)
+        for image in images:
+            bool(Image(image))
```
apahim: This will have no effect... what do you want to do when an image does not exist?

apahim:
Sure, the skopeo inspect is useful and should be kept indeed :)

utils/saasherder.py
```diff
@@ -100,7 +101,8 @@ def _collect_images(self, resources):
     def _check_images(self, resources):
         images = self._collect_images(resources)
         for image in images:
-            bool(Image(image))
+            if not bool(Image(image)):
```
apahim: `if not Image(image):`, in that case :)

maorfr: i'm so distracted

apahim:
lgtm

PR #516, opened 2020-02-17
Change action for s3-cloudfront from *Object to GetObject

rrati:


rrati:
@maorfr Any concerns with this change?

PR #522, opened 2020-02-19
WIP: POC Add affinity rules for spreading pods accross availability zones and …

rrati:
…nodes

### DIFF
--- a/utils/openshift_resource.py
+++ b/utils/openshift_resource.py
@@ -159,6 +159,16 @@ class OpenshiftResource(object):
         now = datetime.datetime.utcnow().replace(microsecond=0).isoformat()
         annotations['qontract.update'] = now
 
+        # create labels if not present
+        body['metadata'].setdefault('labels', {})
+        labels = body['metadata']['labels']
+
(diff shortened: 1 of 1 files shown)


rrati:
Might also want to add a rule with topologyKey failure-domain.beta.kubernetes.io/region if we want to also attempt to ensure pods across regions.  Would we have a single cluster with nodes across regions?

maorfr:
I like this idea a lot!

We are currently deploying our apps via saasherder, so this change will currently not do anything.
After we start using [openshift-saas-deploy](https://issues.redhat.com/browse/APPSRE-1276) this change would totally make sense IMO.

I'd like to get more opinions, and would also like to request that we merge this only after we have enough confidence in openshift-saas-deploy.

thanks @rrati, this is a really good idea!

rrati:
This could be implemented in saasherder really easily as well.  Whether it's worth it or not might depend on when we plan to deprecate saasherder (if at all) and when we want this functionality available.

maorfr:
cleaning the queue a bit, apologies!

PR #528, opened 2020-02-20
WiP: Adding statement timeout for postgres

apahim:


apahim:
@tparikh is 30s a good number?

PR #530, opened 2020-02-21
openshift-saas-deploy to use Github token

apahim:


maorfr:
I know I did it in the past, but I don't like it :)

Totally good enough for now and to enable github-mirror authentication!

apahim:
Sorry, what's it you don't like?

maorfr:
sorry, this:
```
token = github_config['app-sre']['token']
```

using the `app-sre` token explicitly...

apahim:
Got it. It's actually not working. Let me do it the right way.

PR #533, opened 2020-02-23
[openshift] fix infinite reconcile

maorfr:


apahim:
Right, my bad. I somehow considered that entering one of those ifs was a one-way road. Thanks for the fix.

PR #534, opened 2020-02-23
[openshift-saas-deploy] improvements #1

maorfr:
- check image per resource and skip applying if image does not exist
- deleting resources is enabled only if no errors had occurred
- removed default `None` values from SaasHerder init method

### DIFF
--- a/utils/saasherder.py
+++ b/utils/saasherder.py
@@ -12,10 +12,10 @@ class SaasHerder():
     """Wrapper around SaaS deployment actions."""
 
     def __init__(self, saas_files,
-                 github=None,
-                 gitlab=None,
-                 integration=None,
-                 integration_version=None):
+                 github,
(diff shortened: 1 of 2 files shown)


utils/saasherder.py
```diff
@@ -120,13 +122,16 @@ def populate_desired_state(self, ri):
...
                                                target, parameters)
-                    # check images
-                    self._check_images(resources)
                     # add desired resources
                     for resource in resources:
                         resource_kind = resource['kind']
                         if resource_kind not in managed_resource_types:
                             continue
+                        # check images
+                        image_error = self._check_images(resource)
```
apahim: A bit confusing that that `check_images()` returns `True` when there's a problem. Anyway, the fact that you're putting that into a variable called `image_error` mitigates that confusion here.

PR #542, opened 2020-02-26
Add replication configuration to s3 in terraform-resources

rrati:


utils/terrascript_client.py
```diff
@@ -435,17 +436,126 @@ def populate_tf_resource_s3(self, resource, namespace_info):
...
+        sc = common_values.get('storage_class')
+        if sc:
+            rule = {
+                "id": sc + "_storage_class",
+                "enabled": "true",
+                "transition": {
+                    "storage_class": sc
+                }
+            }
+            if values.get('lifecycle_rule'):
```
maorfr: `common_values` already contains `lifecycle_rules`. is that a duplication?

maorfr: and can there be a lifecycle_rule without a storage_class?

utils/terrascript_client.py
```diff
@@ -435,17 +436,126 @@ def populate_tf_resource_s3(self, resource, namespace_info):
...
+                values['lifecycle_rule'].append(rule)
+            else:
+                values['lifecycle_rule'] = rule
         if common_values.get('cors_rules'):
             # common_values['cors_rules'] is a list of cors_rules
             values['cors_rule'] = common_values['cors_rules']
+        if common_values.get('replication_configurations'):
+            rc_configs = []
+            for config in common_values['replication_configurations']:
```
maorfr: can we reduce to a single access to the dict? just to be consistent with the above:
```
sc = common_values.get('storage_class')
        if sc:
```

utils/terrascript_client.py
```diff
@@ -435,17 +436,126 @@ def populate_tf_resource_s3(self, resource, namespace_info):
...
+                tf_resources.append(tf_resource)
+
+                # iam role policy attachment
+                # Terraform resource reference:
+                # https://www.terraform.io/docs/providers/aws/r/iam_policy_attachment.html
+                rc_values.clear()
+                rc_values['role'] = "${aws_iam_role." + id + ".name}"
+                rc_values['policy_arn'] = "${aws_iam_policy." + id + ".arn}"
+                tf_resource = aws_iam_role_policy_attachment(id, **rc_values)
+                tf_resources.append(tf_resource)
```
maorfr: nicely done!

are there dependencies between these resources? if so, we can use `depends_on`

utils/terrascript_client.py
```diff
@@ -435,17 +436,126 @@ def populate_tf_resource_s3(self, resource, namespace_info):
...
+
+                rc_values.clear()
```
maorfr: can you comment this section with what it does?

utils/terrascript_client.py
```diff
@@ -864,18 +974,18 @@ def populate_tf_resource_ecr(self, resource, namespace_info):
...
+                        "ecr:GetDownloadUrlForLayer",
+                        "ecr:GetRepositoryPolicy",
+                        "ecr:DescribeRepositories",
+                        "ecr:ListImages",
+                        "ecr:DescribeImages",
+                        "ecr:BatchGetImage",
+                        "ecr:InitiateLayerUpload",
+                        "ecr:UploadLayerPart",
+                        "ecr:CompleteLayerUpload",
+                        "ecr:PutImage"
```
maorfr: :heart:

utils/terrascript_client.py
```diff
@@ -949,7 +1059,7 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
                 'origin_access_identity':
                     'origin-access-identity/cloudfront/' +
                     '${' + cf_oai_tf_resource.fullname + '.id}'
-                }
+                    }
```
maorfr: :heart:

reconcile/terraform_resources.py
```diff
@@ -211,7 +211,7 @@ def write_outputs_to_vault(vault_path, ri):
     for cluster, namespace, _, data in ri:
         for name, d_item in data['desired'].items():
             secret_path = \
-              f"{vault_path}/{integration_name}/{cluster}/{namespace}/{name}"
+                f"{vault_path}/{integration_name}/{cluster}/{namespace}/{name}"
```
maorfr: :heart:

utils/terrascript_client.py
```diff
@@ -435,17 +436,126 @@ def populate_tf_resource_s3(self, resource, namespace_info):
         # https://www.terraform.io/docs/providers/aws/r/s3_bucket.html
         values = {}
         values['bucket'] = identifier
-        values['versioning'] = {'enabled': True}
+        values['versioning'] = common_values.get('versioning') or False
```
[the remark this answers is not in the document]
rrati: Currently versioning is not enabled for s3 buckets, and enabling it complicates storage.  Every object is kept track of even if deleted, so you have to make sure to have a process that will delete all versions of an object in order to delete it completely.  There are policies we could use that might help, but imo I think we want to default to false for now and tackle changing all buckets to being versioned in a different pr

utils/terrascript_client.py
rrati: I have simplified the means to define a storage class for an s3 bucket done to just a single entry, but in fact it requires a whole lifecycle policy most of which the user doesn't need to know about.  This logic is handling appending the lifecycle rule for storage class if there are or are not already other lifecycle policies defined.

This could probably be removed if we push all the lifecycle definition to the config file.  That would have the benefit of making this simpler and support the few options this implementation isn't supporting, but shifts the knowledge of what is needed to enable a storage class to the user defining the config.

I'm not sure which one I prefer.

utils/terrascript_client.py
rrati: Done

utils/terrascript_client.py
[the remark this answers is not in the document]
maorfr: ack. thanks!

utils/terrascript_client.py
rrati: Done

utils/terrascript_client.py
rrati: Added deps

PR #550, opened 2020-03-03
Added availability zone support to rds instances

rrati:


utils/terrascript_client.py
```diff
@@ -345,6 +345,11 @@ def populate_tf_resource_rds(self, resource, namespace_info,
                                            output_resource_name)
         values['password'] = password
 
+        az = values.get('availability_zone')
+        if az is not None:
+            values['availabliliy_zone'] = az
+            values['provider'] = 'aws.' + az[:-1]
```
maorfr: where is this being set on the providers?

PR #555, opened 2020-03-07
Take OWNERS_ALIASES into account

rporres:
Since we have no repository context, only one OWNERS_ALIASES is allowed.
There seems to be different standards for this file although the most
commonly found is to have just one in the root of the repository. We
will follow this approach.

See https://issues.redhat.com/browse/APPSRE-1484

### DIFF
--- a/reconcile/slack_usergroups.py
+++ b/reconcile/slack_usergroups.py
@@ -60,6 +60,8 @@ ROLES_QUERY = """
         gitlab_owners
         channels
         description
+        github_owners_aliases
+        gitlab_owners_aliases
       }
     }
   }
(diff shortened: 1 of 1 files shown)


rporres:
cc @maorfr

apahim:
We might want to consider refactoring this integration to use the new [RepoOwners](https://github.com/app-sre/qontract-reconcile/blob/master/utils/repo_owners.py#L10) class. It just takes an instance of a git client that implements two specific methods (`get_file()` and `get_repository_tree()`) and gives you all the owners of that repository (`RepoOwners.get_owners()`, respecting the `OWNERS_ALIASES` file.

maorfr:
LGTM

i agree with Amador about the possible future refactor, but let's ship this first.

PR #557, opened 2020-03-08
remove gitlabOwners until schema is updated

maorfr:
this causes some failures.
let's restore this when the schema is updated.

cc @apahim


apahim:
Tks... I was about to merge the schema wen I was requested to change the property name.

PR #561, opened 2020-03-09
Don't set db.user or db.password in terraform-resources if no username is set in the configuration

rrati:


rrati:
Would there ever be a case where we would want password set if username is blank?  I don't think so....

PR #562, opened 2020-03-09
[terraform-vpc-peerings] introduce new integration

maorfr:
this PR adds an integration that creates VPC peering between OSDv4 cluster VPCs and AWS accounts managed in app-interface.

This is currently a manual step done during cluster provisioning.

covers https://issues.redhat.com/browse/APPSRE-1586

matching app-interface MR: https://gitlab.cee.redhat.com/service/app-interface/merge_requests/3503

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -199,7 +199,7 @@ CLUSTERS_QUERY = """
           uid
           terraformUsername
         }
-        id
+        vpc_id
         cidr_block
         region
       }
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -191,6 +191,19 @@ CLUSTERS_QUERY = """
       service
       pod
     }
+    peering {
+      vpc_id
+      connections {
+        account {
+          name
+          uid
+          terraformUsername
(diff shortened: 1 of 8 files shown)


reconcile/queries.py
```diff
@@ -191,6 +191,19 @@ def get_aws_accounts():
       service
       pod
     }
+    peering {
+      vpc_id
+      connections {
```
rrati: Add a connection name to give the peering connection?  I know you generate one, but it's vpc IDs which may not be easy to recognize.  If not a name, then maybe prefix?

maorfr: i think that in general we may want to support an array here in the future.

i agree that we need a connection name. please allow me to save it for a future PR.

reconcile/queries.py
```diff
@@ -191,6 +191,19 @@ def get_aws_accounts():
...
+        account {
+          name
+          uid
+          terraformUsername
+        }
+        id
```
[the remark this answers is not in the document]
maorfr: done!

PR #566, opened 2020-03-11
Adding get_root_owners to the RepoOwners class

apahim:
... and some minor fixes


apahim:
cc @maorfr this is what I meant... we can easily get only the root owners

PR #567, opened 2020-03-11
New integration: gitlab-fork-compliance

apahim:
This re-implements the `test_fork.sh` as an integration.

### DIFF
--- a/reconcile/gitlab_fork_compliance.py
+++ b/reconcile/gitlab_fork_compliance.py
@@ -1,0 +1,83 @@
+import logging
+from gitlab import MAINTAINER_ACCESS
+
+from reconcile import queries
+from utils.gitlab_api import GitLabApi
+
+
+_LOG = logging.getLogger(__name__)
+
+QONTRACT_INTEGRATION = 'gitlab-fork-compliance'
+
+BLOCKED_LABEL = 'blocked/bot-access'
+
+MSG_BRANCH = ('@{user}, this Merge Request is using the "master" '
+              'source branch. Please submit a new Merge Request from another '
+              'branch.')
+
+MSG_ACCESS = ('@{user}, this fork of {project_name} is not shared with '
+              '[{bot}](/{bot}) as "Maintainer". !!Please '
+              '[add the user to the project]'
+              '({source_project_url}/project_members) '
+              'and retest by commenting "[test]" on the merge request.')
+
+
+def run(project_id, mr_id, maintainers_group, dry_run=False):
+    instance = queries.get_gitlab_instance()
+    settings = queries.get_app_interface_settings()
+
+    gitlab_cli = GitLabApi(instance, project_id=project_id,
+                           settings=settings)
+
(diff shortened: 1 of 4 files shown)


reconcile/gitlab_fork_compliance.py
```diff
@@ -0,0 +1,78 @@
...
+from reconcile import queries
+from utils.gitlab_api import GitLabApi
+from utils.gitlab_api import MRState
+
+
+_LOG = logging.getLogger(__name__)
+
+QONTRACT_INTEGRATION = 'gitlab-fork-compliance'
+
+MAINTAINERS_GROUP = 'app-sre'
```
maorfr: can this be an argument?

reconcile/gitlab_fork_compliance.py
```diff
@@ -0,0 +1,78 @@
...
+BOT_USER = 'devtools-bot'
```
maorfr: can we get this from the gitlab cli?

reconcile/gitlab_fork_compliance.py
```diff
@@ -0,0 +1,78 @@
...
+BLOCKED_LABEL = 'blocked/devtools-bot-access'
```
maorfr: let's take this opportunity to change the label to `blocked/bot-access` (i already added the label in the project)

reconcile/gitlab_fork_compliance.py
```diff
@@ -0,0 +1,78 @@
...
+    settings = queries.get_app_interface_settings()
+
+    gitlab_cli = GitLabApi(instance, project_id=project_id,
+                           settings=settings)
+
+    group = gitlab_cli.gl.groups.get(MAINTAINERS_GROUP)
+    maintainers = group.members.list()
+    bot = gitlab_cli.get_user(BOT_USER)
+
+    for mr in gitlab_cli.get_merge_requests(state=MRState.OPENED):
```
maorfr: unlike gitlab-owners, this integration should run as part of a pr check of every MR.
so it should get the merge request id as an argument and only act on that MR.

reconcile/gitlab_fork_compliance.py
```diff
@@ -0,0 +1,78 @@
...
+        source_project = GitLabApi(instance, project_id=mr.source_project_id,
+                                   settings=settings)
+        blocked = BLOCKED_LABEL in gitlab_cli.get_merge_request_labels(mr.iid)
+
+        # If the bot is not a maintainer on the fork,
```
maorfr: there is a missing check here - if the merge request's source branch is `master` - we should block it as well:
https://gitlab.cee.redhat.com/service/app-interface/blob/master/hack/test_fork.sh#L9

reconcile/gitlab_fork_compliance.py
apahim: sure

reconcile/gitlab_fork_compliance.py
apahim: o/ my bad

PR #569, opened 2020-03-11
Add option to run OpenShift related integrations without using a jump host

maorfr:
The OpenShift integrations are using a jump host to be able to reach private clusters.

In case the integrations are running in an environment with access to those clusters (by VPC peering for example), using the jump host is not required.

This PR adds a `--use-jump-host/--no-use-jump-host` option (defaults to True - use a jump host).

Since our jump host implementation relies on SSH, and to use SSH in a container in OpenShift a valid user must exist, this PR removes the need to have such user(s), hence the Dockerfile change.

This just may work!

### DIFF
--- a/helm/qontract-reconcile/values.yaml
+++ b/helm/qontract-reconcile/values.yaml
@@ -102,6 +102,7 @@ integrations:
     limits:
       memory: 300Mi
       cpu: 40m
+  extraArgs: --no-use-jump-host
   logs: true
 - name: openshift-groups
   resources:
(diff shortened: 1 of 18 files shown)


helm/qontract-reconcile/values.yaml
```diff
@@ -191,7 +198,7 @@ integrations:
     limits:
       memory: 1000Mi
       cpu: 400m
-  extraArgs: --external --vault-output-path app-sre/integrations-output
+  extraArgs: --external --no-use-jump-host--vault-output-path app-sre/integrations-output
```
apahim: missing white-space

maorfr: good catch!

PR #570, opened 2020-03-11
change log value

BumbleFeng:
Change logs config in the Helm chart file. Prepare to ship logs to additional providers

https://issues.redhat.com/browse/APPSRE-1618

### DIFF
--- a/helm/qontract-reconcile/templates/template.yaml
+++ b/helm/qontract-reconcile/templates/template.yaml
@@ -23,6 +23,7 @@ objects:
       spec:
         {{- if $integration.logs }}
         initContainers:
+        {{- if $integration.logs.slack }}
         - name: config
           image: quay.io/app-sre/busybox
           resources:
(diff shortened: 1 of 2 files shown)


maorfr:
w00t w00t! your first PR!

The intention is to use a separate sidecar for each location we want to ship logs to.
Your changes seem as if we plan to re-use the existing sidecar/initcontainer, which is a bit different.

In addition, does this pass `make generate`? (you'll need to install the helm 3 binary and call it `helm3`).

Let's also enhance the PR description a bit to indicate what these changes are. You did reference the Jira ticket, which is great, but let's add a description here as well.

mmclanerh:
@BumbleFeng  first PR! so cool! :D

maorfr:
@BumbleFeng please run `make generate` and commit that as well

maorfr:
@BumbleFeng can you please rebase on top of latest `master`?

PR #581, opened 2020-03-17
Replacing direct calls to gl client by our APIs

apahim:


reconcile/gitlab_fork_compliance.py
```diff
@@ -56,18 +54,16 @@ def run(self):
         # At this point, we know that the bot is a maintainer, so
         # we check if all the maintainers are in the fork, adding those
         # who are not
-        group = self.gl_cli.gl.groups.get(self.maintainers_group)
```
maorfr: i like this approach better!

taking us back to a past discussion: https://github.com/app-sre/qontract-reconcile/pull/530

apahim: sure, no worries

PR #586, opened 2020-03-17
[jira-watcher] basic auth instead of oauth

maorfr:
covers https://issues.redhat.com/browse/APPSRE-1630

revive jira-watcher!


maorfr:
works! https://coreos.slack.com/archives/CHVL0B0G7/p1584484382225200

PR #592, opened 2020-03-18
[ocm] only initiate ready clusters

maorfr:


tparikh:
https://issues.redhat.com/browse/APPSRE-1646

PR #601, opened 2020-03-22
[terraform-resources] add support for cloudwatch log groups

maorfr:
Covers https://issues.redhat.com/browse/APPSRE-1614

This PR adds support in terraform-resorces for a new provider: `cloudwatch`.

This provider will add:
- CloudWatch log group
- IAM user (+access key)
- IAM user policy (to allow log group access)

Matching app-interface MR: https://gitlab.cee.redhat.com/service/app-interface/merge_requests/3786

### DIFF
--- a/utils/terrascript_client.py
+++ b/utils/terrascript_client.py
@@ -1291,7 +1291,7 @@ class TerrascriptClient(object):
             "Statement": [
                 {
                     "Action": [
-                        "logs:CreateLogStream"
+                        "logs:CreateLogStream",
                         "logs:PutLogEvents"
                     ],
                     "Effect": "Allow",
--- a/utils/terrascript_client.py
+++ b/utils/terrascript_client.py
@@ -1291,10 +1291,8 @@ class TerrascriptClient(object):
             "Statement": [
                 {
                     "Action": [
-                        "logs:CreateLogStream",
-                        "logs:DescribeLogStreams",
-                        "logs:PutLogEvents",
-                        "logs:GetLogEvents"
+                        "logs:CreateLogStream"
+                        "logs:PutLogEvents"
                     ],
                     "Effect": "Allow",
                     "Resource":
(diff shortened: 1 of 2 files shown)


utils/terrascript_client.py
```diff
@@ -1232,6 +1235,103 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
...
+        # iam user policy for log group
+        policy = {
+            "Version": "2012-10-17",
+            "Statement": [
+                {
+                    "Action": [
+                        "logs:CreateLogStream",
+                        "logs:DescribeLogStreams",
+                        "logs:PutLogEvents",
+                        "logs:GetLogEvents"
```
tparikh: The IAM user does not need `logs:GetLogEvents` permission.

utils/terrascript_client.py
```diff
@@ -1232,6 +1235,103 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
...
+                user_tf_resource, identifier, output_prefix))
+
+        # iam user policy for log group
+        policy = {
+            "Version": "2012-10-17",
+            "Statement": [
+                {
+                    "Action": [
+                        "logs:CreateLogStream",
+                        "logs:DescribeLogStreams",
```
tparikh: The IAM user does not need `logs:DescribeLogStreams` permission.

utils/terrascript_client.py
```diff
@@ -1232,6 +1235,103 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
...
+
+    @staticmethod
+    def _get_retention_in_days(values, account, identifier):
+        default_retention_in_days = 14
+        allowed_retention_in_days = \
+            [1, 3, 5, 7, 14, 30, 60, 90, 120, 150,
+             180, 365, 400, 545, 731, 1827, 3653]
+
+        retention_in_days = \
+            values.get('retention_in_days') or default_retention_in_days
```
tparikh: Would it help to add a check here to make sure the value is a number? What would happen if user provided `14d` instead of `14`? Can we catch this error and show a ERROR in logs?

utils/terrascript_client.py
maorfr: removed

utils/terrascript_client.py
maorfr: removed

utils/terrascript_client.py
maorfr: changed logging from warning to error

PR #602, opened 2020-03-23
handle errors thrown during reconcile run

tparikh:


reconcile/terraform_resources.py
```diff
@@ -227,39 +227,52 @@ def run(dry_run=False, print_only=False,
...
+                     disable_service_account_keys=True)
+
+        if vault_output_path:
+            write_outputs_to_vault(vault_output_path, ri)
+
+    except Exception as e:
+        msg = 'There was problem running terraform resource reconcile.'
+        msg += ' Exception: {}'
+        msg = msg.format(str(e))
+        logging.error(msg)
```
maorfr: should we exit(1) here?

tparikh: updated

PR #607, opened 2020-03-25
handle errors thrown during reconcile run

BumbleFeng:


tparikh:
LGTM

apahim:
This will wake people up. I thought we would select which integrations should page us... Cc @tparikh

tparikh:
@apahim The integration has already failed at this point. We are just catching the exception and logging a better message instead of the stack trace.

apahim:
Yes, the unhandled exception had the potential to CrashLoop the integration. That is what wakes us up. Now we're handling the exception but it can still provoke the CrashLoop. Anyway, that's probably a different discussion.

PR #613, opened 2020-03-26
Default to expiring noncurrent objects after 30 days in versioned 3 b…

rrati:
…uckets

### DIFF
--- a/utils/terrascript_client.py
+++ b/utils/terrascript_client.py
@@ -601,6 +601,19 @@ class TerrascriptClient(object):
         if common_values.get('lifecycle_rules'):
             # common_values['lifecycle_rules'] is a list of lifecycle_rules
             values['lifecycle_rule'] = common_values['lifecycle_rules']
+        if versioning:
+            days = common_values.get('expire_noncurrent_version_days') or 30
+            rule = {
+                "id": "expire_noncurrent_versions",
+                "enabled": "true",
+                "noncurrent_version_expiration": {
+                    "days": days
+                }
+            }
+            if values.get('lifecycle_rule'):
+                values['lifecycle_rule'].append(rule)
+            else:
+                values['lifecycle_rule'] = rule
         sc = common_values.get('storage_class')
         if sc:
             sc = sc.upper()
--- a/utils/terrascript_client.py
+++ b/utils/terrascript_client.py
@@ -602,18 +602,26 @@ class TerrascriptClient(object):
             # common_values['lifecycle_rules'] is a list of lifecycle_rules
             values['lifecycle_rule'] = common_values['lifecycle_rules']
         if versioning:
-            days = common_values.get('expire_noncurrent_version_days') or 30
-            rule = {
-                "id": "expire_noncurrent_versions",
-                "enabled": "true",
-                "noncurrent_version_expiration": {
-                    "days": days
+            lrs = values.get('lifecycle_rule', [])
+            expiration_rule = False
+            for lr in lrs:
+                if "noncurrent_version_expiration" in lr:
+                    expiration_rule = True
+                    break
+            if not expiration_rule:
+                # Add a default noncurrent object expiration rule if
+                # if one isn't already set
+                rule = {
+                    "id": "expire_noncurrent_versions",
+                    "enabled": "true",
+                    "noncurrent_version_expiration": {
+                        "days": 30
+                    }
                 }
-            }
-            if values.get('lifecycle_rule'):
-                values['lifecycle_rule'].append(rule)
-            else:
-                values['lifecycle_rule'] = rule
+                if len(lrs) > 0:
+                    lrs.append(rule)
+                else:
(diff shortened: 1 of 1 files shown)


utils/terrascript_client.py
```diff
@@ -559,6 +559,19 @@ def populate_tf_resource_s3(self, resource, namespace_info):
...
+            days = common_values.get('expire_noncurrent_version_days') or 30
+            rule = {
+                "id": "expire_noncurrent_versions",
+                "enabled": "true",
+                "noncurrent_version_expiration": {
+                    "days": days
+                }
+            }
+            if values.get('lifecycle_rule'):
+                values['lifecycle_rule'].append(rule)
```
maorfr: This may create conflicts with requested policies IMO

rrati: It definitely could, but it's no different than how replication has to handle it.  If someone where to write their own expiration or replication rule in the lifecycle rules AND used the configuration options we provide to simplify then we could have a conflict.

maorfr: The git-keeper bucket may be a good example to look at

jmelis: perhaps I don't understand the problem, but can't we check the contents of `rules` and add it only if there are no conflicting rules?

also @rrati I prefer `common_values.get('expire_noncurrent_version_days', 30)`

rrati: git-keeper just has multiple lifecycle rules, just as this will produce.  I don't see anything there that will prevent anything from stepping on other configs.  The definitions are laid out in the config, but that presumes you understand terraform configuration.  Both storage_class and this I setup to know as little of the terraform syntax and possible, basically reducing several lines of boilerplate config to a single line.  If we don't want that simplicity then I can remove the work to simplify and just make users use lifecycle definitions.

rrati: @jmelis That's possible I suppose.  We'd have to iterate over all existing rules looking for if a field is set in that rule.  I'm not sure that is better though, as it would cause as much or more confusion than an overlap here could imo.  In this case, one can look at the rules on the bucket and see if there's an overlap causing a problem.  If however we look and only add if there's no conflict, then a user will have explicitly defined something that doesn't get applied and they likely won't know why.

IMO, we need to make a choice.  Do we expect our users to know terraform syntax well enough to define their own config per terraform rules/syntax, or do we abstract that knowledge away as much as possible and provide simple configuration options for complex concepts.  There's trade-offs either way.

rrati: In this case, we are actually linking two pieces of functionality.  If s3 versioning is enabled, then we are basically enforcing some kind of object expiration policy for noncurrent objects.  Since we default to enabling versioning, if we don't enforce this then total bucket storage will increase each time an object in an s3 bucket is changed since all versions of objects will be kept.

maorfr: I think that if life cycle rules are defined for a bucket, we don't add the default life cycle rule. This will prevent any possible conflicts. wdyt?

maorfr:
This may actually be fine, since it's not a life cycle rule for objects, but rather to bucket versions.

utils/terrascript_client.py
maorfr: This may actually be fine, since it's not a life cycle rule for objects, but rather to bucket versions.

rrati: If a user knows terraform then they could always define a lifecycle rule that would conflict with this.  I don't think that's especially likely, but it is possible.

The most likely scenario I can think of is someone actually knows terraform, knows that bucket versioning is enabled by default or explicitly enables it, and wants to keep bucket storage to a minimum and so creates their own lifecycle rule to delete noncurrent object versions.  In this scenario the bucket would end up with 2 lifecycle rules to delete noncurrent object versions, and whichever one is shortest would take effect.

maorfr: that is very opaque. what if that someone wants longer retention? we should not override them. I strongly believe we should avoid conflicts when we can.

rrati: If someone wants a longer extension they can set the config option to more than 30 days.

utils/terrascript_client.py
```diff
@@ -601,6 +601,27 @@ def populate_tf_resource_s3(self, resource, namespace_info):
         if common_values.get('lifecycle_rules'):
             # common_values['lifecycle_rules'] is a list of lifecycle_rules
             values['lifecycle_rule'] = common_values['lifecycle_rules']
+        if versioning:
+            lrs = values.get('lifecycle_rule', [])
```
maorfr, line 605: should this be `lifecycle_rules`?

rrati: It's lifecycle_rules in common_values, but just lifecycle_rule in values.  In terraform, it's multiple lifecycle_rule definition stanzas for an s3 object.

maorfr: perfect!

PR #614, opened 2020-03-27
ship log to cloudwatch via syslog

BumbleFeng:
Add Syslog as new logger handler
Add [syslog-cloudwatch-bridge](https://github.com/app-sre/syslog-cloudwatch-bridge) as sidecar to ship log to cloudwatch
https://issues.redhat.com/browse/APPSRE-1664

### DIFF
--- a/helm/qontract-reconcile/templates/template.yaml
+++ b/helm/qontract-reconcile/templates/template.yaml
@@ -22,8 +22,8 @@ objects:
           app: qontract-reconcile
       spec:
         {{- if $integration.logs }}
-        initContainers:
         {{- if $integration.logs.slack }}
+        initContainers:
         - name: config
           image: quay.io/app-sre/busybox
           resources:
@@ -123,9 +123,9 @@ objects:
           - name: qontract-reconcile-toml
             mountPath: /config
         {{- if $integration.logs }}
+        {{- if $integration.logs.slack }}
           - name: logs
             mountPath: /fluentd/log/
-        {{- if $integration.logs.slack }}
         - name: fluentd
           image: quay.io/app-sre/fluentd:latest
           resources:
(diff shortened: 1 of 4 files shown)


reconcile/cli.py
```diff
@@ -219,6 +220,11 @@ def init_log_level(log_level):
     logging.basicConfig(format=format,
                         datefmt='%Y-%m-%d %H:%M:%S',
                         level=level)
+    syslog = SysLogHandler(address=('localhost', 5014))
```
tparikh: make host & port configurable from `config.toml`

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -141,15 +141,68 @@ objects:
...
+                secretKeyRef:
+                  name: ${CLOUDWATCH_SECRET}
+                  key: log_group_name
+          resources:
+            requests:
+              memory: 30Mi
+              cpu: 15m
+            limits:
+              memory: 60Mi
+              cpu: 25m
```
tparikh: Are you sure this is enough? How about `100m`?

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -141,15 +141,68 @@ objects:
...
+              valueFrom:
+                secretKeyRef:
+                  name: ${CLOUDWATCH_SECRET}
+                  key: log_group_name
+          resources:
+            requests:
+              memory: 30Mi
+              cpu: 15m
+            limits:
+              memory: 60Mi
```
tparikh: Are you sure this is enough?

reconcile/cli.py
tparikh: What happens if there is no syslog server like in dev env?

BumbleFeng: Syslog goes through UDP connect. There are any errors or warnings for no Syslog server.

helm/qontract-reconcile/templates/template.yaml
BumbleFeng: I will change it

reconcile/cli.py
tparikh: Do you mean to say there will be no errors?

BumbleFeng: Yes.

reconcile/cli.py
```diff
@@ -219,6 +220,14 @@ def init_log_level(log_level):
     logging.basicConfig(format=format,
                         datefmt='%Y-%m-%d %H:%M:%S',
                         level=level)
+    if 'syslog' in config.get_config():
```
BumbleFeng: Check if Syslog server is configured

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -22,8 +22,8 @@ objects:
           app: qontract-reconcile
       spec:
         {{- if $integration.logs }}
-        initContainers:
         {{- if $integration.logs.slack }}
+        initContainers:
```
maorfr: good catch!

reconcile/cli.py
maorfr: all integrations get the same config file. that means that all integrations will attempt to send logs to syslog.
unless i'm missing something?

reconcile/cli.py
```diff
@@ -219,6 +220,14 @@ def init_log_level(log_level):
...
+        syslog_server = config.get_config()['syslog']['server']
+        syslog_port = int(config.get_config()['syslog']['port'])
```
maorfr: `config.get_config()` should be called once

reconcile/cli.py
```diff
@@ -219,6 +220,14 @@ def init_log_level(log_level):
...
+        syslog = SysLogHandler(address=(syslog_server, syslog_port))
+        formatter = logging.Formatter(format, datefmt='%Y-%m-%d %H:%M:%S')
```
maorfr: can we extract the date format to a variable and re-use it?

maorfr: also, does this mean we are shipping logs from all levels?

reconcile/cli.py
```diff
@@ -229,8 +238,8 @@ def init_log_level(log_level):
 def integration(ctx, configfile, dry_run, log_level):
     ctx.ensure_object(dict)
 
-    init_log_level(log_level)
     config.init_from_toml(configfile)
+    init_log_level(log_level)
```
maorfr: i think we should start thinking about sretoolbox.logging
out of scope of this PR of course.
@apahim, wdyt?

also feel free to take a look at this PR!

apahim:
Hi @BumbleFeng, awesome work here. I like the approach you're taking, using Python logging, but it diverges from what we already have in place. My suggestion is to keep it consistent by using the same approach for both fluentd and cloudwatch.
Question: can we just extend the fluentd configuration to also send the log messages to cloudwatch, without the need for a syslog server side-car? Or even to have a side-car, but a fluentd one that looks almost the same, just with the cloudwatch config instead of a slack config?
I'm thinking of https://github.com/fluent-plugins-nursery/fluent-plugin-cloudwatch-logs

reconcile/cli.py
tparikh: @BumbleFeng @maorfr Raised a good point. We should make log level for CloudWatch configurable. Debug logs may contain messages that should not be stored in external systems. We should make sure only log messages at `INFO` level or above are shipped to CloudWatch.

tparikh:
If we were to use Fluentd, we should look into adding k8s metadata to log messages using filter https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter

reconcile/cli.py
BumbleFeng: `syslog.setLevel(log_level)` should work

reconcile/cli.py
BumbleFeng: Yes. Syslog goes through UDP connect. So there are no errors or warnings for no Syslog server. Maybe we should add a new config flag to make Syslog optional.

BumbleFeng:
Good idea. I will dive in Fluentd config.

BumbleFeng:
Check fluentd solution here
https://github.com/app-sre/qontract-reconcile/pull/620

maorfr:
Why did you choose to use `@copy` over a cloud watch fluentd plugin?

BumbleFeng:
In case we need to ship the same log to both slack and cloudwatch. 

> Why did you choose to use `@copy` over a cloud watch fluentd plugin?

maorfr:
Have you seen this?
https://github.com/fluent-plugins-nursery/fluent-plugin-cloudwatch-logs
As suggested by @apahim

tparikh:
@maorfr he is using the right plugin. See https://github.com/app-sre/qontract-reconcile/pull/620/files#diff-56aa24bda0547277ddfbaffb78f5bf99R96

Copy sends logs to both slack and cloudwatch.
I would propose moving discussion to https://github.com/app-sre/qontract-reconcile/pull/620

maorfr:
Sorry, misread. Ignore me.

PR #620, opened 2020-03-30
ship log to cloudwatch via fluentd

BumbleFeng:
Reuse fluentd to ship log to both slack and cloudwatch.
Need to update fluentd image first. 
https://github.com/app-sre/fluentd/pull/2

### DIFF
--- a/helm/qontract-reconcile/templates/template.yaml
+++ b/helm/qontract-reconcile/templates/template.yaml
@@ -23,7 +23,6 @@ objects:
       spec:
         {{- if $integration.logs }}
         initContainers:
-        {{- if $integration.logs.slack }}
         - name: config
           image: quay.io/app-sre/busybox
           resources:
@@ -34,6 +33,7 @@ objects:
               memory: 20Mi
               cpu: 25m
           env:
+          {{- if $integration.logs.slack }}
           - name: SLACK_WEBHOOK_URL
             valueFrom:
               secretKeyRef:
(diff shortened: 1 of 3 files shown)


tparikh:
Nice work. Seems much simpler than using syslog bridge.

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -67,20 +79,32 @@ objects:
...
+                username sd-app-sre-bot
+                flush_interval 10s
+                message "\`\`\`[{{ $integration.name }}] %s\`\`\`"
+              </store>
+              {{- end }}
+              {{- if $integration.logs.cloudwatch }}
+              <store>
+                @type cloudwatch_logs
+                log_group_name ${LOG_GROUP_NAME}
+                log_stream_name ${LOG_STREAM_NAME}
```
maorfr: You can use integration.name here if it makes sense

tparikh: This approach will create a unique stream for each integration for each run instead of putting logs from various instances in same stream. Everytime pod is restarted, we get a new stream. 

You can still search across streams using CloudWatch.

tparikh: Stream name will be set to the pod name. Pod name will tell us which integration and which instance of it wrote those logs.

BumbleFeng:
> Nice work. Seems much simpler than using syslog bridge.

Yes. And don't need an extra container.

helm/qontract-reconcile/templates/template.yaml
maorfr: Why would it create a log stream per run?

maorfr: I don't think that setting the log stream name to the pod name is desired.

I meant to use `{{ $integration.name }}`

tparikh: Why not?

maorfr: Each pod has a different name. We want to see a stream of events from each integration over time. No?

tparikh: Yes and for that you don't worry about stream name. You can search on a log group in CloudWatch.

PR #626, opened 2020-03-31
Don't override IMAGE_TAG if it's already specified in target params

kylape:
This is useful for the insights migration so that we can update the image tag by hand while the deployment templates are in a different repo from the code.


maorfr:
Thanks @kylape!

PR #632, opened 2020-04-02
Adding reviewers to the loop

apahim:
They don't merge, but the do cleanup reviewers messages.

Example:
https://gitlab.cee.redhat.com/service/managed-tenants/merge_requests/113

### DIFF
--- a/reconcile/gitlab_owners.py
+++ b/reconcile/gitlab_owners.py
@@ -43,15 +43,20 @@ class MRApproval:
         change_owners_map = dict()
         paths = self.gitlab.get_merge_request_changed_paths(self.mr.iid)
         for path in paths:
-            path_owners = self.owners.get_path_owners(path)
-            if not path_owners:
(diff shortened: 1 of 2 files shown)


reconcile/gitlab_owners.py
```diff
@@ -43,15 +43,20 @@ def get_change_owners_map(self):
         change_owners_map = dict()
         paths = self.gitlab.get_merge_request_changed_paths(self.mr.iid)
         for path in paths:
-            path_owners = self.owners.get_path_owners(path)
-            if not path_owners:
+            owners = self.owners.get_path_owners(path)
+            path_approvers = owners['approvers']
+            path_reviewers = owners['reviewers']
```
maorfr, line 48: will both always exist?

apahim: At this point, yes. The `utils.repo_owners.RepoOwners()._get_owners_map()` below will give them an empty list case the section is missing in the file.

PR #627, opened 2020-04-01
Created openshift-performance-parameters integration

rporres:
https://issues.redhat.com/browse/APPSRE-1674

* Great thx to @aditya-konarde for starting this on https://github.com/app-sre/qontract-reconcile/pull/237. It's been a great inspiration
* ~Docker file builds correctly but it is yet untested. It depends on https://github.com/rporres/valet-libsonnet/pull/2 to be reviewed, merged and transferred to app-sre org~
* This needs the schema in https://gitlab.cee.redhat.com/service/app-interface/merge_requests/3950
* There are a number of complex schema validations done here. This ensures consistency and are needed for the app to work, but I'd love to make them in app-interface, where they belong. JSON schema extensions are not so powerful and are meant for a different purpose

Signed-off-by: Rafa Porres Molina <rporresm@redhat.com>

PR #633, opened 2020-04-03
Add performance parameters to openshift template

rporres:
This was missing in https://github.com/app-sre/qontract-reconcile/pull/627

### DIFF
--- a/openshift/qontract-reconcile.yaml
+++ b/openshift/qontract-reconcile.yaml
@@ -3034,6 +3034,162 @@ objects:
           emptyDir: {}
         - name: fluentd-config
           emptyDir: {}
+- apiVersion: extensions/v1beta1
+  kind: Deployment
+  metadata:
+    labels:
+      app: qontract-reconcile
+    name: qontract-reconcile-openshift-performance-parameters
+  spec:
+    replicas: 1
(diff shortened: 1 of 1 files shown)


openshift/qontract-reconcile.yaml
```diff
@@ -3034,6 +3034,162 @@ objects:
...
+            requests:
+              cpu: 100m
+              memory: 300Mi
+          volumeMounts:
+          - name: qontract-reconcile-toml
+            mountPath: /config
+          - name: logs
+            mountPath: /fluentd/log/
+        - name: fluentd
+          image: quay.io/app-sre/fluentd:latest
```
aditya-konarde, line 3156: Can we pin this to a specific image tag instead? Using `latest` is against our own best practices :)

rporres: Unfortunately this is like that for the rest of the deployments. You're 100% right but I won't do it in this PR, as it is unrelated to it. But I have created an issue to track it https://issues.redhat.com/browse/APPSRE-1722

jmelis: I kinda have the feeling I'm the one to blame here

PR #635, opened 2020-04-03
Fix permissions inside jsonnet vendor dir

rporres:


rporres:
cc @aditya-konarde

PR #642, opened 2020-04-06
[WIP][openshift-acme] v2 migration

maorfr:
Part of https://issues.redhat.com/browse/APPSRE-1678

This PR changes the deployment manifests of openshift-acme to match the v2 changes in the upstream repo.

This will require a change to the default image in app-interface.

Still wondering what will be a good way to perform the migration.

### DIFF
--- a/utils/openshift_acme.py
+++ b/utils/openshift_acme.py
@@ -1,76 +1,31 @@
 ACME_DEPLOYMENT = """
-apiVersion: extensions/v1beta1
 kind: Deployment
+apiVersion: apps/v1
 metadata:
   name: %(deployment_name)s
   labels:
     app: openshift-acme
 spec:
-  progressDeadlineSeconds: 600
-  replicas: 1
-  revisionHistoryLimit: 10
+  replicas: 2
   selector:
     matchLabels:
       app: openshift-acme
   strategy:
-    type: Recreate
+    type: RollingUpdate
   template:
     metadata:
-      creationTimestamp: null
       labels:
         app: openshift-acme
     spec:
       containers:
-      - env:
-        - name: OPENSHIFT_ACME_EXPOSER_IP
-          valueFrom:
-            fieldRef:
-              apiVersion: v1
-              fieldPath: status.podIP
-        - name: OPENSHIFT_ACME_ACMEURL
-          value: 'https://acme-v01.api.letsencrypt.org/directory'
-        - name: OPENSHIFT_ACME_LOGLEVEL
-          value: '4'
-        - name: OPENSHIFT_ACME_NAMESPACE
-          valueFrom:
-            fieldRef:
-              apiVersion: v1
-              fieldPath: metadata.namespace
-        image: %(image)s
+      - image: %(image)s
         imagePullPolicy: Always
         name: openshift-acme
-        ports:
-        - containerPort: 5000
-          protocol: TCP
-        resources:
-          limits:
-            cpu: 50m
-            memory: 100Mi
-          requests:
-            cpu: 5m
-            memory: 50Mi
-        terminationMessagePath: /dev/termination-log
-        terminationMessagePolicy: File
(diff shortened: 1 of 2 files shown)


utils/openshift_acme.py
```diff
@@ -1,76 +1,31 @@
...
+apiVersion: apps/v1
 metadata:
   name: %(deployment_name)s
   labels:
     app: openshift-acme
 spec:
-  progressDeadlineSeconds: 600
-  replicas: 1
-  revisionHistoryLimit: 10
+  replicas: 2
```
jfchevrette, line 9: we probably don't need the extra replica here? As long as the pod is up "most of the time" it will be able to handle new certs, renewals, etc...

utils/openshift_acme.py
```diff
@@ -1,76 +1,31 @@
...
-              apiVersion: v1
-              fieldPath: metadata.namespace
-        image: %(image)s
+      - image: %(image)s
         imagePullPolicy: Always
         name: openshift-acme
-        ports:
-        - containerPort: 5000
-          protocol: TCP
-        resources:
```
jfchevrette, line 45: I think it'd be good to retain the resource limits for good measure

utils/openshift_acme.py
```diff
@@ -1,76 +1,31 @@
...
-          readOnly: true
-      dnsPolicy: ClusterFirst
-      restartPolicy: Always
-      schedulerName: default-scheduler
-      securityContext: {}
-      serviceAccount: %(serviceaccount_name)s
+        args:
+        - --exposer-image=quay.io/tnozicka/openshift-acme:exposer
+        - --loglevel=4
+        - --namespace=acme-controller
```
jfchevrette, line 27: I believe we want to template this so that it is watching the correct namespace

utils/openshift_acme.py
maorfr: done

utils/openshift_acme.py
maorfr: done

utils/openshift_acme.py
maorfr: hard one!

utils/openshift_acme.py
```diff
@@ -1,76 +1,38 @@
...
-apiVersion: extensions/v1beta1
 kind: Deployment
+apiVersion: apps/v1
 metadata:
   name: %(deployment_name)s
   labels:
     app: openshift-acme
 spec:
-  progressDeadlineSeconds: 600
   replicas: 1
```
rporres, line 9: Interestingly enough, the templates from the doc set this to 2 whereas in the past it was set to 1...

utils/openshift_acme.py
```diff
@@ -1,76 +1,38 @@
...
-        - mountPath: /dapi
-          name: podinfo
-          readOnly: true
-      dnsPolicy: ClusterFirst
-      restartPolicy: Always
-      schedulerName: default-scheduler
-      securityContext: {}
-      serviceAccount: %(serviceaccount_name)s
+        args:
+        - --exposer-image=quay.io/tnozicka/openshift-acme:exposer
```
rporres, line 32: This version schema is so weird. There are no versions, just `quay.io/tnozicka/openshift-acme:exposer` and `quay.io/tnozicka/openshift-acme:controller`

utils/openshift_acme.py
```diff
@@ -1,76 +1,38 @@
...
+        - --loglevel=4
```
rporres, line 33: you may want to template this too to add more info if we need to debug

rporres:
At least we have a sort of proper version tagging in quay https://quay.io/repository/tnozicka/openshift-acme?tag=latest&tab=tags

maorfr:
we may move openshift-acme to be deployed using openshift-saas-deploy. closing this for now.

PR #664, opened 2020-04-13
[openshift] handle null annotations

maorfr:
example: https://github.com/RedHatInsights/saas-templates/blob/master/templates/inventory/insights-host-delete.yaml#L6


maorfr:
@apahim is there a better way to do this?

PR #670, opened 2020-04-15
Print more useful error message when a deployment template can't be fetched

kylape:
Instead of a stack trace, print the url to the file and the target hash,
along with the error message pulled from the original Exception object.


jmelis:
LGTM!

PR #671, opened 2020-04-15
Log stack trace alongside the error message

tparikh:
When we started catching these exceptions and logging errors, we were not getting stack trace. By using logging.exception() right after except, we will now get the error message and a stack trace.

### DIFF
--- a/reconcile/openshift_acme.py
+++ b/reconcile/openshift_acme.py
@@ -150,9 +150,6 @@ def run(dry_run=False, thread_pool_size=10, internal=None,
 
         ob.realize_data(dry_run, oc_map, ri)
 
-    except Exception as e:
-        msg = 'There was problem running openshift acme reconcile.'
-        msg += ' Exception: {}'
-        msg = msg.format(str(e))
-        logging.error(msg)
+    except Exception:
+        logging.exception('Exception reconciling openshift acme')
         sys.exit(1)
--- a/reconcile/openshift_groups.py
+++ b/reconcile/openshift_groups.py
@@ -237,9 +237,6 @@ def run(dry_run=False, thread_pool_size=10, internal=None,
             if not dry_run:
                 act(diff, oc_map)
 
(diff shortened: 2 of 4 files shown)


reconcile/openshift_acme.py
```diff
@@ -151,8 +151,6 @@ def run(dry_run=False, thread_pool_size=10, internal=None,
         ob.realize_data(dry_run, oc_map, ri)
 
     except Exception as e:
-        msg = 'There was problem running openshift acme reconcile.'
-        msg += ' Exception: {}'
-        msg = msg.format(str(e))
-        logging.error(msg)
+        logging.exception('Exception running openshift acme reconcile. ' +
+                          '{}'.format(str(e)))
```
apahim, line 155: No need to include the `str(e)`. The exception message will be in the traceback.

tparikh: logging.exception needs a message

apahim: Yes it does. The redundant part is only the `str(e)`.

tparikh: PR updated

apahim:
A broader observation: the only difference between the state of this PR and the state we were before the exception handlers is the logging message (e.g. `Exception reconciling openshift acme`).  We now print the traceback and we still interrupt the execution. We are not quieter nor safer.
Question: should we even keep the exception handlers just to have that extra logging message?

tparikh:
> A broader observation: the only difference between the state of this PR and the state we were before the exception handlers is the logging message (e.g. `Exception reconciling openshift acme`). We now print the traceback and we still interrupt the execution. We are not quieter nor safer.
> Question: should we even keep the exception handlers just to have that extra logging message?

Yes 100% agreed. I am looking at ways to improve logging and exception handling. This PR is WIP and not something I am even considering merging right now. I will close this PR as it is causing confusion.

PR #672, opened 2020-04-16
[user validator] introduce new integration

maorfr:
covers https://issues.redhat.com/browse/APPSRE-1760

this PR adds a new validating integration which should only run in pr-checks (same as service-dependencies).

### DIFF
--- a/reconcile/user_validator.py
+++ b/reconcile/user_validator.py
@@ -1,0 +1,53 @@
+import sys
+import logging
+
+from utils.gpg import gpg_key_valid
+import reconcile.queries as queries
+
+
+QONTRACT_INTEGRATION = 'user-validator'
+
+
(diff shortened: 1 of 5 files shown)


reconcile/user_validator.py
```diff
@@ -0,0 +1,53 @@
...
+                ok = False
+
+    return ok
+
+
+def run(dry_run=False):
+    ok = True
+    users = queries.get_users()
+
+    ok = validate_users_single_path(users)
```
jmelis, line 49: this ok will be overridden by the one in the next line

maorfr: https://github.com/app-sre/qontract-reconcile/pull/673

maorfr: :facepalm:

PR #674, opened 2020-04-16
Recycle pods connected to Deployments and DeploymentConfigs using

rrati:
their deployment strategies

### DIFF
--- a/utils/oc.py
+++ b/utils/oc.py
@@ -221,6 +221,9 @@ class OC(object):
 
         recyclables = {}
         for pod in pods_to_recycle:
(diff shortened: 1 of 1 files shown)


maorfr:
@rrati can you describe what this PR is doing with a little more details?
also, on these changes i would suggest we have at least one reviewer.

rrati:
@maorfr This change will find the owner/controller of a pod (deployment, dc) and modify the podspec on that object to ensure pod replacement is done via deployment strategy.  Previous code just killed all pods which could result in downtime.  This will trigger a new deployment that honors the deployment strategy.  RC controller is handled the same as it was, by just killing the pod

PR #682, opened 2020-04-21
Place the try in the correct position

apahim:
The try block is starting too late, after the point where the exception
can actually happen. Fixing, while increasing the logging severity.


apahim:
cc @maorfr

PR #687, opened 2020-04-21
[openshift-saas-deploy] add option to deploy only specific saas-files and specific envs

maorfr:
Part of https://issues.redhat.com/browse/APPSRE-1276

Following the work in https://gitlab.cee.redhat.com/service/app-interface/merge_requests/4267

This PR adds the ability to run the openshift-saas-deploy integration for a specific saas-file and for a specific environment targeted from that saas file.

Each target is a namespace, and each namespace is a part of an environment.

This will allow us to create dedicated jenkins jobs that will deploy using this integration, but without dependency on other saas files.

Example:
```
qontract-reconcile --config config.toml --dry-run openshift-saas-deploy --saas-file-name saas-github-mirror --env-name App-SRE-stage
```

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -714,6 +714,9 @@ SAAS_FILES_QUERY = """
       targets {
         namespace {
           name
+          environment {
+            name
+          }
           cluster {
             name
             serverUrl
@@ -754,10 +757,26 @@ SAAS_FILES_QUERY = """
 """
 
 
-def get_saas_files():
+def get_saas_files(saas_file_name='', env_name=''):
     """ Returns SaasFile resources defined in app-interface """
     gqlapi = gql.get_api()
-    return gqlapi.query(SAAS_FILES_QUERY)['saas_files']
+    saas_files = gqlapi.query(SAAS_FILES_QUERY)['saas_files']
+
+    for saas_file in saas_files[:]:
+        if saas_file_name:
+            if saas_file['name'] != saas_file_name:
+                saas_files.remove(saas_file)
+                continue
+        if env_name:
+            for rt in saas_file['resourceTemplates']:
+                targets = rt['targets']
+                for target in targets[:]:
+                    namespace = target['namespace']
+                    environment = namespace['environment']
+                    if environment['name'] != env_name:
+                        targets.remove(target)
+
+    return saas_files
 
 
 PERFORMANCE_PARAMETERS_QUERY = """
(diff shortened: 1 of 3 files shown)


reconcile/queries.py
```diff
@@ -704,10 +707,26 @@ def get_app_interface_sql_queries():
...
     """ Returns SaasFile resources defined in app-interface """
     gqlapi = gql.get_api()
-    return gqlapi.query(SAAS_FILES_QUERY)['saas_files']
+    saas_files = gqlapi.query(SAAS_FILES_QUERY)['saas_files']
+
+    for saas_file in saas_files[:]:
+        if saas_file_name:
+            if saas_file['name'] != saas_file_name:
+                saas_files.remove(saas_file)
+                continue
```
apahim, line 719: Nice! Very memory effective :)

reconcile/queries.py
```diff
@@ -704,10 +707,26 @@ def get_app_interface_sql_queries():
...
 
-def get_saas_files():
+def get_saas_files(saas_file_name='', env_name=''):
     """ Returns SaasFile resources defined in app-interface """
     gqlapi = gql.get_api()
-    return gqlapi.query(SAAS_FILES_QUERY)['saas_files']
+    saas_files = gqlapi.query(SAAS_FILES_QUERY)['saas_files']
+
+    for saas_file in saas_files[:]:
+        if saas_file_name:
```
apahim, line 716: placing the if before the for would avoid processing the loop

reconcile/queries.py
```diff
@@ -704,10 +707,26 @@ def get_app_interface_sql_queries():
 """
 
 
-def get_saas_files():
+def get_saas_files(saas_file_name='', env_name=''):
```
apahim, line 710: So, when the cli runs:
```
qontract-reconcile --config config.toml --dry-run openshift-saas-deploy --saas-file-name ""
```
Should we find no saas files or all saas files? As of now, we find all saas files.
Not a big deal, since we have the cli under our control... just setting the expectations here.
If we want "no saas files", the default here (and everywhere else) should be `None` and the condition `if saas_file_name is not None`.

maorfr: the default should be all saas files, and that is indicated by not passing `--saas-file-name`.

similar to `kubectl get pods` vs `kubectl get pod <name>`.

reconcile/queries.py
maorfr: i learned from the best!

reconcile/queries.py
maorfr: added a condition before the for loop

reconcile/queries.py
maorfr: i'm going to merge this and we can discuss further. we will have many more iterations :)

PR #691, opened 2020-04-22
added cli helper log line if we're bailing out of openshift-resources

mmclanerh:


mmclanerh:
@maorfr totally reasonable and more appropriate + less ambiguous in terms of tracking down what/where... also, looking now, this isn't necessary. fetch_states() should already be providing exceptional exception logging.

tparikh:
image

PR #696, opened 2020-04-23
[gitlab-housekeeping] handle additional labels

maorfr:
In the case that a MR has additional labels expect for `lgtm`, the automerge doesn't work. This PR handles that.

still very ugly that the labels are hard coded. we'll get there.

@apahim can you please review?

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -8,6 +8,11 @@ import reconcile.queries as queries
 from utils.gitlab_api import GitLabApi
 
 
+MERGE_LABELS = ['lgtm', 'automerge', 'approved']
+HOLD_LABELS = ['awaiting-approval', 'blocked/bot-access',
+               'do-not-merge/hold', 'do-not-merge/pending-review']
+
+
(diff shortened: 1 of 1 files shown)


reconcile/gitlab_housekeeping.py
```diff
@@ -96,7 +105,7 @@ def rebase_merge_requests(dry_run, gl, rebase_limit):
         if not labels:
             continue
 
-        good_to_rebase = all(l in REBASE_LABELS for l in labels)
```
apahim, line 99: good catch

reconcile/gitlab_housekeeping.py
```diff
@@ -76,8 +81,12 @@ def handle_stale_items(dry_run, gl, days_interval, enable_closing, item_type):
                     gl.remove_label(item, item_type, LABEL)
 
 
+def is_good_to_merge(labels):
+    return any(l in MERGE_LABELS for l in labels) and \
+        not any(l in HOLD_LABELS for l in labels)
```
apahim, line 86: yes, makes sense

PR #705, opened 2020-04-27
Make busybox and fluentd images configurable

rporres:
So that we can avoid `latest` tag


rporres:
Then we can set in https://github.com/app-sre/saas-app-interface/blob/master/qontract-reconcile-services/qontract-reconcile.yaml

* [busybox](https://quay.io/repository/app-sre/busybox?tab=tags) to `1.31.1`
* [fluentd](https://quay.io/repository/app-sre/fluentd?tab=tags) to `8e1d986`

and return to the `IfNotPresent` pull policy

rporres:
cc @maorfr @aditya-konarde

PR #707, opened 2020-04-27
network policy for openshift acme

tparikh:


maorfr:
please hold until we have an agreed approach on how we do this (added a discussion item for today's coordination session)

rporres:
I would make this optional as there can be already network policies that cover what this fixes

maorfr:
this is covered by openshift-networkpolicies, so i would recommend we keep the network policy for it.

added docs here: https://gitlab.cee.redhat.com/service/app-interface/blob/master/README.md#enable-network-traffic-between-namespaces-via-app-interface-openshiftnamespace-1yml

PR #709, opened 2020-04-28
[jenkins-job-buidler] dynamically generate openshift-saas-deploy jobs

maorfr:
part of https://issues.redhat.com/browse/APPSRE-1276

This PR adds a dynamic jenkins job generation per saas file per environment.


maorfr:
This is very raw, but works. I'll merge this and have a review at a later point, as other things settle into place.

PR #711, opened 2020-04-30
add null check for keys to delete

tparikh:


mmclanerh:
reproduced error locally and tested patch, seems to be way less mad now.
```
(venv) [mmclanerh.qontract@mmclane qontract-reconcile]$ qontract-reconcile --config config.debug.toml --dry-run aws-support-cases-sos
[2020-04-30 16:06:13] [ERROR] [aws_a...{snip -- expected}
```

PR #713, opened 2020-05-01
elasticsearch integration for app-interface

tparikh:


utils/terrascript_client.py
```diff
@@ -1726,3 +1735,183 @@ def unpack_namespace_info(self, namespace_info):
...
+        service_role = {
+          'aws_service_name': "es.amazonaws.com"
+        }
+        return aws_iam_service_linked_role(identifier, **service_role)
+
+
+    @staticmethod
+    def is_elasticsearch_domain_name_valid(name):
+        """ Handle for Error creating Elasticsearch:
+        InvalidParameterValue: Elasticsearch domain name must must start with a
```
rrati, line 1759: must must ...

utils/terrascript_client.py
```diff
@@ -1726,3 +1735,183 @@ def unpack_namespace_info(self, namespace_info):
...
+
+        account, identifier, values, output_prefix, output_resource_name = \
+            self.init_values(resource, namespace_info)
+
+        tf_resources = []
+
+        self.init_common_outputs(tf_resources, namespace_info,
+                                 output_prefix, output_resource_name)
+
+        if values['identifier'] and \
```
rrati, line 1778: Should just be able to use identifier from the return of init_values.  It also has to exist so should you be checking for len > 0?

utils/terrascript_client.py
```diff
@@ -1726,3 +1735,183 @@ def unpack_namespace_info(self, namespace_info):
...
+           not self.is_elasticsearch_domain_name_valid(values['identifier']):
+            raise ElasticSearchResourceNameInvalidError(
+                f"[{account}] ElasticSearch domain name must must start with" +
+                f" a lowercase letter and must be between 3 and 28 " +
+                f"characters. Valid characters are a-z (lowercase only), 0-9" +
+                f", and - (hyphen). " +
+                f"{values['identifier']}")
+
+        # values['domain_name'] = identifier
+        elasticsearch_version = values.get('elasticsearch_version')
```
rrati, line 1788: I would provide a default or this will throw an exception if elasticsearch_version isn't defined

utils/terrascript_client.py
```diff
@@ -1726,3 +1735,183 @@ def unpack_namespace_info(self, namespace_info):
...
+        if values['identifier'] and \
+           not self.is_elasticsearch_domain_name_valid(values['identifier']):
+            raise ElasticSearchResourceNameInvalidError(
+                f"[{account}] ElasticSearch domain name must must start with" +
+                f" a lowercase letter and must be between 3 and 28 " +
+                f"characters. Valid characters are a-z (lowercase only), 0-9" +
+                f", and - (hyphen). " +
+                f"{values['identifier']}")
+
+        # values['domain_name'] = identifier
```
rrati, line 1787: remove?

utils/terrascript_client.py
```diff
@@ -1726,3 +1735,183 @@ def unpack_namespace_info(self, namespace_info):
...
+                f"[{account}] Invalid ElasticSearch version" +
+                f" {values['elasticsearch_version']} provided" +
+                f" for resource {values['identifier']}.")
+
+        values['cluster_config']["zone_awareness_enabled"] = False
+
+        es_values = {}
+        es_values["domain_name"] = identifier
+        es_values["elasticsearch_version"] = \
+            values.pop('elasticsearch_version')
```
rrati, line 1800: Why pop?  Why not use the variable you set/checked earlier?

utils/terrascript_client.py
```diff
@@ -1726,3 +1735,183 @@ def unpack_namespace_info(self, namespace_info):
...
+        dedicated_master_count = cluster_config.\
+            get("dedicated_master_count", 3)
+        zone_awareness_enabled = cluster_config.\
+            get("zone_awareness_enabled", True)
+
+        es_values["cluster_config"] = {
+            "instance_type": cluster_config.get("instance_type",
+                                                't2.small.elasticsearch'),
+            "instance_count": cluster_config.get("instance_count", 3),
+            "dedicated_master_enabled": dedicated_master_enabled,
```
rrati, line 1843: If dedicated master is disabled, will tf error or create bad resource if master_type/count are set?

utils/terrascript_client.py
tparikh: I had defined a default but removed it. Without the error, user may assume their ES cluster is setup with the version they provided.

utils/terrascript_client.py
tparikh: yep noticed after raising pr

utils/terrascript_client.py
tparikh: thanks for catching this

utils/terrascript_client.py
tparikh: I don't see any other integrations check for length of identifier

tparikh: If we want to check the length of identifier maybe we should add check in the init_values func? WDYT?

utils/terrascript_client.py
tparikh: good catch

utils/terrascript_client.py
tparikh: Now I only add dedicated master config if enabled

utils/terrascript_client.py
rrati: I was just thinking in the context of this check.  You're looking for a non-none identifier in values, but it will always exist as a minimum empty string.  Other integrations don't check for len of identifier, but I was suggesting a check that might be worthwhile.  so long as is_elasticsearch_domain_name_valid can handle an empty string, I don't think you need to add a len check.  Removing the first condition here seems like it would be fine.

utils/terrascript_client.py
rrati: This effectively makes elasticsearch_version a required config param, but it is in the defaults yaml.  I would recommend moving version to the ns yaml level if we want it to be required.  Most required config params don't live in the defaults yaml.

tparikh: Yep that makes sense. I think I will default to latest `7.4` if none is provided. However, a value is provided, I will run the check and make sure it is a valid version.

reconcile/elasticsearch_exceptions.py
```diff
@@ -0,0 +1,19 @@
...
+        super(ElasticSearchResourceMissingSubnetIdError, self).__init__(
+            str(msg)
+        )
+
+
+class ElasticSearchResourceVersionInvalidError(Exception):
+    def __init__(self, msg):
+        super(ElasticSearchResourceVersionInvalidError, self).__init__(
+            str(msg)
+        )
```
maorfr, line 19: I would prefer to have these exceptions under `utils` if that works

utils/terrascript_client.py
maorfr: i would prefer to keep the namespace definition as clean as possible, and add schema validation to the defaults file. does that make sense? we should do that to many other resources as well.

utils/terrascript_client.py
tparikh: updated

utils/terrascript_client.py
tparikh: @maorfr Sounds good. I will refactor the code later once I sync with you on this.

reconcile/elasticsearch_exceptions.py
tparikh: 👍

PR #714, opened 2020-05-02
updated, with suggested fix for monthyl reporting + new jjb parse cha…

mmclanerh:
…nges

@maorfr  this passes my local dry-run..  (no exception), but i'm not sure it's doing so without skipping over other data.. ptal and review before any merge.. thanks!

### DIFF
--- a/utils/jjb_client.py
+++ b/utils/jjb_client.py
@@ -302,7 +302,7 @@ class JJB(object):
             all_jobs[name] = []
             jobs = self.get_jobs(wd, name)
             for job in jobs:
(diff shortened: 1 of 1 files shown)


maorfr:
thanks for this @mmclanerh!

as i said on the slack thread, this will not work (or rather, work partially).
the reason is the same function is also used to get all jobs that has a `build-master` in their name, and these jobs usually start with `gl-` or `gh-`.

your app-interface MR will solve the problem perfectly:
https://gitlab.cee.redhat.com/service/app-interface/merge_requests/4563

thanks again!

PR #717, opened 2020-05-03
[openshift-saas-deploy] enable multiple running instances of the integration

maorfr:
This PR adds support for a new feature - `multiple_callers`.

When openshift integrations call `realize_data` with `multiple_callers=True`, the execution assumes that there may be multiple running instances of the integration, and each instance should only handle its own resources.

In our use case, the openshift-saas-deploy integration can now run in multiple instances, modifying the same namespaces from different saas files, and still not get in the way of itself.

We introduce a new annotation that will be used for that purpose - `qontract.caller_name`. This annotation will only be set by the openshift-saas-deploy integration.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -215,7 +215,8 @@ def check_unused_resource_types(ri):
 
 
 def realize_data(dry_run, oc_map, ri,
-                 take_over=False):
+                 take_over=False,
+                 mutilple_callers=False):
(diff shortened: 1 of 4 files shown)


maorfr:
I'm going to merge this as this, as it only affects openshift-saas-deploy. It is also a breaking change, so I had to run that locally to force the changes.

I would still love a review!

cc @jmelis @apahim

PR #730, opened 2020-05-05
add support for cluster as vpc peering target

jfchevrette:
This updates the terraform_vpc_peerings integration to support targetting clusters as VPC peering targets

### DIFF
--- a/reconcile/terraform_vpc_peerings.py
+++ b/reconcile/terraform_vpc_peerings.py
@@ -20,6 +20,7 @@ def fetch_desired_state(settings):
                 if c.get('peering') is not None]
     ocm_map = OCMMap(clusters=clusters, integration=QONTRACT_INTEGRATION,
                      settings=settings)
+
     for cluster_info in clusters:
         cluster = cluster_info['name']
         ocm = ocm_map.get(cluster)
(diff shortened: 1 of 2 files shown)


reconcile/terraform_vpc_peerings.py
```diff
@@ -32,6 +53,87 @@ def fetch_desired_state(settings):
...
+                      "'network-mgmt' access level on the cluster"
+                return None, msg
+
+            # find role to use for aws access
+            awsAccount['assume_role'] = \
+                ocm.get_aws_infrastructure_access_terraform_assume_role(
+                    cluster,
+                    awsAccount['uid'],
+                    awsAccount['terraformUsername']
+                )
```
rporres, line 102: Is there a possibility that this returns empty? It shouldn't be the case as we should always have a `terraform` user in the cluster's access list... Maybe too defensive to check for it, just worth mentioning :)

jfchevrette: The function should always return a string according to https://github.com/app-sre/qontract-reconcile/blob/ba004224abea6d3b0834a4a353c9873118957a96/utils/ocm.py#L165-L167

rporres: well that `return` is inside a for loop with two  conditions with `continue` statements. I guess there's a chance that it doesn't return anything...

jfchevrette: I see what you mean! I've added a test to check if returned value is None (if it never reaches the return statement)

PR #733, opened 2020-05-06
[openshift-saas-deploy] trigger by moving commits integration

maorfr:
This PR adds an integration called `openshift-saas-deploy-trigger-moving-commits`.

Whenever a commit is changed under a ref in a saas file (a new PR is merged to `master` for example), this integration will trigger the correct jobs in Jenkins to perform the required deployment operations.

If such a job has an `upstream` field, it will not be triggered, but rather listen to another job and execute after it (enabling image build to be synced).

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -370,11 +370,35 @@ def ls(ctx, integration):
 @click.argument('integration')
 @click.argument('key')
 @click.pass_context
+def get(ctx, integration, key):
+    settings = queries.get_app_interface_settings()
+    accounts = queries.get_aws_accounts()
+    state = State(integration, accounts, settings=settings)
+    value = state.get(key)
+    print(value)
+
+
+@state.command()
+@click.argument('integration')
+@click.argument('key')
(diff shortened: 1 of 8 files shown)


tools/qontract_cli.py
```diff
@@ -366,6 +366,18 @@ def ls(ctx, integration):
...
+@state.command()
+@click.argument('integration')
+@click.argument('key')
+@click.pass_context
+def get(ctx, integration, key):
+    settings = queries.get_app_interface_settings()
+    accounts = queries.get_aws_accounts()
+    state = State(integration, accounts, settings=settings)
+    value = state.get(key)
+    print(value)
```
apahim, line 378: leftover?

apahim: ops, actual print :)

apahim:
@maor, the code looks very good. I'm having a hard time to understand all the flows here, so I will have to take more time to understand the bigger picture. But that's on me. As far as I can tell right now, this looks awesome.

PR #762, opened 2020-05-14
introduce feature toggles using unleash

maorfr:
part of https://issues.redhat.com/browse/APPSRE-1728

this PR adds a check for the state of a feature toggle in an unleash instance.
if the feature toggle does not exist, the integration will be executed.

depends on https://gitlab.cee.redhat.com/service/app-interface/merge_requests/4816

### DIFF
--- a/helm/qontract-reconcile/templates/template.yaml
+++ b/helm/qontract-reconcile/templates/template.yaml
@@ -132,6 +132,16 @@ objects:
           - name: APP_INTERFACE_STATE_BUCKET_ACCOUNT
             value: "${APP_INTERFACE_STATE_BUCKET_ACCOUNT}"
           {{- end }}
+          - name: UNLEASH_API_URL
+            valueFrom:
+              secretKeyRef:
+                name: unleash
+                key: API_URL
+          - name: UNLEASH_CLIENT_ACCESS_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: unleash
+                key: CLIENT_ACCESS_TOKEN
           {{- with $integration.extraEnv }}
           {{- range $i, $env := . }}
           - name: {{ $env.secretKey }}
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1,6 +1,9 @@
+import os
 import sys
 import logging
 import click
+
+from UnleashClient import UnleashClient
 
 import utils.config as config
 import utils.gql as gql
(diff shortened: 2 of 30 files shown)


helm/qontract-reconcile/templates/template.yaml
```diff
@@ -132,6 +132,16 @@ objects:
...
           {{- end }}
+          - name: UNLEASH_API_URL
+            valueFrom:
+              secretKeyRef:
+                name: unleash
+                key: API_URL
+          - name: UNLEASH_CLIENT_ACCESS_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: unleash
```
apahim, line 143: `app-interface-token` (or `app-interface-unleash-token`, if you changed it there)

reconcile/cli.py
```diff
@@ -203,9 +207,44 @@ def f(function):
...
+    api_url = os.environ.get('UNLEASH_API_URL')
+    client_access_token = os.environ.get('UNLEASH_CLIENT_ACCESS_TOKEN')
+    if not (api_url and client_access_token):
+        return True
+
+    # hide INFO logging from UnleashClient
+    logger = logging.getLogger()
+    default_logging = logger.level
+    logger.setLevel(logging.ERROR)
+    defer(lambda: logger.setLevel(default_logging))
```
apahim, line 225: we have to start using different logging streams... but that's not only here, just something to keep in mind.

helm/qontract-reconcile/templates/template.yaml
maorfr: answered in https://gitlab.cee.redhat.com/service/app-interface/merge_requests/4816#note_1298642

reconcile/cli.py
```diff
@@ -203,9 +207,44 @@ def f(function):
...
+    headers = {'Authorization': f'Bearer {client_access_token}'}
+    client = UnleashClient(url=api_url,
+                           app_name='qontract-reconcile',
+                           custom_headers=headers)
+    client.initialize_client()
+    defer(lambda: client.destroy())
+
+    state = client.is_enabled(integration_name,
+                              fallback_function=get_feature_toggle_true)
+    return state
```
apahim, line 236: great. clean and efficient :)

reconcile/cli.py
```diff
@@ -203,9 +207,44 @@ def f(function):
     return f
 
 
-def run_integration(func, *args):
+def get_feature_toggle_default(feature_name: str, context: dict) -> bool:
```
apahim, line 210: sorry, overlooked this. can we please omit the type hints?

PR #766, opened 2020-05-15
set default monitoring interval for rds  when enhanced monitoring is true

tparikh:


utils/terrascript_client.py
```diff
@@ -456,7 +456,17 @@ def populate_tf_resource_rds(self, resource, namespace_info,
             values['parameter_group_name'] = pg_name
 
         enhanced_monitoring = values.pop('enhanced_monitoring')
+
+        # monitoring interval should only be set if enhanced monitoring
+        # is true
+        if not enhanced_monitoring and values.get('monitoring_interval'):
```
rrati, line 462: This will except without a default value if monitoring_interval isn't defined.

tparikh: Defaulting to `None`

maorfr: `get` does not throw an exception when the key doesn't exist, that's `pop`:
```
>>> d = {}
>>> value = d.get('key')
>>> value
>>> d.pop('key')
KeyError: 'key'
>>> value = d.pop('key', None)
>>> value
>>>
```

PR #767, opened 2020-05-16
change the way db name is handled

tparikh:

### DIFF
--- a/utils/terrascript_client.py
+++ b/utils/terrascript_client.py
@@ -417,8 +417,11 @@ class TerrascriptClient(object):
                                  output_prefix, output_resource_name)
 
         # we want to allow an empty name, so we
-        # only validate names which are not emtpy
-        if values['name'] and not self.validate_db_name(values['name']):
+        # only validate names which are not empty
+        if (
+            values.get('name', None) and
+            not self.validate_db_name(values['name'])
+        ):
             raise FetchResourceError(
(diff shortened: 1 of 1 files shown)


utils/terrascript_client.py
```diff
@@ -417,8 +417,11 @@ def populate_tf_resource_rds(self, resource, namespace_info,
                                  output_prefix, output_resource_name)
 
         # we want to allow an empty name, so we
-        # only validate names which are not emtpy
-        if values['name'] and not self.validate_db_name(values['name']):
+        # only validate names which are not empty
+        if (
+            values.get('name', None) and
```
maorfr, line 422: `get` does not throw an exception when the key doesn't exist, that's `pop`:
```
>>> d = {}
>>> value = d.get('key')
>>> value
>>> d.pop('key')
KeyError: 'key'
>>> value = d.pop('key', None)
>>> value
>>>
```

utils/terrascript_client.py
```diff
@@ -615,7 +618,7 @@ def populate_tf_resource_rds(self, resource, namespace_info,
         tf_resources.append(output(output_name, value=output_value))
         # db.name
         output_name = output_prefix + '[db.name]'
-        output_value = values['name']
+        output_value = values.get('name', '')
```
maorfr, line 621: same here

PR #778, opened 2020-05-22
adds qontract-cli promquery - runs promql queries

jmelis:


jmelis:
not very resilient yet, it would need to check if the cluster exists, etc... but good enough to get this subcommand started

PR #779, opened 2020-05-27
register error on cross cluster namespace policy

rporres:
Fixes https://issues.redhat.com/browse/APPSRE-1955


rporres:
Tested locally with https://gitlab.cee.redhat.com/service/app-interface/merge_requests/4995

PR #783, opened 2020-05-29
add support for cluster as vpc peering target

jfchevrette:
This PR adds the ability to peer an OCM cluster to another OCM cluster

Notes for the reviewer:
- I changed `populate_vpc_peerings` in `terraform_client.py` so that each the requester and accepter have their own account. This is to accomodate the fact a cluster-to-cluster peering use two different accounts/roles to create the requester and accepter objects on each respective account
- The code makes a few assumptions, such as the presence of some dict keys, which come from the graphql query results and required fields in the schemas. As such the code doesn't currently check for the presence of these keys which IMO would be a bit redundant
- Please checkout and run the integration locally. Note that you will also need https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/4611 as there are schema changes.
- When you run the integration locally, note that a few existing peerings have changes in the resulting terraform spec. This is because of the addition of the `provider` field on the accepter dict in `populate_vpc_peerings`. From my findings this is OK, as without this they defaulted to the `__DEFAULT__` provider, which has the same credentials as the one that is now being specified. The resulting terraform spec doesn not cause a change in the plan.

### DIFF
--- a/reconcile/terraform_vpc_peerings.py
+++ b/reconcile/terraform_vpc_peerings.py
@@ -15,13 +15,156 @@ QONTRACT_INTEGRATION = 'terraform_vpc_peerings'
 QONTRACT_INTEGRATION_VERSION = semver.format_version(0, 1, 0)
 
 
-def fetch_desired_state(settings):
+def find_matching_peering(from_cluster, peering, to_cluster, desired_provider):
+    """
+    Ensures there is a matching peering with the desired provider type
+    going from the destination (to) cluster back to this one (from)
+    """
+    peering_info = to_cluster['peering']
+    peer_connections = peering_info['connections']
+    for peer_connection in peer_connections:
+            if not peer_connection['provider'] == desired_provider:
+                continue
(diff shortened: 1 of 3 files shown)


reconcile/terraform_vpc_peerings.py
```diff
@@ -15,13 +15,156 @@
...
+    """
+    Generate an AWS account object from a cluster's awsInfrastructureAccess
+    groups and access levels
+    """
+    ocm = ocm_map.get(cluster['name'])
+    account = None
+    for awsAccess in cluster['awsInfrastructureAccess']:
+        if awsAccess.get('accessLevel', "") == access_level:
+            account = {
+                'name': awsAccess['awsGroup']['account']['name'],
```
maorfr, line 45: quoting @apahim:
> I'd reduce the number of dict lookups for the same key by placing it in a variable on the first lookup. Not a big difference thoug, just fine tunning.

PR #791, opened 2020-06-03
[aws-ecr-image-pull-secrets] introduce new integration

maorfr:
this integration generates image pull secrets for AWS ECR and stores them in vault.

the plan is to consume these secrets into namespaces using openshift-vault-secrets.

this integration will run once every 6 hours in a CronJob.

### DIFF
--- a/reconcile/aws_ecr_image_pull_secrets.py
+++ b/reconcile/aws_ecr_image_pull_secrets.py
@@ -1,0 +1,52 @@
+import base64
+import json
+import logging
+
+import reconcile.queries as queries
+import utils.vault_client as vault_client
+
+from utils.aws_api import AWSApi
+
+QONTRACT_INTEGRATION = 'aws-ecr-image-pull-secrets'
+
+
+def enc_dec(data):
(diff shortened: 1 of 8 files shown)


reconcile/aws_ecr_image_pull_secrets.py
```diff
@@ -0,0 +1,52 @@
...
+
+import reconcile.queries as queries
+import utils.vault_client as vault_client
+
+from utils.aws_api import AWSApi
+
+QONTRACT_INTEGRATION = 'aws-ecr-image-pull-secrets'
+
+
+def enc_dec(data):
```
jmelis, line 13: can you explain what this is for?

maorfr: this is to handle TypeErrors from base64 encoding and decoding of strings (and thank you to python3).

jmelis: got it, base64encode expects a bytes-like object...

PR #801, opened 2020-06-04
introduce terraform-aws-route53 integration

jfchevrette:


utils/terrascript_client.py
```diff
@@ -306,6 +310,66 @@ def populate_additional_providers(self, accounts):
...
+            values = {
+                'name': origin,
+            }
+
+            zone_resource = aws_route53_zone(zone_id, **values)
+            self.add_resource(account_name, zone_resource)
+
+            counts = defaultdict(int)
+            for record in records:
+                counts[record['name']] += 1
```
apahim, line 335: Nice usage of the defaultdict :)
Have you seen the `collections.Counter`? Just mentioning, no advantages for this case anyway.

utils/terrascript_client.py
```diff
@@ -306,6 +310,66 @@ def populate_additional_providers(self, accounts):
...
+
+                # terraform identifier for record
+                # We need to replace some characters that are not allowed
+                record_id = record_id.replace(".", "__")
+                record_id = record_id.replace("@", "_ORIGIN_")
+                record_id = record_id.replace("*", "_wildcard_")
+
+                values = {
+                    'zone_id': zone_resource.fullname,
+                    'name': record['name'],
```
apahim, line 357: I'd reduce the number of dict lookups for the same key by placing it in a variable on the first lookup. Not a big difference thoug, just fine tunning.

apahim:
Love it! This is going to reduce a lot of toil. Tks for that!

utils/terrascript_client.py
```diff
@@ -306,6 +310,66 @@ def populate_additional_providers(self, accounts):
...
+                }
+
+                if isinstance(record['target'], str):
+                    values['records'] = [record['target']]
+                else:
+                    msg = f"record {record['name']} for zone {origin} " \
+                          f"is not a string (got: {type(record['target'])}"
+                    logging.error(msg)
+
+                values['ttl'] = record.get('ttl') or zone.get('default_ttl')
```
apahim, line 368: According to rfc2181, TTL `0` is valid, so we should use it instead of assuming the `default_ttl` when that's the case.

reconcile/terraform_aws_route53.py
```diff
@@ -0,0 +1,75 @@
...
+    if tf is None:
+        sys.exit(1)
+
+    defer(lambda: tf.cleanup())
+
+    deletions_detected, err = tf.plan(enable_deletion)
+    if err:
+        sys.exit(1)
+    if deletions_detected and not enable_deletion:
+        sys.exit(1)
```
maorfr, line 68: i think that this integration should be allowed to delete resources. DNS holds no data (like RDS).
we can do that at a future iteration of course.

utils/terrascript_client.py
jfchevrette: Thanks for the pointer! I've dropped the need for `defaultdict` or `collections.Counter` in this case but I'll keep it in mind for future needs!

utils/terrascript_client.py
jfchevrette: Easy optimisation. Done!

utils/terrascript_client.py
jfchevrette: The idea of `default_ttl` is to let us define a value to use globally on a zone for records that don't have any. This is to simplify and reduce repetition as we typically have the same TTL on all records.
A TTL of zero is almost always not what you want and I've never seen it used ever. It should not be the default.

reconcile/terraform_aws_route53.py
jfchevrette: Very good point. I've removed the condition and deletion will now occur on apply

utils/terrascript_client.py
apahim: So, if `record.get('ttl')` returns `0`, which is a valid ttl, should we still use the `default_ttl` instead? Just making sure we have the same understanding.

maorfr:
LGTM from my side!

jfchevrette:
Closing this PR and will open a new one for the rewrite using Boto3 instead of terraform

PR #813, opened 2020-06-09
[openshift-saas-deploy] wait for namespace to exist

maorfr:
in the case a namespace and a saas file are created in the same MR, the trigger to the job may arrive before the namespace was created.

this PR adds the option for an integration to request to wait for the namespace to exist.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -8,6 +8,8 @@ from utils.oc import StatusCodeError
 from utils.openshift_resource import (OpenshiftResource as OR,
                                       ResourceInventory)
 
+from sretoolbox.utils import retry
+
 
 class StateSpec(object):
     def __init__(self, type, oc, cluster, namespace, resource, parent=None,
@@ -175,7 +177,14 @@ def fetch_current_state(namespaces=None,
     return ri, oc_map
 
 
-def apply(dry_run, oc_map, cluster, namespace, resource_type, resource):
+@retry(max_attempts=20)
+def wait_for_namespace_exists(oc, namespace):
+    if not oc.project_exists(namespace):
+        raise Exception(f'namespace {namespace} does not exist')
+
+
+def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
+          wait_for_namespace):
     logging.info(['apply', cluster, namespace, resource_type, resource.name])
 
     oc = oc_map.get(cluster)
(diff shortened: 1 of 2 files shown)


reconcile/openshift_base.py
```diff
@@ -185,8 +194,12 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource):
         # do not skip if this is a cluster scoped integration
         if namespace != 'cluster' and not oc.project_exists(namespace):
             msg = f"[{cluster}/{namespace}] namespace does not exist (yet)"
-            logging.warning(msg)
-            return
+            if wait_for_namespace:
+                logging.info(msg)
```
apahim, line 198: we will be up 3,5 minutes waiting here. can we add a msg about that? somthing like `msg += ' - waiting a bit'`

reconcile/openshift_base.py
```diff
@@ -185,8 +194,12 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource):
...
+                wait_for_namespace_exists(oc, namespace)
```
rporres, line 199: Does this mean that the whole integration with `wait_for_namespace`  is going to fail hard in case of one of the namespaces do not exist?

maorfr: and this will only apply to openshift-saas-deploy. we shouldn't enable it for integrations running continuously (although that will not be super harmful).

rporres: Sure, but even for openshift-saas-deploy... We move from warning and moving on to failing hard the whole integration. Why can't we keep the old behaviour with the retry mechanism?

maorfr: we want openshift-saas-deploy to fail hard if it was not able to deploy. unlike other integrations, this will only run once, and we'd like to know that it failed. does that make sense?

rporres: If that's the case, if it's possible I would control the behaviour in the integration, not in a library that's consumed by lots of integrations

maorfr: not possible with the current flow. and we may have a case in the future when we want to have this applied to other integrations as well...

rporres: d'oh... then a nice comment about the intentions and implications of using `wait_for_namespace` in the function declaration for those that actually go and have a look would be nice 😍

reconcile/openshift_base.py
maorfr: added!

reconcile/openshift_base.py
maorfr: added documentation to the `realize_data` method. thanks!

PR #815, opened 2020-06-10
Sharding support

apahim:


apahim:
cc @jmelis @maorfr

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -5,21 +5,22 @@ metadata:
...
 {{- range $i, $integration := .Values.integrations }}
+{{- range $shard_id := until (int $integration.sharding | default 1) }}
 - apiVersion: apps/v1
   kind: Deployment
   metadata:
     labels:
-      app: qontract-reconcile-{{ $integration.name }}
-    name: qontract-reconcile-{{ $integration.name }}
+      app: qontract-reconcile-{{ $integration.name }}-{{ $shard_id }}
+    name: qontract-reconcile-{{ $integration.name }}-{{ $shard_id }}
```
maorfr, line 14: i would only add $shard_id here

maorfr: and maybe not even add it if shards == 1?

utils/sharding.py
```diff
@@ -0,0 +1,13 @@
+import os
+import hashlib
+
+
+SHARDS = os.environ.get('SHARDS', 1)
+SHARD_ID = os.environ.get('SHARD_ID', 0)
+
+
+def is_shard(value):
```
maorfr, line 9: `is_in_shard`?

utils/sharding.py
```diff
@@ -0,0 +1,13 @@
+import os
+import hashlib
+
+
+SHARDS = int(os.environ.get('SHARDS', 1))
+SHARD_ID = int(os.environ.get('SHARD_ID', 0))
+
+
+def is_in_shard(value):
```
jmelis, line 9: would it be cheaper to return earlier if SHARDS is 1?

If we instrument an integration to support shards we will likely use them, so this won’t change anything, but if we ever switch back to one shard this is more correct.

reconcile/quay_mirror.py
```diff
@@ -102,6 +103,13 @@ def process_sync_tasks(self):
         sync_tasks = defaultdict(list)
         for org, data in summary.items():
             for item in data:
+
+                if is_in_shard(item["name"]):
```
jmelis, line 107: what if we move this log statement to is_in_shard? that way we won’t need to repeat ourselves and we can omit the log if SHARDS is 1

helm/qontract-reconcile/values.yaml
```diff
@@ -64,6 +64,7 @@ integrations:
       cpu: 300m
   logs:
     cloudwatch: true
+  shards: 2
```
jmelis, line 67: this may be just here temporarily, but do we intend to shard quay-mirror?  I guess we are waiting for the metrics of the other PR to adjusts the shards here.

utils/sharding.py
apahim: done

reconcile/quay_mirror.py
apahim: done

helm/qontract-reconcile/values.yaml
apahim: yes, removed

PR #817, opened 2020-06-11
[terraform-vpc-peerings] fixes for VPC ID

maorfr:
this PR adds some fixes as per #808.

@rporres can you review?

### DIFF
--- a/utils/aws_api.py
+++ b/utils/aws_api.py
@@ -570,6 +570,25 @@ class AWSApi(object):
         self.auth_tokens = auth_tokens
 
     def get_cluster_vpc_id(self, account):
+        """
+        Returns a cluster VPC ID.
+
+        :param account: a dictionary containing the following keys:
+                        - name - name of the AWS account
+                        - assume_role - role to assume to get access
+                                        to the cluster's AWS account
+                        - assume_region - region in which to operate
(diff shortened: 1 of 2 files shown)


utils/aws_api.py
```diff
@@ -570,6 +570,25 @@ def init_ecr_auth_tokens(self, accounts):
...
+                                        to the cluster's AWS account
+                        - assume_region - region in which to operate
+                        - assume_cidr - CIDR block of the cluster to
+                                        use to find the matching VPC
+        """
+        required_keys = \
+            ['name', 'assume_role', 'assume_region', 'assume_cidr']
+        ok = all(elem in account.keys() for elem in required_keys)
+        if not ok:
+            account_name = account.get('name')
```
rporres, line 588: nit: being strict, you may not have `name` here 😅

maorfr: i know, but it's the best we can do...

rporres: What do you mean by "the best we can do"? If you cannot be sure that you won't have `name` in the input dict, you cannot place it there (unless add more conditions)

maorfr: well, `name` is a mandatory field in the schema. it will be here for sure.

PR #824, opened 2020-06-11
Keep the previous behaviour until we have alerts

apahim:


jmelis:
What is the rationale here?

apahim:
@jmelis updated. The problem is the integration silently failing since we still don't have alerts.

PR #825, opened 2020-06-12
Add sentry

jmelis:


setup.py
```diff
@@ -45,6 +45,7 @@
         "tabulate>=0.8.6,<0.9.0",
         "UnleashClient>=3.4.2,<3.5.0",
         "prometheus-client~=0.8",
+        "sentry-sdk>=0.14.4,<0.15.0",
```
apahim, line 48: I'm trying to create a new, shorter, standard here (`~=0.14`) but it's not being very popular :)

jmelis: applied!

PR #826, opened 2020-06-12
Fix sentry: support multiple instances of the same team

apahim:
When the same team shows up in different places, the data structure is reset
and only the projects from the last definition are taken into account.

This patch add a check to decide if we have to extend the team projects.


apahim:
@rrati

PR #834, opened 2020-06-15
preprocess jinja vault path and key

maorfr:
there are cases when we want to add templating for the vault path or key to use. this PR implements that.

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -154,9 +154,10 @@ class UnknownTemplateTypeError(Exception):
         )
 
 
-def lookup_vault_secret(path, key, version=None, tvars={}):
-    path = process_jinja2_template(path, vars=tvars)
-    key = process_jinja2_template(key, vars=tvars)
+def lookup_vault_secret(path, key, version=None, tvars=None):
+    if tvars is not None:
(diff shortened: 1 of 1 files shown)


reconcile/openshift_resources_base.py
```diff
@@ -167,7 +169,8 @@ def lookup_vault_secret(path, key, version=None):
 
 
 def process_jinja2_template(body, vars={}, env={}):
-    vars.update({'vault': lookup_vault_secret})
+    vars.update({'vault': lambda p, k, v=None:
+        lookup_vault_secret(p, k, v, vars)})
```
apahim: so beatiful :)

reconcile/openshift_resources_base.py
```diff
@@ -154,7 +154,9 @@ def __init__(self, msg):
         )
 
 
-def lookup_vault_secret(path, key, version=None):
+def lookup_vault_secret(path, key, version=None, tvars={}):
```
apahim, line 157: please don't use mutable objects (the dict) as default parameters
this should be:
```
def lookup_vault_secret(path, key, version=None, tvars=None):
    if tvars is None:
        tvars = {}
```

PR #835, opened 2020-06-15
Let's exit on subprocess sigkill

apahim:

### DIFF
--- a/dockerfiles/hack/run-integration.py
+++ b/dockerfiles/hack/run-integration.py
@@ -99,4 +99,7 @@ if __name__ == "__main__":
         if return_code == State.ERROR:
             continue
 
+        if return_code == State.SIGKILL:
+            sys.exit(return_code)
+
         time.sleep(int(SLEEP_DURATION_SECS))
--- a/reconcile/status.py
+++ b/reconcile/status.py
@@ -2,3 +2,4 @@ class State:
     SUCCESS = 0
     ERROR = 1
     DATA_CHANGED = 3
+    SIGKILL = -9


reconcile/status.py
```diff
@@ -2,3 +2,4 @@ class State:
     SUCCESS = 0
     ERROR = 1
     DATA_CHANGED = 3
+    SIGKILL = -9
```
jmelis, line 5: To be honest I don't understand why we are seeing -9 instead of 137. Everywhere I look I see that the PID cause by a signal is 128 + signal code.

jmelis: documented here: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.poll

rporres:
I know I have approved this. But thinking about it, why not a different approach? Why don't we exit in any error code we don't know about. We currently know about `0`, `1` and `3`. Can we continue on those and die in any other status code?

dockerfiles/hack/run-integration.py
```diff
@@ -99,4 +99,7 @@ def run_cmd():
         if return_code == State.ERROR:
             continue
 
+        if return_code == State.SIGKILL:
+            sys.exit(return_code)
```
jmelis, line 103: I think we should return `137` and not `-9`, because `-9` is `247`.

jmelis:
If we do what @rporres is suggesting, which I like, we need to make sure we "correct" the error code.

```
if return_code < 0:
  return_code = 128 - return_code
```

apahim:
@rporres we only know about 3, that is "data changed". 1 can be anything. Meaning we are back to the state where we page people for CrashLooping pods.

apahim:
@jmelis yes, no worries, I will undo the python return code smartness once we agree on the cases to `sys.exit`.

jmelis:
I like @rporres proposal, wdyt

rporres:
We exit with `1` in lots of integrations when we have an error. I understood that was the agreement.

jmelis:
I just had a quick chat with @apahim and basically a week ago we were in the same place you are suggesting @rporres. We were crashlooping on OOM kills. If we do that, we will lose the metrics, and we would prefer to handle this with directed alerts. We can have one kind of alert for OOM kills, another one for unknown errors.

Let's do that for now, we hope to reduce alerting fatigue this way.

rporres:
What I was suggesting shouldn't change the behaviour of getting a `-9`.

Having said that, I don't see any other code that `0`, `1`, `3` or  -9`, so it's not crucial ATM :)

apahim:
The problem is that the fact that we know `1` does not make it special. `1` is a representation that something bad happened, but we don't know that something is given that we use it for handling exceptions. We can, at most, track the exception that generated it.

If we see another code, let say `5`, it has the same meaning of `1`: something bad happened and we can at most track the exception that caused it.

So, why whould we `sys.exit` on `5` and not on `1`?

My problem with `sys.exit()` in the main loop is that we loose the metrics, so we have to act on k8s CrashLooping pods, which is exatly what we were doing until last week. Even worse if we have a mixed behaviour: we have alerts from the metrics for exit code `1` and `-9` and alerts from the CrashLooping pods for exit code != `0`, `1`, `3` and `-9`.

By staying in the loop, we can have the metrics for all the codes and use them for custom alerts.

So, I'd be ok with either one of:

- option 1: never leave the main loop and use `/metrics` for alerting.
- option 2: leave the loop on exit code != `0`|`3` and drive alerts from CrashLooping pods.

rporres:
Thanks for the summary, @apahim 

I think I prefer option 2 as we will need to have CrashLooping alerts anyway to capture other error conditions. I agree that a mixed behaviour is bad.

PR #836, opened 2020-06-15
exit on RUN_ONCE only after return_code checks

maorfr:
in the case of a CronJob, if the data changes during execution, we want to re-try it.

i guess we also want to expose metrics on CronJobs as well...

### DIFF
--- a/dockerfiles/hack/run-integration.py
+++ b/dockerfiles/hack/run-integration.py
@@ -93,10 +93,10 @@ if __name__ == "__main__":
         if return_code == State.DATA_CHANGED:
             continue
 
-        if return_code == State.ERROR:
-            continue
-
         if RUN_ONCE:
             sys.exit(return_code)
+
+        if return_code == State.ERROR:
+            continue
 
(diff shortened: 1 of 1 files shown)


dockerfiles/hack/run-integration.py
```diff
@@ -99,4 +96,7 @@ def run_cmd():
         if return_code == State.ERROR:
             continue
 
+        if RUN_ONCE:
+            sys.exit(return_code)
+
```
apahim, line 101: Why you want to skip the `if RUN_ONCE` check in case of error? Then it will not run only once...

maorfr: i guess that makes sense. another integration that is running once is `github-users` (which is sending emails). we don't want our users to be spammed in case of an error.

moved it to before the error check.

maorfr: the important integration to think of here is `aws-ecr-image-pull-secrets`. but in any case, we can leave it up for the CronJob to retry.

PR #837, opened 2020-06-15
[openshift-saas-deploy-trigger-configs] only trigger on changes in relevant fields

maorfr:
currently, even if there is a change to `managedRoles` in a namespace file, a new deployment will be triggered to that namespace.

this PR sanitizes the namespace object to avoid triggering jobs on unrelated changes.

### DIFF
--- a/utils/saasherder.py
+++ b/utils/saasherder.py
@@ -503,14 +503,8 @@ class SaasHerder():
     def sanitize_namespace(namespace):
         """Only keep fields that should trigger a new job."""
         new_job_fields = {
-            # https://gitlab.cee.redhat.com/service/app-interface/-/blob/master
-            # /schemas/openshift/namespace-1.yml
             'namespace': ['name', 'cluster', 'app'],
-            # https://gitlab.cee.redhat.com/service/app-interface/-/blob/master
-            # /schemas/openshift/cluster-1.yml
             'cluster':  ['name', 'serverUrl'],
-            # https://gitlab.cee.redhat.com/service/app-interface/-/blob/master
-            # /schemas/app-sre/app-1.yml
             'app': ['name']
         }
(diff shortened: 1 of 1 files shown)


utils/saasherder.py
```diff
@@ -497,6 +499,19 @@ def get_configs_diff_saas_file(self, saas_file):
...
+        namespace = {k: v for k, v in namespace.items()
+                     if k in ['name', 'cluster', 'app']}
+        cluster = namespace['cluster']
+        namespace['cluster'] = {k: v for k, v in cluster.items()
+                                if k in ['name', 'serverUrl']}
+        app = namespace['app']
+        namespace['app'] = {k: v for k, v in app.items()
+                            if k in ['name']}
+        return namespace
+
```
apahim, line 514: This looks nice. My only suggestion here is to create a doctionary with the lists os items we want to use for triggering a new job, adding comments pointing to the related schemas to make the maintenance easier. Something like:
```
new_job_fields = {
    # https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/schemas/openshift/namespace-1.yml
    'namespace': ['name', 'cluster', 'app'],
    # https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/schemas/openshift/cluster-1.yml
    'cluster':  ['name', 'serverUrl'],
    # https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/schemas/app-sre/app-1.yml
    'app': ['name'],
}
```
Not a big deal though.

apahim: Well, maybe the comments are not a good idea. Anyway, the dict seems like a nice to have.

maorfr: done

maorfr: i'm easy on this

PR #841, opened 2020-06-16
[openshift-saas-deploy] add option to disable deployment to a target

maorfr:
this PR adds a way to disable deployments in openshift-saas-deploy to specified targets.

the actual job in Jenkins will still be enabled, but the target will be skipped during execution.

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/5471


apahim:
lgtm anyway :)

PR #844, opened 2020-06-16
add support for negative and regexp selectors to the perf params integration

jmelis:
This PR requires a change of schema: `selectors` and `additionalLabels` have been changed from a dict to a list.

I have added some tests to validate the logic.

This integration is backwards compatible with the previous schema.


reconcile/test/fixtures/openshift_perf_params/t1-params.yaml
```diff
@@ -0,0 +1,55 @@
+http_rates:
```
rporres, line 1: I would add a comment explaining this is an internal representation from `build_template_params`

rporres: nit: the fixture directory should be called as the integration

reconcile/test/fixtures/openshift_perf_params/t1-pp.yaml
```diff
@@ -0,0 +1,48 @@
+labels: '{"service":"observability"}'
```
rporres, line 1: i would add a comment explaining this is the app-interface representation

PR #845, opened 2020-06-17
logstoelasticsearch integration for app-interface

BumbleFeng:


utils/terrascript_client.py
```diff
@@ -41,7 +43,11 @@
                                aws_cloudwatch_log_group, aws_kms_key,
                                aws_kms_alias,
                                aws_elasticsearch_domain,
-                               aws_iam_service_linked_role)
+                               aws_iam_service_linked_role,
+                               aws_lambda_function, aws_lambda_permission,
+                               aws_cloudwatch_log_subscription_filter)
+
+LAMBDA_GIT = 'https://raw.githubusercontent.com/app-sre/aws-lambda/master/'
```
tparikh, line 50: I would recommend renaming `aws-lambda` repository. The name is too generic. Maybe something like `cloudwatch-stream-elasticsearch-lambda`

utils/terrascript_client.py
```diff
@@ -1485,6 +1492,163 @@ def populate_tf_resource_cloudwatch(self, resource, namespace_info):
...
+                "Version": "2012-10-17",
+                "Statement": [
+                    {
+                        "Effect": "Allow",
+                        "Action": [
+                            "logs:CreateLogGroup",
+                            "logs:CreateLogStream",
+                            "logs:PutLogEvents"
+                        ],
+                        "Resource": "*"
```
tparikh, line 1533: Can we list specific resources instead of `*`?

utils/terrascript_client.py
```diff
@@ -1485,6 +1492,163 @@ def populate_tf_resource_cloudwatch(self, resource, namespace_info):
...
+                    {
+                        "Effect": "Allow",
+                        "Action": [
+                            "ec2:DescribeNetworkInterfaces",
+                            "ec2:CreateNetworkInterface",
+                            "ec2:DeleteNetworkInterface",
+                            "ec2:DescribeInstances",
+                            "ec2:AttachNetworkInterface"
+                        ],
+                        "Resource": "*"
```
tparikh, line 1549: Can we list specific resources instead of `*`?

BumbleFeng: Find an existing role policy provided by AWS. https://console.aws.amazon.com/iam/home?region=us-east-1#/policies/arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole$jsonEditor. and they use "*"

BumbleFeng: Lambda need this policy to access resources in VPC according to https://docs.aws.amazon.com/lambda/latest/dg/configuration-vpc.html

PR #846, opened 2020-06-17
Moving the shards check

apahim:
Testing with 2 shards, each shard will:

Before
Screenshot from 2020-06-17 15-16-37

After
Screenshot from 2020-06-17 15-18-06

It didn't avoid the peaks though.


maorfr:
it feels like this will eventually land in openshif_base?

https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/openshift_base.py#L27

PR #850, opened 2020-06-18
[gitlab-housekeeping] re-add support for days_interval, limit and closing

maorfr:
depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/5556

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -111,8 +111,8 @@ def rebase_merge_requests(dry_run, gl, rebase_limit):
         if not good_to_rebase:
             continue
 
-        logging.info(['rebase', gl.project.name, mr.iid])
         if not dry_run and rebases < rebase_limit:
+            logging.info(['rebase', gl.project.name, mr.iid])
             try:
                 mr.rebase()
                 rebases += 1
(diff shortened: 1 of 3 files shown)


reconcile/gitlab_housekeeping.py
```diff
@@ -159,25 +159,31 @@ def merge_merge_requests(dry_run, gl, merge_limit):
...
+    default_limit = 8
+    default_enable_closing = False
     instance = queries.get_gitlab_instance()
     settings = queries.get_app_interface_settings()
     repos = queries.get_repos_gitlab_housekeeping(server=instance['url'])
 
     for repo in repos:
+        hk = repo['housekeeping']
         project_url = repo['url']
+        days_interval = hk.get('days_interval') or default_days_interval
```
apahim, line 179: what should happen when `days_interval` is `0`? here it is falling-back to `default_days_interval`.

reconcile/gitlab_housekeeping.py
```diff
@@ -159,25 +159,31 @@ def merge_merge_requests(dry_run, gl, merge_limit):
...
+        enable_closing = hk.get('enable_closing') or default_enable_closing
+        limit = hk.get('limit') or default_limit
```
apahim, line 181: ditto

reconcile/gitlab_housekeeping.py
maorfr: i think this is the desired behavior.

just in case, i'm not allowing 0 in the schema validation:
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/5556/diffs?commit_id=7610d2006302e81ed61617ad521fcb02075128af

reconcile/gitlab_housekeeping.py
maorfr: ditto ;)

PR #851, opened 2020-06-18
Avoid duplicates

apahim:


apahim:
@maorfr @jmelis we have bigger plans for the uniqueness test, but this solves 2 specific use cases and it's a quick win. Mind reviewing?

PR #852, opened 2020-06-18
[slack-usergroups] do not depend on PagerDuty name

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2057

instead of using the user's name field (which may contain special chars) we use the user's id and get his email (organization username).

### DIFF
--- a/reconcile/slack_usergroups.py
+++ b/reconcile/slack_usergroups.py
@@ -38,6 +38,7 @@ ROLES_QUERY = """
       name
       org_username
       slack_username
+      pagerduty_username
     }
     permissions {
       service
@@ -131,7 +132,7 @@ def get_slack_username(user):
 
 
 def get_pagerduty_name(user):
-    return user['org_username']
+    return user['pagerduty_username'] or user['org_username']
 
 
 def get_slack_usernames_from_pagerduty(pagerduties, users, usergroup):
@@ -163,8 +164,9 @@ def get_slack_usernames_from_pagerduty(pagerduties, users, usergroup):
              if pagerduty_name not in all_pagerduty_names]
         if not_found_pagerduty_names:
             msg = (
-                '[{}] org_usernames not found in app-interface: {} '
-                '(hint: create a user file if one does not exist.'
+                '[{}] PagerDuty username not found in app-interface: {} '
+                '(hint: user files should contain '
+                'pagerduty_username if it is different then org_username)'
             ).format(usergroup, not_found_pagerduty_names)
             logging.warning(msg)
         all_slack_usernames.extend(slack_usernames)
--- a/reconcile/slack_usergroups.py
+++ b/reconcile/slack_usergroups.py
@@ -38,7 +38,6 @@ ROLES_QUERY = """
       name
       org_username
       slack_username
-      pagerduty_name
     }
     permissions {
       service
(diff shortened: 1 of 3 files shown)


utils/pagerduty_api.py
```diff
@@ -25,14 +29,22 @@ def get_pagerduty_users(self, resource_type, resource_id):
 
         return users
 
+    def get_user(self, user_id):
+        for user in self.users:
+            if user.id == user_id:
```
rporres, line 34: What's this `id`? I cannot find it in https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/components/schemas/User

utils/pagerduty_api.py
```diff
@@ -25,14 +29,22 @@ def get_pagerduty_users(self, resource_type, resource_id):
...
     def get_schedule_users(self, schedule_id, now):
         s = pypd.Schedule.fetch(
             id=schedule_id,
             since=now,
             until=now,
             time_zone='UTC')
         entries = s['final_schedule']['rendered_schedule_entries']
-        return [entry['user']['summary'] for entry in entries]
+
+        return [self.get_user(entry['user']['id']) for entry in entries]
```
rporres, line 47: What happens if you get `None` here?

utils/pagerduty_api.py
```diff
@@ -51,7 +63,7 @@ def get_escalation_policy_users(self, escalation_policy_id, now):
                         self.get_schedule_users(target['id'], now)
                     users.extend(schedule_users)
                 elif target_type == 'user_reference':
-                    users.append(target['summary'])
+                    users.append(self.get_user(target['id']))
```
rporres, line 66: What happens if you get `None` here?

utils/pagerduty_api.py
maorfr: `id` is a user's ID in PagerDuty. example: `PV1WLMF` - https://redhat.pagerduty.com/users/PV1WLMF

utils/pagerduty_api.py
maorfr: i changed the `return None` to raise an exception to prevent this.

utils/pagerduty_api.py
maorfr: i changed the `return None` to raise an exception to prevent this.

reconcile/slack_usergroups.py
```diff
@@ -164,9 +163,8 @@ def get_slack_usernames_from_pagerduty(pagerduties, users, usergroup):
              if pagerduty_name not in all_pagerduty_names]
         if not_found_pagerduty_names:
             msg = (
-                '[{}] PagerDuty names not found in app-interface: {} '
-                '(hint: user files should contain '
-                'pagerduty_name if it is different then name)'
+                '[{}] org_usernames not found in app-interface: {} '
+                '(hint: create a user file if one does not exist.'
```
rporres, line 167: nit: 😬 
```suggestion
                '(hint: create a user file if one does not exist.)'
```

rporres:
Are you sure we can always find `org_username` from the email? For example, @jmelis uses `jmelis@redhat.com` as his mail in PD but his `org_username` is `jmelisba`

reconcile/slack_usergroups.py
maorfr: thanks!

maorfr:
@rporres solved it in https://github.com/app-sre/qontract-reconcile/pull/852/commits/79fbbf80660b6a9cbd193a5962b3e6a8aa07bb44

PR #855, opened 2020-06-21
[openshift-saas-deploy] improve parallelism model

maorfr:
openshift-saas-deploy currently works in parallel per saas file. this is ~fine for deploying multiple saas files, but has no impact when deploying a single saas file.

this PR updates the behavior to first initiate a list of specs to fetch when populating the desired state, and populate all specs in parallel. each spec represents the `target` level, so in fact this is changing the parallelism from per saas file to per target.

this PR also changes the image check to happen only once per target and not once per every resource in the target.

in addition, images are checked in parallel (a sane pool size is estimated by number of saas files and total number of namespaces, which represent targets). for a saas file with many images such as telemeter, this reduces the execution time from ~2m to ~20s.

### DIFF
--- a/utils/saasherder.py
+++ b/utils/saasherder.py
@@ -334,14 +334,19 @@ class SaasHerder():
         return image_auth
 
     def populate_desired_state(self, ri):
+        results = threaded.run(self.init_populate_desired_state_specs,
+                               self.saas_files,
+                               self.thread_pool_size)
+        desired_state_specs = \
+            [item for sublist in results for item in sublist]
         threaded.run(self.populate_desired_state_saas_file,
-                     self.saas_files,
+                     desired_state_specs,
                      self.thread_pool_size,
                      ri=ri)
 
-    def populate_desired_state_saas_file(self, saas_file, ri):
+    def init_populate_desired_state_specs(self, saas_file):
+        specs = []
         saas_file_name = saas_file['name']
-        logging.debug(f"populating desired state for {saas_file_name}")
         github = self._initiate_github(saas_file)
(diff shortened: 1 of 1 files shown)


utils/saasherder.py
```diff
@@ -87,6 +89,20 @@ def _collect_namespaces(self):
...
+        # each thread can use 1 thread internally.
+        # 
+        # each namespace is in fact a target,
+        # so we can use it to calculate.
+        thread_pool_size = int(
+            self.thread_pool_size /
+            len(self.namespaces)
+        )
+        return max(thread_pool_size, 1)
+
```
apahim, line 105: very efficient usage of the availables threads

utils/saasherder.py
```diff
@@ -245,35 +261,46 @@ def _collect_images(self, resource):
...
-                    error = True
-                    logging.error(
-                        f"{error_prefix} Image does not exist: {image}")
-                    continue
-            except Exception:
-                error = True
-                logging.error(f"{error_prefix} Image is invalid: {image}")
-                continue
+        images_list = threaded.run(self._collect_images, resources,
+                                   self.available_thread_pool_size)
```
apahim, line 296: did you test this without threading? just curious to see how much faster this is with the threaded...

utils/saasherder.py
maorfr: :heart:

utils/saasherder.py
maorfr: this actually doesn't add much :)

apahim: I'd expect that :)

PR #858, opened 2020-06-22
[WiP] Removing the forks check

apahim:
The forks check only takes "master" into account, which is not really usefull
considering that any leaked key would be placed in a fork feature branch AND we
already scan "master" on the main repo.

This will reduce the number of scanned repos from 6820 to 273 as of now.


apahim:
have to check whether the git-secrets really does not check all the branches.

apahim:
indeed it checks all the braches

PR #866, opened 2020-06-24
[saas-file-owners] can't lgtm yourself

maorfr:
unless you are a single owner

### DIFF
--- a/reconcile/saas_file_owners.py
+++ b/reconcile/saas_file_owners.py
@@ -197,6 +197,7 @@ def run(gitlab_project_id, gitlab_merge_request_id, dry_run=False,
             gitlab_merge_request_id, 'approved')
         return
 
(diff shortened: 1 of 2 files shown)


reconcile/saas_file_owners.py
```diff
@@ -205,6 +206,10 @@ def run(gitlab_project_id, gitlab_merge_request_id, dry_run=False,
         # check for a lgtm by an owner of this app
         saas_file_name = diff['saas_file_name']
         saas_file_owners = owners.get(saas_file_name)
+        if author in saas_file_owners:
+            # allow a single owner to lgtm themselves for now
```
apahim, line 210: for now?

maorfr: for now :)
https://github.com/app-sre/qontract-reconcile/blob/master/utils/saasherder.py#L60

PR #875, opened 2020-06-26
Just check for qontract.recycle to be the string "true"

rporres:
It can be `None` if unset, so the checks will become a bit convoluted. It
is valuable to keep the check to be the string `"true"` to give
information in case of error.

### DIFF
--- a/utils/oc.py
+++ b/utils/oc.py
@@ -212,8 +212,6 @@ class OC(object):
         qontract_recycle = dep_annotations.get('qontract.recycle')
         if qontract_recycle is True:
             raise RecyclePodsInvalidAnnotationValue('should be "true"')
-        if not isinstance(qontract_recycle, str):
-            raise RecyclePodsInvalidAnnotationValue('should be a string')
         if qontract_recycle != 'true':
             logging.debug(['skipping_pod_recycle_no_annotation',
                            namespace, dep_kind])


utils/oc.py
```diff
@@ -212,8 +212,6 @@ def recycle_pods(self, dry_run, namespace, dep_kind, dep_resource):
         qontract_recycle = dep_annotations.get('qontract.recycle')
         if qontract_recycle is True:
             raise RecyclePodsInvalidAnnotationValue('should be "true"')
-        if not isinstance(qontract_recycle, str):
-            raise RecyclePodsInvalidAnnotationValue('should be a string')
```
apahim, line 216: Yeah.. python was born to be duck-typed... What about one check to rule them all?
```python
qontract_recycle = dep_annotations.get('qontract.recycle')
if str(qontract_recycle).lower() != 'true':
    return
```

rporres: That's more or less what we had before. I think @maorfr wanted to make sure we warn people that they should not use the boolean `true` but the string `"true"`. He did that in https://github.com/app-sre/qontract-reconcile/pull/874 but went too far 😬 

I like the idea of being nice and warning people as much as possible

apahim: yes..but before we had no handler for `True`... and if users said `qontract.recycle: true` it is quite clear that they mean the recycle to happen. Enforcing types is a waste of tentant cycles imo, but I'm not super strict here :)

rporres: I don't know about the reasons to make it a string, honestly. But he went quite far to make sure it is, so I'd rather keep it that way until he can give more context :-D

maorfr: a ConfigMap will also fail to apply if an annotation value is boolean IIRC

apahim: then forget what I said :)

PR #876, opened 2020-06-27
[openshift-namespaces] performance improvement (and to all openshift integrations as well)

maorfr:
This PR adds 2 things:
1. use threads to initiate OC clients in OC_Map
2. add option to initiate Projects when initiating an oc client
3. if projects are initiated, use it to check if a project exists

This speeds up the openshift-namespaces integration by 80% according to local testing.

It also speeds up many other openshift integrations (not by a lot though).


utils/oc.py
```diff
@@ -379,23 +390,25 @@ class OC_Map(object):
...
         elif clusters:
-            for cluster_info in clusters:
-                self.init_oc_client(cluster_info)
+            threaded.run(self.init_oc_client, clusters, self.thread_pool_size)
         elif namespaces:
-            for namespace_info in namespaces:
-                cluster_info = namespace_info['cluster']
-                self.init_oc_client(cluster_info)
+            clusters = [ns_info['cluster'] for ns_info in namespaces]
+            threaded.run(self.init_oc_client, clusters, self.thread_pool_size)
```
apahim, line 411: the `threaded.run()` can be called after the `if` block to avoid duplication. That will remove the need for `elif clusters:`.

maorfr: i prefer to keep all the cases covered explicitly. wdyt?

apahim: sure

PR #888, opened 2020-07-01
[gitlab-housekeeping] merge and rebase by label priority

maorfr:
in addition to merging in reversed order (FIFO), we want to merge according to label priority.

the most important label is `approved` which indicates a tenant is self-servicing a merge.
`automerge` comes next to allow our automations to accomplish their goals. `lgtm` comes last since this is a manual process anyway.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -8,7 +8,7 @@ import reconcile.queries as queries
 from utils.gitlab_api import GitLabApi
 
 
-MERGE_LABELS = ['lgtm', 'automerge', 'approved']
+MERGE_LABELS_PRIORITY = ['approved', 'automerge', 'lgtm']
 HOLD_LABELS = ['awaiting-approval', 'blocked/bot-access',
                'do-not-merge/hold', 'do-not-merge/pending-review']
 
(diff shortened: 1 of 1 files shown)


reconcile/gitlab_housekeeping.py
```diff
@@ -83,88 +83,90 @@ def handle_stale_items(dry_run, gl, days_interval, enable_closing, item_type):
...
 
-def is_good_to_merge(labels):
-    return any(l in MERGE_LABELS for l in labels) and \
+def is_good_to_merge(merge_label, labels):
+    return merge_label in labels and \
         not any(l in HOLD_LABELS for l in labels)
 
 
 def rebase_merge_requests(dry_run, gl, rebase_limit):
     mrs = gl.get_merge_requests(state='opened')
```
apahim, line 92: Does it make sense to move the `get_merge_requests` inside the `for merge_label in MERGE_LABELS_PRIORITY` and [filter per label](https://docs.gitlab.com/ee/api/merge_requests.html#list-merge-requests)? That would avoid processing MRs without the current label in the loop, but would also generate additional calls to the API to list MRs. Not sure it pays off.

reconcile/gitlab_housekeeping.py
```diff
@@ -83,88 +83,90 @@ def handle_stale_items(dry_run, gl, days_interval, enable_closing, item_type):
...
-        if last_pipeline_result != 'success':
-            continue
-
-        logging.info(['merge', gl.project.name, mr.iid])
-        if not dry_run and merges < merge_limit:
-            mr.merge()
-            if rebase:
-                return
-            merges += 1
+    for merge_label in MERGE_LABELS_PRIORITY:
```
apahim, line 127: same here

reconcile/gitlab_housekeeping.py
maorfr: right now it's only doing internal processing inside the loop, so i think we can keep things as they are to save some API calls

reconcile/gitlab_housekeeping.py
maorfr: same here :)

reconcile/gitlab_housekeeping.py
apahim: There's the `gl.project.commits.list`

PR #903, opened 2020-07-03
Adds dump-schemas functionality

jmelis:
The `--dump-schemas` option will dump all the schemas traversed by the gql api to a file. This is necessary in order to verify what schemas does an integration depend on.

A big refactor has been needed in this PR. The entrypoints for all integrations is `def run(...)` with an arbitrary list of arguments defined per integration. The contract has now changed, so that the entrypoint is always `def run(dry_run, ...)`. I needed to do this because it was necessary in order to pass the context object to the `run_integration` method. With this change we can now add additional global options, like `--dump-schemas`.

This functionality depends on this PR:
https://github.com/app-sre/qontract-server/pull/81

EDIT: I have added an additional functionality. When the integration starts it will query app-interface for the allowed schemas, and if it tries to perform a query that depends on an unlisted schema, the integration will immediately exit with a defined exit code.

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import os
 import sys
@@ -112,6 +113,15 @@ def dry_run(function):
 
     function = click.option('--dry-run/--no-dry-run',
                             default=False,
+                            help=help_msg)(function)
+    return function
+
+
+def dump_schemas(function):
+    help_msg = ('Dump the list of used schemas to a file. It will not'
+                'overwrite the target file')
+
+    function = click.option('--dump-schemas',
                             help=help_msg)(function)
     return function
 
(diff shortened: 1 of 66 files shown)


reconcile/cli.py
```diff
@@ -265,6 +275,11 @@ def run_integration(func_container, ctx, *args):
         else:
             raise e
 
+    dump_schemas_file = ctx.get('dump_schemas')
+    if dump_schemas_file:
+        with open(dump_schemas_file, 'w') as f:
+            f.write(json.dumps(gql.get_called_schemas()))
+
```
apahim, line 282: Is this temporary? what's next after you have the file?

jmelis: No, it's not temporary. The idea is that if you want to know what schemas an integration depends on you add `--dump-schemas <file>` and that will create that file.

I plan to run all integrations with dry-run mode on the pr check for this repository. And compare the content of each file with the list of schemas defined per integration in the integration schema.

jmelis:
@rporres the problem is that the integration already sends output to stdout (the list of actions to take, or list of actions taken). If I also send it to stdout it will mix with that output. Suggestions?

jmelis:
@rporres @apahim: can you take a look at my last commit, I have added some new functionality (described in the `EDIT` section of the description above)

jmelis:
Note: I do want to remove the `TODO` before merging this PR, I'm just preparing all integrations first.

jmelis:
Dropped in favour of:
- https://github.com/app-sre/qontract-reconcile/pull/906
- https://github.com/app-sre/qontract-reconcile/pull/907

PR #907, opened 2020-07-05
validate integrations and schemas (introduces a new integration: integrations-validator) 

jmelis:
The goal of this PR is twofold:

- ensure that all integrations are defined in App-Interface
- blocks integration from querying schemas if the schema is not defined in the corresponding App-Interface integrtaion file

In order to achieve these two goals the following changes are included in this PR:

**Integration**

> `integrations-validator`: Ensures all integrations are defined in App-Interface.

This integration programmatically iterates through all the integration click commands, and ensures that they exist in App-Interface

**Mandatory schema validation**

Integrations now run with the enabled feature of validating schemas: `--validate-schemas` which is set to True by default.

If necessary this can be disabled with the option `--no-validate-schemas`.

Also, in order to make the developer life easier, all queried schemas will be logged with `DEBUG` level, so developers can run their integrations with `--no-validate-schemas --log-level=DEBUG`.

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -122,7 +122,7 @@ def validate_schemas(function):
     help_msg = 'Fail integration if it queries forbidden schemas'
 
     function = click.option('--validate-schemas/--no-validate-schemas',
-                            default=False,
+                            default=True,
                             help=help_msg)(function)
     return function
 
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -74,7 +74,8 @@ import reconcile.user_validator
 
 from reconcile.status import ExitCodes
 
-from utils.gql import GqlApiError
+from utils.gql import (GqlApiError, GqlApiErrorForbiddenSchema,
+                       GqlApiIntegrationNotFound)
 from utils.aggregated_list import RunnerException
 from utils.binary import binary
 from utils.environ import environ
@@ -111,6 +112,15 @@ def dry_run(function):
                 'that would be performed, without executing them.')
 
     function = click.option('--dry-run/--no-dry-run',
+                            default=False,
+                            help=help_msg)(function)
+    return function
+
+
+def validate_schemas(function):
+    help_msg = 'Fail integration if it queries forbidden schemas'
+
+    function = click.option('--validate-schemas/--no-validate-schemas',
                             default=False,
                             help=help_msg)(function)
     return function
@@ -243,8 +253,13 @@ def enable_rebase(**kwargs):
 
 
 def run_integration(func_container, ctx, *args):
-    integration_name = func_container.QONTRACT_INTEGRATION.replace('_', '-')
-    unleash_feature_state = get_feature_toggle_state(integration_name)
+    int_name = func_container.QONTRACT_INTEGRATION.replace('_', '-')
+
+    gql.init_from_config(sha_url=ctx['gql_sha_url'],
+                         integration=int_name,
+                         validate_schemas=ctx['validate_schemas'])
+
+    unleash_feature_state = get_feature_toggle_state(int_name)
     if not unleash_feature_state:
         logging.info('Integration toggle is disabled, skipping integration.')
         sys.exit(ExitCodes.SUCCESS)
@@ -256,6 +271,12 @@ def run_integration(func_container, ctx, *args):
     except RunnerException as e:
         sys.stderr.write(str(e) + "\n")
         sys.exit(ExitCodes.ERROR)
+    except GqlApiIntegrationNotFound as e:
+        sys.stderr.write(str(e) + "\n")
+        sys.exit(ExitCodes.INTEGRATION_NOT_FOUND)
+    except GqlApiErrorForbiddenSchema as e:
+        sys.stderr.write(str(e) + "\n")
+        sys.exit(ExitCodes.FORBIDDEN_SCHEMA)
     except GqlApiError as e:
         if '409' in str(e):
             logging.info(f'Data changed during execution. This is fine.')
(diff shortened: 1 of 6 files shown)


utils/gql.py
```diff
@@ -13,23 +13,58 @@ class GqlApiError(Exception):
     pass
 
 
+class GqlApiIntegrationNotFound(Exception):
+    def __init__(self, integration):
+        super().__init__(f"integration not found: {integration}")
```
maorfr, line 18: can we add a hint here about what can be done? where is the integration missing from and what should a user do?

utils/gql.py
```diff
@@ -13,23 +13,58 @@ class GqlApiError(Exception):
...
+
+
+class GqlApiErrorForbiddenSchema(Exception):
+    def __init__(self, schema):
+        super().__init__(f"forbidden schema: {schema}")
```
maorfr, line 23: same here

utils/gql.py
```diff
@@ -13,23 +13,58 @@ class GqlApiError(Exception):
...
 class GqlGetResourceError(Exception):
     def __init__(self, path, msg):
         super(GqlGetResourceError, self).__init__(
             "error getting resource from path {}: {}".format(path, str(msg))
         )
 
 
+INTEGRATIONS_QUERY = """
+{
+    integrations_v1 {
```
maorfr, line 35: add alias?

reconcile/cli.py
```diff
@@ -883,3 +913,10 @@ def gitlab_fork_compliance(ctx, gitlab_project_id, gitlab_merge_request_id,
     run_integration(reconcile.gitlab_fork_compliance, ctx.obj,
                     gitlab_project_id, gitlab_merge_request_id,
                     gitlab_maintainers_group)
+
+
+@integration.command()
+@click.pass_context
+def integrations_validator(ctx):
+    run_integration(reconcile.integrations_validator, ctx.obj,
+                    reconcile.cli.integration.commands.keys())
```
maorfr, line 922: this is so cool!

reconcile/integrations_validator.py
```diff
@@ -0,0 +1,17 @@
...
+QONTRACT_INTEGRATION = 'integrations-validator'
+
+
+def run(dry_run, integration_commands):
+    desired_integrations = [i['name'] for i in queries.get_integrations()]
+
+    missing = set(integration_commands) - set(desired_integrations)
+
+    for integration in missing:
+        logging.info(['missing_integration', integration])
```
maorfr, line 14: logging.error?

reconcile/queries.py
```diff
@@ -102,6 +102,11 @@ def get_credentials_requests():
     return gqlapi.query(CREDENTIALS_REQUESTS_QUERY)['credentials_requests']
 
 
+def get_integrations():
+    gqlapi = gql.get_api()
+    return gqlapi.query(gql.INTEGRATIONS_QUERY)['integrations']
```
maorfr, line 107: an interesting exception. i can't think of anything better to do here.

reconcile/cli.py
```diff
@@ -243,8 +254,21 @@ def f(function):
...
-    unleash_feature_state = get_feature_toggle_state(integration_name)
+    try:
+        int_name = func_container.QONTRACT_INTEGRATION.replace('_', '-')
+    except AttributeError:
+        sys.stderr.write("Integration missing QONTRACT_INTEGRATION.\n")
+        sys.exit(ExitCodes.ERROR)
+
+    try:
+        gql.init_from_config(sha_url=ctx['gql_sha_url'],
+                             integration=int_name,
```
maorfr, line 265: this is super smart!

utils/gql.py
```diff
@@ -2,34 +2,83 @@
...
 
 class GqlApi(object):
-    def __init__(self, url, token=None):
+    _valid_schemas = None
+
+    def __init__(self, url, token=None, int_name=None, validate_schemas=False):
         self.url = url
         self.token = token
+        self.integration = int_name
+        self.validate_schemas = validate_schemas
```
maorfr, line 64: maybe we can do:
```
self.validate_schemas = validate_schemas and int_name
```

utils/gql.py
```diff
@@ -2,34 +2,83 @@
...
 
         self.client = GraphQLClient(self.url)
 
         if token:
             self.client.inject_token(token)
 
+        if int_name:
```
maorfr, line 71: and use it here:
```
if self.validate_schemas:
```

utils/gql.py
```diff
@@ -43,6 +92,17 @@ def query(self, query, variables=None):
 
         result = json.loads(result_json)
 
+        # show schemas if log level is debug
+        for s in result['extensions']['schemas']:
+            logging.debug(['schema', s])
+
+        # we need to test self._valid_schemas to let through the query that
+        # retrieves the schemas in the __init__ method
+        if self.validate_schemas and self._valid_schemas:
```
maorfr, line 101: and here as well:
```
if self.validate_schemas:
```

reconcile/queries.py
jmelis: yes, I had it the other way round, but it created a circular dependency

utils/gql.py
jmelis: I like it!

jmelis: well, on second thought this is not as easy, I'd rather raise an exception if you try to validate schemas without definiing the integration name?

utils/gql.py
jmelis: the problem here is that you need to let through the first query

PR #914, opened 2020-07-08
[WIP][openshift-base] remove comparison by desired object

maorfr:
in the base of this logic stands the assumption that if a desired resource does not contain a field, the value of that field in the current resource does not matter.

this was done with the intention of supporting resources with multiple default values, such as Deployments.

this does not work for a situation where a key is removed.

merging this will cause quite a few resources to be re-applied, and maybe even to be continuously re-applied. we will need to fix it in following PRs to improve the comparison logic in openshift_resouce.canonicalize.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -268,15 +268,6 @@ def realize_data(dry_run, oc_map, ri,
                         ).format(cluster, namespace, resource_type, name)
                         logging.info(msg)
 
-                    # don't apply if resources match
-                    elif d_item == c_item:
-                        msg = (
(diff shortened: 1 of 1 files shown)


jzelinskie:
Can we create a follow up JIRA ticket for cleaning this up and link it to this PR?

jzelinskie:
Another option is to keep this block but change the conditional to the following psuedocode.

```py
if my_res == cluster_res and \
   len(my_res.all_applied_keys()) == len(cluster_res.previously_applied().all_applied_keys()):
   ...
```

PR #915, opened 2020-07-08
[WIP] Add email notification

violetyao:


utils/gitlab_api.py
```diff
@@ -188,6 +188,60 @@ def create_app_interface_reporter_mr(self, reports, email_schema,
...
+                'content': notification['content'],
+            }
+            for notification in notifications
+        ]
+
+        self.create_commit(branch_name, commit_message, actions)
+
+        msg = 'add email notification'
+        email_path = f"data{notification_path}/emails/{branch_name}.yml"
+        email = {
```
jfchevrette, line 225: the comments in this dict seem a bit off. can you review and remove the comments or parts that are not needed?

utils/gitlab_api.py
```diff
@@ -188,6 +188,46 @@ def create_app_interface_reporter_mr(self, reports, email_schema,
...
+        ts = now.strftime("%Y%m%d%H%M%S")
+        isodate = now.isoformat()
+        short_date = now.strftime('%Y-%m-%d')
+
+        branch_name = 'app-interface-notificator-{}'.format(ts)
+        commit_message = '[{}] notification for {}'.format(
+            prefix, isodate
+        )
+
+        self.create_branch(branch_name, target_branch)
```
apahim, line 211: It's a bit confusing that `create_branch()` takes a `source_branch` argument and for that you provide `target_branch`. I mean, this line could well be written like this: `self.create_branch(new_branch=branch_name, source_branch=target_branch)`. Maybe renaming `target_branch` to `main_branch` or just `branch`?
Anyway, maybe it's just me :)

utils/gitlab_api.py
```diff
@@ -188,6 +188,46 @@ def create_app_interface_reporter_mr(self, reports, email_schema,
...
+
+        msg = 'add email notification'
+        email_path = f"data{notification_path}/emails/{branch_name}.yml"
```
apahim, line 214: Using the `branch_name` to name the email file is a sign that the variable holds more than the branch name :)
Maybe rename `branch_name` to `email_id`?

By the way, you're using f-strings here! You should definetly pick one between f-strings and `format()` so it's all consistent.

Additionally, I'm always afraid of manually manipulating paths, spceially with slashs coming from different places. You forget the slash in the `notification_path` and the thing is broken. My suggestion here is to use either `os.path.join()` or `pathlib.Path()`.

utils/gitlab_api.py
```diff
@@ -188,6 +188,46 @@ def create_app_interface_reporter_mr(self, reports, email_schema,
...
+        email = {
+            '$schema': email_schema,
+            'labels': {},
+            'name': branch_name,
+            'subject': f"[{prefix} {notification['notification_type']}]" +
```
apahim, line 219: The plus sign is not needed here.
Also, it seems like there's a missing white-space after the closing square bracket.
Please align the string continuation line to the beginning of the string in favor or readability.

utils/gitlab_api.py
```diff
@@ -188,6 +188,46 @@ def create_app_interface_reporter_mr(self, reports, email_schema,
...
+        notification_path="/app-interface",
+        email_schema="/app-interface/app-interface-email-1.yml"
+    ):
+        # labels = ['automerge']
+        labels = ['do-not-merge/hold']
+        prefix = 'app-interface-notificator'
+        target_branch = 'master'
+
+        now = datetime.now()
+        ts = now.strftime("%Y%m%d%H%M%S")
```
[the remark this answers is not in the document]
violetyao: Good point! I'll include notification["short_description"] in branch_name in addition to datatime info.

utils/gitlab_api.py
violetyao: Yeah, I agree! It is a confusing. I renamed target_branch to main_branch.

violetyao: * a bit

utils/gitlab_api.py
violetyao: Yeah! I renamed branch_name to email_id. 

Good catch! I'll stick to f-strings then.

I see. Changed to os.path.join(). Thanks for pointing this out! ;D

utils/gitlab_api.py
violetyao: Gotcha!

violetyao:
> Hi @violetyao, awesome work here! Thank you for this.
> I left some comments, most of them being just suggestions. Feel free to just ignore whatever you don't agree with.
> Looking forward to having this available :)

Thank you, Amador! I love your comments. ;D

utils/gitlab_api.py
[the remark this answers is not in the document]
jfchevrette: the short description may be too long for the branch name, it may also contain spaces or characters that are not allowed in branch names.

I guess the branch name could be something like `notificator-<unixtime>` where `<unixtime>` is the result of `int(time.time())`. That would make it short and unique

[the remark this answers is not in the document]
violetyao: I see. I didn't think about that description may contain spaces/ unallowed chars. I will change the branch name to `notificator-<unixtime>`. Thank you, JF!

maorfr:
hey @violetyao!

this looks great! thanks for working on this.
i would like to point out something that you should consider -

the same message may appear in the queue over and over again as the mechanism adding messages to the queue may not be aware that the message was read, an MR was created and is waiting to be merged. not saying it will happen here, but it may.

to handle it, we need to check if the MR already exists before creating it (again?). to do that, you can use the `mr_exists` method that takes the MR title as the unique identifier. this means that the title should be unique (per incident?). from what I saw, the title is the commit message, which contains a formatted version of time.now, which is not unique.

hope I didn't miss anything and that the use case I described is already covered.

again, nicely done!

violetyao:
> hey @violetyao!
> 
> this looks great! thanks for working on this.
> i would like to point out something that you should consider -
> 
> the same message may appear in the queue over and over again as the mechanism adding messages to the queue may not be aware that the message was read, an MR was created and is waiting to be merged. not saying it will happen here, but it may.
> 
> to handle it, we need to check if the MR already exists before creating it (again?). to do that, you can use the `mr_exists` method that takes the MR title as the unique identifier. this means that the title should be unique (per incident?). from what I saw, the title is the commit message, which contains a formatted version of time.now, which is not unique.
> 
> hope I didn't miss anything and that the use case I described is already covered.
> 
> again, nicely done!

It is a great point, Maor! Thank you for pointing this out and explaining it so well. < 3 I will do something like        
```
 if self.mr_exists(title):
      return
```
and rename the title to include both incident and time info. 

Thanks again!!

maorfr:
you are most welcome @violetyao!
just to make sure -
if you include the time info the title will not be unique...

violetyao:
> you are most welcome @violetyao!
> just to make sure -
> if you include the time info the title will not be unique...

I think I get it now. If we include time info, more merge requests about a same incident will have different titles from the first merge request and thus become duplicates. The title should be incident-unique. Thank you!

jfchevrette:
@violetyao I think we're ready to merge this. You will have to rebase and fix conflicts before we can merge. Let me know if you wait to pair on this

violetyao:
> @violetyao I think we're ready to merge this. You will have to rebase and fix conflicts before we can merge. Let me know if you wait to pair on this

Rebased! Thank you JF!

PR #918, opened 2020-07-10
[github] do not depend on automationToken

maorfr:
currently, to run the github integration we need an automationToken so we can initialize an oc client for each cluster.

the reason we are initializing an oc client for each cluster (oc_map) is because the integration relies on logic from other integrations (openshift-rolebindings and openshift-groups) which REQUIRE an oc client.
the github integration, however, does not care about this. it will add users to github teams for cluster access, even if that cluster does not have an automationToken.

this makes sense because in the cluster on-boarding process we first want to allow access for users to the console, and only after that we are able to create the bot SA to add the automationToken.

tl;dr - another chicken/egg situation solved.

### DIFF
--- a/reconcile/openshift_rolebindings.py
+++ b/reconcile/openshift_rolebindings.py
@@ -110,7 +110,7 @@ def fetch_desired_state(ri, oc_map):
 
         for permission in permissions:
             cluster = permission['cluster']
-            if not oc_map.get(cluster):
+            if oc_map and not oc_map.get(cluster):
                 continue
             for user in users:
                 # used by openshift-users and github integrations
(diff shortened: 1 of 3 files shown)


reconcile/openshift_rolebindings.py
```diff
@@ -110,7 +110,7 @@ def fetch_desired_state(ri, oc_map):
 
         for permission in permissions:
             cluster = permission['cluster']
-            if not oc_map.get(cluster):
+            if oc_map and not oc_map.get(cluster):
```
jmelis, line 113: is this correct? if `oc_map` is none we will not enter the `if`.

Perhaps you meant:
```
if not oc_map or (oc_map and not oc_map.get(cluster)):
```

maorfr: I mean to not enter the if when oc_map is none

PR #921, opened 2020-07-12
[openshift-saas-deploy] add support for raw manifests in directory

maorfr:
covers https://issues.redhat.com/browse/APPSRE-1753

this PR adds the ability to specify the `directory` provider in a saas file to indicate that the resources to apply are raw manifests in a directory (applies all resources).

### DIFF
--- a/utils/saasherder.py
+++ b/utils/saasherder.py
@@ -122,18 +122,49 @@ class SaasHerder():
         path = options['path']
         ref = options['ref']
         github = options['github']
+        html_url = os.path.join(url, 'blob', ref, path)
+        content = None
         if 'github' in url:
             repo_name = url.rstrip("/").replace('https://github.com/', '')
             repo = github.get_repo(repo_name)
             f = repo.get_contents(path, ref)
-            return f.decoded_content, f.html_url
+            content = f.decoded_content
         elif 'gitlab' in url:
             if not self.gitlab:
                 raise Exception('gitlab is not initialized')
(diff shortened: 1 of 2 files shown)


utils/saasherder.py
```diff
@@ -122,18 +122,49 @@ def _get_file_contents(self, options):
...
+                resources.append(resource)
+        elif 'gitlab' in url:
+            if not self.gitlab:
+                raise Exception('gitlab is not initialized')
+            project = self.gitlab.get_project(url)
+            for f in project.repository_tree(path=path.lstrip('/'), ref=ref):
+                file_contents = \
+                    project.files.get(file_path=f['path'], ref=ref)
+                resource = yaml.safe_load(file_contents.decode())
+                resources.append(resource)
```
apahim, line 165: Why don't we process them? The use case here is for non-templates? Won't the Deployment be in the same directory?

maorfr: the intention is not to process or apply any templating for such files.

in order to template, the resources should be in an openshift template or a helm chart, or anything that has a defined templating method. for the `directory` provider, just simple plain resources.

makes sense?

apahim: Meaning this is something to be specified in addition to the `openshift-template` provider, not instead of. Got it.

maorfr: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/6177 for more info

PR #927, opened 2020-07-14
[openshift-saas-deploy] skip image check if upstream job is running

maorfr:
when a target specifies an `upstream` job, it is mostly because that upstream job is building an image required for the deployment job.

if we trigger the deployment job while the upstream job is running, the deployment job will wait until the build job is complete. this is thanks to [this block-upstream field](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/resources/jenkins/global/templates.yaml#L211).

this means that it is ok to merge a MR while the upstream job is running, which means that it's pr-check should pass.

this change implements this behavior.

### DIFF
--- a/reconcile/openshift_saas_deploy.py
+++ b/reconcile/openshift_saas_deploy.py
@@ -4,6 +4,7 @@ import logging
 
 import reconcile.queries as queries
 import reconcile.openshift_base as ob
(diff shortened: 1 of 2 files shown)


BumbleFeng:
LGTM

PR #930, opened 2020-07-15
[ocm-clusters] create cluster

maorfr:
This PR adds the ability for the ocm-clusters integration to provision clusters.

Provision only! no deletions, no updates, provision only.

### DIFF
--- a/utils/ocm.py
+++ b/utils/ocm.py
@@ -62,8 +62,7 @@ class OCM(object):
             'spec': {
                 'provider': cluster['cloud_provider']['id'],
                 'region': cluster['region']['id'],
-                'major_version':
-                    int(cluster['openshift_version'].split('.')[0]),
+                'version': cluster['openshift_version'],
                 'multi_az': cluster['multi_az'],
                 'nodes': cluster['nodes']['compute'],
                 'instance_type':
@@ -101,6 +100,7 @@ class OCM(object):
             'region': {
                 'id': cluster_spec['region']
             },
+            'openshift_version': cluster_spec['version'],
             'multi_az': cluster_spec['multi_az'],
             'nodes': {
                 'compute': cluster_spec['nodes'],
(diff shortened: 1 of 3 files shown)


utils/ocm.py
```diff
@@ -77,6 +77,47 @@ def _get_cluster_ocm_spec(cluster):
...
+            },
+            'load_balancer_quota': cluster_spec['load_balancers'],
+            'network': {
+                'machine_cidr': cluster_network['vpc'],
+                'service_cidr': cluster_network['service'],
+                'pod_cidr': cluster_network['pod'],
+            }
+        }
+
+        self._post(api, ocm_spec)
```
aditya-konarde, line 119: There's some cases where it would be useful to have more info than just the status code. Could you please also add the response body to the _post function, based on the current log level?

aditya-konarde: Mostly thinking of the following cases:
- Not enough quota on our org
- Wrong spec in app-interface, thus an incorrect payload gets sent to the create API
- The OCM API changes and we're missing parameters (maybe)

utils/ocm.py
```diff
@@ -77,6 +77,47 @@ def _get_cluster_ocm_spec(cluster):
...
+        :param name: name of the cluster
+        :param cluster: a dictionary representing a cluster desired state
+
+        :type name: string
+        :type cluster: dict
+        """
+        api = f'/api/clusters_mgmt/v1/clusters'
+        cluster_spec = cluster['spec']
+        cluster_network = cluster['network']
+        ocm_spec = {
```
aditya-konarde, line 93: We need to be able to select if a cluster is public/private here. Looking at https://api.openshift.com/#/default/post_api_clusters_mgmt_v1_clusters for the spec

aditya-konarde:
cluster-privacy

utils/ocm.py
maorfr: done!

jfchevrette:
Lgtm!

utils/ocm.py
maorfr: done!

PR #934, opened 2020-07-16
[terraform-vpc-peerings] use default provider for account-vpc accepter

maorfr:
covers https://issues.redhat.com/browse/APPSRE-2234

when the vpc peering is between clusters, we need to use the accepter's alias provider (the cluster's aws account id).
when the vpc peering is between a cluster and an aws account, we need to use the default provider (the aws account).


maorfr:
going to merge this as there are more MRs which may depend on this.

PR #935, opened 2020-07-16
[terraform-vpc-peerings] add routes capability

maorfr:
covers https://issues.redhat.com/browse/APPSRE-1861

this PR adds the ability to manage routes in existing route tables in a cluster's aws account.
that cluster can be a requester, an accepter or the requester in an `account-vpc` connection. it all works.

issues:
- for now, it only works with creating new routes (i am still unable to import all the existing routes which were created manually).
- since we don't have permissions to delete routes, we can't "unmanage" routes (this will involve removing routes from the state).

for that reason i have added a selective `manageRoutes` field to be used.

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -320,6 +320,7 @@ CLUSTERS_QUERY = """
       connections {
         name
         provider
+        manageRoutes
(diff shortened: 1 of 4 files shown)


jfchevrette:
LGTM

PR #941, opened 2020-07-17
Allow users to provide alternate database name

tparikh:
Allow users to provide database name in cases where initial database name was not provided when rds instance was created. Otherwise these databases will have `db.name` set to empty string in such cases.


mmclanerh:
LGTM, other than the build check barfed.

maorfr:
-1 to this proposal. This is coding technical debt. We should instead pay it off.

PR #944, opened 2020-07-19
allow running openshift-resources based integrations per cluster and namespace

maorfr:
this PR adds two new flags to be passed to the following integrations:
- openshift-resources
- openshift-vault-secrets
- openshift-routes

the flags are:
- `cluster-name` - only run for a cluster with this name
- `namespace-name` - only run for namespaces with this name

this allows us to run integrations for any particular namespace we choose.
or for all namespaces with that name (think hive).
or for all namespaces in a cluster (think hive).

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -512,20 +512,13 @@ def fetch_data(namespaces, thread_pool_size, internal, use_jump_host):
 def filter_namespaces_by_cluster_and_namespace(namespaces,
                                                cluster_name,
                                                namespace_name):
-    filtered_namespaces = []
     if cluster_name:
-        filtered_namespaces.extend(
-            [n for n in namespaces
-             if n['cluster']['name'] == cluster_name]
-        )
-    elif namespace_name:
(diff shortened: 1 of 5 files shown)


reconcile/openshift_resources_base.py
```diff
@@ -509,6 +509,25 @@ def fetch_data(namespaces, thread_pool_size, internal, use_jump_host):
...
+    filtered_namespaces = []
+    if cluster_name:
+        filtered_namespaces.extend(
+            [n for n in namespaces
+             if n['cluster']['name'] == cluster_name]
+        )
+    elif namespace_name:
+        filtered_namespaces.extend(
+            [n for n in namespaces
+             if n['name'] == namespace_name]
```
apahim, line 524: When both are provided I'd expect either an error message or to have them both used.
One might want to specify cluster AND namespace.... in this case, you're ignoring the namespace.

maorfr: what about now?

apahim: yep, looks good, thanks!

PR #947, opened 2020-07-20
reduce downlad lambda zip times

BumbleFeng:


utils/terrascript_client.py
```diff
@@ -105,6 +107,18 @@ def __init__(self, integration, integration_prefix,
         self.uids = {a['name']: a['uid'] for a in filtered_accounts}
         self.default_regions = {a['name']: a['resourcesDefaultRegion']
                                 for a in filtered_accounts}
+        self.logtoes_zip = self.download_logtoes_zip(LOGTOES_RELEASE)
+
+    def download_logtoes_zip(self, release_url):
+        r = requests.get(GH_BASE_URL + '/' + release_url)
+        if r.status_code == 404:
+            raise ValueError(f"release is not found")
```
maorfr, line 115: you can use raise_for_status

utils/terrascript_client.py
```diff
@@ -105,6 +107,18 @@ def __init__(self, integration, integration_prefix,
...
+        zip_url = r.json()['assets'][0]['browser_download_url']
+        zip_file = '/tmp/LogsToElasticsearch-'+r.json()['tag_name']+'.zip'
```
maorfr, line 117: we should probably only get r.json() once to avoid accessing it multiple times

utils/terrascript_client.py
```diff
@@ -105,6 +107,18 @@ def __init__(self, integration, integration_prefix,
...
+        if not path.exists(zip_file):
+            r = requests.get(zip_url)
+            open(zip_file, 'wb').write(r.content)
```
maorfr, line 120: why do we need to use the local file system? can we keep the data in-memory?

utils/terrascript_client.py
```diff
@@ -1606,15 +1620,17 @@ def populate_tf_resource_cloudwatch(self, resource, namespace_info):
             tf_resources.append(data('aws_elasticsearch_domain',
                                      es_identifier, **es_domain))
 
-            zip_url = common_values.get('zip_url', LOGTOES_ZIP)
-            r = requests.get(zip_url)
-            open('/tmp/LogsToElasticsearch.zip', 'wb').write(r.content)
+            release_url = common_values.get('release_url', LOGTOES_RELEASE)
```
maorfr, line 1623: i like this! you are allowing users to override the default value. nice!

utils/terrascript_client.py
```diff
@@ -49,8 +51,8 @@
                                aws_cloudwatch_log_subscription_filter,
                                aws_acm_certificate)
 
-LOGTOES_ZIP = 'https://github.com/app-sre/logs-to-elasticsearch-lambda' + \
-              '/releases/latest/download/LogsToElasticsearch.zip'
+GH_BASE_URL = os.environ.get('GITHUB_API', 'https://api.github.com')
```
maorfr, line 54: nice. this way we'll go through github-mirror.

utils/terrascript_client.py
BumbleFeng: The default value will be `repos/app-sre/logs-to-elasticsearch-lambda/releases/latest`, while the user could set to `repos/app-sre/logs-to-elasticsearch-lambda/releases/tags/1.0.2` to pin source code to a certain tag, just as we pin saas file to a certain hash.

utils/terrascript_client.py
BumbleFeng: Tried but failed. Seems terraform only support local file. https://www.terraform.io/docs/providers/aws/r/lambda_function.html#filename

maorfr: ack

PR #949, opened 2020-07-21
[gitlab-housekeeping] do not rebase when pipeline is incomplete

maorfr:


reconcile/gitlab_housekeeping.py
```diff
@@ -113,6 +113,18 @@ def rebase_merge_requests(dry_run, gl, rebase_limit):
             if not good_to_rebase:
                 continue
 
+            pipelines = mr.pipelines()
+            if not pipelines:
+                continue
+
+            # posibble statuses:
```
rrati, line 120: type: possible

maorfr: did you meen typo?

maorfr: was that not funny? :) thanks for the catch!

PR #954, opened 2020-07-23
Updated obsolete apiversions

rrati:


rrati:
Ya, i didn't think it was, but better safe than sorry.  We should probably clean it up at some point. :)

PR #956, opened 2020-07-23
[WIP] Add slack_sender integration

violetyao:


reconcile/slack_sender.py
```diff
@@ -0,0 +1,77 @@
...
+    channels = to.get('channels')
+    if channels:
+        for channel in channels:
+            audience.add(channel)
+
+    return audience
+
+
+def run(dry_run=False):
+    slackapi = SlackApi(username="app-sre-bot", icon_emoji=":eyes:")
```
maorfr, line 40: i would propose that we initiate SlackApi using the method defined in reconcile/slack_base.py.
this would also force you to consume slack from the query as defined in other integrations, which is desirable.

reconcile/slack_sender.py
```diff
@@ -0,0 +1,77 @@
...
+
+import reconcile.queries as queries
+from utils.slack_api import SlackApi
+
+from utils.state import State
+
+QONTRACT_INTEGRATION = 'slack-sender'
+
+
+def collect_to(to):
```
maorfr, line 12: i think we should re-use the existing collect_to function and not include slack channels in it (will also be required due to the previous comment)

reconcile/slack_sender.py
```diff
@@ -0,0 +1,77 @@
...
+
+    notifications_to_send = [n for n in notifications
+                             if not state.exists(n['name'])]
+
+    for notification in notifications_to_send:
+        logging.info(['send_notification', notification['name'],
+                      notification['subject']])
+
+        if not dry_run:
+            names = collect_to(notification['to'])
```
maorfr, line 64: i'm not sure if we want to support sending messages to individuals, but leaving that to others to decide

utils/slack_api.py
```diff
@@ -29,6 +29,14 @@ def chat_post_message(self, text):
             **self.chat_kwargs
         )
 
+    def chat_post_message_to_channel(self, channel, text):
```
maorfr, line 32: can we re-use the `chat_post_message` function?

maorfr:
I like this integration a lot! thanks for working on this @violetyao!

violetyao:
Hi Maor! Thank you for commenting on this! It has been a long way. We finally decided to come back to the qontract-reconcile approach for sending slack notifications instead of sending directly from qontract-api. I'll reply to this thread after meeting with JF tomorrow.

maorfr:
respectfully cleaning the queue. :1st_place_medal:

PR #965, opened 2020-07-29
[openshift-saas-deploy] don't hang when thread exits

maorfr:
since the underlying function is using sys.exit(), the thread is not being closed and the execution of openshift-saas-deploy-wrapper does not end.

### DIFF
--- a/reconcile/openshift_saas_deploy_wrapper.py
+++ b/reconcile/openshift_saas_deploy_wrapper.py
@@ -41,5 +41,5 @@ def run(dry_run, thread_pool_size=10, io_dir='throughput/'):
         available_thread_pool_size=available_thread_pool_size
     )
 
-    if [ec for ec in exit_codes if ec > 0]:
+    if [ec for ec in exit_codes if ec]:
         sys.exit(1)
--- a/reconcile/openshift_saas_deploy_wrapper.py
+++ b/reconcile/openshift_saas_deploy_wrapper.py
@@ -1,3 +1,4 @@
+import sys
 import semver
 
 import reconcile.openshift_saas_deploy as osd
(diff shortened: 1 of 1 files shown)


reconcile/openshift_saas_deploy_wrapper.py
```diff
@@ -13,10 +14,15 @@
...
-            saas_file_name=saas_file_name,
-            env_name=env_name)
+    exit_code = 0
+    try:
+        osd.run(dry_run=dry_run,
+                thread_pool_size=available_thread_pool_size,
+                saas_file_name=saas_file_name,
+                env_name=env_name)
+    except SystemExit as e:
+        exit_code = int(str(e))
```
maorfr, line 24: is there a better way?

jmelis:
Can you elaborate @maorfr ? The description of the issue does not match to what I understand from the code you are changing.

maorfr:
look at this log:
https://ci.int.devshift.net/job/service-app-interface-gl-pr-check/5081/console

when running the wrapper integration - it is using multiple threads all calling to the openshift-saas-deploy run function.
this function is using sys.exit(), which is not closing the thread. so even when running with just 1 saas file as input, the thread is not closing and the wrapper never exists itself.

reconcile/openshift_saas_deploy_wrapper.py
```diff
@@ -29,6 +35,11 @@ def run(dry_run, thread_pool_size=10, io_dir='throughput/'):
             thread_pool_size,
             len(saas_file_owners_diffs))
 
-    threaded.run(osd_run_wrapper, saas_file_owners_diffs, thread_pool_size,
-                 dry_run=dry_run,
-                 available_thread_pool_size=available_thread_pool_size)
+    exit_codes = threaded.run(
```
jmelis, line 38: but will this return at all? won't it just stay hanging here as long as there are running threads?

maorfr: catching `SystemExit` will make the thread return

reconcile/openshift_saas_deploy_wrapper.py
apahim: `return e.code`

maorfr: ty!

reconcile/openshift_saas_deploy_wrapper.py
```diff
@@ -29,6 +35,11 @@ def run(dry_run, thread_pool_size=10, io_dir='throughput/'):
...
+        osd_run_wrapper, saas_file_owners_diffs, thread_pool_size,
+        dry_run=dry_run,
+        available_thread_pool_size=available_thread_pool_size
+    )
+
+    if [ec for ec in exit_codes if ec > 0]:
```
apahim, line 44: `if [ec for ec in exit_codes if ec]:`, so we get any non-0, including negatives.

maorfr: is that not good?

maorfr: done!

jmelis:
After a conversation with @maorfr  this now makes sense to me. LGTM

PR #982, opened 2020-08-04
remove last-applied-configuration annotation on 0x0

maorfr:
fixes https://issues.redhat.com/browse/APPSRE-2064

this PR adds a check if `oc apply` failed due to this issue, and if so - it will remove the annotation and re-apply.


jmelis:
This is brilliant @maorfr ! :rocket:

PR #993, opened 2020-08-11
[openshift-resources] check that kind exists in api-resources

maorfr:
based on the work in #950.

we sometimes want to add resources of CRs which their CRDs have not yet been applied to the cluster.

this PR adds a check that a resource kind under managedResourceTypes is also available in oc api-resources.
if it doesn't, the integration will not attempt to fetch it's current state.
assuming such a resource will only be available in the template after the CRD has been applied.

example issue that this PR solves:
- https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/7101
- https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/7104

### DIFF
--- a/reconcile/openshift_resources.py
+++ b/reconcile/openshift_resources.py
@@ -19,7 +19,8 @@ def run(dry_run, thread_pool_size=10, internal=None, use_jump_host=True,
                  use_jump_host=use_jump_host,
                  providers=providers,
                  cluster_name=cluster_name,
-                 namespace_name=namespace_name)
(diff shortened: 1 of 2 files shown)


jmelis:
Genius. Why don't we want to init_api_resources every single time?

maorfr:
It makes less sense to check this for integrations applying specific resources like Secret/NetworkPolicy/..., and more for integrations which deploy "everything"

maorfr:
merging this, but we can discuss more and change things if needed.

PR #1008, opened 2020-08-19
Automatically add id and external_id to App-Interface

jmelis:


reconcile/ocm_clusters.py
```diff
@@ -51,6 +52,21 @@ def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
                     'from current version %s',
                     cluster_name, desired_version, current_version)
                 error = True
+
+            # id and external_id are present
+            if not desired_spec.get('id') or \
+                    not desired_spec.get('external_id'):
+                cluster_id = current_spec['spec']['id']
+                cluster_external_id = current_spec['spec']['external_id']
+                logging.info(
```
maorfr, line 61: f-strings?

maorfr (approved):

reconcile/ocm_clusters.py
jmelis: done

PR #1022, opened 2020-08-23
validate saas files are reusing environment parameters

maorfr:
we can force reusing parameters defined on the environment instead of redefining them in saas files.
this will help us rebuild complete environments faster and update full environments to point to new endpoints, such as dependencies.

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -517,6 +517,18 @@ def saas_dev(ctx, app_name=None, saas_file_name=None, env_name=None):
                 parameters.update(saas_file_parameters)
                 parameters.update(rt_parameters)
                 parameters.update(target_parameters)
+
+                for replace_key, replace_value in parameters.items():
+                    if not isinstance(replace_value, str):
+                        continue
+                    replace_pattern = '${' + replace_key + '}'
+                    for k, v in parameters.items():
+                        if not isinstance(v, str):
+                            continue
+                        if replace_pattern in v:
+                            parameters[k] = \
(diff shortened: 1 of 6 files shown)


apahim:
on it

maorfr:
i'm aware we can re-use a lot of the code, but i'd rather refactor in a future iteration

tools/qontract_cli.py
```diff
@@ -517,6 +517,18 @@ def saas_dev(ctx, app_name=None, saas_file_name=None, env_name=None):
...
+                for replace_key, replace_value in parameters.items():
+                    if not isinstance(replace_value, str):
+                        continue
+                    replace_pattern = '${' + replace_key + '}'
+                    for k, v in parameters.items():
+                        if not isinstance(v, str):
+                            continue
+                        if replace_pattern in v:
+                            parameters[k] = \
+                                v.replace(replace_pattern, replace_value)
```
apahim, line 530: Not sure I understand what you're trying to achieve here. Mind explaining?

maorfr: if someone uses the following parameter: `AUTHORIZE_URL: ${OCM_BASE_URL}/auth`
we replace the `OCM_BASE_URL` with the actual value coming from the environment.
this is actually the parameter unpacking.

apahim: I see. `validate saas files are reusing environment parameters` or the commit messges do not tell the full story here :)
tks.

apahim (approved):
You know me so well... I was going to comment on the code duplication :)
Please don't loose track of it.

PR #1023, opened 2020-08-24
[sentry-helper] introduce new integration

maorfr:
this integration watches our smtp server and finds messages sent due to user issues (access requests, etc).
the integration will send a slack notification to the user pointing them to the sentry documentation that will assist them.

### DIFF
--- a/reconcile/sentry_helper.py
+++ b/reconcile/sentry_helper.py
@@ -1,0 +1,48 @@
+import logging
+
+import utils.smtp_client as smtp_client
+import reconcile.queries as queries
+
+from reconcile.slack_base import init_slack_workspace
+from utils.state import State
+
+QONTRACT_INTEGRATION = 'sentry-helper'
+
+
+def guess_user(user_name, users):
+    guesses = [
+        u for u in users
+        if user_name.lower() == u['name'].lower()
+        or user_name.lower() == u['org_username']
+        or user_name.lower() == u['github_username'].lower()
+    ]
+    return guesses
+
+
+def run(dry_run):
+    settings = queries.get_app_interface_settings()
+    accounts = queries.get_aws_accounts()
+    users = queries.get_users()
+    state = State(
+        integration=QONTRACT_INTEGRATION,
+        accounts=accounts,
+        settings=settings
+    )
+
+    user_names = smtp_client.get_sentry_users_from_mails(settings=settings)
+    if not dry_run:
+        slack = init_slack_workspace(QONTRACT_INTEGRATION)
+    for user_name in user_names:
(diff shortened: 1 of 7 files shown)


reconcile/slack_base.py
```diff
@@ -24,3 +24,8 @@ def init_slack(slack_info, integration):
                      username=username)
 
     return slack
+
+
+def init_slack_workspace(integration):
```
rporres, line 29: This is doing more than `init`ting. It's actually returning a slack client. I would name it accordingly

rporres: Although I see it's a common pattern in the code 😅

utils/smtp_client.py
```diff
@@ -27,8 +30,14 @@ def init(host, port, username, password):
         s.starttls()
         s.login(username, password)
         _client = s
+    if _server is None:
```
rporres, line 33: Does it make sense to init this every time if it's only going to be used by `get_sentry_users_from_mails`? Is there a pythonic way to do lazy inits? :)

reconcile/sentry_helper.py
```diff
@@ -0,0 +1,48 @@
...
+        settings=settings
+    )
+
+    user_names = smtp_client.get_sentry_users_from_mails(settings=settings)
+    if not dry_run:
+        slack = init_slack_workspace(QONTRACT_INTEGRATION)
+    for user_name in user_names:
+        guesses = guess_user(user_name, users)
+        if not guesses:
+            continue
```
rporres, line 38: Not even a debug log message here? 😢

reconcile/sentry_helper.py
```diff
@@ -0,0 +1,48 @@
...
+        slack_username = \
+            guesses[0].get('slack_username') or guesses[0]['org_username']
+        if state.exists(slack_username):
+            continue
+        logging.info(['help_user', slack_username])
+        if not dry_run:
+            state.add(slack_username)
+            slack.chat_post_message(
+                f'yo <@{slack_username}>! ' +
+                'checkout https://url.corp.redhat.com/sentry-help')
```
rporres, line 48: Shouldn't this contain more information of what's actually happening? Not only for the user but for us

utils/smtp_client.py
rporres: Most likely we would need to turn this into an object to achieve this kind of behaviour

rporres (commented):
Looks good. Comments here and there, nothing blocking.

No doc at all, which forces the reader to go through all the code to understand what functions are actually doing... Not great reviewer experience :(

rporres:
I see no reference to this integration in the README

PR #1027, opened 2020-08-25
informative error message for vpc peering

tparikh:
https://issues.redhat.com/browse/APPSRE-2193

### DIFF
--- a/utils/aws_api.py
+++ b/utils/aws_api.py
@@ -613,7 +613,12 @@ class AWSApi(object):
         session = self.get_session(account['name'])
         sts = session.client('sts')
         role_arn = account['assume_role']
-        role_name = role_arn.split('/')[1]
+
+        try:
+            role_name = role_arn.split('/')[1]
+        except Exception:
+            logging.error("awsInfrastructureAccess missing in cluster.yml")
+
         response = sts.assume_role(
             RoleArn=role_arn,
             RoleSessionName=role_name


utils/aws_api.py
```diff
@@ -613,7 +613,12 @@ def get_cluster_vpc_id(self, account, route_tables=False):
         session = self.get_session(account['name'])
         sts = session.client('sts')
         role_arn = account['assume_role']
-        role_name = role_arn.split('/')[1]
+
+        try:
+            role_name = role_arn.split('/')[1]
+        except Exception:
```
apahim, line 619: This exception is too generic. Please be specific.

utils/aws_api.py
```diff
@@ -613,7 +613,12 @@ def get_cluster_vpc_id(self, account, route_tables=False):
         session = self.get_session(account['name'])
         sts = session.client('sts')
         role_arn = account['assume_role']
-        role_name = role_arn.split('/')[1]
+
+        try:
+            role_name = role_arn.split('/')[1]
```
apahim, line 618: Now `role_name` won't be initialized and the code will break in the `sts.assume_role()` call below.

apahim (changes requested):
I'd avoid the try/except block for a code path that is expected to happen. Anyway, at minimum, please resolve my comments.

apahim:
Ping

PR #1032, opened 2020-08-31
allow running terraform-resources based integrations per account

BumbleFeng:

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -238,9 +238,15 @@ def init_working_dirs(accounts, thread_pool_size,
     return ts, working_dirs
 
 
-def setup(print_only, thread_pool_size, internal, use_jump_host):
+def setup(print_only, thread_pool_size, internal, use_jump_host,
+          account_name):
     gqlapi = gql.get_api()
     accounts = queries.get_aws_accounts()
+    if account_name:
+        accounts = [n for n in accounts
+                    if n['name'] == account_name]
+        if not accounts:
+            raise ValueError(f"aws account {account_name} is not found")
     settings = queries.get_app_interface_settings()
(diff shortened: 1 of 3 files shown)


tparikh:
Nice work @BumbleFeng. @maorfr @jmelis PTAL

reconcile/terraform_resources.py
```diff
@@ -238,9 +238,13 @@ def init_working_dirs(accounts, thread_pool_size,
     return ts, working_dirs
 
 
-def setup(print_only, thread_pool_size, internal, use_jump_host):
+def setup(print_only, thread_pool_size, internal, use_jump_host,
+          account_name):
     gqlapi = gql.get_api()
     accounts = queries.get_aws_accounts()
+    if account_name:
+        accounts = [n for n in accounts
```
maorfr, line 246: should we check that `accounts` is not empty?

maorfr (approved):
LGTM

very nice! is this going to be instead of sharding or in addition?
i can see how this change would make it simpler for our idea of qontract-operator (instead of counting accounts and using that for shards, we just run the integration per account). is this is the intention?

reconcile/terraform_resources.py
BumbleFeng: If it runs `--account-name` with nothing, get `Error: --account-name option requires an argument`.
If it runs with `--account-name ""`, it runs for all accounts.

BumbleFeng: If no account matches, it only runs the oc part.

maorfr: Can we avoid it by checking?

BumbleFeng: Sure, raise an error for account not found

PR #1033, opened 2020-09-01
[gitlab-labeler] introduce new integration

maorfr:
this integration adds labels to MRs based on the changed paths and on the existing labels in the project.

for example, if the `/services/myservice/` path changed and we have a label containing the word `myservice` - that label will be added.

the integration basically guesses labels and we will probably need to modify the logic or the data in app-interface as we find issues.


rporres (approved):
LGTM. As usual in me, I would love if we started acquiring the habit of documenting the functions 😂 

There also no reference in the README about this integration

PR #1036, opened 2020-09-01
retry apply on error if it's not an invalid value error

maorfr:
in some cases when we apply resources we get an invalid value error `Invalid Value 0x0`, which can be handled by removing the last applied configuration annotation.

this PR changes the way the `apply` method works, and makes it retry apply actions which got an error which is not the invalid value error.

this should allow us to apply resource types which depend on the previously applied resources. for example, an operator creating a CR - with this change we should be able to apply both resource types from the same saas file, as long as the dependant CR is listed after the creating resources under `managedResourceTypes`.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -219,7 +219,7 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
 
         try:
             oc.apply(namespace, annotated.toJSON())
-        except InvalidValueApplyError as e:
+        except InvalidValueApplyError:
             oc.remove_last_applied_configuration(
                 namespace, resource_type, resource.name)
             oc.apply(namespace, annotated.toJSON())
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -4,7 +4,7 @@ import utils.threaded as threaded
 import reconcile.queries as queries
 
 from utils.oc import OC_Map
-from utils.oc import StatusCodeError, ApplyError
+from utils.oc import StatusCodeError, InvalidValueApplyError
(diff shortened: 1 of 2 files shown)


reconcile/openshift_base.py
```diff
@@ -219,13 +219,10 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
...
         try:
             oc.apply(namespace, annotated.toJSON())
-        except ApplyError as e:
-            if 'Invalid value: 0x0' in str(e):
-                oc.remove_last_applied_configuration(
-                    namespace, resource_type, resource.name)
-                oc.apply(namespace, annotated.toJSON())
-            else:
-                raise StatusCodeError(str(e))
+        except InvalidValueApplyError as e:
```
apahim, line 222: You don't need the `e` anymore

maorfr: thanks!

reconcile/openshift_base.py
```diff
@@ -219,13 +219,10 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
 
         try:
             oc.apply(namespace, annotated.toJSON())
-        except ApplyError as e:
-            if 'Invalid value: 0x0' in str(e):
-                oc.remove_last_applied_configuration(
-                    namespace, resource_type, resource.name)
-                oc.apply(namespace, annotated.toJSON())
-            else:
-                raise StatusCodeError(str(e))
```
apahim, line 228: Did anything change at all with this pull request?
Before, you'd call `StatusCodeError` here.

apahim (approved):
lgtm

PR #1038, opened 2020-09-02
[terraform-resources] don't collect current state from openshift in dry-run mode

maorfr:
as spotted by @BumbleFeng, collecting the current state from the clusters takes a long time.

this is not required in dry-run mode as the terraform output will hold all the required information (we only use the resources on the cluster as a fall-back).

so if the integration is running in dry-run mode - skip collecting current state.


BumbleFeng (approved):

maorfr:
@BumbleFeng i messed this one up. would you like to submit a new one?

PR #1040, opened 2020-09-03
[terraform-resources] don't collect current state from openshift in dry-run mode

BumbleFeng:
collecting the current state from the clusters takes a long time.

this is not required in dry-run mode as the terraform output will hold all the required information (we only use the resources on the cluster as a fall-back).

so if the integration is running in dry-run mode - skip collecting the current state.

Replacement of https://github.com/app-sre/qontract-reconcile/pull/1038


maorfr (approved):

reconcile/terraform_resources.py
```diff
@@ -207,13 +207,17 @@ def populate_oc_resources(spec, ri):
         logging.error(msg)
 
 
-def fetch_current_state(namespaces, thread_pool_size, internal, use_jump_host):
+def fetch_current_state(dry_run, namespaces, thread_pool_size, internal,
+                        use_jump_host):
     ri = ResourceInventory()
     settings = queries.get_app_interface_settings()
     oc_map = OC_Map(namespaces=namespaces, integration=QONTRACT_INTEGRATION,
```
maorfr, line 214: do we even need to initialize a real oc_map instance?

BumbleFeng: Need for cleanup https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/terraform_resources.py#L302

maorfr (approved):

PR #1043, opened 2020-09-04
pr_gateway refactoring - phase 2

apahim:
Now that the have the qontract-reconcile-gitlab-mr-sqs-consumer in the periodic job, it is time to replace the producers by the new utils.mr module.

Reference: https://github.com/app-sre/qontract-reconcile/blob/master/utils/mr/README.md


maorfr:
the intention of reconcile.gitlab_pr_submitter was to abstract away if the integration is using gitlab or sqs, in order to allow us to run the integration in both ways.

i think we should maintain this ability, so we are able to switch integrations to submit MRs directly to gitlab instead of going through sqs. wdyt?

apahim:
Yeah, removed that because I didn't see any integrations running in both ways... all of them have a well defined target for the MR, but I guess that might be usefull for running locally or in the future, when/if we have an internal cluster.
Restoring that functionality.

apahim:
Done. PTAL

apahim:
Done. PTAL

reconcile/ldap_users.py
```diff
@@ -42,13 +43,11 @@ def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
                        in user_specs if delete]
 
     if not dry_run:
-        instance = queries.get_gitlab_instance()
-        settings = queries.get_app_interface_settings()
-        gl = GitLabApi(instance, project_id=gitlab_project_id,
-                       settings=settings)
+        mr_cli = mr_client_gateway.init(gitlab_project_id=gitlab_project_id)
```
maorfr, line 46: yes. beautiful. if we can find a way to use the external ldap we will even be able to move this integration to run on the cluster as a CronJob if we want.

reconcile/mr_client_gateway.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    Creates the Merge Request client to of a given type.
+
+    :param gitlab_project_id: used when the client type is 'gitlab'
+    :param sqs_or_gitlab: 'gitlab' or 'sqs'
+    :return: an instance of the selected MR client.
+    """
+    if sqs_or_gitlab is None:
+        settings = queries.get_app_interface_settings()
+        client_type = settings.get('pullRequestGateway', 'gitlab')
```
maorfr, line 23: change to mergeRequestGateway while we are renaming this component?

maorfr (approved):
Beautiful. LGTM

reconcile/mr_client_gateway.py
apahim: done

PR #1047, opened 2020-09-08
[unleash-watcher] don't act on empty unleash instances

maorfr:
When there is an error calling the Unleash API, exceptions are caught deep in the client and we can not identify these cases.
The client also returns a fallback value in case issues occurred, so we have no way of knowing if anything went wrong.

We can, however, ignore Unleash instances with 0 feature toggles in this integration.

This integration is also opted in, so this is likely a good enough solution.


apahim (approved):
LGTM

PR #1049, opened 2020-09-08
Don't fail on template missing parameters

apahim:
The parameters are no really mandatory. Let's not fail when they are not
present in the template.

### DIFF
--- a/utils/saasherder.py
+++ b/utils/saasherder.py
@@ -309,23 +309,20 @@ class SaasHerder():
 
             if "IMAGE_TAG" not in consolidated_parameters:
                 template_parameters = template.get('parameters')
-                if not template_parameters:
-                    logging.error(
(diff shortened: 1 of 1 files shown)


maorfr:
Doesn't oc process fails when there are no parameters?

maorfr:
Also look at the history. This error was added due to a problem we had...

apahim:
> Doesn't oc process fails when there are no parameters?

no

> Also look at the history. This error was added due to a problem we had...

history says that much... do you know what was the error?

maorfr:
I only found this PR: #864. I didn't link any issues in it, but i believe it was about trying to look into the parameters which didn't exist.
I think your approach works just as well.

maorfr (approved):
LGTM

apahim:
thanks!

PR #1050, opened 2020-09-08
return an exception if no commits found for ref

jfchevrette:
If no commits are found for a given ref, raise an Exception

Wrap `saasherder.get_moving_commits_diff` in try/catch, log a proper error and exit early


maorfr (approved):

jfchevrette:
In review, I don't think it's a good idea to have the entire integration fail if one repo is badly configured (ie: has an invalid ref)

Should we simply log, let it continue and then exit with a non-zero?

maorfr:
i also prefer #1051 over this one

PR #1051, opened 2020-09-08
saas-deploy-trigger-moving-commits: only consider targets that are not disabled

jfchevrette:
This ensures we don't trigger jobs for saas-file targets that are disabled

### DIFF
--- a/reconcile/openshift_saas_deploy_trigger_moving_commits.py
+++ b/reconcile/openshift_saas_deploy_trigger_moving_commits.py
@@ -20,6 +20,15 @@ def run(dry_run, thread_pool_size=10):
         logging.error('no saas files found')
         sys.exit(1)
 
+    # Remove saas-file targets that are disabled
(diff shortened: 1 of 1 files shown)


maorfr (commented):
Should this be moved into queries.get_saas_files? There is already the loop in there (we can add a kwarg?)

jfchevrette:
@maorfr I originally made the change in that file, but I thought there's possibility of undesired side-effects from other consumers of that query function. Thinking there's a possibility some other integration is retrieving the data from the same function and expects to have the disabled jobs listed in the results

maorfr:
yup, so something along the lines of `saas_files = queries.get_saas_files(disabled=False)` perhaps?

maorfr:
that may prove difficult though since we have the option to pass env_name. i say this is good enough as is.

maorfr (approved):
Ran it locally, works as expected.
LGTM

PR #1053, opened 2020-09-09
[sentry-helper] fixes

maorfr:
fixes comments in #1023.


rporres (approved):
For a moment I was completely puzzled as I wasn't taking into account #1024 😂 

For the rest, thanks for taking care of this.

PR #1062, opened 2020-09-11
[jenkins-job-cleaner] introduce integration

maorfr:
this PR adds a new integration: jenkins-job-cleaner

this integration iterates through jenkins instances with a `deleteMethod: manual` and compares the existing jobs in the instance to the desired jobs calculated by jenkins-job-builder. for each such found job, if the job name starts with one of the `managedProjects` of the intsance - the integration will delete the job.

this ensures that we can only delete our own jobs in a multi-tenant jenkins instance.


apahim (approved):
Looks very nice. I love the unnested conditions.

PR #1067, opened 2020-09-16
enable kms encryption for sqs

BumbleFeng:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-key-management.html

### DIFF
--- a/utils/terrascript_client.py
+++ b/utils/terrascript_client.py
@@ -1231,6 +1231,7 @@ class TerrascriptClient(object):
         region = common_values['region'] or self.default_regions.get(account)
         specs = common_values['specs']
         all_queues_per_spec = []
+        kms_keys = set()
         for spec in specs:
             defaults = self.get_values(spec['defaults'])
             queues = spec.pop('queues', [])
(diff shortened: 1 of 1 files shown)


utils/terrascript_client.py
```diff
@@ -1248,6 +1249,21 @@ def populate_tf_resource_sqs(self, resource, namespace_info):
                 if self._multiregion_account_(account):
                     values['provider'] = 'aws.' + region
                 values.update(defaults)
+                kms_master_key_id = values.pop('kms_master_key_id', None)
+                if kms_master_key_id:
```
apahim, line 1253: Shouldn't we check only agains `None`? Anything else should be considered that the value is there...

utils/terrascript_client.py
```diff
@@ -1248,6 +1249,21 @@ def populate_tf_resource_sqs(self, resource, namespace_info):
...
+                    if not kms_master_key_id.startswith("arn:", 0):
```
apahim, line 1254: 0 is the default index for `startswith()`

utils/terrascript_client.py
```diff
@@ -1248,6 +1249,21 @@ def populate_tf_resource_sqs(self, resource, namespace_info):
...
+                        kms_key = self._find_resource_(
+                            account, kms_master_key_id, 'kms')
+                        if kms_key:
+                            kms_res = "aws_kms_key." + \
+                                kms_key['resource']['identifier']
```
apahim, line 1259: I'd favor f-strings all around. No biggie anyways.

utils/terrascript_client.py
```diff
@@ -1302,6 +1318,13 @@ def populate_tf_resource_sqs(self, resource, namespace_info):
                     }
                 ]
             }
+            if len(kms_keys) > 0:
```
apahim, line 1321: It's a list... `if ksm_keys:` is enough.

utils/terrascript_client.py
```diff
@@ -1302,6 +1318,13 @@ def populate_tf_resource_sqs(self, resource, namespace_info):
...
+                kms_statement = {
+                    "Effect": "Allow",
+                    "Action": ["kms:Decrypt"],
+                    "Resource": list(set(kms_keys))
```
apahim, line 1325: Why don't you make  `kms_keys` a set to begin with? That would make one less type conversion.

apahim (approved):
Just a few nitpicks, nothing really critical. Feel free to ignore them :)

utils/terrascript_client.py
```diff
@@ -1302,6 +1318,13 @@ def populate_tf_resource_sqs(self, resource, namespace_info):
                     }
                 ]
             }
+            if len(kms_keys):
```
apahim, line 1321: It's really just `if kms_keys:`

PR #1068, opened 2020-09-16
Add StatefulSet as recyclable

apahim:


maorfr (approved):

apahim:
ok, verified, seems to work well :)

PR #1069, opened 2020-09-16
secret_reader should raise a useful exception on vault forbidden error

jfchevrette:
At the moment, if we use `secret_reader.read` to read a vault secret and we don't have permission to access the vault path, the `hvac` package will raise an exception that isn't very helpful.

This is especially problematic for integrations or utilities we use that tries to read multiple secrets consecutively. For instance some integrations will iterate through all known clusters and will attempt to get their automation token. Currently if we get a permission denied on reading one of these secrets, we don't know anything on what secret caused the error.

This change aims to catch the generic permission error from `hvac` and return a more useful exception that contains the path that couldn't be read.

### DIFF
--- a/utils/secret_reader.py
+++ b/utils/secret_reader.py
@@ -1,8 +1,13 @@
-from sretoolbox.utils import retry
+import hvac
 
+from sretoolbox.utils import retry
 
 import utils.vault_client as vault_client
 import utils.config as config
+
+
+class VaultForbidden(Exception):
+    pass
 
 
 @retry()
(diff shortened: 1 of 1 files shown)


apahim (approved):
LGTM

utils/secret_reader.py
```diff
@@ -49,6 +54,11 @@ def read_all(secret, settings=None):
     """
 
     if settings and settings.get('vault'):
-        return vault_client.read_all(secret)
+        try:
+            data = vault_client.read_all(secret)
+        except hvac.exceptions.Forbidden:
+            raise VaultForbidden(f'permission denied reading vault secret at '
```
maorfr, line 60: this will still be retried multiple times. should we be specific in the `retry` function to not catch this?

PR #1072, opened 2020-09-17
Fixing the SQS cleaning logic

apahim:
The way SQS works, if we don't delete the messages withing 30s from
the receive_messages(), they messages will come back in SQS ans up
for consuption.

We could just extend that 30s timeout OR, better, we can just delete in
SQS all the messages we just received right away and then process
them. This patch implementes the latter.

Baseline is: no real changes are introduced here, we now just delete
the messages in time.

### DIFF
--- a/reconcile/gitlab_mr_sqs_consumer.py
+++ b/reconcile/gitlab_mr_sqs_consumer.py
@@ -8,7 +8,6 @@ import logging
 import reconcile.queries as queries
 
 from utils import mr
-from utils.defer import defer
 from utils.sqs_gateway import SQSGateway
 from utils.gitlab_api import GitLabApi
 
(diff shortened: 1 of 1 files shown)


maorfr (approved):
LGTM

in the near future we will probably want to send faulty messages to an error queue.
perhaps it would make sense to use `threaded` here to process all received messages at once and keep the deletion "decision" for after the processing.

this is just a proposal. this PR is good to merge as is and we can iterate over it later.

apahim:
I don't think we have to cherish messages, unless we only send them once from the producer side, but that's not the case here... the producers will keep producing messages untill the changes are merged to app-interface. We already deal with broken integrations, so I'd just not care about broken messages in the message level.

PR #1074, opened 2020-09-18
annotation oc secret with managed aws account

BumbleFeng:


maorfr:
what problem is this solving and where can i find it in jira? :)

BumbleFeng:
> what problem is this solving and where can i find it in jira? :)

https://issues.redhat.com/browse/APPSRE-2500

maorfr:
i think this is a good use case for `caller`:

https://github.com/app-sre/qontract-reconcile/blob/aa217a77c8fcb7d83ed98429619bc566952b46ad/reconcile/openshift_base.py#L263-L265

PR #1075, opened 2020-09-18
Fix missing tuple return type in owner lgtm code

kylape:


maorfr (approved):
Thank you!

Please include the new commit sha in .env in the faulty MR.

PR #1076, opened 2020-09-18
aws_route53: exclude records matching items in unmanaged_record_names

jfchevrette:
This PR adds the ability to specify a list of regexes via `unmanaged_record_names` in a zone file.

The regex in this list will be compiled and records will be checked for a `fullmatch` against the list of regexes. If the record name is found, it is then removed from both the current and the desired state.

This will allow us to add records in AWS outside of the integration flow, without having the integration remove them

This is a quick fix to an immediate problem. Longer term we may want to add state management to the integration instead.


maorfr (approved):
seems much more straight forward then a state.

LGTM

PR #1081, opened 2020-09-21
[gitlab-labeler] add 'saas' in ignore_tokens

BumbleFeng:
Fix all saas file change get labeled to `tenant-dsass`


maorfr:
Why not use ignore_tokens?
This lets the labels be whatever they are...

maorfr:
i think we should split this to 2 PRs. one of them to ignore the `saas` token, and another to add the saas-file-update properly (which will cover https://issues.redhat.com/browse/APPSRE-2395).

wdyt?

maorfr (approved):

PR #1082, opened 2020-09-22
[sql-query] check resource info is found

maorfr:
fixes https://issues.redhat.com/browse/APPSRE-2506

this PR adds a check that the terraform resource info is not None, which indicates that the rds identifier does not exist in the specified namespace.


apahim (approved):
LGTM, thank you!

rporres (approved):

PR #1089, opened 2020-09-23
[openshift-upgrade-watcher] introduce new integration

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2514

this integration watches for the UpgradeConfig resource in each cluster and notifies in a slack channel if the cluster is currently being upgraded.

### DIFF
--- a/reconcile/openshift_upgrade_watcher.py
+++ b/reconcile/openshift_upgrade_watcher.py
@@ -1,0 +1,64 @@
+import logging
+
+import reconcile.queries as queries
+
+from datetime import datetime
+
+from reconcile.slack_base import init_slack_workspace
+from utils.oc import OC_Map
+from utils.state import State
+from utils.defer import defer
+
+
+QONTRACT_INTEGRATION = 'openshift-upgrade-watcher'
+
+
+@defer
+def run(dry_run, thread_pool_size=10, internal=None, use_jump_host=True, defer=None):
+    settings = queries.get_app_interface_settings()
+    accounts = queries.get_aws_accounts()
+    clusters = [c for c in queries.get_clusters(minimal=True) if c.get('ocm')]
+    oc_map = OC_Map(clusters=clusters, integration=QONTRACT_INTEGRATION,
+                    settings=settings, internal=internal,
+                    use_jump_host=use_jump_host,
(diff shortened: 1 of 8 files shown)


reconcile/openshift_upgrade_watcher.py
```diff
@@ -0,0 +1,66 @@
...
+            if state.exists(state_key):
+                # already notified
+                continue
+            logging.info(['cluster_upgrade', cluster])
+            if not dry_run:
+                state.add(state_key)
+                slack.chat_post_message(
+                    f'Heads up! cluster `{cluster}` is currently ' +
+                    f'being upgraded to version `{version}`'
+                )
```
apahim, line 66: I love it!
Can we somehow detect when it finishes?

reconcile/openshift_upgrade_watcher.py
```diff
@@ -0,0 +1,66 @@
...
+        state_key = f'{cluster}-{upgrade_at}'
+        # if this is the first iteration in which 'now' had passed
+        # the upgrade at date time, we send a notification
+        if upgrade_at_obj < now:
+            if state.exists(state_key):
+                # already notified
+                continue
+            logging.info(['cluster_upgrade', cluster])
+            if not dry_run:
+                state.add(state_key)
```
rrati, line 62: DO you need to remove this key from the state file at some point or at these keys unique per upgrade?

reconcile/openshift_upgrade_watcher.py
maorfr: we currently detect it and a MR is submitted to app-interface.
we can totally enhance this later on to also report the finish. would you agree to this in a next iteration?

reconcile/openshift_upgrade_watcher.py
maorfr: the key contains a timestamp, so it's unique per upgrade

reconcile/openshift_upgrade_watcher.py
maorfr: although even this MR does not indicate the real finish. nor does the `ClusterVersion` on the cluster. so maybe a bit more work to report the "real" finish time.

apahim (approved):

PR #1090, opened 2020-09-23
skip apply secret when running tf-r per account

BumbleFeng:
Temporary solutions for running tf-r per account locally. 
The integration running on the cluster will manage the secret after any manual running.
Will refactor with `caller` for further operator implement


maorfr (approved):
can you also add a comment in the code itself?
just to save folks time when going through it.

PR #1092, opened 2020-09-24
add rds to ignore_tokens

BumbleFeng:
Fix rds file get labeled with `tenant-telemeter-ocp-dashboards`


maorfr (approved):
let's start thinking about how we make this logic a bit smarter :)

PR #1095, opened 2020-09-24
[openshift-resources] add support for openshift shared resources

maorfr:
Part of https://issues.redhat.com/browse/APPSRE-2524

Dpends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/9822

Now that we are running multiple instances of the same service, we are duplicating a lot of the content of a namespace file. Mostly we are duplicating the openshift resources that we want to apply to the namespace.

This PR allows users to define shared resources that can be added to multiple namespace files, thus removing the need to duplicate this content and allow users to update all their namespaces in a single place.

The lack in this ability recently led to 2 namespace files overriding each other and causing a service misconfiguration.

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -82,7 +82,7 @@ NAMESPACES_QUERY = """
       resource
       resourceNames
     }
-    openshiftSharedResources {
+    sharedResources {
       openshiftResources {
         %s
       }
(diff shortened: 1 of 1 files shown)


apahim (approved):
THis is AWESOME. I was afraid things would get fuzzy, but kept a very clear distinction about what's shared and what's local. Kudos.

aditya-konarde (approved):
Lgtm! :)

rporres:
Arrived late to the party but read it anyway. LGTM, indeed!

maorfr:
thank you all!

PR #1097, opened 2020-09-29
[gitlab-housekeeping] merge without waiting for pending jobs

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2484

if a job is cancelled, in many cases the gitlab pipeline will hang and wait forever. we want to merge based on the latest result only, as each run already cancelled the previous one.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -137,7 +137,8 @@ def rebase_merge_requests(dry_run, gl, rebase_limit, wait_for_pipeline=False):
 
 
 @retry(max_attempts=10)
-def merge_merge_requests(dry_run, gl, merge_limit, rebase, insist=False):
+def merge_merge_requests(dry_run, gl, merge_limit, rebase, insist=False,
+                         wait_for_pipeline=False):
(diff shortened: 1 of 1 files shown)


reconcile/gitlab_housekeeping.py
```diff
@@ -165,16 +166,17 @@ def merge_merge_requests(dry_run, gl, merge_limit, rebase, insist=False):
...
-            # running, pending, success, failed, canceled, skipped
-            incomplete_pipelines = \
-                [p for p in pipelines
-                 if p['status'] in ['running', 'pending']]
-            if incomplete_pipelines:
-                if insist:
-                    raise Exception(f'insisting on {merge_label}')
-                else:
-                    continue
+            if wait_for_pipeline:
```
apahim, line 169: Shouldn't we wait for running?

maorfr: as long as the last one passed, we don't really care about previous ones...
plus, the disconnection may also mean that jobs that are "running" are not really running anymore.

apahim: got it

apahim (approved):
lgtm

PR #1103, opened 2020-10-01
[slack-cluster-usergroups] introduce new integration

maorfr:
this integration adds all users with access to a certain cluster to a slack usergroup named <cluster_name>-cluster (same as .auth.team in a cluster file).

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/10072

### DIFF
--- a/reconcile/slack_cluster_usergroups.py
+++ b/reconcile/slack_cluster_usergroups.py
@@ -64,11 +64,7 @@ def run(dry_run):
     slack.initiate_usergroups()
     desired_state = get_desired_state(slack)
     usergroups = [d['usergroup'] for d in desired_state]
-    print(usergroups)
     current_state = get_current_state(slack, usergroups)
-    print(current_state)
-    import sys
-    sys.exit()
     slack_usergroups.print_diff(current_state, desired_state)
 
     if not dry_run:
--- a/reconcile/slack_cluster_usergroups.py
+++ b/reconcile/slack_cluster_usergroups.py
@@ -1,8 +1,11 @@
+import logging
+
 import reconcile.queries as queries
 import reconcile.openshift_users as openshift_users
 import reconcile.slack_usergroups as slack_usergroups
 
 from reconcile.slack_base import init_slack_workspace
+from utils.slack_api import UsergroupNotFoundException
 
 QONTRACT_INTEGRATION = 'slack-cluster-usergroups'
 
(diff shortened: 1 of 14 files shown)


utils/slack_api.py
```diff
@@ -14,13 +14,14 @@ class SlackApi(object):
...
     def __init__(self, token, settings=None, init_usergroups=True,
-                 **chat_kwargs):
+                 workspace_name=None, **chat_kwargs):
         slack_token = secret_reader.read(token, settings=settings)
         self.sc = SlackClient(slack_token)
         self.results = {}
         self.chat_kwargs = chat_kwargs
         if init_usergroups:
             self.initiate_usergroups()
+        self.workspace_name = workspace_name or ''
```
jfchevrette, line 24: Maybe we should fail if workspace_name is not passed instead of possibly passing an empty string down the line in slack_map or when we build states?

jfchevrette:
Other than the comment above, LGTM

reconcile/slack_cluster_usergroups.py
```diff
@@ -0,0 +1,69 @@
...
+            "channels": channels,
+            "description": description,
+        })
+
+    return current_state
+
+
+def run(dry_run):
+    slack = init_slack_workspace(QONTRACT_INTEGRATION)
+    slack.initiate_usergroups()
```
rporres, line 60: filling part of an object outside the object creation looks like an antipattern. Since `SlackApi` object supports to pass this in the object creation, we should do it in that moment.

I would also make `initiate_usergroups` a "private" method of that class to avoid temptations 😂

utils/slack_api.py
rporres: I agree, it looks weird that it can be empty...

rporres (commented):
A few comments here and there, looks good otherwise

reconcile/slack_cluster_usergroups.py
```diff
@@ -0,0 +1,69 @@
...
+QONTRACT_INTEGRATION = 'slack-cluster-usergroups'
+
+
+def get_desired_state(slack):
+    desired_state = []
+    all_users = queries.get_users()
+    all_clusters = queries.get_clusters(minimal=True)
+    clusters = [c for c in all_clusters
+                if c.get('auth') and c['auth'].get('team')
+                and c.get('ocm')]
```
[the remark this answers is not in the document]
maorfr: this integration is intended to send notifications about cluster events, such as an upgrade. we only want to do that for clusters we manage (in our ocm organization).

reconcile/slack_cluster_usergroups.py
maorfr: updated the defaults of `init_usergroups` to True with overrides where required to align with the SlackApi class init behavior.

utils/slack_api.py
maorfr: make workspace_name a required argument for SlackApi init

PR #1101, opened 2020-10-01
fix oc resources not apply for clean up data field in configmap or secret

BumbleFeng:
Fix https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/9999 not get apply

Signed-off-by: Feng Huang <fehuang@redhat.com>

PR #1104, opened 2020-10-04
fix openshift-resource data compare

BumbleFeng:

### DIFF
--- a/utils/openshift_resource.py
+++ b/utils/openshift_resource.py
@@ -55,7 +55,7 @@ class OpenshiftResource(object):
                     diff = [k for k in obj2_v
                             if k not in obj1_v
                             and k not in IGNORABLE_DATA_FIELDS]
-                    if diff:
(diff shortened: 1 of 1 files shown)


rporres:
Can you give context on this change, @BumbleFeng ? Or an example on when this change was necessary?

BumbleFeng:
> Can you give context on this change, @BumbleFeng ? Or an example on when this change was necessary?

This is a fix for https://github.com/app-sre/qontract-reconcile/pull/1101
The intention is to apply change when deleting field in configmap or secret. However, after I promote to prod, I found out it only compare field name but not values, because I also skip [compare values part](https://github.com/app-sre/qontract-reconcile/blob/master/utils/openshift_resource.py#L74-L75) in my PR. As a result, MR to change values such as https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/10146 did not get applied.

rporres:
great, thx!!

PR #1107, opened 2020-10-04
print unused schemas when running integrations

maorfr:
related to https://issues.redhat.com/browse/APPSRE-2143

this PR adds a check in each integration execution to see if there are any schemas that were not used and log a warning in such case.
we only do this is NON dry-run to prevent cases of an integration only performing certain queries when it is running in full mode.


rporres (approved):
not exactly what I was thinking but it is an interesting addition. Thx for taking care

PR #1109, opened 2020-10-05
Allow users to provide alternate database name

maorfr:
part of https://issues.redhat.com/browse/RHCLOUD-8432

this PR adds the option to define an altername db name in the namespace file, which will only override the value in the output resource.

this PR replaces the suggested implementation in #941. the difference is that 941 implemented this through the defaults file which is not visible enough and will allow us to keep accumulating such debt.

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -34,6 +34,7 @@ TF_NAMESPACES_QUERY = """
         output_resource_name
         enhanced_monitoring
         replica_source
+        alt_db_name
       }
       ... on NamespaceTerraformResourceS3_v1 {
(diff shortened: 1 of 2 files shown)


maorfr:
cc @tparikh

apahim (approved):
LGTM. You know, when `name` is not defined in the defaults file, there's still a database created... that's just how RDS works, meaning I don't get why we don't just specify the right db name from the beginning.
Anyway, just thinking out loud here, nothing related to your PR.

maorfr:
as this is very similar to the approach in #941 i don't imagine we'll have objections. moving forward with this.

PR #1114, opened 2020-10-07
[saas-file-owners] add note when MR can not be self-serviced

maorfr:
if there are changes outside the scope of a saas file owner, let's add a note indicating that the MR can not be self-serviced.


rrati (approved):
lgtm

arilivigni:
lgtm

PR #1121, opened 2020-10-11
[jenkins-job-builder] change from compare/no-compare to state

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2591

the way we currently use the jjb integration (in dry-run mode) is by running it twice: first time to read the current state from the production endpoint and a second time to read the desired state from the local endpoint and print the diffs.

this means that every MR is running the 1st execution even if the jjb integration is not selected for execution afterwards.

this PR changes the way we run the jjb integration - we run it once, like every other integration. that's it.
it will collect the current state from `State` instead.

we tried to read the current state from Jenkins itself, but the JJB defintions and the queried jobs from Jenkins are different to a degree we can't "patch" them to look alike (was attempted a few times in https://issues.redhat.com/browse/APPSRE-1854).

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -492,14 +492,11 @@ def jenkins_plugins(ctx):
 
 
 @integration.command()
+@environ(['APP_INTERFACE_STATE_BUCKET', 'APP_INTERFACE_STATE_BUCKET_ACCOUNT'])
 @throughput
-@click.option('--compare/--no-compare',
-              default=True,
-              help='compare between current and desired state.')
 @click.pass_context
-def jenkins_job_builder(ctx, io_dir, compare):
-    run_integration(reconcile.jenkins_job_builder, ctx.obj,
-                    io_dir, compare)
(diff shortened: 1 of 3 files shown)


apahim:
The change is very welcomed, thank you for that.
I'm thinking that we will have a period of time, from merge to non-dryrun jenkins-job-builder, when all the MRs will tell tenants that the jenkins jobs will be updated. We have to make sure that the AppSRE Team knows what that is about and where to look at when it does not go away.

apahim (approved):

maorfr:
i already ran the integration manually and persisted the state :)

apahim:
I'm talking about new changes to the jobs in app-itnerface.

maorfr:
ah got you. yes, you are correct. to make that time shorter i have separated the jjb integration to its own job and we can iterate over the frequency: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/10498

apahim:
Thanks. Just make sure you communicate that to the team, as you usually do :)

PR #1122, opened 2020-10-11
[saas-file-owners] print if this is a valid saas file changes MR only

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2591

we will use this output to better select which integrations we want to run in https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/hack/select-integrations.py.

for example, if this MR contains valid saas file changes only (updating a `ref` for example), there is no reason to run the jjb integrations, as this MR will not create/delete any jobs.

### DIFF
--- a/reconcile/saas_file_owners.py
+++ b/reconcile/saas_file_owners.py
@@ -268,6 +268,11 @@ def run(dry_run, gitlab_project_id=None, gitlab_merge_request_id=None,
     valid_saas_file_changes_only = is_saas_file_changes_only and is_valid_diff
     write_diffs_to_file(io_dir, diffs, valid_saas_file_changes_only)
 
+    # print 'yes' or 'no' to allow pr-check to understand if changes
+    # are only valid saas file changes (and exclude other integrations)
+    output = 'yes' if valid_saas_file_changes_only else 'no'
+    print(output)
+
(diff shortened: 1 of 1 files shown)


reconcile/saas_file_owners.py
```diff
@@ -268,6 +268,11 @@ def run(dry_run, gitlab_project_id=None, gitlab_merge_request_id=None,
     valid_saas_file_changes_only = is_saas_file_changes_only and is_valid_diff
     write_diffs_to_file(io_dir, diffs, valid_saas_file_changes_only)
 
+    # print 'yes' or 'no' to allow pr-check to understand if changes
+    # are only valid saas file changes (and exclude other integrations)
+    output = 'yes' if valid_saas_file_changes_only else 'no'
+    print(output)
```
apahim, line 274: I'm assumning this is temporary? If not, please use `logging`.

maorfr: this is actually the intention: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/10500/diffs#103cbab4c3108c4f7f19aefb17f404405aaae36c_117_116

apahim: yep, logging ftw

apahim (approved):
Just a suggestion btw... we can always improve later.

PR #1124, opened 2020-10-13
[openshift-saas-deploy] allow non-yaml files in directory provider

maorfr:
this will allow to have .md files in a directory with dashboards (for example)


maorfr:
such as https://github.com/RedHatInsights/topological_inventory-api/tree/master/grafana-dashboards

PR #1128, opened 2020-10-15
[openshift-serviceaccount-tokens] write to shared resources location

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2606

we want to write all output secrets to the shared location so it's easier to migrate apps between clusters.
in a follow up PR we will do the work to support `openshiftServiceAccountTokens` in the integration.

allows us to do: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/10729

### DIFF
--- a/reconcile/openshift_serviceaccount_tokens.py
+++ b/reconcile/openshift_serviceaccount_tokens.py
@@ -60,9 +60,18 @@ def write_outputs_to_vault(vault_path, ri):
     integration_name = QONTRACT_INTEGRATION.replace('_', '-')
     for cluster, namespace, _, data in ri:
         for name, d_item in data['desired'].items():
+            body_data = d_item.body['data']
+            # write secret to per-namespace location
             secret_path = \
-              f"{vault_path}/{integration_name}/{cluster}/{namespace}/{name}"
-            secret = {'path': secret_path, 'data': d_item.body['data']}
+                f"{vault_path}/{integration_name}/" + \
+                f"{cluster}/{namespace}/{name}"
+            secret = {'path': secret_path, 'data': body_data}
+            vault_client.write(secret)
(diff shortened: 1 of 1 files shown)


reconcile/openshift_serviceaccount_tokens.py
```diff
@@ -60,9 +60,18 @@ def write_outputs_to_vault(vault_path, ri):
...
-              f"{vault_path}/{integration_name}/{cluster}/{namespace}/{name}"
-            secret = {'path': secret_path, 'data': d_item.body['data']}
+                f"{vault_path}/{integration_name}/" + \
+                f"{cluster}/{namespace}/{name}"
+            secret = {'path': secret_path, 'data': body_data}
+            vault_client.write(secret)
+            # write secret to shared-resources location
+            secret_path = \
+                f"{vault_path}/{integration_name}/" + \
+                f"shared-resources/{name}"
```
apahim, line 73: Is this path unique enough?

apahim (approved):
Looks good, just one question to be in the safe side.

reconcile/openshift_serviceaccount_tokens.py
maorfr: the `name` consists of <cluster>-<namespace>-<sa_name> of the desired SA token. even if there are duplicates, they are identical.

maorfr: https://github.com/app-sre/qontract-reconcile/blob/de3600f62c5cea8a7bf33b5780a7ecb7945b79a2/reconcile/openshift_serviceaccount_tokens.py#L46-L47

apahim: Cool, that works, thanks!

PR #1131, opened 2020-10-15
[terraform-resources] change resources tags

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2614

as part of moving `terraformResources` to shared resources files, we can no longer use the cluster and namespace tags, as these will be different in each namespace including these resources.

we can assume that a terraform resource will not be used across different environments, and mostly will not be used across different apps. in case it is, we assume there will be a shared parentApp and we will use that.

this change will only change the tags on the resources.

the best reason to use these tags is for billing calculations, and using the `parentApp` is likely the best tag to choose in this case.

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -155,6 +155,15 @@ TF_NAMESPACES_QUERY = """
     terraformResources {
       %s
     }
+    app {
+      name
+      parentApp {
+        name
+      }
+    }
+    environment {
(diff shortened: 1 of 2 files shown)


BumbleFeng (approved):

BumbleFeng:
LGTM

rporres (approved):

rrati (approved):
LGTM

PR #1137, opened 2020-10-20
Convert container base form centos:7 to ubi8:8.2

arilivigni:
Made some modifications to support going from centos:7 to ubi8:8.2 for the base image

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -1,11 +1,13 @@
-FROM quay.io/app-sre/centos:7 AS jsonnet-builder
+FROM registry.access.redhat.com/ubi8/ubi:8.2 AS jsonnet-builder
 
 ENV GOPATH=/go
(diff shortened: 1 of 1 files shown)


arilivigni:
> we should test all our integrations with this change.
> an easy way to do that is to run `make rc`, which will create a new image with these changes (and push it to quay).
> you can then create a MR to app-interface (DO NOT MERGE) with the new image, to see how the integrations are responding.

excellent will do!

arilivigni:
@maorfr can this be merged then?

maorfr:
this has been tested in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/10879.

LGTM from me, but i would still want @jfchevrette to take a look. i don't feel comfortable approving this as i have never used the ubi image myself.

jfchevrette (approved):

jfchevrette:
LGTM as well. Test results look good 👍

PR #1138, opened 2020-10-22
use oc replace when last-applied-configuration annotation is too long

maorfr:
fixes https://issues.redhat.com/browse/APPSRE-2631

this PR adds a handle to the case where the last-applied-configuration annotation is too long, in which case we should use `oc replace` instead of `oc apply`.


BumbleFeng (approved):
LGTM

PR #1142, opened 2020-10-23
[ocm-clusters] submit a single MR for all clusters updates

maorfr:
instead of submitting multiple MRs for multiple clusters which creates an overload, we now submit a single MR for all clusters.


apahim (approved):
Great one! LGTM

PR #1143, opened 2020-10-25
[terraform-resources] do not initiate oc_map on dry-run

maorfr:
since if this is a dry-run we exit before checking openshift resources, there is no need to initiate them.

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -247,8 +247,8 @@ def init_working_dirs(accounts, thread_pool_size,
     return ts, working_dirs
 
 
-def setup(print_only, thread_pool_size, internal, use_jump_host,
-          account_name):
+def setup(dry_run, print_only, thread_pool_size, internal,
+          use_jump_host, account_name):
     gqlapi = gql.get_api()
     accounts = queries.get_aws_accounts()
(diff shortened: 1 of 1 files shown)


reconcile/terraform_resources.py
```diff
@@ -260,8 +260,16 @@ def setup(print_only, thread_pool_size, internal, use_jump_host,
     namespaces = gqlapi.query(TF_NAMESPACES_QUERY)['namespaces']
     tf_namespaces = [namespace_info for namespace_info in namespaces
                      if namespace_info.get('managedTerraformResources')]
-    ri, oc_map = fetch_current_state(tf_namespaces, thread_pool_size,
-                                     internal, use_jump_host)
+    if dry_run:
+        ri = ResourceInventory()
```
apahim, line 264: The `fetch_current_state` should be the function with the knowledge about what to give you in case of `dry_run`. Can you make it accept the `dry_run` argument and return this?

apahim (approved):
Small nitpick, lgtm otherwise.

reconcile/terraform_resources.py
maorfr: that makes a lot of sense! thanks!

PR #1149, opened 2020-10-26
[jira-watcher] board per shard (round robin)

maorfr:
this PR adds a round-robin approach to the sharding module and adds usage of it for jira-watcher


apahim (approved):
Smart :)

PR #1151, opened 2020-10-26
[jira-watcher] change from io-dir to s3 state

maorfr:
instead of using the local fs, we should use s3 state like most other integrations. this will prevent the loss of a state.


apahim (approved):
So much simpler... LGTM

PR #1153, opened 2020-10-27
[jira-watcher] performance improvements

maorfr:
this PR makes it so that jira-watcher only queries for the fields it needs instead of the entire issue.
this reduces the integration's run time to ~1m for 3 boards.

since we no longer hold the entire issue in memory, we do not really need sharding, and we can reduce resources.


apahim (approved):
Wow super nice :)
lgtm

PR #1154, opened 2020-10-27
[openshift-resources] replace instead of apply on UnsupportedMediaType

maorfr:
fixes https://issues.redhat.com/browse/APPSRE-2660

this PR solves an issue with applying EgressNetworkPolicy, as this resource can not be applied, only replaced.


apahim (approved):
LGTM

PR #1155, opened 2020-10-28
[terraform-resources] rds determine password in dry-run without oc

maorfr:
since #1143, we no longer initiate oc_map in dry-run, this is bound to fail for a new DB (that it's password is not in the state yet).


apahim (approved):
lgtm

PR #1157, opened 2020-10-28
Don't fail on unreachable cluster retry & move on

arilivigni:
Instead of raise log the error on an unreachable cluster and move on to the next cluster after retries

### DIFF
--- a/utils/jsonnet.py
+++ b/utils/jsonnet.py
@@ -21,7 +21,7 @@ def generate_object(jsonnet_string):
     try:
         jsonnet_bundler_dir = os.environ['JSONNET_VENDOR_DIR']
     except KeyError as e:
-        raise JsonnetError('JSONNET_VENDOR_DIR not set')
+        raise JsonnetError(f'JSONNET_VENDOR_DIR not set: {e}')
 
     cmd = ['jsonnet', '-J', jsonnet_bundler_dir, path]
     status = run(cmd, stdout=PIPE, stderr=PIPE)
--- a/utils/jsonnet.py
+++ b/utils/jsonnet.py
@@ -20,8 +20,8 @@ def generate_object(jsonnet_string):
 
     try:
         jsonnet_bundler_dir = os.environ['JSONNET_VENDOR_DIR']
-    except KeyError as e:
-        raise JsonnetError(f'JSONNET_VENDOR_DIR not set: {e}')
+    except KeyError:
+        raise JsonnetError('JSONNET_VENDOR_DIR not set')
 
     cmd = ['jsonnet', '-J', jsonnet_bundler_dir, path]
     status = run(cmd, stdout=PIPE, stderr=PIPE)
--- a/utils/oc.py
+++ b/utils/oc.py
@@ -230,7 +230,10 @@ class OC(object):
         # oc api-resources only has name or wide output
         # and we need to get the KIND, which is the last column
         cmd = ['api-resources', '--no-headers']
-        results = self._run(cmd).decode('utf-8').split('\n')
+        if self._run(cmd) == '{}':
+            results = '{}'
+        else:
+            results = self._run(cmd).decode('utf-8').split('\n')   
         return [r.split()[-1] for r in results]
 
     @retry(exceptions=(JobNotRunningError), max_attempts=20)
@@ -423,14 +426,14 @@ class OC(object):
                 raise MetaDataAnnotationsTooLongApplyError(
                     f"[{self.server}]: {err}")
             if not (allow_not_found and 'NotFound' in err):
-                raise StatusCodeError(f"[{self.server}]: {err}")
+                logging.error(StatusCodeError(f"[{self.server}]: {err}"))
 
         if not out:
             if allow_not_found:
                 return '{}'
             else:
-                raise NoOutputError(err)
-
+                logging.error(NoOutputError(err))
+                return '{}'
         return out.strip()
 
     def _run_json(self, cmd, allow_not_found=False):
--- a/utils/oc.py
+++ b/utils/oc.py
@@ -516,14 +516,14 @@ class OC_Map(object):
                 jump_host = None
             try:
                 oc_client = OC(server_url, token, jump_host,
-                    settings=self.settings,
-                    init_projects=self.init_projects,
-                    init_api_resources=self.init_api_resources)
+                               settings=self.settings,
+                               init_projects=self.init_projects,
+                               init_api_resources=self.init_api_resources)
                 self.set_oc(cluster, oc_client)
             except StatusCodeError:
-                    logging.info('Cluster unreachable: ' + cluster)
-                    pass
-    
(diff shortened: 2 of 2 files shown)


arilivigni:
This should address https://issues.redhat.com/browse/APPSRE-2414

utils/oc.py
```diff
@@ -423,14 +426,14 @@ def _run(self, cmd, **kwargs):
                 raise MetaDataAnnotationsTooLongApplyError(
                     f"[{self.server}]: {err}")
             if not (allow_not_found and 'NotFound' in err):
-                raise StatusCodeError(f"[{self.server}]: {err}")
+                logging.error(StatusCodeError(f"[{self.server}]: {err}"))
```
apahim, line 429: This is more conceptual, but when the function raises an exception, the function is already telling all it should tell. Logging the exception message is a decision for the caller, if it wants to do it as part of the exception handling.

utils/oc.py
```diff
@@ -230,7 +230,10 @@ def get_api_resources(self):
         # oc api-resources only has name or wide output
         # and we need to get the KIND, which is the last column
         cmd = ['api-resources', '--no-headers']
-        results = self._run(cmd).decode('utf-8').split('\n')
+        if self._run(cmd) == '{}':
+            results = '{}'
+        else:
+            results = self._run(cmd).decode('utf-8').split('\n')
```
apahim, line 236: This is basically saying that the cluster has no resources in it, but it will still allow the openshift-resources to apply whaterver is specified in app-interface, which it shouldn't.
Also, this mechanism depends on `init_api_resources` to be True on OC init.
Last but not least, this is breaking things:
```
...
  File "/home/apahim/git/qontract-reconcile/reconcile/openshift_base.py", line 231, in apply
    oc.recycle_pods(dry_run, namespace, resource_type, resource)
  File "/home/apahim/git/qontract-reconcile/utils/oc.py", line 291, in recycle_pods
    pods = self.get(namespace, 'Pods')['items']
KeyError: 'items'
```
Did you test this properly?

apahim (changes requested):
Hi @arilivigni, thanks for tackling this.

I believe that you're hooking into the wrong place here. The key functionality here would be to avoid the `OC_Map` of creating the client for the cluster that is offline. And that logic goes in the `OC_Map.init_oc_client()`, when it calls `self.set_oc()`. That call should not happen if the cluster is not online.

As for testing if the cluster is online, I reccomend adding a new method to the `OC` class. That method would be called from the `OC.__init__()`, but it would always be called (unlike the `self.get_api_resources()`, which depends on `init_api_resources` to be `True`).

Then, if the new `OC` method detects that the cluster is down, `OC.__init__()` would raise an exception that would ultimatelly make the `OC_Map.init_oc_client()` not to add that cluster mapping, as mentioned before.

Hope that makes sense, feel free to reach me out to discuss that further.

utils/oc.py
arilivigni: When I ran this I got zero errors back.  I assigned a cluster DNS FQDN to point to my laptop.  In both positive and negative testing I never got these errors, but maybe my setup is incorrect.

utils/oc.py
```diff
@@ -503,12 +514,15 @@ def init_oc_client(self, cluster_info):
...
-                   init_api_resources=self.init_api_resources))
+            try:
+                oc_client = OC(server_url, token, jump_host,
+                    settings=self.settings,
+                    init_projects=self.init_projects,
+                    init_api_resources=self.init_api_resources)
+                self.set_oc(cluster, oc_client)
+            except StatusCodeError:
+                logging.info
+                pass
```
apahim, line 525: pushed too early? :)

arilivigni: Yes sorry making a change now and testing :disappointed:

utils/jsonnet.py
```diff
@@ -21,7 +21,7 @@ def generate_object(jsonnet_string):
     try:
         jsonnet_bundler_dir = os.environ['JSONNET_VENDOR_DIR']
     except KeyError as e:
-        raise JsonnetError('JSONNET_VENDOR_DIR not set')
+        raise JsonnetError(f'JSONNET_VENDOR_DIR not set: {e}')
```
apahim, line 24: Is this to make code checker happy? you can remove the ` as e` instead... not need to throw extra info in the message if you're missing it :)

utils/oc.py
```diff
@@ -50,6 +50,10 @@ class JobNotRunningError(Exception):
     pass
 
 
+class ClusterUnreachable(Exception):
+    pass
```
apahim, line 54: I don't see this being used...

utils/oc.py
```diff
@@ -233,6 +239,11 @@ def get_api_resources(self):
         results = self._run(cmd).decode('utf-8').split('\n')
         return [r.split()[-1] for r in results]
 
+    def get_cluster_info(self):
+        # get cluster-info to check that the cluster is reachable
```
apahim, line 243: this comment is a bit misleading.... it makes sense in the line 73 aboce, but it doesn't here... here you're simply calling cluster-info... this method knows nothing about your use case for that, if you know what I mean.

arilivigni: agreed will change

utils/oc.py
```diff
@@ -503,12 +514,15 @@ def init_oc_client(self, cluster_info):
...
-                   init_projects=self.init_projects,
-                   init_api_resources=self.init_api_resources))
+            try:
+                oc_client = OC(server_url, token, jump_host,
+                               settings=self.settings,
+                               init_projects=self.init_projects,
+                               init_api_resources=self.init_api_resources)
+                self.set_oc(cluster, oc_client)
+            except StatusCodeError:
+                logging.info('Cluster unreachable: ' + cluster)
```
apahim, line 523: I'd use `logging.error` here.
Also, small nitpick, please use `('Cluster unreachable: %s', cluster)` for improved performance.

utils/oc.py
arilivigni: correct removed

utils/oc.py
```diff
@@ -503,12 +514,15 @@ def init_oc_client(self, cluster_info):
...
+                pass
```
apahim, line 520: You can safelly drop the `pass` now that you have the `logging` call.

utils/jsonnet.py
arilivigni: I think it is worth matching the lines above so they are consistent

maorfr: let me represent @rporres and say that this is an unrelated change.

apahim: The previous call was made out of `Exception`, which is as generic as you can get, so casting the original exception to a string makes sense as you don't really know what happened there.
Here you are handling `KeyError`, which is specific. You have a message saying exactly what's wrong, so there's no added value.... you're just making more noise.

apahim (approved):
LGTM
Thanks for the patience :)

PR #1161, opened 2020-11-01
Revert "Don't fail on unreachable cluster retry & move on"

maorfr:
Reverts app-sre/qontract-reconcile#1157

cc @arilivigni @apahim


utils/jsonnet.py
```diff
@@ -20,7 +20,7 @@ def generate_object(jsonnet_string):
 
     try:
         jsonnet_bundler_dir = os.environ['JSONNET_VENDOR_DIR']
-    except KeyError:
+    except KeyError as e:
```
arilivigni, line 23: This fails pep8 guidelines that is why it was removed.

PR #1162, opened 2020-11-02
Warn if clusters are unreachable and continue

arilivigni:
Address unreachable clusters
https://issues.redhat.com/browse/APPSRE-2414


apahim (changes requested):
This is proven to be broken. Details on https://issues.redhat.com/browse/APPSRE-2414.

arilivigni:
@apahim I don't think this was my error this seems to be an OC mismatch more than anything because the old OC didn't support cluster-info.  I think I am just going to choose a different command that should work on both, but that said is it really my error or the fact that we had a rather large hole in using the OC 3.11 binary :disappointed:

arilivigni:
@apahim and @maorfr please let me know if this is a better solution.  I got the correct behavior locally and it only marked a stage cluster unreachable when I pointed it to my localhost instead.  So all seems to behave as expected, but once merged will monitor stage.

jfchevrette:
I tested this locally while blocking network to a single cluster and it logged the error 56 times which makes me think we're not skipping the cluster completely and it's trying to re-initialize the oc client as many times during the run. If the cluster becomes available again during the run then that may mean we risk having partially applied state. We really want to skip it completely for all resources if it is unreachable initially

```
[2020-11-02 11:33:17] [ERROR] [oc.py:init_oc_client:525] - Cluster unreachable: foobarbaz
[2020-11-02 11:33:17] [ERROR] [oc.py:init_oc_client:525] - Cluster unreachable: foobarbaz
[2020-11-02 11:33:17] [ERROR] [oc.py:init_oc_client:525] - Cluster unreachable: foobarbaz
[2020-11-02 11:33:18] [ERROR] [oc.py:init_oc_client:525] - Cluster unreachable: foobarbaz
... (repeated 56 times during the run)
```

The integration also returned exit code 0, but I believe it would be preferable to return a non-zero code

arilivigni:
@jfchevrette how many retries are acceptable?  Doesn't this just follow the normal retry logic?  I will change it to exit with non-zero exit code.

PR #1163, opened 2020-11-02
upgrade oc client to v4

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2684

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -28,12 +28,13 @@ FROM registry.access.redhat.com/ubi8/ubi:8.2
 
 ENV LC_ALL=en_US.utf8
 ENV LANG=en_US.utf8
+ENV OC_VERSION=4.6.1
 ENV TF_VERSION=0.11.14
 ENV GIT_SECRETS_VERSION=1.3.0
 ENV JSONNET_VENDOR_DIR=/opt/jsonnet-bundler/vendor
 
 RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && \
-    curl https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.6.1/openshift-client-linux.tar.gz -o oc.tar.gz &&\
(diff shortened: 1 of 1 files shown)


maorfr:
this is being tested using the `make rc` command in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/11374

dockerfiles/Dockerfile
```diff
@@ -33,10 +33,10 @@ ENV GIT_SECRETS_VERSION=1.3.0
 ENV JSONNET_VENDOR_DIR=/opt/jsonnet-bundler/vendor
 
 RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && \
-    curl -L https://github.com/openshift/origin/releases/download/v3.11.0/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz | tar -xvz && \
-    pushd openshift-origin-client* && \
+    curl https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-client-linux.tar.gz -o oc.tar.gz &&\
```
apahim, line 36: We should probably pin-point a version from https://mirror.openshift.com/pub/openshift-v4/clients/ocp/ so we always know what we are running, otherwise any compatibility break from latest will blow in the hands of someone doing unrelated changes.

apahim (approved):
LGTM. Not sure we should go with latest though, but I'd be ok with it.

maorfr:
latest is greatest! pinned to 4.6.1, which is what we are trying to get to for our clusters ATM.

dockerfiles/Dockerfile
maorfr: pinned.

rporres:
Late to the party. A couple of things I don't particularly like:

* A newer client going against older clusters. It can blow in the least expected moment.
* A solution to this different than matching cluster version with client version.

maorfr:
we decided to just go with this approach since it appears that most of us are running oc v4 locally anyway, and no one has seen anything funky. we also only have 2-3 remaining v3 clusters.

rporres:
> no one has seen anything funky.

famous last words

rporres:
https://issues.redhat.com/browse/APPSRE-2724 :)

PR #1173, opened 2020-11-05
[terraform-resources] use oc_map in dry-run

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2717

reverts #1143

run time is less important then correctness. we should get the same results in dry-run.


BumbleFeng:
Should we also revert https://github.com/app-sre/qontract-reconcile/pull/1155 and https://github.com/app-sre/qontract-reconcile/pull/1156/files ?

maorfr:
this is not the right solution. failing forward instead.

PR #1174, opened 2020-11-06
[terraform-resources] reset db password #1

maorfr:
this PR adds an option to define `reset_password` on an `rds` provider, which will lead to the reset of the DB password.
this PR uses the external S3 state common to app-interface.


maorfr:
closing in favor of #1175

PR #1175, opened 2020-11-06
[terraform-resources] reset db password #2

maorfr:
this PR adds an option to define `reset_password` on an rds provider, which will lead to the reset of the DB password.
this PR uses the terraform state.


maorfr:
this is likely better then #1174

rporres (approved):

PR #1177, opened 2020-11-09
[sql-query] add support for CronJob

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2715
depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/11665

this PR adds support for recurring sql queries. if a `schedule` field exists in the sql-query object, the resource will be a `CronJob` instead of a `Job` and will not be deleted.


apahim (approved):
Awesome! LGTM

PR #1180, opened 2020-11-12
[APPSRE-2609] Main loop: drop subprocess

apahim:
This patch will drop the subprocess call from the main loop (used to run
the integrations on kubernetes) in favor of a direct call to the Click
entrypoint.

With that, both the Main lopp and the integration execution will happen
in the same Python session, meaning we can more data from the
integration executiuon in the Main loop.

This sets the foundation for instrumenting the integrations to  report
the reconciliation time or any other executions attributes we want to
track.

### DIFF
--- a/dockerfiles/hack/run-integration.py
+++ b/dockerfiles/hack/run-integration.py
@@ -2,7 +2,6 @@
 
 import logging
 import os
-import subprocess
 import sys
 import time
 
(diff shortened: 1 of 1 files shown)


maorfr (approved):
:heart:

rporres (approved):
Woa, so cool. Thx!

mmclanerh:
we don't ever run qontract endpoints as threaded / subprocess anymore, right? i dont think we do.. does this have any impact on wanting to run an integration as a subscriber of a queue? that would happen serially on at least in line with whatever the integration is, yeh?

apahim:
No impact there, I guess... if anything, this makes things easier to multitask or to parallelize, if we ever come to that.

PR #1181, opened 2020-11-12
[APPSRE-2611] Tracking decorator

apahim:
```
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="10.0",name="utils.oc.OC.apply"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="30.0",name="utils.oc.OC.apply"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="60.0",name="utils.oc.OC.apply"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="150.0",name="utils.oc.OC.apply"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="300.0",name="utils.oc.OC.apply"} 10.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="600.0",name="utils.oc.OC.apply"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="1200.0",name="utils.oc.OC.apply"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="1800.0",name="utils.oc.OC.apply"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="+Inf",name="utils.oc.OC.apply"} 0.0
...
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="10.0",name="utils.oc.OC.delete"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="30.0",name="utils.oc.OC.delete"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="60.0",name="utils.oc.OC.delete"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="150.0",name="utils.oc.OC.delete"} 2.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="300.0",name="utils.oc.OC.delete"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="600.0",name="utils.oc.OC.delete"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="1200.0",name="utils.oc.OC.delete"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="1800.0",name="utils.oc.OC.delete"} 0.0
qontract_reconcile_function_elapsed_seconds_since_bundle_commit_bucket{integration="openshift-resources",le="+Inf",name="utils.oc.OC.delete"} 0.0
```

### DIFF
--- a/utils/metrics.py
+++ b/utils/metrics.py
@@ -1,0 +1,12 @@
+from prometheus_client import Gauge
+from prometheus_client import Counter
+
+
+run_time = Gauge(name='qontract_reconcile_last_run_seconds',
+                 documentation='Last run duration in seconds',
+                 labelnames=['integration', 'shards', 'shard_id'])
+
+run_status = Counter(name='qontract_reconcile_run_status',
+                     documentation='Status of the runs',
+                     labelnames=['integration', 'status',
+                                 'shards', 'shard_id'])
--- a/utils/metrics.py
+++ b/utils/metrics.py
@@ -1,5 +1,6 @@
 from prometheus_client import Gauge
 from prometheus_client import Counter
+from prometheus_client import Histogram
 
 
 run_time = Gauge(name='qontract_reconcile_last_run_seconds',
@@ -10,3 +11,11 @@ run_status = Counter(name='qontract_reconcile_run_status',
                      documentation='Status of the runs',
                      labelnames=['integration', 'status',
                                  'shards', 'shard_id'])
+
+reconcile_time = Histogram(name='qontract_reconcile_function_'
+                                'elapsed_seconds_since_bundle_commit',
+                           documentation='Run time seconds for tracked '
+                                         'functions',
+                           labelnames=['name', 'integration'],
+                           buckets=(10.0, 30.0, 60.0, 150.0, 300.0, 600.0,
+                                    1200.0, 1800.0, float("inf")))
--- a/utils/oc.py
+++ b/utils/oc.py
@@ -4,6 +4,7 @@ import logging
 import tempfile
 
 import utils.threaded as threaded
+from utils.qrtracking import elapsed_seconds_from_commit_metric
 
 from subprocess import Popen, PIPE
 from datetime import datetime
@@ -158,14 +159,17 @@ class OC(object):
                'kubectl.kubernetes.io/last-applied-configuration-']
         self._run(cmd)
 
+    @elapsed_seconds_from_commit_metric
     def apply(self, namespace, resource):
         cmd = ['apply', '-n', namespace, '-f', '-']
         self._run(cmd, stdin=resource, apply=True)
 
+    @elapsed_seconds_from_commit_metric
     def replace(self, namespace, resource):
         cmd = ['replace', '-n', namespace, '-f', '-']
         self._run(cmd, stdin=resource, apply=True)
 
+    @elapsed_seconds_from_commit_metric
     def delete(self, namespace, kind, name):
         cmd = ['delete', '-n', namespace, kind, name]
         self._run(cmd)
--- a/utils/qrtracking.py
+++ b/utils/qrtracking.py
@@ -1,0 +1,27 @@
+import time
+
+from functools import wraps
+
+from reconcile.status import RunningState
+from utils.metrics import reconcile_time
+
+
+def elapsed_seconds_from_commit_metric(function):
+
+    @wraps(function)
+    def wrapper(*args, **kwargs):
+        runing_state = RunningState()
+        result = function(*args, **kwargs)
+
+        commit_time = float(runing_state.timestamp)
+        time_spent = time.time() - commit_time
+
+        name = f'{function.__module__}.{function.__qualname__}'
+        reconcile_time.labels(
+            name=name,
+            integration=runing_state.integration
+        ).observe(amount=time_spent)
+
+        return result
+
+    return wrapper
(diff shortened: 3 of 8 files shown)


mmclanerh:
super cool! :+1:

rporres:
While this is certainly interesting and I love the idea of being able to measure the time that integrations take to run, I fail to understand how it help us to measure the time from merge to apply... Is this part of larger plan?

jmelis:
I'm concerned this may cause a cardinality explosion. It basically implies ~20 new time-series per git commit, which may be a lot.

I forgot the justification of why we need to keep track of the specific git commit.

If we don't really need the values for each git commit, I seem to recall that the last thing we proposed here was to report a histogram (without the `commit` label) in the `oc_apply` method. The value would be `current_time - git_commit_time`. The `git_commit_time` is not yet queriable via qontract-server, but it would be. So everytime you set up and initialize a new `sha` url for the server, you would also fetch the `git_commit_time` associated to it.

BTW, I liked the namedtuple _replace method, pretty cool :)

rporres:
This looks cool. For the purposes of what we want to achieve, most likely we will need to include in the same metric the deletes and the replaces in the same metric, don't you think @jmelis. The annotation name `track` would probably be too generic.

apahim:
As of now, the decorated function name is a dimension to the metric.

rporres:
> As of now, the decorated function name is a dimension to the metric.

Yeah, that was what I understood from the code

apahim:
> most likely we will need to include in the same metric the deletes and the replaces in the same metric

so all the decorated functions will contribute to the same histogram?

rporres:
Forget about it, we can just sum with a regex and that should be it :)

utils/oc.py
```diff
@@ -158,6 +159,7 @@ def remove_last_applied_configuration(self, namespace, kind, name):
                'kubectl.kubernetes.io/last-applied-configuration-']
         self._run(cmd)
 
+    @track
```
rporres, line 162: Please add tracking to `replace` and `delete` too

utils/metrics.py
```diff
@@ -0,0 +1,19 @@
...
+run_time = Gauge(name='qontract_reconcile_last_run_seconds',
+                 documentation='Last run duration in seconds',
+                 labelnames=['integration', 'shards', 'shard_id'])
+
+run_status = Counter(name='qontract_reconcile_run_status',
+                     documentation='Status of the runs',
+                     labelnames=['integration', 'status',
+                                 'shards', 'shard_id'])
+
+func_runtime = Histogram(name='qontract_reconcile_function_runtime',
```
rporres, line 15: would be `qontract_reconcile_function_elapsed_seconds_since_bundle_commit` too long? 😂

utils/metrics.py
```diff
@@ -0,0 +1,19 @@
...
+                         documentation='Run time for tracked functions',
```
rporres, line 16: We should specify the unit here

utils/metrics.py
```diff
@@ -0,0 +1,19 @@
...
+                         labelnames=['name'],
+                         buckets=(.1, .5, 1.0, 5.0, 10.0, 50.0, 100.0, 500.0,
```
rporres, line 18: What do you thing about this?
```suggestion
                         buckets=(10.0, 30.0, 60.0, 150.0, 300.0, 600.0, 1200.0, 1800.0,
```

utils/qrtracking.py
```diff
@@ -0,0 +1,26 @@
+import time
+
+from functools import wraps
+
+from reconcile.status import RunningState
+from utils.metrics import func_runtime
+
+
+def track(function):
```
rporres, line 9: What about? Maybe too long (again)
```suggestion
def elapsed_seconds_from_commit_metric(function):
```

apahim: That's terrible, but I like it :)

utils/metrics.py
apahim: sure +1

utils/metrics.py
apahim: not for me. hopefully also not for prometheus.

rporres: It doesn't look like it : https://prometheus.io/docs/concepts/data_model/

apahim:
@rporres @jmelis updated. The description shows how the metric is looking atm.

utils/oc.py
```diff
@@ -128,6 +129,7 @@ def get_items(self, kind, **kwargs):
 
         return items
 
+    @elapsed_seconds_from_commit_metric
```
rporres, line 132: I don't think we need to track this ATM, only calls that write into k8s

rporres (approved):
Apart from a small details, looks great!

utils/oc.py
apahim: sorry, leftover from testing :]

rporres: thx!

PR #1183, opened 2020-11-15
[ocm-clusters] add usage of dryRun

maorfr:
following https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/commit/ce4bd1ebfdb7af4221ad47c0430726bf0b25b24c

there is now support for quota and preflight checks in OCM. this PR adds usage of those in dry-run.

### DIFF
--- a/utils/ocm.py
+++ b/utils/ocm.py
@@ -85,7 +85,7 @@ class OCM(object):
         }
         return ocm_spec
 
-    def create_cluster(self, name, cluster):
+    def create_cluster(self, name, cluster, dry_run):
         """
         Creates a cluster.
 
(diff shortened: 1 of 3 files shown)


cben:
LGTM.  Creating cluster with `dryRun=true` either returns 204 with empty body, or 400 with error JSON, and [raise_for_status](https://requests.readthedocs.io/en/master/api/#requests.Response.raise_for_status) will raise for 400.

apahim (approved):
Noice

rporres (approved):
This is cool!

utils/ocm.py
```diff
@@ -550,13 +557,20 @@ def _get_json(self, api):
...
+            json=data,
+            params=params
+        )
         try:
             r.raise_for_status()
         except Exception as e:
             logging.error(r.text)
             raise e
+        if r.status_code == requests.codes.no_content:
+            return None
```
cben, line 573: LGTM.  The caller doesn't look at return status anyway.

PR #1187, opened 2020-11-16
[ocm-addons] introduce new integration

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2697
depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/11961

this PR adds a new integration to manage addons installations via OCM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Additional tools that use the libraries created by the reconciliations are also 
 - `ocm-github-idp`: Manage GitHub Identity Providers in OCM.
 - `ocm-machine-pools`: Manage Machine Pools in OCM.
 - `ocm-upgrade-scheduler`: Manage Upgrade Policy schedules in OCM.
-- `ocm-addons`: anages cluster Addons in OCM.
+- `ocm-addons`: Manages cluster Addons in OCM.
 - `email-sender`: Send email notifications to app-interface audience.
 - `requests-sender`: Send emails to users based on requests submitted to app-interface.
 - `service-dependencies`: Validate dependencies are defined for each service.
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Additional tools that use the libraries created by the reconciliations are also 
 - `ocm-github-idp`: Manage GitHub Identity Providers in OCM.
 - `ocm-machine-pools`: Manage Machine Pools in OCM.
 - `ocm-upgrade-scheduler`: Manage Upgrade Policy schedules in OCM.
+- `ocm-addons`: anages cluster Addons in OCM.
 - `email-sender`: Send email notifications to app-interface audience.
 - `requests-sender`: Send emails to users based on requests submitted to app-interface.
 - `service-dependencies`: Validate dependencies are defined for each service.
--- a/reconcile/ocm_addons.py
+++ b/reconcile/ocm_addons.py
@@ -59,12 +59,12 @@ def act(dry_run, diffs, ocm_map):
         cluster = diff.pop('cluster')
         addon_id = diff['id']
         logging.info([action, cluster, addon_id])
+        ocm = ocm_map.get(cluster)
+        if not ocm.get_addon(addon_id):
+            logging.error(f'Addon {addon_id} does not exist')
+            err = True
+            continue
         if not dry_run:
-            ocm = ocm_map.get(cluster)
-            if not ocm.get_addon(addon_id):
-                logging.error(f'Addon {addon_id} does not exist')
-                err = True
-                continue
             if action == 'install':
                 ocm.install_addon(cluster, diff)
             # uninstall is not supported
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -73,6 +73,7 @@ Additional tools that use the libraries created by the reconciliations are also
 - `ocm-github-idp`: Manage GitHub Identity Providers in OCM.
 - `ocm-machine-pools`: Manage Machine Pools in OCM.
 - `ocm-upgrade-scheduler`: Manage Upgrade Policy schedules in OCM.
+- `ocm-addons`: anages cluster Addons in OCM.
```
apahim, line 76: typo: anages

utils/ocm.py
```diff
@@ -491,6 +492,57 @@ def create_kafka_cluster(self, data):
...
+            result = {k: v for k, v in item.items() if k in desired_keys}
+            results.append(result)
+
+        return results
+
+    def install_addon(self, cluster, spec):
+        """ Installs an addon on a cluster
+
+        :param cluster: cluster name
+        :param spec: required information for installation
```
apahim, line 535: hard to keep track of what that would be (the required information).  I see you're using the diff, but that is built, then some keys are popped... as you're at it, could you please document here what is expected in the spec?

apahim (approved):
LGTM. You're weeks ahead of time, considering the addons that triggered this PR are not even close to prod yet :)

reconcile/ocm_addons.py
```diff
@@ -0,0 +1,88 @@
+import sys
+import logging
+
+import reconcile.queries as queries
+
+from reconcile.status import ExitCodes
+from utils.ocm import OCMMap
+
+QONTRACT_INTEGRATION = 'ocm-addons'
+
```
aditya-konarde, line 10: Do we need a `QONTRACT_INTEGRATION_VERSION` here?

reconcile/ocm_addons.py
```diff
@@ -0,0 +1,88 @@
...
+        logging.info([action, cluster, addon_id])
+        ocm = ocm_map.get(cluster)
+        if not ocm.get_addon(addon_id):
+            logging.error(f'Addon {addon_id} does not exist')
+            err = True
+            continue
+        if not dry_run:
+            if action == 'install':
+                ocm.install_addon(cluster, diff)
+            # uninstall is not supported
```
aditya-konarde, line 70: Perhaps log something for uninstalls in this case, just so we know?

utils/ocm.py
aditya-konarde: +1 I wasn't able to find the spec definition. Might be okay to add a link to the api.openshift.com docs here as the spec is evolving?

reconcile/ocm_addons.py
rporres, line 10: no `QONTRACT_INTEGRATION_VERSION` in this integration?

aditya-konarde (changes requested):
Overall lgtm, a couple of enhancements suggested. Thank you for taking this on! :)

reconcile/ocm_addons.py
```diff
@@ -0,0 +1,88 @@
...
+            diffs.append(d)
+
+    return diffs
+
+
+def act(dry_run, diffs, ocm_map):
+    err = False
+    for diff in diffs:
+        action = diff.pop('action')
+        cluster = diff.pop('cluster')
```
rporres, line 59: I'd rather be explicit here and create a new dict to pass to `install_addon` than modifying the given structure while iterating over it... It makes things harder to follow, to debug and most likely they're more prone to errors.

rporres (commented):
So this is is a bit hard to review. All the structures are buried are passed back and forth around the different functions and they're modified in place via `pop`. The lack of tests (not something exclusive to this PR) makes things more difficult to figure out.

reconcile/ocm_addons.py
maorfr: the version is used for integrations that apply resources to OpenShift, so not really needed here.

maorfr: the version is used for integrations that apply resources to OpenShift, so not really needed here.

reconcile/ocm_addons.py
maorfr: i'm tracking when this lands in OCM. once it does, we'll just implement it.

reconcile/ocm_addons.py
maorfr: we add the `cluster` and the `action` in the functions right above. i kind of like this approach, because it allows us to add fields in the app-interface schema without changing the integration's code.

utils/ocm.py
maorfr: updated!

README.md
maorfr: thanks!

maorfr:
going to merge this and improve it in a future iteration. as @apahim says, we're weeks away from operating this.

PR #1189, opened 2020-11-17
[openshift-rolebindings] fix shards

maorfr:
sharding was implemented for openshift-rolebindings in #932 and removed from use in #942.

in this PR we fix the issue we had with calculating the desired state (we rely on roles and permissions and were not filtering out roles for cluster/namespace not in the shard).

also splitting the integration to 4 shards (currently takes ~1k seconds to complete).


apahim (approved):
lgtm

aditya-konarde (approved):
Lgtm

PR #1193, opened 2020-11-18
[gitlab-housekeeping] merge not rebased

maorfr:
if `rebase` is false, we shouldn't continue if the MR is not rebased.


apahim (approved):
lgtm

PR #1188, opened 2020-11-17
Main loop: drop subprocess

apahim:
Trying again. It's the same PR, with one extra commit, that moves the Sentry init to the main loop.

PR #1195, opened 2020-11-18
SmtpClient refactoring

apahim:
This is yet another refactoring to avoid the global initialization of
objects.

### DIFF
--- a/utils/smtp_client.py
+++ b/utils/smtp_client.py
@@ -11,137 +11,79 @@ from email.mime.text import MIMEText
 from email.header import Header
 from email.utils import formataddr
 
-_client = None
-_server = None
-_username = None
-_mail_address = None
 
+class SmtpClient:
+    def __init__(self, settings):
+        config = get_config()
 
-def init(host, port, username, password, client_only):
-    global _client
-    global _server
+        smtp_secret_path = config['smtp']['secret_path']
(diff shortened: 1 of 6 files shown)


apahim:
Related to https://github.com/app-sre/qontract-reconcile/pull/1188

utils/smtp_client.py
```diff
@@ -11,137 +11,81 @@
...
+
+        self.client = smtplib.SMTP(host=self.host,
+                                   port=self.port)
+        self.client.send
+        self.client.starttls()
+        self.client.login(self.user, self.passwd)
+
+        self.server = None
+        if not client_only:
+            self.server = self._init_server()
```
apahim, line 35: this should actually be a lazy init. no reason to have the client_only path.

apahim: will update the PR with that

rporres (approved):
YES! YES! <3

utils/smtp_client.py
rporres: you updated the PR just before I was going to suggest it 😂  Great!

rporres:
Should we keep the teardown code in a destructor?

rporres:
nah, QUIT is apparently useless https://cr.yp.to/smtp/quit.html

PR #1198, opened 2020-11-19
[ocm-clusters] fix provision_shard_id

maorfr:
https://docs.google.com/document/d/1J7HNNFIWXq5gjpsT8QQw8ehkSe40A0N55eX584fdKMA

### DIFF
--- a/utils/ocm.py
+++ b/utils/ocm.py
@@ -134,7 +134,7 @@ class OCM(object):
 
         provision_shard_id = cluster_spec.get('provision_shard_id')
         if provision_shard_id:
-            ocm_spec['properties'] = {}
+            ocm_spec.setdefault('properties', {})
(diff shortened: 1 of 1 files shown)


rporres (approved):

utils/ocm.py
```diff
@@ -134,7 +134,8 @@ def create_cluster(self, name, cluster):
 
         provision_shard_id = cluster_spec.get('provision_shard_id')
         if provision_shard_id:
-            ocm_spec['provision_shard']['id'] = provision_shard_id
+            ocm_spec['properties'] = {}
```
rporres, line 137: I would just verify `ocm_spec['properties']` exists. Just to make sure that someone doesn't fill it in `ocm_spec` and doesn't realize that there's an initialization afterwards

maorfr: done!

PR #1203, opened 2020-11-20
Adding "copy_all" to the skopeo copy call

apahim:
This will make Skopeo copy all the architectures of a given image/tag
instead of only the architecture of the OS Skopeo is running on.

Depends on: https://github.com/app-sre/sretoolbox/pull/22


rporres:
can we do this opt-in? I wonder who's actually going to need this and we could save space in the general case

apahim:
We sure can. I will work on it soon, thanks for the review.

PR #1210, opened 2020-11-23
Vault client optimizations

apahim:
- Reverts commit fbc669aed0713dccb397c5c8d6e403425a2cc735.
- Creates a singletown for the VaultClient. The Class instance is then only renewed when the `is_authenticated()` return false on instance creation.


maorfr (approved):
LGTM

PR #1212, opened 2020-11-24
[VaultClient] add cache

maorfr:
This PR adds a cache to VaultClient, which will prevent getting the same data multiple times.

In openshift-resources alone, this saved ~1k of calls to Vault.

### DIFF
--- a/utils/vault.py
+++ b/utils/vault.py
@@ -29,6 +29,12 @@ class VaultConnectionError(Exception):
 
 
 class _VaultClient:
+    """
+    A class representing a Vault client. Allows read/write operations.
+    The client caches read requests in-memory if the request is made
+    to a versioned KV engine (v2), since that includes both a path
+    and a version (no invalidation required).
+    """
     def __init__(self):
         config = get_config()
 
--- a/utils/vault.py
+++ b/utils/vault.py
@@ -105,18 +105,9 @@ class _VaultClient:
         return data
 
     def _read_all_v1(self, path):
-        cache_key = path
-        cache_value = self._cache.get(cache_key)
-        if cache_value is not None:
-            logging.debug('Vault v1 cache hit')
-            return cache_value
-
         secret = self._client.read(path)
         if secret is None or 'data' not in secret:
             raise SecretNotFound(path)
-
-        data = secret['data']
-        self._cache[cache_key] = data
         return secret['data']
 
     @retry()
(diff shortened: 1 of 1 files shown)


utils/vault.py
```diff
@@ -75,6 +77,12 @@ def read_all(self, secret):
         return data
 
     def _read_all_v2(self, path, version):
+        cache_key = (path, version)
+        cache_value = self._cache.get(cache_key)
+        if cache_value:
```
apahim, line 82: `if cache_value is not None:`

utils/vault.py
```diff
@@ -91,12 +99,24 @@ def _read_all_v2(self, path, version):
...
+        data = secret['data']['data']
+        self._cache[cache_key] = data
+        return data
 
     def _read_all_v1(self, path):
+        cache_key = path
+        cache_value = self._cache.get(cache_key)
+        if cache_value:
+            logging.debug('Vault v1 cache hit')
+            return cache_value
```
apahim, line 112: can we really cache in v1? content will change, key wont...

maorfr: we only cache for the current execution

utils/vault.py
maorfr: fixed

utils/vault.py
apahim: an instance of this class will be used in many executions... we are a singletown now

maorfr: ah crap, that's true.

apahim: wait, cache for v2 is still important... if the content changes, the version also changes there

maorfr: if the data is kept across executions i don't trust it so much...

apahim: can the data change in a v2 secret while not changing the version?

maorfr: not really, but a version can be deleted

apahim: that's fine, I guess

apahim (approved):
lgtm

rporres:
I would capture in a docstring in this class the discussion around that we are caching v2 as we're using both the version and the path as cache keys and we can't cache v1 as we it can change and we have no means to invalidate it.

utils/vault.py
rporres: Not awesome but in our use case, removing a secret version and not updating app-interface would be silly

rporres (approved):

PR #1213, opened 2020-11-24
[service-dependencies] add kafka dependency

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2727
depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/12321

this will add a check that all apps that rely on kafka specify that in their dependencies.


apahim (approved):
lgtm

PR #1214, opened 2020-11-25
qontract-reconcile openshift resources should skip unreachable clusters

arilivigni:
Retry each cluster max attempts and if unreachable mark as failed
- `logging.error('The cluster is unreachable: error %s', str(e))`
- `self.set_oc(cluster, '')`
- `ri.register_error()`


utils/oc.py
```diff
@@ -529,12 +535,15 @@ def init_oc_client(self, cluster_info):
...
-                   init_api_resources=self.init_api_resources))
+            try:
+                oc_client = OC(server_url, token, jump_host,
+                               settings=self.settings,
+                               init_projects=self.init_projects,
+                               init_api_resources=self.init_api_resources)
+                self.set_oc(cluster, oc_client)
+            except StatusCodeError as e:
+                logging.error('The cluster is unreachable: error %s', str(e))
+                self.set_oc(cluster, '')
```
apahim, line 546: We are getting out of options to signal something bad happened. We have `False` in the line 529 above, and we also the check for `None` in the `openshift_base` (which I don't see ever being set, but that's besided my point).
Using an empty string crosses a line that tells me we should actually have a class to signal a bad `OC` instance. That class should be used for all the cases (replacing `None`, `False` and `''`) and carry extra information that will be used in the openshift_base to select the log level, the log message and to decide whether or not to call `ri.register_error()`.

arilivigni: I hope I can avoid writing a whole separate class for a simple catch like this one.  I mean in that case, I might as well set to False and just change the message would be easier TBH.

apahim: A whole separate class is no big deal. In fact, by returning an empty string you're creating an instance of a string type class and setting its value to `''`. Not that far from what I proposed. Same happens to False, which is a subclass of the integer type with value 0.
Anyway, returning False would be fine, even though returning None would be more correct, but that would make things harder given the current code.

maorfr: i completely agree with @apahim.
@arilivigni we have talked about just this and you agreed. what has changed?

arilivigni: I didn't remember us talking about a separate class just to catch the exception.  If I missed that please point me to where that was discussed?
I will implement a class then so it is more robust for what we are trying to achieve.  Thanks for the feedback.

arilivigni: @apahim I added the oc_status class for exit statuses related to output choices of the oc command.  If I remove line 577 in oc it gives me an error `AttributeError: 'int' object has no attribute 'cleanup'`  I am sure maybe there is a better way around this so feedback welcome.

apahim: That's exactly why I mentioned you needed a class. Now, even though you created a class, you're not really passing that new class around, you're just using it to select the right `int` and passing the `int` around. Meaning you can either have an `OC()` instance, which implements `cleanup()`, or an `int` instance, which doesn't.

The way out here is to check the type (with `isinstance()`) to decide whether to call `cleanup()` or not. Very non-pythonic. It would be fine to check types to select a code path, but not to decide which interfaces to use (see EAFP vs LBYL here: https://docs.python.org/3/glossary.html).

Anyway, I still think that you need to pass a custom class instance around (not just an `int`), that carries the error information and implements the minimum interfaces needed not to break the code. In this case, `__bool__` returning `False`, which will also be useful for the check in the `openshift_base.py`.

arilivigni: I agree with you better to pass the class around.  I just made an update to do just that, but not sure if it is 100% what you were looking for.  It does allow to add values for the oc_status, but again open to feedback to make it better.

apahim: You're still not passing a clustom class around... now you're back to passing `False`.

Sorry, I'd should have been more vebose here. What I mean is to have a class like this (created in the `utils/oc.py` module):

```python
class OCFail:
    def __init__(self, message):
        self.message = message

    def __bool__(self):
        return False
```

Then use it like this:

```python
        if automation_token is None:
            self.set_oc(cluster, OCFail(message='something something no automationToken'))
        else
            ...
            try:
                self.set_oc(cluster, OC(...))
            except StatusCodeError:
                self.set_oc(cluster, OCFail(message='something something unreachable'))
```

And check it (in the `openshift_base.py`) with:

```python
            if oc is None:
                msg = f"[{cluster}] cluster skipped."
                logging.debug(msg)
                continue
            if not oc:  # instead of "if oc is False"
                ri.register_error()
                logging.error(oc.message)
                continue
```

In that case, no changes will be required to the `OC_Map.cleanup()`, since it already checks with `if oc`. It will just work.

It can get more complex if you want, also replacing the `None` check by embedding more info into the `OCFail` instance, but that is what I meant from the beginning. 

Not saying this is the only way... ultimatelly, you tell me the approach you want to use and we go from there.

arilivigni: No, you did state that I didn't understand it completely.  Thanks for the more verbose answer now I finally get it :100:

arilivigni: Also cleanup still fails with this approach

arilivigni: Ok @apahim thanks for the solution I believe this is the answer you were looking for.  I will also attempt the `None` case but just wanted to make sure this was more in line.

utils/oc.py
```diff
@@ -256,6 +258,10 @@ def get_api_resources(self):
         results = self._run(cmd).decode('utf-8').split('\n')
         return [r.split()[-1] for r in results]
 
+    def get_oc_version(self):
```
maorfr, line 261: i'm not seeing usage of this function. am i missing it?
and just a nit - can we rename it to `version`? we are in the context of `oc`, and `version` is not `get`.

utils/oc.py
```diff
@@ -566,3 +577,11 @@ def cleanup(self):
         for oc in self.oc_map.values():
             if oc:
                 oc.cleanup()
+
+
+class OCFail:
```
maorfr, line 582: i'm totally fine with the simple approach. however, we will probably generalize this class fairly soon -
in my view, it's more the status of OC rather then a failure. it can also be a success or "nothing" (cluster skipped for example).
the status will indicate what is the proper logging level, return code of the integration, etc.
i'm +1 to iterate over it in future PRs, but just wanted to share my thoughts

reconcile/openshift_base.py
```diff
@@ -113,10 +112,9 @@ def init_specs_to_fetch(ri, oc_map,
                 msg = f"[{cluster}] cluster skipped."
```
maorfr, line 112: this is another situation we would want to handle in the same way IMO (following the previous comment). again, perfectly good to go in a future PR.

maorfr:
this is starting to look pretty good! thank you @arilivigni and @apahim!

utils/oc.py
apahim: If you instrument `oc` for the errors paths and pass an instance of `oc` to `OCFail`, will you need `OCFail` at all? In that case, all you need is to return `oc` and make `oc` return `False` on `bool(oc)` when something bad happens. As I said, you tell me the approach you want to use and we go from there.

utils/oc.py
arilivigni: Sure thing!

utils/oc.py
arilivigni: So remove the check the in openshift_base.py class for oc and handle this as Amadaor mentioned for `None` in the oc.py class?

reconcile/openshift_base.py
arilivigni: Same question as above.

utils/oc.py
maorfr: I'll let @apahim answer

apahim: not sure I got the question...

utils/oc.py
arilivigni: I want to handle all the error paths including `None` so if I remove `OCFail` can this be done directly in `init_oc_client`?  I thought that was the point of a separate class, unless that class is renamed and handles `False` and `None`.

utils/oc.py
arilivigni: Comments above from your analysis of handling all error paths.

utils/oc.py
apahim: I think that what we are proposing here is the latter. To make `OCFail` more generic and cover all the cases where `OC()` won't be instantiated.
Meaning that instead of:
```python
            if oc is None:
                msg = f"[{cluster}] cluster skipped."
                logging.debug(msg)
                continue
            if not oc:
                ri.register_error()
                msg = f"[{cluster}] cluster has no automationToken."
                logging.error(msg)
                continue
```
you would have something like this pseudocode:
```python
            if not oc:
                if oc.am_i_a_bad_boy:
                    ri.register_error()            
                continue
```
But really, the current state is already good enough IMO. Feel free to consider the current code reviewable so we can get this merged and iterate later.

arilivigni: Thanks so much @apahim I implemented this in my latest change.  Now I understand the need for extra info and how to use it.  Let me know your thought on this latest change.

utils/oc.py
arilivigni: @maorfr I took some updated suggestions from @apahim which I think round out this solution even better. Let me know your thoughts.

reconcile/openshift_base.py
arilivigni: The new solution should cover this case as well

apahim (changes requested):
A bit more polishing and we are good to go. I will run some tests on my machine meanwhile.

reconcile/openshift_base.py
```diff
@@ -109,14 +105,10 @@ def init_specs_to_fetch(ri, oc_map,
...
-            if oc is None:
-                msg = f"[{cluster}] cluster skipped."
-                logging.debug(msg)
-                continue
-            if oc is False:
-                ri.register_error()
-                msg = f"[{cluster}] cluster has no automationToken."
-                logging.error(msg)
+            if not oc:
+                if oc.log_level == 40:
```
apahim, line 109: what about CRITICAL? FATAL? Basically anything from ERROR up? THey are even worse and should also make the integration fail.

utils/oc.py
```diff
@@ -566,3 +584,12 @@ def cleanup(self):
         for oc in self.oc_map.values():
             if oc:
                 oc.cleanup()
+
+
+class OCLogMsg:
```
apahim, line 589: Name is still not in line with the pupose here... don't have a better suggestion though. Maybe `OCFalse`? Not sure :)

utils/oc.py
```diff
@@ -529,12 +538,18 @@ def init_oc_client(self, cluster_info):
...
+            try:
+                oc_client = OC(server_url, token, jump_host,
+                               settings=self.settings,
+                               init_projects=self.init_projects,
+                               init_api_resources=self.init_api_resources)
+                self.set_oc(cluster, oc_client)
+            except StatusCodeError as e:
+                logging.error('The cluster is unreachable: error %s', str(e))
+                self.set_oc(cluster,
+                            OCLogMsg(log_level=40,
```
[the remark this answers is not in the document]
arilivigni: That is the log level in python since it takes an integer.  `ERROR=40` `DEBUG=10`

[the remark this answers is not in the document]
arilivigni: @apahim I made the change to use the contanst and convert with getattr

reconcile/openshift_base.py
arilivigni: So is it better to use integer if I am checking for ERROR and higher
`if oc.log_level >= 40`
or preferable to just do
`if oc.log_level == 'ERROR' or oc.log_level == 'CRITICAL' or oc.log_level == 'FATAL':`

utils/oc.py
arilivigni: We are returning a log_level and msg though so why isn't it in line?

reconcile/openshift_base.py
arilivigni: I went with option 2 from above

utils/oc.py
apahim: This class is not a log message, is a class that represents we are not getting an instance of OC. It does carry content used for generating log messages, but that's not the core functionality here. The core functionality is to flag OC was not given.
Anyway, leave it. We can iterate on that, if we ever come to it..

arilivigni: I am fine with any name doesn't matter to me, but it made more sense based on the extra info, but you are correct we only return `False`

reconcile/openshift_base.py
apahim: Neither. I mean, create the instance like this:
```python
OCLogMsg(log_level=logging.ERROR, message=...)
```
and check like this:
```python
if oc.log_level >= logging.ERROR:
    ...
```

arilivigni: `FATAL` is not a logging constant
https://docs.python.org/3/library/logging.html#levels

apahim: It is: https://github.com/python/cpython/blob/master/Lib/logging/__init__.py#L92

arilivigni: :+1:

utils/oc.py
arilivigni: We decided on `OCLogMsg`

apahim:
Looks like it will work. I will verify soon. Meanwhile, please squash/rebase.

jfchevrette:
It seems very non pythonic and "hackish" to encapsulate a bool state and a message in a class like `OCLogMsg` here. Depending on where you look at the code it acts like a bool, or an (log-related) object that has a message in it.

Could you have extended the `OC` class to have `status` and `reason` field in the OC client object and check for that wherever you're checking `oc is None` and `oc is False` (ex: `if oc.state is False` -> log message using `oc.reason`)

Otherwise the code and logic look good and I'm OK with it as it is :+1:

jfchevrette (approved):

apahim:
Then python is non pythonic :)
Bear with me:
```python
>>> bool('foo')
True
>>> bool('')
False
>>> bool(5)
True
>>> bool(0)
False
>>> bool([1,2,3])
True
>>> bool([])
False
```
The "truth value testing" protocol is implemented either explicetly or implicitely on every single object, not only on boolean types, and it always has a meaning. I don't see why implementing it explicitely on a custom class is hackish.
Indeed, there are other ways to solve this. The one you just mentioned was also discussed in this PR at some point, along with 3 or 4 other ways.

apahim:
@arilivigni this is not working... you lost some code after the latest merge. Rebase would have been safer. Please have a look.

utils/oc.py
```diff
@@ -529,12 +538,18 @@ def init_oc_client(self, cluster_info):
...
-                   init_projects=self.init_projects,
-                   init_api_resources=self.init_api_resources))
+            try:
+                oc_client = OC(server_url, token, jump_host,
+                               settings=self.settings,
+                               init_projects=self.init_projects,
+                               init_api_resources=self.init_api_resources)
+                self.set_oc(cluster, oc_client)
+            except StatusCodeError as e:
+                logging.error('The cluster is unreachable: error %s', str(e))
```
apahim, line 548: two log messages? I believe this one can be removed.

apahim:
So, tested it, it looks nice:
```
$ qontract-reconcile --config config.toml --dry-run --log-level INFO openshift-resources
[2020-12-02 12:23:31] [INFO] [gql.py:init_from_config:179] - using gql endpoint http://localhost:4000/graphqlsha/f028436c21f1c3bf50095a0ebbe467864f88de5c9505796f9a52727e301e926e
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
[2020-12-02 12:25:11] [ERROR] [openshift_base.py:init_specs_to_fetch:61] - [app-sre-stage-01] is unreachable
$ echo $?
1
```
Please remove the extra log message and rebase.

apahim (approved):

utils/oc.py
arilivigni: sure thing.  I did like how it listed the full server address but cluster name is sufficient.

apahim: you can still have it, I'm only asking for de-duplication. content is up to you.

PR #1216, opened 2020-11-26
sql-query: multiple queries in the same spec

apahim:
Depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/12421


maorfr (approved):

maorfr:
thank you, as always, for the great commits!

rporres (approved):
I like the idea of multiple `psql|mysql` commands to be able to time every query separately. Good stuff! :)

reconcile/sql_query.py
```diff
@@ -208,45 +217,64 @@ def collect_queries(query_name=None):
...
+
+def filesystem_closing_message():
     return [
-        '> /tmp/query-result.txt;',
         'echo;',
         'echo Get the sql-query results with:;',
         'echo;',
-        'echo  oc rsh --shell=/bin/bash ${HOSTNAME} ',
-        'cat /tmp/query-result.txt;',
+        'echo  oc cp ${HOSTNAME}:/tmp/query-result.txt ',
```
BumbleFeng, line 276: I get a warning when running this command and it looks like it fails
`tar: Removing leading '/' from member names`
Could we remove the leading '/'?

apahim: You tell me :)

PR #1217, opened 2020-12-01
[saasherder] retry image init

maorfr:
this PR will make the `Image` action be retried to retry things like Quay timeouts.
example: https://ci.int.devshift.net/job/openshift-saas-deploy-notifications-insights-stage/8/console

### DIFF
--- a/utils/saasherder.py
+++ b/utils/saasherder.py
@@ -389,7 +389,11 @@ class SaasHerder():
         return images
 
     @staticmethod
-    def _check_image(image, image_patterns, image_auth, error_prefix):
+    @retry()
+    def _is_image_valid(image, image_auth):
+        return Image(image, **image_auth)
+
+    def _check_image(self, image, image_patterns, image_auth, error_prefix):
         error = False
         if image_patterns and \
                 not any(image.startswith(p) for p in image_patterns):
@@ -397,7 +401,7 @@ class SaasHerder():
             logging.error(
                 f"{error_prefix} Image is not in imagePatterns: {image}")
         try:
-            valid = Image(image, **image_auth)
+            valid = self._is_image_valid(image, image_auth)
             if not valid:
                 error = True
                 logging.error(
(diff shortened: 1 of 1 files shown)


utils/saasherder.py
```diff
@@ -389,15 +389,19 @@ def _collect_images(self, resource):
         return images
 
     @staticmethod
-    def _check_image(image, image_patterns, image_auth, error_prefix):
+    @retry()
```
apahim, line 392: we already retry 5x: https://github.com/app-sre/sretoolbox/blob/master/sretoolbox/container/image.py#L276

utils/saasherder.py
```diff
@@ -389,15 +389,19 @@ def _collect_images(self, resource):
...
+    def _is_image_valid(image, image_auth):
+        return Image(image, **image_auth)
```
apahim, line 394: this is not checking is the image is valid. The check happens on bool(Image()), not on instance creation.

utils/saasherder.py
apahim: if you want to retry more, you can just increase that number ^

maorfr: but does it handle `ConnectionError` like observed here?
https://ci.int.devshift.net/job/openshift-saas-deploy-notifications-insights-stage/8/console

utils/saasherder.py
maorfr: thanks! changed to `return bool(Image(image, **image_auth))`. makes sense?

utils/saasherder.py
maorfr: also, would this make sense?
https://github.com/app-sre/sretoolbox/pull/25

apahim: > but does it handle ConnectionError like observed here?

no

> also, would this make sense?

yes!

utils/saasherder.py
apahim: yes!

apahim:
do you still need this after improving the sretoolbox retry?

maorfr:
probably not :)

maorfr:
replaced by https://github.com/app-sre/sretoolbox/pull/25

PR #1220, opened 2020-12-02
[gitlab-owners] ignore 0 commits MRs and run in parallel for all repos

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2768

### DIFF
--- a/reconcile/gitlab_owners.py
+++ b/reconcile/gitlab_owners.py
@@ -35,8 +35,8 @@ class MRApproval:
         self.dry_run = dry_run
 
         commits = self.mr.commits()
-        if len(commits) != 0:
-            top_commit = next(self.mr.commits())
+        if commits:
+            top_commit = next(commits)
             self.top_commit_created_at = \
                 dateparser.parse(top_commit.created_at)
         else:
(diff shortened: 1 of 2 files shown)


reconcile/gitlab_owners.py
```diff
@@ -32,8 +34,13 @@ def __init__(self, gitlab_client, merge_request, owners, dry_run):
         self.owners = owners
         self.dry_run = dry_run
 
-        top_commit = next(self.mr.commits())
-        self.top_commit_created_at = dateparser.parse(top_commit.created_at)
+        commits = self.mr.commits()
+        if len(commits) != 0:
```
apahim, line 38: `if commits`

reconcile/gitlab_owners.py
```diff
@@ -32,8 +34,13 @@ def __init__(self, gitlab_client, merge_request, owners, dry_run):
...
+            top_commit = next(self.mr.commits())
```
apahim, line 39: now you have `commits`... no need to call `mr.commits()` again, I guess

apahim (approved):
lgtm, thanks for this

reconcile/gitlab_owners.py
maorfr: :+1:

reconcile/gitlab_owners.py
maorfr: too early in the morning

PR #1223, opened 2020-12-03
[APPSRE-2772] Fix logging configuration

apahim:
Before, when we were using subprocess, we would collect the subprocess
messages to STDOUT and log them using this module's logger.

Now, running integrations in the same session, the strategy is
different: we setup the root logger handlers and they will tackle
any messages generated.

### DIFF
--- a/dockerfiles/hack/run-integration.py
+++ b/dockerfiles/hack/run-integration.py
@@ -31,15 +31,15 @@ LOG = logging.getLogger(__name__)
 # Messages to stdout
 STREAM_HANDLER = logging.StreamHandler(sys.stdout)
 STREAM_HANDLER.setFormatter(logging.Formatter(fmt='%(message)s'))
-LOG.addHandler(STREAM_HANDLER)
 
 # Messages to the log file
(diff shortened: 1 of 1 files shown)


dockerfiles/hack/run-integration.py
```diff
@@ -31,15 +31,15 @@
...
 # Messages to the log file
 if LOG_FILE is not None:
     FILE_HANDLER = logging.FileHandler(LOG_FILE)
     FILE_HANDLER.setFormatter(logging.Formatter(fmt='%(message)s'))
-    LOG.addHandler(FILE_HANDLER)
 
-LOG.setLevel(logging.INFO)
+# Setting up the root logger
+logging.basicConfig(level=logging.INFO,
+                    handlers=(STREAM_HANDLER, FILE_HANDLER))
```
BumbleFeng, line 42: I have a question. What if LOG_FILE is None? could handlers contain an undefined one?

apahim: I don't think so. Good catch! Feel free to fix it :)

PR #1227, opened 2020-12-04
oc int bug fix for openshift-saas-deploy-wrapper integration

arilivigni:
We were failing with the error below when running the openshift-saas-deploy-wrapper integration.
`utils.oc.StatusCodeError: [server]: Unable to connect to the server: dial tcp: lookup server on 10.5.30.160:53: no such host`

### DIFF
--- a/utils/oc.py
+++ b/utils/oc.py
@@ -73,7 +73,8 @@ class OC(object):
 
         self.oc_base_cmd = oc_base_cmd
         # calling get_version to check if cluster is reachable
-        self.get_version()
+        if server != 'server':
+            self.get_version()
         self.init_projects = init_projects
         if self.init_projects:
             self.projects = [p['metadata']['name']


BumbleFeng:
saasherder need oc cli to test the template. It point to a empty cluster and there is no need for testing reachable in this case
https://github.com/app-sre/qontract-reconcile/blob/master/utils/saasherder.py#L328

Could you also add a comment in the code itself?
just to save folks time when going through it.

utils/oc.py
```diff
@@ -73,7 +73,8 @@ def __init__(self, server, token, jh=None, settings=None,
 
         self.oc_base_cmd = oc_base_cmd
         # calling get_version to check if cluster is reachable
-        self.get_version()
+        if server != 'server':
+            self.get_version()
```
apahim, line 77: This is way too fragile.

apahim:
There is another issue here: https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/openshift_base.py#L201
You can reproduce that issue with the sql-query integration.

utils/oc.py
arilivigni: @BumbleFeng @apahim is not happy with this solution, but when I have a better one I will add comments.

apahim: I mean, that could have been any string... There's nothing special about `'server'`.

BumbleFeng: 'server' comes from saasherder https://github.com/app-sre/qontract-reconcile/blob/master/utils/saasherder.py#L328
It only needs to run oc to verify the template file, doesn't need to actually run towards a real cluster.
We could either skip check  get_version for saasherder or catch and swallow the error when init oc in saasherder.

BumbleFeng: wdyt?
`
        if server.startswith('https://api'):
            self.get_version()
`

apahim: Let's take a step back here for a minute. Do we understand why we are setting the server to `'server' `? Is that the right thing to do?

arilivigni: @apahim  If it is `server` it isn't valid and we would want to skip it.  Is there a positive case where we would just return `server`?

arilivigni: No, I don't think so I am experimenting with a try block now.  I will try this and test `sql-query` and all `openshift-*` integrations.  Unless you think this is incorrect?

arilivigni:
closing this for this PR that @maorfr has done
https://github.com/app-sre/qontract-reconcile/pull/1228

PR #1228, opened 2020-12-06
allow OC init to skip version

maorfr:
fixes #1214 

since we use OC in saasherder process_template to do `oc process`, we are initializing a "local" oc client (without an actual server). this PR will skip `get_version` if OC is initialized with `local=True`.

### DIFF
--- a/utils/saasherder.py
+++ b/utils/saasherder.py
@@ -325,7 +325,7 @@ class SaasHerder():
                                 get_commit_sha_options)
                             consolidated_parameters['IMAGE_TAG'] = image_tag
 
-            oc = OC('server', 'token')
+            oc = OC('server', 'token', local=True)
             try:
                 resources = oc.process(template, consolidated_parameters)
             except StatusCodeError as e:
--- a/utils/oc.py
+++ b/utils/oc.py
@@ -58,7 +58,8 @@ class JobNotRunningError(Exception):
 
 class OC(object):
     def __init__(self, server, token, jh=None, settings=None,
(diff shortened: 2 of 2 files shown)


utils/saasherder.py
```diff
@@ -325,7 +325,7 @@ def _process_template(self, options):
                                 get_commit_sha_options)
                             consolidated_parameters['IMAGE_TAG'] = image_tag
 
-            oc = OC('server', 'token')
+            oc = OC('server', 'token', local=True)
```
apahim, line 328: Can we do `OC(None, None)` and consider is local when server is None?

apahim (approved):
Lgtm. I'd recommend either make server optional in the current implementation, with the extra parameter, or consider local when server is None and avoid the extra param.
Anyway, the current code is already good enough.

utils/saasherder.py
maorfr: i just don't want to dive all the way into https://github.com/app-sre/qontract-reconcile/blob/542b0bbf1df84871ce6d39eb8abce8db6039fe7b/utils/oc.py#L439

apahim: Fair enough.

arilivigni:
@maorfr this looks good to me and easier than what I was trying to implement for this issue. 
I believe there are other issues with unreachable clusters and other integrations.  I run through those with this change to identify the issue for those as well.

arilivigni (approved):
lgtm

PR #1230, opened 2020-12-07
[oc] create group if not exists

maorfr:
the openshift-groups integration thinks that an empty group does not exist.
instead of diving into it, adding a check if the group exists before creating it (which is good practice in any case).


apahim (approved):
lgtm

PR #1232, opened 2020-12-08
catch issues with integrations that use oc_map.get()

arilivigni:
After making the changes for unreachable clusters there are several integrations that end up calling the OC class and specifically `oc_map.get()` without these fixes, these integrations get stack traces and error.

Amador found some of these issues and I also did some searching to find out all the places this method is used.  These incorporate those changes.

Amador also noted that there was an issue with sql-query, but I was unable to reproduce:
I put this in my local app-interface

`data/app-interface/sql-queries/app-sre/2020-12-07-dashdotdb-stage-clusters.yml `
```
---
$schema: /app-interface/app-interface-sql-query-1.yml
labels: {}
name: 2020-12-07-dashdotdb-stage-clusters
namespace:
  $ref: /services/dashdot/namespaces/app-sre-stage-01.yml
identifier: dashdotdb-stage
output: stdout
query: |
  SELECT name from cluster;
```
```
/etc/hosts
127.0.0.1 api.app-sre-stage-0.k3s7.p1.openshiftapps.com
```

`APP_INTERFACE_STATE_BUCKET_ACCOUNT=app-sre  APP_INTERFACE_STATE_BUCKET=app-interface-production qontract-reconcile --config ../config.debug.toml --dry-run --log-level INFO sql-query`

Please provide feedback since I am sure some of this can be done better.

### DIFF
--- a/e2e_tests/create_namespace.py
+++ b/e2e_tests/create_namespace.py
@@ -15,7 +15,7 @@ def test_cluster(cluster, oc_map, ns_under_test):
     oc = oc_map.get(cluster)
     if not oc:
         logging.log(level=oc.log_level, msg=oc.message)
-        return
+        return None
     logging.info("[{}] Creating namespace '{}'".format(
         cluster, ns_under_test
     ))
--- a/e2e_tests/create_namespace.py
+++ b/e2e_tests/create_namespace.py
@@ -13,6 +13,9 @@ QONTRACT_E2E_TEST = 'create-namespace'
 
 def test_cluster(cluster, oc_map, ns_under_test):
     oc = oc_map.get(cluster)
+    if not oc:
+        logging.log(level=oc.log_level, msg=oc.message)
+        return
     logging.info("[{}] Creating namespace '{}'".format(
         cluster, ns_under_test
     ))
--- a/e2e_tests/create_namespace.py
+++ b/e2e_tests/create_namespace.py
@@ -13,9 +13,6 @@ QONTRACT_E2E_TEST = 'create-namespace'
 
 def test_cluster(cluster, oc_map, ns_under_test):
     oc = oc_map.get(cluster)
-    if not oc:
-        logging.log(level=oc.log_level, msg=oc.message)
-        return None
     logging.info("[{}] Creating namespace '{}'".format(
         cluster, ns_under_test
     ))
--- a/e2e_tests/dedicated_admin_rolebindings.py
+++ b/e2e_tests/dedicated_admin_rolebindings.py
@@ -12,6 +12,9 @@ QONTRACT_E2E_TEST = 'dedicated-admin-rolebindings'
 
 def test_cluster(cluster, oc_map, pattern):
     oc = oc_map.get(cluster)
+    if not oc:
+        logging.log(level=oc.log_level, msg=oc.message)
+        return
     logging.info("[{}] validating RoleBindings".format(cluster))
 
     projects = [p['metadata']['name']
--- a/e2e_tests/dedicated_admin_rolebindings.py
+++ b/e2e_tests/dedicated_admin_rolebindings.py
@@ -14,7 +14,7 @@ def test_cluster(cluster, oc_map, pattern):
     oc = oc_map.get(cluster)
     if not oc:
         logging.log(level=oc.log_level, msg=oc.message)
-        return
+        return None
     logging.info("[{}] validating RoleBindings".format(cluster))
 
     projects = [p['metadata']['name']
--- a/e2e_tests/dedicated_admin_rolebindings.py
+++ b/e2e_tests/dedicated_admin_rolebindings.py
@@ -12,9 +12,6 @@ QONTRACT_E2E_TEST = 'dedicated-admin-rolebindings'
 
 def test_cluster(cluster, oc_map, pattern):
     oc = oc_map.get(cluster)
-    if not oc:
-        logging.log(level=oc.log_level, msg=oc.message)
-        return None
     logging.info("[{}] validating RoleBindings".format(cluster))
 
     projects = [p['metadata']['name']
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -201,7 +201,7 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
     oc = oc_map.get(cluster)
     if not oc:
         logging.log(level=oc.log_level, msg=oc.message)
-        return
+        return None
     if not dry_run:
         annotated = resource.annotate()
         # skip if namespace does not exist (as it will soon)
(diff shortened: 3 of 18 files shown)


e2e_tests/create_namespace.py
```diff
@@ -13,6 +13,9 @@
 
 def test_cluster(cluster, oc_map, ns_under_test):
     oc = oc_map.get(cluster)
+    if not oc:
+        logging.log(level=oc.log_level, msg=oc.message)
+        return None
```
apahim, line 18: So, this is a test... when we can't get an actionable `OC()`, doesn't that mean that the test failed?
At the end, this is just about the message we log, so no big deal.

e2e_tests/dedicated_admin_rolebindings.py
```diff
@@ -12,6 +12,9 @@
 
 def test_cluster(cluster, oc_map, pattern):
     oc = oc_map.get(cluster)
+    if not oc:
+        logging.log(level=oc.log_level, msg=oc.message)
```
apahim, line 16: same here

reconcile/openshift_base.py
```diff
@@ -233,6 +236,10 @@ def delete(dry_run, oc_map, cluster, namespace, resource_type, name,
         logging.error('\'delete\' action is disabled due to previous errors.')
         return
 
+    oc = oc_map.get(cluster)
```
apahim, line 239: now you have the `oc` variable...

reconcile/openshift_base.py
```diff
@@ -233,6 +236,10 @@ def delete(dry_run, oc_map, cluster, namespace, resource_type, name,
...
+    if not oc:
+        logging.log(level=oc.log_level, msg=oc.message)
+        return None
     if not dry_run:
         oc_map.get(cluster).delete(namespace, resource_type, name)
```
apahim, line 244: ...you should reuse it here

reconcile/openshift_groups.py
```diff
@@ -44,6 +44,9 @@ def get_cluster_state(group_items, oc_map):
     results = []
     cluster = group_items["cluster"]
     oc = oc_map.get(cluster)
+    if not oc:
+        logging.log(level=oc.log_level, msg=oc.message)
+        return None
```
apahim, line 49: this will break ahead, because the `fetch_current_state()` loops through the return from `get_cluster_state()` , assuming it's a list (line 94 below).

reconcile/openshift_users.py
```diff
@@ -14,6 +14,9 @@
 
 def get_cluster_users(cluster, oc_map):
     oc = oc_map.get(cluster)
+    if not oc:
+        logging.log(level=oc.log_level, msg=oc.message)
+        return None
```
apahim, line 19: this breaks ahead in line 37

reconcile/openshift_serviceaccount_tokens.py
```diff
@@ -42,6 +43,11 @@ def fetch_desired_state(namespaces, ri, oc_map):
             sa_namespace_name = sa_namespace_info['name']
             sa_cluster_name = sa_namespace_info['cluster']['name']
             oc = oc_map.get(sa_cluster_name)
+            if not oc_map.get:
```
apahim, line 46: Huh? wrong call here

utils/openshift_resource.py
```diff
@@ -406,16 +406,24 @@ def initialize_resource_type(self, cluster, namespace, resource_type):
 
     def add_desired(self, cluster, namespace, resource_type, name, value):
         with self._lock:
-            desired = \
-                self._clusters[cluster][namespace][resource_type]['desired']
+            try:
```
apahim, line 409: why do we need this?

apahim: I think we never get to run this if the `oc_map.get(cluster)` gives us an `OcLogMsg` instance.

utils/openshift_resource.py
```diff
@@ -406,16 +406,24 @@ def initialize_resource_type(self, cluster, namespace, resource_type):
...
+                desired = \
+                    (self._clusters[cluster][namespace][resource_type]
+                        ['desired'])
+            except KeyError:
+                return False
```
apahim, line 414: Even if we need it, we should probably return `None`.

apahim (changes requested):
Looks mostly good. A couple of fixes and I believe we will be good to go.

reconcile/openshift_serviceaccount_tokens.py
arilivigni: :facepalm:  oops sorry copy/paste error. I will update

utils/openshift_resource.py
arilivigni: `openshift-saas-deploy-wrapper` integration fails without it.  I will retest without the try/except block, but I got hung up here with an exception until I added this.

utils/openshift_resource.py
arilivigni: I based it on other `KeyError` which used `False` but will switch to `None`

utils/openshift_resource.py
apahim: maybe that was before all the other fixes? yes, check it with all the other fixes here in place please.

utils/openshift_resource.py
apahim: All I care is what the function is giving back to the caller so it does not break.

apahim: But I don't think that the return from `add_desired()` is ever used, so there's no reason to return a boolean.

utils/openshift_resource.py
arilivigni: :+1:

utils/openshift_resource.py
arilivigni: :+1:

reconcile/openshift_namespaces.py
```diff
@@ -103,7 +103,10 @@ def create_new_project(spec, oc_map):
     cluster = spec['cluster']
     namespace = spec['namespace']
 
-    oc_map.get(cluster).new_project(namespace)
+    oc = oc_map.get(cluster).new_project(namespace)
+    if not oc:
+        logging.log(level=oc.log_level, msg=oc.message)
+        return None
```
apahim, line 109: I meant:
```
    oc = oc_map.get(cluster)
    if not oc:
        return None
    oc.new_project(namespace)
```

arilivigni: :+1:

apahim: `oc_map.get(cluster).new_project(namespace)` does not give you an `OC()`/`OcLogMsg()` instance.

arilivigni: Just made this change :+1:

apahim (approved):
lgtm

PR #1233, opened 2020-12-08
Dvo dashdot

mmclanerh:
This adds support for DeplomentValidation metric ingestion and processing.


reconcile/dashdotdb_dvo.py
```diff
@@ -0,0 +1,105 @@
...
+            except requests.exceptions.HTTPError as details:
+                LOG.error('DVO: error posting %s - %s', cluster, details)
+
+        LOG.info('DV: cluster %s synced', cluster)
+        return response
+
+    def promql(url, query, auth=None):
+        url = os.path.join(url, 'api/v1/query')
+        if auth is None:
+            auth = {}
```
apahim, line 55: `auth` can't be a dict. Just leave it as `None`. The only alternative is a tuple.

reconcile/dashdotdb_dvo.py
```diff
@@ -0,0 +1,105 @@
...
+                                           self.dashdotdb_pass))
+            try:
+                response.raise_for_status()
+            except requests.exceptions.HTTPError as details:
+                LOG.error('DVO: error posting %s - %s', cluster, details)
+
+        LOG.info('DV: cluster %s synced', cluster)
+        return response
+
+    def promql(url, query, auth=None):
```
apahim, line 52: Does it make sense to use https://pypi.org/project/prometheus-api-client/?

apahim:
ping

reconcile/dashdotdb_dvo.py
mmclanerh: For sure -- needed to re-do auth anyway to use bearer token vs user+pass auth. This was removed.

reconcile/dashdotdb_dvo.py
mmclanerh: I started out using this but had some issues with how long things were taking / timeouts, unfortunately found these were unrelated to the prom lib but also all I'm doing with prom is fetching a raw json blob vs doing any prom object operations so at this time don't need the entire library anyway.

reconcile/dashdotdb_dvo.py
mmclanerh: resolving, as this was removed a bit ago

reconcile/dashdotdb_dvo.py
mmclanerh: replaced with self._promget()

apahim: fair enough

reconcile/dashdotdb_dvo.py
```diff
@@ -0,0 +1,126 @@
...
+            try:
+                response.raise_for_status()
+            except requests.exceptions.HTTPError as details:
+                LOG.error('%s error posting %s - %s',
+                          self.logmarker, cluster, details)
+
+        LOG.info('%s cluster %s synced', self.logmarker, cluster)
+        return response
+
+    def _promget(self, url, query, token=None):
```
apahim, line 52: token should be mandatory OR the Authorization header should not be defined when the token None.

reconcile/utils/oc.py
```diff
@@ -267,7 +267,8 @@ def get_api_resources(self):
         return [r.split()[-1] for r in results]
 
     def get_version(self):
-        cmd = ['version']
+        # this is actually a 10 second timeout, because: oc reasons
+        cmd = ['version', '--request-timeout=5']
```
apahim, line 271: out of curiosity, how long it takes otherwise?
@arilivigni ptal

reconcile/dashdotdb_dvo.py
```diff
@@ -0,0 +1,126 @@
...
+        LOG.debug('%s processing %s, %s', self.logmarker, cluster, validation)
+        cluster_promurl = "prometheus." + cluster + ".devshift.net"
+        promquery = "deployment_validation_"+validation+"_validation"
+        cluster_promuri = "query=" + promquery
+        cluster_promtoken = self._get_automationtoken(cluster)
+        try:
+            deploymentvalidation = self._promget(url=cluster_promurl,
+                                                 query=cluster_promuri,
+                                                 token=cluster_promtoken)
+        except StatusCodeError:
```
rporres, line 94: I don't understand how `self._promget` is going to raise a `StatusCodeError`. You may not want to wrap `response.raise_for_status()` in line 67 in a try/except block and capture here the `requests.exceptions.HTTPError` exception

reconcile/dashdotdb_dvo.py
```diff
@@ -0,0 +1,126 @@
...
+        return response
+
+    def _get_automationtoken(self, cluster):
+        autotoken_reader = SecretReader(settings=self.settings)
+        autotoken = {'path': "app-sre/creds/kube-configs/" + cluster,
+                     'field': "token"}
+        token = autotoken_reader.read(autotoken)
+        return token
+
+    def _get_deploymentvalidation(self, cluster, validation, oc_map):
```
rporres, line 84: I don't see where you're using `oc_map` in this function.

reconcile/dashdotdb_dvo.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def _promget(self, url, query, token=None):
+        uri = 'api/v1/query'
+        url = urljoin('https://'+url, uri)
+        headers = {
+            "accept": "application/json",
+            "Authorization": "Bearer " + token,
+        }
+        params = {'query': query}
+        LOG.info('%s Fetching prom payload from %s?%s',
```
rporres, line 60: I would do this a debug message or this integration logs are going to be very verbose 😅

reconcile/dashdotdb_dvo.py
```diff
@@ -0,0 +1,126 @@
...
+        self.dashdotdb_user = secret_content['username']
+        self.dashdotdb_pass = secret_content['password']
+        self.logmarker = "DDDB_DVO:"
+
+    def _post(self, deploymentvalidation):
+        if deploymentvalidation is None:
+            return None
+        cluster = deploymentvalidation['cluster']
+        dvdata = deploymentvalidation['data']
+        LOG.info('%s posting validations for %s', self.logmarker, cluster)
```
rporres, line 35: I would do this a debug message or this integration logs are going to be very verbose 😅

rporres (changes requested):
Looks good. A few comments here and there. I like a lot the use of a class for the integration code, it helps simplifying a lot the passing of variables.

reconcile/dashdotdb_dvo.py
mmclanerh: agreed. a carry over from local testing where i wasn't using a bearer token. i'll make this consistent by not adding headers if  token=none.

reconcile/utils/oc.py
mmclanerh: this was taking 10-15 minutes per run locally, as several of the clusters aren't or weren't reachable (when tunnel is off). this also happens in our production code anywhere oc is used and a cluster is not reachable by OC. in nearly all cases a response comes back normally under 4 seconds. the timeout options in the oc client are not granular. also (an openshift "feature"): the server can respond back with a retry timeout that the client honors.. so any cluster can say: hey, there was an error, but retry in 60 seconds. it's a bit of a mess, but a ~10 second timeout (--request-timeout=5) fit the 98 percentile of connect cases observed. this version check also is invoked even if the don't-init-resources parameter is passed: https://github.com/app-sre/qontract-reconcile/blob/42d8ec48b20f5fef7b29bd24ec20397bff44efa4/reconcile/utils/oc.py#L77-L78  

development observation: this didn't have any debug-visbility around it, as the oc call is a silent exec().. so for a while i thought it was either vault or my original usage of a prom lib causing long run times. i would very much like to replace all usage of the oc binary with the maintained openshift py client (which wraps the kube client): https://github.com/openshift/openshift-client-python. obviously the scope of that is well outside this effort. :)

reconcile/dashdotdb_dvo.py
```diff
@@ -0,0 +1,126 @@
...
+    def _get_automationtoken(self, cluster):
+        autotoken_reader = SecretReader(settings=self.settings)
+        autotoken = {'path': "app-sre/creds/kube-configs/" + cluster,
+                     'field': "token"}
+        token = autotoken_reader.read(autotoken)
+        return token
+
+    def _get_deploymentvalidation(self, cluster, validation, oc_map):
+        LOG.debug('%s processing %s, %s', self.logmarker, cluster, validation)
+        cluster_promurl = "prometheus." + cluster + ".devshift.net"
```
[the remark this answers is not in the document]
mmclanerh: you'd think so.. but several of the clusters don't have an application prometheusUrl defined (as they are network-isolated). note: the way prometheus urls are managed in grafana is %100 manually editing the secrets file.. and i wanted to aim for the least hands-on approach for maintaining configuration. this seemed like an okay way to derive the prometheusUrl per cluster.. with the caveat that some urls are invalid and therefore would not resolve or not connect, and can then be skipped.

i could add the prometheusUrl property to the generic cluster-fetch query, and then skip any clusters where this is none. maybe that's better, though i'm not sure if that has impact to anything consuming that query. note: i wish there was a way to schema+version these queries, to prevent backwards compatibility issues:  
https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/queries.py#L268-L270

wdyt? /cc @apahim

reconcile/dashdotdb_dvo.py
```diff
@@ -0,0 +1,126 @@
...
+        promquery = "deployment_validation_"+validation+"_validation"
+        cluster_promuri = "query=" + promquery
```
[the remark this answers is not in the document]
mmclanerh: good point. i will clean that up and/or remove this code in favor of an item discussed in another comment, as well as verify standards are met across this new class.

reconcile/dashdotdb_dvo.py
mmclanerh: oc_map used by the threaded iteration here [1], looked like this was "the way" iterating over the clusters was being done elsewhere.
[1] https://github.com/mmclanerh/qontract-reconcile/blob/b8d5b0f3eeb81de5d6e1e8c0895759089c39150c/reconcile/dashdotdb_dvo.py#L114-L118

reconcile/dashdotdb_dvo.py
```diff
@@ -0,0 +1,126 @@
...
+        cluster_promtoken = self._get_automationtoken(cluster)
+        try:
+            deploymentvalidation = self._promget(url=cluster_promurl,
+                                                 query=cluster_promuri,
+                                                 token=cluster_promtoken)
+        except StatusCodeError:
+            LOG.info('%s Unable to fetch data for %s', self.logmarker, cluster)
+            return None
+
+        if not deploymentvalidation:
```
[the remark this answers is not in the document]
mmclanerh: 🤕 oh, totally don't. good catch. will fix.

reconcile/dashdotdb_dvo.py
mmclanerh: agreed. will change to debug. as one does, littered code for a bit with LOG.info()s during development.. as running with a debug log level was too insanely verbose. development note: this is also how the self.logmarker log prefix came to be, it was easier to filter the debug fire hose that way..

reconcile/dashdotdb_dvo.py
mmclanerh: agreed. will fix. ..and gives me an idea for backlogging: perhaps exposing some runtime activity metrics would be good.

reconcile/dashdotdb_dvo.py
```diff
@@ -0,0 +1,126 @@
...
+        cluster = deploymentvalidation['cluster']
+        dvdata = deploymentvalidation['data']
+        LOG.info('%s posting validations for %s', self.logmarker, cluster)
+        response = None
+        if not self.dry_run:
+            endpoint = (f'{self.dashdotdb_url}/api/v1/'
+                        f'deploymentvalidation/{cluster}')
+            response = requests.post(url=endpoint, json=dvdata,
+                                     auth=(self.dashdotdb_user,
+                                           self.dashdotdb_pass))
```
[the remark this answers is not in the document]
mmclanerh: great question, i'm glad you asked! it's only because this is a magical application that will never encounter failures, ever, of course. ..i'll add a timeout parameter here. :)

reconcile/dashdotdb_dvo.py
```diff
@@ -0,0 +1,126 @@
...
+            try:
+                response.raise_for_status()
+            except requests.exceptions.HTTPError as details:
```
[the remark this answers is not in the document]
mmclanerh: ah, good point. i had started down the path of enumerating different error-types exception handling and (probably) meant to roll those all back to RequestException. the intent here is to be a catchall, so, i'll change to that.

reconcile/dashdotdb_dvo.py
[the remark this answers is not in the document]
rporres: I think it is safer than building it manually as app-interface is the source of truth. If it doesn't exist, it should mean that there are no prometheus there...

reconcile/dashdotdb_dvo.py
rporres: You don't need `oc_map` in the `_get_deploymentvalidation` function. As you can see in https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/threaded.py#L17, `validation` and `oc_map` are the `**kwargs` that are passed to the `func`. You should be able to safely remove from here and from the `threaded` call

reconcile/dashdotdb_dvo.py
mmclanerh: changed: don't pass auth header if token=None

reconcile/utils/oc.py
mmclanerh: @arilivigni  this make sense to you? ok to resolve?

reconcile/dashdotdb_dvo.py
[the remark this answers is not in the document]
mmclanerh: adding property to query.py.. need to retool _get_deploymentvalidation slightly to read that property and skip the cluster if promurl isn't set.

reconcile/dashdotdb_dvo.py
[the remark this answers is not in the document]
mmclanerh: this piece is getting removed, will resolve once that's solid.

reconcile/dashdotdb_dvo.py
mmclanerh: fixed: removed try block from _promget and handling in the caller, with catchall

reconcile/dashdotdb_dvo.py
mmclanerh: yeah reviewing.. this was cruft. removed parameter.

reconcile/utils/oc.py
arilivigni: :+1:  sounds good to me

reconcile/dashdotdb_dvo.py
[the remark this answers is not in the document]
mmclanerh: removed test

reconcile/dashdotdb_dvo.py
mmclanerh: changed to debug

reconcile/dashdotdb_dvo.py
mmclanerh: changed to debug.

reconcile/dashdotdb_dvo.py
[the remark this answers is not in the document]
mmclanerh: added timeouts

reconcile/dashdotdb_dvo.py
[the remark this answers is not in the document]
mmclanerh: changed to catchall exception

reconcile/dashdotdb_dvo.py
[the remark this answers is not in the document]
mmclanerh: rebased this entirely to use data from app-interface. if prometheusUrl isn't set, we skip..  however there's still some cases where the prometheusUrl is set, but can't be reached.. so.. neat.

reconcile/dashdotdb_dvo.py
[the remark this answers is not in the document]
mmclanerh: removed but also added f-strings in many places vs lazy-old-bad-habits format

mmclanerh:
was able to test this code against a prod data set and it works, though there is an issue on the receiver (gunicorn) that limits the max post size.. so that will need to be adjusted in the dashdotdb server. there's also a number of prom urls defined that aren't reachable.. so it is what it is.

image

i believe this is good to merge.

rporres (approved):
Ship it!

```

                       _____|\
                  _.--| RPM |:
                 <____|.----||
                        .---''---,
                         ;..__..'    _...
                       ,'/  ;|/..--''    \
                     ,'_/.-/':            :
                _..-'''/  /  |  \    \   _|/|
               \      /-./_ \;   \    \,;'   \
               ,\    / \:  `:\    \   //    `:`.
             ,'  \  /-._;   | :    : ::    ,.   .
           ,'     ::   /`-._| |    | || ' :  `.`.)
        _,'       |;._:: |  | |    | `|   :    `'
      ,'   `.     /   |`-:_ ; |    |  |  : \
      `--.   )   /|-._:    :          |   \ \
         /  /   :_|   ;`-._;   __..--';    : :
        /  (    ;|;-./_  _/.-:'o |   /     ' |
       /  , \._/_/_./--''/_|:|___|_,'        |
      :  /   `'-'--'----'---------'          |
      | :     O ._O   O_. O ._O   O_.      ; ;
      : `.      //    //    //    //     ,' /
    ~~~`.______//____//____//____//_______,'~
              //    //~   //    //
       ~~   _//   _//   _// ~ _//     ~
     ~     / /   / /   / /   / /  ~      ~~
          ~~~   ~~~   ~~~   ~~~
```

apahim:
LGTM, unless the ` client intended to send too large body: 1819715 bytes` error is caused by this client, then please fix it befire merging this.

mmclanerh:
@apahim issue is not due to the client, but is related to auth-proxy (nginx) under dashdotdb limiting post size to the default for nginx.

apahim:
Sure, what I meant was: don't merge it if you know it won't work :)
Any chance to split the data in the client side? Like, per namespace basis or something like that...

mmclanerh:
DashDotDb stage updated to raise post limit: https://github.com/app-sre/dashdotdb/pull/36

Re-basing with latest updates from master and should be able to merge in.

PR #1235, opened 2020-12-09
Fix missing packages

arilivigni:
This doesn't exist
http://mirror.centos.org/centos/8/BaseOS/x86_64/os/Packages/centos-gpg-keys-8.2-2.2004.0.2.el8.noarch.rpm


arilivigni:
not needed any longer.  [This is better](https://github.com/app-sre/qontract-reconcile/pull/1236)

PR #1239, opened 2020-12-14
new command that displays quay mirrors

jmelis:
Example:

```
$ qontract-cli --config config.prod.toml get mirrors
REPO                                               UPSTREAM                                     PUBLIC
-------------------------------------------------  -------------------------------------------  --------
quay.io/REDACTED/REDACTED                          REDACTED                                     True
```

### DIFF
--- a/reconcile/quay_mirror.py
+++ b/reconcile/quay_mirror.py
@@ -57,7 +57,8 @@ class QuayMirror:
                 except SkopeoCmdError as details:
                     _LOG.error('[%s]', details)
 
-    def process_repos_query(self):
+    @staticmethod
+    def process_repos_query():
         apps = queries.get_quay_repos()
 
         summary = defaultdict(list)
--- a/reconcile/quay_mirror.py
+++ b/reconcile/quay_mirror.py
@@ -37,32 +37,6 @@ class QuayMirror:
     }
     """
 
-    QUAY_REPOS_QUERY = """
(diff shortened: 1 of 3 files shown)


reconcile/quay_mirror.py
```diff
@@ -81,11 +56,11 @@ def run(self):
                     _LOG.error('[%s]', details)
 
     def process_repos_query(self):
-        result = self.gqlapi.query(self.QUAY_REPOS_QUERY)
+        apps = queries.get_quay_repos()
```
apahim, line 59: grat, tks for that :)

tools/qontract_cli.py
```diff
@@ -349,6 +349,40 @@ def integrations(ctx):
     print_output(ctx.obj['output'], environments, columns)
 
 
+@get.command()
+@click.pass_context
+def mirrors(ctx):
```
apahim, line 354: there are other mirrors... should we be specific here? I'd expect the command to be `get quay-mirrors`

apahim (approved):
lgtm, just a naming question.

tools/qontract_cli.py
rporres: +1

rporres (commented):
LGTM. Just the naming thing...

apahim (approved):
lgtm

PR #1246, opened 2020-12-15
[WIP] Refactor aws-route53 integration

jfchevrette:
NOTES
* The schema version was not incremented even though the query and schema has changes as the only consumer of this integration is us and for a single zone at this time 
* Because Terraform retrieves the records one by one from AWS, it almost always hits the Route53 API endpoint rate limit and as such takes longer to execute compared to the existing integration based on `Boto3`. The flipside of this is that we now benefit from a true state-based reconciliation backend provided by terraform as opposed to having to build our own. One big win from this change is this makes dealing with non-unique record names much simpler.

CHANGES
* Remove the OOP code related to AWS Route53 objects - The reconcile loop is now much simpler as we're mapping the fields 1-to-1 with those from the terraform `aws_route53_records` module
* The integration now supports weighted records and healthchecks
* Changed `target` & `targets` fields to `records` to match the terraform field for simplicity. Records are now always a list
* Changed `target_cluster` field to `_target_cluster` (starts with underscore) to denote that it is a special field processed differently by the integration (i.e.: not a terraform field)
* Add `_healthcheck` field to allow defining a healthcheck for a record
* Update README
* Added many examples in the README


utils/terrascript_client.py
```diff
@@ -344,6 +352,58 @@ def populate_additional_providers(self, accounts):
...
+                    f"{record_fqdn}_{record['type'].upper()}")
+
+                # Count record names so we can generate unique IDs
+                if record_id not in counts:
+                    counts[record_id] = 0
+                counts[record_id] += 1
+
+                # If more than one record with a given name, append _{count}
+                if counts[record_id] > 1:
+                    record_id = f"{record_id}_{counts[record_id]}"
```
rporres, line 381: Is it important that this can change if a new record with a given name is added after we already have one? there's little we can do to avoid it...

rporres (commented):
It has indeed simplify things. A small comment. LGTM otherwise

reconcile/terraform_aws_route53.py
```diff
@@ -0,0 +1,137 @@
...
+
+            zone_values['records'].append(record)
+
+        desired_state.append(zone_values)
+    return desired_state
+
+
+@defer
+def run(dry_run=False, print_only=False,
+        enable_deletion=False, thread_pool_size=10, defer=None):
```
maorfr, line 93: i assume we would want to change `enable_deletion` to true once we gain confidence?

maorfr (approved):
LGTM

i like the way you setup the desired state to match the populate function. nice!

utils/terrascript_client.py
jfchevrette: Only the 2nd, 3rd, etc... records are going to have a number appended. The original record won't have its name changes and thus shouldn't cause a resource removal/addition.

Did I accurately understand your point or did you mean something different?

reconcile/terraform_aws_route53.py
jfchevrette: That was a (convenient?) oversight from my part. I think it's safe(r) to leave it disabled for the initial run and I'll enabled it shortly after

utils/terrascript_client.py
rporres: Yes, thanks

PR #1247, opened 2020-12-15
update images for sql-query

arilivigni:
Updated build postgres and mariadb images as part of qontract-reconcile since we want to get away from using
mirror images in app-sre and these are dependencies of the `sql-query` integration


Makefile
```diff
@@ -15,9 +18,21 @@ build:
...
+	@docker build -t $(MDB_IMAGE_NAME):latest -f dockerfiles/Dockerfile.mariadb .
+	@docker tag $(MDB_IMAGE_NAME):latest $(MDB_IMAGE_NAME):$(IMAGE_TAG)
+
 push:
 	@docker --config=$(DOCKER_CONF) push $(IMAGE_NAME):latest
-	@docker --config=$(DOCKER_CONF) push $(IMAGE_NAME):$(IMAGE_TAG)
+	@docker --config=$(DOCKER_CONF) push $(IMAGE_NAME):latest
+	
+push-dbs:	
+	@docker --config=$(DOCKER_CONF) push $(POSTGRES_IMAGE_NAME):$(IMAGE_TAG)
```
mmclanerh, line 32: if i'm parsing this correctly, one of these should be latest?

build_deploy_dbs.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+DOCKER_CONF="$PWD/.docker"
+mkdir -p "$DOCKER_CONF"
+docker --config="$DOCKER_CONF" login -u="$QUAY_USER" -p="$QUAY_TOKEN" quay.io
```
mmclanerh, line 5: just be aware that any ci node this runs on is going to persist this to disk, which can add concurrency woes as well as may not be ideal when it comes to secrets at rest. this is probably okay in this context as long as $PWD/.nnn is the job build path in terms of concurrency.

reconcile/sql_query.py
```diff
@@ -34,7 +34,7 @@
     spec:
       containers:
       - name: {{ JOB_NAME }}
-        image: quay.io/app-sre/{{ ENGINE }}:{{ENGINE_VERSION}}
+        image: quay.io/app-sre/qontract-reconcile-{{ENGINE}}:{{ENGINE_VERSION}}
```
apahim, line 37: What about all the other versions?

arilivigni: Isn't this swapped out for the `ENGINE_VERSION`?

jfchevrette: This would only support postgres 9.6  and mariadb 10.3 for now. We also have packages for postgres 10 and 12 which we can add as separate image tags. Looking at some of the previous uses of sql-query I have only seem 9.6 or 'latest' being used (default). For completeness we should also build postgresl 10 and 12

Makefile
```diff
@@ -15,9 +18,21 @@ build:
...
+build-dbs:
+	@docker build -t $(POSTGRES_IMAGE_NAME):latest -f dockerfiles/Dockerfile.postgres .
+	@docker tag $(POSTGRES_IMAGE_NAME):latest $(POSTGRES_IMAGE_NAME):$(IMAGE_TAG)
+	@docker build -t $(MDB_IMAGE_NAME):latest -f dockerfiles/Dockerfile.mariadb .
+	@docker tag $(MDB_IMAGE_NAME):latest $(MDB_IMAGE_NAME):$(IMAGE_TAG)
+
 push:
 	@docker --config=$(DOCKER_CONF) push $(IMAGE_NAME):latest
-	@docker --config=$(DOCKER_CONF) push $(IMAGE_NAME):$(IMAGE_TAG)
+	@docker --config=$(DOCKER_CONF) push $(IMAGE_NAME):latest
```
jfchevrette, line 29: Is this a typo? We don'T want to change how the qontract-reconcile (IMAGE_NAME) is pushed. This would effectively remove the hash based tag (IMAGE_TAG) and push :latest twice?

Makefile
arilivigni: yes thanks will fix

Makefile
```diff
@@ -5,6 +5,9 @@ IMAGE_TEST := reconcile-test
 IMAGE_NAME := quay.io/app-sre/qontract-reconcile
 IMAGE_TAG := $(shell git rev-parse --short=7 HEAD)
 
+POSTGRES_IMAGE_NAME := quay.io/app-sre/qontract-reconcile-postgres
+MDB_IMAGE_NAME := quay.io/app-sre/qontract-reconcile-mariadb
```
jfchevrette, line 9: Nit: `MARIADB_IMAGE_NAME ` instead of `MDB_IMAGE_NAME` for consistency and clarity

Makefile
arilivigni: I fixed it in the latest commit

Makefile
```diff
@@ -15,9 +18,21 @@ build:
 	@docker build -t $(IMAGE_NAME):latest -f dockerfiles/Dockerfile .
 	@docker tag $(IMAGE_NAME):latest $(IMAGE_NAME):$(IMAGE_TAG)
 
+build-dbs:
```
jfchevrette, line 21: Nit: maybe call this `build-utility-images` for clarity? Similarly for `push-utility-images` vs `push-dbs`

Makefile
arilivigni: Fixed

Makefile
arilivigni: Roger that

reconcile/sql_query.py
arilivigni: If you mean postgres versions we will build the others as well

apahim: We should build at least all versions on app interface. Or, maybe we can use latest for any version? Never tested it.

arilivigni: versions are all built now with the latest changes

arilivigni: Is the solution now ok?

dockerfiles/Dockerfile.mariadb
```diff
@@ -0,0 +1,4 @@
+FROM centos:8
```
maorfr, line 1: `FROM centos:8` sort of defeats the purpose of this entire PR?

maorfr:
I'm not sure about this approach.

Regarding implementation, as it means that we rebuild and push all these images even though nothing changed (likely).
Another way to go would be to create a repo per image, like https://github.com/app-sre/fluentd for example (which is used in this project).

Conceptually, I don't think we want to maintain our own set of images, when the upstream repo (with mirroring, which should be fine IMO) has what we need. It may not be UBI based, but this also doesn't seem to be the way to go these days? :confused: 

I'd prefer that we go with the original plan described here: https://issues.redhat.com/browse/APPSRE-2830?focusedCommentId=15586776&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15586776

arilivigni:
We can make it so it doesn't build everytime and is separate or create a completely separate repo as you suggested.   Centos based is fine IMO or we can pursue a UBI solution as well.   Eventually, we will stop mirroring from what I've been told and we will have to do this regardless or something similar, but not what you want so I'll close it.

PR #1248, opened 2020-12-16
[sql-query] use image pull secret

maorfr:
part of APPSRE-2830

this PR will add an imagePullSecret to the job's pod spec to be able to pull private images.

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/13182

replaces #1247

### DIFF
--- a/reconcile/sql_query.py
+++ b/reconcile/sql_query.py
@@ -8,6 +8,7 @@ import ruamel.yaml as yaml
 from textwrap import indent
 
 from reconcile import openshift_base
+from reconcile import openshift_resources_base as orb
 from reconcile import queries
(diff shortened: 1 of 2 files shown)


reconcile/sql_query.py
```diff
@@ -361,6 +389,16 @@ def run(dry_run, enable_deletion=False):
                         settings=queries.get_app_interface_settings(),
                         internal=None)
 
+        if use_pull_secret:
+            print(secret_resource.body)
```
apahim, line 393: leftover?

maorfr: absolutely

apahim (approved):
lgtm

PR #1254, opened 2020-12-17
[slack-usergroups] allow specifying a branch

maorfr:
this PR enables to pass a `url:ref` in `ownersFromRepos`.

part of https://issues.redhat.com/browse/APPSRE-2838

app-interface MR: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/13264

### DIFF
--- a/reconcile/slack_usergroups.py
+++ b/reconcile/slack_usergroups.py
@@ -183,7 +183,14 @@ def get_slack_usernames_from_owners(owners_from_repo, users, usergroup):
 
     all_slack_usernames = []
 
-    for url in owners_from_repo:
+    for url_ref in owners_from_repo:
+        # allow passing repo_url:ref to select different branch
+        if ':' in url_ref:
+            url, ref = url_ref.split(':')
+        else:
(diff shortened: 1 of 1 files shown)


reconcile/slack_usergroups.py
```diff
@@ -183,7 +183,14 @@ def get_slack_usernames_from_owners(owners_from_repo, users, usergroup):
 
     all_slack_usernames = []
 
-    for url in owners_from_repo:
+    for url_ref in owners_from_repo:
+        # allow passing repo_url:ref to select different branch
+        if ':' in url_ref:
```
apahim: Never saw that format... url:branch. Can we have a separate field to set the branch?

maorfr: not simply, as the `ownersFromRepos` is a simple string.

apahim: Right. I don't love it, but we can live with that for now.

apahim (approved):
Lgtm

PR #1260, opened 2020-12-21
Do not validate schema when querying for resources

rporres:

### DIFF
--- a/utils/gql.py
+++ b/utils/gql.py
@@ -136,7 +136,10 @@ class GqlApi(object):
         """
 
         try:
-            resources = self.query(query, {'path': path})['resources']
+            # Do not validate schema in resources since schema support in the
+            # resources is not complete.
+            resources = self.query(query, {'path': path},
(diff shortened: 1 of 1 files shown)


jmelis (approved):
LGTM

I was going to propose this:

```patch
diff --git a/utils/gql.py b/utils/gql.py
index 8ed83c7..23447ba 100644
--- a/utils/gql.py
+++ b/utils/gql.py
@@ -110,7 +110,9 @@ class GqlApi(object):
 
         if self.validate_schemas and not skip_validation:
             forbidden_schemas = [schema for schema in query_schemas
-                                 if schema not in self._valid_schemas]
+                                 if (schema is not None
+                                     and schema not in self._valid_schemas)]
+
             if forbidden_schemas:
                 raise GqlApiErrorForbiddenSchema(forbidden_schemas)
```

but on second thought this is a bit of dark magic, your option is much cleaner and more future-proof.

PR #1261, opened 2020-12-21
APPSRE-2473 download terraform providers in docker build for qontract…

skryzhny:
…-reconcile


maorfr (approved):
LGTM

this will reduce the run time of the integrations significantly.
we can use this approach for the provider versions that are commonly used, and in addition (if we find this approach hard to maintain) we can enhance with an alternative download url.

thank you for working on this!

PR #1265, opened 2020-12-22
Add prometheus-rules-validator integration

rporres:
Part of https://issues.redhat.com/browse/APPSRE-1146

It is debatable that this is actually an integration, as it will only
run on pr-check. But the code made sense here, as all the utils are here
and I would have needed to copy too much code elsewhere.

The complication here has been to be able to query qontract-server for
PrometheusRules. This has been done to just resolve the templates we
needed, reducing the run time (in my laptop) from 1m20 to 10s. Since
this is an integration that will run with every namespace change (see
app-interface related PR) and we're committed to reduce the PR check
time, this is important.

Related PRs:

* App-interface schema
  * https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/13366

* Add schema to resource in app-interface bundle
  * https://github.com/app-sre/qontract-validator/pull/25
  * https://github.com/app-sre/qontract-validator/pull/29

* Support schema in resources in qontract-server
  * https://github.com/app-sre/qontract-server/pull/100

* Avoid checking schema in resource graphql queries:
  * https://github.com/app-sre/qontract-reconcile/pull/1260

### DIFF
--- a/reconcile/prometheus_rules_validator.py
+++ b/reconcile/prometheus_rules_validator.py
@@ -16,7 +16,8 @@ def check_rule(rule):
     try:
         promtool.check_rule(rule['spec'])
     except Exception as e:
-        return {'path': rule['path'], 'message': str(e).replace('\n', '')}
+        rule['message'] = str(e).replace('\n', '')
+        return rule
 
 
 def run(dry_run, thread_pool_size=10, cluster_name=None):
@@ -28,10 +29,11 @@ def run(dry_run, thread_pool_size=10, cluster_name=None):
 
     rules = []
     for n in gqlapi.query(orb.NAMESPACES_QUERY)['namespaces']:
-        if n['name'] != 'openshift-customer-monitoring':
+        if cluster_name and n['cluster']['name'] != cluster_name:
             continue
 
-        if cluster_name and n['cluster']['name'] != cluster_name:
+        if not n['managedResourceTypes'] or \
+           'PrometheusRule' not in n['managedResourceTypes']:
             continue
 
         openshift_resources = n.get('openshiftResources')
@@ -42,13 +44,17 @@ def run(dry_run, thread_pool_size=10, cluster_name=None):
 
             openshift_resource = orb.fetch_openshift_resource(r, n)
             rules.append({'path': r['path'],
-                          'spec': openshift_resource.body['spec']})
+                          'spec': openshift_resource.body['spec'],
+                          'namespace': n['name'],
+                          'cluster': n['cluster']['name']})
 
     failed = [f for f in threaded.run(check_rule, rules, thread_pool_size)
               if f]
 
     if failed:
         for f in failed:
-            logging.warning(f"Error in rule {f['path']}: {f['message']}")
+            logging.warning(f"Error in rule {f['path']} from namespace "
+                            f"{f['namespace']} in cluster {f['cluster']}: "
+                            f"{f['message']}")
 
         sys.exit(1)
--- a/reconcile/prometheus_rules_validator.py
+++ b/reconcile/prometheus_rules_validator.py
@@ -34,6 +34,10 @@ def run(dry_run, thread_pool_size=10, cluster_name=None):
             continue
 
         openshift_resources = n.get('openshiftResources')
+        if not openshift_resources:
+            logging.warning("No openshiftResources defined for namespace"
+                            f"{n['name']} in cluster {n['cluster']['name']}")
+            continue
 
         for r in openshift_resources:
             if r['path'] not in rules_paths:
--- a/reconcile/prometheus_rules_validator.py
+++ b/reconcile/prometheus_rules_validator.py
@@ -13,11 +13,8 @@ QONTRACT_INTEGRATION_VERSION = semver.format_version(0, 1, 0)
 
 
 def check_rule(rule):
-    try:
-        promtool.check_rule(rule['spec'])
-    except Exception as e:
-        rule['message'] = str(e).replace('\n', '')
-        return rule
+    rule['check_result'] = promtool.check_rule(rule['spec'])
+    return rule
 
 
 def run(dry_run, thread_pool_size=10, cluster_name=None):
@@ -48,13 +45,13 @@ def run(dry_run, thread_pool_size=10, cluster_name=None):
                           'namespace': n['name'],
                           'cluster': n['cluster']['name']})
 
-    failed = [f for f in threaded.run(check_rule, rules, thread_pool_size)
-              if f]
+    failed = [r for r in threaded.run(check_rule, rules, thread_pool_size)
+              if not r['check_result']]
 
     if failed:
         for f in failed:
             logging.warning(f"Error in rule {f['path']} from namespace "
                             f"{f['namespace']} in cluster {f['cluster']}: "
-                            f"{f['message']}")
+                            f"{f['check_result']}")
 
         sys.exit(1)
--- a/reconcile/prometheus_rules_validator.py
+++ b/reconcile/prometheus_rules_validator.py
@@ -1,0 +1,54 @@
+import sys
+import semver
+import logging
+
+import utils.gql as gql
+import utils.threaded as threaded
+import utils.promtool as promtool
+import reconcile.openshift_resources_base as orb
+import reconcile.queries as queries
+
+QONTRACT_INTEGRATION = 'prometheus_rules_validator'
+QONTRACT_INTEGRATION_VERSION = semver.format_version(0, 1, 0)
+
+
+def check_rule(rule):
+    try:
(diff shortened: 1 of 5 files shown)


rporres:
output sample
```
$ qontract-reconcile --dry-run --log-level INFO --config config.development.toml prometheus-rules-validator
[2020-12-22 13:00:45] [INFO] [gql.py:init_from_config:202] - using gql endpoint http://localhost:4000/graphqlsha/b8984ff290137e968fc36e1363d6bad9b38691f5c1d8e3d9863c55af7f4285e9
[2020-12-22 13:00:52] [WARNING] [prometheus_rules_validator.py:run:52] - Error in rule /insights-prod/prometheus/prom.prometheusrules.yaml: Error running promtool:   FAILED:/var/folders/dx/y5_klhc1187gnzzmyb4pswb40000gn/T/tmp5h4q_bo0: 56:11: group "network-check", rule 0, "HttpFailuresLast1Min": could not parse expression: 1:1: parse error: unknown function with name "POLLAS"
[2020-12-22 13:00:52] [WARNING] [prometheus_rules_validator.py:run:52] - Error in rule /insights-prod/prometheus/prom.prometheusapp-slo.yaml: Error running promtool:   FAILED:/var/folders/dx/y5_klhc1187gnzzmyb4pswb40000gn/T/tmpwj2zr4oq: 592:11: group "xavier-sli", rule 4, "service:latency:adjusted:rate5m": could not parse expression: 1:40: parse error: unknown function with name "RATA"
$ echo $?
1

File: reconcile/prometheus_rules_validator.py [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+def run(dry_run, thread_pool_size=10, cluster_name=None):
+    orb.QONTRACT_INTEGRATION = QONTRACT_INTEGRATION
+    orb.QONTRACT_INTEGRATION_VERSION = QONTRACT_INTEGRATION_VERSION
+
+    gqlapi = gql.get_api()
+    rules_paths = queries.get_prometheus_rules_paths()
+
+    rules = []
+    for n in gqlapi.query(orb.NAMESPACES_QUERY)['namespaces']:
+        if n['name'] != 'openshift-customer-monitoring':
```
jmelis, line 31: why this artificial check? we can filter by

```
managedResourceTypes:    
- PrometheusRule    
- ...
```

utils/promtool.py
```diff
@@ -0,0 +1,27 @@
...
+
+class PromtoolError(Exception):
+    pass
+
+
+def check_rule(yaml_spec):
+    try:
+        with tempfile.NamedTemporaryFile() as fp:
+            fp.write(yaml.dump(yaml_spec).encode())
+            fp.seek(0)
```
jmelis, line 14: why do you need to do this?

reconcile/prometheus_rules_validator.py
rporres: Because we don't have rules anywhere else... but I like your condition more. I will need to adjust a bit the error messages too.

reconcile/prometheus_rules_validator.py
```diff
@@ -0,0 +1,54 @@
...
+            continue
+
+        if cluster_name and n['cluster']['name'] != cluster_name:
+            continue
+
+        openshift_resources = n.get('openshiftResources')
+
+        for r in openshift_resources:
```
jmelis, line 39: I think `openshift_resources` can be `None`: https://gitlab.cee.redhat.com/service/app-interface/-/blob/67c684b180aaab5a5ef8b0dd038a9556a8f35b98/schemas/openshift/namespace-1.yml#L189

In that case, this may fail.

utils/promtool.py
rporres: Most likely I was intending to make sure temp file is flushed to disk before, but most likely it is a better idea to add `delete=False` to the temp file creation, keep the file name in a variable and run the command out the context manager

apahim: This is te right thing to do. Outside the context manager, the file is gone.
Alternativelly, you can use `tempfile.mkdtemp()` to hae a new uniqe directory on `/tmp` and then create a file there with a stable name, deferring the cleanup for OS reboot. I like the current approach better :)

reconcile/prometheus_rules_validator.py
```diff
@@ -0,0 +1,54 @@
...
+
+QONTRACT_INTEGRATION = 'prometheus_rules_validator'
+QONTRACT_INTEGRATION_VERSION = semver.format_version(0, 1, 0)
+
+
+def check_rule(rule):
+    try:
+        promtool.check_rule(rule['spec'])
+    except Exception as e:
+        return {'path': rule['path'], 'message': str(e).replace('\n', '')}
```
apahim, line 19: why to wrap `utils.promtool.check_rule` here? can't it give you the right thing to begin with?

utils/promtool.py
rporres: Is it guaranteed that promtool is going to gete properly written? That my main doubt, to make sure nothing gets buffered before passing the file to the command

utils/promtool.py
```diff
@@ -0,0 +1,27 @@
...
+        raise PromtoolError(f'Error building promtool file: {e}')
+
+    if status.returncode != 0:
+        message = 'Error running promtool'
+        if status.stderr:
+            message += ": " + status.stderr.decode()
+
+        raise PromtoolError(message)
+
+    return status.stdout
```
apahim, line 27: This function returns either an string or an exception. Later, you get the exception message and build a dictionary that is returned from the threaded execution... it's a lot of transformations as the data you need goes by.
Can we simplify the transformations? This function could return a dict:
```
{
    valid: True | false
    message: status.stdout
}
```
Or any custom object that carries status + message, being the same object for both good and bad statuses. Then dict (or the custom object) would be directly processed by the comprehension getting the threaded results. Wdyt?

reconcile/prometheus_rules_validator.py
rporres: What do you suggest, that instead of an exception on error it returns something I can inspect, e.g. an object?

utils/promtool.py
rporres: It will simplify things. I will go that route.

utils/promtool.py
apahim: `fp.flush()` to make sure

reconcile/prometheus_rules_validator.py
apahim: You know me... I'd return a custom object from `utils.promtool.check_rule()` tha implements `__bool__` and carries a message :)
But a dict or a custom object with two attributes (status and message) should do it.

apahim (approved):
looks good. dropped some ideas, nothing critical.

reconcile/prometheus_rules_validator.py
rporres: fixed in 210abfb

utils/promtool.py
rporres: fixed in c4d9dab

reconcile/prometheus_rules_validator.py
rporres: fixed in a4b9b09

reconcile/prometheus_rules_validator.py
rporres: fixed in eb8e705. I loved the fancy approach, I hope I implemented it properly

utils/promtool.py
rporres: fixed in eb8e705

rporres:
New error messages
```
[2020-12-22 18:37:48] [WARNING] [prometheus_rules_validator.py:run:57] - Error in rule /insights-prod/prometheus/prom.prometheusrules.yaml from namespace openshift-customer-monitoring in cluster crcp01ue1: Error running promtool:   FAILED:/var/folders/dx/y5_klhc1187gnzzmyb4pswb40000gn/T/tmp0v13s7qr: 56:11: group "network-check", rule 0, "HttpFailuresLast1Min": could not parse expression: 1:1: parse error: unknown function with name "POLLAS"
[2020-12-22 18:37:48] [WARNING] [prometheus_rules_validator.py:run:57] - Error in rule /insights-prod/prometheus/prom.prometheusapp-slo.yaml from namespace openshift-customer-monitoring in cluster crcp01ue1: Error running promtool:   FAILED:/var/folders/dx/y5_klhc1187gnzzmyb4pswb40000gn/T/tmpcvk56yt4: 592:11: group "xavier-sli", rule 4, "service:latency:adjusted:rate5m": could not parse expression: 1:40: parse error: unknown function with name "RATA"

reconcile/prometheus_rules_validator.py
apahim: looks awesome :)

PR #1266, opened 2020-12-24
Adding information about parent owners

apahim:
Instead of exposing only the closest owners in the comment message,
let's add also the parent owners.

Before this patch:

```
[OWNERS] You will need a "/lgtm" from one person from each of these
groups:

* amoran, crarobin, jcueto, mciccone, mfreer, pcremin, raananda,
sclarkso
```

After this patch:

```
[OWNERS] You will need a "/lgtm" from at least one person from the
following group:

* amoran, crarobin, jcueto, mciccone, mfreer, pcremin, raananda,
sclarkso

Override approvers, from parent directories, are:

* asegundo, cbrookes, dffrench, jeder, jmelisba, kasingh, pamccart,
pbrookes, sfrancog, weil

Other relevant reviewers, from parent directories, are:

* mafriedm
```

### DIFF
--- a/reconcile/gitlab_owners.py
+++ b/reconcile/gitlab_owners.py
@@ -116,8 +116,10 @@ class MRApproval:
             report[change_path] = {}
             if not change_approved:
                 approval_status['approved'] = False
-                approvers = change_owners['closest_approvers']
+                approvers = change_owners['approvers']
                 report[change_path]['approvers'] = approvers
+                closest_approvers = change_owners['closest_approvers']
+                report[change_path]['closest_approvers'] = closest_approvers
 
             change_reviewed = False
             for reviewer in change_owners['reviewers']:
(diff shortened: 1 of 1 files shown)


reconcile/gitlab_owners.py
```diff
@@ -187,33 +191,82 @@ def format_report(report):
...
-                                f'groups:\n\n')
-        for group in sorted(approvers):
+        if closest_approvers:
+
+            if len(closest_approvers) == 1:
+                markdown_report += (f'{COMMENT_PREFIX} You will need a '
+                                    f'"/lgtm" from at least one person from '
+                                    f'the following group:\n\n')
+            else:
+                markdown_report += (f'{COMMENT_PREFIX} You will need a '
```
rporres, line 221: when this can happen? Shouldn't the closest approvers group be unique?

rporres (commented):
I like this message more. It reflects clearer the inheritance nature of owners files... 

A small doubt

reconcile/gitlab_owners.py
apahim: It's unique per change. This will happen when you change two files on different directories. Each file can then have its own set of closest approvers. Example:
https://gitlab.cee.redhat.com/service/managed-tenants/-/merge_requests/471

PR #1270, opened 2020-12-28
[terrascript_client] push and pop

maorfr:

this PR changes the way values are collected.

instead of pushing any value, we now avoid adding `None` to the values. In turn, we need to `.pop` with a `None` default value where needed.

this is required to upgrade to terraform 0.13, which fails if there are null values with unknown keys. for example, an rds instance can not be created if it contains an `es_identifier` key.

this PR alone should cause no changes to the integration.

### DIFF
--- a/utils/terrascript_client.py
+++ b/utils/terrascript_client.py
@@ -2577,7 +2577,8 @@ class TerrascriptClient(object):
         es_values['access_policies'] = json.dumps(
             access_policies, sort_keys=True)
 
-        region = values['region'] or self.default_regions.get(account)
+        region = values.get('region') or \
+            self.default_regions.get(account)
         if self._multiregion_account_(account):
             es_values['provider'] = 'aws.' + region
 
--- a/utils/terrascript_client.py
+++ b/utils/terrascript_client.py
@@ -563,7 +563,7 @@ class TerrascriptClient(object):
             not enhanced_monitoring and
             values.get('monitoring_interval', None)
         ):
-            values.pop('monitoring_interval', None)
+            values.pop('monitoring_interval')
 
         if enhanced_monitoring:
             # Set monitoring interval to 60s if it is not set.
@@ -2280,6 +2280,8 @@ class TerrascriptClient(object):
         self.override_values(values, overrides)
         values['identifier'] = identifier
         values['tags'] = self.get_resource_tags(namespace_info)
+        # checking explicitly for not None
+        # to allow passing empty strings, False, etc
         if variables is not None:
             values['variables'] = variables
         if policies is not None:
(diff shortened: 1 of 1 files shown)


utils/terrascript_client.py
```diff
@@ -555,15 +555,15 @@ def populate_tf_resource_rds(self, resource, namespace_info,
...
+        enhanced_monitoring = values.pop('enhanced_monitoring', None)
 
         # monitoring interval should only be set if enhanced monitoring
         # is true
         if (
             not enhanced_monitoring and
             values.get('monitoring_interval', None)
         ):
-            values.pop('monitoring_interval')
+            values.pop('monitoring_interval', None)
```
rporres, line 566: Do you need this? You've already checked `values.get('monitoring_interval')` is true

utils/terrascript_client.py
```diff
@@ -2271,25 +2280,44 @@ def init_values(self, resource, namespace_info):
...
-        values['bucket_policy'] = bucket_policy
-        values['storage_class'] = sc
-        values['enhanced_monitoring'] = enhanced_monitoring
-        values['replica_source'] = replica_source
-        values['es_identifier'] = es_identifier
-        values['filter_pattern'] = filter_pattern
-        values['secret'] = secret
-        values['output_resource_db_name'] = output_resource_db_name
-        values['reset_password'] = reset_password
+        if variables is not None:
```
rporres, line 2283: why not
```
if variables:
```
In fact I don't think you need to initialize to `None` in 2256, e.g
```
>>> h = {'a':1}
>>> c = h.get('c')
>>> type(c)
<class 'NoneType'>
>>> if not c: print("Hello Maor")
...
Hello Maor
```

rporres (commented):
I'm having a bit of a hard time understanding some of the changes related to how they are described. It looks like the main part here is avoiding initialization of `None` values in `init_values`.

The other set of changes are related to avoiding `KeyError` either with `pop` or by accessing elements of dicts via `[]`. But I fail to understand well how they are related 😅  (most likely due to my general lack of knowledge of the code base)

utils/terrascript_client.py
maorfr: doesn't hurt?

utils/terrascript_client.py
maorfr: in some cases we want to pass an empty string, which fails the `if var` check

maorfr: if we don't set a certain key is `values` we can't get it with `[]`. that is the relation. there is nothing smart here :)

utils/terrascript_client.py
rporres: It doesn't, just redundant :)

utils/terrascript_client.py
rporres: ok, understood. I would add a note stating why you want to check explicitly for `None` :)

rporres (approved):

PR #1271, opened 2020-12-28
[terraform] add 0.13 compliant outputs

maorfr:

this PR adds terrraform 0.13 compliant outputs (no `.`, `[`, `]`) which will eventually replace the current outputs.
we need the outputs to exist in parallel to be able to start using the new outputs and only remove the old ones before migrating to terraform 0.13.

### DIFF
--- a/utils/terraform_client.py
+++ b/utils/terraform_client.py
@@ -265,11 +265,10 @@ class TerraformClient(object):
             # of the integration.
             if '[' not in k or ']' not in k:
                 continue
-            # this currently has no effect
-            # it will replace the above condition when we
+            # this will replace the above condition when we
             # upgrade to terraform 0.13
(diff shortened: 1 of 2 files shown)


rporres:
I understand you cannot change output formats for the current integration, correct? Or you see it too risky?

maorfr:
this PR is the first of a few. this will add the new outputs, a following PR will switch to use them, and a 3rd one to remove the old ones.

until all these are done, we can't really upgrade to 0.13.

rporres:
My question is really if you can change current outputs

maorfr:
we can, but only after the outputs exist there (as in, we must merge this and let it run first)

rporres:
understood. Then this approach sounds reasonable :)

PR #1274, opened 2020-12-30
[terraform] use 0.13 outputs and remove 0.11 outputs

maorfr:
following up on #1271, where we added 0.13 compliant outputs, we now need to consume them and remove the 0.11 compliant outputs.

this PR should be promoted to production only after #1271 is already promoted itself (and had a chance to create the 0.13 compliant outputs before we remove the 0.11 ones):
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/13448

### DIFF
--- a/utils/terraform_client.py
+++ b/utils/terraform_client.py
@@ -250,9 +250,9 @@ class TerraformClient(object):
         if output is None:
             return data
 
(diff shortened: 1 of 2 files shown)


maorfr:
testing release candidate in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/13453

PR #1275, opened 2020-12-30
[terraform] upgrade to 0.13

maorfr:

this PR holds the changes required to perform the migration from 0.11 to 0.13:
- upgrade terraform in the Dockerfile
- upgrade terrascript
- update syntax where needed
- use `terraform plan -out` and `terraform show -json` instead of parsing logs
- check for provider alias uniqueness

to merge this, 3 steps are required:
1. https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/13448 (merged), which promotes to production:
- #1270 
- #1271 
2. https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/13463, which promotes to production:
- #1274 
- #1276 
- #1278 
- #1279 
- #1280 
- #1283
3. promote this PR.


utils/lean_terraform_client.py
```diff
@@ -13,3 +15,12 @@ def state_rm_access_key(working_dirs, account, user):
                  cwd=wd, stdout=PIPE, stderr=PIPE)
     proc.communicate()
     return proc.returncode == 0
+
+
+def show_json(working_dir, out_file):
+    proc = Popen(['terraform', 'show', '-json', out_file],
+                 cwd=working_dir, stdout=PIPE, stderr=PIPE)
+    out, _ = proc.communicate()
+    if proc.returncode != 0:
```
apahim, line 24: `if proc.returncode:`

utils/lean_terraform_client.py
```diff
@@ -13,3 +15,12 @@ def state_rm_access_key(working_dirs, account, user):
...
+        raise Exception('terraform show failed')
```
apahim, line 25: not sure how often we will see this, but `subprocess.run()` with `check=True` gives us a very specific exception with all the debug details.

apahim (approved):
Code looks good. I haven't executed it, but I'm sure you're on top of that.
Thank you for tackling this... a tricky work that had to be done.

utils/lean_terraform_client.py
maorfr: i don't think this will ever happen.

utils/lean_terraform_client.py
maorfr: fixed!

utils/lean_terraform_client.py
rporres: Famous last words :-P

utils/terraform_client.py
```diff
@@ -206,7 +198,8 @@ def apply(self):
     def terraform_apply(self, apply_spec):
         name = apply_spec['name']
         tf = apply_spec['tf']
-        return_code, stdout, stderr = tf.apply(auto_approve=True,
+        return_code, stdout, stderr = tf.apply(dir_or_plan=name,
```
rporres, line 201: why do you have a plan and then you do a `skip_plan` afterwards?

maorfr: skip_plan is really just to enable auto-approve:
https://github.com/beelit94/python-terraform/blob/a33aebce1ca34357bb8a999f9853a6f614fba924/python_terraform/__init__.py#L112

utils/lean_terraform_client.py
maorfr: if you say "famous last words" does it cancel the jinx? :)

utils/terraform_client.py
```diff
@@ -134,63 +135,54 @@ def terraform_plan(self, plan_spec, enable_deletion):
...
-                        logging.info(['replace', name,
-                                      resource_type, resource_name])
-                    else:
-                        logging.error(['replace', name,
-                                       resource_type, resource_name])
-                        logging.error('\'replace\' action is not enabled. ' +
-                                      'Please run the integration manually ' +
-                                      'with the \'--enable-deletion\' flag.')
+
+        output = self.terraform_show(name, tf.working_dir)
```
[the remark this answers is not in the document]
maorfr: very good idea! done in https://github.com/app-sre/qontract-reconcile/pull/1275/commits/0c5b6f978f96178680facf44ec375b72c0de6c99

utils/terraform_client.py
rporres: There's no need for auto-approve if you apply a plan, it doesn't ask. But if you skip the plan you've already calculated, I'm afraid it will calculate a new one...

maorfr: that makes sense. i've removed `auto_approve` and `skip_plan`. we'll see if it behaves as we expect.

PR #1277, opened 2020-12-30
[slack-usergroups] ignore username with + in it

maorfr:
fixes the following warning:
```
PagerDuty username not found in app-interface: ['<username>+nobody'] (hint: user files should contain pagerduty_username if it is different then org_username)
```

https://ci.int.devshift.net/job/service-app-interface-gl-pr-check/29175/app-interface_20JSON_20validation/


apahim (approved):
lgtm

PR #1279, opened 2020-12-31
add option to call realize_data with safety nets

maorfr:
with this PR, we can call `ob.realize_data` without allowing resources to be deleted and without recycling pods.

this will be used a safety net for the upcoming terraform upgrade.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -195,7 +195,7 @@ def wait_for_namespace_exists(oc, namespace):
 
 
 def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
-          wait_for_namespace):
+          wait_for_namespace, recycle_pods=True):
     logging.info(['apply', cluster, namespace, resource_type, resource.name])
 
     oc = oc_map.get(cluster)
@@ -225,7 +225,8 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
                 UnsupportedMediaTypeError):
             oc.replace(namespace, annotated.toJSON())
 
-    oc.recycle_pods(dry_run, namespace, resource_type, resource)
+    if recycle_pods:
+        oc.recycle_pods(dry_run, namespace, resource_type, resource)
 
 
 def delete(dry_run, oc_map, cluster, namespace, resource_type, name,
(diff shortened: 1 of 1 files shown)


reconcile/openshift_base.py
```diff
@@ -257,7 +258,9 @@ def realize_data(dry_run, oc_map, ri,
                  take_over=False,
                  caller=None,
                  wait_for_namespace=False,
-                 no_dry_run_skip_compare=False):
+                 no_dry_run_skip_compare=False,
+                 override_enable_deletion=None,
```
rporres, line 262: why this is not `True/False`?

maorfr: whatever the value the user passes is the override value. if a user doesn't pass it, they don't want to override. see the check `if not None` below which matches this idea.

reconcile/openshift_base.py
```diff
@@ -270,9 +273,13 @@ def realize_data(dry_run, oc_map, ri,
                    to deploy to the same namespace
     :param wait_for_namespace: wait for namespace to exist before applying
     :param no_dry_run_skip_compare: when running without dry-run, skip compare
+    :param override_enable_deletion: override calculated enable_deletion value
+    :param recycle_pods: should pods be recycled if a dependency changed
     """
     actions = []
     enable_deletion = False if ri.has_error_registered() else True
+    if override_enable_deletion is not None:
+        enable_deletion = override_enable_deletion
```
rporres, line 282: We should only override if we don't have any errors, correct?. If not, `enable_deletion` should always be `False`

reconcile/openshift_base.py
```diff
@@ -225,7 +225,8 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
                 UnsupportedMediaTypeError):
             oc.replace(namespace, annotated.toJSON())
 
-    oc.recycle_pods(dry_run, namespace, resource_type, resource)
+    if recycle_pods:
```
rporres, line 228: Most likely we want a log message here to tell us that we're not recycling certain pods due to the option. That way we would have a better view of what's happening.

A debug message in `oc.py` telling which pods/deployments/statefulsets/... are going to be affected may also help give visibility here...

maorfr: for terraform-resources it will be all secrets that are getting applied. so if we see an apply log, we know it will be recycled.

reconcile/openshift_base.py
rporres: It is dangerous to override to `True` in case of errors.

PR #1280, opened 2020-12-31
add safety nets to terraform integrations

maorfr:

this PR will prevent terraform integrations from deleting resources. this is a safety net for the duration of the terraform 0.13 upgrade and this PR should be reverted once we gained confidence.

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -357,7 +357,13 @@ def run(dry_run, print_only=False,
 
     tf.populate_desired_state(ri, oc_map)
 
-    ob.realize_data(dry_run, oc_map, ri)
+    # temporarily not allowing resources to be deleted
+    # or for pods to be recycled
+    # this should be removed after we gained confidence
+    # following the terraform 0.13 upgrade
+    ob.realize_data(dry_run, oc_map, ri,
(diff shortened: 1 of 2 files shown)


reconcile/terraform_resources.py
```diff
@@ -357,7 +357,13 @@ def run(dry_run, print_only=False,
...
     tf.populate_desired_state(ri, oc_map)
 
-    ob.realize_data(dry_run, oc_map, ri)
+    # temporarily not allowing resources to be deleted
+    # or for pods to be recycled
+    # this should be removed after we gained confidence
+    # following the terraform 0.13 upgrade
+    ob.realize_data(dry_run, oc_map, ri,
+                    override_enable_deletion=enable_deletion,
+                    recycle_pods=enable_deletion)
```
rporres, line 366: shouldn't we have a separate option for the pods recycling?

maorfr: i thought about it as well, but this is just for a short period. once the upgrade is complete we will remove this safety net, so didn't bother adding the additional option to the integration itself. generally, terraform should always recycle the pods if a secret changes.

rporres: As long as you delete it afterwards, it's ok

PR #1282, opened 2021-01-03
retry gql.query

maorfr:
add resiliency against qontract-server


utils/gql.py
```diff
@@ -88,6 +89,7 @@ def __init__(self, url, token=None, int_name=None, validate_schemas=False):
             if not self._valid_schemas:
                 raise GqlApiIntegrationNotFound(int_name)
 
+    @retry(exceptions=(GqlApiError))
```
apahim, line 92: the parenthesis around `GqlApiError` is redundant

apahim (approved):

PR #1287, opened 2021-01-05
remove cluster token in sentry

BumbleFeng:

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import sys
+import re
 
 import click
 import sentry_sdk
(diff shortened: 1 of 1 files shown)


reconcile/cli.py
```diff
@@ -102,9 +103,21 @@
 from utils.unleash import get_feature_toggle_state
 
 
+def before_breadcrumb(breadcrumb):
```
jfchevrette, line 106: Can you add a docstring to explain what this does? I expect we will eventually want to add more things in there to mask other types of sensitive informaton

jfchevrette:
LGTM from a code/sentry sdk standpoint but I can't think of a way we can test this before deploying it.

OTOH, I don't think the code can do harm so +1 to :ship:

maorfr:
LGTM as well
@jmelis PTAL

maorfr (approved):
this is probably fine to promote and see if it works

PR #1295, opened 2021-01-07
Improve logs returned when validating desired state

jfchevrette:
This improves the log messages when we're validating desired state to give more info to the end users:
* Deployment/DeploymentConfig: tell the user how many replicas are ready vs total desired
* Subscription: tell the user what the actual state is
* Job: tell the user what the conditions are (currently dumped as-is to simplify code)
* ClowApp: tell the user how many deployments are ready vs. total managed

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -433,17 +433,23 @@ def validate_data(oc_map, actions):
                 ready_replicas = status.get('readyReplicas')
                 if not desired_replicas == replicas == \
                         ready_replicas == updated_replicas:
-                    logging.info('new replicas not ready, status is invalid')
+                    logging.info(
+                        f'{kind} {name} has replicas that are not ready '
+                        f'({ready_replicas} ready / {desired_replicas} total)')
                     raise ValidationError(name)
             elif kind == 'Subscription':
                 state = status.get('state')
                 if state != 'AtLatestKnown':
-                    logging.info('Subscription status.state is invalid')
+                    logging.info(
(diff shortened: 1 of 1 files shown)


arilivigni:
lgtm

reconcile/openshift_base.py
```diff
@@ -433,17 +433,23 @@ def validate_data(oc_map, actions):
...
                 state = status.get('state')
                 if state != 'AtLatestKnown':
-                    logging.info('Subscription status.state is invalid')
+                    logging.info(
+                        f'Subscription {name} state is invalid. '
+                        f'Current state: {state}')
                     raise ValidationError(name)
             elif kind == 'Job':
                 succeeded = status.get('succeeded')
+                conditions = status.get('conditions')
```
apahim, line 449: you can move this line inside the if below

apahim (approved):
super nice!

reconcile/openshift_base.py
apahim: btw, can this return `None`? If yes, maybe the default should be and empty string? (to avoid the message `Job conditions are: None`). Not sure if it's any better.

jfchevrette: Thanks for pointing that out.

I moved retrieving the conditions inside of the if and since the value is actually either `None` or a list of dict, instead of changing to default to empty string I checked if there are conditions (not None or empty list) and dump them as yaml for better readability instead of printing them raw

apahim: sounds great, tks!

PR #1302, opened 2021-01-10
[sentry-config] cache some responses

maorfr:
this reduces the number of requests in the integration by about half.

### DIFF
--- a/reconcile/utils/sentry_client.py
+++ b/reconcile/utils/sentry_client.py
@@ -10,6 +10,7 @@ class SentryClient:
     def __init__(self, host, token):
         self.host = host
         self.auth_token = token
+        self._cache = {}
 
     @retry()
     def _do_sentry_api_call_(self, method, path, slugs, payload=None):
@@ -33,7 +34,11 @@ class SentryClient:
 
     # Organization functions
     def get_organizations(self):
+        organizations = self._cache.get('organizations')
+        if organizations:
+            return organizations
         response = self._do_sentry_api_call_("get", "organizations", [])
+        self._cache['organizations'] = response
         return response
 
     def get_organization(self, slug):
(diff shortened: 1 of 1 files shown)


apahim:
This is getting more and more frequent.. can we have our own cache decorator and reuse it everywhere?

reconcile/utils/sentry_client.py
```diff
@@ -33,7 +34,11 @@ def _do_sentry_api_call_(self, method, path, slugs, payload=None):
 
     # Organization functions
     def get_organizations(self):
+        organizations = self._cache.get('organizations')
+        if organizations:
+            return organizations
```
aditya-konarde, line 39: If I understand correct, returning at this point would mean we almost always (as long as we have `organizations` set in the dictionary) return the field directly and never make an API call to sentry. 

In this case, I'm wondering how changes (for example, new organizations getting created in sentry) are being handled? I think we'd need to also have the cache as a write-through in the case of `creates` (users, projects), and then poll for changes in case of things that we don't control (I don't see a create function for `organizations`)

aditya-konarde (changes requested):

reconcile/utils/sentry_client.py
maorfr: with this change we assume there are no changes to organizations throughout the iteration of the integration. every new iteration starts fresh.
i've also only implemented the cache for `get` functions, so no need to worry about create/delete/etc.

apahim: Cache will start fresh on every integration execution. Whatever new shows up during the integration execution, will be tackled in the next execution.

aditya-konarde: Makes sense, thank you for the clarification! :)

aditya-konarde (approved):

maorfr:
@apahim that makes a lot of sense. I've created https://issues.redhat.com/browse/APPSRE-2866.

reconcile/utils/sentry_client.py
apahim: Good point about the setters though. We typically set things as the last act agter getting, but having the setters reseting or updating the cache is important to keep things sane.

PR #1303, opened 2021-01-11
[sentry-config] add pagination

maorfr:
as we grew larger in numbers, some calls now do not return all results and include a `link` header in the response indicating of more pages.
this PR adds usage of these pages to return the complete response (if pages exist, the response is a list of items).

### DIFF
--- a/reconcile/utils/sentry_client.py
+++ b/reconcile/utils/sentry_client.py
@@ -52,7 +52,7 @@ class SentryClient:
             response = \
                 call(f"{url}?&cursor={cursor}", headers=headers, json=payload)
             response.raise_for_status()
-            # if there are pages, each response is a list to append
+            # if there are pages, each response is a list to extend
             all_results += response.json()
 
         return all_results
--- a/reconcile/utils/sentry_client.py
+++ b/reconcile/utils/sentry_client.py
@@ -43,6 +43,9 @@ class SentryClient:
                 break
             # 2nd item is the next page
             next_item = link.split(', ')[1]
+            # copied with love from
+            # https://stackoverflow.com/questions/10663093/
+            # use-python-format-string-in-reverse-for-parsing
             _, rel, results, cursor = parse.parse(item_format, next_item)
             if rel != 'next' or results != 'true':
                 break
--- a/reconcile/utils/sentry_client.py
+++ b/reconcile/utils/sentry_client.py
@@ -1,5 +1,6 @@
 import requests
 import json
+import parse
 
 from sretoolbox.utils import retry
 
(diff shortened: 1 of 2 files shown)


reconcile/utils/sentry_client.py
```diff
@@ -27,10 +28,35 @@ def _do_sentry_api_call_(self, method, path, slugs, payload=None):
         response.raise_for_status()
 
         try:
-            return response.json()
+            all_results = response.json()
         except json.decoder.JSONDecodeError:
             return
 
+        # there may be more pages if the response contains a link header
```
aditya-konarde, line 35: Thank you for adding a note here!

aditya-konarde: Perhaps add an example as well for easier reading?

reconcile/utils/sentry_client.py
```diff
@@ -27,10 +28,35 @@ def _do_sentry_api_call_(self, method, path, slugs, payload=None):
...
+        while True:
+            link = response.headers.get('link')
+            if not link:
+                break
+            # 2nd item is the next page
+            next_item = link.split(', ')[1]
+            # copied with love from
+            # https://stackoverflow.com/questions/10663093/
+            # use-python-format-string-in-reverse-for-parsing
+            _, rel, results, cursor = parse.parse(item_format, next_item)
```
aditya-konarde, line 49: :heart_eyes:

reconcile/utils/sentry_client.py
```diff
@@ -27,10 +28,35 @@ def _do_sentry_api_call_(self, method, path, slugs, payload=None):
...
+            if rel != 'next' or results != 'true':
+                break
+            response = \
+                call(f"{url}?&cursor={cursor}", headers=headers, json=payload)
+            response.raise_for_status()
+            # if there are pages, each response is a list to append
```
apahim, line 55: append -> extend

apahim (approved):
noice

reconcile/utils/sentry_client.py
aditya-konarde: Even better if we add a link to the sentry API documentation here, if available

reconcile/utils/sentry_client.py
```diff
@@ -27,10 +28,35 @@ def _do_sentry_api_call_(self, method, path, slugs, payload=None):
...
+            # 2nd item is the next page
+            next_item = link.split(', ')[1]
+            # copied with love from
+            # https://stackoverflow.com/questions/10663093/
+            # use-python-format-string-in-reverse-for-parsing
+            _, rel, results, cursor = parse.parse(item_format, next_item)
+            if rel != 'next' or results != 'true':
+                break
+            response = \
+                call(f"{url}?&cursor={cursor}", headers=headers, json=payload)
```
aditya-konarde, line 53: Curious: in what situations does 'call' fail? Is it worth wrapping it in a try/catch block?

aditya-konarde (approved):
Thanks for working on this Maor! I have added a few suggestions and questions I had. Overall LGTM, please feel free to move ahead after addressing/acknowledging the suggestions :)

reconcile/utils/sentry_client.py
maorfr: thanks!

reconcile/utils/sentry_client.py
maorfr: this is the example (right here below):
https://github.com/app-sre/qontract-reconcile/pull/1303/files#diff-c4b8fdf06e29c3116b2012bc23e99e3833efbd32539bbdd8b84ea54fc17ccf6bR36-R39

maorfr: the API documentation is for a newer version then the one we are using. I couldn't find this (had to dig through the response to implement this)

reconcile/utils/sentry_client.py
maorfr: the way to test the result of call (`requests.get` for example) is with `response.raise_for_status()` (which will raise an exception we want to see if something happened).

PR #1309, opened 2021-01-11
Removing leading /

BumbleFeng:
avoid a warning `tar: Removing leading / from member names` which may make the user think the command failed


apahim (approved):
lgtm

PR #1310, opened 2021-01-12
APPSRE-2473 Fix permissions for plugin-cache

skryzhny:
For OpenShift giving write permission to group


dockerfiles/Dockerfile
```diff
@@ -63,7 +63,7 @@ RUN find ${JSONNET_VENDOR_DIR} -type d -exec chmod 0755 {} \;
 RUN mkdir -p /.terraform.d/plugin-cache
 COPY --from=terraform-providers-downloader /tmp/terraform-providers/ /.terraform.d/plugins/registry.terraform.io/hashicorp/aws/
 RUN find /.terraform.d/ -type d -exec chmod 0755 {} \;
-
+RUN chmod 0775 /.terraform.d/plugin-cache
```
maorfr, line 66: doesn't `chmod` support a recursive flag? would that help?

skryzhny: `--recursive` flag will change both directories and files.
And last `chmod` is different, it's allow group to write for specific directory, while previous doesn't.

maorfr: as you see fit!

maorfr (approved):

PR #1315, opened 2021-01-12
Added prometheus tests to prometheus validator

rporres:
Rule tests have been added in the same integration as the validations as
we needed the actual rule contents to run the tests, hence separating it
would have meant to fetch the rules twice. The integration has been
renamed to tester since test is more generic as it  includes validation.

Since rules can be templates, this means rules can have a different
shape for different namespaces. This force us to run the tests in every
namespace the rules appear. To make things simpler, I have limited the
number of rules per test to one as allowing for multiple rules
complicated significantly the run of the tests. If we see this is a
limitation, we can revisit this decision.

Related to https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/13709


reconcile/prometheus_rules_tester.py
```diff
@@ -0,0 +1,269 @@
...
+        rules[r['path']] = {}
+
+    for n in gqlapi.query(orb.NAMESPACES_QUERY)['namespaces']:
+        cluster = n['name']
+        namespace = n['cluster']['name']
+
+        if cluster_name and cluster != cluster_name:
+            continue
+
+        if not n['managedResourceTypes'] or \
```
maorfr, line 68: nit: use get?

reconcile/prometheus_rules_tester.py
```diff
@@ -0,0 +1,269 @@
...
+                            f'{namespace} in cluster {cluster}')
+            continue
+
+        for r in openshift_resources:
+            path = r['path']
+            if path not in rules:
+                continue
+
+            # Or we will get an unexepected and confusing html_url annotation
+            if 'add_path_to_prom_rules' not in r:
```
maorfr, line 84: is `add_path_to_prom_rules` ever in `r`?

reconcile/prometheus_rules_tester.py
```diff
@@ -0,0 +1,269 @@
...
+            msg = f"Error in test {f['path']}"
+            if 'cluster' in f:
+                msg += f" from namespace {f['namespace']} in cluster" \
+                       f"{f['cluster']}"
+            msg += f":  {f['check_result']}"
+
+            logging.warning(msg)
+
+    if invalid_rules or failed_tests:
+        sys.exit(1)
```
maorfr, line 269: please use reconcile.status.ExitCodes

reconcile/prometheus_rules_tester.py
```diff
@@ -0,0 +1,269 @@
...
+
+def run(dry_run, thread_pool_size=10, cluster_name=None):
+    orb.QONTRACT_INTEGRATION = QONTRACT_INTEGRATION
+    orb.QONTRACT_INTEGRATION_VERSION = QONTRACT_INTEGRATION_VERSION
+
+    rules = get_prometheus_rules(cluster_name)
+    invalid_rules = check_prometheus_rules(rules, thread_pool_size)
+    if invalid_rules:
+        for i in invalid_rules:
+            logging.warning(f"Error in rule {i['path']} from namespace "
```
maorfr, line 252: if this eventually leads to exit with error, should we log error?

reconcile/prometheus_rules_tester.py
```diff
@@ -0,0 +1,269 @@
...
+    failed_tests = check_prometheus_tests(tests, rules, thread_pool_size)
+    if failed_tests:
+        for f in failed_tests:
+            msg = f"Error in test {f['path']}"
+            if 'cluster' in f:
+                msg += f" from namespace {f['namespace']} in cluster" \
+                       f"{f['cluster']}"
+            msg += f":  {f['check_result']}"
+
+            logging.warning(msg)
```
maorfr, line 266: if this eventually leads to exit with error, should we log error?

maorfr (approved):
LGTM

i've added some minor comments, but you can merge as is as far as i'm concerned.
i love the fact that despite this being very complex, you were still able to write it with such simplicity. kudos for taking the time to go into all the corners cases.

reconcile/prometheus_rules_tester.py
rporres: No, but it is set to `True` in `fetch_provider_resource`. That's why I needed to add this and make sure it is taken into account in `orb. fetch_openshift_resource`

reconcile/prometheus_rules_tester.py
rporres: Yes, it makes sense.

reconcile/prometheus_rules_tester.py
rporres: Yes, it makes sense.

reconcile/prometheus_rules_tester.py
rporres: will do

reconcile/prometheus_rules_tester.py
rporres: yes, either that or `not in` to avoid potential `KeyError` exceptions

rporres: fixed in a937393

reconcile/prometheus_rules_tester.py
rporres: fixed in 248412f

reconcile/prometheus_rules_tester.py
rporres: fixed in 248412f

reconcile/prometheus_rules_tester.py
rporres: fixed in 1927ba1

PR #1318, opened 2021-01-14
Getting credentials from the replica_source for the replica Secret

apahim:
We can't set the username/password for a RDS replica in terraform, but
we still want them in the OpenShift Secret. For that, we find the
username/password from the replica_source in the terraform output and
copy them to the replica Secret data.

### DIFF
--- a/reconcile/utils/terraform_client.py
+++ b/reconcile/utils/terraform_client.py
@@ -4,12 +4,14 @@ import logging
 import json
 import os
 
+from ruamel import yaml
+
 from threading import Lock
 
 from python_terraform import Terraform, IsFlagged, TerraformCommandError
 from sretoolbox.utils import retry
 
-from reconcile.utils import threaded
+from reconcile.utils import threaded, gql
 from reconcile.utils.openshift_resource import OpenshiftResource as OR
 import reconcile.utils.lean_terraform_client as lean_tf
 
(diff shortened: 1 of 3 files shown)


reconcile/utils/terraform_client.py
```diff
@@ -221,13 +221,59 @@ def get_terraform_output_secrets(self):
...
+                # First, we have to find the terraform resources
+                # that have a replica_source defined in app-interface
+                tf_replica_source = tf_resource.get('replica_source')
+                if tf_replica_source is None:
+                    continue
+                # All resources are expected to have an account. This
+                # is just to keep it safe.
+                tf_account = tf_resource.get('account')
+                if tf_account is None:
+                    continue
```
rporres, line 243: nit: Shouldn't we log something here as this should not happen?

rporres:
This looks great to me... If not in a hurry, I'd wait to @maorfr to take a look as he knows much better this code :)

rporres (approved):
This looks great to me... If not in a hurry, I'd wait to @maorfr to take a look as he knows much better this code :)

maorfr (approved):
LGTM

i've also noticed that `replicate_source_db` can be defined in defaults files. how will we cover that scenario?

apahim:
@maorfr mind taking another look? I'm now processing the `replicate_source_db` from the `defaults` and also checking the `overrides`. Even though `replicate_source_db` is not currently allowed in the overrides, if we get to that, it will work out-of-the-box.

maorfr (approved):
LGTM again!

thank you!

PR #1321, opened 2021-01-15
Allow Route host changes

apahim:


jhernand:
I can't say anything useful about the actual code, but we indeed would benefit from being able to change route hosts. So looks good to me.

rporres (approved):
I see no other way to fix this. I don't particularly love that we need to parse the output to do this and we cannot rely on a proper return code, but here we are.

Thanks for doing this.

reconcile/openshift_base.py
```diff
@@ -225,6 +226,10 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
         except (MetaDataAnnotationsTooLongApplyError,
                 UnsupportedMediaTypeError):
             oc.replace(namespace, annotated.toJSON())
+        except FieldIsImmutableError:
```
maorfr, line 229: should we check if `resource_type` is `Route`? this can happen to StatefulSets as well, and deleting those may lead to unintended outages.

maorfr:
in addition to the above comment, here is a related ticket: https://issues.redhat.com/browse/APPSRE-2065

reconcile/openshift_base.py
maorfr: unintentional changes to Routes may lead to outages as well. maybe the author's intention behind immutable fields was exactly that. and perhaps that should be our strategy as well? to create a new Route in one MR and delete the old one in the 2nd? just throwing thoughts.

apahim: unintentional changes? not sure we have sageguards for that anywhere else... this change has the same effect of renaming the Route while changing the host, which can be done in the same MR to app-interface... but with this, tenants don't have to learn that they need to rename the route. I'm ok with just doccumenting that as well, but then it would be nice to fail the MR check when only the host is updated, with no renaming.

maorfr: i've no problem with this PR as is. just wanted to raise awareness around (particularly) StatefulSets.

apahim: Yes yes, no question on that. Fixing.

maorfr (approved):
LGTM

PR #1322, opened 2021-01-15
quay-mirror: exit with error on official images made public

apahim:
The official images can't be mirrored to a public repository. We already
skip them. Let's also fail the integration so that case is caught on MRs
to app-interface.


maorfr (approved):
LGTM

PR #1323, opened 2021-01-17
gate production promotion based on post-deployment test results

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2882

this PR adds the logic to gate promotions based on the success published to a "channel" by a previous job.

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/13877

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -62,6 +62,7 @@ class SaasHerder():
     def _validate_saas_files(self):
         self.valid = True
         saas_file_name_path_map = {}
+        saas_file_promotion_publish_channels = []
         for saas_file in self.saas_files:
             saas_file_name = saas_file['name']
             saas_file_path = saas_file['path']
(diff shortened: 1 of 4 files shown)


reconcile/utils/saasherder.py
```diff
@@ -794,3 +794,25 @@ def update_config(self, job_spec):
...
+        If there were promotion sections in the participating saas files
+        validate that the conditions are met. """
+        for item in self.promotions:
+            # validate that the commit sha being promoted
+            # was succesfully published to the subscribed channel(s)
+            commit_sha = item['commit_sha']
+            subscribe = item.get('subscribe')
+            if subscribe:
+                for channel in subscribe:
+                    state_key = f"{channel}/{commit_sha}"
```
apahim, line 828: I believe we need a common prefix path, otherwise the state will be a mess of keys.

maorfr: like this?
https://github.com/app-sre/qontract-reconcile/blob/58ba6944dd61c810e90e1775e5078c572691a605/reconcile/utils/state.py#L28

maorfr: and each integration "chooses" how it handles it's own "space"

apahim: I was thinking in something like `state_key = f"promotions/{channel}/{commit_sha}"`, but whater aggregates the promotion state keys in the first level would work, I guess

apahim: I'm also trying to avoid overriding keys that belong to other features

maorfr: `promotions/{channel}/{commit_sha}` makes sense. also didn't think about multiple features in the same integration requiring the use of state. all good points!

apahim:
Man, this is super nice. The commits separation made it super-fast to review. It's a huge feature for the tenants.
One use-case that I'm missing here, probably more important for us than for the tenants, is to block saas promotions based on the test failures. But that's probably an easy next step after this work is done.

apahim (approved):
lgtm

PR #1329, opened 2021-01-19
fix static dddb service reference in dddb cronjobs, fixes APPSRE-2886

mmclanerh:
bug in all dashdotdb crons that staticly points config to production services. this prevents the ability to validate in stage before promoting to production. this change removes the static var that references where the crons gets the url/creds for services, to a secret. this secret has already been updated in stage and prod.


apahim:
This is not a bug, it was intentional. The stage for the client is meant for validating the client.

The stage for the server uses its own database instance, so the validation there can be more intrusive, no need to use --dry-run there.

Because I wanted the deeper validation there but not to use the same RDS instance size, I opted for not having a client pointing to the stage server all the time and just parametrize the `DASHDOTDB_SECRET` and the `DRY_RUN` parameters from the saas file when I needed to do some validation.

Anyway, I'm ok with merging this.

mmclanerh:
@apahim ah, that makes sense. thanks for the background and the review. wonder if for cases like this some ephemeral idling/resourcing would be worth exploring. ex: auto-scale deployment to 0 and/or de-comm-db if it's been nn-minutes without requests or new saas updates.. 🤔

apahim (approved):

PR #1330, opened 2021-01-19
monthly validation report metrics for visual-app-interface

mmclanerh:
Added metrics output for validations to monthly/visual-app-interface reports. Not looking to re-invent anything here, so following what has already been done in other existing cases.


apahim (approved):
lgtm, thanks!

PR #1331, opened 2021-01-19
support slow receiver timeouts for dashdotdb in dvo

mmclanerh:
o added chunking of prom data blob into dddb for dvo into optionally configurable size (default: 20 metric/value pairs per post)
o added parameter to run dvo sync for single cluster

### DIFF
--- a/reconcile/dashdotdb_dvo.py
+++ b/reconcile/dashdotdb_dvo.py
@@ -41,7 +41,8 @@ class DashdotdbDVO:
         LOG.info('%s Processing (%s) metrics for: %s', self.logmarker,
                  len(dvresult),
                  cluster)
-        self.chunksize = len(dvresult) if not self.chunksize
+        if not self.chunksize:
+            self.chunksize = len(dvresult) 
         if len(dvresult) <= int(self.chunksize):
             metrics = dvresult
         else:
--- a/reconcile/dashdotdb_dvo.py
+++ b/reconcile/dashdotdb_dvo.py
@@ -23,29 +23,48 @@ class DashdotdbDVO:
         self.dashdotdb_url = secret_content['url']
         self.dashdotdb_user = secret_content['username']
         self.dashdotdb_pass = secret_content['password']
+        self.chunksize = secret_content.get('chunksize') or '20'
         self.logmarker = "DDDB_DVO:"
+
+    def _chunkify(self, data, size):
+        for i in range(0, len(data), int(size)):
+            yield data[i:i+int(size)]
 
     def _post(self, deploymentvalidation):
         if deploymentvalidation is None:
-            return None
+            return
         cluster = deploymentvalidation['cluster']
-        dvdata = deploymentvalidation['data']
-        LOG.info('%s posting validations for %s', self.logmarker, cluster)
-        response = None
-        if not self.dry_run:
-            endpoint = (f'{self.dashdotdb_url}/api/v1/'
-                        f'deploymentvalidation/{cluster}')
-            response = requests.post(url=endpoint, json=dvdata,
-                                     auth=(self.dashdotdb_user,
-                                           self.dashdotdb_pass),
-                                     timeout=(5, 30))
-            try:
-                response.raise_for_status()
-            except requests.exceptions.RequestException as details:
-                LOG.error('%s error posting %s - %s',
-                          self.logmarker, cluster, details)
+        # dvd.data.data.result.[{metric,values}]
+        dvresult = deploymentvalidation.get('data').get('data').get('result')
+        if dvresult is None:
+            return
+        LOG.info('%s Processing (%s) metrics for: %s', self.logmarker,
+                 len(dvresult),
+                 cluster)
+        metrics = list(self._chunkify(dvresult, self.chunksize))
+        # keep everything but metrics from prom blob
+        deploymentvalidation['data']['data']['result'] = []
+        for metric_chunk in metrics:
+            # to keep future-prom-format compatible,
+            # keeping entire prom blob but iterating on metrics by
+            # self.chunksize max metrics in one post
+            dvdata = deploymentvalidation['data']
+            dvdata['data']['result'] = metric_chunk
+            response = None
+            if not self.dry_run:
+                endpoint = (f'{self.dashdotdb_url}/api/v1/'
+                            f'deploymentvalidation/{cluster}')
+                response = requests.post(url=endpoint, json=dvdata,
+                                         auth=(self.dashdotdb_user,
+                                               self.dashdotdb_pass),
+                                         timeout=(5, 120))
+                try:
+                    response.raise_for_status()
+                except requests.exceptions.RequestException as details:
+                    LOG.error('%s error posting DVO data (%s): %s',
+                              self.logmarker, cluster, details)
 
-        LOG.info('%s cluster %s synced', self.logmarker, cluster)
+        LOG.info('%s DVO data for %s synced to dddb', self.logmarker, cluster)
         return response
 
     def _promget(self, url, query, token=None):
@@ -62,7 +81,7 @@ class DashdotdbDVO:
         response = requests.get(url,
                                 params=params,
                                 headers=headers,
-                                timeout=(5, 30))
+                                timeout=(5, 120))
         response.raise_for_status()
 
         response = response.json()
@@ -86,14 +105,14 @@ class DashdotdbDVO:
                                                  query=promquery,
                                                  token=promtoken)
         except requests.exceptions.RequestException as details:
-            LOG.error('%s error accessing prometheus - %s, %s',
+            LOG.error('%s error accessing prometheus (%s): %s',
                       self.logmarker, cluster, details)
             return None
 
         return {'cluster': cluster,
                 'data': deploymentvalidation}
 
-    def _get_clusters(self):
+    def _get_clusters(self, cnfilter=None):
         # 'cluster': 'fooname',
         # 'tokenpath':
         #  'path': 'app-sre/creds/kubeube-configs/barpath',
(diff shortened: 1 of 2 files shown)


mmclanerh:
tested this locally, using the giant blob in stage. also tested new single-cluster run {optional} parameter.
> $ qr dashdotdb-dvo --cluster-name=app-sre-NNNN-01
> 
> [2021-01-19 16:45:46] [INFO] [dashdotdb_dvo.py:_post:43] - DDDB_DVO: Processing (5240) metrics for: app-sre-NNNN-01

reconcile/dashdotdb_dvo.py
```diff
@@ -23,29 +23,48 @@ def __init__(self, dry_run, thread_pool_size):
         self.dashdotdb_url = secret_content['url']
         self.dashdotdb_user = secret_content['username']
         self.dashdotdb_pass = secret_content['password']
+        self.chunksize = secret_content.get('chunksize') or '20'
```
apahim, line 26: I believe that the defaulkt should be `len(data)`, which means you don't set a default here...

reconcile/dashdotdb_dvo.py
```diff
@@ -23,29 +23,48 @@ def __init__(self, dry_run, thread_pool_size):
...
         self.logmarker = "DDDB_DVO:"
 
+    def _chunkify(self, data, size):
```
apahim, line 29: ... but here: `if not size: size = len(data)`.

apahim: or, even better: `if not size: return data`

apahim (approved):
looks good. small nitpick, nothing critical. do we still need bigger body size in nginx after this?

reconcile/dashdotdb_dvo.py
```diff
@@ -23,29 +23,48 @@ def __init__(self, dry_run, thread_pool_size):
...
-
-        LOG.info('%s cluster %s synced', self.logmarker, cluster)
+        # dvd.data.data.result.[{metric,values}]
+        dvresult = deploymentvalidation.get('data').get('data').get('result')
+        if dvresult is None:
+            return
+        LOG.info('%s Processing (%s) metrics for: %s', self.logmarker,
+                 len(dvresult),
+                 cluster)
+        metrics = list(self._chunkify(dvresult, self.chunksize))
```
apahim, line 44: calling list() here just defeats the generator purpose

apahim:
Btw, mixing the cluster filter with the chunks thing in the same commit makes it harder to review :)

reconcile/dashdotdb_dvo.py
mmclanerh: the chunksize is intended to be the max number of metrics we ship within a single post iteration, not the size the overall input. this seems to be working okay. i threw some debugs in and tried input of varying sizes (bigger and smaller than the limit). i'll make a small change in the chunk code, detailed below, to skip the chunking where we don't need it.

an example, showing where the last iteration of a list had a remainder of 2 result elements:
["data"]["result"][{metric{},value{}}]
```
{
	'status': 'success',
	'data': {
		'resultType': 'vector',
		'result': [{
			'metric': {
				'__name__': 'deployment_validation_operator_request_limit_validation',
				'endpoint': 'http-metrics',
				'exported_namespace': 'uhc-stage',
				'instance': '10.130.4.46:8383',
				'job': 'deployment-validation-operator-metrics',
				'kind': 'ReplicaSet',
				'name': 'uhc-acct-mngr-7fc4576947',
				'namespace': 'deployment-validation-operator',
				'pod': 'deployment-validation-operator-59b55df96c-xld4w',
				'service': 'deployment-validation-operator-metrics'
			},
			'value': [1611174950.067, '0']
		}, {
			'metric': {
				'__name__': 'deployment_validation_operator_request_limit_validation',
				'endpoint': 'http-metrics',
				'exported_namespace': 'visual-qontract-stage',
				'instance': '10.130.4.46:8383',
				'job': 'deployment-validation-operator-metrics',
				'kind': 'Deployment',
				'name': 'openshift-acme',
				'namespace': 'deployment-validation-operator',
				'pod': 'deployment-validation-operator-59b55df96c-xld4w',
				'service': 'deployment-validation-operator-metrics'
			},
			'value': [1611174950.067, '1']
		}]
	}
}
```

reconcile/dashdotdb_dvo.py
mmclanerh: size is the max number of elements we want in a list. the typical input looks something like this:

```
data [
     resulttype.vector
     result[0].metric
     result[0].value
     result[1].metric
     result[1].value
     ...
     result[n].metric
     result[n].value
]
```
what this does it keep data.result[n] from being more than chunksize. we replicate all of result[n] to a list of dict (metrics)  where n<chunksize.. we then blow away the original results keeping the rest of the blob, and iterate on the list (metrics) for the post.

hrm. it shouldn't be necessary since the first iteration of the chunk function wouldn't satsify, but i'll add an if to skip running the chunk function if the data we want to chunk up is already less than or equal to the max desired size.

reconcile/dashdotdb_dvo.py
mmclanerh: i thought i had totally missed something, but i've verified up and down it's working how i intended. i'm taking a list of dict and converting it to smaller chunks, then iterating on that list.. combing each element with another (static) list.
```
2021-01-20 14:11:08] [INFO] [dashdotdb_dvo.py:_post:75] - DDDB_DVO: DVO data for app-sre-stage-01 synced to dddb
[2021-01-20 14:11:09] [INFO] [dashdotdb_dvo.py:_post:43] - DDDB_DVO: Processing (5342) metrics for: app-sre-stage-01
[2021-01-20 14:11:09] [INFO] [dashdotdb_dvo.py:_post:51] - DDDB_DVO: Chunked metrics into (268) elements for: app-sre-stage-01
[2021-01-20 14:11:09] [INFO] [dashdotdb_dvo.py:_post:61] - DDDB_DVO: len([data][result]):20
[2021-01-20 14:11:09] [INFO] [dashdotdb_dvo.py:_post:61] - DDDB_DVO: len([data][result]):20
{snip}
[2021-01-20 14:11:09] [INFO] [dashdotdb_dvo.py:_post:61] - DDDB_DVO: len([data][result]):20
[2021-01-20 14:11:09] [INFO] [dashdotdb_dvo.py:_post:61] - DDDB_DVO: len([data][result]):2
[2021-01-20 14:11:09] [INFO] [dashdotdb_dvo.py:_post:75] - DDDB_DVO: DVO data for app-sre-stage-01 synced to dddb
```

reconcile/dashdotdb_dvo.py
mmclanerh: ..just to verify, i'm not mixing language syntax, this is valid:
```
j = {'one':'won', 'two':'too'}                                                  
print(f'j["one"]: {j.get("one") or "missing"}')  # j["one"]: won                
print(f'j["xyz"]: {j.get("xyz") or "missing"}')  # j["xyz"]: missing            
```

mmclanerh:
> Btw, mixing the cluster filter, the timeout (which is the PR title) and the chunks thing in the same commit makes it harder to review :)

yeah, sorry about that. i usually do mean to keep features/fixes isolated. in this case i needed to be able to test the chunking feature more quickly than iterating across all the clusters.. so it ended up helping run tests for this new feature and became it's own new feature at the same time.  note: adding another item, to address private clusters, in it's own PR.. as one should. :P

reconcile/dashdotdb_dvo.py
apahim: I understand the concept... I'm just saying that 20 is a bad default, because it's not in the function signature, which makes it harder for the caller to know what the default is, and because a chunking mechanism should probably not chunk if the chunk size is not given.
As I said, no big deal.

reconcile/dashdotdb_dvo.py
apahim: It's working indeed. All I'm saying is that creating a generator and then casting it into a list defeats its purpose.

reconcile/dashdotdb_dvo.py
mmclanerh: Totally fair. If I was adding something like that to the sre-toolbox, I'd opt for not stuffing a default in the init. Intent: wanted to be able to tune the chunksize at run-time, if we need, without waiting for the mammoth process of pushing a new qontract update. While we're still discovering the seams, in the context of the dvo class here it's probably okay as is. I did pre-emptively add a just-incase check if the property isn't set before we call the function here.

reconcile/dashdotdb_dvo.py
mmclanerh: Maybe I'm using the generator in weird way. I'm using the casting to list() as an intentionally cheap way to trigger iteration. I guess that might not be what one normally expects. I could wrap the post inside of the generator.. but I often find that sometimes harder for folks to read/grok.

apahim: `for metric in self._chunkify(...):` should be enough

mmclanerh:
> looks good. small nitpicks, nothing critical. do we still need bigger body size in nginx after this?

Yeah, I'd still like to understand why dashdotdb ingestion is so crazy slow. It looks to be this way for cso and dvo with some inserts taking seconds to return. I suspect debugging the sql queries and running some explains will highlight some ways to improve. When I was digging around previously more in DashDotDb, I found some of the ways SqlAlchemy was constructing queries to be a bit.. greedy? Some intense join statements.. but I dunno. Maybe there's a bottleneck in the python code itself.. tbd. Ideally, the less http calls we need to make, the better.. as the amount of data we're pushing in grows, I'd rather we're able also reduce the amount of new http session we need to open. ..git checkout -b kafka-dashdotdb-qfeeder? :P 

For now, leaving both ability to set a value and a value over 1MB in place.

PR #1332, opened 2021-01-20
only allow to override enable_deletion if no errors were found

maorfr:
fixes https://github.com/app-sre/qontract-reconcile/pull/1279#discussion_r551238698


apahim (approved):
makes sense

rporres (approved):
Thx for fixing this

PR #1334, opened 2021-01-20
Validate container names against DNS-1123 in OpenshiftResource

jfchevrette:
This PR adds validation for container names in OpenshiftResource. Corresponding tests have also been added.

This also extracts the DNS-1123 based regexes as global variables so they can be reused elsewhere in the future.

I have also changed the error messages slightly for clarity and added a link to the kubernetes docs on object naming

### DIFF
--- a/reconcile/test/fixtures/openshift_resource/invalid_resource_container_name_format.yml
+++ b/reconcile/test/fixtures/openshift_resource/invalid_resource_container_name_format.yml
@@ -1,0 +1,20 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: foo
+  labels:
+    app: foo
+spec:
+  replicas: 3
+  selector:
+    matchLabels:
+      app: foo
+  template:
+    metadata:
+      labels:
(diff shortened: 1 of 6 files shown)


BumbleFeng:
There is also a lengthy limit, IIRC

jfchevrette:
Please hold on reviewing this PR I will change it to do the validation in openshift_base rather than locally in the sql-query integration

maorfr:
i agree with adding this to our integrations. until this is solved, we can put a patch in place: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/14081

jfchevrette:
I've refactored the PR and updated the description accordingly. The validation is now done in OpenshiftResource itself so this should catch issues globally.

jfchevrette:
Added length checks and tests covering all objects names tests we have right now

rporres:
Is this ready to review, @jfchevrette . It definitely looks to me :)

rporres (approved):
LGTM... And with updated tests <3

PR #1335, opened 2021-01-21
Dvo insecureskip

mmclanerh:
Adjusts verify in prometheus request based on whether cluster.spec['private'] returns True or False

### DIFF
--- a/reconcile/dashdotdb_dvo.py
+++ b/reconcile/dashdotdb_dvo.py
@@ -127,7 +127,7 @@ class DashdotdbDVO:
 
     def _get_clusters(self, cnfilter=None):
         # 'cluster': 'fooname'
-        # 'private': 'XX'
+        # 'private': False
         # 'prometheus': 'https://prometheus.baz.tld'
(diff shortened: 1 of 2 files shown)


reconcile/dashdotdb_dvo.py
```diff
@@ -87,9 +87,11 @@ def _promget(self, url, query, token=None):
                   }
         if token:
             headers["Authorization"] = (f"Bearer {token}")
+        ssl_verify = False if is_private else True
```
rporres, line 90: nit: Most likely you make `ssl_verify` an argument of `_promget` and get it via the `private` value so that you don't need to calculate it here. It would also feel more natural to pass that option to an http related function

rporres (approved):
A small thing. Looks good otherwise

reconcile/dashdotdb_dvo.py
mmclanerh: Excellent point. I moved this test out of the request method and made ssl_verify a "more natural" parameter.

PR #1327, opened 2021-01-18
[openshift-saas-deploy] add support for automated promotions

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2244

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/13902

this PR adds support for automated promotions for saas file targets defining `auto: true` in the `promotion` section.
this works by sending messages to the SQS gateway to promote a change and consumed by the MR consumer integration.

PR #1336, opened 2021-01-21
[SQSGateway] only initiate submitter queue account

maorfr:
fixes #1327

this will avoid deployment jobs from initiating tokens for all AWS accounts, and only do it for the AWS account with the MR submitter queue.

fixes https://ci.ext.devshift.net/view/github-mirror/job/openshift-saas-deploy-saas-github-mirror-test-App-SRE-stage/25/console


apahim (approved):
lgtm, thank you!

PR #1338, opened 2021-01-22
New integration to check alertmanager config

rporres:
The approach is not very satisfactory, as we need to know the file where
the config is located. The other approach would be to define a schema
for the alertmanager config in app-interface, but it looks like
duplicating the job as we're going to check the schema anyway. As we
only have one alertmanager config this looks enough for the moment but
we may want to define a very lax schema in app-interface in the future
to better identify these files.

Closes [APPSRE-2419](https://issues.redhat.com/browse/APPSRE-2419)
Needs https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/14174

### DIFF
--- a/reconcile/alertmanager_config_validator.py
+++ b/reconcile/alertmanager_config_validator.py
@@ -1,0 +1,88 @@
+import sys
+import semver
+import base64
+import logging
+
+import reconcile.utils.gql as gql
+import reconcile.utils.threaded as threaded
+import reconcile.utils.amtool as amtool
+import reconcile.openshift_resources_base as orb
+
+from reconcile.status import ExitCodes
+
+QONTRACT_INTEGRATION = 'alertmanager_config_validator'
+QONTRACT_INTEGRATION_VERSION = semver.format_version(0, 1, 0)
+
+ALERTMANAGER_SECRET_PATH = '/observability/alertmanager/' \
+                           'alertmanager-instance.secret.yaml'
+ALERTMANAGER_SECRET_KEY = 'alertmanager.yaml'
+
+
+def check_config(config):
+    '''Helper function to run amtool check-config via threaded.run'''
+    config['check_result'] = amtool.check_config(config['data'])
+    return config
+
+
+def get_config_data(namespace_info):
+    '''Returns a dict with the alertmanager config from the cluster refered
+       in the namespace info'''
+
+    openshift_resources = namespace_info.get('openshiftResources')
+
+    if not openshift_resources:
+        return
+
+    for r in openshift_resources:
+        path = r['path']
+        if path != ALERTMANAGER_SECRET_PATH:
+            continue
+
+        openshift_resource = \
+            orb.fetch_openshift_resource(resource=r, parent=namespace_info)
+
+        try:
+            encoded = openshift_resource.body['data'][ALERTMANAGER_SECRET_KEY]
+            data = base64.b64decode(encoded).decode('utf-8')
+        except KeyError:
+            logging.warning('No data found in config secret for key '
+                            f'{ALERTMANAGER_SECRET_KEY}')
+            return
+
+        return {'cluster': namespace_info['cluster']['name'], 'data': data}
+
+
+def run(dry_run, thread_pool_size=20, cluster_name=None):
+    orb.QONTRACT_INTEGRATION = QONTRACT_INTEGRATION
+    orb.QONTRACT_INTEGRATION_VERSION = QONTRACT_INTEGRATION_VERSION
+
+    gqlapi = gql.get_api()
+
+    configs = []
+    for namespace_info in gqlapi.query(orb.NAMESPACES_QUERY)['namespaces']:
+        if namespace_info['name'] != 'openshift-customer-monitoring':
+            continue
+
+        if cluster_name and namespace_info['cluster']['name'] != cluster_name:
+            continue
+
+        config_data = get_config_data(namespace_info)
+        if config_data:
+            configs.append(config_data)
+
+    if not configs:
+        logging.error(f'No configs found at {ALERTMANAGER_SECRET_PATH}')
+        sys.exit(ExitCodes.ERROR)
+
+    result = threaded.run(func=check_config,
+                          iterable=configs,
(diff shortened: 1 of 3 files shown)


reconcile/alertmanager_config_validator.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def run(dry_run, thread_pool_size=20, cluster_name=None):
+    orb.QONTRACT_INTEGRATION = QONTRACT_INTEGRATION
+    orb.QONTRACT_INTEGRATION_VERSION = QONTRACT_INTEGRATION_VERSION
+
+    gqlapi = gql.get_api()
+
+    configs = []
+    for namespace_info in gqlapi.query(orb.NAMESPACES_QUERY)['namespaces']:
```
maorfr, line 62: this is the 3rd duplicate of `gqlapi.query(orb.NAMESPACES_QUERY)['namespaces']`. is it time to extract this to a `get_namespaces` function in `orb`? just a nit

reconcile/utils/amtool.py
```diff
@@ -0,0 +1,43 @@
...
+        self.message = message
+
+    def __str__(self):
+        return str(self.message).replace('\n', '')
+
+    def __bool__(self):
+        return self.is_ok
+
+
+def check_config(yaml_str):
```
maorfr, line 18: would it make sense to wrap this up into a `Amtool` class? that will help you get rid of the helper function in the integration.

reconcile/alertmanager_config_validator.py
```diff
@@ -0,0 +1,88 @@
...
+import reconcile.utils.threaded as threaded
+import reconcile.utils.amtool as amtool
+import reconcile.openshift_resources_base as orb
+
+from reconcile.status import ExitCodes
+
+QONTRACT_INTEGRATION = 'alertmanager_config_validator'
+QONTRACT_INTEGRATION_VERSION = semver.format_version(0, 1, 0)
+
+ALERTMANAGER_SECRET_PATH = '/observability/alertmanager/' \
```
maorfr, line 16: would it make sense to mark the resources to check in the schema instead? for grafana dashboards we have `validate_json: true`.

reconcile/alertmanager_config_validator.py
```diff
@@ -0,0 +1,88 @@
...
+        path = r['path']
+        if path != ALERTMANAGER_SECRET_PATH:
+            continue
+
+        openshift_resource = \
+            orb.fetch_openshift_resource(resource=r, parent=namespace_info)
+
+        try:
+            encoded = openshift_resource.body['data'][ALERTMANAGER_SECRET_KEY]
+            data = base64.b64decode(encoded).decode('utf-8')
```
maorfr, line 46: being defensive, should we check if the resource is of kind `Secret` and decode based on that? maybe we'd want to validate ConfigMaps at some point?

reconcile/cli.py
```diff
@@ -1184,3 +1185,13 @@ def integrations_validator(ctx):
 def prometheus_rules_tester(ctx, thread_pool_size, cluster_name):
     run_integration(reconcile.prometheus_rules_tester, ctx.obj,
                     thread_pool_size, cluster_name)
+
+
+@integration.command()
+@threaded()
+@binary(['amtool'])
```
maorfr, line 1192: nice! would this be added to qontract-reconcile-base?

maorfr (approved):
LGTM

i've added some comments on possible changes. since things are working as is, i'm fine with iterating over them in a future PR if you find them useful.

thanks for this, this will help us A LOT!

reconcile/alertmanager_config_validator.py
rporres: Good point. I'll do it

reconcile/alertmanager_config_validator.py
rporres: Hmmm, are you proposing that we add a new property, e.g. `validate_alertmanager_config` to trigger this validation instead of the horrible path hardcoding? Because I like that approach much more than the current one 😂

reconcile/alertmanager_config_validator.py
rporres: Yes, it makes sense.

reconcile/cli.py
rporres: Yes, I forgot to create the PR and link it 😅  https://github.com/app-sre/container-images/pull/9

reconcile/utils/amtool.py
rporres: That would be a way to do it, yes. But I don't see the `check_config` helper function out of the integration code as it deals with the data structures that are in the integration and should never be part of the `Amtool` class. So at the moment, I don't see much benefit in changing the shape of this file

reconcile/alertmanager_config_validator.py
maorfr: maybe even only iterate over namespaces with are `observabilityNamespace`?

reconcile/alertmanager_config_validator.py
maorfr: yes! perhaps even `validate_alertmanager_config_key: alertmanager.yaml` to catch 2 birds with 1 stone?

rporres: While the idea of catching two birds with one stone is tempting, I see this more properly done with two properties:

`validate_alertmanager_config`: <boolean>
`validate_alertmanager_config_key`: string with a default of `alertmanager.yaml`

maorfr: leaving the implementation decision to you

reconcile/alertmanager_config_validator.py
rporres: Hmmm, good idea. I didn't remember about that property. I'll take a look into that. That removes another hardcoded value which is always a good idea

rporres:
Thanks a lot for your review and ideas, @maorfr . Thinking about this let's just remove this integration and do this in openshift-resources as we're already doing the hard work (resolving the templates) there...

rporres:
New attempt in https://github.com/app-sre/qontract-reconcile/pull/1340

PR #1341, opened 2021-01-25
Fix dddb crons

mmclanerh:
Ensure DDDB crons don't run multiple of the same job concurrently and add additional success/fail pod history to dddb-dvo job

### DIFF
--- a/helm/qontract-reconcile/templates/template.yaml
+++ b/helm/qontract-reconcile/templates/template.yaml
@@ -241,21 +241,9 @@ objects:
     name: qontract-reconcile-{{ $integration.name }}
   spec:
     schedule: "{{ $integration.cron }}"
-    {{- if $integration.concurrencyPolicy }}
-    concurrencyPolicy: {{ $integration.concurrencyPolicy }}
-    {{- else }}
-    concurrencyPolicy: Allow
-    {{- end }}
-    {{- if $integration.successfulJobHistoryLimit }}
-    successfulJobHistoryLimit: {{ $integration.successfulJobHistoryLimit }}
-    {{- else }}
-    successfulJobHistoryLimit: 3
-    {{- end }}
-    {{- if $integration.failedJobHistoryLimit }}
-    failedJobHistoryLimit: {{ $integration.failedJobHistoryLimit }}
-    {{- else }}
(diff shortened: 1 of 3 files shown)


helm/qontract-reconcile/templates/template.yaml
```diff
@@ -241,6 +241,21 @@ objects:
     name: qontract-reconcile-{{ $integration.name }}
   spec:
     schedule: "{{ $integration.cron }}"
+    {{- if $integration.concurrencyPolicy }}
+    concurrencyPolicy: {{ $integration.concurrencyPolicy }}
```
jfchevrette, line 245: to simplify I think you could do something like `{{ $integration.concurrencyPolicy | default "Allow" }}` and remove the if block completely

same for the other if/else blocks added after this one

jfchevrette:
I've added a suggestion on simplifying if/else blocks for setting default values, other than that this LGTM!

helm/qontract-reconcile/templates/template.yaml
mmclanerh: Awesome suggestion! I'm less familiar with Helm syntax. I'll go with the abbreviated conditional route as it hopefully will lend to easier reading. Thanks!

mmclanerh: @jfchevrette PTAL; TIA!

mmclanerh: new version updated

jfchevrette (approved):

maorfr:
very nice @mmclanerh!
i already used it in #1344 (notice that since the value was `0` it required quotes. maybe that can be improved? leaving it to you).

mmclanerh:
> 
> 
> very nice @mmclanerh!
> i already used it in #1344 (notice that since the value was `0` it required quotes. maybe that can be improved? leaving it to you).

Ah, good catch and thanks for the quick fixup. Yeah in some cases the value is int (no quote) vs string (quoted). I had looked at an existing yaml to verify quoted vs non-quoted values, and maybe I missed that.. Also noted there was a function to add quotes, where needed:  ' | quote '.

PR #1339, opened 2021-01-25
Revert "monthly validation report metrics for visual-app-interface (#…

apahim:
…1330)"

This reverts commit 721c5c70d3c5ab0a3c64c4894c7a2250d3087b6d.

@mmclanerh the reporter is broken in different ways... first one being:
```python
INFO: collecting dashdotdb information for quayio
Traceback (most recent call last):
  File "/home/apahim/git/qontract-reconcile/venv/bin/app-interface-reporter", line 33, in <module>
    sys.exit(load_entry_point('reconcile', 'console_scripts', 'app-interface-reporter')())
  File "/home/apahim/git/qontract-reconcile/venv/lib64/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/apahim/git/qontract-reconcile/venv/lib64/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/apahim/git/qontract-reconcile/venv/lib64/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/apahim/git/qontract-reconcile/venv/lib64/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/apahim/git/qontract-reconcile/tools/app_interface_reporter.py", line 535, in main
    apps = get_apps_data(now)
  File "/home/apahim/git/qontract-reconcile/tools/app_interface_reporter.py", line 484, in get_apps_data
    status = ('Passed', 'Failed')[sample.labels['status']]
TypeError: tuple indices must be integers or slices, not str
```

Unfortunatelly I'm on a tight schedule here, so I'm going to revert this hoping that you can check it out?

PR #1342, opened 2021-01-25
Dddb dvo reporter

mmclanerh:
o fix casting typo in monthly reporter for dddb/dvo
o fix method call name (. vs _) in monthly reporter for dddb/dvo

### DIFF
--- a/tools/app_interface_reporter.py
+++ b/tools/app_interface_reporter.py
@@ -204,6 +204,7 @@ class Report(object):
                                 'post_deploy_job': post_deploy_job})
         return results
 
+    @staticmethod
     def get_validations_content(deployment_validations):
         parsed_metrics = []
         if not deployment_validations:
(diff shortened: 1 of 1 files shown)


tools/app_interface_reporter.py
```diff
@@ -495,33 +515,57 @@ def get_apps_data(date, month_delta=1):
...
+                        cluster = sample.labels['cluster']
+                        if app_namespace['cluster']['name'] != cluster:
+                            continue
+                        namespace = sample.labels['namespace']
+                        if app_namespace['name'] != namespace:
+                            continue
+                        validation = sample.labels['validation']
+                        # dvo: fail == 1, pass == 0, py: true == 1, false == 0
+                        # so: ({false|pass}, {true|fail})
+                        status = ('Passed', 'Failed')[int(sample.labels['status'])]
```
mmclanerh, line 555: @apahim  this was the hiccup causing the casting exception; added int(). ptal -- thanks!

mmclanerh:
note: this is a un-revert + small fix, from https://github.com/app-sre/qontract-reconcile/pull/1339

rporres (approved):

PR #1343, opened 2021-01-25
APPSRE-2919: SendGrid Integration

jmelis:
More info here: https://issues.redhat.com/browse/APPSRE-2919

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -1,4 +1,5 @@
 import logging
+from jinja2 import Template
 
 import reconcile.utils.gql as gql
 
@@ -951,6 +952,8 @@ ROLES_QUERY = """
         }
         role
       }
+
+      {% if aws %}
       aws_groups {
         name
         path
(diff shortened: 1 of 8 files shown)


reconcile/queries.py
```diff
@@ -959,18 +962,32 @@ def get_quay_orgs():
...
   }
 }
 """
 
 
-def get_roles():
+def get_roles(aws=True, saas_files=True, sendgrid=False):
     gqlapi = gql.get_api()
-    return gqlapi.query(ROLES_QUERY)['users']
+    query = Template(ROLES_QUERY).render(aws=aws,
```
jmelis, line 987: Templated queries is the only "interesting" part of this PR.

The rationale behind this is in order to avoid retrieving schemas that aren't necessary to the integration. If you fetch the roles in order to reconcile sendgrid teams, you don't care about the aws groups. If you include it in the query, the schema needs to be declared in the integration schema. Basically, we don't want to trigger the sendgrid PR integration in the PR check if the PR only includes aws groups changes.

maorfr: :heart:
i'm going to adopt this from now on!

reconcile/sendgrid_teammates.py
```diff
@@ -0,0 +1,98 @@
...
+            if not dry_run:
+                req = {
+                    'email': user.email,
+                    'scopes': [],
+                    'is_admin': True,
+                }
+
+                response = sg_client.teammates.post(request_body=req)
+                if int(response.status_code / 100) != 2:
+                    LOG.error(['error inviting user',
```
maorfr, line 79: in addition to logging an error we should eventually exit with an error

maorfr (approved):
LGTM

straight and to the point. this was a very simple review. thank you!
don't forget to add the integration to the helm chart (and thus openshift template).

the internals of the `SendGridAPIClient` would probably end up living in `utils`, which will help us disconnect it from the integration logic. this will probably be needed only when we create the next sendgrid integration, so not needed right now. not like we are not doing this in many places as is.

reconcile/sendgrid_teammates.py
jmelis: thanks for catching that! fixed

jmelis:
> don't forget to add the integration to the helm chart (and thus openshift template).
thanks! I just added it

PR #1345, opened 2021-01-26
get providerVersion from account schema and not vault

arilivigni:
Removed references to vault secret for the `aws_provider_version` to `providerVersion` in the account config. 
Updates for `providerVersion` have already been made on the app-interface side and merged.

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -232,6 +232,7 @@ AWS_ACCOUNTS_QUERY = """
     consoleUrl
     resourcesDefaultRegion
     supportedDeploymentRegions
+    providerVersion
     accountOwners {
       name
       email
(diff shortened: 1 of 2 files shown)


jfchevrette:
LGTM but I don't have a ton of experience around the Terrascript code so I'd like if another reviewer could +1 as well

Thanks @arilivigni!

jfchevrette (approved):

arilivigni:
@maorfr and @apahim your feedback is always welcome.

maorfr (approved):
LGTM

this is a nice approach! what we currently do is we populate the `config` in https://github.com/app-sre/qontract-reconcile/blob/758989149760b810fe234986280f05c5c12d7e9d/reconcile/utils/terrascript_client.py#L169-L170

which of the two do you think is preferable?

PR #1346, opened 2021-01-26
Change bucket distribution for reconcile times

rporres:
We need more resolution in the upper part as it is where data is getting concentrated.

image


jmelis (approved):
LGTM

PR #1348, opened 2021-01-26
Improved detection of Vault KV version and better handling of Forbidden errors

jfchevrette:
- Properly raise vault access forbidden errors (with useful message)
- Refactor the naive try/except for falling back to KV_1 if KV_2 failed into a proper detection of the KV version 

This will hopefuully solve most of the very lengthy stack traces we've been getting in the logs which were caused by bad secrets and/or permission denied. With this we should be able to spot and fit these more easily


maorfr (approved):
LGTM

great usage of the cache!

PR #1349, opened 2021-01-27
[WiP] Don't mirror any docker.io images to public repos

apahim:
Aparently the issue is not only with official docker images, but with
any images hosted in docker.io.


maorfr:
replaced by #2700

PR #1351, opened 2021-01-27
[ocm-clusters] add support for autoscaling and cluster update

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2917

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/14339

### DIFF
--- a/reconcile/ocm_clusters.py
+++ b/reconcile/ocm_clusters.py
@@ -94,6 +94,10 @@ def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
 
             if current_spec != desired_spec:
                 # update cluster
+                logging.debug(
+                    '[%s] desired spec %s is different ' +
+                    'from current spec %s',
+                    cluster_name, desired_spec, current_spec)
                 logging.info(['update_cluster', cluster_name])
                 ocm = ocm_map.get(cluster_name)
                 ocm.update_cluster(cluster_name, desired_spec, dry_run)
--- a/reconcile/ocm_clusters.py
+++ b/reconcile/ocm_clusters.py
@@ -18,6 +18,10 @@ def fetch_desired_state(clusters):
     # remove unused keys
     for desired_spec in desired_state.values():
         desired_spec['spec'].pop('upgrade', None)
+        # remove empty keys in spec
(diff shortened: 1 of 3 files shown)


reconcile/ocm_clusters.py
```diff
@@ -12,12 +12,16 @@
...
 
-def fetch_current_state(clusters):
+def fetch_desired_state(clusters):
     desired_state = {c['name']: {'spec': c['spec'], 'network': c['network']}
                      for c in clusters}
     # remove unused keys
     for desired_spec in desired_state.values():
         desired_spec['spec'].pop('upgrade', None)
+        # remove empty keys in spec
+        empty_keys = [k for k, v in desired_spec['spec'].items() if v is None]
```
apahim, line 22: you can use a dict comprehension and avoid the pop ahead... `{k: n for k, v in d.items() if v is not None }`

reconcile/utils/ocm.py
```diff
@@ -140,12 +147,60 @@ def create_cluster(self, name, cluster, dry_run):
...
+        api = f'/api/clusters_mgmt/v1/clusters/{cluster_id}'
+        cluster_spec = cluster['spec']
+        ocm_spec = {
+            'nodes': {
+                'compute_machine_type': {
+                    'id': cluster_spec['instance_type']
+                }
+            },
+            'storage_quota': {
+                'value': float(cluster_spec['storage'] * pow(1024, 3))
```
apahim, line 182: no need to calc `1024**3` all the time... let's use the actual number and add a commet... `... * 1073741824  # 1024 ** 3`

apahim (approved):
small nitpics that can safelly be ignored. lgtm, thanks!

apahim:
Sorry, for the first time I clicked "Update branch" and it did the worst thing it could do :(
Running `git push -f` should fix it

reconcile/ocm_clusters.py
maorfr: absolutely!

reconcile/utils/ocm.py
maorfr: makes sense

PR #1352, opened 2021-01-27
[jenkins-job-builder] print out job

BumbleFeng:
https://issues.redhat.com/browse/APPSRE-2773

### DIFF
--- a/reconcile/jenkins_job_builder.py
+++ b/reconcile/jenkins_job_builder.py
@@ -31,6 +31,7 @@ QUERY = """
 """
 
 QONTRACT_INTEGRATION = 'jenkins-job-builder'
+GENERATE_TYPE = ['jobs', 'views']
 
 
 def get_openshift_saas_deploy_job_name(saas_file_name, env_name, settings):
@@ -38,7 +39,7 @@ def get_openshift_saas_deploy_job_name(saas_file_name, env_name, settings):
     return f"{job_template_name}-{saas_file_name}-{env_name}"
 
 
-def collect_saas_file_configs():
+def collect_saas_file_configs(settings, instance_name=None):
     # collect a list of jobs per saas file per environment.
     # each saas_file_config should have the structure described
     # in the above query.
@@ -47,11 +48,12 @@ def collect_saas_file_configs():
     saas_file_configs = []
     repo_urls = set()
     saas_files = queries.get_saas_files()
-    settings = queries.get_app_interface_settings()
     job_template_name = settings['saasDeployJobTemplate']
     for saas_file in saas_files:
         saas_file_name = saas_file['name']
         jc_instance = saas_file['instance']
+        if instance_name is not None and jc_instance['name'] != instance_name:
+            continue
         app_name = saas_file['app']['name']
         # currently ignoring the actual Slack workspace
         # as that is configured in Jenkins.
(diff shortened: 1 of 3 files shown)


maorfr:
i admit it took me a while to understand this PR.

you are using `--jenkins-config` to indicate that you want configs of a specific jenkins instance printed. at the same time, since when we print, we don't want to get/print credentials, you are using the existence of `jenkins_config` to decide how to print the ini file. i was also a bit confused reading the code, as there are very different behaviors when `jenkins_config` is defined vs when it's not.

i think there are 2 different things (and hence 2 separate PRs):
1. the ability to run the integration for a specific jenkins instance (`--jenkins-config` is a bit confusing. what about `--instance-name`?)
2. the ability to `--print-only` (let's remain consistent with other integrations that have a print option).

splitting this into 2 PRs will also force you to write consistent (less `if; else` with very different code) and readable code.

it would also make sense to have a good description to the PR so your intentions would not need to be interpreted from the code. it's also a better way to let someone else review your code - you say what you meant to happen, and the reviewer can review with a critical eye if that indeed happens.

what do you think?

BumbleFeng:
Actually, it's not to get configs of a specific jenkins instance printed, It's to get the configs of a specific jenkins config file printed, for example if I run `qontract-reconcile  jenkins-job-builder --jenkins-config ci-ext-unleash-jobs` it will generate only two XML files defined in https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/unleash/cicd/jobs.yaml
```
.
└── desired
    ├── ci-ext
    │   ├── app-sre-unleash-gh-build-master
    │   │   └── config.xml
    │   └── app-sre-unleash-gh-pr-check
    │       └── config.xml
    └── ci-int
```
The original intention of this ticket is to help our tenants examine their jenkins job config before raise MR into app-interface while they may not familiar with all the default templates we have defined. However, there are two problems. First, they don't have vault creds. So I need to avoid using secretreader. Second, the output of all jobs is quite huge. No one can find the target job file easily. Meanwhile, we will only get XML file for jobs and views, so I add a filter to remove other configs but keep all the templates and secrets, which will still be used for generating the XML file.

BumbleFeng:
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/14374 helps me to separate the job file from default config, as my filer base on jenkins-config [type](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/jenkins/cicd/ci-centos/01-ci-centos-defaults.yaml#L17). The output will be much clear without tons of job files.

BumbleFeng:
print out the job in JSON is much more clear than generate the XML file.

BumbleFeng:
JJB manage all jobs which contain "" in Description. As a result, the minimal granularity is an instance.

reconcile/jenkins_job_builder.py
```diff
@@ -182,10 +191,22 @@ def validate_repos_and_admins(jjb, additional_repo_urls):
 
 
 @defer
-def run(dry_run, io_dir='throughput/', defer=None):
-    jjb, additional_repo_urls = init_jjb()
+def run(dry_run, print_only=False, jenkins_config=None, jenkins_job=None,
```
maorfr, line 194: it's customary to add new arguments at the end

reconcile/jenkins_job_builder.py
```diff
@@ -182,10 +191,22 @@ def validate_repos_and_admins(jjb, additional_repo_urls):
...
+        jenkins_instance=None, io_dir='throughput/', defer=None):
+    if jenkins_config or jenkins_job:
+        print_only = True
```
maorfr, line 197: i'm confused if `print-only` is passed in or calculated by the integration. it can get very confusing, there are so many new arguments in this PR...

reconcile/utils/jjb_client.py
```diff
@@ -22,13 +22,18 @@
...
 
+JJB_INI = "[jenkins]\nurl = https://JENKINS_URL"
+
 
 class JJB(object):
     """Wrapper around Jenkins Jobs"""
 
-    def __init__(self, configs, ssl_verify=True, settings=None):
+    def __init__(self, configs, ssl_verify=True,
+                 settings=None, local=False):
```
maorfr, line 32: i very much like the usage of `local` here. it's very understandable. but it's called print-only on the outside and then it's called local on the inside. why not just call in print-only?

maorfr (changes requested):
No doubt that the changes work. It is still a bit confusing to me. can we simplify things a bit?

apahim:
ohh forgot about this one. sorry, will review tomorrow.

reconcile/jenkins_job_builder.py
```diff
@@ -182,10 +194,18 @@ def validate_repos_and_admins(jjb, additional_repo_urls):
 
 
 @defer
-def run(dry_run, io_dir='throughput/', defer=None):
-    jjb, additional_repo_urls = init_jjb()
+def run(dry_run, io_dir='throughput/', print_only=False,
+        config_name=None, job_name=None, instance_name=None, defer=None):
+    if not print_only and config_name:
+        raise Exception("--config-name can only works with --print-only mode")
```
BumbleFeng, line 200: This PR introduces a third working mode for jenkins-job-builder and three flitter level.

flitter | print-only | dry-run | no-dry-run
------------ | -------------  | -------------  | -------------
config_name | print jobs defined in exact config file | generate desired XML file for jobs defined in exact config file and try to delete most other jobs [1] | delete most other jobs [2]
job_name | print jobs match with job_name | no effect  | no effect
instance_name | print jobs run for exact instance | compare jobs for in exact instance | update jobs for exact instance

This line is to avoid running into case [1] and [2]

reconcile/jenkins_job_builder.py
```diff
@@ -182,10 +194,18 @@ def validate_repos_and_admins(jjb, additional_repo_urls):
 
 
 @defer
-def run(dry_run, io_dir='throughput/', defer=None):
-    jjb, additional_repo_urls = init_jjb()
+def run(dry_run, io_dir='throughput/', print_only=False,
+        config_name=None, job_name=None, instance_name=None, defer=None):
+    if not print_only and config_name:
+        raise Exception("--config-name must works with --print-only mode")
```
BumbleFeng, line 200: This PR introduces a third working mode for jenkins-job-builder and three flitter level.

flitter | print-only | dry-run | no-dry-run
------------ | -------------  | -------------  | -------------
config_name | print jobs defined in exact config file | generate desired XML file for jobs defined in exact config file and try to delete most other jobs [1] | delete most other jobs [2]
job_name | print jobs match with job_name | no effect  | no effect
instance_name | print jobs for exact instance | compare jobs for exact instance | update jobs for exact instance

This line is to avoid running into case [1] and [2]

maorfr (approved):
LGTM

w00t w00t!

PR #1358, opened 2021-02-01
move templates into reconcile

maorfr:
this is needed due to the move of `utils` into `reconcile` in #1301.


rporres:
you can then remove the `templates` dir copy from the Dockerfile

rporres:
From the look of the tests you will also need to fix `reconcile/utils/template.py` to import `utils.templates`

PR #1360, opened 2021-02-01
will this fix templates?

maorfr:
should fix #1358

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ setup(
 
     packages=find_packages(exclude=('tests',)),
 
-    data_files=[('templates', glob('reconcile/templates/*.j2'))],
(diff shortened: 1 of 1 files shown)


rporres (approved):

setup.py
```diff
@@ -16,7 +16,7 @@
 
     packages=find_packages(exclude=('tests',)),
 
-    data_files=[('templates', glob('templates/*.j2'))],
+    data_files=[('templates', glob('reconcile/templates/*.j2'))],
```
rporres, line 19: Maybe?

I wonder why they aren't copied in the first place...

```suggestion
    data_files=[('reconcile/templates', glob('reconcile/templates/*.j2'))],
```

PR #1361, opened 2021-02-01
Add decorator for ensuring specific binary versions are available

jfchevrette:
- Add `binary_version` decorator
- Add `binary_version` decorator to all `cli.py` commands that uses `terraform` or `oc`

Solves APPSRE-2932 and should help with APPSRE-2724


maorfr (approved):
LGTM

love the regex approach!

PR #1363, opened 2021-02-02
Instrument oc time to reconcile

rporres:

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -1,5 +1,6 @@
 import os
 import json
+import time
 import logging
 import tempfile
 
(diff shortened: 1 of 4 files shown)


rporres:
I needed to get a bit of code from the decorator... I don't like it a lot, but I don't know how to do ir otherwise

reconcile/utils/oc.py
```diff
@@ -90,6 +92,18 @@ def __init__(self, server, token, jh=None, settings=None,
         else:
             self.api_resources = None
 
+        self.running_state = RunningState()
+
+        self.slow_oc_reconcile_threshold = \
+            float(os.environ.get('SLOW_OC_RECONCILE_THRESHOLD', 600))
```
jmelis, line 98: why float?

jmelis (approved):
LGTM!

PR #1364, opened 2021-02-02
[openshift-saas-deploy] add github large file handler

maorfr:
related to https://issues.redhat.com/browse/APPSRE-2959

upstream issue: https://github.com/PyGithub/PyGithub/issues/661

this PR adds a handle to errors caused by files with a size of over 1MB. the new code can handle files up to 100MB, which i doubt we'll ever reach in text files.

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -2,8 +2,9 @@ import os
 import yaml
 import json
 import logging
+import base64
 
-from github import Github
+from github import Github, GithubException
 from sretoolbox.container import Image
 from sretoolbox.utils import retry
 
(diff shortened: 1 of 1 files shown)


jmelis (approved):
LGTM

reconcile/utils/saasherder.py
```diff
@@ -196,6 +197,38 @@ def _collect_parameters(container):
...
+            #    'message': '<text>',
+            #    'errors': [{
+            #                  'resource': 'Blob',
+            #                  'field': 'data',
+            #                  'code': 'too_large'
+            #               }],
+            #    'documentation_url': '<url>'
+            # }
+            for error in errors:
+                if error['code'] == 'too_large':
```
rporres, line 220: Where does this code come from? Do we know it is not going to change?

maorfr: it comes from the github api (couldn't find the exact location). we are indeed not sure, but if it changes we will know right away since this code will stop functioning. can you think of a better approach?

rporres: I was thinking if we actually need to check for the `too_large`, but since you're raising anyway I guess we will soon know.

I've found tons of mentions to the `too_large` string, I don't think it will easily change.

rporres (approved):

PR #1370, opened 2021-02-07
[sql-query] support password special chars

maorfr:
by adding quotes when assigning the password.

this is an attempt to fix https://issues.redhat.com/browse/APPSRE-2982

### DIFF
--- a/reconcile/sql_query.py
+++ b/reconcile/sql_query.py
@@ -270,12 +270,12 @@ def make_postgres_command(output, sqlqueries, recipient=None):
     command = []
     for query in sqlqueries:
         command.extend([
-            '(PGPASSWORD="$(db.password)"',
-            'time psql',
+            '(time PGPASSWORD="$(db.password)"',
+            'psql',
             '--host=$(db.host)',
             '--port=$(db.port)',
             '--username=$(db.user)',
-            '--dbname=$(db.name)'
+            '--dbname=$(db.name)',
             f'--command "{query}")',
         ])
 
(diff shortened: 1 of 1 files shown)


apahim (approved):
lgtm, thanks!

reconcile/sql_query.py
```diff
@@ -270,9 +270,12 @@ def make_postgres_command(output, sqlqueries, recipient=None):
...
-            '(time psql',
-            'postgres://$(db.user):$(db.password)@'
-            '$(db.host):$(db.port)/$(db.name)',
+            '(PGPASSWORD="$(db.password)"',
+            'time psql',
+            '--host=$(db.host)',
+            '--port=$(db.port)',
+            '--username=$(db.user)',
+            '--dbname=$(db.name)'
             f'--command "{query}")',
```
Red-HAP, line 279: @maorfr 
You have a risk here with "{query}" as well.

If the query itself contains double-quote chars, you could encounter other bash errors. There should be a more secure way to do this.

Ex:
```sql
select 1 as "cool_column"
  from my_table as "eek";
```

maorfr: thanks for raising this! this is not an issue, as the query returns escaped from the graphql server. example:
```
      {
        "name": "2021-01-19-compliance-prod-check-profile-policies",
        "query": "SELECT \"version\" FROM \"schema_migrations\" ORDER BY \"version\" DESC LIMIT 10;"
      }
```

maorfr: thanks for raising this! this is not an issue, as the query returns escaped from the graphql server. example:
```
      {
        "query": "SELECT \"version\" FROM \"schema_migrations\" ORDER BY \"version\" DESC LIMIT 10;"
      }
```

PR #1372, opened 2021-02-08
[ocm-clusters] allow to change cluster channel group

maorfr:
following https://issues.redhat.com/browse/SDA-3297, we can now update cluster channel through the API.

the API will soon change to PATCH `version.channel_group` instead of `upgrade_channel_group`, but that's WIP


apahim (approved):
lgtm

PR #1373, opened 2021-02-08
move self-implemented caches to functools.lru_cache

erdii:
This issue implements the changes from: https://issues.redhat.com/browse/APPSRE-2866
We figured, that this pattern:

```python
class Foo:
  __init__(self):
    self._cache = {}

  get_something(self):
    cache_key = 'something'
    cached = self._cache.get(cache_key)
    if cached is not None:
      return cached
    fresh = do_expensive_operation_to_get_something()
    self._cache[cache_key] = fresh
```

can be expressed in a simpler way using [@functools.lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache):

```python
import functools

class Foo:
  @functools.lru_cache()
  get_something(self):
    return do_expensive_operation_to_get_something()

  # added to explain further usage:

  @functools.lru_cache()
  get_something_by_id(self, id):
    # results are even cached based on function argument values
    return do_expensive_operation_to_get_something_by_id(id)

  update_something(self):
    do_something_that_makes_cached_data_from_get_something_stale()
    # remove all data from `get_something`'s cache
    self.get_something.cache_clear()
```

cc @apahim

### DIFF
--- a/reconcile/utils/vault.py
+++ b/reconcile/utils/vault.py
@@ -1,8 +1,8 @@
-import time
-import requests
-import hvac
 import base64
-import logging
+import functools
+import hvac
+import requests
+import time
 
 from hvac.exceptions import InvalidPath
 from requests.adapters import HTTPAdapter
@@ -55,7 +55,6 @@ class _VaultClient:
                               pool_maxsize=100)
         session.mount('https://', adapter)
         self._client = hvac.Client(url=server, session=session)
-        self._cache = {}
 
         authenticated = False
         for i in range(0, 3):
@@ -81,7 +80,7 @@ class _VaultClient:
         secret_path = secret['path']
         secret_version = secret.get('version')
 
-        kv_version = self._get_mount_version(secret_path)
+        kv_version = self._get_mount_version_by_secret_path(secret_path)
 
         data = None
         if kv_version == 2:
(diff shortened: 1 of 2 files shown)


reconcile/utils/vault.py
```diff
@@ -211,6 +201,7 @@ def write(self, secret):
             self._write_v2(secret_path, data)
 
     def _write_v2(self, path, data):
+        # do not forget to run `self._read_all_v2.cache_clear()` if this ever get's implemented
         raise NotImplementedError('vault_client write v2')
 
     def _write_v1(self, path, data):
```
apahim, line 207: should we `cache_clear` here?

reconcile/utils/vault.py
```diff
@@ -81,7 +81,7 @@ def read_all(self, secret):
         secret_path = secret['path']
         secret_version = secret.get('version')
 
-        kv_version = self._get_mount_version(secret_path)
+        kv_version = self._get_mount_version_by_secret_path(secret_path)
```
apahim, line 84: that makes sense :)

apahim (approved):
Will the cache size ever be a problem? We run this in a deamon-like execution, but not every loop will create a new instance of the SentryClient. The VaultClient instance is reused a lot... we only create a new one when the `is_authenticated()` return false at the beginning of the main loop. Just giving food for thought.

reconcile/utils/vault.py
erdii: if we cache the result of `_read_all_v2` we should propably also invalidate this if we touch a secret in a v2 engine

erdii:
I don't think it will be a problem for these reasons:
The default lru_cache size is 128 keys per decorated function and it will automatically evict keys if it overflows.
Currently the only decorated functions do not take argument to parametrize the query and only cache one big collection result.

reconcile/utils/vault.py
```diff
@@ -94,32 +93,23 @@ def read_all(self, secret):
...
         try:
             self._client.secrets.kv.v2.read_configuration(mount_point)
             version = 2
         except Exception:
             version = 1
 
-        self._cache[cache_key] = version
         return version
 
+    @functools.lru_cache()
```
erdii, line 111: @jmelis mentioned that this may be should be an unbound cache

erdii: pr: https://github.com/app-sre/qontract-reconcile/pull/1378

PR #1378, opened 2021-02-10
make vault read_all_v2 cache unbound to restore previous caching beha…

erdii:
…vior

this is a follow up to https://github.com/app-sre/qontract-reconcile/pull/1373/files#discussion_r573650034
and restores previous unbound caching behavior on the _read_all_v2 function


apahim (approved):
lgtm

PR #1380, opened 2021-02-11
sendgrid-teammates: add logic to delete users

jmelis:


reconcile/sendgrid_teammates.py
```diff
@@ -47,28 +55,52 @@ def fetch_current_state(sg_client):
...
             LOG.info(['delete', user.email])
+            if not dry_run:
+                if user.pending:
+                    delete_method = sg_client.teammates.pending
+                    identifier = user.pending_token
+                else:
+                    delete_method = sg_client.teammates
+                    identifier = user.email
+
+                response = delete_method._(identifier).delete()
```
jmelis, line 96: This underscore thing is tricky. I kind of inferred it from here: https://github.com/sendgrid/sendgrid-python#general-v3-web-api-usage-without-fluent-interface

reconcile/sendgrid_teammates.py
```diff
@@ -47,28 +55,52 @@ def fetch_current_state(sg_client):
...
         state.append(t)
 
     return state
 
 
+def raise_if_error(response):
+    """
+    Raises an SendGridAPIError if the request has returned an error
+    """
+    if int(response.status_code / 100) != 2:
```
apahim, line 68: `if response.status_code >= 300:` is 3x faster :)

reconcile/sendgrid_teammates.py
apahim, line 96: that's interesting... first time I see the `_` used that way.

jmelis: me too, quite ingenious, though

apahim (approved):
lgtm

reconcile/sendgrid_teammates.py
jmelis: good idea, yes :D

PR #1381, opened 2021-02-11
Fix diff check

apahim:


maorfr (approved):
LGTM

so i was right in https://github.com/app-sre/qontract-reconcile/pull/1327#discussion_r560912144?

apahim:
Yes! The check was there and you did find a bug :)

PR #1382, opened 2021-02-11
Added mechanism to avoid reconcile metrics publish

rporres:
This implements a way to avoid certain reconciliation to be published by means of the `qontract.ignore_reconcile_time` annotation.

A few comments

* ~The general approach works, but this code is untested. I wanted to check before testing if there were red flags about the approach as testing this properly is not straight forward and will require some time.~ Now this has been properly tested
* I've moved all the code related to the reconcile time to a decorator in the same module as the `OC` class as it was too specific to be consumed anyywhere else
* I've moved JSON serialization to `OC` as I needed to deserialize again to inspect the object and it didn't make sense. Since the consumers of that class is us, it should be fine
~* Everything the decorator needs is passed in the return which is stripped as every function that modifies stuff in the class does not return anything~


reconcile/utils/oc.py
```diff
@@ -63,6 +64,55 @@ class JobNotRunningError(Exception):
 
 
 class OC:
+    class _Decorators:
+        @classmethod
+        def process_reconcile_time(cls, function):
```
apahim, line 69: Not a big fan of the nested class, but I lack technical arguments against it. Anyway, why not have the function in the module root?

reconcile/utils/oc.py
```diff
@@ -63,6 +64,55 @@ class JobNotRunningError(Exception):
...
+                    kind = result['resource']['kind']
+                    name = result['resource']['metadata']['name']
+                    annotations = \
+                        result['resource']['metadata'].get('annotations')
+                except KeyError as e:
+                    logging.warning(f"Error processing metric: {e}")
+                    return
+
+                ignore_reconcile_time = \
+                    annotations.get('qontract.ignore_reconcile_time') == 'true'
```
apahim, line 89: `annotations` can be None.

reconcile/utils/oc.py
```diff
@@ -63,6 +64,55 @@ class JobNotRunningError(Exception):
...
+                if not ignore_reconcile_time:
+                    name = f'{function.__module__}.{function.__qualname__}'
+                    reconcile_time.labels(
+                        name=name,
+                        integration=running_state.integration
+                    ).observe(amount=time_spent)
+
+                if not result['is_log_slow_oc_reconcile']:
```
apahim, line 97: `is_log_slow_oc_reconcile` might not be there.

reconcile/utils/oc.py
```diff
@@ -63,6 +64,55 @@ class JobNotRunningError(Exception):
...
+                    return
+
+                if time_spent > result['slow_oc_reconcile_threshold']:
```
apahim, line 100: `slow_oc_reconcile_threshold` might not be there.

reconcile/utils/oc.py
```diff
@@ -63,6 +64,55 @@ class JobNotRunningError(Exception):
...
+                          f"reconcile. Commit sha {running_state.commit} " \
+                          f"and commit ts {running_state.timestamp}."
+
+                    if ignore_reconcile_time:
+                        msg += " Ignored in the metric published."
+
+                    logging.info(msg)
+
+                return
+
```
apahim, line 113: so, my comments here are around making a safe decorator. there are 2 ways to approach that: testing each argument (with try/except or with get()/if). Or, the way I preffer in this case, by having the `function(*args, **kwargs)` returning a custom class instance trusted to have all you need, then you don't have to check any arguments in advance, but you do have to check if the function returned the object you need.

apahim: I ok, with both, just wanted to raise the topic.

rporres: So since everything comes from `_msg_to_process_reconcile_time` I was trusting I had everything I can. But I understand where you come from and it makes sense. If we want to make this safe I like the custom class approach better.

reconcile/utils/oc.py
rporres: Hmmm, since this was so related to this class I thought it was a good idea. I don't have any preference for it to be inside the class or in the module root. I do agree that it is not pretty :-D

reconcile/utils/oc.py
rporres: indeed, I somehow thought this was going to work testing the wrong thing.

maorfr:
added the proper annotation to all resources in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/14991

reconcile/utils/oc.py
rporres: Fixed in 69576f7

reconcile/utils/oc.py
rporres: Fixed in 69576f7

reconcile/utils/oc.py
rporres: Fixed in 69576f7

reconcile/utils/oc.py
rporres: Fixed in 69576f7

rporres:
Passed message as an object in the decorator in 69576f7. While testing this I realised we weren't adding namespace reconciliations to our SLO, so I added them too.

reconcile/utils/oc.py
```diff
@@ -178,30 +241,46 @@ def remove_last_applied_configuration(self, namespace, kind, name):
...
-    @elapsed_seconds_from_commit_metric
+    @OCDecorators.process_reconcile_time
     def replace(self, namespace, resource):
-        self._log_slow_oc_reconcile(namespace, resource)
+        """ Runs oc replace. Returns nothing """
         cmd = ['replace', '-n', namespace, '-f', '-']
-        self._run(cmd, stdin=resource, apply=True)
+        self._run(cmd, stdin=resource.toJSON(), apply=True)
+        # This return will be removed by the last decorator
+        return self._msg_to_process_reconcile_time(namespace, resource.body)
```
apahim, line 274: whant heppens if a function returns something that should go through + the message to the decorator? example:
```python
    return actual_return, OCDecoratorMsg(...)
```
I guess what I'm trying to say is: the decorator should be totally transparent. even the message to the decorator should go through it and be present on the other side, but I'd be fine with not having the message going through. In any case, actual returns have to.

rporres: the thing is that no modifying function in this class actually returns anything. I was trying to preserve that behaviour as much as possible. It looked pretty weird to me just return the decorator message but I must admit this is my first time dealing with decorators in Python so I lack the sense of best practices.

What I can do is making the decorator check for the last element of the return value and if it is an instance of the `OCDecoratorMsg` do stuff. I will return the contents of the call unmodified as in this case we don't really care, no one checks it.

I guess what you want as decorator is a function that can be consumed everywhere even by mistake whereas I'm thinking about a function that I know where it is called from. It doesn't hurt to be extra defensive, I guess :)

apahim: What I'm thinking is that we can, at some point, intentionally use the decorator t odecorate a function that returns something. And, even then, both the decorator and the function should keep working as expected.

rporres: I buy your argument ;-)

apahim: To be very accurate, even these functions you're decorating, they return something. They return `None`. Your decorator also returns `None` when there's a message to the decorator, so this case is covered.

My concern is only when we try to decorate a function that has already an explicit return AND we want to track its reconcile time, meaning we have to add the message to the decorator in the existing explicit return.

Does that make sense?

apahim: cool cool :)

rporres: Fixed in https://github.com/app-sre/qontract-reconcile/pull/1382/commits/8204c5c61390d5fe750c2f576631eab860d06b64. This is the more generic I can imagine this 😂

reconcile/utils/oc.py
```diff
@@ -62,6 +63,90 @@ class JobNotRunningError(Exception):
...
+            ignore_reconcile_time = \
+                annotations.get('qontract.ignore_reconcile_time') == 'true'
+            if not ignore_reconcile_time:
+                reconcile_time.labels(
+                    name=function_name,
+                    integration=running_state.integration
+                ).observe(amount=time_spent)
+
+            if not msg.is_log_slow_oc_reconcile:
+                return
```
apahim, line 120: `return result` ?

reconcile/utils/oc.py
```diff
@@ -62,6 +63,90 @@ class JobNotRunningError(Exception):
...
+                          f'{msg.server} took {time_spent} to ' \
+                          f'reconcile. Commit sha {running_state.commit} ' \
+                          f'and commit ts {running_state.timestamp}.'
+
+                if ignore_reconcile_time:
+                    log_msg += ' Ignored in the metric published.'
+
+                logging.info(log_msg)
+
+            return result
```
apahim, line 135: So, the message object is going through.

reconcile/utils/oc.py
apahim: It's nice. And, most importantly, it works :)

reconcile/utils/oc.py
rporres: Yes, we don't really care as it is not being taken into account and it makes things less magical

reconcile/utils/oc.py
rporres: FML. Fixed in 826bfb5

apahim (approved):
lgtm

rporres:
Thx for the help on this, @apahim ;-)

PR #1385, opened 2021-02-14
[saas-file-owners] only add comments when MR is rebased

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3029

we are seeing issues of non-rebased MRs thinking they change a saas file that was recently merged.
to avoid that, we can just not comment on the MR if it is not rebased (while keeping all other behaviors).

### DIFF
--- a/reconcile/utils/gitlab_api.py
+++ b/reconcile/utils/gitlab_api.py
@@ -296,6 +296,15 @@ class GitLabApi(object):
                              'id': note.id})
         return comments
 
+    def is_merge_request_rebased(self, mr_id):
+        merge_request = self.project.mergerequests.get(mr_id)
+        target_branch = merge_request.target_branch
+        head = self.project.commits.list(ref_name=target_branch)[0].id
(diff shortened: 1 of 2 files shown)


reconcile/utils/gitlab_api.py
```diff
@@ -296,6 +296,15 @@ def get_merge_request_comments(self, mr_id):
                              'id': note.id})
         return comments
 
+    def is_merge_request_rebased(self, mr_id):
+        merge_request = self.project.mergerequests.get(mr_id)
+        target_branch = merge_request.target_branch
+        head = self.project.commits.list(ref_name=target_branch)[0].id
+        result = self.project.repository_compare(merge_request.sha, head)
+        if len(result['commits']) == 0:  # rebased
```
apahim, line 304: `return len(result['commits']) == 0` or `return not result['commits']`

apahim (approved):
lgtm (with a small suggestion)

maorfr:
replaced by #1404

PR #1386, opened 2021-02-15
[sentry-config] read all secrets before execution

maorfr:
sentry-config fails often for being forbidden to access secrets in vault.
this is likely due to the sequential execution of the integration.

there are 3 options to handle this:
1. extend token lifetime
2. read all secrets before execution
3. parallelize the execution

i chose 2 to not change the token lifetime (undesired) and to not change the resource consumption of the integration.
if we ever choose to measure SLOs for sentry-config, we'll probably want to go with 3.


apahim (approved):
lgtm

PR #1392, opened 2021-02-16
[openshift-namespaces] use explicit kind

maorfr:
towards using the native kube client we have an ambiguity over `Project`. this should solve it.


rrati (approved):
lgtm

rporres (approved):

PR #1393, opened 2021-02-16
[APPSRE-3917] pylint assignment-from-no-return

rporres:
[`ts.populate_route53`](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/terrascript_client.py#L370-L421) has no explicit return, so no need to check for it.


rporres:
cc @jfchevrette

apahim (approved):
muchos pocos hacen un mucho

PR #1395, opened 2021-02-16
fix recycle_pods

BumbleFeng:
OC.apply can only deal with unserialized obj due to https://github.com/app-sre/qontract-reconcile/pull/1382

cc @rporres 


rporres:
Thx, @BumbleFeng . I didn't get this one while testing this

PR #1396, opened 2021-02-17
Fix pylint wrong-import-order

apahim:
Correct order is:

```python
import bultin
from builtin import function

import external
from external import function

import local
from local import function
```


rporres (approved):

rporres:
If you're doing PyCharm order (which I think is totally fine) I would enforce it via https://github.com/PyCQA/flake8-import-order too

apahim:
@rporres  I reverted to the previous version, with the minimal pylint fixes. I looked into it and it feels wrong to enforce an style beyond pep8. Feel free to withdraw or confrim your approval, I won't merge before you say something :)

rporres (approved):
I agree on not imposing a certain order out of PEP8 unless we find the experience frustrating.

PR #1399, opened 2021-02-18
Fix pylint subprocess-run-check

jfchevrette:
- Adds `check=True` everywhere we use `subprocess.run()`
- Refactors most usage of `subprocess.run()` to wrap them into try/catch and catch `subprocess.CalledProcessError` to handle raising errors from stdout/stderr instead of via the result of `subprocess.run()`

Ref.: APPSRE-3017


apahim (approved):
Thank you for this!
Not sure I understand the refactoring though. I mean, `check=False` would be enough to fix the pylint error and keep the current functionality.
Anyway, it looks safe, just wanted to understand.

jfchevrette:
I first did `check=False` to keep the same behavior but I thought: are we enabling pylint to merely "silence" the checks to make pylint happy or to improve the code?

I chose to use the opportunity to improve the code. Do you think that having it raise an exception and catching it is not better? In my view it is....

apahim:
The main reason to make pylint happy is to improve the code. Sure, more improvements are always welcome.

> Do you think that having it raise an exception and catching it is not better?

No, I don't think it's always the case. Thank you for the rationale. I though I was missing some hidden issue. Ship it!

apahim:
Don't take me wrong... I think it's a good thing here. My concern is because we will have to revisit your changes when we fix the `broad-except` error, but I guess the effort won't be that bigger then.

rporres:
Late to the party, but this makes total sense. Thx, @jfchevrette 🥇

PR #1400, opened 2021-02-19
Use native openshift client for get requests

rrati:


rrati:
Changes in app-interface are needed before merging as well as auth for proxy.

apahim:
Hi @rrati, thank you for this! I guess we all love the idea of a native openshift client and the inital results are promissing!

I wonder if we should have a separate client for that? Hooking into the OC client seems wrong in terms of design... wdyt?

rrati:
Why would we want to re-write a native client when one already exists for this purpose?  Or are you meaning something else?

apahim:
I mean, the class in which we implement this should not be `OC()`, but a new one.

rrati:
The OC class wraps the calls to the OC binary.  My intention/desire is that we get rid of all call outs to OC and replace them with native api calls.  This is a step towards that.  To that end, the OC class would be that wrapper around the openshift dependency.  Does that roadmap meet your concern?  What would be missing?

rporres:
I'd also prefer to have a separate `OC` class so we could switch between one implementation or the other from the outside, e.g. we could separate if you're using the native vs the binary for private clusters. It's also very interesting to have an implementation that we know that it works that we could switch to quickly if we have a problem. At least until we're confident to remove it.

apahim:
You will end-up with an `OC()` class that uses oc binary for some things and api calls for other things. That mix is what I find hard to maintain and to evolve.
On the other hand, if you have an `OC()` that implements `get()` using `oc get` and you have another class, say `Kli()`, that implements `get()` using native api calls, you can keep migrating functionalities from `OC()` to `Kli()` more organically, with easy fallback path.

rrati:
The end goal here would be the elimination of usage of the oc binary entirely.  This is the first step, but once it has been proven we should move all other openshift operations to native api.  We should not aim to maintain oc and native client imo.  That introduces 2 dependencies to keep in line with supported openshift versions.

rrati:
Adding another class complicates the code because now you have to know to transition between 1 class for gets and another for everything else.

IMO unless we need to take on that complexity it isn't worth the effort.  If we can't move everything to native api client then it makes sense to head down this more complex path.

rporres:
It actually makes the code much clearer as you know well how things are going to work. You want native, you consume one class. You want binary fallback, you consume the other. You can be much smarter in the way you handle common operations, but since this is very critical we need a hard and clear separation between both approaches while we transition from one model to the other to be able to rollback easily, to be able to do a smooth rollout in all of our clusters instead of an all-or-nothing model.

apahim:
You can either:
```python
class OCNative(OC):
    def get(self):
        # using the api
```
changing the callers of `OC` to use the new name already, or:

```python
class _OC():  # renamed from OC
    ...

class OC(_OC):  # new OC, using API, to be renamed later
    def get(self):
        # using the api
```

In any case, you're overriding the methods with the new implementation. Callers won't have to know about the two of them.

apahim:
@rporres is taking a more extreme approach there. I don't find it critical to have the callers handling the two classes. In fact I'd argue against that. But I find it critical to have the the implementation done in a separate class and use inheritance to make the migration. Migration as well as eventual rollbacks and future maintenance will be way easier.

rporres:
I don't mind if we don't want the callers decide as long as we can rollout / rollback this easily, ideally via configuration/feature flags, etc...

rrati:
IMO the rollback here is to revert the PR.  I don't think we should add complexity around a solution we wish to get rid of.  If api works, that's the thing we want to use.  If not then roll back the PR.

rrati:
@apahim IIUC, what you are suggesting will still result in a) lots of copied code and/or b) a need to migrate code from the old class to the new.  There's a lot of functionality in OC that is based upon a few actual calls to openshift.  If the native api transition completes, then the old OC class would go away and all those functions would have to migrate to the new native api class.  That's not terrible, but it's amazing what problems can be introduced just by moving code around.

rrati:
I updated to use inheritance, but imo much more than this overcomplicates this transition.  We should be aiming for minimal changes

rporres:
Rollback by reverting the PR assumes that you start seeing problems soon enough so that you can do a clean revert which it's a big assumption. Since we don't have a good mechanism to test in staging, we will need to test in production in a small subset of our clusters by means of configuration to gain confidence. This is not a trivial change and should be handled with lots of care.

rrati:
We can deploy to staging and verify it is detecting the changes correctly.  We can watch prod and if there's an error, fix or roll back.

We don't currently have the mechanism you are wanting and a core principal for this kind of change is to keep it as small as possible to reduce the number of moving parts.  Adding the control creates 2 new pieces that could produce failures.  I would rather do what you want piecemeal.  As in leave this alone as is, merge watch staging.  Implement a control mechanism, deploy to stage and watch.  Then deploy all to prod and slowly scale up clusters.

BumbleFeng:
We can test it on https://github.com/app-sre/app-interface/tree/openshift

maorfr:
we can test it by `make rc` and submit a MR to app-interface without merging. this will provide output of what all integrations will do.

reconcile/utils/oc.py
```diff
@@ -620,6 +628,177 @@ def _run_json(self, cmd, allow_not_found=False):
...
+
+    def _get_client(self, server, token):
+        opts = dict(
+            api_key={'authorization': f'Bearer {token}'},
+            host=server,
+            verify_ssl=True
+        )
+
+        if self.jump_host is not None:
+            opts['proxy'] = f'https://{self.jump_host.get_hostname()}:8888'
```
apahim, line 655: This is not something already in place, right? I mean, I can't test this now?

rrati: It is and it isn't. :) I had a proxy setup where access was controlled via allow clauses, and it was configured to work from my laptop.  That won't scale or manage well, so I am working on changing the proxy to use client certs for authentication.

rrati: @apahim This should now be fully testable now.

reconcile/utils/jump_host.py
```diff
@@ -71,3 +77,19 @@ def get_ssh_base_cmd(self):
...
+                ssh_address_or_host=self.hostname,
+                ssh_port=self.port,
+                ssh_username=self.user,
+                ssh_pkey=self.identity_file,
+                remote_bind_address=(self.hostname, remote_port),
+                local_bind_address=('localhost', local_port)
+            )
+            tunnel.start()
+            JumpHostSSH.bastion_tunnel[key] = tunnel
+        JumpHostSSH.tunnel_lock.release()
```
rrati, line 95: @apahim I would like to use defer here instead, but my attempts to use it failed.  Should just be:

@defer
def create_ssh_tunnel(...):
...
acquire()
defer(JumpHostSSH.tunnel_lock.release())

Right?

apahim: `defer()` takes a callable, not an object instance

apahim: `defer(JumpHostSSH.tunnel_lock.release)` should do it

rrati: Hrm, using defer results in about a 30+s decrease in time to complete.  I'd prefer to use it, but for now I will leave it as is since a goal of this is to improve performance.  Giving back 30+s is a pretty big chunk.

apahim: that doesn't sound right :)

apahim: But agreed, adding 30s is not acceptable, specially when we are fighting for faster executions.

rrati: I was surprised by the degradation too.  I ran it twice, and the 2nd time was a little better but still about 30s slower than without:
1st run:
real	1m56.568s
user	1m6.170s
sys	0m11.752s

2nd run:
real	1m46.512s
user	1m5.351s
sys	0m11.312s

current (no defer):
real	1m15.839s
user	1m4.597s
sys	0m11.181s

rrati:
This is break openshift-saas-deploy as is currently implemented.

rrati:
saas-deploy is fixed

reconcile/utils/oc.py
```diff
@@ -147,23 +154,28 @@ def __init__(self, namespace, resource, server,
...
                  local=False):
         self.server = server
         oc_base_cmd = [
             'oc',
-            '--kubeconfig', '/dev/null',
-            '--server', server,
-            '--token', token
+            '--kubeconfig', '/dev/null'
         ]
+        if server is not None and len(server) > 0:
```
Piojo, line 166: I find `if server:` shorter and easier to read. In English, to me it reads as "if there is a server".

reconcile/utils/oc.py
```diff
@@ -147,23 +154,28 @@ def __init__(self, namespace, resource, server,
...
+            oc_base_cmd.extend(['--server', server])
+
+        if token is not None and len(token) > 0:
```
Piojo, line 169: Ditto, I tend to prefer `if token:`.

rrati:
Time comparison for 1 cluster running `APP_INTERFACE_STATE_BUCKET=app-interface-production APP_INTERFACE_STATE_BUCKET_ACCOUNT=app-sre qontract-reconcile --config ./config.debug.toml --dry-run openshift-resources --cluster-name app-sre-stage-01`:

With this change:
22.24s user 3.60s system 57% cpu 44.561 total

Current code:
70.22s user 51.89s system 147% cpu 1:22.62 total

reconcile/utils/jump_host.py
```diff
@@ -71,3 +78,25 @@ def get_ssh_base_cmd(self):
             '-o', 'StrictHostKeyChecking=yes',
             '-o', 'UserKnownHostsFile={}'.format(self.known_hosts_file),
             '-i', self.identity_file, '-p', str(self.port), user_host]
+
+    def create_ssh_tunnel(self, local_port, remote_port):
+        key = f'{self.hostname}-{local_port}-{remote_port}'
+        JumpHostSSH.tunnel_lock.acquire()
```
jmelis, line 84: nit: why not `self.tunnel_lock....` ?

reconcile/utils/oc.py
```diff
@@ -636,6 +649,210 @@ def _run_json(self, cmd, allow_not_found=False):
...
+        k, group_version = self._parse_kind(kind)
+        obj_client = self._get_obj_client(
+            group_version=group_version, kind=k)
+        try:
+            return obj_client.get().to_dict()
+        except NotFoundError as e:
+            raise StatusCodeError(f"[{self.server}]: {e}")
+
+    @staticmethod
+    def add_group_kind(kind, kgv, new, preferred):
```
jmelis, line 832: I'm not sure I understand this. The api-resources is cluster-specific, but we are storing it as a class variable.

reconcile/utils/oc.py
jmelis: +1 to @Piojo's suggestion.

@rrati I prefer changing the default to `None` as you did, but is there a reason for this? or are you fixing this purely because of style and legibility?

reconcile/utils/jump_host.py
rrati: Both the locks and the tunnel info needs to exist across multiple different instantiations of jump host to avoid tunnels being opened simultaneously or to hosts that already have a tunnel.

reconcile/dashdotdb_cso.py
```diff
@@ -47,7 +47,7 @@ def _post(self, manifest):
 
         for item in imagemanifestvuln['items']:
             endpoint = (f'{self.dashdotdb_url}/api/v1/'
-                        f'imagemanifestvuln/{cluster}')
+                        f'ImageManifestVuln/{cluster}')
```
jmelis, line 50: @rrati can you review this one? It seems to talk to dashdotdb and not to the cluster

reconcile/utils/jump_host.py
jmelis: :facepalm: totally

jmelis, line 84: use `with`

jmelis: https://docs.python.org/3/library/threading.html#using-locks-conditions-and-semaphores-in-the-with-statement

reconcile/dashdotdb_cso.py
rrati: I think this change is incorrect.  It doesn't get executed in the dry-run case, so setting back to original code.

reconcile/utils/jump_host.py
rrati: Done

reconcile/utils/oc.py
rrati: Namely fixing for legibility and consistency with error handling for other empty things.

rrati: Make change suggested by @Piojo

reconcile/utils/oc.py
rrati: Done

reconcile/utils/oc.py
```diff
@@ -636,6 +650,252 @@ def _run_json(self, cmd, allow_not_found=False):
...
+        # above in the format of var_name:value then set them here
+        # in the configuration object with setattr.
+        for k, v in opts.items():
+            setattr(configuration, k, v)
+
+        if self.jump_host is not None:
+            # the ports could be parameterized, but at this point
+            # we only have need of 1 tunnel for 1 service
+            self.jump_host.create_ssh_tunnel(8888, 8888)
+            opts['proxy'] = 'https://localhost:8888'
```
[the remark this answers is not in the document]
rrati: This has been fixed.

reconcile/utils/jump_host.py
```diff
@@ -71,3 +78,24 @@ def get_ssh_base_cmd(self):
...
+
+                tunnel = sshtunnel.SSHTunnelForwarder(
+                    ssh_address_or_host=self.hostname,
+                    ssh_port=self.port,
+                    ssh_username=self.user,
+                    ssh_pkey=self.identity_file,
+                    remote_bind_address=(self.hostname, remote_port),
+                    local_bind_address=('localhost', local_port)
+                )
+                tunnel.start()
```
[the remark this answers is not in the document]
rrati: Done

reconcile/utils/oc.py
```diff
@@ -636,6 +650,252 @@ def _run_json(self, cmd, allow_not_found=False):
...
+        # client via instance variables.  We define a set of options
+        # above in the format of var_name:value then set them here
+        # in the configuration object with setattr.
+        for k, v in opts.items():
+            setattr(configuration, k, v)
+
+        if self.jump_host is not None:
+            # the ports could be parameterized, but at this point
+            # we only have need of 1 tunnel for 1 service
+            self.jump_host.create_ssh_tunnel(8888, 8888)
```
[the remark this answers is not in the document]
rrati: I agree hardcoding the port is not the right way long term.  ATM the jump_host definition doesn't have a remote (or local) port configuration.  We could probably add config options.  I favor adding the information to the jump host config in app-interface.

Can we do this as a later enhancement once this is merged?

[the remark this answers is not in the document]
BumbleFeng: works for me.

reconcile/utils/oc.py
```diff
@@ -668,6 +682,252 @@ def _run_json(self, cmd, allow_not_found=False):
...
 
 
+class OCNative(OCDeprecated):
+    def __init__(self, server, token, jh=None, settings=None,
+                 init_projects=False, init_api_resources=False,
+                 local=False):
+        super().__init__(server, token=token, jh=jh, settings=settings,
+                         init_projects=False, init_api_resources=False,
+                         local=local)
+        if server is not None and len(server) > 0:
```
jmelis, line 692: nit: `if server:`

reconcile/utils/oc.py
```diff
@@ -668,6 +682,252 @@ def _run_json(self, cmd, allow_not_found=False):
...
+            init_api_resources = False
+            init_projects = False
+
+        self.object_clients = {}
+        self.init_projects = init_projects
+        if self.init_projects:
+            self.projects = \
+                [p['metadata']['name']
+                 for p
+                 in self.get_all('Project.project.openshift.io')['items']]
```
jmelis, line 705: do you know if this works in OpenShift v3?

reconcile/utils/oc.py
```diff
@@ -668,6 +682,252 @@ def _run_json(self, cmd, allow_not_found=False):
...
+        opts = dict(
+            api_key={'authorization': f'Bearer {token}'},
+            host=server,
+            verify_ssl=False,
+            # default timeout seems to be 1+ minutes and it can't
+            # be changed, so do no retries on connection failure
+            retries=0
+        )
+
+        if self.jump_host is not None:
```
jmelis, line 722: nit: `if self.jump_host:`

reconcile/utils/oc.py
```diff
@@ -668,6 +682,252 @@ def _run_json(self, cmd, allow_not_found=False):
...
+        kind_group = kind_name.split('.', 1)
+        kind = kind_group[0]
+        if kind in self.api_kind_version:
+            group_version = self.api_kind_version[kind][0]
+        else:
+            raise StatusCodeError(f'{self.server}: {kind} does not exist')
+
+        # if a kind_group has more than 1 entry than the kind_name is in
+        # the format kind.apigroup.  Find the apigroup/version that matches
+        # the apigroup passed with the kind_name
```
jmelis, line 762: what if we don't find it with the algorithm below? will we send a relevant error message?

reconcile/utils/oc.py
```diff
@@ -668,6 +682,252 @@ def _run_json(self, cmd, allow_not_found=False):
...
+            apigroup_override = kind_group[1]
+            for gv in self.api_kind_version[kind]:
+                if apigroup_override in gv:
+                    group_version = gv
+                    break
+        return(kind, group_version)
+
+    # this function returns a kind:apigroup/version map for each kind on the
+    # cluster
+    def get_api_resources(self):
```
jmelis, line 773: We agreed that the native client would not require tests because we agreed on that once this code was already in place, but the logic of this function is particularly hard to follow. Could I convince you to do some quick unit tests for this method? I believe you would need to mock `client.resources.parse_api_groups` and `client_resources.get_resources_for_api_version` with several values and ensure that ensure that the kind_groupversion return is the one expected.

reconcile/utils/oc.py
```diff
@@ -668,6 +682,252 @@ def _run_json(self, cmd, allow_not_found=False):
...
+        k, group_version = self._parse_kind(kind)
+        obj_client = self._get_obj_client(group_version=group_version, kind=k)
+
+        namespace = ''
+        if 'namespace' in kwargs:
+            namespace = kwargs['namespace']
+            # for cluster scoped integrations
+            # currently only openshift-clusterrolebindings
+            if namespace != 'cluster':
+                if not self.project_exists(namespace):
```
jmelis, line 823: weird, why do we return empty list if the project does not exist?

reconcile/utils/oc.py
```diff
@@ -668,6 +682,252 @@ def _run_json(self, cmd, allow_not_found=False):
...
+                            documentation='Cluster is using openshift '
+                            'native client',
+                            labelnames=['cluster_name', 'native_client'])
+
+    def __new__(cls, server, token, jh=None, settings=None,
+                init_projects=False, init_api_resources=False,
+                local=False, cluster_name=None):
+        use_native = os.environ.get('USE_NATIVE_CLIENT', False)
+        if use_native:
+            if cluster_name:
```
jmelis, line 916: can you remind me in what case the `cluster_name` is expected to be None?

reconcile/utils/oc.py
rrati: This should be mimicking the existing get_items functionality.  I'm not sure why this is done, but if this is not the same return as the oc client then it should be fixed.  I think this is correct though.

reconcile/utils/oc.py
rrati: Actually, no we won't.  If we don't find the kind.apigroup then we will return the group_version that is known to the cluster.  For example:

If kind_name = whosit.whatsit
and api_kind_version contains:
{'whosit':["clusterapigroup"]}
then we would return
kind = whosit
group_version = clusterapigroup

This effectively means that the overrided apigroup doesn't exist on the cluster.  Since we are trying to override explicitly, we probably should throw an exception here.

reconcile/utils/oc.py
rrati: I'm not sure.  I've run this against all clusters in app-interface (ie not specified specific cluster) and it hasn't thrown any errors, but I haven't made changes to anything that would apply to a v3 cluster to test it reads everything correctly.  I should do that.

reconcile/utils/oc.py
rrati: IN cases like saas deploy where OC is created just to use the process to generate a template.

reconcile/utils/oc.py
BumbleFeng: Done

reconcile/utils/oc.py
BumbleFeng: tested and it works

reconcile/utils/oc.py
```diff
@@ -890,7 +869,7 @@ def add_group_kind(kind, kgv, new, preferred):
             group = new.split('/', 1)[0]
             new_group = True
             for pos in range(len(kgv[kind])):
-                if group in kgv[kind][pos]:
```
rrati: Why would we do this?  The group should be explicit.

Deployment != DeploymenConfig for example.

reconcile/test/fixtures/oc_native/api.yml
```diff
@@ -0,0 +1,98 @@
...
+        version: v1alpha1
+    - name: managed.openshift.io
+      versions:
+      - groupVersion: managed.openshift.io/v1alpha1
+        version: v1alpha1
+      - groupVersion: managed.openshift.io/v1alpha2
+        version: v1alpha2
+      preferredVersion:
+        groupVersion: managed.openshift.io/v1alpha2
+        version: v1alpha2
```
rrati, line 45: I would add 2 more items here:

- Order so non-preferred version is first
- Multiple versions, but no preferred version

BumbleFeng: Add the fist one
the preferred version can not be empty

reconcile/utils/oc.py
BumbleFeng: Done

BumbleFeng (approved):

PR #1401, opened 2021-02-21
[RunningState] Default value for attributes not explicetly created is None

maorfr:
one of our e2e tests is `create_namespace` which creates a namespace on each cluster that we manage and verifies that it is created according to a predefined template.

this test has been failing:
```
  File "/usr/local/lib/python3.6/site-packages/reconcile-0.2.2-py3.6.egg/reconcile/utils/threaded.py", line 10, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/reconcile-0.2.2-py3.6.egg/e2e_tests/create_namespace.py", line 29, in test_cluster
    oc.delete_project(ns_under_test)
  File "/usr/local/lib/python3.6/site-packages/reconcile-0.2.2-py3.6.egg/reconcile/utils/oc.py", line 116, in wrapper
    integration=running_state.integration
AttributeError: 'RunningState' object has no attribute 'integration'
```

e2e tests should not be publishing metrics (they also do not run in the cluster, so it doesn't make sense at the moment). perhaps they should be running on the cluster just like any other integration? :thinking: 

in any case, this PR will fix this failure: https://ci.int.devshift.net/job/service-app-interface-gl-periodic-job-e2e-tests/256/artifact/reports/e2e_tests_reports_fail/e2e-test-create-namespace.txt

### DIFF
--- a/reconcile/status.py
+++ b/reconcile/status.py
@@ -22,5 +22,5 @@ class RunningState(_RunningState):
     """
     def __getattr__(self, item):
         """
-        Default value for attributes not explicetly created is None.
+        Default value for attributes not explicitly created is None.
         """
--- a/reconcile/status.py
+++ b/reconcile/status.py
@@ -20,3 +20,7 @@ class RunningState(_RunningState):
     the running state. Attributes will be populated
     by the callers.
     """
+    def __getattr__(self, item):
+        """
+        Default value for attributes not explicetly created is None.
+        """


reconcile/utils/oc.py
```diff
@@ -108,8 +108,12 @@ def wrapper(*args, **kwargs):
                 return result
 
             function_name = f'{function.__module__}.{function.__qualname__}'
-            ignore_reconcile_time = \
+            ignore_reconcile_time_annotation_present = \
                 annotations.get('qontract.ignore_reconcile_time') == 'true'
+            is_e2e_test = running_state.get('e2e_test', None)
```
erdii, line 113: what about 

```suggestion
            is_e2e_test = running_state.get('e2e_test', None) is not None
```

to make the intent 100% clear?

erdii: lgtm!

apahim, line 113: `RunningState()` has not `get()` method. You can use `getattr(running_state, 'e2e_test', None)`, but I'd rather just set a default for any `RunningState()` attribute instead of creating multiple execution paths here.

`RunningState()` is super prone to race conditions. Having a default will make it safer.

You can just add to `RunningState()`:

```python
    def __getattr__(self, item):
        """
        Default value for attributes not explicetly created is None.
        """
```

With that, if something breakes because the attibute is `None`, that something has to check the attribute first. ATM, the only thing that could break is the `reconcile_time.labels()` call bellow, but Histograms have no issues with `label=None` as argument, so I'd not touch it.

maorfr: so in fact this change is sufficient and nothing else is required!

reconcile/status.py
```diff
@@ -20,3 +20,7 @@ class RunningState(_RunningState):
     the running state. Attributes will be populated
     by the callers.
     """
+    def __getattr__(self, item):
+        """
+        Default value for attributes not explicetly created is None.
```
rporres, line 25:
```suggestion
        Default value for attributes not explicitly created is None.
```

maorfr: how many engineers does it take to write a single line of code which is at all a comment

rporres (approved):
I like the approach of not adding ad-hoc conditions

apahim (approved):
lgtm!
thank you for taking care of this!

reconcile/status.py
apahim: fun fact: if you drop this comment, you will need a line of code here :)

PR #1402, opened 2021-02-21
[smtp] lazy client init

maorfr:
we have been recently seeing errors with the SMTP client. any such issue is causing the following integrations to fail:
- email-sender
- requests-sender
- sql-query
- terraform-users
- sentry-helper

the integrations are failing to initiate the SMTP client, which is happening on every iteration, even if there are no changes to be made (most of the time). the client should only be initialized when it's needed (to send a mail).

this PR changes the smtp client to a property (also adding consistency to `server`) which means it will only initialize the client when it's needed and not as part of a class instantiation.

### DIFF
--- a/reconcile/utils/smtp_client.py
+++ b/reconcile/utils/smtp_client.py
@@ -23,13 +23,17 @@ class SmtpClient:
         self.passwd = smtp_config['password']
         self.mail_address = config['smtp']['mail_address']
 
-        self.client = smtplib.SMTP(host=self.host,
-                                   port=self.port)
-        self.client.send
-        self.client.starttls()
-        self.client.login(self.user, self.passwd)
-
+        self._client = None
         self._server = None
+
+    @property
+    def client(self):
+        if self._client is None:
+            self._client = smtplib.SMTP(host=self.host, port=self.port)
+            self._client.send
+            self._client.starttls()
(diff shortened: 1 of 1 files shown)


apahim (commented):
Nice!

reconcile/utils/smtp_client.py
```diff
@@ -23,14 +23,18 @@ def __init__(self, settings):
...
         self._server = None
 
+    @property
+    def client(self):
+        if self._client is None:
+            self._client = smtplib.SMTP(host=self.host, port=self.port)
+            self._client.send
+            self._client.starttls()
+            self._client.login(self.user, self.passwd)
+        return self._client
```
erdii, line 36: `@property` re-creates the client every time it is accessed - is this intended?
this is a version where the client is created once per class-instance and then cached:

```suggestion
    @functools.cached_property
    def client(self):
            client = smtplib.SMTP(host=self.host, port=self.port)
            client.send # <- btw, is this a side-effect hidden behind a property access? 🤨
            client.starttls()
            client.login(self.user, self.passwd)
        return client
```

https://docs.python.org/3/library/functools.html#functools.cached_property

maorfr: > @property re-creates the client every time it is accessed
can you expand on that?

PR #1405, opened 2021-02-21
[saas-file-owners] remove hold label when hold is cancelled

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3072

the current behavior will not remove the hold label if there isn't a valid lgtm. this isn't the right behavior, and the hold label should be removed if it's cancelled.


apahim (approved):
lgtm

erdii:
lgtm

PR #1407, opened 2021-02-21
[promtool] fix temporary file scope

maorfr:
a minor fix for #1399 

the change moved the promtool check out of the scope of the temp file context manager, leading to the removal of the temporary file before the execution of `promtool check`.


apahim (commented):
great catch!

erdii:
lgtm

PR #1412, opened 2021-02-23
[gql] retry init from config

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3084

there may be cases where the active sha in one pod is different from another pod, which may lead to inconsistencies in data fetched. adding a retry to this function eventually makes sure that we are using a matching sha and git commit info (even if those were queried between different qontract-server pods).

we also add a `max_attempts` of 5 to allow time for s3-reload to reload data. the default is: https://github.com/app-sre/csnotify/blob/a3522611b4d05c63525529b2a522977fc5aa3478/s3-watcher.go#L105

### DIFF
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -176,7 +176,7 @@ def get_sha(server, token=None):
     return sha
 
 
-@retry(exceptions=GqlApiError, max_attempts=5)
+@retry(exceptions=requests.exceptions.HTTPError, max_attempts=5)
 def get_git_commit_info(sha, server, token=None):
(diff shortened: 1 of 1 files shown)


jmelis:
@maorfr hold this PR until we LGTM

rporres:
I don't understand how this is going to work... when `get_git_commit_info` fails it uses `response.raise_for_status()` to raise an exception, which is not of the `GqlApiError` kind but `requests.exceptions.HTTPError`.

maorfr:
good catch @rporres! fixed it.

rporres (approved):

PR #1414, opened 2021-02-24
[vault] reauthenticate on missing data if not authenticated

maorfr:
following #1348, we now know what secrets are forbidden.

those secrets are usually not really forbidden, and my suspicion is that we are hitting the SA token limit. this makes sense since we recently moved to a singleton vault client, which means we are using the token without refreshing it.

this PR adds a refresh every time we find a secret forbidden and we find that the client is not authenticated.

### DIFF
--- a/reconcile/utils/vault.py
+++ b/reconcile/utils/vault.py
@@ -176,7 +176,7 @@ class _VaultClient:
         if data is None:
             self._refresh_client_auth()
             raise SecretNotFound
-            
+
         return base64.b64decode(data) if secret_format == 'base64' else data
 
     def _read_v2(self, path, field, version):
--- a/reconcile/utils/vault.py
+++ b/reconcile/utils/vault.py
@@ -43,8 +43,8 @@ class _VaultClient:
         config = get_config()
 
         server = config['vault']['server']
-        role_id = config['vault']['role_id']
-        secret_id = config['vault']['secret_id']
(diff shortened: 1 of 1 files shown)


arilivigni:
LGTM

reconcile/utils/vault.py
```diff
@@ -160,10 +164,19 @@ def read(self, secret):
         secret_field = secret['field']
         secret_format = secret.get('format', 'plain')
         secret_version = secret.get('version')
-        try:
+
+        kv_version = self._get_mount_version_by_secret_path(secret_path)
+
+        data = None
+        if kv_version == 2:
```
apahim, line 171: Are you sure that read_v2 won't raise an exception when trying to read from a v1 engine? I believe you verified that... just checking anyway :)

jfchevrette (approved):

reconcile/utils/vault.py
maorfr: with this change we are no longer trying to read v1 from v2 or vice versa

maorfr: same as here: https://github.com/app-sre/qontract-reconcile/pull/1414/files#diff-421d17fb36e3edd0412b07a396bcdb5a518bd98d1abfc44d5df8f6072d589540R86-R92

apahim:
lgtm

PR #1416, opened 2021-02-25
Sticky sessions

jmelis:
The goal of this PR is to enable session reusage in the graphql queries.

We were using the [GraphQLClient](https://pypi.org/project/graphqlclient/), but the codebase seems abandoned, and it doesn't support reusing sessions, so this PR has replaced the external `GraphQLClient` with an internal `QontractServerClient`.

It's important to note that the session will be regenerated every time we `init` again, which is good, because the [run-integration.py](https://github.com/jmelis/qontract-reconcile/blob/sticky-sessions/dockerfiles/hack/run-integration.py#L69) script will reinitialize every time. This has been tested (although manually, no automated tests for this).

### DIFF
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -200,7 +200,7 @@ def init(url, token=None, sha_url=True, integration=None,
     """
     global _gqlapi
     _gqlapi = GqlApi(url, token, integration, validate_schemas,
-                     sha_url=sha_url, use_sessions=True)
+                     sha_url=sha_url, use_sessions=use_sessions)
 
     if print_url:
         url = _gqlapi.client.query_url
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -64,7 +64,8 @@ class GqlApi:
         self.integration = int_name
         self.validate_schemas = validate_schemas
         self.client = QontractServerClient(
-            self.url, token=token, sticky_session=sticky_session, sha_url=sha_url)
+            self.url, token=token, sticky_session=sticky_session,
+            sha_url=sha_url)
 
         if validate_schemas and not int_name:
             raise Exception('Cannot validate schemas if integration name '
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -1,8 +1,6 @@
 import logging
 import textwrap
 
-from sretoolbox.utils import retry
-
 from reconcile.utils.qontract_server_client import QontractServerClient
 from reconcile.utils.config import get_config
 from reconcile.status import RunningState
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -58,13 +58,13 @@ class GqlApi:
     _queried_schemas = set()
 
     def __init__(self, url, token=None, int_name=None, validate_schemas=False,
-                 use_sessions=False, sha_url=False):
+                 sticky_session=False, sha_url=False):
         self.url = url
         self.token = token
         self.integration = int_name
         self.validate_schemas = validate_schemas
         self.client = QontractServerClient(
-            self.url, token=token, use_sessions=use_sessions, sha_url=sha_url)
+            self.url, token=token, sticky_session=sticky_session, sha_url=sha_url)
 
         if validate_schemas and not int_name:
             raise Exception('Cannot validate schemas if integration name '
@@ -146,7 +146,7 @@ class GqlApi:
 
 
 def init_from_config(sha_url=True, integration=None, validate_schemas=False,
-                     print_url=True, use_sessions=True):
+                     print_url=True, sticky_session=True):
     """Inits the GraphQL client based on the server url and token defined in the
     config. This method is a wrapper around `init`.
 
@@ -161,9 +161,9 @@ def init_from_config(sha_url=True, integration=None, validate_schemas=False,
     :param print_url: when a new sha is acquired, it will be printed,
         defaults to True
     :type print_url: bool, optional
-    :param use_sessions: reuse session for all the requests performed.
+    :param sticky_session: reuse session for all the requests performed.
         defaults to True
-    :type use_sessions: bool, optional
+    :type sticky_session: bool, optional
     """
 
     config = get_config()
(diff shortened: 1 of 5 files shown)


reconcile/utils/gql.py
```diff
@@ -91,20 +83,13 @@ def __init__(self, url, token=None, int_name=None, validate_schemas=False):
             if not self._valid_schemas:
                 raise GqlApiIntegrationNotFound(int_name)
 
-    @retry(exceptions=GqlApiError, max_attempts=5)
```
jmelis, line 94: Removing the `@retry` decorator. I feel we should let it crash every time the server is not reachable in order not to obscure the errors.

reconcile/utils/gql.py
```diff
@@ -91,20 +83,13 @@ def __init__(self, url, token=None, int_name=None, validate_schemas=False):
...
     def query(self, query, variables=None, skip_validation=False):
         try:
-            # supress print on HTTP error
-            # https://github.com/prisma-labs/python-graphql-client
-            # /blob/master/graphqlclient/client.py#L32-L33
-            with open(os.devnull, 'w') as f, contextlib.redirect_stdout(f):
-                result_json = self.client.execute(query, variables)
+            result = self.client.query(query, variables)
```
jmelis, line 88: No need to suppress print on HTTP error as the new library is not doing that.

reconcile/utils/qontract_server_client.py
```diff
@@ -0,0 +1,94 @@
...
+        self.use_sessions = use_sessions
+        self.sha = None
+
+        if sha_url:
+            self.set_graphqlsha_url()
+        else:
+            self.query_url = url
+
+        if self.use_sessions:
+            self.session = requests.Session()
```
jmelis, line 23: This is the main purpose of this PR.

reconcile/utils/qontract_server_client.py
```diff
@@ -0,0 +1,94 @@
...
+        :rtype: dict
+        """
+        data = {'query': query, 'variables': variables}
+
+        headers = {'Accept': 'application/json',
+                   'Content-Type': 'application/json'}
+        if self.token:
+            headers['Authorization'] = self.token
+
+        if self.use_sessions:
```
apahim, line 77: You don't really need `self.use_sessions` as a class attribute. You should control that with `self.session`, having it initialized as `None` or as `requests.Session()`. Here you'd check `if/else` `self.session is None`.

jmelis: fixed

reconcile/utils/gql.py
jmelis: @maorfr do you have any reasons for not doing this?

maorfr: network issues?

if we're looking for server issues, let's instrument the server IMO

apahim:
reviewing

reconcile/utils/gql.py
```diff
@@ -162,51 +145,74 @@ def get_queried_schemas(self):
...
+        defaults to True
+    :type print_url: bool, optional
+    :param use_sessions: reuse session for all the requests performed.
+        defaults to True
+    :type use_sessions: bool, optional
+
+    """
+    global _gqlapi
+    _gqlapi = GqlApi(url, token, integration, validate_schemas,
+                     sha_url=sha_url, use_sessions=True)
```
apahim, line 203: `use_sessions=use_sessions` ?

maorfr:
big +1 for moving away from the current gql client

reconcile/utils/qontract_server_client.py
```diff
@@ -0,0 +1,95 @@
+from urllib.parse import urlparse
+
+import requests
+
+
+class QontractServerClient:
+    """REST API for Qontract Server """
+
+    def __init__(self, url, token=None,
+                 use_sessions=False,
```
apahim, line 10: the `use_sessions` argument doesn't tell much & is exposing internal implementation... I'd suggest calling it `sticky_session` instead.

reconcile/utils/qontract_server_client.py
```diff
@@ -0,0 +1,95 @@
...
+
+        if sha_url:
+            self.set_graphqlsha_url()
+        else:
+            self.query_url = url
+
+    def set_graphqlsha_url(self):
+        """Fetches the latest sha and sets the query url to use it.
+        """
+        self.sha = self.get_sha()
```
apahim, line 29: It's a personal choice I guess, but I always find the class variables initialization outside of the `__init__` confusing. My suggestion is to return values from the helpers and to have all the initializations done on the `__init__`.

reconcile/utils/qontract_server_client.py
```diff
@@ -0,0 +1,95 @@
...
+        :type variables: dict, optional
+        :return: graphql response payload. keys: 'data' and '
+        :rtype: dict
+        """
+        data = {'query': query, 'variables': variables}
+
+        headers = {'Accept': 'application/json',
+                   'Content-Type': 'application/json'}
+        if self.token:
+            headers['Authorization'] = self.token
```
apahim, line 76: `headers` never changes... I'd move its initialization to `__init__`.

reconcile/utils/qontract_server_client.py
```diff
@@ -0,0 +1,95 @@
...
+
+        r = post_method(self.query_url, json=data, headers=headers)
+        r.raise_for_status()
+        return r.json()
+
+    def _base_url_path(self, path):
+        return urlparse(self.base_url)._replace(path=path).geturl()
+
+    def _get(self, path):
+        headers = {'Authorization': self.token} if self.token else None
```
apahim, line 91: can we re-use the `headers` from `query()`?

apahim (approved):
This is awesome! Lgtm, with a few small suggestions.

reconcile/utils/gql.py
jmelis: would you be aginst trying this out? I'd like to. We can add the decorator if we see a lot of errors.

apahim:
lgtm! ship it!

jmelis:
thanks!

PR #1418, opened 2021-03-01
Deleted openshift-performance-parameters

rporres:
We no longer find this approach valid as it was too restrictive and too
opinionated.  Current effort for tenants to express their SLOs are
in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/15506
The recording rules that were created by this integration will be
persisted in app-interface by means of that PR in app-interface.


maorfr (approved):
:(

PR #1421, opened 2021-03-01
[JJB] add generate xml file based on config name in print only mode

BumbleFeng:
cc @jmelis 


maorfr (approved):
LGTM

cc @jmelis

PR #1425, opened 2021-03-03
terraform enable deletion per account

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2936

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/15710

this PR will allow deleting resources in terraform integrations from aws accountss where `enableDeletion` is true.
there is a cli argument that can be passed to terraform integrations called `--enable-deletion`. this PR will enable deleting resources even if the default for enable deletion for the entire integration is false.

tl;dr - what we define on an aws account file is stronger.

### DIFF
--- a/reconcile/terraform_vpc_peerings.py
+++ b/reconcile/terraform_vpc_peerings.py
@@ -305,10 +305,10 @@ def run(dry_run, print_only=False,
 
     defer(lambda: tf.cleanup())
 
-    deletions_detected, err = tf.plan(enable_deletion)
+    disabled_deletions_detected, err = tf.plan(enable_deletion)
     if err:
         sys.exit(1)
-    if deletions_detected and not enable_deletion:
+    if disabled_deletions_detected and not enable_deletion:
         sys.exit(1)
 
     if dry_run:
--- a/reconcile/terraform_vpc_peerings.py
+++ b/reconcile/terraform_vpc_peerings.py
@@ -296,6 +296,7 @@ def run(dry_run, print_only=False,
     tf = Terraform(QONTRACT_INTEGRATION,
                    QONTRACT_INTEGRATION_VERSION,
                    "",
+                   accounts,
                    working_dirs,
                    thread_pool_size)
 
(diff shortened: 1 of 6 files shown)


rporres (commented):
The whole idea of disabling from the cli and enabling from the configuration is not pretty to say the least, so I don't really like the whole idea 😂 

As for the current implementation, I find very confusing that the logic of enabling deletion is shared between the client and the integration. I'd really prefer that the decision of deciding if we can delete or not is in the integration based on the combination of the cli flags and the account configuration as it will be very clear what we're trying to do from the integration side.

rporres:
But I sincerely don't see how to do it in the current shape...

reconcile/terraform_vpc_peerings.py
```diff
@@ -304,10 +305,10 @@ def run(dry_run, print_only=False,
 
     defer(lambda: tf.cleanup())
 
-    deletions_detected, err = tf.plan(enable_deletion)
+    disabled_deletions_detected, err = tf.plan(enable_deletion)
     if err:
         sys.exit(1)
-    if deletions_detected and not enable_deletion:
+    if disabled_deletions_detected and not enable_deletion:
```
rporres, line 311: The fact that we're checking again here for `enable_deletion` when we check for it in the client confuses me... Which is the scenario we're trying to cover here?

maorfr: Deletions in an account that doesn't enable deletion and the enable deletion flag is not set.
This is a forbidden (disabled) deletion

rporres: Wouldn't that return a `disabled_deletions_detected` `True` from https://github.com/app-sre/qontract-reconcile/blob/0ba2d18cf9b65b5caf621ad81f8c42a56aba366a/reconcile/utils/terraform_client.py#L185-L186?

maorfr: It's a nit in any case as I see it?
Maybe confusing, but not more...

rporres: yeah, just confusing... The overall approach seems to be the only way to achieve this :)

rporres (approved):
Just some nit-picking above. Not very important, but will definitely help to make the code a bit more understandable

PR #1426, opened 2021-03-04
[ocm-machine-pools] add taints and fix update

BumbleFeng:
change of labels and taints on the current machine-pool will only be applied to new Nodes.

https://github.com/openshift/machine-api-operator/blob/master/FAQ.md

### DIFF
--- a/reconcile/ocm_machine_pools.py
+++ b/reconcile/ocm_machine_pools.py
@@ -33,9 +33,12 @@ def fetch_desired_state(clusters):
         machine_pools = cluster['machinePools']
         for machine_pool in machine_pools:
             machine_pool['cluster'] = cluster_name
-            labels = machine_pool.get('labels')
+            labels = machine_pool.pop('labels')
             if labels:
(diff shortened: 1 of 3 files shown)


reconcile/ocm_machine_pools.py
```diff
@@ -85,6 +93,10 @@ def act(dry_run, diffs, ocm_map):
                 ocm.update_machine_pool(cluster, diff)
             elif action == 'delete':
                 ocm.delete_machine_pool(cluster, diff)
+            elif action == 'recreate':
+                ocm.delete_machine_pool(cluster, diff)
+                time.sleep(1)
```
maorfr, line 98: i don't like using `sleep`. it's something i picked up from my days writing test automation. in this case, i suggest we don't add a `recreate` method, but instead do a `delete` and `create`. it would probably make sense to move `if action == 'delete'` to be first in this switch case. wdyt?

perhaps `ocm.delete_machine_pool` should verify something instead of the sleep?

maorfr (approved):
otherwise LGTM!

PR #1428, opened 2021-03-04
parametrize config env in run-integration

jmelis:
I like running `run-integration.py` locally. In order to run it I usually modify the hardcoded config path, but with this change as a dev I will be able to just do:

```
CONFIG=./config.toml SLEEP_DURATION_SECS=3 DRY_RUN="--dry-run" INTEGRATION_NAME="quay-membership" python dockerfiles/hack/run-integration.py
```


apahim (approved):
lgtm

PR #1430, opened 2021-03-04
instrument gql query retry

jmelis:
I want to make sure we capture all errors on every gql attempt. For that we need the hook capability of the `@retry` method available in a newer sretoolbox release. I was force to update semver as sretoolbox depends on a newer version of semver.


reconcile/utils/gql.py
```diff
@@ -91,7 +92,7 @@ def __init__(self, url, token=None, int_name=None, validate_schemas=False):
             if not self._valid_schemas:
                 raise GqlApiIntegrationNotFound(int_name)
 
-    @retry(exceptions=GqlApiError, max_attempts=5)
+    @retry(exceptions=GqlApiError, max_attempts=5, hook=capture_exception)
```
rporres, line 95: what happens if `capture_exception` fails?

rporres: Or even better, we have to make sure that it is fire and forget.

reconcile/utils/gql.py
```diff
@@ -29,6 +30,19 @@
 """
 
 
+def capture_and_forget(error):
```
jmelis, line 33: is this what you had in mind @rporres?

jmelis:
FYI I had to bump semver to the version that sretoolbox uses, which generated some warnings that I had to fix by stopping using `semver.format_version` and `semver.parse`

reconcile/utils/gql.py
rporres: I guess that makes me feel more relaxed about this  :-D

rporres (approved):

PR #1438, opened 2021-03-10
[slack-usergroups] allow usergroup permission to be defined from multiple roles

maorfr:
due to the way we collect slack user groups (per role, per user) we are not taking into account that the same slack usergroup can be referenced from multiple roles. this leads to people being added and removed from such defined groups.

this PR takes into account that a previous item referencing the same slack usergroup was already added to the desired state. if such item is found, we append the additional users into it instead of adding a new item to the desired state.

### DIFF
--- a/reconcile/slack_usergroups.py
+++ b/reconcile/slack_usergroups.py
@@ -277,14 +277,20 @@ def get_desired_state(slack_map):
             channel_names = [] if p['channels'] is None else p['channels']
             channels = slack.get_channels_by_names(channel_names)
 
-            desired_state.append({
-                "workspace": workspace_name,
-                "usergroup": usergroup,
-                "usergroup_id": ugid,
-                "users": users,
-                "channels": channels,
-                "description": description,
-            })
+            existing_items = [i for i in desired_state
+                              if i['workspace'] == workspace_name
+                              and i['usergroup'] == usergroup]
+            if len(existing_items) == 1:
+                existing_items[0]['users'].update(users)
+            else:
+                desired_state.append({
+                    "workspace": workspace_name,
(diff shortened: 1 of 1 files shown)


reconcile/slack_usergroups.py
```diff
@@ -277,14 +277,20 @@ def get_desired_state(slack_map):
...
-                "usergroup_id": ugid,
-                "users": users,
-                "channels": channels,
-                "description": description,
-            })
+            existing_items = [i for i in desired_state
+                              if i['workspace'] == workspace_name
+                              and i['usergroup'] == usergroup]
+            if len(existing_items) == 1:
+                existing_items[0]['users'].update(users)
```
rporres, line 284: Is there a possibility that the already existing `users` is different to the new one? I'm trying to find the answer for myself but I guess you want a faster review 😅

maorfr: the existing users are likely different. if user files reference roles and roles reference a slack usergroup, different users reference different roles which may reference the same slack usergroup. this is what we are solving.

rporres: As long as we're ok with the last users list winning... :-D

maorfr: but why?
users is a dict, and `update` adds items to the dict...

rporres: because I hadn't properly read the code 😂 😅

rporres (approved):

reconcile/slack_usergroups.py
maorfr: that's a good reason! thanks a lot!

PR #1440, opened 2021-03-11
[vault] retry reading secret when access is forbidden

maorfr:
following #1414, we saw a major decrease in authentication errors. we do see them from time to time still, but they have reduced a lot.

in this PR, we add a retry in case where we get access forbidden for a v1 secret, which is most of what we are currently seeing. this should reduce the noise completely!


apahim (approved):
lgtm

PR #1444, opened 2021-03-16
[ocp-release-ecr-mirror] only mirror stable+enabled ClusterImageSets

maorfr:
we are currently mirroring all images.
Let's avoid some mirroring:
- if a CIS is disabled
- if a CIS is not in the stable channel

future work should include actually performing a reconciliation and not just mirroring, but this is an easy start.
Just with these changes we reduce the number of unique mirrored releases from 141 to 17.


apahim (approved):
lgtm
should we cleanup the repo as it is today, so we start fresh without hitting the 1k limit?

maorfr:
yes, once this is merged and promoted i will cleanup both ecr repos.

PR #1446, opened 2021-03-17
[qontract-cli] add get command for aws infrastructure access switch role links

maorfr:
a utility function to get a list of all the switch role links from the OCM console for all our clusters.

related to https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/FAQ.md#get-access-to-cluster-logs-via-log-forwarding


arilivigni (approved):
LGTM

PR #1449, opened 2021-03-18
[ocm] skip provision shards mapping

maorfr:
querying for each cluster's provision shard is time consuming and only required for the ocm-clusters integration.


BumbleFeng (approved):

arilivigni (approved):
LGTM

PR #1451, opened 2021-03-18
[terraform-tgw-attachments] create resource share and share tgw

maorfr:

data example: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/16637

this PR adds a new integration called terraform-tgw-attachments which will (eventually) create attachments between VPCs and TGWs. at this point, the integration creates a share between the accounts (TGW's account and VPC's account) and adds the TGW to be a shared resource.

follow up PRs will add attachments, etc.


reconcile/terraform_tgw_attachments.py
```diff
@@ -0,0 +1,166 @@
...
+            if accepter_vpc_id is None:
+                logging.error(f'[{cluster} could not find VPC ID for cluster')
+                error = True
+                continue
+            accepter['vpc_id'] = accepter_vpc_id
+            accepter['route_table_ids'] = accepter_route_table_ids
+            accepter['account'] = account
+
+            account_tgws = \
+                aws_api.get_tgws_details(
```
rporres, line 72: this can be empty if it's the first time we're provisioning, right?

reconcile/utils/terrascript_client.py
```diff
@@ -521,6 +528,78 @@ def populate_vpc_peerings(self, desired_state):
...
+            self.add_resource(acc_account_name, tf_resource_share_accepter)
+
+            # until now it was standard sharing
+            # from this line onwards we will be adding content
+            # specific for the TGW attachments integration
+
+            # tgw share association
+            identifier = f"{requester['tgw_id']}-{accepter['vpc_id']}"
+            values = {
+                'resource_arn': requester['tgw_arn'],
```
rporres, line 596: I don't understand well how this has been populated the first time the integration finds a new tgw definition

rporres (commented):
I have a couple of doubts... If you have a couple of minutes to go over them it would be great

reconcile/terraform_tgw_attachments.py
maorfr: we are not responsible for provisioning the TGW. we are just discovering them. this list can come back empty

reconcile/utils/terrascript_client.py
maorfr: same as above

rporres: understood, TGW are already provisioned

reconcile/terraform_tgw_attachments.py
rporres: no, it cannot... 😂

rporres (approved):

maorfr:
at this point, the integration successfully creates the resource share, adds the osd account sharing principal, accepts the resources share and adds the TGW to be shared.

merging.

PR #1456, opened 2021-03-19
[qontract-cli] add aws-route53-zones command

jfchevrette:
The `aws-route53-zones` command is used to list information about DNS zones that are managed in app-interface.

Other than listing the zone name and number of records, it also uses the `dnspython` package to query the nameservers (NS) and return them. This is useful for tenants who are looking for their nameservers to use after adding a new zone to app-interface.

I went with a simple DNS query rather than somehow looking at the terraform state or doing AWS API calls. DNS queries should work just fine and don't require extra privileges

I also added a small change to handle lists in table cells. tabular supports multiline cells natively for some table formats but does not have support for github-flavored markdown multiline cells, hence the addition of `<br>` tags when that format is used

The first use of this will be to output the information in the app-interface-output repo

### DIFF
--- a/reconcile/utils/dnsutils.py
+++ b/reconcile/utils/dnsutils.py
@@ -1,0 +1,9 @@
+from dns import resolver
+
+
+def get_nameservers(domain):
+    records = []
+    answers = resolver.query(domain, 'NS')
+    for rdata in answers:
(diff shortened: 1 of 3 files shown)


maorfr (approved):
❤️

PR #1457, opened 2021-03-19
Selective client

rrati:
Allow using unleash to control which client is used by specific clusters

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -24,6 +24,8 @@ from reconcile.utils.secret_reader import SecretReader
 import reconcile.utils.threaded as threaded
 from openshift.dynamic.exceptions import NotFoundError
 from openshift.dynamic import DynamicClient
+from reconcile.utils.unleash import (get_feature_toggle_strategies,
+                                     get_feature_toggle_state)
 from openshift.dynamic.resource import ResourceList
 
 urllib3.disable_warnings()
@@ -957,8 +959,25 @@ class OC:
     def __new__(cls, cluster_name, server, token, jh=None, settings=None,
                 init_projects=False, init_api_resources=False,
                 local=False):
-        use_native = os.environ.get('USE_NATIVE_CLIENT', '').lower() \
-            in ['true', 'yes']
+        use_native = os.environ.get('USE_NATIVE_CLIENT', '')
+        if not use_native:
+            enable_toggle = 'openshift-resources-native-client'
+            strategies = get_feature_toggle_strategies(
+                enable_toggle, ['perCluster'])
+
+            # only use the native client if the toggle is enabled and this
+            # server is listed in the perCluster strategy
+            cluster_in_strategy = False
+            if strategies:
+                for s in strategies:
+                    if cluster_name in s.parameters['cluster_name'].split(','):
+                        cluster_in_strategy = True
+                        break
+            use_native = get_feature_toggle_state(enable_toggle) and \
+                cluster_in_strategy
+        else:
+            use_native = use_native.lower() in ['true', 'yes']
+
         if use_native:
             OC.client_status.labels(
                 cluster_name=cluster_name, native_client=True).inc()
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -960,7 +960,9 @@ class OC:
                 init_projects=False, init_api_resources=False,
                 local=False):
         use_native = os.environ.get('USE_NATIVE_CLIENT', '')
-        if not use_native:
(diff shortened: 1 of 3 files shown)


rrati:
@rporres This is for you

depends on #1400

reconcile/utils/oc.py
```diff
@@ -620,6 +635,245 @@ def _run_json(self, cmd, allow_not_found=False):
...
+        k, group_version = self._parse_kind(kind)
+        obj_client = self._get_obj_client(group_version=group_version, kind=k)
+        try:
+            obj = obj_client.get(name=name, namespace=namespace)
+            return obj.to_dict()
+        except NotFoundError as e:
+            if allow_not_found:
+                return {}
+            else:
+                raise StatusCodeError(f"[{self.server}]: {e}")
```
Piojo, line 838: Does `e` mention `name` and `namespace`? Unless they come from a vault, they'd be very useful to have when troubleshooting.

reconcile/utils/oc.py
```diff
@@ -620,6 +635,245 @@ def _run_json(self, cmd, allow_not_found=False):
...
+            # this is a new kind so add it
+            kgv[kind] = [new]
+        else:
+            # this kind already exists, so check if this apigroup has already
+            # been added as an option.  If this apigroup is the preferred one,
+            # then replace the apigroup/version so that the preferred
+            # apigroup/version is used instead of a non-preferred one
+            group = new.split('/', 1)[0]
+            new_group = True
+            for pos in range(len(kgv[kind])):
```
Piojo, line 864: Wouldn't this be easier to read?
```python
for pos, item in enumerate(kgv[kind]):
    if group in item:
         new_group = False
         if preferred:
              kgv[kind][pos] = new
```

reconcile/utils/oc.py
```diff
@@ -620,6 +635,245 @@ def _run_json(self, cmd, allow_not_found=False):
...
+                    for kind in r.keys():
+                        if len(apigroup) == 0:
+                            gv = f'{version}'
+                        else:
+                            gv = f'{apigroup}/{version}'
+                        kind_groupversion = self.add_group_kind(
+                            kind, kind_groupversion, gv, obj.preferred)
+        return kind_groupversion
+
+    def get_items(self, kind, **kwargs):
```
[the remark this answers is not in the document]
rrati: I do not plan to change this on this iteration.  The API definitions are copied from the deprecated, and as such I do not want to change anything more than needed.   The big change here is to use native python client, and as such I want to minimize other changes not related to that no matter how much better they may be

reconcile/utils/oc.py
```diff
@@ -620,6 +635,245 @@ def _run_json(self, cmd, allow_not_found=False):
...
+        obj_client = self._get_obj_client(group_version=group_version, kind=k)
+
+        namespace = ''
+        if 'namespace' in kwargs:
+            namespace = kwargs['namespace']
+            # for cluster scoped integrations
+            # currently only openshift-clusterrolebindings
+            if namespace != 'cluster':
+                if not self.project_exists(namespace):
+                    return []
```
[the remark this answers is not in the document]
rrati: See above

reconcile/utils/oc.py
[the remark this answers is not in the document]
jmelis: +1 to not fixing this in this PR. We have this all over the codebase.  But, I do agree with changing our style to stop doing this.

reconcile/utils/oc.py
jmelis: +1, very good point. This goes in #1400 though.

rrati: ATM, this is the same error message raised by the original get function.  I would prefer not to change that here and do it in a separate PR once this is merged.  My intention was to keep inputs/outpus the same.

reconcile/utils/oc.py
```diff
@@ -957,8 +959,25 @@ class OC:
     def __new__(cls, cluster_name, server, token, jh=None, settings=None,
                 init_projects=False, init_api_resources=False,
                 local=False):
-        use_native = os.environ.get('USE_NATIVE_CLIENT', '').lower() \
-            in ['true', 'yes']
+        use_native = os.environ.get('USE_NATIVE_CLIENT', '')
```
BumbleFeng, line 962: current USE_NATIVE_CLIENT will be either "false" or "true", but in both conditions, if use_native will always be true.
```
>>> bool("false")
True
```

rrati: Fixed

BumbleFeng (approved):

PR #1461, opened 2021-03-24
[terraform-tgw-attachments] add support for vpc attachment and accepter

maorfr:

this will add the provisioning of the vpc<->tgw attachment (requesting in the cluster's aws account and accepting in the calling aws account).

### DIFF
--- a/reconcile/utils/aws_api.py
+++ b/reconcile/utils/aws_api.py
@@ -668,7 +668,7 @@ class AWSApi:
                 route_table_ids = [rt['RouteTableId']
                                    for rt in vpc_route_tables['RouteTables']]
             if subnets:
-                vpc_subnets = ec2.describe_subnets(
+                vpc_subnets = assumed_ec2.describe_subnets(
                     Filters=[{'Name': 'vpc-id', 'Values': [vpc_id]}]
                 )
                 subnets_id_az = \
--- a/reconcile/utils/aws_api.py
+++ b/reconcile/utils/aws_api.py
@@ -672,11 +672,12 @@ class AWSApi:
                     Filters=[{'Name': 'vpc-id', 'Values': [vpc_id]}]
                 )
                 subnets_id_az = \
-                    [{
-                        'id': s['SubnetId'],
-                        'az': s['AvailabilityZone']
-                     }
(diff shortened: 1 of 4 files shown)


reconcile/utils/aws_api.py
```diff
@@ -609,7 +609,7 @@ def init_ecr_auth_tokens(self, accounts):
 
         self.auth_tokens = auth_tokens
 
-    def get_cluster_vpc_id(self, account, route_tables=False):
+    def get_cluster_vpc_id(self, account, route_tables=False, subnets=False):
         """
         Returns a cluster VPC ID.
```
rporres, line 614: No longer true, it returns much more :)

Please document with an example what this function returns exactly. It will help a lot to understand the rest of the code

Since this function has grown in scope, I would rename it to something more related to what it currently does

reconcile/utils/terrascript_client.py
```diff
@@ -603,6 +608,54 @@ def populate_tgw_attachments(self, desired_state):
...
+            if self._multiregion_account_(req_account_name):
+                values['provider'] = 'aws.' + requester['region']
+            tf_resource_attachment_accepter = \
+                aws_ec2_transit_gateway_vpc_attachment_accepter(
+                    identifier, **values)
+            self.add_resource(
+                req_account_name, tf_resource_attachment_accepter)
+
+    @staticmethod
+    def get_az_unique_subnet_ids(subnets_id_az):
```
rporres, line 646: what's `subnets_id_az` exactly?

rporres (commented):
Looks good. Some things are quite difficult to really understand without having direct exposure with the AWS objects in place.

Some renaming/doc requests ;-)

reconcile/utils/terrascript_client.py
maorfr: it's a list of items with a subnet's id and availability zone. it is needed later to add subnets to the VPC attachment (and only 1 per AZ is allowed, hence we are passing the AZ).

reconcile/utils/aws_api.py
maorfr: DONE!

PR #1463, opened 2021-03-25
ECR Mirror

apahim:

### DIFF
--- a/reconcile/ecr_mirror.py
+++ b/reconcile/ecr_mirror.py
@@ -1,0 +1,138 @@
+import base64
+import logging
+
+from sretoolbox.container import Image
+from sretoolbox.container import Skopeo
+from sretoolbox.container.skopeo import SkopeoCmdError
+
+import reconcile.utils.threaded as threaded
+
+from reconcile import queries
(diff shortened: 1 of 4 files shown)


reconcile/ecr_mirror.py
```diff
@@ -0,0 +1,138 @@
...
+            account=account,
+            region=region,
+        )
+        self.ecr_auth = f'{self.ecr_username}:{self.ecr_password}'
+
+        self.image_username = None
+        self.image_password = None
+        self.image_auth = None
+        push_secret = self.instance['mirror']['pullCredentials']
+        if push_secret is not None:
```
jmelis, line 59: why this if?

jmelis: if it so happens that push_secret is None, shouldn't we raise an exception?

apahim: it's pull secret... typo. Repo to pull from might be public.

jmelis: nevermind, I understand it now

jmelis (approved):
LGTM

PR #1466, opened 2021-03-26
[terraform-resources] add support for public repositories

maorfr:
since public ecr repositories are now available in terraform, we can provision them.

the dependency is implementation in terrascript, which should be followed up  in https://github.com/mjuenema/python-terrascript/issues/160.
EDIT: i believe i found a hack that allows us to create aws_ecrpublic_repository resources without upgrading terrascript (added comments in the code).

terraform docs: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecrpublic_repository
schema changes: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/16919

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -96,6 +96,7 @@ provider
   identifier
   region
   output_resource_name
+  public
 }
 ... on NamespaceTerraformResourceS3CloudFront_v1 {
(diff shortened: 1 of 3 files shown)


jfchevrette (approved):
Clean & simple. LGTM

PR #1468, opened 2021-03-29
WiP: New integration: osd mirrors data update

apahim:


reconcile/osd_mirrors_data_updater.py
```diff
@@ -0,0 +1,148 @@
...
+    # Now the tricky part. The "OCP Release ECR Mirror" is a stand-alone
+    # object in app-interface. We have to process it so we get the
+    # upstream and the mirror repositories
+    instances = queries.get_ocp_release_ecr_mirror()
+    for instance in instances:
+        namespace = instance['ecrResourcesNamespace']
+        ocp_release_identifier = instance['ocpReleaseEcrIdentifier']
+        ocp_art_dev_identifier = instance['ocpArtDevEcrIdentifier']
+
+        ocp_release_tf_info = get_tf_resource_info(namespace,
```
rporres, line 88: should this be called `get_ecr_tf_resource_info`? You're looking for something very specific here

reconcile/osd_mirrors_data_updater.py
```diff
@@ -0,0 +1,148 @@
...
+                                                   ocp_release_identifier)
+
+        # We get an ECR resource from app-interface, but it has
+        # no mirror since the mirroring is done differently there.
+        # Anyway, let's just inject the mirror information so we process
+        # all the ECR resources the same way
```
rporres, line 94: So we just know this mirror exists? This comment is a bit daunting, you may want to really explain how things work or a reference about how to understand it 😅

reconcile/osd_mirrors_data_updater.py
```diff
@@ -0,0 +1,148 @@
...
+                                                   ocp_art_dev_identifier)
+        ocp_art_dev_tf_info['mirror'] = {
+            'url': 'quay.io/openshift-release-dev/ocp-v4.0-art-dev',
+            'pullCredentials': None,
+            'tags': None,
+            'tagsExclude': None
+        }
+        osd_mirrors.append(ocp_art_dev_tf_info)
+
+    # Initializing the AWS Client for all teh accounts
```
rporres, line 112:
```suggestion
    # Initializing the AWS Client for all the accounts
```

reconcile/osd_mirrors_data_updater.py
```diff
@@ -0,0 +1,148 @@
...
+        if 'name' not in account_info:
+            continue
+        if account_info['name'] != account:
+            continue
+        return account_info
+
+
+def get_image_uri(aws_cli, account, repository):
+    """
+    Finds the repository URL for a given ECR resource.
```
rporres, line 48: why the uri/url mix?

reconcile/osd_mirrors_data_updater.py
```diff
@@ -0,0 +1,148 @@
...
+                     accounts=accounts,
+                     settings=settings,
+                     init_ecr_auth_tokens=True)
+    aws_cli.map_ecr_resources()
+
+    # Building up the mirrors information in the
+    # install-config.yaml compatible format
+    mirrors_info = []
+    for tfr in osd_mirrors:
+        image_url = get_image_uri(
```
rporres, line 129: why the uri/url mix?

rporres (commented):
Code is in general very understandable and clear

I have a couple of doubts:

* The existing mirrors. Maybe a longer explanation is enough
* When/how this integration will be run

apahim:
> The existing mirrors. Maybe a longer explanation is enough

Not sure I understand. Each mirroring mechanist is document by it's own. What are you looking for here?

> When/how this integration will be run

This is a normal integration. I'm just coordinating the deploymet so nothing fails anywhere. THe helm chart bits will come after I merge the app-interface MR.

reconcile/osd_mirrors_data_updater.py
apahim: done

rporres:
> The existing mirrors. Maybe a longer explanation is enough

For those that we don't have the same clear picture that you have 😂 

I just don't understand what you exactly mean in the code comment referenced in my review comment https://github.com/app-sre/qontract-reconcile/pull/1468/files#r603615812

reconcile/utils/mr/cluster_service_install_config.py
```diff
@@ -0,0 +1,73 @@
...
+            },
+
+        }
+
+        config_map['data'] = {'install-config.yaml': yaml.dump(
+            install_config, indent=2, Dumper=yaml.RoundTripDumper
+        )}
+        yaml.scalarstring.walk_tree(config_map)
+
+        content = ("# This file is auto-generated via app-interface. \n"
```
rporres, line 58: nit: app-interface autogenerates? 🤔 😂  Maybe it makes more sense to put here the integration that created the MR. Not critical by any means

rporres (approved):
Thx for documenting the mirroring thing, now it is way clearer

reconcile/utils/mr/cluster_service_install_config.py
apahim: sure, but I can't say which integration is using this lib here.

PR #1470, opened 2021-03-31
Fixing tf-r rds replica probe

apahim:


rporres (approved):
Awesome. Thx for fixing this and the debugging session <3

PR #1471, opened 2021-03-31
introduce integration jenkins-job-builds-cleaner

jfchevrette:
We currently have a cronjob running on jenkins master that looks at the builds folders creation time and deletes them according to timestamps. The behavior is similar in this new integration but while the cronjob is deleting the build folders from the filesystem this integration will do it via the jenkins REST API

The problem with deleting the build folders from the filesystem is that jenkins still keeps all the builds loaded in memory until there is a safe or full restart of jenkins. This causes increased memory and overhead when jobs and build history are queried. Listing build history happens frequently from many places: the web UI, monitoring/prometheus and upon every webhook among others.

This integration is meant to replace the cronjob that we have on the server so that builds are deleted properly via the API.

One alternative to this integration was to use the builds discarder functionality in jenkins but as many have pointed out this has caused issues in the past, thus the reason for wanting to do it via an integration.

### DIFF
--- a/reconcile/jenkins_job_builds_cleaner.py
+++ b/reconcile/jenkins_job_builds_cleaner.py
@@ -1,0 +1,83 @@
+import logging
+import re
+import time
+
+import reconcile.queries as queries
+
+from reconcile.utils.jenkins_api import JenkinsApi
+
+QONTRACT_INTEGRATION = 'jenkins-job-builds-cleaner'
+
+
+def hours_to_ms(hours):
+    return hours * 60 * 60 * 1000
+
+
+def delete_builds(jenkins, builds_todel, dry_run=True):
+    delete_builds_count = len(builds_todel)
+    for idx, build in enumerate(builds_todel, start=1):
+        job_name = build['job_name']
+        build_id = build['build_id']
+        progress_str = f"{idx}/{delete_builds_count}"
+        logging.debug([progress_str, job_name, build['rule_name'],
+                       build['rule_keep_hours'], build_id])
+        if not dry_run:
+            try:
+                jenkins.delete_build(build['job_name'], build['build_id'])
+            except Exception as e:
+                msg = f"failed to delete {job_name}/{build_id}"
+                logging.exception(msg)
+
+
+def find_builds(jenkins, job_names, rules):
+    # Current time in ms
+    time_ms = time.time() * 1000
+
+    builds_found = []
+    for job_name in job_names:
+        builds = None
+        for rule in rules:
+            if rule['name_re'].search(job_name):
+                # Fetch list of builds if we dont have it already
+                if not builds:
+                    builds = jenkins.get_builds(job_name)
+                for build in builds:
+                    if time_ms - rule['keep_ms'] > build['timestamp']:
+                        builds_found.append({
+                            'job_name': job_name,
+                            'rule_name': rule['name'],
+                            'rule_keep_hours': rule['keep_hours'],
+                            'build_id': build['id'],
+                        })
+    return builds_found
+
+
+def run(dry_run):
+    jenkins_instances = queries.get_jenkins_instances()
+    settings = queries.get_app_interface_settings()
+
+    for instance in jenkins_instances:
+        instance_cleanup_rules = instance.get('buildsCleanupRules', [])
+        if not instance_cleanup_rules:
+            # Skip instance if no cleanup rules defined
+            continue
+
+        # Process cleanup rules, pre-compile as regexes
+        cleanup_rules = []
+        for rule in instance_cleanup_rules:
+            cleanup_rules.append({
+                'name': rule['name'],
+                'name_re': re.compile(rule['name']),
+                'keep_hours': rule['keep_hours'],
+                'keep_ms': hours_to_ms(rule['keep_hours'])
+            })
+
+        token = instance['token']
+        jenkins = JenkinsApi(token, ssl_verify=False, settings=settings)
+        all_job_names = jenkins.get_job_names()
+
+        builds_todel = find_builds(jenkins, all_job_names, cleanup_rules)
+
+        logging.info(f'{len(builds_todel)} builds will be deleted')
+        delete_builds(jenkins, builds_todel, dry_run)
+        logging.info('deletion completed')
--- a/reconcile/utils/jenkins_api.py
+++ b/reconcile/utils/jenkins_api.py
@@ -50,6 +50,15 @@ class JenkinsApi:
 
         return jobs_state
 
+    def delete_build(self, job_name, build_id):
+        url = f"{self.url}/job/{job_name}/{build_id}/doDelete"
+        res = requests.post(
+            url,
+            verify=self.ssl_verify,
+            auth=(self.user, self.password),
+        )
+        res.raise_for_status()
+
     def delete_job(self, job_name):
         kwargs = self.get_crumb_kwargs()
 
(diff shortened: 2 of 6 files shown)


reconcile/jenkins_job_builds_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+                            delete_builds.append({
+                                'job_name': job_name,
+                                'rule_name': rule['name'],
+                                'rule_keep_hours': rule['keep_hours'],
+                                'build_id': build['id'],
+                            })
+
+        todel_current = 0
+        todel_total = len(delete_builds)
+        for build in delete_builds:
```
jmelis, line 64: I think the right way to do this is to use https://docs.python.org/3/library/functions.html#enumerate

reconcile/utils/jenkins_api.py
```diff
@@ -33,6 +33,15 @@ def get_job_names(self):
         job_names = [r['name'] for r in res.json()['jobs']]
         return job_names
 
+    def delete_build(self, job_name, build_id):
```
jmelis, line 36: do you want to add a retry decorator here?

jmelis:
@jfchevrette fantastic job! super clear!

I have added a couple of minor comments.

My only concerns are:

- In the past I've seen the delete job api call fail randomly. The api call returned an error code (I don't remember which), but the job was properly deleted. If you've already tried this and haven't seen this, perhaps it's not worth it, but otherwise I would propose to add a mechanism to handle those errors gracefully. Hence my `@retry` comment above, but I believe that won't be enough (as it may fail on subsequent calls, since the jobs may be deleted by the api call that raises the error).

- Do we want to run this continiously or on a schedule? Perhaps we want to run it only at specific hours? Although it may be worth trying this continuously at first, to see if there's any impact.

reconcile/jenkins_job_builds_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+        if not instance_cleanup_rules:
+            # Skip instance if no cleanup rules defined
+            continue
+
+        # Process cleanup rules, pre-compile as regexes
+        cleanup_rules = []
+        for rule in instance_cleanup_rules:
+            cleanup_rules.append({
+                'name': rule['name'],
+                'name_re': re.compile(rule['name']),
```
Piojo, line 33: Why this one? It looks to me like the names are always going to be fixed strings, so searching for them should be very idiomatic.

reconcile/jenkins_job_builds_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+        token = instance['token']
+        jenkins = JenkinsApi(token, ssl_verify=False, settings=settings)
+        all_job_names = jenkins.get_job_names()
+
+        delete_builds = []
+
+        for job_name in all_job_names:
+            builds = None
+            for rule in cleanup_rules:
+                if rule['name_re'].search(job_name):
```
Piojo, line 48: I lack a lot of context, but this looks to me like, as mentioned above.
```
if rule['name'] in job_name:
```

reconcile/jenkins_job_builds_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+            job_name = build['job_name']
+            build_id = build['build_id']
+            progress_str = f"{todel_current}/{todel_total}"
+            logging.info(['clean_job_builds', progress_str,
+                         instance_name, job_name, build['rule_name'],
+                         build['rule_keep_hours'], build_id])
+            if not dry_run:
+                try:
+                    jenkins.delete_build(build['job_name'], build['build_id'])
+                except Exception as e:
```
Piojo, line 75: Can you capture a specific exception class? I've suffered way more pain than I want to remember with code that captured `Exception`.

reconcile/jenkins_job_builds_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+                    msg = f"[{instance_name}] failed to delete " \
+                          f"{job_name}/{build_id}. Error was: {e}"
+                    logging.error(msg)
```
Piojo, line 78: `logger.exception` will print the exception and give a stack trace, which is very handy for debugging:

```python
logging.exception(f"[{instance_name}] failed to delete {job_name}/{build_id}")
```

reconcile/jenkins_job_builds_cleaner.py
```diff
@@ -0,0 +1,78 @@
...
+from reconcile.utils.jenkins_api import JenkinsApi
+
+QONTRACT_INTEGRATION = 'jenkins-job-builds-cleaner'
+
+
+def hours_to_ms(hours):
+    return hours * 60 * 60 * 1000
+
+
+def run(dry_run):
```
Piojo, line 16: A matter of personal preference, but I find this function too long and deeply indented, which makes following very hard for my sleep-deprived brain. Can we split it?

reconcile/jenkins_job_builds_cleaner.py
jfchevrette: enumerate is indeed cleaner! changed in 374eb2bfed28829ec553cb5e7a8d5c5a3f819c1f

reconcile/utils/jenkins_api.py
jfchevrette: I think in this case I would prefer to let it fail early and we catch it where we use the function. Adding a retry to this would make it inconsistent from the rest of the jenkins_api library

In this specific case it doesn't matter, the next integration run will effectively retry the delete

I personally am not in favor the use of the retry decorator in libraries (vs integration / end-user code). For transient issues they tend to mask the underlying issues by retrying and never bubbling up the issues. We have things that use the retry decorator all over the place and in the end the retry times add up and is part of the reason why things get slow(er) over time and we never know why because we silently retry

jfchevrette:
>     * In the past I've seen the delete job api call fail randomly. The api call returned an error code (I don't remember which), but the job was properly deleted. If you've already tried this and haven't seen this, perhaps it's not worth it, but otherwise I would propose to add a mechanism to handle those errors gracefully. Hence my `@retry` comment above, but I believe that won't be enough (as it may fail on subsequent calls, since the jobs may be deleted by the api call that raises the error).

I have seen sporadic errors in my test runs as well, but they were always temporary and the next run would delete the job. I don't think it's worth adding the retry here as it would 1) mask underlying issues that I would rather want to bubble up and be made visible 2) this would make the method behave differently from the rest of the jenkins_api library
 
>     * Do we want to run this continiously or on a schedule? Perhaps we want to run it only at specific hours? Although it may be worth trying this continuously at first, to see if there's any impact.

My idea was to run it continuously, with a SLEEP_DURATION_SECS of 1 hour in between runs. It's fairly lightweight as it deletes the jobs sequentially with no threading or parallelism.

reconcile/jenkins_job_builds_cleaner.py
jfchevrette: I wanted to support both a string and a regex. Some of the patterns we will have are regexes hence why I'm compiling them once here. Literal strings will also compile as regexes and should work as if we were doing a simple string matching, albeit via a regex search

reconcile/jenkins_job_builds_cleaner.py
jfchevrette: unless I'm missing your point, this would not work for regexes? (also see my previous comment)

reconcile/jenkins_job_builds_cleaner.py
jfchevrette: I tried to see what we could do without refactoring too much of the jenkins_api library. 

At the moment the delete_build method (most methods from jenkins_api.py really) will simply return `res.raise_for_status()` from `requests`. As best I can tell I would need to catch an exception from `requests` here but this code isn't aware of requests. I could import it, but really `requests` is an implementation detail of the library.

I think the best solution would be to refactor the jenkins_api.py library and have all methods return proper library-specific exceptions instead of `requests`.

Does that make sense? Can you suggest something I may have missed?

reconcile/jenkins_job_builds_cleaner.py
jfchevrette: thanks for the tip! changed in 881ad7d2b267cf50e37a7e7fc6d7874eada1db77

reconcile/jenkins_job_builds_cleaner.py
jfchevrette: I've split the method in a9b852510c1192fb86483ea091d8a8e343239cbd

PR #1473, opened 2021-03-31
APPSRE-3207: Generate `REPO_DIGEST`

2uasimojo:
Support a new template parameter, `REPO_DIGEST`. If required in the template, we will query the registry data for the image based on `REGISTRY_IMG` and `IMAGE_TAG`, which must also be present (whether specified or generated). `REPO_DIGEST` will be set using the same form as produced by

`docker image inspect ${REGISTRY_IMG}:${IMAGE_TAG} --format '{{index .RepoDigests 0}}'`

(except the image need not be present locally)

That is, `${REGISTRY_IMG}@sha256${DIGEST}`, where `${DIGEST}` is a 64c
hex string. For example, given:
- `REGISTRY_IMG = "quay.io/app-sre/osd-metrics-exporter-registry"`
- `IMAGE_TAG` = "production-2cb14b0"`
...we will set:
- `REPO_DIGEST = "quay.io/app-sre/osd-metrics-exporter-registry@sha256:c94893c302cf5ef8aae0581f60bf59c3f65860aeefdd6046440bdc95f258c26e"`

APPSRE-3207

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -6,6 +6,7 @@ import os
 import yaml
 
 from github import Github, GithubException
+from requests import exceptions as rqexc
 from sretoolbox.container import Image
 from sretoolbox.utils import retry
 
@@ -336,6 +337,7 @@ class SaasHerder():
     def _process_template(self, options):
         saas_file_name = options['saas_file_name']
         resource_template_name = options['resource_template_name']
+        image_auth = options['image_auth']
         url = options['url']
         path = options['path']
         provider = options['provider']
@@ -395,6 +397,38 @@ class SaasHerder():
                             # add IMAGE_TAG only if it is required
                             image_tag = commit_sha[:hash_length]
                             consolidated_parameters['IMAGE_TAG'] = image_tag
+
+            # Do this in a separate loop, since it relies on IMAGE_TAG already
+            # being calculated.
+            # This should almost never already be set in
+            # consolidated_parameters, but just in case...
+            if "REPO_DIGEST" not in consolidated_parameters:
+                for template_parameter in template.get("parameters", {}):
+                    if template_parameter["name"] == "REPO_DIGEST":
+                        try:
+                            logging.debug("Generating REPO_DIGEST.")
+                            registry_image = consolidated_parameters[
+                                "REGISTRY_IMG"]
+                            image_tag = consolidated_parameters["IMAGE_TAG"]
+                        except KeyError as e:
+                            logging.error(
+                                f"[{saas_file_name}/{resource_template_name}] "
+                                + f"{html_url}: error generating REPO_DIGEST. "
+                                + "Is REGISTRY_IMG or IMAGE_TAG missing? "
+                                + f"{str(e)}")
+                            return None, None
+                        try:
+                            img = Image(
+                                f"{registry_image}:{image_tag}", **image_auth)
+                            consolidated_parameters[
+                                "REPO_DIGEST"] = f"{img.url_digest}"
+                        except (rqexc.ConnectionError, rqexc.HTTPError) as e:
+                            logging.error(
+                                f"[{saas_file_name}/{resource_template_name}] "
+                                + f"{html_url}: error generating REPO_DIGEST. "
+                                + "Couldn't talk to the registry: "
+                                + f"{str(e)}")
+                            return None, None
 
             oc = OC('server', 'token', local=True)
             try:
@@ -606,6 +640,7 @@ class SaasHerder():
                 process_template_options = {
                     'saas_file_name': saas_file_name,
                     'resource_template_name': rt_name,
+                    'image_auth': image_auth,
                     'url': url,
                     'path': path,
                     'provider': provider,


reconcile/utils/saasherder.py
```diff
@@ -388,13 +388,38 @@ def _process_template(self, options):
...
 
             if "IMAGE_TAG" not in consolidated_parameters:
-                template_parameters = template.get('parameters')
-                if template_parameters is not None:
-                    for template_parameter in template_parameters:
-                        if template_parameter['name'] == 'IMAGE_TAG':
-                            # add IMAGE_TAG only if it is required
-                            image_tag = commit_sha[:hash_length]
-                            consolidated_parameters['IMAGE_TAG'] = image_tag
+                for template_parameter in template.get("parameters", {}):
```
2uasimojo, line 391: This delta not strictly necessary, just saved an indent and made the two chunks match stylistically.

reconcile/utils/saasherder.py
```diff
@@ -388,13 +388,38 @@ def _process_template(self, options):
...
+            # Do this in a separate loop, since it relies on IMAGE_TAG already
+            # being calculated.
+            # This should almost never already be set in
+            # consolidated_parameters, but just in case...
+            if "REPO_DIGEST" not in consolidated_parameters:
+                for template_parameter in template.get("parameters", {}):
+                    if template_parameter["name"] == "REPO_DIGEST":
+                        try:
+                            logging.info("Generating REPO_DIGEST.")
+                            import docker
```
2uasimojo, line 406: Where do I go to make sure the `docker` package is installed?

jharrington22:
@2uasimojo looks good to me - shouldn't effect existing deploys unless OC templates include this parameter. 

To me IMAGE_DIGEST makes more sense than REPO_DIGEST since we're switching on uses digests for image references.

reconcile/utils/saasherder.py
2uasimojo: Nope. The second argument to `.get()` is the default value to use if the key is not present. The default default is `None`, but here I switched to use `{}` (the empty dict) which will make the loop iterate zero times.

reconcile/utils/saasherder.py
rrati: https://github.com/app-sre/container-images

2uasimojo:
Okay, note to reviewers: The second commit is formatting only. Rather than manually editing to conform to flake8, I ran `black` on the thing. You should review only the first commit for content.

jharrington22:
> > To me IMAGE_DIGEST makes more sense than REPO_DIGEST since we're switching on the use of digests for image references.
> 
> I'm matching the `docker inspect` field, which is (a list) called `RepoDigests`. (Only guessing, but that's probably short for "Repository URI plus image digest".)

Makes sense.

2uasimojo:
> Okay, note to reviewers: The second commit is formatting only. Rather than manually editing to conform to flake8, I ran `black` on the thing. You should review only the first commit for content.

Bleh, I forgot `black` uses line length as a suggestion. I'll go back and reformat by hand :(

reconcile/utils/saasherder.py
2uasimojo: :+1: 

https://github.com/app-sre/container-images/pull/13

2uasimojo:
/hold

Depends-On: https://github.com/app-sre/container-images/pull/13

reconcile/utils/saasherder.py
```diff
@@ -388,13 +388,42 @@ def _process_template(self, options):
...
+                            docker_client = docker.from_env()
```
apahim, line 407: Are we sure that saas deployer has pull credentials in place to ANY images that can ever be templated?

apahim: Maybe what we have to do is to keep IMAGE_TAG unchanged and introduce a new IMAGE_DIGEST tag that will trigger this special behavior.

apahim: On top of that, we need to make sure we have at least the quay.io/app-sre pull credentials in the saas-deployer job env.

2uasimojo: Probably not. Nor do I suppose we can count on that being the case across the board, since the push could technically happen anywhere, not necessarily in chain with this saas deployer run.

I'm going to say we should defer that problem to future, if/when we actually encounter it. In the meantime, if someone tries to request `REPO_DIGEST` for a private image, this will blow up, which is good. (I.e. we don't want it to silently succeed with some bogus value.)

reconcile/utils/saasherder.py
```diff
@@ -388,13 +388,42 @@ def _process_template(self, options):
...
+                            reg_img = consolidated_parameters["REGISTRY_IMG"]
+                            img_tag = consolidated_parameters["IMAGE_TAG"]
+                            reg_data = docker_client.images.get_registry_data(
```
apahim, line 410: We have a lib, already part of the qontract-reconcile, that can grab the digest:
```python
>>> from sretoolbox.container import Image
>>> 
>>> i = Image('quay.io/app-sre/qontract-reconcile:latest', username=None, password=None)
>>> i.url_digest
'quay.io/app-sre/qontract-reconcile@sha256:9d119edb4c906e013c1b72c97cdab3fd26b35bed96f9b436990f43b9dbf9780e'
```

Missing image:

```python
>>> i = Image('quay.io/app-sre/missing:latest', username=None, password=None)
>>> i.url_digest
  File "/home/apahim/git/sretoolbox/sretoolbox/container/image.py", line 336, in url_digest
    url_digest += f'/{self.image}@{self.digest}'
  File "/home/apahim/git/sretoolbox/sretoolbox/container/image.py", line 79, in digest
    manifest = self._get_manifest()
  File "/home/apahim/git/sretoolbox/sretoolbox/container/image.py", line 94, in _get_manifest
    return self._request_get(url)
  File "/home/apahim/git/sretoolbox/sretoolbox/utils/retry.py", line 28, in f_retry
    defaults to ()
  File "/home/apahim/git/sretoolbox/sretoolbox/utils/retry.py", line 23, in f_retry
    :type exceptions: tuple(Exception, ...), optional
  File "/home/apahim/git/sretoolbox/sretoolbox/container/image.py", line 312, in _request_get
    self._raise_for_status(response)
  File "/home/apahim/git/sretoolbox/sretoolbox/container/image.py", line 286, in _raise_for_status
    raise HTTPError(msg)
requests.exceptions.HTTPError: (401) UNAUTHORIZED, access to the requested resource is not authorized
```

reconcile/utils/saasherder.py
```diff
@@ -388,13 +388,42 @@ def _process_template(self, options):
...
+                                f"{reg_img}:{img_tag}")
+                            digest = reg_data.attrs["Descriptor"]["digest"]
+                            # `digest` is of the form "sha256:abcd1234..."
+                            consolidated_parameters[
+                                "REPO_DIGEST"] = f"{reg_img}@{digest}"
+                        except Exception as e:
```
apahim, line 416: Let's be specific about the exceptions we are handling here.

reconcile/utils/saasherder.py
apahim: Not asking you to use it, just presenting you with an option that's already part of the repo :)

reconcile/utils/saasherder.py
2uasimojo: > Maybe what we have to do is to keep IMAGE_TAG unchanged and introduce a new IMAGE_DIGEST tag that will trigger this special behavior.

That's exactly what I've done (except `s/IMAGE_DIGEST/REPO_DIGEST/`).

> On top of that, we need to make sure we have at least the quay.io/app-sre pull credentials in the saas-deployer job env.

I mean, we _can_, but perhaps that's future work? I can make a note (where?) that this currently only works for public images.

reconcile/utils/saasherder.py
2uasimojo: I can peel out `KeyError`, but I'd like to keep the catch-all in any case, because I don't know what kinds of exceptions the docker lib can raise. Acceptable? (There are other spots in this file that use `Exception`...)

apahim: We are cleaning up the pylint errors, generic exceptions being one of them. Having this here now only increases the upcoming amount of work.

reconcile/utils/saasherder.py
2uasimojo: Neat! And it doesn't require the `docker` package. I'll switch it.

reconcile/utils/saasherder.py
2uasimojo: Understood. If I miss one, it'll blow up ugly instead of pretty -- I assume that's an acceptable risk?

reconcile/utils/saasherder.py
jharrington22: I'm assuming is assuming the registry is up. Whats the exception if the registry is down?

jharrington22:
@rrati @maorfr I see this has a hold, I think we need to make an internal decision before this gets merged.

reconcile/utils/saasherder.py
apahim: `requests.exceptions.ConnectionError`

reconcile/utils/saasherder.py
2uasimojo: Update here: I've switched to using `sretoolbox.container.Image` and funneled in the same `image_auth` used elsewhere. So this should be good as far as creds already exist. And I believe that also puts the onus on the client to make sure they have the right creds in place if necessary, right?

reconcile/utils/saasherder.py
2uasimojo: This is done.

reconcile/utils/saasherder.py
2uasimojo: This is done.

reconcile/utils/saasherder.py
```diff
@@ -388,13 +390,41 @@ def _process_template(self, options):
...
+
+            # Do this in a separate loop, since it relies on IMAGE_TAG already
+            # being calculated.
+            # This should almost never already be set in
+            # consolidated_parameters, but just in case...
+            if "REPO_DIGEST" not in consolidated_parameters:
+                for template_parameter in template.get("parameters", {}):
+                    if template_parameter["name"] == "REPO_DIGEST":
+                        try:
+                            logging.info("Generating REPO_DIGEST.")
```
maorfr, line 407: we generally don't log info, as we look at the output of the integration and only want to see actions that occured. i suggest to change this to debug

reconcile/utils/saasherder.py
```diff
@@ -388,13 +390,41 @@ def _process_template(self, options):
...
+                            reg_img = consolidated_parameters["REGISTRY_IMG"]
+                            img_tag = consolidated_parameters["IMAGE_TAG"]
```
maorfr, line 409: sorry to be petty, but can you name variables in the same way we do? `image_tag` over `img_tag`, and no abbreviations in general

reconcile/utils/saasherder.py
```diff
@@ -388,13 +390,41 @@ def _process_template(self, options):
...
+                        except KeyError as e:
+                            logging.error(
+                                f"[{saas_file_name}/{resource_template_name}] "
+                                + f"{html_url}: error generating REPO_DIGEST. "
+                                + f"Is REGISTRY_IMG or IMAGE_TAG missing? "
```
maorfr, line 414: IMAGE_TAG will never be missing

reconcile/utils/saasherder.py
```diff
@@ -388,13 +390,41 @@ def _process_template(self, options):
...
+                    if template_parameter["name"] == "IMAGE_TAG":
+                        # add IMAGE_TAG only if it is required
+                        image_tag = commit_sha[:hash_length]
+                        consolidated_parameters["IMAGE_TAG"] = image_tag
+
+            # Do this in a separate loop, since it relies on IMAGE_TAG already
+            # being calculated.
+            # This should almost never already be set in
+            # consolidated_parameters, but just in case...
+            if "REPO_DIGEST" not in consolidated_parameters:
```
maorfr, line 403: can we use "IMAGE_DIGEST" instead?

reconcile/utils/saasherder.py
```diff
@@ -606,6 +636,7 @@ def init_populate_desired_state_specs(self, saas_file):
                 process_template_options = {
                     'saas_file_name': saas_file_name,
                     'resource_template_name': rt_name,
+                    'image_auth': image_auth,
```
maorfr, line 639: :heart:

maorfr (changes requested):
thanks a lot for this contribution!

reconcile/utils/saasherder.py
2uasimojo: I chose `REPO_DIGEST` to match how `docker image inspect` names the corresponding field (`RepoDigests`, a list) in their JSON output. (I'm assuming that's short for "Repository URI plus image digest".)

maorfr: good thing we are not using docker anymore?

reconcile/utils/saasherder.py
2uasimojo: Reverted.

reconcile/utils/saasherder.py
2uasimojo: Done.

reconcile/utils/saasherder.py
2uasimojo: Done.

reconcile/utils/saasherder.py
2uasimojo: Based on L392-394 -- and reinforced by the comment on L395 -- it looks like it can be missing if it's not included in the saas parameters or the template parameters. Is there somewhere else that it's magically injected in that case?

reconcile/utils/saasherder.py
2uasimojo: Okay, but I agree with them on this point: The "image digest" is the `sha256:{hash}` part, not the whole URI. So IMO `IMAGE_DIGEST` isn't the correct name. And `IMAGE_URI_BY_DIGEST` gets a bit mealy, especially with python line width restrictions enforced at this level of indent :)

reconcile/utils/saasherder.py
maorfr: Ah you are right

reconcile/utils/saasherder.py
maorfr: Fair enough

maorfr (approved):
LGTM!

Thanks again. The result is very clear and to the point. I'll merge and promote with some docs tomorrow.

PR #1478, opened 2021-04-04
[terraform-tgw-attachments] add support for routes

maorfr:

this PR adds support for managing routes in the VPC's route tables (in the cluster's aws account).
the cidr block comes from app-interface and is not known at run time.


maorfr:
i'd appreciate a retrospective review! merging this to move forward

PR #1481, opened 2021-04-05
periodic vault refresh

jmelis:
The vault token has a TTL. Once that TTL has expired, when trying to
access vault secrets it will return a permission error, as the token is
no longer valid.

Our code is currently reactive, we watch when that happens, we refresh
the vault token, and try again.

That approach is not optimal as it requires handling the same error in
multiple places. We want to proactively refresh the token before it
expires in order to avoid the problem altogether.

Note that this is required because our integrations are long running and
our goal is to exceed the TTL.

We were previously assuming SecretAccessForbidden was a symptom of token
expiration. Therefore we were catching that exception, refreshing the
token, and raising a SecretNotFound to get the code to retry the
operation.

With the proactive token refresh we don't need to handle this. If the
vault client raises SecretAccessForbidden it's now likely a problem with
the RBAC of the client, so we don't want to hide it.

Related MRs:
#1414
#1480

### DIFF
--- a/reconcile/test/test_vault_utils.py
+++ b/reconcile/test/test_vault_utils.py
@@ -1,0 +1,45 @@
+import importlib
+import os
+import time
+
+import pytest
+
+from mock import patch, MagicMock
+
+import reconcile.utils.vault as vault
+
+
+class SleepCalled(Exception):
+    pass
+
+
+class testVaultClient(vault._VaultClient):  # pylint: disable=W0223
+    def __init__(self):  # pylint: disable=W0231
+        pass
+
+
+class TestVaultUtils:
+    @staticmethod
+    def test_vault_auto_refresh_env():
+        os.environ['VAULT_AUTO_REFRESH_INTERVAL'] = '1'
+        importlib.reload(vault)
+        assert vault.VAULT_AUTO_REFRESH_INTERVAL == 1
+
+    @staticmethod
+    def test_vault_auto_refresh_no_env():
+        del os.environ['VAULT_AUTO_REFRESH_INTERVAL']
+        assert os.getenv('VAULT_AUTO_REFRESH_INTERVAL') is None
+        importlib.reload(vault)
+        assert vault.VAULT_AUTO_REFRESH_INTERVAL == 600
+
+    @staticmethod
+    @patch.object(time, 'sleep')
+    def test_sleep_is_called(sleep):
+        sleep.side_effect = SleepCalled
+
+        testVaultClient._refresh_client_auth = MagicMock()
+
+        client = testVaultClient()
+
+        with pytest.raises(SleepCalled):
+            client._auto_refresh_client_auth()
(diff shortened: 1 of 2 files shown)


dockerfiles/hack/run-integration.py
```diff
@@ -55,8 +56,13 @@ def build_args():
         args.extend(INTEGRATION_EXTRA_ARGS.split())
     return args
 
+def refresh_vault():
+    while True:
+        VaultClient.refresh_client()
+        time.sleep(60*10)
 
 if __name__ == "__main__":
+    threading.Thread(target=refresh_vault, daemon=True).start()
```
maorfr, line 65: we should do that only if we are using vault: https://gitlab.cee.redhat.com/service/app-interface/-/blob/ae2715fdf47097afa020842697d02d2b7af72a9c/data/app-interface/app-interface-settings.yml#L11

apahim, line 65: This should be a feature of the VaultClient, imo. Callers should not have to worry about that.

jmelis:
@maorfr @apahim I've moved it inside VaultClient which addresses both of your concerns. Could you both review again?

apahim:
Beautiful ❤️
lgtm

apahim (approved):

reconcile/utils/vault.py
```diff
@@ -178,17 +192,9 @@ def read(self, secret):
         if kv_version == 2:
             data = self._read_v2(secret_path, secret_field, secret_version)
         else:
-            try:
-                data = self._read_v1(secret_path, secret_field)
-            except SecretAccessForbidden:
```
apahim, line 183: Yep, I was wondering if we still needed this. Tks for taking care of it.

maorfr (approved):
LGTM

i'd only add that it doesn't hurt to have both methods.

jmelis:
Thank you for your reviews @apahim @maorfr .

@maorfr:

> i'd only add that it doesn't hurt to have both methods.

But if the code in this PR works, this is not needed. I think that by removing unnecessary code, the result will be leaner.

If the doesn't work as expected, I'd like to see the errors and fix it.

maorfr:
in the case a token has a TTL under 10m and the env variable is not set - the integration would fail. although it has the potential to recover.

jmelis:
My assumptions is that our TTL is global, as it's tied to the approle we are using to run all integrations. Is this not right?

maorfr:
that is correct. i'm thinking anyone who may be using this but us (or, us in the future).

reconcile/test/test_vault_utils.py
```diff
@@ -0,0 +1,45 @@
...
+    @patch.object(time, 'sleep')
+    def test_sleep_is_called(sleep):
+        sleep.side_effect = SleepCalled
+
+        testVaultClient._refresh_client_auth = MagicMock()
+
+        client = testVaultClient()
+
+        with pytest.raises(SleepCalled):
+            client._auto_refresh_client_auth()
```
Piojo, line 45: You're missing an assertion about whether sleep was called here...

Piojo:
Thanks for making testing better for all of us.

PR #1482, opened 2021-04-05
APPSRE-3225: support multiple quays

jmelis:
Depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/17290

The `instance { url }` section is now mandatory for all quay_orgs (and `serverUrl` was never used). We will always pass the full api_url to the contructor.

### DIFF
--- a/reconcile/quay_base.py
+++ b/reconcile/quay_base.py
@@ -1,8 +1,11 @@
+from collections import namedtuple
+
 from reconcile.utils.secret_reader import SecretReader
 import reconcile.queries as queries
 
 from reconcile.utils.quay_api import QuayApi
 
+OrgKey = namedtuple('OrgKey', ['instance', 'org_name'])
 
 def get_quay_api_store():
     """
@@ -17,10 +20,10 @@ def get_quay_api_store():
     for org_data in quay_orgs:
         instance_name = org_data['instance']['name']
         org_name = org_data['name']
-        key = (instance_name, org_name)
+        org_key = OrgKey(instance_name, org_name)
         base_url = org_data['instance']['url']
         token = secret_reader.read(org_data['automationToken'])
-        store[key] = {
+        store[org_key] = {
             'api': QuayApi(token, org_name, base_url=base_url),
             'teams': org_data.get('managedTeams')
         }
--- a/reconcile/quay_base.py
+++ b/reconcile/quay_base.py
@@ -16,10 +16,10 @@ def get_quay_api_store():
     store = {}
     for org_data in quay_orgs:
         name = org_data['name']
-        server_url = org_data.get('serverUrl')
+        api_url = org_data['instance']['url']
         token = secret_reader.read(org_data['automationToken'])
         store[name] = {
-            'api': QuayApi(token, name, base_url=server_url),
+            'api': QuayApi(token, name, api_url=api_url),
             'teams': org_data.get('managedTeams')
         }
 
--- a/reconcile/quay_base.py
+++ b/reconcile/quay_base.py
@@ -15,11 +15,13 @@ def get_quay_api_store():
     secret_reader = SecretReader(settings=settings)
     store = {}
     for org_data in quay_orgs:
-        name = org_data['name']
+        instance_name = org_data['instance']['name']
+        org_name = org_data['name']
+        key = (instance_name, org_name)
         api_url = org_data['instance']['url']
         token = secret_reader.read(org_data['automationToken'])
-        store[name] = {
-            'api': QuayApi(token, name, api_url=api_url),
+        store[key] = {
+            'api': QuayApi(token, org_name, api_url=api_url),
             'teams': org_data.get('managedTeams')
         }
 
--- a/reconcile/quay_base.py
+++ b/reconcile/quay_base.py
@@ -7,6 +7,7 @@ from reconcile.utils.quay_api import QuayApi
 
 OrgKey = namedtuple('OrgKey', ['instance', 'org_name'])
 
+
 def get_quay_api_store():
     """
     Returns a dictionary with a key for each Quay organization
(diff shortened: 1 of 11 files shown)


reconcile/quay_base.py
```diff
@@ -16,10 +16,10 @@ def get_quay_api_store():
     store = {}
     for org_data in quay_orgs:
         name = org_data['name']
-        server_url = org_data.get('serverUrl')
+        api_url = org_data['instance']['url']
```
Piojo, line 19: The previous version of this doesn't assume `serverUrl` will exist. Where is it declared that either `instance` or `instance.url` will definitely exist?

Piojo (approved):
seal of approval

Piojo:
Sorry, but some tests are failing on the first commit. It looks like the mock test data doesn't have an `instance` even though we have discussed it is mandatory.

jmelis:
@Piojo oh! I run it locally and it passed, thanks for catching that, I'll patch the tests to include that. I'm still working on patching a couple of integrations to support multi-quay. Let me ping you once I'm finished to review again. Thanks for the seal :smile:

reconcile/quay_base.py
```diff
@@ -15,11 +15,13 @@ def get_quay_api_store():
     secret_reader = SecretReader(settings=settings)
     store = {}
     for org_data in quay_orgs:
-        name = org_data['name']
-        server_url = org_data.get('serverUrl')
+        instance_name = org_data['instance']['name']
+        org_name = org_data['name']
+        key = (instance_name, org_name)
```
maorfr, line 20: i like this approach

reconcile/quay_permissions.py
```diff
@@ -18,12 +18,20 @@
...
+        instance {
+          name
+        }
       }
       teams {
         permissions {
           service
-          ... on PermissionGithubOrgTeam_v1 {
-            org
+          ...on PermissionQuayOrgTeam_v1 {
```
maorfr, line 28: thanks!

reconcile/quay_permissions.py
```diff
@@ -50,14 +58,24 @@ def run(dry_run):
         if not quay_repo_configs:
             continue
         for quay_repo_config in quay_repo_configs:
+            instance_name = quay_repo_config['org']['instance']['name']
             org_name = quay_repo_config['org']['name']
-            quay_api = quay_api_store[org_name]['api']
+            org_key = (instance_name, org_name)
```
maorfr, line 63: i spent the entire review looking for this line!

maorfr (approved):
LGTM

the only thing that bothered me was the re-use of `org` as the org_key, but i understand the time constraints.

reconcile/quay_permissions.py
```diff
@@ -66,17 +84,28 @@ def run(dry_run):
...
                             continue
-                        if permission['org'] != org_name:
+
+                        perm_org_key = (
+                          permission['quayOrg']['instance']['name'],
+                          permission['quayOrg']['name']
+                        )
+
+                        if perm_org_key != org_key:
                             logging.warning('wrong org, ' +
```
Piojo, line 94: You don't need a `+` here:
```
>>> i=42
>>> 'Life, the universe and everything else are ' \
... f'{i}'
'Life, the universe and everything else are 42'
```

Piojo (approved):
Just a few nits. Don't bother acting on them, but they are probably useful for future iterations.

reconcile/quay_mirror.py
```diff
@@ -116,9 +122,10 @@ def process_sync_tasks(self):
 
         summary = self.process_repos_query()
         sync_tasks = defaultdict(list)
-        for org, data in summary.items():
+        for org_key, data in summary.items():
+            org = org_key[1]
```
apahim, line 126: Should we use a NamedTuple for the key? `org_key.org` is better than `org_key[1]`.

apahim (approved):
Lgtm

reconcile/quay_permissions.py
```diff
@@ -66,17 +84,28 @@ def run(dry_run):
...
-                                            f'should be {org_name}')
+                                            f'should be {org_key}')
```
apahim, line 95: You should use the `%s` plus args for logging messages, so the string only gets formatted when the message is actually generated.

reconcile/quay_membership.py
```diff
@@ -59,10 +81,11 @@ def fetch_desired_state():
...
     for role in result['roles']:
-        permissions = list(filter(
-            lambda p: p.get('service') == 'quay-membership',
-            role['permissions']
-        ))
+        permissions = [
+            process_permission(p)
+            for p in role['permissions']
+            if p.get('service') == 'quay-membership'
+        ]
```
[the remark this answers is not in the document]
jmelis: agreed

reconcile/quay_permissions.py
jmelis: fwiw this predates this PR, however I still haven't got a hang on multiline strings, sometimes I enclose them in parenthesis so it allows me to split them into multiple lines. I also dislike the `+` because it creates more objects in memory, vs the interpreter creating a single string

reconcile/quay_membership.py
```diff
@@ -78,7 +101,8 @@ def append_quay_username_members(member):
             for bot in role['bots']:
                 append_quay_username_members(bot)
 
-            list(map(lambda p: state.add(p, members), permissions))
+            for permission in permissions:
```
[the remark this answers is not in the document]
jmelis: @Piojo I like that recommendation, can you submit a MR to our style guide?

fixing

reconcile/quay_permissions.py
jmelis: fixed

reconcile/quay_mirror.py
jmelis: fixed

PR #1486, opened 2021-04-06
APPSRE-3207: Support `use_channel_in_image_tag`

2uasimojo:
Previously some consumers were using `IMAGE_TAG` as only *part* of the actual image tag referenced by a saas template, e.g.:

`  image: ${REGISTRY_IMG}:${CHANNEL}-${IMAGE_TAG}`

This was clearly wrong and misleading.

With this change, if the saas file/template specifies the parameter:

`  USE_CHANNEL_IN_IMAGE_TAG: "true"` (note: `"true"`, not `true`)

...then we will calculate the value of `IMAGE_TAG` to include the channel; so the above example can be changed to:

`  image: ${REGISTRY_IMG}:${IMAGE_TAG}`

This also makes it possible for `REPO_DIGEST` to be used by such consumers. They would specify:

```
parameters:
- name: REGISTRY_IMG
  required: true
- name: CHANNEL
  required: true
- name: USE_CHANNEL_IN_IMAGE_TAG
  value: "true"
- name: IMAGE_TAG
  required: true
- name: REPO_DIGEST
  required: true

...

        image: ${REPO_DIGEST}
```

APPSRE-3207

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -411,10 +411,27 @@ class SaasHerder():
                     f"error fetching template: {str(e)}")
                 return None, None, None
 
+            # add IMAGE_TAG only if it is required and unspecified
             if self._parameter_value_needed(
                     "IMAGE_TAG", consolidated_parameters, template):
-                # add IMAGE_TAG only if it is required
-                image_tag = commit_sha[:hash_length]
+                sha_substring = commit_sha[:hash_length]
+                # IMAGE_TAG takes one of two forms:
+                # - If saas file attribute 'use_channel_in_image_tag' is true,
+                #   it is {CHANNEL}-{SHA}
+                # - Otherwise it is just {SHA}
+                if self._get_saas_file_attribute("use_channel_in_image_tag"):
+                    try:
+                        channel = consolidated_parameters["CHANNEL"]
+                    except KeyError:
+                        logging.error(
+                            f"[{saas_file_name}/{resource_template_name}] "
+                            + f"{html_url}: CHANNEL is required when "
+                            + "'use_channel_in_image_tag' is true."
+                        )
+                        return None, None, None
+                    image_tag = f"{channel}-{sha_substring}"
+                else:
+                    image_tag = sha_substring
                 consolidated_parameters['IMAGE_TAG'] = image_tag
 
             # Do this in a separate loop, since it relies on IMAGE_TAG already
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -334,6 +334,28 @@ class SaasHerder():
                         'could not add html_url annotation to' +
                         resource['name'])
 
+    @staticmethod
+    def _parameter_value_needed(
+            parameter_name, consolidated_parameters, template):
+        """Is a parameter named in the template but unspecified?
+
+        NOTE: This is currently "parameter *named* and absent" -- i.e. we
+        don't care about `required: true`. This is for backward compatibility.
+
+        :param parameter_name: The name (key) of the parameter.
+        :param consolidated_parameters: Dict of parameters already specified/
+                calculated.
+        :param template: The template file in dict form.
+        :return bool: True if the named parameter is named in the template,
+                but not already present in consolidated_parameters.
+        """
+        if parameter_name in consolidated_parameters:
+            return False
+        for template_parameter in template.get("parameters", {}):
+            if template_parameter["name"] == parameter_name:
+                return True
+        return False
+
     def _process_template(self, options):
         saas_file_name = options['saas_file_name']
         resource_template_name = options['resource_template_name']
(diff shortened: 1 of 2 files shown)


2uasimojo:
@jharrington22 @maorfr

2uasimojo:
Note to reviewers: Two commits here. Easier to review separately.

jharrington22:
The additional `USE_CHANNEL_IN_IMAGE_TAG` parameter in the template allows consumers to be specific about whether or not channel is used in the digest fetch which keeps this change to just SREP or specific consumers (can't be added by mistake).

Long term when we switch to OPM builds for operators the catalog image channels go away. 

/lgtm

maorfr (changes requested):
thanks again for this contribution!

this PR adds behavior that is based on parameters. it would be better to control behavior through the fields in the saas file itself rather then the parameters.

i'd look at `compare` as an example (search app-interface for `compare`).

jharrington22:
> thanks again for this contribution!
> 
> this PR adds behavior that is based on parameters. it would be better to control behavior through the fields in the saas file itself rather then the parameters.
> 
> i'd look at `compare` as an example (search app-interface for `compare`).

@maorfr this is similar to `compare` no? Its switching on the digest request to use the channel parameter as part of the image name. Its not an ideal solution but given the time restrictions we don't see a cleaner way. Do you have a suggestion?

2uasimojo:
> it would be better to control behavior through the fields in the saas file

Done, I think. How does that look?

jharrington22:
> > thanks again for this contribution!
> > this PR adds behavior that is based on parameters. it would be better to control behavior through the fields in the saas file itself rather then the parameters.
> > i'd look at `compare` as an example (search app-interface for `compare`).
> 
> @maorfr thanks for the feedback! This is similar to `compare` no? Its switching on the digest request to use the channel parameter as part of the image name. Its not an ideal solution but given the time restrictions we don't see a cleaner way. Do you have a suggestion?

I see move the control into the saas file. Sorry @maorfr I understand now don't mind me.

reconcile/utils/saasherder.py
```diff
@@ -387,48 +406,58 @@ def _process_template(self, options):
...
+
+            # add IMAGE_TAG only if it is required and unspecified
+            if self._parameter_required_and_absent(
+                    "IMAGE_TAG", consolidated_parameters, template):
+                sha_substring = commit_sha[:hash_length]
+                # IMAGE_TAG takes one of two forms:
+                # - If saas file attribute 'use_channel_in_image_tag' is true,
+                #   it is {CHANNEL}-{SHA}
+                # - Otherwise it is just {SHA}
+                if self._get_saas_file_attribute("use_channel_in_image_tag"):
```
maorfr, line 419: You'll need to add this field to the graphql query and set it in the saasherder init I believe?
It would also require a schema change in app-interface to support this change.

reconcile/utils/saasherder.py
```diff
@@ -334,6 +334,25 @@ def _additional_resource_process(resources, html_url):
...
+                calculated.
+        :param template: The template file in dict form.
+        :return bool: True if the named parameter is required by the template,
+                but not already present in consolidated_parameters.
+        """
+        if parameter_name in consolidated_parameters:
+            return False
+        for template_parameter in template.get("parameters", {}):
+            if template_parameter["name"] == parameter_name:
+                return template_parameter.get("required", False)
```
maorfr, line 353: This is new logic. We can't rely on the IMAGE_TAG being required, but we still want to generate it. Can we keep this to a separate PR?

2uasimojo: Reverted the logic and left a TODO.

reconcile/utils/saasherder.py
2uasimojo: Found the query thing. Will look for the schema.

2uasimojo: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/17433 for the schema and documentation change. I suppose that should be merged with/after promoting this PR.

reconcile/utils/saasherder.py
```diff
@@ -453,7 +488,7 @@ def _process_template(self, options):
                 logging.error(
                     f"[{url}/{path}:{target_ref}] " +
                     f"error fetching directory: {str(e)}")
-                return None, None
+                return None, None, None
```
maorfr, line 491: Why is this needed?

2uasimojo: Prevents these error conditions from raising `ValueError: not enough values to unpack (expected 3, got 2)` from the [caller](https://github.com/app-sre/qontract-reconcile/blob/035460ca61d130aab188db6359c602db1b2bf92d/reconcile/utils/saasherder.py#L682-L683).

2uasimojo: (That's the "latent buglet" referenced by the [first commit](https://github.com/app-sre/qontract-reconcile/pull/1486/commits/aa54ee96618804f0f2991e044efd973719154348))

maorfr: OH NICE

reconcile/utils/saasherder.py
```diff
@@ -334,6 +334,31 @@ def _additional_resource_process(resources, html_url):
...
                         resource['name'])
 
+    @staticmethod
+    def _parameter_required_and_absent(
+            parameter_name, consolidated_parameters, template):
+        """Is a parameter is required in the template but unspecified?
+
+        TODO: This is currently "parameter *named* and absent" -- i.e. we
+        don't care about `required: true`. This is for backward compatibility;
+        but should eventually be fixed.
```
maorfr, line 344: i'm not sure this statement is true. i think it's enough to name the parameter. most services just use the value `latest` and not make it required, which helps them deploy the template in their development process.

reconcile/utils/saasherder.py
maorfr: :heart:

maorfr (approved):
otherwise for that comment, LGTM

this is very nice work. very involved and you are doing it effortlessly and with very little hints.
kudos!

reconcile/utils/saasherder.py
2uasimojo: I renamed the method and took out the TODOs.

(Nitpick: If they specify the value, `required` is redundant/unnecessary and wouldn't trigger this function anyway. But after poking around, I think I grok why it will *never* be possible to track down all the template files in existence and make sure they've specified `required: true` for `IMAGE_TAG`.)

maorfr: EXACTLY! if we can't automate it, we can't enforce it.

PR #1495, opened 2021-04-09
More immutable-type errors

rporres:
The case we're specifically addressing here is making a Service
headless. We need to set the clusterIP to either None or "" (empty).
Kubernetes doesn't allow for clusterIP to be unset or modified, hence we
need to delete the Service and recreate it again.

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -31,6 +31,14 @@ class FieldIsImmutableError(Exception):
     pass
 
 
+class MayNotChangeOnceSetError(Exception):
+    pass
+
+
+class PrimaryClusterIPCanNotBeUnsetError(Exception):
+    pass
+
+
 class MetaDataAnnotationsTooLongApplyError(Exception):
     pass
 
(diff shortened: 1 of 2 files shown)


rporres:
cc @jhernand

jhernand:
Looks good to me, thanks @rporres.

apahim:
Can you please give more context in the commit message?

apahim:
I mean, what's the problem we are facing?

rporres:
> Can you please give more context (ideally in the commit message)?

You're totally right. Done!

apahim:
Beatiful, thank you!

apahim (approved):
lgtm

reconcile/utils/oc.py
```diff
@@ -591,8 +599,15 @@ def _run(self, cmd, **kwargs):
             if kwargs.get('apply'):
                 if 'Invalid value: 0x0' in err:
                     raise InvalidValueApplyError(f"[{self.server}]: {err}")
-                if 'Invalid value: ' in err and ': field is immutable' in err:
-                    raise FieldIsImmutableError(f"[{self.server}]: {err}")
+                if 'Invalid value: ' in err:
```
erdii, line 602: What happens if this codebranch is taken but none of the child-ifs get executed? should this if have a corresponding else?

rporres: it goes through the `StatusCodeError` condition

erdii: i see!
lgtm then :)

PR #1496, opened 2021-04-09
option to disable management of quay repos

arilivigni:
This is to add the `managedRepos` option to allow to enable or disable management of quay.io repos

### DIFF
--- a/reconcile/quay_repos.py
+++ b/reconcile/quay_repos.py
@@ -16,6 +16,7 @@ QUAY_REPOS_QUERY = """
     quayRepos {
       org {
         name
+        managedRepos
         instance {
           name
         }
@@ -73,6 +74,9 @@ def fetch_desired_state():
             continue
 
         for quay_repo in quay_repos:
+            if not quay_repo['org']['managedRepos']:
+                continue
+
             instance_name = quay_repo['org']['instance']['name']
             org_name = quay_repo['org']['name']
             org_key = (instance_name, org_name)
@@ -183,7 +187,7 @@ class RunnerAction:
 
 
 def run(dry_run):
-    quay_api_store = get_quay_api_store()
+    quay_api_store = get_quay_api_store(managed_repos=True)
 
     current_state = fetch_current_state(quay_api_store)
     desired_state = fetch_desired_state()
--- a/reconcile/quay_base.py
+++ b/reconcile/quay_base.py
@@ -8,7 +8,7 @@ from reconcile.utils.quay_api import QuayApi
 OrgKey = namedtuple('OrgKey', ['instance', 'org_name'])
 
 
-def get_quay_api_store():
+def get_quay_api_store(managed_repos=False):
     """
     Returns a dictionary with a key for each Quay organization
     managed in app-interface.
@@ -19,6 +19,8 @@ def get_quay_api_store():
     secret_reader = SecretReader(settings=settings)
     store = {}
     for org_data in quay_orgs:
+        if managed_repos and not org_data['managedRepos']:
+            continue
         instance_name = org_data['instance']['name']
         org_name = org_data['name']
         org_key = OrgKey(instance_name, org_name)
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -935,6 +935,7 @@ QUAY_ORGS_QUERY = """
 {
   quay_orgs: quay_orgs_v1 {
     name
+    managedRepos
     instance {
       name
       url


reconcile/test/fixtures/quay_repos/desired_state_simple.yml
```diff
@@ -14,13 +15,13 @@ gql_response:
         public: false
 state:
 - params:
-    org: [my_instance, org1]
+    org: [true, my_instance, org1]
```
BumbleFeng, line 18: remove this

reconcile/test/fixtures/quay_repos/desired_state_simple.yml
```diff
@@ -14,13 +15,13 @@ gql_response:
...
     repo: repo1
   items:
   - description: desc1
     public: true
 - params:
-    org: [my_instance, org1]
+    org: [true, my_instance, org1]
```
BumbleFeng, line 24: remove this

reconcile/quay_repos.py
```diff
@@ -183,7 +187,7 @@ def action(params, items):
 
 
 def run(dry_run):
-    quay_api_store = get_quay_api_store()
+    quay_api_store = get_quay_api_store(managed_repos=True)
```
apahim, line 190: Why do we need this `managed_repos=True`?

arilivigni: @jmelis thought this was the better option and we will update the 6 files on the app-interface side.  Why is there a concern?

BumbleFeng (approved):

reconcile/quay_repos.py
apahim: This is your code, not Jaimes's. I want to understand why do you think this is a good idea.

apahim:
Please don't merge PRs with pending questions.

apahim:
Btw, stage is broken now.

reconcile/quay_repos.py
arilivigni: I think the default should be true that repos are managed if they are in app-interface.  I actually peered with Jaime and I agree this is the right solution.  I mean maybe you can shed some light on your concern, which is missing from your original question.

apahim: When I commented this, there was no app-interface counterpart, so it was just a suspicion. Now there is, and my suspicion confirmed. You made `managedRepos` required there, you made it `True` here and you made it `False` in the function signature. So, again, do we really need this `managed_repos` argument?

apahim: Let me try to be more clear, I have that time now. This code works. That doesn't mean it's doing the right thing.

You're using a `managed_repos` argument to creep into the `get_quay_api_store()` behaviour, making it not initialize all the orgs defined in app-interface, but you do that only when you use a specific integration AND when the org has `managedRepos` set to `false`.

The problem is that we might need the `QuayApi` instances for something else in the future, and they will not be there. Someone else, probably not you, will have to figure out why and also a way to make it do the right thing then.

It's all about well defined interfaces and consistent behaviour. To have well defined interface and consistent behaviour here, the `get_quay_api_store()` should always give you what you expect from it, regarless the integration using it, as defined in the docstring:

```python 
def get_quay_api_store():
    """
    Returns a dictionary with a key for each Quay organization
    managed in app-interface.
    Each key contains an initiated QuayApi instance.
    """
```

So, instead of skipping the initialization of the client for certain orgs when using certain integrations under certain conditions, one option is to just add the `managedRepos` information to the `store` dictionary returned by the `get_quay_api_store`.

Then, the right place to process that `managedRepos` information is where you need it. In this case, in the `fetch_current_state()`, just like you did with the `fetch_desired_state()`. There, you don't add to the current state any orgs that have the `managedRepos` property set to `False`.

I'd love to understand why you think that having the `managed_repos` added to the `get_quay_api_store` is the right thing to do instead. Maybe you can shed some light on your ideas, which is missing from your responses.

Btw, sorry for the short and minimalistic message in the beginning of this discussion. I tend to expect that my comments will ring the right bells, not always the case.

jmelis: @apahim thanks for you review. Maybe I haven't interpreted your message correctly but I think this might be a naming / documentation issue. Consider this diff:

```diff
diff --git a/reconcile/quay_base.py b/reconcile/quay_base.py
index 560df6c..b7f9dc0 100644
--- a/reconcile/quay_base.py
+++ b/reconcile/quay_base.py
@@ -8,18 +8,21 @@ from reconcile.utils.quay_api import QuayApi
 OrgKey = namedtuple('OrgKey', ['instance', 'org_name'])
 
 
-def get_quay_api_store(managed_repos=False):
+def get_quay_api_store(skip_unmanaged_repos=False):
     """
     Returns a dictionary with a key for each Quay organization
     managed in app-interface.
     Each key contains an initiated QuayApi instance.
+
+    :param skip_unmanaged_repos: if true, it will not initialize organizations
+      that have `managedRepos: False`. If ommitted it will return all orgs.
     """
     quay_orgs = queries.get_quay_orgs()
     settings = queries.get_app_interface_settings()
     secret_reader = SecretReader(settings=settings)
     store = {}
     for org_data in quay_orgs:
-        if managed_repos and not org_data['managedRepos']:
+        if skip_unmanaged_repos and org_data['managedRepos'] is False:
             continue
         instance_name = org_data['instance']['name']
         org_name = org_data['name']
diff --git a/reconcile/quay_repos.py b/reconcile/quay_repos.py
index b14a7d5..fa825bb 100644
--- a/reconcile/quay_repos.py
+++ b/reconcile/quay_repos.py
@@ -187,7 +187,7 @@ class RunnerAction:
 
 
 def run(dry_run):
-    quay_api_store = get_quay_api_store(managed_repos=True)
+    quay_api_store = get_quay_api_store(skip_unmanaged_repos=True)
 
     current_state = fetch_current_state(quay_api_store)
     desired_state = fetch_desired_state()
```

I think this would be the right naming to use, plus the docstring clarifies it. Note that we are not changing the default behaviour anywhere and that a new client would receive the expected orgs (unless they intentionally use `skip_unmanaged_repos`.

To be honest though, I agree that this is a bit cutting corners. The reason I initially suggested going this path is to not invest time on the quay-repos integration, since it's one of the first integrations we wrote, and we used a the `AggregatedList` class, which was a design failure: complex and bloated.

All that said, I like a lot your proposal of putting `managedRepos` information in the store itself, that makes a lot of sense.

apahim: Hey @jmelis, thanks for clarifying. I see where you're coming from.
Indeed, documentation would help a lot, but the effort is pretty much the same, and since we seem to be in the same page regarding the alternative I proposed, plus this is already merged, let's wrap up here and let me send a PR with that for you guys to review.
Thank you!

jmelis: that makes a lot of sense, thanks @apahim!

PR #1498, opened 2021-04-11
[terraform-resources] Allow AWS IAM service accounts access to cluster's AWS infrastructure via OCM

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3234

this PR adds the option to define `aws_infrastructure_access` for an IAM service account created with terraform-resources.
this is to allow such users the network-mgmt role via the AWS infrastructure access feature of OCM.

this is done to enable Hive to access it's own cluster's AWS account to perform PrivateLink related operations (https://github.com/openshift/aws-account-operator/pull/539).

the final result of all this is that the IAM user created by terraform-resources will be granted AWS infrastructure access through ocm-aws-infrastructure-access. once that access is granted, a `role_arn` key will be added to the output Secret created by terraform-resources, as required for section (3) of https://github.com/openshift/hive/blob/master/docs/awsprivatelink.md#permissions-required-for-aws-private-link.

### DIFF
--- a/reconcile/ocm_aws_infrastructure_access.py
+++ b/reconcile/ocm_aws_infrastructure_access.py
@@ -1,8 +1,10 @@
 import logging
 
+import reconcile.utils.gql as gql
 import reconcile.queries as queries
 
 from reconcile.utils.ocm import OCMMap
+from reconcile.terraform_resources import TF_NAMESPACES_QUERY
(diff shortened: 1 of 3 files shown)


maorfr:
happy to get post-merge reviews as i want to push this forward!

PR #1499, opened 2021-04-11
[terraform-resources] don't write to vault if no resources were applied

maorfr:
we currently write all output secrets on every iteration of the integrations.

with this PR, we will skip it in case no resources were applied. this should speed up the integrations significantly.

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -396,12 +396,12 @@ def run(dry_run, print_only=False,
     # or for pods to be recycled
     # this should be removed after we gained confidence
     # following the terraform 0.13 upgrade
-    ob.realize_data(dry_run, oc_map, ri)
+    actions = ob.realize_data(dry_run, oc_map, ri)
 
     disable_keys(dry_run, thread_pool_size,
                  disable_service_account_keys=True)
 
-    if vault_output_path:
+    if actions and vault_output_path:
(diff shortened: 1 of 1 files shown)


apahim (approved):
great catch

reconcile/terraform_resources.py
```diff
@@ -396,12 +396,12 @@ def run(dry_run, print_only=False,
...
     # this should be removed after we gained confidence
     # following the terraform 0.13 upgrade
-    ob.realize_data(dry_run, oc_map, ri)
+    actions = ob.realize_data(dry_run, oc_map, ri)
 
     disable_keys(dry_run, thread_pool_size,
                  disable_service_account_keys=True)
 
-    if vault_output_path:
+    if actions and vault_output_path:
```
rporres, line 404: Since we first pass through all the accounts and then we write the outputs, can the following happen?

* The integration runs when it has to do stuff in two different accounts.
* The first update in account `a` succeeds (something that will trigger a secret update)
* The second update in account `b` fails (terraform + AWS madness)
* Then we fix stuff because second update was not fesible (it happened recently, when we tried to shrink a volume).
* In the next iteration there's no change, hence no action and no secret write

maorfr: this is a viable concern and we should address it right when we address only writing the changed the changed secrets and not all of them IMO.

PR #1500, opened 2021-04-11
Move the managedRepos processing to where we need it

apahim:
The `managedRepos` is an option of interest for the quay-repos
integration.

Instead of changing the behaviour of the `get_quay_api_store()` acording
to the integration using it, let' s just make sure it ships enough
information so the quay-repos integration knows what to do.

Then, in the quay-repos intregration, we test the `managedRepos` in the
`fetch_current_state()`, the same way we do for the `fetch_desired_state()`.


jmelis (approved):
LGTM

arilivigni (approved):

PR #1501, opened 2021-04-12
[template/fluentd] do not forward "Certificate did not match expected hostname" messages

maorfr:
we are seeing these messages from time to time in our OpenShift related integrations:
https://coreos.slack.com/archives/CS0E65QCV/p1618200378271100

our integrations are resilient to these errors and it seems there is not much we can do with them.
this PR will cleanup the forwarded logs (slack, cloudwatch).


jfchevrette (approved):
The change LGTM but we should really try and find out why this is happening. It also probably affects other services and not just our integrations

maorfr:
I've revived https://issues.redhat.com/browse/APPSRE-1564

jmelis:
I agree with @jfchevrette. I would rather receive alerts about this than hide them.

maorfr:
#sd-app-sre-reconcile is not alerts :)

PR #1502, opened 2021-04-12
[terraform-tgw-attachments] add support for global routing

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3228

with this PR, we can now manage routes in TGWs that are peered to the TGWs we discover. this is required as AWS does not support cross-region route propagation and we want global routing (if a request goes to some region, it will be routed to the proper TGW and from there to the hive VPC).


maorfr:
again, happy to get review in retrospective!

PR #1504, opened 2021-04-12
update limits/requests for aws-[support|iam]

skryzhny:


maorfr:
You need to make the same change in helm/external-values

maorfr:
And LGTM

PR #1507, opened 2021-04-15
Use the CIS .metadata.name as the target tag

jmelis:
The tag in the releaseImage has been changed from a tag to a digest.
This has caused the mirroring of the ocp-release repo to use the digest
as the tags.

Instead of using the digest as the tag, we are changing the logic to
pick up the CIS name as the tag name for destination ocp release.

### DIFF
--- a/reconcile/ocp_release_mirror.py
+++ b/reconcile/ocp_release_mirror.py
@@ -3,6 +3,7 @@ import logging
 import sys
 
 from urllib.parse import urlparse
+from collections import namedtuple
 
 from sretoolbox.container import Image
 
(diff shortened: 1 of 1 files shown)


reconcile/ocp_release_mirror.py
```diff
@@ -20,6 +21,8 @@
 
 LOG = logging.getLogger(__name__)
 
+OcpReleaseInfo = namedtuple('OcpReleaseInfo', ['ocp_release', 'tag'])
```
apahim, line 24: :heart:

apahim (approved):
lgtm

PR #1508, opened 2021-04-15
rewrite quay-repos without AggregatedList

jmelis:


reconcile/test/test_quay_repos.py
```diff
@@ -1,91 +1,56 @@
...
-    for group in state:
-        this_h = AggregatedList.hash_params(group['params'])
-
-        if h == this_h:
-            return sorted(group['items'])
-    return False
-
+def build_state(fixture_state):
+    return [
+        RepoInfo(OrgKey('instance', 'org'), item[0], item[1], item[2])
```
apahim, line 14: you can increase readability here using
```
RepoInfo(org_key=OrgKey('instance', 'org'), name=item[0], public=item[1], description=item[2])
```

apahim (approved):
beatiful change, thank you for taking the time

reconcile/test/test_quay_repos.py
jmelis: +1

PR #1509, opened 2021-04-15
[MTSRE-36] gitlabRepoOwners now has the branch name

apahim:
gitlabRepoOwners used to be a boolean, but now it brings the branch name
to compare the MR against, as some repos are migrating from "master" to
"main".

app-interface counterpart: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/17914

### DIFF
--- a/reconcile/gitlab_owners.py
+++ b/reconcile/gitlab_owners.py
@@ -299,8 +299,8 @@ class MRApproval:
 
 
 def act(repo, dry_run, instance, settings):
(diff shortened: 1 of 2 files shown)


arilivigni (approved):

maorfr:
can the automation detect the default branch from the gitlab cli?

apahim:
good point, will check that out. tks.

maorfr:
would #1536 work?

PR #1510, opened 2021-04-15
Support for downstream quay org repo management

jmelis:
This PR only addresses the repo management piece (not the content mirroring).

App-Interface counterpart:
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/17927

### DIFF
--- a/reconcile/quay_repos.py
+++ b/reconcile/quay_repos.py
@@ -45,7 +45,7 @@ def fetch_current_state(quay_api_store):
     state = []
 
     for org_key, org_info in quay_api_store.items():
-        if not org_info['managedRepos']:
+        if not org_info['managedRepos'] and not org_info['mirror']:
             continue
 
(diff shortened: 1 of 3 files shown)


reconcile/quay_repos.py
```diff
@@ -45,7 +45,7 @@ def fetch_current_state(quay_api_store):
     state = []
 
     for org_key, org_info in quay_api_store.items():
-        if not org_info['managedRepos']:
+        if not org_info['managedRepos'] and not org_info['mirror']:
```
apahim, line 48: should this and corresponding condition in the `fetch_desired_state()` be the same?

apahim: :thinking:

apahim: never mind

apahim (approved):
lgtm

PR #1513, opened 2021-04-16
update osd-mirrors-data with ocm-quay

BumbleFeng:

### DIFF
--- a/reconcile/osd_mirrors_data_updater.py
+++ b/reconcile/osd_mirrors_data_updater.py
@@ -130,16 +130,21 @@ def run(dry_run, gitlab_project_id=None):
     # install-config.yaml compatible format
     mirrors_info = []
     for tfr in osd_mirrors:
-        image_url = get_image_uri(
-            aws_cli=aws_cli,
-            account=tfr['account'],
-            repository=tfr['identifier']
-        )
+        # ecr image is not in use for now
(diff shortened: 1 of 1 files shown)


reconcile/osd_mirrors_data_updater.py
```diff
@@ -130,16 +130,21 @@ def run(dry_run, gitlab_project_id=None):
...
-        )
+        # ecr image is not in use for now
+        # image_url = get_image_uri(
+        #     aws_cli=aws_cli,
+        #     account=tfr['account'],
+        #     repository=tfr['identifier']
+        # )
+        quayio_url = tfr['mirror']['url']
+        ocmquay_url = quayio_url.\
+            replace('quay.io', 'pull.q1w2.quay.rhcloud.com')
```
apahim, line 141: This integration goes through the ECR resources and get their quay.io mirrors.
I'm afraid that just changing the URL is too fragile. A new org in quay.io won't make this mirror include that org in the list.

apahim: I mean, since ECR is out of the game now, we should completly skip that aws_cli logic.

maorfr:
no longer required: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/26494

PR #1514, opened 2021-04-16
Add support for cloudfront S3 public key to terraform-resources

rrati:
Depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/18029


BumbleFeng:
I think we can also include creating cloudfront_key_group https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_key_group

rrati:
@BumbleFeng I was planning to add that, but in the aws module we are using 3.22.0, cloudfront_key_group doesn't exist, not does the terrascript version we are using support it. :(

Existing naming conventions for the cloudfront all use s3 with it.  The function name should include cloudfront though.  I'll add that

rrati:
@BumbleFeng Updated the function name to include cloudfront

BumbleFeng (approved):

PR #1515, opened 2021-04-19
promote sretoolbox

maorfr:
promotes sretoolbox to version 0.13.0, which includes https://github.com/app-sre/sretoolbox/pull/35


maorfr:
promoting in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/18070

PR #1516, opened 2021-04-19
[openshift-saas-deploy] skip empty promotion

maorfr:
When an Image is not found, `promotion` returns as None, which we do not handle.

This PR changes the behavior to skip validating a `None` promotion. The integration will still exit with an error code and the promotion itself will be published as failed (unable to promote to it, as expected).

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -923,6 +923,8 @@ class SaasHerder():
         If there were promotion sections in the participating saas files
         validate that the conditions are met. """
         for item in self.promotions:
+            if item is None:
+                continue
             # validate that the commit sha being promoted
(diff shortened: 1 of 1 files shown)


maorfr:
it mostly cleans up the exception in cases when the image is not found, such as https://ci.int.devshift.net/job/openshift-saas-deploy-saas-pagerduty-operator-osd-stage-hives02ue1/31/console

maorfr:
related to https://github.com/app-sre/sretoolbox/pull/35, in which we only expect to see the "Image not found" error message and not the promotion validation stack trace.

2uasimojo (commented):
Cool, thanks for this. That stack trace was a tad confusing.

PR #1520, opened 2021-04-19
[saas-file-owners] allow to self-service use_channel_in_image_tag

maorfr:
our integrations are covering our bases very well. if `use_channel_in_image_tag` is used, our integrations are checking all cases for all different targets.

the merge strategy for `use_channel_in_image_tag` is (by @2uasimojo):

The use_channel_in_image_tag attribute needs to be rolled out in
stages. This is because it causes the generated IMAGE_TAG to change
from $hash to $channel-$hash. This is incompatible with older
versions of the in-repo OLM bundle, which specify the image with tag
${CHANNEL}-${IMAGE_TAG}: this would become e.g.
staging-staging-abc1234 if we just landed use_channel_in_image_tag
by itself, resulting in broken builds.
The stages are as follows:

Freeze staging/integration to the hash of the current master (prod is
already frozen). Supply the IMAGE_TAG parameter explicitly,
corresponding to the frozen hash.
Land the in-repo change to cut over to $REPO_DIGEST

Revert the staging/integration parts of step #1. This
should unfreeze staging/integration and cause them to start using
$REPO_DIGEST properly.
During the prod promotion that gets us to/past the commit from #2
above, remove the IMAGE_TAG parameter from the prod envs. This should
cause prod to start using $REPO_DIGEST properly.

### DIFF
--- a/reconcile/saas_file_owners.py
+++ b/reconcile/saas_file_owners.py
@@ -195,11 +195,13 @@ def valid_diff(current_state, desired_state):
     for c in current_state_copy:
         c.pop('ref')
         c.pop('parameters')
+        c['saas_file_definitions'].pop('use_channel_in_image_tag')
         c.pop('disable', None)
     desired_state_copy = copy.deepcopy(desired_state)
     for d in desired_state_copy:
         d.pop('ref')
         d.pop('parameters')
+        d['saas_file_definitions'].pop('use_channel_in_image_tag')
         d.pop('disable', None)
     return current_state_copy == desired_state_copy
 


BumbleFeng (approved):

reconcile/saas_file_owners.py
```diff
@@ -195,11 +195,13 @@ def valid_diff(current_state, desired_state):
     for c in current_state_copy:
         c.pop('ref')
         c.pop('parameters')
+        c['saas_file_definitions'].pop('use_channel_in_image_tag')
```
2uasimojo, line 198: Does this need to be
```suggestion
        c['saas_file_definitions'].pop('use_channel_in_image_tag', None)
```
to avoid raising an exception for saas files where this attribute is absent?

2uasimojo:
```
[efried@efried boilerplate]$ python3
Python 3.6.8 (default, Aug 18 2020, 08:33:21) 
[GCC 8.3.1 20191121 (Red Hat 8.3.1-5)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> d = dict(one=2, three=4)
>>> d.pop('five')
KeyError: 'five'
>>> d.pop('five', None)
>>> 
```

maorfr: it's always there: https://github.com/app-sre/qontract-reconcile/blob/b16e1a014f7942e7cd57f928313be71eac724f25/reconcile/saas_file_owners.py#L65

2uasimojo: Okay.

PR #1526, opened 2021-04-21
[kafka-clusters] add support for multi_az and service accounts

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2728

for a consumer to use a cluster, they need a service account. this PR adds support for creating a service account per cluster (with the cluster name being the service account name) and returns the credentials for that service account in the output Secret to be consumed by applications.

### DIFF
--- a/reconcile/kafka_clusters.py
+++ b/reconcile/kafka_clusters.py
@@ -78,6 +78,7 @@ def run(dry_run, thread_pool_size=10,
     desired_state = {}
     current_state = ocm_map.kafka_cluster_specs()
     desired_state = fetch_desired_state(kafka_clusters)
+    kafka_service_accounts = ocm_map.kafka_service_account_specs()
 
     error = False
     for kafka_cluster in kafka_clusters:
@@ -108,11 +109,26 @@ def run(dry_run, thread_pool_size=10,
         if current_cluster['status'] != 'ready':
             continue
         # we have a ready cluster!
+        # let's get/create a service account for it
+        # we match cluster to service account by name
+        service_accounts = [sa for sa in kafka_service_accounts
+                            if sa['name'] == kafka_cluster_name]
+        if service_accounts:
+            service_account = service_accounts[0]
+        else:
+            service_account = {}
+            logging.info(['create_service_account', kafka_cluster_name])
+            if not dry_run:
+                ocm = ocm_map.get(kafka_cluster_name)
+                service_account = \
+                    ocm.create_kafka_service_account(kafka_cluster_name)
+        
         # let's create a Secret in all referencing namespaces
         kafka_namespaces = kafka_cluster['namespaces']
         secret_fields = ['bootstrapServerHost']
         data = {k: v for k, v in current_cluster.items()
                 if k in secret_fields}
+        data.update(service_account)
         resource = construct_oc_resource(data)
         for namespace_info in kafka_namespaces:
             ri.add_desired(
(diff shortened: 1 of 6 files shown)


reconcile/kafka_clusters.py
```diff
@@ -102,14 +119,15 @@ def run(dry_run, thread_pool_size=10,
             error = True
             continue
         # check if cluster is ready. if not - wait
-        if current_cluster['status'] != 'complete':
+        if current_cluster['status'] != 'ready':
```
Piojo, line 122: Can we convert these magic strings into constants? A mistake on a future iteration, (say, I'm pushing this during an outage and I type `rready`) will waste hours of debugging.

reconcile/utils/ocm.py
```diff
@@ -633,12 +633,35 @@ def get_kafka_clusters(self, fields=None):
...
+    def get_kafka_service_accounts(self, fields=None):
+        """Returns details of the Kafka service accounts """
+        results = []
+        api = '/api/managed-services-api/v1/serviceaccounts'
+        service_accounts = self._get_json(api)['items']
+        for sa in service_accounts:
+            sa_id = sa['id']
+            id_api = f'{api}/{sa_id}'
+            sa_details = self._get_json(id_api)
+            item = {k: v for k, v in sa_details.items() if k in fields}
```
Piojo, line 645: What happens if `fields` has its default value of `None`? Won't this crash?

reconcile/utils/ocm.py
```diff
@@ -633,12 +633,35 @@ def get_kafka_clusters(self, fields=None):
...
+
     def create_kafka_cluster(self, data):
         """Creates (async) a Kafka cluster """
         api = '/api/managed-services-api/v1/kafkas'
         params = {'async': 'true'}
         self._post(api, data, params)
 
+    def create_kafka_service_account(self, name):
+        """ Creates a Kafka service account """
+        api = '/api/managed-services-api/v1/serviceaccounts'
```
Piojo, line 657: This looks like a constant we can share across the entire class, doesn't it?

reconcile/kafka_clusters.py
```diff
@@ -75,10 +76,26 @@ def run(dry_run, thread_pool_size=10,
...
 
     error = False
     for kafka_cluster in kafka_clusters:
         kafka_cluster_name = kafka_cluster['name']
+        # get a service account for the cluster
+        # we match cluster to service account by name
+        service_accounts = [sa for sa in kafka_service_accounts
+                            if sa['name'] == kafka_cluster_name]
+        if service_accounts:
+            service_account = service_accounts[0]
```
Piojo, line 89: Please, please, please, let's avoid variables that differ from just one character. As we get tired we'll confuse them.

In this context, `sa` or `acc` are perfectly valid identifiers.

reconcile/kafka_clusters.py
maorfr: done

reconcile/utils/ocm.py
maorfr: good catch! fixed

reconcile/utils/ocm.py
maorfr: i usually like copying the api directly from https://api.openshift.com, but we can go with constants.

reconcile/kafka_clusters.py
maorfr: ok!

maorfr:
note: more changes are about to be added to handle storing the clientSecret in vault.

maorfr:
image

maorfr:
thanks for the review @Piojo! this is ready for another round.

maorfr:
as I want this promoted before my week ends, i'm going to merge this. please feel free to provide a post-merge review and i promise to address everything. ask @rporres, i'm good for it.

PR #1533, opened 2021-04-25
[openshift-saas-deploy] always generate IMAGE_TAG

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3265

we currently only generate IMAGE_TAG if it is a parameter in the openshift template and it is not explicitly specified in the saas file parameters. there are cases where we need to use the image tag to generate other parameters, such as REPO_DIGEST.

generally, since there are no real operations to generate the image tag, we can just always do it.

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -411,9 +411,9 @@ class SaasHerder():
                     f"error fetching template: {str(e)}")
                 return None, None, None
 
-            # add IMAGE_TAG only if it is required and unspecified
-            if self._parameter_value_needed(
-                    "IMAGE_TAG", consolidated_parameters, template):
+            # add IMAGE_TAG only if it is unspecified
+            image_tag = consolidated_parameters.get('IMAGE_TAG')
+            if not image_tag:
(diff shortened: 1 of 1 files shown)


reconcile/utils/saasherder.py
```diff
@@ -441,12 +441,11 @@ def _process_template(self, options):
                 try:
```
2uasimojo, line 441: The comment on L437 could use some help. (Latent; I should have fixed it when I factored out _parameter_value_needed.)

2uasimojo (commented):
/lgtm

reconcile/utils/saasherder.py
maorfr: what would you change it to? (also, it's Sunday).

2uasimojo: Probably just:

```python
            # This relies on IMAGE_TAG already being calculated.
```

maorfr: done

apahim (approved):
lgtm

PR #1536, opened 2021-04-25
[gitlab-owners] use default branch

maorfr:
replaces #1509?


apahim (approved):
Makes sense. There's no reason to compare against a non-default branch. Thanks!

PR #1541, opened 2021-04-27
Support generating IMAGE_DIGEST

2uasimojo:
Today referencing an image with:

`  image: ${REPO_DIGEST}`

requires designating `REGISTRY_IMG` as a required parameter even if it is not explicitly used anywhere in the template. To make things more scrutable, this commit adds support for generating a new `IMAGE_DIGEST` value, which comprises just the digest (`sha256:xxxx...`) whereas `REPO_DIGEST` comprises the full by-digest URL. With this change, the template file can include:

```
parameters:
- name: REGISTRY_IMG
  required: true
- name: IMAGE_DIGEST
  required: true

...

    image: ${REGISTRY_IMG}@${IMAGE_DIGEST}
```

To wit, you're explicitly using all the parameters you require.

APPSRE-3265

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -435,8 +435,11 @@ class SaasHerder():
                 consolidated_parameters['IMAGE_TAG'] = image_tag
 
             # This relies on IMAGE_TAG already being calculated.
-            if self._parameter_value_needed(
-                    "REPO_DIGEST", consolidated_parameters, template):
+            need_repo_digest = self._parameter_value_needed(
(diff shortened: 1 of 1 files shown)


maorfr (approved):
LGTM

The explicitness makes total sense. I now understand what you were going for.

Thanks!

maorfr:
I'd appreciate a look (and merge) from @apahim when he is awake as I am out for the rest of the week.

apahim:
lgtm. will merge&promote first thing tomorrow.

PR #1543, opened 2021-05-04
Add service-owners-for-rds-instance cli command

rporres:

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -492,6 +492,27 @@ def quay_mirrors(ctx):
     print_output(ctx.obj['output'], mirrors, columns)
 
 
+@get.command()
+@click.argument('aws_account')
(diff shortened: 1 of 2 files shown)


tools/qontract_cli.py
```diff
@@ -492,6 +492,27 @@ def quay_mirrors(ctx):
     print_output(ctx.obj['output'], mirrors, columns)
 
 
+@get.command()
+@click.argument('aws_account')
+@click.argument('db_instance_name')
```
maorfr, line 497: identifier? just to make it consistent with app-interface?

maorfr (approved):

tools/qontract_cli.py
rporres: fixed in 82485ad :)

PR #1546, opened 2021-05-07
Add extension to jinja to raise errors

rporres:


rporres:
This is the kind of things I want to do with this change: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/19047

PR #1550, opened 2021-05-11
add support for group association by org username

maorfr:
this change affects the following integrations:
- openshift-groups
- ocm-groups
- openshift-users

if a cluster is defined as `internal`, instead of using the github username, we use the org username.

this PR is done in the context of https://issues.redhat.com/browse/APPSRE-3261, where we create an internal OSD cluster.

### DIFF
--- a/reconcile/openshift_groups.py
+++ b/reconcile/openshift_groups.py
@@ -13,11 +13,13 @@ ROLES_QUERY = """
   roles: roles_v1 {
     name
     users {
+      org_username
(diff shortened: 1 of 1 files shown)


maorfr:
```
['add_user_to_group', 'appsres03ue1', 'dedicated-admins', 'mafriedm']
['del_user_from_group', 'appsres03ue1', 'dedicated-admins', 'maorfr']
```

PR #1558, opened 2021-05-13
[saasherder] change parameters not reused log from warning to error

maorfr:
openshift-saas-deploy collects parameters for deployment from 2 places:
1. saas file parameters (top level, resource template level, target level)
2. environment parameters

if a parameter is defined with values that can be reused from the environment level, we currently log a warning message.

this PR changes the behavior so that we log an error message and exit with error instead, which means we will enforce reusing parameters.

this is an alternative for #1557

output example: https://ci.int.devshift.net/job/service-app-interface-gl-pr-check/61084/artifact/temp/reports/reconcile_reports_success/reconcile-saas-file-validator.txt


rporres:
I like this one more than #1557

Piojo:
Indeed. If this matters, enforcing it makes all the sense.

PR #1560, opened 2021-05-13
[ocm-clusters] improve cluster update and add validation

maorfr:
this PR changes the way we update a cluster via OCM.
first, we validate that any updated fields are able to be updated in OCM. second, we only pass the updated fields to the clusters PATCH endpoint instead of passing all fields that can be updated, even if they were not updated.

this really solves 2 problems:
1. the integration will return an error when trying to update an invalid field.
2. assuming we want to update cluster storage quota (for example) during an upgrade. in the current situation, since we are passing the `channel` field as well, the cluster update will be rejected. with this change, only updated fields will be passed.

### DIFF
--- a/reconcile/ocm_clusters.py
+++ b/reconcile/ocm_clusters.py
@@ -24,6 +24,42 @@ def fetch_desired_state(clusters):
     return desired_state
 
 
+def get_cluster_update_spec(cluster_name, current_spec, desired_spec):
+    """ Get a cluster spec to update. Returns an error if diff is invalid """
+    allowed_spec_update_fields = [
+        'instance_type',
+        'storage',
+        'load_balancers',
+        'private',
+        'channel',
+        'autoscale',
+        'nodes'
+    ]
+    error = False
+    if current_spec['network'] != desired_spec['network']:
+        error = True
+        logging.error(f'[{cluster_name}] invalid update: network')
+    current_spec_spec = current_spec['spec']
+    desired_spec_spec = desired_spec['spec']
+    updated = {k: desired_spec_spec[k] for k in desired_spec_spec
+               if current_spec_spec[k] != desired_spec_spec[k]}
+
+    # we only need deleted to check if a field removal is valid
+    # we really want to check updated + deleted, and since
+    # we have no further use for deleted -
+    deleted = {k: current_spec_spec[k] for k in current_spec_spec
+               if k not in desired_spec_spec}
+    diffs = deleted
+    diffs.update(updated)
+
+    for k in diffs:
+        if k not in allowed_spec_update_fields:
+            error = True
+            logging.error(f'[{cluster_name}] invalid update: {k}')
+
+    return updated, error
+
+
 def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
     settings = queries.get_app_interface_settings()
     clusters = queries.get_clusters()
@@ -95,6 +131,15 @@ def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
                 desired_spec['spec'].pop(k, None)
 
             if current_spec != desired_spec:
+                # check if cluster update is valid
+                update_spec, err = get_cluster_update_spec(
+                    cluster_name,
+                    current_spec,
+                    desired_spec,
+                )
+                if err:
+                    error = True
+                    continue
                 # update cluster
                 logging.debug(
                     '[%s] desired spec %s is different ' +
@@ -104,7 +149,7 @@ def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
                 # TODO(mafriedm): check dry_run in OCM API patch
                 if not dry_run:
                     ocm = ocm_map.get(cluster_name)
-                    ocm.update_cluster(cluster_name, desired_spec, dry_run)
+                    ocm.update_cluster(cluster_name, update_spec, dry_run)
         else:
             # create cluster
             if cluster_name in pending_state:
--- a/reconcile/ocm_clusters.py
+++ b/reconcile/ocm_clusters.py
@@ -26,7 +26,7 @@ def fetch_desired_state(clusters):
 
 def get_cluster_update_spec(cluster_name, current_spec, desired_spec):
     """ Get a cluster spec to update. Returns an error if diff is invalid """
-    allowed_spec_update_fields = [
+    allowed_spec_update_fields = {
         'instance_type',
         'storage',
         'load_balancers',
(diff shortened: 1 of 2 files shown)


reconcile/ocm_clusters.py
```diff
@@ -24,6 +24,42 @@ def fetch_desired_state(clusters):
...
+        'private',
+        'channel',
+        'autoscale',
+        'nodes'
+    ]
+    error = False
+    if current_spec['network'] != desired_spec['network']:
+        error = True
+        logging.error(f'[{cluster_name}] invalid update: network')
+    current_spec_spec = current_spec['spec']
```
Piojo, line 42: `current_spec_spec` is making me dizzy...

reconcile/ocm_clusters.py
```diff
@@ -24,6 +24,42 @@ def fetch_desired_state(clusters):
     return desired_state
 
 
+def get_cluster_update_spec(cluster_name, current_spec, desired_spec):
+    """ Get a cluster spec to update. Returns an error if diff is invalid """
+    allowed_spec_update_fields = [
```
Piojo, line 29: Make this a set

reconcile/ocm_clusters.py
```diff
@@ -24,6 +24,42 @@ def fetch_desired_state(clusters):
...
+
+    # we only need deleted to check if a field removal is valid
+    # we really want to check updated + deleted, and since
+    # we have no further use for deleted -
+    deleted = {k: current_spec_spec[k] for k in current_spec_spec
+               if k not in desired_spec_spec}
+    diffs = deleted
+    diffs.update(updated)
+
+    for k in diffs:
```
Piojo, line 55: Easier if `allowed_spec_update_fields` is a set:

```python
invalid_fields = set(diffs.keys()) & allowed_spec_update_fields
error = True
logging.error(...)
```

reconcile/ocm_clusters.py
```diff
@@ -24,6 +24,42 @@ def fetch_desired_state(clusters):
...
+        logging.error(f'[{cluster_name}] invalid update: network')
+    current_spec_spec = current_spec['spec']
+    desired_spec_spec = desired_spec['spec']
+    updated = {k: desired_spec_spec[k] for k in desired_spec_spec
+               if current_spec_spec[k] != desired_spec_spec[k]}
+
+    # we only need deleted to check if a field removal is valid
+    # we really want to check updated + deleted, and since
+    # we have no further use for deleted -
+    deleted = {k: current_spec_spec[k] for k in current_spec_spec
```
Piojo, line 50: You're not using the values of this dictionary at all, why not have this a set of keys?

```python
deleted = {k for k in current_spec_spec...}
```

reconcile/ocm_clusters.py
```diff
@@ -24,6 +24,42 @@ def fetch_desired_state(clusters):
...
+               if k not in desired_spec_spec}
+    diffs = deleted
+    diffs.update(updated)
+
+    for k in diffs:
+        if k not in allowed_spec_update_fields:
+            error = True
```
Piojo, line 57: Why not raise `ValueError` instead? This boolean flag feels clunky

reconcile/ocm_clusters.py
```diff
@@ -95,6 +131,15 @@ def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
                 desired_spec['spec'].pop(k, None)
 
             if current_spec != desired_spec:
+                # check if cluster update is valid
+                update_spec, err = get_cluster_update_spec(
```
Piojo, line 135: This is better served with a `try-except`.

reconcile/utils/ocm.py
```diff
@@ -171,12 +171,12 @@ def create_cluster(self, name, cluster, dry_run):
 
         self._post(api, ocm_spec, params)
 
-    def update_cluster(self, name, cluster, dry_run):
+    def update_cluster(self, name, cluster_spec, dry_run):
```
Piojo, line 174: I preferred the `cluster` identifier, but no biggie.

Piojo:
... and I know I'm a painful reviewer, but can we start adding unit tests too? :)

Piojo (changes requested):

Piojo:
Also, one more question, would this barf at my change of the `multi_az` field?

maorfr:
@Piojo it will, as i don't think you can change a cluster from single az to multi az without recreating it.

reconcile/ocm_clusters.py
maorfr: we are using them further down the road when actually updating the cluster

reconcile/ocm_clusters.py
maorfr: "easier" is subjective :)

reconcile/ocm_clusters.py
maorfr: it's some golang-like error handling which we do from time to time instead of raising, catching, logging error, continuing.

reconcile/ocm_clusters.py
maorfr: "better" is subjective :)

reconcile/ocm_clusters.py
maorfr: why would a set be better?

reconcile/ocm_clusters.py
maorfr: me too! i'll take anything else you have to offer

reconcile/ocm_clusters.py
Piojo: What you do below logicallly the difference of the keys of your dictionary and the allowed keys (my first revision stated intersection because I was on a hurry and sleep deprived). You can express the operation with one line of code instead of a loop and an `if`. Then, the cost of the entire difference is O(N) (or O(N*log N), depending on internal implementation details). The cost of each an `in` for a list is O(N), which makes the loop below O(N^2).

I don't care much about speed here, but when you add the reduced code size and the readability, sets are totally the way to go here.

reconcile/ocm_clusters.py
Piojo: That's my point, golang style is best left for golang.

reconcile/ocm_clusters.py
Piojo: Sets are easy. Unless you wonder whether the set of all sets that contain themselves is a member of the set of all sets that contain themselves. ;)

Piojo:
I'm not going to block this on the golang vs pythonic style mix, but I do have hard feelings and a strong love for sets.

reconcile/ocm_clusters.py
maorfr: Russel's set?

maorfr:
changed to a set!

maorfr:
i'm going to let this merge. feel free to test it against your MR changing the multiaz by updating the `.env` file.

PR #1561, opened 2021-05-14
Make jenkins-webhooks a StatefulSet

rporres:


rporres:
This hasn't been tested yet

maorfr (approved):
LGTM as is

just a nit - i usually prefer something more abstract like `cache: true` instead of templating out specific implementation (like `Deployment`). good to merge as is, up to you.

rporres:
I like both approaches, I don't think any good argument against yours except of reworking the files 😂  Since you like one more, I'll go with it 😉

rporres:
It's true that it looks cleaner and nicer :-D

PR #1567, opened 2021-05-18
[terraform-resources] add crendentials file like secret

maorfr:
this will allow using this file-like secret as a mounted volume in pods. this is another option to use the aws credentials chain.

in our case, this key is added to the `instance` secret in openshift-logging namespaces where we use log forwarding and we want to override it to enable https://issues.redhat.com/browse/APPSRE-3284

### DIFF
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -2682,11 +2682,19 @@ class TerrascriptClient:
         # outputs
         # aws_access_key_id
         output_name_0_13 = output_prefix + '__aws_access_key_id'
-        output_value = '${' + tf_resource.id + '}'
-        tf_resources.append(Output(output_name_0_13, value=output_value))
+        output_value_id = '${' + tf_resource.id + '}'
+        tf_resources.append(Output(output_name_0_13, value=output_value_id))
         # aws_secret_access_key
         output_name_0_13 = output_prefix + '__aws_secret_access_key'
-        output_value = '${' + tf_resource.secret + '}'
+        output_value_secret = '${' + tf_resource.secret + '}'
+        tf_resources.append(
+            Output(output_name_0_13, value=output_value_secret))
+        # aws credentials file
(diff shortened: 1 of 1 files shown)


reconcile/utils/terrascript_client.py
```diff
@@ -2682,11 +2682,19 @@ def get_tf_iam_access_key(self, user_tf_resource,
         # outputs
         # aws_access_key_id
         output_name_0_13 = output_prefix + '__aws_access_key_id'
-        output_value = '${' + tf_resource.id + '}'
-        tf_resources.append(Output(output_name_0_13, value=output_value))
+        output_value_id = '${' + tf_resource.id + '}'
```
Piojo, line 2685: f-strings are more readable...
```python
output_value_id = f"${{tf_resource.id}}'
```

reconcile/utils/terrascript_client.py
```diff
@@ -2682,11 +2682,19 @@ def get_tf_iam_access_key(self, user_tf_resource,
...
+        tf_resources.append(Output(output_name_0_13, value=output_value_id))
         # aws_secret_access_key
         output_name_0_13 = output_prefix + '__aws_secret_access_key'
-        output_value = '${' + tf_resource.secret + '}'
+        output_value_secret = '${' + tf_resource.secret + '}'
```
Piojo, line 2689: As above:
```python
output_value_secret = f"${{tf_resource.secret}}"
```

Piojo (approved):
A couple of nits, nothing serious.

reconcile/utils/terrascript_client.py
maorfr: i'd say this is out of scope? there a lot of these in our code base

Piojo: Fine by me, I've approved the changes nevertheless.

rporres (approved):
d'oh, i forgot to give you the +1 when I read this

maorfr:
reverted in #1574

PR #1578, opened 2021-05-25
Quay membership must fail on missing users

rporres:
Nobody cares about warning messages, only about failed PR checks

### DIFF
--- a/reconcile/quay_membership.py
+++ b/reconcile/quay_membership.py
@@ -1,4 +1,5 @@
 import logging
+import sys
 
 import reconcile.utils.gql as gql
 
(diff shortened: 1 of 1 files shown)


maorfr (approved):
LGTM

assuming you tested this locally, since i don't understand the exact behavior of `AggregatedDiffRunner`.

maybe @jmelis also wants to take a look?

rporres:
Yes, I tested it locally. I had my issues understanding that class too. It was going to be your typical 5 minutes fix and then... 😂 

Assigning this to @jmelis , the father of the creature

jmelis:
I will never stop apologizing to the universe for creating that class.

That said: @rporres if there is one missing user, you don't create any user whatsoever and exit. Other integrations apply whatever they can, and then they exit. I assume the reason why you did this is because the first place you are going to catch this is during PR checks with --dry-run, so it doesn't really matter in that scenario, but I think I would prefer to have it work the other way for 1) consistency 2) in case a user is deleted on the fly accidentally.

rporres:
hmmm, we don't create the quay users with this integration, we just create the memberships. Having said that, I can address your (valid) concerns

jmelis (approved):
LGTM

PR #1582, opened 2021-05-25
aws_api: paginate when initializing the list of users

jfchevrette:
The call to `list_users` need to be paginated. At the moment only the first 100 users is returned as this is the default for `iam.list_users()`

Once the fix is deployed it is possible that other functionality that relied on the list of users may trigger changes or errors which were not caught before as we were not evaluating the users after the first 100. After review the following functionality may be impacted:
- Verification of IAM keys
- Deletion of IAM users
- AWS resource owners (resources without owners)

### DIFF
--- a/reconcile/utils/aws_api.py
+++ b/reconcile/utils/aws_api.py
@@ -67,7 +67,8 @@ class AWSApi:
         self.users = {}
         for account, s in self.sessions.items():
             iam = s.client('iam')
-            users = [u['UserName'] for u in iam.list_users()['Users']]
+            users = self.paginate(iam, 'list_users', 'Users')
+            users = [u['UserName'] for u in users]
             self.users[account] = users
 
     def simulate_deleted_users(self, io_dir):


reconcile/utils/aws_api.py
```diff
@@ -67,7 +67,8 @@ def init_users(self):
         self.users = {}
         for account, s in self.sessions.items():
             iam = s.client('iam')
-            users = [u['UserName'] for u in iam.list_users()['Users']]
+            users = self.paginate(iam, 'list_users', 'Users')
+            users = [u['UserName'] for u in users]
```
rporres, line 71: nit: why redefining the `users` variable instead of directly doing
```
users = [u['UserName'] for u in self.paginate(iam, 'list_users', 'Users')]
```

rporres (commented):
Apart from the nit-picking :-D this looks fine. Have you run the affected integrations in dry-run mode to have an idea of the scope of the change?

reconcile/utils/aws_api.py
jfchevrette: I thought redefining users (or another var) is clearer whereas executing the function inside the list comprehension is a bit harder on the eye. also line length (can be split of course)

i'll leave it as it is for the sake of clarity and readability

maorfr (approved):
LGTM

PR #1583, opened 2021-05-25
saasherder: disallow using same parameter name in value

jfchevrette:
I found this corner case while investigating an odd behavior reported by a customer where the parameter value in a saas-file was being replaced by itself, and not the value from the environment file. This is caused by the fact the parameter name is being used in it's own value with the expectation that the value will be the one from the environment file.

# Situation
Environment file
```yaml
---
$schema: /app-sre/environment-1.yml

# ... snip ...

parameters:
  THE_HOST: 'example.com'
```

saas-file
```yaml
---
$schema: /app-sre/saas-file-1.yml

# ... snip ...

resourceTemplates:
- name: myapp
  path: ...
  url: ...
  targets:
  - namespace:
      $ref: ...
    ref: ...
    parameters:
      THE_HOST: 'www.${THE_HOST}
```

Given the above example, when saasherder tried to [parse and render the parameters](https://github.com/app-sre/qontract-reconcile/blob/4e21b455f3f82ae61e5e44131db64214545b1b50/reconcile/utils/saasherder.py#L376-L397) it consolidates the environment, the resource parameters and target parameters (in that order) it ends up overwriting same-name variables from most generic definitions (environment) with the most local ones (target parameters). In many cases this is desired as the users re-define variables for using in specific environment.

In the above example, not only the variable is redefined with the same name but it is also referencing itself with the expectation that the same-name variable from the most generic definitions (environment) will be used. However this doesn't work in the current implementation because we effectively squash all into a single consolidated dict.

If we were to change the order by which the parameters are consolidated we would break the precedence of environment -> resource -> target 

This fix is a bit dumb, but by preventing the usage of a variable within its own value this should cover this undesired side-effect without disrupting breaking existing precedence expectations

As a workaround until we have a fix I have requested that the customer renames the parameter in the target in order to avoid this conflict.


maorfr:
perhaps it would be better to implement this in `_validate_saas_files` instead. with this addition, we will get a conflict.

jfchevrette:
> perhaps it would be better to implement this in `_validate_saas_files` instead. with this addition, we will get a conflict.

I have moved the check to the `_validate_saas_files` as sugested. This makes more sense. Can you PTAL again?

I have tested this locally with `saas-file-validator`

```
❯ qontract-reconcile --config config.debug.toml --dry-run saas-file-validator
[2021-05-26 13:30:18] [INFO] [gql.py:init_from_config:222] - using gql endpoint http://localhost:4000/graphqlsha/c58eccd87034ac2b6d4e315403465245fb69b9e580119c10aa02c7d6528b6558
[2021-05-26 13:30:18] [ERROR] [saasherder.py:_validate_saas_files:120] - [myservice/myservice] recursivity in parameter name and value found: PLATFORM_HOSTNAME: cert.${PLATFORM_HOSTNAME} - this will likely not work as expected. Please consider changing the parameter name
```

maorfr (approved):
LGTM

This is a very good solution to the problem IMO. Thanks for doing this!

PR #1584, opened 2021-05-26
desolve vpc peerings dependency on awsInfrastructureAccess

maorfr:
in order to create a VPC peering through app-interface, the `awsInfrastructureAccess` section has to be merged in advance for terraform to get access to the cluster's aws account.

with this PR, the peering section and the access section can be submitted in the same MR, and the vpc peerings integration will log warnings until the access is granted.


Piojo:
While this is very illustrative, I really don't want to modify existing, working logic with deep `try-except` blocks. Let's find a way that doesn't require that much copy-paste.

PR #1587, opened 2021-05-27
[openshift-saas-deploy] add slack start and end notifications

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3187

this PR adds slack start and end notifications. this functionality replaces the functionality of the Jenkins Slack publisher for v2 saas files. it also adds a console url to open to view the pipelines run.

### DIFF
--- a/reconcile/openshift_saas_deploy.py
+++ b/reconcile/openshift_saas_deploy.py
@@ -20,9 +20,12 @@ QONTRACT_INTEGRATION = 'openshift-saas-deploy'
 QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
 
 
-def slack_notify_result(saas_file_name, env_name, slack, ri):
+def slack_notify(saas_file_name, env_name, slack, ri, in_progress):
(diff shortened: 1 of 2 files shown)


reconcile/openshift_saas_deploy.py
```diff
@@ -30,6 +59,35 @@ def run(dry_run, thread_pool_size=10, io_dir='throughput/',
         logging.error('no saas files found')
         sys.exit(ExitCodes.ERROR)
 
+    # notify different outputs (publish results, slack notifications)
+    # we only do this if:
+    # - this is not a dry run
+    # - there is a single saas file deployed
```
jmelis, line 65: why is this?

maorfr: this integration can run either for all saas files or for a single one. each saas file has a different slack section, so we can't initialize one for all. so we only use this when a single saas file is deployed.

maorfr:
merging to promote to migrate some more :)

PR #1588, opened 2021-05-27
[openshift-saas-deploy-trigger-configs] remove trigger loop

maorfr:
now that this integration runs as a pod, we don't need to enter a loop and trigger at any cost. if an error occurs, it may be fixed in the next iteration. this way we will not be blocked on our problematic trigger and will still trigger other jobs.


maorfr:
going to merge this one since this is currently a bug in production

PR #1593, opened 2021-05-31
[openshift-saas-deploy-trigger-upstream-jobs] introduce integration

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3321

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/19991

this PR adds a new integration - `openshift-saas-deploy-trigger-upstream-jobs` which triggers deployments based on successful builds of jobs defined in jenkins. this is to allow to link a build job with a stage deployment job for example.

### DIFF
--- a/reconcile/openshift_saas_deploy_trigger_upstream_jobs.py
+++ b/reconcile/openshift_saas_deploy_trigger_upstream_jobs.py
@@ -1,0 +1,61 @@
+import sys
+import logging
+
+import reconcile.openshift_saas_deploy_trigger_base as osdt_base
+import reconcile.queries as queries
+
+from reconcile.status import ExitCodes
+from reconcile.utils.defer import defer
+from reconcile.utils.semver_helper import make_semver
+
+
+QONTRACT_INTEGRATION = 'openshift-saas-deploy-trigger-upstream-jobs'
+QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
+
+
+def fetch_current_state(jenkins_map):
+    return {instance_name: jenkins.get_jobs_state()
+            for instance_name, jenkins in jenkins_map.items()}
+
+
+@defer
+def run(dry_run, thread_pool_size=10, internal=None,
+        use_jump_host=True, defer=None):
+    saas_files = queries.get_saas_files(v1=False, v2=True)
+    if not saas_files:
+        logging.error('no saas files found')
+        sys.exit(ExitCodes.ERROR)
+
+    setup_options = {
+        'saas_files': saas_files,
+        'thread_pool_size': thread_pool_size,
+        'internal': internal,
+        'use_jump_host': use_jump_host,
+        'integration': QONTRACT_INTEGRATION,
+        'integration_version': QONTRACT_INTEGRATION_VERSION,
+    }
+    saasherder, jenkins_map, oc_map, settings = osdt_base.setup(setup_options)
+    defer(lambda: oc_map.cleanup())
+
+    current_state = fetch_current_state(jenkins_map)
+    trigger_specs = saasherder.get_upstream_jobs_diff(dry_run, current_state)
+    already_triggered = []
+    error = False
+    for job_spec in trigger_specs:
+        trigger_options = {
+            'dry_run': dry_run,
+            'spec': job_spec,
+            'jenkins_map': jenkins_map,
+            'oc_map': oc_map,
+            'already_triggered': already_triggered,
+            'settings': settings,
+            'state_update_method': saasherder.update_upstream_job,
+            'integration': QONTRACT_INTEGRATION,
+            'integration_version': QONTRACT_INTEGRATION_VERSION,
+        }
+        trigger_error = osdt_base.trigger(trigger_options)
+        if trigger_error:
+            error = True
+
+    if error:
+        sys.exit(ExitCodes.ERROR)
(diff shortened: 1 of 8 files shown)


reconcile/openshift_saas_deploy_trigger_upstream_jobs.py
```diff
@@ -0,0 +1,61 @@
...
+            'spec': job_spec,
+            'jenkins_map': jenkins_map,
+            'oc_map': oc_map,
+            'already_triggered': already_triggered,
+            'settings': settings,
+            'state_update_method': saasherder.update_upstream_job,
+            'integration': QONTRACT_INTEGRATION,
+            'integration_version': QONTRACT_INTEGRATION_VERSION,
+        }
+        trigger_error = osdt_base.trigger(trigger_options)
```
Piojo, line 56: I gotta insist on changing the convention calling to `trigger` to use named arguments. Less code, less bugs, and they can be caught by a run of the linter instead of at runtime.

reconcile/utils/saasherder.py
```diff
@@ -859,6 +859,91 @@ def update_moving_commit(self, job_spec):
             f"{namespace_name}/{env_name}/{ref}"
         self.state.add(key, value=commit_sha, force=True)
 
+    def get_upstream_jobs_diff(self, dry_run, current_state):
+        results = threaded.run(self.get_upstream_jobs_diff_saas_file,
+                               self.saas_files,
+                               self.thread_pool_size,
+                               dry_run=dry_run,
+                               current_state=current_state)
+        return [item for sublist in results for item in sublist]
```
Piojo, line 868: I find `itertools.chain` easier. According to stackoverflow it's also faster.

```python
return itertools.chain.from_iterable(results)
```
(it returns a generator, wrap it onto a `list` if you truly want to have a list)

Piojo: But this is just a nit.

reconcile/utils/saasherder.py
```diff
@@ -859,6 +859,91 @@ def update_moving_commit(self, job_spec):
...
+
+    def get_upstream_jobs_diff_saas_file(self, saas_file, dry_run,
+                                         current_state):
+        saas_file_name = saas_file['name']
+        timeout = saas_file.get('timeout') or None
```
Piojo, line 873: `get` returns `None` if the value doesn't exist. Unless you want 0 replaced with `None`?

reconcile/utils/saasherder.py
```diff
@@ -859,6 +859,91 @@ def update_moving_commit(self, job_spec):
...
+        pipelines_provider = self._get_pipelines_provider(saas_file)
+        trigger_specs = []
+        for rt in saas_file['resourceTemplates']:
+            rt_name = rt['name']
+            for target in rt['targets']:
```
Piojo, line 878: Can we move the body of this loop into its own method for the sake of the tests to come?

reconcile/utils/saasherder.py
```diff
@@ -859,6 +859,91 @@ def update_moving_commit(self, job_spec):
...
+                    if not dry_run:
+                        self.state.add(key, value=last_build_result)
+                    continue
+
+                state_build_result_number = state_build_result['number']
+                for build_result in job_history:
+                    # this is the most important condition
+                    # if there is a successful newer build -
+                    # trigger the deployment ONCE.
+                    if build_result['number'] > state_build_result_number \
```
Piojo, line 913: As above, if this becomes its own method we can test it in isolation.

reconcile/utils/saasherder.py
```diff
@@ -859,6 +859,91 @@ def update_moving_commit(self, job_spec):
...
+                            'timeout': timeout,
+                            'pipelines_provider': pipelines_provider,
+                            'rt_name': rt_name,
+                            'cluster_name': cluster_name,
+                            'namespace_name': namespace_name,
+                            'instance_name': instance_name,
+                            'job_name': job_name,
+                            'last_build_result': last_build_result,
+                        }
+                        trigger_specs.append(job_spec)
```
Piojo, line 928: Let's aim for named arguments, sample PR coming in a few minutes.

reconcile/utils/saasherder.py
maorfr: not with graphql it doesn't :)

reconcile/openshift_saas_deploy_trigger_upstream_jobs.py
maorfr: please take a look around the code base to understand the extent of this form of usage. i suggest to not insist on specific PRs, and instead submit a proposal for the team (or; anyone interested) to discuss.

let's not hold new functionality for things that are "new".

reconcile/utils/saasherder.py
maorfr: this is common around this class and should be handled once for all repetitions. feel free to take a look around and suggest better approaches through PRs.

reconcile/utils/saasherder.py
maorfr: this is not bad. there are a lot of places in the code base that can use this change!

reconcile/utils/saasherder.py
maorfr: maybe i'm confused here

reconcile/utils/saasherder.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/1751

PR #1597, opened 2021-06-02
[APPSRE-3325]Introduce code coverage in tox configs

Piojo:
To get the tests ball rolling. We seem to be using tox for testing, so
I adapt its configuration to generate reports.

The minimum required coverage is 19%. We currently have 19.07%
coverage.

I also disable the `test_make` file, which doesn't work at all on my
system, even without any local changes to my git repository.

### DIFF
--- a/reconcile/test/test_make.py
+++ b/reconcile/test/test_make.py
@@ -1,5 +1,7 @@
 import subprocess
 
+import pytest
+
 
 def run_make(sub_command):
     cmd = ['make', sub_command]
@@ -17,6 +19,9 @@ def has_uncommited_changes():
 
 class TestMake:
     @staticmethod
+    @pytest.mark.skip(
+        "Broken - uncommitted changes are normal during development"
+        )
     def test_make_generate():
         assert not has_uncommited_changes(), ('No uncommited changes must '
                                               'exists')
--- a/reconcile/test/test_make.py
+++ b/reconcile/test/test_make.py
@@ -1,3 +1,4 @@
+import os
 import subprocess
 
 import pytest
@@ -19,9 +20,10 @@ def has_uncommited_changes():
 
 class TestMake:
     @staticmethod
-    @pytest.mark.skip(
-        "Broken - uncommitted changes are normal during development"
-        )
+    @pytest.mark.skipif(
+        not os.environ.get("HUDSON_HOME"),
+        reason="This test is only for CI environments",
+    )
     def test_make_generate():
         assert not has_uncommited_changes(), ('No uncommited changes must '
                                               'exists')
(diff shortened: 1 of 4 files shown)


reconcile/test/test_make.py
```diff
@@ -17,6 +19,9 @@ def has_uncommited_changes():
 
 class TestMake:
     @staticmethod
+    @pytest.mark.skip(
+        "Broken - uncommitted changes are normal during development"
+        )
```
maorfr, line 24: i'd advise not to disable this. this test is here due to past issues with folks not generating the openshift templates using `make generate`.

Piojo:
> i'd advise not to disable this. this test is here due to past issues
> with folks not generating the openshift templates using `make
> generate`.

I was sure there was a good reason for this test. But it doesn't even
pass on master.

If we want to run it as part of a pipeline, we may use a `skipIf` gate
instead, but then, how do you recommend to identify that we are running
in a CI environment?

maorfr:
perhaps there is an environment variable we can use that exists when running in jenkins?

tox.ini
```diff
@@ -5,10 +5,23 @@ envlist = py36
 commands =
     flake8 reconcile tools e2e_tests
     pylint -j0 reconcile tools e2e_tests
-    pytest
+    pytest --cov=reconcile
```
apahim, line 8: let's also add `--cov-report=term-missing`. That's shows the gaps.

Piojo:
New version undergoing CI. Will `HUDSON_HOME` stay even as pipelines move to Tekton?

tox.ini
Piojo: Fixed!

maorfr:
i don't think we will find something that will appear both in jenkins and in tekton. we can control what will go into tekton, and that is a long way ahead anyway, so i would separate the two concerns.

HUDSON_HONE is jenkins specific and so is JOB_*. but than again, probably everything else. there are other CI systems which expose something like `CI=true`, but we don't have it here.

i think we have 2 options:
1. choose something generic, like GIT_BRANCH for example (even though it will not be compatible with tekton)
2. enhance all our jobs to have something generic like `CI=true` (we can implement the same in tekton for future compatibility)

Piojo:
Ok. These tests are going to run in jenkins for the time being (what are the plans for tekton and ci-ext?), so I propose we move forward as it is and revisit when we change the CI service for this repo.

maorfr (approved):
LGTM

maorfr:
these are the current plans for migrating this sort of job: https://issues.redhat.com/browse/APPSRE-3184

PR #1602, opened 2021-06-04
[openshift-saas-deploy-triggers] threaded run + refactor

maorfr:
this PR started out as adding usage of multi threading.
from there, it was difficult not to refactor the trigger function to lock for as little time as possible.
from there, it was difficult not to make the different trigger provider sections look as similar as possible.
from there, it was difficult not to change `already_triggered` to a set (also was a debt to @Piojo).
from there, it was difficult not to extract them to functions (also was a debt to @rporres).
from there, it was difficult not to update `setup` from options to **options (also was a debt to @Piojo).
from there, it was difficult not to drop one line in the `oc_map` initiation (also was a debt to my OCD).

note to the wise reviewers: commit by commit.
note to future selves: this is likely not a complete refactor (thoughts of more OOP come to mind), but it makes things much simpler to understand and discuss.

### DIFF
--- a/reconcile/openshift_saas_deploy_trigger_base.py
+++ b/reconcile/openshift_saas_deploy_trigger_base.py
@@ -119,8 +119,8 @@ def construct_tekton_trigger_resource(saas_file_name,
               error_details=name), long_name
 
 
-def trigger(dry_run,
-            spec,
+def trigger(spec,
+            dry_run,
             jenkins_map,
             oc_map,
             already_triggered,
@@ -131,8 +131,8 @@ def trigger(dry_run,
     """Trigger a deployment according to the specified pipelines provider
 
     Args:
-            dry_run (bool): Is this a dry run
             spec (dict): A trigger spec as created by saasherder
+            dry_run (bool): Is this a dry run
             jenkins_map (dict): Instance names with JenkinsApi instances
             oc_map (OC_Map): a dictionary of OC clients per cluster
             already_triggered (list): A list of already triggered deployments.
--- a/reconcile/openshift_saas_deploy_trigger_base.py
+++ b/reconcile/openshift_saas_deploy_trigger_base.py
@@ -66,62 +66,6 @@ def setup(options):
     return saasherder, jenkins_map, oc_map, settings
 
 
-def construct_tekton_trigger_resource(saas_file_name,
-                                      env_name,
-                                      timeout,
-                                      settings,
-                                      integration,
-                                      integration_version):
-    """Construct a resource (PipelineRun) to trigger a deployment via Tekton.
-
-    Args:
-        saas_file_name (string): SaaS file name
-        env_name (string): Environment name
-        timeout (int): Timeout in minutes before the PipelineRun fails
-        settings (dict): App-interface settings
-        integration (string): Name of calling integration
-        integration_version (string): Version of calling integration
-
-    Returns:
-        OpenshiftResource: OpenShift resource to be applied
-    """
-    long_name = f"{saas_file_name}-{env_name}".lower()
-    # using a timestamp to make the resource name unique.
-    # we may want to revisit traceability, but this is compatible
-    # with what we currently have in Jenkins.
-    ts = datetime.datetime.utcnow().strftime('%Y%m%d%H%M')  # len 12
-    # max name length can be 63. leaving 12 for the timestamp - 51
-    name = f"{long_name[:50]}-{ts}"
-    body = {
-        "apiVersion": "tekton.dev/v1beta1",
-        "kind": "PipelineRun",
-        "metadata": {
-            "name": name
-        },
-        "spec": {
-            "pipelineRef": {
-                "name": settings['saasDeployJobTemplate']
-            },
-            "params": [
-                {
-                    "name": "saas_file_name",
-                    "value": saas_file_name
-                },
-                {
-                    "name": "env_name",
-                    "value": env_name
-                }
-            ]
-        }
-    }
-    if timeout:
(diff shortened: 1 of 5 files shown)


reconcile/openshift_saas_deploy_trigger_base.py
```diff
@@ -63,12 +67,158 @@ def setup(options):
...
+                      oc_map=oc_map,
+                      cluster=tkn_cluster_name,
+                      namespace=tkn_namespace_name,
+                      resource_type=tkn_trigger_resource.kind,
+                      resource=tkn_trigger_resource,
+                      wait_for_namespace=False)
+            if not dry_run:
+                state_update_method(spec)
+        except Exception as e:
+            error = True
```
Piojo, line 206: We have a stylistic disagreement here. If you let the exception go, the caller can try to implement some automatic remediation on its own. It is also far less code to drop this exception handling and let the top level do the final handling and reporting. Also, IMHO and FWIIW such top level is `run_integration` and not `run`.

reconcile/openshift_saas_deploy_trigger_configs.py
```diff
@@ -23,27 +24,28 @@ def run(dry_run, thread_pool_size=10, internal=None,
...
     # This will be populated by osdt_base.trigger in the below loop and
     # we need it to be consistent across all iterations
-    already_triggered = []
+    already_triggered = set()
 
-    error = False
-    for job_spec in trigger_specs:
-        trigger_error = osdt_base.trigger(
+    trigger_errors = \
+        threaded.run(
```
Piojo, line 44: Let's hold all this threading for now. If we want to run stuff concurrently, we can use `asyncio`, which is also far more readable.  Maybe its futures are easiest to understand and adapt for now. Multi-threading in Python is but an illusion, thanks to the GIL.

Let's have that conversation over slack.

reconcile/openshift_saas_deploy_trigger_configs.py
```diff
@@ -52,8 +54,7 @@ def run(dry_run, thread_pool_size=10, internal=None,
             integration=QONTRACT_INTEGRATION,
             integration_version=QONTRACT_INTEGRATION_VERSION
         )
-        if trigger_error:
-            error = True
+    error = True in trigger_errors
```
Piojo, line 57: This would be more pythonic:

```
error = any(trigger_errors)
```

reconcile/openshift_saas_deploy_trigger_base.py
```diff
@@ -63,12 +67,158 @@ def setup(options):
...
+            oc_map,
+            already_triggered,
+            settings,
+            state_update_method,
+            integration,
+            integration_version):
+    """Trigger a deployment according to the specified pipelines provider
+
+    Args:
+        spec (dict): A trigger spec as created by saasherder
```
Piojo, line 82: Open question for the future: how do you feel about type hints? At least with them we could get a static analyzer to tell us when the callers don't obey this specification.

reconcile/openshift_saas_deploy_trigger_base.py
```diff
@@ -63,12 +67,158 @@ def setup(options):
...
+    Returns:
+        bool: True if there was an error, False otherwise
+    """
+
+    # TODO: Convert these into a dataclass.
+    saas_file_name = spec['saas_file_name']
+    provider_name = spec['pipelines_provider']['provider']
+
+    error = False
+    if provider_name == 'jenkins':
```
Piojo, line 102: Can we have an `Enum` with the providers?

In some file:

```python
import enum
class Providers(enum.Enum):
    JENKINS = "jenkins"
    TEKTON = "tekton"
```
and here it becomes:
```python
if provider_name == Providers.JENKINS.value:
    ...
```

Or at least have some named constants. My concern is that we will eventually evolve this and somebody will type `jeenkins` here.

reconcile/openshift_saas_deploy_trigger_configs.py
jmelis: Our code is massively io-bound. Threading is working very well for us.

reconcile/openshift_saas_deploy_trigger_base.py
jmelis: One of the things we decided a while back was that even if specific actions fail, we still want to reconcile as much as possible, therefore the agreement here was to continue the execution but report an error (so we can be aware of this).

I agree with what you said the other day about not doing sys.exit in the integrations (in the run methods) but rather in the run_integration, which makes sense, but I think this example here is a bit different. It's an interesting conversation to have because maybe I'm not visualizing your proposal for this particular situation, but since this is the current agreement, I would suggest to not change this in this PR, but rather come up with a proposal and an action plan, if we agree to.

reconcile/openshift_saas_deploy_trigger_base.py
```diff
@@ -119,98 +269,23 @@ def construct_tekton_trigger_resource(saas_file_name,
...
-            state_update_method (function): A method to call to update state
-            integration (string): Name of calling integration
-            integration_version (string): Version of calling integration
+        name (str): unique trigger name to check and
+        already_triggered (set): A set of already triggered deployments.
+                                 It will get populated by this function.
 
     Returns:
-        bool: True if there was an error, False otherwise
+        bool: to trigger or not to trigger
```
jmelis, line 281: that is the question

maorfr: glad someone benefited from this :)

maorfr:
followed up by #1608

PR #1610, opened 2021-06-07
register error on ResourceKeyExistsError

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3331

as some consumers of the ResourceInventory class rely on it's error registering, we need to start registering more.
in [today's example](https://console-openshift-console.apps.appsrep05ue1.zqxk.p1.openshiftapps.com/k8s/ns/app-sre-pipelines/tekton.dev~v1beta1~PipelineRun/saas-grafana-app-sre-stage-202106070212/logs/openshift-saas-deploy), a deployment reported success while it actually failed due to this exception being raised.

### DIFF
--- a/reconcile/utils/openshift_resource.py
+++ b/reconcile/utils/openshift_resource.py
@@ -10,6 +10,9 @@ import semver
 
 
 class ResourceKeyExistsError(Exception):
+    """ Exception to indicate an attempt to add a desired resource
+    with the same name and the same type as was already added previously
+    """
     pass
 
 
@@ -452,6 +455,7 @@ class ResourceInventory:
             except KeyError:
                 return None
             if name in desired:
+                self.register_error()
                 raise ResourceKeyExistsError(name)
             desired[name] = value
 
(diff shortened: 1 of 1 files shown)


reconcile/utils/openshift_resource.py
```diff
@@ -452,6 +454,7 @@ def add_desired(self, cluster, namespace, resource_type, name, value):
             except KeyError:
                 return None
             if name in desired:
+                self.register_error()
```
rporres, line 457: why not?
```suggestion
                self.register_error(cluster)
```

maorfr: this is to indicate that a whole cluster is having issues. add_desired is at the namespace level. does that make sense?

rporres: Well, now that you say it, it emerges a pattern 😂 

Most likely this usage that you describe would be best served with different error methods, but it is out of the scope of this PR :)

rporres (approved):

PR #1614, opened 2021-06-08
Add tkn cluster info to saas-deploy PipelineRun

rporres:
This will help us building URLs from AlertManager

Part of https://issues.redhat.com/browse/APPSRE-3320


rporres:
I've tested that the approach is viable, to add new parameters to a pipeline run without modifying the pipeline definition. It doesn't break.

The change in the integration is not tested yet, I still need to figure out how to run any of the affected integrations locally 😅

rporres:
The other part is in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/20396

maorfr (approved):
NICE

rporres:
Tested that `pipelines_provider` key of the saasherder spec contains the desired information :)

PR #1617, opened 2021-06-08
[openshift-saas-deploy-trigger-cleaner] introduce new integration

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3340

this PR introduces a new integration called `openshift-saas-deploy-trigger-cleaner`.
this integration goes over all pipelines providers, and for each one of type `tekton` it performs a cleanup according to an optinal `retention` section.

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/20431

### DIFF
--- a/reconcile/openshift_saas_deploy_trigger_cleaner.py
+++ b/reconcile/openshift_saas_deploy_trigger_cleaner.py
@@ -1,0 +1,71 @@
+import logging
+
+from datetime import datetime, timedelta
+from dateutil import parser
+
+import reconcile.queries as queries
+
+from reconcile.utils.oc import OC_Map
+from reconcile.utils.defer import defer
+from reconcile.utils.semver_helper import make_semver
+from reconcile.utils.saasherder import Providers
+
+
+QONTRACT_INTEGRATION = 'openshift-saas-deploy-trigger-cleaner'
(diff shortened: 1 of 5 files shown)


reconcile/openshift_saas_deploy_trigger_cleaner.py
```diff
@@ -0,0 +1,71 @@
...
+        integration=QONTRACT_INTEGRATION,
+        settings=settings, internal=internal,
+        use_jump_host=use_jump_host,
+        thread_pool_size=thread_pool_size)
+    defer(lambda: oc_map.cleanup())
+
+    for pp in pipelines_providers:
+        retention = pp.get('retention')
+        if not retention:
+            continue
```
BumbleFeng, line 45: Should we hava default retention setting instead of ignoring those pipelines?

maorfr: good next iteration? for now we have very few namespaces that have A LOT of PipelineRuns.

rrati: Next iteration wfm

jmelis: +1

PR #1621, opened 2021-06-09
[ocm-clusters] another fix to update logic

maorfr:
to avoid KeyError on `nodes` and to only add the autoscale/nodes if it is being updated

follow up on #1612

### DIFF
--- a/reconcile/utils/ocm.py
+++ b/reconcile/utils/ocm.py
@@ -217,9 +217,15 @@ class OCM:
 
         autoscale = cluster_spec.get('autoscale')
         if autoscale is not None:
-            ocm_spec['nodes']['autoscale_compute'] = autoscale
-        else:
-            ocm_spec['nodes']['compute'] = cluster_spec['nodes']
+            ocm_spec['nodes'] = {
+                'autoscale_compute': autoscale
+            }
+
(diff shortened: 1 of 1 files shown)


reconcile/utils/ocm.py
```diff
@@ -217,9 +217,15 @@ def update_cluster(self, name, cluster_spec, dry_run):
...
-        else:
-            ocm_spec['nodes']['compute'] = cluster_spec['nodes']
+            ocm_spec['nodes'] = {
+                'autoscale_compute': autoscale
+            }
+
+        nodes = cluster_spec.get('nodes')
+        if nodes:
+            ocm_spec['nodes'] = {
+                'compute': cluster_spec['nodes']
```
Piojo, line 227: If I'm reading this right, you will have both `autoscale_compute` and `compute` in the same OCM spec. Is this desired?

maorfr: i guess so. but this is prevented here: https://gitlab.cee.redhat.com/service/app-interface/-/blob/d9300f9c214b60d3a07be1ae544ea989d436aae1/schemas/openshift/cluster-1.yml#L193-197

Piojo (approved):

PR #1615, opened 2021-06-08
Revert "register error on ResourceKeyExistsError (#1610)"

rporres:
This reverts commit d9c7d2ab8e9f9ad21d27e72c1e6af2356e9aec17.

This is making openshift-rolebindings and openshift-clusterrolebindings silently fail

PR #1622, opened 2021-06-09
[saasherder] add handle for ResourceKeyExistsError

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3331

this is an alternate approach for #1610 which was reverted in #1615 

cc @rporres


rporres:
This looks more targeted to what you are fixing, looks a better approach 👍

PR #1627, opened 2021-06-10
[openshift] add support for recycling StatefulSet

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3351

```
Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden
```

this PR adds support to apply every additional changes to StatefulSets without incurring any downtime.

When we detect a fields immutable error for a StatefulSet:
1. delete statefulset with cascade false to prevent the pods from being deleted
2. apply the updated statefulset
3. delete one pod at a time and wait for the new statefulset to bring it up

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -246,7 +246,7 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
 
             oc.delete(namespace=namespace, kind=resource_type,
                       name=resource.name, cascade=cascade)
-            oc.apply(namespace=namespace, resource=annotated
+            oc.apply(namespace=namespace, resource=annotated)
 
             if not cascade:
                 oc.recycle_orphan_pods(dry_run, namespace, resource)
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -12,6 +12,7 @@ from reconcile.utils.oc import MayNotChangeOnceSetError
 from reconcile.utils.oc import PrimaryClusterIPCanNotBeUnsetError
 from reconcile.utils.oc import InvalidValueApplyError
 from reconcile.utils.oc import MetaDataAnnotationsTooLongApplyError
+from reconcile.utils.oc import StatefulSetUpdateForbidden
 from reconcile.utils.oc import OC_Map
 from reconcile.utils.oc import StatusCodeError
 from reconcile.utils.oc import UnsupportedMediaTypeError
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -254,7 +254,7 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
                       name=resource.name)
             oc.apply(namespace=namespace, resource=annotated)
         except StatefulSetUpdateForbidden:
-            if resource_type not in ['StatefulSet']:
+            if resource_type != 'StatefulSet':
                 raise
 
             logging.info(['delete_sts_and_apply', cluster, namespace,
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -260,7 +260,7 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
             oc.apply(namespace=namespace, resource=annotated)
 
             if not cascade:
-                oc.recycle_orphan_pods(dry_run, namespace, resource)
+                oc.recycle_sts_orphan_pods(dry_run, namespace, resource)
         except (MayNotChangeOnceSetError, PrimaryClusterIPCanNotBeUnsetError):
             if resource_type not in ['Service']:
                 raise
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -257,13 +257,13 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
             if resource_type not in ['StatefulSet']:
                 raise
 
-            logging.info(['delete_and_apply', cluster, namespace,
+            logging.info(['delete_sts_and_apply', cluster, namespace,
                           resource_type, resource.name])
             owned_pods = oc.get_owned_pods(namespace, resource)
             oc.delete(namespace=namespace, kind=resource_type,
                       name=resource.name, cascade=False)
             oc.apply(namespace=namespace, resource=annotated)
-            logging.info(['recycle_pods', cluster, namespace,
+            logging.info(['recycle_sts_pods', cluster, namespace,
                           resource_type, resource.name])
             # the resource was applied without cascading, we proceed
             # to recycle the pods belonging to the old resource.
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -239,10 +239,21 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
         except FieldIsImmutableError:
             # Add more resources types to the list when you're
             # sure they're safe.
-            if resource_type not in ['Route', 'Service', 'StatefulSet']:
+            supported_resource_types = {
+                'Route': {
+                    'cascade': True
+                },
+                'Service': {
+                    'cascade': True
+                },
+                'StatefulSet': {
+                    'cascade': False
+                },
+            }
+            if resource_type not in supported_resource_types:
                 raise
 
-            cascade = resource_type not in ['StatefulSet']
+            cascade = supported_resource_types[resource_type]['cascade']
 
             oc.delete(namespace=namespace, kind=resource_type,
                       name=resource.name, cascade=cascade)
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -237,46 +237,14 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
                 oc.create(namespace, annotated)
             oc.replace(namespace, annotated)
         except FieldIsImmutableError:
-            # when we detect a fields immutable error for a resource
-            # that is supported for being replaced.
-            # we delete it. if the resource can be applied
-            # without cascading, we do it to avoid replacing pods.
-            # we then apply the new resource.
             # Add more resources types to the list when you're
             # sure they're safe.
-            supported_resource_types = {
-                'Route': {
-                    'cascade': True
-                },
-                'Service': {
-                    'cascade': True
-                },
-                'StatefulSet': {
-                    'cascade': False
-                },
-            }
-            if resource_type not in supported_resource_types:
+            if resource_type not in ['Route', 'Service']:
                 raise
 
-            cascade = supported_resource_types[resource_type]['cascade']
-
-            logging.info(['delete_and_apply', cluster, namespace,
-                          resource_type, resource.name])
-            owned_pods = \
-                None if cascade else oc.get_owned_pods(namespace, resource)
             oc.delete(namespace=namespace, kind=resource_type,
-                      name=resource.name, cascade=cascade)
+                      name=resource.name)
             oc.apply(namespace=namespace, resource=annotated)
-
-            if not cascade:
-                # if the resource was applied without cascading, we proceed
-                # to recycle the pods belonging to the old resource.
-                # note: we really just delete pods and let the new resource
-                # recreate them. we delete one by one and wait for a new
-                # pod to become ready before proceeding to the next one.
-                logging.info(['recycle_pods', cluster, namespace,
-                              resource_type, resource.name])
-                oc.recycle_orphan_pods(namespace, owned_pods)
         except (MayNotChangeOnceSetError, PrimaryClusterIPCanNotBeUnsetError):
             if resource_type not in ['Service']:
                 raise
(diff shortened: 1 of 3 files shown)


reconcile/utils/oc.py
```diff
@@ -439,6 +440,26 @@ def get_service_account_username(user):
         name = user.split('/')[1]
         return "system:serviceaccount:{}:{}".format(namespace, name)
 
+    def recycle_orphan_pods(self, dry_run, namespace, resource):
```
rporres, line 443: Seeing `validate_pod_ready` function, I don't see where this method of recycling would work apart from StatefulSets so I would name the functions accordingly so they don't give the impression of being more general than they currently are.

reconcile/openshift_base.py
```diff
@@ -239,12 +239,17 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
         except FieldIsImmutableError:
             # Add more resources types to the list when you're
             # sure they're safe.
-            if resource_type not in ['Route', 'Service']:
+            if resource_type not in ['Route', 'Service', 'StatefulSet']:
                 raise
 
+            cascade = resource_type not in ['StatefulSet']
```
apahim, line 245: This is begging for a mapping...
```python
supported_resource_types = {
    'Route': {
        'cascade': False
    },
    'Service': {
        'cascade': False
    },
    'StatefulSet': {
        'cascade': True
    },
}
```
No need to maintain 2 lists of resources_types

reconcile/utils/oc.py
```diff
@@ -439,6 +440,26 @@ def get_service_account_username(user):
...
+                continue
+            if resource.name != owner['metadata']['name']:
+                continue
+            # we found a pod we want to recycle
+            name = p['metadata']['name']
+            self.delete(namespace, 'Pod', name)
+            self.validate_pod_ready(namespace, name)
+
+    @retry(max_attempts=20)
+    def validate_pod_ready(self, namespace, name):
```
rporres, line 457: Maybe 
```suggestion
    def validate_recycled_statefuset_pod_ready(self, namespace, name):
```
Sorry for the long names, I must have been a java developer in a previous life

apahim:
I don't see tests, still, CI seems happy. Is that expected? Saw some discussion about not reducing the coverage, this clearly reduces it. Are we in a grace period or something?

reconcile/utils/oc.py
```diff
@@ -439,6 +440,26 @@ def get_service_account_username(user):
...
+        pod = self.get(namespace, 'Pod', name)
+        for status in pod['status']['containerStatuses']:
+            if not status['ready']:
+                raise PodNotReadyError(name)
```
rporres, line 461: If we arrive here, it means that we're bound to manual intervention. I would write a nice and long message describing the situation for the poor soul that has to fix this :)

reconcile/utils/oc.py
```diff
@@ -439,6 +440,26 @@ def get_service_account_username(user):
...
 
+    def recycle_orphan_pods(self, dry_run, namespace, resource):
+        pods = self.get(namespace, 'Pod')['items']
+        for p in pods:
+            owner = self.get_obj_root_owner(namespace, p)
+            if resource.kind != owner['kind']:
+                continue
+            if resource.name != owner['metadata']['name']:
+                continue
+            # we found a pod we want to recycle
```
rporres, line 451: I would write an INFO message here to that says which one we're recycling. I'm thinking in a situation where you have 5 pods, and the third one fails to get recycled. Then at least you know that whenever you have to fix this manually, you don't have to touch certain pods

reconcile/openshift_base.py
maorfr: very nice. implemented.

rporres (commented):
Doing this with StatefulSets scares the **** out of me, but this approach is cool. The new SS doesn't do anything as it already has the number of pods it needs. Great.

I would add a good chunk of documentation (basically what you have in the description)

Another thing I don't know how to fix is pods getting a long time to get deleted. I remember Telemeter pods taking a long time to get rebooted because they needed to wait a lot to upload something. If that's something that is in the Pod spec, at least we should wait that time before declaring the Pod not ready...

reconcile/utils/oc.py
maorfr: renamed to `recycle_sts_orphan_pods` and added protection

reconcile/utils/oc.py
maorfr: this is a generic function, the pod just has to become ready

reconcile/utils/oc.py
maorfr: this is raised in the context of a deployment. so before this happens, we will see a log that a StatefulSet is being applied. i've added more logging to indicate that we are deleting and applying, and more logging to indicate that we are recycling pods and validating them. so the exception alone will not say much, but it will within the context.

and just a nit: we will reach here multiple times but will only raise it after the retry max attempts.

reconcile/utils/oc.py
maorfr: added logging!

maorfr:
@rporres do note that when we apply the new STS, it is not happy. this is why we need to delete the pods one by one. after each deletion, the new STS will kick in and create the pod (we are just deleting an orphaned one from the previous STS).

about pods that take a long time to be deleted - this is why i use `20` as the max attempts for the pod ready validation. 20 attempts is 210 seconds (1 + 2 + ... + 20) which i hope would be enough. if not, we can either extend the attempts here or add support through the saas file for a configurable number (maybe even use `timeout`).

reconcile/utils/oc.py
rporres: Hmmm, ok, you're right. This can be indeed a function that doesn't have to do with stateful sets

reconcile/utils/oc.py
rporres: As long as we know that we're recycling statefulsets pods in the context of an immutable change and which pod we're recycling I'm fine.

maorfr: :kissing:

rporres:
> do note that when we apply the new STS, it is not happy. this is why we need to delete the pods one by one. after each deletion, the new STS will kick in and create the pod (we are just deleting an orphaned one from the previous STS).

That's my understanding too of what you're trying to achieve here

maorfr:
@apahim `Required test coverage of 19.0% reached. Total coverage: 19.14%`

rporres:
> about pods that take a long time to be deleted - this is why i use 20 as the max attempts for the pod ready validation. 20 attempts is 210 seconds (1 + 2 + ... + 20) which i hope would be enough. if not, we can either extend the attempts here or add support through the saas file for a configurable number (maybe even use timeout).

Thanos receive pods have a `terminationGracePeriodSeconds` set to 900 seconds, so our retry time has to take that into account or we will declare pods as not ready while they're behaving as configured.

I don't think retry time should be configurable as it would mean duplicating configuration that the pods already have

apahim:
> @apahim Required test coverage of 19.0% reached. Total coverage: 19.14%

oh, so we have some fat to burn, got it. Thanks!

rporres:
What I don't know yet how to achieve is to avoid waiting the `terminationGracePeriodSeconds` if the pod has already been properly terminated so that we don't have to wait more than we need 🤔

maorfr:
@rporres i think i gave you a wrong answer. the waiting time is for a new pod to become ready. waiting for the pod to be deleted is a part of the `oc delete` step, and i believe we get that for free. i mean - `oc delete` itself will wait for the pod to terminate before returning.

maorfr:
changed approach a bit to simplify and not not decide between cascading or not.

rporres:
> waiting for the pod to be deleted is a part of the oc delete step, and i believe we get that for free

Yes, I also think you're right. `oc` waits for the pod to be deleted to return, so I also believe that we get that for free

The other times we could take into account is the readiness/startup probes ones, but I would leave that for another iteration as it can be quite complicated to set up.

reconcile/openshift_base.py
```diff
@@ -252,6 +253,24 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
             oc.delete(namespace=namespace, kind=resource_type,
                       name=resource.name)
             oc.apply(namespace=namespace, resource=annotated)
+        except StatefulSetUpdateForbidden:
+            if resource_type not in ['StatefulSet']:
+                raise
+
+            logging.info(['delete_and_apply', cluster, namespace,
```
rporres, line 260: nit: since you know where you are, I would make this more explicit so it's easier to find the place where this is written when debugging something like this
```suggestion
            logging.info(['delete_sts_and_apply', cluster, namespace,
```

reconcile/openshift_base.py
```diff
@@ -252,6 +253,24 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
...
+                          resource_type, resource.name])
+            owned_pods = oc.get_owned_pods(namespace, resource)
+            oc.delete(namespace=namespace, kind=resource_type,
+                      name=resource.name, cascade=False)
+            oc.apply(namespace=namespace, resource=annotated)
+            logging.info(['recycle_pods', cluster, namespace,
```
rporres, line 266: For the same above reasons
```suggestion
            logging.info(['recycle_sts_pods', cluster, namespace,
```

rporres (approved):
two nits, this looks good otherwise :)

Piojo (commented):
By any reasonable standard, `apply` has [become too long](https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#limit-method-size). We need to plan to split it.

Then, I am very uncomfortable with the fact that code reductions in other areas are being used as an excuse to not have a single test here. There is **a lot** of value for future iterations on testing at least `get_owned_pods` and `validate_pod_ready`. Simple tests (warning! untested!):

```python
from unittest import TestCase
from unittest.mock import patch, MagicMock

import reconcile.utils.oc as oc

class TestGetOwnedPods(TestCase):
    @mock.patch.object(oc.OC, 'get')
    @mock.patch.object(oc.OC, 'get_obj_root_owner')
    def test_get_owned_pods(self, get_owner, oc_get):

        oc_get.return_value = {
            'items': ['pod1', 'pod2', 'pod3']
        }
        get_owner.side_effect = [
            {
                'kind': 'mykind',
                'metadata': {'name': 'myname'}
            },
            {
                'kind': 'notmykind',
                'metadata': {'name': 'notmyname'},
            }
            {
                'kind': 'mykind',
                'metadata': {'name': 'notmyname'}
            }
        ]

        resource = MagicMock(kind='mykind', name='myname')

        o = OC()
        pods = o.get_owned_pods('namespace', resource)
        self.assertEqual(pods, ['pod1'])

class TestValidatePodReady(TestCase):
    @mock.patch.object(oc.OC, 'get')
    def test_validate_pod_ready_all_good(self, oc_get):
        oc_get.return_value = {
            'status': {
                'containerStatuses': [
                    {
                        'name': 'container1',
                        'ready': True,
                    },
                    {
                        'name': 'container2',
                        'ready': True
                    }
                ]
            }
        }
        o = OC()
        # Bypass the retry stuff
        o.validate_pod_ready.__wrapped__('namespace', 'podname')
        # Feel free to add assertions around the call to oc_get, but
        # they aren't important on this test

    @mock.patch.object(oc.OC, 'get')
    def test_validate_pod_ready_one_missing(self, oc_get):
        oc_get.return_value = {
            'status': {
                'containerStatuses': [
                    {
                        'name': 'container1',
                        'ready': True,
                    },
                    {
                        'name': 'container2',
                        'ready': False
                    }
                ]
            }
        }

        o = OC()
        with self.assertRaises(oc.PodNotReadyError):
            o.validate_pod_ready.__wrapped__('namespace', 'podname')
```
Simple stuff like this will give the next person opening this file (which may be you in 6 months) a lot of confidence on what they're doing. And they will also speed up your own development iterations massively.

Would you consider these 2 test cases for inclusion?

reconcile/openshift_base.py
```diff
@@ -252,6 +253,24 @@ def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
             oc.delete(namespace=namespace, kind=resource_type,
                       name=resource.name)
             oc.apply(namespace=namespace, resource=annotated)
+        except StatefulSetUpdateForbidden:
+            if resource_type not in ['StatefulSet']:
```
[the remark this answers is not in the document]
maorfr: this was meant to leave room for future resource types (look above). but i guess this is STS specific.

reconcile/utils/oc.py
```diff
@@ -439,6 +444,34 @@ def get_service_account_username(user):
...
+        pods = self.get(namespace, 'Pod')['items']
+        owned_pods = []
+        for p in pods:
+            owner = self.get_obj_root_owner(namespace, p)
+            if resource.kind != owner['kind']:
+                continue
+            if resource.name != owner['metadata']['name']:
+                continue
+            # we found an owned pod
+            owned_pods.append(p)
```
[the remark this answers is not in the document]
maorfr: :+1:

reconcile/utils/oc.py
```diff
@@ -439,6 +444,34 @@ def get_service_account_username(user):
...
+
+    def recycle_orphan_pods(self, namespace, pods):
+        for p in pods:
+            name = p['metadata']['name']
+            self.delete(namespace, 'Pod', name)
+            self.validate_pod_ready(namespace, name)
+
+    @retry(max_attempts=20)
+    def validate_pod_ready(self, namespace, name):
+        logging.info(['validate_pod_ready', namespace, name])
```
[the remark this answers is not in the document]
maorfr: i'll go with you

reconcile/utils/oc.py
```diff
@@ -439,6 +444,34 @@ def get_service_account_username(user):
...
+        return owned_pods
+
+    def recycle_orphan_pods(self, namespace, pods):
+        for p in pods:
+            name = p['metadata']['name']
+            self.delete(namespace, 'Pod', name)
+            self.validate_pod_ready(namespace, name)
+
+    @retry(max_attempts=20)
+    def validate_pod_ready(self, namespace, name):
```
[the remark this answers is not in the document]
maorfr: i have an old itch from my days as a test automation developer that assertions only happen in tests, and using "assert" in a function name indicates using an assertion?

maorfr:
added tests! thanks for helping get them started!

about `apply`'s length - can you say exactly what makes it too long?
it does one thing - `oc apply`, and it does it well (including catching errors and overcoming them), it has very few local variables, it has less than 4 levels of indentation. it does have a lot of arguments, but not too many?

i agree that this section is not pretty: https://github.com/app-sre/qontract-reconcile/blob/4f9dcbda1153f1324eb580e3fdf82f15a0897ca9/reconcile/openshift_base.py#L216-L225

but otherwise, how would you refactor it?

maorfr:
in any case, we can discuss this in jira: https://issues.redhat.com/browse/APPSRE-3355

proceeding to merge this.

reconcile/utils/oc.py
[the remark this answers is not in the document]
Piojo: Yeah, in my eyes this acts like one: it is called in the middle of some processing and breaks the flow entirely by raising an exception/aborting whatever. It is true that this one isn't meant to be disabled with `-DNDEBUG`.

It was a nit, just that I like it when the naming makes it clear that this is an integrity check and that normal flow of the program can break here.

PR #1628, opened 2021-06-10
Add an option to threaded.run to return all exceptions

Piojo:
It is based on `asyncio.gather`.  It will run the given function on
the entire iterable, but any failures will become part of the result
list. The caller will tell good results from errors by inspecting on
instances of `Exception`.

The idea is to simplify exception handling by not having to convert exceptions in underlying code into some return values that will then be converted into something else, which makes us lose information along the way.

Also an example of code with tests.

### DIFF
--- a/reconcile/test/test_utils_threaded.py
+++ b/reconcile/test/test_utils_threaded.py
@@ -1,0 +1,54 @@
+import unittest
+import reconcile.utils.threaded as threaded
+
+
+def identity(x):
+    return x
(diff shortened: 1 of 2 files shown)


maorfr:
related to https://issues.redhat.com/browse/APPSRE-3335?

maorfr (approved):
LGTM

read the tests thoroughly. very educating. thanks a lot!

PR #1633, opened 2021-06-13
[utils/gpg] add tests

maorfr:
add some basic tests for gpg


Piojo (approved):
Thanks for testing! :))

PR #1635, opened 2021-06-14
[openshift-saas-deploy] validate StatefulSet after deployment

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3351

this PR adds validation that a deployment on a StatefulSet was successful.

follow up on #1627

about keeping a state in case the deployment stops mid-run -
in such case, it is expected that the next deployment realizes that the StatefulSet is not in a healthy state and fail.
with this PR - this will now be the reality.


jmelis (approved):
LGTM

PR #1636, opened 2021-06-14
add qontract-cli get sre-checkpoints

jmelis:


jmelis:
@Piojo can you review?

tools/qontract_cli.py
```diff
@@ -491,6 +491,57 @@ def service_owners_for_rds_instance(ctx, aws_account, identifier):
...
+
+    app_checkpoints = []
+    for app in apps:
+        # skip parent apps
+        if app['path'] in parent_apps:
+            continue
+
+        # consider only onboardedApps
+        if app['onboardingStatus'] != 'OnBoarded':
+            continue
```
Piojo, line 532: Shorter:

```python
if app['path'] not in parent_apps and app['onboardingStatus'] == 'OnBoarded':
    app_checkpoints.append(...)
```
which may lend itself to a list comprehension.

tools/qontract_cli.py
```diff
@@ -491,6 +491,57 @@ def service_owners_for_rds_instance(ctx, aws_account, identifier):
...
+    sre_checkpoints = {}
+    for checkpoint in sre_checkpoints_raw:
+        app = checkpoint['app']
+        name = full_name(app)
+
+        if name not in sre_checkpoints:
+            sre_checkpoints[name] = checkpoint['date']
+        else:
+            if checkpoint['date'] > sre_checkpoints[name]:
+                sre_checkpoints[name] = checkpoint['date']
```
Piojo, line 516: A smart trick:

```python
sre_checkpoint[name] = max(sre_checkpoints.get(name, 0), checkpoint['date'])
```

tools/qontract_cli.py
```diff
@@ -491,6 +491,57 @@ def service_owners_for_rds_instance(ctx, aws_account, identifier):
     print_output(ctx.obj['output'], service_owners, columns)
 
 
+@get.command()
+@click.pass_context
+def sre_checkpoints(ctx):
+    sre_checkpoints_raw = queries.get_sre_checkpoints()
+    apps = queries.get_apps()
+
+    def full_name(app):
```
Piojo, line 500: Not a big deal here, but internal functions are very hard to test. We can accept it here.

tools/qontract_cli.py
```diff
@@ -491,6 +491,57 @@ def service_owners_for_rds_instance(ctx, aws_account, identifier):
...
+        name = app['name']
+        if app.get('parentApp'):
+            parent_app = app['parentApp']['name']
+            name = f"{parent_app}/{name}"
+        return name
+
+    sre_checkpoints = {}
+    for checkpoint in sre_checkpoints_raw:
```
Piojo, line 508: I'm very confused with all these `sre_checkpoints`, `checkpoint`, `sre_checkpoints_raw`... a wise man told me that naming is one of the two hardest things in computing. The others being where to put things and off-by-one errors.

Piojo (commented):
... and I know I'm annoying, I can haz tests? Since you have two layers of decorators I'd guess you'll have to do `method_to_test = sre_checkpoints.__wrapped__.__wrapped__`. Or something along those lines. 

And I guess for the time being we'll have to mock away all the graphql calls in our tests. In due time we can plan to write a stub graphql for these queries.

tools/qontract_cli.py
jmelis: someone likes functional programming

tools/qontract_cli.py
jmelis: In general, in terms of legibility, do you prefer your way? I think in terms of performance is the same, and with regard to legibility, I believe I would understand the separated if clauses better than a long if. Also the long if holds because there are just two terms, but for more terms I really prefer to separate them.

Piojo: Yeah, I find a single `and` expression much easier to understand than a string of if-continues. Also, if this were a complicated expression, I doubt 10 2-line blocks, spanning 20 lines would be easier to read than a single if with 10 conditions.

tools/qontract_cli.py
Piojo: That someone also hates `if`. ;)

jmelis:
@Piojo ready for a second review

Piojo (approved):

tools/test/test_sre_checkpoints.py
```diff
@@ -0,0 +1,96 @@
+from unittest.mock import patch
```
jmelis, line 1: for @Piojo with love

PR #1637, opened 2021-06-15
[sendgrid-teammates] fix user deletion

maorfr:
deleting a user is done with the `username`: https://sendgrid.com/docs/api-reference/
a username can be retrieved from the api, and will sometimes differ from the email name.


Piojo (approved):
It's so simple I'm not going to pester you with tests. But if you could...

maorfr:
will add tests at a later time, but at least getting this fix in (i am completely out of context by now)

PR #1638, opened 2021-06-15
override sleep_duration_secs for jenkins-job-builds-cleaner

jfchevrette:


jfchevrette:
@maorfr wdyt about doing this as a helm value?

It could also be a new OpenShift template parameter but if eventually we want to tweak the sleep time of more integrations we would need one parameter per integration we want to tweak. However doing this allows us to modify the value via app-interface which is more typical.

maorfr:
perhaps this is better served as a CronJob?
https://github.com/app-sre/qontract-reconcile/blob/e5abbd734741a307502abb40a3123bdec44718ec/helm/qontract-reconcile/values-external.yaml#L513

jfchevrette:
Closing in favor of #1643 and #1644

PR #1639, opened 2021-06-15
introduce integration ocm-additional-router

BumbleFeng:
https://issues.redhat.com/browse/APPSRE-3062

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/20784

### DIFF
--- a/reconcile/test/test_ocm_additional_routers.py
+++ b/reconcile/test/test_ocm_additional_routers.py
@@ -1,0 +1,128 @@
+from unittest.mock import patch, call
+from unittest import TestCase
+
+from reconcile.utils.ocm import OCMMap
+import reconcile.queries as queries
+import reconcile.ocm_additional_routers as integ
+
+from .fixtures import Fixtures
+
+
+fxt = Fixtures('ocm_additional_routers')
+
+
+class TestOCMAdditionalRouters(TestCase):
+
+    # integration test
+    @patch.object(queries, 'get_clusters')
+    def test_integ_fail(self, get_clusters):
+        fixture = fxt.get_anymarkup('state.yml')
+
+        clusters = fixture['gql_response']
+        for c in clusters:
+            c.pop('additionalRouters')
+        get_clusters.return_value = clusters
+
+        with self.assertRaises(SystemExit):
+            integ.run(False)
+
+    @patch.object(queries, 'get_app_interface_settings')
+    @patch.object(queries, 'get_clusters')
+    @patch.object(OCMMap, 'init_ocm_client')
+    @patch.object(OCMMap, 'get')
+    def test_integ(self, get, init_ocm_client, get_clusters,
+                   get_app_interface_settings):
+        fixture = fxt.get_anymarkup('state.yml')
+
+        get_clusters.return_value = fixture['gql_response']
+        ocm = get.return_value
+        ocm.get_additional_routers.side_effect = \
+            lambda x: fixture['ocm_api'][x]
+
+        integ.run(False)
+
+        ocm_act = fixture['ocm_act']
+
+        router_create = ocm_act['create']
+        expected = []
+        for c in router_create.keys():
+            expected.append(call(c, router_create[c]))
+        calls = ocm.create_additional_router.call_args_list
+        self.assertEqual(calls, expected)
+
+        router_delete = ocm_act['delete']
+        expected = []
+        for c in router_delete.keys():
+            expected.append(call(c, router_delete[c]))
+        calls = ocm.delete_additional_router.call_args_list
+        self.assertEqual(calls, expected)
+
+    # unit test
+    @patch.object(queries, 'get_app_interface_settings')
+    @patch.object(OCMMap, 'init_ocm_client')
+    @patch.object(OCMMap, 'get')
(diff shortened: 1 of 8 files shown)


jfchevrette (approved):
This looks very clean. LGTM!

maorfr:
following https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/20784#note_2556356, i'd consider renaming this integration to `ocm-additional-routers`?

reconcile/utils/ocm.py
```diff
@@ -623,6 +623,63 @@ def delete_upgrade_policy(self, cluster, spec):
...
+            f'{CS_API_BASE}/v1/clusters/{cluster_id}/ingresses'
+        items = self._get_json(api).get('items')
+        if not items:
+            return results
+
+        for item in items:
+            # filter out default router
+            if item['default']:
+                continue
+            desired_keys = ['id', 'listening', 'dns_name', 'route_selectors']
```
Piojo, line 647: Move this out of this block, or you'll be generating this list over and over on each iteration. Ideally, make it a constant.

reconcile/ocm_additional_router.py
```diff
@@ -0,0 +1,105 @@
...
+    """ Sort diffs so we delete first and create later """
+    if diff['action'] == 'delete':
+        return 1
+    else:
+        return 2
+
+
+def act(dry_run, diffs, ocm_map):
+    diffs.sort(key=sort_diffs)
+    for diff in diffs:
```
Piojo, line 78: Please, don't use loop variables that differ from the iterated object by just one character. In this context, `for d in diffs` is perfectly understandable.

reconcile/ocm_additional_router.py
```diff
@@ -0,0 +1,105 @@
...
+
+def sort_diffs(diff):
+    """ Sort diffs so we delete first and create later """
+    if diff['action'] == 'delete':
+        return 1
+    else:
+        return 2
+
+
+def act(dry_run, diffs, ocm_map):
```
Piojo, line 76: `act` is not a terribly useful name here. How about `update_routers`?

reconcile/ocm_additional_router.py
```diff
@@ -0,0 +1,105 @@
...
+    desired_state = []
+    for cluster in clusters:
+        cluster_name = cluster['name']
+        router = cluster['additionalRouter']
+        listening = 'internal' if router['private'] else 'external'
+        item = {
+            'listening': listening,
+            'cluster': cluster_name
+        }
+        selectors = cluster['additionalRouter']['route_selectors']
```
Piojo, line 40: This is `router['route_selectors']`, isn't it? Either use `router` all the way or drop this variable and use `cluster['additionalRouter']`. In any case, do the same things the same way always.

Piojo (changes requested):
This is pretty big and meaningful, so I can haz tests? I don't need a very comprehensive suite, but I do need some degree of automated testing here. Feel free to reach out if you want help.

reconcile/ocm_additional_router.py
maorfr: fwiw, this is very common across the code base, and specifically for ocm integrations

reconcile/ocm_additional_router.py
maorfr: fwiw, `act` is a generic name for the different actions that we may perform, which differ between integrations

Piojo: ... but we already have `run` as our entry point for the "API", at some point we have to become specific, don't we?  Anyways, this isn't a big concern.

reconcile/test/test_ocm_additional_routers.py
```diff
@@ -0,0 +1,118 @@
...
+        diffs = fixture['diffs']
+        integ.act(False, diffs, ocm_map)
+
+        ocm_act = fixture['ocm_act']
+
+        router_create = ocm_act['create']
+        expected = []
+        for c in router_create.keys():
+            expected.append(call(c, router_create[c]))
+        calls = ocm.create_additional_router.call_args_list
```
Piojo, line 110: Why not:

```python
for c, route in router_create.items():
    ocm.create_additional_router.assert_called_with(c, route)
```
?

reconcile/test/test_ocm_additional_routers.py
```diff
@@ -0,0 +1,118 @@
...
+        self.assertEqual(calls, expected)
+
+        router_delete = ocm_act['delete']
+        expected = []
+        for c in router_delete.keys():
+            expected.append(call(c, router_delete[c]))
+        calls = ocm.delete_additional_router.call_args_list
```
Piojo, line 117: As above.

Piojo (approved):
Awesome stuff. Thanks!

reconcile/test/test_ocm_additional_routers.py
BumbleFeng: assert_called_with can only get info of the last call.

Piojo: `assert_any_call` is the right call here.

PR #1640, opened 2021-06-16
allow providing resources values from saas file and not only through helm chart

maorfr:
currently, whenever we want to add resources to an integration, we must make the changes in the repo and promote them to production.

with this PR, we can now override any resources that we want directly from the saas file.

### DIFF
--- a/helm/qontract-reconcile/templates/template.yaml
+++ b/helm/qontract-reconcile/templates/template.yaml
@@ -220,8 +220,27 @@ objects:
                 key: {{ $env.secretKey }}
           {{- end }}
           {{- end }}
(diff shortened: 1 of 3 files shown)


rporres:
Great stuff, very much needed and nicely implemented :)

I would call the variables using singular, not plural: `xxx_CPU_LIMIT`, `xxx_MEMORY_LIMIT`, `xxx_CPU_REQUEST`, `xxx_MEMORY_REQUEST` as I guess `limits` and `requests` keys are plural because we have two of them but inside each of them is a resource request/limit

rporres (approved):

PR #1643, opened 2021-06-16
[JJB cleaner] update rule match strategy & change deploy to cronjob

BumbleFeng:


openshift/qontract-reconcile-internal.yaml
```diff
@@ -6961,6 +6778,63 @@ objects:
...
           emptyDir: {}
+- apiVersion: batch/v1beta1
+  kind: CronJob
+  metadata:
+    labels:
+      app: qontract-reconcile-jenkins-job-builds-cleaner
+    name: qontract-reconcile-jenkins-job-builds-cleaner
+  spec:
+    schedule: "0 * * * *"
+    concurrencyPolicy: Allow
```
jfchevrette, line 6789: Should we set this to `Forbid` or maybe `Replace` here? We wouldn't want multiple jobs to run at the same time

reconcile/jenkins_job_builds_cleaner.py
```diff
@@ -49,6 +46,8 @@ def find_builds(jenkins, job_names, rules):
                             'rule_keep_hours': rule['keep_hours'],
                             'build_id': build['id'],
                         })
+            # Only act on the first rule matched
+            break
```
jfchevrette, line 50: Shouldn't this be indented a level further inside of (at the end of) `if rule['name_re'].search(job_name)`?

jfchevrette (commented):
Nice optimization! Also thank you for moving this to a CronJob!

jfchevrette (approved):

maorfr (approved):
LGTM

PR #1644, opened 2021-06-17
allow override sleep_duration_secs

BumbleFeng:
relate to https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/20844

### DIFF
--- a/helm/qontract-reconcile/values-internal.yaml
+++ b/helm/qontract-reconcile/values-internal.yaml
@@ -217,6 +217,7 @@ integrations:
       memory: 200Mi
       cpu: 200m
 - name: gitlab-housekeeping
+  sleepDurationSecs: 600
   resources:
     requests:
       memory: 100Mi
--- a/openshift/qontract-reconcile-internal.yaml
+++ b/openshift/qontract-reconcile-internal.yaml
@@ -5672,7 +5672,7 @@ objects:
           - name: INTEGRATION_EXTRA_ARGS
             value: "--wait-for-pipeline"
           - name: SLEEP_DURATION_SECS
-            value: ${SLEEP_DURATION_SECS}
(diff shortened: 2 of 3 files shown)


openshift/qontract-reconcile-internal.yaml
```diff
@@ -7172,6 +7172,8 @@ parameters:
   value: 100m
 - name: GITLAB_OWNERS_MEMORY_REQUEST
   value: 100Mi
+  name: GITLAB_HOUSEKEEPING_SLEEP_DURATION_SECS
+  value: "0"
```
jfchevrette, line 7176: Should we have default that isn't zero?

helm/qontract-reconcile/values-internal.yaml
```diff
@@ -204,6 +204,7 @@ integrations:
       memory: 200Mi
       cpu: 200m
 - name: gitlab-housekeeping
+  sleep_duration_secs: "0"
```
jfchevrette, line 207: This look like this will run continually (no sleep). Is this desired?

BumbleFeng: GitLab housekeeping can only merge one MR for app-interface in one loop. The faster it iterates, the more quickly MR can be merged. In the case of foolproof, I have changed the default value in template to 600 and override the parameter in the saas file.

jfchevrette (approved):

helm/qontract-reconcile/values-internal.yaml
```diff
@@ -204,6 +204,7 @@ integrations:
       memory: 200Mi
       cpu: 200m
 - name: gitlab-housekeeping
+  sleep_duration_secs: 600
```
maorfr, line 207: in helm we should use `sleepDurationSecs`

maorfr (approved):
LGTM

PR #1645, opened 2021-06-17
[APPSRE-3369]Improve error handling and reporting in requests_sender

Piojo:
If, instead of relying on return values we use exceptions, we can
pass a lot of context to the part of the stack best suited to handle
it or to produce the most useful reporting.

So, I have replaced `Popen` with `run` in `gpg_encrypt`, so that I can
get such an exception while removing code.  I have not updated
`gpg_key_valid` since I'd need to update its callers and they are out
of the scope of this task.

I have also replaced some calls to `defer` with a more Pythonic
context manager.  In this case I have adapted `gpg_key_valid`, for
consistency and because it doesn't change its behaviour or API.  This
removes some code, more notably two imports.

Finally, I have tweaked the integration's `run` to keep the happy path
uninterrupted - that's the beauty of exceptions.  We also guarantee
that we perform as much work as possible, reporting failures but
continuing processing.

### DIFF
--- a/reconcile/requests_sender.py
+++ b/reconcile/requests_sender.py
@@ -77,13 +77,13 @@ def run(dry_run):
             org_username = user['org_username']
             logging.info(['send_credentials', org_username, credentials_name])
 
+            request_name = credentials_request_to_send['name']
+            names = [org_username]
+            subject = request_name
+            encrypted_credentials = get_encrypted_credentials(
+                credentials_name, user, settings, smtp_client
+            )
             if not dry_run:
-                request_name = credentials_request_to_send['name']
-                names = [org_username]
-                subject = request_name
-                encrypted_credentials = get_encrypted_credentials(
-                    credentials_name, user, settings, smtp_client
-                )
                 body = MESSAGE_TEMPLATE.format(
                     request_name, credentials_name, encrypted_credentials
                 )
--- a/reconcile/requests_sender.py
+++ b/reconcile/requests_sender.py
@@ -30,7 +30,7 @@ Encrypted credentials:
 """
 
 
-def get_ecrypted_credentials(credentials_name, user, settings, smtp_client):
+def get_encrypted_credentials(credentials_name, user, settings, smtp_client):
     credentials_map = settings['credentials']
     credentials_map_item = \
         [c for c in credentials_map if c['name'] == credentials_name]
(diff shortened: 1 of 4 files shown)


reconcile/test/test_requests_sender.py
```diff
@@ -0,0 +1,107 @@
+from unittest import TestCase
+from unittest.mock import patch, MagicMock
+
+import reconcile.queries as queries
+import reconcile.requests_sender as integ
+import sys
+
+
+class TestRunInteg(TestCase):
```
maorfr, line 10: our first tests for an integration! this is very cool and educating.

reconcile/requests_sender.py
```diff
@@ -69,33 +71,35 @@ def run(dry_run):
...
+                subject = request_name
+                encrypted_credentials = get_encrypted_credentials(
+                    credentials_name, user, settings, smtp_client
+                )
+                body = MESSAGE_TEMPLATE.format(
+                    request_name, credentials_name, encrypted_credentials
+                )
+                smtp_client.send_mail(names, subject, body)
+                state.add(request_name)
+        except KeyError:
```
maorfr, line 92: i understand where you are coming from here. fwiw, we have schema validation that verifies all our assumptions, and only `public_gpg_key` is not mandatory. it's only a matter of taste, but i personally prefer short(er?) try blocks, and only where an issue may occur. leaving it for your judgement, as this is perfectly fine the way it is.

reconcile/utils/gpg.py
```diff
@@ -29,18 +28,18 @@ def gpg_key_valid(public_gpg_key, defer=None):
...
-    gnupg_home_dir = tempfile.mkdtemp()
-    defer(lambda: shutil.rmtree(gnupg_home_dir))
-    proc = Popen(['gpg', '--homedir', gnupg_home_dir],
-                 stdin=PIPE,
-                 stdout=PIPE,
-                 stderr=STDOUT)
-    out = proc.communicate(public_gpg_key_dec)
-    if proc.returncode != 0:
-        return False, out
+    with tempfile.TemporaryDirectory() as gnupg_home_dir:
```
maorfr, line 28: i couldn't agree with this change more

maorfr (approved):
LGTM

your commits are amazing.

reconcile/requests_sender.py
Piojo: Yeah, but if I capture the exception here I can avoid an ugly if that breaks the flow of this block. This `try` is on the long end, also on the limit of what I find acceptable, TBH.

PR #1646, opened 2021-06-17
[openshift-saas-deploy-trigger] update state even if not triggered

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3375

in all openshift-saas-deploy-trigger integrations, we update a state based on the last triggered spec.
sometimes, there will be multiple specs that would trigger the same deployment. in such cases, we only want to trigger the deployment once. however, we still need to persist the spec to the state so this deployment is not triggered again in the next iteration.

this is actually a regression. here is the a previous working version: https://github.com/app-sre/qontract-reconcile/blob/864a5d500da0300177004d1ef411013297a8e6d3/reconcile/openshift_saas_deploy_trigger_configs.py#L60-L63

summary: even if a deployment is skipped because it was already triggered, we still need to persist the spec to the state.

### DIFF
--- a/reconcile/openshift_saas_deploy_trigger_base.py
+++ b/reconcile/openshift_saas_deploy_trigger_base.py
@@ -295,8 +295,6 @@ def _trigger_tekton(spec,
                       resource_type=tkn_trigger_resource.kind,
                       resource=tkn_trigger_resource,
                       wait_for_namespace=False)
-            if not dry_run:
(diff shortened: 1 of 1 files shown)


maorfr:
note for the reviewer: yes, we need tests for this. it will take me a long time to write them properly and this is quite urgent.

Piojo (approved):

rporres (commented):
LGTM too

PR #1649, opened 2021-06-21
quay-mirror: increase default sleep duration

jfchevrette:
Increase the default sync interval from 30 seconds to 300 (5 minutes)


jmelis (approved):
LGTM

PR #1650, opened 2021-06-21
Slow down Quay mirror variable even more

Piojo:
Because 300 seconds may not be enough.


jfchevrette:
The limit is 50k per 24h so I think 30s to 300s (10x less requests give or take) will be fine for now and have minimal disruption for any tenant who might rely on images being available fast 

https://www.docker.com/increase-rate-limits

jmelis:
closing this PR

PR #1651, opened 2021-06-22
run deep sync every 24 hours

jmelis:


apahim:
lgtm

apahim:
should we make that an env var?

jmelis:
> should we make that an env var?
It should be, yes, but let's deal with that separately.
@apahim can you approve?

apahim (approved):

PR #1660, opened 2021-06-25
[user-validator] key singer should match with user org email

BumbleFeng:

### DIFF
--- a/reconcile/user_validator.py
+++ b/reconcile/user_validator.py
@@ -3,6 +3,7 @@ import logging
 
 from reconcile.utils.gpg import gpg_key_valid
 import reconcile.queries as queries
+from reconcile.utils.smtp_client import SmtpClient
 
 
 QONTRACT_INTEGRATION = 'user-validator'
@@ -28,10 +29,13 @@ def validate_users_single_path(users):
 
 def validate_users_gpg_key(users):
     ok = True
+    settings = queries.get_app_interface_settings()
+    smtp_client = SmtpClient(settings=settings)
     for user in users:
         public_gpg_key = user.get('public_gpg_key')
         if public_gpg_key:
-            gpg_ok, error_message = gpg_key_valid(public_gpg_key)
+            recipient = smtp_client.get_recipient(user['org_username'])
+            gpg_ok, error_message = gpg_key_valid(public_gpg_key, recipient)
             if not gpg_ok:
                 msg = \
                     'invalid public gpg key for user {}: {}'.format(
--- a/reconcile/user_validator.py
+++ b/reconcile/user_validator.py
@@ -35,11 +35,12 @@ def validate_users_gpg_key(users):
         public_gpg_key = user.get('public_gpg_key')
         if public_gpg_key:
             recipient = smtp_client.get_recipient(user['org_username'])
-            gpg_ok, error_message = gpg_key_valid(public_gpg_key, recipient)
-            if not gpg_ok:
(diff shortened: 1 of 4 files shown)


reconcile/utils/gpg.py
```diff
@@ -6,9 +6,10 @@
 ERR_SPACES = 'key has spaces in it'
 ERR_EQUAL_SIGNS = 'equal signs should only appear at the end of the key'
 ERR_BASE64 = 'could not perform base64 decode of key'
+ERR_SIGNER = 'key signer dose not match with user org email address '
 
 
-def gpg_key_valid(public_gpg_key):
+def gpg_key_valid(public_gpg_key, recepient=None):
```
Piojo, line 12: s/recepient/recipient/

reconcile/user_validator.py
```diff
@@ -44,9 +46,11 @@ def validate_users_gpg_key(users):
 
 def run(dry_run):
     users = queries.get_users()
+    settings = queries.get_app_interface_settings()
+    smtp_client = SmtpClient(settings=settings)
 
     single_path_ok = validate_users_single_path(users)
-    gpg_ok = validate_users_gpg_key(users)
+    gpg_ok = validate_users_gpg_key(users, smtp_client)
```
Piojo, line 53: Pass the `org_username` directly, the gpg-handling methods have no business dealing with SMTP clients.

reconcile/user_validator.py
```diff
@@ -26,12 +27,13 @@ def validate_users_single_path(users):
...
 
-def validate_users_gpg_key(users):
+def validate_users_gpg_key(users, smtp_client):
     ok = True
     for user in users:
         public_gpg_key = user.get('public_gpg_key')
         if public_gpg_key:
-            gpg_ok, error_message = gpg_key_valid(public_gpg_key)
+            recepient = smtp_client.get_recipient(user['org_username'])
+            gpg_ok, error_message = gpg_key_valid(public_gpg_key, recepient)
```
Piojo, line 36: The interface for `gpg_key_valid` kills me. In many other places we return the `ok` flag as the last element of the result tuple. And, if the next caller makes that assumption and does:

```python
message, ok = gpg_key_valid(public_gpg_key, recipient)
```
there is nothing to detect the mistake. No linter, no crashing until much farther in the code... just wrong results, and most tests wouldn't even catch it.

reconcile/utils/gpg.py
```diff
@@ -34,6 +35,10 @@ def gpg_key_valid(public_gpg_key):
         if proc.returncode != 0:
```
Piojo, line 35: I'd very much prefer if this were:
```python
if ' ' in stripped_public_gpg_key:
    raise ValueError(ERR_SPACES)
```

reconcile/utils/gpg.py
```diff
@@ -34,6 +35,10 @@ def gpg_key_valid(public_gpg_key):
...
             return False, out
 
+        if recepient and recepient not in str(out):
```
Piojo, line 38: And I'd very much rather this be:

```python
if recipient and recipient not in str(out):
    raise ValueError(f"{ERR_SIGNER} {recipient}")
```

Piojo: But I guess this is a useless rant right now, because we'd have to adapt other callers of this function.

Piojo (approved):
Fix the nits before merging, please.

PR #1664, opened 2021-06-28
[terraform-resources] add support for a full run with account_name

maorfr:

depends on #1666

until now we couldn't run the integration fully with `--account-name` as it would have removed Secrets from other accounts.
with this PR, we now filter the Secrets based on their called (the name of the aws account), and it enables to run the integration per account.

future work:
1. create terraform-resources-wrapper, which will run the integration in parallel per account (makes tf init, tf plan, tf apply run per account and not wait for all accounts to finish each stage)
2. implement a sharded run per account (https://issues.redhat.com/browse/APPSRE-2963)

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -254,11 +254,9 @@ def populate_oc_resources(spec, ri, account_name):
         logging.error(msg)
 
 
-def fetch_current_state(dry_run, namespaces, thread_pool_size,
+def fetch_current_state(namespaces, thread_pool_size,
                         internal, use_jump_host, account_name):
     ri = ResourceInventory()
-    if dry_run:
-        return ri, None
     settings = queries.get_app_interface_settings()
     oc_map = OC_Map(namespaces=namespaces, integration=QONTRACT_INTEGRATION,
                     settings=settings, internal=internal,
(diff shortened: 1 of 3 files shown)


reconcile/utils/terraform_client.py
```diff
@@ -231,13 +231,16 @@ def get_terraform_output_secrets(self):
...
+    def populate_desired_state(self, ri, oc_map, tf_namespaces, account_name):
         self.init_outputs()  # get updated output
 
         # Dealing with credentials for RDS replicas
         replicas_info = self.get_replicas_info(namespaces=tf_namespaces)
 
         for account, output in self.outputs.items():
+            if account_name and account != account_name:
+                continue
+
```
Piojo, line 243: It'd be worth adding a test for this. According to the PR description this is pretty critical.

reconcile/terraform_resources.py
```diff
@@ -250,11 +250,9 @@ def populate_oc_resources(spec, ri):
         logging.error(msg)
 
 
-def fetch_current_state(dry_run, namespaces, thread_pool_size,
+def fetch_current_state(namespaces, thread_pool_size,
                         internal, use_jump_host):
     ri = ResourceInventory()
-    if dry_run:
```
Piojo, line 256: +1 to reading stuff even under dry_run. :)

Piojo (approved):

reconcile/terraform_resources.py
```diff
@@ -386,29 +390,19 @@ def run(dry_run, print_only=False,
...
-    if account_name:
-        cleanup_and_exit(tf)
-
-    tf.populate_desired_state(ri, oc_map, tf_namespaces)
+    tf.populate_desired_state(ri, oc_map, tf_namespaces, account_name)
 
-    actions = ob.realize_data(dry_run, oc_map, ri)
+    actions = ob.realize_data(dry_run, oc_map, ri, caller=account_name)
 
     disable_keys(dry_run, thread_pool_size,
```
BumbleFeng, line 402: We should also add support for  aws_iam_keys to run with account_name

maorfr: very good idea! #1673

maorfr: done!

reconcile/terraform_resources.py
maorfr: reverting this as we can't really compare the diffs in secrets in dry-run mode (they rely on the terraform state)

maorfr:
now that #1666 is merged and promoted, this one is ready to go!

rporres (approved):

PR #1665, opened 2021-06-28
[terraform-resources] introduce wrapper integration

maorfr:

A wrapper integration is such that runs the wrapped integration in parallel per some data.
In this PR, we add a wrapper integration for terraform-resources to run it per aws account.

depends on #1664 to be able to run terraform-resources per account.

the limitation in terraform-resources is that it runs `terraform init`, `terraform plan` and `terraform apply` for all accounts in parallel, and waits for all of them to finish each stage before moving on to the next stage.

with this PR, we now run init+plan+apply per aws account, reaching faster results per account (the waiting time moves to the end of the integration).

If we use the round robin sharding method with the accounts sorted, we guarantee that each shard will work on only one AWS account (as long as the number of shards equals the number of accounts). If a new AWS account is added to app-interface, it will run in the first shard again, until we increase the number of shards.


reconcile/terraform_resources_wrapper.py
```diff
@@ -0,0 +1,63 @@
...
+        exit_code = e.code
+    return exit_code
+
+
+def run(dry_run, print_only=False,
+        enable_deletion=False, io_dir='throughput/',
+        thread_pool_size=10, internal=None, use_jump_host=True,
+        light=False, vault_output_path=''):
+    account_names = [a['name'] for a in queries.get_aws_accounts()]
+    if len(account_names) == 0:
```
Piojo, line 46: It is shorter and I find it more readable:
```python
if not account_names:
```
(which in English reads almost as "if there are no account names").

Piojo (approved):

reconcile/terraform_resources_wrapper.py
maorfr: done!

PR #1672, opened 2021-06-30
Add instrumented classes for common usage

Piojo:
As part of APPSRE-3385 we need to review measure how often we reach
to the external registry, and how effective our cache is.

For the first part, we keep `sretools.container.Image` decoupled from
Prometheus by subclassing it inside this repository, and instrumenting
the method that matters in this context: `_get_requests`. We make
`quay_mirror` use this instrumented version.

Then, we add a simple dictionary-like class that reports on hits,
misses and size. This will become our cache. We don't subclass `dict`
here since we may need to use a more sophisticated, persistent cache
with the same look and feel.

TODO:

* Is this the right way to use Prometheus metrics?

### DIFF
--- a/reconcile/utils/instrumented_wrappers.py
+++ b/reconcile/utils/instrumented_wrappers.py
@@ -1,0 +1,81 @@
+from prometheus_client import Counter, Gauge
+
+from sretoolbox.container import Image
+
+
+class InstrumentedImage(Image):
+    """Normal Image that exposes the count of reachouts to external
+    registries.
+
+    It helps us understand the performance of our caches and predict
+    our mirroring-related costs.
+
+    """
+    _registry_reachouts = Counter(
+        name='qontract_reconcile_registry_reachouts',
+        documentation='Number GET requests on public image registries',
+    )
+    #    labelnames=['integrations', 'shards', 'shard_id'])
+
+    def _request_get(self, url):
+        # TODO: Do I need to raise by labels? I don't think so
+        self._registry_reachouts.inc()
+        super()._request_get(url)
+
+
+class InstrumentedCache:
+    _cache_hits = Counter(
+        name='qontract_reconcile_cache_hits',
+        documentation='Number of hits to this cache',
+        labelnames=['integrations', 'shards', 'shard_id']
+    )
+
+    _cache_misses = Counter(
+        name='qontract_reconcile_cache_misses',
+        documentation='Number of misses on this cache',
+        labelnames=['integrations', 'shards', 'shard_id']
+    )
+
+    _cache_size = Gauge(
+        name='qontract_reconcile_cache_size',
+        documentation='Size of the cache',
+        labelnames=['integration', 'shards', 'shard_id']
+    )
+
+    def __init__(self, integration_name, shards, shard_id):
+        self.integraton_name = integration_name
+        self.shards = shards
+        self.shard_id = shard_id
+
+        self._hits = self._cache_hits.labels(
+            integration=integration_name,
+            shards=shards,
+            shard_id=shard_id
+        )
+        self._misses = self._cache_misses.labels(
+            integration=integration_name,
+            shards=shards,
+            shard_id=shard_id
+        )
+        self._size = self._cache_size.labels(
+            integration=integration_name,
+            shards=shards,
+            shard_id=shard_id
+        )
+
+        self._cache = {}
+
+    def __getitem__(self, item):
+        if item in self._cache:
+            self._hits.inc()
+        else:
+            self._misses.inc()
+        return self._cache[item]
+
+    def __setitem__(self, key, value):
+        self._cache[key] = value
+        self._size.inc()
+
+    def __delitem__(self, key):
+        del self._cache[key]
+        self._size.dec(1)
--- a/reconcile/utils/instrumented_wrappers.py
+++ b/reconcile/utils/instrumented_wrappers.py
@@ -42,7 +42,7 @@ class InstrumentedCache:
     )
 
     _cache_misses = Counter(
-        name='qontract_reconcile_cache_misses',
+        name='qontract_reconcile_cache_misses_total',
         documentation='Number of misses on this cache',
         labelnames=['integration', 'shards', 'shard_id']
     )
--- a/reconcile/utils/instrumented_wrappers.py
+++ b/reconcile/utils/instrumented_wrappers.py
@@ -23,13 +23,14 @@ class InstrumentedImage(Image):
     _registry_reachouts = Counter(
         name='qontract_reconcile_registry_get_manifest_total',
         documentation='Number of GET requests on public image registries',
-        labelnames=['integration', 'shard', 'shard_id'])
+        labelnames=['integration', 'shard', 'shard_id', 'registry'])
 
     def _get_manifest(self):
         self._registry_reachouts.labels(
             integration=INTEGRATION_NAME,
             shard=SHARDS,
-            shard_id=SHARD_ID
+            shard_id=SHARD_ID,
+            registry=self.registry,
         ).inc()
         super()._get_manifest()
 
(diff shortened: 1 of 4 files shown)


reconcile/utils/instrumented_wrappers.py
```diff
@@ -0,0 +1,81 @@
...
+    our mirroring-related costs.
+
+    """
+    _registry_reachouts = Counter(
+        name='qontract_reconcile_registry_reachouts',
+        documentation='Number GET requests on public image registries',
+    )
+    #    labelnames=['integrations', 'shards', 'shard_id'])
+
+    def _request_get(self, url):
```
apahim, line 20: We found out that not all requests count agains the quota. requests to `https://auth.docker.io/token` and `https://registry-1.docker.io/v2/<repo>/<image>/tags/list` won't, but requests to `https://registry-1.docker.io/v2/<repo>/<image>/manifests/<tag>` will.

We might then want to either split the caller function per endpoint or track the different endpoints with different metric labels. Main point being to understand how many calls that count against quota we are doing.

Piojo: Which requests do? Just manifest ones? In that case I could override just `_get_manifest`. But if there are any other calls that count against our quota, I'd rather override this one and accept a non-zero baseline. It keeps the code simple and avoids any bugs from delivering a misleading metric.

apahim: `_get_manifest` calls `_request_get`, which then calls both `/token` and `/manifests`. But only the call against `/manifests` will count against the quota, aparently.

apahim: >  I'd rather override this one and accept a non-zero baseline. 

I believe it depends on what you want. Do you want to track the number of total requests to the docker API or the number of requests that count against 5k/day quota?

Piojo: In that case, I'll override `_get_manifest` - it'll give the correct numbers. Give me a jiffy.

apahim: I don't think it will. Now you left out /tags and the related /token calls. For all the API calls, code was good before.

Piojo: > Do you want to track the total number of requests to the docker API or the number of requests that count against 5k/day quota?

It depends. I prefer to count calls that don't add to the quota than leave out calls that do. If we're confident that out of the methods in `Image` only `_get_manifest` adds to the quota, my current version is the right one. Otherwise, indeed, I'll go back to override `_request_get` - and get a number that isn't perfect but is strongly correlated with our quota consumption.

apahim: `_get_manifest()` calls `/token` and `/manifest`. `/token` doesn't count, `/manifest` does count.

apahim: What'I'm trying to say is: the way `_request_get` is implemented, it calls `/token` whenever it needs to, and that call doesn't count for the quota.

apahim: Actually, I was checking the current code. we should be fine now. Sorry about the noise, I will review the latest version when you drop the WiP from it.

reconcile/utils/instrumented_wrappers.py
```diff
@@ -0,0 +1,109 @@
...
+class InstrumentedImage(Image):
+    """Normal Image that exposes the count of reachouts to external
+    registries.
+
+    It helps us understand the performance of our caches and predict
+    our mirroring-related costs.
+
+    """
+    _registry_reachouts = Counter(
+        name='qontract_reconcile_registry_reachouts',
```
rporres, line 24: Not a fan of this name. If you're measuring the number of GET requests, then that is how you metric should be called. A proposal follows:
```suggestion
        name='qontract_reconcile_registry_get_manifest_total',
```

reconcile/utils/instrumented_wrappers.py
```diff
@@ -0,0 +1,109 @@
...
+        documentation='Number of GET requests on public image registries',
```
rporres, line 25: why public?

reconcile/utils/instrumented_wrappers.py
```diff
@@ -0,0 +1,109 @@
...
+            integration=INTEGRATION_NAME,
+            shard=SHARDS,
+            shard_id=SHARD_ID
+        ).inc()
+        super()._get_manifest()
+
+
+class InstrumentedCache:
+    _cache_hits = Counter(
+        name='qontract_reconcile_cache_hits',
```
rporres, line 39: Prometheus convention is for counters to end in `_total`
```suggestion
        name='qontract_reconcile_cache_hits_total',
```

reconcile/utils/instrumented_wrappers.py
```diff
@@ -0,0 +1,109 @@
...
+        documentation='Number of hits to this cache',
+        labelnames=['integration', 'shards', 'shard_id']
+    )
+
+    _cache_misses = Counter(
+        name='qontract_reconcile_cache_misses',
```
rporres, line 45:
```suggestion
        name='qontract_reconcile_cache_misses_total',
```

reconcile/utils/instrumented_wrappers.py
```diff
@@ -0,0 +1,109 @@
...
+        documentation='Number of misses on this cache',
+        labelnames=['integration', 'shards', 'shard_id']
+    )
+
+    _cache_size = Gauge(
+        name='qontract_reconcile_cache_size',
+        documentation='Size of the cache',
```
rporres, line 52: Reading the below code this appears to be the number of entries in the cache.  I would name this metric accordingly

reconcile/utils/instrumented_wrappers.py
```diff
@@ -0,0 +1,109 @@
...
+    _registry_reachouts = Counter(
+        name='qontract_reconcile_registry_reachouts',
+        documentation='Number of GET requests on public image registries',
+        labelnames=['integration', 'shard', 'shard_id'])
+
+    def _get_manifest(self):
+        self._registry_reachouts.labels(
+            integration=INTEGRATION_NAME,
+            shard=SHARDS,
+            shard_id=SHARD_ID
```
rporres, line 32: Since we may mirror from multiple registries, I would add the `Image.registry` label here.

rporres (changes requested):
The implementation looks great, only comments about metric names.

In order to follow the current code shape, I would place the metric definitions in `reconcile/util/metrics` so that we have only one place where metrics are defined and it's easier to track

reconcile/utils/instrumented_wrappers.py
```diff
@@ -0,0 +1,109 @@
...
+        self._size.inc()
+
+    def __delitem__(self, key):
+        del self._cache[key]
+        self._size.dec(1)
+
+
+class InstrumentedSkopeo(Skopeo):
+    _copy_count = Counter(
+        name='qontract_reconcile_skopeo_copy',
```
rporres, line 97:
```suggestion
        name='qontract_reconcile_skopeo_copy_total',
```

reconcile/utils/instrumented_wrappers.py
Piojo: I can change to `external`, the wording is poor indeed.

Piojo:
> I would place the metric definitions in reconcile/util/metrics so that we have only one place where metrics are defined and it's easier to track

I considered it, but then every integration would expose these metrics, even those that don't need this cache or this specialized `Image`. If you think that overhead is still worth it, I'll move them.

reconcile/utils/metrics.py
```diff
@@ -17,3 +16,32 @@
...
                            buckets=(60.0, 150.0, 300.0, 600.0, 1200.0, 1800.0,
                                     2400.0, 3000.0, float("inf")))
+
+registry_reachouts = Counter(
+    name='qontract_reconcile_registry_get_manifest_total',
+    documentation='Number of GET requests on image registries',
+    labelnames=['integration', 'shard', 'shard_id', 'registry'])
+
+cache_hits = Counter(
+    name='qontract_reconcile_cache_hits_tocal',
```
rporres, line 26:
```suggestion
    name='qontract_reconcile_cache_hits_total',
```

rporres (approved):
I see that you finally moved the metrics. I'm fine with both approaches, I get your point.

PR #1673, opened 2021-06-30
[aws-iam-keys] add support to run per account

maorfr:

required for #1664

this PR enables running the `aws-iam-keys` integration for a specific AWS account. since this integration is called from within `terraform-resources`, which can be run for a specific account, this will be beneficial.

### DIFF
--- a/reconcile/aws_iam_keys.py
+++ b/reconcile/aws_iam_keys.py
@@ -43,8 +43,13 @@ def cleanup(working_dirs):
 
 @defer
 def run(dry_run, thread_pool_size=10,
-        disable_service_account_keys=False, defer=None):
+        disable_service_account_keys=False, account_name=None, defer=None):
     accounts = queries.get_aws_accounts()
+    if account_name:
+        accounts = [a for a in accounts if a['name'] == account_name]
+        if not accounts:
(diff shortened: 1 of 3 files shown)


reconcile/aws_iam_keys.py
```diff
@@ -43,8 +43,13 @@ def cleanup(working_dirs):
 
 @defer
 def run(dry_run, thread_pool_size=10,
-        disable_service_account_keys=False, defer=None):
+        disable_service_account_keys=False, account_name=None, defer=None):
     accounts = queries.get_aws_accounts()
+    if account_name:
```
Piojo, line 48: Can we please add a test for this kind of filter? Other than this, it looks good.

maorfr: can we do it without writing a test for the entire integration?

BumbleFeng (approved):

reconcile/aws_iam_keys.py
Piojo: Mmm... you could move this logic to its own method, say `filter_accounts` and test that. Or it's not worth it and we can write a full test suite in a different PR. Up to you.

maorfr: extracted to function and added tests!

PR #1674, opened 2021-06-30
add support for additional labels in metrics (+implement for terraform-resources)

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3424

It will be useful to be able to expose additional labels on metrics based on the data that is being processed.
One use case would be - if terraform-resources is running per account and is running slowly, we would want to know which is the problematic account.

this is just an implementation proposal that works. feel free to review it and suggest improvements to things i'm obviously missing.

note: this works.

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1051,7 +1051,8 @@ def terraform_resources(ctx, print_only, enable_deletion,
                     ctx.obj, print_only,
                     enable_deletion, io_dir, thread_pool_size,
                     internal, use_jump_host, light, vault_output_path,
-                    account_name=account_name)
+                    account_name=account_name,
+                    extra_labels=ctx.obj['extra_labels'])
 
 
 @integration.command()
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1052,7 +1052,7 @@ def terraform_resources(ctx, print_only, enable_deletion,
                     enable_deletion, io_dir, thread_pool_size,
                     internal, use_jump_host, light, vault_output_path,
                     account_name=account_name,
-                    extra_labels=ctx.obj['extra_labels'])
+                    extra_labels=ctx.obj.get('extra_labels', {}))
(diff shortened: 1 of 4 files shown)


maorfr:
sharding_key is not bad at all!

reconcile/cli.py
```diff
@@ -1049,7 +1049,8 @@ def terraform_resources(ctx, print_only, enable_deletion,
                     ctx.obj, print_only,
                     enable_deletion, io_dir, thread_pool_size,
                     internal, use_jump_host, light, vault_output_path,
-                    account_name=account_name)
+                    account_name=account_name,
+                    extra_labels=ctx.obj['extra_labels'])
```
rporres, line 1053: Just one doubt. If we run the integration manually, is this set to `None` by some Click internal magic?

reconcile/utils/metrics.py
```diff
@@ -2,13 +2,18 @@
 from prometheus_client import Histogram
 
 
-run_time = Gauge(name='qontract_reconcile_last_run_seconds',
-                 documentation='Last run duration in seconds',
-                 labelnames=['integration', 'shards', 'shard_id'])
+extra_labels = {'shard_key': None}
```
maorfr, line 5: @rporres it's set to None right here

reconcile/cli.py
maorfr: i'll take another look at this

reconcile/utils/metrics.py
rporres: If I understand correctly, that is imported from `run-integegration.py` which is not in the manual execution path.

reconcile/cli.py
maorfr: you are correct. to make things simpler, i changed this to `ctx.obj.get('extra_labels', {})`, as all we're doing inside the integration is setting a value on a dict.

PR #1680, opened 2021-07-04
[openshift-saas-deploy] add support to delete target

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3425

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/21589

this PR adds support for a `delete` boolean in a saas file target to indicate that resources associated to that target should be removed.

the implementation is quite simple: if the target has `delete: true`, we don't add it's resources to the desired state. since they will be found in the current state and not in the desired state, they will be removed.

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -1285,6 +1285,7 @@ SAAS_FILES_QUERY_V1 = """
         parameters
         upstream
         disable
+        delete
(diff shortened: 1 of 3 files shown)


jmelis (approved):
LGTM

PR #1685, opened 2021-07-06
[ldap-users] fix unable to send message, socket is not open error

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3388

was able to reproduce and isolate the issue to the connection (client) being bound or not.
```
>>> _client.bind()
True
>>> _client.bound
True
>>> _client.search(search_filter, '(objectclass=person)')
True
>>> _client.unbind()
True
>>> _client.bound
False
>>> _client.search(search_filter, '(objectclass=person)')
  File "/home/mafriedm/.local/lib/python3.6/site-packages/ldap3/core/connection.py", line 853, in search
    response = self.post_send_search(self.send('searchRequest', request, controls))
  File "/home/mafriedm/.local/lib/python3.6/site-packages/ldap3/strategy/base.py", line 336, in send
    raise LDAPSocketOpenError(self.connection.last_error)
ldap3.core.exceptions.LDAPSocketOpenError: unable to send message, socket is not open
```


rporres:
I guess this happens because there's an idle timeout somewhere. Since the connection is not recreated every time that the integration starts, the idle timeout kicks as soon as the integration does something and stays without doing something for a while.

I think this is going to work, but has the problem of needing to be repeated for every method in the module. In the past I've fixed this kind of issue turning the client into a class and defining a `client` method that implements the keep alive logic by calling the `bind` method. Since I don't think this is going to be needed here as no more methods are likely to be needed, please add a comment in the code stating that in case that a new method is added, it should implement the reconnect logic (unless you want to refactor this code and convert into a class).

maorfr:
going to close this and hide it to enable a learning experice in APPSRE-3430

PR #1687, opened 2021-07-07
[cluster-deploymet-mapper] introduce new integration

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3458

this command will get a list of all customer clusters (external id: provision shard) and put it in a vault secret.
this is required for an SREP breakglass process which we are assisting with (OCM is down is the scenario).

this integration will run once a day as a CronJob and will save the mapping to Vault.


Piojo (commented):
Can we add a small help text?

maorfr:
changed directions a bit and made this an integration instead of a cli command

jmelis:
+1 to the approach, but haven't reviewed the code

PR #1688, opened 2021-07-11
[openshift-upgrade-watcher] fix missing upgrades due to resource name

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3376

due to the UpgradeConfig resource changing it's name in https://github.com/openshift/managed-upgrade-operator/pull/224, we are no longer getting notifications for upgrades.
we also added internal OSD clusters, which means this integration needs to run internally.


rporres:
LGTM, but I'd very much appreciate this being two different PRs, as 04a14cb doesn't have anything to do with this PR title

rporres (approved):

PR #1697, opened 2021-07-14
[openshift-saas-deploy] add support for cluster-admin automation token

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3457

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/22078

in order to be able to apply resources to facilitate this ticket, we need to apply them as cluster admins. with this PR, we can add `clusterAdmin: true` to a saas file and the credentials will be used from the `clusterAdminAutomationToken` key instead of the `automationToken` key of a cluster.


maorfr:
merging to unblock further work. happy to get a post-merge review!

PR #1701, opened 2021-07-18
do not use jump host in internal openshift integrations

maorfr:
we use VPC peerings when running integrations inside our clusters instead of a jump host.


BumbleFeng:
Can we retry this now since hive-v3 clusters are all gone now? @maorfr

maorfr:
sure @BumbleFeng

PR #1703, opened 2021-07-19
[openshift-saas-deploy-trigger] error handle

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3484

this PR helps the trigger integrations (specifically upstream-jobs) to handle an unreachable instance and still trigger deployments based on jobs from reachable instances.


rrati (approved):
lgtm

PR #1706, opened 2021-07-20
[MTSRE-117] MRs to managed tenants must be merged with "/lgtm" even after rebase

mayankshah1607:
Depends on - https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/23349

### Problem 

Currently, an ‘/lgtm’ comment from an approver is not honored after the bot performs a rebase on a MR. As a result, approvers are required to manually issue additional ‘/lgtm’s until the MR is automatically merged.

Why does this happen? The bot takes into consideration only those ‘/lgtm’ which were commented after the top commit was pushed. A rebase causes the timestamp on the commits to change to when the MR was rebased.

### Solution

Each time an ‘/lgtm’ comment is added, the changes (or the diff) that it is approving should be stored on AWS S3. When the integration collects valid ‘/lgtm’ comments, it must check if the diff corresponding to that ‘/lgtm’ matches the diff at the time of merging. This way, the bot can differentiate between a rebase and an actual change being pushed.

Signed-off-by: Mayank Shah <m.shah@redhat.com>

### DIFF
--- a/reconcile/gitlab_owners.py
+++ b/reconcile/gitlab_owners.py
@@ -1,4 +1,5 @@
 import logging
+import json
 
 from dateutil import parser as dateparser
 
@@ -8,6 +9,7 @@ from reconcile import queries
 from reconcile.utils.gitlab_api import GitLabApi
 from reconcile.utils.gitlab_api import MRState
 from reconcile.utils.repo_owners import RepoOwners
+from reconcile.utils.state import State
 
 QONTRACT_INTEGRATION = 'gitlab-owners'
 
@@ -28,19 +30,19 @@ class MRApproval:
     Processes a Merge Request, looking for matches
     between the approval messages the the project owners.
     """
-    def __init__(self, gitlab_client, merge_request, owners, dry_run):
+
+    def __init__(self, gitlab_client, state, merge_request,
+                 persist_lgtm, owners, dry_run):
         self.gitlab = gitlab_client
         self.mr = merge_request
         self.owners = owners
         self.dry_run = dry_run
+        self.state = state
+        self.persist_lgtm = persist_lgtm
 
-        commits = self.mr.commits()
-        if commits:
-            top_commit = next(commits)
-            self.top_commit_created_at = \
-                dateparser.parse(top_commit.created_at)
-        else:
-            self.top_commit_created_at = None
+        self.top_commit_created_at = \
+            self.gitlab.get_merge_request_top_commit_timestamp(
+                mr=merge_request)
 
     def get_change_owners_map(self):
         """
@@ -78,11 +80,26 @@ class MRApproval:
             if comment['body'] != '/lgtm':
                 continue
 
-            # Only interested in comments created after the top commit
-            # creation time
-            comment_created_at = dateparser.parse(comment['created_at'])
-            if comment_created_at < self.top_commit_created_at:
-                continue
+            # Check against saved diffs
+            if self.persist_lgtm:
+                current_diff = self.gitlab.get_merge_request_diffs(self.mr.iid)
+                state_key = generate_state_key(self.mr, comment)
+                try:
+                    diff = self.state.get(key=state_key)
+                    decoded_diff = json.loads(diff)
+                    if current_diff != decoded_diff:
+                        continue
+                except KeyError:
+                    continue
+                lgtms.append(comment['username'])
+
+            # Check using commit creation timestamp
+            else:
+                # Only interested in comments created after the top commit
+                # creation time
+                comment_created_at = dateparser.parse(comment['created_at'])
+                if comment_created_at < self.top_commit_created_at:
+                    continue
 
             lgtms.append(comment['username'])
         return lgtms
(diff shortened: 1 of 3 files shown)


maorfr:
i think that maybe you should approach this from `gitlab-owners`. `gitlab-housekeeping` is pretty simple - has label? merge.
what do you think?

reconcile/gitlab_housekeeping.py
```diff
@@ -209,13 +212,54 @@ def merge_merge_requests(dry_run, gl, merge_limit, rebase, insist=False,
                     logging.error('unable to merge {}: {}'.format(mr.iid, e))
 
 
+# Each '/lgtm' must store the changes (diff) it is approving
+def store_diff_when_lgtm(dry_run, gl, accounts, settings):
```
mayankshah1607, line 216: > i think that maybe you should approach this from gitlab-owners

From what I can understand, `gitlab-owners`,  has the primary responsibility of adding the approval label on MRs. It should use the LGTM -> diff mapping (on S3) to decide if an MR should have an approval label. Creating this mapping itself shouldn't be `gitlab-owner`'s responsibility IMO. 

How about creating a separate integration for mapping LGTMs to thier corresponding diffs?

maorfr: gitlab-owners works in the context of a MR. it can be the one to check if there was no new code (or, rebase) after the last lgtm label. this may help in not keeping anything in a state at all.

my proposal is - only add the label when all conditions are true. this should also be opt-in of course through the app-interface schema.

maorfr:
while this is wip, i'm seeing that you are still using a state. if you are looking at a MR, you should be able to decide if a label should be added or not just by checking if there are no new code changes after the last valid lgtm?

reconcile/gitlab_owners.py
```diff
@@ -68,20 +67,27 @@ def get_change_owners_map(self):
...
+            user = comment['user']
+            comment_id = comment['id']
+            state_key = generate_state_key(self.mr, comment)
+
+            try:
+                diff = self.state.get(key=state_key)
+                decoded_diff = json.loads(diff)
+                if current_diff != decoded_diff:
+                    continue
+            except:
```
mayankshah1607, line 90: @maorfr How we're doing it right now is using the commit creation time-stamp of the top most commit. While this approach will work when there are legitimate code changes after the last lgtm, the problem arises when the bot needs to auto-rebase an MR (because when such a rebase happens, the top commits are re-created, causing the top commit's creation time stamp to change). Even though the rebase doesn't introduce any code changes, the lgtm is simply discarded because the commits were re-created.

The challenge here is to have the integration differentiate between a legitimate code change v/s an auto-rebase (currently the bot sees both the scenarios as the same thing) so that approvers don't have to issue redundant lgtms just because there was a bot rebase.

> if a label should be added or not just by checking if there are no new code changes after the last valid lgtm?

After discussing this idea with @apahim , we decided to try out using state. How this PR checks if new changes are real changes (or just bot rebase), is that each time there is a lgtm, the diff (which this lgtm is approving) is stored as a state. At the time of decided if a label should be added, the current diff is compared with the saved diff. Now the current diff will only change if there is a real code change (and not just a bot rebase).

Do you have any ideas on how we can achieve this (differentiating between a bot rebase and legitimate changes) without creating a mapping between LGTM -> diff?

We've discussed some ideas on [this](https://docs.google.com/document/d/1yJmgDM9PNQfwMHifn4mxENsj5hsRvEx6Do5yK4iFzoQ/edit?usp=sharing) doc that do not require using State. Let me know your thoughts. :)

reconcile/gitlab_owners.py
```diff
@@ -298,15 +302,52 @@ def format_report(report):
...
+        comment_created_at = dateparser.parse(comment['created_at'])
+
+        # Store diffs for only those comments that were created after the latest
+        # commit was pushed (the top commit)
+        # This was done in case, for some reason, the diff for an '/lgtm' is not
+        # immediately stored, during which it could be possible to accidentally add
+        # commits which have not been reviewed / approved
+        if comment_created_at < top_commit_created_at:
+            continue
+
```
apahim, line 331: We should check if the key exists before trying to add it... it's cheaper (`self.exists(key)`)

reconcile/gitlab_owners.py
```diff
@@ -298,15 +302,52 @@ def format_report(report):
...
+        diffs = gl.get_merge_request_diffs(mr.iid)
+        if not dry_run:
+            try:
+                state.add(key=state_key, value=json.dumps(diffs))
+            except:
```
apahim, line 336: Let's capture specific exceptions... `except KeyError`, in this case

apahim: but this exception block is likely to disapear :)

reconcile/gitlab_owners.py
```diff
@@ -68,20 +67,25 @@ def get_change_owners_map(self):
...
-            # creation time
-            comment_created_at = dateparser.parse(comment['created_at'])
-            if comment_created_at < self.top_commit_created_at:
+            state_key = generate_state_key(self.mr, comment)
+
+            try:
+                diff = self.state.get(key=state_key)
+                decoded_diff = json.loads(diff)
+                if current_diff != decoded_diff:
+                    continue
```
apahim, line 87:
```
           diff = self.state[state_key]
     ....
            except KeyError:
```

reconcile/queries.py
```diff
@@ -911,7 +912,9 @@ def get_repos_gitlab_owner(server=''):
     server: url of the server to return. for example: https://github.com
     """
     code_components = get_code_components()
-    return [c['url'] for c in code_components
+
+    return [([c['url']], c['gitlabRepoOwners']['persistLgtm'])
```
apahim, line 916: this shoudl be a dictionary, or a nametuple, or a custom class... for the sake of readability. I mean, `('htttps://foo/bar', True)` doesn't tell much outside of this method.

mayankshah1607:
Closing it for now. Will reopen it when we can prioritize https://issues.redhat.com/browse/MTSRE-156

PR #1707, opened 2021-07-20
add logging of cluster name from within oc

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3326

some actions are logged from within `oc.py`, which is unaware of the name of the cluster. this PR adds the name of the cluster to the OC initiation and adds it to logging messages.

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -503,7 +503,7 @@ class OC:
             raise RecyclePodsInvalidAnnotationValue('should be "true"')
         if qontract_recycle != 'true':
             logging.debug(['skipping_pod_recycle_no_annotation',
-                           self.name, namespace, dep_kind])
+                           self.cluster_name, namespace, dep_kind])
             return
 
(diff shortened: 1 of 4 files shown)


reconcile/utils/oc.py
```diff
@@ -160,9 +160,10 @@ def __init__(self, namespace, resource, server,
 
 
 class OC:
-    def __init__(self, server, token, jh=None, settings=None,
+    def __init__(self, name, server, token, jh=None, settings=None,
```
rporres, line 163: `name`? That can be anything... Maybe `cluster_name` or `cluster`.

Since you're modifying the constructor for this class, it is a fantastic moment to document its usage <3

rporres (commented):
It's unfortunate that we cannot get the cluster name from the url...

reconcile/utils/oc.py
maorfr: i love adding docs

PR #1708, opened 2021-07-20
[oc] recycle pods allow controller object to not be found

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3326

from the time we collect pods and until we look for their owner some time may pass. during that time, the pod may be deleted or its owner can be deleted (like we are seeing here: https://coreos.slack.com/archives/CS0E65QCV/p1626791826140700).

to handle that we allow the owner to not be found, which will allow us to move on to the next object and successfully recycle pods.

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -561,6 +561,23 @@ class OC:
                     self._run(cmd, stdin=stdin, apply=True)
 
     def get_obj_root_owner(self, ns, obj, allow_not_found=False):
+        """Get object root owner (recursively find the top level owner).
+        - Returns obj if it has no ownerReferences
+        - Returns obj if all ownerReferences have controller set to false
+        - Returns obj if controller is true, allow_not_found is true,
+          but referenced object does not exist
+        - Throws an exception if controller is true, allow_not_found false,
+          but referenced object does not exist
+        - Recurses if controller is true and referenced object exists
+
+        Args:
+            ns (string): namespace of the object
+            obj (dict): representation of the object
+            allow_not_found (bool, optional): allow owner to be not found
+
+        Returns:
+            dict: representation of the object's owner
+        """
         refs = obj['metadata'].get('ownerReferences', [])
         for r in refs:
             if r.get('controller'):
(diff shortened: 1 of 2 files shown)


reconcile/utils/oc.py
```diff
@@ -544,12 +545,18 @@ def recycle_pods(self, dry_run, namespace, dep_kind, dep_resource):
                     stdin = json.dumps(obj, sort_keys=True)
                     self._run(cmd, stdin=stdin, apply=True)
 
-    def get_obj_root_owner(self, ns, obj):
+    def get_obj_root_owner(self, ns, obj, allow_not_found=False):
```
rporres, line 548: If I understand correctly this function, it will:

1. return `obj` if no `ownerReferences`
2. return `obj` if `ownerReferences` are all set to `false`
3. return `obj` if `controller` exists and `allow_not_found` is set to `true` but referenced object does not exist
4. throw an exception  if `controller` exists and `allow_not_found` is set to `false` but referenced object does not exist
5. recurse if `controller` is true and referenced object exists

1 and 2 are fine if we're recursing, but it looks weird if the original object has no owner references. Most likely that edge case should be controlled in the caller

3 is weird. We find that the owner of our pod is no longer there so we return the pod `obj`? Shouldn't we return `{}` to tell the caller that the owner has been deleted?

As a final comment an as a testimony to the minutes that I have been reading this, this function:

* should have a nice piece of documentation.
* deserves a test since it is a critical one and we're inside a class where we can mock methods  😍

maorfr: about 3 - it's also fine. if an object had a parent and the parent is gone, it's like the object doesn't have a parent anymore, so it's like 1/2.

i'll add docs and tests

rporres: Not having a parent is a different situation than having a deleted parent. In your case the caller may want to do the same, but they are different situations :)

maorfr: the caller did specify `allow_not_found=True`, so it means they want this behavior.

rporres: no, they want to avoid non-existing parent make the method throw an error 😂 

Anyway, I already got some tests and documentation, so I'll give you this ;-)

rporres (approved):
There are no tests on recursion, but those look quite hard to write, so let's merge this. We're in a better position than we were.

PR #1709, opened 2021-07-21
[openshift-saas-deploy] validate_data add stop condition to Job validation

maorfr:
in the openshift-saas-deploy integration, after we deploy resources we validate some of them. one of the validated kinds is `Job`.

the current validation assums a happy path where all jobs succeed eventually. so in this optimistic spirit, it waits for the job to succeed, even when the job had stopped retrying (due to a backoff policy for example).

this PR adds this stop condition to the Job validation and raises an exception that is not retried, which will lead to the integration finishing. this will avoid situations like the one described in https://coreos.slack.com/archives/CCRND57FW/p1626810703004500?thread_ts=1626803756.494800&cid=CCRND57FW.


reconcile/openshift_base.py
```diff
@@ -511,6 +515,9 @@ def validate_data(oc_map, actions):
                     if conditions:
                         logging.info(f'Job conditions are: {conditions}')
                         logging.info(yaml.safe_dump(conditions))
+                        for c in conditions:
+                            if c.get('type') == 'Failed':
+                                raise ValidationErrorJobFailed(name)
```
Piojo, line 520: I tend to prefer this style:

```
if any(c.get('type') == 'Failed') for c in conditions):
    raise...
```
Also, is there any other information about this failed job that may help the on call/IC troubleshoot problems? I really like descriptive exceptions.

maorfr: added the reason field. fwiw, this will mostly not be for IC/oncall to debug, but a result of a deployment of a service, which will be debugged by the tenants. but a reason makes sense nonetheless.

PR #1710, opened 2021-07-21
[APPSRE-3481] Change teams invitations endpoint on github integration

andreugallofre:
The Github_org integration is using an old endpoint to check the teams invitations that is deprecated by github 
https://docs.github.com/en/rest/reference/teams#list-pending-team-invitations-legacy

We found a problem with the endpoint and the github-mirror that seems that do not return information about the next pages of the request. 

This PR changes the endpoint to use the new endpoint 

https://docs.github.com/en/rest/reference/teams#list-pending-team-invitations

### DIFF
--- a/reconcile/github_org.py
+++ b/reconcile/github_org.py
@@ -120,7 +120,7 @@ def fetch_current_state(gh_api_store):
                 continue
 
             members = [member.login for member in team.get_members()]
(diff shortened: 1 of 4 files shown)


maorfr (approved):
LGTM

nice and clean!

PR #1711, opened 2021-07-22
add instructions to install python-devel and gcc-c++ packages

esron:
Adding instructions to install `python-devel` and `gcc-c++` packages for developers having trouble in the development environment setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,8 @@ python3 setup.py install
 python3 setup.py develop
 ```
 
+If the commands above don't work maybe you need to install the `python-devel` and `gcc-c++` packages.
+
 ### Requirements
 
 Please see [setup.py](setup.py).


README.md
```diff
@@ -136,6 +136,8 @@ python3 setup.py install
 python3 setup.py develop
 ```
 
+If the commands above don't work maybe you need to install the `python-devel` and `gcc-c++` packages.
+
```
rporres, line 140: Thsese instructions are too specific for your case. Other Linux distributions or Mac OS X users won't be able to use them. I would go for something more generic, stating the need of a compiler and python headers to run the last command successfully.

rporres (changes requested):

README.md
```diff
@@ -136,6 +136,8 @@ python3 setup.py install
 python3 setup.py develop
 ```
 
+If the commands above don't work maybe you need to install the `python-devel` and `gcc-c++` packages.
```
esron, line 139:
```suggestion
If the commands above don't work maybe you need a C++ compiler and python headers.
```

@rporres , Does that covers your suggestions?

rporres: indeed, thx!

rporres (approved):

PR #1712, opened 2021-07-22
Ability to add annotations to terraform openshift secrets

kylape:
Adds a terraform output field for annotations.  The value is a base64-encoded JSON document containing the annotations.

### DIFF
--- a/reconcile/utils/terraform_client.py
+++ b/reconcile/utils/terraform_client.py
@@ -266,9 +266,11 @@ class TerraformClient:
                 resource = data['{}_resource'.format(self.integration_prefix)]
                 output_resource_name = data['{}_output_resource_name'.format(
                     self.integration_prefix)]
+                annotations = data['{}_annotations'.format(
+                    self.integration_prefix)]
                 oc_resource = \
                     self.construct_oc_resource(output_resource_name, data,
-                                               account)
+                                               annotations, account)
                 ri.add_desired(
                     cluster,
                     namespace,
@@ -400,7 +402,7 @@ class TerraformClient:
             return data[list(data.keys())[0]]
         return data
 
-    def construct_oc_resource(self, name, data, account):
+    def construct_oc_resource(self, name, data, annotations, account):
         body = {
             "apiVersion": "v1",
             "kind": "Secret",
@@ -413,6 +415,11 @@ class TerraformClient:
             },
             "data": {}
         }
+
+        anno_dict = json.loads(base64.b64decode(annotations))
+
+        for k, v in anno_dict.items():
+            body["metadata"]["annotations"][k] = v
 
         for k, v in data.items():
             if self.integration_prefix in k:
--- a/reconcile/utils/terraform_client.py
+++ b/reconcile/utils/terraform_client.py
@@ -270,7 +270,7 @@ class TerraformClient:
                     self.integration_prefix)]
                 oc_resource = \
                     self.construct_oc_resource(output_resource_name, data,
-                                               annotations, account)
+                                               account, annotations)
                 ri.add_desired(
                     cluster,
                     namespace,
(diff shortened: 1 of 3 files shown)


reconcile/utils/terraform_client.py
```diff
@@ -400,7 +402,7 @@ def format_output(self, output, type):
             return data[list(data.keys())[0]]
         return data
 
-    def construct_oc_resource(self, name, data, account):
+    def construct_oc_resource(self, name, data, annotations, account):
```
Piojo, line 405: Why aren't we putting this annotations field at the end of the argument list? It's pretty easy to get confused and update callers the wrong way otherwise...

Also, in all the other updates you've chosen to return or pass annotations at the end of the relevant list.

reconcile/utils/terraform_client.py
```diff
@@ -414,6 +416,11 @@ def construct_oc_resource(self, name, data, account):
             "data": {}
         }
 
+        anno_dict = json.loads(base64.b64decode(annotations))
+
+        for k, v in anno_dict.items():
+            body["metadata"]["annotations"][k] = v
```
Piojo, line 422: Easier:
```suggestion
        body["metadata"]["annotations"][k] = anno_dict
```

reconcile/utils/terrascript_client.py
```diff
@@ -1192,13 +1193,13 @@ def generate_random_password(string_length=20):
                        for i in range(string_length))
 
     def populate_tf_resource_s3(self, resource, namespace_info):
-        account, identifier, common_values, \
-            output_prefix, output_resource_name = \
+        account, identifier, common_values, output_prefix, \
+            output_resource_name, annotations = \
```
Piojo, line 1197: OMG...

Can you please cut me a task to simplify all this duplicated code?

Piojo (commented):
I don't have enough experience with this integration to approve/reject. Please let's plan to refactor all those almost-identical methods and to stop returning 6-tuples and expecting that the next caller won't screw up.

reconcile/utils/terraform_client.py
maskarb: Perhaps this suggestion instead? Otherwise we would overwrite whatever existed in `annotations`.
```suggestion
        body["metadata"]["annotations"].update(anno_dict)
```

maorfr:
i understand what you want to achieve - adding annotations to Secrets resulting from terraformResources. that is great!
passing the values base64 encoded through the state (output) is also very clever.

is there an app-interface counterpart MR? i think it will help you understand something that you are missing (queries).

thanks for this contribution!

i'll review more thoroughly as well next week.

reconcile/utils/terraform_client.py
Piojo: Even better. :)

reconcile/utils/terraform_client.py
kylape: Happy to move it to the end.

kylape:
I don't have a corresponding app-interface MR, but I'll work on that today.  And thanks for pointing the missing query changes.

reconcile/utils/terrascript_client.py
kylape: https://issues.redhat.com/browse/APPSRE-3505

kylape:
app-interface MR: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/22555

maorfr (approved):
LGTM

thank you for this. I will test it locally a bit and pick it up from here.

nicely done!

maorfr:
i've added a small commit to be a little more defensive (and added a missing json.loads)

maorfr:
tested this with the app-interface dev repo: https://github.com/app-sre/app-interface/commit/9af84870851cb10fd64c0e9ef66398e1a37f34a8

PR #1713, opened 2021-07-23
Implement thread safe LDAP client

TGPSKI:
This PR updates the `ldap_users` integration to support thread safety. Currently, `ldap_client.py` is invoked in a multi-threaded fashion during reconciliation. However, the previous implementation of the ldap3 package did not include provisions for thread safety when interacting with an LDAP server. See the note in the ldap3 documentation [here](https://ldap3.readthedocs.io/en/latest/index.html).

The current usage of the ldap client is not particularly sophisticated - the integration simply checks with the ldap server to determine if a user record exists, but does nothing with the returned data and only operates on the boolean status variable. The thread safe connection returns a tuple instead of assigning found data to its object properties. Therefore, I ignore the additional return values and keep the integration (`ldap_users.py`) working as written.


maorfr (approved):
LGTM

Does this solve the socket errors we are seeing?

TGPSKI:
I have been trying to replicate the problem, but I haven't had luck. I am running docker-test-openldap (https://github.com/rroemhild/docker-test-openldap) on my local k3s cluster, and I've been trying to test multithreaded socket errors, but I haven't seen it hit yet.

Given we are multithreaded, that we are dealing with upcalls for socket access, and we aren't using thread safe patterns in our current implementation, this is my best guess. We will validate experimentally.

PR #1714, opened 2021-07-23
[openshift-saas-deploy] default to compare (true) resources

maorfr:
When the saasherder -> openshift-saas-deploy migration began, we wanted to give a similar experience in the logs, which means to see all resources that are being deployed. This meant we do not compare resources in `openshift-saas-deploy` by default, and `compare: true` can be defined in a saas file to force comparing. For context, other integration compare resources.

Over time, we have recommended that services move their ConfigMaps to be deployed through their saas files, and we kept supporting the `qontract.recycle: "true"` annotation (https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/sop/move-configmap-to-upstream.md).

Putting these two together, means that every ConfigMap which should cause a recycle, and is deployed through a saas file - is causing a recycle of the pods on EVERY deployment. This leads to weird situations, where if a `Deployment` is applied first, the recycle fails, and if a `ConfigMap` is deployed first, the recycle will be essentially cancelled. Or any other weird error in between. The simplest workaround to this problem is to add `compare: true` to such saas files.

This PR changes the default `compare` behavior to be "on", which solves the problem altogether. The only change in behavior is that the logs will show less resources being applied, which means other resources were not changed. In the current behavior, unchanged resources are always applied, but remain unchanged. So we are essentially changing only the behavior of the logs.

This fixes a big problem at the cost of an announcement.


reconcile/test/test_saasherder.py
```diff
@@ -277,3 +277,120 @@ def test_populate_desired_state_saas_file_delete(self):
...
+
+        saasherder = SaasHerder(
+            saas_files,
+            thread_pool_size=1,
+            gitlab=None,
+            integration='',
+            integration_version='',
+            settings={}
+        )
+        self.assertFalse(saasherder._get_saas_file_feature_enabled('attrib'))
```
Piojo, line 396: Tests!! :heart: :heart: :heart:

Piojo (approved):
I **think** I have spotted a bug, but I may be wrong. Can you please add a test for multiple SaaS files **and** a default value?

```python
def test_attribute_multiple_saas_files_default(self):
    saas_files = [
      { 'attrib': 'ute', ...},
     { 'attrib': 'ute', ...}
    ]
    s = SaasHerder(...)
    self.assertFalse(s._get_saas_file_feature_enabled('attrib', 42))
```
if I'm right, please fix it before pushing, if I'm wrong, push ahead. :)

PR #1716, opened 2021-07-23
fix ocp-release-mirror keeps mirroring to ocm-quay

BumbleFeng:
https://issues.redhat.com/browse/APPSRE-3526

This pr fix 2 problems in ocp-release-mirror which results in it keeps pushing images to ocm-quay.

1. urlparse('quay.io').netloc return empty

According to https://docs.python.org/3/library/urllib.parse.html

> Following the syntax specifications in RFC 1808, urlparse recognizes a net2. 
loc only if it is properly introduced by ‘//’. Otherwise the input is presumed to be a relative URL and thus to start with a path component.

2. _build_quay_auths() does not include username and password


jfchevrette (approved):
LGTM

PR #1717, opened 2021-07-24
Refactor terraform_vpc_peerings.build_desired_*state

Piojo:
This is the first step to prevent this integration from crashing when
some resource is not available. First I add tests, which has helped me
understand the code (which is complicated), next I can modify it and
refactor whatever's necessary.

It introduces a testing dependency on
[testslide](https://testslide.readthedocs.io).

Some of the mocked-out classes have lots of side effects in their
constructor, and mocking that out is a royal pain. Instead, I can rely
on `mock_constructor` to do the right thing, and I also gain far
better and stricter mocks.

### DIFF
--- a/reconcile/terraform_vpc_peerings.py
+++ b/reconcile/terraform_vpc_peerings.py
@@ -244,7 +244,7 @@ def build_desired_state_vpc_mesh(clusters, ocm_map, settings):
             account_vpcs = \
                 aws_api.get_vpcs_details(
                     account,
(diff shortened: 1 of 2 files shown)


maorfr (approved):

maorfr:
i wanted to merge it and i messed it up :|

Piojo:
There are a few conflicts since you also introduced some tests last week. I'll handle them, no biggie.

PR #1724, opened 2021-07-27
[kafka-clusters] disable deletion

maorfr:
we've observed cases where output Secrets were deleted. this PR denies that.
until we can further investigate - let's add this safety net.


BumbleFeng (approved):

maorfr:
https://issues.redhat.com/browse/APPSRE-3901

PR #1725, opened 2021-07-28
[openshift] ignore owned resources

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3356

With this PR, we can both act on owned resources (take over them) or ignore them (so they are not deleted)


jfchevrette:
Simple and elegant!

I would make the argument default to `True` so the default is the same as what we will be running in prod (ignore owned, the least dangerous & most common case) and allow the rarer case of not wanting to ignore owner references via an argument (`--include-owned`?)

maorfr:
@jfchevrette i've made this much simpler. care to take another look?

jfchevrette:
@maorfr Beautiful! We can add the argument back if we ever need to act upon routes with owner refs

maorfr:
With this, we can both act on them (take over them) and ignore them (so they are not deleted)!!

BumbleFeng (approved):

reconcile/utils/openshift_resource.py
```diff
@@ -226,6 +226,9 @@ def has_qontract_annotations(self):
 
         return True
 
+    def has_owner_reference(self):
+        return len(self.body['metadata'].get('ownerReferences', [])) > 0
```
Piojo, line 230: Simple nit:
```suggestion
        return bool(self.body['metadata'].get('ownerReferences', []))
```
which is closer to English' "the metadata has owner deferences".

Piojo (approved):
The nit is unimportant. Looks good, and thanks for the tests. :)

reconcile/utils/openshift_resource.py
maorfr: makes sense. i really was going towards the bool direction. thanks!

PR #1727, opened 2021-07-28
Support to cancel pending gitlab pipelines

esron:
Related MR:
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/22859

Jira Ticket:
https://issues.redhat.com/browse/APPSRE-3358

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -212,10 +212,10 @@ def merge_merge_requests(dry_run, gl, merge_limit, rebase, insist=False,
             if wait_for_pipeline:
                 # possible statuses:
                 # running, pending, success, failed, canceled, skipped
-                incomplete_pipelines = \
+                running_pipelines = \
                     [p for p in pipelines
                      if p['status'] in ['running']]
-                if incomplete_pipelines:
+                if running_pipelines:
                     if insist:
                         raise Exception(f'insisting on {merge_label}')
                     else:
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -250,7 +250,9 @@ def run(dry_run, wait_for_pipeline):
         days_interval = hk.get('days_interval') or default_days_interval
         enable_closing = hk.get('enable_closing') or default_enable_closing
         limit = hk.get('limit') or default_limit
+        pipeline_timeout = hk.get('pipeline_timeout')
         gl = GitLabApi(instance, project_url=project_url, settings=settings)
+
         handle_stale_items(dry_run, gl, days_interval, enable_closing,
                            'issue')
         handle_stale_items(dry_run, gl, days_interval, enable_closing,
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -225,6 +225,7 @@ def merge_merge_requests(dry_run, gl, merge_limit, rebase,
                                        settings=gl_settings,).project
                 clean_pipelines(dry_run, gl_project, pipelines,
                                 pipeline_timeout)
+
             if wait_for_pipeline:
                 # possible statuses:
                 # running, pending, success, failed, canceled, skipped
@@ -277,11 +278,14 @@ def run(dry_run, wait_for_pipeline):
         try:
             merge_merge_requests(dry_run, gl, limit, rebase, pipeline_timeout,
                                  insist=True,
-                                 wait_for_pipeline=wait_for_pipeline)
+                                 wait_for_pipeline=wait_for_pipeline,
+                                 gl_instance=instance, gl_settings=settings)
         except Exception:
             merge_merge_requests(dry_run, gl, limit, rebase, pipeline_timeout,
-                                 wait_for_pipeline=wait_for_pipeline)
+                                 wait_for_pipeline=wait_for_pipeline,
+                                 gl_instance=instance, gl_settings=settings)
         if rebase:
             rebase_merge_requests(dry_run, gl, limit,
                                   pipeline_timeout=pipeline_timeout,
-                                  wait_for_pipeline=wait_for_pipeline)
+                                  wait_for_pipeline=wait_for_pipeline,
+                                  gl_instance=instance, gl_settings=settings)
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -179,7 +179,8 @@ def rebase_merge_requests(dry_run, gl, rebase_limit, pipeline_timeout=None,
 @retry(max_attempts=10)
 def merge_merge_requests(dry_run, gl, merge_limit, rebase,
                          pipeline_timeout=None, insist=False,
-                         wait_for_pipeline=False):
+                         wait_for_pipeline=False, gl_instance=None,
+                         gl_settings=None):
     mrs = gl.get_merge_requests(state='opened')
     merges = 0
     for merge_label in MERGE_LABELS_PRIORITY:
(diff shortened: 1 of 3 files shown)


BumbleFeng:
Hi, I know there are a lot of pipelines in https://gitlab.cee.redhat.com/service/app-interface/-/pipelines?page=1&scope=all&status=pending. However, these pipelines will not actually block an MR from merging, we could safely ignore them. 

Meanwhile, the pipelines in a certain MR e.g https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/22859/pipelines once you click the pipeline link and it will redirect you to https://gitlab.cee.redhat.com/essilva/app-interface/-/pipelines/1214675. Notice the project has changed to your own fork repo. These pipelines wouldn't be gathered by your function as they are in a different repo and these pipelines are what we care about. 

So you will still need to check the pipelines from https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/gitlab_housekeeping.py#L123 and cancel them if timeout.

BumbleFeng:
And a hit, we usually rebase master instead of merge mater, so we can have a clean commit history.

reconcile/gitlab_housekeeping.py
```diff
@@ -19,6 +19,26 @@
 
 QONTRACT_INTEGRATION = 'gitlab-housekeeping'
 
+def clean_pipelines(dry_run, gl, pipeline_timeout=60):
+    DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
+    now = datetime.utcnow()
+
+    pipelines = gl.get_pipelines(status='pending')
+
+    for pipeline in pipelines:
```
Piojo, line 28: You'll see me complain about loop variables that differ from the iterated object by just one character. Try making sense of them on long loops after 6 hours of work and interruptions...
```suggestion
    for p in pipelines:
```

Piojo (changes requested):
Nothing major, but tests are failing - and I'd appreciate a couple of unit tests for the new `clean_pipelines` method. Thanks!

reconcile/gitlab_housekeeping.py
esron: In my understanding variables with one letter names are bad for readability, but if this is a project standard I can change that.

Piojo: Well, that depends on the scope. The adagio is "use short names for locals, long for globals". You don't want a single-letter class name or method name. But for a very short-lived loop variable, whose "main" object of interest is 3 characters away, the shorter the better. Also, if this loop ever grows, you'll *really* mistake `pipeline` and `pipeline` (and I've seen the one different character placed in the middle of the word, try to read that).

Longer discussion: https://gitlab.cee.redhat.com/service/dev-guidelines/-/merge_requests/63 .

reconcile/gitlab_housekeeping.py
```diff
@@ -19,6 +19,26 @@
...
+def clean_pipelines(dry_run, gl, pipeline_timeout=60):
+    DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
+    now = datetime.utcnow()
+
+    pipelines = gl.get_pipelines(status='pending')
+
+    for p in pipelines:
+        last_update_date = datetime.strptime(p.updated_at, DATE_FORMAT)
+
+        pipeline_elapsed_time = now.date() - last_update_date.date()
```
Piojo, line 31: Don't change it, but for the future, consider whether for this short-lived object `elapsed` wouldn't be simpler to understand.

reconcile/gitlab_housekeeping.py
```diff
@@ -19,6 +19,26 @@
 
 QONTRACT_INTEGRATION = 'gitlab-housekeeping'
 
+def clean_pipelines(dry_run, gl, pipeline_timeout=60):
+    DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
+    now = datetime.utcnow()
```
Piojo, line 24: Mmmm... Don't mess with date parsing, it's confusing. Prefer time for this:

```python
now = time.time()
```

reconcile/gitlab_housekeeping.py
```diff
@@ -19,6 +19,26 @@
...
+
+    pipelines = gl.get_pipelines(status='pending')
+
+    for p in pipelines:
+        last_update_date = datetime.strptime(p.updated_at, DATE_FORMAT)
```
Piojo, line 29: What is the data type of this `p`? I'd suggest converting it to seconds since epoch and just do normal integer operations. The `datetime` stuff is great if you have to print stuff to humans, or do calculations of the form "the last Monday of next month, two years from now". But for the duration of a computer operation, stick to `time.time`.

reconcile/gitlab_housekeeping.py
Piojo: But do try to change it to seconds and integer operations. It'll be less code and it will also be faster.

reconcile/gitlab_housekeeping.py
```diff
@@ -19,6 +19,26 @@
...
+
+        pipeline_elapsed_time = now.date() - last_update_date.date()
+
+        if pipeline_elapsed_time > timedelta(minutes=pipeline_timeout):
```
Piojo, line 33: Because if you use seconds since epoch you don't need the `timedelta`.

reconcile/gitlab_housekeeping.py
```diff
@@ -19,6 +19,26 @@
 
 QONTRACT_INTEGRATION = 'gitlab-housekeeping'
 
+def clean_pipelines(dry_run, gl, pipeline_timeout=60):
+    DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
```
Piojo, line 23: This is a constant you share with many other methods. If you really need it, make it global.

esron:
@Piojo I've changed things a little bit but your input is still valid, thank you. I will work on it next Monday :)

reconcile/gitlab_housekeeping.py
```diff
@@ -18,10 +20,38 @@
                'do-not-merge/hold', 'do-not-merge/pending-review']
 
 QONTRACT_INTEGRATION = 'gitlab-housekeeping'
+DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
+
+
+def clean_pipelines(dry_run, project_id, pipelines, pipeline_timeout=60):
+    instance = queries.get_gitlab_instance()
+    settings = queries.get_app_interface_settings()
```
BumbleFeng, line 28: instance and settings should get from arguments instead of query every time.

reconcile/gitlab_housekeeping.py
```diff
@@ -18,10 +20,38 @@
...
+    now = time.time()
+
+    pending_pipelines = [p for p in pipelines if p['status'] == 'pending']
```
BumbleFeng, line 31: We should also cancel running pipelines

reconcile/gitlab_housekeeping.py
```diff
@@ -18,10 +20,35 @@
...
+
+        elapsed = now - time.mktime(update_time)
+
+        # pipeline_timeout converted in seconds
+        if elapsed > pipeline_timeout * 60:
+            logging.info(['canceling', p['web_url']])
+            if not dry_run:
+                try:
+                    gl_project.pipelines.get(p['id']).cancel()
+                except gitlab.exceptions.GitlabPipelineCancelError as err:
```
Piojo, line 46: I'm not sure I'd catch the exception here. This hides any failures from the caller, and prevents them from taking any corrective action. Why not let it pass?

Piojo (commented):
Minor nit, I can be convinced it doesn't matter.

PR #1728, opened 2021-07-29
[openshift] reconcile away deleted annotations

maorfr:
our compare logic mostly checks if all desired keys exist and ignores existing keys (which usually come from k8s defaults).
this PR adds a special behavior for annotations - to compare them and delete redundant keys.

### DIFF
--- a/reconcile/utils/openshift_resource.py
+++ b/reconcile/utils/openshift_resource.py
@@ -59,7 +59,7 @@ class OpenshiftResource:
                     pass
                 elif self.ignorable_key_value_pair(obj1_k, obj1_v):
                     pass
-                elif obj1_k in ['data', 'labels']:
+                elif obj1_k in ['data', 'labels', 'annotations']:
                     diff = [k for k in obj2_v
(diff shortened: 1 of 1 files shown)


reconcile/utils/openshift_resource.py
```diff
@@ -59,7 +59,7 @@ def obj_intersect_equal(self, obj1, obj2):
                     pass
                 elif self.ignorable_key_value_pair(obj1_k, obj1_v):
                     pass
-                elif obj1_k in ['data', 'labels']:
+                elif obj1_k in ['data', 'labels', 'annotations']:
```
Piojo, line 62: You know how much I like sets, this looks to me like (warning: untested!):
```python
diff = set(keys(obj2_v)) - (set(keys(obj1_v) - set(IGNORABLE_DATA_FIELDS))
```

... but I guess that requires more changes than this one-liner.

Piojo (approved):

BumbleFeng:
let me test it a little bit before merging it.

PR #1729, opened 2021-07-29
[terraform-vpc-peerings] set assume_region and assume_cidr only once

maorfr:
a minor refactor. also adds tests.

### DIFF
--- a/reconcile/test/test_terraform_vpc_peerings.py
+++ b/reconcile/test/test_terraform_vpc_peerings.py
@@ -9,6 +9,7 @@ class MockOCM:
                                                             tf_user):
         return f"{cluster}/{tf_account_id}/{tf_user}"
 
+
 class TestAWSAccountFromInfrastructureAccess(TestCase):
     def test_aws_account_from_infrastructure_access(self):
         cluster = {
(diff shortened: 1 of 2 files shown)


reconcile/test/test_terraform_vpc_peerings.py
```diff
@@ -0,0 +1,81 @@
...
+            'assume_role': 'cluster/uid/terraform',
+            'assume_region': 'region',
+            'assume_cidr': 'vpc'
+        }
+        account = integ.aws_account_from_infrastructure_access(
+            cluster, 'read-only', ocm_map)
+        self.assertEqual(account, expected_result)
+
+    def test_aws_account_from_infrastructure_access_none(self):
+        cluster = {
```
Piojo, line 54: If you don't want to repeat yourself that much, you can move this to the `setUp` method and only do small modifications in here.

Piojo (approved):
DRY the code!!

reconcile/test/test_terraform_vpc_peerings.py
maorfr: nice. i'll do that.

PR #1736, opened 2021-07-29
Add a tox config friendlier to development

Piojo:
To speed up development, it only runs pytest, and we can pass
positional arguments to narrow down the tests we want executed, like:

```
tox -c tox-devel.ini -- reconcile/test/test_utils_gpg.py
tox -c tox-devel.ini -- reconcile/test/test_utils_gpg.py::TestGpgKeyValid
tox -c tox-devel.ini
```

### DIFF
--- a/tox-devel.ini
+++ b/tox-devel.ini
@@ -1,0 +1,13 @@
+[tox]
+envlist = py36
+
+[testenv]
+commands =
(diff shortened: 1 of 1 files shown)


esron:
lgtm

esron (approved):

PR #1738, opened 2021-07-30
ldap-users: Use thread safe client strategy, use manual binds

TGPSKI:
This PR updates the `ldap_users` integration to support thread safety. Currently, `ldap_client.py` is invoked in a multi-threaded fashion during reconciliation. However, the previous implementation of the ldap3 package did not include provisions for thread safety when interacting with an LDAP server. See the note in the ldap3 documentation [here](https://ldap3.readthedocs.io/en/latest/index.html).

The current usage of the ldap client is not particularly sophisticated - the integration simply checks with the ldap server to determine if a user record exists, but does nothing with the returned data and only operates on the boolean status variable. The thread safe connection returns a tuple instead of assigning found data to its object properties. Therefore, I ignore the additional return values and keep the integration (`ldap_users.py`) working as written.

Changes:

* ldap client is moved from global scope to function scope
* connection uses `client_strategy=SAFE_SYNC`
* connection is manually bound and is now returned from init, instead of assigning to global _client
* during usage of ldap connection, client bind status is checked before proceeding, if not bound, `client.bind()` is executed
* after each search operation, the client is unbound

This integration ran on stage for 3-4 hours on Friday with no socket close errors reported.


reconcile/utils/ldap_client.py
```diff
@@ -22,16 +20,20 @@ def init_from_config():
...
+    if not client.bound:
+        client.bind()
 
     search_filter = "uid={},{}".format(username, _base_dn)
+    result, _, _, _ = client.search(search_filter, '(objectclass=person)')
+
+    client.unbind()
 
-    return _client.search(search_filter, '(objectclass=person)')
+    return result
```
jfchevrette, line 39: previously we returned the result of client.search() now it looks like we're unpacking what it returns and only returning one of it's elements (result). that makes me wonder if the consumers of this function will still function?

or is it that the search() method return type changed in the new version?

can we add a test?

TGPSKI: The change in the output is a result of using a different client_strategy. Searches run with a connection parameter `client_strategy=SYNC` will return a single boolean value - did the search find a result or not? The actual found data is then accessible as additional parameters bound to the client object.

Moving to `client_strategy=SAFE_SYNC`, a tuple is returned instead of a boolean. The values that were written to the client object are instead returned to the caller.

The usage in this integration does not care about any of the data associated with a particular search. The only thing the integration cares about is if a user is present in ldap or not. If the user is not present, then the integration triggers a gitlab remove user method.

TGPSKI: The current usage of the ldap client is not particularly sophisticated - the integration simply checks with the ldap server to determine if a user record exists, but does nothing with the returned data and only operates on the boolean status variable. The thread safe connection returns a tuple instead of assigning found data to its object properties. Therefore, I ignore the additional return values and keep the integration (ldap_users.py) working as written.

TGPSKI: Re: testing, the only test I can think of would be to mock an ldap server and verify that we get correct booleans out of a user search for present / not present. This would duplicate tests in ldap3 package, no? Is there a different angle you can think of?

TGPSKI:
@maorfr adding you as I think JFC is OOO.

New changes:

* added the context manager approach

This pod is running on stage @ `ldap-users-6cd84b9898-jd9p8`

reconcile/utils/ldap_client.py
```diff
@@ -22,16 +24,13 @@ def init_from_config():
...
-    global _client
     global _base_dn
 
-    init_from_config()
-
-    search_filter = "uid={},{}".format(username, _base_dn)
-
-    return _client.search(search_filter, '(objectclass=person)')
+    with init_from_config() as client:
+        search_filter = "uid={},{}".format(username, _base_dn)
```
Piojo, line 34: This will totally work, but it'd be better if you left the base DN here and...

reconcile/utils/ldap_client.py
```diff
@@ -22,16 +24,13 @@ def init_from_config():
...
+        result, _, _, _ = client.search(search_filter, '(objectclass=person)')
```
Piojo, line 35:
```suggestion
        result, _, _, _ = client.search(base_dn, f'(&(uid={username})(objectclass=person))')
```
because that way, you can move towards having far more `uid` s in a single filter, and reduce the amount of round-trips you do to LDAP in any integration.

maorfr:
LGTM from me for anything that works, we've put enough effort into this already. let's get +1s from others who have been involved here and get this out of our way :)

Piojo (approved):

PR #1745, opened 2021-08-04
Implement tokens for CSO and SLO DashDotDB types

rrati:
This is dependent upon 
https://github.com/app-sre/dashdotdb/pull/46

### DIFF
--- a/reconcile/dashdotdb_base.py
+++ b/reconcile/dashdotdb_base.py
@@ -15,62 +15,64 @@ DASHDOTDB_SECRET = os.environ.get('DASHDOTDB_SECRET',
 
 
 class DashdotdbBase:
-    def __init__(self, dry_run, thread_pool_size, marker):
+    def __init__(self, dry_run, thread_pool_size, marker, scope):
         self.dry_run = dry_run
         self.thread_pool_size = thread_pool_size
         self.settings = queries.get_app_interface_settings()
         self.secret_reader = SecretReader(settings=self.settings)
         self.secret_content = self.secret_reader.read_all(
             {'path': DASHDOTDB_SECRET})
-        self.dashdotdb_url = self.secret_content['url']
-        self.dashdotdb_user = self.secret_content['username']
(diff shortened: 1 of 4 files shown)


reconcile/dashdotdb_base.py
```diff
@@ -0,0 +1,101 @@
...
+        self.dry_run = dry_run
+        self.thread_pool_size = thread_pool_size
+        self.settings = queries.get_app_interface_settings()
+        self.secret_reader = SecretReader(settings=self.settings)
+        self.secret_content = self.secret_reader.read_all(
+            {'path': DASHDOTDB_SECRET})
+        self.dashdotdb_url = self.secret_content['url']
+        self.dashdotdb_user = self.secret_content['username']
+        self.dashdotdb_pass = self.secret_content['password']
+        self.logmarker = marker
```
BumbleFeng, line 28: I think we could init and verify the scope and set to self.scope here

BumbleFeng (approved):

reconcile/dashdotdb_base.py
rrati: Yes, good idea.  Done

reconcile/dashdotdb_base.py
```diff
@@ -15,62 +15,64 @@
...
         self.dry_run = dry_run
         self.thread_pool_size = thread_pool_size
         self.settings = queries.get_app_interface_settings()
         self.secret_reader = SecretReader(settings=self.settings)
         self.secret_content = self.secret_reader.read_all(
             {'path': DASHDOTDB_SECRET})
-        self.dashdotdb_url = self.secret_content['url']
-        self.dashdotdb_user = self.secret_content['username']
-        self.dashdotdb_pass = self.secret_content['password']
+        self.dashdotdb_url = 'http://localhost:8080'
```
BumbleFeng, line 25: I think you forget to revert the change of using localhost dddb

PR #1747, opened 2021-08-05
APPSRE-3559: Recorded SLO values in Dash.DB don't have enough precision

jmelis:
Currently the data in Dash.DB doesn't have enough precision:

```
serviceslometrics{cluster="...",name="Tenant Manager API availability",namespace="...",service="...",slitype="availability",type="slo_target"} 95.0
```

This is because we are doing an `int()` on the SLO values. This PR changes that to a float.

### DIFF
--- a/reconcile/dashdotdb_slo.py
+++ b/reconcile/dashdotdb_slo.py
@@ -109,15 +109,19 @@ class DashdotdbSLO:
                 prom_result = prom_response['data']['result']
                 if not prom_result:
                     continue
+
                 slo_value = prom_result[0]['value']
                 if not slo_value:
                     continue
+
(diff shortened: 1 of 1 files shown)


jmelis:
Note for the reviewer, in the previous code, the datatypes were different:

```
                result.append({
                    "name": slo['name'],
                    "SLIType": slo['SLIType'],
                    "namespace": ns,
                    "cluster": ns['cluster'],
                    "service": ns['app'],
                    "value": slo_value,
                    "target": slo_target,
                })
```

- `slo_value` was an `int`, now it's a `float`.
- `slo_target` was a `string`, now it's a `float`.

BumbleFeng (approved):

PR #1748, opened 2021-08-05
[ldap-users] remove threading

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3547

instead of searching user by users, we get all users and find missing ones in memory.

### DIFF
--- a/reconcile/utils/ldap_client.py
+++ b/reconcile/utils/ldap_client.py
@@ -31,8 +31,7 @@ def users_exist(users):
     global _base_dn
 
     with init_from_config() as client:
-        _, _, results, _ = client.search(_base_dn,
-                                        '(&(objectclass=person))')
+        _, _, results, _ = client.search(_base_dn, '(&(objectclass=person))')
         existing_users = set()
         for r in results:
             dn = r['dn']
--- a/reconcile/utils/ldap_client.py
+++ b/reconcile/utils/ldap_client.py
@@ -27,11 +27,15 @@ def init_from_config():
     return init(serverUrl)
 
 
-def get_users():
+def get_users(uids):
     global _base_dn
 
     with init_from_config() as client:
-        _, _, results, _ = client.search(_base_dn, '(&(objectclass=person))',
-                                         attributes=['uid'])
+        user_filter = "".join((f"(uid={u})" for u in uids))
+        _, _, results, _ = client.search(
+            _base_dn,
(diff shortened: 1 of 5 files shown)


reconcile/utils/ldap_client.py
```diff
@@ -27,11 +27,15 @@ def init_from_config():
...
 
     with init_from_config() as client:
-        result, _, _, _ = client.search(_base_dn,
-                                        f'(&(uid={username})' +
-                                        '(objectclass=person))')
-        return result
+        _, _, results, _ = client.search(_base_dn, '(&(objectclass=person))')
+        users = set()
+        for r in results:
+            dn = r['dn']
```
Piojo, line 37: Just extract the uid field here.
```suggestion
            uid = r['uid']
```

Piojo (changes requested):
Please note that, while all my comments are correct from an LDAP perspective, I am not familiar with the Python modules - the return structure of `search` may need some adjustments. API docs: https://ldap3.readthedocs.io/en/latest/searches.html

Piojo:
I'd be curious to know the new timings. :)

reconcile/utils/ldap_client.py
```diff
@@ -27,11 +27,15 @@ def init_from_config():
...
-def user_exists(username):
+def get_users():
     global _base_dn
 
     with init_from_config() as client:
-        result, _, _, _ = client.search(_base_dn,
-                                        f'(&(uid={username})' +
-                                        '(objectclass=person))')
-        return result
+        _, _, results, _ = client.search(_base_dn, '(&(objectclass=person))')
```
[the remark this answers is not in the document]
maorfr: added `attributes=["uid"]`, thanks! there is no significant difference passing uids or not (4-5 seconds for both ways).

reconcile/utils/ldap_client.py
maorfr: thanks!

maorfr:
time improvement: 2m39s -> 8s (!!!)

Piojo (approved):
LGTM. A future optimization is to query only for the 700-or-so users known to app-interface instead of the ~10K IDs we may get from querying the entire LDAP subtree. But for now, this is awesome and I bet the LDAP owners will be pleased as well.

maorfr:
changed the query to only be for the known users, as you suggested. run time is still ~8s, but it's obviously better this way.

PR #1756, opened 2021-08-11
[qontract-cli] add set slack-usergroup command

maorfr:
with this command, we can add users to a usergroup:
```
qontract-cli --config config.debug.toml set slack-usergroup <workspace> <usergroup> <username>
```
adding someone to a usergroup is easier through the slack UI, but this command is mostly useful to empty a slack usergroup:
```
qontract-cli --config config.debug.toml set slack-usergroup <workspace> <usergroup> ''
```

this will help us with adding/removing ourselves from usergroups, such as the upcoming @app-sre-onboarding-ic.

### DIFF
--- a/reconcile/utils/slack_api.py
+++ b/reconcile/utils/slack_api.py
@@ -8,6 +8,10 @@ from reconcile.utils.secret_reader import SecretReader
 from reconcile.utils.config import get_config
 
 
+class UserNotFoundException(Exception):
+    pass
+
+
 class UsergroupNotFoundException(Exception):
     pass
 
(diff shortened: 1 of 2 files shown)


reconcile/utils/slack_api.py
```diff
@@ -98,6 +99,17 @@ def get_random_deleted_user(self):
...
 
+    def get_user_id_by_name(self, user_name):
+        config = get_config()
+        mail_address = config['smtp']['mail_address']
+        result = self.sc.api_call(
+            "users.lookupByEmail",
+            email=f"{user_name}@{mail_address}"
+        )
+        if not result['ok']:
+            raise Exception(result['error'])
```
Piojo, line 110: Please, report at least the user name that is having the error, ideally also their email address. Also, I'd appreciate it if we raised either `ValueError` or a custom `SlackApiError`

maorfr: i've added a custom exception. about reporting the user name, this is raised when someone runs the command with a specific user, so i'd expect it is understood what is the problematic user?

PR #1764, opened 2021-08-16
APPSRE-3563 move dashdotdb components to the internal cluster

patjlm:
Moving those integrations on the private cluster:
* dashdotdb-cso
* dashdotdb-dvo
* dashdotdb-slo

Newbie comments:
- since this is managed by helm (apparently), I suppose statefulsets in the external cluster will be automatically removed by Helm
- no services exposed by integrations (except local metrics endpoint) so no routing to update elsewhere
- Keeping all integration settings as is. I don't see why they would be different on the internal cluster


maorfr:
this test should have failed: https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/test_make.py

patjlm:
`test_make.py` fixed and manifests re-synchronized as part of PR #1765. Thanks @Piojo !

maorfr (approved):
LGTM

just get the tests passing and you are clear to merge!

PR #1767, opened 2021-08-17
fix openshift_saas_deploy_trigger_cleaner PipelineRun sort order

erdii:
This integration cleans up old PipelineRuns but keeps the newest `minimum` amount of runs
and additionally all runs that are younger than `days` days.

There was a bug sorting the PipelineRuns in the wrong order (from oldest to newest) which then resulted in the `minimum`cut-off producing the `minimum` OLDEST PipelineRuns to be kept.

There was another bug in the `within_retention_days` function which compared date differences and not hours passed (the actual timedelta), which made python say that everything from yesterday was older than 1 day even though it was just, say, 5 hours old.

Co-authored-by: Mayank Shah <mashah@redhat.com>
Co-authored-by: Nico Schieder <nschiede@redhat.com>
Signed-off-by: Josh Gwosdz <jgwosdz@redhat.com>

### DIFF
--- a/reconcile/openshift_saas_deploy_trigger_cleaner.py
+++ b/reconcile/openshift_saas_deploy_trigger_cleaner.py
@@ -1,6 +1,6 @@
 import logging
 
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from dateutil import parser
 
 import reconcile.queries as queries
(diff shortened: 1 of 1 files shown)


maorfr (approved):
LGTM

thanks a lot!

PR #1768, opened 2021-08-17
[openshift] redact error when applying secrets

maorfr:
with this PR, when there is an error applying a Secret (when there is a possibility the error contains the Secret itself), we will not print the error.

this will allow us to add slack output for other integrations, such as openshift-secrets.


jfchevrette:
The code change LGTM, but...

I would leave out enabling slack logging for other integrations for a separate PR... Could there be other sources of logging in those integrations other than applying openshift secrets that may result in potentially leaking other kinds of sensitive information to slack?

Piojo (approved):
Approving to unblock, but @jfchevrette's suggestions make all the sense in the world. We should review other sources of potential bad logging in a separate PR.

PR #1770, opened 2021-08-18
Create Quay teams that don't already exist

steveteahan:
This change creates a Quay team when it doesn't exist and a member needs to be added to the team.  A default role of `member` is assigned to the team as a sane default.  Prior to this change an exception would be thrown and the team would need to be created manually.

### DIFF
--- a/reconcile/quay_membership.py
+++ b/reconcile/quay_membership.py
@@ -70,8 +70,9 @@ def fetch_current_state(quay_api_store):
             try:
                 members = quay_api.list_team_members(team)
             except QuayTeamNotFoundException:
-                logging.warning("Attempted to list members for team %s/%s, "
-                                "but it doesn't exist", org_key.org_name, team)
+                logging.warning("Attempted to list members for team %s in "
+                                "org %s/%s, but it doesn't exist", team,
+                                org_key.instance, org_key.org_name)
             else:
                 # Teams are only added to the state if they exist so that
                 # there is a proper diff between the desired and current state.
--- a/reconcile/quay_membership.py
+++ b/reconcile/quay_membership.py
@@ -8,7 +8,7 @@ from reconcile.utils.aggregated_list import (AggregatedList,
                                              RunnerException)
 from reconcile.quay_base import get_quay_api_store
 from reconcile.status import ExitCodes
-
+from reconcile.utils.quay_api import QuayTeamNotFoundException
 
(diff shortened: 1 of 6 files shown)


steveteahan:
Additional app-sre specific testing is documented in: https://issues.redhat.com/browse/APPSRE-3488

reconcile/quay_membership.py
```diff
@@ -67,12 +67,19 @@ def fetch_current_state(quay_api_store):
...
-            state.add({
-                'service': 'quay-membership',
-                'org': org_key,
-                'team': team
-            }, members)
+            try:
+                members = quay_api.list_team_members(team)
+            except QuayTeamNotFoundException:
+                logging.warning("Attempted to list members for team %s/%s, "
+                                "but it doesn't exist", org_key.org_name, team)
```
BumbleFeng, line 74: Could you include org_key.instance in the log?

reconcile/utils/quay_api.py
```diff
@@ -90,6 +101,33 @@ def add_user_to_team(self, user, team):
             raise RequestsException(r)
         return True
 
+    def create_or_update_team(self, team: str, role="member",
+                              description=None) -> None:
```
BumbleFeng, line 105: Could you keep the same format with other methods and remove the team type and return type? I just want or keep consistency.

BumbleFeng (approved):
Generally LGTM, good job

reconcile/utils/quay_api.py
steveteahan: You'd like me to remove the docstring and type hints?  My reasons for adding those were:

- The docstring provides a link to the Quay API docs which might be useful for quick references?  It also documents which exceptions can be raised.
- Gradually introducing the type hints can assist IDEs in auto-completion and prepare the code base for the opportunity to use tools like mypy, pyre, etc. in the future.

I understand this affects the consistency of the code, but it opens up the opportunity to improve the documentation of the other methods next time a change needs to be made to them.

reconcile/quay_membership.py
steveteahan: Good call, I'll fix that!

steveteahan:
Thank you for the review @BumbleFeng!

reconcile/utils/quay_api.py
BumbleFeng: works for me

PR #1772, opened 2021-08-19
APPSRE-3564: Get promotion history for tekton pipeline

esron:
Jira ticket: https://issues.redhat.com/browse/APPSRE-3564

Related MR on App Interface: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/24187

### DIFF
--- a/tools/app_interface_reporter.py
+++ b/tools/app_interface_reporter.py
@@ -329,6 +329,9 @@ def get_apps_data(date, month_delta=1, thread_pool_size=10):
     )
     query = 'app_sre_tekton_pipelinerun_task_status'
 
+    metrics = promql(settings['pushGatewayCluster']['prometheusUrl'],
+                     query, token)
+
     jenkins_job_history = get_build_history_pool(
                             jenkins_map,
                             jobs_to_get,
--- a/tools/app_interface_reporter.py
+++ b/tools/app_interface_reporter.py
@@ -3,6 +3,7 @@ import os
 import textwrap
 
 from datetime import datetime, timezone
+from urllib.parse import urljoin
 
 import click
 import requests
@@ -36,7 +37,7 @@ DASHDOTDB_SECRET = os.environ.get('DASHDOTDB_SECRET',
                                   'app-sre/dashdot/auth-proxy-production')
 
 
-def promql(url, query, auth=None):
+def promql(url, query, token):
     """
     Run an instant-query on the prometheus instance.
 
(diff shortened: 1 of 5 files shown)


tools/app_interface_reporter.py
```diff
@@ -34,9 +35,13 @@
 CONTENT_FORMAT_VERSION = '1.0.0'
 DASHDOTDB_SECRET = os.environ.get('DASHDOTDB_SECRET',
                                   'app-sre/dashdot/auth-proxy-production')
+APP_SRE_PROD_01_PROMETHEUS_URL = \
```
jfchevrette, line 38: You can get this URL from a graphql query instead of hardcoding it here

tools/app_interface_reporter.py
```diff
@@ -270,7 +325,19 @@ def get_apps_data(date, month_delta=1, thread_pool_size=10):
                 else:
                     jobs_to_get[job['instance']].append(job)
 
-    job_history = get_build_history_pool(
+    # This is temporary, it can go into a secret or can come from
+    # queries.get_pipelines_providers()
+    token = secret_reader.read({
```
jfchevrette, line 330: This will be available in the cluster schema via the `automationToken` field once you find out what the tekton cluster is through graphql queries as suggested previously

tools/app_interface_reporter.py
```diff
@@ -34,9 +35,13 @@
 CONTENT_FORMAT_VERSION = '1.0.0'
 DASHDOTDB_SECRET = os.environ.get('DASHDOTDB_SECRET',
                                   'app-sre/dashdot/auth-proxy-production')
+APP_SRE_PROD_01_PROMETHEUS_URL = \
+    os.environ.get('APP_SRE_PROD_01_PROMETHEUS_URL',
+                   'https://prometheus.app-sre-prod-01.devshift.net')
+TEKTON_JOBS_CLUSTER = os.environ.get('TEKTON_JOBS_CLUSTER', 'app-sre-prod-01')
```
jfchevrette, line 41: Likewise here I believe you can query pipelines-providers with `provider: tekton` to get this and then get to the cluster schema which will have a `prometheusUrl` field

jfchevrette (changes requested):

tools/app_interface_reporter.py
esron: Yeah, I know, I was suggested by @BumbleFeng to hardcode this for now just to simplify the process. As we know that tekton jobs only run on app-sre-prod-01.

Piojo (commented):
I can haz unit tests? :))

maorfr:
i +1 previous comments about not using app-sre-prod-01 in any way. this is just the cluster with the prometheus instance that has the push gateway metrics. i'd consider passing it through app-interface-settings and querying for it.

esron:
> I can haz unit tests? :))

This can be a little challenge. There are to many side effects to the command. But maybe I should export my code to a method to make things easier?

tools/app_interface_reporter.py
```diff
@@ -222,9 +231,55 @@ def get_activity_content(activity):
...
 
 
+def prometheus_metrics_to_job_history(metrics):
+    """
+    Returns an object of objects containing the results of the tekton pipelines
+    grouped by environments and saas_file names.
+    """
+    job_history = {}
+
+    def inverter(x):
```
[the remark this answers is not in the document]
esron: Agreed.

esron:
> i +1 previous comments about not using app-sre-prod-01 in any way. this is just the cluster with the prometheus instance that has the push gateway metrics. i'd consider passing it through app-interface-settings and querying for it.
> 
> side note: how do you override environment variables in an integration? cc @BumbleFeng

@maorfr , how would that work? I was thinking about creating a new secret on vault.

maorfr:
everything should be query-able from app-interface. what you need is to know the cluster and that's it. maybe something like:
in app-interface-settings:
```
pushGatewayCluster:
  $ref: /openshift/app-sre-prod-01/cluster.yml
```

Piojo:
Splitting code may definitely make tests simpler. Otherwise, we'll have to mock stuff, and I know it can get annoying.

PR #1773, opened 2021-08-19
de-duplication cluster when init oc_map from namespaces

BumbleFeng:
init_oc_client is running in different threads. If there are multi namespaces from the same cluster, we actually will init oc_client for that cluster multi times.


rrati (approved):
lgtm

PR #1775, opened 2021-08-20
Set higher result limit for conversations.list Slack API call

steveteahan:
conversations.list is a Tier 2 API call which allows 20+ calls per
minute. By default, the result limit was 100 records, which could
easily exceed the API limit when there are many channels. This change
increases the limit to 1000, but also provides the option to override.


maorfr:
from reading the method's pagination section: https://api.slack.com/methods/conversations.list
it seems that using 1000 will not get us the pagination we need.
from reading the rate limit's pagination section: https://api.slack.com/docs/rate-limits
it seems that the rate limit will be more strict?

reconcile/utils/slack_api.py
```diff
@@ -1,12 +1,18 @@
...
+from typing import Dict
 
 from slackclient import SlackClient
 from sretoolbox.utils import retry
 
 from reconcile.utils.secret_reader import SecretReader
 from reconcile.utils.config import get_config
 
+# Default result sizes to return from APIs
+CHANNELS_LIMIT = os.environ.get('SLACK_API_CHANNELS_LIMIT', 1000)
```
maorfr, line 13: i'm not sure how we'd override these environment variables. in any case, will it be possible to pass them through app-interface? maybe enhance the slack workspace schema?

reconcile/utils/slack_api.py
```diff
@@ -114,24 +120,36 @@ def get_user_id_by_name(self, user_name):
...
-    def get_users_by_ids(self, users_ids):
-        return {k: v['name'] for k, v in self._get('users').items()
+    def get_users_by_ids(self, users_ids, limit=USERS_LIMIT):
+        return {k: v['name'] for k, v in self._get('users',
+                                                   limit=limit).items()
                 if k in users_ids}
 
     @retry()
-    def _get(self, type):
+    def _get(self, type: str, **kwargs) -> Dict:
```
maorfr, line 144: the current consensus is not to use type hints: https://gitlab.cee.redhat.com/service/dev-guidelines/-/merge_requests/76

reconcile/utils/slack_api.py
steveteahan: I'm mostly implementing that to keep the option open without changing code in qontract-reconcile.  Environment variables seemed to be a common pattern for handling this in this project.  In reality, it is probably unlikely that we'll need to set them.  Even if Slack API lowered their `limit`, the current behavior is to ignore the value higher than the limit in favor of the limit, without any errors.

steveteahan: I discussed this further with @maorfr and this wouldn't be the way to handle the configuration.  I'll remove this for now to get the initial fix out to reduce the run time, but I'll follow-up with another PR for handling the configuration via app-interface.

reconcile/utils/slack_api.py
steveteahan: I can remove these if the team has decided to prohibit type annotations.  I'll follow-up with the team to figure out the best path forward for considering an update to this policy.

reconcile/utils/slack_api.py
```diff
@@ -114,24 +119,36 @@ def get_user_id_by_name(self, user_name):
             raise UserNotFoundException(result['error'])
         return result['user']['id']
 
-    def get_channels_by_names(self, channels_names):
-        return {k: v['name'] for k, v in self._get('channels').items()
+    def get_channels_by_names(self, channels_names, limit=CHANNELS_LIMIT):
```
maorfr, line 122: since your next step would be to implement this through app-interface, it would probably make sense to decide the limits inside `_get` (i suppose the limits will be in `self`)

maorfr (approved):
LGTM

good enough to start without app-interface. i think you'll find the `_get_api_results_limit` method redundant once you consume data from app-interface, but we'll get there.

let the iterations begin!

steveteahan:
Thanks!

PR #1776, opened 2021-08-21
use random local port

BumbleFeng:
openshift-saas-deploy-wrapper run multiple OC_Map instances, and they all using the same ssh tunnel for the native client.

When one OC_Map starts to clean up, it will stop the tunnel and the rest instances will fail to connect.

This PR allows each ocm_map to have its own local ports for the tunnels, so it will clean up its own tunnels. Meanwhile, JumpHostSSH will store all the ports globally and generate a unique one when required.

### DIFF
--- a/reconcile/utils/jump_host.py
+++ b/reconcile/utils/jump_host.py
@@ -3,13 +3,17 @@ import shutil
 import os
 import threading
 import logging
-
+import random
 from sshtunnel import SSHTunnelForwarder
 
 import reconcile.utils.gql as gql
 from reconcile.utils.secret_reader import SecretReader
 
 from reconcile.exceptions import FetchResourceError
+
+# https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml
+DYNAMIC_PORT_MIN = 49152
+DYNAMIC_PORT_MAX = 65535
 
 
 class HTTPStatusCodeError(Exception):
@@ -41,6 +45,7 @@ class JumpHostBase:
 
 class JumpHostSSH(JumpHostBase):
     bastion_tunnel = {}
+    local_ports = []
     tunnel_lock = threading.Lock()
 
     def __init__(self, jh, settings=None):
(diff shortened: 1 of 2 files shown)


reconcile/utils/jump_host.py
```diff
@@ -41,13 +41,29 @@ def cleanup(self):
...
     bastion_tunnel = {}
+    local_ports = []
     tunnel_lock = threading.Lock()
 
     def __init__(self, jh, settings=None):
         JumpHostBase.__init__(self, jh, settings=settings)
 
         self.known_hosts = self.get_known_hosts(jh)
         self.init_known_hosts_file()
+        self.local_port = self.get_random_port() \
```
steveteahan, line 52: If you have a separate getter for this variable, should this be considered internal?  ex. `self._local_port` and `self._remote_port`?  Alternatively is there any reason not to allow direct access and you can always use properties if needed later?

reconcile/utils/jump_host.py
```diff
@@ -41,13 +41,29 @@ def cleanup(self):
...
+            if jh['localPort'] is None else jh['localPort']
```
steveteahan, line 53: Could the `localPort` key not exist leading to a `KeyError`?  If so, should you use `jh.get('localPort')`?

reconcile/utils/jump_host.py
```diff
@@ -41,13 +41,29 @@ def cleanup(self):
...
+        self.remote_port = jh['remotePort']
+
+    def get_local_port(self):
+        return self.local_port
+
+    @staticmethod
+    def get_random_port():
```
rrati, line 60: Just for clarity, suggest get_unique_random_port

Purely cosmetic though.  If it causes pain or a bunch of retesting it's not worth it

rrati (approved):
lgtm

reconcile/utils/jump_host.py
rrati: This shouldn't be empty since it is initialized with other jh values, but imo safer is always better.

reconcile/utils/oc.py
```diff
@@ -1038,6 +1040,18 @@ def __init__(self, clusters=None, namespaces=None,
         else:
             raise KeyError('expected one of clusters or namespaces.')
 
+    def set_jh_ports(self, jh):
+        # This will be replaced with getting the data from app-interface in
+        # a future PR.
+        jh['remotePort'] = 8888
+        key = f"{jh['hostname']}-{jh['remotePort']}"
```
Piojo, line 1047: Why `-` instead of `:` in the key? It is a perfectly valid component of a FQDN and can get confusing...

Piojo (commented):
Can we have unit tests for these please? :)

reconcile/utils/jump_host.py
```diff
@@ -41,13 +41,26 @@ def cleanup(self):
...
         self.known_hosts = self.get_known_hosts(jh)
         self.init_known_hosts_file()
+        self.local_port = self.get_random_port() \
+            if jh['localPort'] is None else jh['localPort']
+        self.remote_port = jh['remotePort']
+
+    @staticmethod
+    def get_unique_random_port():
+        with JumpHostSSH.tunnel_lock:
+            port = random.randint(49152, 65535)
```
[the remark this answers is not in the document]
BumbleFeng: This is dynamic port range suggested by https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml

BumbleFeng:
I will add tests for JumpHostSSH later. Currently, this issue is blocking us apply oc naive client on prod cluster.

PR #1780, opened 2021-08-24
peers work alot

Piojo:
- Something something0
- Fix terraform-vpc-peerings crashes under errors in desired state


reconcile/terraform_vpc_peerings.py
```diff
@@ -7,9 +7,9 @@
 
 import reconcile.utils.aws_api as awsapi
 from reconcile.utils.defer import defer
-from reconcile.utils.ocm import OCMMap
-from reconcile.utils.terraform_client import TerraformClient as Terraform
-from reconcile.utils.terrascript_client import TerrascriptClient as Terrascript
+import reconcile.utils.ocm as ocm
```
steveteahan, line 10: I've noticed a pattern where we're getting rid of `from module import Class` imports.  These imports are super common in Python code bases and even suggested in [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports) except when there are name clashes.  Are these being removed because of [this section in our style guide](https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#imports), or some other advantage that I'm missing?

Piojo: I need it to allow testslide to do its magic when it comes to mocking constructors. The way Ptyhon handles namespaces makes it impossible for testslide to reliably mock stuff that is imported into the namespace as `from foo import bar`.

steveteahan: Thanks for the explanation.  So, generally speaking the `from package import Class` imports are fine as long as we avoid `__init__()` with side effects?

Piojo: Yeah, they should be fine as long as there are no side effects.

maorfr:
we should not proceed with `terraform apply` if there were errors. terraform creates/deletes/updates everything in a single command - `terraform apply`.

again, the goal is ONLY to pass dry-run on the specific use case of - missing ARN.

maorfr:
any chance to keep the refactor for a separate PR?

reconcile/terraform_vpc_peerings.py
```diff
@@ -431,8 +429,8 @@ def run(dry_run, print_only=False,
     ts.populate_vpc_peerings(desired_state)
     working_dirs = ts.dump(print_only=print_only)
 
-    if print_only:
-        sys.exit()
+    if print_only or any(errors):
+        sys.exit(0 if dry_run else 1)
```
maorfr, line 433: i think this introduces unwanted behavior. we only want to exit if this is a print-only.
i would change this line to `sys.exit(int(any(errors)))` like you did below, but not change the condition.

reconcile/terraform_vpc_peerings.py
```diff
@@ -182,7 +182,7 @@ def build_desired_state_all_clusters(clusters, ocm_map, settings):
                 cluster_info, ocm_map, settings
             )
             desired_state.extend(items)
-        except Exception:
+        except (KeyError, BadTerraformPeeringState, awsapi.MissingARNError):
```
maorfr, line 185: I think the only exception to catch here is BadTerraformPeeringState, but I may be wrong.
I would still catch Exception after this block and set error to True. Maybe check here if this is a dry-run, and only set error to True if it is. This way we it is guaranteed that we only exit with success in the very particular case of bad terraform peering state AND dry-run.

reconcile/terraform_vpc_peerings.py
```diff
@@ -281,7 +281,7 @@ def build_desired_state_vpc_mesh(clusters, ocm_map, settings):
                 cluster_info, ocm, settings
             )
             desired_state.extend(items)
-        except Exception:
+        except (KeyError, BadTerraformPeeringState, awsapi.MissingARNError):
```
maorfr, line 284: same as above

reconcile/terraform_vpc_peerings.py
```diff
@@ -367,7 +367,7 @@ def build_desired_state_vpc(clusters, ocm_map, settings):
...
```
maorfr, line 370: same as above

reconcile/terraform_vpc_peerings.py
```diff
@@ -454,8 +452,9 @@ def run(dry_run, print_only=False,
         sys.exit(1)
 
     if dry_run:
-        return
-
-    err = tf.apply()
-    if err:
+        sys.exit(0)
```
maorfr, line 455: according to my comments above, i think we should exit with any(errors) and not with `0`.

maorfr: in general, as we are only focusing on dry-run, i don't think we should change this block of code.

maorfr (changes requested):
This is starting to be much clearer. Thanks for making the review simpler!

reconcile/terraform_vpc_peerings.py
Piojo: Not quite. This is a bit complicated (you know best that this is a complex integration!) and I should have explained it properly in the PR description, my apologies. Let me break it one exception class at a time:

### `KeyError`

New clusters have no `id` field, so `ocm.get_aws_infrastructure_access_terraform_assume_role` raises this exception. The first line of this method is:

```python
        cluster_id = self.cluster_ids[cluster]
```

### `BadTerraformPeeringState`

Bad inputs on app-interface lead to either there not being:
* A matching peering for the given provider (`find_matching_peering` returns `None`)
* An AWS account for infrastructure access for the given provider (`aws_account_from_infrastructure_access` returns `None`)
* The requester or accepter VPC IDs not existing or not being accessible in AWS (`get_cluster_vpc_details` returning an empty requester VPC ID)

In these cases, I believe (but may be wrong, this is complicated) that the integration should not fail a PR check.

### `awsapi.MissingARNError`

This is what you've cared about all along. In this case the ARN for the `assume_role` doesn't exist in AWS.

reconcile/terraform_vpc_peerings.py
Piojo: Just for my understanding, do we want to call `tf.plan` under errors?

* If `dry_run` ?
* If `not dry_run` ?

The current version doesn't plan if there are errors, `dry_run` or `not dry_run`. I'm getting confused here.

Piojo:
One final source of confusion remains: do we want to call tf.plan under errors?

* If dry_run ?
* If not dry_run ?

I've been changing this behaviour across iterations and I'm getting very confused here. The current version does `plan` under errors. This was *not* the previous behaviour, since all errors when fetching state lead to an immediate `sys.exit(1)`.

maorfr:
running tf plan under the issue we are trying to solve will not work, as the integration shouldn't have succeeded in getting an ARN to assume. hence it will not be able to check what's going on in the assumed AWS account.

reconcile/utils/aws_api.py
```diff
@@ -633,9 +637,10 @@ def _get_assume_role_session(self, account):
         sts = session.client('sts')
         role_arn = account['assume_role']
         if not role_arn:
-            raise KeyError(
-                'Could not find Role ARN. This is likely caused '
-                'due to a missing awsInfrastructureAccess section.'
+            raise MissingARNError(
```
maorfr, line 640: ok, this makes sense. did i just miss it all along?!

maorfr (approved):
feel free to proceed once you meet this condition:
we only exit with 0 when this is dry-run AND missing ARN. in the case of a missing ARN, we should not run tf plan, as it would fail anyway.

thanks for working so hard on this, most definitely one of the more complex pieces we have.

Piojo:
Cool. Thanks a ton for the detailed review and the clarifications. :)

PR #1782, opened 2021-08-24
allow delete key/value in matchLabels

BumbleFeng:


maorfr (approved):

rporres:
Can you add a test around this? I was trying to set it up but failed to find the problem there.

BumbleFeng:
Sure

PR #1793, opened 2021-08-29
[ocm-upgrade-scheduler] keep track of workload types and days of usage per version

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3638

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/24369

In order to be able to upgrade based on our first condition (soak days of a version), we need to keep track of versions running on our clusters, what workloads each cluster is running.

With this information, we can start building our logic of "to upgrade or not to upgrade" based on the conditions described in the Epic's design document.

### DIFF
--- a/reconcile/ocm_upgrade_scheduler.py
+++ b/reconcile/ocm_upgrade_scheduler.py
@@ -52,19 +52,23 @@ def update_history(history, upgrade_policies):
         history (dict): history in the following format:
         {
           "check_in": "2021-08-29 18:01:27.730441",
-          "version1": {
-            "workload1": {
-              "soak_days": 21,
-              "reporting": [
-                "cluster1",
-                "cluster2"
-              ]
-            },
-            "workload2": {
-              "soak_days": 6,
-              "reporting": [
-                "cluster3"
-              ]
+          "versions": {
+            "version1": {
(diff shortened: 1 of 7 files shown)


esron (approved):

reconcile/ocm_upgrade_scheduler.py
```diff
@@ -87,6 +188,7 @@ def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
 
     ocm_map, current_state = fetch_current_state(clusters)
     desired_state = fetch_desired_state(clusters)
+    # versions_history = get_version_history(dry_run, desired_state, ocm_map)
```
bkez322, line 191: Code and tests for `get_version_history()` look good to me.

I don't understand though how the function is intended to actually be used within the codebase. Is the versions_history output supposed to uploaded and saved somewhere?

maorfr: version_history is already uploaded to a remote state in an s3 bucket. the usage of this output is in #1795

bkez322: Got it 👍

PR #1794, opened 2021-08-30
[qontract-cli] add get version-history command

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3639

depends on #1793 (so only review the last 2 commits)

as version history changes, we want to be able to see information about versions that we are using or have used. based on this information, we will be able to understand why an upgrade is scheduled or not.

the intention is to invoke this command as part of the app-interface-output repository.

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -14,6 +14,7 @@ import reconcile.queries as queries
 import reconcile.openshift_resources_base as orb
 import reconcile.terraform_users as tfu
 import reconcile.terraform_vpc_peerings as tfvpc
+import reconcile.ocm_upgrade_scheduler as ous
 
 from reconcile.slack_base import init_slack_workspace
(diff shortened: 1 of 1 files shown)


reconcile/ocm_upgrade_scheduler.py
```diff
@@ -88,6 +182,7 @@ def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
 
     ocm_map, current_state = fetch_current_state(clusters)
     desired_state = fetch_desired_state(clusters)
+    # versions_history = get_version_history(dry_run, desired_state, ocm_map)
```
esron, line 191: I there a reason for this commented code?

esron: Oh, that is not in the last two commits, sorry,

esron (approved):

PR #1795, opened 2021-08-30
[ocm-upgrade-scheduler] add support to schedule upgrades based on conditions

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3640

depends on #1793

Conditions are defined in an `upgradePolicy` of a cluster. in this ticket, we will switch from using automatic upgrades to manual upgrades that will be set by the ocm-upgrade-scheduler integration if conditions are met for a specific version that is available for upgrade.

### DIFF
--- a/reconcile/ocm_upgrade_scheduler.py
+++ b/reconcile/ocm_upgrade_scheduler.py
@@ -39,6 +39,7 @@ def fetch_desired_state(clusters):
         upgrade_policy = cluster['upgradePolicy']
         upgrade_policy['cluster'] = cluster_name
         upgrade_policy['current_version'] = cluster['spec']['version']
+        upgrade_policy['channel'] = cluster['spec']['channel']
         desired_state.append(upgrade_policy)
 
     return desired_state
--- a/reconcile/ocm_upgrade_scheduler.py
+++ b/reconcile/ocm_upgrade_scheduler.py
@@ -204,7 +204,7 @@ def calculate_diff(current_state, desired_state, ocm_map, version_history):
             ocm.get_available_upgrades(d['current_version'], d['channel'])
         for version in reversed(available_upgrades):
             if version_conditions_met(version, version_history, ocm.name,
-                                       d['workloads'], d['conditions']):
+                                      d['workloads'], d['conditions']):
                 item = {
                     'action': 'create',
                     'cluster': cluster,
--- a/reconcile/ocm_upgrade_scheduler.py
+++ b/reconcile/ocm_upgrade_scheduler.py
@@ -223,9 +223,6 @@ def calculate_diff(current_state, desired_state, ocm_map, version_history):
                 }
                 diffs.append(item)
                 break
-            else:
-                logging.debug(
-                    f'[{cluster}] conditions not met for version {version}')
 
     return diffs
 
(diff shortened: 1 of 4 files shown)


reconcile/utils/ocm.py
```diff
@@ -579,6 +582,28 @@ def delete_machine_pool(self, cluster, spec):
...
+
+        Args:
+            version (string): OpenShift version ID
+            channel (string): Upgrade channel
+
+        Raises:
+            KeyError: if specified channel is not valid
+
+        Returns:
+            list: available versions to upgrade to
```
bkez322, line 597: Does this return the versions in a particular order (ascending numerically, descending numerically)?

reconcile/ocm_upgrade_scheduler.py
```diff
@@ -138,45 +140,93 @@ def get_version_history(dry_run, upgrade_policies, ocm_map):
...
+                                      d['workloads'], d['conditions']):
+                item = {
+                    'action': 'create',
+                    'cluster': cluster,
+                    'version': version,
+                    'schedule_type': 'manual',
+                    'next_run': next_schedule.strftime('%Y-%m-%dT%H:%M:%SZ'),
+                }
+                diffs.append(item)
+                break
```
bkez322, line 216: Due to my confusion [here](https://github.com/app-sre/qontract-reconcile/pull/1795/files#diff-035de9609235cce528e8b2280f5b2d18e688e611ac923346eb60c8ff2c5f9f77R596-R597), it's unclear to me if this is trying to upgrade to the newest or oldest available version.

reconcile/test/test_ocm_upgrade_scheduler.py
```diff
@@ -74,3 +74,65 @@ def test_update_history(self):
             }
         }
         self.assertEqual(expected, history)
+
+
+class TestVersionConditionsMet(TestCase):
```
bkez322, line 79: Potentially may be worth considering an additional test-case proving the outcome if upgrade-condition-soak-days == actual-soak-days (exactly equals).

Otherwise I think this looks good.

reconcile/utils/ocm.py
```diff
@@ -579,6 +582,28 @@ def delete_machine_pool(self, cluster, spec):
...
+        """
+        if channel not in UPGRADE_CHANNELS:
+            raise KeyError(f'channel should be one of {UPGRADE_CHANNELS}')
+        version_id = f'openshift-v{version}'
+        if channel != 'stable':
+            version_id = f'{version_id}-{channel}'
+        api = f'{CS_API_BASE}/v1/versions/{version_id}'
+        return self._get_json(api).get('available_upgrades', [])
```
Piojo, line 605: The caller assumes these versions will always be sorted. Can you specify here or in the docstring if that is guaranteed to always be the case?

reconcile/ocm_upgrade_scheduler.py
maorfr: we're trying to upgrade to the latest version

reconcile/utils/ocm.py
maorfr: it returns it just as the OCM API returns it

reconcile/utils/ocm.py
maorfr: i've changed the caller to use `reversed(sorted(available_upgrades))`, so there is no assumption.

reconcile/test/test_ocm_upgrade_scheduler.py
maorfr: added an equality test. in any case, this case is very unlikely since soak days are a float.

PR #1796, opened 2021-08-30
[ocm-upgrade-scheduler] add support to reject specific versions

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3641

schema: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/24403

as we discover versions which are not working well for us, we need to have a way to define a list of versions we shouldn't upgrade to.

### DIFF
--- a/reconcile/utils/ocm.py
+++ b/reconcile/utils/ocm.py
@@ -855,10 +855,10 @@ class OCM:
         self._post(api, data)
 
     def _init_blocked_versions(self, blocked_versions):
-        if blocked_versions is None:
-            self.blocked_versions = set()
-        else:
+        try:
             self.blocked_versions = set(blocked_versions)
+        except TypeError:
+            self.blocked_versions = set()
 
     @retry(max_attempts=10)
     def _get_json(self, api):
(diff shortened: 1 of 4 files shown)


reconcile/utils/ocm.py
```diff
@@ -25,23 +26,27 @@ class OCM:
...
     :type offline_token: string
     :type init_provision_shards: bool
     :type init_addons: bool
+    :type blocked_version: list
     """
-    def __init__(self, url, access_token_client_id, access_token_url,
+    def __init__(self, name, url, access_token_client_id, access_token_url,
                  offline_token, init_provision_shards=False,
-                 init_addons=False):
+                 init_addons=False, blocked_versions=None):
```
Piojo, line 47: Given how this one is being used later on, can we convert it to a set?

maorfr: it is being passed as a list from app-interface

jmelis (approved):
LGTM!

We did a sync review, and we agreed to:
- add the `set` as suggested by luis
- add an  additional test for multiple blocked versions
- ensure that there is only one upgrade policy as returned by the API

reconcile/utils/ocm.py
maorfr: added handling for a set

jmelis:
For completeness we also reviewed the logic and we agreed that:
- there is no possibility of race condition here.
- there will be an SOP to know how to block versions.

PR #1799, opened 2021-08-31
[ocm-upgrade-scheduler] add support to reject versions based on regex expressions

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3646

follows up on #1796

this PR adds support to define a regex expression in `blockedVersions` in addition to a specific version.
this will allow us to avoid upgrades to feature candidate versions (for example).


rporres (approved):
Looks great. Please document https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/schemas/openshift/openshift-cluster-manager-1.yml#L28 stating that it can be a regex too :)

rporres:
Hmmmm, thinking a bit more about this, I think we should compile the regex somewhere before using it to validate it. At least we could catch regex mistakes at a PR level

rporres:
nit: I've just realised that `version_blocked` should also be updated :)

PR #1802, opened 2021-08-31
[ocm-upgrade-scheduler] properly sort versions

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3647

We are currently using python's builtin `sorted` function, which incorrectly sorts versions when there are feature/release candidates involved. We want release candidates to be considered earlier from a release. In this example, we want a sorting function that sorts `4.8.0-rc.2` to be before `4.8.0`.

### DIFF
--- a/reconcile/ocm_upgrade_scheduler.py
+++ b/reconcile/ocm_upgrade_scheduler.py
@@ -11,6 +11,7 @@ import reconcile.queries as queries
 from reconcile.utils.ocm import OCMMap
 from reconcile.utils.state import State
 from reconcile.utils.data_structures import get_or_init
+from reconcile.utils.semver_helper import sort_versions
 
 QONTRACT_INTEGRATION = 'ocm-upgrade-scheduler'
 
@@ -226,7 +227,7 @@ def calculate_diff(current_state, desired_state, ocm_map, version_history):
         # choose version that meets the conditions and add it to the diffs
         available_upgrades = \
             ocm.get_available_upgrades(d['current_version'], d['channel'])
-        for version in reversed(sorted(available_upgrades)):
+        for version in reversed(sort_versions(available_upgrades)):
             logging.debug(
                 f'[{cluster}] checking conditions for version {version}')
             if ocm.version_blocked(version):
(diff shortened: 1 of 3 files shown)


steveteahan (approved):

bkez322 (commented):
Looks good to me.

The only minor nitpick I can find is that the [semver module](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/semver_helper.py#L1) [being used](https://github.com/app-sre/qontract-reconcile/blob/master/setup.py#L34) here [isn't considered production-stable yet by its maintainers](https://github.com/python-semver/python-semver/blame/master/README.rst#L1-L5). (It says to use version 3.0 which doesn't exist yet as a non-dev release).

But it looks like Q-R has been using this for a while presumably without issue, so whatever.

maorfr:
and in any case i've created https://issues.redhat.com/browse/APPSRE-3648 as a follow up

steveteahan:
> Looks good to me.
> 
> The only minor nitpick I can find is that the [semver module](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/semver_helper.py#L1) [being used](https://github.com/app-sre/qontract-reconcile/blob/master/setup.py#L34) here [isn't considered production-stable yet by its maintainers](https://github.com/python-semver/python-semver/blame/master/README.rst#L1-L5). (It says to use version 3.0 which doesn't exist yet as a non-dev release).
> 
> (or maybe the readme is just super confusing to me, and that readme line specifically is referring to the 3.00-dev versions, and 2.X is fine)
> 
> But it looks like Q-R has been using this for a while presumably without issue, so whatever.

That line is associated with the 3.0-dev versions.  It looks like 2.13.0 is their stable version as per the [docs](https://python-semver.readthedocs.io/en/stable/) whereas [3.0 is still in active development](https://python-semver.readthedocs.io/en/latest/).  Their master branch is [tracking dev](https://github.com/python-semver/python-semver/blob/master/src/semver/__about__.py#L19) and there is a separate branch for v2.

PR #1803, opened 2021-08-31
[ocm] validate blocked versions regex expressions

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3646

follows up in #1799

this PR adds validation that all blocked versions are valid regex expressions.

### DIFF
--- a/reconcile/test/test_utils_ocm.py
+++ b/reconcile/test/test_utils_ocm.py
@@ -1,3 +1,5 @@
+import pytest
+
 from unittest import TestCase
 from unittest.mock import patch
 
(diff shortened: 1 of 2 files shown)


reconcile/test/test_utils_ocm.py
```diff
@@ -41,3 +43,16 @@ def test_version_not_blocked_regex(self):
...
+
+class TestVersionRegex(TestCase):
+    @patch.object(OCM, '_init_access_token')
+    @patch.object(OCM, '_init_request_headers')
+    @patch.object(OCM, '_init_clusters')
+    # pylint: disable=arguments-differ
+    # pylint: disable=no-self-use
+    def test_invalid_regex(self, ocm_init_access_token,
+                           ocm_init_request_headers, ocm_init_clusters):
+        with pytest.raises(TypeError):
```
rporres, line 56: why not using [`assertRaises`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) directly and sticking to `unittest` here?

maorfr: just because i'm a noob.

Piojo: Yeah, as Rafa says,
```python
with self.assertRaises(TypeError):
    ...
```
is the unittest way.

Piojo (approved):

reconcile/test/test_utils_ocm.py
rporres: thx for fixing it.

PR #1805, opened 2021-09-01
[qontract-cli] sort output

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3572

this PR adds an option to sort (or not) the output of `get` commands.

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -53,14 +53,18 @@ def root(ctx, configfile):
 @output
 @click.pass_context
 def get(ctx, output):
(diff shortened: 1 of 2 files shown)


Piojo:
While it looks good to me, tests are failing:
```
10:39:32         columns = ['name', 'latest']
10:39:32 >       print_output(ctx.obj['options'], checkpoints_data, columns)
10:39:32 E       KeyError: 'options'
10:39:32 
10:39:32 tools/qontract_cli.py:602: KeyError
```
looks like some click command needs updating.

PR #1806, opened 2021-09-01
[terraform-tgw-attachments] add support for non OCM environments

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3645

this PR will enable the `terraform-tgw-attachments` integration to operate in an environment without OCM (OCP clusters).
in such an environment, a role ARN to assume should be supplied in the TGW definitions in the cluster file.

app-interface counterpart: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/24596
(even though we are not going to implement this into this instance of app-interface, but only for fedramp's root environment).

### DIFF
--- a/reconcile/terraform_tgw_attachments.py
+++ b/reconcile/terraform_tgw_attachments.py
@@ -121,7 +121,7 @@ def run(dry_run, print_only=False,
     settings = queries.get_app_interface_settings()
     clusters = [c for c in queries.get_clusters()
                 if c.get('peering') is not None]
-    with_ocm = any([c.get('ocm') for c in clusters])
+    with_ocm = any(c.get('ocm') for c in clusters)
     if with_ocm:
         ocm_map = OCMMap(clusters=clusters, integration=QONTRACT_INTEGRATION,
                          settings=settings)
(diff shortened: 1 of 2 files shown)


andreugallofre (approved):

reconcile/terraform_tgw_attachments.py
```diff
@@ -114,8 +121,15 @@ def run(dry_run, print_only=False,
     settings = queries.get_app_interface_settings()
     clusters = [c for c in queries.get_clusters()
                 if c.get('peering') is not None]
-    ocm_map = OCMMap(clusters=clusters, integration=QONTRACT_INTEGRATION,
-                     settings=settings)
+    with_ocm = any([c.get('ocm') for c in clusters])
```
Piojo, line 124: Nit: you don't really need to build the intermediate list:
```suggestion
    with_ocm = any(c.get('ocm') for c in clusters)
```

Piojo (approved):

reconcile/terraform_tgw_attachments.py
jmelis: Interesting! what is the intermediate data type? a set?

jmelis: Hmm I guess it just calls `__iter__`

PR #1810, opened 2021-09-02
APPSRE-3522 openshift namespace labels integration

patjlm:
I believe this could be much simpler and to the point, but i don't see how for now.
Creating the PR before diving into other potential changes, to get reviews.
I'd be happy to get feedback and revisit as needed.

### DIFF
--- a/reconcile/openshift_namespace_labels.py
+++ b/reconcile/openshift_namespace_labels.py
@@ -16,7 +16,6 @@ from reconcile.utils.sharding import is_in_shard
 from reconcile.utils.state import State
 
 _LOG = logging.getLogger(__name__)
-# _LOG.setLevel(logging.DEBUG)
 
 QONTRACT_INTEGRATION = 'openshift-namespace-labels'
 
@@ -97,7 +96,7 @@ class LabelInventory():
             self._ns(cluster, namespace)[type] = labels
             return labels
 
-    def delete(self, cluster, namespace):
+    def delete(self, cluster: str, namespace: str):
         """Delete the given cluster / namespace from the inventory"""
         with self._lock:
             self._inv.get(cluster, {}).pop(namespace, None)
@@ -109,7 +108,7 @@ class LabelInventory():
             for namespace, types in namespaces.items():
                 yield cluster, namespace, types
 
-    def update_managed_keys(self, cluster, namespace, key):
+    def update_managed_keys(self, cluster: str, namespace: str, key: str):
         """
         Add or remove a key from the managed key list.
         This actually handles a copy of the managed keys dict and updates it.
@@ -117,8 +116,12 @@ class LabelInventory():
         If it was not, it will get added
         """
         managed = self.get(cluster, namespace, MANAGED, [])
+        if managed is None:
+            managed = []
         upd_managed = self.setdefault(cluster, namespace, UPDATED_MANAGED,
                                       managed.copy())
+
+        assert isinstance(upd_managed, list)  # make mypy happy
         if key in managed:
             upd_managed.remove(key)
         else:
@@ -167,7 +170,7 @@ class LabelInventory():
                     changed[k] = None
 
 
-def get_names_for_namespace(namespace) -> Tuple[str, str]:
+def get_names_for_namespace(namespace: Dict[str, Any]) -> Tuple[str, str]:
     """
     Get the cluster and namespace names from the provided
     namespace qontract info
@@ -186,8 +189,8 @@ def get_gql_namespaces_in_shard() -> List[Any]:
             if is_in_shard(f"{ns['cluster']['name']}/{ns['name']}")]
 
 
-def get_oc_map(namespaces, internal,
-               use_jump_host, thread_pool_size) -> OC_Map:
+def get_oc_map(namespaces: List[Any], internal: Optional[bool],
+               use_jump_host: bool, thread_pool_size: int) -> OC_Map:
     """
     Get an OC_Map for our namespaces
     """
@@ -199,7 +202,8 @@ def get_oc_map(namespaces, internal,
                   init_projects=True)
 
 
-def get_desired(inventory: LabelInventory, oc_map: OC_Map, namespaces) -> None:
+def get_desired(inventory: LabelInventory, oc_map: OC_Map,
+                namespaces: List[Any]) -> None:
     """
     Fill the provided label inventory with every desired info from the
     input namespaces. Ocm_map is used to not register clusters which are
@@ -241,11 +245,11 @@ def get_desired(inventory: LabelInventory, oc_map: OC_Map, namespaces) -> None:
         inventory.delete(cluster, ns_name)
 
 
-def state_key(cluster, namespace):
+def state_key(cluster: str, namespace: str):
     return f"{cluster}/{namespace}-managed-labels"
 
 
-def get_managed(inventory: LabelInventory, namespaces, state: State) -> None:
+def get_managed(inventory: LabelInventory, state: State) -> None:
     """
     Fill the label inventory with the list of currently managed labels
     for each cluster & namespace. This information is retrieved from the state
@@ -253,7 +257,6 @@ def get_managed(inventory: LabelInventory, namespaces, state: State) -> None:
     """
     keys = state.ls()
     # We could run threaded here: probably faster but more parallel requests.
-    # for namespace in namespaces:
     for cluster, ns_name, types in inventory:
         if types.get(DESIRED) is None:
             continue
@@ -265,7 +268,7 @@ def get_managed(inventory: LabelInventory, namespaces, state: State) -> None:
         inventory.set(cluster, ns_name, MANAGED, managed)
 
 
-def lookup_namespaces(cluster, oc_map: OC_Map):
+def lookup_namespaces(cluster: str, oc_map: OC_Map):
     """
     Retrieve all namespaces from the given cluster
     """
@@ -372,7 +375,7 @@ def run(dry_run: bool, thread_pool_size: int = 10, internal: bool = None,
     state = State(integration=QONTRACT_INTEGRATION,
                   accounts=accounts, settings=settings)
     _LOG.debug('Collecting managed state ...')
-    get_managed(inventory, namespaces, state)
+    get_managed(inventory, state)
 
     _LOG.debug('Collecting current state ...')
     get_current(inventory, oc_map, thread_pool_size)
--- a/reconcile/openshift_namespace_labels.py
+++ b/reconcile/openshift_namespace_labels.py
@@ -205,26 +205,40 @@ def get_desired(inventory: LabelInventory, oc_map: OC_Map, namespaces) -> None:
     input namespaces. Ocm_map is used to not register clusters which are
     unreachable or not configured (due to --internal / --external)
     """
+    to_be_ignored = []
     for ns in namespaces:
         if 'labels' not in ns:
             continue
+
         cluster, ns_name = get_names_for_namespace(ns)
         # Skip unreachable / non-hanlded clusters
         # eg: internal settings may not match --internal / --external param
         if cluster not in oc_map.clusters():
             continue
         labels = json.loads(ns['labels'])
+
         validation_errors = validate_labels(labels)
         for err in validation_errors:
             inventory.add_error(cluster, ns_name, err)
-        if inventory.get(cluster, ns_name, DESIRED) is not None:
-            inventory.add_error(cluster, ns_name,
-                                'Found several namespaces for ' +
-                                f'{cluster}/{ns_name}')
-            inventory.delete(cluster, ns_name)
         if inventory.errors(cluster, ns_name):
             continue
+
+        if inventory.get(cluster, ns_name, DESIRED) is not None:
+            # delete at the end of the loop to avoid having a reinsertion at
+            # the third/fifth/.. occurrences
+            to_be_ignored.append((cluster, ns_name))
+            continue
+
         inventory.set(cluster, ns_name, DESIRED, labels)
+
+    for cluster, ns_name in to_be_ignored:
+        # Log only a warning here and do not report errors nor fail the
+        # integration.
+        # A dedicated integration or PR check will be done to ensure this
+        # case does not occur (anymore)
+        _LOG.warning(f'Found several namespace defintions for '
+                     f'{cluster}/{ns_name}. Ignoring')
+        inventory.delete(cluster, ns_name)
 
 
 def state_key(cluster, namespace):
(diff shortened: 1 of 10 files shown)


reconcile/utils/oc.py
```diff
@@ -339,6 +339,18 @@ def delete(self, namespace, kind, name, cascade=True):
         resource = {'kind': kind, 'metadata': {'name': name}}
         return self._msg_to_process_reconcile_time(namespace, resource)
 
+    @OCDecorators.process_reconcile_time
+    def label(self, namespace, kind, name, labels, overwrite=False):
+        ns = ['-n', namespace] if namespace else []
+        added = [f"{k}='{v}'" for k, v in labels.items() if v is not None]
+        removed = [f"{k}-" for k, v in labels.items() if v is None]
```
maorfr, line 346: :heart:

reconcile/openshift_namespace_labels.py
```diff
@@ -0,0 +1,419 @@
...
+from reconcile.utils.defer import defer
+from reconcile.utils.oc import OC_Map, OCNative, StatusCodeError
+
+from reconcile.utils.sharding import is_in_shard
+from reconcile.utils.state import State
+
+_LOG = logging.getLogger(__name__)
+_LOG.setLevel(logging.DEBUG)
+
+QUERY = """
```
maorfr, line 20: can we extend `queries.get_namespaces` and use it instead?

reconcile/openshift_namespace_labels.py
```diff
@@ -0,0 +1,419 @@
...
+                  use_jump_host=use_jump_host,
+                  thread_pool_size=thread_pool_size,
+                  init_projects=True)
+
+
+class LabelError(Exception):
+    pass
+
+
+def validate_labels(labels) -> Iterable[str]:
```
maorfr, line 214: perhaps this can fit into `utils/oc`? just a thought

reconcile/openshift_namespace_labels.py
```diff
@@ -0,0 +1,419 @@
...
+        managed = state.get(key, [])
+        inventory.set(cluster, ns_name, MANAGED, managed)
+
+
+def lookup_namespaces(cluster, oc_map: OC_Map):
+    """
+    Retrieve all namespaces from the given cluster
+    """
+    try:
+        oc: OCNative = oc_map.get(cluster)
```
maorfr, line 310: what is the purpose of `oc: OCNative`?

reconcile/test/test_openshift_namespace_labels.py
```diff
@@ -0,0 +1,295 @@
+import json
```
rporres, line 1: This test is very hard to follow in its current shape. It might be easier to understand if you separated it smaller pieces that contain the different scenarios which are easier to digest. If you need to prepare similar scenarios, the use of set up / tear down unit test functions to prepare the expected data that will be fed into a smaller unit test. A small comment about the general test strategy would be helpful.

The use of a unit testing framework such as unittest or pytest is definitely recommend while writing tests. It could also be helpful with your mocking strategy.

rporres (changes requested):
The code follows what the design doc says. It's not straight-forward but it's clear. There's some missing docstrings in the Inventory class, but I assume this is part of the WIP status. I personally don't see how to approach it differently, maybe others will.

I would rework the tests to handle separately every case as in its current shape it's hard to follow.

patjlm (commented):
~~Need to `make generate` but first trying to check memory consumption.~~
Done

reconcile/openshift_namespace_labels.py
```diff
@@ -0,0 +1,419 @@
...
+        if cluster not in oc_map.clusters():
+            continue
+        labels = json.loads(ns['labels'])
+        validation_errors = validate_labels(labels)
+        for err in validation_errors:
+            inventory.add_error(cluster, ns_name, err)
+        if inventory.get(cluster, ns_name, DESIRED) is not None:
+            inventory.add_error(cluster, ns_name,
+                                'Found several namespaces for ' +
+                                f'{cluster}/{ns_name}')
```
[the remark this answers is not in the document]
rporres: From: https://coreos.slack.com/archives/GGC2A0MS8/p1631028931078100

To avoid cluttering the output you could just see if you have any and log an info message saying that we have reached this case and to run the integration in debug mode.

Until we can enforce uniqueness in app-interface namespace objects, we cannot do better, I won't fail the integration for this, it is not its job.

reconcile/test/test_openshift_namespace_labels.py
rporres: The new tests look great. Things are much easier to understand, I love the usage of set up and tear down functions to avoid the repetition of annotations.

I agree, we don't need more tests for `reconcile()`

reconcile/test/test_openshift_namespace_labels.py
```diff
@@ -0,0 +1,297 @@
...
+from unittest import TestCase
+from unittest.mock import Mock, call, patch
+
+from reconcile import openshift_namespace_labels
+from reconcile.openshift_namespace_labels import NamespaceLabelError, state_key
+
+
+# TODO: use a dataclass when python > 3.6
+class NS():
+    def __init__(self, cluster: str, name: str,
```
rporres, line 13: can you add a small docstring/comment explaining what's this class about?

rporres:
This looks great. I guess the only thing missing is the duplicated namespaces handling, correct?

reconcile/utils/oc.py
```diff
@@ -1138,3 +1152,42 @@ def __init__(self, log_level, message):
 
     def __bool__(self):
         return False
+
+
+def validate_labels(labels) -> Iterable[str]:
```
steveteahan, line 1157: If we're going to type the output, worth typing the input as well?

Also, worthy of a unit test?  Seems like something that is simple enough to test (no external dependencies to mock).

reconcile/test/test_openshift_namespace_labels.py
```diff
@@ -0,0 +1,297 @@
...
+        oc.label.assert_has_calls([call(None, 'Namespace', 'add', k2v2,
+                                        overwrite=True)])
+
+    def test_add_from_none(self):
+        """ addition of a label from none on the current namespace"""
+        self.test_ns = [
+            NS(c1, 'add-from-none', {}, None, k1v1),
+        ]
+        run_integration()
+        self.state.add.assert_called_once()
```
steveteahan, line 188: nit: Could you get rid of a bunch of lines with `assert_called_once_with`?

steveteahan (commented):
A couple minor points for feedback.  I like the usage of typing!

reconcile/utils/oc.py
```diff
@@ -1138,3 +1152,42 @@ def __init__(self, log_level, message):
...
+    https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
+    """
+    if labels is None:
+        return []
+
+    err = []
+    for k, v in labels.items():
+        if len(v) > 63:
+            err.append(f'Label value longer than 63 chars: {v}')
+        pattern = '^[A-Za-z0-9][A-Za-z0-9-_\\.]{0,61}[A-Za-z0-9]$'
```
Piojo, line 1169: Can you compile this regex outside the loop? It'll be easier to follow later on.

reconcile/utils/oc.py
```diff
@@ -1138,3 +1152,42 @@ def __init__(self, log_level, message):
...
+        if not re.match(pattern, v):
+            err.append(f'Label value is invalid: {v}')
+
+        prefix, name = '', k
+        if '/' in k:
+            split = k.split('/')
+            if len(split) > 3:
+                err.append(f'Only one "/" allowed in label keys: {k}')
+            prefix, name = split[0], split[1]
+        if len(name) > 63:
```
Piojo, line 1179: Please turn this magic number into a constant.

reconcile/utils/oc.py
```diff
@@ -1138,3 +1152,42 @@ def __init__(self, log_level, message):
...
+            err.append(f'Label key name is longer than 63 chars: {name}')
+        if not re.match(pattern, name):
+            err.append(f'Label key name is invalid: {name}')
+        if prefix:
+            if len(prefix) > 253:
```
Piojo, line 1184: 253 must be the weirdest magic number I've ever found. :| Can you turn it into a constant? :)

reconcile/utils/oc.py
```diff
@@ -1138,3 +1152,42 @@ def __init__(self, log_level, message):
...
+                err.append(f'Label key prefix longer than 253 chars: {prefix}')
+            hostname = ('^(([a-zA-Z]|[a-zA-Z][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*'
```
Piojo, line 1186: Compile this regexp outside the loop, please.

reconcile/utils/oc.py
patjlm: yes, you're right, i'll do that

reconcile/openshift_namespace_labels.py
```diff
@@ -0,0 +1,419 @@
...
+    """
+    Validate a label key/value against some rules from
+    https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
+    """
+    if labels is None:
+        return []
+
+    err = []
+    for k, v in labels.items():
+        if len(v) > 63:
```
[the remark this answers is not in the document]
patjlm: that's the spec. I can make a constant

maorfr (approved):

rporres (approved):

reconcile/openshift_namespace_labels.py
```diff
@@ -0,0 +1,393 @@
...
+import reconcile.utils.threaded as threaded
+from reconcile.utils.defer import defer
+from reconcile.utils.oc import OC_Map, OCNative, StatusCodeError
+from reconcile.utils.oc import validate_labels
+
+from reconcile.utils.sharding import is_in_shard
+from reconcile.utils.state import State
+
+_LOG = logging.getLogger(__name__)
+# _LOG.setLevel(logging.DEBUG)
```
esron, line 19: I think that you forgot to delete this line.

reconcile/openshift_namespace_labels.py
```diff
@@ -0,0 +1,395 @@
...
+                yield cluster, namespace, types
+
+    def update_managed_keys(self, cluster: str, namespace: str, key: str):
+        """
+        Add or remove a key from the managed key list.
+        This actually handles a copy of the managed keys dict and updates it.
+        If the key was managed, it will get removed.
+        If it was not, it will get added
+        """
+        managed = self.get(cluster, namespace, MANAGED, [])
```
steveteahan, line 118: So, I thought about this more... what about this case:

```
inv = {'cluster': {'namespace': {'managed': None}}}
>>> type(inv.get('cluster').get('namespace').get('managed', []))
<class 'NoneType'>
```

The key is set, therefore `None` can still be returned instead of an empty list.  So, adding the code below would make a None value impossible while making Mypy happy.

```
        if managed is None:
            managed = []
```

The same idea would apply with `setdefault`.  That won't prevent a `None` if the key already exists.

steveteahan: Also, more on this here: https://mypy.readthedocs.io/en/stable/kinds_of_types.html#strict-optional

reconcile/openshift_namespace_labels.py
```diff
@@ -0,0 +1,395 @@
...
+
+DESIRED = 'desired'
+MANAGED = 'managed'
+CURRENT = 'current'
+CHANGED = 'changed'
+UPDATED_MANAGED = 'updated-managed'
+
+Labels = Dict[str, Optional[str]]
+LabelKeys = List[str]
+LabelsOrKeys = Union[Labels, LabelKeys, None]
```
steveteahan, line 30: One other minor thing to consider here is that unless `None` is really useful in this case for a non-obvious reason, it could make more sense to just use `Optional[LabelsOrKeys]` where you need to while dropping None from the union.  I'll leave that up to you, but something like:

```
def some_func() -> Optional[LabelsOrKeys]
```

is much more obvious to me that the function can return `None`.  Mypy doesn't care, just more of a consideration for people reading through the code.

reconcile/openshift_namespace_labels.py
```diff
@@ -0,0 +1,395 @@
...
+        If the key was managed, it will get removed.
+        If it was not, it will get added
+        """
+        managed = self.get(cluster, namespace, MANAGED, [])
+        if managed is None:
+            managed = []
+        upd_managed = self.setdefault(cluster, namespace, UPDATED_MANAGED,
+                                      managed.copy())
+
+        assert isinstance(upd_managed, list)  # we never get a Dict here
```
steveteahan, line 123: All of the comments below are just ideas how we might be able to avoid this case in the future.  I am not suggesting that they are necessary for this PR at all.  The occasional `assert` where we know that something truly can't happen is reasonable, but I'd be more concerned if this becomes a regular thing.

For what it's worth, I found: https://pypi.org/project/typing-extensions/

That allows us to import `TypedDict` ([docs](https://docs.python.org/3/library/typing.html#typing.TypedDict)) and others more recent features from the `typing` module before we upgrade to 3.9.  I think that I'll go ahead and add that to my Mypy PR.  A `TypedDict` would provide the link between the `UPDATE_MANAGED` key and the `list` type, compared the other keys that are returning `dict`.

It is worth mentioning that `TypedDict`s are fairly strict by design.  You'd need to use `total=False` if all of the keys aren't guaranteed to be present and you can't use a variable typed as a `str` during a lookup without an error, so you'd need to do something like:

```
StateKeys = Literal['desired', 'managed', 'current', 'changed',
                    'update_managed']
```

Thanks for using this as an opportunity to add types to this new integration!

steveteahan: Just FYI, I went down the rabbit hole and even with a `TypedDict`, I get the same error message if I remove your `assert`:

```
reconcile/openshift_namespace_labels.py:135: error: Item "Dict[str, Optional[str]]" of "Union[Dict[str, Optional[str]], List[str]]" has no attribute "remove"
reconcile/openshift_namespace_labels.py:137: error: Item "Dict[str, Optional[str]]" of "Union[Dict[str, Optional[str]], List[str]]" has no attribute "append"
```

The Types dict would look like:

```
class Types(TypedDict, total=False):
    desired: Labels
    managed: LabelKeys
    current: Labels
    changed: Labels
    updated_managed: LabelKeys
```


But if you change the code to:

```
        managed = self._inv.get(cluster, {}).get(namespace, {}).get('managed')
        if managed is None:
            managed = []
        upd_managed = self._ns(cluster, namespace).setdefault(
            "updated_managed", managed.copy())
```

Then, Mypy is happy (with this issue, but there were a few others).  So, basically if the method can return A or B, and the `TypedDict` indicates that given a certain key A is returned, Mypy will still assume A or B because that's what your method signature indicates.  You'd need to directly access the `TypedDict` from what I'm seeing (kind of makes sense).

I'll dig into this again when I have some free time because I think these sorts of issues are going to be where we feel the most pain.  We may just find `dataclass` is the way to go where the level of nesting is not large and the attributes are well-known.

PR #1813, opened 2021-09-03
Re-enable deep-syncs

Piojo:
To be merged after we have verified the effectiveness of the caching
strategy.

### DIFF
--- a/reconcile/quay_mirror.py
+++ b/reconcile/quay_mirror.py
@@ -210,9 +210,6 @@ class QuayMirror:
         return sync_tasks
 
     def _is_deep_sync(self, interval):
-        # pylint: disable=unreachable
-        return False  # temporary
-
         control_file_name = 'qontract-reconcile-quay-mirror.timestamp'
         control_file_path = os.path.join(tempfile.gettempdir(),
                                          control_file_name)


rporres:
Where can we see the results of that verification?

Piojo:
Mmm... for what I see, we only make a significant use of manifests during deep syncs. In the mean time, it is "safe" because the integration [isn't crashing](https://prometheus.app-sre-prod-01.devshift.net/new/graph?g0.expr=max(qontract_reconcile_last_run_status%7Bintegration%3D%22quay-mirror%22%7D)&g0.tab=1&g0.stacked=0&g0.range_input=1w&g0.end_input=2021-09-06%2012%3A50%3A23&g0.moment_input=2021-09-06%2012%3A50%3A23) and because it's [faster now](https://prometheus.app-sre-prod-01.devshift.net/new/graph?g0.expr=qontract_reconcile_last_run_seconds%7Bintegration%3D%22quay-mirror%22%7D&g0.tab=0&g0.stacked=0&g0.range_input=1w&g0.end_input=2021-09-06%2012%3A50%3A23&g0.moment_input=2021-09-06%2012%3A50%3A23).  Since we barely download manifests outside deep sync windows, the metrics for caches are empty. :(

On a local test with deep sync, I didn't see any crashes and on a cold cache the hit rate rose quickly to about 62%. I guess (but this is just a guess) it'll be higher as the integration runs multiple times.

rporres:
ok, then go for it :)

rporres (approved):

PR #1814, opened 2021-09-05
get_repo_team_permissions message may not exist

maorfr:
in the quay_api, when we get a 403, the `message` key may not exist, causing us to raise the wrong exception.


patjlm:
lgtm

PR #1815, opened 2021-09-05
[terraform-resources] add support for iam role

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3669

in order to be able to provision iam roles.

this is required for https://issues.redhat.com/browse/APPSRE-3645.

### DIFF
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -1686,6 +1686,11 @@ class TerrascriptClient:
         role_tf_resource = aws_iam_role(identifier, **values)
         tf_resources.append(role_tf_resource)
 
+        # output role arn
+        output_name_0_13 = output_prefix + '__role_arn'
+        output_value = '${' + role_tf_resource.arn + '}'
+        tf_resources.append(Output(output_name_0_13, value=output_value))
+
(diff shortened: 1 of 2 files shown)


reconcile/utils/terrascript_client.py
```diff
@@ -803,6 +803,8 @@ def populate_tf_resources(self, populate_spec, existing_secrets,
             self.populate_tf_resource_service_account(resource,
                                                       namespace_info,
                                                       ocm_map=ocm_map)
+        elif provider == 'aws-iam-role':
+            self.populate_tf_resource_role(resource, namespace_info)
```
Piojo, line 807: I guess we could turn this into a dictionary lookup and a dispatcher, as part of https://issues.redhat.com/projects/APPSRE/issues/APPSRE-3505 .

Piojo (approved):
For consistency, I'd appreciate a unit test or two. In any case, it looks good.

reconcile/utils/terrascript_client.py
maorfr: for consistency, this class has no tests until now. not easy to test. happy to do this separately though.

PR #1816, opened 2021-09-06
Improve error handling in quay-permissions

Piojo:
There are more methods that can raise exceptions, so we cover each
iteration (i.e, each reviewed repository) in a try-except block.

Also, we replace the old `RequestsException` with the `HTTPError`
class supplied by `requests`. This gives us more information,
including access to the failed response and request.

This way, we can get complete error messages, such as:

```[2021-09-06 22:12:54] [ERROR] [quay_permissions.py:run:116] - Could not manage repo permissions on repo $REPO_NAME,team $TEAM_NAME - status code: 403 - resource URL: https://quay.io/api/v1/repository/$REPOSITORY_SHA/$REPOSITORY_NAME/permissions/team/$TEAM_NAME
[...]
requests.exceptions.HTTPError: 403 Client Error: FORBIDDEN for url: ...
```

### DIFF
--- a/reconcile/quay_permissions.py
+++ b/reconcile/quay_permissions.py
@@ -86,39 +86,38 @@ def run(dry_run):
                             continue
 
                         perm_org_key = (
-                          permission['quayOrg']['instance']['name'],
-                          permission['quayOrg']['name']
+                            permission['quayOrg']['instance']['name'],
+                            permission['quayOrg']['name']
                         )
 
                         if perm_org_key != org_key:
-                            logging.warning('wrong org, should be %s', org_key)
+                            logging.warning(
+                                f'wrong org, should be {org_key}'
+                            )
                             continue
 
                         team_name = permission['team']
 
                         # processing team section
                         logging.debug(['team', team_name])
-
-                        current_role = \
-                            quay_api.get_repo_team_permissions(
-                                repo_name, team_name)
-                        if current_role != role:
-                            logging.info(
-                                ['update_role', org_key, repo_name,
-                                 team_name, role])
-                            if not dry_run:
-                                try:
+                        try:
+                            current_role = \
+                                quay_api.get_repo_team_permissions(
+                                    repo_name, team_name)
+                            if current_role != role:
+                                logging.info(
+                                    ['update_role', org_key, repo_name,
+                                     team_name, role])
+                                if not dry_run:
                                     quay_api.set_repo_team_permissions(
                                         repo_name, team_name, role)
-                                except Exception as e:
-                                    error = True
-                                    logging.error(
-                                        'could not set repo permissions: '
-                                        'repo name: %s, '
-                                        'team name: %s. '
-                                        'details: {%s}',
-                                        repo_name, team_name, e
-                                    )
+                        except Exception:
+                            error = True
+                            logging.exception(
+                                'could not manage repo permissions: '
+                                f'repo name: {repo_name}, '
+                                f'team name: {team_name}'
+                            )
 
     if error:
         sys.exit(ExitCodes.ERROR)
--- a/reconcile/utils/quay_api.py
+++ b/reconcile/utils/quay_api.py
@@ -59,6 +59,10 @@ class QuayApi:
         return True
 
     def remove_user_from_team(self, user, team):
+        """Deletes an user from a team.
+
+        :raises HTTPError if there are any problems with the request
+        """
         url_team = "{}/organization/{}/team/{}/members/{}".format(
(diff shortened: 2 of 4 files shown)


reconcile/utils/quay_api.py
```diff
@@ -124,9 +116,7 @@ def create_or_update_team(self, team: str, role="member",
             payload.update({'description': description})
 
         r = requests.put(url, headers=self.auth_header, json=payload)
-
-        if not r.ok:
-            raise RequestsException(r)
+        r.raise_for_status()
```
rporres, line 119: Please update docstring to reflect this change

reconcile/utils/quay_api.py
```diff
@@ -97,8 +90,7 @@ def add_user_to_team(self, user, team):
         url = "{}/organization/{}/team/{}/members/{}".format(
             self.api_url, self.organization, team, user)
         r = requests.put(url, headers=self.auth_header)
-        if not r.ok:
-            raise RequestsException(r)
+        r.raise_for_status()
```
rporres, line 93: can you add a docstring describing this?

reconcile/utils/quay_api.py
```diff
@@ -43,8 +37,7 @@ def list_team_members(self, team, **kwargs):
                 f"org {self.organization}. "
                 f"contact org owner to create the "
                 f"team manually.")
-        elif not r.ok:
-            raise RequestsException(r)
+        r.raise_for_status()
```
rporres, line 40: Please update docstring to reflect this change

reconcile/quay_permissions.py
```diff
@@ -77,48 +78,54 @@ def run(dry_run):
...
-                        current_role = \
-                            quay_api.get_repo_team_permissions(
-                                repo_name, team_name)
-                        if current_role != role:
-                            logging.info(
-                                ['update_role', org_key, repo_name,
-                                 team_name, role])
-                            if not dry_run:
-                                try:
+                    try:
```
rporres, line 81: This `try` block is either navigating through dicts or doing calls. I don't get understand need for such a broad scope. Could we just start it in `get_repo_team_permissions`?

reconcile/quay_permissions.py
```diff
@@ -77,48 +78,54 @@ def run(dry_run):
...
-                                        'details: {%s}',
-                                        repo_name, team_name, e
-                                    )
+                    except HTTPError:
+                        error = True
+                        logging.exception(
+                            f'Could not manage repo permissions on {repo_name}'
+                            f' - team {team_name}'
+                        )
+                    except Exception:
```
rporres, line 122: I see no different treatment in the catch-all or in the `HTTPError`. Why the difference? If we reduced the scope of the `try` block, could we narrow the exceptions here?

Piojo: It started with somewhat different messages, and then converged, sorry, I'll simplify this.

reconcile/quay_permissions.py
Piojo: The goal here is to do as much work as possible even if some repositories or teams fail.

If we can't even get repo permissions we can't continue the work of the iteration. Just having `try: except: log;continue` is more and harder-to-read code. Just having the happy path inside the `try` block leads to less code, and makes sure any other mistake I may have forgotten to consider is handled. This is the top-level in which we want unexpected exceptions handled and reported grafecully.

reconcile/quay_permissions.py
```diff
@@ -86,39 +86,38 @@ def run(dry_run):
                             continue
 
                         perm_org_key = (
-                          permission['quayOrg']['instance']['name'],
-                          permission['quayOrg']['name']
+                            permission['quayOrg']['instance']['name'],
+                            permission['quayOrg']['name']
```
rporres, line 90: please avoid unrelated linting changes.

reconcile/quay_permissions.py
```diff
@@ -86,39 +86,38 @@ def run(dry_run):
...
                         )
 
                         if perm_org_key != org_key:
-                            logging.warning('wrong org, should be %s', org_key)
+                            logging.warning(
+                                f'wrong org, should be {org_key}'
+                            )
```
rporres, line 96: There are endless discussions about avoiding interpolation in logging or not... In this case it should not matter but I'm all for avoiding unrelated changes in a PR 😂

rporres (approved):
a couple of nits, nothing important

Piojo:
I hadn't pushed my fixes to your docstring requests. Sorry!  They are in now.

PR #1817, opened 2021-09-06
Increase memory allocation for quay-mirror

Piojo:
To store the in-memory cache we introduce as part of
https://issues.redhat.com/browse/APPSRE-3385 . The initial estimate is
50,000 failed requests at 20KB each - manifests are usually around
10KB, so this should give us plenty of room.

### DIFF
--- a/helm/qontract-reconcile/values-external.yaml
+++ b/helm/qontract-reconcile/values-external.yaml
@@ -55,10 +55,10 @@ integrations:
   sleepDurationSecs: 300
   resources:
     requests:
-      memory: 150Mi
+      memory: 970Mi
       cpu: 200m
     limits:
-      memory: 200Mi
+      memory: 1Gi
       cpu: 300m
 - name: quay-mirror-org
   resources:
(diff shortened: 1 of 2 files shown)


helm/qontract-reconcile/values-external.yaml
```diff
@@ -55,10 +55,10 @@ integrations:
   sleepDurationSecs: 300
   resources:
     requests:
-      memory: 150Mi
+      memory: 970Mi
```
rporres, line 58: nit: why not just guaranteeing 1Gb of memory for our cache? Do we ever delete things from it?

rporres (approved):
A small nit, looks good otherwise

helm/qontract-reconcile/values-external.yaml
Piojo: We don't delete from the cache. It's what my initial calculation gave me. And I'm pretty sure I'll end up reducing this allocation soon.

rporres: sure, I meant why making requests and limits different... but it's only 30Mb, no biggie 😂

PR #1819, opened 2021-09-08
[qontract-cli] properly sort version history

maorfr:
to properly sort the version, we need the sorting to happen according to semver.
in this PR, we change the `version` field to be semver sortable. to properly print, we need to stringify the values beforehand.


patjlm:
LGTM

PR #1820, opened 2021-09-08
get_dependencies() refactor

bkez322:
This is a small refactor of terraform-dependency-related code in `terrascript_client.py`. Functionally, nothing is changed. This PR is only intended to improve readability and consistency.

Most significantly, the function `get_dependencies()` is renamed to `get_names_from_tf_resources()`. The intention here is to more accurately reflect what the function is doing.

In the current code, the `populate_tf_resource_rds()` function involves an array 'deps' which contains names of terraform resources, and the `populate_tf_resource_s3()` function involves an array 'deps' which contains objects representing terraform objects. With this PR, both functions now consistently only store the names and not full objects.

Some things had to move around slightly awkwardly on account of the 79 character limit.

Running locally, I used this code to successfully provision and deprovision real RDS terraform resources.


maorfr (approved):
LGTM

bkez322:
It looks like this PR caused some issues for QR in our staging environment.

I'm unsure still why this didn't cause problems in my local testing, but I believe I identified the issue.

I mistakenly thought [this function](https://github.com/bkez322/qontract-reconcile/blob/f91d107f061915c3d9a5745aacd6535d80d97308/reconcile/utils/terrascript_client.py#L1130) was returning terraform-resource objects, but in fact it is returning [dictionary wrappers](https://github.com/bkez322/qontract-reconcile/blob/f91d107f061915c3d9a5745aacd6535d80d97308/reconcile/utils/terrascript_client.py#L783-L784) around these terraform-resource objects.

The PR code was then erroneously [trying to read](https://github.com/app-sre/qontract-reconcile/pull/1820/files#diff-2e9fe6fa68837df3e3628e8acea384a1427c2a2b4c5b45a1a96135ef3038eec9R2945-R2948) non-existent properties from these dictionary wrappers.

Oops.

PR #1823, opened 2021-09-09
[qontract-cli] get root-owner

maorfr:
this command finds the top-most owner of an object (usually a Pod).
for example, the root owner of a Pod may be a ClowdApp!


tools/qontract_cli.py
```diff
@@ -576,6 +577,36 @@ def quay_mirrors(ctx):
...
+                    thread_pool_size=1,
+                    settings=settings,
+                    init_api_resources=True)
+    oc = oc_map.get(cluster)
+    obj = oc.get(namespace, kind, name)
+    root_owner = oc.get_obj_root_owner(namespace, obj, allow_not_found=True,
+                                       allow_not_controller=True)
+
+    # TODO(mafriedm): fix this
+    # do not sort
```
Piojo, line 601: Can you clarify why you don't want this sorted?

tools/qontract_cli.py
```diff
@@ -576,6 +577,36 @@ def quay_mirrors(ctx):
...
+    ctx.obj['options']['sort'] = False
+    # a bit hacky, but ¯\_(ツ)_/¯
```
Piojo, line 603: Fine by me!

reconcile/utils/oc.py
```diff
@@ -591,21 +592,23 @@ def get_obj_root_owner(self, ns, obj, allow_not_found=False):
             ns (string): namespace of the object
             obj (dict): representation of the object
             allow_not_found (bool, optional): allow owner to be not found
+            allow_not_controller (bool, optional): allow non-controller owner
```
Piojo, line 595: This would be a good place to experiment with type hints. ;)

Piojo (approved):
LGTM. Thanks for the test.

reconcile/utils/oc.py
maorfr: i'm not there yet :)

tools/qontract_cli.py
maorfr: sorting is a bit buggy and will not work with just any object

PR #1824, opened 2021-09-09
Upgrade Slack SDK from v1 to v3

steveteahan:
slackclient was deprecated, so it made sense to get to a supported
version of the SDK. In addition, we get these other benefits:

  1. Exceptions are now raised on failure, rather than needing to
     manually parse the response dict.
  2. Access to higher-level methods (ex. chat_postMessage) instead
     of just api_call(). This makes it easier to validate that the
     correct args have been passed in.
  3. Built-in retry handling for throttled requests. In addition,
     a simple mechanism for adding custom retry handlers (which was
     done for 5xx responses).


steveteahan:
app-sre specific testing documented in: https://issues.redhat.com/browse/APPSRE-3625

reconcile/test/test_utils_slack_api.py
```diff
@@ -1,19 +1,25 @@
...
+from reconcile.utils.slack_api import SlackApi, MAX_RETRIES, \
+    UserNotFoundException
 
 
 @pytest.fixture
 def slack_api():
     with patch('reconcile.utils.slack_api.SecretReader', autospec=True) as \
             mock_secret_reader, \
-            patch('reconcile.utils.slack_api.SlackClient', autospec=True) as \
+            patch('reconcile.utils.slack_api.WebClient', autospec=True) as \
```
Piojo, line 17: I tend to prefer `patch.object`
```suggestion
            patch.object(reconcile.utils.slack_api, 'WebClient', autospec=True) as \
```
but no biggie.

reconcile/utils/slack_api.py
```diff
@@ -16,36 +20,84 @@ class UsergroupNotFoundException(Exception):
...
+        :param settings: settings to pass to SecretReader
+        :param init_usergroups: whether or not to get a list of all Slack
+        usergroups when instantiated
+        :param channel: the Slack channel to post messages to
+        :param chat_kwargs: any other kwargs that can be used to post Slack
+        channel messages
+        """
         self.workspace_name = workspace_name
+
         secret_reader = SecretReader(settings=settings)
```
Piojo, line 58: This one tears me apart. Is this the right moment to move all side effects from the constructor and use this API in a context manager?

reconcile/utils/slack_api.py
```diff
@@ -16,36 +20,84 @@ class UsergroupNotFoundException(Exception):
...
 
-    def __init__(self, workspace_name, token,
-                 settings=None,
+    def __init__(self,
+                 workspace_name: str,
+                 token: Mapping[str, str],
+                 settings: Optional[Mapping[str, Any]] = None,
                  init_usergroups=True,
-                 **chat_kwargs):
+                 channel: Optional[str] = None,
```
Piojo, line 44: Why is this one optional?

Piojo (approved):
I don't know the Slack SDK enough to make a decision, but if nobody objects...

reconcile/utils/slack_api.py
steveteahan: What are you thinking with respect to the context manager?

I think the issue of `SecretReader` and more generally creating resources in `__init__` is more widespread and might be best handled in a more general way.  It isn't clear to me that it's much better to have `SecretReader` created by `__init__` indirectly (by calling another method) than directly.

Instead of creating 100 SecretReaders, one for each class, and dealing with mocks everywhere, we could handle this in a settings module or some equivalent that moves in the direction of dependency injection.  A dummy example:

```python
"""settings.py"""

class Settings:

@property
def slack_token() -> str:
    # Ignoring what __init__ of this would look like for now, assume there
    # is a SecretReader created somehow
    self._secret_reader.read(SLACK_SECRET_LOCATION)

settings = Settings()



"""slack_usergroups.py (or anywhere else SlackApi is used)"""

from reconcile.settings import settings

slack_api = SlackApi('workspace', settings.slack_token)
```

The SlackApi class no longer needs "settings" directly passed into it, nor does it need a `SecretReader` object at all.  I think this would really lead to cleaner and more testable code.  I'm willing to put together a quick PR in the next few days to show how we could do this to clean up the `__init__` in a few classes.  It won't solve every issues in the constructors, but it could be a move in the right direction.

reconcile/test/test_utils_slack_api.py
steveteahan: I've thought about this for a while and can't think of any immediate downsides.  I like typing fewer strings.  I'll change this and replace the context manager usage.

reconcile/utils/slack_api.py
steveteahan: I'll update docstring and we can address again if still unclear.

reconcile/utils/slack_api.py
steveteahan: Spoke to @Piojo about this offline.  He might also respond, but he agreed that passing the token in directly would be better.  I'll keep this as a separate task for now to avoid delaying some of the benefits that we'll get from this upgrade.

PR #1825, opened 2021-09-09
get_dependencies() refactor (v2)

bkez322:
This is a second attempt at refactoring some of the code in `terrascript_client.py`.

The code changes are almost exactly the same as [this PR](https://github.com/app-sre/qontract-reconcile/pull/1820), except with the complete removal of 3 changes around where the `_find_resource_()` function is used. I had misunderstood the data structures being returned from this function, and I now realize changing these spots of code will involve more work than I originally thought. 

These spots were changed in the first PR attempt, but are now unmodified:
* [spot 1](https://github.com/app-sre/qontract-reconcile/blob/8ae151173048fcca8e90271f40ec06c6898a0e00/reconcile/utils/terrascript_client.py#L982-L984)
* [spot 2](https://github.com/app-sre/qontract-reconcile/blob/8ae151173048fcca8e90271f40ec06c6898a0e00/reconcile/utils/terrascript_client.py#L1041-L1046)
* [spot 3](https://github.com/app-sre/qontract-reconcile/blob/8ae151173048fcca8e90271f40ec06c6898a0e00/reconcile/utils/terrascript_client.py#L1782-L1789)

There are also now mypy-compatible type hints [here](https://github.com/app-sre/qontract-reconcile/blob/e01cd2e7c83630a5a46ba566f4baba67e45c244e/reconcile/utils/terrascript_client.py#L1132-L1136), [here](https://github.com/app-sre/qontract-reconcile/blob/e01cd2e7c83630a5a46ba566f4baba67e45c244e/reconcile/utils/terrascript_client.py#L2948), and [here](https://github.com/app-sre/qontract-reconcile/blob/e01cd2e7c83630a5a46ba566f4baba67e45c244e/reconcile/utils/terrascript_client.py#L2953).

These alterations are in response to [this problem](https://github.com/app-sre/qontract-reconcile/pull/1820#issuecomment-916400265) with the first refactor attempt.


Piojo:
While this looks acceptable to me, is there any meaningful test we can add to detect the bug that caused the previous regression? Or is it just a matter of adding type hints?

bkez322:
> While this looks acceptable to me, is there any meaningful test we can add to detect the bug that caused the previous regression? Or is it just a matter of adding type hints?

I was planning to try adding type hints to this PR before removing the "draft" state.

reconcile/utils/terrascript_client.py
```diff
@@ -2900,10 +2945,14 @@ def unpack_namespace_info(namespace_info):
         return cluster, namespace
 
     @staticmethod
-    def get_dependencies(tf_resources):
+    def get_names_from_tf_resources(tf_resources: List[Resource]) -> List[str]:
```
steveteahan, line 2948: Minor point here, but it's a good practice to use more generic types for parameters if possible.  In this case, you should be able to use `Iterable[Resource]` which should work given that you're only looping through the results.

This is covered on the third bullet point [here](https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#type-hints).

The abstract types that are preferred can be found [here](https://docs.python.org/3/library/typing.html#typing.List).  Likewise, `(Mutable)Mapping` is preferred to `Dict` (again mostly for inputs).

Other than that the type usage looks good!

bkez322: Now using Iterable.

https://github.com/bkez322/qontract-reconcile/commit/55e6f94f04bcf98d3b2985a02cb6269c8b3666a7

Piojo (approved):

PR #1826, opened 2021-09-10
Update tox configs and test dependencies

steveteahan:
* Separate linting and test tox environments so that they can be run
  independently (ex. tox -e lint or tox -e py36). This replaces the
  separate tox-devel.ini config.
* Move testing dependencies to requirements-dev.txt to make it easier
  to install them outside of tox environments
* Update Pylint to a later version that has better performance
  (finishes in 1/3 the time)
* Fix coverage data that was previously including the test/ directory
  due to an invalid configuration. This required reducing the minimum
  coverage configuration.
* Remove dependency on mock because it's now part of the standard
  library

Before this is merged, I'll notify the team that `tox-devel.ini` is being removed, as well as how to replace it.  With this change, the tests go from 130 to 90 seconds.  Using `tox -p` on my laptop, it only takes 55 seconds.

This change also prepares qontract-reconcile to add a new Mypy environment.  This will make it run type checking, linting, or unit tests without having to run them all at once.


requirements-dev.txt
```diff
@@ -0,0 +1,8 @@
+anymarkup==0.7.0
+flake8==3.5.0
+kubernetes~=12.0
+pylint~=2.6
+pytest~=6.2
+pytest-cov~=2.12
+responses
+testslide==2.6.3
```
Piojo, line 8: Feel free to make the testslide dependency of type `~=` too.

reconcile/utils/terrascript_client.py
```diff
@@ -173,6 +173,7 @@ def download_logtoes_zip(self, release_url):
         if not os.path.exists(zip_file):
             r = requests.get(zip_url)
             r.raise_for_status()
+            # pylint: disable=consider-using-with
```
Piojo, line 176: Can you file a task to handle these consider-using-with cases? It's an excellent first issue for a newcomer.

Piojo (approved):

requirements-dev.txt
steveteahan: All set.

reconcile/utils/terrascript_client.py
steveteahan: https://issues.redhat.com/browse/APPSRE-3686

PR #1827, opened 2021-09-10
[terraform-resources] add support to provision aws load balancer

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3695

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/25145

this PR adds support to provision LBs using terraform: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb

this PR initially adds support for Application LB only, and only for HTTPS (with an HTTP redirection). we can refactor when future use cases come along.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,4 +3,4 @@ branch = True
 omit = reconcile/test/*
 
 [report]
-fail_under = 23.1
+fail_under = 23.0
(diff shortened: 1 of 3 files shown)


patjlm:
Any thoughts on using the official aws load balancer controller https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/ ?
Or do we plan to use this for non-kube workload ?

maorfr:
nothing against it. it just doesn't support the rest of the terraform providers (s3, etc).

PR #1828, opened 2021-09-10
Enable Mypy type checker during CI builds

steveteahan:
Type hints in code will be optional for now, but this change
will make Mypy check any methods that are using type hints.
This will enable team members to begin adding type hints to
their code. The requirements files were also separated between
type and test dependencies. The dev file will install both
test and type dependencies for convenience.


Piojo (approved):

rporres:
I come a bit late to the party, but this looks great. Thx also for the detailed README :)

PR #1829, opened 2021-09-10
Reduce the memory allocations for quay mirror

Piojo:
This pod is using about 420MB, let's give it 550 for good measure.

### DIFF
--- a/helm/qontract-reconcile/values-external.yaml
+++ b/helm/qontract-reconcile/values-external.yaml
@@ -55,10 +55,10 @@ integrations:
   sleepDurationSecs: 300
   resources:
     requests:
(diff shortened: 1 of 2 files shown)


rporres:
Why do you need different requests and limits for memory? Do you expect any kind of memory bursts in this integration. I assume they are not going to be, since we don't delete from cache...

Piojo:
Nits are most appreciated. To be honest, I don't expect any bursts here, and I'm leaving a good 20% of leeway to our memory allocation. I was following the previous standard where requests and limits were different, but that doesn't matter in this context. Let me fix it.

rporres (approved):

PR #1836, opened 2021-09-13
Replace easy_install usage with the pip equivalents

steveteahan:
easy_install (used with `setup.py install`) will install pre-releases
which is unexpected and undesirable behavior.

More information:

https://github.com/pypa/setuptools/issues/855

I tested this by creating a venv with `python setup.py install` and another with `pip install .`  The diff of the `pip freeze` output was:

```
$ diff easy_install.txt pip.txt 
29c29
< jira==2.0.1.0rc1
---
> jira==2.0.0
56c56
< python-jenkins==1.8.0.0a0
---
> python-jenkins==1.7.0
86c86
< wrapt==1.13.0rc3
---
> wrapt==1.12.1

```

This is exactly what I expect.  The `pip` install ignored alpha and release candidates while easy_install was happy to install versions that we don't want to rely on in production.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,16 +126,17 @@ python3 -m venv venv
 source venv/bin/activate
 
 # make sure you are running the latest setuptools
-python3 -m pip install --upgrade pip setuptools
+pip install --upgrade pip setuptools
 ```
 
 Install the package:
 
 ```sh
-python3 setup.py install
+pip install .
 
-# or alternatively use this for a devel environment
-python3 setup.py develop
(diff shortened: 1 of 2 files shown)


steveteahan:
I also built the container before and after with `dockerfiles/Dockerfile` and confirmed that pre-release packages were being installed for jira, python-jenkins, and wrapt.

Piojo (approved):
LGTM.

rporres (approved):

jfchevrette (approved):

jmelis (approved):
LGTM

PR #1837, opened 2021-09-13
Fix linters on Python 3.9

Piojo:
Use a version of flake8 that doesn't crash, fix any new errors, and
make pylint ignore an instantiation that has always worked.


jfchevrette:
Note our production image runs on UBI 8.2 which provides python 3.6

Piojo:
Please note that these changes will work on both 3.6 and 3.9, and we have [3.9](https://issues.redhat.com/browse/APPSRE-3634) scheduled to happen in the next sprint.

esron (approved):

PR #1840, opened 2021-09-14
[terraform-resources] only apply if apply is needed

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3702

We are seeing this issue multiple times in pr-checks:
```
Error: Saved plan is stale

The given plan file can no longer be applied because the state was changed by
another operation after the plan was created.
```
This is probably because the actual terraform integration(s) are running continuously and performing `terraform apply` even without anything to change.
Let's try to avoid `terraform apply` if no changes are going to be made. This should alleviate this error and cause it to appear much less (and indirectly, speeding up app-interface pr-check time)

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -422,7 +422,7 @@ def run(dry_run, print_only=False,
     if dry_run:
         cleanup_and_exit(tf)
 
-    if not light:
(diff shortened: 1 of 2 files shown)


patjlm (approved):
LGTM

rporres (approved):

PR #1842, opened 2021-09-14
[oc] retry recycle

maorfr:
related to https://issues.redhat.com/browse/ASIC-86

in some cases, a `Deployment` was applied along (and before) a `ConfigMap`. we may be trying to apply an outdated `Deployment`, so let's get it before we act on it and retry for this particular case:
```
Operation cannot be fulfilled on deployments.apps "<deployment_name>": the object has been modified; please apply your changes to the latest version and try again
```

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -585,11 +585,10 @@ class OCDeprecated:
     def recycle(self, dry_run, namespace, kind, obj):
         """Recycles an object by adding a recycle.time annotation
 
-        Args:
-            dry_run (bool): Is this a dry run
-            namespace (string): Namespace to work in
-            kind (string): Object kind
-            obj (dict): Object to recycle
+        :param dry_run: Is this a dry run
+        :param namespace: Namespace to work in
+        :param kind: Object kind
+        :param obj: Object to recycle
         """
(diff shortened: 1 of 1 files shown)


reconcile/utils/oc.py
```diff
@@ -575,22 +579,36 @@ def recycle_pods(self, dry_run, namespace, dep_kind, dep_resource):
...
+    @retry(exceptions=ObjectHasBeenModifiedError)
+    def recycle(self, dry_run, namespace, kind, obj):
+        """Recycles an object by adding a recycle.time annotation
+
+        Args:
+            dry_run (bool): Is this a dry run
+            namespace (string): Namespace to work in
+            kind (string): Object kind
+            obj (dict): Object to recycle
+        """
```
andreugallofre, line 593: NIT: (And not sure if we agreed that we can use both google and sphinx) But the docstring is not following the guidelines on the dev guide https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#docstrings

andreugallofre: Besides of that lgtm

BumbleFeng (approved):

PR #1843, opened 2021-09-14
Minor Refactor: Explicit enhanced monitoring dependencies

bkez322:
Functionally, in my testing, nothing here is changed. 

Terraform is supposed to (although is not necessarily guaranteed to) be able to automatically resolve the dependency order of these resources on its own: A role, A role-policy-attachment, and a RDS-instance.

This is a "just in case for the future" refactor that explicitly tells terraform to provision the enhanced-monitoring-role and enhanced-monitoring-policy-attachment before attempting to provision an enhanced-monitoring RDS instance.

Note: This PR will not fix [APPSRE-3511](https://issues.redhat.com/browse/APPSRE-3511).


esron (approved):
Whey pushing this to master please try to change the commit description with just the modifications.

reconcile/utils/terrascript_client.py
```diff
@@ -933,21 +933,26 @@ def populate_tf_resource_rds(self, resource, namespace_info,
             }
             role_tf_resource = aws_iam_role(em_identifier, **em_values)
             tf_resources.append(role_tf_resource)
+            values['monitoring_role_arn'] = \
+                "${" + role_tf_resource.arn + "}"
```
esron, line 937: Since you are here, can you change this into `f'` strings?

esron (changes requested):

reconcile/utils/terrascript_client.py
bkez322: Done.
https://github.com/app-sre/qontract-reconcile/blob/98ef424662e180f647c1bcae93db8178184d600b/reconcile/utils/terrascript_client.py#L946

esron (approved):

PR #1844, opened 2021-09-14
only init GitLabApi when need clean_pipelines

BumbleFeng:


reconcile/gitlab_housekeeping.py
```diff
@@ -37,13 +41,26 @@ def clean_pipelines(dry_run, gl_project, pipelines, pipeline_timeout=60):
...
+                                  settings=gl_settings).project.pipelines
+
+    for p in pipelines:
+        logging.info(['canceling', p['web_url']])
+        if not dry_run:
+            try:
+                gl_piplelines.get(p['id']).cancel()
+            except gitlab.exceptions.GitlabPipelineCancelError as err:
+                logging.error('unable to cancel {} - error message {}'
+                              .format(p['web_url'], err.error_message))
```
Piojo, line 63: Please, use f-strings
```suggestion
                logging.error(f'unable to cancel {p["web_url"]} - "
                       f"error message {err.error_message}')
```

reconcile/gitlab_housekeeping.py
```diff
@@ -150,11 +167,12 @@ def rebase_merge_requests(dry_run, gl, rebase_limit, pipeline_timeout=None,
...
             # If pipeline_timeout is None no pipeline will be canceled
             if pipeline_timeout is not None:
-                gl_project = GitLabApi(gl_instance,
-                                       project_id=mr.source_project_id,
-                                       settings=gl_settings).project
-                clean_pipelines(dry_run, gl_project, pipelines,
-                                pipeline_timeout)
+                timeout_pipelines = \
+                    get_timeout_pipelines(pipelines, pipeline_timeout)
+                if timeout_pipelines:
```
Piojo, line 172: You don't need this `if`. The inner loop won't do anything if the list is empty.

reconcile/gitlab_housekeeping.py
```diff
@@ -220,11 +238,12 @@ def merge_merge_requests(dry_run, gl, merge_limit, rebase,
...
```
Piojo, line 243: Ditto.

reconcile/test/test_gitlab_housekeeping.py
```diff
@@ -52,14 +65,18 @@ def test_clean_pipelines_happy_path():
...
 
-        gl_h.clean_pipelines(dry_run, gl_project_mock, pipelines, timeout)
+        timeout_pipelines = gl_h.get_timeout_pipelines(pipelines, timeout)
+        gl_h.clean_pipelines(dry_run, instance, '1', '', timeout_pipelines)
 
         # Test if mock have this exact calls
-        kall = call(47).cancel()
-        assert gl_project_mock.pipelines.get.mock_calls == kall.call_list()
+        kall = call('/projects/1/pipelines/47/cancel')
+        assert http_post.mock_calls == kall.call_list()
```
Piojo, line 82: Better:
```suggestion
        http_post.assert_called_once_with('projects/1/pipelines/47/cancel')
```

reconcile/gitlab_housekeeping.py
```diff
@@ -37,13 +41,26 @@ def clean_pipelines(dry_run, gl_project, pipelines, pipeline_timeout=60):
...
+    if not dry_run:
+        gl_piplelines = GitLabApi(gl_instance,
+                                  project_id=gl_project_id,
+                                  settings=gl_settings).project.pipelines
+
+    for p in pipelines:
+        logging.info(['canceling', p['web_url']])
+        if not dry_run:
+            try:
+                gl_piplelines.get(p['id']).cancel()
```
Piojo, line 60: What happens if the pipeline finishes or crashes in between the call to `get` and the cancel operation? Is it still going to raise a `GitlabPipelineCancelError`?

Piojo: Also, I wonder if it wouldn't be easier to just have one `dry_run` path. Say:

```python
if dry_run:
    logging.info(["cancelling", [p['web_url'] for p in pipeilnes])
    return

for p in pipelines:
   try:
     ...
...
```

reconcile/gitlab_housekeeping.py
```diff
@@ -21,14 +22,17 @@
 DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
 
 
-def clean_pipelines(dry_run, gl_project, pipelines, pipeline_timeout=60):
+def get_timeout_pipelines(pipelines: List[Any],
```
Piojo, line 25: Your pipelines look like `Dict` to me.

Piojo: Also, naming nit, this sounds like you're getting the time out parameter for the pipelines instead of the ones that have timed out.
```suggestion
def get_timed_out_pipelines(pipelines: List[Dict],
```

reconcile/test/test_gitlab_housekeeping.py
```diff
@@ -1,17 +1,30 @@
...
+    data = fixture.get('gitlab').get(path)
+    return data
+
 
 class TestGitLabHousekeeping:
     @staticmethod
-    def test_clean_pipelines_happy_path():
+    @patch.object(SecretReader, 'read')
+    @patch.object(Gitlab, 'http_get')
+    @patch.object(Gitlab, 'http_post')
```
Piojo, line 25: ... this is why I am using testslide. Let's leave it alone for now, though.

reconcile/gitlab_housekeeping.py
esron: Won't this prevent logging when not in dry_run?

BumbleFeng: > What happens if the pipeline finishes or crashes in between the call to `get` and the cancel operation? Is it still going to raise a `GitlabPipelineCancelError`?

No, it will return 200, but pipeline will not actually have any change.

reconcile/gitlab_housekeeping.py
BumbleFeng: I want to avoid init GitLabApi if timeout_pipelines is empty.

Piojo (approved):
Thanks for the explanations. This looks good to go.

PR #1848, opened 2021-09-14
Fix ls method in State util class

esron:
JIRA: [APPSRE-3703](https://issues.redhat.com/browse/APPSRE-3703)

This adds two more development dependencies, [python-mock](https://pypi.org/project/pytest-mock/) and [moto](https://pypi.org/project/moto/).

### DIFF
--- a/reconcile/test/test_utils_state.py
+++ b/reconcile/test/test_utils_state.py
@@ -48,3 +48,18 @@ def test_ls_returns_correct_file(accounts, s3_client, mocker):
     expected = ['/some-file-1']
 
     assert keys == expected
+
+
+def test_ls_when_state_is_empty(accounts, s3_client, mocker):
+    s3_client.create_bucket(Bucket='some-bucket')
+
+    mock_aws_api = mocker.patch('reconcile.utils.state.AWSApi', autospec=True)
+    mock_aws_api.return_value \
+        .get_session.return_value \
+        .client.return_value = s3_client
+
+    state = State('integration-name', accounts)
+
+    keys = state.ls()
+
+    assert keys == []
--- a/reconcile/test/test_utils_state.py
+++ b/reconcile/test/test_utils_state.py
@@ -10,8 +10,8 @@ def accounts():
     return [{'name': 'some-account'}]
 
 
-@mock_s3
-def test_ls(accounts, monkeypatch, mocker):
+@pytest.fixture
+def s3_client(monkeypatch):
     monkeypatch.setenv('AWS_ACCESS_KEY_ID', 'testing')
     monkeypatch.setenv('AWS_SECRET_ACCESS_KEY', 'testing')
     monkeypatch.setenv('AWS_SECURITY_TOKEN', 'testing')
@@ -19,9 +19,13 @@ def test_ls(accounts, monkeypatch, mocker):
     monkeypatch.setenv('APP_INTERFACE_STATE_BUCKET', 'some-bucket')
     monkeypatch.setenv('APP_INTERFACE_STATE_BUCKET_ACCOUNT', 'some-account')
 
-    s3_client = boto3.client('s3', region_name='us-east-1')
-    s3_client.create_bucket(Bucket='some-bucket')
+    with mock_s3():
+        s3_client = boto3.client('s3', region_name='us-east-1')
+        yield s3_client
 
+
+def test_ls_returns_correct_file(accounts, s3_client, mocker):
+    s3_client.create_bucket(Bucket='some-bucket')
     s3_client.put_object(Bucket='some-bucket',
                          Key='state/integration-name/some-file-1',
                          Body='test')
--- a/reconcile/test/test_utils_state.py
+++ b/reconcile/test/test_utils_state.py
@@ -1,0 +1,46 @@
+import pytest
+import boto3
+from moto import mock_s3
+from reconcile.utils.state import State
+
+
+@pytest.fixture
+def accounts():
+    """Account name is needed to instantiate a State"""
+    return [{'name': 'some-account'}]
+
+
+@mock_s3
+def test_ls(accounts, monkeypatch, mocker):
+    monkeypatch.setenv('AWS_ACCESS_KEY_ID', 'testing')
+    monkeypatch.setenv('AWS_SECRET_ACCESS_KEY', 'testing')
+    monkeypatch.setenv('AWS_SECURITY_TOKEN', 'testing')
+    monkeypatch.setenv('AWS_SESSION_TOKEN', 'testing')
+    monkeypatch.setenv('APP_INTERFACE_STATE_BUCKET', 'some-bucket')
+    monkeypatch.setenv('APP_INTERFACE_STATE_BUCKET_ACCOUNT', 'some-account')
+
+    s3_client = boto3.client('s3', region_name='us-east-1')
+    s3_client.create_bucket(Bucket='some-bucket')
+
+    s3_client.put_object(Bucket='some-bucket',
+                         Key='state/integration-name/some-file-1',
+                         Body='test')
+
+    # Creating some-file-2 to identify when two or more integrations have
+    # similar names
+    s3_client.put_object(Bucket='some-bucket',
+                         Key='state/integration-name-2/some-file-2',
+                         Body='test')
+
+    mock_aws_api = mocker.patch('reconcile.utils.state.AWSApi', autospec=True)
+    mock_aws_api.return_value \
+        .get_session.return_value \
+        .client.return_value = s3_client
+
+    state = State('integration-name', accounts)
+
+    keys = state.ls()
+
+    expected = ['/some-file-1']
+
+    assert keys == expected
--- a/reconcile/test/test_utils_state.py
+++ b/reconcile/test/test_utils_state.py
@@ -50,6 +50,36 @@ def test_ls_returns_correct_file(accounts, s3_client, mocker):
     assert keys == expected
 
 
+def test_ls_when_integration_is_empty_string(accounts, s3_client, mocker):
+    s3_client.create_bucket(Bucket='some-bucket')
+    s3_client.put_object(Bucket='some-bucket',
+                         Key='state/integration-name-1/some-file-1',
+                         Body='test')
+    s3_client.put_object(Bucket='some-bucket',
+                         Key='state/integration-name-2/some-file-2',
+                         Body='test')
+    s3_client.put_object(Bucket='some-bucket',
+                         Key='state/integration-name-3/nested/some-file-2',
+                         Body='test')
+
+    mock_aws_api = mocker.patch('reconcile.utils.state.AWSApi', autospec=True)
+    mock_aws_api.return_value \
+        .get_session.return_value \
+        .client.return_value = s3_client
+
+    state = State('', accounts)
+
+    keys = state.ls()
+
+    expected = [
+        '/integration-name-1/some-file-1',
+        '/integration-name-2/some-file-2',
+        '/integration-name-3/nested/some-file-2',
+    ]
+
+    assert keys == expected
+
+
 def test_ls_when_state_is_empty(accounts, s3_client, mocker):
     s3_client.create_bucket(Bucket='some-bucket')
 
--- a/reconcile/utils/state.py
+++ b/reconcile/utils/state.py
@@ -1,6 +1,8 @@
 import os
 import json
 
+from typing import Any, Iterable, Mapping, Optional
+
 from botocore.errorfactory import ClientError
 
 from reconcile.utils.aws_api import AWSApi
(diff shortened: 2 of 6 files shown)


reconcile/test/test_utils_state.py
```diff
@@ -0,0 +1,65 @@
...
+@pytest.fixture
+def accounts():
+    """Account name is needed to instantiate a State"""
+    return [{'name': 'some-account'}]
+
+
+@pytest.fixture
+def aws_credentials():
+    """Mocked AWS Credentials for moto."""
+    os.environ['AWS_ACCESS_KEY_ID'] = 'testing'
```
steveteahan, line 19: I don't think these direct `os.environ` modifications are working as you expect they would.  Consider this:

```
def test_set_os_environ():
    os.environ['SOME_VAR'] = 'this-is-set'
    assert os.environ['SOME_VAR'] == 'this-is-set'

def test_get_os_environ():
    assert os.environ['SOME_VAR'] == 'this-is-set'
```

These environment variables persist beyond the test your're setting them in.  You might consider `patch.dict` and patching `os.environ` directly instead.  I'll only comment on this once, but this applies to all usage of `os.environ` in this PR.

reconcile/test/test_utils_state.py
```diff
@@ -0,0 +1,65 @@
...
+    os.environ['APP_INTERFACE_STATE_BUCKET_ACCOUNT'] = 'some-account'
+    s3_client = boto3.client('s3', region_name='us-east-1')
+    s3_client.create_bucket(Bucket='some-bucket')
+
+    file_text = 'test'
+    with NamedTemporaryFile(delete=True, suffix='.txt') as tmp:
+        with open(tmp.name, 'w', encoding='UTF-8') as f:
+            f.write(file_text)
+
+        s3_client.upload_file(tmp.name, 'some-bucket',
```
steveteahan, line 37: You might consider `put_object` instead of `upload_file` to avoid needing to deal with the filesystem at all.  You'll see this is also how `__setitem__` is implemented for this class.

reconcile/test/test_utils_state.py
```diff
@@ -0,0 +1,65 @@
...
+
+        @staticmethod
+        def get_session(account):
+            class Session:
+                @staticmethod
+                def client(type):
+                    return s3_client
+            return Session()
+
+    mocker.patch('reconcile.utils.state.AWSApi', MockAWSAPi)
```
steveteahan, line 57: What is your thinking about taking this approach with creating `MockAWSAPi` versus using an [autospec mock](https://docs.python.org/3/library/unittest.mock.html#auto-speccing)?

reconcile/utils/state.py
```diff
@@ -54,7 +54,7 @@ def ls(self):
         Returns a list of keys in the state
         """
         objects = self.client.list_objects(Bucket=self.bucket,
-                                           Prefix=self.state_path)
+                                           Prefix=f'{self.state_path}/')
```
steveteahan, line 57: Is there any chance that parts of the code base rely on this less strict form as it's being used today?  In theory, they shouldn't be pulling state from other integrations, but I think it's worth verifying.

requirements-test.txt
```diff
@@ -4,5 +4,7 @@ kubernetes~=12.0
 pylint~=2.6
 pytest~=6.2
 pytest-cov~=2.12
+pytest-mock~=3.6
```
steveteahan, line 7: Good addition - reminded me to use this in one of my other PRs.

setup.cfg
```diff
@@ -100,3 +100,6 @@ ignore_missing_imports = True
 ; Supported with update
 [mypy-urllib3.*]
 ignore_missing_imports = True
+
+[mypy-moto.*]
```
steveteahan, line 104: nit: can we keep this alphabetical? 

Also `Supported with update` means that a newer versions includes type hints, which I don't think is the case here?

reconcile/test/test_utils_state.py
```diff
@@ -0,0 +1,65 @@
...
+def aws_credentials():
+    """Mocked AWS Credentials for moto."""
+    os.environ['AWS_ACCESS_KEY_ID'] = 'testing'
+    os.environ['AWS_SECRET_ACCESS_KEY'] = 'testing'
+    os.environ['AWS_SECURITY_TOKEN'] = 'testing'
+    os.environ['AWS_SESSION_TOKEN'] = 'testing'
+
+
+@mock_s3
+def test_ls(accounts, aws_credentials, mocker):
```
steveteahan, line 26: While you're in here, worth adding one more quick test for where the contents can be empty?

steveteahan (changes requested):
Good find and nice job fixing it!

reconcile/test/test_utils_state.py
esron: It was the first in the first examples for a google search for "pytest mock" 😅

steveteahan: That's fair.  If you wanted to do this with Python mocks it would look something like this:

```python
mock_session = create_autospec(Session)

# This could also be mocker.patch.object(reconcile.utils.state, 'AWSApi', autospec=True)
mock_aws_api = mocker.patch('reconcile.utils.state.AWSApi', autospec=True)

mock_aws_api.return_value.get_session.return_value = mock_session
```

The nice thing about autospec is that it returns an object that matches the API of a real object, without needing to write and maintain additional classes.  If `Session` or `AWSApi` changed in this case, we'd need to update any stubs that have been created.  The nice thing about a stub or strict mock (see testing guide about testslide) is that any undefined behavior will fail, whereas MagicMocks will happily return for any method access.  This also makes the tests more prone to break where a new method is accessed.

For this use case, either way is probably fine because you really aren't using any other features that a mock provides (checking behavior), but for other more complex scenarios, a mock could be more appropriate.

reconcile/utils/state.py
esron: Now that you mentioned, if someone instantiate a `State` passing `integration` as `None` or and empty string this can cause new issues.

Gonna come up with more tests.

steveteahan: It'd be good to check for that.  Additionally, you might consider moving the type hints from the docstring to inline type hints.  That would also improve the likelihood of catching a `None` passed to integration (although it also wouldn't hurt to check the value of `integration`.

reconcile/test/test_utils_state.py
esron: I've done some tests with [Monkeypatch](https://docs.pytest.org/en/6.2.x/monkeypatch.html#monkeypatching-environment-variables) and worked fine. Gona push it.

reconcile/test/test_utils_state.py
esron: Done.

setup.cfg
esron: Done.

reconcile/test/test_utils_state.py
```diff
@@ -0,0 +1,56 @@
...
+                    return s3_client
+            return Session()
+
+    mocker.patch('reconcile.utils.state.AWSApi', MockAWSAPi)
+
+    state = State('integration-name', accounts)
+
+    objects = state.ls()
+
+    expected = ['/some-file-1']
```
steveteahan, line 54: The keys having leading forward slashes is odd to me.  It seems that all the other methods take a bare key without the forward slash?  Also, `get_all` is removing that forward slash?  I'd expect to be able to do something like:

```
state_keys = state.ls()

for key in state_keys:
    state.get(key)
```

From quickly reading the code, it seems that you'd need to account for the slash to make that work.  I can see the couple integrations using `ls()` dealing with this, but wonder if this makes sense or whether it should be addressed in a separate PR?

Also, `qontract_cli.py` has some interesting usage of this:

```
    keys = state.ls()
    # if 'integration' is defined, the 0th token is empty
    table_content = [
        {'integration': k.split('/')[0] or integration,
         'key': '/'.join(k.split('/')[1:])}
        for k in keys]
```

Did you already come across that?

esron: > Did you already come across that?

Yes, that was why I commented [here](https://github.com/app-sre/qontract-reconcile/pull/1848#discussion_r709484431)

IMO a refactoring that has the potential to change that much files should be in another PR.

steveteahan: I agree and that's why I mentioned the separate PR.  This class doesn't seem to be widely used yet, so making a breaking change (returning without forward slash) would only require fixes in a couple of places, rather than later it could be many.  I think it's worth a tracking ticket in any case?

reconcile/test/test_utils_state.py
esron: Thanks for the suggestions, I've pushed a new version.

reconcile/test/test_utils_state.py
esron: Done.

reconcile/test/test_utils_state.py
```diff
@@ -0,0 +1,65 @@
...
+    monkeypatch.setenv('AWS_ACCESS_KEY_ID', 'testing')
+    monkeypatch.setenv('AWS_SECRET_ACCESS_KEY', 'testing')
+    monkeypatch.setenv('AWS_SECURITY_TOKEN', 'testing')
+    monkeypatch.setenv('AWS_SESSION_TOKEN', 'testing')
+    monkeypatch.setenv('APP_INTERFACE_STATE_BUCKET', 'some-bucket')
+    monkeypatch.setenv('APP_INTERFACE_STATE_BUCKET_ACCOUNT', 'some-account')
+
+    with mock_s3():
+        s3_client = boto3.client('s3', region_name='us-east-1')
+        yield s3_client
```
steveteahan, line 24: Any need to `yield` here versus just `return`?

reconcile/test/test_utils_state.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def test_ls_returns_correct_file(accounts, s3_client, mocker):
+    s3_client.create_bucket(Bucket='some-bucket')
+    s3_client.put_object(Bucket='some-bucket',
+                         Key='state/integration-name/some-file-1',
+                         Body='test')
+
+    """Creating some-file-2 to identify when two or more integrations have
```
steveteahan, line 33: nit: `"""` is generally only used for docstrings as far as I'm aware.  I wouldn't make a habit of doing this and would just use `#` instead.  If linters aren't catching this now, they might in the future.

reconcile/test/test_utils_state.py
esron: Yes, if I just `return` the `s3_client` turns back to the boto3 client and tries to access the real infrastructure.

steveteahan: Got it, so I assume that's because with the context manager returning would be out of scope of mock_s3.  Makes sense.

reconcile/test/test_utils_state.py
esron: Done.

reconcile/utils/state.py
esron: I've searched through the code and only found one occasion where integration can be `''`. It is in `qontract_cli.py` as mentioned in a previous comment.

reconcile/utils/state.py
```diff
@@ -18,12 +20,10 @@ class State:
...
     :param settings: App Interface settings
-
-    :type integration: string
-    :type accounts: list
-    :type settings: dict
     """
-    def __init__(self, integration, accounts, settings=None):
+
+    def __init__(self, integration: str, accounts: List[Dict],
+                 settings: Union[Dict, Optional[None]] = None) -> None:
```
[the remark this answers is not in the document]
esron: I think I'm gonna go with that:

```python
def __init__(self, integration: str, accounts: Iterable[Mapping[str, Any]],
             settings: Optional[Mapping[str, Any]] = None) -> None:
```

What you think?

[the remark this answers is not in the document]
steveteahan: If both mappings are going to be nested several levels deep and the type aren't predictable, the use of `Any` is reasonable enough to me.

reconcile/utils/state.py
steveteahan: The fix looks good to me, but I'm just wondering does the last point above mean that this will break some `qontract-cli` usage?  If so, what path would you suggest to move forward?

esron: To be honest, I don't know.

This method in `qontract-cli` is already "broken" because of types.

I fear that trying to fix that would make this PR too big.

Maybe I should try to fix `qontract-cli` first in another PR.

esron: In fact just did in #1855

esron: #1855 Has been merged. I'll see what does to this.

esron: I've update some things, including a new test to deal with `qontract-cli`

steveteahan (approved):

reconcile/test/test_utils_state.py
esron: Done: https://issues.redhat.com/browse/APPSRE-3781

PR #1855, opened 2021-09-17
Fix argument for print_output in qontract-cli

esron:
Due to recent changes on `print_output` function  two sub commands of `qontract-cli` are not working `qontract-cli state ls` and `qontract-cli template`

This fix it by sending an object as argument instead of a string.

I've choose to type content as `List[Dict]` Because `Iterable` doesn't let you use `.sort` in objects.

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -718,7 +718,8 @@ def ls(ctx, integration):
         {'integration': k.split('/')[0] or integration,
          'key': '/'.join(k.split('/')[1:])}
         for k in keys]
-    print_output('table', table_content, ['integration', 'key'])
+    print_output({'output': 'table', 'sort': False},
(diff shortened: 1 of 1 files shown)


tools/qontract_cli.py
```diff
@@ -626,7 +627,8 @@ def sre_checkpoints(ctx):
     print_output(ctx.obj['options'], checkpoints_data, columns)
 
 
-def print_output(options, content, columns=[]):
+def print_output(options: Mapping[str, Union[str, bool]],
+                 content: List[Dict], columns: Iterable[str] = []):
```
Piojo, line 631:
```suggestion
                 content: List[Dict], columns: Iterable[str] = ()):
```
don't use a mutable value as a default. Empty tuples usually work just as well.

Piojo (approved):
Push after fixing my nit.

PR #1857, opened 2021-09-19
WIP aws-iam-service-account handle supplied assume role

maorfr:
instead of supplying the spec for aws infrastructure access (OCM feature) we will be able to supply an ARN to assume (no OCM scenario).


patjlm (approved):
LGTM

PR #1858, opened 2021-09-19
[repo_owners] except ruamel parse error

maorfr:
will handle an issue such as https://gitlab.cee.redhat.com/mk-ci-cd/kas-upgrade-configurations/-/merge_requests/32


reconcile/utils/repo_owners.py
```diff
@@ -133,7 +133,10 @@ def _get_owners_map(self):
         for owner_file in owner_files:
             raw_owners = self._git_cli.get_file(path=owner_file['path'],
                                                 ref=self._ref)
-            owners = yaml.safe_load(raw_owners.decode())
+            try:
+                owners = yaml.safe_load(raw_owners.decode())
+            except yaml.parser.ParserError:
```
esron, line 138: Is there another before this one that guarantees that the error mentioned in this PR is the only one that can happens? If not, maybe we should check the error message.

reconcile/utils/repo_owners.py
```diff
@@ -133,7 +133,10 @@ def _get_owners_map(self):
...
+                owners = None
```
Piojo, line 139: Please log this one, it may surprise users.

Piojo (approved):

PR #1860, opened 2021-09-20
Add --skip-tf-provider-list flag

bkez322:
Adds flag `--skip-tf-provider-list` which accepts a comma-seperated list of terraform providers that qontract-reconcile should skip doing anything with.

Currently supports only the `terraform_resources` and `aws_iam_keys` integrations.

The primary motivation for this is to provide a more convenient way of dealing with [this line of code](https://github.com/app-sre/qontract-reconcile/blob/b3366922af4231074c748a832a0243e096eeeac2/reconcile/utils/terrascript_client.py#L156). A 'GitHub config' from graphQL is only involved with managing cloudwatch tf-resources. Previously, I had been commenting out that line of code when testing non-cloudwatch tf-resource integrations. It got annoying making sure I didn't accidentally commit that.

With this PR, I can locally run non-cloudwatch tf-resources integrations like this:

` qontract-reconcile --config $QR_DEV_CONFIG terraform-resources --skip-tf-provider-list=cloudwatch`

### DIFF
--- a/reconcile/aws_iam_keys.py
+++ b/reconcile/aws_iam_keys.py
@@ -61,7 +61,8 @@ def run(dry_run, thread_pool_size=10,
     settings = queries.get_app_interface_settings()
     aws = AWSApi(thread_pool_size, accounts, settings=settings)
     keys_to_delete = get_keys_to_delete(accounts)
-    working_dirs = init_tf_working_dirs(accounts, thread_pool_size, skip_tf_providers_list, settings)
+    working_dirs = init_tf_working_dirs(accounts, thread_pool_size,
+                                        skip_tf_providers_list, settings)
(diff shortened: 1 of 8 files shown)


maorfr:
the problem here is that we are hard-coding `app-sre`. can we find a way to avoid it? maybe setting something in the org file in app-interface to indicate this is the default org token to be used?

bkez322:
> the problem here is that we are hard-coding `app-sre`. can we find a way to avoid it? maybe setting something in the org file in app-interface to indicate this is the default org token to be used?
@maorfr 
That sounds like that may be a better way to solve the problem. That seems fundamentally different than the feature implemented by this PR, so I will close this PR.

PR #1866, opened 2021-09-22
terraform_aws_route53: safe_resource_id should sanitize * characters

jfchevrette:
Because the integration uses the record name to build the terraform resource ID we need to sanitize * characters such that the resulting resource ID is valid

Ref APPSRE-3776


maorfr (approved):

Piojo (approved):
LGTM.

PR #1868, opened 2021-09-22
Re-add logging of skipped managedResourceTypes

Piojo:
When managedResourceNames or managedResourceTypeOverride are empty,
the graphql query still returns the field, but its value is None.
    
We handle it correctly now. I have added a unit test as well.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -1,5 +1,8 @@
 import logging
 import itertools
+
+from typing import Optional, List, Iterable, Mapping
+
 import yaml
 
 from sretoolbox.utils import retry
@@ -45,11 +48,12 @@ class StateSpec:
         self.resource_names = resource_names
 
 
-def init_specs_to_fetch(ri, oc_map,
-                        namespaces=None,
-                        clusters=None,
-                        override_managed_types=None,
-                        managed_types_key='managedResourceTypes'):
+def init_specs_to_fetch(ri: ResourceInventory, oc_map: OC_Map,
+                        namespaces: Optional[Iterable[Mapping]] = None,
+                        clusters: Optional[Iterable[Mapping]] = None,
+                        override_managed_types: Optional[Iterable[str]] = None,
+                        managed_types_key: str = 'managedResourceTypes'
+                        ) -> List[StateSpec]:
     state_specs = []
 
     if clusters and namespaces:
@@ -57,9 +61,9 @@ def init_specs_to_fetch(ri, oc_map,
     elif namespaces:
         for namespace_info in namespaces:
             if override_managed_types is None:
-                managed_types = namespace_info.get(managed_types_key)
+                managed_types = set(namespace_info.get(managed_types_key, []))
             else:
-                managed_types = override_managed_types
+                managed_types = set(override_managed_types)
 
             if not managed_types:
                 continue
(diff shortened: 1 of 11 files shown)


maorfr:
can you separate the typing changes and list to set and such to separate commits? (if not separate PRs?)

this is very difficult to review.

Piojo:
This is two commits: 

1. Revert your revert. The original PR is in #1851 and it was indeed split by commit. I also made sure to have a clean history on that one so that reviewers didn't have to review stuff that I had reworked afterwards. I hope that one is easier to review and I'll definitely address any comments you make there. Type hints were extremely useful in my little refactor.
2. Replace some defaults in `get` calls with `or []`. I didn't know our graphql queries would return all fields even if they weren't present in app-interface. TBH I think I should simply replace the `get` calls with direct dictionary accesses.

Do you mind doing the piecemeal review on #1851 ?

reconcile/openshift_base.py
```diff
@@ -73,35 +78,44 @@ def init_specs_to_fetch(ri, oc_map,
...
-                    [resource_type_override] = resource_type_override
+            resource_names = {}
+            resource_type_overrides = {}
+            for mrn in managed_resource_names:
+                # Current implementation guarantees only one
+                # managed_resource_name of each managed type
+                if mrn['resource'] in managed_types:
+                    resource_names[mrn['resource']] = mrn['resourceNames']
+                else:
+                    logging.info(
```
maorfr, line 98: in this case, we'd like to restore this: https://github.com/app-sre/qontract-reconcile/blob/7e79c0926afb02b7aa5fcaa56213002f86a64c0d/reconcile/openshift_resources_base.py#L564-L575

this should mean that you will not need to make any changes to any integrations, only to the base.

maorfr (changes requested):
the only integration that really needs handling is openshift-resources. limit ranges, resource quotas, network policies - they don't really care about `managedResourceTypes`.

reconcile/openshift_base.py
Piojo: Totally. And the change that restores it is 65fa050b247e4f8f5836dfc85dd09c4f42eff8ab . In particular, go down to the changes to `reconcile/utils/openshift_resource.py`. The problem is that the `KeyError` there used to be dropped on the floor, so this `except` block was never reached.

Once I restore it, we have a bunch of additional callers that don't seem to want that exception at all and that risk crashing on it - limit ranges, network policies an so on. That's why I adapt all of them, simply with an `with suppress(KeyError):` block - it's the largest change for which I'm kinda confident I don't need to write unit tests. I agree that it'd be a misconfiguration if there were such an exception, but our current behaviour is to drop it, so in here I'm following suit. 

It'd be very easy to let them crash and get the configs fixed, but if we decide to do so it should be in a different PR.

Finally, there's an extra caveat: this block applies only to resources that are listed under `openshiftResources` in the namespace or cluster.

However, `init_specs_to_fetch` does an earlier filtering of `managedResourceNames` and `managedResourceTypeOverrides`. If a `managedResourceName` is of a type not in `managedResourceTypes` it is currently dropped and we don't even get to generate a `StateSpec` that could be used in `ri.add_current` or `ri.add_desired`. Those have to be logged here.

This one was harder than it looked, I really needed to research the code. ;)

rporres:
FWIW, openshift-tekton-resources uses `managedResourceTypes` internally, but calls the openshift_resources_base functions to do the actual work

Piojo:
As discussed I'm splitting this PR for easier review.

PR #1869, opened 2021-09-22
intro gabi-authorized-users

BumbleFeng:


maorfr:
can you add a description and a link to a jira ticket you are working on?

reconcile/gabi_authorized_users.py
```diff
@@ -0,0 +1,98 @@
...
+def fetch_desired_state(gabi_instances: List[Dict], ri: RI) -> None:
+    for g in gabi_instances:
+        exp_date = datetime.strptime(g['expirationDate'], '%Y-%m-%d').date()
+        if exp_date < date.today():
+            continue
+        elif (exp_date - date.today()).days > EXPIRATION_MAX:
+            logging.error(
+                f'The maximum expiration date of {g["name"]} '
+                f'shall not exceed {EXPIRATION_MAX} days form today')
+            sys.exit(1)
```
steveteahan, line 45: Mostly just for my knowledge, is there a strong reason to use `sys.exit` versus `RunnerException`?  I see the former much more in the code, but the latter will exit 1 while logging the message in the exception via `cli.run_integration`.  Using `RunnerException` seems like it'd be easier to test, particularly if we care about the error message.

reconcile/gabi_authorized_users.py
```diff
@@ -0,0 +1,98 @@
...
+                namespace['name'],
+                'ConfigMap',
+                g['name'],
+                resource
+            )
+
+
+@defer
+def run(dry_run: bool, thread_pool_size: int = 10,
+        internal: Optional[bool] = None, use_jump_host: bool = True,
```
steveteahan, line 77: nit: Just to save you some typing in the future, `use_jump_host` doesn't need a type annotation, [it will be inferred](https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html) because of the default boolean value.

reconcile/test/test_gabi_authorized_users.py
```diff
@@ -0,0 +1,125 @@
...
+def mock_get_gabi_instances(expirationDate: str) -> List[Dict]:
+    gabi_instances = apply['gql_response']
+    gabi_instances[0]['expirationDate'] = str(expirationDate)
+    return gabi_instances
+
+
+@patch.dict(os.environ, {"USE_NATIVE_CLIENT": "True"}, clear=True)
+class TestGabiAuthorizedUser(TestCase):
+
+    @patch.object(queries, 'get_app_interface_settings')
```
steveteahan, line 53: I'd [autospec](https://docs.python.org/3/library/unittest.mock.html#auto-speccing) every patch unless there is a good reason not to.  Without autospec, your tests could pass for cases where you're passing in the wrong number of arguments, calling methods that don't exist, etc.

steveteahan (commented):
There are people close to gabi than me to review that aspect of things, but just adding some comments related to type hints/test best practices.

esron:
Shouldn't that also add the new integration and description to README?

reconcile/gabi_authorized_users.py
```diff
@@ -0,0 +1,96 @@
...
+        "apiVersion": "v1",
+        "kind": "ConfigMap",
+        "metadata": {
+            "name": name,
+            "annotations": {
+                "qontract.recycle": "true"
+            }
+        },
+        "data": {
+            "authorized-users.yaml": '\n'.join([u for u in users])
```
Piojo, line 29: You don't need the comprehension here, `join` takes an iterable:
```suggestion
            "authorized-users.yaml": '\n'.join(users)
```

Piojo: Also, will this work? Won't this produce a badly indented YAML? Shouldn't this be at least
```yaml
- user1
- user2
- user3
```
?

Piojo: Since JSON lists are compatible with YAML, maybe you'd prefer `json.dumps(users)` here?

reconcile/test/test_gabi_authorized_users.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+@patch.dict(os.environ, {"USE_NATIVE_CLIENT": "True"}, clear=True)
+class TestGabiAuthorizedUser(TestCase):
+
+    @patch.object(queries, 'get_app_interface_settings', autospec=True)
+    @patch.object(SecretReader, 'read', autospec=True)
+    @patch.object(OCDeprecated, 'get_version', autospec=True)
+    @patch.object(queries, 'get_gabi_instances', autospec=True)
+    @patch.object(ApiClient, 'request')
```
Piojo, line 59: You can patch the entire class instead.

reconcile/gabi_authorized_users.py
```diff
@@ -0,0 +1,96 @@
...
+                        tf_resource['account'] == account and \
+                        tf_resource['identifier'] == identifier:
+                    found = True
+                    break
+            if not found:
+                raise RunnerException(
+                    f'Could not find rds identifier {identifier} '
+                    f'for account {account} in namespace {namespace["name"]}')
+            cluster = namespace['cluster']['name']
+            ri.add_desired(
```
Piojo, line 64: Please remember that `add_desired` can raise a `KeyError`, if you really don't care, please suppress it.

PR #1870, opened 2021-09-23
[WIP] Add pip-tools for better dependency management

steveteahan:
requirements.txt will be used for pinning dependencies moving
forward. setup.py will be used for the abstract dependencies. For
now, setup.py retains the same versions, but with time the versions
can be less restrictive (moving towards allowing minor version
upgrades for packages following semver).

This is an early version of what could be an alternative to using pipenv to lock our dependencies for deployment.  With this change, builds will be more reproducible because versions are locked down further and include checksums.  Additionally, there is a simpler mechanism for updating dependencies than having to manually check whether packages have updated and manually update `setup.py`.

There are still some open questions and things to figure out:

* Do we run unit tests against `requirements.txt` dependencies, or those in `setup.py`?
* Does it make sense to integrate `pip-sync` in the workflow, or continue to use `pip`?
* How do we handle cross-platform? Non-Linux platforms are currently used in dev. This could require needing to maintain multiple lock files (one per platform), which we'd like to avoid.

I'll run some comparisons with `pipenv`.  My biggest concerns with pipenv at the moment are:

1. It's still fairly slow, it seems to be taking almost twice as long to install dependencies as pip with a requirements.txt file.  Locking files can still take a long time as well.
2. You can only define regular dependencies and dev dependencies, which makes it harder to have separate dependencies for different tox environments (resulting in slower feedback during dev and slower builds).


rporres:
woa, this would so cool to have, being able to pin all the dependency tree...

PR #1874, opened 2021-09-23
Fix list more than 1000 keys

esron:
JIRA: https://issues.redhat.com/browse/APPSRE-3779

Docs on AWS S3: 
https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html
https://docs.aws.amazon.com/AmazonS3/latest/userguide/ListingKeysUsingAPIs.html

Docs in boto3:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.list_objects_v2

### DIFF
--- a/reconcile/utils/state.py
+++ b/reconcile/utils/state.py
@@ -53,14 +53,24 @@ class State:
         """
         Returns a list of keys in the state
         """
-        objects = self.client.list_objects(Bucket=self.bucket,
-                                           Prefix=f'{self.state_path}/')
+        objects = self.client.list_objects_v2(Bucket=self.bucket,
+                                              Prefix=f'{self.state_path}/')
 
         if 'Contents' not in objects:
             return []
 
-        return [o['Key'].replace(self.state_path, '')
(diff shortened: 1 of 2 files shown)


reconcile/utils/state.py
```diff
@@ -53,14 +53,26 @@ def ls(self):
...
+                    for o in objects['Contents']]
+
+        while objects['IsTruncated']:
+            objects = self.client.list_objects_v2(
+                Bucket=self.bucket,
+                Prefix=f'{self.state_path}/',
+                ContinuationToken=objects['NextContinuationToken']
+            )
+
+            response += [o['Key'].replace(self.state_path, '')
```
Piojo, line 72: Why not do this replacement only once and at the very end of the method?

Piojo (approved):
LGTM.

reconcile/utils/state.py
esron: Done. And thanks.

PR #1875, opened 2021-09-24
Fix missing error key in GH exception

rporres:
We cannot assume errors is always going to appear. It is not there when
the GH error is a 404.

Related to https://issues.redhat.com/browse/APPSRE-3838

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -271,7 +271,7 @@ class SaasHerder():
         except GithubException as e:
             # slightly copied with love from
             # https://github.com/PyGithub/PyGithub/issues/661
-            errors = e.data['errors']
+            errors = e.data.get('errors', [])
             # example errors dict that we are looking for
             # {
             #    'message': '<text>',


rporres:
We get a very criptic message without this:

```
15:10:39 [2021-09-23 13:09:29] [ERROR] [saasherder.py:_process_template:475] - [https://github.com/RedHatInsights/notifications-backend//.rhcicd/deploy-acceptance.yml:bef7a4e193dbba41bf4dcee2c5ad664b5ea36ef0] error fetching template: 'errors'

```

See: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/25478

reconcile/utils/saasherder.py
```diff
@@ -271,27 +271,29 @@ def _get_file_contents_github(repo, path, commit_sha):
...
-                    # get large files
-                    tree = repo.get_git_tree(
-                        commit_sha, recursive='/' in path).tree
-                    for x in tree:
-                        if x.path != path.lstrip('/'):
-                            continue
-                        blob = repo.get_git_blob(x.sha)
-                        return base64.b64decode(blob.content).decode("utf8")
+            if 'errors' in e.data:
+                errors = e.data['errors']
```
Piojo, line 275:
```suggestion
            errors = e.data.get('errors', [])
```
let's try to keep this big-ass arrow a little flatter.

Piojo (approved):
Fix the nit before pushing, please.

reconcile/utils/saasherder.py
rporres: Apologies if I don't understand well the suggestion, but unfortunately that's not going to make things less nested as there's a chance that we don't return from inside the condition. If I go with your suggestion I would just write:

```python
errors = e.data.get('errors', [])
if errors:
    # (... we may return here or not ...)

raise e
```

PR #1879, opened 2021-09-28
APPSRE-3079: Dockerfile image size reduction

lechuk47:


rporres:
This looks cool. The way to avoid getting all the gcc part is very interesting. What's the status of the testing of this?

dockerfiles/Dockerfile
```diff
@@ -1,15 +1,25 @@
...
+     /work/
+
+RUN python3 -m pip install --upgrade pip setuptools wheel && \
+    python3 -m pip wheel . --wheel-dir /work/wheels
+
+
+FROM quay.io/app-sre/qontract-reconcile-base:0.6.0
+
+RUN --mount=type=cache,target=/tmp/work/,from=build-image,source=/work \
+    microdnf upgrade -y && \
```
rporres, line 19: I like the idea of the upgrade every time we build. I would try to make sure that someone else reads this in case there's a strong disagreement :)

dockerfiles/structure-test.yaml
```diff
@@ -0,0 +1,86 @@
+schemaVersion: "2.0.0"
```
rporres, line 1: this is so cool ❤️

rporres (changes requested):
I really like all the work here. I have left a couple of comments.

Any particular reason for not doing `Dockerfile.test` too?

Makefile
```diff
@@ -11,6 +11,8 @@ else
 	DOCKER_CONF := $(HOME)/.docker
 endif
 
+CTR_STRUCTURE_IMG := gcr.io/gcp-runtimes/container-structure-test:latest
```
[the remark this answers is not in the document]
lechuk47: makes sense. On it.

PR #1880, opened 2021-09-29
[openshift_resource] raise KeyError when adding an item

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3668

attempts to replace #1868 (which reverts a revert of #1851).
re-adds functionality which was lost in #1232.

when a KeyError is raised, it means that the resource kind that is referenced from a namespace file under `openshiftResources` is not mentioned in `managedResourceTypes`.

the only integrations which takes the `managedResourceTypes` into account is openshift-resources.


maorfr:
@Piojo how would you like to proceed?

Piojo:
I'll have to insist on #1868, I gave the full reasoning in our slack thread: https://coreos.slack.com/archives/GGC2A0MS8/p1632895377320900 .

maorfr:
agreed on the same slack thread to merge this one and split the other changes to separate PRs

PR #1883, opened 2021-09-29
[terraform-resources] add support for read/write path split in alb

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3695

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/26162

this PR adds support to control the weights of traffic split between read and write paths (as opposed to only write path currently).

@jfchevrette PTAL

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -221,7 +221,10 @@ provider
   targets {
     name
     default
-    weight
+    weights {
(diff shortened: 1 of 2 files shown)


maorfr:
merging and iterating with https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/26162

PR #1884, opened 2021-09-29
saasherder: Adding GenericTemplateExecutor provider + populate desired state test

bwplotka:
Motivation and task: https://issues.redhat.com/browse/MON-1699

Idea: Add docker template executor that takes you can configure in Saas file:
* image with something that takes input Yaml + templates and produces YAML resources 
* where to take templates from
* input YAML with all parameters 

Currently adding tests for basic openshift template first, then extending with docker executor. Let me know if this idea makes sense (: cc @maorfr @jmelis @rporres

### DIFF
--- a/reconcile/test/fixtures/saasfiles/rendered_resources.yaml
+++ b/reconcile/test/fixtures/saasfiles/rendered_resources.yaml
@@ -1,0 +1,194 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/component: api
+    app.kubernetes.io/instance: observatorium
+    app.kubernetes.io/name: observatorium-api
+    app.kubernetes.io/part-of: observatorium
+    app.kubernetes.io/version: main-2021-09-10-v0.1.2-52-ga24d7d2
+  name: observatorium-observatorium-mst-api
+spec:
+  replicas: 3
+  selector:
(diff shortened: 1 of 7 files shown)


maorfr:
thanks for your contribution! i think supporting an additional provider (jsonnet) as we've discussed in our syncs make sense.

waiting for the core functionality to be added :1st_place_medal:

rporres:
First of all, thanks for contributing to this project.

If I understand correctly you want be able to execute a container inside the existing qontract-reconcile container to generate the openshift manifests based on a certain piece of yaml configuration. That would basically allow you to make the development of the template processing completely independent from this repository which has certain limitations on the testing side for those external to app-sre. That would make the contribution path easier which is a good thing. A few comments from my understanding:

1. Docker in docker is always challenging. We would need to check feasibility
2. The input yaml would need to work with the current [saas file schema](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/schemas/app-sre/saas-file-2.yml#L132). Since `parameters` is an object, this should not be a problem.
3. My main worry of the input yaml is to understand which are your ideas about it. The current logic of the saas file is very hierarchic: a template parameter can be defined from the target namespace, the resourceTemplate, the saas file or the environment. There are good reasons for it as it allows us to avoid duplication of variables in many app-interface services.

This is all that I can think ATM :)

PR #1888, opened 2021-09-30
Raise coverage once again

Piojo:


Piojo:
Nope. Different Python versions report different coverage numbers. :(

PR #1891, opened 2021-10-03
[vault] implement kv v2 write

maorfr:
to be able to write integrations' output to vault kv v2 secret engine (currently required for fedramp effort).


reconcile/utils/vault.py
```diff
@@ -238,9 +238,19 @@ def write(self, secret, decode_base64=True):
             self._write_v1(secret_path, data)
 
     def _write_v2(self, path, data):
-        # do not forget to run `self._read_all_v2.cache_clear()`
-        # if this ever get's implemented
-        raise NotImplementedError('vault_client write v2')
+        path_split = path.split('/')
+        mount_point = path_split[0]
+        write_path = '/'.join(path_split[1:])
+
```
rporres, line 244: I'm not very familiar with this class, but since reads in v2 are cached, I understand that the cache clear comments is pertinent and we would need to clear the lru cache in this method.

rporres:
On top of that, tests would never hurt :)

reconcile/utils/vault.py
maorfr: added!

Piojo (approved):

reconcile/utils/vault.py
```diff
@@ -238,9 +238,20 @@ def write(self, secret, decode_base64=True):
...
+        try:
+            self._client.secrets.kv.v2.create_or_update_secret(
+                mount_point=mount_point,
+                path=write_path,
+                secret=data,
+            )
+            self._read_all_v2.cache_clear()
+        except hvac.exceptions.Forbidden:
+            msg = f"permission denied accessing secret '{path}'"
+            raise SecretAccessForbidden(msg)
```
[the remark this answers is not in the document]
maorfr: the vault client is abstracted under `secret_reader` which attempts to be agnostic to the underlying exceptions. this specific one is not handled in `secret_reader`, but it can be at any point.

maorfr:
whlie there are no new tests, i'd like to get this merged to simplify the fedramp handover effort.

PR #1894, opened 2021-10-05
add support to mirror multiple ocp release channels

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3880

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/26503

back when we created this integration, it was a choice to only mirror stable (due to ECR limitations?).
as ocm-quay has turned into our fallback, we want to mirror more release channels into it (no space/other limitations).


patjlm:
LGTM

PR #1900, opened 2021-10-06
APPSRE-3518: Enable namespace deletion

lechuk47:
## Overview 
This PR enables the namespace deletion with the DELETE pattern describe here. Some basic tests for the integration have been included as well. 

## Notes: 
- The testing strategy it's based in the same approach as openshift_namespace_labels 
- Type hints have been set 
- Namespace reconciliations now run un parallel

### DIFF
--- a/reconcile/openshift_namespaces.py
+++ b/reconcile/openshift_namespaces.py
@@ -1,7 +1,7 @@
 import logging
 import sys
 
-from typing import List, Dict, Optional, Any, Iterable, Mapping, Tuple
+from typing import List, Dict, Optional, Any, Iterable, Mapping, Tuple, cast
 import reconcile.utils.threaded as threaded
 import reconcile.queries as queries
 
@@ -54,11 +54,11 @@ def get_shard_namespaces(
 
         if is_in_shard(key):
             if key not in filtered_ns:
-                filtered_ns[key] = dict(ns)
+                filtered_ns[key] = cast(dict, ns)
             else:
                 # Duplicated NS
                 dupe_list_by_key = duplicates.setdefault(key, [])
-                dupe_list_by_key.append(dict(ns))
+                dupe_list_by_key.append(cast(dict, ns))
 
     for key, dupe_list in duplicates.items():
         dupe_list.append(filtered_ns[key])


steveteahan (changes requested):
Took a look at this PR to get more exposure to this part of the code base, but it's probably better to have someone else with more knowledge of this integration to give the final approval. I think the typing and autospec changes would be good to make. Thank you for adding the types and tests!

reconcile/openshift_namespaces.py
```diff
@@ -1,127 +1,135 @@
...
+NS_ACTION_CREATE = "create"
+
+
+def get_desired_state(
+        namespaces: List[Dict[str, Any]]) -> List[Dict[str, str]]:
+
+    desired_state: List[Dict[str, str]] = []
+    for namespace in namespaces:
+        state = NS_STATUS_PRESENT
+        if namespace.get("delete"):
```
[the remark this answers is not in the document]
lechuk47: I think this is not the same case. The problem was that if you request an attribute in a Graphql query that does not exist in the data, the returned dict have the attribute with a None value assigned. In this case the condition checks if the "delete" attribute is True, so if the attribute does not exist or if it's None, the condition result is False.

reconcile/openshift_namespaces.py
```diff
@@ -1,127 +1,135 @@
...
 
+NS_STATUS_PRESENT = "present"
+NS_STATUS_ABSENT = "absent"
+
+NS_ACTION_DELETE = "delete"
+NS_ACTION_CREATE = "create"
+
+
+def get_desired_state(
+        namespaces: List[Dict[str, Any]]) -> List[Dict[str, str]]:
```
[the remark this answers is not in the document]
lechuk47: Thanks for the hint. Fixed all occurrences.

reconcile/test/test_openshift_namespaces.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def setUp(self):
+        """ Setup GQL, State and Openshift mocks, using self.test_ns data """
+        self.oc_clients = {}
+        self.test_ns = []
+
+        module = 'reconcile.openshift_namespaces'
+
+        self.queries_patcher = patch(f'{module}.queries')
+        self.queries = self.queries_patcher.start()
```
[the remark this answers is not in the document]
lechuk47: Fixed with a tearDown method.

reconcile/openshift_namespaces.py
```diff
@@ -1,127 +1,134 @@
...
+
+    if action == NS_ACTION_CREATE:
+        try:
+            oc.new_project(namespace)
+        except StatusCodeError as e:
+            msg = (
+                f'cluster: {cluster}, namespace: {namespace}, '
+                f'exception: {str(e)}'
+            )
+            logging.error(msg)
```
maorfr, line 92: we'd need to indicate that there was an error to make the integration do it's best but eventually fail

reconcile/openshift_namespaces.py
```diff
@@ -1,127 +1,134 @@
...
+                namespaces[shard_key] = namespace
+            else:
+                to_be_ignored.append(shard_key)
+    for shard_key in to_be_ignored:
+        logging.debug(
+            f"Found multiple definitions for the namespace {shard_key};"
+            " Ignoring")
+        del namespaces[shard_key]
+
+    return namespaces.values()
```
maorfr: now i understand why `namespaces` is a dict. neat!

reconcile/test/test_openshift_namespaces.py
```diff
@@ -0,0 +1,111 @@
...
+        self.oc_map.get.side_effect = self._oc_map_get
+
+    def testCreateNamespace(self):
+        self.test_ns = [
+            NS(c1, n1, False, False),
+        ]
+
+        openshift_namespaces.run(False, thread_pool_size=1)
+        oc = self.oc_clients[c1]
+        oc.new_project.assert_called_with(n1)
```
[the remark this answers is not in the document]
lechuk47: Added.

reconcile/openshift_namespaces.py
```diff
@@ -1,127 +1,136 @@
...
 
+NS_STATUS_PRESENT = "present"
+NS_STATUS_ABSENT = "absent"
+
+NS_ACTION_DELETE = "delete"
+NS_ACTION_CREATE = "create"
+
+
+def get_desired_state(
+        namespaces: Iterable[Dict[str, Any]]) -> List[Dict[str, str]]:
```
steveteahan, line 23: Can this be `Iterable[Mapping[str, Any]]` or was there a specific need to leave this one?

reconcile/openshift_namespaces.py
```diff
@@ -1,127 +1,136 @@
...
+            " Ignoring")
+        del namespaces[shard_key]
 
-    return oc_map, desired_state
+    return list(namespaces.values())
 
 
-def check_ns_exists(spec, oc_map):
+def check_ns_exists(spec: Mapping[str, str],
+                    oc_map: OC_Map) -> Tuple[Mapping[str, str], Any]:
```
steveteahan, line 59: Just to make sure I didn't cause any confusion, you don't need to use `Mapping` for the return value, the idea is just to be as general as is possible for inputs, but we can be specific for return types.

reconcile/openshift_namespaces.py
```diff
@@ -1,127 +1,134 @@
...
     results = threaded.run(check_ns_exists, desired_state, thread_pool_size,
                            oc_map=oc_map)
-    specs_to_create = [spec for spec, create in results if create]
 
-    for spec in specs_to_create:
-        logging.info(['create', spec['cluster'], spec['namespace']])
+    ns_to_create = []
+    ns_to_delete = []
+    for ns, exists in results:
+        if exists is None:
```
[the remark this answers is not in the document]
lechuk47: `project_exists` could raise a `StatusCodeError` exception, in that case `check_ns_exists` returns None.

reconcile/openshift_namespaces.py
```diff
@@ -1,127 +1,134 @@
...
+        desired_state.append({"cluster": namespace["cluster"]["name"],
+                              "namespace": namespace["name"],
+                              "state": state})
+
+    return desired_state
 
-    return oc_map, desired_state
 
+def get_shard_namespaces() -> List[Dict[str, str]]:
+    all_namespaces = queries.get_namespaces()
```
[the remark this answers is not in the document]
lechuk47: Added  minimal parameter with a minimal QUERY. I've seen this is a common pattern with other queries.

reconcile/openshift_namespaces.py
lechuk47: Sorry. I missed this one. Fixed.

reconcile/openshift_namespaces.py
```diff
@@ -1,127 +1,137 @@
...
+                              "namespace": namespace["name"],
+                              "state": state})
+
+    return desired_state
+
+
+def get_shard_namespaces() -> List[Dict[str, str]]:
+    all_namespaces = queries.get_namespaces(minimal=True)
+    namespaces = {}
+    to_be_ignored = set()
```
Piojo, line 41: Mmm... given the use you do below `duplicates` is probably a better name.

reconcile/openshift_namespaces.py
```diff
@@ -1,127 +1,137 @@
...
     for namespace in all_namespaces:
         shard_key = f'{namespace["cluster"]["name"]}/{namespace["name"]}'
         if is_in_shard(shard_key):
-            namespaces.append(namespace)
+            if shard_key not in namespaces:
+                namespaces[shard_key] = namespace
+            else:
+                to_be_ignored.add(shard_key)
+    for shard_key in to_be_ignored:
+        logging.debug(
```
Piojo, line 50: Is this a debugging condition or an error? This code is dropping on the floor these namespaces and, in production, it won't leave any traces of why they were dismissed.

reconcile/openshift_namespaces.py
Piojo: Why do you return the spec here? This looks like a condition that's best treated as a boolean:

```python
def ns_exists(...) -> bool:
```

and the callers do:

```python
if ns_exists(spec, oc_map):
    ...
```
I see you use it in the context of a `threaded.run`, you can pair the results with the specs with `zip`, see the comment down there.

reconcile/openshift_namespaces.py
```diff
@@ -1,127 +1,137 @@
...
+    desired_state = get_desired_state(namespaces)
+
     results = threaded.run(check_ns_exists, desired_state, thread_pool_size,
                            oc_map=oc_map)
-    specs_to_create = [spec for spec, create in results if create]
 
-    for spec in specs_to_create:
-        logging.info(['create', spec['cluster'], spec['namespace']])
+    specs = []
+    for spec, exists in results:
```
Piojo, line 125: This is better served with
```python
for spec, exists in zip(desired_state, results):
```

reconcile/openshift_namespaces.py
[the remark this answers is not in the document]
Piojo: Why not ignore the exception in `check_ns_exists` and pass `return_exceptions=True` to `threaded.run`?

reconcile/openshift_namespaces.py
```diff
@@ -1,127 +1,137 @@
...
+                oc.new_project(namespace)
+        except StatusCodeError as e:
+            msg = (
+                f'cluster: {cluster}, namespace: {namespace}, '
+                f'exception: {str(e)}'
+            )
+            logging.error(msg)
+
+    elif action == NS_ACTION_DELETE:
+        logging.info(['create', cluster, namespace])
```
Piojo, line 98: Typo: `delete`.

I find a dispatching table cleaner and less error-prone here:

```python
MANAGED_ACTIONS = {
   NS_ACTION_CREATE: oc.new_project,
   NS_ACTION_DELETE: oc.delete_project
}
```
and in here
```python
    try:
       logging.info([action, cluster, namespace])
       if not dry_run:
           MANAGED_ACTIONS[action](namespace)
    except ...
```

(and I'm a big proponent of not catching exceptions but I'm not gonna block this PR on that philosophical item).

reconcile/test/test_openshift_namespaces.py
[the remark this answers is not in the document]
Piojo: Even stricter, `assert_called_once_with`?

reconcile/openshift_namespaces.py
lechuk47: Not sure if I understand what you mean. I see this error message format is common in a lot of integrations. Any idea or example on what you have in mind?

reconcile/openshift_namespaces.py
lechuk47: Agree on this one. Changed.

reconcile/openshift_namespaces.py
lechuk47: OK. I wanted to manage this corner case the same way it's managed in the namespace_labels integration. I understood that this will be controlled with other mechanism and this code will be removed in the near future. I think it makes sense to report error in this case but we should do the same in the other integration as well.

reconcile/openshift_namespaces.py
```diff
@@ -55,57 +54,51 @@ def get_shard_namespaces() -> List[Dict[str, str]]:
...
-            )
-            logging.error(msg)
+    exists = oc.project_exists(namespace)
 
-    elif action == NS_ACTION_DELETE:
+    if not exists and desired_state == NS_STATE_PRESENT:
+        logging.info(['create', cluster, namespace])
+        if not dry_run:
+            oc.new_project(namespace)
+    elif exists and desired_state == NS_STATE_ABSENT:
```
Piojo, line 74: You still have the typo in the log statement below. That's why I wanted the dispatch table. You DRY and thus make fewer mistakes.

reconcile/openshift_namespaces.py
```diff
@@ -55,57 +54,51 @@ def get_shard_namespaces() -> List[Dict[str, str]]:
...
-            logging.info(['create', cluster, namespace])
-            if not dry_run:
-                oc.new_project(namespace)
-        except StatusCodeError as e:
-            msg = (
-                f'cluster: {cluster}, namespace: {namespace}, '
-                f'exception: {str(e)}'
-            )
-            logging.error(msg)
+    exists = oc.project_exists(namespace)
```
Piojo, line 68: This makes more sense than the old `check_ns_exists`, indeed.

reconcile/openshift_namespaces.py
Piojo: He returns the error code and lets `run` do `sys.exit` as in all other integrations. This looks solid to me.

reconcile/openshift_namespaces.py
Piojo: The problem is, we don't print debug messages in production, but this is extremely useful when a user has submitted a bad configuraton and needs to understand why.

reconcile/openshift_namespaces.py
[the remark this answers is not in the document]
lechuk47: check the other thread please. let's follow the conversation on the other one

reconcile/test/test_openshift_namespaces.py
```diff
@@ -116,3 +121,17 @@ def testDeleteAbsentNamespace(self):
         oc = self.oc_clients[c1]
         oc.delete_project.assert_not_called()
         oc.new_project.assert_not_called()
+
+    def testErrorHandlingProjectExists(self):
```
Piojo, line 125: No camelCase, for methods, please:
```suggestion
    def test_error_handling_project_exists(self):
```

reconcile/test/test_openshift_namespaces.py
```diff
@@ -116,3 +121,17 @@ def testDeleteAbsentNamespace(self):
...
+        oc = self.oc_clients.setdefault(c1, Mock(name=f'oc_{c1}'))
+        oc.project_exists.side_effect = StatusCodeError("SomeError")
+        self.oc_map.get.return_value = oc
+
+        self.test_ns = [
+            NS(c1, "project_raises_exception", delete=True, exists=False),
+        ]
+        f = io.StringIO()
+        with self.assertRaises(SystemExit), contextlib.redirect_stderr(f):
```
Piojo, line 134: This use of contextlib is very neat. I'll use it on other tests. :)

reconcile/openshift_namespaces.py
lechuk47: check the other thread as I stated before.

reconcile/test/test_openshift_namespaces.py
```diff
@@ -116,3 +121,17 @@ def testDeleteAbsentNamespace(self):
...
+            openshift_namespaces.run(False, thread_pool_size=1)
+
+            self.assertTrue("SomeError" in f.getvalue())
```
Piojo, line 137: Please use `self.assertIn` instead:
```python
self.assertIn("SomeError", f.getvalue())
```
so that when this test fails we get `f.getvalue()` in the error message and have an easier time fixing it.

reconcile/test/test_openshift_namespaces.py
lechuk47: Fixed in all cases.

reconcile/openshift_namespaces.py
maorfr: indeed, we just need the integration to exit with an error when this happens.

reconcile/openshift_namespaces.py
lechuk47: Then, is it ok in this case to print error messages but not fail the integration?

lechuk47: I think it makes sense and I've changed it with a slightly consideration. It throws error if there are discrepancies in the namespace definitions. 
- If all the definitions have the delete set, the ns is deleted 
- if all of them don't have the delete set and the namespace does not exist,  it's created ( I think this last one should never happen, but it's covered the same way)

reconcile/openshift_namespaces.py
```diff
@@ -38,41 +39,32 @@ def get_desired_state(
...
-    for ns in namespaces:
-        shard_key = f'{ns["cluster"]["name"]}/{ns["name"]}'
-        if is_in_shard(shard_key):
-            shard_namespaces.append(ns)
-
-    return shard_namespaces
-
-
-def manage_duplicated_namespaces(namespaces: Any) \
+def get_shard_namespaces(namespaces: Any) \
```
Piojo, line 42: This is probably an `Iterable[Dict]` or something like that, isn't it?

reconcile/openshift_namespaces.py
```diff
@@ -38,41 +39,32 @@ def get_desired_state(
...
+                # Duplicated NS
+                dupe_list_by_key = duplicates.setdefault(key, [])
+                dupe_list_by_key.append(ns)
 
-    for key in duplicated_ns:
-        dupe_ns_list = duplicated_ns[key]
-        dupe_ns_list.append(filtered_ns[key])
-        deletes_in_list = [ns["delete"] for ns in dupe_ns_list]
+    for key in duplicates:
+        dupe_list_by_key = duplicates[key]
```
Piojo, line 63: Use `items` here:
```python
for key, dupe_ls in duplicates.items():
    dupe_ls.append(filtered_ns[key])
    ...
```

reconcile/openshift_namespaces.py
```diff
@@ -39,29 +39,29 @@ def get_desired_state(
...
+    filtered_ns: Dict[str, Dict[str, Any]] = {}
 
     err = False
     for ns in namespaces:
         key = f'{ns["cluster"]["name"]}/{ns["name"]}'
 
         if is_in_shard(key):
             if key not in filtered_ns:
-                filtered_ns[key] = ns
+                filtered_ns[key] = dict(ns)
```
Piojo, line 57: Why this extra copy? If the type checker was giving you headaches, `cast` would be a better option wit no run-time cost:

```python
from typing import cast
...
filtered_ns[key] = cast(dict, ns)
```

lechuk47: 👍

maorfr (approved):
LGTM

just a reminder, deleting namespaces is one of the most sensitive things we have. wrong namespace gets deleted - a service goes down.
thanks for working on this so thoroughly!

PR #1901, opened 2021-10-06
Log skipped managedResources on openshift_resource_base

Piojo:
Log in `init_specs_to_fetch` any `managedResourceNames` or
`managedResourceTypeOverrides` not processed because they are not of a
kind listed in `managedResourceTypes`.

These configurations are currently dropped on the floor, making bad
configurations extremely hard to detect.

Continuation of APPSRE-3668

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -1,5 +1,8 @@
 import logging
 import itertools
+
+from typing import Optional, List, Iterable, Mapping
+
 import yaml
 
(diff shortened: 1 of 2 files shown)


Piojo:
That's a great idea. However, I'd need to understand the integrations a little bit better so that a bad namespace on a tenant doesn't block the entire integration. Let's follow up on https://issues.redhat.com/browse/APPSRE-3890 .

PR #1902, opened 2021-10-06
Report skipped resources in saasherder

Piojo:


reconcile/utils/saasherder.py
```diff
@@ -795,10 +795,22 @@ def populate_desired_state_saas_file(self, spec, ri):
...
         # filter resources
-        resources = [resource for resource in resources
-                     if isinstance(resource, dict)
-                     and resource.get('kind') in managed_resource_types]
+        rs = []
+        for r in resources:
+            if isinstance(r, dict):
+                kind = r.get('kind')
+                if kind in managed_resource_types:
+                    resources.append(r)
```
maorfr, line 803: rs?

Piojo: Totally. Unit testing this one was very hard but I should have tried. :(

PR #1904, opened 2021-10-06
Add detailed boto3 ClientError handling in State.exists

geoberle:
Distinguish between 404 errorcodes for missing key and others like 403, NoSuchBucket or
general AWS errors. While 404 results in a False return value (indicating a missing key),
the others yield a StateInaccessibleException exception to indicate an unknown result.

Refs APPSRE-3889


maorfr:
all in all, the PR looks great!

one question -
i played with `head_object` locally and saw that when a bucket doesn't exist i get 404. are you seeing the same?

geoberle:
checking the existence of a bucket as a byproduct of calling head_object was not yielding the consistent results I was hoping for. therefore I implemented the bucket resistance check with head_bucket [1] and moved it to the constructor of the State class. A bucket misconfiguration results in an exception during State instantiation.

[1] https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html?#S3.Client.head_bucket

reconcile/utils/state.py
```diff
@@ -34,20 +41,44 @@ def __init__(self, integration: str, accounts: Iterable[Mapping[str, Any]],
 
         self.client = session.client('s3')
 
+        # check if the bucket exists
+        try:
+            self.client.head_bucket(Bucket=self.bucket)
+        except ClientError as details:
+            error_code = details.response.get('Error', {}).get('Code', None)
+            if error_code == '404':
```
maorfr, line 49: i am seeing cases where a 403 is returned. do we care about error_code here? bucket is just inaccessible...

reconcile/utils/state.py
```diff
@@ -34,20 +41,44 @@ def __init__(self, integration: str, accounts: Iterable[Mapping[str, Any]],
...
+        :raises StateInaccessibleException: if the bucket is missing or
+        permissions are insufficient or a general AWS error occurred
         """
         try:
             self.client.head_object(
                 Bucket=self.bucket, Key=f"{self.state_path}/{key}")
             return True
-        except ClientError:
-            return False
+        except ClientError as details:
```
maorfr, line 69: with the head_bucket check, do we really expect a 403?

reconcile/utils/state.py
maorfr: you can add `details` to the exception message you are raising here?

reconcile/utils/state.py
geoberle: theoretically, we could see 403 when fine-grained permissions are in place (e.g. prefix conditions), but again, we really don't have to distinguish too much beyond 404. I will clean this up a bit and push another proposal

maorfr (approved):
LGTM

kudos on your first PR!

PR #1907, opened 2021-10-07
saasherder: Added tests for openshift template remote desired state populate.

bwplotka:
This is a prerequisite for the reliable addition of jsonnet executor in https://github.com/app-sre/qontract-reconcile/pull/1884

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>


bwplotka:
>   FileNotFoundError: [Errno 2] No such file or directory: 'oc': 'oc'
>  resources = oc.process(template, consolidated_parameters)

Does it mean, we cannot run test that uses `oc ` or in future `jsonnet` process? Or can we add it to test runner somehow. Any pointers? (:

reconcile/utils/saasherder.py
```diff
@@ -436,6 +436,7 @@ def _process_template(self, options):
         html_url = None
         commit_sha = None
 
+        # TODO(bwplotka): Add docker executor.
```
bwplotka, line 439: Will remove in next iteration.

Piojo:
Thanks for the tests - this is not an easy module to write tests for.

We've never had the need to install `oc` on our test images. We can do it next week if needed, but for the time being it'd be best to mark those tests with `@skipIf(os.getuid() == 0)` or to override `Saasherder._process_template` with calls to `helm`.

lechuk47:
@bwplotka the test image container has been updated to contain all the tooling binaries. Could you please try again?

Piojo:
Your tests have passed but now you have linting errors. You can reproduce locally with `tox -e lint`, so that you needn't wait for Jenkins round trips.

maorfr:
:heart:

bwplotka:
Is there any automatic linter for all of this?

bwplotka:
linter with `--fix` option

bwplotka:
Friendly ping @Piojo

Piojo:
We don't have any automatic formatting enabled. We probably should try to make Jenkins report inline any lint issues, but that's not in our roadmap for now.

Piojo (approved):
This is awesome. Thanks for persevering on it.

bwplotka:
Is there any automatic formatting dev can use at least?

Piojo:
Some team members use autopep-8, but you'll find not all files comply nicely with it.

It is still a matter of judgement, and I really don't want a round of formatting-only PRs, since they will break the ability to revert buggy commits in a timely manner.

bwplotka:
Thanks

PR #1910, opened 2021-10-10
[openshift_base] init state specs by managed_types and not by resource_names

maorfr:
fixes a bug introduced in #1901

by iterating over `resource_names` we are only initiating the current state for resources listed in `managedResourceNames` and not over all other resources.


maorfr:
as this causes tests to fail, i assume they were written after the change and not before it.
reverting instead of failing forward: #1911

Piojo:
Tests were written before the change. Please try `git checkout 'master@{2 weeks ago}' reconcile/openshift_resources_base.py 'master@{2 weeks ago}'; tox -e py36`.

PR #1914, opened 2021-10-11
[slack-usergroups] add support for schedule

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3894

depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/26847

this PR adds support to define a schedule in app-interface as the source of truth for a slack usergroup.


reconcile/slack_usergroups.py
```diff
@@ -70,12 +71,23 @@
...
+            }
+          }
+        }
       }
     }
   }
 }
 """
 
+DATE_FORMAT = '%Y-%m-%d %H:%M'
```
Piojo, line 90: We have a lot of `DATE_FORMAT` constants across this repository. Can we either centralize them or agree on using `isoformat` everywhere?

reconcile/slack_usergroups.py
```diff
@@ -270,6 +282,20 @@ def get_slack_usernames_from_owners(owners_from_repo, users, usergroup):
     return all_slack_usernames
 
 
+def get_slack_usernames_from_schedule(schedule):
+    if schedule is None:
+        return []
+    now = datetime.utcnow()
+    all_slack_usernames = []
```
Piojo, line 289: This one is probably better served as a `set`, I can imagine there being duplicates.

PR #1919, opened 2021-10-11
APPSRE-3868: Remove keys with None value 

OliviaHY:
### What
Remove keys with None value from graphql results to avoid integration failure.
### Why
Integrations expect the field to have valid value when it's in the data, otherwise the integration will failed. 
Note that GraphQL has all types by default nullable, so they will be null unless we declare it as not-nullable. But we are not declare any type that way as far as I can tell. Therefore in order to overcome this without changing all integration code, we filter all the keys that has a null value once we got the result.
### Validation
I ran a few queries including 
```
{
  namespaces_v1 {
    managedResourceTypes
    path
  }
}
```
and
```
{
  saas_files_v2(name: "saas-github-mirror") {
    name
    takeover
    path
  }
}
```
The null valued keys has been removed in the result of this code change.

### DIFF
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -141,7 +141,20 @@ class GqlApi:
                 "`data` field missing from GraphQL"
                 "server response."))
 
+        if result['data'] is not None:
+            self.filter_null(result['data'])
+
         return result['data']
+
+    def filter_null(self, data):
+        if not isinstance(data, (str, int)):
+            for k, v in data.copy().items():
+                if isinstance(v, dict):
+                    self.filter_null(v)
+                elif isinstance(v, list):
+                    [self.filter_null(x) for x in v]
+                elif v is None:
+                    del data[k]
 
     def get_resource(self, path):
         query = """


OliviaHY:
I am a little concerned on the performance implication of this change, after all it is going through the whole result and doing "in place" update to it. Although locally it has been very fast when I tested with a json that' 4692 rows and 151 kb.

Piojo (changes requested):
The method looks good, but can you please add some unit tests for `filter_null` ? Start a `test_utils_gql.py` file in `reconcile/test`.

Regarding the performance impact, I'd suggest we keep an eye in the `qontract_reconcile_last_run_seconds` metric for all integrations - and establish a limit to the slow-down (say, 2 seconds?). Let's use an absolute value instead of a relative one since the GQL processing is typically short in comparison with all the other operations that integrations have to do.

maorfr:
respectfully closing so it's not accidentally merged (https://issues.redhat.com/browse/APPSRE-3868)

PR #1928, opened 2021-10-14
Improve handling of empty serverUrl in OC_Map

steveteahan:
Also add a related comment to indicate why server can be a falsy value
in OCNative. Changes related to this were unnecessary to fix the
immediate concern, but a follow-up ticket was created in APPSRE-3926.

A follow-up app-interface change will be made to prevent serverUrl from being empty if automationToken is set.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -1,6 +1,7 @@
 import logging
 
 from datetime import datetime, timedelta
+from typing import Dict, List
 
 import gitlab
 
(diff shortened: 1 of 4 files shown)


reconcile/test/test_utils_oc.py
```diff
@@ -320,3 +321,43 @@ def test_many_wrong(self):
...
+        should be skipped.
+        """
+        cluster = {
+            'name': 'test-1',
+            'serverUrl': 'http://localhost',
+            'automationToken': None
+        }
+        oc_map = OC_Map(clusters=[cluster])
+
+        self.assertTrue(isinstance(oc_map.get(cluster['name']), OCLogMsg))
```
Piojo, line 360:
```suggestion
        self.assertIsInstance(oc_map.get(cluster['name']), OCLogMsg))
```

steveteahan: Good catch - I fixed these.

PR #1933, opened 2021-10-18
Autofill more details with ocm-clusters

Piojo:
Fill in consoleUrl, serverUrl, prometheusUrl, alertmanagerUrl and
elbFQDN.

We need to split the changes in the integration itself into changes
for the `spec` subtree and changes directly to the root tree, hence
the changes to `utils.mr.create_...`. In addition to that, the `OCM`
class now stores a few more fields from OCM.

It has unit tests for all the affected modules, and I've done a local
dry-run (after removing these fields from appsres03ue1), whose output
is in https://gitlab.cee.redhat.com/-/snippets/4353 .

Fixes https://issues.redhat.com/browse/APPSRE-3610

### DIFF
--- a/reconcile/ocm_clusters.py
+++ b/reconcile/ocm_clusters.py
@@ -138,14 +138,6 @@ def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
                 clusters_updates[cluster_name]['root']['elbFQDN'] = \
                     f"elb.apps.{cluster_name}.{current_spec['domain']}"
 
-            if not desired_spec.get('prometheusUrl'):
-                clusters_updates[cluster_name]['root']['prometheusUrl'] = \
-                    f"https://prometheus.{cluster_name}.devshift.net"
-
-            if not desired_spec.get('alertmanagerUrl'):
-                clusters_updates[cluster_name]['root']['alertmanagerUrl'] = \
-                    f"https://alertmanager.{cluster_name}.devshift.net"
-
             desired_provision_shard_id = \
                 desired_spec['spec'].get('provision_shard_id')
             current_provision_shard_id = \


reconcile/ocm_clusters.py
```diff
@@ -5,16 +5,35 @@
...
+    'storage',
+    'load_balancers',
+    'private',
+    'channel',
+    'autoscale',
+    'nodes'
+}
+
+OCM_GENERATED_FIELDS = ['network', 'consoleUrl', 'serverUrl', 'elbFQDN']
+PREDICTABLE_FIELDS = ['prometheusUrl', 'alertmanagerUrl', ]
```
steveteahan, line 25: Is there a missing field here, or is the comma accidental?

reconcile/ocm_clusters.py
```diff
@@ -75,9 +87,10 @@ def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
     error = False
     clusters_updates = {}
     for cluster_name, desired_spec in desired_state.items():
+        logging.info(f"Processing cluster {cluster_name}")
```
steveteahan, line 90: Would this log unconditionally on each execution, or am I missing something? Basically wondering if this would be extra logging in Slack that isn't necessarily associated with a change?

reconcile/test/test_ocm_clusters.py
```diff
@@ -0,0 +1,224 @@
...
+                self.clusters[0],
+                self.clusters[0]
+            ),
+            ({}, False)
+        )
+
+    def test_valid_change(self):
+        desired = deepcopy(self.clusters[0])
+        desired['spec']['instance_type'] = 'm42.superlarge'
+        print(desired)
```
steveteahan, line 63: Were these just for debugging?

reconcile/test/test_ocm_clusters.py
```diff
@@ -0,0 +1,224 @@
...
+        self.mock_constructor(ocmmod, 'OCMMap').to_return_value(self.ocmmap)
+        self.mock_callable(
+            self.ocmmap, 'get'
+        ).for_call('cluster1').to_return_value(self.ocm)
+        self.update_cluster = self.mock_callable(
+            self.ocm, 'update_cluster'
+        ).to_return_value(None)
+        self.mock_callable(
+            sys, 'exit'
+        ).to_raise(ValueError)
```
steveteahan, line 117: This is because sys.exit will exit the test execution? The only thing I'm wondering is if it would make sense to raise some less common exception in it's place to avoid cases where a `ValueError` could really be legitimately raise? Definitely a solid case for avoiding `sys.exit` in the future as we've discussed.

reconcile/test/test_utils_mr_clusters_updates.py
```diff
@@ -0,0 +1,84 @@
...
+
+class TestProcess(TestCase):
+    def setUp(self):
+        self.clusters = [
+            fxt.get_anymarkup('cluster1.yml')
+        ]
+        self.raw_clusters = fxt.get('cluster1.yml')
+
+    def test_no_changes(self):
+        cli = MagicMock()
```
steveteahan, line 35: Does it make sense to `create_autospec` here to catch cases where we're calling methods that don't even exist, etc? I won't comment on each instance of this.

reconcile/test/test_utils_mr_clusters_updates.py
```diff
@@ -0,0 +1,84 @@
...
+from ruamel import yaml
+
+import reconcile.utils.mr.clusters_updates as sut
+
+from .fixtures import Fixtures
+
+fxt = Fixtures('clusters')
+
+
+class TrivialClustersUpdates(sut.CreateClustersUpdates):
```
steveteahan, line 13: The approach here wasn't clear to me until I looked a little deeper into this. What's the reason behind creating this new object from `CreateClusterUpdates` versus mocking cancel, or checking for `CancelMergeRequest` to be raised? I don't think one versus the other is right or wrong, but some comments might help others to understand this decision.

PR #1934, opened 2021-10-19
reduce some integration run durations, using api caches

patjlm:
APPSRE-3935

Some timings, from my laptop:

```
time qontract-reconcile --config config.debug.toml --dry-run terraform-tgw-attachments --print-only
before: 7m15.120s
after:  2m33.083s
```
```
time qontract-reconcile --config config.debug.toml --dry-run terraform-vpc-peerings --print-only
before: 5m43.625s
after:  1m54.702s
```

### DIFF
--- a/reconcile/utils/aws_api.py
+++ b/reconcile/utils/aws_api.py
@@ -6,21 +6,26 @@ import time
 
 from datetime import datetime
 from threading import Lock
+from typing import TYPE_CHECKING
 from typing import Any, Dict, Iterable, List, Mapping, Optional, Tuple
 
 from boto3 import Session
 import botocore
 
-from mypy_boto3_ec2 import EC2Client
-from mypy_boto3_ec2.type_defs import (
-    RouteTableTypeDef, SubnetTypeDef,
-    TransitGatewayTypeDef, TransitGatewayVpcAttachmentTypeDef, VpcTypeDef
-)
-
 from reconcile.utils import threaded
 import reconcile.utils.lean_terraform_client as terraform
 
 from reconcile.utils.secret_reader import SecretReader
+
+if TYPE_CHECKING:
+    from mypy_boto3_ec2 import EC2Client
+    from mypy_boto3_ec2.type_defs import (
+        RouteTableTypeDef, SubnetTypeDef, TransitGatewayTypeDef,
+        TransitGatewayVpcAttachmentTypeDef, VpcTypeDef
+    )
+else:
+    EC2Client = RouteTableTypeDef = SubnetTypeDef = TransitGatewayTypeDef = \
+        TransitGatewayVpcAttachmentTypeDef = VpcTypeDef = object
 
 
 class InvalidResourceTypeError(Exception):
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,6 @@ setup(
         "python-terraform>=0.10.0,<0.11.0",
         "boto3>=1.17.49,<=1.18.0",
         "botocore>=1.20.49,<=1.21.0",
-        "boto3-stubs[ec2,s3,rds]~=1.18",
         "urllib3>=1.25.4,<1.26.0",
         "slack_sdk>=3.10,<4.0",
         "pypd>=1.1.0,<1.2.0",
--- a/requirements-type.txt
+++ b/requirements-type.txt
@@ -1,5 +1,4 @@
 mypy
-types-boto3
 types-click
 types-croniter
 types-Jinja2
@@ -10,3 +9,4 @@ types-PyYAML
 types-requests
 types-tabulate
 types-toml
+boto3-stubs[ec2,s3,rds]


reconcile/utils/aws_api.py
```diff
@@ -118,12 +143,12 @@ def map_s3_resources(self):
             if 'Buckets' not in buckets_list:
                 continue
             buckets = [b['Name'] for b in buckets_list['Buckets']]
-            self.set_resouces(account, 's3', buckets)
+            self.set_resources(account, 's3', buckets)
```
Piojo, line 146: ewww... thank you.

setup.py
```diff
@@ -33,6 +33,7 @@
         "python-terraform>=0.10.0,<0.11.0",
         "boto3>=1.17.49,<=1.18.0",
         "botocore>=1.20.49,<=1.21.0",
+        "boto3-stubs[ec2,s3,rds]~=1.18",
```
steveteahan, line 36: This can go in [requirements-type.txt](https://github.com/app-sre/qontract-reconcile/blob/master/requirements-type.txt) because it's only needed during type checking.

reconcile/terraform_tgw_attachments.py
```diff
@@ -17,7 +17,7 @@
 QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
 
 
-def build_desired_state_tgw_attachments(clusters, ocm_map, settings):
+def build_desired_state_tgw_attachments(clusters, ocm_map, awsapi):
```
steveteahan, line 20: nit: maybe not a bad time to add a type hint now, even if just for `awsapi`? Adding a single type hint will make Mypy actually check the cfunction.

reconcile/utils/aws_api.py
patjlm: 😄

setup.py
steveteahan: It looks like `types-boto3` is already there, which is a proxy package for `boto3-stubs`. I wonder if that might have been related in some way. In any case, I would try removing `types-boto3` and putting this in its place, then seeing if you can reproduce. Stub and type packages shouldn't be imported directly, so shouldn't need to be runtime dependencies.

patjlm:
> have you considered using `@functools.lru_cache()`? it's already in use across the repo and may save some repetition?

Oh my, did not think at all about this. Maybe even `@functools.cache()` given we should not have too many things in there ?

setup.py
patjlm: same errors during unit tests, looking like:
```
________ ERROR collecting reconcile/test/test_terraform_vpc_peerings.py ________
ImportError while importing test module '/package/reconcile/test/test_terraform_vpc_peerings.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib64/python3.6/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
reconcile/test/test_terraform_vpc_peerings.py:4: in <module>
    import reconcile.terraform_vpc_peerings as integ
reconcile/terraform_vpc_peerings.py:6: in <module>
    import reconcile.utils.aws_api as awsapi
reconcile/utils/aws_api.py:14: in <module>
    from mypy_boto3_ec2 import EC2Client
E   ModuleNotFoundError: No module named 'mypy_boto3_ec2'
```
I am keeping it in `setup.py` for now, until a better / proper way is found

steveteahan: I see the issue now. It wasn't loading that file for me due to a large number of changes. You are importing the type library, which is unique to boto3 due to the dynamic nature of that library.

This is covered in the [docs](https://pypi.org/project/boto3-stubs/) under `Pylint compatibility` section. A conditional with `typing.TYPE_CHECKING` is traditionally how you'd import only for type checking. The only thing that might be better than what they've described is representing the types as strings (ex. `my_func() -> "EC2Client"`). The string version can be used when you're referring to a type before it is defined (think self reference in classes), but might also be useful here. I believe that'd work with 3.6, but what they've described in the docs would work too (although it's a bit less nice).

steveteahan: Also, the string representation and forward references are covered here: https://www.python.org/dev/peps/pep-0484/#forward-references

This won't be required in Python 3.7+, but for now it's fine.

reconcile/terraform_vpc_peerings.py
```diff
@@ -57,15 +57,15 @@ def aws_account_from_infrastructure_access(cluster, access_level, ocm_map):
                     ocm.get_aws_infrastructure_access_terraform_assume_role(
                         cluster['name'],
                         awsAccess['awsGroup']['account']['uid'],
-                        awsAccess['awsGroup']['account']['terraformUsername'],
+                        awsAccess['awsGroup']['account']['terraformUsername']
```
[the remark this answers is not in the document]
maorfr: some nits keep everything from falling apart!

rporres (approved):
Thanks a lot for splitting the changes in different commits. That has made the review feasible for me :)

LGTM

PR #1938, opened 2021-10-19
Log updates to clusters only if there are updated specs

Piojo:
Now, the desired and current specs contain different fields, such as
consoleUrl on one or domain on the other. The previous difference
doesn't work anymore. However, the results to
`get_cluster_update_spec` remain the same and are the true measure of
whether any changes need to be applied to the cluster itself.

Adjusted unit tests and ran locally. Clusters with all fields correct
are not logged, but clusters with changes in their spec still display
this message.


jfchevrette (approved):
LGTM!

PR #1941, opened 2021-10-20
Handle SecretNotFound exceptions for cluster automationTokens

steveteahan:
This change allows the automationToken entry to be added to the
cluster.yml file earlier in the process without failing the
integration due to the SecretNotFound exception. This provides
the opportunity to further streamline the cluster provisioning
process.

Related ticket with testing: https://issues.redhat.com/browse/APPSRE-3442

### DIFF
--- a/reconcile/utils/secret_reader.py
+++ b/reconcile/utils/secret_reader.py
@@ -1,7 +1,9 @@
+from typing import Mapping, Optional
+
 from hvac.exceptions import Forbidden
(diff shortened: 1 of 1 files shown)


steveteahan:
See the associated WIP PR to see if that approach would be acceptable. If so, I'll add unit tests to it and get that merged, then loop back to this PR to catch `secret_reader.SecretNotFound` instead of the vault version.

PR #1943, opened 2021-10-21
Raise single type of SecretNotFound exception in SecretReader

steveteahan:
This is WIP and relates to changes in #1941. I want to make sure that the approach is what we want before I write a number of unit tests for `SecretReader`. An example of a stack trace with this change is:

```
>>> sr.read({'path': 'a', 'field': 'test'})
  File "/home/steahan/Code/app-sre/qontract-reconcile/reconcile/utils/config.py", line 32, in read
    config = config[t]
TypeError: 'NoneType' object is not subscriptable

During handling of the above exception, another exception occurred:

  File "/home/steahan/Code/app-sre/qontract-reconcile/reconcile/utils/secret_reader.py", line 59, in read
    data = config.read(secret)
  File "/home/steahan/Code/app-sre/qontract-reconcile/reconcile/utils/config.py", line 36, in read
    f'key not found in config file {path}: {str(e)}')
reconcile.utils.config.SecretNotFound: key not found in config file a: 'NoneType' object is not subscriptable

The above exception was the direct cause of the following exception:

  File "/home/steahan/Code/app-sre/qontract-reconcile/venv/lib64/python3.6/site-packages/sretoolbox/utils/retry.py", line 78, in f_retry
    raise exception
  File "/home/steahan/Code/app-sre/qontract-reconcile/venv/lib64/python3.6/site-packages/sretoolbox/utils/retry.py", line 73, in f_retry
    return function(*args, **kwargs)
  File "/home/steahan/Code/app-sre/qontract-reconcile/reconcile/utils/secret_reader.py", line 61, in read
    raise SecretNotFound(*e.args) from e
reconcile.utils.secret_reader.SecretNotFound: key not found in config file a: 'NoneType' object is not subscriptable
```
So, only `secret_reader.SecretNotFound` needs to be handled regardless of whether the Vault or config is used.


steveteahan:
> the approach looks wonderful. this feels like a use case for a decorator (catch these exceptions and raise this one instead)? will that make testing easier or more difficult?

I thought about the decorator, but for only two methods it didn't feel necessary. I don't think it'd make testing easier or harder, but it'd be another thing to test (the decorator itself). I think the decorator would help with reducing code duplication, but that's the part I wasn't very concerned about with two methods. I don't have a strong opinion either way, just wanted to make sure this is what you had in mind rather than catching both types of exceptions in the other PR.

maorfr:
more use cases that this decorator will cover can be found in utils/vault.py, and probably more places.

maorfr:
i'm just nitting and playing as if i know python. please feel free to choose what you prefer!

maorfr (approved):
LGTM

implementation up to you!

PR #1956, opened 2021-10-28
APPSRE-3784: Use permissions instead of roles

OliviaHY:
### Why:
Originally slack-usergroups integration query the roles and loop through it to update permission, which means when there is no role for a permission, it will be skipped. And we want to avoid that. So the proposal was to query permissions and iterate those instead, with the schema change to include the information about roles there.
### What:
- Update the query of permissions with the updated schema; 
- Use permissions instead of roles to generate desired state so no permission update will be skipped due to role not exist; 
- Fail integration instead of log warning when the user group is not in managedUsergroup.
### Validation:
Dry-run success before adding the failing mechanism, which result in a failing for crc-bop-team. Asking what to do there right now.
The desired_state data is mostly the same after this change as far as I can tell, will be comparing them further.
### Dependency: 
This shoudn't be merged till the [schema change](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/27719) is.

### DIFF
--- a/reconcile/test/test_queris.py
+++ b/reconcile/test/test_queris.py
@@ -1,0 +1,82 @@
+from unittest import TestCase
+from unittest.mock import create_autospec, patch
+
+from reconcile import queries
+from reconcile.utils import gql
+
+from .fixtures import Fixtures
+
+PERMISSIONS_QUERY = """
+{
+  permissions: permissions_v1 {
+    service
+    ...on PermissionSlackUsergroup_v1 {
+      channels
+      description
+      handle
+      ownersFromRepos
+      pagerduty {
+          name
+          instance {
+            name
+          }
+          scheduleID
+          escalationPolicyID
+        }
+      roles {
+        users {
+            name
+            org_username
+            slack_username
+            pagerduty_username
+        }
+    }
+      schedule {
+          schedule {
+            start
+            end
+            users {
+              org_username
+              slack_username
+            }
+          }
+        }
+      workspace {
+        name
+        token {
+          path
+          field
+        }
+        api_client {
+          global {
+            max_retries
+            timeout
+          }
+          methods {
+            name
+            args
+          }
+        }
+        managedUsergroups
+      }
+    }
+  }
+}
+"""
+
+
+class TestQueris(TestCase):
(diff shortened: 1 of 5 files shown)


OliviaHY:
> the group causing the failure should be added to the managed groups list. the changes look good overall. make things green and you'll get a review :)

reconcile/test/test_slack_usergroups.py
```diff
@@ -86,6 +89,35 @@ def test_get_pagerduty_username_slack_username(self):
...
+                    'app-sre-team',
+                    'app-sre-ic'
+                ]
+            }
+        }
+        result = integ.get_slack_map()
+        slackApi.assert_called_once()
+        self.assertEqual(result['coreos']['managed_usergroups'],
+                         expected_slack_map['coreos']['managed_usergroups'])
+        self.assertIn('slack', result['coreos'].keys())
```
Piojo, line 113: Nit: you don't need `keys()` here:

```
self.assertIn('slack', result['coreos'])
```

reconcile/test/test_queris.py
```diff
@@ -0,0 +1,82 @@
...
+        }
+        managedUsergroups
+      }
+    }
+  }
+}
+"""
+
+
+class TestQueris(TestCase):
```
Piojo, line 68: Typo: `TestQueries`

reconcile/test/test_queris.py
```diff
@@ -0,0 +1,82 @@
...
+    @patch.object(gql, "get_api")
+    def test_get_permissions_return_all_slack_usergroup(self, get_api):
+        gqlapi_mock = create_autospec(gql.GqlApi)
+        gql.get_api.return_value = gqlapi_mock
+        gqlapi_mock.query(PERMISSIONS_QUERY).return_value = \
+            self.get_permissions_fixture()
+        result = queries.get_permissions()
+        self.assertTrue(all(x['service'] == 'slack-usergroup' for x in result))
```
Piojo, line 76: While I love the FP approach here, maybe `self.assertEqual({x['service'] for x in result}, {'slack-usergroup'})` would yield a more informative message when the test fails?

PR #1962, opened 2021-10-28
[terraform-resources] add support for route53 zone

maorfr:
related to https://issues.redhat.com/browse/OSD-8770

depends on https://github.com/app-sre/qontract-schemas/pull/5

This PR adds support to provision Route53 DNS zones using terraform-resources.
We do have a specific terraform integration to manage Route53 zones, but the usage proposed in this PR only creates the resource for a different service to manage.

This use case applies to:
1. In the FedRamp environment, certman-operator will need to manage DNS challenges with LE via such a zone.
2. Hive managed cluster records via such a zone.

Note:
exact permissions we need to grant the IAM service user: https://github.com/openshift/hive/blob/master/docs/using-hive.md#managed-dns-1


reconcile/utils/terrascript_client.py
```diff
@@ -3605,3 +3607,62 @@ def populate_tf_resource_alb(self, resource, namespace_info):
...
+            'name': identifier,
+            'tags': common_values['tags'],
+        }
+        zone_id = safe_resource_id(identifier)
+        zone_tf_resource = aws_route53_zone(zone_id, **values)
+        tf_resources.append(zone_tf_resource)
+
+        # outputs
+        # zone id
+        output_name_0_13 = output_prefix + '__zone_id'
```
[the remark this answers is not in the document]
maorfr: a lot of information here: https://issues.redhat.com/browse/APPSRE-2269

reconcile/utils/terrascript_client.py
```diff
@@ -3624,6 +3625,16 @@ def populate_tf_resource_route53_zone(self, resource, namespace_info):
...
         tf_resources.append(zone_tf_resource)
 
+        # outputs
+        # zone id
+        output_name_0_13 = output_prefix + '__zone_id'
+        output_value = f'${{{zone_tf_resource.zone_id}}}'
+        tf_resources.append(Output(output_name_0_13, value=output_value))
+        # name servers
+        output_name_0_13 = output_prefix + '__name_servers'
+        output_value = f'${{{zone_tf_resource.name_servers}}}'
```
geoberle, line 3635: reusing variables is a source of confusion.

maorfr: who's going to be the unlucky person who gets to refactor this entire file? :)

geoberle: always assume it is future you :)

maorfr: damn! ok, at least we agree it is _future_ me :)

Piojo (approved):

reconcile/utils/terrascript_client.py
```diff
@@ -3605,3 +3607,62 @@ def populate_tf_resource_alb(self, resource, namespace_info):
...
+            self.get_tf_iam_service_user(
+                zone_tf_resource,
+                identifier,
+                policy,
+                common_values['tags'],
+                output_prefix,
+            )
+        )
+
+        for tf_resource in tf_resources:
```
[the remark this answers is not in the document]
maorfr: this is consistent with the entire file. if we want to make changes, IMO we need to refactor and not enforce this on particular PRs.

reconcile/utils/terrascript_client.py
```diff
@@ -3991,5 +3979,73 @@ def populate_tf_resource_asg(self, resource, namespace_info):
...
+        tf_resources.append(Output(output_name_0_13, value=output_value))
+
+        policy = {
+            "Version": "2012-10-17",
+            "Statement": [
+                {
+                    "Effect": "Allow",
+                    "Action": [
+                        "route53:Change*",
+                        "route53:Create*",
```
boranx, line 4018: speaking about roles, seems like we're missing one?

`route53:DeleteHostedZone` or  `route53:Delete*` as what I see from https://github.com/openshift/hive/blob/master/docs/using-hive.md#managed-dns-1

Sorry for the noise if this is covered already :) great job :+1:

maorfr: thanks!

i did this intentionally so Hive doesn't accidentally delete the zone. it is created through app-interface, so i was also reluctant to give `Create`, but i don't mind if it attempts to create something that exists.

boranx: How can hive accidentally delete a zone? I would assume that it could be only on a case where hive would intentionally delete a record under the hosted zone, are there other exceptions?
Speaking about giving create, I'm not sure either. We should be ok as long as we _update_/ _populate_ records under the hosted zone

maorfr: i don't know how hive can delete a zone, but we are not allowing that anyway :)

boranx: sounds cool :) lgtm :+1:

PR #1967, opened 2021-11-01
Fix various auto-promotion issues

geoberle:
This MR addresses multiple issues in the qontract-reconcile auto-promotion feature

## Prevent auto promoter from reverting changes when processing multiple promotions per saas file

When the auto-promote processes more than one target-promotion in a saas file
at the same time, the second promotion reverts the first one.

walkthrough:
* while processing the first promotion, the saas file from the master branch
  is loaded, the target ref is updated, and a commit to the temporary branch
  is created
* when another promotion for the same saas file is processed, the saas file
  is loaded again from the master branch, which does not contain the change
  from the first promotion. then the change from the promotion is applied and
  a commit is added to the temporary branch. this commit effectively reverts
  the change from previous commits in the temporary branch to the saas file.

the behaviour can be prevented by reading the saas file from the temporary
branch. for the first promotion to process, this does not make a difference,
because the temporary branch has just been created from master. for all
consecutive commits, it is the right thing to do, to prevent the revert and
build upon the previous changes.

This has been part of the problem in APPSRE-4001 where the customer
observed such a ref revert.

## Prevent empty commits for auto promoter

This code change prevents empty commits in situations where a commit sha
has already been promoted to all subscribed targets in a saas file. Those
empty commits are noisy and confusing in the MRs and could also be the
reason for certain unmergable auto-promote MRs (unconfirmed).

## Ensure unique auto promoter MR title
The title of auto promoter MRs have been static, which resulted in dropped
promotions because MergeRequestBase prevents the creation of MRs with
duplicate names.

This has been part of the problem for APPSRE-4001 where the logs of
gitlab-mr-sqs-consumer showed multiple promotions close after one another
for different saas files, and one being dropped because of the
duplicate title.

### DIFF
--- a/reconcile/utils/mr/auto_promoter.py
+++ b/reconcile/utils/mr/auto_promoter.py
@@ -1,7 +1,11 @@
+import logging
 from ruamel import yaml
 
 from reconcile.utils.mr.base import MergeRequestBase
 from reconcile.utils.mr.labels import AUTO_MERGE
+
+
+LOG = logging.getLogger(__name__)
 
 
 class AutoPromoter(MergeRequestBase):
@@ -32,6 +36,7 @@ class AutoPromoter(MergeRequestBase):
             if not commit_sha:
                 continue
             for saas_file_path in saas_file_paths:
+                saas_file_updated = False
                 raw_file = gitlab_cli.project.files.get(
                     file_path=saas_file_path,
                     ref=self.branch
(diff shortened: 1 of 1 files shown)


maorfr (approved):
LGTM

iterations FTW

patjlm (approved):
Great, thanks !

PR #1968, opened 2021-11-02
[state] don't initiate AWS users

maorfr:
when integrations are using `State`, they only care about the actual state (stored in S3) and not about users in the account.

https://coreos.slack.com/archives/C024FNN7202/p1635838811182800


geoberle:
LGTM

geoberle (approved):

PR #1969, opened 2021-11-02
allow config file to be defined through environment variable

maorfr:
i never want to type `--config` again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,8 @@ qontract-reconcile --config config.toml --dry-run <subcommand>
 qontract-reconcile --config config.toml <subcommand>
 ```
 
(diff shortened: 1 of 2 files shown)


patjlm (approved):
LGTM

patjlm:
I was going to mention we could have the same in qontract-cli 😃

patjlm:
> I was going to mention we could have the same in qontract-cli 😃

or does it use the same @config_file already ?

maorfr:
same one indeed!
https://github.com/app-sre/qontract-reconcile/blob/747b0d85a43ec60b08a23793650cc5c3ba2f8fd1/tools/qontract_cli.py#L30

PR #1971, opened 2021-11-02
[terraform] fix default region provider

maorfr:
resources which do not specify a `region` are supposed to be created in the default region defined at the account level.
until now, they were created in the terraform state bucket region instead.


jfchevrette (approved):
LGTM

rrati:
LGTM

PR #1972, opened 2021-11-02
[terraform-resources] support aws partition (in sqs)

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4008

depends on https://github.com/app-sre/qontract-schemas/pull/8

this PR adds an option to define a `partition` on an aws account file. this is required for policies and other aws resources to be compliant with the aws account type (regular (`aws`) vs GovCloud (`aws-us-gov`)).

### DIFF
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -160,7 +160,7 @@ class TerrascriptClient:
                 access_key=config['aws_access_key_id'],
                 secret_key=config['aws_secret_access_key'],
                 version=self.versions.get(name),
-                region=config['region'])
+                region=config['resourcesDefaultRegion'])
 
             # the time provider can be removed if all AWS accounts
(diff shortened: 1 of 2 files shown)


patjlm (approved):

reconcile/utils/terrascript_client.py
```diff
@@ -160,7 +160,7 @@ def __init__(self, integration, integration_prefix,
                 access_key=config['aws_access_key_id'],
                 secret_key=config['aws_secret_access_key'],
                 version=self.versions.get(name),
-                region=config['resourcesDefaultRegion'])
+                region=config['region'])
```
patjlm, line 163: aren't you reverting the previous fix here ?

maorfr: good catch!

patjlm (approved):
LGTM

PR #1975, opened 2021-11-03
Add tooling binaries to Dockerfile.test

lechuk47:
* Helm installation moved to qontract-reconcile-builder image. 
* Now the dockerfile.test image extends from qontract-reconcile-base. Now all tooling binaries are available for testing purposes.


rporres:
I assume that later you will also change the main Dockerfile to use builder 0.2.0, correct?

rporres (approved):

PR #1976, opened 2021-11-03
Remove py36 tests

Piojo:
We are py39 from now on


steveteahan:
I think we should update the classifiers in `setup.py` to reflect only supporting 3.9.

steveteahan (approved):

PR #1978, opened 2021-11-04
Add default timeout values for gitlab/github api clients

lechuk47:
APPSRE-3888

### DIFF
--- a/reconcile/test/test_utils_github_api.py
+++ b/reconcile/test/test_utils_github_api.py
@@ -1,4 +1,3 @@
-import os
 from unittest import TestCase
 from unittest.mock import patch
 from requests.exceptions import ConnectTimeout
--- a/reconcile/test/test_utils_github_api.py
+++ b/reconcile/test/test_utils_github_api.py
@@ -1,5 +1,5 @@
 from unittest import TestCase
-from unittest.mock import patch
+from unittest.mock import patch, MagicMock
 from requests.exceptions import ConnectTimeout
 from reconcile.utils.github_api import GithubApi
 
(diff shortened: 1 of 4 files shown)


rporres (commented):
Looks great, I love the tests

Just nit-picking: can we make the timeouts a float, e.g. `0.1` so that they fail faster?

reconcile/test/test_utils_github_api.py
```diff
@@ -0,0 +1,18 @@
...
+from unittest.mock import patch
+from requests.exceptions import ConnectTimeout
+from reconcile.utils.github_api import GithubApi
+
+
+class TestGithubApi(TestCase):
+
+    @patch("reconcile.utils.github_api.GH_BASE_URL", "http://198.18.0.1")
+    def test_github_client_timeout(self):
+        with patch('reconcile.utils.github_api.SecretReader', autospec=True) \
```
Piojo, line 11: I tend to dislike using this `with` format, any changes to indentation will make your tests use the real class and it'll be very confusing. Why don't you decorate this method with the new base URL and the `SecretReader`?

reconcile/test/test_utils_gitlab_api.py
```diff
@@ -0,0 +1,19 @@
+from unittest import TestCase
+from unittest.mock import patch
+from requests.exceptions import ConnectTimeout
+from reconcile.utils.gitlab_api import GitLabApi
+
+
+class TestGitlabApi(TestCase):
+
+    def test_gitlab_client_timeout(self):
+        with patch('reconcile.utils.gitlab_api.SecretReader', autospec=True) \
```
Piojo, line 10: As above.

Piojo (changes requested):
Trivial requests, but they can cause problems in the near future.

reconcile/test/test_utils_gitlab_api.py
lechuk47: answered above.

reconcile/test/test_utils_github_api.py
lechuk47: I think this concern applies in a more general python level, or any other languages that manage the code structures with indents (like nested loops, etc). 
Changed to a decorator flavour.

lechuk47:
> Looks great, I love the tests
> 
> Just nit-picking: can we make the timeouts a float, e.g. `0.1` so that they fail faster?

@rporres Changed in the gitlab test, github requires a integer so `1` is the minimum.

Piojo (approved):
This is awesome. Thanks.

rporres:
thx for taking a look @lechuk47 . This looks great, indeed

PR #1981, opened 2021-11-05
APPSRE-3784: Use permissions instead of roles

OliviaHY:
### Why:
Originally slack-usergroups integration query the roles and loop through it to update permission, which means when there is no role for a permission, it will be skipped. And we want to avoid that. So the proposal was to query permissions and iterate those instead, with the schema change to include the information about roles there.
### What:
- Update the query of permissions with the updated schema; 
- Use permissions instead of roles to generate desired state so no permission update will be skipped due to role not exist; 
- Fail integration instead of log warning when the user group is not in managedUsergroup.
- pull permission query to queries module
- Add tests
### Validation:
Dry-run success before adding the failing mechanism, which result in a failing for crc-bop-team. Asking what to do there right now.
Added tests and all tests passed.

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -2132,3 +2132,68 @@ GABI_INSTANCES_QUERY = """
 def get_gabi_instances():
     gqlapi = gql.get_api()
     return gqlapi.query(GABI_INSTANCES_QUERY)['gabi_instances']
+
+
+PERMISSIONS_QUERY = """
+{
+  permissions: permissions_v1 {
+    service
+    ...on PermissionSlackUsergroup_v1 {
+      channels
+      description
+      handle
+      ownersFromRepos
+      pagerduty {
+          name
+          instance {
+            name
+          }
+          scheduleID
+          escalationPolicyID
+        }
+      roles {
+        users {
+            name
+            org_username
+            slack_username
+            pagerduty_username
+        }
+    }
+      schedule {
+          schedule {
+            start
+            end
+            users {
+              org_username
+              slack_username
+            }
+          }
+        }
+      workspace {
+        name
+        token {
+          path
+          field
+        }
+        api_client {
+          global {
+            max_retries
+            timeout
+          }
+          methods {
+            name
+            args
+          }
+        }
+        managedUsergroups
+      }
+    }
+  }
+}
+"""
+
+
+def get_permissions_for_slack_usergroup():
+    gqlapi = gql.get_api()
+    permissions = gqlapi.query(PERMISSIONS_QUERY)['permissions']
+    return [p for p in permissions if p['service'] == 'slack-usergroup']
--- a/reconcile/test/test_slack_usergroups.py
+++ b/reconcile/test/test_slack_usergroups.py
@@ -1,13 +1,16 @@
 import copy
 from datetime import datetime, timedelta
 from unittest import TestCase
-from unittest.mock import create_autospec, call
+from unittest.mock import create_autospec, call, patch
 
 import pytest
 
 import reconcile.slack_usergroups as integ
 from reconcile.slack_usergroups import act
 from reconcile.utils.slack_api import SlackApi
+from reconcile import queries
+
+from .fixtures import Fixtures
 
 
 @pytest.fixture
@@ -85,6 +88,36 @@ class TestSupportFunctions(TestCase):
         }
         result = integ.get_pagerduty_name(user)
         self.assertEqual(result, 'pd')
+
+    @patch.object(integ, 'SlackApi', autospec=True)
+    @patch.object(queries, 'get_app_interface_settings', autospec=True)
+    @patch.object(queries, 'get_permissions_for_slack_usergroup',
+                  autospec=True)
+    def test_get_slack_map_return_expected(self, mock_get_permissions,
+                                           mock_get_app_interface_settings,
+                                           mock_slack_api):
+        mock_get_permissions.return_value = self.get_permissions_fixture()
+        slack_api_mock = create_autospec(SlackApi)
+        expected_slack_map = {
+            'coreos': {
+                'slack': slack_api_mock,
+                'managed_usergroups': [
+                    'app-sre-team',
+                    'app-sre-ic'
+                ]
+            }
+        }
+        result = integ.get_slack_map()
+        mock_slack_api.assert_called_once()
+        self.assertEqual(result['coreos']['managed_usergroups'],
+                         expected_slack_map['coreos']['managed_usergroups'])
+        self.assertIn('slack', result['coreos'])
+
+    @staticmethod
+    def get_permissions_fixture():
+        fxt = Fixtures('slack_usergroups')
+        permissions = fxt.get_anymarkup('permissions.yml')['permissions']
+        return [p for p in permissions if p['service'] == 'slack-usergroup']
 
 
 def test_act_no_changes_detected(base_state):
--- a/reconcile/slack_usergroups.py
+++ b/reconcile/slack_usergroups.py
@@ -4,7 +4,6 @@ from datetime import datetime
 from urllib.parse import urlparse
 from sretoolbox.utils import retry
 
-from reconcile.utils import gql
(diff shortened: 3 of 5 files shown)


reconcile/test/test_queris.py
```diff
@@ -0,0 +1,82 @@
...
+        }
+        managedUsergroups
+      }
+    }
+  }
+}
+"""
+
+
+class TestQueris(TestCase):
```
Piojo, line 68: Typo: `TestQueries`

reconcile/test/test_queris.py
```diff
@@ -0,0 +1,82 @@
...
+    @patch.object(gql, "get_api")
+    def test_get_permissions_return_all_slack_usergroup(self, get_api):
+        gqlapi_mock = create_autospec(gql.GqlApi)
+        gql.get_api.return_value = gqlapi_mock
+        gqlapi_mock.query(PERMISSIONS_QUERY).return_value = \
+            self.get_permissions_fixture()
+        result = queries.get_permissions()
+        self.assertTrue(all(x['service'] == 'slack-usergroup' for x in result))
```
Piojo, line 76: While I love the FP approach, in the event of bugs we won't see the bad results, how about:
```suggestion
        self.assertEqual({x['service'] for x in result}, {'slack-usergroup'})
```

reconcile/test/test_slack_usergroups.py
```diff
@@ -86,6 +89,35 @@ def test_get_pagerduty_username_slack_username(self):
...
+                    'app-sre-team',
+                    'app-sre-ic'
+                ]
+            }
+        }
+        result = integ.get_slack_map()
+        slackApi.assert_called_once()
+        self.assertEqual(result['coreos']['managed_usergroups'],
+                         expected_slack_map['coreos']['managed_usergroups'])
+        self.assertIn('slack', result['coreos'].keys())
```
Piojo, line 113: Nit: You don't need the `keys` argument here, which in the event of bugs will give a more detailed ourput:
```suggestion
        self.assertIn('slack', result['coreos'])
```

Piojo (approved):
A few nits, let's wait till Monday in case Maor has anything to add.

reconcile/test/test_queris.py
```diff
@@ -0,0 +1,82 @@
+from unittest import TestCase
+from unittest.mock import create_autospec, patch
+
+from reconcile import queries
+from reconcile.utils import gql
+
+from .fixtures import Fixtures
+
+PERMISSIONS_QUERY = """
```
steveteahan, line 9: Is this different from the `PERMISSIONS_QUERY` in `queries.py`? If so, probably worth adding a comment, otherwise can it just be imported?

reconcile/test/test_queris.py
```diff
@@ -0,0 +1,82 @@
...
+}
+"""
+
+
+class TestQueris(TestCase):
+    @patch.object(gql, "get_api")
+    def test_get_permissions_return_all_slack_usergroup(self, get_api):
+        gqlapi_mock = create_autospec(gql.GqlApi)
+        gql.get_api.return_value = gqlapi_mock
+        gqlapi_mock.query(PERMISSIONS_QUERY).return_value = \
```
steveteahan, line 73: I don't think this is doing what you expect it to do. Consider this example:

```
>>> from unittest.mock import MagicMock
>>> mock = MagicMock()

# What you're doing currently
>>> mock.some_method('test').return_value = 'some-value'
>>> mock.some_method()
<MagicMock name='mock.some_method()' id='139985539334496'>
>>> mock.some_method('test')
<MagicMock name='mock.some_method()' id='139985539334496'>

# Without any arg, you're actually setting the return value of the method
# (but it doesn't care about args)
>>> mock.some_method.return_value = 'some-value'
>>> mock.some_method('test')
'some-value'
>>> mock.some_method()
'some-value'
```

If you're trying to return a certain value based on a given input, you want `side_effect` instead. I'm a bit surprised that this would be working at all, so it might be worth taking a closer look (or telling me why I'm wrong).

reconcile/test/test_queris.py
```diff
@@ -0,0 +1,82 @@
...
+        managedUsergroups
+      }
+    }
+  }
+}
+"""
+
+
+class TestQueris(TestCase):
+    @patch.object(gql, "get_api")
```
steveteahan, line 69: I'd suggest you almost always want `autospec=True` anytime that you're using patch. See [here for more info](https://docs.python.org/3/library/unittest.mock.html#auto-speccing).

reconcile/test/test_slack_usergroups.py
```diff
@@ -86,6 +89,35 @@ def test_get_pagerduty_username_slack_username(self):
         result = integ.get_pagerduty_name(user)
         self.assertEqual(result, 'pd')
 
+    @patch.object(integ, 'SlackApi')
```
steveteahan, line 92: Same comment on `autospec` - I won't add comments for other occurrences, but please add it for all of them unless there's a good reason not to.

steveteahan (changes requested):
Looks good overall. There are some changes that you'll want to make on the testing side.

reconcile/test/test_queris.py
OliviaHY: While I'm trying what you suggested, I found out that the result is actually empty. Trying to figure out why. And the fact that this test didn't detect that proved exactly your point.

reconcile/test/test_queris.py
OliviaHY: It is the same! Thank you for point this out

reconcile/test/test_slack_usergroups.py
```diff
@@ -86,6 +89,35 @@ def test_get_pagerduty_username_slack_username(self):
         result = integ.get_pagerduty_name(user)
         self.assertEqual(result, 'pd')
 
+    @patch.object(integ, 'SlackApi', autospec=True)
+    @patch.object(queries, 'get_app_interface_settings', autospec=True)
+    @patch.object(queries, 'get_permissions', autospec=True)
+    def test_get_slack_map_return_expected(self, get_permissions,
```
steveteahan, line 95: nit: calling these `mock_<object_name>` makes things a bit clearer if testing methods get bigger. Maybe we can get that documented as a convention in the future, but you'll see the Python docs do similar.

reconcile/test/test_slack_usergroups.py
```diff
@@ -86,6 +89,35 @@ def test_get_pagerduty_username_slack_username(self):
...
+            'coreos': {
+                'slack': slack_api_mock,
+                'managed_usergroups': [
+                    'app-sre-team',
+                    'app-sre-ic'
+                ]
+            }
+        }
+        result = integ.get_slack_map()
+        slackApi.assert_called_once()
```
steveteahan, line 110: nit: snake case? A little bit odd that our linter doesn't pick up on that, but maybe disabled.

steveteahan (approved):

steveteahan:
Couple of final nits, but non-blocking.

reconcile/queries.py
```diff
@@ -2132,3 +2132,68 @@ def get_pagerduty_instances():
...
+    }
+  }
+}
+"""
+
+
+def get_permissions():
+    gqlapi = gql.get_api()
+    permissions = gqlapi.query(PERMISSIONS_QUERY)['permissions']
+    return [p for p in permissions if p['service'] == 'slack-usergroup']
```
BumbleFeng, line 2199: Since the func name is get_permissions. I suggest we return all permissions here and filter slack-usergroup permissions inside the integration.

OliviaHY: Well the query itself is tailored for getting those information that's related to slack_usergroup integration, i.e. toke, api_client, especially it is querying `PermissionSlackUsergroup_v1`.  There could be a use case that need to query permission with more information. How do you feel about renaming this function to get_permissions_for_slack_usergroup?

BumbleFeng: get_permissions_for_slack_usergroup works for me

PR #1984, opened 2021-11-08
Skip construct desired_state for app-sre-escalation-undefined usergroup

OliviaHY:
### Why
There is a [permission](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/teams/placeholder-escalation-permission.yml) defined as placeholder for app-sre-escalation-undefined usergroup. Now that we loop through permissions, the integration breaks when it run to it because the usergroup has no actual usergroup etc
### What:
Skipping it.

### DIFF
--- a/reconcile/slack_usergroups.py
+++ b/reconcile/slack_usergroups.py
@@ -235,6 +235,9 @@ def get_desired_state(slack_map, pagerduty_map):
     for p in permissions:
         if p['service'] != 'slack-usergroup':
             continue
+        skip_flag = p['skip']
+        if skip_flag:
+            continue
         workspace = p['workspace']
         managed_usergroups = workspace['managedUsergroups']
         if managed_usergroups is None:
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -2143,6 +2143,7 @@ PERMISSIONS_QUERY = """
       description
       handle
       ownersFromRepos
+      skip
       pagerduty {
           name
           instance {


reconcile/slack_usergroups.py
```diff
@@ -243,6 +243,8 @@ def get_desired_state(slack_map, pagerduty_map):
         workspace_name = workspace['name']
         usergroup = p['handle']
         description = p['description']
+        if usergroup == 'app-sre-escalation-undefined':
```
steveteahan, line 246: I'm not a huge fan of handling specific usergroups in different ways in our code. Is there anyway this could be done in a more general way that would make it obvious that it is ignored when looking at the resource from app-interface?

geoberle: this permission seems to be the only one with an empty array for the channel field. can you leverage that?

disclaimer: I don't know too much about permissions yet. so this is more of an observation than a real suggestion

OliviaHY: Good point, I did a quick filter in the code, and that empty channels does seem to be only this one. How does this sound to you @steveteahan ?

reconcile/slack_usergroups.py
```diff
@@ -243,6 +243,11 @@ def get_desired_state(slack_map, pagerduty_map):
         workspace_name = workspace['name']
         usergroup = p['handle']
         description = p['description']
+        channel_names = [] if p['channels'] is None else p['channels']
+        if channel_names == []:
```
steveteahan, line 247: You can just make this `if not channel_names` which is a pretty standard way to handle any falsy values in Python - `None`, `[]`, `0`, etc - unless you really care whether it's one of those specific types (sometimes we care about `None` to indicate something is unset).

Also, I don't even think that you need to make `channel_names` and empty list anymore because you're not passing it in anywhere that expects a list. As mentioned above, checking `if not <object>` will handle `None` just fine.

reconcile/slack_usergroups.py
```diff
@@ -243,6 +243,11 @@ def get_desired_state(slack_map, pagerduty_map):
...
+            # This is to get pass the app-sre-escalation-undefined
+            # usergroup in placeholder-escalation-permission.
```
steveteahan, line 249: nit: maybe worth adding a DEBUG log message here? It never hurts to have more information if someone is troubleshooting unexpected conditions.

steveteahan (commented):
A couple minor edits, but looks good otherwise!

reconcile/slack_usergroups.py
geoberle: +1

Piojo (approved):

geoberle:
LGTM

reconcile/slack_usergroups.py
```diff
@@ -237,6 +237,9 @@ def get_desired_state(slack_map, pagerduty_map):
             continue
         workspace = p['workspace']
         managed_usergroups = workspace['managedUsergroups']
+        skip_flag = p['skip']
```
maorfr, line 240: just a really stupid nit - if we're doing `continue` on this condition, we can probably move it two lines up to avoid the checking `workspace` and `managed_usergroups`

maorfr (approved):
LGTM

maorfr (approved):
LGTM!

PR #1986, opened 2021-11-09
Fail when namespaces list resource names or overrides of unmanaged types

Piojo:
It is a configuration error for a namespace to list entries in
`managedResourceNames` or `managedResourceTypeOverrides` of kinds that
are not in `managedResourceTypes`.

The first step here is to re-add #1901 - it was doing most of the work
and had all the tests.

Then, we fix the bug that forced [its revert](#1911), together with
better tests.  In this case, when a namespace lists
`managedResourceTypes` that have no match in `managedResourceNames` we
still want to initialize an `StateSpec` - we may be adding resources
of this kind without modifying or overriding their names.

Finally, instead of logging stuff we raise `KeyError`, to let the
integration handle the error.

TODO: Adjust schemas to forbid the combination of empty/non-existing
`managedResourceTypes` with non-empty
`managedResourceNames`/`managedResourceTypeOverrides`.

To test this, I removed the `Template` entry from
`managedResourceTypes` in the openshift-config namespace of one of our
clusters, which has `Template` among its `managedResourceNames` - got
the expected exception:

    KeyError: "Non-managed resource {'resource': 'Template', 'resourceNames': ['project-request']} listed on app-sre-prod-01/openshift-config"

and the integration crashes. Do we want the integration to crash on
this? Do we want to capture the exception somewhere?

Fixes APPSRE-3890.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -1,7 +1,7 @@
 import logging
 import itertools
 
-from typing import Optional, List, Iterable, Mapping
+from typing import Optional, Iterable, Mapping
 
 import yaml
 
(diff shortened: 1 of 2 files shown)


geoberle:
LGTM

maorfr (approved):
LGTM

thanks for chasing this down!

PR #1892, opened 2021-10-04
Add openshift-tekton-resources integration

rporres:
https://issues.redhat.com/browse/APPSRE-3389
Design doc: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/openshift-tekton-resources-integration.md

This work has been split in two

1) The new integration in this PR
2) The openshift-saas-deploy bits in #1998

We will first merge this one, verify that everything goes well and then merge the other.

## Related schema changes

See https://github.com/app-sre/qontract-schemas/pull/9

## Current testing includes:

* Creation/modification/deletion of Pipelines and Tasks
* Coexistence with current coexistence of openshift-resources managed objects
* Dry-run creation of openshit-saas-deploy related PipelineRuns
* Do not create PipelineRuns if no correspondent openshift-saas-deploy objects do not exist

## Current testing does not include

* Creation of PipelineRuns and check that deployments actually work, since it needs access to an up to date qontract-server

## Request and limits defaults

The variability is always in the openshift-saas-deploy step task. The rest consume a very low amount of resources. A minimum of 20Mb of memory has been assigned to avoid pod creation issues.

I have gathered a fair amount of openshift-saas-deploy metric pods examples that have been used to generate defaults. The outliers will need to be configured at a saas file level. Find examples in the related JIRA issue

## Method documentation

Since the new integration uses type hints, I haven't added parameter related documentation of the methods. Let me know if that's a problem and I will add it.

Signed-off-by: Rafa Porres Molina <rporresm@redhat.com>

EDIT: Split PR in two.

PR #1998, opened 2021-11-15
Openshift-tekton-resources: the saasherder part

rporres:
We have divided the work for openshift-tekton-resources in two parts:

1) The new integration in #1892
2) The saasherder bits here. They include logic to make sure that names
   for the resources created by the new integration are correctly
   referenced from openshift-saas-deploy integrations and dealing with a
   race condition on the new objects not yet existing (a recently
   created saas file).

This is part of APPSRE-3389


rporres:
weird... I'll take a look when the time comes.
```
FAIL Required test coverage of 27.23% not reached. Total coverage: 9.97%
```

PR #2001, opened 2021-11-16
Add onboarding status on MR as Label

andreugallofre:
As the number of apps managed by us and that the team is growing fast, a nice improvement would be to tag every MR on app interface with the onboarding status of the service just as we do with the tenant.


reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,48 @@
...
+    MR.
+    """
+    apps = queries.get_apps()
+    #   logging.info(apps)
+    app_name = get_app_name(changed_paths)
+    if app_name is not None:
+        app = None
+        for a in apps:
+            if a['name'] == app_name:
+                app = a
```
Piojo, line 26: You want to `break` here. Ideally you'd mark the `name` field with `isSearchable: true` in the schema, and save yourself this linear search.

reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,48 @@
...
+def get_app_name(changed_paths: List[str]) -> Optional[str]:
+    """
+    Guess the service name of a given MR from the changed paths of the
+    MR. This will allow to add the onboarding status to the MR's as label
+    """
+    for path in changed_paths:
+        if 'data/services/' in path:
+            path_slices = path.split('/')
+            try:
+                if path_slices[2] == "insights":
```
maorfr, line 44: i think we shouldn't hard code service names here

reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,48 @@
...
+    #   logging.info(apps)
+    app_name = get_app_name(changed_paths)
+    if app_name is not None:
+        app = None
+        for a in apps:
+            if a['name'] == app_name:
+                app = a
+
+    if app is not None:
+        logging.info(app['onboardingStatus'])
```
rporres, line 29: Is this a leftover? In any case, you could keep it with a nicer message in debug mode, not info

reconcile/gitlab_labeler.py
rporres: As the parent app pattern is not only on insights, you may find it in other places as ocm or rhobs.

I acknowledge that this is hard, but you must avoid hardcoding and try to come with a series of heuristics that can give you a hint of the service by inspecting the path and not by trusting in positions.

rporres: I've discussed a possible approach to fix this with @andreugallofre . The idea would be to query qontract-server to have an idea of which apps are children apps or parent apps or none. Having that it would be possible to know which positions are relevant. He will experiment with it.

reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,48 @@
...
+        return app['onboardingStatus']
+
+    return None
+
+
+def get_app_name(changed_paths: List[str]) -> Optional[str]:
```
steveteahan, line 35: `Iterable[str]` is more general here and appropriate for cases where you're just looping through a list, tuple, set, etc. See [this doc](https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#abstract-collection-types) for more info. For method parameters, concrete type usage is generally reserved for cases where you rely on the actual `list` implementation. Using concrete types is fine for method returns types.

reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,48 @@
 QONTRACT_INTEGRATION = 'gitlab-labeler'
 
 
+def guess_onboarding_status(changed_paths: List[str]) -> Optional[str]:
```
steveteahan, line 14: Similar to the comment on `get_app_name`, `Iterable[str]` might be better here.

steveteahan (commented):
This is great, the more IC work that we can automate the better!

reconcile/gitlab_labeler.py
```diff
@@ -45,6 +92,6 @@ def run(dry_run, gitlab_project_id=None, gitlab_merge_request_id=None):
         gl.get_merge_request_changed_paths(gitlab_merge_request_id)
     guessed_labels = guess_labels(project_labels, changed_paths)
     labels_to_add = [b for b in guessed_labels if b not in labels]
-    if labels_to_add:
+    if labels_to_add and not dry_run:
         logging.info(['add_labels', labels_to_add])
```
Piojo, line 96: Please, do log even under `dry_run` - that's the whole point of a dry run.

reconcile/gitlab_labeler.py
maorfr: what if the change set includes changes for several apps?

Piojo: Fair comment, then we should return `list[str]` instead of just `str`.

reconcile/gitlab_labeler.py
rporres: I understand that you're fixing a bug in the integration. While I agree 100% with this comment, you may want to address it in a separate PR

andreugallofre: This will be addressed in a new pull request 👍

reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,60 @@
...
+    Get a list of all services managed by app-interface with the parent app
+    and the onboarding status of the service.
+    """
+    apps = queries.get_apps()
+    app_info = {a['name']: {'onboardingStatus': a['onboardingStatus'],
+                'parentApp': a['parentApp']} for a in apps}
+    return app_info
+
+
+def get_parents_list() -> List[str]:
```
Piojo, line 25: Since we're using Python 3.9 it's OK to return `list[str]` here.

reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,60 @@
...
+                    if app is not None:
+                        labels.add(app['onboardingStatus'])
+                else:
+                    app = apps[path_slices[2]]
+                    if app is not None:
+                        labels.add(app['onboardingStatus'])
+                    else:
+                        return None
+            except (IndexError, KeyError):
+                logging.info("Error getting app name")
```
Piojo, line 61: Log `path` here, at least.

reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,60 @@
...
+                return None
+
+    labels_list = list(labels)
+    return labels_list[0] if len(labels_list) == 1 else None
```
Piojo, line 65: Easier:
```suggestion
    if len(labels) == 1:
        return labels.pop()
    else:
        return None
```

reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,60 @@
...
+                if path_slices[2] in parent_apps:
+                    app = apps[path_slices[3]]
+                    if app is not None:
+                        labels.add(app['onboardingStatus'])
+                else:
+                    app = apps[path_slices[2]]
+                    if app is not None:
+                        labels.add(app['onboardingStatus'])
+                    else:
+                        return None
```
Piojo, line 59: This method assumes you will only be processing one app, why do you continue iterating over all `changed_paths` instead of `break`-ing here?

reconcile/test/test_gitlab_labeler.py
```diff
@@ -0,0 +1,142 @@
...
+    def test_get_app_list(self):
+        self.test_data.apps = [self.app1]
+        t = gl.get_app_list()
+        assert t == {'normalapp': {'onboardingStatus': 'BestEffort',
+                     'parentApp': None}}
+
+    def test_get_app_list_2(self):
+        self.test_data.apps = [self.app1, self.app2]
+        t = gl.get_app_list()
+        assert t == {'normalapp': {'onboardingStatus': 'BestEffort',
```
Piojo, line 93: If you're using `unittest.TestCase` please do `self.assertEqual(t, {...})`

reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,60 @@
...
+                            apps: Dict[str, Dict], parent_apps:
+                            Set[str]) -> Optional[str]:
+    """
+    Guess the service name of a given MR from the changed paths of the
+    MR. This will allow to add the onboarding status to the MR's as label
+    """
+    labels = set()
+    for path in changed_paths:
+        if 'data/services/' in path:
+            path_slices = path.split('/')
```
maorfr, line 48: i recommend to assign the different slices to variables with meaningful names instead of using `path_slices[2]` (for example)

reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,60 @@
...
+            try:
+                if path_slices[2] in parent_apps:
+                    app = apps[path_slices[3]]
```
maorfr, line 51: no risk of getting a KeyError here?

reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,60 @@
...
+                    if app is not None:
+                        labels.add(app['onboardingStatus'])
+                else:
+                    app = apps[path_slices[2]]
```
maorfr, line 55: no risk of getting a KeyError here?

reconcile/gitlab_labeler.py
maorfr: i think this try-except block is too wide. IndexError can be handled above (see my comment about assigning to variables). KeyError can only happen in the `else`. you can be more defensive instead and check for indexes and keys. i'd also not use `info` (probably `debug`), as this means that we couldn't find an app, which can happen.

reconcile/gitlab_labeler.py
maorfr: there could be multiple apps involved in a MR, and thus multiple onboarding statuses?

reconcile/gitlab_labeler.py
```diff
@@ -30,6 +85,13 @@ def guess_labels(project_labels, changed_paths):
                        and t not in ignore_tokens]
             if matches:
                 guesses.add(label)
+    apps = get_app_list()
+    parent_apps = get_parents_list()
+    onboarding_status = guess_onboarding_status(changed_paths, apps,
+                                                parent_apps)
+    logging.info(onboarding_status)
```
maorfr, line 92: we are already logging in line 111, no need to log in internal functions. essentially we log only when the integration does something (add labels in this example).

maorfr:
i've added a few minor comments, but otherwise this looks good to me!

Piojo:
@andreugallofre what is the desired behaviour when a change affects multiple apps? (say, an automatic codemod)

reconcile/gitlab_labeler.py
andreugallofre: As part of this effort we are trying to add the information only when we know that is 100% correct. 

We are iterating over all the different changed paths because we don't want to add partial information, If there are two different applications but we only add a tag for one, the information is incomplete.

Also, we don't want (or that was the idea)  to have two different onboarding status labels on a given MR that's why we are only adding it in case we have one application.

reconcile/gitlab_labeler.py
andreugallofre: Addressed this in https://github.com/app-sre/qontract-reconcile/pull/2001#discussion_r760291999

reconcile/gitlab_labeler.py
andreugallofre: @Piojo Added above the information about what we thought about the expected behavior on a multiple app MR

Piojo: Mmm... in that case I do believe this method is doing too many things. Identifying the app for each path and assigning the onboarding status are different things, and separating them will yield more obvious code.

Can you try that small refactor?

reconcile/gitlab_labeler.py
```diff
@@ -30,6 +86,12 @@ def guess_labels(project_labels, changed_paths):
                        and t not in ignore_tokens]
             if matches:
                 guesses.add(label)
```
rporres, line 88: We discussed offline yesterday but writing it here for the benefit of the rest of the reviewers. I think you have to remove this old logic of finding apps. If not, we could end with many `tenant-insights` MRs with different onboarding-status, which is not what we want.

For those apps with parent app, I would include both labels, e.g. `tenant-insights` `tenant-rbac` so that it is clearer (especially since there are so many insights apps).

rporres:
Another aspect not covered here is the missing labels. If I understand how things work correctly, only existing labels will be used which leaves with the manual creation of labels. How about creating the labels if they do not exist so that we don't have to maintain manually this thing?

reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,61 @@
...
+
+    return list(parent_set)
+
+
+def guess_onboarding_status(changed_paths: Iterable[str],
+                            apps: Dict[str, Dict], parent_apps:
+                            Set[str]) -> Optional[str]:
+    """
+    Guess the service name of a given MR from the changed paths of the
+    MR. This will allow to add the onboarding status to the MR's as label
```
rporres, line 43: I'd add that in case that multiple apps are found, you will return None

reconcile/gitlab_labeler.py
```diff
@@ -18,18 +74,35 @@ def guess_labels(project_labels, changed_paths):
...
+        if t not in not_allowed_labels:
+            guesses.add("tenant-" + t)
+
+    for m in matches:
+        if m not in not_allowed_labels:
+            guesses.add(m)
+
+    onboarding_status = guess_onboarding_status(changed_paths, apps,
+                                                parent_apps)
+    if onboarding_status is not None:
```
rporres, line 104: nit: `guess_onboarding_status` is going return either a string or none... I guess you just need to do
```
if onboarding_status:
```

reconcile/gitlab_labeler.py
```diff
@@ -10,6 +11,61 @@
...
+                else:
+                    logging.debug("Error getting app name " + path)
+
+    if len(labels) == 1:
+        return labels.pop()
+    else:
+        return None
+
+
 def guess_labels(project_labels, changed_paths):
```
rporres, line 69: missing type hints

reconcile/test/test_gitlab_labeler.py
```diff
@@ -0,0 +1,142 @@
+from unittest import TestCase
```
rporres, line 1: we have agreed into writing tests with pytest. It looks like a lot of work but actually, it isn't... with very little effort you can transform this into a pure pytest test. See https://github.com/app-sre/qontract-reconcile/pull/2077 as example

reconcile/test/test_gitlab_labeler.py
```diff
@@ -0,0 +1,142 @@
...
+        self._apps = apps
+
+
+class TestOnboardingGuesser(TestCase):
+
+    def setUp(self) -> None:
+        self.test_data = TestData()
+
+        self.fxt = Fixtures('apps')
+        self.app1 = {'path': '/services/parent/normalapp/app.yml',
```
rporres, line 34: why don't you get these from the above yaml fixtures?

reconcile/test/test_gitlab_labeler.py
```diff
@@ -0,0 +1,142 @@
...
+        if query == APPS_QUERY:
+            return {'apps': self.test_data.apps}
+        else:
+            return None
+
+    def tearDown(self) -> None:
+        """ cleanup patches created in self.setUp"""
+        self.gql_patcher.stop()
+
+    def test_get_app_list(self):
```
rporres, line 84: missing type hints.

rporres (changes requested):
Looks great. A few comments here and there but this should be very close to be ready :)

reconcile/test/test_gitlab_labeler.py
andreugallofre: It's possible that I implemented something wrong, but with the fixtures I'm not capable of expanding the parent app of an app as the real graphql is doing, that's why I ended up using "hardcoded" apps

reconcile/gitlab_labeler.py
```diff
@@ -45,6 +119,14 @@ def run(dry_run, gitlab_project_id=None, gitlab_merge_request_id=None):
...
+
+    if labels_to_create:
+        logging.info(['create_labels', labels_to_create])
+        if not dry_run:
+            for label in labels_to_create:
+                gl.create_label(label, LABEL_COLOR)
+
     if labels_to_add:
         logging.info(['add_labels', labels_to_add])
         gl.add_labels_to_merge_request(gitlab_merge_request_id, labels_to_add)
```
rporres, line 132: missing `dry_run` here?

reconcile/test/test_gitlab_labeler.py
rporres: Then you will need to return exactly what qontract-server is returning for the query you're mocking, `APPS_QUERY`. In order to do that I get the exact json doc that q-s returns.

reconcile/test/fixtures/apps/app1.json
```diff
@@ -0,0 +1,21 @@
...
+    "codeComponents": [
+        {
+            "url": "https://github.com/test",
+            "resource": "upstream",
+            "gitlabRepoOwners": null,
+            "gitlabHousekeeping": null,
+            "jira": null
+        }
+    ]
+}
```
rporres, line 21: nit: missing newline at EOL

rporres (approved):

rporres:
since type hints are still optional, I'm not going to insist more :-P

reconcile/test/fixtures/apps/parentApp.yml
```diff
@@ -0,0 +1,26 @@
+---
```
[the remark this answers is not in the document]
andreugallofre: No, I added all the yml, but as it's not capable of combining them as the graphql I was not using it, better if I remove them?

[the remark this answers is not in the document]
rporres: yes, remove it if it's not used

PR #2005, opened 2021-11-18
APPSRE-3446: Improve SAAS triggers with config changes

lechuk47:
## Overview. 
When a SAAS deployment (SAAS-2) is dependant on the execution of another deployment (SAAS-1), it's required to update  the `REF` parameter on the SAAS-2 target to trigger its deployment. The flow is as follows:

```
SAAS-1-deployment --> Promotion(Updates REF on subscribed job manifests) --> Config Changes Triggers SAAS-2
```

This approach does not trigger SAAS-2 if a configuration update is made to SAAS-1 manifest as it does not update the SAAS-2 manifest with a new `REF`. 

## Proposed solution
To trigger subscribed jobs on parent configuration changes,  a new attribute  named `parent_config_hash` is added to the  promotion section. The `parent_config_hash` is a hash of the effective target configuration on SAAS-1. Adding this parameter, the AutoPromoter will change the SAAS-2 manifest with a new hash every time a configuration change happens in SAAS-1. 
Also, the promotion validation takes into account this parameter to ensure the parent job has succeed with the configuration expected. 

## Additional Notes
* The  config changes trigger method has been changed to be reused by this logic. 
* Tests added to ensure the logic works as expected
* I did some additional manual tests to ensure the new code does not trigger unintended deployments 
* Schemas PR: https://github.com/app-sre/qontract-schemas/pull/18


reconcile/utils/saasherder.py
```diff
@@ -1082,22 +1101,54 @@ def get_configs_diff(self):
...
+
         for rt in saas_file['resourceTemplates']:
             rt_name = rt['name']
             url = rt['url']
             path = rt['path']
             rt_parameters = rt.get('parameters')
-            for desired_target_config in rt['targets']:
+            for v in rt['targets']:
+                # Create a copy to avoid modifying saas_file values
+                desired_target_config = copy.deepcopy(v)
```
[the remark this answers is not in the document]
lechuk47: I've modified the `deepcopy` to use `chainmap`. I think it's not conceptually easier but I feel a lot more clever implementation. Thanks for the hint.

Piojo (commented):
LGTM, but as we've discussed over Slack, we'll give some time for people with better knowledge in saasherder to chime in.

maorfr:
cc @geoberle

lechuk47:
The config_sha approach has problems with some possible configurations. Consider the next example.

- Just 1 SAAS file to simplify (deploy and tests)
- Stage deploy publishes state to channel C
- Int deploy publishes state to C'
- State and Int starts at the same time.
- Prod deploy is subscribed to C and C'
- Int deploy takes longer than stage deploy (due e2e tests or whatever)

Right now with the current approach based on REF updates this is what happens:

- Stage and Integration jobs start
- Stage ends with success --> publishes State to C --> Promotes Prod (PR1) --> PR1 fails at PR_CHECK because REF is not published at C'
- Integration ends with success --> publishes State to C1 --> Promotes Prod(PR2) --> PR2 PR_CHECK ok --> PR2 merged --> Prod Deployment --> PR1 has no changes.
- This ends up with an empty PR but the flow is correct.
- There is a chance of PR1 getting merged instead of PR2 as this PRs checks are continuously running with new changes introduced by the bot. 


With the config_sha approach and validating the config_hash at PR_CHECK, if the same flow starts by a config change in the SAAS file, PR1 and PR2 will only have the new config_hash introduced by its target deployment:
PR1 --> new config hash of Stage target
PR2 --> new config hash of Int deployment
So PR1 will not fail checking the state of C', because REF and the config_sha for C' are not updated and the state is already published with the old values.

So to continue with this approach, we need a way to ensure PR1 is aware of the incoming changes on C'.

reconcile/utils/saasherder.py
```diff
@@ -1228,7 +1327,9 @@ def publish_promotions(self, success, saas_files, mr_cli):
                     # publish to state to pass promotion gate
                     state_key = f"promotions/{channel}/{commit_sha}"
                     value = {
-                        'success': success
+                        'success': success,
+                        'saas_file': item["saas_file"],
+                        TARGET_CONFIG_HASH: item.get(TARGET_CONFIG_HASH)
```
geoberle, line 1332: nit: this could be outside of the loop since `value` does not depend on the channel

reconcile/utils/mr/auto_promoter.py
```diff
@@ -34,6 +44,47 @@ def title(self):
...
+        # Existent subscribe data channel data
+        promotion_data = {v["channel"]: v["data"]
+                          for v in target_promotion.get('promotion_data', [])
+                          if v["channel"] in target_channels}
+
+        for channel in target_channels:
+            modified = False
+            channel_data = promotion_data.get(channel)
+            if channel_data is None:
+                channel_data = \
```
geoberle, line 75: i might miss or misunderstand something here. shouldn't the channel_data be applied back to the target_promotion somehow?

reconcile/utils/mr/auto_promoter.py
```diff
@@ -34,6 +44,47 @@ def title(self):
...
+                    AutoPromoter.init_promotion_data(channel, promotion)
+                modified = True
+            else:
+                for item in channel_data:
+                    if item["type"] == ParentSaasConfigPromotion.TYPE:
+                        psc = ParentSaasConfigPromotion(**item)
+                        if psc.target_config_hash != \
+                                promotion[TARGET_CONFIG_HASH]:
+                            modified = True
```
geoberle, line 84: shouldn't there be some target_config modification going on here?

reconcile/utils/saasherder.py
lechuk47: you are right, fixed.

reconcile/utils/mr/auto_promoter.py
```diff
@@ -34,6 +44,47 @@ def title(self):
...
+
+            return modified
```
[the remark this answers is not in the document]
lechuk47: moved `modified`variable outside of the loop.

reconcile/utils/mr/auto_promoter.py
lechuk47: oh my, you are completely right. I fixed it and extend the tests to cover this

reconcile/utils/mr/auto_promoter.py
lechuk47: same as the other comment, don't know what I was thinking when I did this ...

geoberle (approved):
LGTM

PR #2008, opened 2021-11-23
[terraform-resources] support openshift service for alb

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4107

depends on https://github.com/app-sre/qontract-schemas/pull/20

This allows for setting an OpenShift Service name as the target for an ALB.
With `openshift_service` defined, the integration will search the cluster's AWS account for a load balancer matching the OpenShift service, find its network interfaces and get their private IP (instead of manually doing it and hard coding the IP).

it is strongly recommended to review commit by commit.

### DIFF
--- a/reconcile/utils/aws_api.py
+++ b/reconcile/utils/aws_api.py
@@ -8,7 +8,6 @@ from datetime import datetime
 from threading import Lock
 from typing import TYPE_CHECKING
 from typing import Any, Dict, Iterable, List, Mapping, Optional, Tuple
-from typing_extensions import Literal
 
 from boto3 import Session
 from sretoolbox.utils import threaded
--- a/reconcile/utils/aws_api.py
+++ b/reconcile/utils/aws_api.py
@@ -8,6 +8,7 @@ from datetime import datetime
 from threading import Lock
 from typing import TYPE_CHECKING
 from typing import Any, Dict, Iterable, List, Mapping, Optional, Tuple
+from typing_extensions import Literal
 
 from boto3 import Session
 from sretoolbox.utils import threaded
@@ -716,7 +717,7 @@ class AWSApi:
     # pylint: disable=method-hidden
     def _get_assumed_role_client(self, account_name: str, assume_role: str,
                                  assume_region: str,
-                                 client_type: str = 'ec2') -> EC2Client:
+                                 client_type='ec2') -> EC2Client:
         assumed_session = self._get_assume_role_session(account_name,
                                                         assume_role,
                                                         assume_region)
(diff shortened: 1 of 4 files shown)


reconcile/utils/aws_api.py
```diff
@@ -715,11 +715,12 @@ def _get_assume_role_session(self, account_name: str, assume_role: str,
...
     # pylint: disable=method-hidden
     def _get_assumed_role_client(self, account_name: str, assume_role: str,
-                                 assume_region: str) -> EC2Client:
+                                 assume_region: str,
+                                 client_type: str = 'ec2') -> EC2Client:
         assumed_session = self._get_assume_role_session(account_name,
                                                         assume_role,
                                                         assume_region)
-        return assumed_session.client('ec2')
+        return assumed_session.client(client_type)
```
maorfr, line 723:
```
reconcile/utils/aws_api.py:723: error: No overload variant of "client" of "Session" matches argument type "str"
```

maorfr: removed the type hint to get rid of this (https://coreos.slack.com/archives/GGC2A0MS8/p1637668588271300)

reconcile/utils/terrascript_client.py
```diff
@@ -3374,7 +3378,33 @@ def populate_tf_resource_s3_cloudfront_public_key(self, resource,
...
+        account['assume_role'] = \
+            ocm.get_aws_infrastructure_access_terraform_assume_role(
+                cluster['name'],
+                account['uid'],
+                account['terraformUsername'],
+            )
+        account['assume_region'] = cluster['spec']['region']
+        service_name = \
+            f"{namespace_info['name']}/{openshift_service}"
+        return awsapi.get_alb_network_interface_ips(
```
jfchevrette, line 3401: Should we add a safeguard to verify that `get_alb_network_interface_ips` did not return an empty set? Services should always have at least 1 IP address (single-az) and I'm worried that if one day the tags change or maybe if the AWS API call is rate limited or something else that doesn't raise an exception we could end up removing all IP addresses from the ALB

geoberle (approved):

geoberle:
lgtm

reconcile/utils/terrascript_client.py
maorfr: handled in https://github.com/app-sre/qontract-reconcile/pull/2008/commits/b0e9d09383fe9f79c0488e3186840bef984de6e9

jfchevrette (approved):
LGTM

PR #2009, opened 2021-11-23
[terraform] add support for always enabled deletions for specific resource types

maorfr:
follow up on:
- #1992
- #2008

with this PR, resources that are expected to get re-created will not cause an error when attempted to be deleted.

### DIFF
--- a/reconcile/utils/terraform_client.py
+++ b/reconcile/utils/terraform_client.py
@@ -184,6 +184,11 @@ class TerraformClient:
         if resource_changes is None:
             return disabled_deletion_detected, deleted_users
 
+        always_enabled_deletions = {
+            'random_id',
+            'aws_lb_target_group_attachment',
+        }
+
         # https://www.terraform.io/docs/internals/json-format.html
         for resource_change in resource_changes:
             resource_type = resource_change['type']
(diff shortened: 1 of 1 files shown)


reconcile/utils/terraform_client.py
```diff
@@ -184,6 +184,11 @@ def log_plan_diff(self, name, tf, enable_deletion):
         if resource_changes is None:
             return disabled_deletion_detected, deleted_users
 
+        always_enabled_deletions = {
+            'random_id',
+            'aws_lb_target_group_attachment',
+        }
```
jfchevrette, line 190: we need to add `aws_lb_target_group` to properly address the specific case this is trying to address as the group is being re-created when the list of IPs change

jfchevrette: probably better: if we remove [this](https://github.com/app-sre/qontract-reconcile/blob/7cf76d53a5d1b09aa8a503eaa2fdbc0041b16e83/reconcile/utils/terrascript_client.py#L3531) we don't need to add `aws_lb_target_group` as the IPs apparently can be added/removed without re-creating the group

jfchevrette: I will submit a different MR to address this

jfchevrette (approved):
LGTM

PR #2011, opened 2021-11-25
add integration for status page component management

geoberle:
### Overview
the new integration can create, update and delete status page components based on the qontract-schemas `dependencies/status-page-1.yml` and `dependencies/status-page-component-1.yml``

### atlassian page provider
The integration is currently implementing page management for Atlassian status pages (statuspage.io). Other status page provided can be added as necessary. The provider of a page is defined in the `provider` field of `dependencies/status-page-1.yml`

### pydantic
I wanted to point out the use of pydantic and dataclasses since this integration is also a POC for that approach. Please note the following:
* Field aliases are used to bridge the camelCase world of schemas with the snake_case world of python

Please also note that VaultSecretRef is not using pydantic but plain python dataclasses. I've chosen this approach to isolate the experiment with pydantic to the status_page_components.py file. 

### VaultSecretRef
I was playing a bit with consuming vault secrets a bit more conveniently directly from a declaration in a schema. So have a look at vaultsecretref.py and how it is used in status_page_component.py#StatusPage

### DIFF
--- a/reconcile/status_page_components.py
+++ b/reconcile/status_page_components.py
@@ -83,9 +83,11 @@ class StatusPage(BaseModel):
 
     def get_page_provider(self) -> StatusPageProvider:
         if self.provider == "atlassian":
-            return AtlassianStatusPage(page_id=self.page_id,
-                                       api_url=self.api_url,
-                                       token=self.credentials.get("token"))
+            provider = AtlassianStatusPage(page_id=self.page_id,
+                                           api_url=self.api_url,
+                                           token=self.credentials.get("token"))
+            provider.rebuild_state()
+            return provider
         else:
             raise ValueError(f"provider {self.provider} is not supported")
 
@@ -139,11 +141,7 @@ class AtlassianStatusPage(StatusPageProvider):
     components_by_displayname: dict[str, AtlassianComponent] = {}
     group_name_to_id: dict[str, str] = {}
 
-    def __init__(self, **data):
-        super().__init__(**data)
-        self._rebuild_state()
-
-    def _rebuild_state(self):
+    def rebuild_state(self):
         components = self._fetch_components()
         self.components_by_id = {c.id: c for c in components}
         self.components_by_displayname = {c.name: c for c in components}
@@ -215,7 +213,7 @@ class AtlassianStatusPage(StatusPageProvider):
     def delete_component(self, dry_run: bool, id: str) -> None:
         if not dry_run:
             self._client().components.delete(id)
-            self._rebuild_state()
+            self.rebuild_state()
 
     def _fetch_components(self) -> list[AtlassianComponent]:
         raw_components = self._client().components.list()
--- a/reconcile/status_page_components.py
+++ b/reconcile/status_page_components.py
@@ -4,7 +4,7 @@ from typing import Optional, Any
 from collections.abc import Iterable
 import sys
 
-from pydantic import BaseModel, Field, validate_arguments
+from pydantic import BaseModel, Field
 from pydantic.networks import HttpUrl
 import statuspageio  # type: ignore
 
--- a/reconcile/status_page_components.py
+++ b/reconcile/status_page_components.py
@@ -1,0 +1,261 @@
+from abc import abstractmethod
+import logging
+from typing import Optional, Any
+from collections.abc import Iterable
+import sys
+
+from pydantic import BaseModel, Field, validate_arguments
+from pydantic.networks import HttpUrl
+import statuspageio  # type: ignore
+
+from reconcile import queries
+from reconcile.utils.semver_helper import make_semver
+from reconcile.utils.state import State
+from reconcile.utils.vaultsecretref import VaultSecretRef
+
+
+QONTRACT_INTEGRATION = "status-page-components"
+QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
+
+
+LOG = logging.getLogger(__name__)
+
+
+class AtlassianComponent(BaseModel):
+    """
+    atlassian status page REST schema for component
+    """
+
+    id: str
+    name: str
+    description: Optional[str]
+    position: int
+    status: str
+    automation_email: Optional[str]
+    group_id: Optional[str]
+    group: Optional[bool]
+    group_name: Optional[str]
+
+
+class StatusComponent(BaseModel):
+    """
+    app-interface schema dependencies/status-page-component-1
+    """
+
+    name: str
+    display_name: str = Field(..., alias="displayName")
+    description: Optional[str]
+    group_name: Optional[str] = Field(..., alias="groupName")
+    component_id: Optional[str]
+
+
+class StatusPageProvider(BaseModel):
+    """
+    Provider specific status page reconcile implementation is implemented
+    as a subclass of `StatusPageProvider`
+    """
+
+    @abstractmethod
+    def component_ids(self) -> Iterable[str]:
+        return []
+
+    @abstractmethod
+    def apply_component(self, dry_run: bool,
+                        desired: StatusComponent) -> Optional[str]:
+        return None
+
+    @abstractmethod
+    def delete_component(self, dry_run: bool, id: str):
+        return None
+
+
+class StatusPage(BaseModel):
+    """
+    app-interface schema dependencies/status-page-1
+    """
+
+    name: str
+    page_id: str = Field(..., alias="pageId")
+    api_url: HttpUrl = Field(..., alias="apiUrl")
+    credentials: VaultSecretRef
+    components: list[StatusComponent]
+    provider: str
+
+    def get_page_provider(self) -> StatusPageProvider:
+        if self.provider == "atlassian":
+            return AtlassianStatusPage(page_id=self.page_id,
+                                       api_url=self.api_url,
+                                       token=self.credentials.get("token"))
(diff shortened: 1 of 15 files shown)


lechuk47:
I like the approach to manage the data with model classes, but should we consider to move model classes out of reconcile?

geoberle:
> I like the approach to manage the data with model classes, but should we consider to move model classes out of reconcile?

that boils down to the question of fat models vs thin models, so having logic in the model or not. a fat model is tightly bound to the integration itself and it would not make sense to extract it.

Personally, I use both styles and choose what fits the situation. in this case I've decided on fat models because in that way it is clear what functions belong to the general flow of control (StatusPage) and which ones are specific for Atlassian status pages (AtlassianStatusPage).

I also don't see those model classes as something compared to what is usually done in the ORM field. I think it is absolutely legit to have multiple models for various schemas, each one specifically designed to serve the purpose of the integration at hand.

I think there are definitely cases where a model is used in enough integrations that it makes sense to build them as thin models, though.

So TLDR - yes, no, maybe :)

reconcile/status_page_components.py
```diff
@@ -0,0 +1,261 @@
...
+    def component_ids(self) -> Iterable[str]:
+        return []
+
+    @abstractmethod
+    def apply_component(self, dry_run: bool,
+                        desired: StatusComponent) -> Optional[str]:
+        return None
+
+    @abstractmethod
+    def delete_component(self, dry_run: bool, id: str):
```
Piojo, line 68: Make the return type explicit here:
```suggestion
    def delete_component(self, dry_run: bool, id: str) -> None:
```

reconcile/status_page_components.py
```diff
@@ -0,0 +1,261 @@
...
+            # if the component is known to the state management and if it is
+            # not known to the desired state, it was once managed by this
+            # integration but was delete from app-interface -> delete from page
+            name_for_current_component = id_to_name_state.get(current_id)
+            if (name_for_current_component and
+               name_for_current_component not in desired_component_names):
+                LOG.info(f"delete component {name_for_current_component} "
+                         f"from page {self.name}")
+                page_provider.delete_component(dry_run, current_id)
+                state.rm(name_for_current_component)
```
Piojo, line 113: Do you want to modify the state even under `dry_run`?

reconcile/status_page_components.py
```diff
@@ -0,0 +1,261 @@
...
+    api_url: str
+    token: str
+
+    components_by_id: dict[str, AtlassianComponent] = {}
+    components_by_displayname: dict[str, AtlassianComponent] = {}
+    group_name_to_id: dict[str, str] = {}
+
+    def __init__(self, **data):
+        super().__init__(**data)
+        self._rebuild_state()
```
Piojo, line 143: Can we rebuild the state outside of `__init__`? I see this ends up calling `self._client()` which means the constructor will have side effects, which makes modification and testing harder. We document it [in the style guide](https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#avoid-side-effects-in-constructors) and yes, I know, we have a lot of code breaking this convention but this is a good time to move to a better style.

reconcile/status_page_components.py
```diff
@@ -0,0 +1,261 @@
...
+    return State(integration=QONTRACT_INTEGRATION,
+                 accounts=accounts,
+                 settings=settings)
+
+
+def run(dry_run: bool = False):
+    state = get_state()
+    status_pages = fetch_pages()
+
+    page_reconcile_error_occured = False
```
Piojo, line 252: `error` is fine IMO. It's also more consistent with the naming in other integrations.

Piojo (commented):
Nothing terrible stands out.

Question about pydantic: which is the source of truth here? The schemas in the qontract-schemas repository? Or this one? Can we have one auto-generate the other?

reconcile/status_page_components.py
geoberle: oh dear good catch. i will fix this

reconcile/status_page_components.py
geoberle: fixed

reconcile/status_page_components.py
geoberle: fixed

reconcile/status_page_components.py
geoberle: fixed

reconcile/status_page_components.py
geoberle: fixed

geoberle:
rebased to incorporate the pylint 2.12.x fixes in master

geoberle:
> Nothing terrible stands out.
> 
> Question about pydantic: which is the source of truth here? The schemas in the qontract-schemas repository? Or this one? Can we have one auto-generate the other?

to me the schema is the source of truth and the model just a reflection of it in the code. or even: the model is a companion of the query. code generating sounds intriguing though.

lechuk47:
> > I like the approach to manage the data with model classes, but should we consider to move model classes out of reconcile?
> 
> that boils down to the question of fat models vs thin models, so having logic in the model or not. a fat model is tightly bound to the integration itself and it would not make sense to extract it.
> 
> Personally, I use both styles and choose what fits the situation. in this case I've decided on fat models because in that way it is clear what functions belong to the general flow of control (StatusPage) and which ones are specific for Atlassian status pages (AtlassianStatusPage).
> 
> I also don't see those model classes as something compared to what is usually done in the ORM field. I think it is absolutely legit to have multiple models for various schemas, each one specifically designed to serve the purpose of the integration at hand.
> 
> I think there are definitely cases where a model is used in enough integrations that it makes sense to build them as thin models, though.
> 
> So TLDR - yes, no, maybe - this is still fresh - we need to figure it out

I was thinking in a wider approach, for example, this status page models could be reused in `qontract-cli` for quering the api or other purposes, not just for `qontract-reconcile`. I also think that the "ideal state" for all this is the auto-generation of model claases from `qontract-schemas` so having the logic separated from the model will facilitate that if we get there in the future. 

Finally, and this could be seen as a personal preference, I think that keeping the the single responsibility for classes is always desirable and somehow here the logic and the data representation are mixed in the same object/class. 

Don't take this as a stopper, just wanted to express my point of view, if you think this is a better approach for this purpose will be fine for me.

geoberle:
> I was thinking in a wider approach, for example, this status page models could be reused in `qontract-cli` for quering the api or other purposes, not just for `qontract-reconcile`. I also think that the "ideal state" for all this is the auto-generation of model claases from `qontract-schemas` so having the logic separated from the model will facilitate that if we get there in the future.
> 
> Finally, and this could be seen as a personal preference, I think that keeping the the single responsibility for classes is always desirable and somehow here the logic and the data representation are mixed in the same object/class.
> 
> Don't take this as a stopper, just wanted to express my point of view, if you think this is a better approach for this purpose will be fine for me.

resue is good because it reduces the toil of writing the same thing over and over again with the potential of human error. still, I'm not a huge fan of 100% DRY for models. models always need to be seen in the context of their use. also in qontract-reconciles codebase, the same schema object is often queried differently (fetching only a subset of fields or following relations extensively).

That said, I completely agree that manually writing the same model object over and over again is definitely something to be avoided. auto-generation is a good idea and also @Piojo and @steveteahan suggested that. I'm intrigued by the idea to generate a model based on a query instead of the schema though. that way we can ensure that the model always fits the intended use. this does not imply that every use of a model should result in a dedicated query and auto-generated companion model, but I would prefer to have the possibility to have a model generated more than once depending on context and use case.

I understand why you prefer thin models. Thick models make more sense when you really want to leverage some of the features from the OOP paradigm, e.g. polymorphism. For this new integration, it made sense because I wanted to leave the door open for other status page providers where the implementation might differ. For other integrations, a thin model might be the perfectly right choice. auto-generation and thick models can be combined quite well, e.g. by generating a thin model and then subclassing it to add the functionality that should be bound to the class. So I don't see any conflict in having both.

Let's try another angle: Let's say we have an autogenerated model from the schema that contains all fields and relations so we can a schema type represented by just one model class. The perfect DRY. I'm unsure how we could handle optional and mandatory fields when we want to keep writing queries that return only the required amount of data. We could define all fields as optional to allow arbitrary queries but that defeats the purpose of having those models as a verification layer.

Please consider all I'm writing as stuff that swirls around in my head. I have not set my mind on anything yet. So let's keep brainstorming. For the matter of this new integration, I would prefer not to change anything in that regard within this PR. auto-generation or rewriting to thin models can be done in follow-up work.

geoberle:
or maybe i'm just overreacting on having queries minimal and not requesting any fields that are not required for the job at hand. a valid solution could also be to always fetch all fields so models can really act as a verification layer. relations could be optional.

i'm going to start a design-doc about this where we can dive into questions like
* source of truth for schemas
 * generating models from schemas vs. generating schemas from models
 * how would thinking about source of truth be affected by the thought of not having all integrations in qontract-reconcile and not written in python
* thin vs thick vs mixed models
* full models vs sparse model (having all fields in the model and querying them all)

maorfr (approved):
LGTM

as noted above, every piece of this PR can be reverted or updated according to any future decision we make. i propose we proceed with this one to enable the functionality for our tenants and take the discussion offline.

rporres:
Late to the party but this also LGTM. I agree with @maorfr statement

PR #2015, opened 2021-11-28
[terraform-resources] only add ready network interfaces to ALB private IPs

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4117

this PR adds logic to skip "unready" network interfaces from being added to an ALB's IP list.


patjlm (approved):
LGTM

PR #2016, opened 2021-11-28
[terraform-users] add support to define alternate aws username

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4122

depends on https://github.com/app-sre/qontract-schemas/pull/21

this PR adds support to define `aws_username` in a user file to create aws IAM users with user name different from `org_username`.


geoberle (approved):

geoberle:
LGTM

PR #2017, opened 2021-11-28
[terraform-resources] consider elasticsearch_version

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4123

this field is currently not obtained from the `elasticsearch` provider defaults file. let's use it and remove any default values.

### DIFF
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -94,7 +94,7 @@ VARIABLE_KEYS = ['region', 'availability_zone', 'parameter_group',
                  'variables', 'policies', 'user_policy',
                  'es_identifier', 'filter_pattern',
                  'specs', 'secret', 'public', 'domain',
-                 'aws_infrastructure_access']
+                 'aws_infrastructure_access', 'elasticsearch_version']
 
 
 class UnknownProviderError(Exception):
@@ -3027,7 +3027,7 @@ class TerrascriptClient:
     @staticmethod
     def validate_elasticsearch_version(version):
         """ Validate ElasticSearch version. """
-        return version in [7.7, 7.4, 7.1,
+        return version in [7.10, 7.9, 7.8, 7.7, 7.4, 7.1,
                            6.8, 6.7, 6.5, 6.4, 6.3, 6.2, 6.0,
                            5.6, 5.5, 5.3, 5.1,
                            2.3,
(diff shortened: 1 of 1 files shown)


reconcile/utils/terrascript_client.py
```diff
@@ -3071,11 +3071,11 @@ def populate_tf_resource_elasticsearch(self, resource, namespace_info):
                 ", and - (hyphen). " +
                 f"{values['identifier']}")
 
-        elasticsearch_version = values.get('elasticsearch_version', 7.7)
+        elasticsearch_version = values.get('elasticsearch_version')
```
romfreiman, line 3074: no default? in TF it's an optional parameter.
On the other hand, better to fail if you expect all apps-sre customers to specify it.

reconcile/utils/terrascript_client.py
```diff
@@ -3027,7 +3027,7 @@ def get_dependencies(tf_resources: Iterable[Resource]
     @staticmethod
     def validate_elasticsearch_version(version):
         """ Validate ElasticSearch version. """
-        return version in [7.7, 7.4, 7.1,
+        return version in [7.10, 7.9, 7.8, 7.7, 7.4, 7.1,
```
romfreiman, line 3030: isnt it better to have it as a global env?

maorfr: its better not to have it here at all :)

geoberle (approved):

geoberle:
LGTM

reconcile/utils/terrascript_client.py
BumbleFeng: We need to skip 
```
es_values["elasticsearch_version"] = elasticsearch_version
```
if elasticsearch_version is none.

reconcile/utils/terrascript_client.py
```diff
@@ -94,7 +94,7 @@
                  'variables', 'policies', 'user_policy',
                  'es_identifier', 'filter_pattern',
                  'specs', 'secret', 'public', 'domain',
-                 'aws_infrastructure_access']
+                 'aws_infrastructure_access', 'elasticsearch_version']
```
BumbleFeng, line 97: We only add `elasticsearch_version` in the default file. These keys are used for values comes form gql queries, so we don't need to add `elasticsearch_version` here.

PR #2018, opened 2021-11-29
[terrascript] init log to ES zip only lazy and once

maorfr:
this PR changes the initiation of the lambda zip to only be initiated when needed and to only be initiated once.
we use a lock because there is concurrency.

the implementation of hard coding `app-sre` is wrong, but this is not what this change is about.

### DIFF
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -190,12 +190,12 @@ class TerrascriptClient:
                                 for a in filtered_accounts}
         self.partitions = {a['name']: a.get('partition') or 'aws'
                            for a in filtered_accounts}
-        github_config = get_config()['github']
-        self.token = github_config['app-sre']['token']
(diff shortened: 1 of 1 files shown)


reconcile/utils/terrascript_client.py
```diff
@@ -190,13 +190,17 @@ def __init__(self, integration, integration_prefix,
...
         self.partitions = {a['name']: a.get('partition') or 'aws'
                            for a in filtered_accounts}
-        github_config = get_config()['github']
-        self.token = github_config['app-sre']['token']
         self.logtoes_zip = ''
 
     def get_logtoes_zip(self, release_url):
         if not self.logtoes_zip:
-            self.logtoes_zip = self.download_logtoes_zip(LOGTOES_RELEASE)
+            with Lock():
```
BumbleFeng, line 197: We need a global lock here.

PR #2021, opened 2021-12-01
add tf provider secrets-manager

BumbleFeng:

depends on https://github.com/app-sre/qontract-schemas/pull/24

### DIFF
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -3733,10 +3733,6 @@ class TerrascriptClient:
         tf_resources.append(aws_version_resource)
 
         # outputs
-        for k, v in secret_data.items():
-            output_name_0_13 = output_prefix + '__' + k
-            output_value = v
-            tf_resources.append(Output(output_name_0_13, value=output_value))
         output_name_0_13 = output_prefix + '__arn'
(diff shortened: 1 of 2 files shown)


reconcile/utils/terrascript_client.py
```diff
@@ -3691,3 +3695,54 @@ def populate_tf_resource_alb(self, resource, namespace_info,
...
+
+        aws_version_resource = \
+            aws_secretsmanager_secret_version(identifier, **version_values)
+        tf_resources.append(aws_version_resource)
+
+        # outputs
+        for k, v in secret_data.items():
+            output_name_0_13 = output_prefix + '__' + k
+            output_value = v
+            tf_resources.append(Output(output_name_0_13, value=output_value))
```
maorfr, line 3739: why do you need to keep the secret information in terraform state?

BumbleFeng: The intention is to send secret information to the namespace, in case the service can get it in cluster.

maorfr: if a service is running in a cluster, it is not likely to need to consume a secret from aws. this will be the case for openshift-vault-secrets. i'm not against this, but i do propose that we add this only when we find that it is needed. leaving it to you.

maorfr (approved):

PR #2023, opened 2021-12-03
Make the openshift-tekton-resources objects be used by default

rporres:
The pipelines used from now on will be those created from
openshift-tekton-resources

APPSRE-3389


Piojo (approved):
Simpler code is better code. :)

PR #2026, opened 2021-12-06
[qontract-cli] add aws-creds get subcommand

maorfr:
similar to `qontract-cli get bot-login`, but for AWS accounts. this will result in the credentials of the terraform user.
these creds should be used in case someone can't access their own creds or aws account and a swift action is required.

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -484,6 +484,25 @@ def bot_login(ctx, cluster_name):
 
 
 @get.command()
+@click.argument('account_name')
+@click.pass_context
+def aws_creds(ctx, account_name):
+    settings = queries.get_app_interface_settings()
+    secret_reader = SecretReader(settings=settings)
+    accounts = queries.get_aws_accounts()
+    accounts = [a for a in accounts if a['name'] == account_name]
+    if len(accounts) == 0:
+        print(f"{account_name} not found.")
+        sys.exit(1)
+
+    account = accounts[0]
+    secret = secret_reader.read_all(account['automationToken'])
+    print(f"export AWS_REGION={account['resourcesDefaultRegion']}")
+    print(f"export AWS_ACCESS_KEY_ID={secret['aws_access_key_id']}")
+    print(f"export AWS_SECRET_ACCESS_KEY={secret['aws_secret_access_key']}")
+
+
+@get.command()
 @click.argument('name', default='')
 @click.pass_context
 def namespaces(ctx, name):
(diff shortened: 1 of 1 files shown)


tools/qontract_cli.py
```diff
@@ -478,6 +478,25 @@ def bot_login(ctx, cluster_name):
...
 
+@get.command()
+@click.argument('account_name')
+@click.pass_context
+def aws_creds(ctx, account_name):
+    settings = queries.get_app_interface_settings()
+    secret_reader = SecretReader(settings=settings)
+    accounts = queries.get_aws_accounts()
+    accounts = [a for a in accounts if a['name'] == account_name]
+    if len(accounts) == 0:
```
Piojo, line 489: `if not accounts:` is closer to English, and more flexible as a general pattern.

tools/qontract_cli.py
```diff
@@ -478,6 +478,25 @@ def bot_login(ctx, cluster_name):
...
 
 
+@get.command()
+@click.argument('account_name')
+@click.pass_context
+def aws_creds(ctx, account_name):
+    settings = queries.get_app_interface_settings()
+    secret_reader = SecretReader(settings=settings)
+    accounts = queries.get_aws_accounts()
+    accounts = [a for a in accounts if a['name'] == account_name]
```
Piojo, line 488: For a future iteration, I think we should do the filtering on the server side. It'll be faster and lead to less code around here.

tools/qontract_cli.py
```diff
@@ -478,6 +478,25 @@ def bot_login(ctx, cluster_name):
...
+    if len(accounts) == 0:
+        print(f"{account_name} not found.")
+        sys.exit(1)
+
+    account = accounts[0]
+    secret = secret_reader.read_all(account['automationToken'])
+    print(f"export AWS_REGION={account['resourcesDefaultRegion']}")
+    print(f"export AWS_ACCESS_KEY_ID={secret['aws_access_key_id']}")
+    print(f"export AWS_SECRET_ACCESS_KEY={secret['aws_secret_access_key']}")
```
Piojo, line 497: This is very neat, can we put in the help of this command that `$(qontract-cli aws_creds --account-name foo")` will set up the environment as expected?

Piojo (approved):

tools/qontract_cli.py
maorfr: i agree

geoberle (approved):
LGTM - cool feature!

tools/qontract_cli.py
maorfr: done

tools/qontract_cli.py
maorfr: done

tools/qontract_cli.py
maorfr: done

PR #2027, opened 2021-12-06
bump sretoolbox to 1.2.2

maorfr:
follows up on:
- https://github.com/app-sre/sretoolbox/pull/65
- https://github.com/app-sre/sretoolbox/pull/66

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ setup(
     package_data={'reconcile': ['templates/*.j2']},
 
     install_requires=[
-        "sretoolbox==1.1.0",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -19,7 +19,7 @@
     package_data={'reconcile': ['templates/*.j2']},
 
     install_requires=[
-        "sretoolbox==1.1.0",
+        "sretoolbox==1.2.2",
```
Piojo, line 22: Mmm... should we use `~=1.2` instead?

maorfr: image

PR #2028, opened 2021-12-07
[oc] get projects consistently

maorfr:
from time to time, we are observing the `openshift-namespaces` integration attempting to create a namespace that already exists:
```
[openshift-namespaces] ['create', 'REDACTED', 'openshift-config']
[openshift-namespaces] cluster: REDACTED, namespace: openshift-config, exception: [https://api.REDACTED.openshiftapps.com:6443]: Error from server (Forbidden): project.project.openshift.io "openshift-config" is forbidden: cannot request a project starting with "openshift-"
```
https://coreos.slack.com/archives/CS0E65QCV/p1638868906389900

we can hard code `openshift` to avoid attempting to create these namespaces, but that's not very generic.
since the integration attempts to create a namespace that exists and shows no previous errors, my assumption is that the call to get projects comes back with missing results but without an error.

this PR adds an option to get projects consistently (if 2 attempts do not return the same number of items, we retry).

if this idea seems right, i will add tests and docstrings.

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -101,6 +101,10 @@ class UnableToApplyError(Exception):
     pass
 
 
+class InconsistentProjectCountError(Exception):
+    pass
+
+
 class OCDecorators:
     @classmethod
     def process_reconcile_time(cls, function):
@@ -225,10 +229,7 @@ class OCDeprecated:
             self.get_version()
         self.init_projects = init_projects
         if self.init_projects:
-            self.projects = \
-                [p['metadata']['name']
-                 for p
-                 in self.get_all('Project.project.openshift.io')['items']]
+            self.projects = self.get_project_names(consistent=True)
         self.init_api_resources = init_api_resources
         if self.init_api_resources:
             self.api_resources = self.get_api_resources()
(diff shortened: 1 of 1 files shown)


geoberle (approved):

geoberle:
LGTM but getting different results back from the API server makes me nervous. especially since the `openshift-config` namespace should always exist

rporres:
Shouldn't we first confirm what the problem is before attempting a solution? If this is happening frequently enough that we think it deserves to deal with it, we could first start instrumenting more the part of the code where this happens so that we have a real insight on the error.

reconcile/utils/oc.py
```diff
@@ -472,6 +473,22 @@ def get_version(self):
         cmd = ['version', '--request-timeout=5']
         return self._run(cmd)
 
+    @retry(exceptions=InconsistentProjectCountError)
+    def get_project_names(self, consistent=False):
+        projects = self.get_projects()
+        if consistent:
+            projects_attempt_2 = self.get_projects()
```
Piojo, line 480: Don't we have a race condition here? A different piece of automation can add/remove projects that we're not allowed to modify (i.e, somebody else might be operating outside our limited `managedResourceNames`).

Piojo (commented):
How big of a deal is it if we try to create a project/namespace that already exists or to delete one that is already gone?

For the slack link you give, it seems like this is a recurring problem with just this one cluster, I'd escalate to the owners of said cluster, or discuss with them how we can diagnose it is safe for our integrations to run.

maorfr:
i'll just close this one. it doesn't happen often, and the impact is insignificant.

PR #2029, opened 2021-12-07
[terraform-users] improve new users detection

maorfr:
we currently rely on the terraform state to understand if a new user was created.
in #1897 we introduced a way to reset user passwords by deleting the login profile.

this PR detects when a new login profile is created and sends an email as a result.

### DIFF
--- a/reconcile/utils/terraform_client.py
+++ b/reconcile/utils/terraform_client.py
@@ -65,7 +65,6 @@ class TerraformClient:
         new_users = []
         self.init_outputs()  # get updated output
         for account, output in self.outputs.items():
-            existing_users = self.users[account]
             user_passwords = self.format_output(
                 output, self.OUTPUT_TYPE_PASSWORDS)
             console_urls = self.format_output(
--- a/reconcile/utils/terraform_client.py
+++ b/reconcile/utils/terraform_client.py
@@ -19,6 +19,12 @@ from reconcile.utils.openshift_resource import OpenshiftResource as OR
 ALLOWED_TF_SHOW_FORMAT_VERSION = "0.1"
 
 
+class AccountUser:
+    def __init__(self, account, user):
+        self.account = account
+        self.user = user
+
+
 class TerraformClient:
     def __init__(self, integration, integration_version,
                  integration_prefix, accounts, working_dirs, thread_pool_size,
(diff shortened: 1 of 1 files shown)


reconcile/utils/terraform_client.py
```diff
@@ -19,6 +19,12 @@
 ALLOWED_TF_SHOW_FORMAT_VERSION = "0.1"
 
 
+class AccountUser:
+    def __init__(self, account, user):
+        self.account = account
+        self.user = user
```
Piojo, line 25: We have dataclasses now! :smiley: 

```python

@dataclass
class AccountUser:
    account: str
    user: str
```

reconcile/utils/terraform_client.py
```diff
@@ -203,6 +213,9 @@ def log_plan_diff(self, name, tf, enable_deletion):
                 with self._log_lock:
                     logging.info([action, name, resource_type, resource_name])
                     self.should_apply = True
+                if action == 'create':
+                    if resource_type == 'aws_iam_user_login_profile':
+                        created_users.append(AccountUser(name, resource_name))
```
Piojo, line 218: This is an `and`, which is easier to read.
```suggestion
                if action == 'create' and resource_type == 'aws_iam_user_login_profile':
                    created_users.append(AccountUser(name, resource_name))
```

reconcile/utils/terraform_client.py
```diff
@@ -159,6 +168,7 @@ def log_plan_diff(self, name, tf, enable_deletion):
         # or if the integration's enable_deletion is true
         deletions_allowed = enable_deletion or account_enable_deletion
         deleted_users = []
+        created_users = []
```
Piojo, line 171: Are these users already created or to be created at a different point of the integration?

reconcile/utils/terraform_client.py
maorfr: perhaps, but more cases will probably be added over time (look at other sections above and below)

reconcile/utils/terraform_client.py
maorfr: we are in the context of `terraform plan`, so they are not yet created, but they are as good as created.

reconcile/utils/terraform_client.py
maorfr: :heart:

geoberle (approved):

PR #2033, opened 2021-12-08
enabling expiration to temp roles in app-interface

suzana-nesic:
Handling temp roles in app-interface by implementing an expirationDate field. Work for [APPSRE-2284](https://issues.redhat.com/browse/APPSRE-2284)

### DIFF
--- a/reconcile/openshift_groups.py
+++ b/reconcile/openshift_groups.py
@@ -120,7 +120,7 @@ def fetch_desired_state(oc_map):
                 f'expirationDate field is not formatted as YYYY-MM-DD, '
                 f'currently set as {r["expirationDate"]}'
             )
-        if not openshift_rolebindings.role_still_valid(r):
+        if not openshift_rolebindings.role_still_valid(r['expirationDate']):
             logging.warning(
                 f'The maximum expiration date of {r["name"]} '
                 f'has passed today '
--- a/reconcile/openshift_groups.py
+++ b/reconcile/openshift_groups.py
@@ -116,7 +116,7 @@ def fetch_desired_state(oc_map):
     for r in roles:
         if not openshift_rolebindings \
                 .has_valid_expiration_date(r['expirationDate']):
-            logging.warning(
+            raise ValueError(
                 f'expirationDate field is not formatted as YYYY-MM-DD, '
                 f'currently set as {r["expirationDate"]}'
             )
@@ -125,6 +125,7 @@ def fetch_desired_state(oc_map):
                 f'The maximum expiration date of {r["name"]} '
                 f'must not exceed '
                 f'{openshift_rolebindings.EXPIRATION_MAX} days from today')
+            continue
         for a in r['access'] or []:
             if None in [a['cluster'], a['group']]:
                 continue
--- a/reconcile/openshift_groups.py
+++ b/reconcile/openshift_groups.py
@@ -4,14 +4,13 @@ import itertools
 
 from sretoolbox.utils import threaded
 
-from reconcile.utils import openshift_resource
 from reconcile.utils import gql
 from reconcile import queries
+from reconcile import openshift_rolebindings
 from reconcile.utils.oc import OC_Map
 from reconcile.utils.defer import defer
 from reconcile.utils.sharding import is_in_shard
 
-EXPIRATION_MAX = 90
 
 ROLES_QUERY = """
 {
@@ -116,10 +115,15 @@ def fetch_desired_state(oc_map):
 
     for r in roles:
         gql.get_api
-        if not openshift_resource.role_still_valid(r):
+        if not openshift_rolebindings \
+                .has_valid_expiration_date(r['expirationDate']):
+            raise ValueError(
+                f'expirationDate field is not formatted as YYYY-MM-DD, '
+                f'currently set as {r["expirationDate"]}')
+        if not openshift_rolebindings.role_still_valid(r):
             raise ValueError(
                 f'The maximum expiration date of {r["name"]} '
-                f'shall not exceed {openshift_resource.EXPIRATION_MAX} \
+                f'shall not exceed {openshift_rolebindings.EXPIRATION_MAX} \
                     days from today')
         for a in r['access'] or []:
             if None in [a['cluster'], a['group']]:
--- a/reconcile/openshift_groups.py
+++ b/reconcile/openshift_groups.py
@@ -2,6 +2,7 @@ import sys
 import logging
 import itertools
 
+from datetime import datetime, date
 from sretoolbox.utils import threaded
 
 from reconcile.utils import gql
@@ -9,6 +10,9 @@ from reconcile import queries
 from reconcile.utils.oc import OC_Map
 from reconcile.utils.defer import defer
 from reconcile.utils.sharding import is_in_shard
+from reconcile.utils.aggregated_list import RunnerException
+
+EXPIRATION_MAX = 90
 
 ROLES_QUERY = """
 {
@@ -17,6 +21,7 @@ ROLES_QUERY = """
     users {
       github_username
     }
+    expirationDate
     access {
       cluster {
         name
@@ -111,6 +116,13 @@ def fetch_desired_state(oc_map):
     desired_state = []
 
     for r in roles:
+        exp_date = datetime.strptime(r['expirationDate'], '%Y-%m-%d').date()
+        if exp_date < date.today():
+            continue
+        elif (exp_date - date.today()).days > EXPIRATION_MAX:
+            raise RunnerException(
+                f'The maximum expiration date of {r["name"]} '
+                f'shall not exceed {EXPIRATION_MAX} days form today')
         for a in r['access'] or []:
             if None in [a['cluster'], a['group']]:
                 continue
--- a/reconcile/openshift_groups.py
+++ b/reconcile/openshift_groups.py
@@ -2,16 +2,15 @@ import sys
 import logging
 import itertools
 
-from datetime import datetime, date
 from sretoolbox.utils import threaded
 
+from reconcile.utils import openshift_resource
 from reconcile.utils import gql
 from reconcile import queries
 from reconcile import openshift_rolebindings
 from reconcile.utils.oc import OC_Map
 from reconcile.utils.defer import defer
 from reconcile.utils.sharding import is_in_shard
-from reconcile.utils.aggregated_list import RunnerException
 
 EXPIRATION_MAX = 90
 
--- a/reconcile/openshift_groups.py
+++ b/reconcile/openshift_groups.py
@@ -116,14 +116,15 @@ def fetch_desired_state(oc_map):
     for r in roles:
         if not openshift_rolebindings \
                 .has_valid_expiration_date(r['expirationDate']):
-            raise ValueError(
+            logging.warning(
                 f'expirationDate field is not formatted as YYYY-MM-DD, '
-                f'currently set as {r["expirationDate"]}')
+                f'currently set as {r["expirationDate"]}'
+            )
         if not openshift_rolebindings.role_still_valid(r):
-            raise ValueError(
+            logging.warning(
                 f'The maximum expiration date of {r["name"]} '
-                f'shall not exceed {openshift_rolebindings.EXPIRATION_MAX} \
-                    days from today')
+                f'must not exceed '
+                f'{openshift_rolebindings.EXPIRATION_MAX} days from today')
         for a in r['access'] or []:
             if None in [a['cluster'], a['group']]:
                 continue
--- a/reconcile/openshift_groups.py
+++ b/reconcile/openshift_groups.py
@@ -123,8 +123,7 @@ def fetch_desired_state(oc_map):
         if not openshift_rolebindings.role_still_valid(r):
             logging.warning(
                 f'The maximum expiration date of {r["name"]} '
-                f'was either over {openshift_rolebindings.EXPIRATION_MAX} days ago or over '
-                f'{openshift_rolebindings.EXPIRATION_MAX} days old from now'
+                f'has passed today '
             )
             continue
         for a in r['access'] or []:
--- a/reconcile/openshift_groups.py
+++ b/reconcile/openshift_groups.py
@@ -4,15 +4,12 @@ import itertools
 
 from sretoolbox.utils import threaded
 
-from reconcile.utils import openshift_resource
 from reconcile.utils import gql
 from reconcile import queries
 from reconcile import openshift_rolebindings
 from reconcile.utils.oc import OC_Map
 from reconcile.utils.defer import defer
 from reconcile.utils.sharding import is_in_shard
-
-EXPIRATION_MAX = 90
 
 
 ROLES_QUERY = """
--- a/reconcile/openshift_groups.py
+++ b/reconcile/openshift_groups.py
@@ -116,13 +116,18 @@ def fetch_desired_state(oc_map):
 
     for r in roles:
         gql.get_api
-        checkExpirationDate = openshift_resource.checkExpirationDate(r)
-        if checkExpirationDate is True:
-            continue
-        elif checkExpirationDate is False:
+        # checkExpirationDate = openshift_resource.roll_still_valid(r)
+        if not openshift_resource.role_still_valid(r):
             raise ValueError(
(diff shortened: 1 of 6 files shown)


reconcile/openshift_groups.py
```diff
@@ -111,6 +116,13 @@ def fetch_desired_state(oc_map):
     desired_state = []
 
     for r in roles:
+        exp_date = datetime.strptime(r['expirationDate'], '%Y-%m-%d').date()
+        if exp_date < date.today():
+            continue
+        elif (exp_date - date.today()).days > EXPIRATION_MAX:
+            raise RunnerException(
+                f'The maximum expiration date of {r["name"]} '
+                f'shall not exceed {EXPIRATION_MAX} days form today')
```
Piojo, line 125: Typo
```suggestion
                f'shall not exceed {EXPIRATION_MAX} days fromtoday')
```

reconcile/openshift_rolebindings.py
```diff
@@ -90,6 +93,13 @@ def fetch_desired_state(ri, oc_map):
     roles = gqlapi.query(ROLES_QUERY)['roles']
     users_desired_state = []
     for role in roles:
+        exp_date = datetime.strptime(role['expirationDate'], '%Y-%m-%d').date()
```
Piojo, line 96: This code is duplicated from above, can you centralize it in some method? Maybe somewhere in `reconcile.utils`?

reconcile/openshift_groups.py
```diff
@@ -111,6 +116,13 @@ def fetch_desired_state(oc_map):
     desired_state = []
 
     for r in roles:
+        exp_date = datetime.strptime(r['expirationDate'], '%Y-%m-%d').date()
+        if exp_date < date.today():
+            continue
+        elif (exp_date - date.today()).days > EXPIRATION_MAX:
+            raise RunnerException(
```
Piojo, line 123: Is this a `RunnerException` or just a good old `ValueError`?

Piojo (changes requested):
I can haz unit tests? :)

reconcile/openshift_groups.py
suzana-nesic: Updated, check out latest commit.

reconcile/openshift_rolebindings.py
suzana-nesic: Updated, went ahead and moved the method into specified path.

reconcile/openshift_groups.py
suzana-nesic: Yup, went ahead and updated that for you.

reconcile/utils/openshift_resource.py
```diff
@@ -431,6 +433,15 @@ def calculate_sha256sum(body):
...
 
 
+def checkExpirationDate(role):
+    exp_date = datetime.datetime \
+        .strptime(role['expirationDate'], '%Y-%m-%d').date()
+    if exp_date < datetime.date.today():
+        return True
+    elif (exp_date - datetime.date.today()).days > EXPIRATION_MAX:
+        return False
+
```
geoberle, line 443: you don't cover the the situation where the expiration date is in less than 90d from now, which results in a return value of None.

if the intent of this function is to detect expiration dates, that are more than 90d in the future, you could simply like this

return (exp_date - datetime.date.today()).days <= EXPIRATION_MAX

but please verify!

reconcile/utils/openshift_resource.py
```diff
@@ -431,6 +433,15 @@ def calculate_sha256sum(body):
         return m.hexdigest()
 
 
+def checkExpirationDate(role):
```
geoberle, line 436: i'm not sure this file is the right place for this function because it generically handles openshift/kubernetes resouces stuff, while this function is very specific for roles (at least the current usage and the name of the variable indicate that). 

please don't consider this a blocker!

reconcile/openshift_rolebindings.py
```diff
@@ -90,6 +92,14 @@ def fetch_desired_state(ri, oc_map):
...
     users_desired_state = []
     for role in roles:
+        expDate = openshift_resource.checkExpirationDate(role)
+        if expDate is True:
+            continue
+        elif expDate is False:
+            raise ValueError(
+                f'The maximum expiration date of {role["name"]} '
+                f'shall not exceed {openshift_resource.EXPIRATION_MAX} \
+                    days from today')
```
geoberle, line 102: by failing here for a wrongly configured expiration date, other good rolebindings are not reconciled for the time being. i know we do this in other places as well but if you could find a way to not generally fail the reconcile run but do whatever work can be done that would be great. 

BUT i haven't looked at how easy that can be done and i don't want to block a great enhancement to qontract-reconcile. so if you decide not to implement this right now, it is fine with me but leave a comment in the code and create a follow up jira ticket.

reconcile/utils/openshift_resource.py
```diff
@@ -431,6 +433,15 @@ def calculate_sha256sum(body):
...
+    exp_date = datetime.datetime \
+        .strptime(role['expirationDate'], '%Y-%m-%d').date()
```
geoberle, line 438: as far as i have seen the schema change, expirationDate is an optional field so you need to check if it is really in the role dict. otherwise you get a KeyError.

you should also make sure to handle the exception from an expirationDate that is not following the defined format.

can you also write tests for that?

geoberle (changes requested):

reconcile/openshift_groups.py
```diff
@@ -111,6 +115,14 @@ def fetch_desired_state(oc_map):
     desired_state = []
 
     for r in roles:
+        gql.get_api
+        checkExpirationDate = openshift_resource.checkExpirationDate(r)
+        if checkExpirationDate is True:
+            continue
+        elif checkExpirationDate is False:
```
Piojo, line 122: PEP8 requires snake_case instead of camelCase, so please `check_expiration_date()` and `check_expiration_date`.

reconcile/openshift_groups.py
```diff
@@ -111,6 +115,14 @@ def fetch_desired_state(oc_map):
...
+            raise ValueError(
+                f'The maximum expiration date of {r["name"]} '
+                f'shall not exceed {EXPIRATION_MAX} days from today')
```
Piojo, line 125: Also, this is shorter and easier:
```suggestion
        if  not openshift_resource.check_expiration_date(r):
            raise ValueError(
                f'The maximum expiration date of {r["name"]} '
                f'shall not exceed {EXPIRATION_MAX} days from today')
```

reconcile/openshift_rolebindings.py
Piojo: Same comment as above. Plus, what @geoberle mentioned.

reconcile/utils/openshift_resource.py
Piojo: If this comes from the graphql query, the field will always exist but may be `None`. You are more likely to get a `TypeError`. :(
As @geoberle said, let's have a test for this.

reconcile/utils/openshift_resource.py
```diff
@@ -431,6 +433,15 @@ def calculate_sha256sum(body):
...
+    if exp_date < datetime.date.today():
```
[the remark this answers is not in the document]
Piojo: We have had problems with code that didn't use `utcnow`, so please change this.

reconcile/utils/openshift_resource.py
Piojo: Please add a type hint here. Also, as I mentioned above, prefer `snake_case` names instead of `camelCase` for methods, in this case `def check_expiration_date`. Finally, this name doesn't really tell much about the return, why not call it `role_still_valid`? It is totally fine for predicates to be adjectives, and this one tells exactly what to expect as a return value, and calls like above become simpler, like
```python
if not role_still_valid(r):
    raise ValueError(...)
```

reconcile/utils/openshift_resource.py
Piojo: As an alternative (pick the one you prefer), all the callers in this PR are raising an exception in case the role has expired which means you can make this method act like an assertion. That is:

```python
if (exp_date - datetime.date.today()).days > EXPIRATION_MAX:
    raise ...
```
and you save yourself a few ifs in the callers.

reconcile/utils/openshift_resource.py
geoberle: i always forgett about graphql returning None for those kind of things :(
thanks for the reminder!

reconcile/openshift_groups.py
suzana-nesic: Thank you! Went ahead and fixed that.

reconcile/openshift_rolebindings.py
suzana-nesic: I will go ahead and follow that up with a jira ticket.

reconcile/utils/openshift_resource.py
suzana-nesic: I wrote some tests for this but I don't believe I wrote is really checking what we want? Can anyone confirm if my latest commits are showing this?

reconcile/utils/openshift_resource.py
```diff
@@ -280,6 +293,13 @@ def sha256sum(self):
         annotations = body['metadata']['annotations']
         return annotations['qontract.sha256sum']
 
+    def expiration_date(self):
+        body = self.annotate().body
+
+        # annotations = body['expirationDate']
+        # return annotations['expirationDate']
+        return body['expirationDate']
+
```
geoberle, line 302: the class `OpenshiftResource` really represents Kubernetes objects and not the schema object from qontract-schema. therefore `expirationDate` does not exist within this context.

please have a closer look at this class. it represents Kubernetes objects in a very generic way, so not specific to any concrete K8S type. having handling for a very specific type (e.g. Role) in here does not fit well the purpose of this class.

the `expirationDate` only exists within qontract-schema, so please move the methods/functions `expiration_date`,  `has_valid_expiration_date`, `has_expiration_date`, `role_still_valid` and the `EXPIRATION_DATE` constant to a more appropriate context. how about `reconcile/openshift_rolebindings.py` or `reconcile/openshift_groups.py`?

reconcile/utils/openshift_resource.py
geoberle: i saw that you introduced some methods to check the validity of the expiration date but i can only see them used from within the tests

reconcile/utils/openshift_resource.py
```diff
@@ -237,6 +242,14 @@ def has_valid_sha256sum(self):
...
             return False
 
+    def has_valid_expiration_date(self):
+        try:
+            current_expiration_date = \
+                self.body['expirationDate']
+            return current_expiration_date == self.expiration_date()
+        except KeyError:
+            return False
+
```
geoberle, line 252: i can't find usage of this method anywhere besides the test you wrote for it itself. from the name of the method, i expect also something like checking if the date adheres to the format

reconcile/test/test_openshift_resource.py
```diff
@@ -132,3 +139,67 @@ def test_has_owner_reference_false():
...
+
+        with pytest.raises(ConstructResourceError):
+            openshift_resource = OR(resource, TEST_INT, TEST_INT_VER)
+            assert openshift_resource.has_valid_expiration_date() is None
+
+    @staticmethod
+    def test_has_expiration_date():
+        resource = {
+            'kind': 'kind',
+            'expirationDate': '01-01-2022',
```
geoberle, line 172: as already mentioned on the change in `reconcile/utils/openshift_resource.py`, `expirationDate` is not a valid field on the K8s resources representing roles and rolebindings

geoberle (changes requested):

reconcile/openshift_groups.py
```diff
@@ -111,6 +114,17 @@ def fetch_desired_state(oc_map):
     desired_state = []
 
     for r in roles:
+        gql.get_api
```
geoberle, line 117:
```suggestion
```

reconcile/openshift_rolebindings.py
```diff
@@ -90,6 +92,15 @@ def fetch_desired_state(ri, oc_map):
...
     users_desired_state = []
     for role in roles:
+        if not has_valid_expiration_date(role['expirationDate']):
+            raise ValueError(
+                f'expirationDate field is not formatted as YYYY-MM-DD, '
+                f'currently set as {role["expirationDate"]}')
+        if not role_still_valid(role['expirationDate']):
+            raise ValueError(
+                f'The maximum expiration date of {role["name"]} '
+                f'shall not exceed {EXPIRATION_MAX} \
```
Piojo, line 102: Better to use "must" instead of "shall"?

reconcile/openshift_rolebindings.py
```diff
@@ -185,3 +196,31 @@ def run(dry_run, thread_pool_size=10, internal=None,
...
     if ri.has_error_registered():
         sys.exit(1)
+
+
+def has_valid_expiration_date(role):
+    date_bool = True
+    try:
+        role.get('expirationDate')
+    except AttributeError:
+        return date_bool
```
Piojo, line 206: The point of `.get` is not to have an exception. Either:

```python
try:
    role['expirationDate']
except KeyError:
    ...
```

or 
```python
if not role.get('expirationDate'):
    return date_bool
```

Given how graphql introduces the key even if it doesn't exist in app-interface, I'd recommend the latter.

reconcile/openshift_rolebindings.py
```diff
@@ -185,3 +196,31 @@ def run(dry_run, thread_pool_size=10, internal=None,
...
+    try:
+        date_bool = \
+            bool(datetime.datetime.strptime(exp_date_role, date_format))
+    except ValueError:
+        date_bool = False
+    return date_bool
+
+
+def role_still_valid(role):
+    try:
```
Piojo, line 218: Same as above.

reconcile/openshift_rolebindings.py
```diff
@@ -185,3 +196,31 @@ def run(dry_run, thread_pool_size=10, internal=None,
 
     if ri.has_error_registered():
         sys.exit(1)
+
+
+def has_valid_expiration_date(role):
```
Piojo, line 201: Type hints, please. In this case, I believe it's 
```suggestion
def has_valid_expiration_date(role: dict[str, Any]) -> bool:
```

reconcile/test/test_openshift_resource.py
```diff
@@ -132,3 +139,61 @@ def test_has_owner_reference_false():
...
         openshift_resource = OR(resource, TEST_INT, TEST_INT_VER)
         assert not openshift_resource.has_owner_reference()
+
+    @staticmethod
+    def test_check_temp_role_after_expiration_date():
+        expiration_date = date.today() + \
+                timedelta(days=(openshift_rolebindings.EXPIRATION_MAX+1))
+        resource = mock_openshift_role_bindings(expiration_date)
+        for r in resource:
+            assert openshift_rolebindings.role_still_valid(r) is False
```
Piojo, line 149: Easier to read:

```suggestion
            assert not openshift_rolebindings.role_still_valid(r)
```

reconcile/test/test_openshift_resource.py
```diff
@@ -132,3 +139,61 @@ def test_has_owner_reference_false():
...
+        resource = mock_openshift_role_bindings(expiration_date)
+        for r in resource:
+            assert openshift_rolebindings.role_still_valid(r) is True
+
+    @staticmethod
+    def test_check_temp_role_no_expiration_date():
+        resource = mock_openshift_role_bindings_no_expiration_date()
+        for r in resource:
+            assert openshift_rolebindings \
+                .has_valid_expiration_date(r) is True
```
Piojo, line 164: Easier:
```suggestion
            assert openshift_rolebindings \
                .has_valid_expiration_date(r)
```
(maybe this fits in one line now)

reconcile/openshift_rolebindings.py
suzana-nesic: I was having a lot of trouble getting either of those two options to work, I have something similar to the first option that does work with my tests and with local q-r but let me know if it can be condensed down.

geoberle (approved):
LGTM - congratulations!! 🥳 🎉

PR #2034, opened 2021-12-08
Create dependabot.yml

Piojo:
Try to automate some dependency maintenance. It should result in pull requests for outdated packages with known vulnerabilities.

This is an experiment for now, it needs to be integrated in the security efforts for the team.


Piojo:
Refs https://issues.redhat.com/browse/APPSRE-4170

Piojo:
As discussed in slac, we are going to give this a try and see how noisy its PRs are.

PR #2035, opened 2021-12-09
[saasherder] add support to collect images from ClowdApp

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4171

we want to validate images being used before deploying them for ClowdApp resources.

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -591,9 +591,16 @@ class SaasHerder():
             images.add(resource["spec"]["image"])
         except KeyError:
             pass
-        # ClowdApp
+        # ClowdApp deployments
         try:
             deployments = resource["spec"]["deployments"]
+            for d in deployments:
+                images.add(d["podSpec"]["image"])
+        except KeyError:
+            pass
+        # ClowdApp jobs
+        try:
+            deployments = resource["spec"]["jobs"]
             for d in deployments:
                 images.add(d["podSpec"]["image"])
         except KeyError:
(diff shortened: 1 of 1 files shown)


reconcile/utils/saasherder.py
```diff
@@ -591,6 +591,20 @@ def _collect_images(resource):
             images.add(resource["spec"]["image"])
         except KeyError:
             pass
+        # ClowdApp deployments
+        try:
+            deployments = resource["spec"]["deployments"]
+            for d in deployments:
+                images.add(d["podSpec"]["image"])
+        except KeyError:
+            pass
```
Piojo, line 600:
```suggestion
        with suppress(KeyError):
            deployments = resource["spec"]["deployments"]
            for d in deployments:
                images.add(d["podSpec"]["image"])
```

reconcile/utils/saasherder.py
```diff
@@ -591,6 +591,20 @@ def _collect_images(resource):
...
+        # ClowdApp jobs
+        try:
+            deployments = resource["spec"]["jobs"]
+            for d in deployments:
+                images.add(d["podSpec"]["image"])
+        except KeyError:
+            pass
```
Piojo, line 607:
```suggestion
        with suppress(KeyError):
            deployments = resource["spec"]["jobs"]
            for d in deployments:
                images.add(d["podSpec"]["image"])
```

Piojo (changes requested):
Please, use `contextlib.suppress` instead.

Piojo (approved):

PR #2043, opened 2021-12-13
add idle_timeout support for AWS ALB

jfchevrette:
Related to APPSRE-3695

Requires https://github.com/app-sre/qontract-schemas/pull/31

### DIFF
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -3493,6 +3493,10 @@ class TerrascriptClient:
             'tags': common_values['tags'],
             'depends_on': self.get_dependencies([sg_tf_resource]),
         }
+
+        if 'idle_timeout' in resource:
+            values['idle_timeout'] = resource['idle_timeout']
+
         lb_tf_resource = aws_lb(identifier, **values)
         tf_resources.append(lb_tf_resource)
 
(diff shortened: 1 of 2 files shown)


maorfr (approved):

reconcile/utils/terrascript_client.py
```diff
@@ -3493,6 +3493,10 @@ def populate_tf_resource_alb(self, resource, namespace_info,
             'tags': common_values['tags'],
             'depends_on': self.get_dependencies([sg_tf_resource]),
         }
+
+        if 'idle_timeout' in resource:
+            values['idle_timeout'] = resource['idle_timeout']
```
Piojo, line 3498: If this comes from a graphql query this path will always exist, see https://issues.redhat.com/browse/APPSRE-3868 . In that case, this is equivalent:
```python
        values['idle_timeout'] = resource['idle_timeout']
```

and, if we fix the task above, then 
```suggestion
        values['idle_timeout'] = resource.get('idle_timeout')
```

Piojo (approved):

reconcile/utils/terrascript_client.py
jfchevrette: Ah yes I forgot about that graphql behavior 

I adjusted the condition in 0d91787

rporres:
late to the party but LGTM :)

PR #2046, opened 2021-12-14
[sqs-gateway] do not init aws users

maorfr:
related to https://issues.redhat.com/browse/APPSRE-4141

sqs-gateway is used by openshift-saas-deploy to publish promotions. without init_users, we are actually initiating all users, which takes a lot of memory.

example: saas-app-sre-observability-per-cluster -
with `init_users=True`:
image

with `init_users=False`:
image


maorfr:
this is likely not really the cause for the spike, but still worth adding.

PR #2047, opened 2021-12-15
add terrafrom asg

BumbleFeng:

depens on https://github.com/app-sre/qontract-schemas/pull/34


maorfr (approved):
LGTM

since this is a new provider, we can be less strict in the review and allow to iterate while this is already promoted.

i'd suggest to start splitting your work to separate commits, but i won't force it here as this is hard work as it is. something to consider in your future PRs to make them approachable for others.

PR #2049, opened 2021-12-16
query for aws accounts by name or uid

maorfr:
related to https://issues.redhat.com/browse/APPSRE-2694

depends on https://github.com/app-sre/qontract-schemas/pull/35

this PR adds the ability to query for aws accounts by name of uid. this means we will only get the response of what we queried for (less transmitted data).
it also adds usage of those functions in many integrations.

better to review commit by commit, even though at some point i got sloppy.


Piojo (approved):
LGTM

PR #2050, opened 2021-12-16
[openshift-base] statefulset pvc resize

maorfr:
related to https://issues.redhat.com/browse/ASIC-145

during the incident, we noticed that pods were being picked up by the newly applied statefulset without being restarted, including picking up an updated storage size.

this needs to be validated through docs or an experiment.

### DIFF
--- a/reconcile/test/test_utils_oc.py
+++ b/reconcile/test/test_utils_oc.py
@@ -267,6 +267,57 @@ class TestGetObjRootOwner(TestCase):
             oc.get_obj_root_owner('namespace', obj, allow_not_found=False)
 
 
+class TestPodOwnedPVCNames(TestCase):
+    def setUp(self):
+        self.oc = OC('cluster', 'server', 'token', local=True)
+
+    def test_no_volumes(self):
+        pods = [{'volumes': []}]
+        owned_pvc_names = self.oc.get_pod_owned_pvc_names(pods)
+        self.assertEqual(len(owned_pvc_names), 0)
+
+    def test_other_volumes(self):
+        pods = [{'volumes': [{'configMap': {'name': 'cm'}}]}]
+        owned_pvc_names = self.oc.get_pod_owned_pvc_names(pods)
+        self.assertEqual(len(owned_pvc_names), 0)
+
+    def test_ok(self):
+        pods = [{'volumes': [{'persistentVolumeClaim': {'claimName': 'cm'}}]}]
+        owned_pvc_names = self.oc.get_pod_owned_pvc_names(pods)
+        self.assertEqual(len(owned_pvc_names), 1)
+        self.assertEqual(owned_pvc_names[0], 'cm')
+
+
+class TestGetStorage(TestCase):
+    def setUp(self):
+        self.oc = OC('cluster', 'server', 'token', local=True)
+
+    def test_none(self):
+        resource = {'spec': 'whatever'}
+        storage = self.oc.get_storage(resource)
+        self.assertIsNone(storage)
+
+    def test_ok(self):
+        size = "100Gi"
+        resource = {
+            'spec': {
+                'volumeClaimTemplates': [
+                    {
+                        'spec': {
+                            'resources': {
+                                'requests': {
+                                    'storage': size
+                                }
+                            }
+                        }
+                    }
+                ]
+            }
+        }
+        result = self.oc.get_storage(resource)
+        self.assertEqual(result, size)
+
+
 class TestValidateLabels(TestCase):
     def test_ok(self):
         self.assertFalse(validate_labels({'my.company.com/key-name': 'value'}))
--- a/reconcile/test/test_utils_oc.py
+++ b/reconcile/test/test_utils_oc.py
@@ -267,6 +267,7 @@ class TestGetObjRootOwner(TestCase):
             oc.get_obj_root_owner('namespace', obj, allow_not_found=False)
 
 
+@patch.dict(os.environ, {"USE_NATIVE_CLIENT": "False"}, clear=True)
 class TestPodOwnedPVCNames(TestCase):
     def setUp(self):
         self.oc = OC('cluster', 'server', 'token', local=True)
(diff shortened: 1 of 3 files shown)


geoberle:
# StatefulSet PVC resize

The “delete statefulset —cascade=orphaned and reapply statefulset with increased storage request”-strategy does not have any effect on the PVCs. The recreated StatefulSet just adopts existing Pods and existing PVCs and [does not change them](https://github.com/kubernetes/kubernetes/blob/1df88a8a42952d4b0f393c1d25b1d1c992bf72c3/pkg/controller/statefulset/stateful_pod_control.go#L342). The statefulset controller has some sort of [volumeClaimTemplate conformance check](https://github.com/kubernetes/kubernetes/blob/1df88a8a42952d4b0f393c1d25b1d1c992bf72c3/pkg/controller/statefulset/stateful_pod_control.go#L150-L157) for the pods owned by a statefulset but it does not incorporate sizing and access modes and instead checks only [basic things like existence](https://github.com/kubernetes/kubernetes/blob/1df88a8a42952d4b0f393c1d25b1d1c992bf72c3/pkg/controller/statefulset/stateful_set_utils.go#L115).

It is strange, that the deletion of the `owned_pods` did not happen (are we 100% sure about this?). As far as i remember, we resized the PVCs manually and Pods gain immediate access to the increased storage once the PVC has been resized (see next section)


# PVC resize without Pod restart

TL;DR; expansion of gp2 backed in-use PVCs works on Openshift 4.8 (I did not test on older versions) without Pod or Container restarts. Resizing of EBS volumes is only supported once every 6 hours though.

The `ExpandInUsePersistentVolumes` feature gate is enabled since Kubernetes 1.15 by default and enables us to resize PVCs without Pod restarts if supported by the storage class. Luckily `gp2` supports this. 

## Experiment: Resize a PVC used by a Pod owned by a StatefulSet

```
oc describe pvc (right after resize request)
FileSystemResizePending   True    Mon, 01 Jan 0001 00:00:00 +0000   Tue, 28 Dec 2021 07:34:41 +0100           Waiting for user to (re-)start a pod to finish file system resize of volume on node.
```

And after 1 minute

```
Normal  FileSystemResizeSuccessful  7s (x2 over 7h52m)  kubelet  MountVolume.NodeExpandVolume succeeded for volume "pvc-d75ffd88-e145-44aa-bc75-76eb850921ae"
```

No Pod restart was required even though the initial event indicated that.

```
oc describe pod
…
Normal  FileSystemResizeSuccessful  2m40s (x2 over 7h55m)  kubelet  MountVolume.NodeExpandVolume succeeded for volume "pvc-d75ffd88-e145-44aa-bc75-76eb850921ae"
```

Pod was not recreated and no containers within the pod have been restarted

Not directly related to this but good to know: an EBS volume resize can only be done once every 6 hours (see https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifyVolume.html). If a gp2 PVC resize is started before that, `oc describe pvc` shows something like this:

```
Warning  VolumeResizeFailed  0s  volume_expand  (combined from similar events): error expanding volume "gerd/www-web-0" of plugin "kubernetes.io/aws-ebs": AWS modifyVolume failed for vol-06164ef6eae80f9ec with VolumeModificationRateExceeded: You've reached the maximum modification rate per volume limit. Wait at least 6 hours between modifications per EBS volume.
```

But good news: after the required time has passed, the PVC resize operation reconciles automatically.

# Summary
If we want to enable qontract-reconcile to take care of PVC resizes in StatefulSets, we need to take care of resizing the actual PVCs in our code - the “statefulset-dance” is not going to help. BUT: there is an ongoing [KEP-0661](https://github.com/kubernetes/enhancements/pull/2842) to enable StatefulSets to directly take care of PVC resizing.

maorfr:
great investigation! implemented exactly what you have proposed.

reconcile/test/test_utils_oc.py
```diff
@@ -267,6 +267,58 @@ def test_controller_true_allow_false_ref_not_found_raise(self, oc_get):
             oc.get_obj_root_owner('namespace', obj, allow_not_found=False)
 
 
+@patch.dict(os.environ, {"USE_NATIVE_CLIENT": "False"}, clear=True)
+class TestPodOwnedPVCNames(TestCase):
+    def test_no_volumes(self):
+        pods = [{'volumes': []}]
+        oc = OC('cluster', 'server', 'token', local=True)
+        owned_pvc_names = oc.get_pod_owned_pvc_names(pods)
+        self.assertEqual(len(owned_pvc_names), 0)
```
Piojo, line 276: Better:
```python
self.assertFalse(owned_pvc_names)
```
it'll show how `owned_pvc_names` is true-y, making debugging easier.

reconcile/utils/oc.py
```diff
@@ -514,6 +522,32 @@ def get_owned_pods(self, namespace, resource):
 
         return owned_pods
 
+    @staticmethod
+    def get_pod_owned_pvc_names(pods):
```
Piojo, line 526: Type hints please? It'll make things easier for me to understand. In this case I believe it's:
```suggestion
    def get_pod_owned_pvc_names(pods: Iterable[dict[str, dict]]) -> set[str]:
```
(but double-check with `tox -e type`)

reconcile/utils/oc.py
```diff
@@ -514,6 +522,32 @@ def get_owned_pods(self, namespace, resource):
...
+                continue
+            for v in vols:
+                with suppress(KeyError):
+                    cn = v['persistentVolumeClaim']['claimName']
+                    owned_pvc_names.add(cn)
+
+        return owned_pvc_names
+
+    @staticmethod
+    def get_storage(resource):
```
Piojo, line 540: In this case, it looks like the type hint is
```suggestion
    def get_storage(resource: dict) -> Optional[str]:
```

reconcile/test/test_utils_oc.py
maorfr: not sure i agree with that (or maybe i'm not understanding something).
the test is called `test_no_volumes`, which to me makes a lot of sense that it means that the length of the returned list will be 0.

reconcile/utils/oc.py
maorfr: thanks!

reconcile/utils/oc.py
maorfr: thanks!

geoberle (approved):
lgtm

reconcile/utils/oc.py
maorfr: this caused a failure due to a missing return statement. reverted it.

maorfr:
as type hints are optional (and i am fulfilling my right to not use them), and the open comment is a nit, i'm going to merge this to get the functionality promoted.

PR #2051, opened 2021-12-16
add support to delete specific aws resources

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4186

depends on https://github.com/app-sre/qontract-schemas/pull/38

this PR adds an ability to define resources that are approved for deletion, even for an account that does not have deletions enabled.

### DIFF
--- a/reconcile/test/test_utils_terraform_client.py
+++ b/reconcile/test/test_utils_terraform_client.py
@@ -1,0 +1,87 @@
+from unittest import TestCase
+import reconcile.utils.terraform_client as tfclient
+
+
+class TestDeletionApproved(TestCase):
+
+    def test_no_deletion_approvals(self):
+        account = {
+            'name': 'a1',
+            'deletionApprovals': []
+        }
+        tf = tfclient.TerraformClient(
+            'integ',
+            'v1',
+            'integ_pfx',
+            [account],
+            {},
+            1
+        )
+        result = tf.deletion_approved('a1', 't1', 'n1')
+        self.assertFalse(result)
+
+    def test_deletion_not_approved(self):
+        account = {
+            'name': 'a1',
+            'deletionApprovals': [
+                {
+                    'type': 't1',
+                    'name': 'n1',
+                    'expiration': '2000-01-01'
+                }
+            ]
+        }
+        tf = tfclient.TerraformClient(
+            'integ',
+            'v1',
+            'integ_pfx',
+            [account],
+            {},
+            1
+        )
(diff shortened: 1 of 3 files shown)


reconcile/utils/terraform_client.py
```diff
@@ -246,6 +251,21 @@ def log_plan_diff(self, name, tf, enable_deletion):
...
         return disabled_deletion_detected, deleted_users, created_users
 
+    def deletion_approved(self, account_name, resource_type, resource_name):
+        account = self.accounts[account_name]
+        deletion_approvals = account.get('deletionApprovals')
+        if not deletion_approvals:
+            return False
+        now = datetime.utcnow()
+        for da in deletion_approvals:
+            expiration = datetime.strptime(da['expiration'], DATE_FORMAT)
```
geoberle, line 261: please handle the potential ValueError for `da['expiration']` values that don't match `DATE_FORMAT`

reconcile/utils/terraform_client.py
```diff
@@ -246,6 +251,21 @@ def log_plan_diff(self, name, tf, enable_deletion):
...
+            if resource_type == da['type'] \
+                    and resource_name == da['name'] \
+                    and expiration <= now:
```
geoberle: the other way around now <= expiration

maorfr: yup, already caught that with testing. i started with a `if not confition: continue` structure, so that's where this bug came from.

reconcile/test/test_utils_terraform_client.py
```diff
@@ -0,0 +1,87 @@
...
+        self.assertFalse(result)
+
+    def test_deletion_approved(self):
+        account = {
+            'name': 'a1',
+            'deletionApprovals': [
+                {
+                    'type': 't1',
+                    'name': 'n1',
+                    'expiration': '2500-01-01'
```
geoberle, line 74: in 2500 people are going to freak out because of the economy crashing effect of "The Maor Bug"

reconcile/utils/terraform_client.py
maorfr: done, and added a test as well

reconcile/test/test_utils_terraform_client.py
maorfr: i figured its easier to use the year 2500 and not patch datetime :)

geoberle (approved):
LGTM

PR #2053, opened 2021-12-17
Add PATCH as a write method to ALB config

syed:


syed:
closing in lieu of https://github.com/app-sre/qontract-reconcile/pull/2057

PR #2056, opened 2021-12-19
[openshift-serviceaccount-tokens] add support to define resource output name

maorfr:
depends on https://github.com/app-sre/qontract-schemas/pull/39

this will allow an ability to define the name of the output Secret (both in OpenShift and in Vault).


geoberle (approved):

patjlm:
LGTM

PR #2057, opened 2021-12-20
Add per-path targets for ALB

syed:
schema changes: https://github.com/app-sre/qontract-schemas/pull/40
JIRA: APPSRE-3695

This change supports per-path ALB targets. Sample config:
```
provider: alb
...
targets:
- name: py2
  openshift_service: py2-svc
  default: true
- name: py3
  openshift_service: py3-svc
rules:
- condition:
    path: /v2/auth*
    methods: read
  action:
  - target: py2
    weight: 100
- condition:
    path: /v2*
    methods: read
  action:
  - target: py3
    weight: 100
  - target: py2
    weight: 0
```

### DIFF
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -3714,6 +3714,8 @@ class TerrascriptClient:
             'read': ['GET', 'HEAD', 'OPTIONS'],
             'write': ['POST', 'PUT', 'PATCH', 'DELETE']
         }
+        http_methods_lookup['all'] = http_methods_lookup['read'] \
+            + http_methods_lookup['write']
 
         for rule_num, rule in enumerate(resource['rules']):
             condition = rule['condition']
@@ -3732,7 +3734,7 @@ class TerrascriptClient:
                 'action': {
                     'type': 'forward',
                     'forward': {
-                        'target_group':[]
+                        'target_group': []
                     },
                 },
                 'condition': [
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -3722,6 +3722,7 @@ class TerrascriptClient:
 
             values = {
                 'listener_arn': f'${{{forward_lbl_tf_resource.arn}}}',
+                'priority': rule_num + 1,
                 'action': {
                     'type': 'forward',
                     'forward': {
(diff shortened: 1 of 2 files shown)


reconcile/utils/terrascript_client.py
```diff
@@ -3636,64 +3622,64 @@ def populate_tf_resource_alb(self, resource, namespace_info,
...
+                    'target_group': {
+                        'arn': f'${{{target_resource.arn}}}',
+                        'weight': a['weight']
+                    }
+                })
+                weight_sum += a['weight']
+
+            if weight_sum != 100:
+                raise ValueError(
+                    'sum of weights for a rule should be 100'
```
Piojo, line 3675: Log `weight_sum` here too, please.

Piojo (commented):
Looks solid to me. I won't force you to start writing unit tests for `terrascript_client` - but if you feel inclined, you're welcome. @jfchevrette can you have a look here since you know ALB way better than I do?

reconcile/utils/terrascript_client.py
```diff
@@ -3651,64 +3637,65 @@ def populate_tf_resource_alb(self, resource, namespace_info,
...
+                    )
+
+                target_resource = valid_targets[target_name]
+
+                values['action']['forward'].append({
+                    'target_group': {
+                        'arn': f'${{{target_resource.arn}}}',
+                        'weight': a['weight']
+                    }
+                })
```
jfchevrette, line 3685: according to the terraform docs and how previous code worked, `forward` here is a dict and the list of target groups arn/weight is actually in `target_group`

syed: @jfchevrette  this is what I have from the docs

```
# Weighted Forward action

resource "aws_lb_listener_rule" "host_based_routing" {
  listener_arn = aws_lb_listener.front_end.arn
  priority     = 99

  action {
    type = "forward"
    forward {
      target_group {
        arn    = aws_lb_target_group.main.arn
        weight = 80
      }

      target_group {
        arn    = aws_lb_target_group.canary.arn
        weight = 20
      }

      stickiness {
        enabled  = true
        duration = 600
      }
    }
  }

  condition {
    host_header {
      values = ["my-service.*.terraform.io"]
    }
  }
}
```

I had assumed that the list has each entry as `target_group` as the dict key. I have fixed it so now the list is inside `target_group`

PR #2060, opened 2021-12-21
Fix update_state methodç

lechuk47:
Update method call to work with a Chainmap


Piojo:
What is the issue with the current code?

PR #2063, opened 2021-12-22
APPSRE-3446 Improve Saas configs trigger 2

lechuk47:
## Overview

This PR re-introduces the changes we made here: #2005. 
There were 2 problems with the 1st PR. 

### Problem 1: Introduction of ChainMap to manage target configs. 
ChainMap is a useful data structure to manage changes on dictionaries without changing them. It basically works having layers of dictionaries and setting the changes in a new dictionary instead of changing the original.  When a key is get  from the ChainMap, it returns the modified data if exists. This is useful to manage the target config dictionaries in SaasHerder without modifying the original target config dictionary in the saas_file structures. 

The problem is that ChainMaps are not JSON serializable and SaasHerder needs to serialize them in 2 places: 
 * To generate the target_config_hash value (sha hash): This one was controlled converting the CM to a dict 
 * To store the config definition on S3 (config state). This was not controlled and the mock in the tests can not detect this behavior easily. 

This caused a loop of PipelineRuns. The new configuration could not be stored, so every reconcile loop was detecting configuration changes on the Saas files.  

To solve this problem, the ChainMap is just used to generate the changes without changing the data in the saas_file. Once the the CM is filled, is stored as a plain dict. Check the comment below to show the code change. 

### Problem 2: GrapQL schema changes caused differences in all target configs. 
Current target config (the one existent in the state) and the new Target Config (From GQL query) were compared directly over the two dictionaries to check if there were changes in the configuration. If a schema change is introduced in the Saas files schema, the comparison detects changes because graphQL queries returns all the query attributes, whether are set or not. With the new promotion_data schema change, all the saas files has configuration changes.

To solve this, the comparison between configurations has been changed to remove all None values in the dictionaries. This way we ensure we are checking the effective configuration set instead of all the structure stored with null values. Check the comment above to see the code. 

## Notes. 
 * Testing has been improved to detect this details and with comments.


reconcile/utils/saasherder.py
```diff
@@ -1112,26 +1193,15 @@ def get_configs_diff_saas_file(self, saas_file):
...
-                    'cluster_name': cluster_name,
-                    'namespace_name': namespace_name,
-                    'target_config': desired_target_config
-                }
-                trigger_specs.append(job_spec)
-
-        return trigger_specs
+                )
+                # Convert to dict, ChainMap is not JSON serializable
+                # desired_target_config needs to be serialized to generate
```
lechuk47, line 1201: Problem 1 Fix.

reconcile/utils/saasherder.py
```diff
@@ -1084,23 +1105,83 @@ def get_configs_diff(self):
...
+        # Dict by key
+        targets = self.get_saas_targets_config(saas_file)
+
+        pipelines_provider = self._get_pipelines_provider(saas_file)
+        configurable_resources = saas_file.get('configurableResources', False)
+        trigger_specs = []
+
+        for key, desired_target_config in targets.items():
+            current_target_config = self.state.get(key, None)
+            # Continue if there are no diffs between configs.
```
lechuk47, line 1128: Problem 2 Fix

reconcile/utils/saasherder.py
```diff
@@ -1084,23 +1105,83 @@ def get_configs_diff(self):
...
+            # Compare existent values only, gql queries return None
+            # values for non set attributes so any change in the saas
+            # schema will trigger a job even though the saas file does
+            # not have the new parameters set.
+            ctc = SaasHerder.removeNoneValues(current_target_config)
+            dtc = SaasHerder.removeNoneValues(desired_target_config)
+            if ctc == dtc:
+                continue
+
```
lechuk47, line 1137: Problem 2 Fix.

reconcile/utils/saasherder.py
```diff
@@ -1112,26 +1193,15 @@ def get_configs_diff_saas_file(self, saas_file):
...
-                }
-                trigger_specs.append(job_spec)
-
-        return trigger_specs
+                )
+                # Convert to dict, ChainMap is not JSON serializable
+                # desired_target_config needs to be serialized to generate
+                # its config hash and to be stored in S3
+                configs[key] = dict(desired_target_config)
+        return configs
```
lechuk47, line 1204: Problem 1 Fix.

reconcile/utils/saasherder.py
```diff
@@ -1084,23 +1105,83 @@ def get_configs_diff(self):
...
+            current_target_config = self.state.get(key, None)
+            # Continue if there are no diffs between configs.
+            # Compare existent values only, gql queries return None
+            # values for non set attributes so any change in the saas
+            # schema will trigger a job even though the saas file does
+            # not have the new parameters set.
+            ctc = SaasHerder.remove_none_values(current_target_config)
+            dtc = SaasHerder.remove_none_values(desired_target_config)
+            if ctc == dtc:
+                continue
```
lechuk47, line 1136: Problem 2 fix.

geoberle (approved):
lgtm

PR #2067, opened 2021-12-27
allow namespaces to use cluster-admin token for openshift resource mgmt

geoberle:
# Summary
This change introduces a field `clusterAdmin` in a `namespace-1.yml` file similar to how `saas` files use this flag. When it is set to true, the cluster-admin token of a cluster is used to reconcile resources.

~~This change needs more testing, i opened this MR to get a feeling what people think of the general approach~~

# Approach
Two things needed touching to implement this change: the OC_Map and the ResourceInventory initialisation and usage

## OC_Map
Instead of tracking all clusters in one dict and having either all of them authenticated with dedicated-admin or cluster-admin, this change introduces the ability for clusters being present in a cluster map both ways at the same time. This is mostly relevant (and necessary) for `OC_Map` initialisation based on a set of namespaces, where one namespace requires cluster-admin while another one on the same cluster can work with dedicated-admin.

An alternate approach would have been to initialise the cluster with the highest required permission, but that had apparent security problems, allowing a non-admin namespace to escalate permissions beyond what has been declared in the namespace file.

Due to this change, the API of `OC_Map` changed, specifically the `get(cluster)` method got a new optional parameter `privileged` (defaults to false). Client code using the get method needs to be aware what permissions are required and needs to provide the `privileged` parameter accordingly. The `openshift-saas-deploy` and `openshift-resources` integrations are the only ones leveraging cluster-admin permissions right now, and both of them have been changed to pass the `privileged` parameter based on the `clusterAdmin` field in saas and namespace files. All other integrations do not required cluster-admin permissions and will still automatically use the dedicated-admin authenticated oc client.

## ResourceInventory
The `ResourceInventory` will also keep track of which resource in the `desired` state can use privileged cluster access for reconciliation. For this the API of `ResourceInventory` changed: the `add_desired` method got an additional optional parameter `privileged` (defaults to false). `openshift_resource_base.py` and the `SaasHerder` have been changed to feed information based on their respective `clusterAdmin` sources through the state spec to the `privileged` parameter of `add_desired`.

Related schema change: https://github.com/app-sre/qontract-schemas/pull/41
Part of: https://issues.redhat.com/browse/APPSRE-4195

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -36,7 +36,8 @@ class ValidationErrorJobFailed(Exception):
 
 class StateSpec:
     def __init__(self, type, oc, cluster, namespace, resource, parent=None,
-                 resource_type_override=None, resource_names=None):
+                 resource_type_override=None, resource_names=None,
+                 privileged=False):
         self.type = type
         self.oc = oc
         self.cluster = cluster
@@ -45,6 +46,7 @@ class StateSpec:
         self.parent = parent
         self.resource_type_override = resource_type_override
         self.resource_names = resource_names
+        self.privileged = privileged
 
 
 def init_specs_to_fetch(ri: ResourceInventory, oc_map: OC_Map,
@@ -69,7 +71,8 @@ def init_specs_to_fetch(ri: ResourceInventory, oc_map: OC_Map,
                 continue
 
             cluster = namespace_info['cluster']['name']
-            oc = oc_map.get(cluster)
+            privileged = namespace_info.get("clusterAdmin", False)
+            oc = oc_map.get(cluster, privileged)
             if not oc:
                 if oc.log_level >= logging.ERROR:
                     ri.register_error()
@@ -148,7 +151,8 @@ def init_specs_to_fetch(ri: ResourceInventory, oc_map: OC_Map,
             openshift_resources = namespace_info.get('openshiftResources')
             for openshift_resource in openshift_resources or []:
                 d_spec = StateSpec("desired", oc, cluster, namespace,
-                                   openshift_resource, namespace_info)
+                                   openshift_resource, namespace_info,
+                                   privileged=privileged)
                 state_specs.append(d_spec)
     elif clusters:
         # set namespace to something indicative
@@ -250,11 +254,11 @@ def wait_for_namespace_exists(oc, namespace):
         raise Exception(f'namespace {namespace} does not exist')
 
 
-def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
-          wait_for_namespace, recycle_pods=True):
+def apply(dry_run, oc_map: OC_Map, cluster, namespace, resource_type, resource,
+          wait_for_namespace, privileged: bool, recycle_pods=True):
     logging.info(['apply', cluster, namespace, resource_type, resource.name])
 
-    oc = oc_map.get(cluster)
+    oc = oc_map.get(cluster, privileged)
     if not oc:
         logging.log(level=oc.log_level, msg=oc.message)
         return None
@@ -335,14 +339,14 @@ def create(dry_run, oc_map, cluster, namespace, resource_type, resource):
 
 
 def delete(dry_run, oc_map, cluster, namespace, resource_type, name,
-           enable_deletion):
+           enable_deletion, privileged: bool):
     logging.info(['delete', cluster, namespace, resource_type, name])
 
     if not enable_deletion:
         logging.error('\'delete\' action is disabled due to previous errors.')
         return
 
-    oc = oc_map.get(cluster)
+    oc = oc_map.get(cluster, privileged)
     if not oc:
         logging.log(level=oc.log_level, msg=oc.message)
         return None
@@ -360,7 +364,8 @@ def check_unused_resource_types(ri):
 
 
 def _realize_resource_data(unpacked_ri_item,
-                           dry_run, oc_map, ri,
+                           dry_run, oc_map: OC_Map,
+                           ri: ResourceInventory,
                            take_over,
                            caller,
                            wait_for_namespace,
@@ -439,15 +444,17 @@ def _realize_resource_data(unpacked_ri_item,
                       OR.serialize(OR.canonicalize(d_item.body)))
 
         try:
+            privileged = data['use_admin_token'].get(name, False)
             apply(dry_run, oc_map, cluster, namespace,
                   resource_type, d_item, wait_for_namespace,
-                  recycle_pods=recycle_pods)
+                  privileged, recycle_pods=recycle_pods)
             action = {
                 'action': ACTION_APPLIED,
                 'cluster': cluster,
                 'namespace': namespace,
                 'kind': resource_type,
-                'name': d_item.name
+                'name': d_item.name,
+                'privileged': privileged
             }
             actions.append(action)
         except StatusCodeError as e:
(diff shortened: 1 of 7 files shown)


reconcile/utils/oc.py
```diff
@@ -1087,18 +1088,31 @@ def __init__(self, clusters=None, namespaces=None,
...
+                # declares clusterAdmin token usage, many of those places are
+                # happy with regular dedicated-admin and will request a cluster
+                # with oc_map.get(cluster) without specifying privileged access
+                # specifically
+                c = ns_info['cluster']
+                clusters[c['name']] = c
+                privileged = ns_info.get("clusterAdmin", False) or \
+                    cluster_admin
+                if privileged:
+                    privileged_clusters[c['name']] = c
```
patjlm, line 1107: Would it make sense to raise an error here if `clusterAdminAutomationToken` is missing from the cluster ? This would allow MR/dry-runs to fail if `clusterAdmin: True` but the token is not configured in the cluster.

patjlm (commented):
The idea looks great to me. I just added a comment on early failure in case of misconfiguration. Otherwise, LVGTM !

reconcile/utils/oc.py
geoberle: the integration will fail if a required token is missing (regardless of dedicated-admin or cluster-admin) but it will do whatever work can be done before failing. the path is like this for the openshift-resources integration:

* during oc initialisation, a [missing token results in registering an OCLogMsg](https://github.com/geoberle/qontract-reconcile/blob/c4405e8d90ce45dd371f6f53cc06e8424d84e0af/reconcile/utils/oc.py#L1146-L1161) instead of an oc client
* during state spec initialisation, the problem with a cluster is [detected and registered](https://github.com/geoberle/qontract-reconcile/blob/c4405e8d90ce45dd371f6f53cc06e8424d84e0af/reconcile/openshift_base.py#L75-L80) as an error in the ResourceInventory, but reconciliation is progressing with other resources and clusters
* when all remaining work is done, the integration [checks for registered problems](https://github.com/geoberle/qontract-reconcile/blob/c4405e8d90ce45dd371f6f53cc06e8424d84e0af/reconcile/openshift_resources_base.py#L713-L714) in the ResourceInventory and fails if any have been found

for openshift-saas-deploy, the general path is the same

patjlm: I see, so it's really up to each oc_map caller to handle the case, often via the resourceinventory error handling. This allows to go as far as possible.
Makes sense, thanks!

reconcile/openshift_base.py
```diff
@@ -36,7 +36,8 @@ class ValidationErrorJobFailed(Exception):
 
 class StateSpec:
     def __init__(self, type, oc, cluster, namespace, resource, parent=None,
-                 resource_type_override=None, resource_names=None):
+                 resource_type_override=None, resource_names=None,
+                 privileged=False):
```
Piojo, line 40: It is about time we convert this one into a dataclass, isn't it?

reconcile/openshift_base.py
```diff
@@ -250,11 +254,12 @@ def wait_for_namespace_exists(oc, namespace):
         raise Exception(f'namespace {namespace} does not exist')
 
 
-def apply(dry_run, oc_map, cluster, namespace, resource_type, resource,
-          wait_for_namespace, recycle_pods=True):
+def apply(dry_run, oc_map: OC_Map, cluster, namespace, resource_type, resource,
+          wait_for_namespace, recycle_pods: bool = True,
+          privileged: bool = False):
```
Piojo, line 259: While you're at it, can you please complete the type signature here? At least the return type?

reconcile/test/test_utils_oc.py
```diff
@@ -431,3 +431,100 @@ def test_clusters_errors_with_include_errors(
...
+        )
+
+        self.assertIsInstance(
+            oc_map.get(cluster_1['name']),
+            OC
+        )
+        self.assertIsInstance(
+            oc_map.get(cluster_1['name'], privileged=True),
+            OCLogMsg
+        )
```
Piojo, line 530: Considering that the code uses this as a boolean, wouldn't it make sense to do `self.assertFalse(oc_map.get(cluster_1['name'], privileged=True))`
?

rporres (commented):
I really like the idea of enhancing OC_Map so that it only uses elevated privileges when needed. LGTM!

Piojo (commented):
LGTM. Can we have unit tests for the very simple `set_oc` and friends, please?

reconcile/openshift_base.py
Piojo: On a second thought, don't change it on this PR, let's do it at some point in the near future.

reconcile/test/test_utils_oc.py
geoberle: fixed in 2e55eb1

geoberle:
> LGTM. Can we have unit tests for the very simple `set_oc` and friends, please?

@Piojo can you be a bit more specific? `set_oc` gets called during the `OC_Map` initialisation that the correct behaviour is tested several times in test_utils_oc.py, e.g. within `test_namespace_with_cluster_admin`. or do you mean a few edge cases around `set_oc` like `cluster_disabled` or `internal` clusters?

Piojo:
> `set_oc` gets called during the `OC_Map` initialisation and the correct behaviour is tested several times in test_utils_oc.py, e.g. within `test_namespace_with_cluster_admin`

That's a bit obscure, but I guess we can accept it.

>  do you mean a few edge cases around set_oc like cluster_disabled or internal clusters?

Yes, please. Let's cover these edge cases as well.

geoberle:
@Piojo wdyt?

Piojo (approved):
Works for me. Thanks!

reconcile/openshift_base.py
```diff
@@ -69,7 +71,8 @@ def init_specs_to_fetch(ri: ResourceInventory, oc_map: OC_Map,
                 continue
 
             cluster = namespace_info['cluster']['name']
-            oc = oc_map.get(cluster)
+            privileged = namespace_info.get("clusterAdmin", False)
```
maorfr: since this is a graphql response, the default value will not be returned even if `clusterAdmin` is not defined. the returned value will be None.
you'd probably want `namespace_info.get("clusterAdmin") or False`

reconcile/utils/openshift_resource.py
```diff
@@ -443,17 +443,23 @@ def initialize_resource_type(self, cluster, namespace, resource_type):
...
             desired = \
                 (self._clusters[cluster][namespace][resource_type]
                     ['desired'])
             if name in desired:
                 raise ResourceKeyExistsError(name)
             desired[name] = value
+            admin_token_usage = \
+                (self._clusters[cluster][namespace][resource_type]
+                    ['use_admin_token'])
+            admin_token_usage[name] = privileged
```
maorfr, line 462: from reading this i understand that we are maintaining "privileged" per resource? and since many resources are within the same namespace it means we will do this multiple times? shouldn't it be per namespace?

maorfr (approved):
LGTM

some minor nits for your consideration.

reconcile/utils/openshift_resource.py
geoberle: yes the current schema change attached to this PR is managing privileged access on the namespace level but the change to the `ResourceInventory` allows per-resource permission tracking.

when i started to write this patch, my knowledge about resource management was only basic and i made the decision to handle this task per-namespace. while digging deeper, i found that the ResourceInventory was already pretty close to what is needed to track this per-resource. the door is open to also do the last step to allow per-resource declaration in the schema.

maybe in a follow up PR? how do you feel about this?

reconcile/openshift_base.py
geoberle: yes i fell for this every time :) but it is already fixed in 3bb6e27aab73df04125eb94ea7dd3668cd44b16a

reconcile/utils/openshift_resource.py
maorfr: i was actually suggesting the other way - to do it only per namespace. happy to talk about this more if you'd like, but you can also decide yourself since things will work the same either way.

geoberle: left a comment about this in e5151b6

PR #2071, opened 2021-12-28
[terraform-aws-route53] add identifier to support split view dns

maorfr:
depends on https://github.com/app-sre/qontract-schemas/pull/43

with a unique identifier, we can specify multiple zones of the same name to support features such as split view DNS: https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/hosted-zone-private-considerations.html#hosted-zone-private-considerations-split-view-dns

### DIFF
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -488,7 +488,7 @@ class TerrascriptClient:
             acct_name = zone['account_name']
 
             # Ensure zone is in the state for the given account
-            zone_id = safe_resource_id(f"{zone['name']}")
+            zone_id = safe_resource_id(f"{zone.pop('identifier')}")
             zone_values = {
                 'name': zone['name'],
                 'vpc': zone.get('vpc'),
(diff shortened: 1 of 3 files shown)


reconcile/utils/terrascript_client.py
```diff
@@ -488,7 +488,7 @@ def populate_route53(self, desired_state, default_ttl=300):
             acct_name = zone['account_name']
 
             # Ensure zone is in the state for the given account
-            zone_id = safe_resource_id(f"{zone['name']}")
+            zone_id = safe_resource_id(f"{zone.pop('identifier')}")
```
Piojo, line 491: Modifying `zone` inside the f-string - or a read-only function argument for that matter - is going to cause confusion. 
```suggestion
            zone_id = zone.pop('identifier')
            zone_id = safe_resource_id(zone_id)
```
(or whatever if `zone_id` becomes `None`.

Piojo (changes requested):

reconcile/utils/terrascript_client.py
maorfr: fair enough. https://github.com/app-sre/qontract-reconcile/pull/2071/commits/779a6a07f3efa1546ac9a004504192229fa674c8

jfchevrette (approved):

Piojo (approved):
Thanks for addressing this, and sorry for the late review.

PR #2073, opened 2021-12-30
[terraform-aws-route53] enable deletion

maorfr:
the integration should be able to delete resources by default. it already does.
the deletion detection is not used in this integration (`disabled_deletions_detected` is dropped in the return values from `tf.plan`).

the only problem is that we are printing a warning. this PR will remove the warning.


Piojo (approved):
LGTM

PR #2077, opened 2022-01-03
Move openshift-tekton-resources tests to pytest

rporres:
We're standarizing our test framework.

part of [APPSRE-4190](https://issues.redhat.com/browse/APPSRE-4190)


rporres:
If I started this from scratch, I might use pytest native fixtures, but at this point I don't feel like rewriting it completely since pytest supports setup/teardown methods...

Piojo (approved):
LGTM

PR #2078, opened 2022-01-04
Fix remove_none_values method

lechuk47:
## Description 

remove_none_value method has been changed to work with a None value input. This could happen when a new saas target is deployed and there is no current configuration stored in the State


geoberle (approved):
LGTM

PR #2081, opened 2022-01-05
Create SNS event notification for S3 

OliviaHY:
### Why:
Be able to create SNS notification through App Interface for existing topic and bucket.
### What:
1. Change the query to get the event_notification information;
2. Add it in the terraform-resources integration to loop through event_notification and create SNS topic;
### Validation:
Local run was able to create it.

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -52,6 +52,13 @@ provider
   overrides
   sqs_identifier
   s3_events
+  event_notifications {
+    destination_type
+    destination
+    event_type
+    filter_prefix
+    filter_suffix
+  }
   bucket_policy
   output_resource_name
   storage_class
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -97,7 +97,7 @@ VARIABLE_KEYS = ['region', 'availability_zone', 'parameter_group',
                  'enhanced_monitoring', 'replica_source',
                  'output_resource_db_name', 'reset_password', 'ca_cert',
                  'sqs_identifier', 's3_events', 'bucket_policy',
-                 'storage_class', 'kms_encryption',
+                 'event_notifications', 'storage_class', 'kms_encryption',
                  'variables', 'policies', 'user_policy',
                  'es_identifier', 'filter_pattern',
                  'specs', 'secret', 'public', 'domain',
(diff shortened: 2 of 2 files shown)


reconcile/utils/terrascript_client.py
```diff
@@ -1527,7 +1527,7 @@ def populate_tf_resource_s3(self, resource, namespace_info):
             tf_resources.append(sqs_data)
 
             s3_events = common_values.get('s3_events', ["s3:ObjectCreated:*"])
-            notification_values = {
+            notification = {
```
OliviaHY, line 1530: Eventually this part will be deleted, and its functionality will be take care by the loop added at line 1549. This is mostly trading duplicated code for backwards compatibility. Plan to migrate the date, then consolidate the code.

reconcile/terraform_resources.py
```diff
@@ -52,6 +52,11 @@
   overrides
   sqs_identifier
   s3_events
+  event_notifications {
+    destination_type
+    destination
+    event_type
```
BumbleFeng, line 58: I think you need to add filter_prefix and filter_suffix here.

reconcile/utils/terrascript_client.py
```diff
@@ -1536,17 +1536,72 @@ def populate_tf_resource_s3(self, resource, namespace_info):
...
+
+            if detination_identifier is not None:
+                identifier_is_arn = True \
+                    if detination_identifier.startswith('arn:') else False
+            if identifier_is_arn:
+                resource_name = detination_identifier.split(':')[-1]
+                resource_arn = detination_identifier
+            else:
+                resource_name = detination_identifier
+                resource_arn = '${data.aws_sns_topic.' \
```
BumbleFeng, line 1566: data.aws_sns_topic or data.aws_sqs_queue

need to be  determined by destination_type

reconcile/utils/terrascript_client.py
```diff
@@ -1536,17 +1536,72 @@ def populate_tf_resource_s3(self, resource, namespace_info):
...
+
+            prefix = event_notification.get('filter_prefix', None)
+            if prefix is not None:
+                notification[notification_type][0]['filter_prefix'] = prefix
+            suffix = event_notification.get('filter_suffix', None)
+            if suffix is not None:
+                notification[notification_type][0]['filter_suffix'] = suffix
+
+            notification_tf_resource = aws_s3_bucket_notification(
+                identifier+'-'+resource_name, **notification)
```
BumbleFeng, line 1604: Ideally, we can append all notification into one aws_s3_bucket_notification resource with the values of
```
            notification = {
                'bucket': '${' + bucket_tf_resource.id + '}',
                'topic': [list of SNS Topic],
                'queue': [list of SQS Queue]
            }

```
I think this will be more clear, wdyt?

OliviaHY: Good point, I'll make this part explicit and a list that can be appended.

reconcile/utils/terrascript_client.py
```diff
@@ -1549,6 +1549,77 @@ def populate_tf_resource_s3(self, resource, namespace_info):
                 sqs_identifier, **notification_values)
             tf_resources.append(notification_tf_resource)
 
+        event_notifications = common_values.get('event_notifications')
+        sns_notifications = []
+        sqs_notifications = []
+
+        for event_notification in event_notifications:
+            destination_type = event_notification['destination_type']
+            detination_identifier = event_notification['destination']
```
fishi0x01, line 1558: Small typo - `detination_` -> `destination_` (not important though :smiley:  )

reconcile/utils/terrascript_client.py
```diff
@@ -1549,6 +1549,77 @@ def populate_tf_resource_s3(self, resource, namespace_info):
...
+
+            if destination_type == 'sns':
+                notification_type = 'topic'
+                resource_arn_data = 'data.aws_sns_topic'
+            elif destination_type == 'sqs':
+                notification_type = 'queue'
+                resource_arn_data = 'data.aws_sqs_queue'
+
+            if detination_identifier is not None:
+                identifier_is_arn = True \
```
fishi0x01, line 1569: Not sure if Im overlooking something here, but I think the `if else` is redundant. The following should produce the same

```
identifier_is_arn = destination_identifier.startswith('arn:')
```

OliviaHY: Good catch!

reconcile/utils/terrascript_client.py
```diff
@@ -1549,6 +1549,74 @@ def populate_tf_resource_s3(self, resource, namespace_info):
...
+
+            resource_values = {
+                'name': resource_name
+            }
+            resource_provider = values.get('provider')
+            if resource_provider:
+                resource_values['provider'] = resource_provider
+
+            if destination_type == 'sns':
+                resource_data = data.aws_sns_topic(resource_name,
```
BumbleFeng, line 1597: I think we need to add resource_data only if destination_identifier does not startswith('arn:'), otherwise terraform can not find the target resource.

BumbleFeng (approved):

PR #2082, opened 2022-01-06
[endpoint monitoring] integration backed by blackbox-exporter

geoberle:
## Scope
this new integration `blackbox-exporter-endpoint-monitoring` implements the blackbox-exporter part (phase 1) of service
endpoint monitoring as defined in this [design doc](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/service-endpoint-monitoring.md)

Schema change: https://github.com/app-sre/qontract-schemas/pull/45
Part of: https://issues.redhat.com/browse/APPSRE-4209

## What was introduced
With this change, tenants can declare blackbox-exporter backed `EndpointMonitoringProviders` and attach app endpoints to it to setup basic closed box monitoring. This concept of `EndpointMonitoringProviders` will be enhanced in a followup PR where Catchpoint probes are going to be implemented as an additional provider (phase 2 of design doc).

## Divergence from design docs
* use `Probe` instead of `ServiceMonitor` to declare scrape configs for Prometheus because referencing the in-cluster blackbox-exporter via `ServiceMonitor` would result in scraping all blackbox-exporter pods
* the schema `/dependencies/endpoint-monitoring-provider-1.yml` defines an additional field (dict) `metricLabels`. those are added as target labels while scraping to give a bit more context to define alerts

## Example

```yaml
---
$schema: /dependencies/endpoint-monitoring-provider-1.yml

name: blackbox-exporter-http-2xx-stage
description: Checks a URL for HTTP 2xx status codes via blackbox-exporter
provider: blackbox-exporter
metricLabels:
  environment: staging
blackboxExporter:
  namespace:
    $ref: /services/observability/namespaces/openshift-customer-monitoring.xxxx.yml
  exporterUrl: xxxx.svc.cluster.local:9115/probe
  module: http_2xx
```

and

```yaml
---
$schema: /app-sre/app-1.yml
...
endPoints:
- name: test
  description: test
  url: https://httpstat.us/200
  monitoring:
    provider:
      $ref: /dependencies/monitoring/blackbox-exporter-http-2xx-stage.yml
```

this would result in a `Probe` CR (similar to a `ServiceMonitor` but with concrete URLs instead of `Endpoint` discovery)

```yaml
apiVersion: monitoring.coreos.com/v1
kind: Probe
metadata:
  name: blackbox-exporter-http-2xx-stage
  namespace: blackbox_exporter.namespace.name
  labels:
    prometheus: app-sre
spec: 
  jobName: blackbox-exporter-http-2xx-stage
  module: http_2xx
  prober:
    scheme: http
    url: xxxx.svc.cluster.local:9115
    path: /probe
  targets:
    staticConfig: 
      relabelingConfigs:
        action: labeldrop
        regex: namespace
      labels:
        environment: staging
      static: 
      - https://httpstat.us/200
```

This will then result in the following metrics produced by scraping blackbox-exporter

```
up{job="${provider.name}", ${metricLabels}, instance="${app.endPoint.url}"}
probe_*{job="${provider.name}", ${metricLabels}, instance="{app.endPoint.url}", ...}
```

which boils down to

```
up{job="blackbox-exporter-http-2xx-stage", environment="staging", instance="https://httpstat.us/200"}
probe_*{job="blackbox-exporter-http-2xx-stage", environment="staging", instance="https://httpstat.us/200", ...}
```

### DIFF
--- a/reconcile/blackbox_exporter_endpoint_monitoring.py
+++ b/reconcile/blackbox_exporter_endpoint_monitoring.py
@@ -80,7 +80,7 @@ def build_probe(provider: EndpointMonitoringProvider,
     blackbox_exporter = provider.blackboxExporter
     if blackbox_exporter:
         prober_url = parse_prober_url(blackbox_exporter.exporterUrl)
-        body = {
+        body: dict[str, Any] = {
             "apiVersion": "monitoring.coreos.com/v1",
             "kind": "Probe",
             "metadata": {
@@ -92,15 +92,17 @@ def build_probe(provider: EndpointMonitoringProvider,
             },
             "spec": {
                 "jobName": f"endpoint-monitoring-{provider.name}",
-                "interval": provider.checkInterval,
+                "interval": provider.checkInterval or "10s",
                 "module": blackbox_exporter.module,
                 "prober": prober_url,
                 "targets": {
                     "staticConfig": {
-                        "relabelingConfigs": {
-                            "action": "labeldrop",
-                            "regex": "namespace"
-                        },
+                        "relabelingConfigs": [
+                            {
+                                "action": "labeldrop",
+                                "regex": "namespace"
+                            }
+                        ],
                         "labels": provider.metric_labels,
                         "static": [
                             ep.url for ep in endpoints
@@ -109,6 +111,8 @@ def build_probe(provider: EndpointMonitoringProvider,
                 }
             }
         }
+        if provider.timeout:
+            body["spec"]["scrapeTimeout"] = provider.timeout
         return OpenshiftResource(
             body,
             QONTRACT_INTEGRATION,
--- a/reconcile/blackbox_exporter_endpoint_monitoring.py
+++ b/reconcile/blackbox_exporter_endpoint_monitoring.py
@@ -130,8 +130,7 @@ def get_endpoints() -> dict[EndpointMonitoringProvider, list[Endpoint]]:
             if monitoring:
                 if monitoring["provider"]["provider"] == "blackbox-exporter":
                     ep = Endpoint(**ep_data)
-                    if ep.monitoring.provider not in endpoints:
-                        endpoints[ep.monitoring.provider] = []
+                    endpoints.setdefault(ep.monitoring.provider, [])
                     endpoints[ep.monitoring.provider].append(ep)
 
     return endpoints
--- a/reconcile/blackbox_exporter_endpoint_monitoring.py
+++ b/reconcile/blackbox_exporter_endpoint_monitoring.py
@@ -1,0 +1,176 @@
+import logging
+import sys
+from urllib.parse import urlparse
+from typing import Any, Optional
+import json
+from dataclasses import field
+
+from pydantic.dataclasses import dataclass
+
+from reconcile import queries
+from reconcile.utils.openshift_resource import (
+    OpenshiftResource, ResourceInventory)
+from reconcile.utils.defer import defer
+from reconcile.utils.semver_helper import make_semver
+import reconcile.openshift_base as ob
+
+QONTRACT_INTEGRATION = "blackbox-exporter-endpoint-monitoring"
+QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
+
+LOG = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, eq=True)
+class BlackboxMonitoringProvider:
+
+    module: str
+    namespace: dict[str, Any] = field(compare=False, hash=False)
+    exporterUrl: str
+
+
+@dataclass(frozen=True, eq=True)
+class EndpointMonitoringProvider:
+
+    name: str
+    provider: str
+    description: str
+    timeout: Optional[str] = None
+    checkInterval: Optional[str] = None
+    blackboxExporter: Optional[BlackboxMonitoringProvider] = None
+    metricLabels: Optional[str] = None
+
+    @property
+    def metric_labels(self):
+        return json.loads(self.metricLabels) if self.metricLabels else {}
+
+
+@dataclass
+class Endpoint:
+
+    name: str
+    description: str
+    url: str
+
+    @dataclass
+    class Monitoring:
+
+        provider: EndpointMonitoringProvider
+
+    monitoring: Monitoring
+
+
+def parse_prober_url(url: str) -> dict[str, str]:
+    parsed_url = urlparse(url)
+    if parsed_url.scheme not in ["http", "https"]:
+        raise ValueError(
+            "the prober URL needs to be an http:// or https:// one "
+            f"but is {url}"
+        )
+    data = {
+        "url": parsed_url.netloc,
+        "scheme": parsed_url.scheme
+    }
+    if parsed_url.path:
+        data["path"] = parsed_url.path
+    return data
+
+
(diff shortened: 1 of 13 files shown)


reconcile/blackbox_exporter_endpoint_monitoring.py
```diff
@@ -0,0 +1,135 @@
...
+    if blackbox_exporter:
+        prober_url = parse_prober_url(blackbox_exporter.exporterUrl)
+        body = {
+            "apiVersion": "monitoring.coreos.com/v1",
+            "kind": "Probe",
+            "metadata": {
+                "name": f"endpoint-monitoring-{provider.name}",
+                "namespace": blackbox_exporter.namespace.name,
+                "labels": {
+                    "prometheus": "app-sre"
```
geoberle, line 49: probe labels (prometheus selector) should not be hard coded. this is too specific for AppSRE

reconcile/blackbox_exporter_endpoint_monitoring.py
```diff
@@ -0,0 +1,180 @@
...
+    if blackbox_exporter:
+        prober_url = parse_prober_url(blackbox_exporter.exporterUrl)
+        body: dict[str, Any] = {
+            "apiVersion": "monitoring.coreos.com/v1",
+            "kind": "Probe",
+            "metadata": {
+                "name": f"endpoint-monitoring-{provider.name}",
+                "namespace": blackbox_exporter.namespace.get("name"),
+                "labels": {
+                    "prometheus": "app-sre"
```
geoberle, line 90: tight too much towards app-sre. should we introduce prometheus selector labels in the [EndpointMonitoringProviderBlackboxExporterSettings_v1](https://github.com/app-sre/qontract-schemas/pull/45/files#diff-6bad3cee90319b7494304e3ce4e439992becd03f9d8825eca7f44687f8184ee9R2260) schema?

reconcile/queries.py
```diff
@@ -2434,3 +2434,53 @@ def get_dyn_traffic_directors():
 def get_status_pages():
     gqlapi = gql.get_api()
     return gqlapi.query(STATUS_PAGE_QUERY)['status_pages']
+
+
+CLOSED_BOX_MONITORING_PROBES_QUERY = """
+{
+  apps: apps_v1 {
```
maorfr, line 2441: why are you querying apps and not the endpoint resources directly?

geoberle: the endpoints are not dedicated "objects" but embedded into an app. i'm not aware on how to query those without starting from apps_v1

but what i can do is remove the fields from app because i don't need them

maorfr: oh, sorry. i misunderstood. please disregard!

reconcile/blackbox_exporter_endpoint_monitoring.py
```diff
@@ -0,0 +1,180 @@
+import logging
+import sys
+from urllib.parse import urlparse
+from typing import Any, Optional
+import json
+from dataclasses import field
+
+from pydantic.dataclasses import dataclass
```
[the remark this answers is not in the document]
geoberle: regular python dataclasses can only handle top level objects on object instantiation. all nested dataclass objects just get mapped as dict. pydantic dataclasses handle correctly trees of objects as we have them a lot in app-interface.

reconcile/blackbox_exporter_endpoint_monitoring.py
rporres: There's only other prometheus that we manage, the prometheus aggregator instance in Quay, and we're not going to use it for this. So I don't think it is worth the effort. If we ever need it, let's add it.

rporres:
This looks great. I really like the usage of pydantic dataclasses to be able to better navigate into complex structures.

My main doubt is if there's any problem in this growing and having many probes only with one target, while we can have a list of `static` targets. I was wondering if it would be worth to group probes per common labels or if it is premature optimization.

geoberle:
> My main doubt is if there's any problem in this growing and having many probes only with one target, while we can have a list of `static` targets. I was wondering if it would be worth to group probes per common labels or if it is premature optimization.

This was also my concern. All endpoints referencing the same provider are grouped as targets under the same `Probe` and each `Probe` is added to the prometheus config as one job. As long as we don't have a really high number of providers of type blackbox-exporter, we should be fine - and i also don't expect too many of those providers.

For Catchpoint, the approach is going to be different anyways. Monitoring is done in Catchpoint and the result metrics are going to be scraped by just one prometheus job.

rporres (approved):
awesome!

reconcile/blackbox_exporter_endpoint_monitoring.py
```diff
@@ -0,0 +1,180 @@
...
+QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
+
+LOG = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, eq=True)
+class BlackboxMonitoringProvider:
+
+    module: str
+    namespace: dict[str, Any] = field(compare=False, hash=False)
```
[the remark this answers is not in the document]
Piojo: Not really, but then add a comment here please.

Piojo (approved):
LGTM. Thanks for the explanations.

reconcile/blackbox_exporter_endpoint_monitoring.py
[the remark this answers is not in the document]
geoberle: comment added in 20ebc3c

patjlm:
Just wondering: would it make sense to support multiple `provider/$ref` under endpoint `monitoring` in `/app-sre/app-1.yml` ?

geoberle:
> Just wondering: would it make sense to support multiple `provider/$ref` under endpoint `monitoring` in `/app-sre/app-1.yml` ?

could be interesting for cases where multiple different tests go towards the same target. you mean something like this?

```yaml
---
$schema: /app-sre/app-1.yml
...
endPoints:
- name: test
  description: test
  url: https://httpstat.us/200
  monitoring:
  - provider:
       $ref: /dependencies/monitoring/blackbox-exporter-http-2xx.yml
  - provider:
       $ref: /dependencies/monitoring/catchpoint_extravaganza.yml
```

patjlm:
> But for those cases we could also use only one "powerful" test that covers multiple things at the same time, e.g. with Catchpoints scripted tests.

This could mean a bit a repetition/duplication of code, right ? simple url ping, cert validation would be duplicated in those tests.
I am not familiar with Catchpoint. If it is possible to have something like a library of 'unit' tests that we can aggregate into a 'scripted test', I am fine ;-)

geoberle:
> This could mean a bit a repetition/duplication of code, right ? simple url ping, cert validation would be duplicated in those tests. I am not familiar with Catchpoint. If it is possible to have something like a library of 'unit' tests that we can aggregate into a 'scripted test', I am fine ;-)

i'm not aware of elborate libraries. but they have javascript functions available to make the HTTP callout to a URL. by doing that, you get roughly the same metrics as a blackbox_exporter check with the http_xxx modules (including metrics about ssl and certs). the part that justifies catchpoint usually comes after that, verifying the results in a scripted way etc.

geoberle:
@patjlm thank for discussing this with me. multiple providers are now supported as proposed above

PR #2083, opened 2022-01-07
Larger defaults for tekton tasks

rporres:
We have actually overriden a large amount of files. It is time to
avoid problems once and for all.


geoberle (approved):
lgtm

PR #2084, opened 2022-01-08
move statuspage-component tests to pytest

geoberle:
We're standarizing our test framework.

part of APPSRE-4190

### DIFF
--- a/reconcile/test/test_status_page_components.py
+++ b/reconcile/test/test_status_page_components.py
@@ -1,4 +1,3 @@
-from unittest.mock import patch
 from typing import Optional
 import re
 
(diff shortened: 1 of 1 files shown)


reconcile/test/test_status_page_components.py
```diff
@@ -67,296 +69,275 @@ def f(self):
...
-            get_atlassian_component_fixtures(fixture_name, "page_1")
-        create_mock.return_value = None
-        page = get_page_fixtures(fixture_name)[0]
-
-        page.reconcile(False, StateStub())
-
-        create_mock.assert_called_with(
-            component_to_dict(page.components[0], "group_id_1")
-        )
+@patch.object(VaultSecretRef, '_resolve_secret',
```
steveteahan, line 72: I'd suggest moving these to use the `mocker` pytest fixture. The same goes for other tests, but I won't repeat the same comment elsewhere. This is covered more here: https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-testing-guide/#pytest-best-practices

It just helps get away from using decorators because pytest uses those for other uses cases like parameterizing tests, defining fixtures, etc. You'll run into issues of ordering of decorators with the args passed in the test function and it just isn't ideal. The change should be minimal, it'd just look something like:

```python
def test_create_component(mocker):
    fetch_mock = mocker.patch.object(AtlassianStatusPage, '_fetch_components')
    create_mock = ...
    vault_mock = ...

    [the rest of your code stays the same]
    ...
```

Also, while we're fixing these, adding `autospec=True` might not be a bad idea.

steveteahan (commented):
Thanks for updating these to use pytest!

reconcile/test/test_status_page_components.py
geoberle: Thanks for your feedback. I will update accordingly!

geoberle: remove decorators in 1be1e45
use autospec in 2225936

reconcile/test/test_status_page_components.py
```diff
@@ -67,296 +69,295 @@ def f(self):
...
+    provider.apply_component(dry_run, component)
+
+    if dry_run:
+        create_mock.assert_not_called()
+    else:
+        create_mock.assert_called()
+
+
+def test_dry_run_on_update(mocker):
+    _exec_dry_run_test_on_update(mocker, True)
```
steveteahan, line 280: This isn't really specific to moving to pytest and therefore not a blocker, but I think tests like this are a bit obscure. It's not really clear what's going on here until I go off and read `_exec_dry_run_test_on_update`. I'm not sure whether that level of DRY is really helpful or not.

steveteahan (approved):

reconcile/test/test_status_page_components.py
geoberle: i know what you mean. at the same time i always have this nagging feeling at the back of my head what bad things can happen when we have a bug somewhere not propagating dry_run information properly.

PR #2088, opened 2022-01-12
Improve gitlab-labeler

andreugallofre:
* correctly add onboarding status on parent apps without "child" updates on the MR
* Only add tenant information if path is under data/services

### DIFF
--- a/reconcile/gitlab_labeler.py
+++ b/reconcile/gitlab_labeler.py
@@ -54,6 +54,9 @@ def guess_onboarding_status(changed_paths: Iterable[str],
                 if child_app in apps:
                     app = apps[child_app]
                     labels.add(app['onboardingStatus'])
+                else:
+                    app = apps[app_name]
+                    labels.add(app['onboardingStatus'])
             else:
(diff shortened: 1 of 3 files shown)


reconcile/gitlab_labeler.py
```diff
@@ -88,8 +91,10 @@ def guess_labels(project_labels: Iterable[str],
         path_dir = os.path.dirname(path)
         path_tokens = path_dir.split('/')
 
-        tenants += [t for t in path_tokens if t not in ignore_tokens and
-                    t in apps]
+        if 'data/services/' in path:
+            tenants += [t for t in path_tokens if t not in ignore_tokens and
+                        t in apps]
+
```
rporres, line 97: Although I would love if we could find more heuristics to find labels, this should be enough for now, at least it will be real.

rporres:
> correctly add onboarding status on parent apps without "child" updates

Do you mean "apps without children"? Because you cannot be a parent if you don't have children 😬

rporres (approved):

PR #2089, opened 2022-01-13
add stickiness config for forwarding rule

jfchevrette:
It appears AWS is requiring this stickiness parameter even though the terraform docs mark it as optional

Fixes
```
[terraform-resources-wrapper] error: b'\nError: Error modifying LB Listener Rule: ValidationError: Target group stickiness duration must be between 1 and 604800 seconds\n\tstatus code: 400, request id: 676c4a12-20c3-430d-b254-a95f3dd20564\n\n  on config.tf.json line 2487, in resource.aws_lb_listener_rule.foo-alb01-rule-01:\n2487:       },\n\n\n'
```

### DIFF
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -3726,7 +3726,11 @@ class TerrascriptClient:
                 'action': {
                     'type': 'forward',
                     'forward': {
-                        'target_group': []
+                        'target_group': [],
+                        'stickiness': {
+                            'enabled': False,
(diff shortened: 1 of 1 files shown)


jfchevrette:
The setting had been removed [here](https://github.com/app-sre/qontract-reconcile/pull/2057/files#diff-2e9fe6fa68837df3e3628e8acea384a1427c2a2b4c5b45a1a96135ef3038eec9L3766) as part of #2057

syed (approved):

PR #2094, opened 2022-01-17
Make pipelines defaults mandatory

rporres:
We will make sure that all pipelines providers always have what they need
to work properly and will fail at a schema level without the need of any
ad-hoc checks in the the code.

part of APPSRE-4165

schema changes: https://github.com/app-sre/qontract-schemas/pull/49


andreugallofre (approved):
LGTM

maorfr (approved):

PR #2099, opened 2022-01-18
APPSRE-4229 ocm-clusters network type

patjlm:
Enable network type selection clusters
Linked to https://github.com/app-sre/qontract-schemas/pull/52
The default is the current only value: `OpenshiftSDN`


geoberle (approved):
LGTM

maorfr (approved):
i think it may be better to not add a default value if `type` is not specified and by that using OCM's default?
perhaps an approach more similar to how we handle `provision_shard_id` is more suitable?

in any case, this will work for sure, so LGTM

PR #2100, opened 2022-01-18
[tf-r]auto promote asg ami id

BumbleFeng:
Depends on https://github.com/app-sre/qontract-schemas/pull/53

Part of https://issues.redhat.com/browse/APPSRE-3925

Design doc: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/34670

First implementation. Tf-r will fail when the packer job is running or failed, as the commit has been pushed but the AMI has not been built. Will add upstream info for checking Jenkins job state in a second PR.

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -279,7 +279,9 @@ provider
   }
   variables
   image {
-    id
+    tag_name
+    url
+    ref
   }
   output_resource_name
   annotations
--- a/reconcile/utils/aws_api.py
+++ b/reconcile/utils/aws_api.py
@@ -23,7 +23,7 @@ if TYPE_CHECKING:
     from mypy_boto3_ec2.type_defs import (
         RouteTableTypeDef, SubnetTypeDef, TransitGatewayTypeDef,
         TransitGatewayVpcAttachmentTypeDef, VpcTypeDef, ImageTypeDef,
-        LaunchPermissionModificationsTypeDef, TagTypeDef
+        LaunchPermissionModificationsTypeDef, TagTypeDef, FilterTypeDef
     )
     from mypy_boto3_iam import IAMClient
     from mypy_boto3_iam.type_defs import AccessKeyMetadataTypeDef
@@ -31,7 +31,7 @@ else:
     EC2Client = EC2ServiceResource = RouteTableTypeDef = SubnetTypeDef = TransitGatewayTypeDef = \
         TransitGatewayVpcAttachmentTypeDef = VpcTypeDef = IAMClient = \
         AccessKeyMetadataTypeDef = ImageTypeDef = TagTypeDef = \
-        LaunchPermissionModificationsTypeDef = object
+        LaunchPermissionModificationsTypeDef = FilterTypeDef = object
 
 
 class InvalidResourceTypeError(Exception):
@@ -1281,3 +1281,26 @@ class AWSApi:  # pylint: disable=too-many-public-methods
                           f'unknown DNS zone {zone_id}')
         except Exception as e:
             logging.error(f'[{account_name}] unhandled exception: {e}')
+
+    def get_image_id(self, account_name: str, region_name: str,
+                     tag: Mapping[str, str]) -> str:
+        """
+        Get AMI ID matching the specified criteria.
+
+        :param account_name: the account name to operate on
+        :param region_name: aws region name to operate on
+        https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-images.html
+        """
+        ec2 = self._account_ec2_client(account_name, region_name)
+        filter_type_def: FilterTypeDef = {
+            'Name': 'tag:' + tag['Key'],
+            'Values': [tag['Value']]
+        }
+        images = ec2.describe_images(Filters=[filter_type_def])['Images']
+        if len(images) > 1:
+            raise ValueError(
+                f"found multiple AMI with tag {tag} " +
+                f"in account {account_name}")
+        elif not images:
+            return ''
+        return images[0]['ImageId']
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -12,12 +12,14 @@ import tempfile
 from threading import Lock
 
 from typing import (
-    Any, Dict, List, Iterable, Mapping, MutableMapping, Optional
+    Any, Dict, List, Iterable, Mapping, MutableMapping, Optional, Tuple
 )
 from ipaddress import ip_network, ip_address
 
 import anymarkup
 import requests
+from github import Github
+
 
 from terrascript import (Terrascript, provider, Provider, Terraform,
                          Backend, Output, data)
@@ -223,6 +225,8 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
                            for a in filtered_accounts}
         self.logtoes_zip = ''
         self.logtoes_zip_lock = Lock()
+        self.github = None
+        self.github_lock = Lock()
 
     def get_logtoes_zip(self, release_url):
         if not self.logtoes_zip:
@@ -250,6 +254,14 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
             # pylint: disable=consider-using-with
             open(zip_file, 'wb').write(r.content)
         return zip_file
+
+    def init_github(self) -> Github:
+        if not self.github:
+            with self.github_lock:
+                if not self.github:
+                    token = get_default_config()['token']
+                    self.github = Github(token, base_url=GH_BASE_URL)
+        return self.github
 
     def filter_disabled_accounts(self, accounts):
         filtered_accounts = []
@@ -959,7 +971,8 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
         elif provider == 'secrets-manager':
             self.populate_tf_resource_secrets_manager(resource, namespace_info)
         elif provider == 'asg':
-            self.populate_tf_resource_asg(resource, namespace_info)
+            self.populate_tf_resource_asg(resource, namespace_info,
+                                          existing_secrets)
         elif provider == 'route53-zone':
             self.populate_tf_resource_route53_zone(resource, namespace_info)
         else:
@@ -4073,17 +4086,68 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
 
         self.add_resources(account, tf_resources)
 
-    def populate_tf_resource_asg(self, resource, namespace_info):
+    def _get_commit_sha(self, repo_info: dict) -> str:
+        url = repo_info['url']
+        ref = repo_info['ref']
+        pattern = r'^[0-9a-f]{40}$'
+        # get commit_sha from ref
+        if re.match(pattern, ref):
+            return ref
+        # get commit_sha from branch
+        elif 'github' in url:
+            github = self.init_github()
+            repo_name = url.rstrip("/").replace('https://github.com/', '')
+            repo = github.get_repo(repo_name)
+            commit = repo.get_commit(sha=ref)
+            return commit.sha
+        elif 'gitlab' in url:
+            raise NotImplementedError("dose not support gitlab repo for now")
+
+        return ''
+
+    def _get_asg_image_id(self, image: dict,
+                          account: str, region: str) -> Tuple[str, str]:
+        """
+        AMI ID comes form AWS Api filter result.
+        AMI needs to be shared by integration aws-ami-share.
+        AMI needs to be taged with a tag_name and
+        its value need to be the commit sha comes from upstream repo.
+        """
+        commit_sha = self._get_commit_sha(image)
+        tag_name = image['tag_name']
+
+        # Get the most recent AMI id
+        aws_account = self.accounts[account]
+        aws = AWSApi(1, [aws_account],
+                     settings=self.settings,
+                     init_users=False)
+        commit_sha = 'test'
+        tag = {
+            'Key': tag_name,
+            'Value': commit_sha
+        }
+        image_id = aws.get_image_id(account, region, tag)
+        if not image_id:
+            raise ValueError(f"could not find ami with tag {tag}"
+                             f"in account {account}")
+
+        return image_id, commit_sha
+
+    def populate_tf_resource_asg(self, resource: dict,
+                                 namespace_info: dict,
+                                 existing_secrets: dict) -> None:
         account, identifier, common_values, \
             output_prefix, output_resource_name, annotations = \
             self.init_values(resource, namespace_info)
 
-        tf_resources = []
+        tf_resources: List[Any] = []
         self.init_common_outputs(tf_resources, namespace_info, output_prefix,
                                  output_resource_name, annotations)
 
         tags = common_values['tags']
         tags['Name'] = identifier
+        region = common_values.get('region') or \
+            self.default_regions.get(account)
 
         template_values = {
             "name": identifier,
@@ -4106,11 +4170,10 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
         }
 
         image = common_values.get('image')
-        image_id = image.get('id')
+        image_id, commit_sha = \
+            self._get_asg_image_id(image, account, region)
         template_values['image_id'] = image_id
 
-        region = common_values.get('region') or \
-            self.default_regions.get(account)
         if self._multiregion_account(account):
             template_values['provider'] = 'aws.' + region
 
@@ -4205,6 +4268,9 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
         tf_resources.append(Output(output_name_0_13, value=output_value))
         output_name_0_13 = output_prefix + '__image_id'
         output_value = image_id
+        tf_resources.append(Output(output_name_0_13, value=output_value))
+        output_name_0_13 = output_prefix + '__commit_sha'
+        output_value = commit_sha
         tf_resources.append(Output(output_name_0_13, value=output_value))
 
         self.add_resources(account, tf_resources)


maorfr:
since git is our source of truth, and we probably want to make this a little more similar to the saas file `ref` field, perhaps we can do something along the lines of:

somehow link the asg definition to a git repository and a ref (that can be either a branch name or a specific commit). the integration will query github/gitlab for the commit behind the ref (similar to what saasherder does). when a new commit is found, the integration will check if the AMI was already created (i liked this part of the implementation a lot, because it removes the race condition where a new commit is pushed but not yet bulit) and update the terraform configuration as you did in this PR. in case a new commit is found but no AMI is present yet, the integration should use the last used commit (store it in terraform state?).

think of this structure in an `asg` definition:
```
- provider: asg
  ...
  image:
    url: https://github.com/org/repo
    ref: master / main / commit
```

this keeps us truly gitops. i know it complicates things a lot.

wdyt?

maorfr:
starting to realize this is going to be extra fun for private repos.

maorfr:
and another thought - if there is a jenkins job building this, perhaps we are better off using a logic similar to the `upstream` approach of saas files.
this means we will not need to check AMIs in AWS at all?

BumbleFeng:
AMI may be owned by another account(e.g build by image-build-stage account and use it in image-build-prod), and when we share AMI across different accounts, we can only tell the difference of the AMI by its name. Tags can not share across accounts. 
So We have to get the AMI id from AWS by filtering its name. However, we still need the `upstream` job to store the last used AMI id, it seems very hard to get from terraform state.

reconcile/utils/aws_api.py
```diff
@@ -1209,3 +1212,30 @@ def upsert_route53_record(self, account_name, zone_id, recordset):
...
+                           by AWS.
+        :param sort_ascending: used to sort AMIs by creation time.
+        """
+        ec2 = self._account_ec2_client(account_name, region_name)
+        images = ec2.describe_images(
+            Owners=owners, Filters=filters)['Images']
+        if name_regex:
+            images = [i for i in images if re.match(name_regex, i['Name'])]
+        images = sorted(images,
+                        key=lambda i: datetime.strptime(i['CreationDate'],
```
Piojo, line 1238: The good news is that the date format you declare (RFC-3339) works perfectly with string comparisons. You can just do
```python
sorted(images, key=lambda i: i['CreationDate'])
```
(unless this is user-controlled input and you want an exception if they pass shit to us).

reconcile/utils/terrascript_client.py
```diff
@@ -3892,12 +3917,72 @@ def populate_tf_resource_asg(self, resource, namespace_info):
...
                     "resource_type": "volume",
                     "tags": tags
-                }
-            ]
+                }]
         }
 
-        region = common_values.get('region') or \
-            self.default_regions.get(account)
+        image = common_values.get('image')
```
Piojo, line 3923: OMG, this block is HUGE. Can you please split it into smaller chunks we can read in isolation?

reconcile/utils/terrascript_client.py
```diff
@@ -3892,12 +3917,72 @@ def populate_tf_resource_asg(self, resource, namespace_info):
...
+                                           settings=self.settings)
+                jenkins_instance = self.jenkins_map[instance_name]
+                job_name = upstream['name']
+                commit_sha = \
+                    jenkins_instance.get_last_success_build_commit(job_name)
+            # get commit_sha from branch
+            elif 'github' in url:
+                if not self.github:
+                    with self.github_lock:
+                        if not self.github:
```
Piojo, line 3960: Why are you rechecking for `self.github` 2 lines after checking? Can multiple threads really work on this same object?

reconcile/utils/terrascript_client.py
```diff
@@ -3892,12 +3917,72 @@ def populate_tf_resource_asg(self, resource, namespace_info):
...
+        image = common_values.get('image')
+        image_id = image.get('id')
+        if not image_id:
+            accounts = [a for a in self.filtered_accounts
+                        if a['name'] == account]
+            awsapi = AWSApi(1, accounts,
+                            settings=self.settings,
+                            init_users=False)
+            owner = image.get('owner')
+            if owner in self.uids.keys():
```
Piojo, line 3932: Shorter:
```python
if owner in self.uids:
```

reconcile/utils/terrascript_client.py
```diff
@@ -3892,12 +3917,72 @@ def populate_tf_resource_asg(self, resource, namespace_info):
...
+                owner = self.uids.get(owner)
```
Piojo, line 3933: And in that case, don't do `get` but a direct access. Even better:
```python
owner = image.get('owner')
if owner:
    owner = self.uids[owner]
```

reconcile/utils/terrascript_client.py
```diff
@@ -3892,12 +3917,72 @@ def populate_tf_resource_asg(self, resource, namespace_info):
...
+
+            url = image.get('url')
+            ref = image.get('ref')
```
Piojo, line 3936: What happens if there is no `url` or no `ref` in the image? Either handle the case for these calls to `get` returning `None` or, if cannot happen, just do normal dictionary accesses.

reconcile/utils/terrascript_client.py
```diff
@@ -3892,12 +3917,72 @@ def populate_tf_resource_asg(self, resource, namespace_info):
...
+            pattern = r'^[0-9a-f]{40}$'
+            upstream = image.get('upstream')
+            # get commit_sha from ref
+            if re.search(pattern, ref):
```
Piojo, line 3940: Nit: `re.match`.

reconcile/utils/terrascript_client.py
```diff
@@ -3892,12 +3917,72 @@ def populate_tf_resource_asg(self, resource, namespace_info):
...
+                if not self.gitlab:
+                    with self.gitlab_lock:
+                        if not self.gitlab:
+                            self.gitlab = self.init_gitlab()
+                project = self.gitlab.get_project(url)
+                commits = project.commits.list(ref_name=ref)
+                commit_sha = commits[0].id
+            # Get the most recent AMI named end with commit sha
+            name_regex = f".*{commit_sha}$"
+            image_ids = awsapi.get_image_ids(account, region, [owner],
```
Piojo, line 3976: According to the way you fetch the owner, this can result in a call to `get_image_ids(account, region, [None], ...)`. Is this a valid call? What do you expect in such a case?

reconcile/utils/terrascript_client.py
```diff
@@ -3892,12 +3917,72 @@ def populate_tf_resource_asg(self, resource, namespace_info):
...
-            self.default_regions.get(account)
+        image = common_values.get('image')
+        image_id = image.get('id')
+        if not image_id:
+            accounts = [a for a in self.filtered_accounts
+                        if a['name'] == account]
+            awsapi = AWSApi(1, accounts,
+                            settings=self.settings,
+                            init_users=False)
+            owner = image.get('owner')
```
Piojo, line 3931: What should happen if the image doesn't have an `owner` key?

reconcile/terraform_resources.py
```diff
@@ -394,32 +418,57 @@ def fetch_current_state(dry_run, namespaces, thread_pool_size,
...
+    ci_accounts = [a for a in accounts if a.get('defaultCI')]
+    if len(ci_accounts) == 0:
+        raise KeyError('default ci accounts not found')
+    if len(ci_accounts) > 1:
+        account_names = [a['name'] for a in ci_accounts]
+        raise KeyError(
+            'multiple default ci accounts found: '
+            f'{account_names}'
+        )
+    return ci_accounts[0]
```
maorfr, line 445: this can probably move to queries? maybe around here: https://github.com/app-sre/qontract-reconcile/blob/8a673b521dfedea89fc6a3be1cebccddd6d01879/reconcile/queries.py#L329-L338

reconcile/utils/terrascript_client.py
```diff
@@ -4085,7 +4130,71 @@ def populate_tf_resource_secrets_manager(self, resource, namespace_info):
...
+            instance = upstream['instance']
+            jenkins_instance = self.init_jenkins_instance(instance)
+            job_name = upstream['name']
+            if jenkins_instance.is_job_running(job_name):
+                try:
+                    existing_image = \
+                        existing_secrets[account][output_prefix]
+                except KeyError:
+                    raise ValueError("could not find existing image info")
+                return existing_image['image_id'], existing_image['commit_sha']
```
maorfr, line 4151: you can move this return statement inside the try block

reconcile/utils/terrascript_client.py
```diff
@@ -4085,7 +4130,71 @@ def populate_tf_resource_secrets_manager(self, resource, namespace_info):
...
+            github = self.init_github(authentication)
+            repo_name = url.rstrip("/").replace('https://github.com/', '')
+            repo = github.get_repo(repo_name)
+            commit = repo.get_commit(sha=ref)
+            commit_sha = commit.sha
+        elif 'gitlab' in url:
+            gitlab = self.init_gitlab()
+            project = gitlab.get_project(url)
+            commits = project.commits.list(ref_name=ref)
+            commit_sha = commits[0].id
```
maorfr, line 4172: https://github.com/app-sre/qontract-reconcile/blob/8a673b521dfedea89fc6a3be1cebccddd6d01879/reconcile/utils/saasherder.py#L464-L485

reconcile/utils/terrascript_client.py
```diff
@@ -4085,7 +4130,71 @@ def populate_tf_resource_secrets_manager(self, resource, namespace_info):
         for tf_resource in tf_resources:
             self.add_resource(account, tf_resource)
 
-    def populate_tf_resource_asg(self, resource, namespace_info):
+    def get_asg_image_id(self, image, existing_secrets, output_prefix,
```
maorfr, line 4133: i think you can simplify things a bit here...
if the "last known commit" (existing secrets) matches what is currently in the repo - we are done.
else, if there is a new commit, check if the image exists and shared, etc. if any of these conditions is not met - use the "last known commit".

reconcile/utils/terrascript_client.py
BumbleFeng: I copied from https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/terrascript_client.py#L228-L231

reconcile/utils/terrascript_client.py
BumbleFeng: they are required by schema https://github.com/app-sre/qontract-schemas/pull/53/files#diff-74d10b90e8f97c2e79cc05df43a72a2c59295a88871d5a1f6bbb7af4be7f98d4

reconcile/utils/terrascript_client.py
BumbleFeng: if owner is none, aws will return all ami (public and private)

reconcile/utils/terrascript_client.py
```diff
@@ -251,6 +262,14 @@ def download_logtoes_zip(self, release_url):
             open(zip_file, 'wb').write(r.content)
         return zip_file
 
+    def init_github(self):
```
Piojo, line 265: Type hint here, please.

reconcile/utils/terrascript_client.py
```diff
@@ -4073,7 +4093,63 @@ def populate_tf_resource_secrets_manager(self, resource, namespace_info):
...
+    def get_asg_image_id(self, image, existing_secrets, output_prefix,
+                         account, region):
+        image_id = image.get('id')
+        if image_id:
+            return image_id, ''
+
+        image_repo = image.get('repo')
+        tag_name = image_repo.get('tag_name')
+        url = image_repo.get('url')
+        ref = image_repo.get('ref')
```
Piojo, line 4105: As you mentioned that these fields are mandatory in the schema, remove `get` here:
```suggestion
        image_repo = image['repo']
        tag_name = image_repo['tag_name']
        url = image_repo['url']
        ref = image_repo['ref']
```

reconcile/utils/terrascript_client.py
```diff
@@ -4073,7 +4093,63 @@ def populate_tf_resource_secrets_manager(self, resource, namespace_info):
...
+                                         filters)
+        if not image_id:
+            logging.warning(f"could not find ami with filters {filters}"
+                            f"in account {ci_account_name}, "
+                            f"use existing asg's ami info")
+            try:
+                existing_image = \
+                    existing_secrets[account][output_prefix]
+            except KeyError:
+                raise ValueError("could not find existing asg's ami info")
```
Piojo, line 4137: Give more details here:
```suggestion
                raise ValueError(f"could not find existing asg's ami info for {account} with prefix {output_prefix}")
```
or something liek that

reconcile/utils/terrascript_client.py
```diff
@@ -4073,7 +4093,63 @@ def populate_tf_resource_secrets_manager(self, resource, namespace_info):
...
+
+        # Check if the ami has been shared with target account
+        filters = [{
+            'Name': 'image-id',
+            'Values': image_id
+        }]
+        share_id = self.aws.get_image_id(account, region, owner, filters)
+        if not share_id:
+            raise ValueError(f'ami {image_id} has not been shared'
+                             f'with account {account}')
```
Piojo, line 4148: This is an excellent exception.

Piojo (commented):
Looks solid to me. I'll let you handle @steahan's remaining comments before approving.

reconcile/utils/terrascript_client.py
```diff
@@ -4073,7 +4093,63 @@ def populate_tf_resource_secrets_manager(self, resource, namespace_info):
...
+        return image_id[0], commit_sha
+
+    def populate_tf_resource_asg(self, resource, namespace_info,
```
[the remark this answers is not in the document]
BumbleFeng: Do you know how to annotate the type of `self`?

reconcile/utils/terrascript_client.py
BumbleFeng: Do you know how to annotate the type of `self`?

reconcile/utils/terrascript_client.py
```diff
@@ -4073,17 +4093,86 @@ def populate_tf_resource_secrets_manager(self, resource, namespace_info):
...
+            commit_sha = ref
+        # get commit_sha from branch
+        elif 'github' in url:
+            github = self.init_github()
+            repo_name = url.rstrip("/").replace('https://github.com/', '')
+            repo = github.get_repo(repo_name)
+            commit = repo.get_commit(sha=ref)
+            commit_sha = commit.sha
+        elif 'gitlab' in url:
+            raise ValueError("dose not support gitlab repo for now")
```
maorfr, line 4127: can this be extracted to a function?
just thinking of the **future** refactor of this and saasherder (which has nearly the exact function _get_commit_sha).

reconcile/utils/terrascript_client.py
```diff
@@ -155,12 +157,19 @@ class ElasticSearchLogGroupInfo:
 
 class TerrascriptClient:
     def __init__(self, integration, integration_prefix,
-                 thread_pool_size, accounts, settings=None):
+                 thread_pool_size, accounts, settings=None, init_aws=False):
```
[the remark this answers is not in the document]
BumbleFeng: Actually, I also need to init AWSApi for asg's target account. While the `account` comes from `self.init_values(resource, namespace_info)` only have a name, I need to get complete accounts info from init TerrascriptClient.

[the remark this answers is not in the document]
BumbleFeng: never mind. Found a way doing that.

reconcile/utils/terrascript_client.py
```diff
@@ -4073,17 +4093,98 @@ def populate_tf_resource_secrets_manager(self, resource, namespace_info):
...
+                     settings=self.settings,
+                     init_users=False)
+        owner = [ci_account['uid']]
+        filters: Sequence[FilterTypeDef] = [{
+            'Name': 'tag:' + tag_name,
+            'Values': [commit_sha]
+        }]
+        image_id = aws.get_image_id(ci_account_name, region, owner,
+                                    filters)
+        if not image_id:
```
maorfr, line 4150: i saw that you imported queries to this file. we generally don't want to involve our graphql schema with any of our utils. this is why queries lives in reconcile and not in reconcile.utils.

that got me thinking - why do we even need queries? to get the ci account. why do we even need the ci account? to check if the image exists there. why do we even care if it exists there if it's not shared with the account we are in the context of?

we probably don't.
this means you can remove this whole part and do something like:
Check if the ami has been shared with target account. if not - log a warning and try to return the commit sha from existing_secrets.

this may simplify things significantly?

maorfr: assuming we can get `image_id` without looking into the ci account...

maorfr: we've discussed it this morning. the `image_id` is obtained from the source account by tag. these tags are not available in the shared account. so we decided to automate the sharing and copying of tags!
see #2216

reconcile/utils/terrascript_client.py
Piojo: I don't need it for `self` but at leat the return value: `init_github(self) -> None:` - this way we get mypy to check its callees.

reconcile/utils/terrascript_client.py
```diff
@@ -4073,17 +4091,84 @@ def populate_tf_resource_secrets_manager(self, resource, namespace_info):
...
+            return ref
+        # get commit_sha from branch
+        elif 'github' in url:
+            github = self.init_github()
+            repo_name = url.rstrip("/").replace('https://github.com/', '')
+            repo = github.get_repo(repo_name)
+            commit = repo.get_commit(sha=ref)
+            return commit.sha
+        elif 'gitlab' in url:
+            raise ValueError("dose not support gitlab repo for now")
```
maorfr, line 4109: NotImplementedError?

reconcile/utils/aws_api.py
```diff
@@ -1215,3 +1215,24 @@ def upsert_route53_record(self, account_name, zone_id, recordset):
...
+        https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-images.html
+        """
+        ec2 = self._account_ec2_client(account_name, region_name)
+        images = ec2.describe_images(
+            Owners=owners, Filters=filters)['Images']
+        if len(images) > 1:
+            raise ValueError(
+                f"found multiple AMI with filters {filters} " +
+                f"owned by {owners} in account {account_name}")
+        return [i['ImageId'] for i in images]
```
maorfr, line 1238: should we return index 0 instead of a list or None if image is not found?

reconcile/utils/terrascript_client.py
```diff
@@ -4073,17 +4091,84 @@ def populate_tf_resource_secrets_manager(self, resource, namespace_info):
...
+        its value need to be the commit sha comes from upstream repo.
+        In case of there is a new commit while the packer job is still running,
+        use the AMI form existing asg.
+        """
+        image_id = image.get('id')
+        if image_id:
+            return image_id, ''
+
+        image_repo = image['repo']
+        commit_sha = self._get_commit_sha(image_repo)
```
maorfr, line 4129: if image_repo['ref'] == commit_sha:
    # this is a specific commit and not main/master or any other branch

maorfr (approved):

PR #2101, opened 2022-01-18
If a RS is resized but there are pods with ownerrefs where the owner …

rrati:
…object does not exist, then the openshift-resources will error out.  The code desires to find the pods that are controlled by the RS being resizedso they can be restarted, so having the controller not exist shouldn't impact how things work and allow handling the above issue.

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -515,7 +515,7 @@ class OCDeprecated:
         pods = self.get(namespace, 'Pod')['items']
         owned_pods = []
         for p in pods:
-            owner = self.get_obj_root_owner(namespace, p)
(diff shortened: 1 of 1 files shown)


jfchevrette (approved):

maorfr:
this was supposed to work after #1725 :(

maorfr:
i'm not sure i understand what this PR is supposed to achieve. the `get_owned_pods` method is only called when we need to rezize a PVC. how is that related to RS (which we are not explicitly managing)?

perhaps i am missing something.

maorfr:
https://issues.redhat.com/browse/ASIC-163

maorfr (approved):
LGTM

PR #2102, opened 2022-01-20
terrascript client make get partition succeed

maorfr:
even if the account has integrations disabled (avoid KeyError)


BumbleFeng:
lgtm

PR #2104, opened 2022-01-20
move blackbox-exporter integration to external cluster

geoberle:
the blackbox-exporter-endpoint-monitoring integration was added to the internal app-interface environment, but it betters fits the external one.

this patch moves the integration from internal to external and also introduces the usual `--internal` and `--use-jump-host` flags to correctly handle communication with clusters


reconcile/blackbox_exporter_endpoint_monitoring.py
```diff
@@ -166,19 +167,24 @@ def run(dry_run: bool, thread_pool_size: int, defer=None) -> None:
...
+        defer(oc_map.cleanup)
+
+        # reconcile
+        for provider, endpoints in desired_endpoints.items():
+            fill_desired_state(provider, endpoints, ri)
+        ob.realize_data(dry_run, oc_map, ri, thread_pool_size,
+                        recycle_pods=False)
+
+        if ri.has_error_registered():
+            sys.exit(1)
```
fishi0x01, line 190: Is a registered error also logged somewhere before exit?

fishi0x01: Ah ok [looks like](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/openshift_resource.py#L484-L492) a registered error is anyways just a boolean flag - not an error object. So anyways not a concern at this stage :)

rporres (approved):

reconcile/blackbox_exporter_endpoint_monitoring.py
geoberle: the errors related to `ResourceInventory a.k.a. ri` are usually logged where they occur, e.g. [here](https://github.com/app-sre/qontract-reconcile/blob/407eda78e4b326396a316b969bd7af149ed7a776/reconcile/openshift_base.py#L469-L474)

PR #2105, opened 2022-01-21
Fix bug in elasticache output variable selection

TGPSKI:
This MR fixes a bug in terrascript_client.py related to the creation of Elasticache Resources with terraform.

Currently, EC clusters are created properly in all cases. However, in the case of non-HA cluster creation, the output attributes for the db_endpoint Vault variable are not properly selected.

In cases of both HA and non-HA EC resources, the key "cluster_mode" is included in the terraform attribute outputs, therefore a check for `values.get('cluster_mode')` will always return true.

The logic has been changed to check for `cluster_enabled`. If `cluster_enabled` is True, then use the primary endpoint address.

My only question is how is terrascript processing booleans. Can I do a boolean comparison on output attribute values? Or do I have to do a string comparison. @maorfr 

Terraform state output for non-HA EC resource:
```
"cluster_enabled": false,
"cluster_mode": [
    {
    "num_node_groups": 1,
    "replicas_per_node_group": 1
    }
],
"configuration_endpoint_address": null,
"engine": "redis",
"engine_version": "6.0",
"final_snapshot_identifier": null,
"id": "test",
"kms_key_id": "",
"maintenance_window": "wed:10:00-wed:11:00",
"member_clusters": [
    "test"
],
"multi_az_enabled": false,
"node_type": "cache.t2.small",
"notification_topic_arn": null,
"number_cache_clusters": 2,
"parameter_group_name": "test",
"port": 6379,
"primary_endpoint_address": "foo.use1.cache.amazonaws.com",
"reader_endpoint_address": "replica.foo.use1.cache.amazonaws.com",
```


steveteahan:
I believe that the logic here is correct. I know that this particular method is non-trivial to unit test. I was wondering if we might consider running the integration in dry-run and determining what diff would be produced in the JSON output by this change? That should help to improve confidence about the affect of the change. The outputs could be included in Jira. I think this might also answer the open question about boolean behavior?

TGPSKI:
My first attempt won't work based on some digging into the parameters available in this method. I will revise.

reconcile/utils/terrascript_client.py
```diff
@@ -1742,7 +1742,7 @@ def populate_tf_resource_elasticache(self, resource, namespace_info,
         output_name_0_13 = output_prefix + '__db_endpoint'
         # https://docs.aws.amazon.com/AmazonElastiCache/
         # latest/red-ug/Endpoints.html
-        if values.get('cluster_mode'):
```
maorfr, line 1745: the `cluster_mode` field is coming from the defaults file defined for the resources in app-interface fwiw

maorfr:
commented on https://issues.redhat.com/browse/APPSRE-4300 according to these docs: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticache_replication_group

the decision on what endpoint to use should be based on the `cluster-enabled` parameter of the parameter group and not based on the values supplied in the defaults file.

PR #2108, opened 2022-01-23
[ocm-aws-infrastructure-access] do not delete deleting role grants

maorfr:
related to:
- https://issues.redhat.com/browse/SDA-4422
- https://issues.redhat.com/browse/OHSS-9286

some role grants are "stuck" in Deleting. if we try to delete those, we get a 5xx error from OCM.
this PR updates the behavior to avoid deleting role grants in the "deleting" state.

the implementation is to remove any "Deleting" role grants from the current state, preventing them from being attempted to be deleted. because we remove these role grants from the current state, they will be attempted to be added because they may be in the desired state. we prevent that by removing any role grants in "Deleting" from the role grants to be added.

### DIFF
--- a/reconcile/ocm_aws_infrastructure_access.py
+++ b/reconcile/ocm_aws_infrastructure_access.py
@@ -3,7 +3,7 @@ import logging
 from reconcile.utils import gql
 from reconcile import queries
 
-from reconcile.utils.ocm import OCMMap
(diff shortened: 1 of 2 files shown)


maorfr:
this change is quite straight forward. even though the code may not be perfect, i'm going to merge and promote it to stop the integration from failing. happy to get post merge reviews!

PR #2109, opened 2022-01-24
Added AWS ES advanced security options support

rporres:
part of [APPSRE-4219](https://issues.redhat.com/browse/APPSRE-4219)

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticsearch_domain#advanced_security_options

schema changes: https://github.com/app-sre/qontract-schemas/pull/56

### DIFF
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -9,7 +9,7 @@ import tempfile
 
 from threading import Lock
 
-from typing import Dict, List, Iterable, Optional
+from typing import Any, Dict, List, Iterable, MutableMapping, Optional
 from ipaddress import ip_network, ip_address
 
 import anymarkup
@@ -3299,6 +3299,12 @@ class TerrascriptClient:
         if self._multiregion_account_(account):
             es_values['provider'] = 'aws.' + region
 
+        advanced_security_options = values.get('advanced_security_options', {})
+        if advanced_security_options:
+            es_values['advanced_security_options'] = \
+                self._build_es_advanced_security_options(
+                    advanced_security_options)
+
         es_tf_resource = aws_elasticsearch_domain(identifier, **es_values)
         tf_resources.append(es_tf_resource)
 
@@ -3333,6 +3339,26 @@ class TerrascriptClient:
 
         for tf_resource in tf_resources:
             self.add_resource(account, tf_resource)
+
+    def _build_es_advanced_security_options(
+            self, advanced_security_options: MutableMapping[str, Any]) \
+            -> MutableMapping[str, Any]:
+        master_user_options = advanced_security_options.pop(
+            'master_user_options', {})
+
+        if master_user_options:
+            master_user_secret = master_user_options['master_user_secret']
+            secret_data = self.secret_reader.read_all(master_user_secret)
+
+            required_keys = {'master_user_name', 'master_user_password'}
+            if secret_data.keys() != required_keys:
+                raise KeyError(
+                    f"vault secret '{master_user_secret['path']}' must "
+                    f"exactly contain these keys: {', '.join(required_keys)}")
+
+            advanced_security_options['master_user_options'] = secret_data
+
+        return advanced_security_options
 
     def populate_tf_resource_acm(self, resource, namespace_info):
         account, identifier, common_values, \


Piojo (commented):
We don't have unit tests for terrascript client, so can you give some details of how this was tested?

reconcile/utils/terrascript_client.py
```diff
@@ -3299,6 +3299,32 @@ def populate_tf_resource_elasticsearch(self, resource, namespace_info):
         if self._multiregion_account_(account):
             es_values['provider'] = 'aws.' + region
 
+        advanced_security_options = values.get('advanced_security_options', {})
+        if advanced_security_options:
```
fishi0x01, line 3303: Not mandatory, just small suggestion: The function is quite humongous - would it make sense to migrate this section to a new function, like:

```
if advanced_security_options:
    es_values['advanced_security_options'] = extract_advanced_security_options(advanced_security_options)
    
...

def extract_advanced_security_options(Dict options) -> Dict:  # Not sure if the types could be narrowed down a bit more here maybe
    pass
```

reconcile/utils/terrascript_client.py
```diff
@@ -3299,6 +3299,32 @@ def populate_tf_resource_elasticsearch(self, resource, namespace_info):
...
+                master_user_secret = master_user_options['master_user_secret']
+                secret_data = self.secret_reader.read_all(master_user_secret)
+
+                required_keys = ['master_user_name', 'master_user_password']
+                found = 0
+                for key in required_keys:
+                    if key in secret_data:
+                        found += 1
+
+                if len(required_keys) != found:
```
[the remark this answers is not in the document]
rporres: sure, less variables... :)

rporres:
> We don't have unit tests for terrascript client, so can you give some details of how this was tested?

Using the `--print-to-file` option with different local configurations, inspecting the resulting json, making sure a `terraform plan` works as expected, etc...

reconcile/utils/terrascript_client.py
rporres: it is humongouns, indeed... I'll follow your advise

rporres: Fixed in 6c9a83d

reconcile/utils/terrascript_client.py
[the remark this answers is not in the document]
rporres: hmmm, your proposal not working as I expected. This should not succeed:
```
>>> secret_data = {'master_user_name': "u", 'master_user_password': "p"}
>>> required_keys = {'master_user_name', 'master_user_password'}
>>> not required_keys < secret_data.keys()
True
```
But mine is not working too. It would appear that it is:
```
>>> secret_data.keys() < required_keys
False
```
But it actually fails as this should return True
```
>>> secret_data = {'u': "u", 'p': "p"}
>>> secret_data.keys() < required_keys
False
```
I guess it is actually comparing lengths.

So if we want to continue with the `required_keys` set approach, we could use intersections
```
>>> secret_data = {'master_user_name': "u", 'master_user_password': "p"}
>>> required_keys = {'master_user_name', 'master_user_password'}
>>> secret_data.keys() & required_keys != required_keys
False
>>> secret_data = {'u': "u", 'p': "p"}
>>> secret_data.keys() & required_keys != required_keys
True
>>> secret_data = {'master_user_name': "u"}
>>> secret_data.keys() & required_keys != required_keys
True
```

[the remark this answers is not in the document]
rporres: Fixed in 9e087c6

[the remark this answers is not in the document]
rporres: It was easier, for sure. A simple comparison would have sufficed:
```
>>> secret_data = {'master_user_name': "u", 'master_user_password': "p", "this": "should not be here"}
>>> required_keys = {'master_user_name', 'master_user_password'}
>>> secret_data.keys() != required_keys
True

>>> secret_data = {'master_user_name': "u", 'master_user_password': "p"}
>>> required_keys = {'master_user_name', 'master_user_password'}
>>> secret_data.keys() != required_keys
False
```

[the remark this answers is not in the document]
rporres: Fixed in 2e7228b

reconcile/utils/terrascript_client.py
```diff
@@ -3334,6 +3340,25 @@ def populate_tf_resource_elasticsearch(self, resource, namespace_info):
...
+            self, advanced_security_options: dict[str, Any]) -> dict[str, Any]:
+        master_user_options = advanced_security_options.pop(
+            'master_user_options', {})
+
+        if master_user_options:
+            master_user_secret = master_user_options['master_user_secret']
+            secret_data = self.secret_reader.read_all(master_user_secret)
+
+            required_keys = {'master_user_name', 'master_user_password'}
+            if secret_data.keys() != required_keys:
```
Piojo, line 3353: This is even more intuitive than what I had proposed at the beginning.

reconcile/utils/terrascript_client.py
```diff
@@ -3334,6 +3340,25 @@ def populate_tf_resource_elasticsearch(self, resource, namespace_info):
...
+    def _build_es_advanced_security_options(
+            self, advanced_security_options: dict[str, Any]) -> dict[str, Any]:
+        master_user_options = advanced_security_options.pop(
+            'master_user_options', {})
+
+        if master_user_options:
+            master_user_secret = master_user_options['master_user_secret']
+            secret_data = self.secret_reader.read_all(master_user_secret)
+
+            required_keys = {'master_user_name', 'master_user_password'}
```
Piojo, line 3352: You may want to make this a constant on a higher level. No biggie.

Piojo (approved):
LGTM. Thanks for refactoring.

reconcile/utils/terrascript_client.py
```diff
@@ -3334,6 +3340,25 @@ def populate_tf_resource_elasticsearch(self, resource, namespace_info):
         for tf_resource in tf_resources:
             self.add_resource(account, tf_resource)
 
+    def _build_es_advanced_security_options(
+            self, advanced_security_options: dict[str, Any]) -> dict[str, Any]:
```
fishi0x01, line 3344: According to [dev-guidelines](https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#abstract-collection-types), my understanding here would be:

```
def _build_es_advanced_security_options(
            self, advanced_security_options: Mapping[str, Any]) -> Mapping[str, Any]:
```

rporres: It actually needs to be a mutable mapping but you're right. The method signature helps understand better the code now

Fixed in 3186d1e

reconcile/utils/terrascript_client.py
rporres: cool!

reconcile/utils/terrascript_client.py
rporres: Not really, this is unlikely to be reused anywhere else so I think it makes things clearer to leave it here.

PR #2111, opened 2022-01-24
Add cronjobs to run e2e-tests; Add dry_run mode to e2e_tests.

OliviaHY:
### Why
Moving cronjobs from Jenkins to Openshift. Ref [APPSRE-3615](https://issues.redhat.com/browse/APPSRE-3615)
### What
* Add dry_run mode to e2e_tests scripts.
* Add cronjobs to run e2e_tests.
* Minor tweak to helm template to allow `e2e_tesets` prefix

### DIFF
--- a/e2e_tests/default_network_policies.py
+++ b/e2e_tests/default_network_policies.py
@@ -17,7 +17,6 @@ def test_cluster(cluster, oc_map, pattern):
         logging.log(level=oc.log_level, msg=oc.message)
         return None
     logging.info("[{}] validating default NetworkPolicies".format(cluster))
-
     projects = [p['metadata']['name']
                 for p in oc.get_all('Project')['items']
                 if p['status']['phase'] != 'Terminating' and
--- a/e2e_tests/cli.py
+++ b/e2e_tests/cli.py
@@ -33,12 +33,16 @@ def run_test(func, *args):
                   'WARNING',
                   'ERROR',
                   'CRITICAL']))
+@click.option('--dry-run/--no-dry-run', 'dry_run', default=False,
+              help='Only print the planned actions if `true`')
 @click.pass_context
-def test(ctx, configfile, log_level):
+def test(ctx, configfile, log_level, dry_run):
     ctx.ensure_object(dict)
 
     level = getattr(logging, log_level) if log_level else logging.INFO
     logging.basicConfig(format='%(levelname)s: %(message)s', level=level)
+
+    ctx.obj['dry_run'] = dry_run
 
     config.init_from_toml(configfile)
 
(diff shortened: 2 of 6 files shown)


e2e_tests/dedicated_admin_rolebindings.py
```diff
@@ -32,10 +32,11 @@ def test_cluster(cluster, oc_map, pattern):
...
-def run(thread_pool_size=10, defer=None):
+def run(thread_pool_size=10, dry_run=False, defer=None):
     oc_map = tb.get_oc_map(QONTRACT_E2E_TEST)
     defer(oc_map.cleanup)
     pattern = tb.get_namespaces_pattern()
     threaded.run(test_cluster, oc_map.clusters(), thread_pool_size,
                  oc_map=oc_map,
-                 pattern=pattern)
+                 pattern=pattern,
+                 dry_run=dry_run)
```
rporres, line 42: I don't think you need to pass this option to `test_cluster` if there's nothing being done there with it

e2e_tests/default_network_policies.py
```diff
@@ -35,10 +34,11 @@ def test_cluster(cluster, oc_map, pattern):
...
-def run(thread_pool_size=10, defer=None):
+def run(thread_pool_size=10, dry_run=False, defer=None):
     oc_map = tb.get_oc_map(QONTRACT_E2E_TEST)
     defer(oc_map.cleanup)
     pattern = tb.get_namespaces_pattern()
     threaded.run(test_cluster, oc_map.clusters(), thread_pool_size,
                  oc_map=oc_map,
-                 pattern=pattern)
+                 pattern=pattern,
+                 dry_run=dry_run)
```
rporres, line 44: I don't think you need to pass this option to `test_cluster` if there's nothing being done there with it

e2e_tests/default_project_labels.py
```diff
@@ -36,10 +36,11 @@ def test_cluster(cluster, oc_map, pattern):
...
-def run(thread_pool_size=10, defer=None):
+def run(thread_pool_size=10, dry_run=False, defer=None):
     oc_map = tb.get_oc_map(QONTRACT_E2E_TEST)
     defer(oc_map.cleanup)
     pattern = tb.get_namespaces_pattern()
     threaded.run(test_cluster, oc_map.clusters(), thread_pool_size,
                  oc_map=oc_map,
-                 pattern=pattern)
+                 pattern=pattern,
+                 dry_run=dry_run)
```
rporres, line 46: I don't think you need to pass this option to `test_cluster` if there's nothing being done there with it

rporres (commented):
So there's a missing aspect here. The app-interface script running this test evaluates the results at the end of the run and exits if there's any error. There's an alert rule that checks if e2e tests are failing: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/resources/observability/prometheusrules/ci-int.prometheusrules.yaml

So an additional aspect to consider is the alert creation. I can imagine two ways:

* Use the Prometheus Push Gateway
* Use the `kube_job_*` and `kube_cronjob_*` metrics very cleverly

The first one looks easier to get data from in Prometheus but it requires additional code here that we will need to maintain. The second one looks that it is going to give you a few PromQL headaches but it's worth exploring as you don't need to do anything else code-wise...

OliviaHY:
Close this for now until the alert situation is sorted.

PR #2115, opened 2022-01-25
Update SlackApi

janboll:
* Minor refactoring: move calls to _get into seperate methods for easier
  readability
* Refactor slack api creation, remove init_slack functions, create class
  functions for instance creation

Refactoring improvements where discovered while working on: https://issues.redhat.com/browse/APPSRE-3988

### DIFF
--- a/reconcile/openshift_upgrade_watcher.py
+++ b/reconcile/openshift_upgrade_watcher.py
@@ -3,8 +3,8 @@ import logging
 from datetime import datetime
 
 from reconcile import queries
+from reconcile.slack_base import slackapi_from_queries
 
-from reconcile.slack_base import init_slack_workspace
 from reconcile.utils.oc import OC_Map
 from reconcile.utils.state import State
 from reconcile.utils.defer import defer
@@ -31,7 +31,7 @@ def run(dry_run, thread_pool_size=10, internal=None, use_jump_host=True,
     )
 
     if not dry_run:
-        slack = init_slack_workspace(QONTRACT_INTEGRATION)
+        slack = slackapi_from_queries(QONTRACT_INTEGRATION)
 
     now = datetime.utcnow()
     for cluster in oc_map.clusters(include_errors=True):
--- a/reconcile/utils/slack_api.py
+++ b/reconcile/utils/slack_api.py
@@ -1,14 +1,15 @@
 import json
 import logging
-from typing import Sequence, Dict, Any, Mapping, Optional, Union
+from typing import Sequence, Dict, Any, Mapping, Optional, Union, ItemsView, \
+    Iterable
 
 from slack_sdk import WebClient
 from slack_sdk.errors import SlackApiError
 from slack_sdk.http_retry import RateLimitErrorRetryHandler, RetryHandler, \
     RetryState, HttpRequest, HttpResponse
 
-from reconcile.utils.secret_reader import SecretReader
 from reconcile.utils.config import get_config
+from reconcile.utils.secret_reader import SecretReader
 
 MAX_RETRIES = 5
 TIMEOUT = 30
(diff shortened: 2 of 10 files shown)


reconcile/utils/slack_api.py
```diff
@@ -285,22 +292,28 @@ def get_user_id_by_name(self, user_name: str) -> str:
 
         return result['user']['id']
 
-    def get_channels_by_names(self, channels_names):
-        return {k: v['name'] for k, v in self._get('channels').items()
+    def get_channels_by_names(self, channels_names: str):
```
Piojo, line 295: Type hint the return value, please: `-> dict:`

reconcile/utils/slack_api.py
```diff
@@ -285,22 +292,28 @@ def get_user_id_by_name(self, user_name: str) -> str:
...
+        return {k: v['name'] for k, v in self.get_channels()
                 if v['name'] in channels_names}
 
-    def get_channels_by_ids(self, channels_ids):
-        return {k: v['name'] for k, v in self._get('channels').items()
+    def get_channels_by_ids(self, channels_ids: str):
```
Piojo, line 299: Is this a string? I'd expect some `Iterable[str]` given the name of the channel and the arguments. Also, type hint for the return value.

reconcile/utils/slack_api.py
```diff
@@ -342,3 +355,48 @@ def _get(self, resource: str) -> Dict[str, Any]:
 
         self._results[resource] = results
         return results
+
+    @classmethod
+    def create_using_queries(cls, integration_name: str,
```
Piojo, line 360: Wouldn't `create_from_queries` be more consistent with `create_from_dict`?

reconcile/utils/slack_api.py
```diff
@@ -285,22 +292,28 @@ def get_user_id_by_name(self, user_name: str) -> str:
...
-    def get_users_by_ids(self, users_ids):
-        return {k: v['name'] for k, v in self._get('users').items()
+    def get_users_by_ids(self, users_ids: str):
+        return {k: v['name'] for k, v in self.get_users()
                 if k in users_ids}
 
+    def get_channels(self):
+        return self._get(self.RESOURCE_GET_CHANNELS).items()
+
+    def get_users(self):
```
Piojo, line 314: These return `dict_items`.

maorfr:
first round from me:
the new methods introduced to SlackApi to initiate from dict or from a query are creating a hard dependency between our own graphql schema and the slack logic.
in my view, the api/clients under `utils` should be as decoupled as possible from our schema.
i'd recommend to move this logic to slack_base perhaps?

i'm sure we are not following this 100% right now, but still worth the thought.

wdyt?

janboll:
Thanks for the feedback, I'll move them back to  `slack_base.py`. I didn't saw the need for this file it, but I get it now with your  point of not querying server from the `utils` package.

janboll:
> While I understand the intent of saving code and repetition, I find this new API confusing: some methods return dictionaries, some others return items to be iterated with.

Thanks for the feedback. There's no use case for the two new methods outside the class, since that I hid them.

reconcile/utils/slack_api.py
janboll: --> moved to slack_base.py and renamed

reconcile/openshift_upgrade_watcher.py
```diff
@@ -31,7 +31,7 @@ def run(dry_run, thread_pool_size=10, internal=None, use_jump_host=True,
     )
 
     if not dry_run:
-        slack = init_slack_workspace(QONTRACT_INTEGRATION)
+        slack = SlackApi.create_using_queries(QONTRACT_INTEGRATION)
```
maorfr, line 34: still needs changing?

tools/qontract_cli.py
```diff
@@ -857,7 +857,7 @@ def slack_usergroup(ctx, workspace, usergroup, username):
     Use an org_username as the username.
     To empty a slack usergroup, pass '' (empty string) as the username.
     """
-    slack = init_slack_workspace('qontract-cli')
+    slack = SlackApi.create_using_queries('qontract-cli')
```
maorfr, line 860: still needs changing?

reconcile/utils/slack_api.py
```diff
@@ -252,7 +259,7 @@ def update_usergroup_users(self, id: str,
                 raise
 
     def get_random_deleted_user(self):
-        for user_id, user_data in self._get('users').items():
+        for user_id, user_data in self.get_users():
```
maorfr, line 262: i don't see the `get_users` method... am i missing it?

janboll: Let me add a test for this...

PR #2117, opened 2022-01-26
Container integration loop for local development

fishi0x01:
**Motivation**

The reconcile loop can be triggered in local dev env, however, it requires proper binaries to be in `PATH`. I see 2 ways to automate this issue away:

1. Run reconcile dev loop inside a container
2. Add make target which downloads binaries to repo local `./bin` directory and add an .env script to source those bins as prefix to PATH

This PR targets the first approach.

**Proposed Changes**

- added a make target to run reconcile loop inside a docker container
- added documentation about that new make target
- set 3.9.x as compatibility target (same as in `setup.py`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build push rc build-test test-app test-container-image test clean
+.PHONY: help build push rc build-test test-app test-container-image test clean
 
 CONTAINER_ENGINE ?= $(shell which podman >/dev/null 2>&1 && echo podman || echo docker)
 IMAGE_TEST := reconcile-test
@@ -13,6 +13,10 @@ else
 endif
 
 CTR_STRUCTURE_IMG := quay.io/app-sre/container-structure-test:latest
+
+help: ## Prints help for targets with comments
+	@grep -E '^[a-zA-Z0-9.\ _-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
 build:
 	@DOCKER_BUILDKIT=1 $(CONTAINER_ENGINE) build -t $(IMAGE_NAME):latest -f dockerfiles/Dockerfile . --progress=plain
 	@$(CONTAINER_ENGINE) tag $(IMAGE_NAME):latest $(IMAGE_NAME):$(IMAGE_TAG)
@@ -34,12 +38,10 @@ generate:
 build-test:
 	@$(CONTAINER_ENGINE) build -t $(IMAGE_TEST) -f dockerfiles/Dockerfile.test .
 
-test-app: build-test
-#	Target to test app with tox on docker
+test-app: build-test ## Target to test app with tox on docker
 	@$(CONTAINER_ENGINE) run --rm $(IMAGE_TEST)
 
-test-container-image: build
-#	Target to test the final image
+test-container-image: build ## Target to test the final image
 	@$(CONTAINER_ENGINE) run --rm \
 		-v /var/run/docker.sock:/var/run/docker.sock \
 		-v $(CURDIR):/work \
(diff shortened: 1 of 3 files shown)


dockerfiles/hack/run-integration.py
```diff
@@ -55,8 +55,8 @@ def build_entry_point_args(command: click.Command, config: str,
                            dry_run: Optional[str], integration_name: str,
                            extra_args: Optional[str]) -> list[str]:
     args = ['--config', config]
-    if dry_run is not None:
-        args.append(dry_run)
+    if dry_run != '':
+        args.append('--dry-run')
```
maorfr, line 59: can `dry_run` also be `--no-dry-run`?

steveteahan:
I mentioned this in Slack as well, but this is something that will simplify #1870 because we'd no longer need to worry about handling multiple platforms if team members were running integrations in containers. It was something I was meaning to get to, but thank you for doing this!

dockerfiles/hack/run-integration.py
fishi0x01: My understanding is that `--no-dry-run` is the default behavior. In case I do not want to have a dry-run, then I dont set the `DRY_RUN` variable. I.e., I do not really see a use-case now for `--no-dry-run`. (I might be missing sth here though)

On the other hand one could argue that it is not a bad practice to allow very explicit flags - just to be sure things run as expected if the default changes one day silently. So I guess I will follow your advice here :+1:

fishi0x01: Fixed

Makefile
```diff
@@ -34,12 +38,10 @@ generate:
...
 	@$(CONTAINER_ENGINE) build -t $(IMAGE_TEST) -f dockerfiles/Dockerfile.test .
 
-test-app: build-test
-#	Target to test app with tox on docker
+test-app: build-test ## Target to test app with tox on docker
 	@$(CONTAINER_ENGINE) run --rm $(IMAGE_TEST)
 
-test-container-image: build
-#	Target to test the final image
+test-container-image: build ## Target to test the final image
```
Piojo, line 44: Leftover from a previous PR? Please rebase. :)

dockerfiles/hack/run-integration.py
```diff
@@ -55,7 +55,7 @@ def build_entry_point_args(command: click.Command, config: str,
                            dry_run: Optional[str], integration_name: str,
                            extra_args: Optional[str]) -> list[str]:
     args = ['--config', config]
-    if dry_run is not None:
+    if dry_run in ['--dry-run', '--no-dry-run']:
```
Piojo, line 58: Does this change belong in this PR?

README.md
```diff
@@ -208,7 +208,33 @@ tox -e type
 tox -e type -- reconcile/utils/slack_api.py
 ```
 
+## Run reconcile loop for an integration locally in a container
 
+ This is currently only tested with the docker container engine.
+
+### Prepare config.toml
+
+Make sure the file  [`./config.dev.toml`](https://vault.devshift.net/ui/vault/secrets/app-sre/show/creds/app-interface-dev-config-toml) exists and contains your current configuration.
```
Piojo, line 217: I'd rather not put private vault paths in public documentation. They won't help anyone trying to test q-r for any other purposes. This is probably best inside our own repos.

How about:

```
Make sure there is a ./config.dev.toml` file and that its pointing to the following qontract-server address:
...
```

Makefile
fishi0x01: This is actually intended. This comment gets printed on the `help` target. The regex currently requires the comment to be on the same line as the target.

README.md
fishi0x01: I fully agree. Will be more conscious in the future about that stuff :+1: 

Fixed

dockerfiles/hack/run-integration.py
fishi0x01: Was hoping it could :) 
But I will create a separate PR for this now :+1:

fishi0x01: Fixed :+1: 

Created new commits now to make review easier -> will squash after review

fishi0x01: The dry-run change lives here now :)
https://github.com/app-sre/qontract-reconcile/pull/2119

dockerfiles/hack/run-integration.py
fishi0x01: Dry-run change was moved to this PR https://github.com/app-sre/qontract-reconcile/pull/2119

Piojo (approved):

PR #2118, opened 2022-01-26
increase blackbox-exporter-end-point-monitoring resources

geoberle:
oc client gets oom killed on apply

Part of APPSRE-4209


rporres (approved):
You can also set the values in the saas file if you want...

PR #2119, opened 2022-01-26
enhanced dry_run parsing

fishi0x01:
**Motivation**

When I started using the DRY_RUN environment variable, I assumed that it just needs to be declared in order to activate dry run mode, like `DRY_RUN=1`. However, the script exited with a command error. 

**Change**

This PR validates that a proper option is passed to the script. Further, an explanation is printed if a bad option is passed.

### DIFF
--- a/dockerfiles/hack/run-integration.py
+++ b/dockerfiles/hack/run-integration.py
@@ -50,13 +50,17 @@ if LOG_FILE is not None:
 logging.basicConfig(level=logging.INFO,
                     handlers=HANDLERS)
 
+
 def _parse_dry_run_flag(dry_run: str) -> Optional[str]:
     dry_run_options = ['--dry-run', '--no-dry-run']
     if dry_run is not None and dry_run not in dry_run_options:
-        logging.error('Invalid DRY_RUN option given: "%s". Only the following options are allowed: %s', dry_run, dry_run_options)
+        logging.error(
+          f'Invalid DRY_RUN option given: "{dry_run}".'
+          f'Only the following options are allowed: {dry_run_options}'
+        )
         sys.exit(1)
-    if dry_run in ['--dry-run', '--no-dry-run']:
-        return dry_run
+    return dry_run if dry_run else None
+
 
 def build_entry_point_args(command: click.Command, config: str,
                            dry_run: Optional[str], integration_name: str,
@@ -65,9 +69,10 @@ def build_entry_point_args(command: click.Command, config: str,
     if dry_run_flag := _parse_dry_run_flag(dry_run):
         args.append(dry_run_flag)
 
-    # if the integration_name is a known sub command, we add it right before the extra_args
+    # if the integration_name is a known sub command,
+    # we add it right before the extra_args
     if integration_name and isinstance(command, click.MultiCommand) and \
-            command.get_command(None, integration_name): # type: ignore
+            command.get_command(None, integration_name):  # type: ignore
         args.append(integration_name)
 
     if extra_args is not None:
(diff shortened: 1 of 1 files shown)


dockerfiles/hack/run-integration.py
```diff
@@ -50,13 +50,20 @@
...
+    if dry_run in ['--dry-run', '--no-dry-run']:
+        return dry_run
 
 def build_entry_point_args(command: click.Command, config: str,
                            dry_run: Optional[str], integration_name: str,
                            extra_args: Optional[str]) -> list[str]:
     args = ['--config', config]
-    if dry_run is not None:
-        args.append(dry_run)
+    if dry_run_flag := _parse_dry_run_flag(dry_run):
```
Piojo, line 65: Niice, I keep forgetting the `:=` operator in Python 3.8. :)

dockerfiles/hack/run-integration.py
```diff
@@ -50,13 +50,20 @@
 logging.basicConfig(level=logging.INFO,
                     handlers=HANDLERS)
 
+def _parse_dry_run_flag(dry_run: str) -> Optional[str]:
+    dry_run_options = ['--dry-run', '--no-dry-run']
+    if dry_run is not None and dry_run not in dry_run_options:
+        logging.error('Invalid DRY_RUN option given: "%s". Only the following options are allowed: %s', dry_run, dry_run_options)
+        sys.exit(1)
+    if dry_run in ['--dry-run', '--no-dry-run']:
```
Piojo, line 58: DRY:
```suggestion
    if dry_run in dry_run_options:
```
(or better, `if dry_run:` since you've checked everything else in the block above)

dockerfiles/hack/run-integration.py
```diff
@@ -50,13 +50,20 @@
 logging.basicConfig(level=logging.INFO,
                     handlers=HANDLERS)
 
+def _parse_dry_run_flag(dry_run: str) -> Optional[str]:
+    dry_run_options = ['--dry-run', '--no-dry-run']
+    if dry_run is not None and dry_run not in dry_run_options:
+        logging.error('Invalid DRY_RUN option given: "%s". Only the following options are allowed: %s', dry_run, dry_run_options)
+        sys.exit(1)
```
Piojo, line 57: Can we avoid a `sys.exit` at this level?

dockerfiles/hack/run-integration.py
```diff
@@ -50,13 +50,20 @@
 logging.basicConfig(level=logging.INFO,
                     handlers=HANDLERS)
 
+def _parse_dry_run_flag(dry_run: str) -> Optional[str]:
+    dry_run_options = ['--dry-run', '--no-dry-run']
+    if dry_run is not None and dry_run not in dry_run_options:
+        logging.error('Invalid DRY_RUN option given: "%s". Only the following options are allowed: %s', dry_run, dry_run_options)
```
Piojo, line 56: Line too long (how came `flake8` doesn't complain)? Also, we are currently using `f-strings` instead of `%s` in log messages - the performance hit is minimal but the messages become easier to read.

dockerfiles/hack/run-integration.py
fishi0x01: Good catch!
Fixed

dockerfiles/hack/run-integration.py
fishi0x01: Oha. Linter wasnt activated :see_no_evil: 

Great catch! :+1: 
Fixed

dockerfiles/hack/run-integration.py
fishi0x01: Fair point. Replacing with `ValueError` instead

Piojo (approved):

PR #2120, opened 2022-01-28
Update validate_promotions to use the state instead of graphql

lechuk47:
This PR solves a problem with the current saas file deployments based on config_changes. 

```
DeployTarget ------[promotes]---------> TestTarget
                                        - parent_config_hash
```

When a automatic promotion happens, a hash from the DeployTarget configuration is added to TestTarget  in an automated MR (AutoPromote). This MR validation uses `saasherder.validate_promotions` method to ensure the parent_config_hash set in the promotion matches the **current** DeployTarget configuration in app-interface. 

This is causing issues in pipelines that could run in parallel because only the last promotion is valid (configuration is matched with app-interface state).

To solve this, the validation is changed to match the hash in the success state of DeployTarget instead of app-interface. The state is stored for every ref(commit) where DeployTarget has run so every step will be matched with its real parent state insteda of the last one.

### DIFF
--- a/reconcile/openshift_saas_deploy.py
+++ b/reconcile/openshift_saas_deploy.py
@@ -142,7 +142,7 @@ def run(dry_run, thread_pool_size=10, io_dir='throughput/',
 
     # validate that this deployment is valid
     # based on promotion information in targets
-    if not saasherder.validate_promotions(all_saas_files):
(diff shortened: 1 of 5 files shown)


geoberle (approved):
LGTM

PR #2121, opened 2022-01-28
APPSRE-4330 debugging environment

fishi0x01:
Adding debugging to containerized reconcile loop


dockerfiles/Dockerfile.dev
```diff
@@ -0,0 +1,34 @@
...
+# This layer is not needed in the final image, so we can leverage a cache mount
+# to get rid of it. Implement the cache mount un the RUN command when we are able to
+# to use the buildkit features.
+COPY --from=build-image /work/ /tmp/work/
+#RUN --mount=type=cache,target=/tmp/work/,from=build-image,source=/work \
+
+COPY . /work
+WORKDIR /work
+
+RUN microdnf upgrade -y && \
```
Piojo, line 28: If this is for local development, please don't upgrade on every run. It wastes about one minute of execution time, and developers are impatient.

We have an ongoing discussion on how to keep our images reasonably up-to-date without requiring to upgrade on every development build.

Makefile
```diff
@@ -51,18 +55,43 @@ test-container-image: build ## Target to test the final image
...
 	@rm -rf .tox .eggs reconcile.egg-info build .pytest_cache
 	@find . -name "__pycache__" -type d -print0 | xargs -0 rm -rf
 	@find . -name "*.pyc" -delete
+
+dev-clean:
+	@$(DEV_CONF_EXISTS)
+	$(DEV_CONF) docker-compose down
+
+dev-venv: ## Create a local venv for your IDE and remote debugging
+	rm -rf venv
```
Piojo, line 68: Shouldn't this be part of either `clean` or `dev-clean`?

.gitignore
```diff
@@ -12,3 +12,5 @@ coverage.xml
 dist
 venv
 throughput
+config.dev.toml
```
Piojo, line 15: Maybe add `config.debug.toml` too?

Piojo (commented):
Engineers who can `make` are **rare** nowadays!!

maorfr:
maybe a bit late to the party, but there are 2 things that we need to consider:
1. we want to improve our usage of https://github.com/app-sre/app-interface as our dev environment data. this will "force" us to add more test data to that repo, also helping it become a very good demo repo.
2. qontract-reconcile has a vault-less option (see the app-interface settings). we don't need a local vault instance just to develop.

.gitignore
fishi0x01: Added

dockerfiles/Dockerfile.dev
fishi0x01: Good point. Removed it.

Makefile
fishi0x01: Moved it to `clean` now

fishi0x01:
@maorfr Good points :+1: Im all in for https://github.com/app-sre/app-interface. Im currently using that test data for tf integration development.

I reduced now what this PR does. No more docker-compose and/or glueing - just some addition to the containerized reconcile loop, mainly for debugging purposes. 

I moved the core idea [here](https://gitlab.cee.redhat.com/kfischer/qontract-development) instead. If I still deem it useful after using it some time, I will throw it in the ring again :)

maorfr (approved):
LGTM

PR #2125, opened 2022-01-30
[terraform-users] add support for role expiration

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4338

follows up on #2033

this PR adds support for role expiration to terraform-users, by filtering out roles that are beyond the expiration date.

recommendation: review commit by commit.


reconcile/utils/expiration.py
```diff
@@ -1,3 +1,4 @@
+from ast import List
```
geoberle: do you mean `from typing import List`?

reconcile/utils/expiration.py
```diff
@@ -0,0 +1,45 @@
...
+    else:
+        exp_date = datetime.datetime \
+            .strptime(role, '%Y-%m-%d').date()
+        current_date = datetime.datetime.utcnow().date()
+        if current_date < exp_date:
+            return True
+        return False
+
+
+def filter(roles: List[dict], key: str = 'expirationDate') -> List[dict]:
```
geoberle, line 32: you can also use `list[dict]` in current python so you don't need the import at all

reconcile/utils/expiration.py
maorfr: i did!

reconcile/utils/expiration.py
maorfr: done! thank you!

patjlm (approved):
LGTM

PR #2129, opened 2022-02-01
Proposal: add docs about integrations

maorfr:
to make qontract-reconcile and app-interface truly open source, we need to add docs on usage.

this is the first PR (out of hopefully many) adding such docs.


reconcile/cli.py
```diff
@@ -1458,11 +1458,11 @@ def gabi_authorized_users(ctx, thread_pool_size, internal, use_jump_host):
...
 
-@integration.command()
-@enable_deletion(default=False)
-@click.pass_context
-def dyn_traffic_director(ctx, enable_deletion):
-    run_integration(reconcile.dyn_traffic_director, ctx.obj, enable_deletion)
+# @integration.command()
+# @enable_deletion(default=False)
+# @click.pass_context
+# def dyn_traffic_director(ctx, enable_deletion):
```
fishi0x01, line 1464: Is removing this intended? Not mentioned in the PR description

README.md
```diff
@@ -46,20 +46,20 @@ Additional tools that use the libraries created by the reconciliations are also
 - `kafka-clusters`: Manages Kafka clusters via OCM.
 - `ldap-users`: Removes users which are not found in LDAP search.
 - `openshift-clusterrolebindings`: Configures ClusterRolebindings in OpenShift clusters.
-- `openshift-groups`: Manages OpenShift Groups.
+- `openshift-groups`: [Manages OpenShift Groups.](/docs/integreations/openshift_groups.md)
```
rporres, line 49: while I love the word `integreations` as it is indeed, great, this should be `integrations`

maorfr: oops!

reconcile/cli.py
maorfr: no no, it was a mistake

docs/integrations/openshift_network_policies.md
```diff
@@ -0,0 +1,10 @@
+# Enable network traffic between Namespaces via App-Interface (`/openshift/namespace-1.yml`)
+
+To enable network traffic between namespaces, a user must add the following to a namespace declaration:
```
geoberle, line 3: in line 10 it is correctly stated that the traffic is allowed from the source namespace to "this" namespace, but this line suggests something different with the wording "between"

geoberle (approved):
thank you for starting this!

maorfr:
@patjlm could this PR inspire https://github.com/app-sre/qontract-docs?

PR #2131, opened 2022-02-02
Join slack channels

janboll:
Join Slack channels, whenever required and possible.
Join only when initializing instances of SlackApi.

Related ticket: [APPSRE-3988](https://issues.redhat.com/browse/APPSRE-3988)

### DIFF
--- a/reconcile/test/test_utils_slack_api.py
+++ b/reconcile/test/test_utils_slack_api.py
@@ -1,10 +1,11 @@
 import json
 from collections import namedtuple
-from unittest.mock import call, patch
+from unittest.mock import call, patch, MagicMock
 
 import httpretty
 import pytest
 from slack_sdk.errors import SlackApiError
+from slack_sdk.web import SlackResponse
 
 import reconcile
 from reconcile.utils.slack_api import SlackApi, MAX_RETRIES, \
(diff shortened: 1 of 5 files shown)


bkez322:
LGTM (thanks for the test)

geoberle (approved):
LGTM

reconcile/test/test_utils_slack_api.py
```diff
@@ -167,6 +168,35 @@ def test_chat_post_message_missing_channel(slack_api):
         slack_api.client.chat_post_message('test')
 
 
+def test_join_channel_missing_channel(slack_api):
+    """Don't raise an exception when the channel is set."""
```
steveteahan, line 172: The comment should really be raise an exception when the channel isn't set?

reconcile/test/test_utils_slack_api.py
```diff
@@ -167,6 +168,35 @@ def test_chat_post_message_missing_channel(slack_api):
...
+    slack_api.client.channel = None
+    with pytest.raises(ValueError):
+        slack_api.client.join_channel()
+
+
+@patch.object(SlackApi, 'get_channels_by_names')
```
steveteahan, line 178: I'd suggest replacing this with the `mocker` fixture as per: https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-testing-guide/#pytest-best-practices

It should be a simple change because `mocker` is just exposing much of the `unittest.mock` API via a pytest fixture.

steveteahan (approved):
A couple minor things related to the tests, but other than that looks good!

PR #2132, opened 2022-02-02
APPSRE-4119 ES log publishing feature

fishi0x01:
**Motivation**

Allow to specify which kinds of ES logs should be published.
Related schema change https://github.com/app-sre/qontract-schemas/pull/63

**Changes**

- support log_types: [`INDEX_SLOW_LOGS`, `SEARCH_SLOW_LOGS`, `ES_APPLICATION_LOGS`]
- create one dedicated cloudwatch loggroup per log_type per ES domain (AWS recommended)
- create ES resource policy to allow ES to create a logstream. Single policy per account required, as there is a limit of 10 per account.

**Tests**

Test data: https://github.com/app-sre/app-interface/pull/15
3 ES domains bootstrapped in dev account. 1 had a single log. 1 had all 3 logs. 1 had zerio logs. Logs (where expected) were successfully published.

Further, there was no plan diff on prod data.

### DIFF
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -3134,6 +3134,7 @@ class TerrascriptClient:
         self, identifier, account,
         resource, values, output_prefix
     ):
+        ES_LOG_GROUP_RETENTION_DAYS = 180
         tf_resources = []
         publishing_options = []
         # https://docs.aws.amazon.com/opensearch-service/latest/developerguide/createdomain-configure-slow-logs.html
@@ -3175,7 +3176,7 @@ class TerrascriptClient:
             log_group_values = {
                 'name': log_type_identifier,
                 'tags': {},
-                'retention_in_days': 90,
+                'retention_in_days': ES_LOG_GROUP_RETENTION_DAYS,
             }
             region = values.get('region') or \
                 self.default_regions.get(account)
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -1,4 +1,5 @@
 import base64
+from dataclasses import dataclass
 import enum
 import json
 import logging
@@ -11,7 +12,7 @@ import tempfile
 from threading import Lock
 
 from typing import (
-    Any, Dict, List, Iterable, Mapping, MutableMapping, Optional, Tuple
+    Any, Dict, List, Iterable, Mapping, MutableMapping, Optional
 )
 from ipaddress import ip_network, ip_address
 
@@ -142,6 +143,14 @@ class ElasticSearchLogGroupType(enum.Enum):
     INDEX_SLOW_LOGS = 'INDEX_SLOW_LOGS'
     SEARCH_SLOW_LOGS = 'SEARCH_SLOW_LOGS'
     ES_APPLICATION_LOGS = 'ES_APPLICATION_LOGS'
+
+
+@dataclass
+class ElasticSearchLogGroupInfo:
+    account: str
+    account_id: str
+    region: str
+    log_group_identifier: str
 
 
 class TerrascriptClient:
@@ -3149,12 +3158,12 @@ class TerrascriptClient:
 
     def _elasticsearch_aggregate_log_groups_per_account(
         self
-    ) -> Iterable[Tuple[str, str, str, str]]:
-        log_group_identifiers = []
+    ) -> list[ElasticSearchLogGroupInfo]:
+        log_group_infos = []
         for resources in self.account_resources.values():
             for i in resources:
-                res = i.get('resource')
-                ns = i.get('namespace_info')
+                res = i['resource']
+                ns = i['namespace_info']
                 if not ns.get('managedTerraformResources'):
                     continue
                 if res.get('provider') != 'elasticsearch':
@@ -3165,7 +3174,7 @@ class TerrascriptClient:
                 if not log_types:
                     log_types = []
                 for log_type in log_types:
-                    region = ns.get('cluster').get('spec').get('region')
+                    region = ns['cluster']['spec']['region']
                     account = res['account']
                     account_id = self.accounts[account]['uid']
                     lg_identifier = \
@@ -3173,9 +3182,15 @@ class TerrascriptClient:
                             domain_identifier=res['identifier'],
                             log_type=ElasticSearchLogGroupType(log_type),
                         )
-                    tup = (account, region, account_id, lg_identifier)
-                    log_group_identifiers.append(tup)
-        return log_group_identifiers
+                    log_group_infos.append(
+                        ElasticSearchLogGroupInfo(
+                            account=account,
+                            account_id=account_id,
+                            region=region,
+                            log_group_identifier=lg_identifier,
+                        )
+                    )
+        return log_group_infos
 
     def _get_tf_resource_elasticsearch_resource_policy(
         self, account: str
@@ -3189,7 +3204,7 @@ class TerrascriptClient:
         I.e., ideally we aggregate ALL log group identifiers for each
         account first.
         '''
-        log_group_identifiers = \
+        log_group_infos = \
             self._elasticsearch_aggregate_log_groups_per_account()
 
         log_groups_policy = {
@@ -3204,8 +3219,11 @@ class TerrascriptClient:
                     'logs:CreateLogStream',
                 ],
                 'Resource': [
-                    f'arn:aws:logs:{tup[1]}:{tup[2]}:log-group:{tup[3]}:*'
-                    for tup in log_group_identifiers if tup[0] == account
+                    (
+                        f'arn:aws:logs:{info.region}:{info.account_id}'
+                        f':log-group:{info.log_group_identifier}:*'
+                    )
+                    for info in log_group_infos if info.account == account
                 ],
             }]
         }
@@ -3223,7 +3241,7 @@ class TerrascriptClient:
         self, identifier: str, account: str,
         resource: Mapping[str, Any], values: Mapping[str, Any],
         output_prefix: str
-    ) -> Tuple[Iterable[Mapping[str, Any]], Iterable[Mapping[str, str]]]:
+    ) -> tuple[list[dict[str, Any]], list[dict[str, str]]]:
         ES_LOG_GROUP_RETENTION_DAYS = 180
         tf_resources = []
         publishing_options = []
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -3168,9 +3168,7 @@ class TerrascriptClient:
                     continue
                 # res.get('', []) won't work, as publish_log_types is
                 # explicitly set to None if not set
-                log_types = res['publish_log_types']
-                if not log_types:
-                    log_types = []
+                log_types = res['publish_log_types'] or []
                 for log_type in log_types:
                     region = ns['cluster']['spec']['region']
                     account = res['account']
@@ -3240,15 +3238,13 @@ class TerrascriptClient:
         resource: Mapping[str, Any], values: Mapping[str, Any],
         output_prefix: str
     ) -> tuple[list[dict[str, Any]], list[dict[str, str]]]:
-        ES_LOG_GROUP_RETENTION_DAYS = 180
+        ES_LOG_GROUP_RETENTION_DAYS = 90
         tf_resources = []
         publishing_options = []
 
         # res.get('', []) won't work, as publish_log_types is
         # explicitly set to None if not set
-        log_types = resource['publish_log_types']
-        if not log_types:
-            log_types = []
+        log_types = resource['publish_log_types'] or []
         for log_type in log_types:
             log_type_identifier = \
                 TerrascriptClient.elasticsearch_log_group_identifier(
@@ -3268,7 +3264,7 @@ class TerrascriptClient:
                 aws_cloudwatch_log_group(log_type_identifier,
                                          **log_group_values)
             tf_resources.append(log_group_tf_resource)
-            arn = f'${{aws_cloudwatch_log_group.{log_type_identifier}.arn}}'
+            arn = f'${{{log_group_tf_resource.arn}}}'
 
             # add arn to output
             output_name_0_13 = (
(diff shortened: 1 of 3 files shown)


reconcile/utils/terrascript_client.py
```diff
@@ -3155,6 +3166,145 @@ def is_elasticsearch_domain_name_valid(name):
...
+    def elasticsearch_log_group_identifier(
+        domain_identifier: str,
+        log_type: ElasticSearchLogGroupType
+    ) -> str:
+        log_type_name = log_type.value.lower()
+        return f'OpenSearchService__{domain_identifier}__{log_type_name}'
+
+    def _elasticsearch_aggregate_log_groups_per_account(
+        self
+    ) -> Iterable[Tuple[str, str, str, str]]:
```
Piojo, line 3179: Be concrete with the return types. In this case, `list[tuple[str, str, str, str]]` looks more appropriate.

reconcile/utils/terrascript_client.py
```diff
@@ -3155,6 +3166,145 @@ def is_elasticsearch_domain_name_valid(name):
...
+                    continue
+                if res.get('provider') != 'elasticsearch':
+                    continue
+                # res.get('', []) won't work, as publish_log_types is
+                # explicitly set to None if not set
+                log_types = res['publish_log_types']
+                if not log_types:
+                    log_types = []
+                for log_type in log_types:
+                    region = ns.get('cluster').get('spec').get('region')
```
Piojo, line 3195: What is the source for `ns`? Is it a graphql query? If so, you'll be getting weird exceptions and `get` will only be more surprising, see https://issues.redhat.com/browse/APPSRE-3868 . I'd suggest you go with normal dictionary accesses:
```python
ns['cluster']['spec']['region'`]
```

reconcile/utils/terrascript_client.py
```diff
@@ -3155,6 +3166,145 @@ def is_elasticsearch_domain_name_valid(name):
...
+                    account = res['account']
+                    account_id = self.accounts[account]['uid']
+                    lg_identifier = \
+                        TerrascriptClient.elasticsearch_log_group_identifier(
+                            domain_identifier=res['identifier'],
+                            log_type=ElasticSearchLogGroupType(log_type),
+                        )
+                    tup = (account, region, account_id, lg_identifier)
```
Piojo, line 3203: For easier understanding of the logging policy below, may I suggest you convert these into a `dataclass`...

reconcile/utils/terrascript_client.py
```diff
@@ -3155,6 +3166,145 @@ def is_elasticsearch_domain_name_valid(name):
...
+                'Effect': 'Allow',
+                'Principal': {
+                    'Service': 'es.amazonaws.com',
+                },
+                'Action': [
+                    'logs:PutLogEvents',
+                    'logs:CreateLogStream',
+                ],
+                'Resource': [
+                    f'arn:aws:logs:{tup[1]}:{tup[2]}:log-group:{tup[3]}:*'
```
Piojo, line 3234: ... so that these become `resource.region, resource.account_id, resource.lg_identifier` or something like that?

reconcile/utils/terrascript_client.py
```diff
@@ -3155,6 +3166,145 @@ def is_elasticsearch_domain_name_valid(name):
...
+            'es_log_publishing_resource_policy',
+            **log_groups_policy_values,
+        )
+        return resource_policy
+
+    def _get_tf_resource_elasticsearch_log_groups(
+        self, identifier: str, account: str,
+        resource: Mapping[str, Any], values: Mapping[str, Any],
+        output_prefix: str
+    ) -> Tuple[Iterable[Mapping[str, Any]], Iterable[Mapping[str, str]]]:
```
Piojo, line 3253: As above, you want the types for your return values to be as precise as possible.

Piojo (commented):
Overall looks good. I'll let somebody with better understanding of the AWS definitions chime in before approving.

reconcile/utils/terrascript_client.py
fishi0x01: You are right. I misread our dev-guidelines. Abstract types should only be favored for arguments; not return types - makes more sense :+1: 

Fixed

reconcile/utils/terrascript_client.py
fishi0x01: Yeah good point. Already stumbled across that issue during development.

Converted to standard dict access.

reconcile/utils/terrascript_client.py
fishi0x01: Very good suggestion. Fully agree!

Fixed.

reconcile/utils/terrascript_client.py
fishi0x01: Fixed

reconcile/utils/terrascript_client.py
```diff
@@ -3155,6 +3175,154 @@ def is_elasticsearch_domain_name_valid(name):
...
+            **log_groups_policy_values,
+        )
+        return resource_policy
+
+    def _get_tf_resource_elasticsearch_log_groups(
+        self, identifier: str, account: str,
+        resource: Mapping[str, Any], values: Mapping[str, Any],
+        output_prefix: str
+    ) -> tuple[list[dict[str, Any]], list[dict[str, str]]]:
+        ES_LOG_GROUP_RETENTION_DAYS = 180
```
fishi0x01, line 3272: Can someone confirm that 180 days retention period makes sense here?

reconcile/utils/terrascript_client.py
```diff
@@ -3155,6 +3175,154 @@ def is_elasticsearch_domain_name_valid(name):
...
+    def _elasticsearch_aggregate_log_groups_per_account(
+        self
+    ) -> list[ElasticSearchLogGroupInfo]:
+        log_group_infos = []
+        for resources in self.account_resources.values():
+            for i in resources:
+                res = i['resource']
+                ns = i['namespace_info']
+                if not ns.get('managedTerraformResources'):
+                    continue
```
geoberle, line 3195: do you need the check here again? i think it is check already earlier in the program flow [here](https://github.com/app-sre/qontract-reconcile/blob/b066ce7373248e991029378c29a0f560c662be08/reconcile/terraform_resources.py#L446)

reconcile/utils/terrascript_client.py
```diff
@@ -3155,6 +3175,154 @@ def is_elasticsearch_domain_name_valid(name):
...
 
+    @staticmethod
+    def elasticsearch_log_group_identifier(
+        domain_identifier: str,
+        log_type: ElasticSearchLogGroupType
+    ) -> str:
+        log_type_name = log_type.value.lower()
+        return f'OpenSearchService__{domain_identifier}__{log_type_name}'
+
+    def _elasticsearch_aggregate_log_groups_per_account(
```
geoberle, line 3186: nit: from the method name i would have expected some aggregated return per account or filtered ElasticSearchLogGroupInfo objects for a specific account.

reconcile/utils/terrascript_client.py
fishi0x01: Great catch!!

Removed the redundant check :+1:

reconcile/utils/terrascript_client.py
fishi0x01: You are absolutely right! The method name does not make any sense.

Renamed it to `_elasticsearch_filter_all_log_group_infos`.

reconcile/utils/terrascript_client.py
maorfr: can you make 180 the default and also allow to define this value in app-interface?

reconcile/utils/terrascript_client.py
```diff
@@ -3155,6 +3175,152 @@ def is_elasticsearch_domain_name_valid(name):
...
+            }
+            region = values.get('region') or \
+                self.default_regions.get(account)
+            if self._multiregion_account(account):
+                log_group_values['provider'] = f'aws.{region}'
+            log_group_tf_resource = \
+                aws_cloudwatch_log_group(log_type_identifier,
+                                         **log_group_values)
+            tf_resources.append(log_group_tf_resource)
+            arn = f'${{aws_cloudwatch_log_group.{log_type_identifier}.arn}}'
```
maorfr, line 3298: i believe you can use something simpler like `f'${{{aws_cloudwatch_log_group.arn}}}'`?

reconcile/utils/terrascript_client.py
```diff
@@ -3179,6 +3345,22 @@ def populate_tf_resource_elasticsearch(self, resource, namespace_info):
...
+            self._get_tf_resource_elasticsearch_log_groups(
+                identifier=identifier,
+                account=account,
+                resource=resource,
+                values=values,
+                output_prefix=output_prefix
+            )
+        tf_resources += log_group_resources
+
+        resource_policy = self._get_tf_resource_elasticsearch_resource_policy(
```
maorfr, line 3358: are we actually creating this policy only once per account or once per account per usage of the elasticsearch provider?

either way, i think terraform covers us and will only create a resource once, even if it is defined multiple times.

maybe i'm just not reading this right?

reconcile/utils/terrascript_client.py
```diff
@@ -3155,6 +3175,152 @@ def is_elasticsearch_domain_name_valid(name):
...
+    ) -> tuple[list[dict[str, Any]], list[dict[str, str]]]:
+        ES_LOG_GROUP_RETENTION_DAYS = 180
+        tf_resources = []
+        publishing_options = []
+
+        # res.get('', []) won't work, as publish_log_types is
+        # explicitly set to None if not set
+        log_types = resource['publish_log_types']
+        if not log_types:
+            log_types = []
```
maorfr, line 3278: `log_types = resource['publish_log_types'] or []` perhaps?

reconcile/utils/terrascript_client.py
```diff
@@ -3155,6 +3175,152 @@ def is_elasticsearch_domain_name_valid(name):
...
+            for i in resources:
+                res = i['resource']
+                ns = i['namespace_info']
+                if res.get('provider') != 'elasticsearch':
+                    continue
+                # res.get('', []) won't work, as publish_log_types is
+                # explicitly set to None if not set
+                log_types = res['publish_log_types']
+                if not log_types:
+                    log_types = []
```
maorfr, line 3200: log_types = res['publish_log_types'] or []

reconcile/utils/terrascript_client.py
maorfr: in addition, would it make sense to call `_get_tf_resource_elasticsearch_resource_policy` from within `_get_tf_resource_elasticsearch_log_groups`?

reconcile/utils/terrascript_client.py
```diff
@@ -3277,15 +3459,23 @@ def populate_tf_resource_elasticsearch(self, resource, namespace_info):
             es_values["vpc_options"]['security_group_ids'] = security_group_ids
 
         advanced_options = values.get('advanced_options', None)
+        if advanced_options and \
+           not advanced_options.get('override_main_response_version', None):
+            '''
+            By default terraform expects this to be False. If not set (null),
+            then this results in an 'Update' request on each plan/apply.
+            '''
+            advanced_options['override_main_response_version'] = False
```
maorfr, line 3468: not sure i understand this part. does it have to be a part of this PR?

fishi0x01: There is currently an issue that on each plan/apply loop, this change appears:

```
~ advanced_options      = {
            "indices.query.bool.max_clause_count"    = "4000"
          - "override_main_response_version"         = "false" -> null
            "rest.action.multi.allow_explicit_index" = "true"
        }
```

The reason for that is, that terraform tries to set `override_main_response_version` to `null`, because gql returns it in my test data as `None`. However, it won't actually be set to `None`, but to `false` instead, which means this diff will appear on every loop run. By setting it to `false` explicitly, the diff on each subsequent run is clean.

But you are right that this is not necessarily an issue for this ticket. Just sth I stumbled on along the way. Will open a follow up PR afterwards.

reconcile/utils/terrascript_client.py
fishi0x01: Way better. Thx!

reconcile/utils/terrascript_client.py
fishi0x01: Way better. Thx!

reconcile/utils/terrascript_client.py
fishi0x01: Neat trick! Thx!

reconcile/utils/terrascript_client.py
fishi0x01: Reduced it to 90 days now. 

I am wondering if for `AUDIT_LOGS` a longer period might be necessary. Anyways that can be tackled in the follow-up ticket dedicated to `AUDIT_LOGS`.

reconcile/utils/terrascript_client.py
fishi0x01: Tbh I do not see a reason to call `_get_tf_resource_elasticsearch_resource_policy` from within `_get_tf_resource_elasticsearch_log_groups`, because:

- `_get_tf_resource_elasticsearch_log_groups` returns the log_groups of that specific terraform resource. 
- `_get_tf_resource_elasticsearch_resource_policy` returns a policy for the whole AWS account.

This is not well reflected on the naming though. Will rename the function to `_get_elasticsearch_account_wide_resource_policy` to highlight that better.

fishi0x01: > are we actually creating this policy only once per account or once per account per usage of the elasticsearch provider?

To make sure we do not hit any limits, I suggest to have a single resource policy per account for the log-groups. In the current implementation, `_get_tf_resource_elasticsearch_resource_policy` is called on every ES terraform resource. The policy name is static though, which also makes its terraform identifier static:

```
'policy_name': 'es-log-publishing-permissions'
```

By doing so, the resource appears only once in the final terraform `config.json`.

Still, this is kind of hacky as we redundantly re-compute the policy for every ES resource. It would be nice to do it in a single run, but I am currently not aware of a smoother approach.

maorfr: i believe this actually happens a lot throughout our code. you can proceed with that from my perspective!

maorfr (approved):
LGTM

PR #2133, opened 2022-02-03
[openshift-saas-deploy-trigger-upstream-jobs] handle loss of job history

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4364

in case we lose job history, or if the state becomes corrupt due to changing the upstream back and forth between ci instances, we currently fail to trigger new deployments.

this PR solves it essentially by checking if the last build number is different from the stored value in the state, instead of checking if the last build number is larger from what is stored in the state.


lechuk47 (approved):
LGTM

PR #2136, opened 2022-02-03
[saas-file-validator] validate that upstream is not used when ref is a commit sha

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4366

it makes less sense to specify a commit sha as `ref` and using `upstream`, because even if we trigger a deployment after each build, the deployment will change nothing in the environment. the template is identical, and the image that is deployed is either automatically generated according to the `ref` or is hard coded as `IMAGE_TAG` in the parameters. in either case, a new deployment will not change anything.

### DIFF
--- a/reconcile/test/test_saasherder.py
+++ b/reconcile/test/test_saasherder.py
@@ -189,66 +189,8 @@ class TestSaasFileValid(TestCase):
         self.assertFalse(saasherder.valid)
 
     def test_check_saas_file_upstream_not_used_with_commit_sha(self):
-        saas_files = [
-            {
-                'path': 'path1',
-                'name': 'a1',
-                'managedResourceTypes': [],
-                'resourceTemplates':
-                [
-                    {
-                        'name': 'rt',
-                        'url': 'url',
-                        'targets':
-                        [
-                            {
-                                'namespace': {
-                                    'name': 'ns',
-                                    'environment': {
-                                        'name': 'env1'
-                                    },
-                                    'cluster': {
-                                        'name': 'cluster'
-                                    }
-                                },
-                                'ref': 'main',
-                                'upstream': {
-                                    'instance': {
-                                        'name': 'ci'
-                                    },
-                                    'name': 'job'
-                                },
-                                'parameters': {}
-                            },
-                            {
-                                'namespace': {
-                                    'name': 'ns',
-                                    'environment': {
-                                        'name': 'env2'
-                                    },
-                                    'cluster': {
-                                        'name': 'cluster'
-                                    }
-                                },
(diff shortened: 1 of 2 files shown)


reconcile/utils/saasherder.py
```diff
@@ -131,6 +132,12 @@ def _validate_saas_files(self):
                         saas_file_name,
                         environment_name
                     )
+                    # validate upstream not used with commit sha
```
Piojo, line 135: You don't really need the comment here. The method name is self-explanatory. ;)

reconcile/utils/saasherder.py
```diff
@@ -131,6 +132,12 @@ def _validate_saas_files(self):
...
+                    self._validate_upstream_not_used_with_commit_sha(
```
Piojo, line 136: You only log the error here? This is still trying to apply the promotion.

I'd expect you to handle the `False` return path, or otherwise have this method raise an exception...

maorfr: this is handled here: https://github.com/app-sre/qontract-reconcile/blob/b3a8e9bec00ae27203786e243d9a69663a874267/reconcile/saas_file_validator.py#L30-L31

reconcile/utils/saasherder.py
```diff
@@ -226,6 +233,22 @@ def _check_saas_file_env_combo_unique(self, saas_file_name, env_name):
         else:
             self.tkn_unique_pipelineruns[tkn_name] = tkn_long_name
 
+    def _validate_upstream_not_used_with_commit_sha(
+            self,
+            saas_file_name: str,
+            resource_template_name: str,
+            target: dict,
```
Piojo, line 240: We'd usually mark `target` as `Mapping` - maybe even `Mapping[str, Any]`, but this module is already hard enough to make a fuss here.

Piojo (approved):
Minor nits, up to your discretion to handle them.

reconcile/utils/saasherder.py
maorfr: :+1:

reconcile/test/test_saasherder.py
```diff
@@ -129,6 +129,152 @@ def test_check_saas_file_env_combo_not_unique(self):
         self.assertFalse(saasherder.valid)
 
 
+class TestCheckSaasFileUpstreamNotUsedWithCommitSha(TestCase):
+    def test_check_saas_file_upstream_not_used_with_commit_sha(self):
+        saas_files = [
```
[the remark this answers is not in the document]
maorfr: done, and even did it for more tests above. thanks!

PR #2137, opened 2022-02-03
Only join on error

janboll:
Do join channels only if an error up joining slack channels is received

### DIFF
--- a/reconcile/test/test_utils_slack_api.py
+++ b/reconcile/test/test_utils_slack_api.py
@@ -1,5 +1,6 @@
 import json
 from collections import namedtuple
+from typing import Union, Dict
 from unittest.mock import call, patch, MagicMock
 
 import httpretty
@@ -62,6 +63,12 @@ def test_slack_api_config_from_dict():
 
     assert slack_api_config.max_retries == 1
     assert slack_api_config.timeout == 5
+
+
+def new_slack_response(data: Dict[str, Union[bool, str]]):
+    return SlackResponse(client='', http_verb='', api_url='',
+                         req_args={}, data=data, headers={},
+                         status_code=0)
 
 
 def test_instantiate_slack_api_with_config(mocker):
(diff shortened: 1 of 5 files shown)


maorfr (approved):

reconcile/test/test_utils_slack_api.py
```diff
@@ -168,6 +175,39 @@ def test_chat_post_message_missing_channel(slack_api):
         slack_api.client.chat_post_message('test')
 
 
+def test_chat_post_message_channel_not_found(mocker, slack_api):
+    slack_api.client.channel = 'test'
+    mock_join = mocker.patch('reconcile.utils.slack_api.SlackApi.join_channel')
```
steveteahan, line 180: I'd generally use `autospec=True` on all patches unless there is a good reason not to. This might be one of those good reasons not to, but just pointing it out in case it isn't.

reconcile/test/test_utils_slack_api.py
```diff
@@ -168,6 +175,39 @@ def test_chat_post_message_missing_channel(slack_api):
...
+
+
+def test_chat_post_message_ok(slack_api):
+    slack_api.client.channel = 'test'
+    ok_resp = new_slack_response({'ok': True})
+    slack_api.mock_slack_client.return_value.chat_postMessage.side_effect = \
+        ok_resp
+    slack_api.client.chat_post_message('foo')
+    slack_api.mock_slack_client.return_value.chat_postMessage. \
+        assert_called_once
```
steveteahan, line 197: This should be a function call? `assert_called_once()`?

steveteahan (commented):
Two quick fixes related to testing.

reconcile/test/test_utils_slack_api.py
Piojo: While we're at it, do we have a `patch.object` for this `mocker` object? It's usually even more reliable than bare `patch`.

janboll: @Piojo good point, I'll add this to this ticket: APPSRE-4341

reconcile/test/test_utils_slack_api.py
janboll: :facepalm:

steveteahan (approved):

maorfr:
try to use a branch of your fork please: https://service.pages.redhat.com/dev-guidelines/docs/appsre/git/#forks

maorfr:
i have tested this locally and things seem to work well! https://coreos.slack.com/archives/C031E9BRTJB

PR #2139, opened 2022-02-03
Support Python 3.9+ hashlib usage

TGPSKI:
This MR fixes a bug seen on fedramp qontract-reconcile-saas-deploy-trigger-configs with an executor python3.9.6. 

When FIPS is enabled on a system, python restricts usage of hashlib.md5() as it is insecure. This MR adds usedforsecurity=false to the hashlib.md5 invocation, allowing correct operation of the integration.

```
Changed in version 3.9: All hashlib constructors take a keyword-only argument usedforsecurity with default value True. A false value allows the use of insecure and blocked hashing algorithms in restricted environments. False indicates that the hashing algorithm is not used in a security context, e.g. as a non-cryptographic one-way compression function.
```

https://docs.python.org/3.9/library/hashlib.html


reconcile/utils/sharding.py
```diff
@@ -13,7 +13,8 @@ def is_in_shard(value):
     if SHARDS == 1:
         return True
 
-    value_md5 = hashlib.md5(value.encode())
+    h = hashlib.new('md5', usedforsecurity=False)
```
steveteahan, line 16: I think that this change could break the method?

```python
>>> value = 'test'
>>> import hashlib
>>> h = hashlib.new('md5', usedforsecurity=False)
>>> value_md5 = h.update(value.encode())
>>> type(value_md5)
<class 'NoneType'>
>>> value_hex = value_md5.hexdigest()
AttributeError: 'NoneType' object has no attribute 'hexdigest'
```

I think it might be worthwhile to add a quick unit test test or two for this function?

steveteahan: Also, just as an aside, it looks like pylint was complaining about unknown kwargs before. You could keep the `hashlib.md5` call if you just add a comment like:

```python
value = doing_something_that_pylint_doesnt_like()  # pylint: disable=<check-name>
```

TGPSKI: Yes you are 100% right\

steveteahan (approved):
LGTM - thanks for adding the tests!

PR #2141, opened 2022-02-04
Automate metadata verification for SRE checkpoints

Piojo:
Our current process for checkpoints requires reviewing a lot of
metadata manually and cutting tickets for stuff that doesn't meet our
expectations.  In this PR we validate a few fields of each application
and cut a ticket for every field that doesn't meet our requirements.

The idea is to call this as

``` bash
qontract-cli --config config.toml sre-checkpoint-metadata --parent-ticket='CHECKPOINT_TICKET' --app-path='/a/path/to/app.yml'
```

and let it flag any fields that don't make sense and, more imporantly,
do the awful JIRA dance for us.

An example ticket is https://issues.redhat.com/browse/APPSRE-4390


reconcile/checkpoint.py
```diff
@@ -0,0 +1,123 @@
...
+    # shouldn't let important service information in shady self-signed
+    # servers
+    rs = requests.get(url, verify=False)
+    # Codes above NOT_FOUND mean the URL to the document doesn't
+    # exist, that the URL is very malformed or that it points to a
+    # broken resource
+    return (rs.status_code < HTTPStatus.NOT_FOUND)
+
+
+def valid_owners(owners: Iterable[Mapping[str, str]]) -> bool:
```
geoberle, line 53: 💡 we could have a look at an identity store (LDAP, something else?) to figure out if the mail address exists and belongs to a person that still works for the company.

reconcile/checkpoint.py
```diff
@@ -0,0 +1,123 @@
...
+MISSING_DATA_TEMPLATE = PROJ_ROOT / 'templates' / \
+    'jira-checkpoint-missinginfo.j2'
+
+
+def url_makes_sense(url: str) -> bool:
+    """Guesses whether the URL may have a meaningful document.
+
+    Obvious cases are if the document can be fully downloaded, but we
+    also accept that the given document may require credentials that
+    we don't have.
```
geoberle, line 38: depending on the actual URL field that gets checked here, i wonder if we really are fine with not having access to a document (e.g. sopsURL). i'm not talking about implementing authentication in this context, but rather highlight 401s to the caller of qontract-cli

geoberle (commented):
i ❤️ the idea!!

reconcile/checkpoint.py
Piojo: ... after some thougth, I don't have a good generic solution in these cases. And right now, the priority is to assist onboardings report quickly on services that list `http://TODO`. Can we leave docs that require authentication to a next iteration?

reconcile/test/test_checkpoint.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def test_valid_owner(valid_owner) -> None:
+    """Confirm that the valid owner is recognized as such."""
+    assert sut.valid_owners([valid_owner])
+
+
+def test_invalid_owners(invalid_owners):
+    """Confirm that the invalid owners are flagged."""
+    for o in invalid_owners:
```
steveteahan, line 43: Might make more sense to use parameterization here: https://docs.pytest.org/en/6.2.x/parametrize.html

If you parametrize the test, they'll show up in the results as distinct tests, and they'll all run instead of short-circuiting. You could just make this `INVALID_OWNERS` and even `return INVALID_OWNERS` if you really needed the fixture too for some reason.

reconcile/utils/jira_client.py
```diff
@@ -15,21 +18,41 @@ def __init__(self, jira_board, settings=None):
         token_auth = self.secret_reader.read(token)
         self.jira = JIRA(self.server, token_auth=token_auth)
 
-    def get_issues(self, fields=None):
+    def get_issues(self, fields: Optional[Mapping] = None) -> list:
         block_size = 100
         block_num = 0
-        all_issues = []
+
+        all_issues = []  # type: ignore
```
steveteahan, line 25: Just curious why this needs to be ignored?

reconcile/utils/jira_client.py
```diff
@@ -15,21 +18,41 @@ def __init__(self, jira_board, settings=None):
...
             all_issues.extend(issues)
             if len(issues) < block_size:
                 break
             block_num += 1
 
         return all_issues
+
+    def create_issue(self, summary: str, body: str,
+                     labels: Optional[Iterable[str]] = None,
+                     links: Iterable[str] = ()) -> Issue:
```
steveteahan, line 43: Does this not suffer from [this issue](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) because tuples aren't mutable?

reconcile/checkpoint.py
```diff
@@ -0,0 +1,149 @@
...
+    doesn't exist or the specified URL can't be even probed with GET.
+    """
+    # TODO: we really should verify SSL certificates here. We
+    # shouldn't let important service information in shady self-signed
+    # servers
+    rs = requests.get(url)
+    # Codes above NOT_FOUND mean the URL to the document doesn't
+    # exist, that the URL is very malformed or that it points to a
+    # broken resource
+    return rs.status_code < HTTPStatus.NOT_FOUND
```
steveteahan, line 51: Will this handle urls like `https://TODO` or do we need more exception handling?

steveteahan (commented):
Great work on putting this together, should save some time for sure, and encourage others to iterate and improve this process more!

reconcile/checkpoint.py
steveteahan: I missed that this is handled via catching the exception elsewhere, so resolving.

reconcile/utils/jira_client.py
Piojo: mypy is pretty unespecific: 
```
reconcile/utils/jira_client.py:25: error: Need type annotation for "all_issues" (hint: "all_issues: List[<type>] = ...")
```
(and if I try to cast it to `list[Issue]`  I get a suggestion for something truly awful)

reconcile/utils/jira_client.py
Piojo: Exactly.

reconcile/checkpoint.py
Piojo: `requests.get` will raise an exception. We handle all such exceptions below.

reconcile/checkpoint.py
```diff
@@ -0,0 +1,149 @@
...
+    """Guesses whether the URL may have a meaningful document.
+
+    Obvious cases are if the document can be fully downloaded, but we
+    also accept that the given document may require credentials that
+    we don't have.
+
+    The URL is non-sensical if the server is crashing, the document
+    doesn't exist or the specified URL can't be even probed with GET.
+    """
+    # TODO: we really should verify SSL certificates here. We
```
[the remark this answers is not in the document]
Piojo: Yeah, I didn't know I needed to specify `REQUESTS_CA_BUNDLE` in the environment. I've removed this comment.

reconcile/queries.py
```diff
@@ -1944,11 +1945,36 @@ def get_pipelines_providers():
...
+      }
+    }
+  }
+}
+"""
+
+
+def get_simple_jira_boards(app_path: str):
+    gqlapi = gql.get_api()
+    query = JIRA_BOARDS_QUICK_QUERY.replace("APATH", shlex.quote(app_path))
```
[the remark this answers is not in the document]
Piojo: The restriction is that the query has a lot of curly brackets already. I could use `%s` instead, but since it's been deprecated for a long time I settled with replace.

reconcile/utils/jira_client.py
steveteahan: That's the expected behavior for collections: https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html#explicit-types-for-collections

I added `list[Issue]` and see:

```
reconcile/utils/jira_client.py:32: error: Argument 4 to "search_issues" of "JIRA" has incompatible type "**Dict[str, str]"; expected "bool"
reconcile/utils/jira_client.py:33: error: Argument 1 to "extend" of "list" has incompatible type "Union[List[Dict[str, Any]], ResultList[Issue]]"; expected "Iterable[Issue]"
```

I think the first issue is valid. You didn't add a type to `kwargs`, so `dict[str, str]` is inferred because of `kwargs['fields'] = ','.join(fields)`. If you change this to `kwargs: dict[str, Any]`, then that error goes away.

The remaining issue is that `search_issues` return type is not `Issue`, it's: `-> Union[List[Dict[str, Any]], ResultList[Issue]]`.

This can be fixed by changing `all_issues` to: `all_issues: list[Union[dict, Issue]] = []` and the return type to the same.

It takes a little bit of time to run these down, but Mypy is doing the right thing here. Our expectations are not inline with what the Jira client is actually returning. If we're certain that `ResultList[Issue]` will always be returned, despite what the `search_issues` types say, then [this might be helpful](https://mypy.readthedocs.io/en/stable/common_issues.html).

steveteahan (approved):

maorfr:
https://issues.redhat.com/browse/APPSRE-3765

PR #2150, opened 2022-02-07
APPSRE-4330 use host uid for easy vol mounts

fishi0x01:
Ease mounting for dev containers by using host's UID


Piojo (approved):
Looks convenient. Thanks.

PR #2151, opened 2022-02-07
Capture output from jjb runs and log it when needed

lechuk47:
## Overview 

Jira: APPSRE-42245

Changes in this PR capture the output for jjb update method and writes through logging facility. Thi will cause 
the update command logs will be pushed when the command fails or when there changes in the jobs/views etc.


geoberle (approved):
LGTM

PR #2155, opened 2022-02-09
Add --no-send-mails to terraform-users in fedramp

TGPSKI:
This MR allows automated reconciliation of terraform-users without email support.


maorfr:
with this MR we now have the ability to send emails from fedramp: https://rh-osd-gitlab.redhat-osd.local/app-sre/app-interface/-/merge_requests/117

maorfr:
no longer required

PR #2158, opened 2022-02-09
Refactoring SlackApi

janboll:
* Minor refactoring: move calls to _get into seperate methods for easier
  readability, add type hints to these methods
* Refactor slack api creation, remove init_slack functions, create class
  functions for instance creation


janboll:
after short discussion with @steveteahan: going to split this up

PR #2159, opened 2022-02-09
APPSRE-4225: Improve Exception handling

lechuk47:
Improved two aspects: 
* Set `encoding` to `subprocess.run()` to avoid calling decode. 
* Add generic Exception handling. CalledProcessError is only raised if the process could run. If, for example, `jenkins-jobs` executable file does not exist, a `FileNotFoundException` is raised instead of CPE.

### DIFF
--- a/reconcile/utils/jjb_client.py
+++ b/reconcile/utils/jjb_client.py
@@ -238,13 +238,16 @@ class JJB:
                 result = subprocess.run(cmd,
                                         check=True,
                                         stdout=PIPE,
-                                        stderr=STDOUT)
-                out_str = result.stdout.decode("utf-8")
-                if re.search("updated: [1-9]", out_str):
-                    logging.info(out_str)
+                                        stderr=STDOUT,
+                                        encoding="utf-8")
+                if re.search("updated: [1-9]", result.stdout):
+                    logging.info(result.stdout)
             except CalledProcessError as ex:
-                msg = ex.stdout.decode("utf-8")
-                logging.error(msg)
+                logging.exception(ex.stdout)
+                raise
+            except Exception as e:
+                logging.exception(e)
+                raise
 
     @staticmethod
     def get_jjb(args):
--- a/reconcile/utils/jjb_client.py
+++ b/reconcile/utils/jjb_client.py
@@ -223,7 +223,7 @@ class JJB:
         else:
             return file_name.replace('current', 'desired')
 
-    def update(self):
+    def update(self) -> None:
         for name, wd in self.working_dirs.items():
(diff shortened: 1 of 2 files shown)


jfchevrette (approved):
LGTM!

maorfr:
please try to use your fork :)

reconcile/utils/jjb_client.py
```diff
@@ -238,13 +238,16 @@ def update(self):
                 result = subprocess.run(cmd,
                                         check=True,
                                         stdout=PIPE,
-                                        stderr=STDOUT)
-                out_str = result.stdout.decode("utf-8")
-                if re.search("updated: [1-9]", out_str):
-                    logging.info(out_str)
+                                        stderr=STDOUT,
+                                        encoding="utf-8")
+                if re.search("updated: [1-9]", result.stdout):
```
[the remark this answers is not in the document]
lechuk47: Added the type hint, I tried it and it works as expected. Thanks.

reconcile/utils/jjb_client.py
```diff
@@ -238,13 +238,16 @@ def update(self):
...
+                    logging.info(result.stdout)
             except CalledProcessError as ex:
-                msg = ex.stdout.decode("utf-8")
-                logging.error(msg)
+                logging.exception(ex.stdout)
+                raise
```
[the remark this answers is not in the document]
lechuk47: Exactly. I changed to logging.error to just log the stdout attr and let run-integration to log the exception.

steveteahan (approved):

reconcile/utils/jjb_client.py
```diff
@@ -238,13 +238,13 @@ def update(self):
...
-                if re.search("updated: [1-9]", out_str):
-                    logging.info(out_str)
+                                        stderr=STDOUT,
+                                        encoding="utf-8")
+                if re.search("updated: [1-9]", result.stdout):
+                    logging.info(result.stdout)
             except CalledProcessError as ex:
-                msg = ex.stdout.decode("utf-8")
-                logging.error(msg)
+                logging.error(ex.stdout)
```
Piojo, line 246: Why bother? Why not just remove the `try` block here and let somebody up the stack do the logging?

lechuk47: CalledProcessError stdout attribute contains the output of the failed command and this data is not printed with the base exception handling. I think it's worth to log the command's output and not just `command returned with non-zero error`. The exception is raised and managed up in the stack though.

Piojo: Fair enough. I'm convinced.

Piojo (approved):

PR #2160, opened 2022-02-09
add support for default github org

maorfr:
depends on https://github.com/app-sre/qontract-schemas/pull/74

instead of hard coding the name of our org, we can define it as `default: true` in app-interface, allowing other consumers to do the same.

### DIFF
--- a/reconcile/github_org.py
+++ b/reconcile/github_org.py
@@ -29,6 +29,7 @@ ORGS_QUERY = """
       version
       format
     }
+    default
     managedTeams
   }
 }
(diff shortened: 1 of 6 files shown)


reconcile/github_org.py
```diff
@@ -83,23 +84,36 @@
...
     config = {'github': {}}
+    default_found = False
     for org in orgs:
         org_name = org['name']
-        if desired_org_name and org_name != desired_org_name:
+        is_default_org = org.get('default')
+        if default and not is_default_org:
             continue
+        if default_found:
+            raise KeyError('multiple default github org configs found')
```
Piojo, line 100: Print the found defaults in the exception, it'll make troubleshooting much easier.

Piojo: OTOH It'd be even easier if you added all the found defaults to a list and raised at the end if `len(all_defaults) > 1`.

maorfr: done in https://github.com/app-sre/qontract-reconcile/pull/2160/commits/fac34618dd6d40ab51fa31ea2d9e8338eb0fbed9

please take another look

maorfr:
cc @TGPSKI 

once you promote this in fedramp, add `default: true` to the github org file.

PR #2162, opened 2022-02-09
Refactor SlackApi creation

janboll:
Remove code duplication, move creation of SlackApi instances into slack_base.py

This is part of: https://issues.redhat.com/browse/APPSRE-4341


reconcile/test/test_slack_base.py
```diff
@@ -0,0 +1,166 @@
...
+def test_slack_workspace_ok(mocker, slack_workspace):
+    mock_rs = mocker.patch('reconcile.utils.secret_reader.SecretReader.read',
+                           return_value='secret')
+    mock_iu = mocker.patch(
+        'reconcile.utils.slack_api.SlackApi._initiate_usergroups')
+    a = slackapi_from_slack_workspace(slack_workspace, {}, 'dummy')
+    mock_rs.assert_called_once()
+    mock_iu.assert_called_once()
+
+    assert a.channel == 'test'
```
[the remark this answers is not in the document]
janboll: I get you point, however the three fixtures differ not only in content, but also in the structure. It's due to subtle differences in the schema....

reconcile/test/test_slack_base.py
```diff
@@ -0,0 +1,166 @@
...
+    with pytest.raises(ValueError):
+        slackapi_from_slack_workspace({}, {}, 'foo')
+
+
+def test_slack_workspace_ok(mocker, slack_workspace):
+    mock_rs = mocker.patch('reconcile.utils.secret_reader.SecretReader.read',
+                           return_value='secret')
+    mock_iu = mocker.patch(
+        'reconcile.utils.slack_api.SlackApi._initiate_usergroups')
+    a = slackapi_from_slack_workspace(slack_workspace, {}, 'dummy')
```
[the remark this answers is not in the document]
janboll: Good point, changed!

reconcile/test/test_slack_base.py
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+
+def test_slack_workspace_raises():
+    with pytest.raises(ValueError):
+        slackapi_from_slack_workspace({}, {}, 'foo')
+
+
+def test_slack_workspace_ok(mocker, slack_workspace):
+    mock_rs = mocker.patch('reconcile.utils.secret_reader.SecretReader.read',
```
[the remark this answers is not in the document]
janboll: Absolutely right, added!

janboll:
I tested this manually by removing code that is not related in `unleash_watcher`, `jira_watcher` and `slack_usergroups`.
Meaning I used the old and new version of creating slack instances and comparing them with debugger.

reconcile/unleash_watcher.py
```diff
@@ -68,40 +69,16 @@ def calculate_diff(current_state, previous_state):
     return diffs
 
 
-def init_slack_map(unleash_instance):
+def init_slack_map(unleash_instance) -> Dict[str, SlackApi]:
```
geoberle, line 72: nit: you can user the lower case `dict` type hint starting from py3.9

reconcile/unleash_watcher.py
```diff
@@ -68,40 +69,16 @@ def calculate_diff(current_state, previous_state):
...
-            slack_api_kwargs['api_config'] = \
-                SlackApiConfig.from_dict(client_config)
-
-        slack = SlackApi(workspace_name, token, **slack_api_kwargs)
-
-        slack_map[channel] = slack
+        slack_map[channel] = \
+            slackapi_from_slack_workspace(slack_info, settings,
+                                          QONTRACT_INTEGRATION,
+                                          init_usergroups=False)
```
geoberle, line 81: nite: since you got rid of everything in this loop but the core, you could also use a dict comprehension to build the slack_map, but this is more about taste and completely up to you

geoberle (commented):
i like the change - only minor nits

reconcile/test/test_slack_base.py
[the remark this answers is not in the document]
steveteahan: That's fair, I reviewed very quickly and missed that the two that weren't related to permissions were a bit different. In this case, I'm not opinionated because there isn't duplication of fixtures (yet). You still have the problem of the values that are pulled from fixtures which aren't obvious from within the test. That is a bit more of a nit, however, but something to consider in the future.

steveteahan (approved):

reconcile/test/test_slack_base.py
```diff
@@ -0,0 +1,140 @@
...
+            "api_client": create_api_config(),
+            "managedUsergroups": [
+                "foo"
+            ]
+        }
+    }
+
+
+@pytest.fixture
+def slack_base(mocker):
```
steveteahan, line 84: I missed that this was changed. I think there are at least a few approaches to taking care of patching these dependencies in `__init__`:

1. A single patch per fixture, which is very explicit and simple to follow if named well (ex. `patch_secret_reader_read` and `patch_initiate_usergroups`
2. Use xunit style setup/teardown - the setup is now implicit, which might not be as easy to read from within the test, but at least it's a standard convention and can be easily referenced from classes using `self`
3. Doing a lot of setup in a fixture - my biggest concern here is you get the implicit nature of xunit setup, but also have to start doing things like returning tuples to pass the mocked objects. You'll end up with `ValueError` if the fixture ever changes to return more objects.

My general preference is 1 and I'm okay with 2. The third option I'm a little hesitant with unless it's named and documented that people won't use it for a purpose other than it was intended. In this case, may `patch_slack_api_init_deps` and a docstring indicating that this is only meant to patch those dependencies that SlackApi is creating in `__init__` that need to be patched to be able to unit test the class. You might also consider using a NamedTuple or dataclass if you feel strongly about doing it this way. That might help reduce the need to refactor tests in the fixture changes, and reduce the chance that someone pulls the mocks in the wrong order, etc.

janboll: Yeah, I saw that in slack_api and thought that's the way to go :) 
I'll create two individual fixtures.

steveteahan: The difference with the `slack_api` fixture is that it's actually returning an object more than it's intended to just patch objects. It's unlikely that fixture ends up becoming very general use (although a docstring could make the intent clearer for `slack_api`). Also, it uses namedtuple to expose everything, so adding a new mock wouldn't require refactoring all of the tests that don't care about that new mock. Hopefully this reasoning makes sense. I'm just really trying to avoid very general fixtures that end up being used for different use cases and are changed as a result, which could affect existing tests.

PR #2165, opened 2022-02-10
aws api use the account default region instead of terraform state bucket region as default

maorfr:
similar to #1971

the default region to initiate the Session should be the account's default region and not the region where the terraform state bucket is located. since this is only the default region, this change is not expected to cause any issues (famous last words).

### DIFF
--- a/reconcile/test/test_utils_aws_api.py
+++ b/reconcile/test/test_utils_aws_api.py
@@ -12,7 +12,7 @@ def accounts():
             'automationToken': {
                 'path': 'path',
             },
-            'resourcesDefaultRegion': 'region'
+            'resourcesDefaultRegion': 'default-region'
         }
     ]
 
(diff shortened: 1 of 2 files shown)


reconcile/test/test_utils_aws_api.py
```diff
@@ -11,7 +11,8 @@ def accounts():
             'name': 'some-account',
             'automationToken': {
                 'path': 'path',
-            }
+            },
+            'resourcesDefaultRegion': 'region'
```
geoberle, line 15: please change the value "region" to something different from the tf state bucket. how do you feel about creating a test that makes sure the session is really using the accounts region?

maorfr: added a test in https://github.com/app-sre/qontract-reconcile/pull/2165/commits/abc4d9681ff1ae2b60813713a867a81c404a6cb4

geoberle (approved):
LGTM

PR #2168, opened 2022-02-11
APPSRE-4422: Check saas promotions source respositories

lechuk47:
This PR adds additional validations on saas files at PR_CHECK level. 

* Check that channels set in `subscribe` are published by any target
* Check that source repository is equal for publisher and subscriber targets

Check the JIRA ticket for additional information

### DIFF
--- a/reconcile/test/test_saasherder.py
+++ b/reconcile/test/test_saasherder.py
@@ -129,6 +129,59 @@ class TestSaasFileValid(TestCase):
 
         self.assertFalse(saasherder.valid)
 
+    def test_check_saas_file_promotion_same_source(self):
+        rts = [
+            {
+                'name': 'rt_publisher',
+                'url': 'repo_publisher',
+                'targets': [
+                    {
+                        'namespace': {
+                            'name': 'ns',
+                            'environment': {'name': 'env1'},
+                            'cluster': {'name': 'cluster'}
+                        },
+                        "parameters": {},
+                        'ref': '0000000000000',
+                        'promotion': {
+                            'publish': ['channel-1'],
+                        }
+                    }
+                ]
+            },
+            {
+                'name': 'rt_subscriber',
+                'url': 'this-repo-will-not-match-the-publisher',
+                'targets': [
+                    {
+                        'namespace': {
+                            'name': 'ns2',
+                            'environment': {'name': 'env1'},
+                            'cluster': {'name': 'cluster'}
+                        },
+                        "parameters": {},
+                        'ref': '0000000000000',
+                        'promotion': {
+                            'auto': 'true',
+                            'subscribe': ['channel-1'],
+                        }
+                    }
+                ]
+            }
+        ]
+        self.saas_files[0]['resourceTemplates'] = rts
+        saasherder = SaasHerder(
+            self.saas_files,
+            thread_pool_size=1,
+            gitlab=None,
+            integration='',
+            integration_version='',
+            settings={},
+            validate=True
+        )
+        self.assertFalse(saasherder.valid)
+
+
 
 class TestGetMovingCommitsDiffSaasFile(TestCase):
     def setUp(self):
--- a/reconcile/test/test_saasherder.py
+++ b/reconcile/test/test_saasherder.py
@@ -182,7 +182,6 @@ class TestSaasFileValid(TestCase):
         self.assertFalse(saasherder.valid)
 
 
-
 class TestGetMovingCommitsDiffSaasFile(TestCase):
     def setUp(self):
         self.saas_files = [
--- a/reconcile/test/test_saasherder.py
+++ b/reconcile/test/test_saasherder.py
@@ -141,7 +141,7 @@ class TestSaasFileValid(TestCase):
                             'environment': {'name': 'env1'},
                             'cluster': {'name': 'cluster'}
                         },
-                        "parameters": {},
+                        'parameters': {},
                         'ref': '0000000000000',
                         'promotion': {
                             'publish': ['channel-1'],
@@ -159,7 +159,7 @@ class TestSaasFileValid(TestCase):
                             'environment': {'name': 'env1'},
                             'cluster': {'name': 'cluster'}
                         },
-                        "parameters": {},
+                        'parameters': {},
                         'ref': '0000000000000',
                         'promotion': {
                             'auto': 'true',
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -218,7 +218,6 @@ class SaasHerder():
             for saas_file_name, saas_file_paths in duplicates.items():
                 logging.error(msg.format(saas_file_name, saas_file_paths))
 
-        # Promotions have the same source repository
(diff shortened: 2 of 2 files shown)


reconcile/utils/saasherder.py
```diff
@@ -141,10 +145,29 @@ def _validate_saas_files(self):
...
+                                    )
+                                    continue
+                                publications[channel] = rt_ref
+
+                        subscribe = promotion.get('subscribe')
+                        if subscribe:
+                            for channel in subscribe:
+                                subscriptions.setdefault(channel, [])
+                                subscriptions[channel].append(rt_ref)
+
```
fishi0x01, line 170: The function is quite humongous already. Does it make sense to move this portion into a dedicated function, like

```
# input target
# output subscriptions[channel, list[]]
# output publications[channel, RTRef]
def _get_channels_from_pomotion(self, target: Mapping[str, Any]) -> dict[str, list[str]], dict[str, RTRef]
```

That way, we could also remove the comment `# promotion publish channels`

Probably, instead of having `rt_ref` passed around as tuple, we could use a dataclass

```
@dataclass
class RTRef:
    saas_file_path: str
    resource_template_name: str
    resource_template_url: str
```

reconcile/utils/saasherder.py
```diff
@@ -206,14 +229,38 @@ def _validate_saas_files(self):
...
+                            "source repositories\n"
+                            "publisher_saas: {}\n"
+                            "publisher_rt: {}\n"
+                            "publisher_repo: {}\n"
+                            "subscriber_saas: {}\n"
+                            "subscriber_rt: {}\n"
+                            "subscriber_repo: {}\n"
+                            .format(pub_saas, pub_rt_name, pub_rt_url,
+                                    sub_saas, sub_rt_name, sub_rt_url)
+                        )
```
fishi0x01, line 263: Similarly like above: would move this to a dedicated function

```
def _validate_source_repositories(self, subscriptions: Mapping[str, list[str]], publications: Mapping[str, RTRef]) -> bool
```

reconcile/utils/saasherder.py
lechuk47: TBH I think this code needs a refactor. 
Separating this part into a function requires more that what you pointed out, for example, RT_Ref attributes are not in the object, are local to the function, etc. This PR intention is to add a validation to ensure saas specs are correct, not to improve the python code.

About the dataclass, I prefer to maintain the same style on the whole file and not introduce a new concept just for this part, I think it's better to have a model on this that covers the whole saas spec, not just a local ref.

fishi0x01: > This PR intention is to add a validation to ensure saas specs are correct, not to improve the python code.

Only concern is that in my experience dedicated refactor PRs never come. Usually what this means is that stuff just grows out of hand over time to a point were any refactor will be huge amount of work (looking at `terrascript_client.py` for instance :eyes: ). Personally, I feel like every **feature** PR's goal should also be to (slightly) improve the code it affects (if possible). That's why Im always in favor of making small enhancements as you touch a portion of code - and annoying people with it :sweat_smile: . E.g., in this case it would allow us to add some type safety via the function signatures on top of de-cluttering the method. 

That being said, thank you for bearing with me :) I might as well be too idealistic about this. As you said, the effort is bigger than what I initially anticipated and might not really be pragmatic right now.

reconcile/utils/saasherder.py
fishi0x01: Same as comment above :top:

reconcile/utils/saasherder.py
lechuk47: I completely agree with you. 

I'm going to separate the logic you mention. I won't add the dataclass bc I feel it needs a broader change scope, but I think it's worth to slim down this function a bit. 

You made me feel bad... 😝

reconcile/test/test_saasherder.py
```diff
@@ -129,6 +129,58 @@ def test_check_saas_file_upstream_used_with_commit_sha(self):
...
+                'name': 'rt_publisher',
+                'url': 'repo_publisher',
+                'targets': [
+                    {
+                        'namespace': {
+                            'name': 'ns',
+                            'environment': {'name': 'env1'},
+                            'cluster': {'name': 'cluster'}
+                        },
+                        "parameters": {},
```
fishi0x01, line 144: nit: single quotes

reconcile/test/test_saasherder.py
```diff
@@ -129,6 +129,58 @@ def test_check_saas_file_upstream_used_with_commit_sha(self):
...
+                'name': 'rt_subscriber',
+                'url': 'this-repo-will-not-match-the-publisher',
+                'targets': [
+                    {
+                        'namespace': {
+                            'name': 'ns2',
+                            'environment': {'name': 'env1'},
+                            'cluster': {'name': 'cluster'}
+                        },
+                        "parameters": {},
```
fishi0x01, line 162: nit: single quotes

reconcile/utils/saasherder.py
```diff
@@ -206,14 +218,78 @@ def _validate_saas_files(self):
...
 
-        # promotion publish channel duplicates
-        duplicates = [p for p in saas_file_promotion_publish_channels
-                      if saas_file_promotion_publish_channels.count(p) > 1]
-        if duplicates:
-            self.valid = False
-            msg = 'saas file promotion publish channel is not unique: {}'
-            for duplicate in duplicates:
-                logging.error(msg.format(duplicate))
+        # Promotions have the same source repository
```
fishi0x01, line 221: The comment feels redundant here. The function name describes it accurately already.

reconcile/test/test_saasherder.py
lechuk47: changed to single quotes.

reconcile/test/test_saasherder.py
lechuk47: changed to single quotes.

reconcile/utils/saasherder.py
lechuk47: Commend removed.

reconcile/utils/saasherder.py
```diff
@@ -206,14 +218,77 @@ def _validate_saas_files(self):
...
+            rt_ref: Tuple,
+            promotion: dict[str, Any],
+            publications: MutableMapping[str, Tuple],
+            subscriptions: MutableMapping[str, list[Tuple]]):
+        """
+        Function to gather promotion publish and subcribe configurations
+        It validates a publish channel is unique across all publis targets.
+        """
+        publish = promotion.get('publish')
+        if publish:
```
fishi0x01, line 234: I think we can safe one indention here. A neat trick I recently learned from Maor :)

```
publish = promotion.get('publish') or []
for channel in publish:
```

fishi0x01 (approved):
LGTM! Thanks a lot for bearing with me here :bow:

reconcile/utils/saasherder.py
lechuk47: changed. thanks

PR #2172, opened 2022-02-14
APPSRE-3528: Auto-promote targets only if there are changes

lechuk47:
## Overview 
This PR introduces a control to do automatic promotions only if there are changes in the promoting stage. 

## Why? 
Saas deployments are basically a desired/current resources reconciliation. If a deployment succeeds, it can trigger promotions to other saas deployments that comes next in the pipeline. If a deployment succeeds without doing anything (no changes to reconcile) it will trigger its promotions causing the pipeline to advance to the next steps. This can cause undesired effects when a target needs to run before moving to the next stage. 

### Example: 
Deployment(stage) --- > Test(stage) ---> Deployment(production)  --> ... 

Imagine this example pipeline runs and fails at `Test(stage)`. The pipeline won't continue because failed, but if `Test(Stage)` is rerun from the console, tests won't run because the test `Job` already exists in the cluster, but the stage will succeed and promote the production deployment ☠️ 

Introducing this change will prevent this behavior because a change is required to go further in the pipeline. The case explained above could be avoided adding a random suffix on the  `Job` names that is being documented, but this step is required to prevent this to happen if that workaround is not implemented.

### DIFF
--- a/reconcile/openshift_saas_deploy.py
+++ b/reconcile/openshift_saas_deploy.py
@@ -175,8 +175,12 @@ def run(dry_run, thread_pool_size=10, io_dir='throughput/',
     success = not ri.has_error_registered()
     # only publish promotions for deployment jobs (a single saas file)
     if notify:
+        # Auto-promote next stages onlye if there are changes in the
+        # current stage. This prevents promotions made on job re-runs
+        auto_promote = len(actions) > 0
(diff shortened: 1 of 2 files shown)


maorfr (approved):
LGTM

even though our comparison (current vs desired) is not perfect, this is absolutely a good addition. thank you!

PR #2173, opened 2022-02-14
[jira-watcher] post link for deleted issues

maorfr:
even if an issue was deleted, post the message with a link. it's likely it was moved, and a link will still lead to the moved ticket.

worst case - it's a dead link.


geoberle (approved):
SGTM

PR #2176, opened 2022-02-15
Add terraform users wrapper

andreugallofre:
Ticket: APPSRE-4168

### DIFF
--- a/reconcile/terraform_users.py
+++ b/reconcile/terraform_users.py
@@ -1,7 +1,7 @@
 import sys
 
 from textwrap import indent
-from typing import Any
+from typing import Any, Optional
 
 from reconcile.utils import expiration
 from reconcile.utils import gql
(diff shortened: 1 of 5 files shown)


maorfr (approved):
LGTM!

reconcile/terraform_users_wrapper.py
```diff
@@ -0,0 +1,77 @@
...
+from reconcile.utils.sharding import is_in_shard_round_robin
+from reconcile.utils.semver_helper import make_semver
+from reconcile.utils.terrascript_client import TerrascriptClient as Terrascript
+from reconcile.status import ExitCodes
+
+QONTRACT_INTEGRATION = 'terraform-users-wrapper'
+QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
+
+
+def get_accounts_names():
```
rporres, line 18: Please use type hints

reconcile/terraform_users.py
```diff
@@ -57,10 +57,15 @@
 QONTRACT_TF_PREFIX = 'qrtf'
 
 
-def setup(print_to_file, thread_pool_size: int) \
+def setup(print_to_file, thread_pool_size: int, account_name: str) \
```
rporres, line 60: Since `account_name` is not mandatory, shouldn't this be?

```suggestion
def setup(print_to_file, thread_pool_size: int, account_name: Optional[str] = None) \
```

rporres (changes requested):
We agreed on type hints being mandatory

The rest looks good :)

andreugallofre:
@rporres PTAL when you can :)

rporres (approved):

PR #2179, opened 2022-02-16
fix aws-iam-password-reset accounts

maorfr:
we need to iterate over all accounts, and not only the account storing the state.

this is a bug introduced in #2049


lechuk47 (approved):
LGTM

PR #2180, opened 2022-02-16
ensure elasticache parameter group identifier is unique

jfchevrette:
Ref.: https://issues.redhat.com/browse/APPSRE-4517

Given the scenario of two elasticache instances both using the same parameter group definition:

The terraform/terrascript integration fails to create the elasticache parameter group when a region is specified and is different from the default region and there is another parameter group by the same name in the default region. Users expect to be able to define a single parameter group and re-use it for multiple instances, sometimes in different regions.

This is due to the fact the terraform resource identifiers are based on the resource name only and so because the name is not unique it thinks the resource already exists when it tries to create it for a 2nd region.

By appending the region name to the resource identifier (terraform identifier, not AWS resource name) we ensure uniqueness.

This change only applies for when the desired region is different from the default region. This is to avoid updating existing resources that are already deployed and present in the state file, all of which are deployed in the default region already and are not exposed to this issue.


fishi0x01 (approved):
LGTM

PR #2184, opened 2022-02-17
check if blackbox-exporter monitoring provider uses only allowed modules

geoberle:
if a blackbox-exporter endpoint monitoring provider uses a module not listed in app-interface-settings, fail the integration (and as such also PR checks introducing invalid changes). if the app-interface-settings do not mention any allowed blackbox-exporter modules, no check will be conducted.

Part of: https://issues.redhat.com/browse/APPSRE-4523
Depends on schema change: https://github.com/app-sre/qontract-schemas/pull/80

### DIFF
--- a/reconcile/blackbox_exporter_endpoint_monitoring.py
+++ b/reconcile/blackbox_exporter_endpoint_monitoring.py
@@ -141,7 +141,7 @@ def get_endpoints() -> dict[EndpointMonitoringProvider, list[Endpoint]]:
     return endpoints
 
 
-def get_blackbox_providers() -> list:
+def get_blackbox_providers() -> list[EndpointMonitoringProvider]:
     return [
         EndpointMonitoringProvider(**p)
         for p in queries.get_blackbox_exporter_monitoring_provider()
@@ -173,7 +173,8 @@ def run(dry_run: bool, thread_pool_size: int, internal: bool,
     verification_errors = False
     if allowed_modules:
         for p in get_blackbox_providers():
-            if p.blackboxExporter.module not in allowed_modules:
+            if p.blackboxExporter and \
+                 p.blackboxExporter.module not in allowed_modules:
                 LOG.error(
                     f"endpoint monitoring provider {p.name} uses "
                     f"blackbox-exporter module {p.blackboxExporter.module} "
--- a/reconcile/blackbox_exporter_endpoint_monitoring.py
+++ b/reconcile/blackbox_exporter_endpoint_monitoring.py
@@ -141,6 +141,14 @@ def get_endpoints() -> dict[EndpointMonitoringProvider, list[Endpoint]]:
     return endpoints
 
 
+def get_blackbox_providers() -> list:
+    return [
+        EndpointMonitoringProvider(**p)
+        for p in queries.get_blackbox_exporter_monitoring_provider()
+        if p["provider"] == "blackbox-exporter"
+    ]
+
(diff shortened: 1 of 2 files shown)


patjlm (approved):
Nice, LGTM

reconcile/blackbox_exporter_endpoint_monitoring.py
```diff
@@ -141,6 +141,14 @@ def get_endpoints() -> dict[EndpointMonitoringProvider, list[Endpoint]]:
     return endpoints
 
 
+def get_blackbox_providers() -> list:
```
steveteahan, line 144: Any reason not to make this `list[EndpointMonitoringProvider]`? By default it will be `list[Any]` which probably isn't going to help with autocompletion or catching bugs where this is being used?

reconcile/queries.py
```diff
@@ -2568,3 +2569,28 @@ def get_status_pages():
...
+        }
+        exporterUrl
+      }
+    }
+  }
+}
+"""
+
+
+def get_blackbox_exporter_monitoring_provider():
```
steveteahan, line 2594: Should we add a return type here too?

reconcile/blackbox_exporter_endpoint_monitoring.py
geoberle: no reason, just an oversight. thanks for pointing it out. fixed in [9a865e5](https://github.com/app-sre/qontract-reconcile/pull/2184/commits/9a865e5920715fd7e5de89cfdb36a1dcb79b2500)

reconcile/queries.py
geoberle: same

rporres (approved):

reconcile/blackbox_exporter_endpoint_monitoring.py
```diff
@@ -159,6 +167,24 @@ def fill_desired_state(provider: EndpointMonitoringProvider,
 @defer
 def run(dry_run: bool, thread_pool_size: int, internal: bool,
         use_jump_host: bool, defer=None) -> None:
+    # verify blackbox-exporter modules
+    settings = queries.get_app_interface_settings()
+    allowed_modules = settings["endpointMonitoringBlackboxExporterModules"]
```
Piojo, line 172: Given how you use this below, you may want to turn it into a `set`.

reconcile/blackbox_exporter_endpoint_monitoring.py
```diff
@@ -159,6 +167,24 @@ def fill_desired_state(provider: EndpointMonitoringProvider,
...
+    verification_errors = False
+    if allowed_modules:
+        for p in get_blackbox_providers():
+            if p.blackboxExporter and \
+                 p.blackboxExporter.module not in allowed_modules:
```
Piojo, line 177: How long is `allowed_modules` in a typical case? This loop will be `O(len(providers)*len(allowed_modules))`

geoberle: i can make allowed_modules a set, i guess that will make it `O(len(providers)*some_constant)`

reconcile/blackbox_exporter_endpoint_monitoring.py
geoberle: i'm not using `sets` often enough. thanks for the reminder

PR #2187, opened 2022-02-20
[terraform-users] replace inline policies with managed policies

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4505

AWS best practices recommends using customer managed policies instead of inline policies: https://docs.aws.amazon.com/IAM/latest/UserGuide/best-practices.html#best-practice-managed-vs-inline

with this approach, we are no longer using inline policies, which have a limit of 2048 chars:
```
LimitExceeded: Maximum policy size of 2048 bytes exceeded for user
```


geoberle (approved):
LGTM

PR #2189, opened 2022-02-21
add cli help to qontract-reconcile and copy to README

maorfr:
this will make maintenance of README much simpler.


geoberle:
so the idea is to run `qontract-reconcile --help` and put the output into the README?

geoberle (approved):
❤️

maorfr:
> so the idea is to run qontract-reconcile --help and put the output into the README?
indeed!

PR #2191, opened 2022-02-21
Validate mandatory policies

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3251

depends on https://github.com/app-sre/qontract-schemas/pull/81

this PR adds a validation that mandatory policies are attached with each role that has aws groups for an account.
if a role references an aws group in an account that has mandatory policies, and those policies are not referenced in the same role, the integration will fail.


reconcile/utils/terrascript_client.py
```diff
@@ -342,17 +342,41 @@ def populate_iam_groups(self, roles):
...
-                account_name = aws_group['account']['name']
-                account_console_url = aws_group['account']['consoleUrl']
+                account = aws_group['account']
+                account_name = account['name']
+                account_console_url = account['consoleUrl']
+
+                ok = self._validate_mandatory_policies(
+                    account, user_policies, role['name'])
+                if not ok:
+                    raise ValueError('mandatory policies missing')
```
geoberle, line 379: optimization nit: instead of failing on the first group that does not attach all mandatory policies, this code could check all groups and fail afterwards if at least one did not satisfy the rule

feel free to ignore - consecutive PR fixes and PR checks will lead to clean configs as well

reconcile/utils/terrascript_client.py
```diff
@@ -342,17 +342,42 @@ def populate_iam_groups(self, roles):
     def _get_aws_username(user):
         return user.get('aws_username') or user['org_username']
 
+    @staticmethod
+    def _validate_mandatory_policies(account: dict,
```
fishi0x01, line 346: I think we should favor `Mapping[str, Any]` over concrete implementation `dict` according to [styleguide](https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#abstract-collection-types).

reconcile/utils/terrascript_client.py
maorfr: i didn't ignore and see where that got us

reconcile/utils/terrascript_client.py
maorfr: thanks!

reconcile/utils/terrascript_client.py
```diff
@@ -342,17 +342,42 @@ def populate_iam_groups(self, roles):
...
+                                     user_policies: list[dict],
```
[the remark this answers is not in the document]
maorfr: thanks!

geoberle (approved):
LGTM

PR #2192, opened 2022-02-21
[openshift-resources] add support to template resources from github files

maorfr:
similar to the way we can use `vault()`, this PR adds support to template using `github()`. example:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: example
data:
  public_policy.json: |-
    {{ github('org/repo', 'path/to/file', 'ref') }}
```
```


patjlm (approved):
Nice! I see we use the default authentication token from the reconciliation. I am guessing this one has access to all repos declared in app-interface ?
Other than that question, LGTM!

maorfr:
the default token is mostly for making the call with an authenticated users. the assumption is that the content is public. if it's not, _our_ default user is most likely able to get the info.

PR #2197, opened 2022-02-23
[terraform] detect deleted outputs

maorfr:
in some cases, there are outputs to be deleted without any resources being deleted. we currently can not detect that.
we decide whether or not to `terraform apply` based on resources changing, and a partial ability to find values changing for outputs.
this means that the outputs will never be removed. or, only removed the next time a resource is removed.

this PR adds a way to detect outputs being deleted and as a result, perform `terraform apply`.


geoberle (commented):
LGTM

geoberle (approved):
LGTM

PR #2198, opened 2022-02-24
[ocm-aws-infrastructure-access] explicitly define terraform user access

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4397

this PR changes the way we (implicitly) add `network-mgmt` role grants to terraform users, to (explicitly) define accounts that their terraform user should get a role grant.


geoberle (approved):
LGTM 🎉

PR #2202, opened 2022-02-25
[WIP] Check bot access from gitlab-housekeeping if pipelines aren't running

steveteahan:
This is a very early proof of concept MR, so not worth commenting on any best practices, etc. I can probably reuse more of `GitlabForkCompliance` with a bit of refactoring. It's also possible that `gitlab-housekeeping` isn't the right place, but that is where we check if there are any `pipelines` and a common issue there is the missing permissions for the bot.

The motivation here is to fix the common issue of forks not having the correct permissions, which results in the MR failing to merge silently. One way to fix this is to add `gitlab-fork-compliance` to `pr_check.sh` in the tenants repos, but that isn't ideal because we need to expose credentials to them for both app-interface and GitLab.

**Open Questions**

* Do we actually need the other bits in `GitlabForkCompliance` like adding members of the maintainers group to the fork.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -4,6 +4,8 @@ from datetime import datetime, timedelta
 from typing import Dict, List
 
 import gitlab
+from gitlab import GitlabGetError
+from gitlab.const import MAINTAINER_ACCESS
 
 from sretoolbox.utils import retry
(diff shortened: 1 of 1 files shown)


maorfr:
closing for https://issues.redhat.com/browse/APPSRE-4607

PR #2203, opened 2022-02-28
[terraform-resources] introduce terraformProviderResources

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4611

replaces #2201

depends on https://github.com/app-sre/qontract-schemas/pull/87

this PR adds support to provision resources using `terraformProviderResources`, and prepares for addition of new provision providers, such as gcp (or CNA).

review commit by commit as one of them is a black format.


reconcile/terraform_resources.py
```diff
@@ -130,15 +131,15 @@
...
+  {% if account %}account{% endif %}
   identifier
   region
   output_resource_name
   public
   annotations
 }
 ... on NamespaceTerraformResourceS3CloudFront_v1 {
-  account
+  {% if account %}account{% endif %}
```
Piojo, line 142: This is an interesting style. But I see this Jinja-style rendering mixed with `%s` in [other places](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/queries.py#L425) . Should we agree (and then document) that this is the preferred way when it comes to formatting queries?

reconcile/terraform_resources.py
```diff
@@ -300,6 +301,17 @@
...
     name
     managedTerraformResources
+    terraformProviderResources {
+      provider
+      provisioner {
+        name
+      }
+      ... on NamespaceTerraformProviderResourceAWS_v1 {
+        resources {
+          %s
```
Piojo, line 311: As above, we should aim to unify these rendering styles.

reconcile/terraform_resources.py
```diff
@@ -335,7 +347,11 @@
     }
   }
 }
-""" % (indent(TF_RESOURCE, 6*' '))
+""" % (
+    indent(Template(TF_RESOURCE_AWS).render(account=False), 6*' '),
+    indent(Template(TF_RESOURCE_AWS).render(account=True), 6*' ')
+)
+
```
Piojo, line 354: ... even if it means investigating how to deal with nested templates. Or refactoring them.

reconcile/terraform_resources.py
```diff
@@ -458,10 +474,16 @@ def filter_tf_namespaces(namespaces, account_name):
...
-        if not tf_resources:
+        tf_provider_resources = \
+            namespace_info.get('terraformProviderResources') or []
+        tf_resources = namespace_info.get('terraformResources') or []
+        if not (tf_provider_resources or tf_resources):
             tf_namespaces.append(namespace_info)
             continue
+        for provider_resource in tf_provider_resources:
+            if provider_resource['provisioner']['name'] == account_name:
+                tf_namespaces.append(namespace_info)
```
Piojo, line 485: Can this lead to `namespace_info` being appended twice? Once for `tf_profided_resources` and another one for `tf_resources`?

reconcile/utils/terrascript_aws_client.py
```diff
@@ -0,0 +1,4377 @@
...
+                # Skip if account_name is specified
+                if account_name and account != account_name:
+                    continue
+                if account not in self.account_resources:
+                    self.account_resources[account] = []
+                for resource in provider_resource["resources"]:
+                    resource["account"] = account
+                    populate_spec = {
+                        "resource": resource,
+                        "provision_provider": provision_provider,
```
geoberle, line 934: why is provision_provider stored in the populate_spec? it is always "aws" because non-aws resources are rightfully discarded in this context.

do i miss a place it is used?

geoberle (approved):
i love how this change is minimal invasive to the terrascript_(aws)_client.py!
the backwards compatibility of the terraformProviderResources was a good decision

reconcile/terraform_resources.py
maorfr: since there are multiple occurrences in this file already

reconcile/terraform_resources.py
maorfr: especially if it means to to rabbit hole into this topic as part of this PR?

rporres:
I guess this will be updated with the new name `externalResources`? correct?

reconcile/utils/terrascript_aws_client.py
maorfr: future compatibility with additional providers?

maorfr:
replaced by #2441

PR #2204, opened 2022-02-28
APPSRE-4403 Update ldap_client to use a-i settings

bhushanthakur93:
This change is dependent on this [schema](https://github.com/app-sre/qontract-schemas/pull/90) and should not be merged in before schema update is merged. 

Definitely appreciate any feedback as this is my first time using python and also the testing framework.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ throughput
 config.toml
 config.*.toml
 .container_tmp
+.idea
--- a/reconcile/test/test_ldap_users.py
+++ b/reconcile/test/test_ldap_users.py
@@ -40,15 +40,15 @@ def patched_queries_get_app_interface_settings(mocker):
     queries_get_app_interface_settings = mocker.patch.object(
         ldap_users.queries, "get_app_interface_settings", autospec=True
     )
-    queries_get_app_interface_settings.return_value = {}
+    queries_get_app_interface_settings.return_value = {
+        "ldap": {"baseDn": "test", "serverUrl": "testUrl"}
+    }
     return queries_get_app_interface_settings
 
 
 @pytest.fixture
 def mocked_ldap_client(mocker):
-    mock_ldap_client = mocker.patch.object(
-        ldap_users.LdapClient, "from_settings", autospec=True
-    )
+    mock_ldap_client = mocker.patch("reconcile.ldap_users.LdapClient", autospec=True)
     dummy_ldap_client = mocker.Mock(spec=ldap_users.LdapClient)
     dummy_ldap_client.__enter__ = dummy_ldap_client
     dummy_ldap_client.__exit__ = mocker.Mock(return_value=None)
--- a/reconcile/test/test_ldap_users.py
+++ b/reconcile/test/test_ldap_users.py
@@ -40,15 +40,15 @@ def patched_queries_get_app_interface_settings(mocker):
     queries_get_app_interface_settings = mocker.patch.object(
         ldap_users.queries, "get_app_interface_settings", autospec=True
     )
-    queries_get_app_interface_settings.return_value = {
-        "ldap": {"baseDn": "test", "serverUrl": "testUrl"}
-    }
+    queries_get_app_interface_settings.return_value = {}
     return queries_get_app_interface_settings
 
 
 @pytest.fixture
 def mocked_ldap_client(mocker):
-    mock_ldap_client = mocker.patch("reconcile.ldap_users.LdapClient", autospec=True)
+    mock_ldap_client = mocker.patch.object(
+        ldap_users.LdapClient, "from_settings", autospec=True
+    )
     dummy_ldap_client = mocker.Mock(spec=ldap_users.LdapClient)
     dummy_ldap_client.__enter__ = dummy_ldap_client
     dummy_ldap_client.__exit__ = mocker.Mock(return_value=None)
--- a/reconcile/test/test_ldap_users.py
+++ b/reconcile/test/test_ldap_users.py
@@ -85,14 +85,13 @@ def test_ldap_users_no_dry_run(
     ldap_users.run(False, None)
 
     patched_mr_client_gateway.assert_called_once()
-    patched_create_delete_user.assert_called_once()
-    assert patched_create_delete_user.call_args == mocker.call(
+    patched_create_delete_user.assert_called_once_with(
         "username3",
         [
-            {"type": 0, "path": "datablah"},
-            {"type": 1, "path": "datatest_path3"},
-            {"type": 2, "path": "dataanother_test_path3"},
-            {"type": 3, "path": "datayet_another_test_path3"},
+            {"path": "datablah", "type": 0},
+            {"path": "datatest_path3", "type": 1},
+            {"path": "dataanother_test_path3", "type": 2},
+            {"path": "datayet_another_test_path3", "type": 3},
         ],
     )
     assert patched_create_delete_user.method_calls[0][0] == "().submit"
--- a/reconcile/test/test_ldap_users.py
+++ b/reconcile/test/test_ldap_users.py
@@ -5,26 +5,31 @@ from reconcile import ldap_users
 
 @pytest.fixture
 def mocked_queries_get_users(mocker):
-    queries_get_users_mock = mocker.patch.object(ldap_users.queries, "get_users", autospec=True)
+    queries_get_users_mock = mocker.patch.object(
+        ldap_users.queries, "get_users", autospec=True
+    )
     queries_get_users_mock.return_value = [
-        {'org_username': 'username1',
-         'requests': [{'path': 'test_path1'}],
-         'queries': [{'path': 'another_test_path1'}],
-         'gabi_instances': [{'path': 'yet_another_test_path1'}],
-         'path': 'blah'
-         },
-        {'org_username': 'username2',
-         'requests': [{'path': 'test_path2'}],
-         'queries': [{'path': 'another_test_path2'}],
-         'gabi_instances': [{'path': 'yet_another_test_path2'}],
-         'path': 'blah'
-         },
-        {'org_username': 'username3',
-         'requests': [{'path': 'test_path3'}],
-         'queries': [{'path': 'another_test_path3'}],
-         'gabi_instances': [{'path': 'yet_another_test_path3'}],
-         'path': 'blah'
-         }
+        {
+            "org_username": "username1",
+            "requests": [{"path": "test_path1"}],
+            "queries": [{"path": "another_test_path1"}],
+            "gabi_instances": [{"path": "yet_another_test_path1"}],
+            "path": "blah",
+        },
+        {
+            "org_username": "username2",
+            "requests": [{"path": "test_path2"}],
+            "queries": [{"path": "another_test_path2"}],
+            "gabi_instances": [{"path": "yet_another_test_path2"}],
+            "path": "blah",
+        },
+        {
+            "org_username": "username3",
+            "requests": [{"path": "test_path3"}],
+            "queries": [{"path": "another_test_path3"}],
+            "gabi_instances": [{"path": "yet_another_test_path3"}],
+            "path": "blah",
+        },
     ]
 
     return queries_get_users_mock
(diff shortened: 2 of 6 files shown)


reconcile/test/test_utils_ldap_client.py
```diff
@@ -0,0 +1,23 @@
...
+        {"attributes": {"uid": ["user3"]}},
+    ]
+    mock_connection_search.return_value = None, None, result, None
+
+    settings = {"ldap": {"baseDn": "test", "serverUrl": "testUrl"}}
+    with LdapClient.from_settings(settings) as ldap_client:
+        uids = ["user1", "user2", "user3"]
+        ldap_client.get_users(uids)
+
+    mock_connection_search.assert_called_once()
```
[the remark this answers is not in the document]
bhushanthakur93: Yes! Thank you for pointing it out.

reconcile/utils/ldap_client.py
```diff
@@ -1,39 +1,39 @@
...
+class LdapClient:
+    """
+    LdapClient that wraps search functionality from ldap3 library
+    and exposes through its own method. The client should be used
+    `with` statement to allow context manager to release connection resource
+    appropriately.
+    """
 
-    config = get_config()
+    def __init__(self, base_dn, connection):
```
steveteahan, line 12: Please add type hints here.

reconcile/utils/ldap_client.py
```diff
@@ -1,39 +1,39 @@
...
+        return self
 
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.connection.unbind()
 
-def get_users(uids):
-    global _base_dn
-
-    with init_from_config() as client:
+    def get_users(self, uids: list) -> set:
```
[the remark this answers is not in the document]
steveteahan: +1 also to adding the type within the collection as @Piojo pointed out.

reconcile/utils/ldap_client.py
```diff
@@ -1,39 +1,39 @@
...
-        _, _, results, _ = client.search(
-            _base_dn, f"(&(objectclass=person)(|{user_filter}))", attributes=["uid"]
+        _, _, results, _ = self.connection.search(
+            self.base_dn, f"(&(objectclass=person)(|{user_filter}))", attributes=["uid"]
         )
-        # pylint: disable=not-an-iterable
         return set(r["attributes"]["uid"][0] for r in results)
+
+    @staticmethod
+    def from_settings(settings):
```
steveteahan, line 31: Please add type hints, noting that you can add forward references to the class name in quotations:

```python
@staticmethod
def from_settings(settings: <type>) -> "LdapClient":
```

I'd also put together a doc string indicating what settings is expecting.

reconcile/utils/ldap_client.py
```diff
@@ -1,39 +1,39 @@
...
+        connection = Connection(
+            Server(settings["ldap"]["serverUrl"], get_info=ALL),
+            None,
+            None,
+            client_strategy=SAFE_SYNC,
+        )
+        base_dn = settings["ldap"]["baseDn"]
+        return LdapClient(base_dn, connection)
```
steveteahan, line 39: You can actually change this to a `classmethod`, your first argument will be `cls` which is a reference to the class, and then return `cls(base_dn, connection)`. Not the biggest deal, but avoids hard-coding the name of the class and maybe could lead to fewer issues down the road with subclasses.

reconcile/test/test_utils_ldap_client.py
```diff
@@ -0,0 +1,23 @@
+from ldap3 import Connection
+from reconcile.utils.ldap_client import LdapClient
+
+
+def test_ldap_client(mocker):
```
steveteahan, line 5: Maybe this is `test_ldap_client_from_settings` and we also want to test that it works as expected when initialized directly?

reconcile/test/test_ldap_users.py
```diff
@@ -0,0 +1,112 @@
...
+            {"type": 0, "path": "datablah"},
+            {"type": 1, "path": "datatest_path3"},
+            {"type": 2, "path": "dataanother_test_path3"},
+            {"type": 3, "path": "datayet_another_test_path3"},
+        ],
+    )
+    assert mocked_create_delete_user.method_calls[0][0] == "().submit"
+
+
+def test_ldap_users_dry_run(
```
steveteahan, line 101: These sorts of tests provide some basic level of testing, but so much has been patched that in reality I'd just be careful about relying on them. To that end, how have you tested this change manually?

steveteahan (commented):
Looks good overall! Most of the feedback is related to some Python best practices and style stuff which you'll pick up quickly.

reconcile/test/test_ldap_users.py
bhushanthakur93: Yes, I agree. Too much mocking or patching in this case actually can masks certain issues that can rise when the stuff actually runs. I believe an integration test with some fake LDAP server with some user data prepopulated would be better in this case given it's an external dependency and that can involve lot of failure points which aren't usually covered by unit tests in general. 

This is how I went with testing my changes.
- Updated qontract-schema with LDAP schema per [commit](https://github.com/app-sre/qontract-schemas/commit/7329bcf5b4c12cf1fcf49460aed8d7f7f70eb02a)
- Update app-interface to define those settings. You can see the MR in the gitlab project.
- Ran the qontract-server locally followed by a dry run through qontract-server cli. 

Let me know if that makes sense.

bhushanthakur93:
@steveteahan Excellent feedback. I'll work through the feedback now.

.gitignore
```diff
@@ -15,3 +15,4 @@ throughput
 config.toml
 config.*.toml
 .container_tmp
+.idea
```
maorfr, line 18: this probably belongs in your global gitignore

reconcile/utils/ldap_client.py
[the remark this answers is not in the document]
bhushanthakur93: This is addressed now.

reconcile/test/test_utils_ldap_client.py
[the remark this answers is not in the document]
bhushanthakur93: This was a bit tricky given the patching involved with `Connection`. There was a way where I could do argument matching by looking at `call_args.args` and `kwargs` but it seemed like hacky and not readable compared to just using pure Mock object. 
Can you think of a better way with this?

reconcile/test/test_ldap_users.py
```diff
@@ -0,0 +1,112 @@
...
+    mocked_queries_get_app_interface_settings,
+    mocked_ldap_client,
+    mocked_mr_client_gateway,
+    mocked_create_delete_user,
+):
+
+    ldap_users.run(True, None)
+
+    assert mocked_mr_client_gateway.call_count == 0
+    assert mocked_create_delete_user.call_count == 0
```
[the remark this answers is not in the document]
bhushanthakur93: addressed.

reconcile/test/test_ldap_users.py
```diff
@@ -0,0 +1,112 @@
+import pytest
+
+from reconcile import ldap_users
+
+
+@pytest.fixture
+def mocked_queries_get_users(mocker):
```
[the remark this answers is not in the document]
bhushanthakur93: Good suggestion. Made the change.

reconcile/test/test_utils_ldap_client.py
bhushanthakur93: I added a test for direct initialization but I'm not really sure the value it adds.

.gitignore
bhushanthakur93: Good point, I'll add it in my configuration.

reconcile/utils/ldap_client.py
bhushanthakur93: Good suggestion! Made the change.

reconcile/test/test_ldap_users.py
Piojo: We could import [ldap-test](https://pypi.org/project/python-ldap-test/) to mock a local LDAP server. I most definitely don't want our tests to try to poke RH's LDAP, but a local server with mock data can help here. Up to you to decide if it's worth the effort.

reconcile/test/test_utils_ldap_client.py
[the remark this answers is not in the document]
Piojo: In this particular case I care a lot that we are producing the right filter, so how about:
```suggestion
    mock_connection_search.assert_called_once_with(
        'test',
        '(&(objectClass=person)(|(uid=user1)(uid=user2)(uid=user3)))'
    )
```
(warning: I haven't really tried to balance the parens)

reconcile/test/test_utils_ldap_client.py
```diff
@@ -0,0 +1,43 @@
...
+    mock_connection_search.return_value = None, None, connection_search_result, None
+
+    settings = {"ldap": {"baseDn": "test", "serverUrl": "testUrl"}}
+    with LdapClient.from_settings(settings) as ldap_client:
+        uids = ["user1", "user2", "user3"]
+        ldap_client.get_users(uids)
+
+    mock_connection_search.assert_called_once()
+    mock_connection_bind.assert_called_once()
+    mock_connection_unbind.assert_called_once()
```
Piojo, line 30: And then these don't seem to be taking any arguments, so:
```suggestion
    mock_connection_bind.assert_called_once_with()
    mock_connection_unbind.assert_called_once_with()
```

reconcile/test/test_utils_ldap_client.py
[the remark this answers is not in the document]
Piojo: On a second thought, you are testing here that you can instantiate the object, so just remove the call to `get_users` altogether and move it to its own test method. Just
```python
with LdapClient.from_settings(settings) as c:
    assert c.base_dn == 'test'
    assert c.connection.server == 'testUrl'
```
and we can see how we test the search method in a separate function.

reconcile/test/test_utils_ldap_client.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def test_ldap_client(mocker, connection_search_result):
+    mocked_connection = mocker.Mock(spec=ldap3.Connection)
+    mocked_connection.search.return_value = None, None, connection_search_result, None
+
+    with LdapClient("test", mocked_connection) as ldap_client:
```
Piojo, line 37: Same comment as above, in this case you care that you are attempting to connect to the right serverurl and have stored the right base DN. Add assertions for those.

bhushanthakur93: The refactor addresses this now.

reconcile/utils/ldap_client.py
```diff
@@ -1,39 +1,40 @@
...
+    def get_users(self, uids: Iterable[str]) -> set[str]:
+        search_filter = self.apply_search_filter_by_person(uids)
+        _, _, results, _ = self.connection.search(
+            self.base_dn, search_filter, attributes=["uid"]
         )
-        # pylint: disable=not-an-iterable
         return set(r["attributes"]["uid"][0] for r in results)
+
+    @staticmethod
+    def apply_search_filter_by_person(uids: Iterable[str]) -> str:
```
Piojo, line 38: Nit: this is more like `build_search_filter`, naming is hard.

bhushanthakur93:
> I share Steve's concerns. We had a module in which we could inject dependencies, which made testing easier, especially for `__init__`. I'd rather keep that initial design.

cc @steveteahan 
Yes, I realized it after I made those changes. Initially I wanted to go with the route of injecting the dependency but later realized got rid of it altogether. Let me take another look at previous commit.

bhushanthakur93:
> > @Piojo I made another attempt at refactoring based on the feedback. On a high level this is what I did.
> > 
> > * Remove `LdapClient.from_settings`, instead made constructor take in the `settings` argument.
> > * Exposed another method `apply_search_filter_by_person` for just the filter to facilitate unit testing to ensure we provide right filter. It also helped decouple test from the implementation detail as compared to earlier when we were verifying `connection.search` arguments.
> > 
> > I think overall this attempt at refactor seems better compared to the earlier one. Let me know what you think.
> 
> I have some concerns with the refactor:
> 
> 1. `LdapClient` creates its dependencies (`Connection`), so we cannot easily customize this dependency or mock it (so we need to patch it instead)
Agreed. I realized this after making the change.
> 2. `LdapClient` takes a `settings` argument which is basically a dict tied to the implementation of some graphql queries. I have to read through `__init__` to have an idea of what to pass here, and if this class grows it's going to become increasingly difficult to not make a mistake here
Agreed as well.
> 3. We've added `apply_search_filter_by_person` to make things easier to test, but it should probably be private (leading underscore like `_apply_search_filter_by_person`)? That being said, I don't think the original test of verifying `connection.search` arguments is a terrible thing. I'd expect to be verifying behavior like this in classes that are really just coordinating some calls to an external service.
Yes, I wanted this to be private but for sake of testability I did it this way. The piece that I didn't like about verifying method call is it makes test coupled to the implementation. For e.g if we were to change the `connection.search` call, we would be needing to update the test as well. This blog helped me with the decision https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html
> 
> The original code should have been testable if you focus on the `__init__` path rather than `from_settings`. The whole purpose of `from_settings` was to offload the parts that make this class difficult to use and test. You could have equally relied on the caller of the code (`ldap_users` integration) to build `Connection` and parse the results of the graphql queries. I think that if you take the approach of:
> 
> 1. Test most functionality through `__init__`
> 2. Focus the testing of `from_settings` on what it's actually doing: parses the settings properly, creates a `Connection`, and that your `LdapClient` is initialized
> 
> It might also help to think of `from_settings` as just a factory method. I think in Java this would be a separate class, but since Python doesn't force class usage, you could have a function called `create_ldap_client` that took care of parsing settings and creating the required dependencies. If `create_ldap_client` existed, I think you'd probably focus your testing on `LdapClient` directly? If you had a separate function for the factory that'd be fine too, but the classmethod `from_<something>` pattern is something that you'll come across in Python.
> 
> This may all sound a bit nitpicky, but some of the patterns introduced here are similar to what has made many other classes difficult to test and use. I understand the motivation behind them, but I think it may just reinforce bad patterns elsewhere. I'm happy to discuss further in Slack or we can jump on a quick call.

I prefer nitpicking so that I get to learn more :)

steveteahan:
> Yes, I wanted this to be private but for sake of testability I did it this way. The piece that I didn't like about verifying method call is it makes test coupled to the implementation. For e.g if we were to change the connection.search call, we would be needing to update the test as well. This blog helped me with the decision https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html

I don't think that we should make methods public just so that we can justify testing them. I also agree that verifying state is generally simpler and less likely to break than behavior, but in this case you're still testing the behavior by asserting that mocks were called. On top of that behavior verification, we've also added a public method to the API for people to use when they really don't need to.

reconcile/test/test_ldap_users.py
```diff
@@ -0,0 +1,112 @@
...
+    patched_create_delete_user.assert_called_once()
+    assert patched_create_delete_user.call_args == mocker.call(
+        "username3",
+        [
+            {"type": 0, "path": "datablah"},
+            {"type": 1, "path": "datatest_path3"},
+            {"type": 2, "path": "dataanother_test_path3"},
+            {"type": 3, "path": "datayet_another_test_path3"},
+        ],
+    )
```
Piojo, line 97: Isn't this the same as
```suggestion
    patched_create_delete_user.assert_called_once_with(
        "username3",
        [
            {"type": 0, "path": "datablah"},
            {"type": 1, "path": "datatest_path3"},
            {"type": 2, "path": "dataanother_test_path3"},
            {"type": 3, "path": "datayet_another_test_path3"},
        ],
    )
```

reconcile/test/test_utils_ldap_client.py
```diff
@@ -0,0 +1,50 @@
...
+        uids = ["user1", "user2", "user3"]
+        ldap_client.get_users(uids)
+
+    assert mocked_connection.search.call_args == mocker.call(
+        "test",
+        "(&(objectclass=person)(|(uid=user1)(uid=user2)(uid=user3)))",
+        attributes=["uid"],
+    )
```
Piojo, line 50:
```suggestion
    mocked_connection.search.assert_called_once_with(
        "test",
        "(&(objectclass=person)(|(uid=user1)(uid=user2)(uid=user3)))",
        attributes=["uid"],
    )
```

Piojo (approved):
It's much nicer now. Thanks. :)

reconcile/test/test_utils_ldap_client.py
```diff
@@ -0,0 +1,50 @@
...
+
+    settings = {"ldap": {"baseDn": "test", "serverUrl": "testUrl"}}
+    with LdapClient.from_settings(settings) as ldap_client:
+        assert ldap_client.base_dn == "test"
+        assert mock_connection_bind.call_args.args[0].host == "testUrl"
+
+
+def test_ldap_client(mocker, connection_search_result):
+    mocked_connection = mocker.Mock(spec=ldap3.Connection)
```
steveteahan, line 28: No patching, I like it!

I think you just want `mocker.create_autospec` here. Check out the docs in `unittest.mock`.

steveteahan (approved):
Minor note on using `create_autospec`, but besides that lgtm!

PR #2205, opened 2022-02-28
Enable untyped def checks in Mypy

steveteahan:
Mypy only checks typed methods by default. This can be a sane default
for teams starting out with types. Now that we're enforcing types, it
makes more sense to enable a stricter mode so that Mypy will always
check functions for errors. There are >80 files that have errors that
Mypy would detect currently, but they were being ignored because there
weren't type hints. This should stop that behavior for new code and for
files that don't have existing errors.

See this link for more information on the motivation: https://mypy.readthedocs.io/en/stable/common_issues.html#no-errors-reported-for-obviously-wrong-code

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,6 +8,271 @@ plugins = pydantic.mypy
 ; More context here: https://github.com/python/mypy/issues/9091
 no_implicit_optional = True
 
+check_untyped_defs = True
+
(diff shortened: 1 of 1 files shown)


geoberle (approved):
LGTM

steveteahan:
Related ticket: https://issues.redhat.com/browse/APPSRE-4614

PR #2207, opened 2022-02-28
Improve logging of validation issues

Piojo:
We expect bad inputs, and we expect to report it as tickets. The stack
traces of `logging.exception` make it hard to understand what went
wrong or to spot the ticket that was actually created. We can still
get it under debug, if we want it.


jfchevrette (approved):
LGTM

PR #2208, opened 2022-02-28
Enable warn_unused_ignores option in Mypy

steveteahan:
This changes ensures that there aren't Mypy ignores that are stale
that could mask other type issues in the future. This also includes
the removal of several ignore statements that no longer appear to be
necessary, or were made unnecessary via other methods (ignoring
missing imports).

### DIFF
--- a/reconcile/checkpoint.py
+++ b/reconcile/checkpoint.py
@@ -133,7 +133,7 @@ def report_invalid_metadata(
     """
     if dry_run:
         do_cut = partial(
-            render_template,  # type: ignore
+            render_template,
             template=MISSING_DATA_TEMPLATE,
             name=app["name"],
             path=path,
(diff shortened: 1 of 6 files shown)


reconcile/checkpoint.py
```diff
@@ -133,7 +133,7 @@ def report_invalid_metadata(
     """
     if dry_run:
         do_cut = partial(
-            render_template,  # type: ignore
+            render_template,
```
Piojo, line 136: Weird. Mypy was totally bitching about this partial because on the other branch of the `if` the called function has a different signature (but same return type). Glad to see they corrected it.

Piojo (approved):

reconcile/checkpoint.py
steveteahan: I was surprised by how many of these ignores were already stale as well, which goes to show you it's good that we have a method for catching them. Thanks for the review!

PR #2209, opened 2022-02-28
Add cronjobs to run e2e-tests; Add dry_run mode to e2e_tests.

OliviaHY:
#### What:
* Add dry_run mode to e2e_tests.
* Add cronjobs to run e2e-tests in cluster.
 
#### Why:
As part of efforts to pivoting away from Jenkins.

#### Tickets:
APPSRE-3615

#### Validation:
Deployed yaml contains only new e2e related cronjobs to Staging and they ran successfully.

### DIFF
--- a/helm/qontract-reconcile/values-internal.yaml
+++ b/helm/qontract-reconcile/values-internal.yaml
@@ -461,3 +461,55 @@ cronjobs:
   - name: XDG_CACHE_HOME
     value: /tmp/.cache
   extraArgs: --gitlab-project-id=${APP_INTERFACE_PROJECT_ID} --reports-path=/tmp/report
+- name: create-namespace
+  command: e2e-tests
+  concurrencyPolicy: Replace
+  restartPolicy: Never
+  resources:
+    requests:
+      memory: 300Mi
+      cpu: 100m
+    limits:
+      memory: 400Mi
+      cpu: 200m
+  cron: '0 8,20 * * *'
+  extraArgs: --thread-pool-size 1
+- name: dedicated-admin-rolebindings
+  command: e2e-tests
+  concurrencyPolicy: Replace
+  restartPolicy: Never
+  resources:
+    requests:
+      memory: 300Mi
+      cpu: 100m
+    limits:
+      memory: 400Mi
+      cpu: 200m
+  cron: '0 8,20 * * *'
+  extraArgs: --thread-pool-size 1
+- name: default-project-labels
+  command: e2e-tests
+  concurrencyPolicy: Replace
+  restartPolicy: Never
+  resources:
+    requests:
+      memory: 300Mi
+      cpu: 100m
+    limits:
+      memory: 400Mi
+      cpu: 200m
+  cron: '0 8,20 * * *'
+  extraArgs: --thread-pool-size 1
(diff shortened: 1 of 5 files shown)


OliviaHY:
@rafael-azevedo  [MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/34446) created to migrate the alerts.

helm/qontract-reconcile/values-internal.yaml
```diff
@@ -452,3 +452,55 @@ cronjobs:
...
+  concurrencyPolicy: Replace
+  restartPolicy: Never
+  resources:
+    requests:
+      memory: 300Mi
+      cpu: 100m
+    limits:
+      memory: 400Mi
+      cpu: 200m
+  cron: '0 10 * * *'
```
Piojo, line 466: Is this every 10 hours? This would make some runs of the alerts have 3 data points and others 2, wouldn't it be confusing?

Piojo (commented):
LGTM. I'll let somebody with better knowledge of the Jenkins side of things stamp it, though.

helm/qontract-reconcile/values-internal.yaml
OliviaHY: This means the jobs gets scheduled every day at 10h00, it's the same as the original jenkins job: https://gitlab.cee.redhat.com/service/app-interface/-/blob/42b2523b851da1bf2384ad7ed9389feb1677d363/data/services/app-interface/cicd/ci-int/jobs.yaml#L44-50

Piojo: D'uh, I'm getting rusty here. In this case, I'm even more worried about the monitoring here: there *will* be spurious failures, and we'll get unnecessary alerts. How about running these at least 4 times per day?

rporres: Does it make sense to run them that often? Since we push mainly on EMEA/NASA time, most likely twice a day will make a difference and we could catch potential problems addressing the @Piojo concerns

Piojo:
(this extra test is because I need to check stuff in Jenkins, not because of this PR)

helm/qontract-reconcile/values-internal.yaml
OliviaHY: Twice a day sounds good. However assuming Jenkins is on UTC, running twice means runs again at 22h00 UTC, which would be 11pm CET and 5pm EST. And that's out of work hour for all EMEA and most of NASA east coast. How about 8h00 UTC( which is 9am CET, 3am EST) and then again at 20h00 UTC(which is 9pm CET and 3pm EST)?

Piojo: I don't mind when the checks run. I want the signal to be reasonably good, let's see if 2 checks per day deliver low noise.

As for the exact times, I don't really care - if it's broken over night, we can fix it in the morning. Whatever you feel more comfortable with.

rporres: I agree with the proposed times, @OliviaHY

Piojo (approved):
We seem to have consensus here. Thanks for improving e2e tests!

PR #2210, opened 2022-02-28
Force rebuild of qontract-reconcile image for sretoolbox upgrade

suzana-nesic:
Creating an empty commit to update the image. Follows this [PR work](https://github.com/app-sre/sretoolbox/pull/68).


steveteahan:
nit: naming the commit something like "Force rebuild of qontract-reconcile image for sretoolbox upgrade" might make people wonder less about why this was done, even if it is just an empty commit.

PR #2212, opened 2022-03-01
Pylint fixes

jfchevrette:
Fixing a couple of pylint issues and "un-disabling" the corresponding checks
* bad-super-call
* cell-var-from-loop
* consider-using-dict-items
* dangerous-default-value

This PR is best reviewed commit by commit

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -242,7 +242,9 @@ def process_jinja2_template(body, vars=None, env=None):
     return r
 
 
-def process_extracurlyjinja2_template(body, vars={}):
+def process_extracurlyjinja2_template(body, vars=None):
+    if vars is None:
+        vars = {}
     env = {
         'block_start_string': '{{%',
         'block_end_string': '%}}',
@@ -700,10 +702,12 @@ def canonicalize_namespaces(
 
 @defer
 def run(dry_run, thread_pool_size=10, internal=None,
-        use_jump_host=True, providers=[],
+        use_jump_host=True, providers=None,
         cluster_name=None, namespace_name=None,
         init_api_resources=False,
         defer=None):
+    if providers is None:
+        providers = []
     gqlapi = gql.get_api()
     namespaces = [namespace_info for namespace_info
                   in gqlapi.query(NAMESPACES_QUERY)['namespaces']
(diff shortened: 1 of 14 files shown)


reconcile/openshift_resources_base.py
```diff
@@ -242,7 +242,9 @@ def process_jinja2_template(body, vars=None, env=None):
     return r
 
 
-def process_extracurlyjinja2_template(body, vars={}):
+def process_extracurlyjinja2_template(body, vars=None):
+    if not vars:
```
steveteahan, line 246: This may be a nit for these specific use cases, but I generally err on the side of checking for `None` directly with `if vars is None` because we care about whether it was set or not more than what it is a truthy value or not?

reconcile/utils/jjb_client.py
```diff
@@ -383,8 +383,10 @@ def get_repo_url(job):
         repo_url_raw = job['properties'][0]['github']['url']
         return repo_url_raw.strip('/').replace('.git', '')
 
-    def get_all_jobs(self, job_types=[''], instance_name=None,
+    def get_all_jobs(self, job_types=None, instance_name=None,
                      include_test=False):
+        if not job_types:
+            job_types = []
```
steveteahan, line 389: Just pointing out that this was previously a list with an empty string, but it looks like that probably doesn't matter?

steveteahan (approved):
Really like the idea of starting to enable many of the Pylint checks that were disabled. In this case it's great that you could fix them, but I think that event explicit ignores would be better than the blanket disabling that we're doing.

My only other concern here is that we are relying on the behavior of the [mutable default args](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). I'll assume that the combination of unit tests, monitoring this in stage, and promoting this to look for any changes to resources would cover this?

reconcile/openshift_resources_base.py
jfchevrette: I believe you are right it should be best to check if the value is None. I'll make the change across all occurences

reconcile/utils/jjb_client.py
jfchevrette: Good catch and indeed I did look at the code and in places where `job_types` is used I don't see how it would matter that the "default" list has a single empty string element

steveteahan: Figured you caught it, but changes like these can make it easy to miss one outlier. Thanks!

reconcile/openshift_resources_base.py
jfchevrette: fixed!

PR #2214, opened 2022-03-01
Re-enable too-many-public-methods check for Pylint

steveteahan:
This check is reasonable given that classes that have lots of methods
are probably trying to do too many things. As classes get bigger,
they're also generally harder to test, which is clear by looking at
the coverage of these modules.


Piojo (approved):
Makes sens.e

PR #2216, opened 2022-03-02
[aws-ami-share] introduce new integration

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4621

depends on https://github.com/app-sre/qontract-schemas/pull/91

design doc: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/34669

this integration is an idea from @BumbleFeng and will help simplify this: https://github.com/app-sre/qontract-reconcile/pull/2100#discussion_r817030105

this PR introduces a new integration that shares AMIs between AWS accounts.
by defining a `sharing` section in an AWS account file, we turn on sharing, according to the `provider`.
we start out by only supporting the `ami` provider, which will share images and copy image tags.

still TODO: tests and type hints

### DIFF
--- a/reconcile/aws_ami_share.py
+++ b/reconcile/aws_ami_share.py
@@ -21,6 +21,16 @@ def filter_accounts(accounts: list[dict[str, Any]]):
     return [a for a in accounts if a["name"] in sharing_account_names]
 
 
+def get_region(
+    share: dict[str, Any], src_account: dict[str, Any], dst_account: dict[str, Any]
+) -> str:
+    region = share.get("region") or src_account["resourcesDefaultRegion"]
+    if region not in dst_account["supportedDeploymentRegions"]:
+        raise ValueError(f"region {region} is not supported in {dst_account['name']}")
+
+    return region
+
+
 def run(dry_run):
     accounts = queries.get_aws_accounts(sharing=True)
     sharing_accounts = filter_accounts(accounts)
@@ -36,8 +46,9 @@ def run(dry_run):
                 continue
             dst_account = share["account"]
             regex = share["regex"]
-            src_amis = aws_api.get_amis_details(src_account, src_account, regex)
-            dst_amis = aws_api.get_amis_details(dst_account, src_account, regex)
+            region = get_region(share, src_account, dst_account)
+            src_amis = aws_api.get_amis_details(src_account, src_account, regex, region)
+            dst_amis = aws_api.get_amis_details(dst_account, src_account, regex, region)
 
             for src_ami in src_amis:
                 src_ami_id = src_ami["image_id"]
--- a/reconcile/aws_ami_share.py
+++ b/reconcile/aws_ami_share.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any
+from typing import Any, Iterable, Mapping
 
 from reconcile import queries
 
@@ -10,7 +10,7 @@ QONTRACT_INTEGRATION = "aws-ami-share"
 MANAGED_TAG = TagTypeDef(Key="managed_by_integration", Value=QONTRACT_INTEGRATION)
 
 
-def filter_accounts(accounts: list[dict[str, Any]]):
+def filter_accounts(accounts: Iterable[Mapping[str, Any]]) -> list[dict[str, Any]]:
     sharing_account_names = set()
     for a in accounts:
         sharing = a.get("sharing")
@@ -23,7 +23,9 @@ def filter_accounts(accounts: list[dict[str, Any]]):
 
 
 def get_region(
-    share: dict[str, Any], src_account: dict[str, Any], dst_account: dict[str, Any]
+    share: Mapping[str, Any],
+    src_account: Mapping[str, Any],
+    dst_account: Mapping[str, Any],
 ) -> str:
     region = share.get("region") or src_account["resourcesDefaultRegion"]
     if region not in dst_account["supportedDeploymentRegions"]:
--- a/reconcile/aws_ami_share.py
+++ b/reconcile/aws_ami_share.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Any
 
 from reconcile import queries
 
@@ -8,7 +9,7 @@ QONTRACT_INTEGRATION = "aws-ami-share"
 MANAGED_TAG = {"Key": "managed_by_integration", "Value": QONTRACT_INTEGRATION}
 
 
-def filter_accounts(accounts):
+def filter_accounts(accounts: list[dict[str, Any]]):
     sharing_account_names = set()
     for a in accounts:
         sharing = a.get("sharing")
--- a/reconcile/aws_ami_share.py
+++ b/reconcile/aws_ami_share.py
@@ -1,0 +1,74 @@
+import logging
+
+from reconcile import queries
+
+from reconcile.utils.aws_api import AWSApi
+
+QONTRACT_INTEGRATION = "aws-ami-share"
+MANAGED_TAG = {"Key": "managed_by_integration", "Value": QONTRACT_INTEGRATION}
+
+
+def filter_accounts(accounts):
+    sharing_account_names = set()
+    for a in accounts:
+        sharing = a.get("sharing")
+        if sharing:
+            sharing_account_names.add(a["name"])
+            for s in sharing:
+                sharing_account_names.add(s["account"]["name"])
+
+    return [a for a in accounts if a["name"] in sharing_account_names]
+
+
+def run(dry_run):
+    accounts = queries.get_aws_accounts(sharing=True)
+    sharing_accounts = filter_accounts(accounts)
+    settings = queries.get_app_interface_settings()
+    aws_api = AWSApi(1, sharing_accounts, settings=settings)
+
+    for src_account in sharing_accounts:
+        sharing = src_account.get("sharing")
+        if not sharing:
+            continue
+        for share in sharing:
+            if share["provider"] != "ami":
+                continue
+            dst_account = share["account"]
+            regex = share["regex"]
+            src_amis = aws_api.get_amis_details(src_account, src_account, regex)
+            dst_amis = aws_api.get_amis_details(dst_account, src_account, regex)
+
+            for src_ami in src_amis:
+                src_ami_id = src_ami["image_id"]
+                found_dst_amis = [d for d in dst_amis if d["image_id"] == src_ami_id]
+                if not found_dst_amis:
+                    logging.info(
+                        [
+                            "share_ami",
+                            src_account["name"],
+                            dst_account["name"],
+                            src_ami_id,
+                        ]
+                    )
+                    if not dry_run:
+                        aws_api.share_ami(src_account, dst_account, src_ami_id)
+                    # we assume an unshared ami does not have tags
+                    found_dst_amis = [{"image_id": src_ami_id, "tags": []}]
+
+                dst_ami = found_dst_amis[0]
+                dst_ami_id = dst_ami["image_id"]
+                dst_ami_tags = dst_ami["tags"]
+                if MANAGED_TAG not in dst_ami_tags:
+                    logging.info(
+                        ["tag_shared_ami", dst_account["name"], dst_ami_id, MANAGED_TAG]
+                    )
+                    if not dry_run:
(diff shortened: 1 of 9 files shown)


reconcile/utils/aws_api.py
```diff
@@ -874,6 +883,32 @@ def get_vpcs_details(self, account, tags=None, route_tables=False):
 
         return results
 
+    def get_amis_details(self, account, owner_account, regex):
```
Piojo, line 886: Can we have type hints here, please?

reconcile/aws_ami_share.py
```diff
@@ -0,0 +1,74 @@
...
+
+from reconcile import queries
+
+from reconcile.utils.aws_api import AWSApi
+
+QONTRACT_INTEGRATION = "aws-ami-share"
+MANAGED_TAG = {"Key": "managed_by_integration", "Value": QONTRACT_INTEGRATION}
+
+
+def filter_accounts(accounts):
```
Piojo, line 11: Type hints, please. It took me a while to understand why you're not returning `sharing_accont_names` in the first place.

reconcile/queries.py
```diff
@@ -309,12 +309,25 @@ def get_github_orgs():
       region
     }
     partition
+    {% if sharing %}
```
jfchevrette, line 312: I don't understand why this templating is needed, can you explain? It seems like `get_aws_accounts` could simply always return the sharing data and then in the integration since you're testing if ` sharing` is falsy (you could also check that length is > ?) the code would end up the same

Perhaps the only reason I can see is to make all the other `get_aws_accounts` graphql queries a little bit smaller and more efficient by not requesting unnecessary data?

jfchevrette (changes requested):
LGTM but marking as "changes requested" until tests and type hints are added

reconcile/utils/aws_api.py
maorfr: i'll add type hints, also wrote it in the description

reconcile/aws_ami_share.py
maorfr: will do

maorfr:
@Piojo i recently added tests for aws api: https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/test_utils_aws_api.py

reconcile/queries.py
maorfr: exactly. in the future, we may add more graphql objects under this section. this templating will make sure we do not need to update the allowed schemas list for other integrations.

Piojo:
Woohoo! I missed that one!

steveteahan:
I think I understand the problem pretty well. I see that tests/types will be added. Should we also add this new integration to `README.md` and what is the plan for user-facing docs?

geoberle (approved):
LGTM

maorfr:
@steveteahan i've added a design doc and user facing docs in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/34669. also added the command to readme.

lechuk47:
AMIS are regional resources. They are  only usable in the same region they belong as described here[1]. Maybe we should ensure that the accounts have at least 1 common `supportedDeploymentRegion`. 

* This logic could be better placed in the app-interface pr_check. 

[1]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sharingamis-explicit.html

maorfr:
good point @lechuk47! added support for defining a region.
note: if the integration fails due to it, it fails app-interface pr-check.

geoberle:
> good point @lechuk47! added support for defining a region. note: if the integration fails due to it, it fails app-interface pr-check.
so this imply that the AMI must be build in each region it is required? or copied there so it can be shared?

maorfr:
> so this imply that the AMI must be build in each region it is required? or copied there so it can be shared?

indeed. but this is out of scope here IMO, as we only share between accounts, and not replicate across regions.

geoberle:
> indeed. but this is out of scope here IMO, as we only share between accounts, and not replicate across regions.

sounds ok for the time being. if the requirement of providing cross region would come up, we can deal with that since the region in the schema change is optional.

reconcile/utils/aws_api.py
```diff
@@ -889,7 +889,7 @@ def get_amis_details(self,
                          account: dict[str, Any],
                          owner_account: dict[str, Any],
                          regex: str,
-                         region: Optional[str] = None) -> List[Dict[str, Sequence[Collection[str]]]]:
+                         region: Optional[str] = None) -> List[Dict[str, Any]]:
```
geoberle, line 892: Any is a valid choice here in my opinion. the `item` constructed below has mixed type contents so automatic inference is hard. you could theoretically introduce a dataclass to represent the item but that would be a bit overengineered.

reconcile/aws_ami_share.py
```diff
@@ -0,0 +1,86 @@
...
+
+from reconcile import queries
+
+from reconcile.utils.aws_api import AWSApi
+
+QONTRACT_INTEGRATION = "aws-ami-share"
+MANAGED_TAG = {"Key": "managed_by_integration", "Value": QONTRACT_INTEGRATION}
+
+
+def filter_accounts(accounts: list[dict[str, Any]]):
```
fishi0x01, line 12: According to style guide, abstract collection types should be used in favor of concrete collection types for method/function arguments.

```
def filter_accounts(accounts: Iterable[Mapping[str, Any]]) ->  list[dict[str, Any]]:
```

reconcile/aws_ami_share.py
```diff
@@ -0,0 +1,86 @@
...
+        if sharing:
+            sharing_account_names.add(a["name"])
+            for s in sharing:
+                sharing_account_names.add(s["account"]["name"])
+
+    return [a for a in accounts if a["name"] in sharing_account_names]
+
+
+def get_region(
+    share: dict[str, Any], src_account: dict[str, Any], dst_account: dict[str, Any]
```
fishi0x01, line 26:
```
def get_region(
    share: Mapping[str, Any], src_account: Mapping[str, Any], dst_account: Mapping[str, Any]
) -> str
```

reconcile/aws_ami_share.py
```diff
@@ -0,0 +1,86 @@
...
+                dst_ami = found_dst_amis[0]
+                dst_ami_id = dst_ami["image_id"]
+                dst_ami_tags = dst_ami["tags"]
+                if MANAGED_TAG not in dst_ami_tags:
+                    logging.info(
+                        ["tag_shared_ami", dst_account["name"], dst_ami_id, MANAGED_TAG]
+                    )
+                    if not dry_run:
+                        aws_api.create_tag(dst_account, dst_ami_id, MANAGED_TAG)
+                src_ami_tags = src_ami["tags"]
```
BumbleFeng, line 80:
```suggestion
                src_ami_tags: Sequence[TagTypeDef] = src_ami["tags"]
```

reconcile/aws_ami_share.py
```diff
@@ -0,0 +1,87 @@
+import logging
+from typing import Any
+
+from reconcile import queries
+
+from reconcile.utils.aws_api import AWSApi
+from mypy_boto3_ec2.type_defs import TagTypeDef
+
+QONTRACT_INTEGRATION = "aws-ami-share"
+MANAGED_TAG = TagTypeDef(Key="managed_by_integration", Value=QONTRACT_INTEGRATION)
```
BumbleFeng, line 10:
```suggestion
if TYPE_CHECKING:
    from mypy_boto3_ec2.type_defs import TagTypeDef
else:
    TagTypeDef = object

MANAGED_TAG: TagTypeDef = {"Key": "managed_by_integration", "Value": QONTRACT_INTEGRATION}
```

reconcile/aws_ami_share.py
maorfr: thanks!

reconcile/aws_ami_share.py
maorfr: thanks! https://github.com/app-sre/qontract-reconcile/pull/2216/commits/eae668d0c98f58cb0d5fe532616cd0c0a8bf8f48

reconcile/aws_ami_share.py
maorfr: went in a different direction. thanks!

reconcile/aws_ami_share.py
maorfr: went in a different direction. thanks!

reconcile/aws_ami_share.py
maorfr: updated in https://github.com/app-sre/qontract-reconcile/pull/2216/commits/e406c89616e1fa6f666b74e75f3815c0524c1b08

reconcile/aws_ami_share.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def run(dry_run):
+    accounts = queries.get_aws_accounts(sharing=True)
+    sharing_accounts = filter_accounts(accounts)
+    settings = queries.get_app_interface_settings()
+    aws_api = AWSApi(1, sharing_accounts, settings=settings)
+
+    for src_account in sharing_accounts:
+        sharing = src_account.get("sharing")
```
fishi0x01, line 43: nit: sth I learned from you some weeks ago

```
sharing = src_account.get("sharing") or []
```

jfchevrette (approved):

reconcile/aws_ami_share.py
maorfr: indeed. but i have learned in the past that it's more efficient to check `if sharing` and not to instantiate an empty list.

it's also more suitable here IMO because it allows us to add more code after the following loop without "missing" that there was no `sharing` section.

reconcile/utils/aws_api.py
```diff
@@ -874,6 +893,47 @@ def get_vpcs_details(self, account, tags=None, route_tables=False):
...
+                  account: dict[str, Any],
+                  share_account_uid: str,
+                  image_id: str,
+                  region: Optional[str] = None):
+        ec2 = self._account_ec2_resource(account['name'], region)
+        image = ec2.Image(image_id)
+        launch_permission: LaunchPermissionModificationsTypeDef = {'Add': [{'UserId': share_account_uid}]}
+        image.modify_attribute(LaunchPermission=launch_permission)
+
+    def create_tag(self,
```
fishi0x01, line 929:
```
def create_tag(self, account: Mapping[str, Any], resource_id: str, tag: Mapping[str, str]):
```

reconcile/utils/aws_api.py
```diff
@@ -874,6 +893,47 @@ def get_vpcs_details(self, account, tags=None, route_tables=False):
...
+                         account: dict[str, Any],
+                         owner_account: dict[str, Any],
+                         regex: str,
+                         region: Optional[str] = None) -> List[Dict[str, Any]]:
+        ec2 = self._account_ec2_client(account['name'], region_name=region)
+        images = self.get_account_amis(ec2, owner=owner_account['uid'])
+        return self._filter_amis(images, regex)
+
+    def share_ami(self,
+                  account: dict[str, Any],
```
fishi0x01, line 920:
```
account: Mapping[str, Any]
```

Piojo (approved):
Minor nits, but it LVGTM.

reconcile/utils/aws_api.py
maorfr: thanks! https://github.com/app-sre/qontract-reconcile/pull/2216/commits/537ebe5c23a1c57fc7f9911237a801a7a840a926

reconcile/utils/aws_api.py
maorfr: thanks! https://github.com/app-sre/qontract-reconcile/pull/2216/commits/537ebe5c23a1c57fc7f9911237a801a7a840a926

reconcile/utils/aws_api.py
```diff
@@ -874,6 +893,47 @@ def get_vpcs_details(self, account, tags=None, route_tables=False):
 
         return results
 
+    @staticmethod
+    def _filter_amis(images: List[ImageTypeDef], regex: str) -> List[Dict[str, Any]]:
```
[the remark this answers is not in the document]
maorfr: thanks! https://github.com/app-sre/qontract-reconcile/pull/2216/commits/537ebe5c23a1c57fc7f9911237a801a7a840a926

PR #2218, opened 2022-03-04
fix loop on None

patjlm:
will fix https://ci.int.devshift.net/job/service-app-interface-gl-pr-check/117473/app-interface_20JSON_20validation/


geoberle (approved):
LGTM

PR #2219, opened 2022-03-08
fix ami tag type error

BumbleFeng:
https://coreos.slack.com/archives/CS0E65QCV/p1646713787105329

```
[aws-ami-share]     tag_type_def = TagTypeDef(Key=tag['Key'], Value=tag['Value'])
[aws-ami-share] TypeError: object() takes no arguments
```


maorfr (approved):
thank you!

PR #2224, opened 2022-03-09
[saas-file-validator] fail validation when IMAGE_TAG matches ref

maorfr:
as the IMAGE_TAG parameter is automatically generated, we should enforce cleaner configuration.
if IMAGE_TAG is the same as the ref, IMAGE_TAG is not required.

https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/continuous-delivery-in-app-interface.md#automatically-generated-parameters

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -344,7 +344,6 @@ class SaasHerder():
             )
             self.valid = False
 
-
     def _collect_namespaces(self):
         # namespaces may appear more then once in the result
         namespaces = []
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -162,6 +162,12 @@ class SaasHerder():
                     if not target_parameters:
                         continue
                     target_parameters = json.loads(target_parameters)
+                    self._validate_image_tag_not_equals_ref(
+                        saas_file_name,
+                        resource_template_name,
+                        target['ref'],
+                        target_parameters
+                    )
                     environment_parameters = environment['parameters']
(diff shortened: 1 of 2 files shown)


reconcile/utils/saasherder.py
```diff
@@ -322,6 +328,22 @@ def _validate_upstream_not_used_with_commit_sha(
...
 
+    def _validate_image_tag_not_equals_ref(
+            self,
+            saas_file_name: str,
+            resource_template_name: str,
+            ref: str,
+            parameters: dict,
+    ):
+        image_tag = parameters.get('IMAGE_TAG')
+        if image_tag and str(ref).startswith(str(image_tag)):
```
fishi0x01, line 339: Is a prefix match accurate here? Judging from function and PR description I'd assume to check for an exact match rather than a prefix, e.g.,

```
ref.strip() == image_tag.strip()
```

fishi0x01: Ah I see the use-case now in the test. All good.

maorfr: the IMAGE_TAG is the first <app-interface settings.hashLength>, so a prefix is what we need.

rporres (approved):

geoberle (approved):
LGTM

nit if you want to go for it but feel free to ignore: you could check with the correct prefix length (global one or saas local one)

maorfr:
do we really care about the hash length? we are not checking that the image exists, but only that IMAGE_TAG is the same as ref?

PR #2225, opened 2022-03-09
Add --no-github flag to openshift -rolebindings and -clusterrolebindings

TGPSKI:
This MR adds support for general IDP providers outside of github when associating users with rolebindings and clusterrolebindings.

The option `--no-github` has been added to both click commands for the integrations. The graphql query has been updated for both users and bots, and `fetch_desired_state` branches to either github_username or org_username depending on the flag.

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -553,8 +553,8 @@ def github_validator(ctx):
 )
 @click.option('--no-github',
               is_flag=True,
-              help='use org_username instead of github_username for ' + 
-                    'user <> clusterrolebinding associations')
+              help='use org_username instead of github_username for ' +
+              'user <> clusterrolebinding associations')
 @threaded()
 @binary(['oc', 'ssh'])
 @binary_version('oc', ['version', '--client'], OC_VERSION_REGEX, OC_VERSION)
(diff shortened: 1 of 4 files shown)


maorfr:
how about we generalize this a bit?
perhaps instead of `--no-github` we can use something like `--auth-method` (github / sso?) or `--user-key` (org_username / github_username)?

rporres:
I like Maor's idea. I prefer the `--auth-method` as it is more descriptive and hides implementation details, but I can live with both Maor's proposals.

steveteahan:
Should this be configurable per cluster? We already have `auth` in the `/openshift/cluster-1.yml` schema? If it's just simpler to implement `--auth-method` for now, that makes sense to me if there isn't any known use case where we need to support >1 auth method in a given environment.

reconcile/cli.py
```diff
@@ -551,31 +551,41 @@ def github_validator(ctx):
...
+              'user <> clusterrolebinding associations')
 @threaded()
 @binary(['oc', 'ssh'])
 @binary_version('oc', ['version', '--client'], OC_VERSION_REGEX, OC_VERSION)
 @internal()
 @use_jump_host()
 @click.pass_context
-def openshift_clusterrolebindings(ctx, thread_pool_size, internal,
-                                  use_jump_host):
+def openshift_clusterrolebindings(ctx, no_github, thread_pool_size,
```
steveteahan, line 564: Just a general note that for any new parameters please add types. For anything existing, it doesn't hurt to if you've already figured out what they are. I'll just make this comment once rather than for each instance.

maorfr:
i completely agree with @steveteahan!
the `auth` section of a cluster has a `service` field, which can describe what auth method to use.

reconcile/openshift_rolebindings.py
```diff
@@ -99,11 +101,22 @@ def fetch_desired_state(ri, oc_map):
...
 
-        users = [user['github_username']
-                 for user in role['users']]
-        bot_users = [bot['github_username']
-                     for bot in role['bots']
-                     if bot.get('github_username')]
+        users = []
+        bot_users = []
+
+        if no_github is True:
```
Piojo, line 107: A fair bit of duplicated code here. Can we refactor it?

PR #2230, opened 2022-03-10
[aws-iam-keys] filter accounts without delete keys

maorfr:
related to https://issues.redhat.com/browse/APPSRE-4621

if there is nothing to do, why not filter it.


geoberle (approved):
LGTM

PR #2231, opened 2022-03-10
[aws-support-cases-sos] filter accounts without premium support

maorfr:
related to https://issues.redhat.com/browse/APPSRE-4621

if premium support is not enabled, the integration can't do it's work. it's worthless to try :)


geoberle (approved):
LGTM

PR #2235, opened 2022-03-11
More help for sre checkpoint metadata

steveteahan:
Related to this Slack thread: https://coreos.slack.com/archives/GGC2A0MS8/p1647030227605789

I also was unlucky and lost the CI dependencies changing game, so two minor fixes there to unblock CI. They were minor enough that it didn't make much sense to create a separate PR.


TGPSKI:
LGTM

PR #2236, opened 2022-03-13
[saas-file-validator] validate upstream jobs

maorfr:
if a job that does not exist is defined in an `upstream` section, the deployment will never be triggered.

### DIFF
--- a/reconcile/utils/jjb_client.py
+++ b/reconcile/utils/jjb_client.py
@@ -383,6 +383,10 @@ class JJB:  # pylint: disable=too-many-public-methods
         repo_url_raw = job['properties'][0]['github']['url']
         return repo_url_raw.strip('/').replace('.git', '')
 
+    @staticmethod
+    def get_ref(job):
+        return job['scm'][0]['git']['branches'][0]
+
     def get_all_jobs(self, job_types=None, instance_name=None,
                      include_test=False):
         if job_types is None:
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -6,7 +6,7 @@ import itertools
 import hashlib
 import re
 from collections import ChainMap
-from typing import Mapping, Any, MutableMapping, Tuple
+from typing import Mapping, Any, MutableMapping, Tuple, Union
 
 from contextlib import suppress
 import yaml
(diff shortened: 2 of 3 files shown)


maorfr:
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/35247

reconcile/utils/jjb_client.py
```diff
@@ -383,6 +383,10 @@ def get_repo_url(job):
         repo_url_raw = job['properties'][0]['github']['url']
         return repo_url_raw.strip('/').replace('.git', '')
 
+    @staticmethod
+    def get_ref(job: dict) -> str:
```
geoberle, line 387: i'm going to adopt this style where i will use dict without further key/value type hints when we talk about a generic dict.

reconcile/utils/saasherder.py
```diff
@@ -344,6 +364,62 @@ def _validate_image_tag_not_equals_ref(
...
+            upstream = t.get("upstream")
+            if upstream:
+                if isinstance(upstream, str):
+                    # skip v1 saas files
+                    continue
+                upstream_job = UpstreamJob(upstream['instance']['name'], upstream['name'])
+                possible_upstream_jobs = self._get_upstream_jobs(jjb, all_jobs, url, ref)
+                found_jobs = [j for j in all_jobs[upstream_job.instance] if j["name"] == upstream_job.job]
+                if found_jobs:
+                    if upstream_job not in possible_upstream_jobs:
```
geoberle, line 408: to verify i understand this condition: this is true if the upstream job exists by name but the url and/or ref do not match - so basically a configuration error.

reconcile/utils/saasherder.py
```diff
@@ -344,6 +364,62 @@ def _validate_image_tag_not_equals_ref(
...
+            jjb: JJB,
+    ):
+        all_jobs = jjb.get_all_jobs(job_types=["build"])
+        pattern = r'^[0-9a-f]{40}$'
+        for sf, rt, t in self:
+            sf_name = sf['name']
+            rt_name = rt['name']
+            url = rt['url']
+            ref = t['ref']
+            if re.search(pattern, ref):
```
geoberle, line 397: is the purpose of this to exclude build jobs with repo refs that are SHAs? that implies we have such configs. what is the story about this?

geoberle (approved):
LGTM - my comments are only questions to deepen my learning about upstream jobs

reconcile/utils/saasherder.py
maorfr: correct

reconcile/utils/saasherder.py
maorfr: we don't have such configs (#2136), and this is why we "filter" these out.

geoberle: oh sry i read that wrong. ref is from the target and not from the job. that makes sense in regards to #2136

PR #2237, opened 2022-03-14
[terraform-resources] ASG add support for upstream job

maorfr:

depends on https://github.com/app-sre/qontract-schemas/pull/95

now that we have support for continuous deployment of AMIs, we want to avoid failing under healthy conditions.
for example, once a new commit is pushed, the integration tries to use a corresponding AMI. initially, that AMI is being built, which means it doesn't exist yet.

to be able to not fail in this condition, and to improve visibility from the ASG definition, this PR adds support for an upstream job to wait for (like saas files).

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -282,6 +282,17 @@ provider
     tag_name
     url
     ref
+    upstream {
+      instance {
+        token {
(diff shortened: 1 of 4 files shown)


geoberle (approved):
LGTM

PR #2238, opened 2022-03-14
APPSRE-4029 simultaneous cluster upgrades

patjlm:
depends on https://github.com/app-sre/qontract-schemas/pull/96

### DIFF
--- a/reconcile/ocm_upgrade_scheduler.py
+++ b/reconcile/ocm_upgrade_scheduler.py
@@ -167,6 +167,11 @@ def version_conditions_met(version, history, ocm_name, workloads, upgrade_condit
     return conditions_met
 
 
+def cluster_mutexes(policy: dict) -> list[str]:
+    """List all mutex locks for the given cluster"""
+    return (policy.get("conditions") or {}).get("mutexes") or []
+
+
 def calculate_diff(current_state, desired_state, ocm_map, version_history):
     """Check available upgrades for each cluster in the desired state
     according to upgrade conditions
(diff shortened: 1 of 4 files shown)


patjlm:
https://issues.redhat.com/browse/APPSRE-4029

maorfr:
i have a couple of recommendations:
1. `ClusterUpgradeLockMap` - perhaps this can be generalized a bit? this is essentially a lock map using the State. maybe generalize the logic within it a bit and extract clusters specific information to the integration? i'd even go as far as implementing this `StateLockMap` in sretoolbox?
2. to avoid deadlocks i'd recommend we use `defer` for each unlock (immediately after locking).

maorfr:
> Am I missing something

no, i was.

patjlm:
@geoberle raised a very valid point that we may be able to do without a state to hold the lock map and rely exclusively on 
- upgradepolicies currently active in OCM
- the same upcoming locks mechanism implemented here (`new_locks`) to avoid 2 triggers in the same loop
This would simplify the overall implementation. I'll see what can be done.

Setting this as WIP in the meanwhile

patjlm:
Following a discussion with @geoberle, I removed the need for a state-based lock map. The lock is now based on the simple presence of an `upgradePolicy` in OCM for each clusters. It should be much easier/leaner/cleaner this way.

geoberle (approved):
LVGTM

patjlm:
renamed `locks` -> `mutexes` after discussion in https://github.com/app-sre/qontract-schemas/pull/96

rporres (approved):
Looks great. My only concern is allowing a free text for the mutexes and getting typo ruin the whole thing. As commented in https://github.com/app-sre/qontract-schemas/pull/96#pullrequestreview-908948880

You may want to create a ticket a solve it in subsequent PR

patjlm:
> You may want to create a ticket a solve it in subsequent PR

https://issues.redhat.com/browse/APPSRE-4709

PR #2239, opened 2022-03-15
APPSRE-4404 move data from toml to app-interface

fishi0x01:
**Motivation**

We currently store configuration for the github-repo-invites integration in our config.toml. This PR makes `github-repo-invites` read the configuration from `app-interface`, making the `[github-repo-invites]` section redundant.

Further, this PR refactors the `github-repo-invites` integration and adds unit tests for it.

~This PR keeps backwards compatibility by using `config.toml` if gql does not return required data. Once all app-interfaces are migrated, we can remove this backwards compatibility option.~ Backwards compatibility is not required, as there is only one app-interface using this integration.

**Test**

- Successfully ran with `--dry-run` for corresponding schema https://github.com/app-sre/qontract-schemas/pull/97
- added unit tests to cover the important logic

### DIFF
--- a/reconcile/github_repo_invites.py
+++ b/reconcile/github_repo_invites.py
@@ -1,6 +1,9 @@
 import logging
 import os
 
+from dataclasses import dataclass
+from typing import Any, Iterable, Mapping, Optional
+
 from reconcile.utils import gql
 from reconcile.utils import raw_github_api
 from reconcile.utils.secret_reader import SecretReader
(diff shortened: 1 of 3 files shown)


reconcile/github_repo_invites.py
```diff
@@ -57,11 +69,35 @@ def run(dry_run):
...
+
+def run(dry_run):
+    gqlapi = gql.get_api()
+    result = gqlapi.query(REPOS_QUERY)
+    settings = queries.get_app_interface_settings()
+    secret_reader = SecretReader(settings=settings)
+
+    # TODO kfischer: This is placed for backwards compatibility.
+    # Once all app-interfaces added the new 'githubRepoInvites' section,
+    # then we can remove this.
```
maorfr, line 89: feel free to remove this, this integration is only used in a single app-interface instance (gitlab.cee)

fishi0x01: Thx for the info! Removed the backwards compat block :+1:

reconcile/test/test_github_repo_invites.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+    assert github_repo_invites._parse_code_components(raw_code_components) == expected
+
+
+def test_parse_valid_code_components():
+    raw_code_components: Iterable[Mapping[str, Any]] = [
+        {
+            "codeComponents": [
+                {
+                    "url": "org1/project1",
```
rporres, line 24: why not using a url as usually found in app-interface here? e.g.

```suggestion
                    "url": "https://github.com/org1/project1",
```

rporres (commented):
Looks great. Thx for the refactor which adds the possibility of writing those nice tests. Just a nit comment in the test data for urls that could be extended to all the tests written as they follow the same pattern

reconcile/test/test_github_repo_invites.py
fishi0x01: Absolutely! Good point!

Added more realistic urls as test input.

rporres (approved):

PR #2240, opened 2022-03-15
[terraform-aws-route53] add support for target namespace zone

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4705

depends on https://github.com/app-sre/qontract-schemas/pull/98

similar to using `_target_cluster` to obtain the elb fqdn, we can target a namespace with a route53-zone provisioned in it to automatically obtain NS records of that zone (instead of hard coding them): `target_namespace_zone`.

example: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/35369

### DIFF
--- a/reconcile/terraform_aws_route53.py
+++ b/reconcile/terraform_aws_route53.py
@@ -19,10 +19,8 @@ QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
 
 
 def build_desired_state(
-        zones: list[dict],
-        all_accounts: list[dict],
-        settings: dict
-    ) -> list[dict]:
+    zones: list[dict], all_accounts: list[dict], settings: dict
+) -> list[dict]:
     """
     Build the desired state from the app-interface resources
 
--- a/reconcile/terraform_aws_route53.py
+++ b/reconcile/terraform_aws_route53.py
@@ -7,6 +7,7 @@ from reconcile import queries
 
 from reconcile.status import ExitCodes
 from reconcile.utils import dnsutils
+from reconcile.utils.aws_api import AWSApi
 from reconcile.utils.semver_helper import make_semver
 from reconcile.utils.defer import defer
 from reconcile.utils.terrascript_client import TerrascriptClient as Terrascript
@@ -17,7 +18,7 @@ QONTRACT_INTEGRATION = "terraform_aws_route53"
 QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
 
 
-def build_desired_state(zones):
+def build_desired_state(zones, all_accounts, settings):
     """
     Build the desired state from the app-interface resources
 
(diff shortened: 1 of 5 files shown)


reconcile/terraform_aws_route53.py
```diff
@@ -17,7 +18,9 @@
 QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
 
 
-def build_desired_state(zones):
+def build_desired_state(
+    zones: list[dict], all_accounts: list[dict], settings: dict
+) -> list[dict]:
```
fishi0x01, line 23: According to dev guidelines:

```
def build_desired_state(
    zones: Iterable[Mapping], all_accounts: Iterable[Mapping], settings: Mapping
) -> list[dict]:
```

reconcile/terraform_aws_route53.py
```diff
@@ -116,6 +119,50 @@ def build_desired_state(zones):
...
+                    for tfr in tf_resources
+                    if tfr["provider"] == "route53-zone" and tfr["name"] == tf_zone_name
+                ]
+                if not tf_zone_resources:
+                    logging.error(
+                        f"{zone_name}: field `_target_namespace_zone` found "
+                        f"for record {record_name}, but target zone not found: "
+                        f"{tf_zone_name}"
+                    )
+                    sys.exit(ExitCodes.ERROR)
```
fishi0x01, line 138: Not sure, but would it make sense to raise an exception here and handle this in the caller?

fishi0x01: NVM. I see the same style is used in the beginning of the function.

reconcile/terraform_aws_route53.py
```diff
@@ -116,6 +119,50 @@ def build_desired_state(zones):
...
+                tf_zone_resource = tf_zone_resources[0]
+                tf_zone_account_name = tf_zone_resource["account"]
+                zone_account = [
+                    a for a in all_accounts if a["name"] == tf_zone_account_name
+                ][0]
+                awsapi = AWSApi(1, [zone_account], settings=settings, init_users=False)
```
fishi0x01, line 144: Not sure if its feasible, but this function is quite humongous. Would it make sense to separate this parsing logic into its own function? That might also help in the unit testing later.

```
@dataclass
class ZoneAccount:
    aws_api: AWSApi
    name: str
    account_name: str
    region: str
    default_region: str

def _parse_zones(raw_target_namespace_zone: Mapping[str, Any]) -> ZoneAccount:
    pass
```

reconcile/terraform_aws_route53.py
```diff
@@ -116,6 +119,50 @@ def build_desired_state(zones):
...
+                        f"for record {record_name}, but target zone not found (yet): "
+                        f"{tf_zone_name}"
+                    )
+                    continue
+                logging.debug(
+                    f"{zone_name}: field `_target_namespace_zone` found "
+                    f"for record {record_name}, Values are: "
+                    f"{tf_zone_ns_records}"
+                )
+                record["records"] = tf_zone_ns_records
```
fishi0x01, line 164: Similarly, the output of the above extracted function could then be used to query and aggregate the real-world zone records.

```
def _query_zone_info(zone_account: ZoneAccount) -> list[tf_zone_ns_records?]:
    pass
```

requirements/requirements-type.txt
```diff
@@ -9,4 +9,4 @@ types-PyYAML
 types-requests
 types-tabulate
 types-toml
-boto3-stubs[ec2,s3,rds,iam]
+boto3-stubs[ec2,s3,rds,iam,route53]
```
fishi0x01, line 12: :heart:

reconcile/utils/aws_api.py
```diff
@@ -1155,6 +1167,38 @@ def get_tgws_details(self, account, region_name, routes_cidr_block,
...
+        return route53.list_resource_record_sets(HostedZoneId=zone_id)["ResourceRecordSets"]
+
+    @staticmethod
+    def _filter_record_sets(record_sets: List[ResourceRecordSetTypeDef],
+                            zone_name: str,
+                            zone_type: str) -> List[ResourceRecordSetTypeDef]:
+        return [r for r in record_sets if r["Name"] == f"{zone_name}." and r["Type"] == zone_type]
+
+    @staticmethod
+    def _extract_records(resource_records: List[ResourceRecordTypeDef]) -> list[str]:
```
geoberle, line 1191: i hate to nitpick on style. would you consider using `list` instead of `List` 😘

reconcile/terraform_aws_route53.py
maorfr: since this is relevant to the entire integration, would it make sense to keep the same style for this PR and create a ticket to refactor?

reconcile/utils/aws_api.py
```diff
@@ -1155,6 +1167,38 @@ def get_tgws_details(self, account, region_name, routes_cidr_block,
...
+        # [{'Value': 'ns.example.com.'}, ...]
+        return [r['Value'].rstrip(".") for r in resource_records]
+
+    def get_route53_zone_ns_records(self, account_name, zone_name, region):
+        route53 = self._account_route53_client(account_name, region)
+        record_sets = self._get_hosted_zone_record_sets(route53, zone_name)
+        resource_records = self._filter_record_sets(record_sets, zone_name, "NS")[0]["ResourceRecords"]
```
geoberle, line 1198: leaving this comment for other reviewers: i was first puzzled about picking just the first NS record but then i learned what a record set is. _filter_record_sets returns a "set of records for a zone and a record type", in this case it returns all NS records for the zone at hand. so this is absolutely fine and i started my learning path for route 53

geoberle (approved):

reconcile/terraform_aws_route53.py
fishi0x01: The only concern here is that in my experience dedicated refactor tickets/PRs never come. Usually stuff then just grows over time to a point where any refactor will be a huge amount of work (looking at terrascript_client.py for instance :eyes: ). Personally, I feel like every feature PR's goal should also be to (slightly) improve the code it affects (if possible and reasonable in given time). That's why Im always in favor of making small enhancements as you touch a portion of code - and annoying people with it :sweat_smile: E.g., in this case splitting into functions would also help to accommodate for some tests.

That being said, thank you for bearing with me :) I know I can be annoying with this stuff and I might as well be too idealistic about this. This effort might not be worth it now. It would be nice though to really have a ticket in the backlog then to take a look at this refactoring situation. I am also happy to pick it up :)

reconcile/utils/aws_api.py
geoberle: good point. then please change the return typehint to `List[str]`

reconcile/terraform_aws_route53.py
maorfr: i've created https://issues.redhat.com/browse/APPSRE-4720

i'm also happy to pick it up!

PR #2244, opened 2022-03-16
openshift support for Secret stringData

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4714

this PR adds support for Secret stringData, which will not cause endless reconciliation.


geoberle (approved):
😘

PR #2245, opened 2022-03-16
canonicalize ACM ManagedCluster CR

geoberle:
the ACM ManagedCluster CR has lots fields/labels/annotations that
are applied and removed dynamically by ACM controllers. removing
them in the canonicalization step (and thus removing them from the
comparision logic to detect drift) allows the a-i management of those
CRs.

part of https://issues.redhat.com/browse/APPSRE-4619

### DIFF
--- a/reconcile/test/test_openshift_resource.py
+++ b/reconcile/test/test_openshift_resource.py
@@ -156,3 +156,55 @@ def test_secret_string_data():
     }
     result = OR.canonicalize(resource)
     assert result == expected
+
+
+def test_managed_cluster_drift():
+    d1 = {
+        "apiVersion": "cluster.open-cluster-management.io/v1",
+        "kind": "ManagedCluster",
+        "metadata": {
+            "labels": {
+                "cluster.open-cluster-management.io/clusterset": "default",
+            },
+            "name": "xxx"
+        },
+        "spec": {
+            "hubAcceptsClient": True
+        }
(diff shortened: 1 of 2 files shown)


maorfr:
why isn't this caught by https://github.com/app-sre/qontract-reconcile/blob/364e6d60f2d8dd21405d7ac770338fa4b4b7bacc/reconcile/utils/openshift_resource.py#L47-L48?

geoberle:
> why isn't this caught by
> 
> https://github.com/app-sre/qontract-reconcile/blob/364e6d60f2d8dd21405d7ac770338fa4b4b7bacc/reconcile/utils/openshift_resource.py#L47-L48

good question. i will investigate

geoberle:
> > why isn't this caught by
> > https://github.com/app-sre/qontract-reconcile/blob/364e6d60f2d8dd21405d7ac770338fa4b4b7bacc/reconcile/utils/openshift_resource.py#L47-L48

it is because of the labels. those are changed as well by ACM, e.g. a dynamic clusterId ID label is added after the `ManagedCluster` has been processed

https://github.com/app-sre/qontract-reconcile/blob/364e6d60f2d8dd21405d7ac770338fa4b4b7bacc/reconcile/utils/openshift_resource.py#L66

the current state of this PR handles only those labels

maorfr (approved):

PR #2252, opened 2022-03-17
controller managed labels per kind

geoberle:
cluster controllers that are labeling resources break our drift detection
for openshift_resources. this MR introduces a way to declare
which labels on which kind are supposed to be ignored because
their existance or values are managed by a cluster controller.

usually additional fields in the current state of a resource are ignored by
the `OpenshiftResource``eq` implementation, but labels drift is checked
specifically.

e.g. ACM (Advanced Cluster Manager) controllers put dynamic labels on
the `ManagedCluster` CRs, e.g. the generated ID of a cluster or the
detected Openshift `version` running on the cluster.

### DIFF
--- a/reconcile/utils/openshift_resource.py
+++ b/reconcile/utils/openshift_resource.py
@@ -4,6 +4,7 @@ import datetime
 import hashlib
 import json
 import re
+from typing import Union
 
 from threading import Lock
 
@@ -31,6 +32,17 @@ DNS_NAMES_URL = (
 )
 
 IGNORABLE_DATA_FIELDS = ["service-ca.crt"]
+# these labels existance and/or value is determined by a controller running
+# on the cluster. we need to ignore their existance in the current state,
+# otherwise we will deal with constant reconciliation
+CONTROLLER_MANAGED_LABELS: dict[str, set[Union[str, re.Pattern]]] = {
+    "ManagedCluster": {
+        "clusterID",
+        "managed-by",
+        "openshiftVersion",
+        re.compile("feature.open-cluster-management.io/.*"),
+    }
+}
 
 
 class OpenshiftResource:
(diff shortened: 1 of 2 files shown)


reconcile/utils/openshift_resource.py
```diff
@@ -63,7 +75,18 @@ def obj_intersect_equal(self, obj1, obj2):
                     pass
                 elif self.ignorable_key_value_pair(obj1_k, obj1_v):
                     pass
-                elif obj1_k in ["data", "labels", "matchLabels"]:
+                elif obj1_k == "labels":
```
patjlm, line 78: is there any risk we have labels out of metadata ?

reconcile/utils/openshift_resource.py
```diff
@@ -326,6 +358,12 @@ def canonicalize(body):
         # remove status
         body.pop("status", None)
 
+        # remove controller managed labels
+        labels = body["metadata"].get("labels", {})
+        for label in set(labels.keys()):
+            if OpenshiftResource.is_controller_managed_label(body["kind"], label):
+                labels.pop(label)
```
patjlm, line 365: this means the controlled label has been set in app-interface right ? Should we log a kind of warning or even fail ?

patjlm: or is that the 'current' resource ?

reconcile/utils/openshift_resource.py
patjlm: It was like this before so probably not an issue

patjlm (approved):
LGTM

reconcile/utils/openshift_resource.py
geoberle: a label that is a `is_controller_managed_label` is one that is completely managed (existence and/or value) by something other then a-i. e.g. the `clusterId` label on a `ManagedCluster`. 

those labels should not be defined in a-i. we could also fail a PR check if such a label is present.

by removing those labels from the state (mostly relevant for current state), resources will still be considered equal in the sense of `__eq__` and `sha256sum()`

reconcile/utils/openshift_resource.py
geoberle: the logic for labels is still the same with the exception of the `IGNORABLE_DATA_FIELDS` `service-ca.crt` but is usually within the data section of a secret and not within a label.

PR #2253, opened 2022-03-17
Add some type hints and docs for Terrascript/Terraform related code

steveteahan:
This isn't a huge change, but hopefully it will help reduce the time
to understand the code and catch type-related issues that could be
introduced in the future.

Note that all cases of using concrete types in parameter definitions
is probably intentional. In these cases we're probably passing the
inputs as outputs directly, which makes it impossible to use abstract
types while returning concrete types. Casting to the expected return
type could have been used, but I went out of my way to avoid changing
the code due to limited test coverage.


reconcile/terraform_resources.py
```diff
@@ -415,8 +417,8 @@ def fetch_current_state(dry_run, namespaces, thread_pool_size,
     return ri, oc_map
 
 
-def init_working_dirs(accounts, thread_pool_size,
-                      oc_map=None, settings=None):
+def init_working_dirs(accounts: list[dict[str, Any]], thread_pool_size: int,
+                      oc_map: Optional[OCMMap] = None, settings: Optional[Mapping[str, Any]] = None) -> tuple[Terrascript, dict[str, str]]:
```
Piojo, line 421: Line too long?

Piojo (approved):

reconcile/terraform_resources.py
steveteahan: Let's see if Black catches it. Everything is green on my laptop.

steveteahan: It looks like black isn't running against `reconcile/utils`, possibly because it doesn't follow subdirectories? I wonder if @fishi0x01 ran into anything like this when originally adding it? I've fixed the lines manually for now.

steveteahan: Nevermind... these are some of the files that are being excluded due to being changed in the last 30 days.

fishi0x01: You are right - we are in the process of reducing that list :)

reconcile/terraform_resources.py
```diff
@@ -415,8 +417,11 @@ def fetch_current_state(dry_run, namespaces, thread_pool_size,
     return ri, oc_map
 
 
-def init_working_dirs(accounts, thread_pool_size,
```
fishi0x01, line 418: According to [style guide ](https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#abstract-collection-types) this should be `Iterable[Mapping[str, Any]]`?

reconcile/utils/terrascript_client.py
```diff
@@ -159,8 +160,21 @@ class ElasticSearchLogGroupInfo:
...
+    JSON format from app-interface schemas/openshift/terraform-resource-1.yml objects.
+
+    Usage example (mostly to demonstrate API):
+
+    ts = TerrascriptClient("terraform_resources", "qrtf", 20, accounts, settings)
+    ts.populate_resources(tf_namespaces, existing_secrets, account_name, ocm_map=ocm_map)
+    ts.dump(print_to_file, existing_dirs=working_dirs)
+
+    More information on Terrascript: https://python-terrascript.readthedocs.io/en/develop/
+    """
```
fishi0x01, line 174: :heart:

reconcile/terraform_resources.py
steveteahan: This was done on purpose. We have some code that takes an input, say a `dict` and returns that object directly (rather than creating a new data structure). Doing this makes it difficult to use abstract types for parameters and concrete for return.

For an example, see `dump` where we're doing:

```python
    def dump(self, print_to_file=None, existing_dirs=None):
        if existing_dirs is None:
            working_dirs = {}
        else:
            working_dirs = existing_dirs

       ....

       return working_dirs
```

In this scenario, the return can either be a `dict`, or whatever we passed into `existing_dirs`. If I restrict `existing_dirs` to a `Mapping`, then the return value can either be a `Mapping` or `dict`, so I can't set the return value to `dict`. I'd really have to change how the code works to fix this, which I don't want to do given the limited coverage here. I only broke that rule twice for a couple of `cast` calls which return the data unmodified.

Anyone writing this function today with type hints would discover this issue and then probably call `dict(existing_dirs)` to cast the input to `dict`, but unless we feel strongly about this, I'd rather keep any actual changes to how the code could behave out of this PR. I think this would be best fixed in a separate PR with the addition of tests (if that is practical).

fishi0x01: Thx a lot for the detailed explanation! :bow: Makes sense now

fishi0x01 (approved):
Great enhancement! Thank you!

PR #2254, opened 2022-03-17
Skip Terraform apply when the only RDS update would be engine_version

steveteahan:
RDS engine_version changes are unique because a tenant may want to change
this value, but allow RDS to upgrade the instance during the next maintenance
window. This window could be up to 7 days away, which in the meantime would
cause Terraform to apply the change every time that the integration runs (as
well as log the change, which is more annoying than harmful). Unnecessary
apply operations will result in state changes and any instances of
terraform-resources running will need to restart (think PR checks).

The manual tests that I performed so far were:

1. Upgrade RDS `engine_version` without `apply_immediately` and confirm that the next time terraform-resources runs, it skips the Terraform `apply`
2. After doing above, set `apply_immediately: true` and run again, then RDS is upgraded immediately. This tests both that >1 change will still cause an apply, but also the special case where we decide after the fact that RDS should be upgraded immediately.

The output from these tests:

```
# Update the database without apply_immediately
$ qontract-reconcile --config config.dev.toml --log-level INFO terraform-resources --account-name app-int-example-01
[2022-03-22 12:38:08] [INFO] [gql.py:init_from_config:220] - using gql endpoint http://localhost:4000/graphqlsha/9326e854260d26c1e7c03e9c37cebe68d766b240c9d45f7f740d20c96ca930a6
[2022-03-22 12:38:33] [INFO] [terraform_client.py:log_plan_diff:249] - ['update', 'app-int-example-01', 'aws_db_instance', 'app-int-example-01-rds']

# Run again with no changes to app-interface
$ qontract-reconcile --config config.dev.toml --log-level DEBUG terraform-resources --account-name app-int-example-01 2>&1 | grep 'INFO\|Not setting should_apply'
[2022-03-22 12:40:21] [INFO] [gql.py:init_from_config:220] - using gql endpoint http://localhost:4000/graphqlsha/9326e854260d26c1e7c03e9c37cebe68d766b240c9d45f7f740d20c96ca930a6
[2022-03-22 12:40:47] [DEBUG] [terraform_client.py:log_plan_diff:243] - Not setting should_apply for app-int-example-01-rds because the only change is EngineVersion and that setting is in PendingModifiedValues

# Add apply_immediately: true and the upgrade occurs
$ qontract-reconcile --config config.dev.toml --log-level DEBUG terraform-resources --account-name app-int-example-01 2>&1 | grep 'INFO\|Not setting should_apply'
[2022-03-22 12:43:05] [INFO] [gql.py:init_from_config:220] - using gql endpoint http://localhost:4000/graphqlsha/808f987d6009ec56c55c2e01b5355275486e9d5b94b413aea32f62fe7d791b50
[2022-03-22 12:43:30] [INFO] [terraform_client.py:log_plan_diff:249] - ['update', 'app-int-example-01', 'aws_db_instance', 'app-int-example-01-rds']
```

Additionally, I ran the `terraform-*` integrations in dry-run mode with `app-sre/app-interface` data. This provides some basic reassurances. Ultimately, these changes will be tested in staging as well because the creation of `AWSApi` and passing of the object into `TerraformClient` happen in dry-run mode.

Related: [APPSRE-4645](https://issues.redhat.com/browse/APPSRE-4645)

### DIFF
--- a/reconcile/utils/terraform_client.py
+++ b/reconcile/utils/terraform_client.py
@@ -18,6 +18,7 @@ from sretoolbox.utils import threaded
 import reconcile.utils.lean_terraform_client as lean_tf
 
 from reconcile.utils import gql
+from reconcile.utils.aws_api import AWSApi
 from reconcile.utils.openshift_resource import OpenshiftResource as OR
 
 ALLOWED_TF_SHOW_FORMAT_VERSION = "0.1"
@@ -38,7 +39,7 @@ class TerraformClient:  # pylint: disable=too-many-public-methods
     def __init__(self, integration: str, integration_version: str,
                  integration_prefix: str, accounts: Iterable[Mapping[str, Any]],
                  working_dirs: Mapping[str, str], thread_pool_size: int,
-                 init_users=False):
+                 aws_api: AWSApi, init_users=False):
         self.integration = integration
         self.integration_version = integration_version
         self.integration_prefix = integration_prefix
@@ -46,6 +47,7 @@ class TerraformClient:  # pylint: disable=too-many-public-methods
         self.accounts = {a['name']: a for a in accounts}
         self.parallelism = thread_pool_size
         self.thread_pool_size = thread_pool_size
+        self._aws_api = aws_api
         self._log_lock = Lock()
         self.should_apply = False
 
@@ -237,7 +239,7 @@ class TerraformClient:  # pylint: disable=too-many-public-methods
                 # unnecessary Terraform state updates until they complete.
                 if action == 'update' and resource_type == 'aws_db_instance' and \
                         self._is_ignored_rds_modification(
-                            resource_name, resource_change, output):
+                            name, resource_name, resource_change, output):
                     logging.warning(
                         f"Not setting should_apply for {resource_name} because the "
                         f"only change is EngineVersion and that setting is in "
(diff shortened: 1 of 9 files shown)


reconcile/utils/terraform_client.py
```diff
@@ -561,3 +575,37 @@ def split_to_lines(*outputs):
...
+            # TODO: I'm open to better ways to get creds. I looked at AWSApi, but I
+            #  wasn't convinced it was much better with the dependency on settings,
+            #  accounts, etc. that aren't accessible from this class
+            #  (without modification)
+            aws_config = output['configuration']['provider_config'][
+                'aws']['expressions']
+            access_key = aws_config['access_key']['constant_value']
+            secret_key = aws_config['secret_key']['constant_value']
+            rds_client = boto3.client('rds', aws_access_key_id=access_key,
+                                      aws_secret_access_key=secret_key)
```
maorfr, line 602: `self.accounts` has the accounts, `name` in this scope is the account name. so the only thing missing is really `settings`, which we can initiate the class with?

otherwise we'd start building a new aws_api within this class...

there will be more use cases in the future, so we might as well do it "right"? :)

one use case i can think of right now is to check if a parameter group is used when trying to delete it, for example.

steveteahan: I ended up directly injecting `AWSApi` because that helps to avoid the need to `patch` to test this class. I didn't see any downside to this yet from initial testing. I still need to test the integrations that aren't `terraform_resources`.

maorfr (approved):
LGTM

this is pretty cool!

PR #2256, opened 2022-03-18
consolidate dockerfiles

fishi0x01:
**What**

Consolidate `Dockerfile.dev` and `Dockerfile`.

**Motivation**

`Dockerfile.dev` and `Dockerfile` share the same `build-image`. Merging both files reduces the need to update 2 files if sth changes in `Dockerfile`.

**Test**

Using `dev-image` for local development worked fine.


janboll (approved):
:+1:

dockerfiles/Dockerfile
```diff
@@ -12,7 +12,36 @@ RUN python3 -m pip install --no-cache-dir --upgrade pip setuptools wheel && \
...
+COPY . /work
+WORKDIR /work
+
+RUN python3 -m pip install --no-cache-dir --no-index --find-links=/tmp/work/wheels qontract-reconcile
+RUN python3 -m pip install -e .
+
+RUN chown -R reconcile /work && \
+    chown -R reconcile /.terraform.d
+
+USER reconcile
```
Piojo, line 39: I **love** this user. But please note there is at least one test that run only under jenkins, and the way they confirm if it's a Jenkins build is whether UID is 0. Please update `test_make.py`.

Piojo (changes requested):

dockerfiles/Dockerfile
fishi0x01: By idea, Jenkins should not be bothered here. The idea is that `dev-image` is only used for local development, i.e., it will not be used by CI. Not sure if Im missing sth here?
CI make targets use the `--target prod-image`, which does not use this user (yet).

PR #2257, opened 2022-03-19
[bug] terrascript_client initiates terraform state backend with wrong region when aws_api is initiated

maorfr:
if the state bucket region (coming from the vault secret) is different from the account's resources default region (coming from the account file) - the region selected for the terraform backend configuration is the wrong one. it should be the region of the state bucket, coming from the vault secret.
when `aws_api` and `terrascript_client` are initiated, they interfere with each other if `resourcesDefaultRegion` is different from the terraform bucket region.

this issue was introduced in #2165 (famous last words indeed), where we actually started overriding the state bucket region with the resources default region.

this PR starts by adding tests that are expected to pass. the fact that they are failing proves the existence of the bug.

i've isolated the "fix" to this single commit: https://github.com/app-sre/qontract-reconcile/pull/2257/commits/cab02840957cb4a39c02f2441abda7f8e9d01eb3

but if i'm honest, i have no idea why this bug happens, and this fix is a workaround.
i'm not sure if this bug has more symptoms, so one may wonder if there is really a bug...
i believe the bug has something to do with `functools.lru_cache` and AWSApi leaks. more information in #1947.

### DIFF
--- a/reconcile/test/test_utils_aws_helper.py
+++ b/reconcile/test/test_utils_aws_helper.py
@@ -37,29 +37,29 @@ def test_get_account_uid_from_role_link():
     assert result == expected
 
 
-class MockSecretReader():
+class MockSecretReader:
     def read_all(self, token):
         return token
 
 
 def test_get_tf_secrets():
-    account = {'name': 'a', 'automationToken': 'at'}
+    account = {"name": "a", "automationToken": "at"}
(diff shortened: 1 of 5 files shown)


geoberle (approved):
thanks for walking me though the change. LGTM but this drives me crazy

PR #2258, opened 2022-03-21
reproduce wrong region bug

maorfr:
reproducing the bug that was "fixed" in #2257


maorfr:
since vault uses lru_cache:
https://github.com/app-sre/qontract-reconcile/blob/0489cff416c4fc368c3f483afeaf010a2a789cfc/reconcile/utils/vault.py#L140
the two calls to secret_reader.read_all actually return the same in-memory object, which is why modifying it in one place will affect the other.

kudos @geoberle :1st_place_medal:

PR #2259, opened 2022-03-21
[openshift-saas-deploy] do not skip checking images

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4365

reverts #927

replaces #2134

this PR removes the possibility to skip checking images. this feature was relying heavily on jenkins functionality which is not present for tekton. more details in the ticket.

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -1052,11 +1052,11 @@ class SaasHerder():
                 else:
                     logging.info(
                         f"Skipping resource of kind {kind} on "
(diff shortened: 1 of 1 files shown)


geoberle:
for my understanding: the result of this change is that PR checks for saas files would fail if the referenced image does not exist.

maorfr:
this is correct for saas files v1. for saas files v2 this is already the case since the condition is broken (#2134)

geoberle (approved):
LGTM

PR #2260, opened 2022-03-21
[oc] refactor secret/configmap used in pod

maorfr:
instead of searching for a specific name, process the entire pod spec to find used resources and check if the requested name exists in them.

TODO: tests.

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -701,53 +701,37 @@ class OCDeprecated:  # pylint: disable=too-many-public-methods
     @staticmethod
     def get_resources_used_in_pod_spec(spec: Dict[str, Any], kind: str) -> Dict[str, Set[str]]:
         if kind not in ("Secret", "ConfigMap"):
-            raise KeyError(f"unsupported kind for get_resources_used_in_pod_spec: {kind}")
+            raise KeyError(f"unsupported resource kind: {kind}")
         if kind == "Secret":
-            volume_kind = "secret"
-            volume_kind_ref = "secretName"
-            env_from_kind = "secretRef"
-            env_kind = "secretKeyRef"
-            env_ref = "name"
+            volume_kind, volume_kind_ref, env_from_kind, env_kind, env_ref = \
+                "secret", "secretName", "secretRef", "secretKeyRef", "name"
         elif kind == "ConfigMap":
-            volume_kind = "configMap"
-            volume_kind_ref = "name"
-            env_from_kind = "configMapRef"
-            env_kind = "configMapKeyRef"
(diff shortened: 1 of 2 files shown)


reconcile/utils/oc.py
```diff
@@ -690,57 +690,66 @@ def get_obj_root_owner(self, ns, obj, allow_not_found=False,
...
+            env_ref = "name"
+        elif kind == "ConfigMap":
+            volume_kind = "configMap"
+            volume_kind_ref = "name"
+            env_from_kind = "configMapRef"
+            env_kind = "configMapKeyRef"
+            env_ref = "name"
+
+        resources = {}
+        volumes = spec.get("volumes")
```
Piojo, line 719: If this comes from GraphQl, then
```python
spec['volumes'] or []
```

allows you to get rid of the `if` below.

reconcile/utils/oc.py
```diff
@@ -690,57 +690,66 @@ def get_obj_root_owner(self, ns, obj, allow_not_found=False,
...
+                    resources.setdefault(resource_name, set())
+                except (KeyError, TypeError):
                     continue
-        return False
+        containers = spec["containers"]
+        init_containers = spec.get("initContainers")
+        if init_containers:
+            containers += init_containers
+        for c in containers:
+            env_from = c.get("envFrom")
```
Piojo, line 732: Same as above, either of the options below saves you one `if`, which makes the code easier to read.
```python
            env_from = c.get("envFrom", [])
            env_from = c["envFrom"] or []
```

reconcile/utils/oc.py
```diff
@@ -690,57 +690,66 @@ def get_obj_root_owner(self, ns, obj, allow_not_found=False,
...
-            for e in c.get('envFrom', []):
-                try:
-                    if e['configMapRef']['name'] == name:
-                        return True
-                except KeyError:
-                    continue
-            for e in c.get('env', []):
+    def get_resources_used_in_pod_spec(spec: Dict[str, Any], kind: str) -> Dict[str, Set[str]]:
+        if kind not in ("Secret", "ConfigMap"):
+            raise KeyError(f"unsupported kind for get_resources_used_in_pod_spec: {kind}")
```
Piojo, line 704: If this function changes its name you'll have to adapt the exception string too. Since you are going to have it in the stack trace anyways, why not either refer to the objective of the function or do away with it entirely?

```suggestion
            raise KeyError(f"unsupported resource kind {kind}")
```

(you could use `inspect.currentframe().f_code.co_name` but I doubt the extra import helps much here)

Piojo (approved):
Great!

PR #2261, opened 2022-03-21
[openshift-saas-deploy] validate planned data

maorfr:
depends on #2260

this PR adds logic to validate resources to be deployed before they are actually deployed.
this allows us to gradually add checks as we discover use cases that can be validated pre-deployment.

this PR adds the first condition - validate that all Secrets[1] used by a Deployment[2] exist[3] along with the required keys[4].

[1] or ConfigMaps
[2] or DeploymentConfig
[3] or will exist
[4] where applicable

this will reduce load from AppSRE by automating a pretty routine check, and will help onboarding tenants by validating that what they are trying to deploy makes sense.
in addition, we see multiple MRs to app-interface which promote a saas file target together with a secret from vault. as these secrets need to be updated in advance to contain any new keys, this PR will force a split of the MR - secret first, saas file after. this alone will reduce:
- review time of the IC
- debug time of tenant and the IC in case the Deployment gets updated before the Secret and the deployment pipeline fails.

TODO: tests

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -2,7 +2,7 @@ from dataclasses import dataclass, field
 import logging
 import itertools
 
-from typing import Any, Optional, Iterable, Mapping, Union
+from typing import Any, Dict, Optional, Iterable, Mapping, Union
 
 import yaml
 
@@ -21,7 +21,7 @@ from reconcile.utils.oc import PrimaryClusterIPCanNotBeUnsetError
 from reconcile.utils.oc import InvalidValueApplyError
 from reconcile.utils.oc import MetaDataAnnotationsTooLongApplyError
 from reconcile.utils.oc import StatefulSetUpdateForbidden
-from reconcile.utils.oc import OC_Map
+from reconcile.utils.oc import OCDeprecated, OC_Map
 from reconcile.utils.oc import StatusCodeError
 from reconcile.utils.oc import UnsupportedMediaTypeError
 from reconcile.utils.openshift_resource import OpenshiftResource as OR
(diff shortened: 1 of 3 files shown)


reconcile/openshift_base.py
```diff
@@ -621,8 +621,58 @@ def realize_data(
...
+def _validate_resources_used_exist(
+    ri: ResourceInventory,
+    oc: OC,
+    spec: Dict[str, Any],
+    cluster: str,
+    namespace: str,
+    kind: str,
+    name: str,
+    used_kind: str,
+):
```
Piojo, line 633: This seems to return `None`:
```python
) -> None:
```

reconcile/openshift_base.py
```diff
@@ -621,8 +621,58 @@ def realize_data(
...
+                ri.register_error()
+                continue
+
+
+def validate_planned_data(ri: ResourceInventory, oc_map: OC_Map):
+    for cluster, namespace, kind, data in ri:
+        oc = oc_map.get(cluster)
+
+        for name, d_item in data["desired"].items():
+            if kind in ("Deployment", "DeploymentConfig"):
```
Piojo, line 664: This looks worth moving up and into a `continue`, right?

```python
if kind not in ("Deployment", "DeploymentConfig"):
    continue

for name, d_item in data["desired"].items():
    ...
```

reconcile/openshift_base.py
```diff
@@ -621,8 +621,58 @@ def realize_data(
...
+            continue
+        # here it is! let's make sure it has all the required keys
+        for k in used_keys:
+            if k not in resource["data"]:
+                logging.error(f"{err_base} does not contain key: {k}")
+                ri.register_error()
+                continue
+
+
+def validate_planned_data(ri: ResourceInventory, oc_map: OC_Map):
```
Piojo, line 659:
```python
 def validate_planned_data(ri: ResourceInventory, oc_map: OC_Map) -> None: 
```

reconcile/openshift_base.py
```diff
@@ -621,8 +621,58 @@ def realize_data(
...
+        if not resource:
+            # no. where is used resource hiding? we can't find it anywhere
+            logging.error(f"{err_base} does not exist")
+            ri.register_error()
+            continue
+        # here it is! let's make sure it has all the required keys
+        for k in used_keys:
+            if k not in resource["data"]:
+                logging.error(f"{err_base} does not contain key: {k}")
+                ri.register_error()
```
Piojo, line 655: Since `keys()` gives you back a set-like object:
```python
missing_keys = used_keys - resource.data_keys()
if missing_keys:
    logging.error(f"{err_base} does not contain required keys: {missing_keys}")
    ri.register_error()
```

Piojo (approved):
Looks good. Please fix the tests before pushing.

geoberle (approved):
nice!

reconcile/openshift_base.py
maorfr: i'd prefer leaving things the way they are to allow room for future enhancements for additional kinds. does that make sense?

Piojo: Up to you. It's a minor nit, indeed.

maorfr:
i'll add tests some other time. promise.

PR #2263, opened 2022-03-22
[WIP] APPSRE-3652 add cli command gpg-encrypt

fishi0x01:
**Motivation**

Encrypt secrets from file/vault path with target user's gpg key and print out/save to local file.
This command will help us to deliver secret to a tenant who doesn't have certain vault access, especially during an incident.

**Changes**

- add new command:
```
qontract-cli gpg-encrypt --for-user orguser --vault-secret-path a/vault/path
```

- allow to run cli via `python tools/qontract_cli.py`, which makes this easier to use for local development
- allow qontract-cli to be executed from dev container image

**Test**

- running locally
- adding unit tests


tools/cli_commands/gpg_encrypt.py
```diff
@@ -0,0 +1,107 @@
...
+                f"No argument given which defines how to fetch the secret {self._command_data}"
+            )
+
+    def _get_gpg_key(self) -> Optional[str]:
+        try:
+            return [
+                user.get("public_gpg_key")
+                for user in self._users
+                if user.get("org_username") == self._command_data.target_user
+            ][0]
```
geoberle, line 64: the org_username in the schema is unique, so maybe we could make it searchable. this would enable this code to only load the one user that is required.

https://github.com/app-sre/qontract-schemas/blob/86ba651bdc8b93880a5fed2466acc4cdf6aeeb0b/graphql-schemas/schema.yml#L1987

maorfr:
when we merged the MR adding the above mentioned script, we already said that such commands are more suited for qontract-cli: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/21295

we can query a single user by making as pointed out by @geoberle

maorfr:
https://issues.redhat.com/browse/APPSRE-3652

tools/cli_commands/gpg_encrypt.py
fishi0x01: This is a great idea Gerd! Will try that now, thanks! :bow:

fishi0x01: Fixed :+1:

reconcile/queries.py
```diff
@@ -697,13 +712,31 @@ def get_queue_aws_accounts():
...
+    tmpl = CLUSTERS_MINIMAL_QUERY if minimal else CLUSTERS_QUERY
+    query = Template(tmpl).render(
+      filter=None,
+    )
+    return gqlapi.query(query)['clusters']
+
+
+@dataclass
+class ClusterFilter:
+    name: str = ""
```
geoberle, line 730: ❤️ i love filter objects and clean function arg lists

geoberle (approved):
LGTM

PR #2267, opened 2022-03-23
APPSRE-4731: Use official graphql library

bhushanthakur93:
This MR replaces `graphqlclient` which is a library unmaintained since 2018 with the official graphql client per https://github.com/graphql-python/gql

I tested some of the integrations locally and they seem to work well. However to get 100% confident I suggest we deploy this in our staging and let it observe for couple of days just to ensure there aren't any missing edge cases. 

I have verified the client displays detailed error message which was hampering our troubleshooting earlier.
```
reconcile.utils.gql.GqlApiError: Could not connect to GraphQL server ({'message': 'Cannot query field "booh" on type "User_v1".', 'locations': [{'line': 10, 'column': 5}], 'extensions': {'code': 'GRAPHQL_VALIDATION_FAILED'}})
```

### DIFF
--- a/reconcile/test/test_utils_gql.py
+++ b/reconcile/test/test_utils_gql.py
@@ -23,7 +23,7 @@ TEST_QUERY = """
 
 @pytest.fixture
 def mocked_client(mocker):
-    mocked_client = mocker.Mock(spec=Client, create_autospec=True)
+    mocked_client = mocker.create_autospec(Client)
     return mocked_client
 
 
--- a/reconcile/test/test_utils_gql.py
+++ b/reconcile/test/test_utils_gql.py
@@ -6,8 +6,8 @@ from reconcile.utils.gql import (
     GqlApiErrorForbiddenSchema,
     GqlApiIntegrationNotFound,
 )
-from gql import Client  # type: ignore
-from gql.transport.exceptions import TransportQueryError  # type: ignore
+from gql import Client
+from gql.transport.exceptions import TransportQueryError
 import pytest
 
 TEST_QUERY = """
--- a/reconcile/test/test_utils_gql.py
+++ b/reconcile/test/test_utils_gql.py
@@ -6,8 +6,8 @@ from reconcile.utils.gql import (
     GqlApiErrorForbiddenSchema,
     GqlApiIntegrationNotFound,
 )
-from gql import Client
-from gql.transport.exceptions import TransportQueryError
+from gql import Client  # type: ignore
+from gql.transport.exceptions import TransportQueryError  # type: ignore
 import pytest
 
 TEST_QUERY = """
--- a/reconcile/test/test_utils_gql.py
+++ b/reconcile/test/test_utils_gql.py
@@ -1,0 +1,72 @@
+import requests
+
+from reconcile.utils.gql import (
+    GqlApi,
+    GqlApiError,
+    GqlApiErrorForbiddenSchema,
+    GqlApiIntegrationNotFound,
+)
+from gql import Client
+from gql.transport.exceptions import TransportQueryError
+import pytest
+
+TEST_QUERY = """
+{
+    integrations: integrations_v1 {
+        name
+        description
+        schemas
+    }
+}
+"""
+
+
+def test_gqlapi_throws_gqlapierror_exception(mocker):
+    mocked_client = mocker.Mock(spec=Client)
+
+    mocked_client.execute.side_effect = Exception("Something went wrong!")
+    with pytest.raises(GqlApiError):
+        gql_api = GqlApi(mocked_client, validate_schemas=False)
+        gql_api.query(TEST_QUERY)
+
+    mocked_client.execute.side_effect = requests.exceptions.ConnectionError(
+        "Could not connect with GraphQL API"
+    )
+    with pytest.raises(GqlApiError):
+        gql_api = GqlApi(mocked_client, validate_schemas=False)
+        gql_api.query(TEST_QUERY)
+
+    mocked_client.execute.side_effect = TransportQueryError("Error in GraphQL payload")
+    with pytest.raises(GqlApiError):
+        gql_api = GqlApi(mocked_client, validate_schemas=False)
+        gql_api.query(TEST_QUERY)
+
+    mocked_client.execute.side_effect = AssertionError(
+        "Transport returned an ExecutionResult without data or errors"
+    )
+    with pytest.raises(GqlApiError):
+        gql_api = GqlApi(mocked_client, validate_schemas=False)
+        gql_api.query(TEST_QUERY)
+
+
+def test_gqlapi_throws_gqlapiintegrationnotfound_exception(mocker):
+    mocked_client = mocker.Mock(spec=Client)
+    mocked_client.execute.return_value.formatted = {
+        "data": {"integrations": [{"name": "INTEGRATION", "schemas": "TEST_SCHEMA"}]}
+    }
+
+    with pytest.raises(GqlApiIntegrationNotFound):
+        gql_api = GqlApi(mocked_client, "INTEGRATION_NOT_FOUND", validate_schemas=True)
+        gql_api.query(TEST_QUERY)
+
+
+def test_gqlapi_throws_gqlapierrorforbiddenschema_exception(mocker):
+    mocked_client = mocker.Mock(spec=Client)
+    mocked_client.execute.return_value.formatted = {
+        "data": {"integrations": [{"name": "INTEGRATION", "schemas": "TEST_SCHEMA"}]},
+        "extensions": {"schemas": ["TEST_SCHEMA", "FORBIDDEN_TEST_SCHEMA"]},
+    }
+
+    with pytest.raises(GqlApiErrorForbiddenSchema):
+        gql_api = GqlApi(mocked_client, "INTEGRATION", validate_schemas=True)
+        gql_api.query(TEST_QUERY)
--- a/reconcile/test/test_utils_gql.py
+++ b/reconcile/test/test_utils_gql.py
@@ -21,14 +21,21 @@ TEST_QUERY = """
 """
 
 
-def test_gqlapi_throws_gqlapierror_exception(mocker):
-    mocked_client = mocker.Mock(spec=Client)
+@pytest.fixture
+def mocked_client(mocker):
+    mocked_client = mocker.Mock(spec=Client, create_autospec=True)
+    return mocked_client
+
+
+def test_gqlapi_throws_gqlapierror_when_generic_exception_thrown(mocked_client):
 
     mocked_client.execute.side_effect = Exception("Something went wrong!")
     with pytest.raises(GqlApiError):
         gql_api = GqlApi(mocked_client, validate_schemas=False)
         gql_api.query(TEST_QUERY)
 
+
+def test_gqlapi_throws_gqlapierror_when_connectionerror_exception_thrown(mocked_client):
     mocked_client.execute.side_effect = requests.exceptions.ConnectionError(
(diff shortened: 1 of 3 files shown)


reconcile/utils/gql.py
```diff
@@ -134,13 +135,7 @@ def query(self, query, variables=None, skip_validation=False):
...
                 raise GqlApiErrorForbiddenSchema(forbidden_schemas)
 
-        if "errors" in result:
-            raise GqlApiError(result["errors"])
-
-        if "data" not in result:
-            raise GqlApiError(("`data` field missing from GraphQL" "server response."))
-
-        return result["data"]
+        return result.get("data")
```
steveteahan, line 138: It looks like if the `data` key didn't exist before, we'd raise a `GqlApiError`, but `get()` will return `None` instead if the key doesn't exist. Is there a specific reason to change this?

steveteahan (commented):
Being able to actually see the errors is going to be great! Just a couple other high-level thoughts here:

1. I don't think it'd be the worst idea to add some type hints to the methods that you're modifying.
2. It doesn't look like we have any tests for this module currently? This leads into the question that I had about the API changing by returning `None` instead of raising an exception. Maybe worth adding some tests for the parts that are changing?

reconcile/utils/gql.py
bhushanthakur93: The client library now raises AssertionError if data key doesn't exist. If you follow the trail of calls happening within you should see this block of code. I have added a comment when the exception is caught first and rethrown as `GqlApiError`

```
        assert (
            result.data is not None
        ), "Transport returned an ExecutionResult without data or errors"
```

reconcile/utils/gql.py
```diff
@@ -86,16 +85,20 @@ def __init__(self, url, token=None, int_name=None, validate_schemas=False):
         self.token = token
         self.integration = int_name
         self.validate_schemas = validate_schemas
-        self.client = GraphQLClient(self.url)
+        self.req_headers = None
```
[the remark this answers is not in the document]
bhushanthakur93: Good point! Still getting used to best practices.

reconcile/utils/gql.py
```diff
@@ -108,18 +114,16 @@ def __init__(self, url, token=None, int_name=None, validate_schemas=False):
                 raise GqlApiIntegrationNotFound(int_name)
 
     @retry(exceptions=GqlApiError, max_attempts=5, hook=capture_and_forget)
-    def query(self, query, variables=None, skip_validation=False):
+    def query(self, query, variables=None, skip_validation=False) -> Dict[str, Dict]:
```
Piojo, line 117: No need for the abstract `Dict` here. Use `dict` as the return type.

reconcile/utils/gql.py
steveteahan: So, `data` is guaranteed to be in the dictionary by the time you get this far? It looks like you switched it back to the original dict access method.

You may already be aware of this, but you should take care when using `some_dict['key']` versus `some_dict.get('key')`. If the key doesn't exist, with the former you'll get a `KeyError` exceptions, whereas the latter will just return `None`. If a key should absolutely exist, then don't use `get()`. If you're going to use `get()`, then the method type hints should indicate that it can return `None` and that has to be handled everywhere.

bhushanthakur93: > So, `data` is guaranteed to be in the dictionary by the time you get this far? It looks like you switched it back to the original dict access method.
> 
Yes, based on the library code `data` is guaranteed to be there.
> You may already be aware of this, but you should take care when using `some_dict['key']` versus `some_dict.get('key')`. If the key doesn't exist, with the former you'll get a `KeyError` exceptions, whereas the latter will just return `None`. If a key should absolutely exist, then don't use `get()`. If you're going to use `get()`, then the method type hints should indicate that it can return `None` and that has to be handled everywhere.
Yep, learned this detail recently :) I reverted to previous access method just to minimize the change.

reconcile/utils/gql.py
```diff
@@ -108,18 +114,16 @@ def __init__(self, url, token=None, int_name=None, validate_schemas=False):
...
-            # /blob/master/graphqlclient/client.py#L32-L33
-            with open(os.devnull, "w") as f, contextlib.redirect_stdout(f):
-                result_json = self.client.execute(query, variables)
+            result = self.client.execute(
+                gql(query), variables, get_execution_result=True
+            ).formatted
         except Exception as e:
+            # The client raises TransportQueryError if there is error returned with payload
+            # or AssertionError when no data is returned
             raise GqlApiError("Could not connect to GraphQL server ({})".format(e))
```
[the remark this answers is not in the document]
bhushanthakur93: I kept it as generic so as to avoid digging into implementation the next time we update the library if the contract changes for some reason. WDYT?
Yeah, the `AssertionError` was bit weird to me as well. I thought maybe this is a python thing :smile:

reconcile/utils/gql.py
```diff
@@ -86,16 +88,20 @@ def __init__(self, url, token=None, int_name=None, validate_schemas=False):
         self.token = token
         self.integration = int_name
         self.validate_schemas = validate_schemas
-        self.client = GraphQLClient(self.url)
+        self._req_headers = None
```
steveteahan, line 91: I'd put a `None` return type hint on `__init__` so that Mypy will type check this code. I suspect that when you do so, it will probably be unhappy that `_req_headers` starts as `None`, but conditionally becomes a `dict`. The solution would be to type it as an `Optional[dict[str, str]]`.

reconcile/utils/gql.py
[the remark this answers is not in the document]
steveteahan: I think that always returning "Could not connect" is misleading because some of these aren't connection issues. The advantage of returning our own exception is that when the implementation changes, like in this case, we don't leak that change to the clients. I'm not sure that we get that much more benefit from having a catch-all and a single exception message.

You probably don't need to catch `AssertionError`. In Python, `assert` is really only used for things that absolutely shouldn't be happening, or to do things like appease Mypy. You can read up more on why asserts aren't used much in Python and how certain optimizer flags can actually disable them. I would still suggest we catch the other exceptions specifically and keep the exception messages the same as they were before unless we have a good reason to change them. You can also catch `Exception` at the end, but catching the more specific exceptions first will replace the if statements that existed before.

reconcile/utils/gql.py
bhushanthakur93: I added `None` return type hint. Also used DI to inject `Client` for better testability.

reconcile/utils/gql.py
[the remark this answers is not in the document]
bhushanthakur93: I decided to `AssertionError` as the underlying library is throwing it and it felt a bit odd not to catch it but others. I hope the library changes the implementation.

Let me know if this is ok.

reconcile/utils/gql.py
bhushanthakur93: addressed

bhushanthakur93:
@steveteahan I think this in a good state to be merged.

reconcile/test/test_utils_gql.py
```diff
@@ -0,0 +1,72 @@
...
+        name
+        description
+        schemas
+    }
+}
+"""
+
+
+def test_gqlapi_throws_gqlapierror_exception(mocker):
+    mocked_client = mocker.Mock(spec=Client)
```
steveteahan, line 25: You pretty much always want `create_autospec` unless there's a compelling reason not to use it. I won't comment on each instance of this.

reconcile/test/test_utils_gql.py
```diff
@@ -0,0 +1,72 @@
...
+    integrations: integrations_v1 {
+        name
+        description
+        schemas
+    }
+}
+"""
+
+
+def test_gqlapi_throws_gqlapierror_exception(mocker):
```
steveteahan, line 24: I'd separate this out into separate tests rather than changing the mock along the way and making it a long test. Alternatively, you could look at pytest parameterize: https://docs.pytest.org/en/6.2.x/parametrize.html

reconcile/utils/gql.py
```diff
@@ -129,17 +125,11 @@ def query(self, query, variables=None, skip_validation=False):
 
         if self.validate_schemas and not skip_validation:
             forbidden_schemas = [
-                schema for schema in query_schemas if schema not in self._valid_schemas
+                schema for schema in query_schemas if schema not in self._valid_schemas  # type: ignore
```
steveteahan, line 128: I'd ignore the specific codes if possible rather than a blanket ignore: https://mypy.readthedocs.io/en/stable/error_codes.html

What's the reason behind ignoring this one?

reconcile/utils/gql.py
```diff
@@ -173,10 +163,20 @@ def get_queried_schemas(self):
 
 def init(url, token=None, integration=None, validate_schemas=False):
     global _gqlapi
-    _gqlapi = GqlApi(url, token, integration, validate_schemas)
+    client = _init_gql_client(url, token)
+    _gqlapi = GqlApi(client, integration, validate_schemas)
     return _gqlapi
 
 
+def _init_gql_client(url, token) -> Client:
```
steveteahan, line 171: For type hints, you want to add them for all parameters in addition to the return value.

steveteahan (commented):
Looks good overall! Just a couple other minor fixes.

reconcile/utils/gql.py
```diff
@@ -81,21 +80,16 @@ class GqlApi:
     _valid_schemas = None
     _queried_schemas: Set[Any] = set()
 
-    def __init__(self, url, token=None, int_name=None, validate_schemas=False):
-        self.url = url
-        self.token = token
+    def __init__(self, client, int_name=None, validate_schemas=False) -> None:
```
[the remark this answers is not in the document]
bhushanthakur93: I do miss strongly typed language :cry:

reconcile/utils/gql.py
bhushanthakur93: I ran into `Unsupported right operand type for in ("None")  [operator]` . I'm not sure the best way to address this.

bhushanthakur93: Although I'm not sure why I ran into this now given I didn't change the code at all.

reconcile/test/test_utils_gql.py
```diff
@@ -0,0 +1,83 @@
...
+        description
+        schemas
+    }
+}
+"""
+
+
+@pytest.fixture
+def mocked_client(mocker):
+    mocked_client = mocker.Mock(spec=Client, create_autospec=True)
```
steveteahan, line 26: Is `create_autospec` a valid kwarg for `Mock`? Check out: https://docs.python.org/3/library/unittest.mock.html

bhushanthakur93: Good catch! I was blind when I looked at plugin.py for mocker, that had create_autopec.

steveteahan: The last bit is that `spec` is a positional arg in `create_autospec`, so you don't even need the `spec=` part.

reconcile/utils/gql.py
```diff
@@ -129,16 +127,18 @@ def query(self, query, variables=None, skip_validation=False):
 
         if self.validate_schemas and not skip_validation:
             forbidden_schemas = [
-                schema for schema in query_schemas if schema not in self._valid_schemas
+                # Here we are explicitly type casting _valid_schemas to list as mypy is unable to infer the type
```
steveteahan, line 130: This no longer applies?

bhushanthakur93: derp! I'll fix it.

PR #2268, opened 2022-03-23
implement signalfx endpoint monitoring provider integration

geoberle:
## what changed?
this change introduces a new integration `signalfx-prometheus-endpoint-monitoring`. the general idea of this integration is to scrape endpoint related metrics from signalfx using a prometheus compatible exporter (https://github.com/app-sre/signalfx-prometheus-exporter). the method of scraping is exactly the same as with the blackbox-exporter-endpoint-monitoring integration, namely a scrape job defined via a `Probe` custom resource from the Prometheus operator.

## ⚠️ how to review this MR

i highly suggest to have a look at the first commit 6592096e778595f0c3bc28347cb69e9e4617adfb for the refactoring that does not change behaviour and then look at the second commit f6d030173e38dbf734c1c6c0a2455f91caf006db that introduces the new integration. based on the refactoring of blackbox exporter endpoint monitoring provider into closedbox_endpoint_monitoring_base.py, this integrations magic mostly happens within
the probe building. 

## how are metrics scraped?
the targets for the `Probe` CR are not the endpoint URLs but the endpoint names (`app-1.yml#endPoints.name`). when scraping signalfx-prometheus-exporter with those endpoint names, it returns metrics that belong to this identifier, e.g. in case of catchpoint, the endpoint name matches the catchpoint check name and signalfx-prometheus-exporter needs to be configured in a way that it knows how to find the signalfx metrics belonging to that catchpoint check.

## example
```yaml
  ---
  $schema: /dependencies/endpoint-monitoring-provider-1.yml

  labels: {}

  name: catchpoint-http-2xx-prod-critical
  description: Checks a URL for HTTP 2xx status codes via blackbox-exporter
  provider: signalfx
  metricLabels:
    environment: production
  signalFx:
    namespace:
      $ref: /services/observability/namespaces/openshift-customer-monitoring.app-sre-prod-01.yml
    exporterUrl: http://signalfx-prometheus-exporter.xxx.svc.cluster.local:9091/metrics
    targetFilterLabel: probe
```

in this example the signalfx provider is used to scrape catchpoint metrics from signalfx. this configuration expects signalfx-prometheus-exporter to be configured in a way that metrics for a catchpoint check are exposed with a label `probe: catchpoing name`. all metrics with the same label value are considered belong to the same catchpoint check and are related to the respective `app-1.yml#endPoint`

the `Probe` is also configured to rewrite the `instance` label with the endpoint URL. this way the scraped metrics will behave the same as `Probe` metrics scraped from blackbox-exporter. having the actual endpoint URL in the metrics makes it easier to define alerts that also contain the URL and will yield to faster results when relating an alert to an actual service endpoint. the actual endpoint name (e.g. catchpoint check name) will still be available in the`targetFilterLabel` label (e.g. `probe` in this case).

resulting metrics look like this:

```
catchpoint_check_failures_total{environment="stage", instance="endpoint url", job="name of the provider", probe="endpoint name / catchpoint check name"}
catchpoint_checks_total{environment="stage", instance="endpoint url", job="name of the provider", probe="endpoint name / catchpoint check name"}
```

part of https://issues.redhat.com/browse/APPSRE-4700
design doc: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/service-endpoint-monitoring.md


reconcile/signalfx_endpoint_monitoring.py
```diff
@@ -0,0 +1,85 @@
...
+        internal,
+        use_jump_host,
+    )
+
+
+def build_probe(
+    provider: EndpointMonitoringProvider, endpoints: list[Endpoint]
+) -> Optional[OpenshiftResource]:
+    signalfx = provider.signalFx
+    if signalfx:
```
fishi0x01, line 42: nit: walrus code golf :golfing: 

```
if signalfx := provider.signalFx:
```

Or better to avoid extra indent completely:

```
if not signalfx:
    return None
prober_url = parse_prober_url(signalfx.exporterUrl)
...
```

fishi0x01 (commented):
Such a beautiful read!! :heart: :heart: 
Love the code and PR description detail! Such a joy reading it, I got tempted into shameless code golf nits -> please forgive me :see_no_evil:

Since this is quite a lot, I need more time to get into this fully, I think it might be good if someone with more experience also has a look.

fishi0x01:
Also, maybe worth removing the `[WIP]` from PR description

reconcile/blackbox_exporter_endpoint_monitoring.py
```diff
@@ -1,80 +1,53 @@
...
+                 p.blackboxExporter.module not in allowed_modules:
+                LOG.error(
+                    f"endpoint monitoring provider {p.name} uses "
+                    f"blackbox-exporter module {p.blackboxExporter.module} "
+                    f"which is not in the allow list {allowed_modules} of "
+                    "app-interface-settings"
+                )
+                verification_errors = True
+    if verification_errors:
+        sys.exit(1)
```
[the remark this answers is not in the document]
Piojo: I 100% agree but it can't be done here.

Here's the thing: `hack/run_integration.py` runs this integration, and there is an `except SystemExit` somewhere in there. Based on that `SystemExit` we increase the counters for successful or failed runs, which then alert us when integrations are truly hosed.

So, yes, I would **love** to get rid of all these `sys.exit`, but it needs to be done in a different place, and with good thought on how we're going to report if an integration is succeeding or not.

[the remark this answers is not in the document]
geoberle: i will not touch this for now. as @Piojo stated, this needs good thought and a general approach

reconcile/closedbox_endpoint_monitoring_base.py
```diff
@@ -0,0 +1,162 @@
...
+
+        # reconcile
+        for ep_mon_provider, endpoints in desired_endpoints.items():
+            probe = probe_builder(ep_mon_provider, endpoints)
+            if probe:
+                fill_desired_state(ep_mon_provider, probe, ri)
+        ob.realize_data(dry_run, oc_map, ri, thread_pool_size, recycle_pods=False)
+
+        if ri.has_error_registered():
+            sys.exit(1)
```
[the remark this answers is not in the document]
geoberle: i moved the sys.exit to the calling integrations

reconcile/blackbox_exporter_endpoint_monitoring.py
```diff
@@ -1,80 +1,61 @@
...
 
-    monitoring: list[Monitoring]
+    try:
+        run_for_provider(
+            PROVIDER, build_probe,
+            QONTRACT_INTEGRATION, QONTRACT_INTEGRATION_VERSION,
+            dry_run, thread_pool_size, internal, use_jump_host
+        )
```
rporres, line 47: Our style guide favours calling functions with keyword arguments, especially with long argument calls

reconcile/signalfx_endpoint_monitoring.py
```diff
@@ -0,0 +1,85 @@
...
+    try:
+        run_for_provider(
+            PROVIDER,
+            build_probe,
+            QONTRACT_INTEGRATION,
+            QONTRACT_INTEGRATION_VERSION,
+            dry_run,
+            thread_pool_size,
+            internal,
+            use_jump_host,
```
rporres, line 34: please consider the use of keyword arguments, especially for these long calls

geoberle: see [408dde8](https://github.com/app-sre/qontract-reconcile/pull/2268/commits/408dde8472ff90deacb8f955c98fe3c04b9c3eea)

reconcile/blackbox_exporter_endpoint_monitoring.py
geoberle: see [408dde8](https://github.com/app-sre/qontract-reconcile/pull/2268/commits/408dde8472ff90deacb8f955c98fe3c04b9c3eea)

reconcile/signalfx_endpoint_monitoring.py
```diff
@@ -0,0 +1,85 @@
...
+def build_probe(
+    provider: EndpointMonitoringProvider, endpoints: list[Endpoint]
+) -> Optional[OpenshiftResource]:
+    signalfx = provider.signalFx
+    if not signalfx:
+        return None
+    prober_url = parse_prober_url(signalfx.exporterUrl)
+    prober_url["path"] += "/" + signalfx.targetFilterLabel
+
+    relabeling = [
```
rporres, line 50: Maybe a quick comment explaining this nice relabeling configuration?

geoberle: good idea. this needs some explanation! [86f77be](https://github.com/app-sre/qontract-reconcile/pull/2268/commits/86f77befba25dc49e6c6d8ce68df5197726f514d)

PR #2270, opened 2022-03-23
Update AWSApi.get_support_cases to use us-east-1 for the support API

steveteahan:
A recent change, #2257, changed the way that the AWS region is determined for sessions that are created in `AWSApi`. One consequence of this is that some AWS services are only located in `us-east-1`, in this case the support API.

Before this change:

```
$ qontract-reconcile --config config.prod.toml --dry-run --log-level INFO aws-support-cases-sos
[2022-03-23 14:58:12] [INFO] [gql.py:init_from_config:220] - using gql endpoint http://localhost:4000/graphqlsha/514f61fcfa6299538bd9e172211f230df64ae191e53db006d3a87c40e8d558d6
[2022-03-23 14:59:58] [ERROR] [aws_api.py:get_support_cases:696] - [account-name] error getting support cases. details: Could not connect to the endpoint URL: "https://support.eu-central-1.amazonaws.com/"
```

With this change the integration runs without errors.

### DIFF
--- a/reconcile/utils/aws_api.py
+++ b/reconcile/utils/aws_api.py
@@ -51,6 +51,10 @@ class MissingARNError(Exception):
 
 KeyStatus = Union[Literal['Active'], Literal['Inactive']]
 
+# The AWS support API uses a us-east-1 endpoint for all regions
+# https://docs.aws.amazon.com/general/latest/gr/awssupport.html
+AWS_SUPPORT_REGION = 'us-east-1'
+
 
 class AWSApi:  # pylint: disable=too-many-public-methods
     """Wrapper around AWS SDK"""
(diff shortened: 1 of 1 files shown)


maorfr:
Won't work for fedramp

steveteahan:
> Won't work for fedramp, where the region is us-gov-east-1

Good catch, let me take another look.

maorfr:
you can decide if it's fedramp or not according to the partition: https://github.com/app-sre/qontract-schemas/blob/21a4ad60553dd6b3b67c715627f9353707864e95/schemas/aws/account-1.yml#L100-L105

steveteahan:
> you can decide if it's fedramp or not according to the partition: https://github.com/app-sre/qontract-schemas/blob/21a4ad60553dd6b3b67c715627f9353707864e95/schemas/aws/account-1.yml#L100-L105

I think that using `partition` is probably a little bit less likely to break in the future if AWS went for a completely different region naming scheme. It was simple enough to fix, so I did so.

maorfr (approved):

PR #2272, opened 2022-03-24
enable option to template resources with graphql query results

maorfr:

with this PR, we can add sections to be templated based on graphql query results.
for example:
```
{{ query('/queries/jira_boards.graphql') }}
```
this will results in an iterable which we can use to template.
in this example, this is the content of /queries/jira_boards.graphql:
```
{
    jira_boards_v1 {
        name
    }
}
```

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -229,6 +229,13 @@ def lookup_github_file_content(repo, path, ref, tvars=None):
     return c.decode("utf-8")
 
 
+def lookup_graphql_query_results(query: str) -> list[Any]:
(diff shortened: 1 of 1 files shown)


geoberle (approved):
LGTM!

reconcile/openshift_resources_base.py
```diff
@@ -228,6 +228,13 @@ def lookup_github_file_content(repo, path, ref, tvars=None):
     return c.decode('utf-8')
 
 
+def lookup_graphql_query_results(query):
```
steveteahan, line 231: Type hints?

PR #2276, opened 2022-03-24
Parallelize tox runs in Dockerfile.test

steveteahan:
pylint was using all available core due to the '-j0' option. It seems
that running tox environments in parallel will provide better improvements
in the total execution time.

On my laptop, this ran in `7m7.225s` compared to `8m56.947s` without tox parallelization. We can get more real-life data from our builds over the next week or two.


Piojo (approved):
Makes sense to me. We can totally afford this parallelism in our instances.

PR #2277, opened 2022-03-25
add support to pass kwargs to query()

maorfr:

follows up on #2272

this PR adds support to pass keyword arguments to a `query()` function in a templated resource, to allow us (for example) to query a resource by it's unique key.

for example:
```
{{ query('/queries/jira_boards.graphql', name=MYBOARD) }}
```
in this example, this is the content of /queries/jira_boards.graphql:
```
{
    jira_boards_v1
    (
        name: "{{ name }}"
    )
    {
        name
    }
}
```

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -246,9 +246,7 @@ def process_jinja2_template(body, vars=None, env=None):
     vars.update(
         {"github": lambda u, p, r, v=None: lookup_github_file_content(u, p, r, vars)}
     )
-    vars.update(
-        {"query": lambda path, **kwargs: lookup_graphql_query_results(path, **kwargs)}
-    )
+    vars.update({"query": lookup_graphql_query_results})
     try:
         env = jinja2.Environment(
             extensions=[B64EncodeExtension, RaiseErrorExtension],
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -229,9 +229,10 @@ def lookup_github_file_content(repo, path, ref, tvars=None):
     return c.decode("utf-8")
 
 
-def lookup_graphql_query_results(query: str) -> list[Any]:
+def lookup_graphql_query_results(query: str, kwargs: dict[str, Any] = {}) -> list[Any]:
     gqlapi = gql.get_api()
     resource = gqlapi.get_resource(query)["content"]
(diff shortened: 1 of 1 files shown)


patjlm:
I was actually going to ask for that after the query implem! that's great!

reconcile/openshift_resources_base.py
```diff
@@ -229,10 +229,11 @@ def lookup_github_file_content(repo, path, ref, tvars=None):
     return c.decode("utf-8")
 
 
-def lookup_graphql_query_results(query: str) -> list[Any]:
+def lookup_graphql_query_results(query: str, kwargs: dict[str, Any] = {}) -> list[Any]:
     gqlapi = gql.get_api()
     resource = gqlapi.get_resource(query)["content"]
-    results = list(gqlapi.query(resource).values())[0]
+    templated_resource = jinja2.Template(resource).render(kwargs)
```
patjlm, line 235: nit: that's more of a `rendered_resource` i think.

patjlm (approved):
Love it

reconcile/openshift_resources_base.py
```diff
@@ -229,10 +229,11 @@ def lookup_github_file_content(repo, path, ref, tvars=None):
     return c.decode("utf-8")
 
 
-def lookup_graphql_query_results(query: str) -> list[Any]:
+def lookup_graphql_query_results(query: str, kwargs: dict[str, Any] = {}) -> list[Any]:
```
Piojo, line 232: That mutable default is going to be problematic, I'd be surprised if the linter let it pass.
```suggestion
def lookup_graphql_query_results(query: str, kwargs: Optional[dict[str, Any]]= None) -> list[Any]:
    if kwargs is None:
        kwargs = {}
```
Also, kwargs is usually reserved to `def foo(*args, **kwargs)` so that we could call this as `lookup_graphql_query_results('{...}', placeholder1='hello', placeholder2='world')`

That way your `render` would be:

```python
rendered_resource = jinja2.Template(resource).render(**kwargs)
```

Also, in that case I wouldn't bother typing `**kwargs`.

Piojo (changes requested):

reconcile/openshift_resources_base.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/2277/commits/4cb9542b99fc292d32810e7fc02cc09ba1c80c0a

PR #2279, opened 2022-03-25
pin pylint - follow-up pr will remove this

fishi0x01:
Newest pylint release 2.13 currently breaks the linting check. Temporarily pin to lower version to unblock CI.


steveteahan (changes requested):
LGTM with the understanding that we'll unpin once the fixes are in place. Thanks!

steveteahan (approved):

fishi0x01:
No worries @steveteahan .

This never happened :)

PR #2280, opened 2022-03-25
Fix linting errors caused by pylint upgrade

bhushanthakur93:


reconcile/saas_file_owners.py
```diff
@@ -48,7 +48,7 @@ def collect_owners():
 
     # make owners suitable for json dump
     for k in owners:
-        owners[k] = list(owners[k])
+        owners[k] = list(owners[k])  # pylint: disable=modified-iterating-dict
```
fishi0x01, line 51: Lets fix instead:

```
    # make owners suitable for json dump
    ans = {}
    for k, v in owners.items():
        ans[k] = list(v)

    return ans
```

fishi0x01: Or better actually

```
 # make owners suitable for json dump
 return {k: list(v) for k, v in owners.items()}
```

bhushanthakur93: Addressed.

fishi0x01 (approved):
Thanks for fixing this! :bow:

PR #2286, opened 2022-03-25
sort clusters according to their version and soakDays before considering their upgrade

patjlm:
This allows to prioritize clusters sharing the same upgrade mutex.

The cluster with lower version, or lower soakDays (when versions are equal) will get upgraded first

@maorfr PTAL. Do you think that's fine ? I am actually wondering if higher version clusters should not upgrade first. It's a matter of taste..


maorfr:
can you add some tests to help understand?

maorfr (approved):
LGTM

patjlm:
https://issues.redhat.com/browse/APPSRE-4773

PR #2295, opened 2022-03-29
[openshift-resources] add support to test if a url exists within a resource

maorfr:
this will allow us to test the existence of a url and template based on the results.
the `url_makes_sense` function was introduced in https://github.com/app-sre/qontract-reconcile/pull/2141/commits/fcbd19d35d93bcf84d3d9b0707ae221e6e95b4b5.

the intention behind this PR is to use it for the same purpose - to check if an `architectureDocument` link makes sense.

example usage: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/36181

### DIFF
--- a/reconcile/checkpoint.py
+++ b/reconcile/checkpoint.py
@@ -9,7 +9,7 @@ import re
 from functools import partial, lru_cache
 from http import HTTPStatus
 from pathlib import Path
-from typing import Any, Iterable, Mapping, Union
+from typing import Any, Callable, Dict, Iterable, Mapping, Union
 
 import requests
 from jinja2 import Template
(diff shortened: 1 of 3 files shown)


reconcile/checkpoint.py
```diff
@@ -30,6 +30,7 @@
 )
 
 
+@lru_cache
```
Piojo, line 33: We've had a lenghty discussion elsewhere about `lru_cache` leaking memory. Maybe we should prefer `cache`?

Piojo (approved):
Minor nit, but overall LGTM.

reconcile/checkpoint.py
maorfr: wasn't it related to the use of `self`?

PR #2299, opened 2022-03-29
Remove direct dependency on tzlocal

Piojo:
It used to break unleash but it seems to have been fixed. We want to
be able to receive security updates easily.


Piojo:
I'll consider the two plus signs I've gotten in slack as approvals. Merging in...

PR #2301, opened 2022-03-29
prometheus-rules-tester be more forgiving with variables

maorfr:
even more then #2292


geoberle:
> even more then #2292

😂

geoberle (approved):

PR #2302, opened 2022-03-29
[WIP] Fix issues with official gql client and revert graphql earlier commit

bhushanthakur93:
This MR addresses following 
- Reverts cbf420df727929c684d9cf82209baffe0ebbdf75 to add graphql library changes
- Fixes verbose logging by the library per https://gql.readthedocs.io/en/stable/advanced/logging.html#disabling-logs
- Fix `During handling of the above exception, another exception occurred` error
- Attempt to fix `Transport is already connected` by introducing threading.Lock within `query` method. This was difficult to reproduce locally but hoping we don't see this error when we deploy this in staging.


reconcile/utils/gql.py
```diff
@@ -30,6 +33,10 @@
 }
 """
 
+requests_logger.setLevel(logging.WARNING)
+
+_lock = Lock()
```
steveteahan, line 38: Isn't the issue that the `Client` blocks on `execute` and can only run one at a time? If so, shouldn't the `Lock` really be part of the class? I'd also make it clear what the purpose of the lock is.

reconcile/utils/gql.py
```diff
@@ -78,24 +85,19 @@ def __init__(self, path, msg):
...
     _queried_schemas: Set[Any] = set()
 
-    def __init__(self, url, token=None, int_name=None, validate_schemas=False):
-        self.url = url
-        self.token = token
+    def __init__(self, client: Client, int_name=None, validate_schemas=False) -> None:
         self.integration = int_name
         self.validate_schemas = validate_schemas
-        self.client = GraphQLClient(self.url)
+        self.client = client
```
steveteahan, line 94: While I like the direction that you were heading in with this change, maybe passing in the `Client` and relying on a single session breaks too many assumptions with this class? This is compared to initiating a new request on each `query` like [we were doing before](https://github.com/prisma-labs/python-graphql-client/blob/master/graphqlclient/client.py#L26)?

This may or may not be the right solution, but it would maintain the functionality of the old client without needing to worry about a new set of issues (lock contention).

steveteahan (changes requested):

Piojo (commented):
I wonder if that global `gqlapi` isn't doing more harm than good.

Considering how we run things, it may be better to have your `GqlApi` provide a context manager we can query, like:

```python
with GqlApi(client, QONTRACT_RECONCILE_INTEGRATION_NAME, ...) as gql:
    rs1 = gql.query("query1", ...)
    rs2 = gql.query("query2", ...)
```
If we're entering this rabbit hole and we've already had problems with locks and shared instances, this small re-design is worth considering.

bhushanthakur93:
> I wonder if that global `gqlapi` isn't doing more harm than good.
> 
> Considering how we run things, it may be better to have your `GqlApi` provide a context manager we can query, like:
> 
> ```python
> with GqlApi(client, QONTRACT_RECONCILE_INTEGRATION_NAME, ...) as gql:
>     rs1 = gql.query("query1", ...)
>     rs2 = gql.query("query2", ...)
> ```
> 
> If we're entering this rabbit hole and we've already had problems with locks and shared instances, this small re-design is worth considering.

I switched approach after taking a look at how some of our integrations are still not sharded and the lock design may impact run time of those.

reconcile/utils/gql.py
bhushanthakur93: Removed the use of lock.

reconcile/utils/gql.py
bhushanthakur93: Per our discussion, I removed the locks and instead went with reinitializing `Client` every time for new query.

steveteahan (commented):
One open question to address, otherwise lgtm!

reconcile/utils/gql.py
```diff
@@ -78,24 +80,22 @@ def __init__(self, path, msg):
 
 
 class GqlApi:
-    _valid_schemas = None
+    _valid_schemas: list[str] = []
     _queried_schemas: Set[Any] = set()
 
-    def __init__(self, url, token=None, int_name=None, validate_schemas=False):
+    def __init__(
+        self, url: str, token: str, int_name=None, validate_schemas=False
```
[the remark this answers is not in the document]
bhushanthakur93: good catch! I meant to revert to previous contract but I missed it being a kwarg. This is addressed now.

reconcile/utils/gql.py
```diff
@@ -108,17 +108,29 @@ def __init__(self, url, token=None, int_name=None, validate_schemas=False):
                 raise GqlApiIntegrationNotFound(int_name)
 
     @retry(exceptions=GqlApiError, max_attempts=5, hook=capture_and_forget)
-    def query(self, query, variables=None, skip_validation=False):
+    def query(
+        self, query: str, variables=None, skip_validation=False
+    ) -> Optional[dict[str, Any]]:
+
+        # Here we are recreating client on purpose as some integrations such as `openshift-resource` require multiple
```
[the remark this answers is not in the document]
bhushanthakur93: \o/

reconcile/utils/gql.py
```diff
@@ -1,24 +1,27 @@
...
+from gql import gql, Client
+from gql.transport.requests import RequestsHTTPTransport
+from gql.transport.exceptions import TransportQueryError
+from gql.transport.requests import log as requests_logger
+
+from threading import Lock
 
-_gqlapi = None
 
+_gqlapi = None
```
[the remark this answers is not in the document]
steveteahan: I agree that this could probably be done differently, but it also isn't strictly in the scope of this PR, so maybe we take this up in a separate ticket and PR?

steveteahan (approved):
LGTM

PR #2303, opened 2022-03-29
add unittests for RDS get_replicas_info

geoberle:
a unittest a day keeps the doctor away...
... and makes @Piojo happy


reconcile/test/test_utils_terraform_client.py
```diff
@@ -56,3 +57,99 @@ def test_expiration_value_error(aws_api):
...
+    }
+    result = tfclient.TerraformClient.get_replicas_info([namespace])
+    expected = {"acc": {"replica-id-rds": "replica-source-id-rds"}}
+    assert result == expected
+
+
+def test_get_replicas_info_via_replica_source_overrides_present():
+    # this test shows that the direct replica_source on the tf resource
+    # has precendence over overrides
+    namespace = {
```
Piojo, line 82: Wouldn't it be better to convert these `namespace` declarations into a fixture, like:

```python
@pytest.fixture
def namespace():
   return {
     "terraformResources": [
              ...
     ]
   }


def test_get_replicas_info_via_replica_source_overrides_present(namespace):
     result = tfclient.TerraformClient....

def test_get_replicas_info_via_defaults(mocker, namespace):
    del namespace["terraformResources"][0]["replica_source"]
    result =...

...
```

Piojo (approved):
You, sir, are very brave.

My nit regarding using fixtures is optional. Do it or don't, there is no block.

reconcile/utils/gql.py
```diff
@@ -177,6 +177,10 @@ def init(url, token=None, integration=None, validate_schemas=False):
     return _gqlapi
 
 
+def get_resource(path):
```
steveteahan, line 180: I know this is a very simple function, but without types it's still going to be assumed that the return is `Any`. As such, could we add types here?

steveteahan:
Thanks for adding more tests!

reconcile/test/test_utils_terraform_client.py
geoberle: i've chosen the approach of having the "fixture" defined within the test because it makes reviewing them easier. the entire test with all data is in this case small enough to fit on a small part of the screen and i don't have to think about how the real fixture looks like.

i know i'm repeating myself a bit with redeclaring very similar testdata quite some time. it is always a fine line between DRY and DAMP.

i will keep fixture in mind though for other tests

reconcile/utils/gql.py
geoberle: sure steve [a839526](https://github.com/app-sre/qontract-reconcile/pull/2303/commits/a839526175d688d296ed4dcc0b53160eb4506f2c)

PR #2304, opened 2022-03-30
[resource-scraper] introduce utility integration

maorfr:
this is a utility integration that collects resources from specific namespaces of clusters and stores them in vault.

this utility can be used for different purposes, such as disaster recovery.
the current use case is https://issues.redhat.com/browse/OSD-10326

### DIFF
--- a/helm/qontract-reconcile/values-fedramp.yaml
+++ b/helm/qontract-reconcile/values-fedramp.yaml
@@ -213,4 +213,16 @@ integrations:
   logs:
     slack: false
   disableUnleash: true
-
+cronjobs:
+- name: resource-scraper
+  resources:
+    requests:
+      memory: 200Mi
+      cpu: 200m
+    limits:
+      memory: 400Mi
+      cpu: 300m
+  # once a day at midnight
+  cron: '0 0 * * *'
+  extraArgs: --namespace-name openshift-sre-token-scraper --resource-kind Secret --vault-output-path osd-fedramp/integrations-output
+  disableUnleash: true
--- a/reconcile/resource_scraper.py
+++ b/reconcile/resource_scraper.py
@@ -1,11 +1,12 @@
 import sys
 import logging
+from typing import cast
 
 from reconcile import queries
 
 from reconcile.status import ExitCodes
 from reconcile.utils.oc import OC_Map
-from reconcile.utils.vault import VaultClient
+from reconcile.utils.vault import _VaultClient, VaultClient
 
 
 QONTRACT_INTEGRATION = "resource-scraper"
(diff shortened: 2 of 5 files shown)


maorfr:
error:
```
reconcile/resource_scraper.py:57: error: "VaultClient" has no attribute "write"  [attr-defined]
```

`VaultClient` indeed doesn't have a `write` method, `_VaultClient` has it. what would be the best way to overcome this?
perhaps this?
```
$ git diff
diff --git a/reconcile/utils/vault.py b/reconcile/utils/vault.py
index 4106c2f..2e4b989 100644
--- a/reconcile/utils/vault.py
+++ b/reconcile/utils/vault.py
@@ -301,3 +301,6 @@ class VaultClient:
             return cls._instance
 
         return cls._instance
+
+    def write(cls, *args, **kwargs):
+        return cls._instance.write(*args, **kwargs)
```

maorfr:
mypy cast commit details: https://github.com/python/mypy/issues/1020, https://mypy.readthedocs.io/en/stable/common_issues.html#complex-type-tests

helm/qontract-reconcile/values-fedramp.yaml
```diff
@@ -213,4 +213,16 @@ integrations:
...
+  resources:
+    requests:
+      memory: 200Mi
+      cpu: 200m
+    limits:
+      memory: 400Mi
+      cpu: 300m
+  # once a day at midnight
+  cron: '0 0 * * *'
+  extraArgs: --namespace-name openshift-sre-token-scraper --resource-kind Secret --vault-output-path osd-fedramp/integrations-output
```
dustman9000, line 227: Should the namespace and output path be configurable?

reconcile/resource_scraper.py
```diff
@@ -0,0 +1,58 @@
...
+
+from reconcile.status import ExitCodes
+from reconcile.utils.oc import OC_Map
+from reconcile.utils.vault import _VaultClient, VaultClient
+
+
+QONTRACT_INTEGRATION = "resource-scraper"
+
+
+def run(dry_run, namespace_name, resource_kind, vault_output_path):
```
dustman9000, line 15: Thanks for adding this Maor! Can you also add support for filtering on labels? The way we were scraping the secrets was to filter on the label:
```
oc get secret ... -l app=openshift-sre-token-scraper
```

maorfr: contributions are welcome? :)

TGPSKI: After discussing with Dustin we are going to merge as-is, and Dustin will contribute a patch to add filtering on a later MR.

TGPSKI (approved):

PR #2308, opened 2022-03-30
[template-tester] introduce new integration

maorfr:

depends on https://github.com/app-sre/qontract-schemas/pull/117

this PR adds a new integration called `template-tester` which tests templated resources and verifies that their result (rendered resource) matches an expected result.

the approach of this PR is similar to https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/prometheus_rules_tester.py

### DIFF
--- a/reconcile/template_tester.py
+++ b/reconcile/template_tester.py
@@ -45,8 +45,7 @@ def run(dry_run):
     for tt in template_tests:
         resource = tt["resource"]
         resource["add_path_to_prom_rules"] = False
-        parent = load_resource(tt["resourceParent"])
-        openshift_resource = orb.fetch_openshift_resource(tt["resource"], parent)
+        openshift_resource = orb.fetch_openshift_resource(tt["resource"], tt["resourceParent"])
         expected_result = load_resource(tt["expectedResult"])
         if openshift_resource.body != expected_result:
             logging.error(
--- a/reconcile/template_tester.py
+++ b/reconcile/template_tester.py
@@ -45,7 +45,9 @@ def run(dry_run):
     for tt in template_tests:
         resource = tt["resource"]
         resource["add_path_to_prom_rules"] = False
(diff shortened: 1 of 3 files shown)


reconcile/utils/gql.py
```diff
@@ -140,7 +149,11 @@ def query(self, query, variables=None, skip_validation=False):
         if "data" not in result:
             raise GqlApiError(("`data` field missing from GraphQL" "server response."))
 
-        return result["data"]
+        result_data = result["data"]
+        if self.test_data:
+            self.add_test_data(result_data)
+
+        return result_data
```
geoberle, line 156: so the query is really executed, to make sure the query syntax is correct but to make template tests reproducible, the actual data is replaced by the test data. that is already a lot better than what we do in qontract-reconcile with unit tests running on fixture only

reconcile/template_tester.py
```diff
@@ -0,0 +1,60 @@
...
+
+def run(dry_run):
+    gqlapi = gql.get_api()
+    template_tests = gqlapi.query(TEMPLATE_TESTS_QUERY)["tests"]
+    error = False
+    for tt in template_tests:
+        gqlapi.init_test_data(load_resource(tt["testData"]))
+        resource = tt["resource"]
+        resource["add_path_to_prom_rules"] = False
+        parent = DictObj(load_resource(tt["resourceParent"]))
```
geoberle, line 48: what is the reason for the DictObj?

rporres:
Is this actually a draft or is it meant to be reviewed?

reconcile/utils/gql.py
maorfr: no longer relevant with the new approach

reconcile/template_tester.py
maorfr: no longer relevant with the new approach

maorfr:
now it is ready for review @rporres @geoberle

geoberle (approved):

PR #2312, opened 2022-04-01
Add more context about the failed account for stale terraform errors

steveteahan:
Without this change, we don't see which account is causing issues unless
Exception was raised 3 times (due to retry() usage elsewhere in the code).


OliviaHY:
LGTM. Thank you for improving logging!

Piojo (approved):

PR #2313, opened 2022-04-01
Use a newer unleash client

Piojo:
After merging a newer tzinfo, we are stuck with a dependency chain
that relies on a very old API, causing a lot of messages like

```
/usr/local/lib/python3.9/site-packages/apscheduler/triggers/interval.py:66: PytzUsageWarning: The normalize method is no longer necessary, as this time zone supports the fold attribute (PEP 495). For more details on migrating to a PEP 495-compliant implementation, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html
```

A newer UnleashClient fixes these dependencies and uses non-deprecated
APIs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setup(
         "ruamel.yaml>=0.16.5,<0.17.0",
         "terrascript==0.9.0",
         "tabulate>=0.8.6,<0.9.0",
-        "UnleashClient>=3.4.2,<3.5.0",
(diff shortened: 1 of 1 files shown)


rporres (approved):
From a quick look, I haven't found any of the breaking changes in https://github.com/Unleash/unleash-client-python/releases that should affect us.

I guess testing this in stage should be possible, mainly by merging this and disabling an integration manually.

PR #2319, opened 2022-04-05
Do not crash OCM if storage_quota isn't defined

Piojo:
Clusters created with ROSA lack some fields we rely on, such as `storage_quota`. If they are missing, we simply skip them since app-interface won't handle them anyways.

Refs https://issues.redhat.com/browse/APPSRE-4790

### DIFF
--- a/reconcile/utils/ocm.py
+++ b/reconcile/utils/ocm.py
@@ -1,15 +1,12 @@
 import functools
 import logging
 import re
-from typing import Optional
-import requests
-
-from sretoolbox.utils import retry
-
-from reconcile.utils.secret_reader import SecretReader
+from typing import Any, Optional
 
 import reconcile.utils.aws_helper as awsh
-
+import requests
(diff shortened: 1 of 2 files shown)


reconcile/utils/ocm.py
```diff
@@ -99,14 +97,22 @@ def _init_request_headers(self):
             "accept": "application/json",
         }
 
+    @staticmethod
+    def _ready_for_app_interface(cluster: dict[str, Any]) -> bool:
+        return (
+            cluster["managed"]
+            and cluster["state"] == STATUS_READY
+            and "storage_quota" in cluster
```
geoberle, line 105: i'm ok with excluding rosa cluster temporarily to not fail integrations and PR checks. in a follow up PR this needs to be redefined together with some other changes in ocm.py to handle the ROSA differences properly.

geoberle (approved):

PR #2323, opened 2022-04-06
Fix lint error on lru_cache usage

Piojo:
It warns (rightly) that we can leak memory. A maxsize of 2048 items
looks enough for most cases, and will keep the linter happy.


patjlm (approved):
LGTM
this currently fails all q-r builds

PR #2324, opened 2022-04-06
Speed up GQL-related tests by ~300x

Piojo:
Test directly the relevant function instead of the decorated one.

The retry decorator introduces sleeps that slow us down and we don't
want to test within this file.  More reading:
https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-testing-guide/#testing-decorated-methods

To be merged after #2323


bhushanthakur93:
LGTM!

PR #2325, opened 2022-04-06
Filter out clusters missing load_balancer_quota

Piojo:
Skip from ocm.py clusters withot load_balancer_quota, based on the spec we've gotten in APPSRE-4790.

To be merged after #2323.

### DIFF
--- a/reconcile/test/test_utils_ocm.py
+++ b/reconcile/test/test_utils_ocm.py
@@ -89,11 +89,37 @@ def test_get_cluster_aws_account_id_ok(mocker, ocm):
 @pytest.fixture
 def clusters_by_readiness():
     return [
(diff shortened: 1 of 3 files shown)


lechuk47:
Shouldn't be better to use the product attribute in the spec to filter out non-OSD clusters? I think ROSA won't have storage_quota.

Piojo:
If the end goal is to have app-interface manage these clusters, I don't see us filtering ROSA entirely, more learning to cope with missing fields. Let's get this unblocked to gain some safety on OCM while we get a proper design doc on how we'll diagnose and manipulate these clusters.

maorfr:
it was eventually determined that we will spike what would be the changes needed to support ROSA, and that spike will happen in the OCM stage environment.

closing.

PR #2326, opened 2022-04-07
bump version to release to pypi

maorfr:
i think we'd want to start enforcing that every PR requires a bump of the version.

and after that, probably change the release to pypi from tag based to push based.

cc @fishi0x01 @Piojo


maorfr:
released in https://ci.ext.devshift.net/job/app-sre-qontract-reconcile-gh-build-tag/8/

PR #2327, opened 2022-04-07
force setup.py version bump in PRs

maorfr:
follows up on #2326

this PR will force each future PR to bump the version in setup.py to a higher version.

### DIFF
--- a/reconcile/utils/semver_helper.py
+++ b/reconcile/utils/semver_helper.py
@@ -21,3 +21,7 @@ def sort_versions(versions: Iterable[str]) -> List[str]:
     """
     semver_versions = sorted([parse_semver(v) for v in versions])
     return [str(v) for v in semver_versions]
+
+
+def is_version_bumped(current_version: str, previous_version: str) -> bool:
+    return parse_semver(current_version) > parse_semver(previous_version)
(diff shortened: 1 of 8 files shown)


maorfr:
the approach of adding this test as a test and not a make target or a bash script was inspired by #652

reconcile/utils/semver_helper.py
```diff
@@ -21,3 +21,7 @@ def sort_versions(versions: Iterable[str]) -> List[str]:
     """
     semver_versions = sorted([parse_semver(v) for v in versions])
     return [str(v) for v in semver_versions]
+
+
+def is_version_bumped(current_version: str, previous_version: str) -> bool:
```
fishi0x01, line 26: Where is this used?

maorfr:
depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/36837

fishi0x01 (commented):
Lets highlight that this is a hard requirement now: 

Each PR **must** ... CI tests will fail if version is not bumped.

**EDIT** Ref to code passage wasnt linked :thinking: This should refer to README update

maorfr:
@fishi0x01 done

fishi0x01:
Nice :+1: 
Probably also worth a mention in team chat as this will affect everybody from now on - pretty sure this might lead to confusion initially

fishi0x01 (approved):

maorfr:
release to pypi was a success: https://pypi.org/project/qontract-reconcile/0.5.0/

PR #2329, opened 2022-04-07
fix ready upgrade in qontract-cli cluster-upgrade-policies

patjlm:
When several upgrades where possible, with enough soak days, the wrong one could be selected as the next one being performed (the one with most soak days instead of the highest version matching conditions)

### DIFF
--- a/reconcile/ocm_upgrade_scheduler.py
+++ b/reconcile/ocm_upgrade_scheduler.py
@@ -3,7 +3,7 @@ import logging
 import copy
 
 from datetime import datetime
-from typing import Optional
+from typing import Mapping, Optional
 from dateutil import parser
 from croniter import croniter
 
@@ -177,7 +177,7 @@ def version_conditions_met(version, history, ocm_name, workloads, upgrade_condit
     return conditions_met
 
 
-def upgradeable_version(policy: dict, history: dict, ocm: OCM) -> Optional[str]:
+def upgradeable_version(policy: Mapping, history: Mapping, ocm: OCM) -> Optional[str]:
     """Get the highest next version we can upgrade to, fulfilling all conditions"""
     upgrades = ocm.get_available_upgrades(policy["current_version"], policy["channel"])
     for version in reversed(sort_versions(upgrades)):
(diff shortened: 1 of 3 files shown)


reconcile/ocm_upgrade_scheduler.py
```diff
@@ -176,6 +177,23 @@ def version_conditions_met(version, history, ocm_name, workloads, upgrade_condit
     return conditions_met
 
 
+def upgradeable_version(policy: dict, history: dict, ocm: OCM) -> Optional[str]:
```
fishi0x01, line 180: :heart: that this was extracted to an extra method.

small nit: According to style guide we should favor abstract collection types in func args

```
def upgradeable_version(policy: Mapping, history: Mapping, ocm: OCM) -> Optional[str]:
```

reconcile/ocm_upgrade_scheduler.py
```diff
@@ -251,29 +269,18 @@ def calculate_diff(current_state, desired_state, ocm_map, version_history):
...
+        if version:
+            item = {
+                "action": "create",
+                "cluster": cluster,
+                "version": version,
+                "schedule_type": "manual",
+                "next_run": next_schedule.strftime("%Y-%m-%dT%H:%M:%SZ"),
+            }
+            for mutex in cluster_mutexes(d):
+                locked[mutex] = cluster
```
fishi0x01, line 282: Out of curiosity: Doesnt that mean that any already held mutex for that cluster gets overridden? The logic is consistent with the current master state, however, it kinda puzzles me :sweat_smile: Would have thought that if the mutex alrdy exists, there will be a skip.

fishi0x01 (approved):
I like it :) 
But I think Im lacking base knowledge about the overall context - not sure how valuable my review is here.

reconcile/ocm_upgrade_scheduler.py
patjlm: This locks all mutexes for that cluster. The check if the mutexes are locked is at line 264

reconcile/ocm_upgrade_scheduler.py
patjlm: thanks, fixed in 42b01ae

PR #2332, opened 2022-04-12
Quay error on repo actions when managedRepos is set to false

maorfr:
part of https://issues.redhat.com/browse/ASIC-208

we currently fail silently when we should error out.

### DIFF
--- a/reconcile/quay_repos.py
+++ b/reconcile/quay_repos.py
@@ -12,6 +12,7 @@ from reconcile.status import ExitCodes
 QUAY_REPOS_QUERY = """
 {
   apps: apps_v1 {
+    name
     quayRepos {
       org {
         name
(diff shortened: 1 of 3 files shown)


maorfr:
depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/37084

reconcile/quay_repos.py
```diff
@@ -75,11 +76,15 @@ def fetch_desired_state(quay_api_store):
...
 
         for quay_repo in quay_repos:
+            org_name = quay_repo["org"]["name"]
             if not quay_repo["org"]["managedRepos"]:
-                continue
+                logging.error(
+                    f"[{app['name']}] Can not manage repos in {org_name} "
+                    "since managedRepos is set to false."
+                )
+                sys.exit(ExitCodes.ERROR)
```
geoberle, line 85: could you continue processing the rest of the repos with the usual `error = True` strategy and then fail at the end?

maorfr: This integration is quite ancient and is not working this way

geoberle (approved):

PR #2333, opened 2022-04-12
Increase quay permissions memory limit

lechuk47:
Memory limit increase to 300Mb, right now the ingration is using about 200Mb


maorfr:
you'll need to go through the helm chart and `make generate`. you'll get a failure in the build because of that

maorfr (approved):

maorfr:
kind reminder to bump the version ;)

PR #2334, opened 2022-04-12
Using MRState values instead of strings

apahim:


maorfr (approved):
you'll need to bump the version in setup.py

PR #2335, opened 2022-04-13
[PROPOSAL] versioning automation

patjlm:
Releasing proposal
Why:
- avoid merge conflicts due to manual version bumping
- version each commit automatically
- simple X.Y.Z version management: simply git tag & push.

Challenges
- Python does not strictly follow Semver: https://peps.python.org/pep-0440/
- docker image tags don't support `+` signs, so dropping it there
- we run setup.py without git repos in several cases (tox does sdist and then setup again, docker builds don't include .git, rightfully). So the git version is first saved as a file GIT_VERSION.

### DIFF
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -8,8 +8,4 @@ docker --config="$DOCKER_CONF" login -u="$QUAY_USER" -p="$QUAY_TOKEN" quay.io
 make build push
 
 # publish to pypi
-set -e
-
-python3 -m pip install --user twine wheel
-python3 setup.py bdist_wheel
-python3 -m twine upload dist/*
+./build_tag.sh
--- a/build_tag.sh
+++ b/build_tag.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+python3 -m pip install --user twine wheel
+python3 setup.py bdist_wheel
+python3 -m twine upload dist/*
--- a/release/version.py
+++ b/release/version.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+from typing import Optional
+import re
+import subprocess
+from subprocess import PIPE
+
+GIT_VERSION_FILE = "GIT_VERSION"
+
+
+def git() -> str:
+    """get the version from git. Can be
+    - X.Y.Z if a tag is set on the current HEAD
+    - X.Y.Z-<count>-g<commitid> otherwise where
+        - X.Y.Z is the latest version tag found
+        - <count> is the number of commits since then
+        - <commitid> is the current HEAD commitid
+    """
+    cmd = "git describe --tags --match=[0-9]*.[0-9]*.[0-9]*"
+    try:
+        p = subprocess.run(cmd.split(" "), stdout=PIPE, stderr=PIPE, check=True)
+        v = p.stdout.decode("utf-8").strip()
+        # tox is running setup.py sdist from the git repo, and then runs again outside
+        # of the git repo. At this second step, we cannot run git commands.
+        # So we save the git version in a file and include it in the source distribution
+        with open(GIT_VERSION_FILE, "w") as f:
+            f.write(v)
+        return v
+    except subprocess.CalledProcessError as e:
+        # if we're not in a git repo, try reading out from the GIT_VERSION file
+        if os.path.exists(GIT_VERSION_FILE):
+            with open(GIT_VERSION_FILE, "r") as f:
+                return f.read()
+        print(e.stderr)
+        raise e
+
+
+def commit(length: int = 7) -> str:
+    """get the current git commitid"""
+    cmd = f"git rev-parse --short={length} HEAD"
+    p = subprocess.run(cmd.split(" "), stdout=PIPE, stderr=PIPE, check=True)
+    return p.stdout.decode("utf-8").strip()
+
+
+def semver(git_version: Optional[str] = None) -> str:
+    """get a semantic version out of the input git version (see git())
+    - if a X.Y.Z tag is set on the current HEAD, we'll use this
+    - else we'll use X.Y.<Z+1>-<count>+<commitid> to respect semver and version
+      ordering. <count> is the prerelease id, <commitid> is a buildinfo
+    """
+    v = git_version or git()
+    m = re.match(r"(\d+)\.(\d+)\.(\d+)-(\d+)-g(.+)", v)
+    if m:
+        major, minor, patch, prerelease, buildinfo = m.groups()
+        # semver prerelase are supposed to show build increments *prior* to a release.
+        # So we're bumping the patch number to show what the next release would be.
+        # this allows correct version ordering
+        patch = str(int(patch) + 1)
+        # X.Y.Z-<count>-g<commitid> is not a valid as <count>-g<commitid> would be
+        # compared as a string, leading to (count=50) < (count=6)
+        # X.Y.Z-<count>+<commitid> is valid .
+        #   <count> is then the prerelease field, treated as a numeric.
+        #   <commitid> is then a buildinfo string, not used in version comparisons
+        v = f"{major}.{minor}.{patch}-{prerelease}+{buildinfo}"
+    return str(v)
+
+
+def pip(git_version: Optional[str] = None) -> str:
+    """get a pip version out of the input git version (see git()),
+    according to https://peps.python.org/pep-0440/
+    - if a X.Y.Z tag is set on the current HEAD, we'll use this
+    - else we'll use X.Y.(Z+1).pre<count>+<commitid> to respect PEP-0440 versioning
+    """
+    return semver(git_version).replace("-", ".pre")
+    # Alternatively, use postreleases
+    # v = git_version or git()
+    # m = re.match(r"(\d+)\.(\d+)\.(\d+)-(\d+)-g(.+)", v)
+    # if m:
+    #     major, minor, patch, count, commitid = m.groups()
+    #     v = f"{major}.{minor}.{patch}.post{count}+{commitid}"
+    # return str(v)
+
+
+def docker(git_version: Optional[str] = None) -> str:
+    # docker tags don't like '+' characters, let's remove the buildinfo/commitid
+    return pip(git_version).split("+", maxsplit=1)[0]
+
+
+if __name__ == "__main__":
+    type_param = sys.argv[1] if len(sys.argv) > 1 else "--git"
+    v = {
+        "--git": git(),
+        "--commit": commit(),
+        "--semver": semver(),
+        "--pip": pip(),
+        "--docker": docker(),
+    }[type_param]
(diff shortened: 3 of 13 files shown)


patjlm:
The build will fail here until we push a tag on the latest version committed manually

release/version.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+from typing import Optional
+import re
+import subprocess
+
+git_version_file = "GIT_VERSION"
```
Piojo, line 9: Constants are usually in ALL_CAPS.

Piojo, line 9:
```suggestion
GIT_VERSION_FILE = "GIT_VERSION"
```

Piojo (approved):
imagen

release/version.py
patjlm: done

patjlm: sure, done

setup.py
```diff
@@ -57,6 +59,9 @@
         "transity-statuspageio>=0.0.3,<0.1",
         "pydantic~=1.9.0",
         "MarkupSafe==2.1.1",
+        # this is really needed only in lint and type validations.
+        # Is there any better place to put this in?
+        "packaging~=21.3",
```
patjlm, line 64: open to advices on how to handle this dependency better..

rporres:
I like the idea of automating this process as much as possible without any other consideration.

I like the idea of using tags for this, as they're usually linked to versions.

Is this enough or do we need to do something to push when we have a tag?

patjlm:
> Is this enough or do we need to do something to push when we have a tag?

We will need to reinstate the build-tag build that @maorfr removed here: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/36837/diffs

At that point, when a tag is set, that build will trigger and call `build-tag.sh`. This will simply generate a new pypi version, on the same commitid, with a `X.Y.Z` version name.

As discussed during the EMEA call today, we will not tag docker images with these version names. These images are currently only used with the commitid or with `latest`

setup.py
fishi0x01: Does one of the requirements files work? https://github.com/app-sre/qontract-reconcile/tree/master/requirements

They are used by tox envs

release/version.py
```diff
@@ -0,0 +1,99 @@
...
+    try:
+        p = subprocess.run(cmd.split(" "), capture_output=True, check=True)
+        v = p.stdout.decode("utf-8").strip()
+        # tox is running setup.py sdist from the git repo, and then runs again outside
+        # of the git repo. At this second step, we cannot run git commands.
+        # So we save the git version in a file and include it in the source distribution
+        with open(GIT_VERSION_FILE, "w") as f:
+            f.write(v)
+        return v
+    except subprocess.CalledProcessError as e:
```
janboll, line 30: Not 100% sure if I am right here, but given the comment regarding calling tox twice, this looks like you are using exceptions for flow control. I would prefer having the read operation outside the except block since you can clearly specify the conditions for when it should be called.

release/version.py
```diff
@@ -0,0 +1,99 @@
...
+        # if we're not in a git repo, try reading out from the GIT_VERSION file
+        if os.path.exists(GIT_VERSION_FILE):
+            with open(GIT_VERSION_FILE, "r") as f:
+                return f.read()
+        print(e.stderr)
+        raise e
+
+
+def commit(length=7) -> str:
```
janboll, line 39: Is this missing a type for parameter length? Or is this valid for the type checker? Not 100% sure

setup.py
patjlm: it seems a single one is not sufficient for both types of validations.. i'd need to double check

release/version.py
patjlm: not really: i always want to prefer the output of `git describe` than reading the file. How would you specify the conditions (i thought of subprocessing `git rev-parse`, but in the end it's the same than here..
The alternative could be `check=False` and check the `errorcode`. Is that what you think about ?

patjlm: or ?
```python
    except subprocess.CalledProcessError as e:
        if not os.path.exists(GIT_VERSION_FILE):
            print(e.stderr)
            raise e
    with open(GIT_VERSION_FILE, "r") as f:
        return f.read()
```

release/version.py
patjlm: can add ..

release/version.py
janboll: Hm, perhaps check if it's a git repository and if not try reading the file? Bash example for checking if dir is a git repo:  
```
if [ -d .git ]; then
  echo .git;
else
  git rev-parse --git-dir 2> /dev/null;
fi;
```

setup.py
patjlm: No, updating `requirements-test` only fails the `type` check:
```
type create: /Users/patmarti/dev/app-sre/qontract-reconcile/.tox/type
type installdeps: -r/Users/patmarti/dev/app-sre/qontract-reconcile/requirements/requirements-type.txt
type inst: /Users/patmarti/dev/app-sre/qontract-reconcile/.tox/.tmp/package/4/qontract-reconcile-0.6.4rc2+f8028f2.zip
type installed: anymarkup==0.8.1,anymarkup-core==0.8.1,appdirs==1.4.4,APScheduler==3.9.1,bcrypt==3.2.0,boto3==1.18.0,boto3-stubs==1.21.39,botocore==1.21.0,botocore-stubs==1.24.39,cachetools==5.0.0,certifi==2021.10.8,cffi==1.15.0,chardet==3.0.4,click==8.1.2,configobj==5.0.6,croniter==1.0.15,cryptography==36.0.2,decorator==5.1.1,defusedxml==0.7.1,Deprecated==1.2.13,dnspython==2.2.1,dyn==1.8.1,fasteners==0.17.3,fcache==0.4.7,google-auth==2.6.4,gql==3.1.0,graphql-core==3.2.1,hvac==0.7.2,idna==2.8,importlib-metadata==4.11.3,jenkins-job-builder==3.12.0,Jinja2==3.1.1,jira==3.1.1,jmespath==0.10.0,json5==0.9.6,jsonpath-rw==1.4.0,keyring==23.5.0,kubernetes==12.0.1,ldap3==2.9.1,MarkupSafe==2.1.1,mmh3==3.0.0,multi-key-dict==2.0.3,multidict==6.0.2,munch==2.5.0,mypy==0.942,mypy-boto3-ec2==1.21.39,mypy-boto3-iam==1.21.34,mypy-boto3-rds==1.21.34,mypy-boto3-route53==1.21.34,mypy-boto3-s3==1.21.34,mypy-extensions==0.4.3,oauthlib==3.2.0,openshift==0.13.1,paramiko==2.10.3,parse==1.18.0,pbr==5.8.1,ply==3.11,prometheus-client==0.14.1,pyasn1==0.4.8,pyasn1-modules==0.2.8,pycparser==2.21,pydantic==1.9.0,PyGithub==1.55,PyJWT==2.3.0,PyNaCl==1.5.0,pyOpenSSL==21.0.0,pypd==1.1.0,python-dateutil==2.8.2,python-gitlab==1.11.0,python-http-client==3.3.7,python-jenkins==1.7.0,python-json-logger==2.0.2,python-string-utils==1.0.0,python-terraform==0.10.1,pytz==2022.1,pytz-deprecation-shim==0.1.0.post0,PyYAML==5.4.1,qontract-reconcile @ file:///Users/patmarti/dev/app-sre/qontract-reconcile/.tox/.tmp/package/4/qontract-reconcile-0.6.4rc2%2Bf8028f2.zip,requests==2.22.0,requests-oauthlib==1.3.1,requests-toolbelt==0.9.1,rsa==4.8,ruamel.yaml==0.17.21,ruamel.yaml.clib==0.2.6,s3transfer==0.5.2,semver==2.13.0,sendgrid==6.4.8,sentry-sdk==0.20.3,six==1.16.0,slack-sdk==3.15.2,sretoolbox==1.2.4,sshtunnel==0.4.0,starkbank-ecdsa==2.0.3,stevedore==3.5.0,tabulate==0.8.9,terrascript==0.9.0,toml==0.10.2,tomli==2.0.1,transity-statuspageio==0.0.3,types-click==7.1.8,types-croniter==1.0.8,types-cryptography==3.3.19,types-Jinja2==2.11.9,types-MarkupSafe==1.1.10,types-mock==4.0.13,types-pyOpenSSL==22.0.1,types-python-dateutil==2.8.10,types-PyYAML==6.0.5,types-requests==2.27.16,types-setuptools==57.4.12,types-tabulate==0.8.6,types-toml==0.10.4,types-urllib3==1.26.11,typing_extensions==4.1.1,tzdata==2022.1,tzlocal==4.2,UnleashClient==5.1.2,urllib3==1.25.11,websocket-client==0.54.0,wrapt==1.14.0,xmltodict==0.12.0,yarl==1.7.2,zipp==3.8.0
type run-test-pre: PYTHONHASHSEED='1176546710'
type run-test: commands[0] | mypy
reconcile/test/test_version_bump.py:6: error: Cannot find implementation or library stub for module named "packaging.version"  [import]
reconcile/test/test_version_bump.py:6: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
reconcile/test/test_version_bump.py:6: error: Cannot find implementation or library stub for module named "packaging"  [import]
Found 2 errors in 1 file (checked 275 source files)
ERROR: InvocationError for command /Users/patmarti/dev/app-sre/qontract-reconcile/.tox/type/bin/mypy (exited with code 1)

============================================================================================================================================================================ log end ============================================================================================================================================================================
✖ FAIL type in 1 minute, 39.315 seconds
```

release/version.py
patjlm: so in the end we'd run a subcommand and check the outcome anyways. Is it really better ?

rporres (approved):

geoberle (approved):
i like the approach a lot!

build_deploy.sh
```diff
@@ -8,8 +8,4 @@ docker --config="$DOCKER_CONF" login -u="$QUAY_USER" -p="$QUAY_TOKEN" quay.io
 make build push
 
 # publish to pypi
-set -e
-
-python3 -m pip install --user twine wheel
-python3 setup.py bdist_wheel
-python3 -m twine upload dist/*
+./build_tag.sh
```
maorfr, line 11: :heart:

build_tag.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
maorfr, line 1: :heart:

maorfr (approved):

maorfr:
depends on https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/37249

setup.py
fishi0x01: And updating `requirements-test.txt` and `requirements-type.txt`? The type stage of tox only uses `requirements-type.txt` as dep file.

PR #2336, opened 2022-04-13
[openshift-resources] validate managed resource names

maorfr:
this PR applies to all openshift-resources based integrations:
- openshift-resources
- openshift-vault-secrets
- openshift-routes

when a `managedResourceNames` section is defined, we should validate that all desired items under `openshiftResources` are one of the list.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -147,7 +147,7 @@ def init_specs_to_fetch(
                 )
                 state_specs.append(c_spec)
                 managed_types.remove(kind)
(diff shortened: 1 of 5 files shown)


maorfr:
this PR will validate that a MR such as this one fails in dry-run: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/37177

Piojo:
The substance of the change looks good to me. Can we please add some tests to `ResourceInventory` for good measure? At least `add_desired` is totally pure, and doesn't need any mocking.

PR #2339, opened 2022-04-14
APPSRE-3556 remove saas-file-1 logic

fishi0x01:
**Context**

Remove saas-file-1 logic. We only support saas-file-2.

**Test**

Tested integrations with `--dry-run` mentioned in https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/saas-walkthrough.md with app-interface prod data

- openshift-saas-deploy-trigger-upstream-jobs
- openshift-saas-deploy-trigger-moving-commits
- openshift-saas-deploy-trigger-cleaner
- openshift-saas-deploy-trigger-configs


reconcile/utils/saasherder.py
```diff
@@ -1451,26 +1450,7 @@ def get_saas_targets_config(self, saas_file):
...
-            # for backwards compatibility
-            pipelines_provider = {
-                "provider": Providers.JENKINS,
-                "instance": saas_file["instance"],
-            }
-        if saas_file_api_version == "v2":
-            pipelines_provider = saas_file["pipelinesProvider"]
-
-        return pipelines_provider
+        return saas_file["pipelinesProvider"]
```
steveteahan, line 1453: If we're going to remove the docstring, which probably makes sense given the size of the method, then worth moving the type info here to a type hint?

Piojo (approved):
Please add the type hint @steveteahan mentioned, but this looks great. Thanks for the huge clean-up.

reconcile/utils/saasherder.py
fishi0x01: Absolutely! :+1: 

Fixed

PR #2342, opened 2022-04-14
[sql-query] try not to use latest

maorfr:

the sql-query integrations creates Jobs (or CronJobs) in a tenant's namespace to perform a query against their database.
the integration currently tries to use the same image of the database engine, by looking for the version in the `defaults` file. however, the version may not be there, and may be in the `overrides` section.

this PR makes it so the integration searches for the version in the `overrides` section as well, by re-using the logic from terrascript_client:
https://github.com/app-sre/qontract-reconcile/blob/02252e58cf1f4579ffd3815ac12485cab68c17d2/reconcile/utils/terrascript_client.py#L3075-L3113

this will help us in running less with `latest`.

as a result, and this is why the linked ticket is about DVO, this will help remove `deployment_validation_operator_latest_tag` DVO alerts from tenants' namespaces, enabling https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/37255

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -1437,6 +1437,7 @@ APP_INTERFACE_SQL_QUERIES_QUERY = """
         provider
         ... on NamespaceTerraformResourceRDS_v1
         {
+          account
           identifier
           output_resource_name
           defaults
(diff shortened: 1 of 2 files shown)


geoberle (commented):
looks solid. i will lgtm after the announced unittests

rporres:
indeed, looks good. Waiting for those tests <3

maorfr:
@geoberle @rporres updated the PR to use terrascript, where all the required logic already exists (thus not requiring new tests :yay:)

geoberle (approved):
LGTM. even more concise now!

PR #2345, opened 2022-04-19
Create the IAM user even when only user_policies have been defined

steveteahan:
The IAM user creation was only accounted for in the handling of
aws_groups, so it was skipped when only user_policies were defined.

I need to run through a couple of manual tests, but this was discovered when a user that had a role attached that only defined `user_polices`. The error seen was:

```
12:46:58 [2022-04-19 16:46:58] [ERROR] [terraform_client.py:check_output:559] - [account-name - plan] A managed resource "aws_iam_user" "[redacted]" has not been declared in the root
```

Adding this code unblocked the dry-run because the IAM user resource is now being added.

I also tested:
* add a new user with a role assigned containing both `aws_groups` and `user-policies`
* print-to-file before and after included the same output (with the exception of location that the `aws_iam_user` block is on, which didn't produce any diff in dry-run)

### DIFF
--- a/reconcile/utils/terrascript_client.py
+++ b/reconcile/utils/terrascript_client.py
@@ -500,8 +500,11 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
                     policy = \
                         policy.replace('${aws:accountid}', account_uid)
 
-                    # Ref: terraform aws_iam_policy
+                    # Ref: terraform aws iam_user
                     tf_iam_user = self.get_tf_iam_user(user_name)
+                    self.add_resource(account_name, tf_iam_user)
+
+                    # Ref: terraform aws_iam_policy
                     identifier = f'{user_name}-{policy_name}'
                     tf_aws_iam_policy = aws_iam_policy(
                         identifier,


Piojo (approved):

maorfr:
will a user be able to login without membership to a group that grants the IAMUserChangePassword policy?

steveteahan:
> will a user be able to login without membership to a group that grants the IAMUserChangePassword policy?

As mentioned on Slack, this can also be configured at the [account level](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_passwords_enable-user-change.html) as well. The primary motivation here is that [our docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_passwords_enable-user-change.html) suggest that roles require at least one of `aws_groups` or `user_policies`. The reality is that if you only provide `user_policies`, then you get the error in the PR description.

In theory you could have a user policy that allows changing password, or if we're not making specific assumptions, then this could be configured at the account level. Requiring `aws_groups` still wouldn't guarantee the group grants `IAMUserChangePassword`?

steveteahan:
This doesn't seem to be blocking anything for now. With that being said, it results in an error message that wasn't obvious to multiple team members. If we'd prefer to address requiring `IAMUserChangePassword` policy access, and we only want that to come from a group, then I think that we can close this for now.

steveteahan:
Closing this because it may just make it so that users can be created without the minimal permissions required to actually make the account usable. This may just be an edge-case, so for now I'll close this and we can revisit it if we determine it's impactful in the future.

PR #2346, opened 2022-04-20
disable DVO minimum-three-replicas check for integrations

maorfr:

docs: https://www.civo.com/learn/yaml-best-practices-using-kubelinter#ignoring-a-check

this PR disables the DVO check of minimum 3 replicas for integrations, as our scaling mechanism is shards and not replicas.


geoberle (approved):
LGTM

PR #2347, opened 2022-04-20
run dashdotdb-dvo once an hour

maorfr:
as we work on https://issues.redhat.com/browse/APPSRE-4765, we will get more traction on DVO metrics.

this means tenants would want to see their work reflected in dashboards.

with this job running once every 24 hours, the feedback loop will be too slow.

is there a reason not to run this job once an hour?


maorfr:
@bkez322 please review

bkez322 (approved):
Makes sense. I expect this to be fine.

I believe these were set to "daily" only because the other dashdotdb jobs already were.

PR #2349, opened 2022-04-20
consider shared resources in tester integrations

maorfr:

this PR adds the option to test resources referenced by shared resources files referenced from namespace files. without this change, only resources defined in a namespace's `openshiftResources` section can be tested.

this is the function being used:
https://github.com/app-sre/qontract-reconcile/blob/a3285f140702cdd9386a4faa7357b0c7ac149af5/reconcile/openshift_base.py#L775-L799

this PR enables https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/37591

### DIFF
--- a/reconcile/prometheus_rules_tester.py
+++ b/reconcile/prometheus_rules_tester.py
@@ -10,6 +10,7 @@ from sretoolbox.utils import threaded
 from reconcile import queries
 from reconcile.utils import gql
 from reconcile.utils import promtool
+import reconcile.openshift_base as ob
 import reconcile.openshift_resources_base as orb
 
 from reconcile.utils.semver_helper import make_semver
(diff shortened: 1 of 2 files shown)


reconcile/prometheus_rules_tester.py
```diff
@@ -77,6 +78,7 @@ def get_prometheus_rules(cluster_name):
         ):
             continue
 
+        ob.aggregate_shared_resources(n, "openshiftResources")
```
steveteahan, line 81: nit: `"openshiftResources"` is repeated enough that it could probably be a constant.

steveteahan, line 81: nit: `"openshiftResources" is repeated enough that it could probably be a constant.

steveteahan (approved):

steveteahan (approved):

reconcile/prometheus_rules_tester.py
maorfr: i'll keep it for a future iteration

PR #2353, opened 2022-04-22
More files under black formatting

fishi0x01:
Adding more files to the black formatter.

Those files do not subject to any open recent PR - so should not cause too much pain :crossed_fingers:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,6 +4,23 @@ target-version = ['py39']
 include = '^.*\.py$'
 extend-exclude = '''
 ^/(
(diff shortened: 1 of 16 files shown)


geoberle:
please exclude terraform_resources.py and terraform_client.py. this will conflict with my ongoing PR. i will make sure to format them before i merge

fishi0x01:
@geoberle added ` utils/terraform_resources.py` and `terraform_client.py` to exclusion list

geoberle (approved):

PR #2356, opened 2022-04-24
[openshift-resources + saas] allow full-qualified kinds in managedResourceTypes

geoberle:
# What changed?

## Core Change idea
The core change of this PR enables the use of full qualified kube `kinds` (resource types) in `managedResourceTypes` of `saas` files and `namespace` files. This replaces the need for `managedResourceTypeOverrides` in saas files and enables resource handling in saas files where the proposed `kind` is not unique without the `group`.

https://issues.redhat.com/browse/APPSRE-4915

e.g. instead of

```yaml
managedResourceTypeOverrides:
- resource: Subscription
  override: Subscription.operators.coreos.com

managedResourceTypes:
- Subscription
```

this change enables the declaration to look like this

```yaml
managedResourceTypes:
- Subscription.operators.coreos.com
```

Both declarations will lead to the same results but the latter one also enables saas files and namespaces files to managed same-named kinds from different groups, e.g.

```yaml
managedResourceTypes:
- Subscription.operators.coreos.com
- Subscription.open-cluster-management.io
```

Mixed declarations of short and full-qualified kinds are supported for backwards compatibility

```yaml
managedResourceTypes:
- Subscription.operators.coreos.com
- Deployment
```

## Core change implementation details
* in both cases of either a defined override or full-qualified resource type declaration, the kind tracked within the `StateSpec` and `ResourceInventory` will be the full-qualified type
* as such, the `kind` passed to `ResourceInventory.add_current` and `add_desires` need to match the registered `kind` in the inventory
* a convenience method `add_desired_resource` has been added to `ResourceInventory` to automatically use the short or full-qualified `kind` depending on how the `kind` was registered with the inventory in the first place

## Companion changes to support the core change
To make the implementation of the core change easier, a couple of additional changes have been introduced, mostly to make it easier to reason about the data passed between code components. The idea of those changes therefore is:
* have type hints for all relevant parts
* know which fields are nullable and which ones can be relied on

### StateSpec
StateSpec management for openshift resources was refactored into dataclasses. the previous `StateSpec` class tried to serve the purpose of current and desired state at the same, while having different sets of fields only relevant to one of both cases. The new `BaseStateSpec` holds the fields common to current and desired spec, and the respective subclasses cover each case with strongly typed sets of fields where the expectations of what field can be None is explicitely defined.

### OC_Map get_cluster
A new method `get_cluster` was introduced on the OC_Map to have a typehintable version of the `get` method. The `get` method returned either an OC client (one of `OCDeprecated` or `OCNative`) or an `OCLogMsg` object indicating an error and also acting as a boolean so it could be used in places to check for non existing clients or error situations with a `not oc`. This is non-obvious magic going on and  `not oc` vs `oc is not None` behave differently which is very subtle and a trap for developers.

The `get_cluster` method always deliveres a valid oc client or raises `OCLogMsg` as an exception, which can be dealth with directly. The `get` method has not been removed by this change and only the code in the scope of this change was adapted to leverage `get_cluster`.

### Tests
A lot of tests have been introduced for code that was touched. Additionally `TestSlide` tests have been migrated to `pytest` where feasible.

# How was the change tested?

Besides extensive addition of unittests, most integrations dealing with `openshift_base.init_specs_to_fetch` have been dry-runned and results have been verified.
Not all integrations have been tested though. This PR was opened anyways to gather feedback during the testing process.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -1,7 +1,8 @@
+from dataclasses import dataclass
 import logging
 import itertools
 
-from typing import Optional, Iterable, Mapping
+from typing import Any, Optional, Iterable, Mapping, Union
 
 import yaml
 
@@ -9,7 +10,7 @@ from sretoolbox.utils import retry
 from sretoolbox.utils import threaded
 
 from reconcile import queries
-from reconcile.utils.oc import FieldIsImmutableError
+from reconcile.utils.oc import FieldIsImmutableError, OCClient
 from reconcile.utils.oc import MayNotChangeOnceSetError
 from reconcile.utils.oc import PrimaryClusterIPCanNotBeUnsetError
 from reconcile.utils.oc import InvalidValueApplyError
@@ -34,28 +35,41 @@ class ValidationErrorJobFailed(Exception):
     pass
 
 
-class StateSpec:
-    def __init__(
-        self,
-        type,
-        oc,
-        cluster,
-        namespace,
-        resource,
-        parent=None,
-        resource_type_override=None,
-        resource_names=None,
-        privileged=False,
-    ):
-        self.type = type
-        self.oc = oc
-        self.cluster = cluster
-        self.namespace = namespace
-        self.resource = resource
-        self.parent = parent
-        self.resource_type_override = resource_type_override
-        self.resource_names = resource_names
-        self.privileged = privileged
+@dataclass
+class BaseStateSpec:
+
+    oc: OCClient
+    cluster: str
+    namespace: str
+
+    def __str__(self) -> str:
+        return f"cluster: {self.cluster}, namespace: {self.namespace}"
+
+
+@dataclass
+class CurrentStateSpec(BaseStateSpec):
+
+    kind: str
+    resource_names: Optional[Iterable[str]]
+
+    def __str__(self) -> str:
+        return (
+            f"current spec - {super().__str__()}, resource_names: {self.resource_names}"
+        )
+
+
+@dataclass
+class DesiredStateSpec(BaseStateSpec):
+
+    resource: Mapping[str, Any]
+    parent: Mapping[Any, Any]
+    privileged: bool = False
+
+    def __str__(self) -> str:
+        return f"desired spec - {super().__str__()}, privileged: {self.privileged}"
+
+
+StateSpec = Union[CurrentStateSpec, DesiredStateSpec]
 
 
 def init_specs_to_fetch(
@@ -66,7 +80,7 @@ def init_specs_to_fetch(
     override_managed_types: Optional[Iterable[str]] = None,
     managed_types_key: str = "managedResourceTypes",
 ) -> list[StateSpec]:
-    state_specs = []
+    state_specs: list[StateSpec] = []
 
     if clusters and namespaces:
         raise KeyError("expected only one of clusters or namespaces.")
@@ -96,9 +110,7 @@ def init_specs_to_fetch(
                 namespace_info.get("managedResourceTypeOverrides") or []
             )
 
-            # Initialize current state specs
-            for resource_type in managed_types:
-                ri.initialize_resource_type(cluster, namespace, resource_type)
+            # Prepare resource names
             resource_names = {}
             resource_type_overrides = {}
             for mrn in managed_resource_names:
@@ -118,6 +130,7 @@ def init_specs_to_fetch(
                         f"{cluster}/{namespace} (valid kinds: {managed_types})"
                     )
 
+            # Prepare type overrides
             for o in managed_resource_type_overrides:
                 # Current implementation guarantees only one
                 # override of each managed type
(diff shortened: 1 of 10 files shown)


reconcile/openshift_resources_base.py
```diff
@@ -694,12 +683,7 @@ def fetch_states(spec, ri):
 
     except StatusCodeError as e:
         ri.register_error(cluster=spec.cluster)
-        msg = "cluster: {},"
-        msg += "namespace: {},"
-        msg += "resource_names: {},"
-        msg += "exception: {}"
-        msg = msg.format(spec.cluster, spec.namespace, spec.resource_names, str(e))
-        logging.error(msg)
+        logging.error(f"{spec} - exception: {str(e)}")
```
Piojo, line 686: Dataclasses FTW!!

reconcile/utils/openshift_resource.py
```diff
@@ -178,6 +178,10 @@ def name(self):
     def kind(self):
         return self.body["kind"]
 
+    @property
+    def kind_and_group(self):
+        return full_qualified_kind(self.kind, self.body["apiVersion"])
```
Piojo, line 183: Typo: `fully_qualified_kind` ?

reconcile/utils/openshift_resource.py
```diff
@@ -512,6 +524,26 @@ def initialize_resource_type(self, cluster, namespace, resource_type):
             resource_type, {"current": {}, "desired": {}, "use_admin_token": {}}
         )
 
+    def add_desired_resource(
+        self,
+        cluster: str,
+        namespace: str,
+        resource: OpenshiftResource,
+        privileged: bool = False,
+    ):
```
Piojo, line 533: ` -> None:`

reconcile/utils/openshift_resource.py
```diff
@@ -498,6 +502,14 @@ def calculate_sha256sum(body):
         return m.hexdigest()
 
 
+def full_qualified_kind(kind: str, api_version: str) -> str:
```
Piojo, line 505: This is probably `fully_qualified_kind`?

reconcile/openshift_base.py
```diff
@@ -135,94 +149,109 @@ def init_specs_to_fetch(
...
-                    oc,
-                    cluster,
-                    namespace,
-                    t,
-                    resource_type_override=resource_type_overrides.get(t),
-                    resource_names=None,
+            # Initialized current state specs
+            for kind in managed_types:
+                managed_names = resource_names.get(kind)
+                kind = resource_type_overrides.get(kind, kind)
```
Piojo, line 155: I'd rather not override the loop variable, TBH.

reconcile/openshift_base.py
```diff
@@ -135,94 +149,109 @@ def init_specs_to_fetch(
...
+                        namespace=namespace,
+                        kind=kind,
+                        resource_names=[],
+                    )
+                )
                 # Initialize desired state specs
-                d_spec = StateSpec("desired", oc, cluster, namespace, resource_type)
-                state_specs.append(d_spec)
+                # it seems this StateSpec has no effect and can be disabled. the only code path
+                # leading to this place is from openshift_clusterrolebindings > ob.fetch_current_state >
```
Piojo, line 211: This is some serious and brave digging. I really hope we can kill this and make the `cluster` sub-branch trivial.

reconcile/openshift_resources_base.py
```diff
@@ -569,54 +569,46 @@ def fetch_openshift_resource(resource, parent):
...
+    namespace: str,
+    kind: str,
+    resource_names=Iterable[str],
 ):
     global _log_lock
 
-    resource_type_to_use = resource_type_override or resource_type
-
-    msg = "Fetching {}s from {}/{}".format(resource_type_to_use, cluster, namespace)
+    msg = "Fetching {}s from {}/{}".format(kind, cluster, namespace)
```
Piojo, line 581: Prefer f-strings here.

Piojo (commented):
Looks great.

Piojo (commented):
Can we have some YAML fixture with the new capabilities you are proposing? This is a complicated beast, and I'd like to have some code (and reference) on valid and invalid combinations of `managedResourceTypes` and `managedResourceTypeOverrides`.

Synthetic pytest fixtures aren't giving me the full understanding, TBH.

reconcile/openshift_base.py
```diff
@@ -34,28 +35,41 @@ class ValidationErrorJobFailed(Exception):
...
-        self.namespace = namespace
-        self.resource = resource
-        self.parent = parent
-        self.resource_type_override = resource_type_override
-        self.resource_names = resource_names
-        self.privileged = privileged
+@dataclass
+class BaseStateSpec:
+
+    oc: OCClient = field(compare=False)
```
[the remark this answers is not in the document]
geoberle: good point thx

reconcile/utils/openshift_resource.py
geoberle: [2bbbf5e](https://github.com/app-sre/qontract-reconcile/pull/2356/commits/2bbbf5e022941d5ab7e6e03eb8f9751aa36ee8c2)

reconcile/utils/openshift_resource.py
geoberle: [2bbbf5e](https://github.com/app-sre/qontract-reconcile/pull/2356/commits/2bbbf5e022941d5ab7e6e03eb8f9751aa36ee8c2)

reconcile/utils/openshift_resource.py
geoberle: [2bbbf5e](https://github.com/app-sre/qontract-reconcile/pull/2356/commits/2bbbf5e022941d5ab7e6e03eb8f9751aa36ee8c2)

reconcile/openshift_base.py
```diff
@@ -34,28 +35,41 @@ class ValidationErrorJobFailed(Exception):
...
+@dataclass
+class CurrentStateSpec(BaseStateSpec):
+
+    kind: str
+    resource_names: Optional[Iterable[str]]
+
+    def __str__(self) -> str:
+        return (
+            f"current spec - {super().__str__()}, resource_names: {self.resource_names}"
+        )
```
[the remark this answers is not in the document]
geoberle: i followed your advice and used repr=False and did not override __str__ and i like how the result looks like

```
CurrentStateSpec(cluster='xxx', namespace='xxx', kind='Job', resource_names=None)
DesiredStateSpec(cluster='xxx', namespace='xxx', resource={'provider': 'resource-template', 'path': '/xxx', 'type': 'extracurlyjinja2', 'variables': None}, privileged=False)
```

reconcile/openshift_base.py
```diff
@@ -34,28 +35,41 @@ class ValidationErrorJobFailed(Exception):
...
+
+
+@dataclass
+class DesiredStateSpec(BaseStateSpec):
+
+    resource: Mapping[str, Any]
+    parent: Mapping[Any, Any]
```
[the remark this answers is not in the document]
geoberle: parent is too noisy so i set repr=False. `cluster` and `namespace` give enough context for debugging issues

reconcile/openshift_base.py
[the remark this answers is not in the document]
geoberle: [4cd2898](https://github.com/app-sre/qontract-reconcile/pull/2356/commits/4cd28989ffa7c434ed3d0697d6b533da5293b0b0)

reconcile/openshift_resources_base.py
geoberle: 🥳

reconcile/openshift_resources_base.py
geoberle: [7641249](https://github.com/app-sre/qontract-reconcile/pull/2356/commits/7641249f9c0b47138353c5239d33bc819a442b1e)

reconcile/openshift_base.py
geoberle: [7641249](https://github.com/app-sre/qontract-reconcile/pull/2356/commits/7641249f9c0b47138353c5239d33bc819a442b1e)

geoberle:
> Can you add a test to ensure that name filtering `managedResourceNames` still works for openshiftResources in namespaces. Support for this filtering should be added in saas files (https://issues.redhat.com/browse/APPSRE-4355) as well. But it could be in a different PR IMHO

the existing tests show that the `managedResourceNames` makes it into the `Spec` and this new tests shows that the spec influences `populate_current_state` to call `get_items` on the oc client with proper name filters.

https://github.com/app-sre/qontract-reconcile/blob/2261679816e73dd6c0dca18a3a151520eb776756/reconcile/test/test_openshift_base.py#L576

geoberle:
> Can we have some YAML fixture with the new capabilities you are proposing? This is a complicated beast, and I'd like to have some code (and reference) on valid and invalid combinations of `managedResourceTypes` and `managedResourceTypeOverrides`.
> 
> Synthetic pytest fixtures aren't giving me the full understanding, TBH.

PTAL at the restructured way of providing namespace YAML

maorfr (approved):
LGTM

thanks for all this work!

PR #2357, opened 2022-04-25
APPSRE-3556 remove saas-file-1

fishi0x01:
Old `saas-file-1` was used in `service_dependencies` integration. This PR removes it

### DIFF
--- a/reconcile/service_dependencies.py
+++ b/reconcile/service_dependencies.py
@@ -20,11 +20,6 @@ APPS_QUERY = """
         name
       }
     }
-    saasFiles {
-      instance {
-        name
-      }
-    }
     quayRepos {
       org {
         name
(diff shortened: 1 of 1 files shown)


maorfr:
instead of removing, we need to shift this to use saas-file-2.

fishi0x01:
@maorfr I'd rather do that in a follow-up PR to be honest. Removing this section now doesn't change anything in functionality, as it is already dead code. You can verify its dead with this query on app-interface master

```
apps: apps_v1 {
    saasFiles {
      instance {
        name
      }
    }
  }
```

In a follow-up PR I will create new functionality: support for saas-file-2.

maorfr:
of course it's a dead query, because saas file v2 reference jenkins instances from resource template targets, and not from the root of the file:
https://github.com/app-sre/qontract-schemas/blob/30c522a6e960f97e6dbd0ff5442ba25dde7910fa/schemas/app-sre/saas-file-2.yml#L152-L154

fishi0x01:
Created a [new ticket](https://issues.redhat.com/browse/APPSRE-5046) and assigned it to me - will do this once saas-file-1 is purged.

maorfr (approved):

PR #2359, opened 2022-04-25
add support to define region for rds instances

maorfr:
depends on https://github.com/app-sre/qontract-schemas/pull/131

this included some changes around how we handle `availability_zone`, to add a validation that there is no conflict.


bkez322 (approved):
LGTM

PR #2364, opened 2022-04-28
[gabi-authorized-users] add expiration to ConfigMap

maorfr:
when a user attempts to query a Gabi instance, they may not be authorized, or the expiration for the instance had passed:
https://github.com/app-sre/gabi/blob/b4be98ac40fe91c7911f4347339b8ddc7db9e0ee/pkg/handlers/query.go#L49

there is little we can say in addition to "Unauthorized" when a user is unauthorized, but we could check for the instance expiration and output a more informative message.

this PR adds the instance expiration date to the Gabi ConfigMap, which will allow a Gabi instance to output a message about expiration. this part needs to be implemented in Gabi itself, so this PR is only a preparation for that: https://github.com/app-sre/gabi/pull/27

### DIFF
--- a/reconcile/gabi_authorized_users.py
+++ b/reconcile/gabi_authorized_users.py
@@ -16,12 +16,17 @@ QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
 EXPIRATION_MAX = 90
 
 
-def construct_gabi_oc_resource(name: str, users: Iterable[str]) -> OpenshiftResource:
+def construct_gabi_oc_resource(
+    name: str, users: Iterable[str], exp_date: date
+) -> OpenshiftResource:
     body = {
         "apiVersion": "v1",
         "kind": "ConfigMap",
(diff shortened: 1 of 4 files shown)


Piojo:
If we're giving a 403, we shouldn't be giving any more details. We should log as much as possible, and at the beginning of the session we can and probably should report back the exact expiration time (if we don't do that already), but an unauthorized user may be hostile and shouldn't be able to see information about the time windows in which their request might have worked.

Piojo:
Alternatively, if we are assuming the user is authotized but their instance has expired, let's clarify it with a `410 GONE` status code.

maorfr:
@Piojo this makes a lot of sense. for this approach to work, we will need to change the logic of the integration to not remove users from the configmap when expiration passes, and change the order of checks in gabi (user authentication first, user authorization (expiration) second).

geoberle (approved):
Makes sense

PR #2374, opened 2022-05-01
url makes sense no verify

maorfr:
we use this function as part of the openshift resources `url` templating (check if a url exists).
when an SSL issue exists (as happens with gitlab for example), we deem the url bad, while it is in fact good.

instead of getting caught up on SSL (which we care about in other perspectives of course), let's let this through without verifying SSL.


maorfr:
example ticket: https://issues.redhat.com/browse/DEVREG-10

PR #2376, opened 2022-05-02
APPSRE-5476 Add pagination for lists

janboll:
Add pagination for ocm get calls

### DIFF
--- a/reconcile/utils/ocm.py
+++ b/reconcile/utils/ocm.py
@@ -930,11 +930,35 @@ class OCM:  # pylint: disable=too-many-public-methods
             except re.error:
                 raise TypeError(f"blocked version is not a valid regex expression: {b}")
 
-    @retry(max_attempts=10)
-    def _get_json(self, api):
-        r = requests.get(f"{self.url}{api}", headers=self.headers)
+    def _do_get_request(self, api: str, params: dict[str, str]):
+        r = requests.get(f"{self.url}{api}", headers=self.headers, params=params)
         r.raise_for_status()
         return r.json()
+
+    def _get_json(self, api):
+        def response_is_list(rs: dict[str, Any]) -> bool:
+            return rs["kind"].endswith("List")
+
+        responses = []
+        params = {"size": 100}
+        while True:
+            rs = self._do_get_request(api, params=params)
+            responses.append(rs)
+            if response_is_list(rs) and rs["size"] == params["size"]:
+                params["page"] = rs["page"] + 1
+            else:
+                break
+
+        if response_is_list(responses[0]):
+            items = []
+            for resp in responses:
+                items.extend(resp["items"])
+            return {
+                "kind": responses[0]["kind"],
+                "items": items,
+                "total": len(items),
+            }
+        return responses[0]
 
     def _post(self, api, data=None, params=None):
         r = requests.post(
--- a/reconcile/utils/ocm.py
+++ b/reconcile/utils/ocm.py
@@ -930,6 +930,7 @@ class OCM:  # pylint: disable=too-many-public-methods
             except re.error:
                 raise TypeError(f"blocked version is not a valid regex expression: {b}")
 
+    @retry(max_attempts=10)
     def _do_get_request(self, api: str, params: dict[str, str]):
         r = requests.get(f"{self.url}{api}", headers=self.headers, params=params)
         r.raise_for_status()
--- a/reconcile/utils/ocm.py
+++ b/reconcile/utils/ocm.py
@@ -1,7 +1,7 @@
 import functools
 import logging
 import re
-from typing import Any, Optional, Union
+from typing import Any, Optional, Union, Mapping
 
 import reconcile.utils.aws_helper as awsh
 import requests
@@ -937,7 +937,7 @@ class OCM:  # pylint: disable=too-many-public-methods
                 raise TypeError(f"blocked version is not a valid regex expression: {b}")
 
     @retry(max_attempts=10)
-    def _do_get_request(self, api: str, params: dict[str, str]):
+    def _do_get_request(self, api: str, params: Mapping[str, str]) -> Mapping[str, Any]:
         r = self.session.get(
             f"{self.url}{api}",
             params=params,
@@ -947,10 +947,10 @@ class OCM:  # pylint: disable=too-many-public-methods
         return r.json()
 
     @staticmethod
-    def _response_is_list(rs: dict[str, Any]) -> bool:
+    def _response_is_list(rs: Mapping[str, Any]) -> bool:
         return rs["kind"].endswith("List")
 
-    def _get_json(self, api):
+    def _get_json(self, api: str) -> Mapping[str, Any]:
         responses = []
         params = {"size": 100}
         while True:
--- a/reconcile/utils/ocm.py
+++ b/reconcile/utils/ocm.py
@@ -25,6 +25,7 @@ CLUSTER_ADDON_DESIRED_KEYS = {"id", "parameters"}
 
 DISABLE_UWM_ATTR = "disable_user_workload_monitoring"
 BYTES_IN_GIGABYTE = 1024**3
+REQUEST_TIMEOUT_SEC = 60
 
 
 class OCM:  # pylint: disable=too-many-public-methods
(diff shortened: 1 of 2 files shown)


reconcile/utils/ocm.py
```diff
@@ -931,11 +931,36 @@ def _init_blocked_versions(self, blocked_versions):
...
-    def _get_json(self, api):
-        r = requests.get(f"{self.url}{api}", headers=self.headers)
+    def _do_get_request(self, api: str, params: dict[str, str]):
+        r = requests.get(f"{self.url}{api}", headers=self.headers, params=params)
         r.raise_for_status()
         return r.json()
 
+    def _get_json(self, api):
+        def response_is_list(rs: dict[str, Any]) -> bool:
+            return rs["kind"].endswith("List")
```
Piojo, line 941: Move this itnernal method to the class (make it a `@staticmethod` if you need to). It'll make testing easier.

Piojo (commented):
This looks solid, let me know when you consider it ready to merge for a final review.

reconcile/utils/ocm.py
janboll: Good point. done

janboll:
> This looks solid, let me know when you consider it ready to merge for a final review.

Please go ahead!

maorfr:
is this still WIP as the description says?

janboll:
No, forgot to update the description. Did it now :+1:

reconcile/utils/ocm.py
```diff
@@ -931,11 +931,37 @@ def _init_blocked_versions(self, blocked_versions):
...
 
+    @staticmethod
+    def _response_is_list(rs: dict[str, Any]) -> bool:
+        return rs["kind"].endswith("List")
+
+    def _get_json(self, api):
+        responses = []
+        params = {"size": 100}
+        while True:
+            rs = self._do_get_request(api, params=params)
```
steveteahan, line 947: Just curious if this will use keep-alive and how much that matters? ([docs](https://docs.python-requests.org/en/latest/user/advanced/#keep-alive))

reconcile/utils/ocm.py
```diff
@@ -931,11 +931,37 @@ def _init_blocked_versions(self, blocked_versions):
...
+            responses.append(rs)
+            if self._response_is_list(rs) and rs["size"] == params["size"]:
```
steveteahan, line 949: I haven't worked very closely with the OCM API, but I was just curious if the `List` suffix is always going to be general enough? I didn't see an obvious use of this in the `ocm-cli` linked from the ticket (100% possible I missed it).

reconcile/utils/ocm.py
```diff
@@ -931,11 +931,37 @@ def _init_blocked_versions(self, blocked_versions):
...
     @retry(max_attempts=10)
-    def _get_json(self, api):
-        r = requests.get(f"{self.url}{api}", headers=self.headers)
+    def _do_get_request(self, api: str, params: dict[str, str]):
+        r = requests.get(f"{self.url}{api}", headers=self.headers, params=params)
         r.raise_for_status()
         return r.json()
 
+    @staticmethod
+    def _response_is_list(rs: dict[str, Any]) -> bool:
```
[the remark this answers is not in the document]
janboll: As per my understand Mapping should be used for any mappable types, however this is a pure dict, hence usage of `dict[]`

reconcile/utils/ocm.py
janboll: Yes, that's per definition: https://github.com/openshift-online/ocm-api-model

reconcile/utils/ocm.py
```diff
@@ -931,11 +931,37 @@ def _init_blocked_versions(self, blocked_versions):
                 raise TypeError(f"blocked version is not a valid regex expression: {b}")
 
     @retry(max_attempts=10)
-    def _get_json(self, api):
-        r = requests.get(f"{self.url}{api}", headers=self.headers)
+    def _do_get_request(self, api: str, params: dict[str, str]):
+        r = requests.get(f"{self.url}{api}", headers=self.headers, params=params)
```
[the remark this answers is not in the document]
janboll: Valid point, I was just taking over the existing implementation. Let me reconsider this.

reconcile/utils/ocm.py
janboll: See timeout

reconcile/utils/ocm.py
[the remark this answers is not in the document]
steveteahan: For your use here, there isn't anything `dict`-specific though? You're only really doing a `__getitem__`, so we could allow the caller to provide anything that implements that? `Mapping` is just more general, covering `dict` as well as anything else that implements the methods in the [docs](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes).

reconcile/utils/ocm.py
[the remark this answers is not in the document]
steveteahan: I realized that after the fact as well, but probably the time to fix it anyways. I think these `requests` calls without timeouts are a major source of integrations hanging when the API we're calling has any transient issues.

Thanks for handling this one!

reconcile/utils/ocm.py
```diff
@@ -931,11 +931,37 @@ def _init_blocked_versions(self, blocked_versions):
...
         r.raise_for_status()
         return r.json()
 
+    @staticmethod
+    def _response_is_list(rs: dict[str, Any]) -> bool:
+        return rs["kind"].endswith("List")
+
+    def _get_json(self, api):
```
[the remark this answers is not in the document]
janboll: done

reconcile/utils/ocm.py
```diff
@@ -931,11 +931,37 @@ def _init_blocked_versions(self, blocked_versions):
                 raise TypeError(f"blocked version is not a valid regex expression: {b}")
 
     @retry(max_attempts=10)
-    def _get_json(self, api):
-        r = requests.get(f"{self.url}{api}", headers=self.headers)
+    def _do_get_request(self, api: str, params: dict[str, str]):
```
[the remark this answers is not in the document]
janboll: done

reconcile/utils/ocm.py
```diff
@@ -67,6 +68,7 @@ def __init__(
         self.access_token_client_id = access_token_client_id
         self.access_token_url = access_token_url
         self.offline_token = offline_token
+        self.session = requests.Session()
```
steveteahan, line 71: nit: maybe make this `self._session`? I know we don't have the strongest convention of public vs. private in the code base, but not a bad place to start.

reconcile/utils/ocm.py
```diff
@@ -931,14 +937,44 @@ def _init_blocked_versions(self, blocked_versions):
                 raise TypeError(f"blocked version is not a valid regex expression: {b}")
 
     @retry(max_attempts=10)
-    def _get_json(self, api):
-        r = requests.get(f"{self.url}{api}", headers=self.headers)
+    def _do_get_request(self, api: str, params: Mapping[str, str]) -> Mapping[str, Any]:
```
steveteahan, line 940: To avoid confusion, we want to keep inputs general, but for outputs and types on collections, it's encouraged to use concrete types, so:

`_do_get_request(self, api: str, params: Mapping[str, str]) -> dict[str, Any]`

reconcile/utils/ocm.py
```diff
@@ -931,14 +937,44 @@ def _init_blocked_versions(self, blocked_versions):
...
+            timeout=REQUEST_TIMEOUT_SEC,
+        )
         r.raise_for_status()
         return r.json()
 
+    @staticmethod
+    def _response_is_list(rs: Mapping[str, Any]) -> bool:
+        return rs["kind"].endswith("List")
+
+    def _get_json(self, api: str) -> Mapping[str, Any]:
```
steveteahan, line 953: Same as above, `dict` is fine for the return value type.

janboll: done

reconcile/utils/ocm.py
janboll: done

reconcile/utils/ocm.py
janboll: done

steveteahan (approved):
LGTM

PR #2379, opened 2022-05-03
Fix RDS ignored modifications logic to be region aware [APPSRE-5483]

steveteahan:
We are seeing errors when upgrading a database in a region other than `us-east-1`. This is because there is logic to check whether a RDS change should skip `terraform apply`, but that logic doesn't account for RDS instances in regions other than `us-east-1`.

The error that we'd see was:

```
botocore.errorfactory.DBInstanceNotFoundFault: An error occurred (DBInstanceNotFound) when calling the DescribeDBInstances operation: DBInstance <database-name> not found.
```

**Testing**

- Changed the `engine_version` of a database in `us-west-2` and observed the error above
- Applied the code fix in the PR and the error was no longer present in dry-run mode
- Changed the `engine_version` of a database in `us-east-1` and confirmed dry-run finished successfully

### DIFF
--- a/reconcile/utils/aws_api.py
+++ b/reconcile/utils/aws_api.py
@@ -1392,13 +1392,22 @@ class AWSApi:  # pylint: disable=too-many-public-methods
         return images[0]["ImageId"]
 
     def describe_rds_db_instance(
-        self, account_name: str, db_instance_name: str
+        self,
+        account_name: str,
+        db_instance_name: str,
+        region_name: Optional[str] = None,
     ) -> DBInstanceMessageTypeDef:
         """
         Describe a single RDS instance.
         :param account_name: the name of the account in app-interface
         :param db_instance_name: the name of the database (ex. some-database-stage)
+        :param region_name: AWS region name for the resource, otherwise fallback to default
         :return: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBInstances.html#API_DescribeDBInstances_ResponseElements
         """
-        rds = self._account_rds_client(account_name)
+        optional_kwargs = {}
+
+        if region_name:
+            optional_kwargs.update({"region_name": region_name})
+
+        rds = self._account_rds_client(account_name, **optional_kwargs)
         return rds.describe_db_instances(DBInstanceIdentifier=db_instance_name)
--- a/reconcile/utils/aws_api.py
+++ b/reconcile/utils/aws_api.py
@@ -1407,7 +1407,7 @@ class AWSApi:  # pylint: disable=too-many-public-methods
         optional_kwargs = {}
 
         if region_name:
-            optional_kwargs.update({"region_name": region_name})
+            optional_kwargs["region_name"] = region_name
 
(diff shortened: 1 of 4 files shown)


maorfr:
Related.to https://github.com/app-sre/qontract-reconcile/pull/2359?

steveteahan:
> Related.to #2359?

I missed that PR, so thank you for pointing it out. I'm not sure that necessarily changes the approach here (I could very well be missing something) because:

* we determine whether to apply or not from the `TerraformClient` rather than `Terrascript`, so I can't access `region` from our schema in a simple way (without some refactoring)
* the `resource_change` doesn't include data about the `provider` unless I've missed something, but `availability_zone` should be present in any case (`multi_az` or otherwise)

Basically, in this change I have the context of Terraform's `aws_db_instance` rather than the app-interface schema.

reconcile/utils/aws_api.py
```diff
@@ -1392,13 +1392,22 @@ def get_image_id(
...
         :param account_name: the name of the account in app-interface
         :param db_instance_name: the name of the database (ex. some-database-stage)
+        :param region_name: AWS region name for the resource, otherwise fallback to default
         :return: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBInstances.html#API_DescribeDBInstances_ResponseElements
         """
-        rds = self._account_rds_client(account_name)
+        optional_kwargs = {}
+
+        if region_name:
+            optional_kwargs.update({"region_name": region_name})
```
Piojo, line 1410: Isn't this simpler?
```suggestion
            optional_kwargs["region_name"] = region_name
```

reconcile/utils/terraform_client.py
```diff
@@ -519,8 +521,9 @@ def _is_ignored_rds_modification(self, account_name: str, resource_name: str,
         ]
         if len(changed_terraform_args) == 1 \
                 and 'engine_version' in changed_terraform_args:
+            region_name = get_region_from_availability_zone(before['availability_zone'])
             response = \
-                self._aws_api.describe_rds_db_instance(account_name, resource_name)
+                self._aws_api.describe_rds_db_instance(account_name, resource_name, region_name=region_name)
```
Piojo, line 526: Line too long, even for Black standards. Let's configure pylint to whine on these.

reconcile/utils/aws_api.py
steveteahan: I was originally thinking there'd be multiple key updates for some reason, but agree this slightly simpler in this case.

reconcile/utils/terraform_client.py
steveteahan: I wonder if it's just time to remove the excludes unless someone on the team objects strongly because they're in the middle of a large refactor (I'm guessing not the case). Any thoughts @fishi0x01?

steveteahan:
Maor will be out for the rest of the week and I don't want to hold this back given it will block RDS upgrades. If there's a better way to do this, I'm open to revisiting this next week, but from what I can see at the moment this seemed like the simplest fix without too much refactoring.

PR #2381, opened 2022-05-08
add fake empty value for env section compare

maorfr:
another attempt similar to #2375 (which was partially reverted in #2378), but from a different approach.

instead of only "fixing" our own template to not have empty values (which are hard to compare), this PR adds a special handle to compare environment variables.

if an current environment variable only has the `name` key, it means the desired value that was applied is empty (`""`). 
by "faking" an empty `value`, we are able to compare (and not apply when "equal").

### DIFF
--- a/reconcile/utils/openshift_resource.py
+++ b/reconcile/utils/openshift_resource.py
@@ -100,7 +100,6 @@ class OpenshiftResource:
                             v["value"] = ""
                     if not self.obj_intersect_equal(obj1_v, obj2_v):
                         return False
-                    pass
                 elif obj1_k == "cpu":
                     equal = self.cpu_equal(obj1_v, obj2_v)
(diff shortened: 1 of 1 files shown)


reconcile/utils/openshift_resource.py
```diff
@@ -94,6 +94,13 @@ def obj_intersect_equal(self, obj1, obj2):
                     ]
                     if diff or not self.obj_intersect_equal(obj1_v, obj2_v):
                         return False
+                elif obj1_k == "env":
+                    for v in obj2_v:
+                        if "name" in v and len(v) == 1:
```
maorfr, line 99: if the length is 1 it means the `name` is the only key, but this is more readable IMO

geoberle (approved):
LGTM

PR #2383, opened 2022-05-09
saasherder get contents from endpoint by file size

maorfr:
part of https://issues.redhat.com/browse/APPSRE-5500

following up on https://github.blog/changelog/2022-05-03-increased-file-size-limit-when-retrieving-file-contents-via-rest-api

we are seeing a change in github API behavior. namely, a `GithubException` is no longer raised, and we can not detect that we have a problem with a file being too large. while the above blog post indicates the required changes to consume a file from a single endpoint regardless of it's size, that will require a change to headers, which should likely get implemented in PyGithub.

nonetheless, this change will work as is. in fact, this was how we should have done this to begin with (#1364).

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -500,34 +500,16 @@ class SaasHerder:
 
     @staticmethod
     def _get_file_contents_github(repo, path, commit_sha):
-        try:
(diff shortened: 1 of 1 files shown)


pontg:
lgtm

pontg (approved):

PR #2367, opened 2022-04-28
[WIP] 1st PoC for typed gql queries design doc

fishi0x01:
:warning: In Progress

This is a very simple PoC to leverage https://github.com/profusion/sgqlc as a gql client and code generator.

**Background**

[Design Doc](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/38147/diffs)

Based on idea https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/gql-client-for-go-integrations.md

GraphQL data is currently used in un-typed nested dictionaries. Adding type safety will increase maintainability of the code base. Non-compliant schema changes (some integrations are using a schema that you might not be aware of) are more likely to raise errors during development. The idea would be to gradually migrate to typed queries over time, i.e., use the new `typed_gql_query.py` whenever a good occasion occurs. `query.py` and `typed_gql_query.py` can co-exist.

All .py files in `schema` directory are generated. Only .gql query files are manually written to clearly define the query

**Open Issue**

No static type checking currently supported. While theoretically doable, it would require an upstream contribution.

https://github.com/profusion/sgqlc/issues/129#issuecomment-885820088

PR #2389, opened 2022-05-10
2nd PoC for Typed GQL Queries Design Doc

fishi0x01:
:warning: Show-case only! Needs some adjustments before actual merge

This is a PoC to leverage a simple custom code generator for gql queries to python

**Background**

[Design Doc](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/38147/diffs)

GraphQL data is currently used in un-typed nested dictionaries. Adding type safety will increase maintainability of the code base. Non-compliant schema changes (some integrations are using a schema that you might not be aware of) are more likely to raise errors during development.

We use a [similar approach](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/gql-client-for-go-integrations.md) for our go integrations.

**Reviewer Notes**

Query examples and their corresponding generated code can be found in `reconcile/gql_queries`.

The module `code_generator` is a proof the concept, i.e., it is not feature complete with everything GQL has to offer. However, it covers our most common use-cases, like standard queries and inline fragments. If we decide to go forward with a custom code generator approach, then it requires a bit more work to adhere to best practices.

The module `reconcile.gql_queries` contains the query definitions `.gql` files and the corresponding generated `.py` files. Those `.py` files are then consumed in `reconcile`.

**Implementation Details**

A query is defined in a `.gql` file. The content of a `.gql` file is a valid query that could be used with curl against a GQL backend.
All `.gql` files are placed under the `gql_queries` directory.

Running `make generate-queries` will:

1. Run an introspection query against a GQL backend, i.e., fetch the schema
1. Read all `.gql` files under `reconcile/gql_queries`
1. For each `.gql` file, generate the AST and match it with the types from the schema, i.e., generate a typed AST
1. Generate a corresponding `.py` file for each `.gql` file and place it in the same (sub)directory

In the end the `reconcile/gql_queries` directory could look like this:

```
reconcile/gql_queries/
├── __init__.py
└── saas_files
    ├── __init__.py
    ├── saas_files_full.gql
    ├── saas_files_full.py
    ├── saas_files_small_with_provider.gql
    └── saas_files_small_with_provider.py
```

The generated `.py` files can be consumed like:

```python
from gql_queries.saas_files import saas_files_full


with open("reconcile/gql_queries/saas_files/saas_files_full.gql", "r") as f:
    query = f.read()

data: dict[Any, Any] = gqlapi.query(query)
apps: list[AppV1] = saas_files_full.MyQuery(**data).apps_v1 or []
```

**Other**

There is a parallel PoC open for a 3rd party code generator https://github.com/app-sre/qontract-reconcile/pull/2367


setup.py
```diff
@@ -76,6 +76,7 @@
     ],
     entry_points={
         'console_scripts': [
+            'query-generator = code_generator.main:main',
```
janboll, line 79: Wouldn't this better fit in: https://github.com/app-sre/sretoolbox or even its own repository?
It does not directly relate to integrations.

fishi0x01: Fully agree. Would make sense to extract this. 

If we finally decide to use a custom code generator, then I will move the `code_generator` module to a different repo or sretoolbox. Until then I will keep it here to have the whole PoC visible in a single PR.

reconcile/typed_queries.py
```diff
@@ -0,0 +1,52 @@
+from typing import Any
+
+from reconcile.gql_queries.ocp import ocp_auth
+from reconcile.gql_queries.saas_files import saas_files_small_with_provider
+from reconcile.gql_queries.saas_files import saas_files_full
+from reconcile.utils import gql
+
+
+def get_query_data(query_name: str) -> dict[Any, Any]:
+    gqlapi = gql.get_api()
```
janboll, line 10: Slightly of topic: can we try to get rid of using get_api if we refactor this? Having this singleton makes it hard to handle i.e. shared sessions to speed up queries.
I would love to see these as methods of a Gql Client class. Where we create instances of the Gql Client per thread.

fishi0x01:
This PoC is open since a while and there had been a schema change. The change was caught during static type checking phase -> as it should :smile:  :+1:

fishi0x01:
Code is migrated to dedicated repository https://github.com/fishi0x01/qenerate

Keeping this PR open until design doc is merged https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/41063

Will close this PR afterwards

fishi0x01:
Opened https://github.com/app-sre/qontract-reconcile/pull/2495 to use the new code generator on `service_dependencies`.

PR #2393, opened 2022-05-10
run integration with current work dir near source code

maorfr:

the integrations-manager integration is failing with the following message:
```
reconcile.utils.helm.HelmTemplateError: Error running helm template [helm template ./helm/qontract-reconcile -n qontract-reconcile -f /tmp/tmpc9kul414] Error: path "./helm/qontract-reconcile" not found
```

this PR should make the integration run from a location where `./helm` exists.


maorfr:
the script was copied to `/` in #385. it doesn't seem there was a particular reason.

geoberle (approved):
i'm not aware of any requirement to have the run script in `/` so this should be fine

PR #2402, opened 2022-05-11
fix helm chart for CronJobs

maorfr:

this PR updates the apiVersion for CronJobs and fixes typos in the job history limit field names (this never worked :scream:)

### DIFF
--- a/helm/qontract-reconcile/templates/template.yaml
+++ b/helm/qontract-reconcile/templates/template.yaml
@@ -361,7 +361,7 @@ objects:
   spec:
     schedule: "{{ $integration.cron }}"
     concurrencyPolicy: {{ $integration.concurrencyPolicy | default "Allow" }}
-    successfulJobsHistoryLimit: {{ $integration.successfulJobHistoryLimit | default "3" }}
+    successfulJobsHistoryLimit: {{ if not (kindIs "invalid" $integration.successfulJobHistoryLimit) }}{{ $integration.successfulJobHistoryLimit }}{{ else }}{{ 3 }}{{ end }}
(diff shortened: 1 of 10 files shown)


helm/qontract-reconcile/templates/template.yaml
```diff
@@ -361,8 +361,8 @@ objects:
   spec:
     schedule: "{{ $integration.cron }}"
     concurrencyPolicy: {{ $integration.concurrencyPolicy | default "Allow" }}
-    successfulJobHistoryLimit: {{ $integration.successfulJobHistoryLimit | default "3" }}
-    failedJobHistoryLimit: {{ if not (kindIs "invalid" $integration.failedJobHistoryLimit) }}{{ $integration.failedJobHistoryLimit }}{{ else }}{{ 1 }}{{ end }}
+    successfulJobsHistoryLimit: {{ $integration.successfulJobHistoryLimit | default "3" }}
```
Piojo, line 364: Doesn't this have the same typing issue as `failedJobsHistoryLimit`? Shouldn't this have an integer as its default?

maorfr: did the same as #2400 in https://github.com/app-sre/qontract-reconcile/pull/2402/commits/92b66cf7f29abdfef8f97e358299ed6f963acbf4

Piojo (approved):

helm/qontract-reconcile/templates/template.yaml
maorfr: the issue doesn't exist because no integration is using `0` :)

PR #2406, opened 2022-05-12
[WIP] remove prometheus extra labels (shard-key)

geoberle:
extra labels were only used by terraform-resources-wrapper to set the account name of a processing thread into the `shard_key` extra label. this is not working reliably because the extra_label dict is shared between all account threads so metrics reporting is not per account thread but interleaved between all account threads.

even if threading would not be a problem. metrics reporting works on the level of integration runs. in the case of the terraform-resource-wrapper it implies a full run of terraform-resources for each account in the shard. so all accounts run the exact same amount of times and individual reporting is not meaningful.

on hold until integration-manager dynamic sharding is available and proves if it services the tracability purpose given by extra-labels


geoberle:
close in favour of #2459

PR #2407, opened 2022-05-13
Assert that user is AWS account

janboll:
Prevents errors like: An error occurred (NoSuchEntity) when calling the DeleteLoginProfile operation: The user with name foobar cannot be found.

### DIFF
--- a/reconcile/aws_iam_password_reset.py
+++ b/reconcile/aws_iam_password_reset.py
@@ -1,4 +1,6 @@
 import logging
+import sys
+from typing import Any, Mapping, Iterable, Optional
 
 from reconcile import queries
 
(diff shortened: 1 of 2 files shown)


reconcile/aws_iam_password_reset.py
```diff
@@ -8,9 +10,23 @@
 QONTRACT_INTEGRATION = "aws-iam-password-reset"
 
 
+def get_roles(roles: list[dict[str, Any]], org_username: str) -> dict[str, Any]:
+    return [d for d in roles if d["org_username"] == org_username][0]
```
Piojo, line 14: Type mismatch: this is a `list` but you're stating `dict`.

reconcile/aws_iam_password_reset.py
```diff
@@ -8,9 +10,23 @@
 QONTRACT_INTEGRATION = "aws-iam-password-reset"
 
 
+def get_roles(roles: list[dict[str, Any]], org_username: str) -> dict[str, Any]:
```
Piojo, line 13: Try to be generic with the input parameters:
```python
from typing import Mapping, Iterable
....
def get_roles(roles: Iterable[Mapping[str, Any]], ...)
```

reconcile/aws_iam_password_reset.py
```diff
@@ -8,9 +10,23 @@
...
+    return [d for d in roles if d["org_username"] == org_username][0]
+
+
+def account_in_roles(roles: list[dict[str, Any]], aws_account: str) -> bool:
+    for role in roles:
+        if role.get("aws_groups"):
+            for g in role.get("aws_groups"):
```
Piojo, line 20: Shorter:
```suggestion
        for g in role.get("aws_groups") or []:
```
(and re-indent)

fishi0x01 (approved):
Nice! :+1:

PR #2410, opened 2022-05-15
[integrations-manager] add support for image-tag-from-ref

maorfr:
when running integrations-manager against multiple environments, each such environment may have a different version of integrations-manager running. as a result, managed integrations in different environments will be using a different `IMAGE_TAG`.

for this reason, running integrations-manager without environment awareness (`ENVIRONMENT_NAME`, `IMAGE_TAG` env vars) will result in the integration wanting to update all integrations in all environments to use `latest` (the default from the OpenShift template).

this PR adds an option to pass a list of environments, and for each of them - which `ref` of the qontract-reconcile repository should be used.

with this change, integrations-manager can run (specifically in dry-run) without showing any changes when there aren't any changes required.

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1840,10 +1840,11 @@ def parse_image_tag_from_ref(ctx, param, value) -> Optional[Dict[str, str]]:
     if value:
         result = {}
         for v in value:
-            if v.count("=") != 1 or v.find("=") == 0:
-                raise ValueError(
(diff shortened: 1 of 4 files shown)


maorfr:
fulfilled by https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/39056

geoberle (approved):
LGTM
hint: i learned we need this because the integration-manager runs towards multiple envs during PR checks

PR #2419, opened 2022-05-18
Persist session

janboll:
Persisting session eliminates the need to create a new HTTPS session for every query and HTTP lookup. In my tests, this sped up integration prometheus-rule-tester by up to 15%

Ticket: [APPSRE-5616](https://issues.redhat.com/browse/APPSRE-5616)

### DIFF
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -19,8 +19,12 @@ from gql.transport.requests import RequestsHTTPTransport
 from gql.transport.exceptions import TransportQueryError
 from gql.transport.requests import log as requests_logger
 
+from reconcile.utils.requests import global_session_cache
+
 _gqlapi = None
-_local_client = None
+"""Using a thread local variable. This data structure ensures, data is not shared
+ between caches. With the code below this creates one client instance per thread."""
+_local_client = threading.local()
 
 INTEGRATIONS_QUERY = """
 {
@@ -198,33 +202,30 @@ class PersistentRequestsHTTPTransport(RequestsHTTPTransport):
         if self.session is None:
             # Copied over from RequestsHTTPTransport
             self.session = requests.Session()
+            global_session_cache.add_session(self.session)
             # we did not implement this in our copy!
             assert self.retries == 0
 
     def close(self) -> None:
+        # Just pass, since we do not want the session to be disconnected.
+        # This allows us to benefit from TCP keep-alive and reduces SSL
+        # Handshake round trips
         pass
 
 
 def _init_gql_client(url: str, token: Optional[str]) -> Client:
     global _local_client
-    if not _local_client:
-        # Requests.Session is not threadsafe, hence create one session per thread
-        _local_client = threading.local()
-
     req_headers = None
     if token:
         # The token stored in vault is already in the format 'Basic ...'
         req_headers = {"Authorization": token}
     client = getattr(_local_client, "client", None)
     if not client:
-        # Here we are explicitly using sync strategy
-        _local_client.client = Client(
-            transport=PersistentRequestsHTTPTransport(
-                url,
-                headers=req_headers,
-                timeout=30,
-            )
+        transport = PersistentRequestsHTTPTransport(
+            url, headers=req_headers, timeout=30, verify=False
         )
+        # Here we are explicitly using sync strategy
+        _local_client.client = Client(transport=transport)
     return _local_client.client
 
 
@@ -232,7 +233,7 @@ def _init_gql_client(url: str, token: Optional[str]) -> Client:
 def get_sha(server, token=None):
     sha_endpoint = server._replace(path="/sha256")
     headers = {"Authorization": token} if token else None
-    response = requests.get(sha_endpoint.geturl(), headers=headers)
+    response = requests.get(sha_endpoint.geturl(), headers=headers, verify=False)
     response.raise_for_status()
     sha = response.content.decode("utf-8")
     return sha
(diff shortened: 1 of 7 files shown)


reconcile/utils/gql.py
```diff
@@ -198,13 +193,39 @@ def get_resource(path: str) -> dict[str, Any]:
     return get_api().get_resource(path)
 
 
+class PersistentRequestsHTTPTransport(RequestsHTTPTransport):
+    def connect(self):
+        if self.session is None:
+            # Copied over from RequestsHTTPTransport
+            self.session = requests.Session()
+            # we did not implement this in our copy!
+            assert self.retries == 0
```
rporres, line 202: we don't retry here since we have the retry mechanism as a decorator, correct?

reconcile/utils/gql.py
```diff
@@ -198,13 +193,39 @@ def get_resource(path: str) -> dict[str, Any]:
...
+
+    def close(self) -> None:
+        pass
```
rporres, line 205: how are we closing in case the program ends?

reconcile/utils/gql.py
```diff
@@ -198,13 +193,39 @@ def get_resource(path: str) -> dict[str, Any]:
     return get_api().get_resource(path)
 
 
+class PersistentRequestsHTTPTransport(RequestsHTTPTransport):
+    def connect(self):
```
rporres, line 197: Is there a way to reconnect in case of connection being deleted due to TCP idle time out?

reconcile/utils/gql.py
```diff
@@ -198,13 +193,39 @@ def get_resource(path: str) -> dict[str, Any]:
...
+            self.session = requests.Session()
+            # we did not implement this in our copy!
+            assert self.retries == 0
+
+    def close(self) -> None:
+        pass
+
+
 def _init_gql_client(url: str, token: Optional[str]) -> Client:
+    global _local_client
```
rporres, line 209: why is this global if you want to define per thread?

reconcile/utils/gql.py
janboll: Correct, hence I did not copy that feature over

reconcile/utils/gql.py
janboll: That question is spot on with the problems in this code. The issue here is, that the GraphQL API instance is a singleton and not created per thread (which would be the best solution IMHO but would require some (I think a lot) of refactoring). `_local_client` must be global, cause a threading.local instance must be initialized only once and then shared by the threads.

reconcile/utils/gql.py
janboll: I will test this locally. But I think the name of the method chosen here is irritating since HTTP is stateless...  Also, the GQL lib is using requests.Session, which handles the connections.

rporres: It would be great if you could test it locally, e.g. sleeping between the first two queries and making sure it reconnects. How much sleep time? The greater the better. Apparently [the default is 2 hours](https://man7.org/linux/man-pages/man7/tcp.7.html#:~:text=The%20default%20value%20is%207200,timeouts%20may%20be%20much%20shorter.), you may want to change that in your tests 😅

reconcile/utils/gql.py
rporres: Ok. Since you're overriding the `connect` method, you will get rid of the `TransportAlreadyConnected` since you actually want to share the connection between threads... But can it cause problems if two threads need to use the connection at the same time?

janboll: That's the trickery of threading.local. It provides a data structure, where threads don't share data. In the end this means every thread has it's own Client instance

reconcile/utils/gql.py
janboll: TCP connections will just time out....

reconcile/utils/gql.py
rporres: cool!

I guess all this deserves a nice comment in the code for the sake of the future readers :) I'm sure they will have similar doubts 😅

reconcile/utils/gql.py
rporres: what about closing them via `__del__` method?

reconcile/utils/gql.py
janboll: I simply restart nginx proxy during test, looks good:

```
[2022-05-19 14:03:56] [DEBUG] [openshift_resources_base.py:fetch_openshift_resource:478] - Fetching resource: /observability/prometheusrules/mt-sre-ci-int.prometheusrules.yaml
[2022-05-19 14:03:56] [DEBUG] [connectionpool.py:_get_conn:271] - Resetting dropped connection: localhost
[2022-05-19 14:03:57] [DEBUG] [connectionpool.py:_new_conn:939] - Starting new HTTPS connection (2): localhost:40444
```

reconcile/utils/gql.py
janboll: Just checked, we have other places where we use Sessions without caring for threads and do not clean them up... 
I'll follow up on that

reconcile/utils/gql.py
```diff
@@ -198,13 +193,39 @@ def get_resource(path: str) -> dict[str, Any]:
...
+    if not _local_client:
+        # Requests.Session is not threadsafe, hence create one session per thread
+        _local_client = threading.local()
```
bhushanthakur93, line 212: Do we need to account for race condition here?

janboll: Good point, let me check.

reconcile/utils/gql.py
janboll: Added a comment!

janboll: Added a comment!

reconcile/utils/gql.py
janboll: Moved the initialization to global scope

reconcile/utils/gql.py
janboll: Added _SessionClosingCache_ that should take care of closing sessions

rporres:
what happened to this? :(

PR #2422, opened 2022-05-19
[openshift-saas-deploy-trigger] add information on what caused a trigger

maorfr:
part of https://issues.redhat.com/browse/APPSRE-5484

currently we can trace what caused a deployment to be triggered only based on the time.
with this change, we will know what integration triggered a deployment and for what reason.

the reason will be:
- openshift-saas-deploy-trigger-moving-commits - the upstream commit
- openshift-saas-deploy-trigger-upstream-jobs - the jenkins job
- openshift-saas-deploy-trigger-configs - the app-interface commit

parameters are added in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/39361

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -1188,6 +1188,7 @@ class SaasHerder:
                         "namespace_name": namespace_name,
                         "ref": ref,
                         "commit_sha": desired_commit_sha,
+                        "reason": f"{url}/commit/{desired_commit_sha}",
                     }
                     trigger_specs.append(job_spec)
                 except (GithubException, GitlabError):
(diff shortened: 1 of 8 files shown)


maorfr:
this is a breaking change to any app-interface instance using the same templates as the commercial one. specifically, fedramp.
cc @TGPSKI @bkez322 @dwelch0 @cubismod

geoberle (approved):

reconcile/utils/saasherder.py
```diff
@@ -1381,6 +1386,7 @@ def get_configs_diff_saas_file(self, saas_file):
                 "cluster_name": cluster_name,
                 "namespace_name": namespace_name,
                 "target_config": desired_target_config,
+                "reason": f"{self.settings['instanceUrl']}/commit/{RunningState().commit}",
```
maorfr, line 1389: depends on https://github.com/app-sre/qontract-schemas/pull/152

PR #2423, opened 2022-05-19
adding custom bucket acl terraform resource to allow standard logging…

suzana-nesic:
… for cloudfront

We have a tenant that would like to set up logging to a S3 bucket. In order to do this, their policy will need to rely on a bucket ACL in the principal rather than a bucket policy. Part of [APPSRE-5533](https://issues.redhat.com/browse/APPSRE-5533)


reconcile/utils/terrascript_client.py
```diff
@@ -2405,6 +2409,71 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
...
+        values['depends_on'] = self.get_dependencies([bucket_tf_resource])
+        region = common_values.get('region') or \
+            self.default_regions.get(account)
+        if self._multiregion_account(account):
+            values['provider'] = 'aws.' + region
+        bucket_policy_tf_resource = aws_s3_bucket_policy(identifier, **values)
+        tf_resources.append(bucket_policy_tf_resource)
+
         # cloud front distribution
         values = common_values.get('distribution_config', {})
```
maorfr, line 2478: with the current changes, every cloudfront distribution will define s3 logging.
i believe what we would want is to get the logging config, and only if it exists we add resources? that would also tell us the name of the bucket to log to.

maorfr (changes requested):
see above comment. i'm also not convinced the s3 bucket policy is needed?

reconcile/utils/terrascript_client.py
suzana-nesic: That would make sense, it would then look like only quay is affected with this at the moment until others decide to include the logging config. Regarding if the bucket policy is needed or not, would our automation be able to acknowledge the ACL if we don't specify the bucket policy within this PR?

reconcile/utils/terrascript_client.py
```diff
@@ -2450,6 +2454,48 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
             '${' + cf_oai_tf_resource.id + '}'
         tf_resources.append(Output(output_name_0_13, value=output_value))
 
+        # aws_s3_bucket_acl
+        if cf_distribution_tf_resource.logging_config:
+            values = {}
+            values['name'] = identifier
```
maorfr, line 2460: you should probably be using the information from `logging_config`?

reconcile/utils/terrascript_aws_client.py
```diff
@@ -2451,6 +2455,49 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
             '${' + cf_oai_tf_resource.id + '}'
         tf_resources.append(Output(output_name_0_13, value=output_value))
 
+        # aws_s3_bucket_acl
+        values = common_values.get('distribution_config', {})
+        if 'logging_config' in values.keys():
+            values = {}
+            values['name'] = identifier
```
patjlm, line 2462: missing `values["bucket"]`. The value should be taken from `logging_config.bucket`. Beware, the format is probably not the good one.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -2451,6 +2455,49 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
...
+                    'permission': 'READ'
+                }
+            }
+            grant = {
+                'grantee': '${' + aws_s3_bucket_acl_resource.grantee + '}',
+                'permission': '${' + aws_s3_bucket_acl_resource.permission + '}'
+            }
+            owner = {
+                'id': '${' + aws_s3_bucket_acl_resource.id + '}'
+            }
```
patjlm, line 2483: The config above doe not match the example from https://github.com/fmasuhr/terraform-aws-cloudfront-logs/blob/599d948859586dfdaf5189d3be0b1f573cc4be02/s3.tf#L13-L39 . Is that normal ?

reconcile/utils/terrascript_aws_client.py
```diff
@@ -2451,6 +2455,49 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
...
+                    },
+                    'permission': 'READ'
+                }
+            }
+            grant = {
+                'grantee': '${' + aws_s3_bucket_acl_resource.grantee + '}',
+                'permission': '${' + aws_s3_bucket_acl_resource.permission + '}'
+            }
+            owner = {
+                'id': '${' + aws_s3_bucket_acl_resource.id + '}'
```
patjlm, line 2482: Not sure which owner we should put here, but most probably not the `aws_s3_bucket_acl_resource.id`

reconcile/utils/terrascript_aws_client.py
```diff
@@ -2451,6 +2455,49 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
...
 
+        # aws_s3_bucket_acl
+        values = common_values.get('distribution_config', {})
+        if 'logging_config' in values.keys():
+            values = {}
+            values['name'] = identifier
+            values['tags'] = common_values['tags']
+
+            aws_s3_bucket_acl_resource = aws_s3_bucket_acl(identifier, **values)
+            tf_resources.append(aws_s3_bucket_acl_resource)
```
patjlm, line 2466: these 2 lines should be removed. We create a singe `aws_s3_bucket_acl` It will be `tf_resources.append`ed at the end

reconcile/utils/terrascript_aws_client.py
```diff
@@ -2451,6 +2455,49 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
...
+
+            output_name_0_13 = output_prefix + '__acl'
+            output_value = '${' + aws_s3_bucket_acl.acl + '}'
+            tf_resources.append(Output(output_name_0_13, value=output_value))
+
+            if self._multiregion_account(account):
+                values['provider'] = 'aws.' + region
```
patjlm, line 2497: Here you're adding to the `values` which was defined for the `aws_s3_bucket_acl_resource` resource above. Is that what you want ?

patjlm: I am wondergin if this is needed at all. Should probably be removed all together
```
            if self._multiregion_account(account):
                values['provider'] = 'aws.' + region
            bucket_policy_tf_resource = aws_s3_bucket_policy(identifier, **values)
            tf_resources.append(bucket_policy_tf_resource)
```

reconcile/utils/terrascript_aws_client.py
```diff
@@ -2451,6 +2455,42 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
...
+                        'type': 'CanonicalUser',
+                    },
+                    'permission': 'FULL_CONTROL',
+                },
+                'owner': {
+                    'id': '${' + data.aws_canonical_user_id.id + '}',
+                }
+            }
+
+            values['bucket'] = logging_config_bucket.get('bucket')
```
patjlm, line 2479: the output of `logging_config_bucket.get('bucket')` is the bucket url fully-qualified hostname. We need to get only the first part

reconcile/utils/terrascript_aws_client.py
```diff
@@ -2451,6 +2455,42 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
...
+
+            access_control_list = {
+                'grant': {
+                    'grantee': {
+                        'id': '${' + data.aws_canonical_user_id.id + '}',
+                        'type': 'CanonicalUser',
+                    },
+                    'permission': 'FULL_CONTROL',
+                },
+                'owner': {
```
patjlm, line 2474: missign grant section for cloudfront logs delivery
```
    grant {
      # Grant CloudFront logs access to your Amazon S3 Bucket
      # https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html#AccessLogsBucketAndFileOwnership
      grantee {
        id   = "c4c1ede66af53448b93c283ce9448c4ba468c9432aa01d700d3878632f77d2d0"
        type = "CanonicalUser"
      }
      permission = "FULL_CONTROL"
    }
```

patjlm:
@suzana-nesic I added a commit to your branch to fix some stuff.
@geoberle and I also saw that the `aws_s3_bucket_acl` resource depends on the terraform aws provider >3.75.0. @andreugallofre will kindly look at this provider upgrade today. So we depend on this to validate.

patjlm (approved):

reconcile/utils/terrascript_client.py
patjlm: i think this is fine now, right @maorfr ?

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1388,7 +1395,12 @@ def populate_tf_resource_s3(self, resource, namespace_info):
         versioning = common_values.get('versioning') or True
         values['versioning'] = {"enabled": versioning}
         values['tags'] = common_values['tags']
-        values['acl'] = common_values.get('acl') or 'private'
+        if "acl" in common_values:
+            values['acl'] = common_values['acl']
```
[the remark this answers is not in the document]
patjlm: normally the logging bucket is the one where we should remove the `acl: private` and enter this code. Eg https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/41099/diffs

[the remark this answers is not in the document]
maorfr: IIUC this is not the logging bucket, but i may be wrong.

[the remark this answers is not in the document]
maorfr: Ah, never mind. We rely on a bucket definition. Disregard :)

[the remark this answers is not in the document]
patjlm: that worries me :) 
`quayio-stage-cloudfront-logs` it he  logging bucket defined in the s3-cloudfront config
https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/resources/terraform/resources/quayio-stage/s3-cloudfront-us-east-1.yml#L56

PR #2424, opened 2022-05-23
introduce query-validator

maorfr:
part of https://issues.redhat.com/browse/APPSRE-5561

depends on https://github.com/app-sre/qontract-schemas/pull/153

design doc: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/schema-compatibility.md

this PR adds an integration (`query-validator`). This integration acts on `query-validation-1` files. The integration will be pretty straight forward: For each entry in the `queries` list, execute the query. If it fails - fail the integration.

This integration will only run within app-interface pr-check to gate merges of schema promotions that break the tenants' queries. In case this integration fails, we will contact the responsible team to make adjustments to their source code and to the queries in app-interface.


geoberle (approved):
LGTM

PR #2425, opened 2022-05-23
Add ROSA support to OCM clusters integration

lechuk47:
## Overview

Jira: https://issues.redhat.com/browse/APPSRE-4268

This PR is the first step towards support ROSA clusters in app-interface. Changes summary is as follows: 
- Introduced a data model for app-interface ocm specs (pydantic)
- OCM refactor with an adapted bridge pattern to allow the implementation of OCM methods in a product-specific way. 
- OCM-clusters refactor to clarify the steps involved. 
- Added tests to verify rosa clusters integration. 
- Filter clusters in integrations not supported by ROSA.  
    - upgrade-scheduler will be tackled in the next milestone
    - aws_infra_access and additional_routers will be improved in the next iteration

## OCM products implementation

- clusters query will return a `product` attribute depending on the product type `osd` or `rosa`. 
- Each OCM supported product will have an implementation of the parent `OCMProduct` class. These implementations will have all the methods that require a custom implementation to interact with OCM. For example, `create_cluster` and `update_cluster` use different specs depending the cluster type, so each implementation differs in that part.

## Manual tests

Besides the unit tests, other tests have been made to ensure this change does not break the ocm-clusters integration. A ROSA cluster has been deployed in the stage OCM environment and the integration has been tested locally with that cluster in place. 
* OCM gets ROSA specs from OCM and populates the model class as expected (Adding a Rosa cluster to OCM broke this integration a couple of months ago)
* ROSA App-interface specs are populated and compared as expected. 

Schemas PR: https://github.com/app-sre/qontract-schemas/pull/155

### DIFF
--- a/reconcile/utils/ocm.py
+++ b/reconcile/utils/ocm.py
@@ -208,9 +208,9 @@ class OCMProductOsd(OCMProduct):
         }
 
         # Workaround to enable type checks.
-        # cluster.spec is Union of pydantic models Union[OSDClusterSpec, RosaClusterSpec].
+        # cluster.spec is a Union of pydantic models Union[OSDClusterSpec, RosaClusterSpec].
         # In this case, cluster.spec will always be an OSDClusterSpec because the type
-        # assignment is  by pydantic, however, mypy complains if OSD attributes are set
+        # assignment is managed by pydantic, however, mypy complains if OSD attributes are set
         # outside the isinstance check because it checks all the types set in the Union.
         if isinstance(cluster.spec, OSDClusterSpec):
             ocm_spec["storage_quota"] = {


reconcile/ocm/types.py
```diff
@@ -0,0 +1,55 @@
...
+    spec: Union[OSDClusterSpec, ROSAClusterSpec]
+    network: OCMClusterNetwork
+    domain: Optional[str]
+    server_url: str = Field(None, alias="serverUrl")
+    console_url: str = Field(None, alias="consoleUrl")
+    elb_fqdn: str = Field(None, alias="elbFQDN")
+
+    class Config:
+        smart_union = True
+        allow_population_by_field_name = True
```
fishi0x01, line 55: Im struggling to understand why this is needed?

Are the classes intended to be used by multiple queries? In that case:

- will the `Optional` fields be appropriate?
- other queries might use different aliases not appropriate for this model?

lechuk47: I'll try again without this option to see if it works.

reconcile/ocm/types.py
```diff
@@ -0,0 +1,55 @@
...
+    path: Optional[str]
+    spec: Union[OSDClusterSpec, ROSAClusterSpec]
+    network: OCMClusterNetwork
+    domain: Optional[str]
+    server_url: str = Field(None, alias="serverUrl")
+    console_url: str = Field(None, alias="consoleUrl")
+    elb_fqdn: str = Field(None, alias="elbFQDN")
+
+    class Config:
+        smart_union = True
```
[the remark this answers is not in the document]
lechuk47: Added the Option

reconcile/ocm/types.py
```diff
@@ -0,0 +1,54 @@
...
+    initial_version: Optional[str]
+    version: str
+
+
+class OSDClusterSpec(OCMClusterSpec, extra=Extra.forbid):
+    load_balancers: int
+    storage: int
+
+
+class ROSAClusterSpec(OCMClusterSpec, extra=Extra.forbid):
```
fishi0x01, line 40: nit: Sorry for nagging about this again :sweat_smile: Would it make sense to have this parameter set in the `OCMSpec.Config` instead? I feel like having it close to `smart_union = True` makes it easier to understand what this is needed for. Further it wouldn't need to be added to class in the `Union`.

```
class OCMSpec(BaseModel):
    
    ...
    
    class Config:
        smart_union = True
        extra = Extra.forbid
```

Or is there a reason to not forbid extra fields for the whole `OCMSpec` class?

fishi0x01 (approved):
Awesome work! The refactor towards types is great! :heart: 

I think it would be nice to have another opinion here too. This is a big change and Im not a ROSA expert at all.

reconcile/ocm_clusters.py
```diff
@@ -1,76 +1,248 @@
...
+    }
+
+    # Removed attributes in app-interface
+    deleted_attrs = {
+        k: v
+        for k, v in current_ocm_spec.items()
+        if k not in desired_ocm_spec and v is not None
+    }
+
+    diffs = deleted_attrs
```
steveteahan, line 205: It doesn't look like you end up using `deleted_attrs` again, but it's worth noting this behavior (that you're probably already aware of):

```python
>>> deleted_attrs = {'deleted': 'attrs'}
>>> diffs = deleted_attrs
>>> diffs.update({'updated': 'attrs'})
>>> deleted_attrs
{'deleted': 'attrs', 'updated': 'attrs'}
>>> diffs
{'deleted': 'attrs', 'updated': 'attrs'}
```

Normally using `copy.deepcopy()` could help when we want to ensure that we're not using a reference to the existing object, but instead an independent copy, but maybe `diffs = deleted_attrs | updated_attrs` could lead to less unexpected behavior later on (if say we added logging for these vars).

reconcile/ocm_clusters.py
```diff
@@ -83,153 +255,73 @@ def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
         settings=settings,
         init_provision_shards=True,
     )
+
+    # current_state = ocm_state
```
steveteahan, line 259: Is this meant to be a comment for readers, or was this code that was commented out?

reconcile/ocm_clusters.py
```diff
@@ -1,76 +1,248 @@
...
+    deleted_attrs = {
+        k: v
+        for k, v in current_ocm_spec.items()
+        if k not in desired_ocm_spec and v is not None
+    }
+
+    diffs = deleted_attrs
+    diffs.update(updated_attrs)
+
+    not_allowed_updates = set(diffs) - impl.ALLOWED_SPEC_UPDATE_FIELDS
```
geoberle, line 208: can you handle the forbidden network update with the ALLOWED_SPEC_UPDATE_FIELDS approach?

reconcile/ocm_clusters.py
```diff
@@ -83,153 +255,73 @@ def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
...
-                    ocmmod.DISABLE_UWM_ATTR
-                ] = current_uwm  # noqa: E501
-
-            # check if cluster update, if any, is valid
-            update_spec, err = get_cluster_update_spec(
-                cluster_name,
-                current_spec,
-                desired_spec,
+            # OCM API Updates
+            # Changes made to pp-interface manifests that need to be requested
```
geoberle, line 279: typo

reconcile/ocm_clusters.py
```diff
@@ -83,153 +255,73 @@ def run(dry_run, gitlab_project_id=None, thread_pool_size=10):
...
+                        ocm.update_cluster(cluster_name, update_spec, dry_run)
+                    except NotImplementedError:
+                        logging.error(
+                            f"[{cluster_name}] Update clusters is currently not "
+                            "implemented for [{desired_spec.spec.product}] product. "
+                            "Updates to the cluster spec are not supported."
+                        )
+                        # Not marking error as a changer made in ocm could trigger
+                        # this.
+                        # error = True
```
geoberle, line 308: can you elaborate on this?

geoberle (commented):
i love this change a lot. see minor nits

reconcile/ocm/types.py
lechuk47: I followed the example on the docs but I agree with you. I didn't see this option was available in the Config class ...

I can not set it in the OCMSpec class because the dictionary filled with the graphql query has more fields than the class and throws a validation error. I've set it in the OCMClusterSpec which is the base class for the classes in the Union.

reconcile/ocm_aws_infrastructure_access.py
```diff
@@ -3,18 +3,26 @@
...
+SUPPORTED_OCM_PRODUCTS = [OCM_PRODUCT_OSD]
 
 
 def fetch_current_state():
     current_state = []
     current_failed = []
     current_deleting = []
     settings = queries.get_app_interface_settings()
-    clusters = [c for c in queries.get_clusters() if c.get("ocm") is not None]
+    # clusters = [c for c in queries.get_clusters() if c.get("ocm") is not None]
```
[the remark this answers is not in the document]
lechuk47: Removed

reconcile/ocm_aws_infrastructure_access.py
```diff
@@ -45,7 +53,12 @@ def fetch_desired_state():
     # get desired state defined in awsInfrastructureAccess
     # or awsInfrastructureManagementAccounts
     # sections of cluster files
-    clusters = queries.get_clusters()
+    # clusters = queries.get_clusters()
```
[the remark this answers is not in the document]
lechuk47: Removed.

reconcile/ocm_clusters.py
lechuk47: I changed the comment to not seem like a code line.

reconcile/utils/ocm.py
```diff
@@ -28,6 +39,413 @@
...
+OCM_PRODCUT_ROSA = "rosa"
+
+
+class OCMProduct:
+
+    ALLOWED_SPEC_UPDATE_FIELDS: set[str]
+    EXCLUDED_SPEC_FIELDS: set[str]
+
+    @staticmethod
+    def create_cluster(ocm: OCM, name: str, cluster: OCMSpec, dry_run: bool):
```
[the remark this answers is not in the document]
lechuk47: Done. Without raising the NotImplementedError in the base class, I think it's a must. 
Thanks.

reconcile/ocm_additional_routers.py
```diff
@@ -5,10 +5,12 @@
 from reconcile import queries
 
 from reconcile.status import ExitCodes
-from reconcile.utils.ocm import OCMMap
+from reconcile.utils.ocm import OCM_PRODUCT_OSD, OCMMap
 
 QONTRACT_INTEGRATION = "ocm-additional-routers"
 
+SUPPORTED_OCM_PRODUCTS = [OCM_PRODUCT_OSD]
```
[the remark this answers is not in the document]
lechuk47: Thanks!

reconcile/ocm_clusters.py
```diff
@@ -1,76 +1,248 @@
...
+
+def _cluster_version_needs_update(
+    cluster: str, current_version: str, desired_version: str
+) -> bool:
+    """Compares version strings (semver) to determine if a current manifest is required.
+
+    Parameters:
+        cluster (str): cluster name
+        current_version(str): current version string (semver)
+        desired_version(str): desired verseion string (semver)
```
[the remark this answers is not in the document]
lechuk47: Fixed.

reconcile/ocm_clusters.py
lechuk47: The model used is a bit misleading, the network is not part of the cluster spec but the ocm spec. I've changed the code to compare the OCM spec instead of the Cluster spec only, so the network and the cluster spec are joined in a new dict before doing the comparison. 

Thanks, I think this is more elegant now.

reconcile/ocm_clusters.py
lechuk47: Fixed.

reconcile/ocm_clusters.py
lechuk47: I've improved the comment. I want to cover the case where a product does not implement cluster updates. If a change is made in OCM out of app interface, the integration will log the error but the integration will not fail. This could be useful in cases where we want to have the cluster reference but not manage it through ocm-clusters integration.

reconcile/ocm_clusters.py
lechuk47: I just want to be clear on semantics. As you have already seen, this dict is not used further, and IMO is better to read it as `diffs` instead of using `deleted_attrs` or `updated_attrs` as the base to compare with the `ALLOWED_SPEC_UPDATE_FIELDS` constant.

reconcile/ocm_clusters.py
```diff
@@ -1,76 +1,248 @@
...
-        # remove empty keys in spec
-        desired_spec["spec"] = {
-            k: v for k, v in desired_spec["spec"].items() if v is not None
-        }
+def fetch_desired_state(clusters: Iterable[Mapping[str, Any]]) -> dict[str, OCMSpec]:
+    """Builds a dictionary with all clusters retrieved from app-interface. Cluster specs
+    are based on OCMSpec implementations depending the ocm product
 
+    Parameters:
+        clusters (list): list of clusters retrieved from app-interface
```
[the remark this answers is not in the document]
lechuk47: Woah, I did not know this was agreed upon and documented there. I modified the docstrings to that format. 

I'll check the code style guide again, thanks for pointing that out. 

PD: I personally prefer the google format though, I think it's way better.

lechuk47:
> It'd also be helpful for me, given the impact of the change, to understand which manual test cases you've run through.

Added a section with the manual tests I made. I think overall, it's safe to deploy this without breaking anything 🤞

reconcile/ocm_clusters.py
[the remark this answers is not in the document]
steveteahan: I don't have strong opinions on docstring format. I really think consistency, and removing redundant data, is the most important part. Thanks for updating these!

reconcile/ocm_clusters.py
steveteahan: The variable name `diffs` makes sense to me, but really what I'm suggesting is what you're doing is assigning two variable names that reference the same object in memory, yet they have different meanings. `deleted_attrs` makes sense for describing what that object in memory is, until you also create the `diffs` reference and then start changing the underlying data. That's why I was suggesting that `diffs = deleted_attrs | updated_attrs` makes sense because it creates a new object in memory, such that `diffs` and `deleted_attrs` now reference to separate objects.

For now, it's not an issue the way the code is working today, but it could quite easily become problematic if someone changes the code by logging `deleted_attrs` or using it in some other way. This may seem like a nit, but I've seen this issue plenty of times when people who change code later either aren't familiar with how Python variables are references to objects, or they haven't noticed that the same object was referenced with multiple variable names.

reconcile/utils/ocm.py
```diff
@@ -115,171 +503,44 @@ def _ready_for_app_interface(cluster: dict[str, Any]) -> bool:
...
-        params = {}
-        if dry_run:
-            params["dryRun"] = "true"
+    def create_cluster(self, name: str, cluster: OCMSpec, dry_run: bool):
+        impl = OCM_PRODUCTS_IMPL[cluster.spec.product]
+        impl.create_cluster(self, name, cluster, dry_run)
 
-        self._patch(api, ocm_spec, params)
+    def update_cluster(
+        self, cluster_name: str, update_spec: dict[str, Any], dry_run=False
```
steveteahan, line 539: Can `update_spec` be a `Mapping`?

reconcile/utils/ocm.py
```diff
@@ -28,6 +40,379 @@
...
+            console_url=cluster["console"]["url"],
+            server_url=cluster["api"]["url"],
+            domain=cluster["dns"]["base_domain"],
+            spec=spec,
+            network=network,
+        )
+
+        return ocm_spec
+
+    @typing.no_type_check
```
steveteahan, line 187: It would be helpful to have a comment for why this is set.

reconcile/test/test_ocm_clusters.py
```diff
@@ -1,305 +1,464 @@
...
+    with pytest.raises(SystemExit):
+        occ.run(False)
+    # If get has not been called means no action has been performed
+    _post, _patch = ocm_mock
+    assert _post.call_count == 0
+    assert _patch.call_count == 0
+    assert cluster_updates_mr_mock.call_count == 0
+
+
+@patch.object(OCM, "_get_json", autospec=True)
```
steveteahan, line 316: The standard way to patch is to use `mocker` as per [this doc](https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-testing-guide/#pytest-best-practices).

The API of the patched object is the same, so the only thing that should change is adding `mocker` to the arguments for the test function, and then adding a line for: `json = mocker.patch.object(OCM, "_get_json", autospec=True)`

I'll not repeat this comment on the rest of the cases below.

steveteahan:
The remaining points from my side are minor. I don't have any high-level concerns with the change given the manual testing performed, and that it will be confirmed that this doesn't change any existing clusters in the q-r promotion MR.

reconcile/test/test_ocm_clusters.py
lechuk47: I've replaced all the `@patch.object` annotations with a fixture.

reconcile/utils/ocm.py
lechuk47: I've changed the code a bit to enable type checking. In my opinion, this is a problem with how pydantic and mypy interact with Unions. In the end, I think it's better to have type checks even though we add the unnecessary isinstance check.

reconcile/ocm_clusters.py
lechuk47: Fair enough. I changed it with your suggestion.

PR #2428, opened 2022-05-25
[integration-manager] automatic integration sharding based on AWS accounts in app-interface

geoberle:
# Change

enable integration-manager to shard integrations based on a sharding strategy.

# Implementation details
delegate the handling for shards to the `IntegrationShardManager` which knows how to react to different strategies and builds sharding specs for each integration.

a sharding spec is part of the generated `values` file within `integration_spec.shard_specs` which is passed to the helm chart.

## shard spec

a shard spec consists of the following fields:

* `shard_id`, `shards` - the backwards compatible fields for the `SHARD_ID` and `SHARDS` env vars, specifying which shard within a shard set a deployment will represent.
* `shard_name_suffix` - this i used to influence the `name` of the deployment/statefulset
* `extra_args` - contains the aggregation of an integrations global `extraArgs` and the shard specific extra args

## static strategy
this is the default and backward compatible strategy. an integration is running per default with the `static` sharding strategy with 1 shard. the `IntegrationsManager` creates one shard without any `shard_name_suffix`. if `spec.shards > 1` a respective number of `shard_specs` with proper `shard_name_suffixes` (e.g. -0, -1, -2, ...) is created

## per-aws-account-strategy
this strategy creates a shard for each AWS account in app-interface. if an integration is disabled on an account, no shard is created (or an existing one is deleted). the AWS account name is also used as the `shard_name_suffix`, which results in deployments/statefulsets being named `qontract-reconcile-$integration-$account`. this strategy also expects an integration to support an extra arg `--account-name` and provides it to `shard_spec.extra_args`.

## shard_spec use in helm chart
the Helm chart expects a list containing at least one `shard_specs` within an `integration_spec`.

# References
design doc: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/qontract-operator-sharding.md
depends on https://github.com/app-sre/qontract-schemas/pull/157

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1937,12 +1937,3 @@ def get_integration_cli_meta() -> dict[str, IntegrationMeta]:
             short_help=integration_cmd.short_help,  # type: ignore
         )
     return integration_meta
-
-
-if __name__ == "__main__":
-    integrations = get_integration_cli_meta()
-    json.dump(
-        {i: m.to_dict() for i, m in integrations.items()},
-        sys.stdout,
-        indent=4,
-    )
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -126,6 +126,7 @@ from reconcile.utils.gql import GqlApiErrorForbiddenSchema, GqlApiIntegrationNot
 from reconcile.utils.aggregated_list import RunnerException
 from reconcile.utils.binary import binary, binary_version
 from reconcile.utils.environ import environ
+from reconcile.utils.runtime.meta import IntegrationMeta
 from reconcile.utils.unleash import get_feature_toggle_state
 from reconcile.utils.exceptions import PrintToFileInGitRepositoryError
 from reconcile.utils.git import is_file_in_git_repo
@@ -1918,8 +1919,30 @@ def integrations_manager(
         reconcile.integrations_manager,
         ctx.obj,
         environment_name,
+        get_integration_cli_meta(),
         thread_pool_size,
         internal,
         use_jump_host,
         image_tag_from_ref,
+    )
+
+
+def get_integration_cli_meta() -> dict[str, IntegrationMeta]:
+    integration_meta = {}
+    for integration_name in integration.list_commands(None):  # type: ignore
+        integration_cmd = integration.get_command(None, integration_name)  # type: ignore
+        integration_meta[integration_name] = IntegrationMeta(
+            name=integration_name,
+            args=[p.opts[0] for p in integration_cmd.params if p.opts and len(p.opts) > 0],  # type: ignore
+            short_help=integration_cmd.short_help,  # type: ignore
+        )
+    return integration_meta
+
+
+if __name__ == "__main__":
+    integrations = get_integration_cli_meta()
+    json.dump(
+        {i: m.to_dict() for i, m in integrations.items()},
+        sys.stdout,
+        indent=4,
     )
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1928,6 +1928,13 @@ def integrations_manager(
 
 
 def get_integration_cli_meta() -> dict[str, IntegrationMeta]:
+    """
+    returns all integrations known to cli.py via click introspection
+
+    todo(geoberle, janboll) - this needs rework in the long run, especially since go-integrations
+    are becoming relevant and this kind of meta programming just solves the python part, and even
+    the python part is not solved in a robust enough way
+    """
     integration_meta = {}
     for integration_name in integration.list_commands(None):  # type: ignore
         integration_cmd = integration.get_command(None, integration_name)  # type: ignore
--- a/reconcile/integrations_manager.py
+++ b/reconcile/integrations_manager.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from abc import abstractmethod
 import os
 import sys
 import json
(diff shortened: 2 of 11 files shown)


reconcile/test/test_integrations_manager.py
```diff
@@ -193,81 +204,300 @@ def integrations(resources):
...
+        == collected_namespaces_env_test1[0]["integration_specs"][0]["shard_specs"]
+    )
+
+
+def test_initialize_shard_specs_two_shards(
+    collected_namespaces_env_test1: list[dict[str, Any]],
+    shard_manager: intop.IntegrationShardManager,
+):
+    """
+    this test shows how the implicit static sharding strategy creates two shards
```
rporres, line 355: I stayed more minutes than I should have thinking about the `implicit` until I understood. Just in case, maybe `default` is a more used term for what I think you mean

```suggestion
    this test shows how the default static sharding strategy creates two shards
```

reconcile/integrations_manager.py
```diff
@@ -22,10 +24,70 @@
...
+
+    aws_accounts: list[dict[str, Any]]
+    integration_runtime_meta: dict[str, IntegrationMeta]
+
+    def build_integration_shards(
+        self, integration: str, spec: Mapping[str, Any]
+    ) -> list[dict[str, Any]]:
+        extra_args = spec["extraArgs"] or ""
+        sharding_strategy = spec.get("shardingStrategy") or "static"
+        if sharding_strategy == "per-aws-account":
```
rporres, line 38: To avoid the if/elif/else, reduce the size of the function (scrolling hell) to have smaller method and foreseeing  future sharding strategies, this is screaming at me the use of a [dispatch table](https://betterprogramming.pub/dispatch-tables-in-python-d37bcc443b0b)

rporres, line 38: To avoid the if/elif/else, reduce the size of the function (scrolling hell) to have smaller method and foreseeing  future sharding strategies, this is screaming at me the use of a [dispatch table](https://betterprogramming.pub/dispatch-tables-in-python-d37bcc443b0b)

rporres:
Looks great, minor stuff :)

reconcile/cli.py
```diff
@@ -1906,8 +1907,30 @@ def integrations_manager(
...
+            name=integration_name,
+            args=[p.opts[0] for p in integration_cmd.params if p.opts and len(p.opts) > 0],  # type: ignore
+            short_help=integration_cmd.short_help,  # type: ignore
+        )
+    return integration_meta
+
+
+if __name__ == "__main__":
+    integrations = get_integration_cli_meta()
+    json.dump(
```
janboll, line 1932: How does that relate to automatic sharding?
When is that output printed?

reconcile/integrations_manager.py
rporres: that could totally work too :)

reconcile/integrations_manager.py
```diff
@@ -22,10 +24,70 @@
 QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
 
 
+@dataclass
+class IntegrationShardManager:
+
+    aws_accounts: list[dict[str, Any]]
```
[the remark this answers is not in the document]
geoberle: this has been refactored into a dedicated shardstrategy class that is specific to aws accounts. as such, the scope of the aws_accounts has been reduced. see [1f1bf97](https://github.com/app-sre/qontract-reconcile/pull/2428/commits/1f1bf97332aa77176ac6436d13f2f547cd565cb9)

rporres (approved):

reconcile/integrations_manager.py
```diff
@@ -22,10 +25,96 @@
 QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
 
 
+class ShardingStrategy:
```
janboll, line 28: Please derive from ABC

reconcile/cli.py
```diff
@@ -1906,8 +1907,21 @@ def integrations_manager(
...
         thread_pool_size,
         internal,
         use_jump_host,
         image_tag_from_ref,
     )
+
+
+def get_integration_cli_meta() -> dict[str, IntegrationMeta]:
+    integration_meta = {}
+    for integration_name in integration.list_commands(None):  # type: ignore
```
janboll, line 1920: That's some tricky meta programming. Wonder if it should actually be in the schema? Please at least add some comments.

janboll (approved):
Looks good, just left some nits

reconcile/utils/runtime/__init__.py
```diff
@@ -0,0 +1,8 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class IntegrationMeta:
```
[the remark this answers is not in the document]
geoberle: we need to know what cli flags an integration supports. we use this information to fail the integration manager during PR checks if a autoscaling strategy has been selected that is not supported by the integration (e.g. not supporting an `--account-name` flag)

i was thinking about implementing a function `get_metadata()` on each integration (like `run()`) to allow an integration to expose metadata in a more explicit way but that was too big for the scope of this change. i think we need to refactor the integration startup code somehow to make all of this a bit more robust. e.g. cli.py and the `run` function of each integration need to be very much aligned on the params they accept/require, which is redundant :(

we might night something like this for the go integrations repo as well. let's align on how to do this cross-repo/language

if it is ok for you, i'm not going to change this metadata logic for now.

reconcile/cli.py
geoberle: i left a comment

PR #2429, opened 2022-05-25
[openshift-saas-deploy] enable option to not use jump host

maorfr:
right now we are (implicitly) using the jump host for all our deployments.

with this PR, we add an option to _not_ use the jump host.


geoberle (approved):
LGTM

PR #2430, opened 2022-05-25
openshift ignore status on object compare only at the root level

maorfr:
if a resource has a `status` section, it should be ignored when comparing current with desired resource.
if a resource has a `status` section deep under another field, it needs to be compared. example: CRDs.

this PR introduces a concept of `depth` into our compare logic. if `status` is on depth 0, it needs to be ignored.

slack thread: https://coreos.slack.com/archives/CCRND57FW/p1653423187653079

### DIFF
--- a/reconcile/utils/openshift_resource.py
+++ b/reconcile/utils/openshift_resource.py
@@ -78,6 +78,8 @@ class OpenshiftResource:
                     pass
                 elif self.ignorable_key_value_pair(obj1_k, obj1_v):
                     pass
+                elif depth == 0 and obj1_k == "status":
+                    pass
                 elif obj1_k == "labels":
                     diff = [
                         k
@@ -140,7 +142,6 @@ class OpenshiftResource:
             "generation",
             "selfLink",
             "uid",
-            "status",
             "fieldRef",
         ]
         if val in ignorable_fields:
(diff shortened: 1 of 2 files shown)


maorfr:
this is an example for a PR that we failed to apply: https://github.com/openshift/hive/pull/1775

reconcile/utils/openshift_resource.py
```diff
@@ -140,7 +142,6 @@ def ignorable_field(val):
             "generation",
             "selfLink",
             "uid",
-            "status",
```
geoberle, line 143: oh i see. they key point is this here. we ignored status fields whereever they were defined. but what we wanted to do was only ignore it when it was the regular `status` section of a K8S resource.

geoberle (approved):
LGTM

reconcile/utils/openshift_resource.py
```diff
@@ -87,21 +89,21 @@ def obj_intersect_equal(self, obj1, obj2):
                             self.kind, k
                         )
                     ]
-                    if diff or not self.obj_intersect_equal(obj1_v, obj2_v):
+                    if diff or not self.obj_intersect_equal(obj1_v, obj2_v, depth + 1):
```
2uasimojo, line 92: nit: It's a python best practice to use keys for kwargs
```suggestion
                    if diff or not self.obj_intersect_equal(obj1_v, obj2_v, depth=depth + 1):
```

2uasimojo (commented):
Love this!

If we wanted to get really clever (and complicated) we could inspect the CRD for the object in question and ignore all and only fields listed in `subresources` at `depth==0`.

PR #2431, opened 2022-05-26
Added rule length check to prometheus rules tester

rporres:
Prometheus operator establishes a limit on the size of a single
 PrometheusRule based on the maximum size of a ConfigMap. From that
 moment on it will stop processing new rules.

 Since we have rules that are heavily templated via queries, it is not
 that uncommon that we get to that limit. This will protect us from that
 to happen.

Part of APPSRE-5640

### DIFF
--- a/reconcile/prometheus_rules_tester.py
+++ b/reconcile/prometheus_rules_tester.py
@@ -17,6 +17,10 @@ from reconcile.utils.semver_helper import make_semver
 from reconcile.status import ExitCodes
 from reconcile.utils.structs import CommandExecutionResult
 
+# This comes from prometheus-operator. It is the largest configmap that they will
+# create. It is also the largest PrometheusRule yaml file they will process to add it
+# to the configmap.
+MAX_CONFIGMAP_SIZE = 0.5 * 1024 * 1024
 
 QONTRACT_INTEGRATION = "prometheus_rules_tester"
 QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
@@ -66,8 +70,8 @@ def get_prometheus_rules(cluster_name):
         rules[r["path"]] = {}
 
     for n in gqlapi.query(orb.NAMESPACES_QUERY)["namespaces"]:
-        cluster = n["name"]
-        namespace = n["cluster"]["name"]
+        namespace = n["name"]
+        cluster = n["cluster"]["name"]
 
         if cluster_name and cluster != cluster_name:
             continue
@@ -101,7 +105,12 @@ def get_prometheus_rules(cluster_name):
             if cluster not in rules[path]:
                 rules[path][cluster] = {}
 
-            rules[path][cluster][namespace] = {"spec": openshift_resource.body["spec"]}
+            rule = openshift_resource.body
+            rule_length = len(yaml.dump(rule))  # Same as prometheus-operator does it.
+            rules[path][cluster][namespace] = {
+                "spec": rule["spec"],
+                "length": rule_length,
+            }
 
             # we keep variables to use them in the rule tests
             variables = json.loads(r.get("variables") or "{}")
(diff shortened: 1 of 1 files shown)


rporres:
I know this PR doesn't have tests with it, but honestly, it is not worth adding them at this stage. This integration should be refactored a bit. See https://issues.redhat.com/browse/APPSRE-5699 for details

reconcile/prometheus_rules_tester.py
```diff
@@ -66,8 +70,8 @@ def get_prometheus_rules(cluster_name):
         rules[r["path"]] = {}
 
     for n in gqlapi.query(orb.NAMESPACES_QUERY)["namespaces"]:
-        cluster = n["name"]
-        namespace = n["cluster"]["name"]
+        namespace = n["name"]
+        cluster = n["cluster"]["name"]
```
rporres, line 74: I know this doesn't belong to this, but messages are quite confusing and `--cluster-name` option doesn't currently work. I can move it to a different PR if reviewers prefer.

reconcile/prometheus_rules_tester.py
```diff
@@ -17,6 +17,10 @@
 from reconcile.status import ExitCodes
 from reconcile.utils.structs import CommandExecutionResult
 
+# This comes from prometheus-operator. It is the largest configmap that they will
+# create. It is also the largest PrometheusRule yaml file they will process to add it
+# to the configmap.
+MAX_CONFIGMAP_SIZE = 1 * 1024 * 1024 / 2
```
lechuk47, line 23: Is there any reason to multiply it by 1? 
```suggestion
MAX_CONFIGMAP_SIZE = (1024 * 1024) / 2
```

rporres: Not many, it came from the combination kubernetes and prometheus operator code :-D

* https://github.com/kubernetes/kubernetes/blob/976a940f4a4e84fe814583848f97b9aafcdb083f/staging/src/k8s.io/api/core/v1/types.go#L6112
* https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/rules.go#L45

I guess people like to separate unit conversions from values which makes kind of sense. In that spirit I have rewritten it as `0.5 * 1024 * 1024`  in a0a7f9bc

reconcile/prometheus_rules_tester.py
```diff
@@ -155,10 +164,23 @@ def check_valid_services(rule):
     return CommandExecutionResult(True, "")
 
 
+# We return here a CommandExecutionResult as it is what check_rule function has to
+# add to the "check_result" field.
+def check_rule_length(rule_length: int) -> CommandExecutionResult:
+    if rule_length > MAX_CONFIGMAP_SIZE:
+        return CommandExecutionResult(
+            False, f"Rule larger than {MAX_CONFIGMAP_SIZE} bytes."
```
lechuk47, line 172: I think the message could be a bit confusing. I understand the problem is related to all the rules within the configmap (Alerting or recording rules) not a single rule. This could be an already existing typo as `check_rule` also seems to check the whole spec, not a single rule. 

```suggestion
            False, f"Rules spec is larger than {MAX_CONFIGMAP_SIZE} bytes."
```

rporres: I like it. Thx!

reconcile/response.txt
```diff
@@ -0,0 +1,7 @@
+Not trying to engage on the debate, I accept the decision. Just pointing out where I don't agree:
```
fishi0x01, line 1: What is this file about? :sweat_smile:

rporres: error 😂

rporres: It was a text file with arguments for a discussion 😂 

removed

fishi0x01 (approved):

PR #2433, opened 2022-05-26
Include state bucket account parameter

dwelch0:


maorfr:
can you add information on what you are trying to accomplish?

dwelch0:
> can you add information on what you are trying to accomplish?

Deployment of the integration manager within FedRamp caused all stateful integrations to fail due to APP_INTERFACE_STATE_BUCKET_ACCOUNT defaulting to `app-sre`. This variable is explicitly set to `osd-fedramp-app-sre` in the saas file for the fedramp environment but appears to have no effect. I manually edited the integration manager deployment to include APP_INTERFACE_STATE_BUCKET_ACCOUNT within the container and set to `osd-fedramp-app-sre`. This triggered integration manager to roll (?) the effected integrations (all of which were the stateful ones) and set the correct value within each. 

I concede that I still do not fully understand the integration manager relationship so I may be barking up the wrong tree with this PR.

PR #2434, opened 2022-05-29
[terraform-resources] only disable keys when tf should apply

maorfr:
`tf.should_apply` should be true after creating an access key (after a leaked one was removed from state).
in this case, it should disable keys. in any other case, we can skip this call altogether.


geoberle (approved):
MSTM

PR #2436, opened 2022-05-31
Add sharding by saas file

janboll:
Implements sharding as required per: [APPSRE-5705](https://issues.redhat.com/browse/APPSRE-5705)

### DIFF
--- a/reconcile/openshift_tekton_resources.py
+++ b/reconcile/openshift_tekton_resources.py
@@ -14,7 +14,7 @@ from reconcile.utils.defer import defer
 from reconcile.utils.openshift_resource import OpenshiftResource as OR
 from reconcile.utils.semver_helper import make_semver
 from reconcile.utils.saasherder import Providers
-from reconcile.utils.sharding import is_in_shard_round_robin
+from reconcile.utils.sharding import is_in_shard
 
 LOG = logging.getLogger(__name__)
 QONTRACT_INTEGRATION = "openshift-tekton-resources"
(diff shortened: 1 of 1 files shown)


reconcile/openshift_tekton_resources.py
```diff
@@ -14,6 +14,7 @@
 from reconcile.utils.openshift_resource import OpenshiftResource as OR
 from reconcile.utils.semver_helper import make_semver
 from reconcile.utils.saasherder import Providers
+from reconcile.utils.sharding import is_in_shard_round_robin
```
maorfr, line 17: why did you choose round robin?

janboll: You got me thinking about an issue that I was not aware of. The list must be sorted consistently. I did not think about that when opening this PR. Let me reconsider using `is_in_shard` or have the list sorted.

rporres: So assuming qontract-server always returns things in the same order (which is a lot to assume) the potential danger here is the moment which we add a new file and we could have two shards handling the same file, right?

rporres: I agree with @maorfr that we should use `is_in_shard` FWIW

janboll: I updated the PR!

maorfr (approved):

rporres (approved):

PR #2437, opened 2022-05-31
[cli] add stats commands for resource usage

maorfr:
these 2 commands show how many resources are managed via app-interface.


geoberle (approved):
❤️

PR #2442, opened 2022-06-01
Use latest image

janboll:


geoberle:
i think `amtool` has been changed to not fail anymore if not called with expected params or configfiles. instead it exists normally. therefore we need to change `expectedError` to `expectedOutput` in `dockerfiles/structure-test.yaml`

geoberle:
we could promote https://github.com/app-sre/qontract-reconcile/pull/2450 along with this PR

PR #2449, opened 2022-06-02
Logic to retrieve bucket information from new terraform state schema

suzana-nesic:
We have implemented a new schema change that displays terraform state information through an account file

Schema setup for a file tf-state-test.yml:
```
---
$schema: /dependencies/terraform-state-1.yml
provider: s3
bucket: someBucket
region: someRegion
integrations:
- integration: terraform-resources
  key:  qontract-reconcile.tfstate
- integration: terraform-tgw-attachements
  key: qontract-reconcile-tgw-attachments.tfstate
```

AWS account file call:
```
terraformState:
  $ref: /aws/app-sre-stage/tf-state-test.yml
```

This MR will implement logic that will check if the new schema changes are present within app-interface and if it is, it will execute logic to retrieve bucket information the new way (looking within app-interface and extracting the information) or the current way (looking within vault to retrieve that information).

Part of [APPSRE-4516](https://issues.redhat.com/browse/APPSRE-4516)

### DIFF
--- a/reconcile/test/test_utils_terrascript_aws_client.py
+++ b/reconcile/test/test_utils_terrascript_aws_client.py
@@ -145,3 +145,28 @@ def test_resource_specs_with_account_filter(ts):
     specs = ts.resource_spec_inventory
     spec = ExternalResourceSpec(p, pa, ra, ns1)
     assert specs == {ExternalResourceUniqueKey.from_spec(spec): spec}
+
+
+def test_terraform_state(ts):
+    account_name = "some-account"
+    integration_name = "terraform-resources-wrapper"
+    config_test = {
+        "aws_access_key_id": "SOMEKEY123",
+        "aws_secret_access_key": "somesecretkey",
+        "bucket": "some-bucket",
+        "key": "qontract-reconcile.tfstate",
+        "region": "us-east-1",
+        "terraform-resources-wrapper_key": "qontract-reconcile.tfstate",
+        "terraformState": {
+            "provider": "s3",
+            "bucket": "some-bucket",
+            "region": "some-region",
+            "integrations": [
+                {
+                    "key": "terraform-resources-wrapper",
+                    "integration": "qontract-reconcile.tfstate",
+                },
+            ],
+        },
+    }
+    assert ts.state_bucket_for_account(integration_name, account_name, config_test)
--- a/reconcile/test/test_utils_terrascript_aws_client.py
+++ b/reconcile/test/test_utils_terrascript_aws_client.py
@@ -147,10 +147,21 @@ def test_resource_specs_with_account_filter(ts):
     assert specs == {ExternalResourceUniqueKey.from_spec(spec): spec}
 
 
-def test_terraform_state(ts):
+expected_result = {
+    "s3": {
+        "access_key": "SOMEKEY123",
+        "secret_key": "somesecretkey",
+        "bucket": "some-bucket",
+        "key": "qontract-reconcile.tfstate",
+        "region": "us-east-1",
+    }
+}
+
+
+def test_terraform_state_when_present(ts):
     account_name = "some-account"
     integration_name = "terraform-resources-wrapper"
-    config_test = {
+    terraform_state_config_test = {
         "aws_access_key_id": "SOMEKEY123",
         "aws_secret_access_key": "somesecretkey",
         "bucket": "some-bucket",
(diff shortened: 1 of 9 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -224,12 +240,23 @@ def __init__(self, integration: str, integration_prefix: str,
...
+                terraform_state = account.get("terraform_state")
+                if not terraform_state:
+                    b = Backend("s3",
+                                access_key=config['aws_access_key_id'],
+                                secret_key=config['aws_secret_access_key'],
+                                bucket=config['bucket'],
+                                key=config['{}_key'.format(integration)],
+                                region=config['region'])
+                else:
+                    integration = terraform_state['integrations']['integration']
```
geoberle, line 253: `terraform_state["integrations"]` is a list so you need to find the one that matches the currently executed integration provided as the `integration` variable to this function.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -102,6 +102,22 @@
 import reconcile.utils.aws_helper as awsh
 
 
+# TERRAFORM_STATE_QUERY = """
```
janboll, line 105: This comment confused me for a couple of seconds. Do we need to keep it?

reconcile/utils/terrascript_aws_client.py
```diff
@@ -224,12 +240,23 @@ def __init__(self, integration: str, integration_prefix: str,
...
 
-            b = Backend("s3",
-                        access_key=config['aws_access_key_id'],
-                        secret_key=config['aws_secret_access_key'],
-                        bucket=config['bucket'],
-                        key=config['{}_key'.format(integration)],
-                        region=config['region'])
+            for account in filtered_accounts:
+                terraform_state = account.get("terraform_state")
+                if not terraform_state:
```
janboll, line 245: Would suggest to move declaration of:
- access_key
- secret_key

out of if statement, cause they are the same for both cases.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -224,12 +240,23 @@ def __init__(self, integration: str, integration_prefix: str,
...
+                    b = Backend("s3",
+                                access_key=config['aws_access_key_id'],
+                                secret_key=config['aws_secret_access_key'],
+                                bucket=config['bucket'],
+                                key=config['{}_key'.format(integration)],
```
janboll, line 250: Please use f-string syntax

reconcile/utils/terrascript_aws_client.py
```diff
@@ -224,12 +240,23 @@ def __init__(self, integration: str, integration_prefix: str,
...
+                                region=config['region'])
+                else:
+                    integration = terraform_state['integrations']['integration']
+                    b = Backend("s3",
+                                access_key=config['aws_access_key_id'],
+                                secret_key=config['aws_secret_access_key'],
+                                bucket=account.get("bucket"),
+                                key=integration + "_key",
```
janboll, line 258: Please use f-string syntax

janboll (changes requested):

janboll (dismissed):

reconcile/utils/terrascript_aws_client.py
```diff
@@ -231,13 +233,37 @@ def __init__(self, integration: str, integration_prefix: str,
...
-
-            b = Backend("s3",
-                        access_key=config['aws_access_key_id'],
-                        secret_key=config['aws_secret_access_key'],
-                        bucket=config['bucket'],
-                        key=config['{}_key'.format(integration)],
-                        region=config['region'])
+            access_key_value = config['aws_access_key_id']
+            secret_key_value = config['aws_secret_access_key']
+            accounts = queries.get_aws_accounts(terraform_state=True)
```
geoberle, line 238: why query for the accounts again? filtered_accounts already holds the accounts queries in terraform_resources.py and you could query with `terraform_state=True` there

https://github.com/app-sre/qontract-reconcile/blob/826cc8683cba0dccbe9fae677d4c2f04d3c5828c/reconcile/terraform_resources.py#L467

reconcile/utils/terrascript_aws_client.py
```diff
@@ -231,13 +233,37 @@ def __init__(self, integration: str, integration_prefix: str,
...
+            for filtered_account in filtered_accounts:
+                for account in accounts:
```
geoberle, line 240: remember that you are in a loop over `configs` here. everything you do here, you do for each account, e.g. you create the `Backend` filtered_accounts*filtered_accounts times.

have a look at the context here. the variable `name` is bound to the AWS account name your are currently iterating over. so maybe something like the following could work. please verify with a dry run though

```
for acc in filtered_account:
    if acc.get("name") == name:
        # check on terraformState variable in acc (assuming filtered_accounts was created with  terraform_state=True 
        # and create Backend accordingly
```

reconcile/utils/terrascript_aws_client.py
```diff
@@ -288,13 +288,41 @@ def __init__(
 
             ts += provider.template(version="2.2.0")
 
+            access_key_backend_value = config["aws_access_key_id"]
+            secret_key_backend_value = config["aws_secret_access_key"]
+            bucket_backend_value = ""
+            key_backend_value = ""
+            region_backend_value = ""
```
geoberle, line 295: you don't need to declare the variables here. in python, local variables are scoped within the function, so assigning them later in this function will make them available during Backend creation.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -288,13 +288,41 @@ def __init__(
...
+                            transformed_integration = str(integration_value).replace(
+                                "-", "_"
+                            )
+                            if transformed_integration == self.integration:
+                                bucket_backend_value = terraform_state.get("bucket")
+                                key_backend_value = str(key_value)
+                                region_backend_value = terraform_state.get("region")
+                            else:
+                                continue
+
```
geoberle, line 318: consider using the `bucket`, `region` and `key` fields from the account if no 'key' is defined within `terraformState`.
or fail the integration if you `terraformState` is present but no `key` is defined for an integration.

geoberle: with "no key defined" i meant if the terraformState does not contain information for an integration, try to discover the required information the old way

geoberle (commented):
tiny nit on the test. other great and LGTM!

reconcile/test/test_utils_terrascript_aws_client.py
```diff
@@ -145,3 +145,28 @@ def test_resource_specs_with_account_filter(ts):
...
+            "region": "some-region",
+            "integrations": [
+                {
+                    "key": "terraform-resources-wrapper",
+                    "integration": "qontract-reconcile.tfstate",
+                },
+            ],
+        },
+    }
+    assert ts.state_bucket_for_account(integration_name, account_name, config_test)
```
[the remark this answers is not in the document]
suzana-nesic: Added a couple of more tests and changed to compare the values

geoberle (approved):
LGTM

PR #2450, opened 2022-06-02
Bump reconcile builder image

andreugallofre:


BumbleFeng:
I think you also need to update https://github.com/app-sre/qontract-reconcile/blob/master/dockerfiles/Dockerfile#L18 and  https://github.com/app-sre/qontract-reconcile/blob/master/dockerfiles/Dockerfile#L47

andreugallofre:
Ported the changes from #2442 and now we can do the promotion in one go, @janboll ptal when you can 😄

geoberle:
@andreugallofre the PR check is still failing because of `amtool` behaving a bit differently. see https://github.com/app-sre/qontract-reconcile/pull/2442#issuecomment-1149845535

chassing:
superseded by #2463

PR #2452, opened 2022-06-03
APPSRE-5552 Slack notification on cluster upgrade completion

patjlm:
The choice here is to notify the slack channel when we create the MR to bump the cluster version.
This depends on https://github.com/app-sre/qontract-schemas/pull/166 and /data/dependencies/slack/coreos.yml needs to be updated in app-interface to grant ocm-clusters access to slack:

An other option would be to amend the `openshift-ugrade-watcher` integration. That would mean additional calls to OCM to get the version from there.

### DIFF
--- a/reconcile/ocm_clusters.py
+++ b/reconcile/ocm_clusters.py
@@ -5,9 +5,12 @@ import semver
 from reconcile import queries
 
 from reconcile import mr_client_gateway
+from reconcile.slack_base import slackapi_from_queries
 import reconcile.utils.mr.clusters_updates as cu
 
 import reconcile.utils.ocm as ocmmod
+import reconcile.utils.slack_api as slackapimod
+import reconcile.utils.state as statemod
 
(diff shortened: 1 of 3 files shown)


janboll:
I really like the feature! But I find it confusing, that it is not in `openshift-ugrade-watcher`

lechuk47:
I got rid of testlide in #2425 as we are pivoting towards pytest only. could it be considered here?

patjlm:
> I got rid of testlide in #2425 as we are pivoting towards pytest only. could it be considered here?

Sure, I'd adapt on top of your code @lechuk47 .

> I really like the feature! But I find it confusing, that it is not in `openshift-ugrade-watcher`

I agree @janboll and that's why I mentioned it in the description. If many feel like it's not the proper way, I'll be happy to revisit.

patjlm:
Closing this in favor of #2479

PR #2453, opened 2022-06-06
[openshift-network-policies] use kubernetes.io/metadata.name instead of name

maorfr:
to create NetworkPolicies, we rely on a `name` label present on the namespace/project.

this label is added to specific projects (as can be seen in this example: https://github.com/openshift/managed-cluster-config/pull/374)

automatic labeling was added in Kubernetes 1.21: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#automatic-labelling
OpenShift version 4.8 uses Kubernetes 1.21: https://docs.openshift.com/container-platform/4.8/release_notes/ocp-4-8-release-notes.html

this PR changes the openshift-network-policies integration to create a NetworkPolicy using the new automatically added label instead of relying on a `name` label to exist.

### DIFF
--- a/reconcile/openshift_network_policies.py
+++ b/reconcile/openshift_network_policies.py
@@ -61,7 +61,17 @@ def construct_oc_resource(name, source_ns):
         "metadata": {"name": name},
         "spec": {
             "ingress": [
-                {"from": [{"namespaceSelector": {"matchLabels": {"kubernetes.io/metadata.name": source_ns}}}]}
+                {
+                    "from": [
+                        {
+                            "namespaceSelector": {
+                                "matchLabels": {
+                                    "kubernetes.io/metadata.name": source_ns
+                                }
+                            }
+                        }
+                    ]
+                }
             ],
             "podSelector": {},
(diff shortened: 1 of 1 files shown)


steveteahan:
LGTM - the only thing I was wondering was whether we had to handle any older versions of OpenShift. It seems that 4.7 still has maintenance support until August? But that's probably not realistically a concern.

steveteahan (approved):

rporres:
This should work fine for all of our clusters, as we're in 4.10. What do we know about the bring-your-own clusters? app-interface doesn't have the version

maorfr:
fedramp: https://coreos.slack.com/archives/C03AKQD1SHF/p1654522352145289
consoleDot: https://coreos.slack.com/archives/CCRND57FW/p1654606691597589
rhobs: https://coreos.slack.com/archives/CCRND57FW/p1654606961903439

maorfr:
followed up by https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/40481

PR #2457, opened 2022-06-09
Add loggin instance to jb-cleaner

skryzhny:
We need to know what instance has builds to delete

### DIFF
--- a/reconcile/jenkins_job_builds_cleaner.py
+++ b/reconcile/jenkins_job_builds_cleaner.py
@@ -83,7 +83,7 @@ def run(dry_run):
             )
 
         token = instance["token"]
-        instance_name = instance.get("name")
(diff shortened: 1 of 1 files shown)


reconcile/jenkins_job_builds_cleaner.py
```diff
@@ -83,11 +83,12 @@ def run(dry_run):
             )
 
         token = instance["token"]
+        instance_name = instance.get("name")
```
geoberle, line 86: the instance is supposed to always have a name, so you could use `instance["name"]` and let the code fail if that would be not the case (for whatever reason)

geoberle: just a nit - feel free to ignore

geoberle (approved):
lgtm

PR #2458, opened 2022-06-09
Add request timeouts to QuayApi

steveteahan:
Without a timeout in requests, it will default to no timeout, which
can cause requests to hang forever. This is suspected to be the cause
of some of the Quay integrations being stuck for hours.

## Testing

Ran the `quay-mirror-org` integration in dry-run mode to ensure there weren't any major issues.


geoberle (approved):
thank you!

PR #2459, opened 2022-06-11
[integration-manager] shard_id metric label as identifier for workload buckets

geoberle:
shard_id has been an integer to identify a bucket of workloads within a collection of shards. since a hashing strategy is used to distribute workload based on the number of buckets, shard_id does not necessarily represent a stable bucket of workload but needs the `shards` label along with it.

this change seeks to make shard_id a true identifier for a workload bucket by allowing it to be one of the following
* shard-nrOfShards, e.g. 2-18
* sharding key, e.g. an aws account

the first option will be enabled by standard sharding (`static` strategy) while the second one will be served by other sharding strategies like the `per-aws-account` strategy.

alerting rule grouping based on `integration` and `shard_id` will be unaffected by this change (maybe short hickups during an initial rollout). also, the extra label feature, used exclusively by terraform-resources to accommodate for the wrapper functionality, will be obsolete.

obsoletes #2406

### DIFF
--- a/dockerfiles/hack/run-integration.py
+++ b/dockerfiles/hack/run-integration.py
@@ -13,12 +13,13 @@ import click
 from reconcile.status import ExitCodes
 from reconcile.cli import LOG_FMT, LOG_DATEFMT
 from reconcile.utils.metrics import (
-  run_time, run_status, extra_labels, execution_counter
+  run_time, run_status, execution_counter
 )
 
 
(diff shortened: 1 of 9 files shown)


lechuk47:
LGTM

patjlm (approved):
LGTM

maorfr (approved):
LGTM

thank you for reverting all my extra_labels hacks :heart:

PR #2460, opened 2022-06-12
enable terraform-users to be sharded per-aws-account

maorfr:

to be sharded per aws account, an integration should expose a `--account-name` arg. this work was already done in #2176, and this is just exposing the argument for the cli.


maorfr:
required for https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/40615

PR #2461, opened 2022-06-13
remove terraform wrapper integrations

maorfr:

these integrations are no longer in use following #2428


geoberle (approved):
whohooo

PR #2464, opened 2022-06-15
openshift-vault-secrets: do not continuously sync empty secrets

patjlm:
ref APPSRE-5810

**Problem statement**
When a vault secret is empty (or has no field with non-empty values), we currently sync it at each iteration.
That's because we have this in the desired state
```
  data: {}
```
but the data field is not even present in the current state.

Note that this occurs also when all secret 'values' in vault are set to an empty string as we just ignore those today. This could be challenged and part of an other change.

**Implementation proposal**
do not add the `data` field in the desired state when there is no secret key being set

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -403,8 +403,6 @@ def fetch_provider_vault_secret(
     }
     if labels:
         body["metadata"]["labels"] = labels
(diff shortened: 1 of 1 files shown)


jfchevrette (approved):
LGTM! I have also tested this and made the same observations

PR #2447, opened 2022-06-02
External resources enhancement

maorfr:
part of https://issues.redhat.com/browse/APPSRE-5711

following up on prep work in #2444

this PR (mostly) adds the `externalResources` section to queries. this "enhances" the integrations so they can work with both `terraformResources` and `externalResources`.

PR #2465, opened 2022-06-15
deprecate terraformResources

maorfr:
part of https://issues.redhat.com/browse/APPSRE-5726

this is essentially the opposite of #2447

once all accounts in all app-interface instances were migrated to externalResources, we can deprecate them terraformResources (this PR).

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -903,36 +903,6 @@ NAMESPACES_QUERY = """
         email
       }
     }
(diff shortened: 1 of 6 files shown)


maorfr:
this PR is just an initial cleanup to stop supporting terraformResources.

the work to generalize things to support additional providers will be done in https://issues.redhat.com/browse/APPSRE-5833

PR #2467, opened 2022-06-16
use .version.raw_id for OSD cluster version

patjlm:
This allows to update the cluster version only once all nodes are upgraded

### DIFF
--- a/reconcile/utils/ocm.py
+++ b/reconcile/utils/ocm.py
@@ -153,7 +153,7 @@ class OCMProductOsd(OCMProduct):
             provider=cluster["cloud_provider"]["id"],
             region=cluster["region"]["id"],
             channel=cluster["version"]["channel_group"],
-            version=cluster["openshift_version"],
+            version=cluster["version"]["raw_id"],
             multi_az=cluster["multi_az"],
             instance_type=cluster["nodes"]["compute_machine_type"]["id"],
             storage=cluster["storage_quota"]["value"] // BYTES_IN_GIGABYTE,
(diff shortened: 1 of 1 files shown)


reconcile/utils/ocm.py
```diff
@@ -320,6 +320,7 @@ def get_ocm_spec(
             provider=cluster["cloud_provider"]["id"],
             region=cluster["region"]["id"],
             channel=cluster["version"]["channel_group"],
+            # should this be version=cluster["version"]["raw_id"] as for OSD ?
```
lechuk47, line 323: Yes. there are two fixtures with the payloads returned by OCM in the tests dir: https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/fixtures/clusters/rosa_spec.json#L156-L159

patjlm: I was just wondering if the ROSA versioning would  follow the same behavior on upgrade (`.openshift_version` getting bumped before the upgrade is actually finished, `.version.raw_id` getting bumped only once all nodes are upgraded)

patjlm: My guess is that it is the same process being applied, so I'll push the same setting on ROSA.

patjlm: 9ce8dcb

janboll:
What is the difference between the two fields?

patjlm:
> What is the difference between the two fields?

This was already implemented in https://github.com/app-sre/qontract-reconcile/pull/2408 and has been mistakenly reverted as part of the ROSA implementation in https://github.com/app-sre/qontract-reconcile/pull/2425. It will enable us later to notify new version readiness properly.

The big difference is "when" the value gets updated during an upgrade. `.openshift_version` gets updated a bit after the openshift control plane is upgraded while `.version.raw_id` gets updated when the upgrade is fully completed, including node upgrade.

I did not spot any other difference than that.

geoberle (approved):
LGTM

PR #2470, opened 2022-06-17
terraform-resources: set S3 bucket ACL for cloudfront logging

patjlm:
APPSRE-5533

New tentative after the revert of https://github.com/app-sre/qontract-reconcile/pull/2423.

The problem was that `data.aws_canonical_user_id("current")` was added only when `--account-name` was set and not for each account shard.

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -132,6 +132,10 @@ class aws_ecrpublic_repository(Resource):
     pass
 
 
+class aws_s3_bucket_acl(Resource):
+    pass
+
+
 # temporary until we upgrade to a terrascript release
 # that supports this provider
 # https://github.com/mjuenema/python-terrascript/pull/166
@@ -239,6 +243,9 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
 
         self.locks: dict[str, Lock] = locks
         """AWS account name to Lock mapping."""
+
+        for name in self.tss:
+            self.add_resource(name, data.aws_canonical_user_id("current"))
 
         self.accounts = {a['name']: a for a in filtered_accounts}
         self.uids = {a['name']: a['uid'] for a in filtered_accounts}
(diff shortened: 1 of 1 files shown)


geoberle (approved):
LGTM

reconcile/utils/terrascript_aws_client.py
```diff
@@ -2451,6 +2464,40 @@ def populate_tf_resource_s3_cloudfront(self, resource, namespace_info):
...
+                        'permission': 'FULL_CONTROL',
+                    },
+                    {
+                        'grantee': {
+                            # well known id for cloudfront logs delivery:
+                            # https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html#AccessLogsBucketAndFileOwnership
+                            'id': 'c4c1ede66af53448b93c283ce9448c4ba468c9432aa01d700d3878632f77d2d0',
+                            'type': 'CanonicalUser',
+                        },
+                        'permission': 'FULL_CONTROL',
```
lechuk47, line 2491: Is this really needed? 
I understand the ACL is set to the bucket when the CF distribution is created. I think we are using the same user to create the bucket and the distribution so, it won't be enough with the other grant?

patjlm: If we don't set it here, it might be added by CF (not sure with which permissions, it's unclear to me) but will then be removed by terraform runs, right ?

lechuk47: That makes sense

lechuk47 (approved):

patjlm:
The last push [7e17dc1](https://github.com/app-sre/qontract-reconcile/pull/2470/commits/7e17dc1f54a5212e055952806b14bd013d09b27b) uses a terraform datasource which is available [since aws provider 3.60](https://github.com/hashicorp/terraform-provider-aws/pull/15167) to get the cloudfront log delivery canonical user.

Since this would not be working on aws provider <3.60, the datasource is used in the logging_config portion of the s3-cloudfront setup, which already requires >3.75. The name is specific to the s3-cloudfront config, so we can have multiple ones.

PR #2471, opened 2022-06-19
retry harder to get file contents

maorfr:
related to https://issues.redhat.com/browse/APPSRE-4854

whlie we investigate why github connection errors occur, we can retry harder.
the default number of attempts is 3.


chassing:
LGTM

PR #2472, opened 2022-06-19
rename TerraformResourceSpec to ExternalResourceSpec

maorfr:

this PR is a rename with no logic changes. it gets the terraform specific specs to be more similar to external resources implemented in #2447.

in a follow up PR, we will likely combine the two together, forming a winning alliance!


geoberle (approved):
LGTM

PR #2475, opened 2022-06-20
Used black's force-exclude instead of extend-exclude.

chassing:
When using VS code black plugin, because the filename is passed
explicitly, the file referred to in `extend-exclude` is not properly
excluded. Use `force-exclude` instead to say we really mean it.

see https://github.com/psf/black/issues/438


geoberle (approved):
LGTM

PR #2479, opened 2022-06-21
Slack notification on cluster upgrade completion

patjlm:
Ref APPSRE-5552

We currently only notify Slack when a cluster upgrade starts. This adds a notification when the cluster upgrade completes, based on the fact that OCM gets the `version.raw_id` updated only at that time (ref #2408, #2467).

Contrarily to #2452 , this implementation is done in `openshift_upgrade_watcher`, which is more coherent.

Notes:
- during the first production loop, we will get a notification for each cluster we currently run.
- Since we query ocm clusters, the following schemas need to be granted in the `openshift-upgrade-watcher` app-interface integration defintion (https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/41318)
```
- /aws/group-1.yml
- /aws/vpc-1.yml
- /openshift/cluster-addon-1.yml
```

### DIFF
--- a/reconcile/openshift_upgrade_watcher.py
+++ b/reconcile/openshift_upgrade_watcher.py
@@ -1,11 +1,14 @@
 import logging
 
 from datetime import datetime
+from typing import Optional
 
 from reconcile import queries
 
 from reconcile.slack_base import slackapi_from_queries
 from reconcile.utils.oc import OC_Map
+import reconcile.utils.ocm as ocmmod
+from reconcile.utils.slack_api import SlackApi
 from reconcile.utils.state import State
 from reconcile.utils.defer import defer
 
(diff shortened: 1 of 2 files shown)


reconcile/openshift_upgrade_watcher.py
```diff
@@ -1,39 +1,51 @@
...
-    )
+def cluster_slack_handle(cluster: str, slack: Optional[SlackApi]):
+    usergroup = f"{cluster}-cluster"
+    usergroup_id = f"@{usergroup}"
+    if slack:
+        usergroup_id = slack.get_usergroup_id(usergroup)
+    return f"<!subteam^{usergroup_id}>"
+
+
+def slack_notify(
```
rporres, line 27: So this function does more than notify, correct? which makes the name of it quite misleading as I didn't expect it to handle the logic of both the state and the message in one go. 

It is very convenient to have it here as it makes the code very similar in both cases... But I would love a name that describes what it does, something about handling the state and notifying...

Maybe something like this, if you like it

```suggestion
def handle_slack_notification(
    """Check notification status, notify if needed and update the notification status"""
```

rporres (commented):
a nit in a function name, it looks great otherwise

PR #2481, opened 2022-06-22
[terrascript] use external resource spec in populate methods

maorfr:

this PR aims to replace the current arguments that are currently passed to terrascript populate method: resource, namespace_info (existing_secrets will be replaced in a different PR)
instead, we will pass the external resource spec, which will contain all the needed information.


maorfr:
not adding type hints because they create a lot of errors. will do that in a future PR.

geoberle (approved):
LGTM. nice to see the spec object usage!

PR #2485, opened 2022-06-22
APPSRE-5903 fix ES publish_log_type

fishi0x01:
The `account` key is no longer part of all resource specs.
However, the resource specs are grouped by account -> using the grouping key to obtain account.

**Tests**

- successfully executed with dev-data
- dry-run with prod data didnt show any changes


maorfr:
i'd suggest to use `spec.provisioner_name` instead of `account`.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -3385,7 +3385,7 @@ def _elasticsearch_get_all_log_group_infos(
                 log_types = res['publish_log_types'] or []
                 for log_type in log_types:
                     region = ns['cluster']['spec']['region']
-                    account = res['account']
+                    account = res['provisioner_name']
```
maorfr, line 3388:
```suggestion
                    account = spec.provisioner_name
```

reconcile/utils/terrascript_aws_client.py
```diff
@@ -3385,7 +3385,7 @@ def _elasticsearch_get_all_log_group_infos(
...
                     account_id = self.accounts[account]['uid']
```
maorfr, line 3389: and here as well?

maorfr:
you can instead set `account = spec.provisioner_name` somewhere at the top of the loop.

fishi0x01:
@maorfr makes sense. I added it to top of the loop. Ran prod data in dry-run one more time -> no changes :+1:

PR #2486, opened 2022-06-22
Cert-manager certificate along with Routes

lechuk47:


reconcile/openshift_resources_base.py
```diff
@@ -596,6 +602,48 @@ def fetch_current_state(
         )
 
 
+def _post_process_route(
+    oc: OCClient,
+    namespace: str,
+    resource: Mapping[str, Any],
```
geoberle, line 608: is `resource` required here?

reconcile/openshift_resources_base.py
```diff
@@ -622,14 +670,22 @@ def fetch_desired_state(
         _log_lock.release()
         return
 
+    # _post_process_resources
+    # Some resources might need existent data in the cluster to populate all its fields.
```
geoberle, line 674: can you elaborate on this comment?

lechuk47: At first, I wanted to sync the certificate data into the Route with the `_post_process_route` method. Finally, I removed the logic and let `cert-utils-operator` sync the data. I left the comment there because this could be extended in the future by other resource types. 

The main reason not to do the sync is that I want to cover the `saas` deployments too, and with those is not possible to get the secret data because they are not reconciled regularly like `openshift_resources`.

reconcile/openshift_resources_base.py
geoberle: wdyt about keeping them in the method signature but using _ as var name to indicate that they are currently not used?

geoberle: consider this a nit and ignore if you like

geoberle (approved):
just nits. LGTM

PR #2487, opened 2022-06-22
q-cli ocm-aws-infrastructure-access-switch-role-links: login link instead of user_arn

patjlm:
This is only to make the output page more intuitive.

I wonder if we really need to display the account id even.

An example output is available here: https://gitlab.cee.redhat.com/-/snippets/5238


fishi0x01 (approved):
Nice! :heart:

patjlm:
https://issues.redhat.com/browse/APPSRE-5907

PR #2488, opened 2022-06-22
WIP: Create command to test Alertmanager receiver config

rporres:
https://issues.redhat.com/browse/APPSRE-5808


reconcile/test/test_amtool.py
```diff
@@ -0,0 +1,27 @@
+from reconcile.utils import amtool
+from .fixtures import Fixtures
```
fishi0x01, line 2: why not absolute import path?

tools/qontract_cli.py
```diff
@@ -1199,6 +1202,148 @@ def template(ctx, cluster, namespace, kind, name):
...
+
+    if not labels:
+        print(f"Cannot find alert {alert_name} in rules {rules_path}")
+        sys.exit(1)
+
+    for lbl in labels:
+        result = amtool.config_routes_test(am_config, lbl)
+        if not result:
+            print(f"Error running amtool: {result}")
+        print(result)
```
fishi0x01, line 1344: should this only be printed on success? Currently, if there is an error, we print result 2 times in a row.

rporres: Fixed in 5def72c9

reconcile/test/test_amtool.py
rporres: A classic case on copy&pasta. Fixed in 3fc6aa98

fishi0x01 (approved):

2uasimojo:
I'll refrain from reviewing the actual code (unless you want me to), but I would like to play around with this. I pulled it down, built and installed, but there still seems to be some setup necessary: I can't run without a config.toml; and when I try to specify one, the CLI still seems to be trying to contact the server.

Remind me: were we going to get to a place where no server was necessary? I remember you telling me about `make server` from app-interface, but when we tried it we discovered it required some kind of creds I don't have.

rporres:
Talking to @2uasimojo privately :)

PR #2489, opened 2022-06-23
[terrascript] use external resource spec in populate methods #2

maorfr:

follows up on #2481, where we removed the resource and namespace_info arguments.
in this PR, we remove the existing_secrets argument, and instead we rely on the external resource spec secret (which is now initialized earlier).


geoberle (approved):
❤️

PR #2492, opened 2022-06-23
unlink terraform integrations from aws-gc

maorfr:

reverts #94

back at the time, we were running integrations in pr-check with dependencies. this means that aws-garbage-collector would wait for terraform-resources and terraform-users to finish to give a true prediction of the expected changes.

since that time, we are no longer running aws-garbage-collector in pr-checks, so this change can be reverted (unused) and the code can be simplified.

### DIFF
--- a/reconcile/aws_garbage_collector.py
+++ b/reconcile/aws_garbage_collector.py
@@ -5,11 +5,9 @@ from reconcile.utils.aws_api import AWSApi
 QONTRACT_INTEGRATION = "aws-garbage-collector"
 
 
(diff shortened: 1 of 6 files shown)


steveteahan:
LGTM - I guess my only follow-up question would be whether it makes sense to continue to support the `aws-garbage-collector` feature at all? The combination of not running on MR checks and running in a non-obvious way makes this feature feel like it could potentially do more harm than good?

maorfr:
This integration only runs for a single account in the commercial environment, and not meant to run against production environments.

If and when we stop supporting that use case, this integration can be deprecated.

PR #2493, opened 2022-06-23
[terraform-resources] introduce TerraformConfigProvider

maorfr:

in our schemas, we represent external resources as an abstract entity. behind the scenes, the implementation of how to handle each external resource may defer.

this PR is beginning an alignment with this approach.

the reason we use terrascript is because it providers a way to get terraform configuration. so it is essentially a terraform configuration provider. in this PR, we replace terrascript in terraform-resources with a new class, TerraformConfigProvider.

this class exposes the required functionality, and under the hood, it utilizes terrascript itself.

the idea that is started in this PR is to slowly abstract away our usage of terrascript, until it is a detail behind ExternalResouceSpec. at that point, we will be able to use any tool we want to generate terraform configuration and replacing terrascript will be something we can do more easily.

### DIFF
--- a/reconcile/utils/terraform/config_provider.py
+++ b/reconcile/utils/terraform/config_provider.py
@@ -46,11 +46,23 @@ class TerraformConfigProvider:
     def init_spec_inventory(
         self, namespaces: Iterable[Mapping[str, Any]], provisioner_name: Optional[str]
     ) -> None:
-        self.ts.init_populate_specs(namespaces, provisioner_name)
+        """
+        Initiates resource specs from the definitions in app-interface
+        (schemas/openshift/external-resource-1.yml).
+        :param namespaces: schemas/openshift/namespace-1.yml object
+        :param account_name: AWS account name
+        """
 
-    @property
-    def resource_spec_inventory(self):
-        return self.ts.resource_spec_inventory
+        self.resource_spec_inventory: ExternalResourceSpecInventory = {}
+
+        for namespace_info in namespaces:
+            specs = get_external_resource_specs(namespace_info)
+            for spec in specs:
+                if provisioner_name and spec.provisioner_name != provisioner_name:
+                    continue
+                self.resource_spec_inventory[spec.id_object()] = spec
+
+        self.ts.init_populate_specs(namespaces, provisioner_name)
 
     def populate_resources(self, ocm_map: Optional[OCMMap] = None) -> None:
         self.ts.populate_resources(ocm_map=ocm_map)
--- a/reconcile/utils/terraform/config_provider.py
+++ b/reconcile/utils/terraform/config_provider.py
@@ -7,6 +7,10 @@ from reconcile.utils.ocm import OCMMap
 from reconcile.utils.terrascript_aws_client import TerrascriptClient as Terrascript
 
 
+class UnknownProvisionProviderError(Exception):
+    pass
+
+
 @dataclass
 class TerraformConfigProvider:
     """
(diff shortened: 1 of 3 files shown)


reconcile/utils/terraform/config_provider.py
```diff
@@ -0,0 +1,80 @@
...
+
+from reconcile.utils.terrascript_aws_client import TerrascriptClient as Terrascript
+
+
+class UnknownProvisionProviderError(Exception):
+    pass
+
+
+@dataclass
+class TerraformConfigProvider:
```
steveteahan, line 15: Is the idea that `Terrascript` is really `TerrascriptAwsClient` and when there is a `TerraformCloudflareClient` that would also be handled in this class? If so, would those all be initialized by this class?

I'm definitely interested in how we get the right abstraction here that will allow us to write new clients that are easy to test and minimize shared state. I'll start to think about what that could look like with other providers/clients and try to work backwards to this.

maorfr: Yes, i think that is the idea. I'm working towards having terraform-resources being provider agnostic, and TerraformConfigProvider being provider aware.

I will rename TerrascriptClient to TerrascriptAwsClient. Already started that move in https://github.com/app-sre/qontract-reconcile/pull/2440.

Let me know what you think!

reconcile/utils/terraform/config_provider.py
```diff
@@ -0,0 +1,80 @@
...
+        self, namespaces: Iterable[Mapping[str, Any]], provisioner_name: Optional[str]
+    ) -> None:
+        """
+        Initiates resource specs from the definitions in app-interface
+        (schemas/openshift/external-resource-1.yml).
+        :param namespaces: schemas/openshift/namespace-1.yml object
+        :param account_name: AWS account name
+        """
+
+        self.resource_spec_inventory: ExternalResourceSpecInventory = {}
```
steveteahan, line 60: Could this be initialized in `__init__`? Good for documentation, avoiding access to variables before they're assigned, etc.

reconcile/utils/terraform/config_provider.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def dump(
+        self,
+        print_to_file: Optional[str] = None,
+        existing_dirs: Optional[dict[str, str]] = None,
+    ) -> dict[str, str]:
+        return self.ts.dump(print_to_file, existing_dirs)
+
+    def init_spec_inventory(
+        self, namespaces: Iterable[Mapping[str, Any]], provisioner_name: Optional[str]
```
steveteahan, line 51: Does this method really need to care about `namespaces`? Or can it just care about specs? It seems like a one-liner for anything calling this to get the specs from the namespaces.

`specs = [get_external_resource_specs(namespace_info) for namespace_info in namespaces]`

It seems weird that namespaces are finding their way into the Terraform code. I understand why from a schema perspective, but I'm wonder if the simple code to extract the specs needs to happen here. The only other thing I see is that `init_populate_specs` also requires namespaces, but it looks like it's just running the same code to filter the specs?

reconcile/utils/terraform/config_provider.py
```diff
@@ -0,0 +1,80 @@
...
+    def __post_init__(self):
+        self.ts = Terrascript(
+            self.integration,
+            self.integration_prefix,
+            self.thread_pool_size,
+            self.provisioners,
+            settings=self.settings,
+        )
+
+    def dump(
```
steveteahan, line 43: Probably makes sense to have a good docstring here? `print_to_file` and `existing_dirs` don't have obvious meanings in my opinion.

steveteahan (commented):
Providing some additional feedback. This might be outside of the scope of what you were planning for in the PR, and it might be creeping into parts that I should be handling with the ticket that I have. With that being said, in my opinion the time to do this is when we're adding the new class. With that in mind, I'm happy to collaborate on this and not just expect you to handle it all. I'm also open to the idea that any of these thoughts are with far less context about this code than you have.

PR #2494, opened 2022-06-23
CronJob SQL Query Deletion - APPSRE-3787

cubismod:
Depends on https://github.com/app-sre/qontract-schemas/pull/175

Part of [APPSRE-3787](https://issues.redhat.com/browse/APPSRE-3787), this MR adds new behavior to the SQL query feature in App Interface to delete CronJobs after they have been created in OpenShift by adding a `delete: true` flag to the SQL query schema. Meaning that tenants will now be able to delete those queries without manual AppSRE intervention required. Tested this on dashdotdb staging, details in linked Jira issue.

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -1590,6 +1590,7 @@ APP_INTERFACE_SQL_QUERIES_QUERY = """
     }
     output
     schedule
+    delete
     query
     queries
   }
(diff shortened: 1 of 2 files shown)


bkez322:
Thanks for adding the details of how you tested this in the Jira issue.

This LGTM. Great job!

PR #2495, opened 2022-06-24
APPSRE-5046 generated types for service_dependencies

fishi0x01:
We want to introduce `saas_file_v2` in `service_dependencies` checks.

1. we add our new code generator to reconcile.
2. we include an `introspection.json` of our schema. That json is required in order to infer types. The json is generated by the code generator via `make gql-introspection`. Note, that at a later step we might decide to package the introspection schema in a different place/repo. This is the first iteration.
3. we place our gql query into a dedicated `.gql` file.
4. the code generator generates corresponding data classes via `make gql-query-classes`.
5. we migrate existing logic to use the new generated types

From now on, any schema change will also require:

```sh
make gql-introsprection
make gql-query-classes
```

If the schema is not compatible in any way, we will see errors during type checking phase.
A check that the `introspection.json` and generated classes are up-to-date with the latest schema should be part of CI in later step.

~This will be merged after we merged our [design doc](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/41063) for the code generator.~ Design doc is merged.

This PR gives a very practical insight on how type generation will look on an existing integration.

A follow-up PR will then introduce `saas_file_v2` to the query.


pyproject.toml
```diff
@@ -2,3 +2,8 @@
 line-length = 88
 target-version = ['py39']
 include = '^.*\.py$'
+force-exclude = '''
```
chassing, line 5: Why that? Just run "black ..." on all generated gql files after generation?

fishi0x01: currently the generated queries do not obey to line length settings. I.e., black will change the generated classes.

The longterm CI goal:

Run `make gql-introspection && make gql-query-classes` -> if `git diff` detects a change, then CI complains, because a schema change is not taken into account. Running black on generated classes complicates this more. Taking black out of the equation simplifies this a little: at least that was my initial thought.

However, I see your point in that we could also simply include black in the `make gql-query-classes` target and see it as a mandatory step of query class generation. That would be fine for CI too then :+1: Will adjust this.

fishi0x01: Just realizing that there is an issue with using black. Take this example:

```python
from pydantic import BaseModel, Extra, Field, Json # noqa: F401 # pylint: disable=W0611
```

becomes after black formatting

```python
from pydantic import (
    BaseModel,
    Extra,
    Field,
    Json,
)  # noqa: F401  # pylint: disable=W0611
```

The line breaks mess with the `noqa` annotation. Those annotations are only considered for the line they are placed on, i.e., black actually breaks the linting here. We could consider to adjust the generator to add those annotations on the file header, which would make the whole file disregard those linters. This would make linter checks less fine-grained though. Further, I feel that for now it would be easier to just avoid black on generated files.

fishi0x01: Actually I will adjust the generator to output already formatted headers. Should be fine then :+1:

fishi0x01: https://github.com/app-sre/qenerate/pull/20

fishi0x01: Ok generated classes are now also being modified by black. This works for now. If we stumble across another edge case, we can easily remove black formatting again from generated query classes.

chassing: perfect, thank you

fishi0x01:
@geoberle brought up an interesting idea on slack: generating a convenience function to load the query string.

```python
def load_query_string() -> str:
  with open('path_to_file.gql', r) as f:
    return f.read()
```

That way we dont have to mess around with the file paths. I will add this to the generator now.

fishi0x01:
Convenience function to load query strings added here: https://github.com/app-sre/qenerate/pull/21

chassing:
LGTM

chassing (approved):
LGTM

reconcile/service_dependencies.py
```diff
@@ -1,54 +1,20 @@
...
+from reconcile.gql_queries.service_dependencies import service_dependencies
+from reconcile.gql_queries.service_dependencies.service_dependencies import (
+    AppV1,
+    ServiceDependenciesQuery,
+)
 
 from reconcile.utils import gql
 from reconcile import queries
-from reconcile.utils.external_resources import (
-    get_provision_providers,
```
patjlm, line 7: `get_provision_providers` seems to not be used anywhere anymore (except in a test file), can we remove it and clean the code further ?

PR #2501, opened 2022-06-27
upgrade kubernetes python library from v12 to v24

geoberle:
v24 is the first release which contains a fix for payloads failing when they contain keys named `self`.

since the openshift python library relies on the kubernetes library but is outdated and does not work anymore on the recent version, this PR also removes the openshift library altogether. we don't rely on anything to specific within the openshift library except some optimizations on the resource discovery. therefore the essence of that was introduced as `OpenshiftLazyDiscoverer` into oc.py.

i tested this update with local dry-runs and also wet-runs for openshift-resources. since create/update/delete operations are not implemented via the OCNative class but OCDeprecated, extensive wet runs are not necessary.

Ref: https://issues.redhat.com/browse/APPSRE-4867


maorfr (approved):
:heart:

PR #2502, opened 2022-06-27
Clean up Terraform working dirs on print to file

steveteahan:
Temporary directories are not cleaned up when neither tf or working_dirs
kwargs are passed into cleanup_and_exit. This appears to affect print to
file functionality in development, but worth fixing. It's simple enough
to add a breakpoint if team members want to preserve the temporary dirs.

A prefix was added to the temporary directories as well so that it's easier
to identify which client is producing them. This will become increasingly
helpful when there is more than just the Terrascript AWS client.

Discovered as part of: https://issues.redhat.com/browse/APPSRE-5922

## Testing

* Run `terraform-resources` with `--print-to-file` and confirm that a temporary directory was created (`terrascript-aws-_doj_004`) and that it was deleted afterwards
* Run `terraform-resources` in `--dry-run` mode to verify that the change in temporary directory name didn't impact anything


reconcile/utils/terrascript_aws_client.py
```diff
@@ -3106,7 +3108,7 @@ def dump(
                     f.write(str(ts))
                     f.write("\n")
             if existing_dirs is None:
-                wd = tempfile.mkdtemp()
+                wd = tempfile.mkdtemp(prefix=TMP_DIR_PREFIX)
```
bkez322, line 3111: I definitely like this!

reconcile/terraform_resources.py
```diff
@@ -617,7 +617,7 @@ def run(
         defer(oc_map.cleanup)
 
     if print_to_file:
-        cleanup_and_exit()
+        cleanup_and_exit(tf)
```
bkez322, line 620: I also like this. 

It took me a bit of looking to convince myself that leaving the terrascript tmp files around when using --print-to-file definitely wasn't a deliberate choice for debugging. But I'm pretty certain such is not the case.

bkez322 (approved):
LGTM

maorfr (approved):
:heart:

PR #2503, opened 2022-06-28
[cli] add get sshuttle-command command

maorfr:
depends on https://github.com/app-sre/qontract-schemas/pull/180

example usage:
```
$ qontract-cli get sshuttle-command bastion.example.com | bash
```


patjlm (approved):
❤️

PR #2506, opened 2022-06-28
make tox happy about py310

maorfr:
when adding `py310` to tox.ini, these are the only errors it returns.
this is related to https://github.com/python/cpython/pull/23737

does this mean we are python 3.10 compatible?


fishi0x01:
From the linked issue:

```
patch, patch.object and create_autospec silently ignore misspelled
arguments such as autospect, auto_spec and set_spec. This can lead
to tests failing to check what they are supposed to check.
```

Oh wow. Good to have that fixed now :+1:

fishi0x01 (approved):

PR #2507, opened 2022-06-28
change make target dev-venv to be python3

maorfr:
another change we will not need to make when migrating to python 3.10.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ clean:
 	@find . -name "*.pyc" -delete
 
 dev-venv: clean ## Create a local venv for your IDE and remote debugging
-	python3.9 -m venv venv
+	python3 -m venv venv
 	. ./venv/bin/activate && pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -78,7 +78,7 @@ clean:
 	@find . -name "*.pyc" -delete
 
 dev-venv: clean ## Create a local venv for your IDE and remote debugging
-	python3.9 -m venv venv
+	python3 -m venv venv
```
fishi0x01, line 81: This is ambiguous and leads to different outcomes on different machines. If we want to upgrade to 3.10, then lets be explicit:

```sh
python3.10 -m venv venv
```

maorfr: isn't it common that python3 is an alias to whatever the user chooses?

fishi0x01: Absolutely. Which is what makes it ambiguous.

If we say we use python3.10, then we should enforce our dev-env to be 3.10.

PR #2510, opened 2022-06-28
[terraform-vpc-peerings] account filter capability

geoberle:
implement account filtering capabilities for terraform-vpc-peerings to allow account based sharding by integration-manager

the relevant account for filtering is the acceptors account

* for cluster to cluster VPC peering, the infrastructure management account on the acceptor cluster is used for filtering
* for cluster to account (normal peerings and mesh peering), the account plays the role of the acceptor and is used for filtering

Ref: https://issues.redhat.com/browse/APPSRE-5917

### DIFF
--- a/reconcile/terraform_vpc_peerings.py
+++ b/reconcile/terraform_vpc_peerings.py
@@ -132,7 +132,9 @@ def aws_assume_roles_for_cluster_vpc_peering(
     return req_aws, acc_aws
 
 
-def build_desired_state_single_cluster(cluster_info, ocm: OCM, awsapi: AWSApi):
+def build_desired_state_single_cluster(
+    cluster_info, ocm: OCM, awsapi: AWSApi, account_filter: Optional[str]
+):
     cluster_name = cluster_info["name"]
 
     peerings = []
@@ -165,6 +167,9 @@ def build_desired_state_single_cluster(cluster_info, ocm: OCM, awsapi: AWSApi):
         req_aws, acc_aws = aws_assume_roles_for_cluster_vpc_peering(
             cluster_info, peer_info, peer_cluster, ocm
         )
+        # filter on account
+        if account_filter and acc_aws["name"] != account_filter:
+            continue
 
         requester_vpc_id, requester_route_table_ids, _ = awsapi.get_cluster_vpc_details(
             req_aws, route_tables=requester_manage_routes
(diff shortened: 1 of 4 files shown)


maorfr:
you really hate `account_name`, don't you :sweat_smile:

reconcile/terraform_vpc_peerings.py
```diff
@@ -165,6 +167,9 @@ def build_desired_state_single_cluster(cluster_info, ocm: OCM, awsapi: AWSApi):
         req_aws, acc_aws = aws_assume_roles_for_cluster_vpc_peering(
             cluster_info, peer_info, peer_cluster, ocm
         )
+        # filter on account
+        if account_filter and acc_aws["name"] != account_filter:
```
patjlm, line 171: shouldn't we filter on `req_aws` here ? this is the current `cluster_info`, right ?

geoberle: good point.

our aws account sharding strategy circles around into what terraform state we are going to put things.

i had a closer look at how `populate_vpc_peerings` processes the peerings
* it puts the `aws_vpc_peering_connection` into the requesters account terraform state
* it puts the `aws_vpc_peering_connection_accepter` into the accepters account terraform state

this does not work nicely with sharding since a single vpc-peering desired state item might belong to two accounts. to make this work, we need to decide into what state we are going to put both sides of the peering and this side will be the one to be used for account filtering/sharding.

luckily requester and accepter accounts are always the same in app-interface commercial so we are in position where we can choose freely. someone would need to check on the other app-interface instance though.

the target account acts as accepter for the `account-vpc` and `account-vpc-mesh` peering types and sets request account = accepter account, so the accepter is the deciding factor for those types. using the accepters side of a `cluster-vpc` peering would put us in a position where we consistently use the accepters terraform state.

this would also imply that we do account-filtering/sharding on the requesters account.

wdyt?

geoberle: e.g. [fc37ded](https://github.com/app-sre/qontract-reconcile/pull/2510/commits/fc37ded3f1b030ce397241875c0c692eba68731b)

geoberle: i just realized that the way this is built right now, the requester and accepter resources must be in the same terraform state because the accepter references the requester by id.

https://github.com/app-sre/qontract-reconcile/blob/69fba0a1b24dd263580b2ec0982ab0e105090ff6/reconcile/utils/terrascript_aws_client.py#L795-L796

PR #2517, opened 2022-06-30
Force terrascript to calculate lambda_iam_role_resource.arn

TGPSKI:
This MR corrects a bug with generating the lambda resource for rosa-authenticator.


bkez322:
🤞

PR #2518, opened 2022-06-30
Use correct sms role for cognito user pool [rosa-authenticator]

TGPSKI:
This MR fixes a bug with the sms role arn - fixing the reference to point to the correct iam resource.


bkez322:
Looks Grand To Me

PR #2519, opened 2022-07-01
Force Terraform to respect DAG [rosa-authenticator]

TGPSKI:
This MR forces terraform to source dependent values from created resources instead of object parameters, creating an expected dependency tree.


bkez322 (approved):
LGTM 🤞

PR #2522, opened 2022-07-01
APPSRE 5774 sql query large queries

chassing:
### sql_query: large query support

Instead of retrieving SQL queries as command line parameters, get them via a config map file. So it's possible to run queries larger than 4096 characters long.
Additionally, add type annotations and some small refactorings.

see APPSRE-5774

### DIFF
--- a/reconcile/sql_query.py
+++ b/reconcile/sql_query.py
@@ -3,6 +3,7 @@ import sys
 import time
 
 from textwrap import indent
+from typing import Any, Iterable, Mapping, Optional, Union
 
 import jinja2
 from ruamel import yaml
@@ -28,6 +29,11 @@ LOG = logging.getLogger(__name__)
 
 JOB_TTL = 604800  # 7 days
 POD_TTL = 3600  # 1 hour (used only when output is "filesystem")
+QUERY_CONFIG_MAP_CHUNK_SIZE = 512 * 1024  # 512 KB
+
+CONFIG_MAPS_MOUNT_PATH = "/configs"
+GPG_KEY_NAME = "gpg-key"
+GPG_KEY_PATH = f"{CONFIG_MAPS_MOUNT_PATH}/{GPG_KEY_NAME}"
 
 JOB_SPEC = """
 spec:
@@ -39,6 +45,7 @@ spec:
       imagePullSecrets:
       - name: {{ PULL_SECRET }}
       {% endif %}
+      restartPolicy: Never
       containers:
       - name: {{ JOB_NAME }}
         image: {{ IMAGE_REPOSITORY }}/{{ ENGINE }}:{{ENGINE_VERSION}}
@@ -62,16 +69,18 @@ spec:
             value: {{ value }}
           {% endif %}
           {% endfor %}
-      {% if GPG_KEY is defined %}
         volumeMounts:
-        - name: gpg-key
-          mountPath: /gpg
+        - name: configs
+          mountPath: {{ CONFIG_MAPS_MOUNT_PATH }}
+          readOnly: true
       volumes:
-      - name: gpg-key
-        configMap:
-          name: {{ GPG_KEY }}
-      {% endif %}
-      restartPolicy: Never
+        - name: configs
+          projected:
+            sources:
+            {% for cm in CONFIG_MAPS %}
+            - configMap:
+                name: {{ cm }}
+          {% endfor %}
 """
 
 
@@ -109,21 +118,9 @@ spec:
 )
 
 
-CONFIGMAP_TEMPLATE = """
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: {{ GPG_KEY }}
-  labels:
-    app: qontract-reconcile
-    integration: {{ QONTRACT_INTEGRATION }}
-    query-name: {{ QUERY_NAME }}
-data:
-  gpg-key: {{ PUBLIC_GPG_KEY }}
-"""
-
-
-def get_tf_resource_info(terrascript: Terrascript, namespace, identifier):
+def get_tf_resource_info(
+    terrascript: Terrascript, namespace: Mapping[str, Any], identifier: str
+) -> Union[dict[str, str], None]:
     """
     Extracting the external resources information from the namespace
     for a given identifier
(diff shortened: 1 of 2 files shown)


rporres:
Hmmm, the maximum size of a ConfigMap is 1Mb. While I don't expect that kind of queries very often, the current solution is also limited and would have also made fail the query that inspired this work (larger that 1Mb)... 

Maybe chunking the query, creating multiple config maps and then join them before the query via `cat`?

chassing:
> ... the maximum size of a ConfigMap is 1Mb ...

I wasn't aware of this ConfigMap limit, good point. Let me see how can I chunk the query.

rporres:
I learnt about it no so long ago... https://kubernetes.io/docs/concepts/configuration/configmap/#motivation

cubismod:
TIL as well. Tracking multiple ConfigMaps may be a bit difficult as well since the state information for this integration is pretty limited (worked on it recently myself). Changes so far are looking good minus the ConfigMap issue

chassing:
In the end, we're talking about 0.X% of sql-queries, so chunking into several ConfigMaps is easy and pretty straightforward, and it doesn't change the integration's current behavior. It will be implemented like that. Imaging a sql-query yml with three queries and the 1st one is huge, then you'll have these ConfigMaps:

```
$ oc get cm
q0000c0000 -> 1st query (q) 1st chunk (c)
q0000c0001 -> 1st query, 2nd chunk
q0001c0000 -> 2nd query, just one chunk
q0003c0000 -> 3rd query, also just one chunk
```

In 99.X% case we will have just:

```
$ oc get cm
q0000c0000 -> one query without any chunking
```

reconcile/sql_query.py
```diff
@@ -385,112 +377,151 @@ def process_template(query, image_repository, use_pull_secret=False):
...
+        cleanup = False
 
         # Checking the sql-query state:
         # - No state: up for execution.
         # - State is a timestamp: executed and up for removal
         #   after the JOB_TTL
         # - State is 'DONE': executed and removed.
         # - State is not 'DONE' but 'delete:true' and is a cronjob: up for removal
-        try:
+        if state.exists(query_name):
```
rporres, line 504: Do we really need all this state handling refactor for the configmap split? Since I'm not very familiar with the original integration I'm having a hard time following all the changes here 😓 

I understand that the current state logic is a bit convoluted and I would agree in attempting a refactor outside this PR 😄

reconcile/test/test_sql_query.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+
+from reconcile.sql_query import split_long_query
+
+
+@pytest.mark.parametrize(
+    "q, size, excepted",
```
rporres, line 7:
```suggestion
    "q, size, expected",
```

rporres: and below :)

reconcile/sql_query.py
```diff
@@ -28,6 +29,11 @@
 
 JOB_TTL = 604800  # 7 days
 POD_TTL = 3600  # 1 hour (used only when output is "filesystem")
+QUERY_CONFIG_MAP_CHUNK_SIZE = 512 * 1024  # 512 KB
+
+CONFIG_MAPS_MOUNT_PATH = "/configs"
+GPG_KEY = "gpg-key"
```
rporres, line 35: Maybe clearer if
```suggestion
GPG_KEY_NAME = "gpg-key"
```

reconcile/sql_query.py
```diff
@@ -385,112 +377,151 @@ def process_template(query, image_repository, use_pull_secret=False):
     if output not in supported_outputs:
         raise RuntimeError(f"Output {output} not supported")
 
-    make_command = engine_cmd_map[engine]
-
-    command = make_command(
-        output=output, sqlqueries=query["queries"], recipient=query.get("recipient")
+    # concatinate all query files into a single one
```
rporres, line 380:
```suggestion
    # concatenate all query files into a single one
```

reconcile/sql_query.py
```diff
@@ -366,7 +353,12 @@ def encrypted_closing_message(recipient):
     ]
 
 
-def process_template(query, image_repository, use_pull_secret=False):
+def process_template(
+    query: dict[str, Any],
+    image_repository: str,
+    use_pull_secret: bool,
+    config_map_names: list[str],
+) -> str:
```
rporres, line 361: After the change, the below doc is no longer accurate.

reconcile/sql_query.py
```diff
@@ -385,112 +377,151 @@ def process_template(query, image_repository, use_pull_secret=False):
...
+
+def split_long_query(q, size) -> list[str]:
+    return [q[i : i + size] for i in range(0, len(q), size)]
 
 
-def openshift_delete(dry_run, oc_map, query, resource_type, enable_deletion):
-    try:
-        openshift_base.delete(
+def merge_files_command(directory, file_glob, output_file):
+    return f"cat ''{directory}''/{file_glob} > ''{output_file}'';"
```
rporres, line 430: nit: It is a bit confusing that this command ends with `;` whereas the db commands do not and rely in other functions to do it... I would make the concatenation of commands via the `;` in the `process_template`

reconcile/sql_query.py
```diff
@@ -62,16 +69,18 @@
...
+          mountPath: {{ CONFIG_MAPS_MOUNT_PATH }}
+          readOnly: true
       volumes:
-      - name: gpg-key
-        configMap:
-          name: {{ GPG_KEY }}
-      {% endif %}
-      restartPolicy: Never
+        - name: configs
+          projected:
```
rporres, line 78: TIL ❤️

reconcile/sql_query.py
chassing: 👍

reconcile/sql_query.py
chassing: 👍

reconcile/sql_query.py
chassing: good point 👍

reconcile/sql_query.py
```diff
@@ -228,14 +232,12 @@ def collect_queries(query_name=None, settings=None):
...
-        sql_queries = []
+        _queries = []
         if sql_query["query"] is not None:
-            sql_queries.append(sql_query["query"])
+            _queries.append(sql_query["query"])
 
         if sql_query["queries"] is not None:
-            sql_queries.extend(sql_query["queries"])
-
-        sql_queries = [item.replace("'", "''") for item in sql_queries]
```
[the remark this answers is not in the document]
rporres: indeed true, thx

chassing:
Reference [Migration strategy](https://issues.redhat.com/browse/APPSRE-5774?focusedCommentId=20561862&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20561862)

chassing:
@rporres I want to merge this PR tommorow, please approve it :)

rporres (approved):
LVGTM ❤️

PR #2525, opened 2022-07-01
Initial support for generating Cloudflare resource configs via Terraform clients

steveteahan:
Design document: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/terraform-resources-cloudflare-support.md

Related ticket: https://issues.redhat.com/browse/APPSRE-6001


reconcile/utils/terrascript_cloudflare_client.py
```diff
@@ -0,0 +1,213 @@
...
+@dataclass
+class CloudflareAccountConfig:
+    name: str
+    email: str
+    api_key: str
+    account_id: str
+
+
+@dataclass
+class S3BackendConfig:
```
maorfr, line 28: Behind the scenes, is this actually cloudflare R2?

steveteahan: I was thinking we'd probably default to S3 for all cases. It's not clear to me from the list of [available backends](https://www.terraform.io/language/settings/backends/remote) that R2 is available yet, and it's considered beta to Cloudflare.

maorfr: R2 should be S3 compatible IIUC

steveteahan: It looks like the S3 client does allow you to specify a [custom endpoint](https://github.com/hashicorp/terraform/blob/73c4a4c3b04a20831bf0e4d15b11cd5442e643e7/internal/backend/remote-state/s3/backend.go#L397). So, it may be possible. We'd need to extend this data class to also expose the `endpoint` field if we decided to go that path. I think the question is whether we want to bother if R2 is in beta and there's not a compelling reason to use it. It's also unclear to me that it supports versioning of objects from a quick search.

reconcile/utils/terrascript/cloudflare_client.py
```diff
@@ -0,0 +1,211 @@
...
+    account_config: CloudflareAccountConfig, backend_config: S3BackendConfig
+) -> Terrascript:
+    terrascript = Terrascript()
+
+    terrascript += Terraform(
+        required_providers={
+            # TODO: pull this from account file
+            "cloudflare": {"source": "cloudflare/cloudflare", "version": "3.18"}
+        }
+    )
```
jfchevrette, line 46: Would it make sense to leave it to the user to pass in the desired version?

This is what I am playing with locally where the integration code pulls the provider version from the account file and passes it in to the updated method code below:
```suggestion
def create_terrascript_cloudflare(
    account_config: CloudflareAccountConfig,
    backend_config: S3BackendConfig,
    provider_version: str,
) -> Terrascript:
    terrascript = Terrascript()

    terrascript += Terraform(
        required_providers={
            # TODO: pull this from account file
            "cloudflare": {
                "source": "cloudflare/cloudflare",
                "version": provider_version,
            }
        }
    )
```

steveteahan: Yes, that's kind of what I meant to capture with ` # TODO: pull this from account file` -- we'd get this from the account query and inject it in here. There's no reason that I didn't do it already other than that this was just meant to be a placeholder to show that we can offload the creation of `Terrascript` to a separate function rather than including it in the client like we're doing in `TerrascriptClient`.

cubismod:
I really like this approach! Very readable and with inheritance we can remove specific state information from the overall Terrascript client class and just put it into subclasses instead which is easier to grasp and test than the current 5000 line file.

steveteahan:
After integration with the `TerraformClient` class, we can start to see the plan results:

```
INFO:root:['create', 'acct_a', 'cloudflare_zone', 'acct-a-domain-com']
INFO:root:['create', 'acct_a', 'cloudflare_zone_settings_override', 'acct-a-domain-com']
INFO:root:['create', 'acct_b', 'cloudflare_zone', 'acct-b-domain-com']
INFO:root:['create', 'acct_b', 'cloudflare_zone_settings_override', 'acct-b-domain-com']
```

I wanted to get this going sooner rather than later because whether we land on the same integration, or a separate integration, we'll still need make sure that `TerrascriptCloudflareClient` behaves as `TerraformClient` expects with working dirs, account info, etc.

I noted in a comment that some minor refactoring of `TerraformClient` will probably be necessary to pull out the AWS-specific parts, particularly the `AWSApi` object that's being injected in `__init__()`. For now, I can inject a dummy object because those paths won't actually be hit since we're not using AWS resources.

reconcile/utils/terrascript/cloudflare_client.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        Add the resource spec to Terrascript using the resource-specific classes
+        to determine which resources to create.
+        """
+        for spec in self._resource_specs.values():
+            resources_to_add = create_cloudflare_terrascript_resource(spec)
+            self._add_resources(resources_to_add)
+
+    def dump(
+        self, print_to_file: Optional[str] = None, existing_dir: Optional[str] = None
```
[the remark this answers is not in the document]
steveteahan: It was on my TODO list, but probably should have been a comment. It's now implemented in `TerraformConfigClientCollection` and removed everywhere else. This enables the `--print-to-file` option in the terraform integrations.

steveteahan:
I've removed the draft status designation because this is ready for full review. There is one `TODO` related to testing `ResourceValueResolver` that will be done before this is merged, so I'd like to avoid blocking on it.

bhushanthakur93:
LGTM

bhushanthakur93 (approved):

steveteahan:
> There is one TODO related to testing ResourceValueResolver that will be done before this is merged, so I'd like to avoid blocking on it.

This is now done.

reconcile/utils/terrascript_cloudflare_client.py
jfchevrette: I think it makes more sense to keep the state in S3, regardless of the provider. What if we eventually add a provider that does not have file storage?

reconcile/utils/external_resources.py
```diff
@@ -47,3 +52,122 @@ def managed_external_resources(namespace_info: Mapping[str, Any]) -> bool:
...
+            values[primary_key][secondary_key] = v
+            values.pop(k, None)
+
+    @staticmethod
+    def _override_values(values, overrides):
+        if overrides is None:
+            return
+        data = json.loads(overrides)
+        for k, v in data.items():
+            values[k] = v
```
jfchevrette, line 173: We probably want to add type hints to these methods, even though they are a direct copy from TerrascriptAwsClient

jfchevrette (approved):

reconcile/utils/external_resources.py
steveteahan: These were just accidentally missed. I've added the following to prevent missing types in new modules:

```
[mypy-reconcile.utils.external_resources.*]
disallow_untyped_defs = True
```

cubismod (approved):

PR #2539, opened 2022-07-03
add bots to be saas file owners

maorfr:
part of https://issues.redhat.com/browse/APPSRE-5940

this change is required to support external flows to app-interface, and allow teams to use their own bots to submit saas-file-update MRs, and to be able to `/lgtm` them.

tested in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/42134


geoberle (approved):
LGTM

PR #1404, opened 2021-02-21
[saas-file-owners] check that diff is in changed paths

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3029

replaces #1385

we are seeing issues of incorrect comments on MRs, where the saas-file-owners integration believes it is changing saas files which it does not. I've investigated a bit and my current working assumption is that this happens when a merge was made to master but the build-master job has not yet completed or has completed and the data was not yet reloaded in the server. This causes a MR to believe it is the one changing the other saas file.

PR #2541, opened 2022-07-03
[saas-file-owners] fix diff not found in changed paths bug

maorfr:
related to https://issues.redhat.com/browse/APPSRE-3029

fixes a bug that was introduced in #1404

this bug shows itself when there are multiple promotions within the same saas file to different environments.

since we use `changed_paths` and update it while we iterate over the `diffs`, if there are 2 changes to the same saas file to different environments - they are considered 2 differents `diffs`, but only one changed path. when we remove this path from `changed_paths`, and get to the 2nd iteration - the integration is mistaken to think it is not a part of the MR, but it is.

the outcome of this bug is not harmful at all - only logging a warning.

it is, however, very confusing.

to resolve it, we will keep the current behavior and usage of `changed_paths` exactly as is, but when checking if a diff is a part of the changed paths, we will compare it against a copy of `changed_paths`, which we will not modify.

### DIFF
--- a/reconcile/saas_file_owners.py
+++ b/reconcile/saas_file_owners.py
@@ -327,10 +327,13 @@ def run(
     )
     comment_lines = {}
     hold = False
+    changed_paths_copy = changed_paths.copy()
     for diff in diffs:
         # check if this diff was actually changed in the current MR
         saas_file_path = diff["saas_file_path"]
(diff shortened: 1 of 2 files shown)


geoberle (approved):
LGTM

PR #2544, opened 2022-07-04
print warning when currently upgrading to a blocked version

patjlm:
This avoids a hard failure of the integration due to an OCM 400:
```
{
  "kind": "Error",
  "id": "400",
  "href": "/api/clusters_mgmt/v1/errors/400",
  "code": "CLUSTERS-MGMT-400",
  "reason": "Upgrade policy 'xxx-xxx' is in 'started' state and cannot be updated",
  "operation_id": "xxx-xxx"
}
```


fishi0x01 (approved):
Not an expert in this domain, but LGTM :+1:

geoberle (approved):
LGTM

PR #2546, opened 2022-07-05
APPSRE-5948 add schema proxy

fishi0x01:
According to our [migration strategy](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/custom-python-code-generator-for-gql-queries.md#leverage-a-proxy-class) towards typed gql queries, we add a `SchemaProxy` class to transparently add types. 

The `SchemaProxy` behaves like a dict, i.e., existing logic is not affected by that change. However, an integration can use the data from a `SchemaProxy` object to achieve safe types.


fishi0x01:
@janboll and me had a chat about this and decided to discard the `SchemaProxy` approach for now. Main reason being the amount of change required on the code base.

PR #2547, opened 2022-07-05
[jenkins-job-builder] reduce required schemas for integration run

geoberle:
the jenkins-job-builder integration declares a lot more schemas than it actually queries, because it uses some generic app-interface-settings and aws-account queries that are too broad.

this change introduces a `get_secret_reader_settings()` query that fetches minimal information for secret reader initialization. also the `State` can be initialized with a new function `init_state` that requires minimal data about the involved account.

as a result, a lot of schemas can be removed from the jenkins-jobs-builders `integration-1.schemas` list in app-interface, which has positive side effects for PR checks. also the `integration-1.pr_check.run_for_valid_saas_file_changes` option can be removed for this integration, since the saas-file-2 schema is not required to be in the schema list anymore. the remaining schemas are

```yaml
    schemas:
    - /app-sre/integration-1.yml
    - /app-interface/app-interface-settings-1.yml # for secret reader
    - /dependencies/jenkins-config-1.yml # for jobs
    - /dependencies/jenkins-instance-1.yml # for jobs
    - /aws/account-1.yml # for state
    - /app-sre/app-1.yml # for code components
    - /access/user-1.yml # for crosscheck repo admins
    - /access/external-user-1.yml # for crosscheck repo admins
    - /access/bot-1.yml # for crosscheck repo admins
```

this commit touches several integration to provide the required `SecretReader` to `init_jjb`. all these integrations and their dependencies will also be reworked to require less schemas. but that will be work for followup PRs.

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -5,11 +5,30 @@ import itertools
 import shlex
 
 from textwrap import indent
-from typing import Any, Optional
+from typing import Any, Mapping, Optional
 
 from jinja2 import Template
 
 from reconcile.utils import gql
+
+
+SECRET_READER_SETTINGS = """
+{
+  settings: app_interface_settings_v1 {
+    vault
+  }
+}
+"""
+
+
+def get_secret_reader_settings() -> Optional[Mapping[str, Any]]:
+    """Returns SecretReader settings"""
+    gqlapi = gql.get_api()
+    settings = gqlapi.query(SECRET_READER_SETTINGS)["settings"]
+    if settings:
+        # assuming a single settings file for now
+        return settings[0]
+    return None
 
 
 APP_INTERFACE_SETTINGS_QUERY = """
--- a/reconcile/utils/state.py
+++ b/reconcile/utils/state.py
@@ -4,12 +4,41 @@ import json
 from typing import Any, Iterable, Mapping, Optional
 
 from botocore.errorfactory import ClientError
+from jinja2 import Template
+from reconcile.utils import gql
 
 from reconcile.utils.aws_api import AWSApi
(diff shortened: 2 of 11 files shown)


reconcile/queries.py
```diff
@@ -5,13 +5,32 @@
...
 from reconcile.utils import gql
 
 
+SECRET_READER_SETTINGS = """
+{
+  settings: app_interface_settings_v1 {
+    vault
+  }
+}
+"""
```
fishi0x01, line 21: If its not too much of a hassle, can we use our new query class generator for this? 

I can also have a look at it after this is merged.

reconcile/utils/state.py
```diff
@@ -4,14 +4,43 @@
...
+    resourcesDefaultRegion
+    automationToken {
+      path
+      field
+      version
+      format
+    }
+  }
+}
+"""
```
fishi0x01, line 32: This query would be a great new example for the query class generation. It contains an input variable and we dont have an example for that yet. 

If its ok for you, I will create a follow-up PR to convert this.

fishi0x01 (approved):
Nice. If you dont mind, I will create a follow-up PR after this is merged, in order to convert this towards the new query class generator workflow. This PR has some really nice examples :)

reconcile/utils/state.py
geoberle: i want to learn about this from the master personally so wdyt about peering on this?

geoberle: thanks for the session today. it turned out quite easy to implement but would require a lot of code to be touched (AWSApi, State, SecretReader are very widely used central pieces of code). i'm going to postpone that to another PR if that is ok with you.

PR #2549, opened 2022-07-05
template-renderer: show diff of rendered and expected outputs

jfchevrette:
Currently, the `template-tester` will output the full content of both the rendered and the expected resources, leaving it to the user to find differences

This changes it so that using `difflib` we only show the differences in a familiar unified diff output

Sample of current output:
```
expected:
some: var
another:
  thing: inside
  thang: again
the: end

rendered:
some: var
another:
  thing: inside
  new: thing
  thang: again
the: end
```

Output after this change
```
--- expected
+++ rendered
@@ -2,5 +2,6 @@
 some: var
 another:
   thing: inside
+  new: thing
   thang: again
 the: end
```

### DIFF
--- a/reconcile/template_tester.py
+++ b/reconcile/template_tester.py
@@ -1,3 +1,4 @@
+import difflib
 import logging
 import sys
 import yaml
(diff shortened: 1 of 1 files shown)


geoberle (approved):
nice. thank you ❤️

jfchevrette:
Fixed two small oversights that triggered black/flake errors

rporres (approved):
so nice 😍

PR #2556, opened 2022-07-06
Make alert name optional in alert-to-receiver

rporres:
The command will show all the alerts in the file along with its
receivers. It will make easier to check how a certain change in routing
tree affects all the alerts in a rules file.

Example of the new behaviour:

```
qontract-cli --config config.local.toml alert-to-receiver -s config \
  crcp01ue1 \
  openshift-customer-monitoring \
 /insights-prod/ros-prod/ros.prometheusrules.yml 
App-insights-ros-backend-api-service-In-ros-prod-Absent|slack-team-consoledot-alert
App-insights-ros-backend-processor-service-In-ros-prod-Absent|slack-team-consoledot-alert
ROSHttp5xx|slack-team-consoledot-alert
```

Output has also changed to always print the alert name. This will also clarify the case in which we have multiple alerts of the same name in a rules file.

Part of APPSRE-5808

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -1359,7 +1359,7 @@ def alert_to_receiver(
         )
         sys.exit(1)
 
-    alert_labels: list[dict] = [] 
+    alert_labels: list[dict] = []
     for group in rule_spec["groups"]:
         for rule in group["rules"]:
             alert_labels.append(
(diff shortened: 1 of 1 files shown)


tools/qontract_cli.py
```diff
@@ -1353,23 +1358,29 @@ def alert_to_receiver(
         )
         sys.exit(1)
 
-    labels = []  # array of dicts
+    alert_labels = []  # array of dicts
```
geoberle, line 1361:
```suggestion
    alert_labels: list[dict] = [] 
```

geoberle (approved):
LGTM

PR #2540, opened 2022-07-03
[saas-file-owners] consider MR description as comment

maorfr:
part of https://issues.redhat.com/browse/APPSRE-5940

allow creating a MR, including a `/lgtm` in the MR description, which will be considered as a `/lgtm` comment.

this will be mostly used for creating promotion MRs that will be automatically merged. the main use case is usage by bots (following up on #2539).

PR #2563, opened 2022-07-07
[saas-file-owners] handle null comment body

maorfr:
following #2540 where we started treating the MR description as a comment, this MR allows for the MR body to be null, to avoid crashing on such a scenario.


maorfr:
https://ci.int.devshift.net/job/service-app-interface-gl-pr-check/143465/

BumbleFeng (approved):

PR #2566, opened 2022-07-08
Add support for creating SNS topic and RDS event notification

OliviaHY:
#### What:
For terraform-resources integration,
* Add support for creating SNS topics
* Add support for creating  RDS event subscription to SNS topics

#### Why:

We want to provide tenants the ability to be able subscribe to email notifications about their RDS instances's activity. 
Design doc [here](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/rds-upgrades-and-maintenance.md#database-maintenance-notifications).

#### Validation:
* Running terraform-resources integration through dry-run mode and check the output consistent with the yml defined in open source App Interface.
* Running the integration without dry-run mode against ter-int-dev AWS account and was able to see SNS topics and RDS event subscriptions created( and deleted) as intended. This includes multiple subscriptions, multiple event notification, with optional fields and without them.
Core yml that was used:
```
externalResources:
  - provider: aws
    provisioner:
      $ref: /aws/app-int-example/account.yml
    resources:
    - provider: rds
      identifier: app-int-example-01
      defaults: /terraform/rds/free-tier-1.yml
      output_resource_name: db-creds
      overrides:
        publicly_accessible: true
      event_notifications:
        - destination: test-sns-1
          source_type: db-instance
          event_categories: 
            - creation 
            - deletion
    - provider: sns
      identifier: test-sns-1
      defaults: /terraform/sns/sns-1.yml
      subscriptions:
        - protocol: email
          endpoint: foo@redhat.com
```
#### Dependencies: 
Depending on https://github.com/app-sre/qontract-schemas/pull/188

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -46,6 +46,7 @@ from terrascript.resource import (
     aws_api_gateway_base_path_mapping,
     aws_db_instance,
     aws_db_parameter_group,
+    aws_db_event_subscription,
     aws_s3_bucket,
     aws_iam_user,
     aws_s3_bucket_notification,
@@ -63,6 +64,8 @@ from terrascript.resource import (
     aws_elasticache_parameter_group,
     aws_iam_user_policy_attachment,
     aws_sqs_queue,
+    aws_sns_topic,
+    aws_sns_topic_subscription,
     aws_dynamodb_table,
     aws_ecr_repository,
     aws_s3_bucket_policy,
@@ -207,6 +210,8 @@ VARIABLE_KEYS = [
     "subnet_ids",
     "network_interface_ids",
     "vpce_id",
+    "fifo_topic",
+    "subscriptions",
 ]
 
 TMP_DIR_PREFIX = "terrascript-aws-"
@@ -1123,6 +1128,8 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
             self.populate_tf_resource_role(spec)
         elif provider == "sqs":
             self.populate_tf_resource_sqs(spec)
+        elif provider == "sns":
+            self.populate_tf_resource_sns(spec)
         elif provider == "dynamodb":
             self.populate_tf_resource_dynamodb(spec)
         elif provider == "ecr":
@@ -1409,10 +1416,26 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
         # this will only affect the output Secret
         output_resource_db_name = values.pop("output_resource_db_name", None)
 
+        if "event_notifications" in values:
+            event_notifications = values.pop("event_notifications")
+            for e_n in event_notifications:
+                sns_topic_name = e_n.pop("destination")
+                if sns_topic_name.startswith("arn:"):
+                    raise (ValueError("destination should not be an arn"))
+                e_n["sns_topic"] = "${aws_sns_topic" + "." + sns_topic_name + ".arn}"
+                e_n["source_ids"] = ["${aws_db_instance" + "." + identifier + ".id}"]
+                source_type = e_n.get("source_type", "all")
+                e_n_identifier = (
+                    f"{sns_topic_name}_{source_type}_aws_db_event_subscription"
+                )
+                e_n_tf_resource = aws_db_event_subscription(e_n_identifier, **e_n)
+                tf_resources.append(e_n_tf_resource)
+
         # rds instance
         # Ref: https://www.terraform.io/docs/providers/aws/r/db_instance.html
         tf_resource = aws_db_instance(identifier, **values)
         tf_resources.append(tf_resource)
+
         # rds outputs
         # we want the outputs to be formed into an OpenShift Secret
         # with the following fields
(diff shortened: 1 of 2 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -1391,15 +1399,29 @@ def populate_tf_resource_rds(self, spec):
 
         if len(deps) > 0:
             values["depends_on"] = deps
-
+        
         # pop alternate db name from value before creating the db instance
         # this will only affect the output Secret
         output_resource_db_name = values.pop("output_resource_db_name", None)
 
+        if "event_notifications" in values.keys():
```
steveteahan, line 1407: You can drop `keys()` here because that's the default behavior.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1391,15 +1399,29 @@ def populate_tf_resource_rds(self, spec):
...
+            event_notifications = values.pop("event_notifications") 
+            counter = 0
+            for e_n in event_notifications:
+                sns_topic_name = e_n.pop("destination")
+                e_n["sns_topic"] = "${aws_sns_topic"+ "." + sns_topic_name + ".arn}"
+                e_n["source_ids"] = ["${aws_db_instance"+ "." + identifier + ".id}"]
+                print(e_n["source_ids"])
+                counter += 1
+                e_n_identifier = f"{identifier}_aws_db_event_subscription_{counter}"
```
steveteahan, line 1416: Does this mean that if we removed say the first subscription that every other subscription would be recreated?

reconcile/utils/terrascript_aws_client.py
```diff
@@ -2255,7 +2277,54 @@ def populate_tf_resource_sqs(self, spec):
...
+    def populate_tf_resource_sns(self, spec):
+        account = spec.provisioner_name
+        identifier = spec.identifier
+        common_values = self.init_values(spec)
+        output_prefix = spec.output_prefix
+        policy = common_values.get("inline_policy")
+        region = common_values.get("region") or self.default_regions.get(account)
+        
+        values = {}
+        if 'fifo_topic' in common_values.keys():
```
steveteahan, line 2290: What about replacing this with:

```python
common_values.get("fifo_topic", False)
```
That'd also set the value to False if tthat key is missing?

reconcile/utils/terrascript_aws_client.py
```diff
@@ -2255,7 +2277,54 @@ def populate_tf_resource_sqs(self, spec):
...
+            fifo_topic = common_values.get("fifo_topic")
+        else:
+            fifo_topic = False
+        if fifo_topic:
```
steveteahan, line 2294: What if the user already added a `.fifo` suffix? Should we error here? Or should we require the user to suffix the name with `.fifo` to make it explicit?

reconcile/utils/terrascript_aws_client.py
```diff
@@ -2255,7 +2277,54 @@ def populate_tf_resource_sqs(self, spec):
...
+        if "subscriptions" in common_values.keys():
+            subscriptions = common_values.get("subscriptions")
+            counter = 0
+            for sub in subscriptions:
+                counter += 1
+                sub_values = {}
+                sub_values["topic_arn"] = "${aws_sns_topic"+ "." + identifier + ".arn}"
+                sub_values["protocol"] = sub["protocol"]
+                sub_values["endpoint"] = sub["endpoint"]
+                sub_identifier = f"{identifier}_aws_sns_topic_subscription_{counter}"
```
steveteahan, line 2317: Much like the other case, would these be recreated if a subscription was removed?

steveteahan (commented):
Looks good at a high-level. I'll take another closer look after the first round of feedback.

Can you include any details about manual testing that was performed?

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1391,15 +1399,29 @@ def populate_tf_resource_rds(self, spec):
...
         # pop alternate db name from value before creating the db instance
         # this will only affect the output Secret
         output_resource_db_name = values.pop("output_resource_db_name", None)
 
+        if "event_notifications" in values.keys():
+            event_notifications = values.pop("event_notifications") 
+            counter = 0
+            for e_n in event_notifications:
+                sns_topic_name = e_n.pop("destination")
+                e_n["sns_topic"] = "${aws_sns_topic"+ "." + sns_topic_name + ".arn}"
```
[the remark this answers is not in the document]
OliviaHY: Actually previously we supported both arn and the name as input. Seems reasonable here too.

reconcile/utils/terrascript_aws_client.py
OliviaHY: Since the identifier is of type "$ref": "/common-1.json#/definitions/longIdentifier", which doesn't allow . q-validator actually would throw an error, for example:
```
"{'provider': 'sns', 'identifier': 'test-sns-1.fifo', 'defaults': '/terraform/sns/sns-1.yml', 'subscriptions': [{'protocol': 'email', 'endpoint': 'yihuang@redhat.com'}]} is not valid under any of the given schemas\n\nFailed validating 'oneOf' in schema['properties']['externalResources']['items']['properties']['resources']['items']:\n    {'oneOf': [{'$ref': '/aws/terraform-resource-2.yml'}
```
I thought about making this a string instead and defer it to the integration to handle this so we support both arn or name as input. But I think since we are asking tenant to create SNS topic through a-i and their arn really isn't visible. So I decided go with this. Also more consistent with all the existing identifier.

reconcile/utils/terrascript_aws_client.py
OliviaHY: I would think so. But let me test it out.

reconcile/utils/terrascript_aws_client.py
OliviaHY: Actually I realize that the primary key to the resource can be the SNS topic name combined with `source_type` in this case. Tenant shouldn't be creating separate event notifications for the same `source_type` but different `event_categories`.

reconcile/utils/terrascript_aws_client.py
OliviaHY: Similar to `event_notifications`.  The primary key should be combination of topic name and `protocol`.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1409,10 +1416,26 @@ def populate_tf_resource_rds(self, spec):
         # this will only affect the output Secret
         output_resource_db_name = values.pop("output_resource_db_name", None)
 
+        if "event_notifications" in values:
+            event_notifications = values.pop("event_notifications")
+            for e_n in event_notifications:
+                sns_topic_name = e_n.pop("destination")
+                if sns_topic_name.startswith("arn:"):
+                    raise (ValueError("destination should not be an arn"))
```
bhushanthakur93, line 1424: Will this error be thrown during dry-run mode so we can catch it during PR check?

bhushanthakur93 (approved):
LGTM. Just a small question regarding runtime exception.

reconcile/utils/terrascript_aws_client.py
OliviaHY: Correct! This will be thrown when --dry-run is enabled.

PR #2569, opened 2022-07-11
Enable clusterAdmin attribute for terraform-resources integration

lechuk47:
Enable cluster-admin token for terraform-resources inteagration. There are some cases where this integration need cluster-admin access. 
e.g: `aws-iam-service-account` within a `openshift-*` namespace needs cluster-admin permissions to populate the secret.


maorfr (approved):
7 lines!

PR #2570, opened 2022-07-11
Fix: Set privileged attribute when populating desired state in tf integration

lechuk47:
Missing part of https://github.com/app-sre/qontract-reconcile/pull/2569

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -606,6 +606,7 @@ def populate_desired_state(
                 resource_type=oc_resource.kind,
                 name=spec.output_resource_name,
                 value=oc_resource,
+                privileged=spec.namespace.get("clusterAdmin") or False,
             )
 
 


bkez322 (approved):
LGTM

reconcile/terraform_resources.py
```diff
@@ -606,6 +606,7 @@ def populate_desired_state(
                 resource_type=oc_resource.kind,
                 name=spec.output_resource_name,
                 value=oc_resource,
+                privileged=spec.namespace.get("clusterAdmin", False),
```
geoberle, line 609: when `clusterAdmin` is not defined in a namespace file, the graphql query will return null for it and `get("clusterAdmin", False)` will result in None. 

```suggestion
                privileged=spec.namespace.get("clusterAdmin") or False,
```

lechuk47: You are right. Even though the result is the same is better set this way. 
Fixed.

geoberle: absolutely. but explicit types are always nice :)
thanks for fixing it!

PR #2573, opened 2022-07-12
Locally scope cli.py imports to save memory usage

steveteahan:
Every import in cli.py has to complete before any integration begins
running. This is because import statements are executed when the cli
module is imported, which happens as a result of console_scripts in
setup.py. Currently, memory usage can reach ~120MB before an integration
ever starts running. This change scopes the imports inside the integration
command functions so that the import only happens when the function
is executed (we only expect to execute a single integration function per
container).

There are probably more elegant ways to fix this issue, but for now
this should work and saves us 50MB or more of imports for many
integrations that don't require the importing of the majority of
qontract-reconcile modules.

### Memory usage

Here are some rough estimates of how much memory usage we'll save from some of the tested integrations:

| Integration        | Memory           |
| ------------- |:-------------:|
| quay-membership      | >70MB |
| gitlab-housekeeping      | >50MB      | 
| slack-usergroups      | >50MB      | 
| terraform-resources | <10MB      |
| openshift-* | <10MB      |

For some integrations like quay-membership or gitlab-housekeeping, this is essentially cutting the memory in half. Some of the more memory hungry integrations, like terraform and openshift integrations, do not benefit a lot from this change. It seems that we have some memory hungry imports, in particular imports from `kubernetes.client` alone can account for 40+ MB.

The memory usage will only get worse as new integrations are added, so even if we only get modest decreases in usage for some of the memory hungry integrations, this will at least avoid piling on to the problem.

### Running integrations

All of the integrations above where run in dry-run mode to ensure there wasn't any change in behavior.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,4 +3,4 @@ branch = True
 omit = reconcile/test/*
 
 [report]
-fail_under = 34
+fail_under = 32.5
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -11,111 +11,6 @@ import sentry_sdk
 
 from reconcile.utils import config
 from reconcile.utils import gql
-import reconcile.dyn_traffic_director
-import reconcile.github_org
-import reconcile.github_owners
-import reconcile.github_users
-import reconcile.github_scanner
-import reconcile.github_validator
-import reconcile.openshift_clusterrolebindings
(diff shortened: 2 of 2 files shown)


geoberle (approved):
LGTM ❤️

patjlm (approved):

PR #2574, opened 2022-07-13
[jjb] config_path loading via resource resolver

geoberle:
the jenkins job builder integration loads config_path resources one by one which results in slower reconciliation runs.
to improve that, we use the resource reference resolver feature from qontract-server introduced with https://github.com/app-sre/qontract-server/pull/147

while the impact on integration speed is not that hight for jenkins job builder, the resource resolution capability can be easily showcased on this integration so others like terraform-resources have a working example

part of https://issues.redhat.com/browse/APPSRE-6003
depends on schema change https://github.com/app-sre/qontract-schemas/pull/190

### DIFF
--- a/reconcile/jenkins_job_builder.py
+++ b/reconcile/jenkins_job_builder.py
@@ -28,7 +28,9 @@ QUERY = """
     }
     type
     config
-    config_path
+    config_path {
+      content
+    }
   }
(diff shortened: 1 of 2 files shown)


maorfr:
this is amazing :heart:

PR #2575, opened 2022-07-13
Support updates to a Deployment's label selector.

chassing:
Procedure:
* openshift_base.apply handles FieldIsImmutableError for Deployments
* Checks that the update strategy is RollingUpdate for the current running Deployment; otherwise, reraise FieldIsImmutableError
* Remove old Deployments without removing the last ReplicaSet and Pods
* Apply new Deployment
* Adapt ownership of old ReplicaSet, it'll be removed automatically afterward when the new Deployment is ready

See APPSRE-5728

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -10,7 +10,12 @@ from sretoolbox.utils import retry
 from sretoolbox.utils import threaded
 
 from reconcile import queries
-from reconcile.utils.oc import FieldIsImmutableError, OCClient, OCLogMsg
+from reconcile.utils.oc import (
+    DeploymentFieldIsImmutableError,
+    FieldIsImmutableError,
+    OCClient,
+    OCLogMsg,
+)
 from reconcile.utils.oc import MayNotChangeOnceSetError
 from reconcile.utils.oc import PrimaryClusterIPCanNotBeUnsetError
 from reconcile.utils.oc import InvalidValueApplyError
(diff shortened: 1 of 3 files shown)


reconcile/openshift_base.py
```diff
@@ -345,11 +345,47 @@ def apply(
         except FieldIsImmutableError:
             # Add more resources types to the list when you're
             # sure they're safe.
-            if resource_type not in ["Route", "Service", "Secret"]:
+            if resource_type not in ["Route", "Service", "Secret", "Deployment"]:
```
maorfr, line 348: would it make sense to raise a different exception and handle this in a separate section?

reconcile/openshift_base.py
```diff
@@ -345,11 +345,47 @@ def apply(
...
+                # create new one
+                oc.apply(namespace=namespace, resource=annotated)
+                if obsolete_rs:
+                    oc.wait_for_deployment(namespace, resource.name)
+                    # delete old ReplicSet
+                    oc.delete(
+                        namespace=namespace,
+                        kind=obsolete_rs["kind"],
+                        name=obsolete_rs["metadata"]["name"],
+                    )
```
geoberle, line 385: i would have expected that the new deployment takes ownership of the old ReplicaSet and handles scaledown of its pods during the RollingUpgrade. if that is true (i honestly don't know for sure), the deletion of the replicaset would not be necessary. i'm happy to learn something new here

geoberle: after discussing with @chassing i learned that the ownership reference of the old ReplicaSet does not only include name and kind but also the uid, which changes when the Deployment is recreated.

idea: after replacing the Deployment, update the old ReplicaSet ownership reference with the uid of the new Deployment. this way no waiting for Deployment success and no deletion of the old ReplicaSet "should" be necessary. worth testing because it would imply to get rid of the waiting.

chassing: I tried it, but it doesn't work :( After deletion of the old Deployment, the `ownerReference` disappeared in the old ReplicaSet; trying to add it again, but it was removed immediately.

geoberle: there is a way to do it. when you update the old `ReplicaSet` with the `ownerReference` (with `controller: true`), you also need to update the `metadata.labels` of it to match the `spec.selector.matchLabels` of the `Deployment` (mins the `pod-template-hash`). this is the roughly the way operator SDK does it behind the scenes in its adoption logic. 🤯 

it is a lot more effort but could be still worth it in order to get rid of the waiting.

it is completely up to you if you want to go that road.

chassing: Yeah, that's working like a charm. I'll adapt the code.

chassing: @geoberle PTAL

geoberle: nice. LVGTM

reconcile/openshift_base.py
chassing: @maorfr now ok for you?

geoberle (approved):
LGTM

reconcile/openshift_base.py
maorfr: Yes! Thank you!

PR #2576, opened 2022-07-13
Add website values support for s3 provider [rosa-authenticator]

TGPSKI:
This MR adds support for the "website" and "request_payer" parameters for the S3 provider in terrascrit_aws_client.py.

This property is passed to the provider via a customized defaults file, but is not assigned because of specific property callouts in the S3 provider pattern.


bkez322 (approved):
LGTM

bkez322:
Documenting in case anyone is curious to see-

This is how our RDS defaults file looks for the FedRAMP RDS instance that will be taking advantage of this code change:

```
website:
  index_document: "token.html"
request_payer: "BucketOwner"
```

PR #2585, opened 2022-07-18
[qontract-cli] add app-interface-review-queue get command

maorfr:
following up on the qontract-cli command app-interface-merge-queue (#2577), which shows the queue of merges that are about to happen -
this PR adds a command that shows the MRs that are up for review.

the list is ordered by the last update date, so that first come is first served.
in addition, if there is a label that requests changes or additional attention, in case there was no further responses from the tenant - the MR is filtered out.


chassing:
The lines are very long, mainly due to the HTML links. Did you consider using [Rich](https://rich.readthedocs.io/en/stable/markup.html#links) to print the links?

chassing:
If you'd like, I can also pimp up the output afterward.

maorfr:
this command will result as a markdown file in app-interface-output, so you will not see the link.
does that make sense?

chassing:
Ah I see, yes makes sense

chassing (approved):
LGTM

PR #2593, opened 2022-07-19
sql_query: large query support (#2522)

chassing:
Instead of giving SQL query as a command line parameter, get it via a config map file.
find and delete k8s objects by labels

see APPSRE-5774


chassing:
@rporres please approve this one again.

rporres (approved):
This is the same as #2522, so LGTM :)

PR #2594, opened 2022-07-19
Add rosa-authenticator-vpce-provider

TGPSKI:
This MR creates a new provider for rosa-authenticator in FedRAMP. The networking topology supplied by this provider will be consumed by rosa-authenticator, and achieves the goal of ensuring private networking for sensitive endpoints in the pursuit of FedRAMP high certification.

Todos:
- Make this work
- Add defaults file
- Add schemas

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -121,6 +121,8 @@ from terrascript.resource import (
     aws_api_gateway_integration_response,
     aws_wafv2_web_acl,
     aws_wafv2_web_acl_association,
+    aws_vpc_endpoint,
+    aws_vpc_endpoint_subnet_association,
     random_id,
 )
 
@@ -200,6 +202,8 @@ VARIABLE_KEYS = [
     "vpc_arn",
     "domain_name",
     "certificate_arn",
+    "subnet_ids",
+    "vpc_ids",
 ]
 
 TMP_DIR_PREFIX = "terrascript-aws-"
@@ -1140,6 +1144,8 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
             self.populate_tf_resource_route53_zone(spec)
         elif provider == "rosa-authenticator":
             self.populate_tf_resource_rosa_authenticator(spec)
+        elif provider == "rosa-authenticator-vpce":
+            self.populate_tf_resource_rosa_authenticator_vpce(spec)
         else:
             raise UnknownProviderError(provider)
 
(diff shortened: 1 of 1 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -5152,3 +5158,68 @@ def populate_tf_resource_rosa_authenticator(self, spec):
...
+            cidr_blocks=["10.0.0.0/8"],
+            **vpce_security_group_rule_common_args,
+        )
+        tf_resources.append(aws_security_group_rule_inbound_resource)
+
+        aws_security_group_rule_outbound_resource = aws_security_group_rule(
+            "vpce_outbound",
+            type="egress",
+            security_group_id=f"${{{aws_security_group_resource.id}}}",
+            cidr_blocks=["0.0.0.0/8"],
```
dustman9000, line 5201: this is an invalid cidr block
```suggestion
            cidr_blocks=["0.0.0.0/0"],
```

reconcile/utils/terrascript_aws_client.py
```diff
@@ -5152,3 +5158,68 @@ def populate_tf_resource_rosa_authenticator(self, spec):
...
+            tags={"Name": f"ocm-{identifier}-api-gateway-vpce-sg"},
+        )
+        tf_resources.append(aws_security_group_resource)
+
+        # SG RULES
+        aws_security_group_rule_inbound_resource = aws_security_group_rule(
+            "vpce_inbound",
+            type="ingress",
+            security_group_id=f"${{{aws_security_group_resource.id}}}",
+            cidr_blocks=["10.0.0.0/8"],
```
dustman9000, line 5192: This will need to match the VPC CIDR blocks

TGPSKI: This is how the prototype is written, we're just implementing it. Example with a data source / variable input?

TGPSKI:
@bkez322 these were removed from rosa-authenticator defaults file:

```
vpce_security_group_rule_common_properties:
  to_port: 443
  from_port: 443
  protocol: "tcp"

vpc_endpoint_properties:
  service_name: "com.amazonaws.us-gov-west-1.execute-api"
  vpc_endpoint_type: "Interface"
  private_dns_enabled: true
```

reconcile/utils/terrascript_aws_client.py
bkez322: https://coreos.slack.com/archives/C027UG79RR9/p1658263031244359?thread_ts=1658253407.763229&cid=C027UG79RR9

I'm interpreting this to mean that the current value in the code is fine.

reconcile/utils/terrascript_aws_client.py
bkez322: Updated to match suggestion.

reconcile/utils/terrascript_aws_client.py
dustman9000: Yes, `10.0.0.0/8` should be ok for the current stage/production VPCs, but future VPCs could have different CIDRs.

bkez322 (approved):
This is basically "my" PR now but LGTM!

PR #2595, opened 2022-07-20
[saas-file-owners] support referenced targets

maorfr:
part of https://issues.redhat.com/browse/APPSRE-6028

this PR introduces support for the `saas-file-owners` integration such that tenants can self-service valid saas file changes, even if those changes happen in a referenced saas file target, rather then an inlined one (https://github.com/app-sre/qontract-schemas/pull/200).

### DIFF
--- a/reconcile/saas_file_owners.py
+++ b/reconcile/saas_file_owners.py
@@ -343,8 +343,8 @@ def run(
         saas_file_target_path = diff["target_path"]
         if saas_file_target_path:
             changed_path_matches = [
-            c for c in changed_paths_copy if c.endswith(saas_file_target_path)
-        ]
+                c for c in changed_paths_copy if c.endswith(saas_file_target_path)
+            ]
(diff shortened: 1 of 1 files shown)


reconcile/saas_file_owners.py
```diff
@@ -334,6 +340,11 @@ def run(
         changed_path_matches = [
             c for c in changed_paths_copy if c.endswith(saas_file_path)
         ]
+        saas_file_target_path = diff["target_path"]
+        if saas_file_target_path:
+            changed_path_matches = [
```
janboll, line 345: Perhaps I am not understanding this fully, Is overwriting   `changed_path_matches` really intended? What if I change both?

maorfr: great catch! https://github.com/app-sre/qontract-reconcile/pull/2595/commits/ae50a1633a2788faeb119224afb2be5815f7776a

PR #2596, opened 2022-07-20
Refactor unleash implementation

janboll:
Keep the client, do not create one per Request. This reduces the number
of requests to unleash. This reduces likelyhood of DNS errors due to
failed lookups during cluster upgrades.

Related ticket: APPSRE-5719

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -1231,7 +1231,7 @@ class OC:
             use_native = use_native.lower() in ["true", "yes"]
         else:
             enable_toggle = "openshift-resources-native-client"
-            strategies = get_feature_toggle_strategies(enable_toggle, ["perCluster"])
+            strategies = get_feature_toggle_strategies(enable_toggle)
 
             # only use the native client if the toggle is enabled and this
             # server is listed in the perCluster strategy
--- a/reconcile/utils/unleash.py
+++ b/reconcile/utils/unleash.py
@@ -1,117 +1,94 @@
 import os
-import tempfile
-import shutil
 import threading
+from typing import Mapping, Optional, Any
 
-from UnleashClient import UnleashClient
+from UnleashClient import UnleashClient, BaseCache
 from UnleashClient import strategies
+from UnleashClient.strategies import Strategy
 
-from reconcile.utils.defer import defer
 from reconcile.utils.helpers import toggle_logger
 
-log_lock = threading.Lock()
+client_local = threading.local()
+
+custom_strategies = ["perCluster", "perClusterNamespace"]
+
+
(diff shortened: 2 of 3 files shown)


janboll:
Tested by comparing results with old and new implementation using a couple of feature toggles. Including `openshift-resources-native-client`

reconcile/utils/unleash.py
```diff
@@ -1,117 +1,94 @@
...
-from UnleashClient import UnleashClient
+from UnleashClient import UnleashClient, BaseCache
 from UnleashClient import strategies
+from UnleashClient.strategies import Strategy
 
-from reconcile.utils.defer import defer
 from reconcile.utils.helpers import toggle_logger
 
-log_lock = threading.Lock()
+client_local = threading.local()
```
geoberle, line 11: So I understand correctly that there will be one unleash client per thread?

janboll: Correct, I added a test to make that clearer.

reconcile/utils/oc.py
```diff
@@ -1231,7 +1231,7 @@ def __new__(
             use_native = use_native.lower() in ["true", "yes"]
         else:
             enable_toggle = "openshift-resources-native-client"
-            strategies = get_feature_toggle_strategies(enable_toggle, ["perCluster"])
+            strategies = get_feature_toggle_strategies(enable_toggle)
```
maorfr, line 1234: why is this being removed?

reconcile/utils/unleash.py
geoberle: What is the reason to have one per thread?

janboll: I kept it the way it was. The previous implementation had the same behavior but destroyed the client after every request. 

Let me test and read up if the client is thread safe.

geoberle: thank you for taking additional time looking into this ❤️

reconcile/utils/unleash.py
```diff
@@ -1,117 +1,96 @@
...
 
-from UnleashClient import UnleashClient
+from UnleashClient import UnleashClient, BaseCache
 from UnleashClient import strategies
+from UnleashClient.strategies import Strategy
 
-from reconcile.utils.defer import defer
 from reconcile.utils.helpers import toggle_logger
 
-log_lock = threading.Lock()
```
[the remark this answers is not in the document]
janboll: I'll leave it removed...
https://docs.python.org/2/library/logging.html#thread-safety

fishi0x01 (approved):
Nice refactor! :+1:

PR #2601, opened 2022-07-21
[automated promotions] support referenced targets

maorfr:
part of https://issues.redhat.com/browse/APPSRE-6028

this MR introduces handling for automated promotions in targets which are referenced from saas files and not inlined.

### DIFF
--- a/reconcile/utils/mr/auto_promoter.py
+++ b/reconcile/utils/mr/auto_promoter.py
@@ -87,8 +87,8 @@ class AutoPromoter(MergeRequestBase):
 
             return modified
 
-    @staticmethod
     def process_target(
+        self,
         target: MutableMapping[str, Any],
         promotion_item: Mapping[str, Any],
         commit_sha: str,
@@ -112,7 +112,7 @@ class AutoPromoter(MergeRequestBase):
                 target_updated = True
 
             # Update Promotion data
-            modified = AutoPromoter.process_promotion(
+            modified = self.process_promotion(
                 promotion_item, target_promotion, channels
             )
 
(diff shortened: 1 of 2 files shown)


reconcile/utils/mr/auto_promoter.py
```diff
@@ -114,30 +149,8 @@ def process(self, gitlab_cli):
...
-
-                            # Update Promotion data
-                            modified = AutoPromoter.process_promotion(
-                                item, target_promotion, channels
-                            )
-
-                            if modified:
-                                saas_file_updated = True
+                        if self.process_target(target, item, commit_sha):
+                            saas_file_updated = True
```
geoberle, line 153: refactoring out the logic make the main method real slick ❤️

reconcile/utils/saasherder.py
```diff
@@ -7,7 +7,7 @@
 import hashlib
 import re
 from collections import ChainMap
-from typing import Iterable, Mapping, Any, MutableMapping, Tuple, cast
```
geoberle, line 10: nit: you can use lower case `dict` and `set` without importing them

reconcile/utils/saasherder.py
```diff
@@ -1634,8 +1646,14 @@ def _get_subscribe_saas_file_path_map(saas_files, auto_only=False):
                     subscribe = target_promotion.get("subscribe")
                     if not subscribe:
                         continue
+                    target_path = target.get("path")
```
geoberle, line 1649: for future "us" leave a comment stating that if target has a path it is a dedicated file
it seems obvious now but might be hard to read in the future

geoberle (approved):
i love this. LGTM

PR #2602, opened 2022-07-21
Limit the number of revisions per deployment

lechuk47:
```
❯ time oc get rs | wc -l
   27254

oc get rs  15.03s user 1.09s system 70% cpu 22.773 total
wc -l  0.11s user 0.22s system 1% cpu 22.772 total
```


cubismod (approved):
LGTM, minus those test errors. Not sure why they're failing with such a simple change :thinking:

PR #2603, opened 2022-07-21
Refactor terraform plan

janboll:
Refactor terraform plan. Split up log_plan_diff, cause it has a way to many responsibilities.
Moving the loop, that inspects resource changes into the newly created methods, cause it makes it more readable. However, this causes the code to loop over the same structure multiple times. But, that should be fine, since usually there are not many changed resources 
Related ticket: APPSRE-5975


chassing (approved):
As discussed during our call, please add some comments regarding your test cases. LGTM :)

janboll:
Test cases using dev data:

- Create RDS instance
- Run delete without enabling deletion
- Delete RDS instance using `--enable-deletion` 
- Modify RDS instance
- Create IAM user, ensure `AccountUser` instance is returned

maorfr:
i'm having a hard time understanding the first commit. any chance you will invest some time in making the commits easier to understand?

janboll:
I updated the commit message, does it help? We can also jump on a call to outline my motivation for this a little more if you like.

maorfr:
the motivation is understood: you want to refactor the plan so you could enhance it with more verbose messages of what is going to change.

it's just hard to review when a single commit has so many changes. any chance to break it into smaller pieces? for example, a single commit that extracts some logic to a function is the easiest thing to review. if you can give a series of commits where each one is very easy to review - reviews will be much simpler.

janboll:
Okay, will try to do that.

reconcile/utils/terraform_client.py
```diff
@@ -151,44 +151,53 @@ def plan(self, enable_deletion):
             self.created_users.extend(created_users)
         return disabled_deletions_detected, errors
 
-    @retry()
```
janboll, line 154: @maorfr Do you perhaps remember why this retry was added? I'd assume if there was an API error, i.e. AWS terraform itself would handle this with retries/timeouts. Thus if there was an error during terraform plan my assumption would be that it's more substantial and retries wouldn't matter.

maorfr: git blame for the rescue: #1289

janboll:
Reran the manual tests :white_check_mark:

reconcile/utils/terraform_client.py
```diff
@@ -155,26 +155,85 @@ def plan(self, enable_deletion):
...
         )
         error = self.check_output(name, "plan", return_code, stdout, stderr)
-        disabled_deletion_detected, created_users = self.log_plan_diff(
-            name, tf, enable_deletion
-        )
+        plan = self._get_json_plan(name, tf)
+
+        resource_changes = plan.get("resource_changes")
+        if resource_changes is not None:
+            disabled_deletion_detected = self._detect_disabled_deletion(
```
maorfr, line 173: will `disabled_deletion_detected` not jump between being True and False?

janboll: Right, that would be expected in case a disabled deletion is detected. `disabled_deletion_detected` is assigned two times, first during initialization providing the default value (line 158) and then in this line where we get the value from `_detect_disabled_deletion`. 
Did I get the question right?

janboll: @maorfr can you please elaborate? I don't understand your comment :question:

maorfr: once `disabled_deletion_detected` has been set to True, it should never be False again.

janboll: Got it. That's the case here. False is only assigned once in line 158/201 to declare the variable.

maorfr: but `_detect_disabled_deletion` can return either True or False, and you are setting the "global" variable...
am i just not seeing it?

maorfr: after discussing, the conclusion is that since there is no loop anymore, there is no risk.

PR #2606, opened 2022-07-21
remove es log publishing explicitly with enabled = false

BumbleFeng:
Follow up of https://github.com/app-sre/qontract-reconcile/pull/2588

Due to a [aws API bug ](https://github.com/hashicorp/terraform-provider-aws/issues/5752), terraform can not remove it once it gets added.

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -3477,7 +3477,7 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
         resource: Mapping[str, Any],
         values: Mapping[str, Any],
         output_prefix: str,
-    ) -> tuple[list[dict[str, Any]], list[dict[str, str]]]:
+    ) -> tuple[list[dict[str, Any]], list[dict[str, object]]]:
         """
         Generate cloud_watch_log_group terraform_resources
         for the given resource. Further, generate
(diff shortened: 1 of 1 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -3490,11 +3490,22 @@ def _get_tf_resource_elasticsearch_log_groups(
...
-        log_types = resource["publish_log_types"] or []
-        for log_type in log_types:
+        publish_log_types = resource["publish_log_types"] or []
+        for t in ElasticSearchLogGroupType:
+            log_type = t.value
+            if log_type not in publish_log_types:
+                publishing_options.append(
+                    {
+                        "log_type": log_type,
+                        "enabled": False,
```
fishi0x01, line 3500: Would it make sense to have an explicit `"enabled": True` in corresponding line 3537?
I.e., if we activate a `log_type`, will TF automatically set it to `enabled` or will it keep the latest setting?

fishi0x01 (commented):
Thank you for the fix! Was this tested with our dev-data or similar?

reconcile/utils/terrascript_aws_client.py
BumbleFeng: TF will automatically set it to enable if cloudwatch_log_group_arn is not empty. But I agree to add it. This will make the code more readable.

BumbleFeng:
Yes. I have test it.

fishi0x01 (approved):
:tada:

PR #2609, opened 2022-07-22
Remove plurals in rosa cloudwatch iam policy

dwelch0:


bkez322 (approved):
That is the single most removed comma I've ever seen in my life.

PR #2611, opened 2022-07-22
Use sort key pattern within rosa auth cloudwatch policy

dwelch0:


mjlshen (commented):
This one's the culprit - the assume_role_policy actually exists in AWS atm

bkez322 (approved):
This LGTM.

PR #2612, opened 2022-07-22
APPSRE-4543 - Replace inline policies for managed policies

esron:
JIRA - [APPSRE-4543](https://issues.redhat.com/browse/APPSRE-4543)

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -50,7 +50,6 @@ from terrascript.resource import (
     aws_iam_user,
     aws_s3_bucket_notification,
     aws_iam_access_key,
-    aws_iam_user_policy,
     aws_iam_group,
     aws_iam_group_policy_attachment,
     aws_iam_user_group_membership,
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -2438,7 +2438,6 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
 
         # iam user policy for bucket
         values = {}
-        values["user"] = identifier
         values["name"] = identifier
         policy = {
             "Version": "2012-10-17",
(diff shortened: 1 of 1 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -1863,8 +1861,16 @@ def populate_tf_resource_s3(self, spec):
...
         values["policy"] = json.dumps(policy, sort_keys=True)
         values["depends_on"] = self.get_dependencies([user_tf_resource])
-        tf_resource = aws_iam_user_policy(identifier, **values)
-        tf_resources.append(tf_resource)
+        tf_aws_iam_policy = aws_iam_policy(identifier, **values)
+        tf_resources.append(tf_aws_iam_policy)
+
+        tf_user_policy_attachment = aws_iam_user_policy_attachment(
+            identifier,
+            user=identifier,
```
TGPSKI, line 1869: These values are identical, so there's not much practical difference between referencing the object or the identifier.

I'm a fan of:

```
user=f"{{{user_tf_resource.name}}}",
```

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1863,8 +1861,16 @@ def populate_tf_resource_s3(self, spec):
         }
         values["policy"] = json.dumps(policy, sort_keys=True)
         values["depends_on"] = self.get_dependencies([user_tf_resource])
-        tf_resource = aws_iam_user_policy(identifier, **values)
-        tf_resources.append(tf_resource)
+        tf_aws_iam_policy = aws_iam_policy(identifier, **values)
+        tf_resources.append(tf_aws_iam_policy)
+
+        tf_user_policy_attachment = aws_iam_user_policy_attachment(
```
TGPSKI, line 1867: One thing to note here is that we are mixing resource invocation styles. Older code in this file follows this pattern:

```
values = {}
values['resource-property']
...
tf_aws_iam_policy = aws_iam_policy(identifier, **values)
```

Where as our style is more HCL-like:

```
tf_aws_iam_policy_attachment = aws_iam_user_policy_attachment(
                identifier,
                user=identifier,
                policy_arn=f"{{{tf_aws_iam_policy.arn}}}",
                depends_on=self.get_dependencies([user_tf_resource, tf_aws_iam_policy]),
            )
tf_resources.append(tf_aws_iam_policy_attachment)
```

Personally, I think the second style is much easier to read and makes objects look more like HCL, which is welcome in my opinion.

I don't have any direct actions out of this comment, just a note. If I had it my way, I would make everything look more like HCL and get rid of building value dictionaries line by line.

Really, if I had it my way, every provider would be a module written in HCL and instantiated in each provider creation method.

reconcile/utils/terrascript_aws_client.py
TGPSKI: This being said, its not a must change or needs to be fixed for me to approve it.

TGPSKI (approved):
The blast radius of these changes seems large. I would recommend not merging until Monday.

PR #2619, opened 2022-07-25
Create local OC class to allow for meaningful errors

suzana-nesic:
Part of [APPSRE-3866](https://issues.redhat.com/browse/APPSRE-3866)

We have two integrations that rely on server having the value of None in the OC class since it relies on one method in said class. The goal of this MR is to take that method out into its own class and refactor OC to return meaningful errors whenever client/api_kind_version need to be set

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -370,9 +370,7 @@ class OCDeprecated:  # pylint: disable=too-many-public-methods
             "-f",
             "-",
         ] + parameters_to_process
-        result = self._run(
-            self, cmd, stdin=json.dumps(template, sort_keys=True)
-        )
+        result = self._run(cmd, stdin=json.dumps(template, sort_keys=True))
         return json.loads(result)["items"]
 
     def release_mirror(self, from_release, to, to_release, dockerconfig):
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -964,9 +964,14 @@ class OCNative(OCDeprecated):
         if server:
             self.client = self._get_client(server, token)
             self.api_kind_version = self.get_api_resources()
-        else:
+        elif not server and not token and local:
             init_api_resources = False
             init_projects = False
+        else:
+            raise Exception("a server value is needed")
+        # else:
+        #     init_api_resources = False
+        #     init_projects = False
 
         self.object_clients = {}
         self.init_projects = init_projects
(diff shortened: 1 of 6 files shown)


reconcile/utils/oc.py
```diff
@@ -1207,6 +1195,43 @@ def is_kind_supported(self, kind: str) -> bool:
...
+    ):
+        self.cluster_name = cluster_name
+        oc_base_cmd = ["oc", "--kubeconfig", "/dev/null"]
+        if insecure_skip_tls_verify:
+            oc_base_cmd.extend(["--insecure-skip-tls-verify"])
+
+        self.jump_host = None
+        if jh is not None:
+            self.jump_host = JumpHostSSH(jh, settings=settings)
+            oc_base_cmd = self.jump_host.get_ssh_base_cmd() + oc_base_cmd
```
maorfr, line 1214: are these needed for local operations?

reconcile/utils/oc.py
```diff
@@ -1207,6 +1195,43 @@ def is_kind_supported(self, kind: str) -> bool:
...
+            parameters = {}
+        parameters_to_process = [f"{k}={v}" for k, v in parameters.items()]
+        cmd = [
+            "process",
+            "--local",
+            "--ignore-unknown-parameters",
+            "-f",
+            "-",
+        ] + parameters_to_process
+        result = OCDeprecated._run(
```
janboll, line 1229: Calling a hidden method in another class is a little odd, even though it looks like a static one.  Should process perhaps be part of that OCDreprecated class?

dwelch0: A potential solution would be to have the new `OCLocal` inherit from `OCDeprecated` in the same fashion as `OCNative` currently does.

reconcile/utils/oc.py
```diff
@@ -373,6 +359,46 @@ def _msg_to_process_reconcile_time(self, namespace, resource):
...
```
dwelch0, line 373: now that `process()` is back within `OCDeprecated`, I believe this can be changed back to original `self._run()`

reconcile/utils/oc.py
```diff
@@ -1208,6 +1201,20 @@ def is_kind_supported(self, kind: str) -> bool:
...
+    def __init__(
+        self,
+        cluster_name,
+        server,
+        token,
+    ):
+        super().__init__(
+            cluster_name=cluster_name,
+            server=server,
+            token=token,
```
dwelch0, line 1214: `local=True`

PR #2624, opened 2022-07-27
[openshift-saas-deploy] consider planned data optional resources

maorfr:
follows up on #2261

dependent resources can be defined as optional: https://kubernetes.io/docs/concepts/configuration/secret/#restriction-secret-must-exist

in such cases, we don't want to fail a deployment.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -704,7 +704,9 @@ def _validate_resources_used_exist(
     name: str,
     used_kind: str,
 ) -> None:
-    used_resources = oc.get_resources_used_in_pod_spec(spec, used_kind)
+    used_resources = oc.get_resources_used_in_pod_spec(
+        spec, used_kind, include_optional=False
+    )
     for used_name, used_keys in used_resources.items():
         # perhaps used resource is deployed together with the using resource?
         resource = ri.get_desired(cluster, namespace, used_kind, used_name)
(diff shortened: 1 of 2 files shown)


maorfr:
for completeness, here is the result of `oc explain pod --recursive`:
```
KIND:     Pod
VERSION:  v1

DESCRIPTION:
     Pod is a collection of containers that can run on a host. This resource is
     created by clients and scheduled onto hosts.

FIELDS:
   ...
   spec	<Object>
      ...
      containers	<[]Object>
         args	<[]string>
         command	<[]string>
         env	<[]Object>
            name	<string>
            value	<string>
            valueFrom	<Object>
               configMapKeyRef	<Object>
                  key	<string>
                  name	<string>
                  optional	<boolean>
               fieldRef	<Object>
                  apiVersion	<string>
                  fieldPath	<string>
               resourceFieldRef	<Object>
                  containerName	<string>
                  divisor	<string>
                  resource	<string>
               secretKeyRef	<Object>
                  key	<string>
                  name	<string>
                  optional	<boolean>
         envFrom	<[]Object>
            configMapRef	<Object>
               name	<string>
               optional	<boolean>
            prefix	<string>
            secretRef	<Object>
               name	<string>
               optional	<boolean>

      ...
      volumes	<[]Object>
         ...
         configMap	<Object>
            defaultMode	<integer>
            items	<[]Object>
               key	<string>
               mode	<integer>
               path	<string>
            name	<string>
            optional	<boolean>
         ...
         secret	<Object>
            defaultMode	<integer>
            items	<[]Object>
               key	<string>
               mode	<integer>
               path	<string>
            optional	<boolean>
            secretName	<string>
```

rporres (approved):

PR #2625, opened 2022-07-27
APPSRE-4543 - Add managed user policies

esron:
This is the first step in migrating from inline policies to managed policies.
[Details](https://issues.redhat.com/browse/APPSRE-4543)
[Reference](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_managed-vs-inline.html#convert-inline-to-managed-policy)


maorfr (approved):
LGTM

PR #2631, opened 2022-07-28
run make test as part of a push

maorfr:
if tests are failing, we shouldn't proceed to build and push an image.

this is a safety net for merges with failing tests (like #2626 :facepalm:)


janboll (approved):

lechuk47 (approved):
LGTM

PR #2633, opened 2022-07-28
github-repo-permissions-validator validate repo permissions for more jobs in more instances

maorfr:
this integration was assuming that jobs in github can only be in a single instance. that is not true.

it would also be beneficial to enforce write permissions in build jobs, so we will be able to update commit status after a push, and not only in pr-checks.


maorfr:
do not merge until permission to all repos has been granted, so this can be merged and promoted.

PR #2616, opened 2022-07-25
Unleash client refactor

janboll:
Fixing previous PR, that required rollback: https://github.com/app-sre/qontract-reconcile/pull/2596

PR #2636, opened 2022-07-28
Unleash client refactor

janboll:
Standard Strategies used in the last try, see https://github.com/app-sre/qontract-reconcile/pull/2616 returned wrong results.

Added `DisableNativeClientStrategy`, which returns True, unless a cluster is added to the cluster_name list of the feature.  If the feature is disabled it is also set to false (see https://github.com/Unleash/unleash-client-python/blob/main/UnleashClient/features/Feature.py#L64). Additionally, I did add some tests, that use httpretty to ensure this behavior.

### DIFF
--- a/reconcile/utils/unleash.py
+++ b/reconcile/utils/unleash.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import threading
 from typing import Mapping, Optional, Any
@@ -6,7 +7,6 @@ from UnleashClient import UnleashClient, BaseCache
 from UnleashClient import strategies
 from UnleashClient.strategies import Strategy
 
-from reconcile.utils.helpers import toggle_logger
 
 client: Optional[UnleashClient] = None
 client_lock = threading.Lock()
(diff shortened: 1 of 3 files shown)


BumbleFeng:
Since we are here. I propose we rename `perCluster` to `disableCluster` and add `DisableClusterStrategy`. It will be much clear on the unleash side while we may reuse the strategy for other integrations.

janboll:
Very good point, I will update this PR.

reconcile/utils/unleash.py
```diff
@@ -1,117 +1,92 @@
...
-        with toggle_logger():
-            # create Unleash client
-            headers = {"Authorization": f"Bearer {token}"}
+    def destroy(self):
+        self.cache = {}
+
+
+class DisableClusterStrategy(Strategy):
+    def load_provisioning(self) -> list:
+        return [x.strip() for x in self.parameters["cluster_name"].split(",")]
```
geoberle, line 36: tiny nit: since `cluster_name` can hold a list of clusters we could name it `clusters_names` or `cluster_names_csv`

geoberle (approved):
LGTM

reconcile/utils/unleash.py
janboll: The unleash UI abstracts the list, however, that name is directly displayed there. You enter one cluster name at a time. If I change it to be plular, it might not be easy to understand in Unleash, since the name would suggest you can enter multiple names.
So, I'd prefer to leave it as is...

geoberle: thanks for the clarification!

janboll:
relates to: APPSRE-5719

PR #2637, opened 2022-07-28
qontract-cli review-queue: consider trigger phrases

chassing:
Don't skip MRs when the last comment is from AppSRE but a trigger phrase.

See APPSRE-6079


cubismod (approved):
LGTM

PR #2640, opened 2022-07-29
Revert "APPSRE-4543 - Add managed user policies (#2625)"

fishi0x01:
We have an open alert for this. Revert.

```
Error: Error attaching policy {aws_iam_policy.ocm-sts-access.arn} to IAM User {ocm-sts-access}: ValidationError: The specified value for userName is invalid. It must contain only alphanumeric characters and/or the following: +=,.@_-
```

This reverts commit f1df0cd2ef28a53f9b7f0b15592450a3dcea11f4.

cc @esron


chassing (approved):
LGTM

PR #2642, opened 2022-07-29
Add LOG_LEVEL to dockerfile run-integration script

rporres:


cubismod:
In slack you said you wanted to default to DEBUG level in Docker but this defaults to INFO unless I'm missing something

rporres:
most likely a bad phrase on my side.  We want the possibility of setting DEBUG if we want to. The default should not change 

I've corrected the Slack one :)

cubismod (approved):
LGTM then!

geoberle (approved):

PR #2643, opened 2022-07-29
APPSRE-4543 Fix - Add inline user policies

esron:
[JIRA](https://issues.redhat.com/browse/APPSRE-4543)

This time I tested against our dev environment with plan and apply.


maorfr (approved):
3rd time is the charm!

PR #2247, opened 2022-03-16
saasherder add support for secret parameters

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4716

depends on https://github.com/app-sre/qontract-schemas/pull/101

PR #2646, opened 2022-08-01
[saas-file-owners] add control over secretParameters

maorfr:
a very late follow up on #2247

with this change, saas file owners will be able to self-service changes in `secretParameters` sections.
example: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/44240

to restrict users from using any secret path they want: #2648


cubismod (approved):
LGTM

PR #2650, opened 2022-08-01
expiration date for tf-r deletionApprovals should contain current date

BumbleFeng:


bhushanthakur93 (approved):
LGTM!

I had run into this during my IC shift and it confused me initially. Thank you for fixing it.

rporres (approved):

PR #2651, opened 2022-08-02
terraform-resources: s3 bucket versioning

chassing:
fix typo in getting s3 versioning parameter

APPSRE-6048


rporres:
Can you explain what's going on? Because if there's no `versioning` in `common_values`, both expressions return the same...

```python
>>> common_values = {}
>>> common_values.get("versioning") or True
True
>>> common_values.get("versioning", True)
True
```

chassing:
you're right and that's the problem. see here

```
>>> common_values = {"versioning": False}
>>> common_values.get("versioning") or True
True
>>> common_values.get("versioning", True)
False
```

rporres (approved):
Ahh, ok, I wasn't looking at the actual problem you were trying to solve 😂 

LGTM

PR #2653, opened 2022-08-02
[gitlab-housekeeping] remove labels when not added by allowed users

maorfr:
depends on https://github.com/app-sre/qontract-schemas/pull/212

example usage in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/44367

with this PR, if a user adds a label but they are not within the allowed labels roles, the label will be removed. this is to handle users with Developer access adding labels to MRs in gitlab.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -203,9 +203,11 @@ def get_merge_requests(dry_run: bool, gl: GitLabApi) -> list:
         )
         label_events = mr.resourcelabelevents.list()
         for label in reversed(label_events):
-            if label.action == "add" and label.label["name"] in MERGE_LABELS_PRIORITY:
+            if label.action == "add":
+                added_by = label.user["username"]
+                if label.label["name"] in MERGE_LABELS_PRIORITY:
                 approved_at = label.created_at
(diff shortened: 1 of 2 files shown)


reconcile/gitlab_housekeeping.py
```diff
@@ -196,17 +198,34 @@ def get_merge_requests(dry_run: bool, gl: GitLabApi) -> list:
...
             continue
 
+        label_events = mr.resourcelabelevents.list()
+        approval_found = False
+        for label in reversed(label_events):
+            if label.action == "add":
+                label_name = label.label["name"]
+                added_by = label.user["username"]
+                if users_allowed_to_label:
+                    if added_by not in users_allowed_to_label:
```
chassing, line 208: Can be in one line and would be easier to read :)

maorfr: thanks! https://github.com/app-sre/qontract-reconcile/pull/2653/commits/4a06dcc73f37c9d2fa4c0b28faea43ffa96d15c5

chassing (approved):
LGTM

rporres:
late to the party, looks good, though :)

PR #2656, opened 2022-08-02
Remove inline user policies

esron:
JIRA - [APPSRE-4543](https://issues.redhat.com/browse/APPSRE-4543)

This one is removing the inline user policies referenced in the ticket.


maorfr:
this PR will cause many resources to be deleted. in deletion protected accounts, this would mean a ton of deletionApprovals. is there a way to avoid it? something like "resources of type aws_iam_user_policy are always allowed for deletion"?

esron:
@maorfr since [APPSRE-6100](https://issues.redhat.com/browse/APPSRE-6100) is done can I merge this one?

maorfr (approved):

PR #2658, opened 2022-08-03
[saasherder] check images when imagePatterns is empty

maorfr:
we are currently skipping this test if imagePatterns is empty.

example: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/44340


chassing (approved):
LGTM

PR #2662, opened 2022-08-03
if MR is not assigned it can not be last assigned

maorfr:
example where this is failing: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/43616


chassing (approved):
LGTM

PR #2671, opened 2022-08-04
[saasherder] fix url paths in error messages

maorfr:
fix the url from logs to something that is a valid link.
we've seen tenants assuming that the broken link means a problem with the integration, so let's at least get that out of the way.


chassing (approved):
LGTM

PR #2672, opened 2022-08-04
[pr-check] early exit for integration dry-runs when no change in desired state

geoberle:
introduce an early exit run mode for integrations during dry-runs that checks if the desired state has changed between two bundles. if no change in desired state can be found, the integration exits early and saves time and resources.

if a cli flag `--early-exit-compare-sha` provides a bundle sha, integrations supporting early exit will fetch the desired state from the comparison and the current bundle and compare them. if no difference an be found, the integration exits early.

this early exit feature is supposed to improve speed for PR checks. early exit will only be considered if the integration has been started in dry-run mode.

an integration supports early exit, if the integration module provides a function

```
def early_exit_desired_state(*args, **kwargs) -> dict[str, Any]
```

it is crucial that the `early_exit_desired_state` leverages the same functionality/code to fetch the desired state, as the `run` method does, in order to prevent a drift between those two. this function also needs to incorporate hidden desired state like vault secrets from v1 secret engines etc.

if any error occurs during the desired state fetching or the comparison, the integration will not exit early and will instead proceed with its regular execution.

this PR also showcases the usage of early exit for `terraform-vpc-peering`. the cluster query for this integration has been stripped down to the required fields to have a minimal desired state that does not prevent early exit unnecessarily.

the app-interface pr-check coordination logic will be adapted to pass the --early-exit-compare-sha to those integrations that will have an `early_exit` flag enabled in the `integration-1.yml` schema (see https://github.com/app-sre/qontract-schemas/pull/217)

i suggest reviewing this PR commit by commit 

how was the terraform-vpc-peering change tested?
generated a tf file before and after the change for all accounts and it was exactly the same

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -477,10 +477,7 @@ def early_exit_integration(
     from deepdiff import DeepDiff
 
     diff = DeepDiff(previous_desired_state, current_desired_state, ignore_order=True)
-    if diff:
-        return False
-    else:
-        return True
+    return not diff
 
 
 def init_log_level(log_level):
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -393,28 +393,31 @@ def run_integration(func_container, ctx, *args, **kwargs):
     early_exit_compare_sha = ctx.get("early_exit_compare_sha")
 
     try:
-        if early_exit_compare_sha:
-            # check if the integration can exit early because there is no difference
-            # in desired state compared to the provided comparison bundle sha
-            # early exit is only supported when the integration is started in
-            # dry-run mode
-            can_exit_early = dry_run and early_exit_integration(
+        # check if the integration can exit early because there is no difference
+        # in desired state compared to the provided comparison bundle sha
+        # early exit is only supported when the integration is started in
+        # dry-run mode
+        can_exit_early = (
+            dry_run
+            and early_exit_compare_sha
+            and early_exit_integration(
                 int_name, early_exit_compare_sha, func_container, args, kwargs
             )
-            if can_exit_early:
-                logging.info("No changes in desired state. Exit PR check early.")
-            else:
-                try:
-                    gql.init_from_config(
-                        autodetect_sha=ctx["gql_sha_url"],
-                        integration=int_name,
-                        validate_schemas=ctx["validate_schemas"],
-                        print_url=ctx["gql_url_print"],
-                    )
-                except GqlApiIntegrationNotFound as e:
-                    sys.stderr.write(str(e) + "\n")
-                    sys.exit(ExitCodes.INTEGRATION_NOT_FOUND)
-                func_container.run(dry_run, *args, **kwargs)
+        )
+        if can_exit_early:
+            logging.info("No changes in desired state. Exit PR check early.")
+        else:
+            try:
+                gql.init_from_config(
+                    autodetect_sha=ctx["gql_sha_url"],
+                    integration=int_name,
+                    validate_schemas=ctx["validate_schemas"],
+                    print_url=ctx["gql_url_print"],
+                )
+            except GqlApiIntegrationNotFound as e:
+                sys.stderr.write(str(e) + "\n")
+                sys.exit(ExitCodes.INTEGRATION_NOT_FOUND)
+            func_container.run(dry_run, *args, **kwargs)
     except RunnerException as e:
         sys.stderr.write(str(e) + "\n")
         sys.exit(ExitCodes.ERROR)
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -75,6 +75,21 @@ def log_level(function):
     return function
 
 
+def early_exit(function):
+    help_msg = (
+        "Runs integration in early exit mode. If the observed desired state of "
+        "an integration does not change between the provided bundle SHA "
+        "--early-exit-compare-sha and the current bundle sha, the integration "
+        "exits without spending more time with reconciling."
+    )
+    function = click.option(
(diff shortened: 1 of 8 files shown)


reconcile/cli.py
```diff
@@ -369,26 +384,37 @@ def run_integration(func_container, ctx, *args, **kwargs):
...
     if not unleash_feature_state:
         logging.info("Integration toggle is disabled, skipping integration.")
         sys.exit(ExitCodes.SUCCESS)
 
     dry_run = ctx.get("dry_run", False)
+    early_exit_compare_sha = ctx.get("early_exit_compare_sha")
 
     try:
-        func_container.run(dry_run, *args, **kwargs)
+        if early_exit_compare_sha:
```
patjlm, line 396: there is no `else` here, so we will never run if this param is not set ?

reconcile/cli.py
```diff
@@ -402,6 +428,58 @@ def run_integration(func_container, ctx, *args, **kwargs):
...
+        current_desired_state = func_container.early_exit_desired_state(*args, **kwargs)
+    except Exception:
+        logging.exception("Failed to fetch desired state for current bundle failed")
+        return False
+
+    # compare
+    from deepdiff import DeepDiff
+
+    diff = DeepDiff(previous_desired_state, current_desired_state, ignore_order=True)
+    if diff:
```
patjlm, line 480: nit: `return not diff` ?

reconcile/cli.py
geoberle: [49883f4](https://github.com/app-sre/qontract-reconcile/pull/2672/commits/49883f48add320f7ec73933c80efae056ab63c6f)

reconcile/cli.py
geoberle: [29dea50](https://github.com/app-sre/qontract-reconcile/pull/2672/commits/29dea5026825be99a02c48411d1cce88f62c932b)

patjlm (approved):
This is simply amazing and very well laid out. Thanks!

reconcile/terraform_vpc_peerings.py
```diff
@@ -456,6 +456,208 @@ def build_desired_state_vpc(
...
+    disable {
+      integrations
+    }
+  }
+}
+"""
+
+
+def fetch_clusters_with_peerings() -> list[dict[str, Any]]:
+    clusters = gql.get_api().query(CLUSTER_QUERY)["clusters"]
```
fishi0x01, line 657: why are you re-defining the `CLUSTER_QUERY` here? Why not use `queries.get_clusters()`?

reconcile/utils/gql.py
```diff
@@ -229,26 +229,38 @@ def get_git_commit_info(sha, server, token=None):
...
     token = config["graphql"].get("token")
-    if sha_url:
+    if sha:
+        server = server_url._replace(path=f"/graphqlsha/{sha}").geturl()
+    elif autodetect_sha:
         sha = get_sha(server_url, token)
         server = server_url._replace(path=f"/graphqlsha/{sha}").geturl()
-
+    if sha:
         runing_state = RunningState()
```
cubismod, line 259:
```suggestion
        running_state = RunningState()
```

reconcile/utils/gql.py
```diff
@@ -229,26 +229,38 @@ def get_git_commit_info(sha, server, token=None):
...
         git_commit_info = get_git_commit_info(sha, server_url, token)
-        runing_state.timestamp = git_commit_info.get("timestamp")
-        runing_state.commit = git_commit_info.get("commit")
-
-    if print_url:
-        logging.info(f"using gql endpoint {server}")
-    return init(server, token, integration, validate_schemas)
+        runing_state.timestamp = git_commit_info.get("timestamp")  # type: ignore[attr-defined]
+        runing_state.commit = git_commit_info.get("commit")  # type: ignore[attr-defined]
```
cubismod, line 262:
```suggestion
        running_state.timestamp = git_commit_info.get("timestamp")  # type: ignore[attr-defined]
        running_state.commit = git_commit_info.get("commit")  # type: ignore[attr-defined]
```

cubismod (approved):
changes LGTM :)
just found some small typos prior to your changes

reconcile/terraform_vpc_peerings.py
geoberle: The other query is too big and returns way too many details which has the potential to prevent early exit unnecessarily. 

Does that make sense?

geoberle: Schemasparsamkeit 😜

reconcile/utils/gql.py
maorfr: one would say this is unrelated to the changes in the PR :sweat_smile:

reconcile/utils/gql.py
maorfr: but since these lines are changed, ii could make sense to fix the typo :smile:

fishi0x01 (approved):

reconcile/utils/gql.py
geoberle: [f6aedd9](https://github.com/app-sre/qontract-reconcile/pull/2672/commits/f6aedd9113e00d0faa506d6c6a2f34087a3a7d59)

reconcile/utils/gql.py
geoberle: [f6aedd9](https://github.com/app-sre/qontract-reconcile/pull/2672/commits/f6aedd9113e00d0faa506d6c6a2f34087a3a7d59)

PR #2673, opened 2022-08-04
[gitlab-housekeeping] automatically close auto merge MRs that cannot be merged

maorfr:
MRs that are created with a `bot/automerge` label are automated, and are not looked at after their creation.
some of these MRs get to a merge conflict situation and can not be merged.

let's close them automatically!

example: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/44593

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -120,6 +120,8 @@ def handle_stale_items(dry_run, gl, days_interval, enable_closing, item_type):
     for item in items:
         item_iid = item.attributes.get("iid")
         item_labels = item.attributes.get("labels")
+        if AUTO_MERGE in item_labels and item.merge_status == "cannot_be_merged":
+            close_item(dry_run, gl, enable_closing, item_type, item)
         notes = item.notes.list()
         note_dates = [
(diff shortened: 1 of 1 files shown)


reconcile/gitlab_housekeeping.py
```diff
@@ -99,6 +120,8 @@ def handle_stale_items(dry_run, gl, days_interval, enable_closing, item_type):
     for item in items:
         item_iid = item.attributes.get("iid")
         item_labels = item.attributes.get("labels")
+        if AUTO_MERGE in item_labels and item.merge_status == "cannot_be_merged":
```
rporres, line 123: This `cannot_be_merged` looks a bit magic but it actually comes from the API: https://docs.gitlab.com/ee/api/merge_requests.html#single-merge-request-response-notes

I would make it a constant and add the API reference link in a comment so that it looks less magical :)

rporres (approved):
A minor thing, looks good otherwise.

reconcile/gitlab_housekeeping.py
maorfr: i'll do it in a follow up PR, since there are more occurences.

maorfr: #2681

PR #2674, opened 2022-08-04
[qontract-cli] add get app-interface-merge-history command

maorfr:
this command (running in app-interface-output) will allow us to look into intervals between merges to get an understanding of how slow/fast things are going.

this is a complement to `qontract-cli get app-interface-merge-queue`.

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -1174,6 +1174,35 @@ def app_interface_review_queue(ctx):
     print_output(ctx.obj["options"], queue_data, columns)
 
 
+@get.command()
+@click.pass_context
+def app_interface_merge_history(ctx):
+    settings = queries.get_app_interface_settings()
+    instance = queries.get_gitlab_instance()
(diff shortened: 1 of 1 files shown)


maorfr:
app-interface-output MR: https://gitlab.cee.redhat.com/service/app-interface-output/-/merge_requests/20

rporres:
So you're just getting the last 100, right? Every time we run it, we will get potentially a lot of merge requests that were in the previous run and there's a slight possibility that if we run not enough frequently, we miss something.

How about adding a date so that we know how many have been merged since a moment in time instead of the arbitrary 100? We will then be able to avoid overlaps

rporres:
ok, so you're rewriting the whole thing with each run and you will see the "MERGED_AT" column to get the idea.

rporres (approved):
Ok, so you rewrite each time the final page and you take a look into the `MERGED_AT` column to get an idea of the speed...

I'm not sure how useful will be since I guess it would be interesting to compare days, etc.. But it is out of the scope of the MR. Let's try this.

PR #2675, opened 2022-08-04
Temp fix to handle the comparation of OCI images

rporres:
sretoolbox currently doesn't handle well the comparation of OCI images
with content-type 'application/vnd.oci.image.index.v1+json'. These are
manifest list type of images that don't have `mediaType` in their
manifest so sretoolbox logic fails.

The real fix is in sretoolbox.container.image module, but that's going
to take a bit longer since we will need to change how image comparation
is done as it will need to inspect the manifest headers instead of the
manifest itself.

In the meantime, we have this to avoid the integration to crash.

### DIFF
--- a/reconcile/quay_mirror.py
+++ b/reconcile/quay_mirror.py
@@ -232,8 +232,26 @@ class QuayMirror:
                             )
                             continue
                     except ImageComparisonError as details:
-                        _LOG.error("[%s]", details)
+                        _LOG.error(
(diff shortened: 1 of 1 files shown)


rporres:
A failed attempt on fixing this from sretoolbox. https://github.com/app-sre/sretoolbox/pull/74

maorfr (approved):

PR #2677, opened 2022-08-04
[terraform-aws-client] validate email for sns topic subscriptions

jfchevrette:
Terraform plan will not do any validation with AWS if the email address is invalid

However when applying, AWS will complain and crash the integration if the email address is not valid

Given the email address `jfchevretteredhat.com` (missing an @ sign)
```
[terraform-resources] error: b'\nError: error creating SNS Topic Subscription: InvalidParameter: Invalid parameter: Email address\n\tstatus code: 400, request id: REDACTED\n\n  on config.tf.json line 15213, in resource.aws_sns_topic_subscription.REDACTED:\n15213:       },\n\n\n'
[terraform-resources] [REDACTED - apply] Error: error creating SNS Topic Subscription: InvalidParameter: Invalid parameter: Email address
```

This change will do a regex-based validation of the email address


OliviaHY (approved):
Thanks for catching and fixing this!

bhushanthakur93:
Alternatively, we can also do this validation on json-schema through https://json-schema.org/understanding-json-schema/reference/string.html#built-in-formats

jfchevrette:
@bhushanthakur93 good point but after researching this a bit:
- [qontract-validator](https://github.com/app-sre/qontract-validator/blob/6a2cc1fbe04e0b0c4131e8fa48fdbdf53501c2ac/validator/validator.py) does not do format checking. so currently it appears that any use of json-schema's `format` field will pass validation (we use none at the moment)
- If we do email validation at the schema level, the user experience for an invalid email will be a schema validation error (currently, but could be improved of course) which does not say anything but "the whole schema does not validate"
- Checking the format in the integration makes it possible to return a more user friendly error. The only downside I see is that checking at the integration level we catch the issue at a much later time in the process which can be annoying

I'll go ahead and merge this as-is and we can decide to implement format checking at a later time

PR #2678, opened 2022-08-05
APPSRE-6103 check saas upstream deps present

fishi0x01:
If a saas file has an upstream dependency to ci-int/ci-ext, then it should be verified by `service_dependencies`.

### DIFF
--- a/reconcile/gql_queries/service_dependencies/service_dependencies.py
+++ b/reconcile/gql_queries/service_dependencies/service_dependencies.py
@@ -32,6 +32,15 @@ query ServiceDependencies {
       pipelinesProvider {
         provider
       }
+      resourceTemplates {
+        targets {
+          upstream {
+            instance {
+              name
+            }
+          }
+        }
+      }
     }
     quayRepos {
       org {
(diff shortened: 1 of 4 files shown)


reconcile/gql_queries/service_dependencies/service_dependencies.py
```diff
@@ -32,6 +32,15 @@
...
       }
+      resourceTemplates {
+        targets {
+          upstream {
+            instance {
+              name
+            }
+          }
+        }
+      }
```
chassing, line 43: Would it make sense to switch to the new query_string() method?

fishi0x01: Im not sure I can follow. This file is generated by qenerate.

chassing: latest qenerate version generates a `query_string` method instead of having a QUERY variable. see https://github.com/chassing/qontract-reconcile/blob/APPSRE-6089__quay_team_creation/reconcile/gql_queries/quay_membership/quay_membership.py#L14

fishi0x01: We decided against `query_string()` because of potential path mismatches. Depending on how you include the module, the path to the file might not work properly.

The latest qenerate version should not have `query_string` anylonger :thinking:

fishi0x01: https://github.com/app-sre/qenerate/pull/23

chassing: Ok, I have an older version installed which still uses the `query_string` method :( Then I'm ok here. I'll reintroduce the `query_string` method in generate, but with a different file path; I like this method :)

chassing (approved):
LGTM

reconcile/gql_queries/service_dependencies/service_dependencies.py
fishi0x01: When you rebase your PR to latest master, then the CI should also fail because if generation mismatch. There was a new CI check added recently https://github.com/app-sre/qontract-reconcile/pull/2663

PR #2679, opened 2022-08-05
upgrade to latest qenerate version

chassing:
update service_dependencies and vpc_peerings_validator to be compliant with the latest qenerate version


fishi0x01:
Great thank you! :heart: 

I think it makes sense to pin the version now https://github.com/app-sre/qontract-reconcile/blob/master/requirements/requirements-type.txt#L14

The CI check might fail unrelated PRs if it always pulls from latest

chassing:
Yes, you're right. We should pin the version right now.

fishi0x01 (approved):

PR #2682, opened 2022-08-07
[aws-iam-keys] store deleted keys in state

maorfr:
part of https://issues.redhat.com/browse/APPSRE-6092

the integration will now store a state per account with the list of keys that were deleted.
if the integration is executed and no new keys are found - it avoids initiating aws api.
in turn, this will save a lot of calls to IAM, and probably stop rate limiting issues.


cubismod (approved):
This is a pretty novel approach to the problem, using State, nice work!

PR #2683, opened 2022-08-08
Set OVNKubernetes as the default CNI

patjlm:
https://issues.redhat.com/browse/APPSRE-6108

`OVNKubernetes` is becoming the default CNI as of 4.11. All app-sre new clusters should use it.

This depends on all app-interfaces managed clusters to be updated to not rely on the old default. In commercial: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/44807

@TGPSKI @cubismod @bkez322 Could we sync for fedramp ?


reconcile/test/fixtures/clusters/cluster1.yml
```diff
@@ -28,8 +28,8 @@ spec:
   private: false
   provision_shard_id: 'ashardid'
 network:
+  type: OpenShiftSDN
```
fishi0x01, line 31: Potentially a stupid question, but why we test with `OpenShiftSDN` instead of `OVNKubernetes`?

patjlm: I actually left the tests as is. They were relying on the default `OpenshiftSDN` being used, and were testing upgrades with the network type set to `OpenshiftSDN`. With the new default, this causes a test error stating the upgrade is not allowed.

fishi0x01: To avoid confusion, do you think it makes sense to add a comment there to highlight that this was the previous default?

fishi0x01 (approved):

cubismod:
Changes to FR AI to add that label are pretty easy so I've opened an MR for that in fedramp.

patjlm:
> Changes to FR AI to add that label are pretty easy so I've opened an MR for that in fedramp.

Thanks @cubismod . I'll get this merged and promoted in commercial tomorrow then.

bkez322:
@patjlm I think I get what's going on here, but I would appreciate you confirming that I understand exactly what we need to do here... 😅

So the steps we'll want to follow in FedRAMP (which should be the same as commercial) are:

1. Merge an app-interface change like [this](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/44807/diffs) to avoid "change an immutable field" integration-run errors. (To be honest I don't entirely understand why this error would get thrown in the first place- but I trust you that it does). As @cubismod mentioned, he opened a fedramp MR like this already, we just need to merge it- which I'll assume we will do today. Presumably, there will be no actual changes to anything as a result of merging this.

2. After this Q-R PR is merged, promote a new QR version with this PR included in it. Presumably, once again, there will be no actual changes to anything as a result of merging this.

3. Merge another app-interface change again similar to [this](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/44807/diffs), but set `OpenShiftSDN` to `OVNKubernetes` everywhere. This WILL cause actual changes to every single cluster... but we expect that nothing of real relevance to us will change or break.

cubismod:
@bkez322 I think that last bit is what we have to figure out, how to do a migration between `OpenShiftSDN` to `OVNKubernetes` for our existing fedramp clusters.

patjlm:
1 and 2 are correct, but not 3. There is no migration path from openshiftsdn to ovnkubernetes for now. This is why we have this immutable field story btw.
So existing clusters running openshiftsdn will keep running with it. We will probably plan to replace all clusters with on based ones, but that is much longer term and will be progressive of course.

bkez322:
@patjlm

Got it, we will entirely skip my "step 3" for now, and we will follow commercial's lead on probably entirely replacing every cluster to use `OVNKubernetes` in the future as a longish term goal.

Thanks! 👍👍👍

bkez322:
We merged our FedRAMP "step 1" change so we're ready for this PR to be merged anytime.

reconcile/test/fixtures/clusters/cluster1.yml
patjlm: In the context of this PR, it makes sense to state it, and your comment makes so perfectly fine, thanks you!
In the context of the code & test sample, I don't really see what is changes whether that matches the current default value or not. So I would not add a comment in the code I think. Does that make sense?

PR #2685, opened 2022-08-08
jira-client ignore arg-type

maorfr:
not able to reproduce this locally, but we need CI passing.
```
reconcile/utils/jira_client.py:37: error: Argument 1 to "extend" of "list" has incompatible type "Union[Dict[str, Any], ResultList[Issue]]"; expected "Iterable[Dict[str, Any]]"  [arg-type]
```

### DIFF
--- a/reconcile/utils/jira_client.py
+++ b/reconcile/utils/jira_client.py
@@ -34,7 +34,7 @@ class JiraClient:
         while True:
             index = block_num * block_size
             issues = self.jira.search_issues(jql, index, block_size, **kwargs)
-            all_issues.extend(issues)
+            all_issues.extend(issues)  # type: ignore[arg-type]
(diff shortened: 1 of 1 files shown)


reconcile/utils/jira_client.py
```diff
@@ -34,7 +34,7 @@ def get_issues(self, fields: Optional[Mapping] = None) -> GottenIssue:
         while True:
             index = block_num * block_size
             issues = self.jira.search_issues(jql, index, block_size, **kwargs)
-            all_issues.extend(issues)
+            all_issues.extend(issues)  # type: ignore[arg-type]
```
fishi0x01, line 37: Can we add a `TODO` here to keep this marked as under investigation / not working properly?

PR #2688, opened 2022-08-08
Properly handle return type from jira_client

fishi0x01:
The `jira` module had a wrong type hint which was recently fixed https://github.com/pycontribs/jira/commit/cc2508485c12232a4a9c4a56ee74175ed818ee20

Our current logic expects a `ResultList` to be returned.

This PR ensures that only `ResultList` is processed and occurrences of `Dict` log a warning.

**Local Test on Prod Data**

```
qontract-reconcile_1  | [2022-08-08 10:57:57] [INFO] [gql.py:init_from_config:250] - using gql endpoint http://qontract-server:4000/graphqlsha/eb59c0d13fd7232954c7aa3eb3fd74509a9bece78a486261f9b49864cf669578
qontract-reconcile_1  | [2022-08-08 10:58:01] [INFO] [jira_watcher.py:act:84] - https://issues.redhat.com/browse/ASIC-273 ([FIRING:1] QontractReconcileIntegrationStuck2hours app-sre-prod-01 app-interface) status change: Done -> To Do
qontract-reconcile_1  | [2022-08-08 10:58:01] [INFO] [jira_watcher.py:act:84] - https://issues.redhat.com/browse/ASIC-283 ([FIRING:1] QontractReconcileIntegrationStuck2hours appsrep05ue1 app-interface) status change: Done -> To Do
```

### DIFF
--- a/reconcile/utils/jira_client.py
+++ b/reconcile/utils/jira_client.py
@@ -1,11 +1,11 @@
+import logging
+
 from jira import JIRA, Issue
 from jira.client import ResultList
 
 from reconcile.utils.secret_reader import SecretReader
 
 from typing import Any, Iterable, Mapping, Optional, Union
-
-GottenIssue = Union[list[dict[str, Any]], ResultList[Issue]]
 
 
 class JiraClient:
@@ -22,20 +22,31 @@ class JiraClient:
         token_auth = self.secret_reader.read(token)
         self.jira = JIRA(self.server, token_auth=token_auth)
 
-    def get_issues(self, fields: Optional[Mapping] = None) -> GottenIssue:
+    def get_issues(self, fields: Optional[Mapping] = None) -> list[Issue]:
         block_size = 100
         block_num = 0
 
-        all_issues: GottenIssue = []
+        all_issues: list[Issue] = []
         jql = "project={}".format(self.project)
         kwargs: dict[str, Any] = {}
         if fields:
             kwargs["fields"] = ",".join(fields)
         while True:
             index = block_num * block_size
-            issues = self.jira.search_issues(jql, index, block_size, **kwargs)
-            # TODO: investigate why the arg-type ignore is needed in CI
-            all_issues.extend(issues)  # type: ignore[arg-type]
+            issues: Union[dict[str, Any], ResultList[Issue]] = self.jira.search_issues(
+                jql, index, block_size, **kwargs
+            )
+            if not isinstance(issues, ResultList):
+                # if search_issues was executed with json_result=True, then we have a Dict.
+                # However, we require a ResultList.
+                # See https://github.com/pycontribs/jira/commit/cc2508485c12232a4a9c4a56ee74175ed818ee20
+                logging.warning(
+                    "Jira client did not receive a ResultList."
+                    " Maybe the call was made with json_result=True which is"
+                    " currently not supported."
+                )
+                continue
+            all_issues.extend(issues)
             if len(issues) < block_size:
                 break
             block_num += 1
--- a/reconcile/utils/jira_client.py
+++ b/reconcile/utils/jira_client.py
@@ -38,10 +38,10 @@ class JiraClient:
             )
             if not isinstance(issues, ResultList):
                 # if search_issues was executed with json_result=True, then we have a Dict.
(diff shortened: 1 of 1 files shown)


reconcile/utils/jira_client.py
```diff
@@ -22,20 +22,31 @@ def __init__(
...
+                jql, index, block_size, **kwargs
+            )
+            if not isinstance(issues, ResultList):
+                # if search_issues was executed with json_result=True, then we have a Dict.
+                # However, we require a ResultList.
+                # See https://github.com/pycontribs/jira/commit/cc2508485c12232a4a9c4a56ee74175ed818ee20
+                logging.warning(
+                    "Jira client did not receive a ResultList."
+                    " Maybe the call was made with json_result=True which is"
+                    " currently not supported."
```
chassing, line 46: It would be better to analyze `kwargs` for the unsupported `json_result` parameter and raise an exception. Then you can always expect `ResultList[Issue]` as the result type.

fishi0x01: My understanding is that we should not raise exceptions in the reconcile loop, because of the way reconcile counts the errors in some integrations. Since this is a utils class, I was trying to avoid bringing in exceptions in potentially unexpected places.

reconcile/utils/jira_client.py
```diff
@@ -22,20 +22,31 @@ def __init__(
...
+                )
+                continue
```
janboll, line 48: Adding continue here changes the control structure of that loop. Before the if-condition on line 50 was always reached. Could that be an issue?

fishi0x01: Before the if condition would not have reached if a non-ResultList was returned. The `extend()` call would break on non-iterable dict

reconcile/utils/jira_client.py
chassing: Ah, ok, then the warning instead of the exception but expecting `ResultList[Issue]` as the result type is easier to understand.

janboll: API documentation says it could be both. So +1 on checking the type this way.

reconcile/utils/jira_client.py
janboll: That makes sense! one last question, what if nothing but ResultLists are returned?

reconcile/utils/jira_client.py
fishi0x01: Adjusted `ResultList` -> `ResultList[Issue]`.

reconcile/utils/jira_client.py
fishi0x01: > what if nothing but ResultLists are returned?

Do you by chance mean "nothing but Dicts are returned"? Currently we only have ResultLists, because `json_result=True` is nowhere being used. Our logic works for that use-case. Dicts are not supported by us and you would just see a lot of warnings in the logs.

If we want to support `json_result=True`, then we must make adjustments here - it would require us to understand what those Dicts consist of. However, the purpose of this PR was solely to make sure we are working on only objects we expect, without breaking already existing behavior.

fishi0x01: To be honest, I feel like this is a weakness of the jira module API to return a union like `Union[Dict, List]`. Those types are fundamentally different and should not be returned by the same function. But yeah this is an naive outside view - might be there is a good reason for doing this :)

reconcile/utils/jira_client.py
chassing: Have a look here #2689. Depending on the type of the input it returns either a `list[dict]` or `list[RoleProtocol]` which is very similar here too. 
Defining the return type as `Union[Dict, List]` looks strange to me, therefore I would do something like

```python
while True:
    if kwargs.get("json_result"):
        logging.warning("...")
        continue
    issues: ResultList[Issue] = self.jira.search_issues(
                 jql, index, block_size, **kwargs
             )
    ...

`´`

fishi0x01: Further, Im not sure if mypy will be happy with the above suggestion. It has no knowledge of `json_result` -> it will expect a Union type as a return.

chassing (approved):
LGTM

PR #2689, opened 2022-08-08
expiration.filter: gql classes support

chassing:
Refactor utils.expiration.filter to support roles as a list of generated GQL classes or dictionaries.

BREAKING CHANGE: remove unused `key` parameter 

prerequisite for #2664


reconcile/test/test_utils_expiration.py
```diff
@@ -1,79 +1,109 @@
 from datetime import date, timedelta
-from typing import Dict, List
+from typing import Optional
+from pydantic import BaseModel
+
+# from typing import Dict, List
```
fishi0x01, line 5: left over

reconcile/utils/expiration.py
```diff
@@ -1,41 +1,41 @@
 import datetime
+from typing import Iterable, Optional, Protocol, TypeVar, Union, cast
 
+DATE_FORMAT = "%Y-%m-%d"
+
+
+class RoleProtocol(Protocol):
```
fishi0x01, line 7: nit: suggesting `FilterableRole`?

reconcile/utils/expiration.py
```diff
@@ -1,41 +1,41 @@
...
-    for r in roles:
-        expiration_date = r[key]
-        if not has_valid_expiration_date(expiration_date):
+    for r in roles or []:
+        if isinstance(r, dict):
+            key = "expirationDate"
+            expiration_date = r.get(key)
+        else:
+            key = "expiration_date"
+            expiration_date = getattr(r, key, None)
```
fishi0x01, line 31: Why not `expiration_date = r.expiration_date`?

fishi0x01 (commented):
Nice refactor :heart: 
Some small questions/comments

reconcile/test/test_utils_expiration.py
chassing: jupp 🤦

reconcile/utils/expiration.py
chassing: 👍 much better naming

reconcile/utils/expiration.py
chassing: That's a very good question! :)

reconcile/utils/expiration.py
```diff
@@ -1,41 +1,41 @@
...
-    if role is None:
-        return True
-    else:
-        exp_date = datetime.datetime.strptime(role, "%Y-%m-%d").date()
-        current_date = datetime.datetime.utcnow().date()
-        if current_date < exp_date:
-            return True
-        return False
 
+def date_expired(role: str) -> bool:
```
fishi0x01, line 16: nit: `role` -> `date`

reconcile/utils/expiration.py
```diff
@@ -1,41 +1,41 @@
...
             raise ValueError(
-                f"{key} field is not formatted as YYYY-MM-DD, "
-                f"currently set as {expiration_date}"
+                f"{key} field is not formatted as YYYY-MM-DD, currently set as {expiration_date}"
             )
-        if role_still_valid(expiration_date):
-            filtered.append(r)
 
-    return filtered
+    return cast(DictsOrRoles, filtered)
```
fishi0x01, line 41: Out of curiosity: would this work without a `cast`, by declaring

```
filtered: DictsOrRoles = []
```

above?

fishi0x01 (commented):
Just to be sure: was this tested on some integrations locally with `dry-run`?

+2 more small nit things - feel free to neglect :)

reconcile/utils/expiration.py
chassing: 👍

reconcile/utils/expiration.py
chassing: That was my first thought as well but it didn't work because `DictsOrRoles` is `Iterable` and it doesn't support `append`. I tried it with `list` but then mypy throws this error :(

```
reconcile/quay_membership.py:97: error: <nothing> has no attribute "__iter__" (not iterable)  [attr-defined]
reconcile/quay_membership.py:97: error: Argument 1 to "filter" has incompatible type "Optional[List[RoleV1]]"; expected "None"  [arg-type]
```

fishi0x01: Interesting. By removing the `Iterable` from the type alias it seems to work locally for me. But I had to resort to `getattr()` again, because `Protocol` apparently doesnt like `isinstance` checks - its like turning in circles :sweat_smile: 

```python
DictOrRole = TypeVar(
    "DictOrRole", bound=Union[FilterableRole, dict]
)


def filter(roles: Optional[Iterable[DictOrRole]]) -> list[DictOrRole]:
    """Filters roles and returns the ones which are not yet expired."""
    filtered: list[DictOrRole] = []
    for r in roles or []:
        if isinstance(r, dict):
            key = "expirationDate"
            expiration_date = r.get(key)
        else:
            key = "expiration_date"
            expiration_date = getattr(r, key)

        try:
            if not expiration_date or not date_expired(expiration_date):
                filtered.append(r)
        except ValueError:
            raise ValueError(
                f"{key} field is not formatted as YYYY-MM-DD, currently set as {expiration_date}"
            )

    return filtered
```

Basically, Im trying to find a way to avoid `getattr()` and `cast()` because they are not ideal for catching stuff in static type checking phase. But I fail to find a way xD

fishi0x01: I think it would be a good idea to also test the dict path. Just in case, because this potentially affects quite a few integrations.

chassing: I've edited my comment. Tested both, class-based with #2664 and the dict-based path with the current implementation of `quay-membership`

fishi0x01: Apparently this works:

```python
@runtime_checkable
class FilterableRole(Protocol):
    expiration_date: Optional[str]


DictOrRole = TypeVar(
    "DictOrRole", bound=Union[FilterableRole, dict]
)


def filter(roles: Optional[Iterable[DictOrRole]]) -> list[DictOrRole]:
    """Filters roles and returns the ones which are not yet expired."""
    filtered: list[DictOrRole] = []
    for r in roles or []:
        if isinstance(r, FilterableRole):
            key = "expiration_date"
            expiration_date = r.expiration_date
        else:
            key = "expirationDate"
            expiration_date = r.get(key)

        try:
            if not expiration_date or not date_expired(expiration_date):
                filtered.append(r)
        except ValueError:
            raise ValueError(
                f"{key} field is not formatted as YYYY-MM-DD, currently set as {expiration_date}"
            )

    return filtered
```

No `cast()`, no `getattr()`. Wdyt?

fishi0x01 (approved):

reconcile/utils/expiration.py
chassing: Are you sure that `Iterable[DictOrRole]]` is semantic correct? `roles` should be either a list of `FilterableRole`s or a list of `dict`s, but `Iterable[DictOrRole]]` has mixed elements; am I right?

chassing: I get this error:

```
reconcile/utils/expiration.py:32: error: Item "FilterableRole" of the upper bound "Union[FilterableRole, Dict[Any, Any]]" of type variable "DictOrRole" has no attribute "get"  [union-attr]
```

fishi0x01: > I get this error:

Note: I changed the branches of the if-statement. I think you still might have the old order, which uses `.get()` in the first `if` branch

fishi0x01: > Are you sure that Iterable[DictOrRole]] is semantic correct? roles should be either a list of FilterableRole's or a list of dict's, but Iterable[DictOrRole]] has mixed elements; am I right?

Good point. Yeah semantically this is flawed :sweat:

PR #2690, opened 2022-08-08
Fix memory requests for SQL jobs / cronJobs

andreugallofre:
With the default limits from app interface the ratio between the memory request and the limit should be
4 or less. With request_mem at 64Mi the ratio is 8, so in some namespaces, this leads to errors

https://coreos.slack.com/archives/CCRND57FW/p1659958449416839


chassing (approved):
LGTM! I didn't know this. Thx

cubismod (approved):
LGTM

PR #2694, opened 2022-08-09
openshift realize data error when caller is different from resource caller and this is not a take over

maorfr:
to prevent cases of two saas files fighting over the same resources -
if this is a saas file (caller) and it's not a take over, error out when checking a current item that is also desired.

related to https://issues.redhat.com/browse/OHSS-14340


geoberle (approved):
lgtm

PR #2696, opened 2022-08-11
terraform-tgw-attachment early exit and schema reduction

geoberle:
enable early exit dry-run for`terraform-tgw-attachment`

also refactor the cluster query used by `terraform-vpc-peering` to be usable by `terraform-tgw-attachment` while reducing unrequired schemas. especially the `/access/user-1.yml` schema is not required anymore, which means that PR checks for users (quite frequent) will not even trigger a PR check for the mentioned integrations anymore.


maorfr:
does this need a rebase? i'm seeing vpc-peerings changes but i believe these were merged already?

geoberle:
> does this need a rebase? i'm seeing vpc-peerings changes but i believe these were merged already?

true :) rebased!

maorfr (approved):
too many changes in a single PR, but i'll let it slide

LGTM :heart:

PR #2698, opened 2022-08-12
[github-org] early exit pr-check dry-runs

geoberle:

### DIFF
--- a/reconcile/github_org.py
+++ b/reconcile/github_org.py
@@ -90,9 +90,13 @@ CLUSTERS_QUERY = """
 QONTRACT_INTEGRATION = "github"
 
 
-def get_config(default=False):
+def get_orgs():
     gqlapi = gql.get_api()
-    orgs = gqlapi.query(ORGS_QUERY)["orgs"]
+    return gqlapi.query(ORGS_QUERY)["orgs"]
+
+
+def get_config(default=False):
+    orgs = get_orgs()
     secret_reader = SecretReader(queries.get_secret_reader_settings())
     config = {"github": {}}
     found_defaults = []
@@ -503,6 +507,6 @@ def run(dry_run):
 
 def early_exit_desired_state(*args, **kwargs) -> dict[str, Any]:
     return {
-        "github_org_config": get_config(),
+        "github_orgs": get_orgs(),
         "github_org_members": fetch_desired_state().dump(),
     }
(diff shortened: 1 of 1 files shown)


reconcile/github_org.py
```diff
@@ -499,3 +499,10 @@ def run(dry_run):
     )
 
     runner.run()
+
+
+def early_exit_desired_state(*args, **kwargs) -> dict[str, Any]:
+    return {
+        "github_org_config": get_config(),
```
maorfr, line 506: why isn't this enough? https://github.com/app-sre/qontract-reconcile/blob/4c5749bb25ebb08741c3f1579a259801c5f53fad/reconcile/github_org.py#L95

reconcile/github_org.py
```diff
@@ -499,3 +499,10 @@ def run(dry_run):
...
+        "github_org_members": fetch_desired_state().dump(),
```
maorfr, line 507: wouldn't you prefer to replace this with `permissions` and `clusters`?

maorfr:
both these comments are nits, but i'm trying to help us understand what is the bare minimum we need to do in early exit.

reconcile/github_org.py
maorfr: do you mean managed teams?

maorfr: if so, the list of managed teams is a part of the query: https://github.com/app-sre/qontract-reconcile/blob/4c5749bb25ebb08741c3f1579a259801c5f53fad/reconcile/github_org.py#L32

geoberle: yes sry i meant `managedTeams`. the query you posted is exactly used by `get_config()`

geoberle: maybe i misunderstood you question. let me revisit

geoberle: ah ok i get it. you mean "why the additional processing of `get_config` if the core information required is get result of the query. good point. i will refactor it

geoberle: see [34feef4](https://github.com/app-sre/qontract-reconcile/pull/2698/commits/34feef40903e8a03583bbeb0ddcef7fc7dae39d4)

reconcile/github_org.py
geoberle: queried `clusters` are intertwined with `openshift_users.fetch_desired_state(oc_map=None)` so this processing done in `fetch_desired_state` seems to meaningful for getting the picture about the desired state.

maorfr (approved):

PR #2699, opened 2022-08-15
gitlab-housekeeping if MR status is unchcked for automerge - check it harder

maorfr:
when listing MRs, some come back with `merge_status: unchcked`. calling the merge request endpoint gets this information.

follows up on #2673

ref: https://docs.gitlab.com/ee/api/merge_requests.html#merge-requests-list-response-notes

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -124,7 +124,7 @@ def handle_stale_items(dry_run, gl, days_interval, enable_closing, item_type):
                 # this call triggers a status recheck
                 item = gl.get_merge_request(item_iid)
             if item.merge_status == MRStatus.CANNOT_BE_MERGED:
-            close_item(dry_run, gl, enable_closing, item_type, item)
+                close_item(dry_run, gl, enable_closing, item_type, item)
         notes = item.notes.list()
         note_dates = [
             datetime.strptime(note.attributes.get("updated_at"), DATE_FORMAT)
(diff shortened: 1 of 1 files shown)


maorfr:
this MR was closed by a local run with these changes: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/45183

reconcile/gitlab_housekeeping.py
```diff
@@ -119,8 +119,12 @@ def handle_stale_items(dry_run, gl, days_interval, enable_closing, item_type):
     for item in items:
         item_iid = item.attributes.get("iid")
         item_labels = item.attributes.get("labels")
-        if AUTO_MERGE in item_labels and item.merge_status == MRStatus.CANNOT_BE_MERGED:
-            close_item(dry_run, gl, enable_closing, item_type, item)
+        if AUTO_MERGE in item_labels:
+            if item.merge_status == MRStatus.UNCHECKED:
+                # this call triggers a status recheck
+                item = gl.get_merge_request(item_iid)
+            if item.merge_status == MRStatus.CANNOT_BE_MERGED:
```
janboll, line 126: nit: IMHO `elif` would make it a little easier to read

maorfr: we are getting a new item to process, so an `elif` would not be logically correct?

janboll (approved):

reconcile/gitlab_housekeeping.py
janboll: Whoops. Never would have expected that and thus overlooked it.

PR #2701, opened 2022-08-15
[quay-mirror] early exit

maorfr:
based on #2672

following #2700, quay-mirror now fails when it attempts to mirror a repo from docker hub publicly. this failure will happen when the integration runs.
this integration was disabled from running on pr-checks in the commercial app-interface in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/21108, to remediate extensive calls to docker hub.

this means that MRs with public mirroring may be merged, and only later quay-mirror will start failing. we need to catch these cases in pr-check.

at the same time, running quay-mirror on pr-check is time consuming. the integration takes a lot of time to run. if only there was a way to run this integration only when relevant content is being changed... oh, wait! early exit!

this PR implements early exit for quay-mirror, such that it fails (really really) fast in case someone tries to publicly mirror a repo, and it will actually run in dry-run only when relevant content is changed.

### DIFF
--- a/reconcile/quay_mirror.py
+++ b/reconcile/quay_mirror.py
@@ -6,6 +6,7 @@ import tempfile
 import time
 
 from collections import defaultdict, namedtuple
+from typing import Any
 
 from sretoolbox.container.image import ImageComparisonError
 from sretoolbox.container.skopeo import SkopeoCmdError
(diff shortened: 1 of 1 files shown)


geoberle (approved):
❤️

maorfr:
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/45217

maorfr:

PR #2705, opened 2022-08-16
extract safe_resource_id helper method to reconcile.utils.terraform

jfchevrette:
Small change to extract the helper method `safe_resource_id` into `reconcile.utils.terraform` as suggested here https://github.com/app-sre/qontract-reconcile/pull/2691#discussion_r945878684

PTAL @bhushanthakur93


bhushanthakur93 (approved):
LGTM. Thank you for the change.

PR #2706, opened 2022-08-17
[openshift-resources*] early exit prepare

maorfr:
some refactoring (no logic change) to prepare for early exit for openshift-resources based integrations.

PR #2707, opened 2022-08-17
[openshift-vault-secrets] early exit

maorfr:
with this integration, the desired state is captured within namespace files, so the early exit is quite simple.

depends on #2706


geoberle (approved):
nice. lgtm

maorfr:

PR #2708, opened 2022-08-17
[openshift-routes] early exit

maorfr:
the desired state for routes is captured within namespaces and referenced resources.
there is also a part if vault (TLS), but in case the vault path/version does not change, we can safely ignore it.

using `fetch_provider_resource` fetches the resource, and is really the only relevant part of `fetch_provider_route`: https://github.com/app-sre/qontract-reconcile/blob/af0a7fcdf102686075bc79450e0fd73710c1db33/reconcile/openshift_resources_base.py#L441

the rest is validations and adding tls data from vault.

depends on https://github.com/app-sre/qontract-reconcile/pull/2706

### DIFF
--- a/reconcile/openshift_routes.py
+++ b/reconcile/openshift_routes.py
@@ -1,3 +1,5 @@
+from typing import Any
+
 import reconcile.openshift_resources_base as orb
 
 from reconcile.utils.semver_helper import make_semver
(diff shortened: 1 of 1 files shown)


reconcile/openshift_routes.py
```diff
@@ -28,3 +30,17 @@ def run(
         cluster_name=cluster_name,
         namespace_name=namespace_name,
     )
+
+
+def early_exit_desired_state(*args, **kwargs) -> dict[str, Any]:
+    namespaces, _ = orb.get_namespaces(PROVIDERS)
+    resources = [
+        orb.fetch_provider_resource(r["path"]).body
```
geoberle, line 38: shouldn't you filter on the `route` provider here?

maorfr: it's filtered in `orb.get_namespaces`.

geoberle: oh i just realized `canonicalize_namespace` is doing in-place filtering based on the providers. look good them

geoberle (approved):

maorfr:

PR #2711, opened 2022-08-17
[openshift-resources] early exit

maorfr:
still WIP, but this is already working.

PR #2712, opened 2022-08-17
Add early exit support to prometheus-rules-tester

rporres:
State changes from this integration will come from rules changing
(either directly or via jinja parameters) or tests themselves changing.
This will certainly save some time, but not as much as we would like
since `get_prometheus_tests` is quite slow.

I will come back to the effort of refactoring the integration, or at
least the part that deals with fetching the prometheus rules.

### DIFF
--- a/reconcile/prometheus_rules_tester.py
+++ b/reconcile/prometheus_rules_tester.py
@@ -4,6 +4,7 @@ import re
 import sys
 import traceback
 import yaml
+from typing import Any
 
 from sretoolbox.utils import threaded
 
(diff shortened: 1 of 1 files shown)


geoberle:
you mentioned `get_prometheus_tests` being slow. can you put down some numbers?

geoberle:
is the performance issue at the core of this the same as in https://github.com/app-sre/qontract-reconcile/pull/2711#discussion_r947853769, namely `orb.fetch_openshift_resource` ?

rporres:
In my laptop, it takes about 2 minutes to execute `early_exit_desired_state`.

From the experiments in https://issues.redhat.com/browse/APPSRE-5699, I was able to reduce substantially the time that `get_prometheus_tests` was taking by using multiprocessing, hence my suspicion about the serializing being the bottleneck.

maorfr (approved):

maorfr:

PR #2714, opened 2022-08-17
[terraform-resources] early exit

maorfr:
the desired state is represented by:
- aws accounts
- namespaces with external resources
- resources referenced from defaults path (or parameter_group for rds).

with no yaml loading, an early exit takes.................. 6 seconds!

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -7,6 +7,7 @@ from typing import Any, Iterable, Optional, Mapping, Tuple, cast
 
 from sretoolbox.utils import threaded
 from reconcile.utils.external_resources import (
+    PROVIDER_AWS,
     get_external_resource_specs,
     managed_external_resources,
 )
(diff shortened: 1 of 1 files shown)


reconcile/terraform_resources.py
```diff
@@ -716,3 +721,25 @@ def run(
...
+
+def early_exit_desired_state(*args, **kwargs) -> dict[str, Any]:
+    gqlapi = gql.get_api()
+    namespaces = get_tf_namespaces()
+    resources = []
+    for ns_info in namespaces:
+        for spec in get_external_resource_specs(
+            ns_info, provision_provider=PROVIDER_AWS
+        ):
+            defaults = spec.resource.get("defaults")
```
geoberle, line 734: SQS and dynamodb have their defaults nested

reconcile/terraform_resources.py
```diff
@@ -716,3 +721,25 @@ def run(
...
+            if defaults:
+                resources.append(gqlapi.get_resource(defaults))
```
geoberle, line 736: Nit: This is again my opportunity to shamelessly plug resource ref loading 😊

We should consider it for a followup PR

reconcile/terraform_resources.py
maorfr: good catch! https://github.com/app-sre/qontract-reconcile/pull/2714/commits/3a50e9c0509c30753d27eab8100e2c3e89a4494d

geoberle (approved):
I'm not entirely happy with the "hardcoded" default fields but this approach works correct and gets the feature out there way faster than refactoring for resource ref loading. Therefore Kudos for this nice adoption of early exit. LGTM ❤️

maorfr:

PR #2715, opened 2022-08-17
APPSRE-5972 obey to default GQL nullable standard

fishi0x01:
qenerate received an [update](https://github.com/app-sre/qenerate/pull/34) which aligns it closer to the GQL standard.

Currently, our GQL schema allows null elements in lists in many places. Sadly, this results in more `Optional` fields and more null checks in the code.

Note, that in order to avoid these `Optional`s, we must adjust our GQL schema properly to enforce `NonNull` fields in lists (see https://issues.redhat.com/browse/APPSRE-6153).

**Validation**

Ran integrations locally in dry-run mode

### DIFF
--- a/reconcile/gql_queries/introspection.json
+++ b/reconcile/gql_queries/introspection.json
@@ -4144,9 +4144,89 @@
                         }
                     ],
                     "inputFields": null,
-                    "interfaces": [],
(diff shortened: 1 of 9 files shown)


chassing (approved):
LGTM

PR #2716, opened 2022-08-17
[openshift-resources] early exit monkey patch to context manager

maorfr:

fixes #2711

monkey patching is causing a faulty behavior in case early exit determined a full dry-run is needed.
this PR makes the monkey patching use a context manager, to restore the previous functions.

### DIFF
--- a/reconcile/openshift_resources.py
+++ b/reconcile/openshift_resources.py
@@ -1,3 +1,4 @@
+from contextlib import contextmanager
 from typing import Any
 
 import reconcile.openshift_base as ob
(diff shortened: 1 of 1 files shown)


reconcile/openshift_resources.py
```diff
@@ -40,30 +41,52 @@ def run(
...
+    check_alertmanager_config = orb.check_alertmanager_config
+
+    try:
+        yield early_exit_monkey_patch_assign(
+            lambda path, key, version=None, tvars=None, settings=None: f"vault({path}, {key}, {version})",
+            lambda repo, path, ref, tvars=None, settings=None: f"github({repo}, {path}, {ref})",
+            lambda url: False,
+            lambda data, path, alertmanager_config_key, decode_base64=False: True,
+        )
+    finally:
```
geoberle, line 74: Ah yes resetting the monkey patch. Sry i did not catch this during the review.

geoberle (approved):
LGTM

reconcile/openshift_resources.py
maorfr: i didn't even think about it

PR #2717, opened 2022-08-17
[terraform-users] early exit

maorfr:
we only care about accounts and the roles query (covering roles, users, aws accounts, groups)


geoberle (approved):

maorfr:

PR #2721, opened 2022-08-18
gitlab-housingkeeping: don't fight with gitlab-labeler

chassing:
Keep track of who added a label and remove it when no one else with proper permissions added it as well.


maorfr (approved):
nice implementation!

PR #2722, opened 2022-08-18
[openshift-resources] early exit enable query support

maorfr:
related to https://issues.redhat.com/browse/APPSRE-6098

depends on https://github.com/app-sre/qontract-schemas/pull/225

resources with dynamic content (generated based on graphql query results) must be checked, but only them.
with this field, we will know exactly the ones to check. we will add a validation that if the query() function is used, this option must be set to true.


chassing (approved):
I hope everything is thread safe 😁 

LGTM

PR #2723, opened 2022-08-18
gitlab-housekeeping: harder label retrieval

chassing:
Sometimes the label attribute is empty but shouldn't. Try it again by fetching the MR separately

similar to https://github.com/app-sre/qontract-reconcile/pull/2699


maorfr:
There are cases of missing labels as well (while other labels exist). But this is a good direction!

chassing:
Oh yes 🤦‍♂️ let me refactor that.

geoberle (approved):

PR #2724, opened 2022-08-18
[terraform-resources] cloudwatch validate elasticsearch domain exists

maorfr:
the cloudwatch provider accepts a field called `es_identifier` and assumes this resource exists. when it doesn't exist, apply fails, but plan passes.

with this PR, plan will fail.


geoberle (approved):
nice

PR #2732, opened 2022-08-23
early early-exit desired state diff performance

geoberle:
the current way to find out if the desired state of both bundles of an early exit check uses too much resources because it tries to ignore the order of elements in a list. ignoring the order is not really necessary (or might even be hurtful in some situations). this PR disables this feature.

tests showed the increased performance while not yielding false differences between the states.

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -476,7 +476,7 @@ def early_exit_integration(
     # compare
     from deepdiff import DeepDiff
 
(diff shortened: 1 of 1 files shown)


chassing (approved):
LGTM

patjlm (approved):
Nice catch. Not sure I understand the title of the PR, but LGTM :)

geoberle:
> Nice catch. Not sure I understand the title of the PR, but LGTM :)

the title happened before the first coffee :D

PR #2733, opened 2022-08-23
terraform-resource: enhance aws asg ami image selection

chassing:
design doc: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/terraform-asg-ami-selection.md
needs: https://github.com/app-sre/qontract-schemas/pull/226

APPSRE-6126

### DIFF
--- a/reconcile/test/test_utils_terrascript_aws_client.py
+++ b/reconcile/test/test_utils_terrascript_aws_client.py
@@ -1,4 +1,5 @@
 import pytest
+from reconcile.utils.aws_api import AmiTag
 
 import reconcile.utils.terrascript_aws_client as tsclient
 from reconcile.utils.external_resource_spec import (
@@ -52,8 +53,8 @@ def test_use_previous_image_id_no_upstream(ts):
     assert ts._use_previous_image_id([]) is False
 
 
-def test_use_previous_image_id_false(mocker, ts):
-    result = False
+@pytest.mark.parametrize("result", [True, False])
+def test_use_previous_image_id(mocker, ts, result):
     mocker.patch(
         "reconcile.utils.terrascript_aws_client.TerrascriptClient.init_jenkins",
         return_value=MockJenkinsApi(result),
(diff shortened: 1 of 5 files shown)


reconcile/utils/aws_api.py
```diff
@@ -1358,7 +1358,7 @@ def upsert_route53_record(self, account_name, zone_id, recordset):
             logging.error(f"[{account_name}] unhandled exception: {e}")
 
     def get_image_id(
```
janboll, line 1360: Can you please add tests for this method?

reconcile/utils/terrascript_aws_client.py
```diff
@@ -4514,32 +4513,37 @@ def _get_commit_sha(self, repo_info: Mapping) -> str:
         return ""
 
     def _get_asg_image_id(
-        self, image: Mapping, account: str, region: str
-    ) -> Tuple[Optional[str], str]:
+        self, filters: Iterable[dict[str, Any]], account: str, region: str
```
janboll, line 4516: I miss this in my PRs as well, but our guidelines state to use abstract types in favor of concrete ones: https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#abstract-collection-types

janboll (changes requested):

janboll (approved):

reconcile/utils/terrascript_aws_client.py
```diff
@@ -4514,32 +4513,37 @@ def _get_commit_sha(self, repo_info: Mapping) -> str:
...
-        its value need to be the commit sha comes from upstream repo.
+        AMI needs to be taged either with:
+          * a tag_name and its value need to be the commit sha comes from upstream repo.
+          * tag_name and value
         """
-        commit_sha = self._get_commit_sha(image)
-        tag_name = image["tag_name"]
+        tags: list[AmiTag] = []
+        for f in filters:
+            if f["provider"] == "git":
```
rporres, line 4527: It may be a bit too protective, but in our current model, does it make sense that we have two `git` providers? Would that break things? Should we limit it somehow?

reconcile/test/test_utils_terrascript_aws_client.py
```diff
@@ -49,27 +50,79 @@ def is_job_running(self, name):
...
         return_value=MockJenkinsApi(result),
     )
-    image = {"upstream": {"instance": {"name": "ci"}, "name": "job"}}
+    image = [{"upstream": {"instance": {"name": "ci"}, "name": "job"}}]
     assert ts._use_previous_image_id(image) == result
 
 
-def test_use_previous_image_id_true(mocker, ts):
-    result = True
+def test_get_asg_image_id(mocker, ts: tsclient.TerrascriptClient):
```
rporres, line 66: hmm, if you're writing tests for it, should `_get_asg_image_id` be public?

reconcile/test/test_utils_terrascript_aws_client.py
```diff
@@ -49,27 +50,79 @@ def is_job_running(self, name):
...
+    awsapi = mocker.patch("reconcile.utils.terrascript_aws_client.AWSApi")
+    get_image_id_mock = awsapi().get_image_id
+    get_image_id_mock.return_value = "ami-123456"
+
+    ref = "sha-12345"
     mocker.patch(
-        "reconcile.utils.terrascript_aws_client.TerrascriptClient.init_jenkins",
-        return_value=MockJenkinsApi(result),
+        "reconcile.utils.terrascript_aws_client.TerrascriptClient._get_commit_sha",
```
rporres, line 73: not a fan of mocking private methods :(

reconcile/test/test_utils_terrascript_aws_client.py
```diff
@@ -49,27 +50,79 @@ def is_job_running(self, name):
...
+    ],
+)
+def test_get_commit_sha(mocker, ts: tsclient.TerrascriptClient, repo_info, expected):
+    init_github = mocker.patch(
+        "reconcile.utils.terrascript_aws_client.TerrascriptClient.init_github"
+    )
+    init_github.return_value.get_repo.return_value.get_commit.return_value.sha = (
+        expected
+    )
+    assert ts._get_commit_sha(repo_info) == expected
```
rporres, line 125: same thing about public/private methods

rporres (commented):
The PR looks good. Some doubts about mocking/testing private methods and about limiting the number of `git` providers.

reconcile/utils/terrascript_aws_client.py
chassing: Allowing multiple git providers won't break anything but doesn't make absolute sense.

chassing: I don't see a need to limit it

reconcile/test/test_utils_terrascript_aws_client.py
chassing: Why should it be public then? Currently, `_get_asg_image_id` is just used internally, making sense to be private.

reconcile/test/test_utils_terrascript_aws_client.py
chassing: I'm never a big fan of mocking at all, but I don't want to test `_get_commit_sha` implicitly. Do you have an idea for a better approach?

reconcile/test/test_utils_terrascript_aws_client.py
chassing: same as above ;)

reconcile/test/test_utils_terrascript_aws_client.py
rporres: I understand that this is an internal method but I'm usually under the assumption that private methods are implementation details and there's no expectation of how they are going to behave exactly and they can be refactored without any fear. The moment you create a test, you create an expectation of it staying in a certain way. Hence my point.

Maybe others will disagree so I'm happy to accept this as it is if people don't mind it. I won't die on this hill 😂

reconcile/utils/terrascript_aws_client.py
rporres: fair enough.

reconcile/test/test_utils_terrascript_aws_client.py
rporres: make it public :-D

rporres: Or maybe mocking `github.get_repo.get_commit`?

reconcile/test/test_utils_terrascript_aws_client.py
chassing: I'll make it public, and you can sleep in peace :)

BumbleFeng (approved):

rporres (approved):

PR #2734, opened 2022-08-23
openshift-resource/route/prometheus-rule-tester resource resolution and early-exit performance

geoberle:
this PR refactors the `openshift-resources`, `openshift-route` and `prometheus-rules-tester` to work with qontract-server resource resolution (instead of sequential fetching). the `openshift-resources` early-exit code has been refactored so it can be reused by the other mentioned integrations as well.

at the same time, general performance issues for early-exit have been adressed by:
* not parsing resource YAML during early exit - this is not required because the content of a resource is sufficient to drive a decision for early-exit
* jinja2 template compilation caching

(!) review commit by commit

depends on https://github.com/app-sre/qontract-schemas/pull/227
relates to relates to https://issues.redhat.com/browse/APPSRE-6098

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -871,6 +871,10 @@ def early_exit_desired_state(
         (r, ns_info) for ns_info in namespaces for r in ns_info["openshiftResources"]
     ]
 
+    # this context manager patches functions used during jinja templating
+    # to ignore data that is not part of the desired state in app-interface.
+    # the context manager also ensures this function patching is
+    # reverted afterwards
     with _early_exit_monkey_patch():
         resources = threaded.run(
             _early_exit_fetch_resource,
@@ -889,10 +893,16 @@ def _early_exit_fetch_resource(spec, settings):
     resource = spec[0]
     ns_info = spec[1]
     if resource.get("enable_query_support"):
+        # use the regular resource processing functionality that evaluates templates
+        # and inline queries, if the resource is allowed to use this inline query
+        # functionality. this is crucial in such situations because the result of
+        # the template processing depends heavily on other data in app-interface
         c = fetch_openshift_resource(
             resource, ns_info, skip_validation=True, settings=settings
         ).body
     else:
+        # for regular resources, the plain content is sufficient enough to
+        # detect changes in desired state
         c = resource["resource"].get("content")
     del resource["resource"]
     return c
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -1,6 +1,6 @@
 import base64
 from contextlib import contextmanager
-from functools import lru_cache
+from functools import cache
 import json
 import logging
 import sys
@@ -248,7 +248,7 @@ def lookup_graphql_query_results(query: str, **kwargs) -> list[Any]:
     return results
 
 
-@lru_cache
+@cache
 def compile_jinja2_template(body, extra_curly: bool = False):
     env: dict = {}
     if extra_curly:
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -322,12 +322,22 @@ def fetch_provider_resource(
     validate_alertmanager_config=False,
     alertmanager_config_key="alertmanager.yaml",
     add_path_to_prom_rules=True,
+    skip_validation=False,
     settings=None,
 ) -> OR:
     path = resource["path"]
     content = resource["content"]
     if tfunc:
         content = tfunc(body=content, vars=tvars, settings=settings)
+
+    if skip_validation:
+        return OR(
+            content,
+            QONTRACT_INTEGRATION,
+            QONTRACT_INTEGRATION_VERSION,
+            error_details=path,
+            validate_k8s_object=False,
+        )
 
     try:
         body = anymarkup.parse(content, force_types=None)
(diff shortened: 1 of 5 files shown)


reconcile/openshift_resources_base.py
```diff
@@ -234,11 +248,30 @@ def lookup_graphql_query_results(query: str, **kwargs) -> list[Any]:
...
+
+    jinja_env = jinja2.Environment(
+        extensions=[B64EncodeExtension, RaiseErrorExtension],
+        undefined=jinja2.StrictUndefined,
+        **env,
+    )
+    return jinja_env.from_string(body)
+
+
+def process_jinja2_template(body, vars=None, extra_curl: bool = False, settings=None):
```
rporres, line 272:
```suggestion
def process_jinja2_template(body, vars=None, extra_curly: bool = False, settings=None):
```

reconcile/openshift_resources_base.py
```diff
@@ -256,12 +289,7 @@ def process_jinja2_template(body, vars=None, env=None, settings=None):
     vars.update({"query": lookup_graphql_query_results})
     vars.update({"url": url_makes_sense})
     try:
-        env = jinja2.Environment(
-            extensions=[B64EncodeExtension, RaiseErrorExtension],
-            undefined=jinja2.StrictUndefined,
-            **env,
-        )
-        template = env.from_string(body)
+        template = compile_jinja2_template(body, extra_curl)
```
rporres, line 292:
```suggestion
        template = compile_jinja2_template(body, extra_curly)
```

reconcile/openshift_resources_base.py
```diff
@@ -234,11 +248,30 @@ def lookup_graphql_query_results(query: str, **kwargs) -> list[Any]:
     return results
 
 
-def process_jinja2_template(body, vars=None, env=None, settings=None):
+@lru_cache
```
rporres, line 251: I love this idea. Have you tried with `@cache`? Since we have already a lot of things in memory, we may want to add a few more with an unbound cache 😬

reconcile/openshift_resources_base.py
```diff
@@ -817,3 +858,79 @@ def run(
         sys.exit(1)
 
     return ri
+
+
+def early_exit_desired_state(
```
rporres, line 863: This function is a complicated beast. I think it is a good moment to add a comment to guide the new reader of this.

reconcile/openshift_resources_base.py
```diff
@@ -817,3 +858,79 @@ def run(
...
+def early_exit_monkey_patch_assign(
+    lookup_secret,
+    lookup_github_file_content,
+    url_makes_sense,
+    check_alertmanager_config,
+):
+    sys.modules[__name__].lookup_secret = lookup_secret
+    sys.modules[__name__].lookup_github_file_content = lookup_github_file_content
+    sys.modules[__name__].url_makes_sense = url_makes_sense
+    sys.modules[__name__].check_alertmanager_config = check_alertmanager_config
```
rporres, line 936: why this change?

reconcile/openshift_resources_base.py
```diff
@@ -769,6 +809,7 @@ def get_namespaces(
     providers: Optional[list[str]] = None,
     cluster_name: Optional[str] = None,
     namespace_name: Optional[str] = None,
+    resource_schema_filter: Optional[str] = None,
```
rporres, line 812: I love this <3

reconcile/openshift_resources_base.py
geoberle: the original code looked like
```
orb.lookup_secret = lookup_secret
```
since we are now in the `orb` module itself, it needed to be rewritten. so i tried
```
lookup_secret = patched_lookup_secret
```
but then vscode syntax highlighting was broken. it did not show, that `lookup_secret` would be the module function that is currently being overwritten. this might be just me but i rely heavily on syntax and color highlighting to get a feeling what is going on.

so i decided to be very explicit and that is how i landed with 

```
sys.modules[__name__].lookup_secret = lookup_secret
```

i'm open for suggestions

reconcile/openshift_resources_base.py
geoberle: fixed

reconcile/openshift_resources_base.py
geoberle: fixed

reconcile/openshift_resources_base.py
geoberle: fixed

reconcile/openshift_resources_base.py
geoberle: i did not consider an unbounded cache until now. thinking about this, i'm not too much concerned about this cache filling up uncontrolled. we don't have that many template changes that we would be hit by this. we restart reconciler way more often than we change templates.

additionally, cache eviction with lru_cache might cost performance gains. so i'm in favour of your proposal!

geoberle: changed in [7a13b0d](https://github.com/app-sre/qontract-reconcile/pull/2734/commits/7a13b0d87d07f4d90205529e58852b580805d24a)

reconcile/openshift_resources_base.py
rporres: Fair enough :)

reconcile/openshift_resources_base.py
geoberle: added comments in [c22bed7](https://github.com/app-sre/qontract-reconcile/pull/2734/commits/c22bed7b258faa8f4afd65c148e1d8377a64bc8a)

rporres: wunderbar! ❤️

rporres (approved):

PR #2735, opened 2022-08-23
[APPSRE-6069] add secretsmanager for es

BumbleFeng:
store es creds in secrets manager for lambda.

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -3935,8 +3935,10 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
         es_values["access_policies"] = json.dumps(access_policies, sort_keys=True)
 
         region = values.get("region") or self.default_regions.get(account)
+        provider = ""
         if self._multiregion_account(account):
-            es_values["provider"] = "aws." + region
+            provider = "aws." + region
+            es_values["provider"] = provider
 
         auth_options = values.get("auth", {})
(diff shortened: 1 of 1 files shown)


esron (approved):

reconcile/utils/terrascript_aws_client.py
```diff
@@ -3994,6 +3996,51 @@ def populate_tf_resource_elasticsearch(self, spec):
             "internal_user_database_enabled", False
         ):
             master_user = security_options["master_user_options"]
+            secret_values = {"name": identifier}
```
fishi0x01, line 3999: I assume the `name` must be unique across the region? Would it make sense to add a prefix for the integration to avoid potential collisions with future identifiers from other integrations? 

```python
secret_values = {"name": f"qr-es_{identifier}"}
```

reconcile/utils/terrascript_aws_client.py
```diff
@@ -3994,6 +3996,51 @@ def populate_tf_resource_elasticsearch(self, spec):
...
+            if provider:
+                secret_values["provider"] = provider
+            aws_secret_resource = aws_secretsmanager_secret(identifier, **secret_values)
```
fishi0x01, line 4002: Similar concern as above: tf secret resources can be created by multiple integrations within the same account, i.e., there is a chance identifiers collide w/o prefixing. But maybe in this case this concern is too far fetched.

fishi0x01 (commented):
Nice enhancement! :tada: 
Some small comments below

PR #2736, opened 2022-08-24
Revert "openshift-resource/route/prometheus-rule-tester resource resolution and early-exit performance"

geoberle:
Reverts app-sre/qontract-reconcile#2734
detected a bug during rollout. reverting for now


chassing (approved):
LGTM

PR #2738, opened 2022-08-25
[template-tester][qontract-cli alert-to-receiver] adapt to resource resolution

geoberle:
adapt template tester and qontract-cli alert-to-receiver to cope with https://github.com/app-sre/qontract-schemas/pull/229


chassing (approved):
LGTM

PR #2568, opened 2022-07-11
[WIP] APPSRE-5972 support fragment queries

fishi0x01:
This is an example of how to leverage [GQL fragments ](https://graphql.org/learn/queries/#fragments) to reduce the amount of generated classes.

Fragments are stored in a dedicated module, e.g., `reconcile/gql_fragments`.

Fragments are then consumed in different queries in `reconcile/gql_queries`.

Current limitations:

- every `.gql` file in `gql_queries` must contain exactly one `query` definition
- every `.gql` file in `gql_fragments` must contain exactly one `fragment` definition
- fragments cannot consume other fragments. E.g., this is not possible:

```graphql
fragment MyHero on Hero {
  name
}

fragment MyEpisode on Episode {
  hero {
    ... MyHero
  }
  name
}
```

Currently, you would need to re-write the previous to:

```graphql
fragment MyEpisode on Episode {
  hero {
    name
  }
  name
}
```

PR #2739, opened 2022-08-25
APPSRE-5972 initial fragment support

fishi0x01:
## Changes

- add convenience function to query and parse data with generated classes
- rename `reconcile/gql_queries` -> `reconcile/gql_definitions`, because it now also holds fragments
- add examples on how [GQL fragments ](https://graphql.org/learn/queries/#fragments) can be used

This PR may look daunting, but vast majority of changes are in generated files. The QR logic is rarely touched.

## Explanation

GQL fragments  can be used to reduce the amount of generated classes. Examples on how fragments can be used are added in `reconcile/gql_definitions/examples`

Current limitations:

- every .gql file must contain exactly one definition (`fragment` or `query`)
- fragments within an inline fragments block are undefined
- fragments cannot consume other fragments (nesting). E.g., this is not possible:

```graphql
fragment MyHero on Hero {
  name
}

fragment MyEpisode on Episode {
  hero {
    ... MyHero
  }
  name
}
```

Supporting nested fragments will be subject to a future version of qenerate.

Currently, you would need to re-write the previous to:

```graphql
fragment MyEpisode on Episode {
  hero {
    name
  }
  name
}
```

This PR will obsolete previous PoC https://github.com/app-sre/qontract-reconcile/pull/2568

## Validation

Executed affected integrations locally in dry-run mode


chassing (approved):
Awesome. LGTM

PR #2740, opened 2022-08-25
update es log group name

BumbleFeng:
AWS suggests using a path kind name to manage cloudwatch log group.

WX20220825-160847@2x


fishi0x01 (approved):
Nice! It looks like `publish_log_types` is not used yet in app-interface, so this change should not be a problem :+1:

PR #2744, opened 2022-08-26
fix pylint missing-timeout

BumbleFeng:
add in https://github.com/PyCQA/pylint/releases/tag/v2.15.0

### DIFF
--- a/reconcile/checkpoint.py
+++ b/reconcile/checkpoint.py
@@ -42,7 +42,7 @@ def url_makes_sense(url: str) -> bool:
     if not url:
         return False
     try:
-        rs = requests.get(url, verify=False)
+        rs = requests.get(url, verify=False, timeout=60)
     except requests.exceptions.ConnectionError:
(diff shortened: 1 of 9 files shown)


steveteahan:
Are there many instances of this? I'd prefer to ignore inline if not because this is a very common cause of issues. The default is no timeout which can lead to calls hanging for a very long time.

BumbleFeng:
about 30 instances. Do you have a suggestion for timeout value? I can just fix them all.

steveteahan:
I usually default to something like 60 seconds unless there is a good reason to have a lower value. Fixing these and keeping the check disabled is best if you have the time. If not, we can disable it per file by adding `# pylint: disable=missing-timeout` at the top of the file. Both would result in a much better chance of catching this issue going forward for new code, rather than having to disable at the project level.

Thanks for doing this!

steveteahan (approved):
Thanks again!

PR #2745, opened 2022-08-29
[gitlab-housekeeper] non-rebasable MRs blocking merge queue

geoberle:
If there is a MR in front of the app-interface merge queue, which has a merge conflict and can't be rebased or merged, it will still count into the rebase attempts of gitlab-housekeeper and slows down or even blocks the entire MR queue.

this change skips such MRs and logs appropriately

Fixes https://issues.redhat.com/browse/APPSRE-6201

Signed-off-by: Gerd Oberlechner <goberlec@redhat.com>

PR #2747, opened 2022-08-29
gitlab-housekeeper fix check for unmergable MRs

geoberle:
the previous attempt to check for unmergable MRs failed because the attribute `merge_error` is not present all the time in MR objects. additionally the place to check is the `get_merge_request` function.

this change removes the check for `merge_error` in favour of the `merge_status` check and adds a check for the gitlab status `cannot_be_merged_recheck` (a.k.a. last time i checked it was not mergable but currently checking)

fixes https://github.com/app-sre/qontract-reconcile/pull/2745


patjlm (approved):
LGTM

PR #2750, opened 2022-08-30
Split DNS support by separating zone name and ID

cubismod:
[APPSRE-6143](https://issues.redhat.com/browse/APPSRE-6143)

Schemas update here: https://github.com/app-sre/qontract-schemas/pull/232

Optionally split the Terraform/App Interface identifier for a Route 53 zone from the name of the zone as defined in App Interface. With this change we should be able to define both a private and public DNS zone within the same AWS account that refer to the same domain name.

This will allow us to do [split view DNS](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/hosted-zone-private-considerations.html#hosted-zone-private-considerations-split-view-dns) within the Fedramp environment for our work to get certificates working.

### DIFF
--- a/reconcile/terraform_aws_route53.py
+++ b/reconcile/terraform_aws_route53.py
@@ -43,14 +43,14 @@ def build_desired_state(
         domain_name = zone.get("domain_name")
         if domain_name:
             zone_name = domain_name
-            zone_id = zone["name"]
+            resource_name = zone["name"]
         else:
             zone_name = zone["name"]
-            zone_id = zone["name"]
+            resource_name = zone["name"]
 
         zone_values = {
             "name": zone_name,
-            "id": zone_id,
+            "resource_name": resource_name,
             "account_name": account_name,
             "records": [],
         }
--- a/reconcile/terraform_aws_route53.py
+++ b/reconcile/terraform_aws_route53.py
@@ -37,8 +37,23 @@ def build_desired_state(
         account = zone["account"]
         account_name = account["name"]
 
-        zone_name = zone["name"]
-        zone_values = {"name": zone_name, "account_name": account_name, "records": []}
+        # optionally decouple the name of the DNS file from the domain
+        # it is creating records for to allow split view DNS
+        # https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/hosted-zone-private-considerations.html#hosted-zone-private-considerations-split-view-dns
+        domain_name = zone.get("domain_name")
+        if domain_name:
+            zone_name = domain_name
+            zone_id = zone["name"]
+        else:
+            zone_name = zone["name"]
+            zone_id = zone["name"]
+
+        zone_values = {
+            "name": zone_name,
+            "id": zone_id,
+            "account_name": account_name,
+            "records": [],
+        }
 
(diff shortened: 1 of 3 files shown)


reconcile/terraform_aws_route53.py
```diff
@@ -37,8 +37,23 @@ def build_desired_state(
...
+        # optionally decouple the name of the DNS file from the domain
+        # it is creating records for to allow split view DNS
+        # https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/hosted-zone-private-considerations.html#hosted-zone-private-considerations-split-view-dns
+        domain_name = zone.get("domain_name")
+        if domain_name:
+            zone_name = domain_name
+            zone_id = zone["name"]
+        else:
+            zone_name = zone["name"]
+            zone_id = zone["name"]
```
bkez322, line 49: To avoid continued confusion with the lack of actual zone-ids being used, could we use variable name `resource_name` instead of `zone_id`?

reconcile/terraform_aws_route53.py
```diff
@@ -37,8 +37,23 @@ def build_desired_state(
...
+
+        zone_values = {
+            "name": zone_name,
+            "id": zone_id,
```
bkez322, line 53: And then here we could have

`"resource_name": resource_name,`

reconcile/utils/terrascript_aws_client.py
```diff
@@ -781,7 +781,9 @@ def populate_route53(self, desired_state, default_ttl=300):
             acct_name = zone["account_name"]
 
             # Ensure zone is in the state for the given account
-            zone_id = safe_resource_id(f"{zone['name']}")
+            # zone_id refers to the terraform identifier, not the AWS
+            # route 53 zone ID that is generated after the zone is created
+            zone_id = safe_resource_id(f"{zone['id']}")
```
bkez322, line 786:
```suggestion
            zone_res_name = safe_resource_id(f"{zone['resource_name']}")
```

And then you could probably remove all 3 comments above here (i think the original one is something leftover?)

reconcile/utils/terrascript_aws_client.py
```diff
@@ -781,7 +781,9 @@ def populate_route53(self, desired_state, default_ttl=300):
...
             zone_values = {
                 "name": zone["name"],
                 "vpc": zone.get("vpc"),
```
bkez322, line 789: And then on line `792` (I'm having trouble selecting it, replace `zone_id` with `zone_res_name`.

bkez322 (commented):
I would like it if we could remove the misleading usage of the `zone_id` variable name entirely, but conceptually this LGTM.

reconcile/terraform_aws_route53.py
cubismod: That sounds good yeah, I agree that `id` is being used wayyyy too much here

bkez322 (approved):
LGTM, thanks for the variable name change 🎉

cubismod:
Just to be super sure, I checked these QR changes against AI commercial in my local env and there were no terraform changes planned with the update as I expected.

PR #2751, opened 2022-08-31
consider templated secrets stringData when checking valid keys

patjlm:
If the Secret is embedded in the openshift template, it may contain keys in `stringData` instead of `data`.


chassing (approved):
LGTM

PR #2754, opened 2022-09-01
Stop unleash threads after test

janboll:
UnleashClient creates background threads during test. Calling destroy() will shut down these threads.


chassing:
TBH I don't like code that is just there for testing purposes. Looks like the global variable, which is also a bad smell, bits you somehow :)

janboll:
:100: agreed, but I don't see this in the scope of this PR.

chassing (approved):
LGTM

PR #2755, opened 2022-09-01
jira_client: fix type error

jfchevrette:
Fixes a typing error where the variable type did not match the method's return type, causing mypy to complain. we can let it infer from the method return type

```
reconcile/utils/jira_client.py:44: error: Incompatible types in assignment (expression has type "Union[List[Dict[str, Any]], ResultList[Issue]]", variable has type "Union[Dict[str, Any], ResultList[Issue]]")  [assignment]
```

Because we check that the variable is of instance ResultList this was never an issue other than a type mismatch

### DIFF
--- a/reconcile/utils/jira_client.py
+++ b/reconcile/utils/jira_client.py
@@ -5,7 +5,7 @@ from jira.client import ResultList
 
 from reconcile.utils.secret_reader import SecretReader
 
-from typing import Any, Iterable, Mapping, Optional, Union
+from typing import Any, Iterable, Mapping, Optional
 
 
 class JiraClient:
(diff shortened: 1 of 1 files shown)


reconcile/utils/jira_client.py
```diff
@@ -41,9 +41,7 @@ def get_issues(self, fields: Optional[Mapping] = None) -> list[Issue]:
             kwargs["fields"] = ",".join(fields)
         while True:
             index = block_num * block_size
-            issues: Union[dict[str, Any], ResultList[Issue]] = self.jira.search_issues(
-                jql, index, block_size, **kwargs
-            )
+            issues = self.jira.search_issues(jql, index, block_size, **kwargs)
```
janboll, line 44: That's pretty confusing error message, since the type is just copied over from search_issue. 
Wonder if there is some special case, where only a dict is returned? :confused:

janboll: If I read the error correctly, there are times where just a dict is returned, couldn't that break `all_issues.extend(issues)` ?

janboll: Okay, got it now. Think we should be safe then. 
What about bumping the dependency?

janboll (approved):

PR #2757, opened 2022-09-02
Fix deep sync mode

rporres:
Compare tags (previously known as deep-sync) control file was updated
at the beginning of the run. Hence, even if the run was unsuccesful,
the next one would have seen that the previous execution of the
integration was a compare-tags one and would run in non-compare-tags
mode, which would mask any potential issue with the slow mode. This
patch fixes by changing the moment of the update of the control file to
the end of the execution of the integration.

In order to be able to properly test these changes without needing to
modify the source code, a few options have been added to the
integration to control:

* where the control file is created.
* how often to compare tags.
* to force the comparation of tags.
* to specify a certain image.

The first one is particularly interesting as it can allow us to create
the control file outside the container, which can allow us to avoid a
compare tags run every time the pod is restarted/recreated.

part of APPSRE-6102

### DIFF
--- a/reconcile/quay_mirror.py
+++ b/reconcile/quay_mirror.py
@@ -6,7 +6,7 @@ import tempfile
 import time
 
 from collections import defaultdict, namedtuple
-from typing import Any
+from typing import Any, Iterable, Optional
 
 from sretoolbox.container.image import ImageComparisonError, ImageContainsError
 from sretoolbox.container.skopeo import SkopeoCmdError
@@ -24,6 +24,7 @@ from reconcile.utils.instrumented_wrappers import (
 _LOG = logging.getLogger(__name__)
 
 QONTRACT_INTEGRATION = "quay-mirror"
+CONTROL_FILE_NAME = "qontract-reconcile-quay-mirror.timestamp"
 
 OrgKey = namedtuple("OrgKey", ["instance", "org_name"])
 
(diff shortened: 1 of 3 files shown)


reconcile/quay_mirror.py
```diff
@@ -54,15 +54,41 @@ class QuayMirror:
         shard_id=sharding.SHARD_ID,
     )
 
-    def __init__(self, dry_run=False):
+    def __init__(
+        self,
+        dry_run: bool = False,
+        control_file_dir: Optional[str] = None,
+        force_compare_tags: bool = False,
+        compare_tags_interval: int = 86400,
```
janboll, line 62: Aren't defaults provided via click?

reconcile/quay_mirror.py
```diff
@@ -261,26 +281,21 @@ def process_sync_tasks(self):
 
         return sync_tasks
 
-    def _is_deep_sync(self, interval):
-        control_file_name = "qontract-reconcile-quay-mirror.timestamp"
-        control_file_path = os.path.join(tempfile.gettempdir(), control_file_name)
+    def _check_compare_tags_elapsed_time(self, interval: int) -> bool:
```
janboll, line 284: Isn't that a good candidate for a unit test? :)

rporres: heheh, yes, it is... I will add them :)

reconcile/quay_mirror.py
```diff
@@ -54,15 +54,41 @@ class QuayMirror:
...
+            self.control_file_path = os.path.join(control_file_dir, control_file_name)
+        else:
+            self.control_file_path = os.path.join(
+                tempfile.gettempdir(), control_file_name
+            )
+
+        self._is_compare_tags = (
+            force_compare_tags
+            or self._check_compare_tags_elapsed_time(compare_tags_interval)
+        )
```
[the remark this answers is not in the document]
rporres: I added it in the constructor as it is used in two places:

* When deciding if comparing tasks in `process_sync_tasks`
* When updating the control a file at the end of `run`

But I agree with the whole idea of avoiding side effects in the constructor, so what I will is to create a new accessor `is_deep_sync` that caches the result.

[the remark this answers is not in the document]
rporres: done

reconcile/quay_mirror.py
rporres: fixed

reconcile/quay_mirror.py
rporres: fixed

rporres:
@janboll I've addressed your comments. I accidentaly rebased the whole thing again in the middle of your review, apologies :(

reconcile/quay_mirror.py
janboll: Just realized reading your comment again, will removing the defaults from Click make this harder to use? Or asked in another way, is Click now complaining if I do not specify the compare_tags_interval?

rporres: Click doesn't complain but if I don't specify defaults, it will pass `None` explicitly, the defaults in the signature won't be considered and the program will fail. So I needed to put them back again in db05bd38

janboll (approved):
:rocket:

PR #2760, opened 2022-09-04
[terraform-users] add instructions to enable MFA in email invitation

maorfr:
this will prevent many repeating questions on MFA, thus reducing IC toil.


patjlm (approved):
LGTM

PR #2767, opened 2022-09-06
Sync changes to quay-mirror quay-mirror org

rporres:
quay-mirror-org is quite similar but subtly different to quay-mirror.
It was written in a rush and some parts were taken and other don't.
I've tried to refactor the whole thing so that we can have something
like QuayMirrorBase and QuayMirror and QuayMirrorOrg inheriting, but I
started to get into a trap that is totally out of the scope of the work
I'm doing, so I'll leave it to a subsequent effort.

This patch brings the fixes and new features of quay-mirror to
quay-mirror-org acknowledging the fact that some of the code has been
copied verbatim.

related to APPSRE-6102 and APPSRE-6101.


rporres:
The jira ticket to refactor quay-mirror-org properly: https://issues.redhat.com/browse/APPSRE-6251

PR #2768, opened 2022-09-07
[dyn-traffic-director] early exit

maorfr:
since this integration has a tendency to fail from time to time and potentially blocking MRs, implementing early exit to prevent it from running in the first place.


geoberle (approved):
❤️

PR #2770, opened 2022-09-08
introduce saas file deprecation

maorfr:
related to https://issues.redhat.com/browse/APPSRE-6021

depends on https://github.com/app-sre/qontract-schemas/pull/236

with this PR, we can mark a saas file as `deprecated`. as a result, when MRs are submitted against that saas file, instead of returning the list of approvers, we return a message asking to migrate resources away from this saas file (saas-file-owners).
at the same time, we can choose to `/lgtm` to let things through anyway.

in addition, resources from deprecated saas files will be considered up for grabs (openshift-saas-deploy) by not passing the name of the saas file to the `all_callers` list:
https://github.com/app-sre/qontract-reconcile/blob/e7e025662ff8c6e6fca4ef6efd332646fd507189/reconcile/openshift_base.py#L546-L556


reconcile/saas_file_owners.py
```diff
@@ -377,10 +378,16 @@ def run(
...
             gl.remove_label_from_merge_request(gitlab_merge_request_id, APPROVED)
-            comment_line_body = (
-                f"- changes to saas file '{saas_file_name}' "
-                + f"require approval (`/lgtm`) from one of: {saas_file_owners}."
-            )
+            if diff["deprecated"]:
+                comment_line_body = (
+                    f"- saas file '{saas_file_name}' is deprecated. "
+                    + "please migrate resources to a different saas file."
+                )
```
chassing, line 385: I would like to see the '/lgtm' message as well, maybe something like that:

```
- changes to saas file '{saas_file_name}' ...
- [bold red]saas file 'XXX' is deprecated. Please migrate resources to a different saas file.[/]
```

chassing: otherwise is hard to see, that we can still `/lgtm` that

maorfr: if the owners of the saas file decided it's deprecated, why should their names be shown? :sweat_smile:

maorfr: removing this part from this PR

chassing (approved):
LGTM

PR #2775, opened 2022-09-09
[qontract-cli] Make template usable by non app-sre

rporres:
Added local secret config support and a way to filter results by path
that will avoid the need to render templates to inspect their names.
This will make calling Vault not needed for many templates, easying the
use for tenants and making it considerably faster for everyone.

The new option --path may seem redundant but it has benefits and we are
keeping backwards compatibility just in case.

Signed-off-by: Rafa Porres Molina rporresm@redhat.com

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -1314,8 +1314,20 @@ def rm(ctx, integration, key):
 @click.argument("namespace")
 @click.argument("kind")
 @click.argument("name")
+@click.option(
+    "-p",
+    "--path",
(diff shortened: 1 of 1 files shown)


rporres:
The other approach would be to forget about backwards compatibility and move `kind` and `name` to be filters too. We would only then match the resources that match the filters given. I prefer that approach, honestly. I don't feel that backwards compatibility is that important for this command, but maybe I'm wrong 😅

maorfr (approved):

PR #2776, opened 2022-09-11
add option to template resources based on qontract-cli commands output

maorfr:
we have a lot of useful information gathered as qontract-cli commands and displayed in app-interface-output.
it could probably be useful to use such commands to template resources.

### DIFF
--- a/reconcile/utils/output.py
+++ b/reconcile/utils/output.py
@@ -26,6 +26,8 @@ def print_output(
         print(json.dumps(content))
     elif output == "yaml":
         print(yaml.dump(content))
+    elif output == "return":
+        return content
     else:
         pass  # error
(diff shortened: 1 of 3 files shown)


rporres:
Do you have any use-case in mind so that this is easier to understand?

reconcile/utils/output.py
```diff
@@ -26,6 +26,8 @@ def print_output(
         print(json.dumps(content))
     elif output == "yaml":
         print(yaml.dump(content))
+    elif output == "return":
+        return content
```
rporres, line 30: I don't think a function called `print_output` should either print or return stuff based on arguments... you may want to create a separate function that returns content used by this one...

PR #2777, opened 2022-09-11
[change-owners] faster diffs and better results with context identifiers

geoberle:
force deepdiff to leverage context identifiers from app-interface bundles to speed up diff calculation and yield better results. deepdiff usually can heuristically decide when to use context identifiers. with `cutoff_intersection_for_pairs=1` we can force it to use them. this is a valid approach in our case since our data follows a schema and context identifiers are schema enforced, e.g. guaranteeing uniqueness within the context.

### Example

```
resourceTemplates:
- name: tmpl-1
  url: url
  path: path
  targets:
  - namespace: xxx
     ref: ref
     ...
  - namespace: yyy
     ref: new-value   <----
     ....
- name: tmpl-2
  url: url
  path: path
  targets:
  - namespace: xxx
     ref: ref
     ....
  - namespace: yyy
     ref: new-value   <----
    ......
```

calculating proper diffs in saas-file-2 file where the ref of two targets from different resourceTemplates changed, requires 73 diff iterations and detects 1193 intermediary diffs before settling on the final results. this is due to nested lists which can look very similar in the case of saas files. search complexity can be reduced by accepting a lower quality of diffs, e.g. the path of the diff would be something like `resourceTemplates[0]`, indicating that the difference is "somewhere" within the template. not very helpful.

with context identifiers, proper diff paths can be achieved while keeping search complexity low. with 13 diff iterations and 161 intermediary diffs, it can pinpoint the changes to something that looks like`resourceTemplates.[0].targets.[1].ref` and `resourceTemplates.[1].targets.[1].ref` while being a lot faster!

### References
https://github.com/app-sre/qontract-schemas/pull/237

### DIFF
--- a/reconcile/change_owners.py
+++ b/reconcile/change_owners.py
@@ -22,6 +22,7 @@ from reconcile.utils.semver_helper import make_semver
 
 from deepdiff import DeepDiff
 from deepdiff.helper import CannotCompare
+from deepdiff.model import DiffLevel
 
 import jsonpath_ng
 import jsonpath_ng.ext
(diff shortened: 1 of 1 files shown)


maorfr (approved):

patjlm:
❤️

PR #2781, opened 2022-09-12
[change-owners] deal with new or deleted files

geoberle:
if files were added or deleted entirely, diff detection would fail in the case of resourcefiles. this change makes sure this does not happen.

as a nice sideeffect, the old and new colums of a detected diff now show more reliable output in the case of added or removed dict elements.

additionally: exceptions during diff detection will not fail PR checks anymore for now, while the diff logic is still stabilizing

### DIFF
--- a/reconcile/change_owners.py
+++ b/reconcile/change_owners.py
@@ -687,9 +687,14 @@ def run(
                     "file": c.fileref.path,
                     "schema": c.fileref.schema,
                     "changed path": d.path,
-                    "old value": d.old,
-                    "new value": d.new,
                 }
+                if str(d.path) != "$":
+                    item.update(
+                        {
+                            "old value": d.old,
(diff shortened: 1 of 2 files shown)


reconcile/change_owners.py
```diff
@@ -366,7 +378,7 @@ def create_bundle_file_change(
 DEEP_DIFF_RE = re.compile(r"\['?(.*?)'?\]")
 
 
-def deepdiff_path_to_jsonpath(deep_diff_path: str) -> str:
+def deepdiff_path_to_jsonpath(deep_diff_path: str) -> jsonpath_ng.JSONPath:
```
geoberle, line 381: i am a bit puzzled why mypy did not report this

chassing (approved):
LGTM

PR #2782, opened 2022-09-12
[openshift-saas-deploy-trigger] introduce trigger spec class

maorfr:
instead of using dicts, we can use classes.

this is a refactor only, and should not change any logic.

### DIFF
--- a/reconcile/openshift_saas_deploy_trigger_base.py
+++ b/reconcile/openshift_saas_deploy_trigger_base.py
@@ -15,6 +15,7 @@ from reconcile.utils.saasherder import (
     SaasHerder,
     Providers,
     UNIQUE_SAAS_FILE_ENV_COMBO_LEN,
+    TriggerSpecUnion,
 )
 from reconcile.utils.sharding import is_in_shard
 from reconcile.utils.defer import defer
@@ -158,14 +159,14 @@ def setup(
 
 
 def trigger(
-    spec,
-    dry_run,
-    saasherder,
-    oc_map,
-    already_triggered,
-    integration,
-    integration_version,
-):
+    spec: TriggerSpecUnion,
+    dry_run: bool,
+    saasherder: SaasHerder,
+    oc_map: OC_Map,
+    already_triggered: set[str],
+    integration: str,
+    integration_version: str,
+) -> bool:
     """Trigger a deployment according to the specified pipelines provider
 
     Args:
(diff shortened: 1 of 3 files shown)


reconcile/openshift_saas_deploy_trigger_base.py
```diff
@@ -177,17 +177,14 @@ def trigger(
...
                                     It will get populated by this function.
-        trigger_type (string): Indicates which method to call to update state
         integration (string): Name of calling integration
         integration_version (string): Version of calling integration
 
     Returns:
         bool: True if there was an error, False otherwise
     """
-
-    # TODO: Convert these into a dataclass.
```
geoberle, line 188: todo done :)

reconcile/utils/saasherder.py
```diff
@@ -58,6 +69,87 @@ def __repr__(self):
...
+@dataclass
+class TriggerSpecMovingCommit(TriggerSpecBase):
+    ref: str
+    reason: Optional[str] = None
+
+    @property
+    def state_key(self):
+        key = (
+            f"{self.saas_file_name}/{self.resource_template_name}/{self.cluster_name}/"
+            + f"{self.namespace_name}/{self.env_name}/{self.ref}"
```
geoberle, line 129: nit: i don't think you need the + symbol here. strings in () will be concatenated also without the +

reconcile/utils/saasherder.py
```diff
@@ -1038,8 +1130,19 @@ def init_populate_desired_state_specs(self, saas_file):
...
+                    timeout=None,
+                    pipelines_provider=None,
+                    resource_template_name=rt_name,
+                    cluster_name=cluster,
+                    namespace_name=namespace,
+                    state_content=None,
+                )
+                digest = SaasHerder.get_target_config_hash(
+                    target_configs[dummy_job_spec.state_key]
+                )
```
geoberle, line 1145: another nit: if this dumm job spec has no other purpose than creating the state_key, i would prefer to use it without assigning it. less mental state

```suggestion
                state_key = TriggerSpecConfig(
                    saas_file_name=saas_file_name,
                    env_name=env_name,
                    timeout=None,
                    pipelines_provider=None,
                    resource_template_name=rt_name,
                    cluster_name=cluster,
                    namespace_name=namespace,
                    state_content=None,
                ).state_key
                digest = SaasHerder.get_target_config_hash(
                    target_configs[state_key]
                )
```

geoberle (approved):
only nits. nice to read!!

reconcile/utils/saasherder.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/2782/commits/9ac938b064f4294f42c51eb397b6543f5a2972f1

reconcile/utils/saasherder.py
maorfr: great idea!
https://github.com/app-sre/qontract-reconcile/pull/2782/commits/34a08a219b3ab8a5251b0caa9a4c7f4a35750c34

PR #2783, opened 2022-09-12
Vault replication integration

andreugallofre:
[APPSRE-6137](https://issues.redhat.com/browse/APPSRE-6137)
[Design doc](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/vault-replica-integration.md#proposal)
[Schema changes](https://github.com/app-sre/qontract-schemas/pull/241)


reconcile/gql_definitions/vault_instances/vault_instances.gql
```diff
@@ -0,0 +1,50 @@
...
+      policy {
+        name
+        instance {
+          name
+          address
+        }
+      }
+    }
+  }
+}
```
rporres, line 50: nit: please make sure that all files have an ending newline 🙏

reconcile/utils/vault.py
```diff
@@ -166,7 +195,8 @@ def _read_all_v2(self, path, version):
             raise SecretNotFound(path)
 
         data = secret["data"]["data"]
-        return data
+        secret_verion = secret["data"]["metadata"]["version"]
```
rporres, line 198:
```suggestion
        secret_version = secret["data"]["metadata"]["version"]
```

reconcile/utils/vault.py
```diff
@@ -166,7 +195,8 @@ def _read_all_v2(self, path, version):
...
+        return data, secret_verion
```
rporres, line 199: hmmm, you should not modify the `read_all` but use the new method...

on top of that... logic should not be duplicated. The old method should call the new one returning just the data.

reconcile/vault_replication.py
```diff
@@ -0,0 +1,176 @@
...
+                )
+
+                provider = replication.provider
+
+                if provider == "jenkins":
+                    if replication.policy is not None:
+                        policy_paths = get_policy_paths(
+                            replication.policy.name,
+                            replication.policy.instance.name,
+                        )
```
rporres, line 169: Hmmm, I don't really understand why you're copying the policy paths too... My understanding is that we were only going to copy the jenkins secrets... 🤔

reconcile/vault_replication.py
```diff
@@ -0,0 +1,176 @@
...
+                    dest_vault.write(write_dict)
+            else:
+                logging.info(["replicate_vault_secret", dest_version, version, secret])
+        except SecretNotFound:
+            logging.info(["replicate_vault_secret", "Secret not found", path])
+            if not dry_run:
+                dest_vault.write(write_dict)
+
+
+def copy_vault_secrets(
```
rporres, line 52: can you change a bit the name of either this or the above function? too similar 😅

reconcile/vault_replication.py
```diff
@@ -0,0 +1,176 @@
...
+    dry_run: bool,
+    source_vault: VaultClient,
+    dest_vault: VaultClient,
+    path_list: List[str],
+    policy_paths: List[str],
+) -> None:
+
+    for path in path_list:
+        if is_valid_path(path, policy_paths):
```
rporres, line 61: if I understand correctly this correct, this is going to silently discard the secrets it cannot copy, which is going to cause all kinds of misunderstandings... Most likely we want to collect all the secrets the policy doesn't allow us and fail or this is going to quickly blow on us.

reconcile/vault_replication.py
```diff
@@ -0,0 +1,176 @@
...
+QONTRACT_INTEGRATION = "vault-replication"
+
+
+def copy_vault_secret(
+    dry_run: bool, source_vault: VaultClient, dest_vault: VaultClient, path: str
+) -> None:
+
+    secret_dict = {"path": path, "version": "LATEST"}
+    secrets, version = source_vault.read_all_version(secret_dict)
+    for secret in secrets:
```
geoberle, line 29: it might be well true that my mental model of vault secrets is wrong, so please please let me know if that is the case :)

i thought the variable `secrets` contains all the `keys` that make up a single secret. if that is true (if not please let me know), updating the destination vault with `dest_vault.write` should be a single operation and not done within the loop for each key within the secret?

reconcile/vault_replication.py
```diff
@@ -0,0 +1,176 @@
...
+        write_dict = {"path": path, "data": secret}
+        try:
+            _, dest_version = dest_vault.read_all(secret_dict)
+            if dest_version < version:
```
geoberle, line 33: consider this. both source and dest vault are in sync and both contain the secret version 3 for a certain path.
then the secret gets updated twice in source vault:
* version 4 has content x
* version 5 has content y

then the reconcile loop starts, detects that the source secret version is newer than the dest secret version and updates the secret, resulting in `version 4 with content y` in dest vault.

then the reconcile loop starts again, still detecting that the version in source is newer than in dest and updates the dest secret to `version 5 with content y`

at this point we have in source
* version 1,2,3
* version 4 with content x
* version 5 with content y

and in dest
* version 1,2,3
* version 4 with content y
* version 5 with content y

depending on how we want to use synced secrets this might be a concern or not

reconcile/utils/vault.py
```diff
@@ -114,6 +117,32 @@ def read_all(self, secret):
         data = None
         if kv_version == 2:
             data = self._read_all_v2(secret_path, secret_version)
+            logging.info(["vault_helper", data])
```
geoberle, line 120: this would log the contents of the secret

reconcile/utils/vault.py
andreugallofre: Fixed!

reconcile/vault_replication.py
andreugallofre: 100% True, I misunderstood the implementation of the vault client that we have in qontract-reconcile, changed to match the expected behaviour.

reconcile/vault_replication.py
andreugallofre: When I was writing it I didn't think about it, but makes complete sense. I did a refactor to sync all the versions of a secret in any of the cases. With this we will have a complete mirror between vaults.

reconcile/vault_replication.py
```diff
@@ -20,33 +20,36 @@
 QONTRACT_INTEGRATION = "vault-replication"
 
 
+def deep_copy_versions(
+    dry_run: bool, source_vault, dest_vault, initial_version, last_version, path
+):
+    for version in range(initial_version, last_version + 1):
+        secret_dict = {"path": path, "version": version}
+        secret, version = source_vault.read_all_version(secret_dict)
```
geoberle, line 28: initial_version is the current version in dest, right? i think your index range starts 1 too soon

so lets say dest is at version 3 and source at version 5
initial_version=3, last_version=5 so you loop over [3,4,5]
you read version 3 from source and write it to dest, but it will become version 4 in dest because 3 exists already

andreugallofre: Changed the loop variables to clarify what was the actual values, and also added the +1 to match the expected outcome of copying all the versions

Thanks!

andreugallofre:
Integration and query classes need to be adapted to schema changes, but besides this, everything else is more or less ready

reconcile/vault_replication.py
```diff
@@ -0,0 +1,187 @@
+import logging
```
janboll, line 1: Since this is an entirely new integration, can you please create tests for it?

janboll (changes requested):

reconcile/vault_replication.py
andreugallofre: Solved on latest changes, now we fail with the complete list of secrets that can not be copied

reconcile/vault_replication.py
```diff
@@ -0,0 +1,225 @@
...
+                dry_run, source_vault, dest_vault, path_list, policy_paths
+            )
+
+
+def run(dry_run: bool) -> None:
+
+    query_data = vault_instances.query(query_func=gql.get_api().query)
+
+    for instance in query_data.vault_instances:
+        if isinstance(instance, VaultInstanceV1) and instance.replication:
```
rporres, line 203: why checking for `VaultInstanceV1`?

reconcile/vault_replication.py
```diff
@@ -0,0 +1,225 @@
...
+            for replication in instance.replication:
+                source_creds = get_vault_credentials(instance)
+                dest_creds = get_vault_credentials(replication.vault_instance)
+
+                source_vault = cast(
+                    _VaultClient,
```
rporres, line 209: I don't understand this very well. Why do you cast instead of using directly `_VaultClient`? If I understand correctly, you need to do it because `VaultClient` is a singleton.

reconcile/vault_replication.py
```diff
@@ -0,0 +1,225 @@
...
+    invalid_paths = []
+    for path in path_list:
+        if not policy_contais_path(path, policy_paths):
+            invalid_paths.append(path)
+
+    return invalid_paths
+
+
+def policy_contais_path(path: str, policy_paths: List[str]) -> bool:
+    if policy_paths is None:
```
rporres, line 106: this is weird. We should not call this function if policy_paths is None

reconcile/vault_replication.py
```diff
@@ -0,0 +1,225 @@
...
+) -> None:
+
+    secret_dict = {"path": path, "version": "LATEST"}
+    _, version = source_vault.read_all_version(secret_dict)
+
+    try:
+        _, dest_version = dest_vault.read_all_version(secret_dict)
+        if dest_version is None and version is None:
+            # Secret is a v1 and does not return version
+            copy_secret(dry_run, source_vault, dest_vault, path, 1, secret_dict)
```
rporres, line 66: Please use named arguments. See https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#function-arguments

reconcile/vault_replication.py
```diff
@@ -0,0 +1,225 @@
...
+                dry_run, source_vault, dest_vault, dest_version, version, path
+            )
+        else:
+            logging.info(["replicate_vault_secret", dest_version, version, path])
+    except SecretNotFound:
+        logging.info(["replicate_vault_secret", "Secret not found", path])
+        deep_copy_versions(dry_run, source_vault, dest_vault, 0, version, path)
+
+
+def check_copy_secret_list(
```
rporres, line 78: nit: I don't really like functions that do two different things. I'd rather see one function to check and one function to copy. That would even make clearer that we won't move on if there are paths the approle cannot copy.

reconcile/vault_replication.py
```diff
@@ -0,0 +1,225 @@
...
+from typing import List, cast, Optional
+
+QONTRACT_INTEGRATION = "vault-replication"
+
+
+def deep_copy_versions(
+    dry_run: bool,
+    source_vault,
+    dest_vault,
+    current_dest_version,
```
rporres, line 31: Since version has a meaning in terms of Vault secrets, it would be more exact that this is `None` in the case of non-existing dest secret instead of `0`.

rporres (changes requested):

reconcile/utils/vault.py
```diff
@@ -114,6 +117,32 @@ def read_all(self, secret):
...
+            secret_data = self._read_all_v1(secret_path)
+            data = secret_data, None
+
+        if data is None:
+            raise SecretNotFound
+
+        return data
+
+    @retry()
+    def read_all(self, secret) -> dict:
```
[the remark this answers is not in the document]
andreugallofre: Changed the function to use read_all_with_version and returning only the secret

reconcile/vault_replication.py
```diff
@@ -0,0 +1,225 @@
...
+    dest_vault: _VaultClient,
+    path_list: List[str],
+    policy_paths: Optional[List[str]],
+) -> None:
+
+    invalid_paths = []
+    if policy_paths is not None:
+        invalid_paths = list_invalid_paths(path_list, policy_paths)
+    if invalid_paths:
+        logging.error(["replicate_vault_secret", "Invalid paths", invalid_paths])
```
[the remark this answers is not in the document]
andreugallofre: Added a sys.exit, to actually fail if there is any path not allowed by the policy

reconcile/vault_replication.py
andreugallofre: Changed the check outside the function instead of checking it for every path

reconcile/vault_replication.py
```diff
@@ -0,0 +1,225 @@
...
+    policy_paths = []
+
+    for policy in query_data.policy:
+        if (
+            isinstance(policy, VaultPolicyV1)
+            and policy.name == policy_name
+            and policy.instance.name == instance_name
+        ):
+            for line in policy.rules.split("\n"):
+                res = re.search(r"path \s*[\'\"](.*)[\'\"]", line)
```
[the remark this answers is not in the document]
andreugallofre: Changed the regex, will check named captures and apply it

reconcile/vault_replication.py
andreugallofre: Simplified the function changing it for two different with clear functionality

reconcile/vault_replication.py
```diff
@@ -0,0 +1,225 @@
...
+    dry_run: bool, source_vault: _VaultClient, dest_vault: _VaultClient, path: str
+) -> None:
+
+    secret_dict = {"path": path, "version": "LATEST"}
+    _, version = source_vault.read_all_version(secret_dict)
+
+    try:
+        _, dest_version = dest_vault.read_all_version(secret_dict)
+        if dest_version is None and version is None:
+            # Secret is a v1 and does not return version
```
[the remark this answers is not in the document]
andreugallofre: 👍

reconcile/vault_replication.py
```diff
@@ -0,0 +1,225 @@
...
+        secret_dict = {"path": path, "version": version}
+        copy_secret(dry_run, source_vault, dest_vault, path, version, secret_dict)
+
+
+def copy_secret(
+    dry_run: bool,
+    source_vault: _VaultClient,
+    dest_vault: _VaultClient,
+    path: str,
+    version: int,
```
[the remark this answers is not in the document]
andreugallofre: Removed the version parameter from the function as is not needed

reconcile/vault_replication.py
andreugallofre: Changed to None

reconcile/vault_replication.py
rporres: Do we really need this function? It's just looping over the secrets and it's used only once. But if we do, please change the name as it is too close to `copy_vault_secret`

reconcile/vault_replication.py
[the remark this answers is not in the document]
rporres: what about raising a exception? It is a more standard way to end a program. It is not great to sys.exit in the middle of a function.

reconcile/vault_replication.py
```diff
@@ -0,0 +1,258 @@
...
+    dest_vault: _VaultClient,
+    path_list: List[str],
+) -> None:
+    for path in path_list:
+        copy_vault_secret(dry_run, source_vault, dest_vault, path)
+
+
+def list_invalid_paths(path_list: List[str], policy_paths: List[str]) -> List[str]:
+    invalid_paths = []
+    if policy_paths is None:
```
rporres, line 127: From the function signature this cannot happen, as it is not `Optional`. And you're only calling `list_invalid_paths` if policy_paths is not None.

reconcile/vault_replication.py
```diff
@@ -0,0 +1,258 @@
...
+
+def check_invalid_paths(
+    path_list: List[str],
+    policy_paths: Optional[List[str]],
+) -> None:
+
+    invalid_paths = []
+    if policy_paths is not None:
+        invalid_paths = list_invalid_paths(path_list, policy_paths)
+    if invalid_paths:
```
rporres, line 110: You can put this inside the `if policy_paths is not None` condition

reconcile/vault_replication.py
```diff
@@ -0,0 +1,258 @@
...
+def copy_secret(
+    dry_run: bool,
+    source_vault: _VaultClient,
+    dest_vault: _VaultClient,
+    path: str,
+    secret_dict: dict,
+) -> None:
+    secret, src_version = source_vault.read_all_with_version(secret_dict)
+    write_dict = {"path": path, "data": secret}
+    logging.info(["replicate_vault_secret", src_version, path])
```
rporres, line 56: how is this going to work with `v1` secrets that return `src_version == None`

reconcile/vault_replication.py
andreugallofre: 👍

reconcile/vault_replication.py
andreugallofre: Changed!

reconcile/vault_replication.py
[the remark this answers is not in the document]
andreugallofre: Changed to rise an exception, I used the sys.exit because i saw it on other parts of QR, but make sense to raise a custom exception

reconcile/vault_replication.py
andreugallofre: Solved, it was a problem with the qenerate versions and nullable fields, should be ok now.

reconcile/utils/vault.py
```diff
@@ -166,7 +181,8 @@ def _read_all_v2(self, path, version):
             raise SecretNotFound(path)
 
         data = secret["data"]["data"]
-        return data
+        secret_version = secret["data"]["metadata"]["version"]
+        return data, secret_version
```
rporres, line 185: nit: since you have modified this function, please add type hints.

reconcile/vault_replication.py
```diff
@@ -0,0 +1,232 @@
...
+            path=path,
+        )
+
+
+def check_invalid_paths(
+    path_list: List[str],
+    policy_paths: Optional[List[str]],
+) -> None:
+
+    invalid_paths = []
```
rporres, line 84: Do you need to define this outside the condition?

reconcile/vault_replication.py
rporres: But there's no version in v1 secrets.

reconcile/vault_replication.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def get_jenkins_secret_list(jenkins_instance: str) -> List[str]:
+    secret_list = []
+    query_data = jenkins_configs.query(query_func=gql.get_api().query)
+
+    if query_data.jenkins_configs:
+        for p in query_data.jenkins_configs:
+            if (
+                isinstance(p, JenkinsConfigV1_JenkinsConfigV1)
```
rporres, line 139: do we need the `isinstance` bit?

reconcile/vault_replication.py
```diff
@@ -0,0 +1,232 @@
...
+    secret_dict = {"path": path, "version": "LATEST"}
+    _, version = source_vault.read_all_with_version(secret_dict)
+
+    try:
+        _, dest_version = dest_vault.read_all_with_version(secret_dict)
+        if dest_version is None and version is None:
+            # v1 secrets don't have version
+            secret, src_version = source_vault.read_all_with_version(secret_dict)
+            write_dict = {"path": path, "data": secret}
+            logging.info(["replicate_vault_secret", src_version, path])
```
rporres, line 53: v1 secrets don't have version

reconcile/vault_replication.py
rporres: hmm, you're still calling it using `0` as `current_dest_version` when it is missing.

reconcile/vault_replication.py
```diff
@@ -0,0 +1,232 @@
...
+from typing import List, cast, Optional
+
+QONTRACT_INTEGRATION = "vault-replication"
+
+
+class VaultInvalidPaths(Exception):
+    pass
+
+
+def deep_copy_versions(
```
rporres, line 23: missing some type hints here.

reconcile/vault_replication.py
```diff
@@ -0,0 +1,232 @@
...
+                ]
+                for line in secret_paths:
+                    res = re.search(r"secret-path:\s*[\'\"](.+)[\'\"]", line)
+                    if res is not None:
+                        secret_list.append(res.group(1))
+
+    return secret_list
+
+
+def get_vault_credentials(vault_instance):
```
rporres, line 156: missing type hints.

rporres (changes requested):

reconcile/vault_replication.py
andreugallofre: Type checking complains if I don't make sure that is a JenkinsConfigV1_JenkinsConfigV1 type, not sure if there is another way to do it

reconcile/vault_replication.py
andreugallofre: Solved

reconcile/vault_replication.py
andreugallofre: Added some tests, will add more later today :)

reconcile/utils/vault.py
andreugallofre: Added type hints

reconcile/vault_replication.py
andreugallofre: Had to work around this type hints as mypy was complaining, not sure if we like this changes

reconcile/vault_replication.py
andreugallofre: Removed the function

reconcile/utils/vault.py
```diff
@@ -115,13 +118,25 @@ def read_all(self, secret):
         if kv_version == 2:
             data = self._read_all_v2(secret_path, secret_version)
         else:
-            data = self._read_all_v1(secret_path)
+            secret_data = self._read_all_v1(secret_path)
+            data = secret_data, None
```
rporres, line 122: I think we talked about this or a I put it in a comment, but I'd honestly prefer that `data` is still a dict and this function explicitly returned a tuple in the form of

```
return data, version
```

So that it is even clearer how different it is from `read_all`.

Please also fix the docstring, which is no longer exact.

reconcile/utils/vault.py
rporres: Hmmm, I think the was left in your computer, since I see no type hints in `_read_all_v2` 😅

rporres (changes requested):

reconcile/vault_replication.py
rporres: It is not complete, though... Still missing type hints for the return.

reconcile/vault_replication.py
```diff
@@ -0,0 +1,231 @@
...
+                for line in secret_paths:
+                    res = re.search(r"secret-path:\s*[\'\"](.+)[\'\"]", line)
+                    if res is not None:
+                        secret_list.append(res.group(1))
+
+    return secret_list
+
+
+def get_vault_credentials(vault_instance: VaultInstanceV1):
+    vault_creds = {"server": "", "role_id": None, "secret_id": None}
```
rporres, line 153: why do you need to initialize the keys if you're defining them afterwards?

reconcile/vault_replication.py
```diff
@@ -0,0 +1,231 @@
...
+    vault = cast(_VaultClient, VaultClient())
```
rporres, line 154: which client is this? Why do you need it?

reconcile/vault_replication.py
```diff
@@ -0,0 +1,231 @@
...
+    vault_instance_auth = cast(VaultInstanceAuthApproleV1, vault_instance.auth)
+
+    role_id = {
+        "path": vault_instance_auth.role_id.path,
+        "field": vault_instance_auth.role_id.field,
+    }
+    secret_id = {
+        "path": vault_instance_auth.secret_id.path,
+        "field": vault_instance_auth.secret_id.field,
+    }
```
rporres, line 165: Can't you use directly `vault_instance.auth`?

reconcile/vault_replication.py
andreugallofre: This is a client for the "main" vault, the vault definitions on app-interface have the path and version used for the secrets, but to authenticate it we need to retrieve the actual value.

reconcile/utils/vault.py
andreugallofre: Submitted 😢

reconcile/vault_replication.py
rporres: Understood. Please make sure that you add a nice docstring and a quick comment explaining what this function does. It is a bit confusing at first that you have to create yet another Vault client here.

andreugallofre: Added the docstring to the function to clarify what it does :)

reconcile/vault_replication.py
andreugallofre: Changed, not sure why initially started with that

reconcile/utils/vault.py
andreugallofre: Fixed both to match the current usecase

reconcile/gql_definitions/vault_instances/vault_instances.gql
```diff
@@ -0,0 +1,58 @@
...
+  vault_instances: vault_instances_v1 {
+    name
+    address
+    auth {
+      provider
+      secretEngine
+      ... on VaultInstanceAuthApprole_v1 {
+      roleID {
+        path
+        field
```
fishi0x01, line 13: Any reason to not use fragment `... VaultSecret` here? It has 2 additional fields `version` and `format`, which are redundant in your case, but you can neglect them. The extra "gql query overhead" wouldn't hurt and you'd greatly reduce the number of generated classes.

fishi0x01: (Same applies to all other VaultSecret occurrences in this query)

reconcile/vault_replication.py
fishi0x01: The reason lies here:

```gql
          auth {
            provider
            secretEngine
            ... on VaultInstanceAuthApprole_v1 {
              roleID {
                path
                field
              }
              secretID {
                path
                field
              }
            }
          }
```

By graphql standard, this inline query can return 2 kinds of auths: one is `VaultInstanceAuthApprole_v1` and the other is the interface itself (which is everything except `VaultInstanceAuthApprole_v1`). I.e., your auth object must be a Union to properly abstract this. The underlying problem with inline queries is, that they are not guaranteed to only return objects that are defined by the inline - they still return anything that is not defined by the inline.

My understanding is you could also make mypy happy by using `isinstance()` to filter any none `VaultInstanceAuthApprole_v1`

rporres: d'oh, that lack of guarantee sucks. I mean, the whole point of using the `... on VaultInstanceAuthApprole_v1` is to only return objects of that kind 😂

fishi0x01: Another funny aspect of gql is, even if you pull in all attributes into the inlined fragment like this:

```gql
    auth {
            ... on VaultInstanceAuthApprole_v1 {
              provider
              secretEngine
              roleID {
                path
                field
              }
              secretID {
                path
                field
              }
            }
          }
```

You'd still require a Union with the Interface (which has no fields then), because the gql server will still return `{}` for any none `VaultInstanceAuthApprole_v1` object. I.e., once you go inline fragment query, you must deal with the Union.

fishi0x01: My understanding is that in this specific scenario we only have a single auth provider: approle. I verified by firing these 2 queries:

```gql
query VaultInstances {
  vault_instances: vault_instances_v1 {
    name
    address
    auth {
      provider
      secretEngine
      ... on VaultInstanceAuthApprole_v1 {
        roleID {
          path
          field
        }
      }
    }
  }
}
```

```gql
query VaultInstances {
  vault_instances: vault_instances_v1 {
    name
    address
    auth {
      provider
      secretEngine
    }
  }
}
```

Both return exactly the same objects, but with different fields.

I.e., in this specific case the Unions seems redundant indeed, because we only have approle providers. However, if in app-interface we are to add another auth provider, then the above **fully inlined** query would return `{}` or `{"provider": "sth", "name": "sth"}` for anything which is not an approle provider. That will lead to issues if there is no Union to catch it.

rporres: hmm. the schema allows for two types: https://github.com/app-sre/qontract-schemas/blob/main/graphql-schemas/schema.yml#L2310-L2320

But we only have approle in app-interface. I don't know why we support both. Maybe fedramp 🤷

rporres: So your recommendation is to use `isinstance` instead of casting, correct @fishi0x01 ?

andreugallofre: Yes, we talked about it on the call this morning, I'm doing the changes :)

fishi0x01: Yes `isinstance` will make mypy happy because it filters the desired objects in a safe way. `cast` will lead to undesired results, once we have different auth provider classes returned.

reconcile/vault_replication.py
andreugallofre: Fixed the function to use _VaultClient, didn't noticed that was a singleton, but make sense for the previous use cases

reconcile/vault_replication.py
andreugallofre: Changed cast to `isinstance()` now we should be safe, thanks!

reconcile/gql_definitions/vault_instances/vault_instances.gql
andreugallofre: Changed all the occurrences of the VaultSecret fragment in all the queries

reconcile/vault_replication.py
andreugallofre: Changed the logic to not call `deep_copy_versions` when the secret is a v1, now it's a simple write

reconcile/vault_replication.py
andreugallofre: Fixed, I was not understanding the comment, sorry!

reconcile/vault_replication.py
rporres: great. Please add a comment explaining why you need to do it.

reconcile/vault_replication.py
```diff
@@ -0,0 +1,252 @@
...
+    * vault_instance - VaultInstanceV1 object from AppInterface Data.
+    """
+    vault_creds = {}
+    vault = cast(_VaultClient, VaultClient())
+
+    if not isinstance(
+        vault_instance.auth, VaultInstanceAuthApproleV1
+    ) and not isinstance(
+        vault_instance.auth,
+        VaultReplicationConfigV1_VaultInstanceV1_VaultInstanceAuthV1_VaultInstanceAuthApproleV1,
```
rporres, line 183: mypy complaining again? 😢

andreugallofre: Since I removed the cast to `VaultInstanceV1` the object can be two different things, the `VaultInstanceV1` itself, or the VaultInstance defined inside the replication part of the schema. That second one is the type of `VaultReplicationConfigV1_VaultInstanceV1` and mypy / the isinstance was not happy

rporres: ok!

reconcile/vault_replication.py
```diff
@@ -0,0 +1,255 @@
...
+            deep_copy_versions(
+                dry_run=dry_run,
+                source_vault=source_vault,
+                dest_vault=dest_vault,
+                current_dest_version=dest_version,
+                current_source_version=version,
+                path=path,
+            )
+        else:
+            logging.info(["replicate_vault_secret", dest_version, version, path])
```
rporres, line 76: is this log line necessary? This is going to cover the case dest_version == version, in which we don't replicate anything, isn't it?

reconcile/test/test_vault_replication.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def test_check_invalid_paths_ok(self):
+        path_list = ["path1", "path2"]
+        policy_paths = ["path1", "path2"]
+        integ.check_invalid_paths(path_list, policy_paths)
+
+    def test_list_invalid_paths(self):
+        path_list = ["path1", "path3"]
+        policy_paths = ["path1", "path2"]
+        self.assertEqual(integ.list_invalid_paths(path_list, policy_paths), ["path3"])
```
rporres, line 31: any chance to test the rest of the integration? Maybe mocking vault client?

rporres (commented):
On top of that and now that the integration bit is stable and looking good, can you add details in the description on how have you tested this in real life?

reconcile/test/test_vault_replication.py
steveteahan: To add a related question, what percentage of the new code is covered, and what are we targeting? A percentage of coverage isn't the end goal by any means, but it's a reasonable start.

reconcile/gql_definitions/vault_instances/vault_instances.py
```diff
@@ -0,0 +1,234 @@
+"""
+Generated by qenerate plugin=pydantic_v1. DO NOT MODIFY MANUALLY!
+"""
+from enum import Enum  # noqa: F401 # pylint: disable=W0611
+from typing import (  # noqa: F401 # pylint: disable=W0611
+    Any,
+    Callable,
```
rporres, line 7: you will need to regenerate your classes with the latest qenerate to make sure this is imported from `collections.abc`

reconcile/utils/vault.py
```diff
@@ -8,6 +8,7 @@
 import hvac
 import requests
 
+from typing import Tuple, Optional, Mapping
```
rporres, line 11: * no need to import `Tuple`, just use `tuple` directly
* `Mapping` should be imported from `collections.abc`

reconcile/vault_replication.py
```diff
@@ -0,0 +1,332 @@
...
+from reconcile.gql_definitions.vault_instances import vault_instances
+from reconcile.gql_definitions.vault_instances.vault_instances import (
+    VaultInstanceAuthApproleV1,
+    VaultInstanceV1,
+    VaultReplicationConfigV1_VaultInstanceV1_VaultInstanceAuthV1_VaultInstanceAuthApproleV1,
+    VaultReplicationConfigV1_VaultInstanceV1,
+)
+
+from reconcile.utils import gql
+from typing import cast, Optional, Union, Iterable, Mapping
```
rporres, line 31: * `Mapping` and `Iterable` should be imported from `collections.abc`

reconcile/vault_replication.py
```diff
@@ -0,0 +1,255 @@
...
+                current_dest_version=dest_version,
+                current_source_version=version,
+                path=path,
+            )
+        else:
+            logging.info(["replicate_vault_secret", dest_version, version, path])
+    except SecretNotFound:
+        logging.info(["replicate_vault_secret", "Secret not found", path])
+        # Handle v1 secrets where version is None and we don't need to deep sync.
+        if version is None:
```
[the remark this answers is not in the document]
andreugallofre: Added the logging line, you are right, the "secret not found" was not clear

reconcile/vault_replication.py
andreugallofre: Removed the else condition

reconcile/test/test_vault_replication.py
```diff
@@ -0,0 +1,31 @@
+from unittest import TestCase
```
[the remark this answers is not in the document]
andreugallofre: Changed the tests to pytest

reconcile/test/test_vault_replication.py
andreugallofre: Added some more tests for the integration, now we are at 70% coverage or so, I'll add a few more tests for the placeholder replacement function that I lastly added

reconcile/utils/vault.py
```diff
@@ -98,8 +101,9 @@ def _refresh_client_auth(self):
         self._client.auth_approle(self.role_id, self.secret_id)
 
     @retry()
-    def read_all(self, secret):
-        """Returns a dictionary of keys and values in a Vault secret.
+    def read_all_with_version(self, secret: dict) -> Tuple[dict, Optional[str]]:
```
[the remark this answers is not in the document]
andreugallofre: Changed the types on the parameters of the functions to abstract collection types

reconcile/gql_definitions/vault_instances/vault_instances.py
andreugallofre: Done!

reconcile/utils/vault.py
andreugallofre: Changed!

reconcile/vault_replication.py
andreugallofre: Changed!

reconcile/vault_replication.py
```diff
@@ -0,0 +1,386 @@
...
+
+def process_wildcard_paths_with_text(
+    vault_instance: _VaultClient, path: str
+) -> list[str]:
+    """Returns a list of secrets that match a wildcard path
+
+    The input secret is the name of a jenkins instance to filter
+    the secrets:
+    * path - Vault path
+    """
```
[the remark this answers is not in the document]
rporres: or maybe something shorter like `get_secrets_from_templated_path` 😂

reconcile/test/test_vault_replication.py
```diff
@@ -0,0 +1,182 @@
...
+                ),
+                type="secrets",
+                config=None,
+                config_path=ResourceV1(
+                    content="name: 'test_data_name'\n    secret-path: 'this/is/a/path'"
+                ),
+            ),
+        ]
+    )
+
```
rporres, line 80: you don't need to build this manually, that's what qenerate does for you. You should be able to get the raw JSON, add it as fixture and do `JenkinsConfigsQueryData(**raw_data)`. This is what the generated code does.

reconcile/test/test_vault_replication.py
```diff
@@ -0,0 +1,182 @@
...
+
+@pytest.fixture
+def vault_client_test() -> _VaultClient:
+    return cast(_VaultClient, None)
+
+
+def test_get_jenkins_secret_list_w_content(
```
rporres, line 87: apart from this one, I guess you will need to test the case where you have a templated secret, right?

reconcile/test/test_vault_replication.py
```diff
@@ -0,0 +1,182 @@
...
+    integ.check_invalid_paths(path_list, policy_paths)
+
+
+def test_list_invalid_paths():
+    path_list = ["path1", "path3"]
+    policy_paths = ["path1", "path2"]
+    assert integ.list_invalid_paths(path_list, policy_paths) == ["path3"]
+
+
+@pytest.fixture
```
rporres, line 55: nit: usually we declare all fixtures at the beginning of a test file and then we write all the tests

reconcile/vault_replication.py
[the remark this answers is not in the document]
andreugallofre: Changed!

reconcile/vault_replication.py
```diff
@@ -0,0 +1,392 @@
...
+        try:
+            secret, src_version = source_vault.read_all_with_version(secret_dict)
+            write_dict = {"path": path, "data": secret}
+            logging.info(["replicate_vault_secret", src_version, path])
+            if not dry_run:
+                dest_vault.write(
+                    secret=write_dict, decode_base64=False, force_write=True
+                )
+
+        except SecretNotFound:
```
rporres, line 68: You have two calls in the `try` block, one to read, one to write. I would catch this only in one of the two calls so that it is clearer what you're trying.

reconcile/vault_replication.py
```diff
@@ -0,0 +1,392 @@
...
+            # Handle the case where the version is already deleted from source, and
+            # we need to write a dummy version to the destination to keep the sync
+            # with versions in the source.
+            write_dummy_versions(
+                dry_run=dry_run,
+                dest_vault=dest_vault,
+                version_needed=version,
+                path=path,
+            )
```
rporres, line 77: If I understand correctly, this is going to happen every time we have a source version > 1, correct? Can we simplify this? I mean, would it be simpler to just create the missing secrets whenever source version is > 1?

reconcile/vault_replication.py
```diff
@@ -0,0 +1,392 @@
...
+def get_secrets_from_templated_path(
+    vault_instance: _VaultClient, path: str
+) -> list[str]:
+    """Returns a list of secrets that result from a templated path
+    expansion.
+
+    The input secret is the name of a jenkins instance to filter
+    the secrets:
+    * path - Vault path
+    """
```
rporres, line 332:
```suggestion
    """Connects to the given Vault instance and returns a list of secrets
    that result from a templated path expansion.

    The input secret is the name of a jenkins instance to filter
    the secrets:
    * path - Vault path
    """
```

reconcile/vault_replication.py
```diff
@@ -0,0 +1,392 @@
...
+    for s in path_slices:
+        if "{" in s and "}" in s:
+            template = s
+
+    cap_groups = re.search(r"(.*)(\{.*\})(.*)", template)
+    if cap_groups is not None:
+        prefix = cap_groups.group(1)
+        suffix = cap_groups.group(3)
+    else:
+        raise VaultInvalidPaths
```
rporres, line 345: Please add nice and helpful comments to every exception you raise.

reconcile/utils/vault.py
```diff
@@ -244,21 +277,23 @@ def write(self, secret, decode_base64=True):
...
         try:
-            current_data = self._read_all_v2(path, version=SECRET_VERSION_LATEST)
-            if current_data == data:
+            current_data, _ = self._read_all_v2(path, version=SECRET_VERSION_LATEST)
+            if current_data == data and not force_write:
                 logging.debug(f"current data is up-to-date, skipping {path}")
                 return
         except SecretVersionNotFound:
+            pass
+        except SecretAccessForbidden:
```
rporres, line 296: why have you changed the logic of how this works? The debug message is now misleading...

reconcile/vault_replication.py
```diff
@@ -0,0 +1,392 @@
...
+    path: str,
+) -> None:
+
+    for version in range(0, version_needed):
+        write_dict = {"path": path, "data": {"dummy": "data"}}
+        logging.info(
+            ["replicate_vault_secret", "generate_dummy_data", version + 1, path]
+        )
+        if not dry_run:
+            dest_vault.write(secret=write_dict, decode_base64=False, force_write=True)
```
rporres, line 93: can you add a comment on why you need force write here?

reconcile/vault_replication.py
```diff
@@ -0,0 +1,392 @@
...
+    try:
+        _, dest_version = dest_vault.read_all_with_version(secret_dict)
+        if dest_version is None and version is None:
+            # v1 secrets don't have version
+            secret, _ = source_vault.read_all_with_version(secret_dict)
+            write_dict = {"path": path, "data": secret}
+            logging.info(["replicate_vault_secret", path])
+            if not dry_run:
+                dest_vault.write(
+                    secret=write_dict, decode_base64=False, force_write=True
```
rporres, line 116: ditto wrt `force_write`

reconcile/vault_replication.py
andreugallofre: Changed!

reconcile/utils/vault.py
andreugallofre: Pushed by error, I restored the original logic

reconcile/vault_replication.py
andreugallofre: Done!

reconcile/vault_replication.py
andreugallofre: Done!

reconcile/vault_replication.py
```diff
@@ -0,0 +1,392 @@
...
+                dest_vault=dest_vault,
+                version_needed=version,
+                path=path,
+            )
+
+
+def write_dummy_versions(
+    dry_run: bool,
+    dest_vault: _VaultClient,
+    version_needed: int,
```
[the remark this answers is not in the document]
andreugallofre: Done!

reconcile/test/test_vault_replication.py
andreugallofre: Changed!

reconcile/vault_replication.py
rporres: this is not going to work like this. Looks better the way it is.

rporres (approved):
Since you have tested this extensively and we're approaching change freeze, let's merge and add the missing tests later.

Ship it!

```
           __  _
       .-.'  `; `-._  __  _
      (_,         .-:'  `; `-._
    ,'o"(        (_,           )
   (__,-'      ,'o"(            )>
      (       (__,-'            )
       `-'._.--._(             )
          |||  |||`-'._.--._.-'
                     |||  |||
```

geoberle (approved):
@andreugallofre  it is fantastic to see how this integration evolved and how many edge cases it can handle reliably. you had your first real-life test scenarios covered already when you migrated data between the two vaults. so also LGTM from me under the premise that tests will be added before new changes are introduced to the code.

rporres:
it looks one of the new tests is failing

rporres (approved):

PR #2785, opened 2022-09-12
[APPSRE-4798] Add support for new parameter_group for RDS upgrade while keeping the old one intact.

bhushanthakur93:
Related: APPSRE-4798

This change depends on https://github.com/app-sre/qontract-schemas/pull/242 

Background: The current process for major version upgrade involves updating parameter_group in place with new engine_version. Due to existing implementation this causes terraform to delete the old parameter group first before the upgrade is complete resulting into the error. 

This MR addresses the issue by allowing to specify new parameter group which will be used for upgrade while keeping the old parameter group intact.

Test cases using app-interface-dev-data:

1.  Create a new RDS instance. This is to ensure we don't break existing process.

```
  - provider: rds
    identifier: dev-bhthakur-rds
    defaults: /terraform/rds/bhthakur-rds-defaults.yml
    parameter_group: /terraform/rds/bhthakur-rds-params.yml
    output_resource_name: dev-bhthakur-rds-creds
```

2. Upgrade to new RDS version. This involves specifying `old_parameter_group`

```
  - provider: rds
    identifier: dev-bhthakur-rds
    defaults: /terraform/rds/bhthakur-rds-defaults.yml
    parameter_group: /terraform/rds/bhthakur-rds-params-1.yml
    old_parameter_group: /terraform/rds/bhthakur-rds-params.yml
    output_resource_name: dev-bhthakur-rds-creds
    overrides:
      engine_version: '12.11'
      apply_immediately: true
```
3. Do another consecutive upgrade

```
  - provider: rds
    identifier: dev-bhthakur-rds
    defaults: /terraform/rds/bhthakur-rds-defaults.yml
    parameter_group: /terraform/rds/bhthakur-rds-params-2.yml ## 
    old_parameter_group: /terraform/rds/bhthakur-rds-params-1.yml ## this refers to old_parameter_group from ex 2.
    output_resource_name: dev-bhthakur-rds-creds
    overrides:
      engine_version: '13.7'
      apply_immediately: true
```

A small refactor has been done to consolidate parameter_group terraform logic into a inner function. In addition, some safety checks have been implemented to ensure we have a unique name and `old_parameter_group` is used only when `parameter_group` is defined.

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -46,7 +46,7 @@ provider
   defaults
   availability_zone
   parameter_group
-  new_parameter_group
+  old_parameter_group
   overrides
   output_resource_name
   enhanced_monitoring
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -46,6 +46,7 @@ provider
   defaults
   availability_zone
   parameter_group
+  new_parameter_group
   overrides
   output_resource_name
   enhanced_monitoring
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -178,6 +178,7 @@ VARIABLE_KEYS = [
     "region",
     "availability_zone",
     "parameter_group",
+    "new_parameter_group",
     "name",
     "enhanced_monitoring",
     "replica_source",
@@ -1240,13 +1241,8 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
                 elif provider != provider_region:
                     raise ValueError("region does not match availability zone")
 
-        # 'deps' should contain a list of terraform resource names
-        # (not full objects) that must be created
-        # before the actual RDS instance should be created
-        deps = []
-        parameter_group = values.pop("parameter_group", None)
-        if parameter_group:
-            pg_values = self.get_values(parameter_group)
+        def populate_parameter_group(name: str) -> aws_db_parameter_group:
+            pg_values = self.get_values(name)
             # Parameter group name is not required by terraform.
             # However, our integration has it marked as required.
             # If user does not provide a name, we will use the rds identifier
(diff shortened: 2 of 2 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -1258,10 +1254,42 @@ def populate_tf_resource_rds(self, spec):
...
+            values["parameter_group_name"] = pg_tf_resource.get("name")
+
+        new_parameter_group = values.pop("new_parameter_group", None)
+        if new_parameter_group:
+            # discourage using new_parameter_group if parameter_group field is not utilized.
+            if parameter_group is None:
+                raise ValueError(
+                    "Cannot use new_parameter_group field without parameter_group."
+                    "This field is only used during RDS major version upgrade"
+                )
```
maorfr, line 1279: this is a nice addition. you can implement also implement a dependency in the schemas such that `new_parameter_group` depends on `parameter_group`.

maorfr:
i like this approach, because it means an upgrade can be done in a single MR.
i was sure that to perform a major version upgrade, one must switch to using the `default` parameter group first. do things work properly when performing a major version upgrade and specifying a parameter group of the new version, at the same time?

some more options that you may have considered, but i'll put them here for visibility (not saying they are better, just that they exist):
1. error if a parameter group is being deleted but is still used around here: https://github.com/app-sre/qontract-reconcile/blob/63797e6528b8bb0dd99b6d24cef80d4adb03cb9b/reconcile/utils/terraform_client.py#L266-L278. that will force users to explicitly specify the default parameter group in a separate MR.
2. in case a parameter group is not defined, set it to use the default. something like https://github.com/app-sre/qontract-reconcile/blob/63797e6528b8bb0dd99b6d24cef80d4adb03cb9b/reconcile/utils/terrascript_aws_client.py#L1218-L1219

reconcile/utils/terrascript_aws_client.py
bhushanthakur93: That is a good idea! Let me look into this.

bhushanthakur93:
> i like this approach, because it means an upgrade can be done in a single MR. i was sure that to perform a major version upgrade, one must switch to using the `default` parameter group first. do things work properly when performing a major version upgrade and specifying a parameter group of the new version, at the same time?
Yes! As long as we keep the old parameter group we can do the upgrade in a single MR.
> 
> some more options that you may have considered, but i'll put them here for visibility (not saying they are better, just that they exist):
> 
> 1. error if a parameter group is being deleted but is still used around here: https://github.com/app-sre/qontract-reconcile/blob/63797e6528b8bb0dd99b6d24cef80d4adb03cb9b/reconcile/utils/terraform_client.py#L266-L278
>    . that will force users to explicitly specify the default parameter group in a separate MR.
Glad that you brought up this case. I'll need to document this in SOP because the #3 case deletes the very first parameter group and whoever is running the SOP should not be surprised by this.
> 2. in case a parameter group is not defined, set it to use the default. something like https://github.com/app-sre/qontract-reconcile/blob/63797e6528b8bb0dd99b6d24cef80d4adb03cb9b/reconcile/utils/terrascript_aws_client.py#L1218-L1219

reconcile/utils/terrascript_aws_client.py
bhushanthakur93: I updated qontract-schemas to add dependency on the field. I'm keeping this check so this is compatible with older schema versions if we need to.

steveteahan:
> i was sure that to perform a major version upgrade, one must switch to using the `default` parameter group first

If I'm not mistaken, there's not anything special about the `default` parameter group, it's just that changing the instance to use any other parameter group means that you're no longer using the parameter group defined in app-interface. As a result, the old parameter group could be deleted, which would unblock the upgrade. This change avoids that by deferring the deletion of the original parameter group.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1240,13 +1241,8 @@ def populate_tf_resource_rds(self, spec):
...
                     raise ValueError("region does not match availability zone")
 
-        # 'deps' should contain a list of terraform resource names
-        # (not full objects) that must be created
-        # before the actual RDS instance should be created
-        deps = []
-        parameter_group = values.pop("parameter_group", None)
-        if parameter_group:
-            pg_values = self.get_values(parameter_group)
+        def populate_parameter_group(name: str) -> aws_db_parameter_group:
```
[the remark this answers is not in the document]
bhushanthakur93: Since this is an inner method the scope is local.

reconcile/terraform_resources.py
```diff
@@ -46,6 +46,7 @@
   defaults
   availability_zone
   parameter_group
+  new_parameter_group
```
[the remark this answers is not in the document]
bhushanthakur93: This was very good suggestion! The user experience is less confusing now :)

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1258,10 +1254,36 @@ def populate_tf_resource_rds(self, spec):
...
+            if parameter_group is None:
+                raise ValueError(
+                    "Cannot use old_parameter_group field without parameter_group."
+                    "This field is only used during RDS major version upgrade"
+                )
+
+            old_pg_tf_resource = populate_parameter_group(old_parameter_group)
+
+            if old_pg_tf_resource.get("name") == pg_tf_resource.get("name"):
+                raise ValueError("Must supply a unique name value for parameter group")
```
steveteahan, line 1284: The only other thing that I would add here is that it might save us time to make this error message a little bit more descriptive, particularly about how to fix it.

steveteahan (approved):

PR #2790, opened 2022-09-13
APPSRE-5962 qenerate naming collision strategy

fishi0x01:
qenerate now supports different naming collision strategies. By default, the parent class name is prefixed if a collision is detected. However, the `vpc_peerings_validator` enumerates now, instead of prefixing the parent class name.

cc @maorfr


maorfr (approved):
I like it!

PR #2791, opened 2022-09-14
gitlab_housekeeping: fix paging bug in mr pipeline retrieval

chassing:


reconcile/utils/gitlab_api.py
```diff
@@ -301,6 +301,9 @@ def get_merge_requests(self, state):
     def get_merge_request_label_events(self, mr: ProjectMergeRequest):
         return self.get_items(mr.resourcelabelevents.list)
 
+    def get_merge_request_pipelines(self, mr: ProjectMergeRequest):
```
fishi0x01, line 304: Should this have a return type?

geoberle (approved):
every gitlab API bugfix revolves around pagination it seems :)

PR #2794, opened 2022-09-15
adapt to non-nullable lists

chassing:
regenerate (qenerate) all gql classes to adapt latest qontract-server changes
see https://github.com/app-sre/qontract-server/commit/bb640ea8951d8c1875943ff81ffd2651cdb0ec19


jfchevrette:
What about the other uses of `filter_null`? Ex: https://github.com/app-sre/qontract-reconcile/blob/c7c522c4c47f90fe47520fef63a7c2ab66b169a6/reconcile/vpc_peerings_validator.py#L26

geoberle (approved):
thanks!

chassing:
> What about the other uses of `filter_null`? Ex:

Will be addressed by @geoberle and @fishi0x01 afterward.

geoberle:
i will start removing unnecessary None checks in `change_owners` once this is merged

PR #2796, opened 2022-09-16
[change-owners] remove unnecessary none checks

geoberle:
due to the recent change in qontract-server, all `isRequired` lists are non NonNull from a GraphGL perspective. also the elements of a list are NonNull as well. based on that, qenerate comes up with much simpler type hints on the generated data classes, which results in a lot less None checks within integrations.


fishi0x01 (approved):
:tada:

PR #2798, opened 2022-09-16
gitlab-housekeeping: fix for deleted label

chassing:
when a label has been removed, the label attribute in label events can be None


geoberle (approved):
thx!

PR #2801, opened 2022-09-20
Add google chat as fluentd log store option

TGPSKI:
This MR updates the helm template to include a conditional google chat log store option. This option is used in FedRAMP environment for chatops.

This MR also adds a new values file, values-manager-fedramp.yaml. Because slack is prohibited due to FR controls, we must have a way to disable slack reporting for all integrations. Every downstream integration can manage this on its own, but integrations-manager itself must be tempalted with `slack: false` and `googleChat: true`.

This MR depends on https://github.com/app-sre/fluentd/pull/4 merged + a new "latest" tag pushed to the quay repository. The image is already validated + tested in isolated workloads on the FR cluster.


janboll (commented):
Can you please consider adding a test here? https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/test_utils_helm.py

TGPSKI:
I added the tests, however I still need to update the template properly. I will be able to resume work on this next Tuesday (I'm out Thurs-Mon).

TGPSKI:
@janboll any hints on generating a passing test yaml? I have spent about an hour trying to generate a passing test with the following values:

```
integrations:
- name: integ
  resources:
    requests:
      memory: 45Mi
      cpu: 123
    limits:
      memory: 90Mi
      cpu: 678
  logs:
    googleChat: true
  shard_specs:
  - shard_id: 0
    shards: 1
    shard_name_suffix: ""
```

I am unclear of the value of this test but I am happy to continue to try to get it to pass.

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -138,6 +149,13 @@ objects:
                 message "\`\`\`[{{ $integration.name }}] %s\`\`\`"
               </store>
               {{- end }}
+              {{- if $logs.googleChat }}
+              <store>
+                @type teams
```
janboll, line 154: I am confused by type teams here.

janboll (approved):
:+1:

helm/qontract-reconcile/templates/template.yaml
TGPSKI: Check the fluentd PR for a more detailed explanation: https://github.com/app-sre/fluentd/pull/4

Perhaps I should include a comment here for visibility?

janboll: That would be helpful I think.
So googlechat is compatible with Teams API? cause the github page of plugins says, it's for MS Teams: https://github.com/nyamairi/fluent-plugin-teams

TGPSKI: They should rename this plugin generalized-webhook-format-push-http-plugin. It's  type converting and json packaging log messages to be sent to http webook apis.

The http plugin with fluentd - I spent many days with this trying to get json payloads formatted properly. Did not have much look.

The fluent-plugin-google-chat requires gcp account  access - not available in FedRAMP and also this level of complexity is not needed 
https://github.com/treasuryspring/fluent-plugin-google-chat

This left me to search for other fluentd plugins which could be used in a generalized webhook api fashion - after review the teams one is doing the trick.

janboll: Thank you for the explanation! I really think that's worth a comment :smiley:

PR #2802, opened 2022-09-20
[change-owners] approval-flow dry-run

geoberle:
this PR adds to following functionality to change-owners

* parse basic decision commands from MR comments (`/lgtm`, `/lgtm cancel`, `/hold`, `/hold cancel`) and manage `bot/approved` `bot/hold` and `awaiting-approval` labels when not running in dry-run mode
* support bots as approvers
* honor `tag_on_merge_requests` flag when mentioning users in MR comments
* add `self-servicable` and `not-self-serviceable` labels to MRs to indicate if all changes are approvable by tenants (a.k.a. covered by change-types)
* manage bot-approved, bot-hold
* write MR decision state command if not running in dry-run mode
* add new integration parameter `--change-type-processing-mode` and supported `limited` and `authorative`
* make `changes.context.when` to optional and react according to the semantics described in https://github.com/app-sre/qontract-schemas/pull/248

### details on change type processing mode

change-owners acts only on bundle content and is therefore unaware if something changed in an app-interface MR that can't be covered by change-types, e.g. .env, scripts, ... 
in such situations the integration is started with `--change-type-processing-mode limited` which prevents the integration to act on the MR. even if all bundle changes would be self-servicable via change-types, the `not-self-servicable` label is added.

if the MR contains only changes that are reflected in the bundle, ´--change-type-processing-mode authorative` is provided to the integration to let it know that it is safe to act.

theoretically we could have queried the changed paths of the MR and figure it out within the integration itself but that would imply that change-owners would need to know details about the app-interface repo structure, which is best to avoid to keep the flexibility in the repo layout. it will be `select-integrations.py` responsibility to set this flag appropriately.

Depends on https://github.com/app-sre/qontract-schemas/pull/248
Must be promoted along with https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/48000

### DIFF
--- a/reconcile/change_owners.py
+++ b/reconcile/change_owners.py
@@ -8,7 +8,7 @@ import logging
 import traceback
 import re
 
-from reconcile.utils.output import print_table
+from reconcile.utils.output import format_table
 from reconcile.utils import gql
 from reconcile.gql_definitions.change_owners.queries.self_service_roles import RoleV1
 from reconcile.gql_definitions.change_owners.queries.change_types import (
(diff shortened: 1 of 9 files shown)


reconcile/change_owners.py
```diff
@@ -705,36 +757,220 @@ def run(
...
+        commenter = c["username"]
+        for line in c.get("body", "").split("\n"):
+            if line == DecisionCommand.APPROVED.value:
+                decisions_by_users[commenter].approve = True
+            if line == DecisionCommand.CANCEL_APPROVED.value:
+                decisions_by_users[commenter].approve = False
+            if line == DecisionCommand.HOLD.value:
+                decisions_by_users[commenter].hold = True
+            if line == DecisionCommand.CANCEL_HOLD.value:
+                decisions_by_users[commenter].hold = False
```
chassing, line 879: Use elif? Do you want to support multiple commands in one comment?

reconcile/change_owners.py
```diff
@@ -705,36 +757,220 @@ def run(
...
+    diff_decisions = []
+    for c in changes:
+        for d in c.diffs:
+            dc = ChangeDecision(file=c.fileref, diff=d, decision=Decision())
+            diff_decisions.append(dc)
+            for change_type_context in d.covered_by:
+                for approver in change_type_context.approvers:
+                    if approver_decisions[approver.org_username].approve:
+                        dc.decision.approve |= True
+                    if approver_decisions[approver.org_username].hold:
```
chassing, line 895: KeyError possible? Is `approver.org_username` always in `approver_decisions`?

reconcile/change_owners.py
```diff
@@ -705,36 +757,220 @@ def run(
...
+            if line == DecisionCommand.CANCEL_APPROVED.value:
+                decisions_by_users[commenter].approve = False
+            if line == DecisionCommand.HOLD.value:
+                decisions_by_users[commenter].hold = True
+            if line == DecisionCommand.CANCEL_HOLD.value:
+                decisions_by_users[commenter].hold = False
+    return decisions_by_users
+
+
+def decide_on_changes(
```
chassing, line 883: Please add a `__doc__` here. What does this method?

reconcile/change_owners.py
geoberle: `decisions_by_users` is a `defaultdict(Decision)`

reconcile/change_owners.py
geoberle: the dict is a `defaultdict(Decision)`

reconcile/change_owners.py
geoberle: this replicates the current behaviour found in `saas-file-owners`. while `change-owners` runs in sort-of-a-sry-run mode next to `saas-file-owners` i would like their comment parsing behaviour to be the same, so we can compare results.

in general you have a point: most likely it does not make a lot of sense to have multiple (conflicting) commands in one comment.

reconcile/change_owners.py
geoberle: but you are right. within this context this might be very well a normal dict and KeyErrors might occure. it is somehow hidden that it is a defaultdict. for future safety i will rework this

maorfr (approved):

reconcile/change_owners.py
geoberle: [ad45422](https://github.com/app-sre/qontract-reconcile/pull/2802/commits/ad4542237036e32e505231c7099d9af42ca95634)

PR #2803, opened 2022-09-20
Introduce target_clusters to prometheus_rules_tester

lechuk47:
APPSRE-6316

Schemas: https://github.com/app-sre/qontract-schemas/pull/249/files

### DIFF
--- a/reconcile/prometheus_rules_tester.py
+++ b/reconcile/prometheus_rules_tester.py
@@ -221,40 +221,55 @@ def run_test(test):
     return test
 
 
-def get_rule_files_from_jinja_test_template(template):
-    """Parse test template to get prometheus rule files paths"""
-    rule_files = []
+def get_data_from_jinja_test_template(
+    template: str, desired_lists: list[str]
+) -> dict[str, Any]:
+
+    # Sort the list to allow the comparison with parsed_lists later
+    desired_lists.sort()
+
+    data: dict[str, Any] = {k: [] for k in desired_lists}
+
     try:
         parsed = yaml.safe_load(template)
-        rule_files = parsed["rule_files"]
+        return {k: v for k, v in parsed.items() if k in desired_lists}
+
     except Exception:
         # The jinja template is not valid yaml :(
-        # A poor man's approach is used to get the rule files
+        # A poor man's approach is used to get the required data
         # Let's assume people will follow the examples and use yaml lists as:
         # rule_files:
         # - file
         # - file
-        in_rule_files = False
-        in_re = re.compile(r"^rule_files:\s*$")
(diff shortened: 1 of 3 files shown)


reconcile/prometheus_rules_tester.py
```diff
@@ -275,6 +306,8 @@ def check_prometheus_tests(tests, rules, thread_pool_size, settings):
     for path, test in tests.items():
         test_to_run = {"path": path}
 
+        target_clusters = get_target_clusters_from_jinja_test_template(test)
+
```
rporres, line 310: can we make something nicer to avoid parsing the same file twice? e.g. a function that returns both the target clusters and the rule files?

rporres, line 310: Since clusters are simple strings, we should check they actually correspond to real clusters and fail if the cluster does not exist. I don't think we can do that while bundling because refs are not resolved in resources, so we will need to do it here.

rporres (commented):
I really like the idea of adding the `target_clusters`. A couple of suggestions/comments.

reconcile/prometheus_rules_tester.py
```diff
@@ -275,7 +290,22 @@ def check_prometheus_tests(tests, rules, thread_pool_size, settings):
...
+        )
+        target_clusters = data["target_clusters"]
+        rule_files = data["rule_files"]
+
+        missing_target_clusters = [c for c in target_clusters if c not in clusters]
+        if len(missing_target_clusters) > 0:
+            failed_tests.append(
+                {
+                    **test_to_run,
+                    "check_result": f"There are missing clusters in target_clusters list: {missing_target_clusters}",
```
rporres, line 304: missing or non-existing?

reconcile/prometheus_rules_tester.py
```diff
@@ -297,6 +327,11 @@ def check_prometheus_tests(tests, rules, thread_pool_size, settings):
 
         try:
             for cluster, namespaces in rules[rule_files[0]].items():
+                if len(target_clusters) > 0 and cluster not in target_clusters:
+                    logging.info(
```
rporres, line 331:
```suggestion
                    logging.debug(
```
I think we tend to be more minimal on the INFO lines...

rporres (approved):
minor nits

reconcile/prometheus_rules_tester.py
lechuk47: changed to debug.

reconcile/prometheus_rules_tester.py
lechuk47: Changed to non-existent.

PR #2807, opened 2022-09-22
[query-validator] add option to validate resources with queries

maorfr:
related to https://issues.redhat.com/browse/APPSRE-4709

depends on https://github.com/app-sre/qontract-schemas/pull/255

query-validator currently only validates that queries are executed successfully.

with the addition of resources, we can do everything we already can in openshiftResources (jinja2 for example). with this addition, we can validate the actual data, and not only the query results.


patjlm (approved):
LGTM ❤️

PR #2809, opened 2022-09-22
fix list comprehension returning empty dict when no accounts

jfchevrette:
We want a list comprehension, not a list with a dict comprehension

The current behavior if no resources or account are present is this returns a list that contains an empty dict `[{}]` which is not what we want

The change will return an empty list if no resources are found, otherwise a list of dicts

### DIFF
--- a/reconcile/terraform_cloudflare_resources.py
+++ b/reconcile/terraform_cloudflare_resources.py
@@ -156,7 +156,7 @@ def run(
         QONTRACT_INTEGRATION,
         QONTRACT_INTEGRATION_VERSION,
         QONTRACT_TF_PREFIX,
-        [{"name": name for name in cf_clients.dump()}],
+        [{"name": name} for name in cf_clients.dump()],
         working_dirs,
         thread_pool_size,
     )


reconcile/terraform_cloudflare_resources.py
```diff
@@ -156,7 +156,7 @@ def run(
         QONTRACT_INTEGRATION,
         QONTRACT_INTEGRATION_VERSION,
         QONTRACT_TF_PREFIX,
-        [{"name": name for name in cf_clients.dump()}],
+        [{"name": name} for name in cf_clients.dump()],
```
steveteahan, line 159: nit: maybe something to consider in the future, does it even make sense to get this far if there aren't any accounts? Or should we error earlier?

steveteahan (approved):

reconcile/terraform_cloudflare_resources.py
jfchevrette: I thought the same, but I have another fix to make later that will also address this

PR #2811, opened 2022-09-23
app interface settings gql

chassing:
Generate GQL classes for `AppInterfaceSettings_v1`

- add common `VaultSecret` fragments
- add app_interface_settings_gql support to SecretReader
- introduce `get_app_interface_settings_class` as convinent method similar to `get_app_interface_settings`


reconcile/gql_definitions/common/app_interface_settings.py
```diff
@@ -0,0 +1,244 @@
...
+    BaseModel,
+    Extra,
+    Field,
+    Json,
+)
+
+from reconcile.gql_definitions.fragments.vault_secret import VaultSecret
+from reconcile.gql_definitions.fragments.vault_secret import VaultSecret
+from reconcile.gql_definitions.fragments.vault_secret import VaultSecret
+from reconcile.gql_definitions.fragments.vault_secret import VaultSecret
```
chassing, line 21: Duplicated `VaultSecret` imports are a bug in https://github.com/app-sre/qenerate/. I will address later

chassing: Build is failing because of this

fishi0x01 (commented):
Thank you for this nice refactor! :heart: 

This is a very interesting case, because it shows a generated class which is shared across many integrations. I know that @steveteahan proposed the use of `Protocols` here, i.e., `class AppInterfaceSettingsV1(Protocol)`. ~~This opens up the question, if all generated classes should inherit `Protocol`, which could easily be adjusted in qenerate~~ (thought mistake - generated classes stay as they are). Here is some more detailed explanation https://gitlab.cee.redhat.com/service/dev-guidelines/-/merge_requests/134/diffs

reconcile/blackbox_exporter_endpoint_monitoring.py
```diff
@@ -27,6 +27,8 @@ def run(
 ) -> None:
     # verify that only allowed blackbox-exporter modules are used
     settings = queries.get_app_interface_settings()
+    if not settings:
```
steveteahan, line 30: nit: could we change `get_app_interface_settings()` to have a kwarg `raise_if_empty=False` and set it to true for the cases that we want an exception to be raised? Also, would there be any value in changing the exception to be called something like `AppInterfaceEmptySettingsError` in case it's helpful to handle that different than a general `Exception`?

reconcile/gql_definitions/examples/queries/ocm.py
```diff
@@ -15,13 +15,9 @@
...
 )
 
-from reconcile.gql_definitions.examples.fragments.vault_secret import VaultSecret
-from reconcile.gql_definitions.examples.fragments.vault_secret_version import (
-    VaultSecretVersion,
-)
-from reconcile.gql_definitions.examples.fragments.vault_secret_path import (
-    VaultSecretPath,
-)
+from reconcile.gql_definitions.fragments.vault_secret import VaultSecret
```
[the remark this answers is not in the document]
chassing: Stupid mistake 🤦 will move it back

reconcile/queries.py
```diff
@@ -117,6 +45,16 @@ def get_app_interface_settings():
...
 
 
+def get_app_interface_settings_class() -> Optional[AppInterfaceSettingsV1]:
+    """Returns App Interface settings as GQL class"""
+    gqlapi = gql.get_api()
+    settings = app_interface_settings_query(query_func=gqlapi.query).settings
+    if settings:
+        # assuming a single settings file for now
+        return settings[0]
+    return None
```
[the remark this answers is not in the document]
chassing: Let's raise an exception, this will make the code much more readable

reconcile/utils/secret_reader.py
```diff
@@ -18,7 +21,10 @@ class SecretNotFound(Exception):
 class SecretReader:
     """Read secrets from either Vault or a config file."""
 
-    def __init__(self, settings: Optional[Mapping] = None) -> None:
+    def __init__(
+        self,
+        settings: Optional[Union[Mapping, AppInterfaceSettingsV1]] = None,
```
steveteahan, line 26: The MR for best practices when using data query classes is still[ in progress](https://gitlab.cee.redhat.com/service/dev-guidelines/-/merge_requests/134/diffs) and I've done a poor job of contributing to it. With that being said, anything it `/utils/` feels like a `Inside shared functions` case from that doc. I understand the fragments are meant to minimize having many versions of gql data classes that all need to be supported by our libraries, but it's still coupled to both gql and the current implementation of how we generate data classes.

I'll contribute a better example for `Protocols` to our docs, but I wanted to get your thoughts on something like:

```python
# Open to better naming, many Protocol examples in the PEP use the "Supports" prefix when checking
# for the existence of a certain method, but this could also be HasVaultToggle or anything else.
# This protocol would be defined in the same module that uses it.
class SupportsVault(Protocol):
    vault: bool

# And then your __init__ changes to:
settings: Optional[Union[Mapping, SupportsVault]] = None,
```
The primary benefit being that you've now defined exactly what your code is looking for and the type definition would only ever need to change if your code changes. If some subtle variation of `AppInterfaceSettingsV1` is added (I know we'd prefer this doesn't happen, but it could), then there's no need to change this code as long as `vault` is defined and is a `bool`.

reconcile/utils/secret_reader.py
```diff
@@ -32,6 +38,16 @@ def vault_client(self):
             self._vault_client = VaultClient()
         return self._vault_client
 
+    @property
+    def vault_enabled(self):
```
steveteahan, line 42: Can we add a `bool` return type definition?

reconcile/utils/secret_reader.py
```diff
@@ -32,6 +38,16 @@ def vault_client(self):
...
+        if not self.settings:
+            return False
+
+        if isinstance(self.settings, Mapping):
+            return self.settings.get("vault")
```
steveteahan, line 47: If settings is a mapping, but doesn't have a `vault` key, then this could return `None`? Is that expected or if not we could change this to `get("vault", False)`?

reconcile/utils/secret_reader.py
chassing: sure 🤦

reconcile/blackbox_exporter_endpoint_monitoring.py
chassing: I'll refactor the raise of the exception into `queries.get_app_interface_settings` and `queries.get_app_interface_settings_class`. this makes the code much clearer on the usage site. And yes I'll create a separate exception class for it

reconcile/utils/secret_reader.py
chassing: Make total sense to me, and I see the pros and cons of both ways. Let's use the `Protocol` pattern, for now, which seems to be the better one :) TBH, I don't have a strong opinion on that, and I won't die on that hill ;)

reconcile/utils/secret_reader.py
chassing: good catch!!!

fishi0x01 (approved):
:tada:

reconcile/queries.py
```diff
@@ -21,100 +26,34 @@
 """
 
 
-def get_secret_reader_settings() -> Optional[Mapping[str, Any]]:
+def get_secret_reader_settings() -> Mapping[str, Any]:
```
steveteahan, line 29: nit: typically the idea is accept general inputs (`Mapping`), but be more specific on outputs (`dict`). This is covered here: https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#type-hints

Not going to block on that change.

steveteahan (approved):

chassing:
I close this PR without merging it because we should use integration-specific settings queries instead of this general one.

PR #2812, opened 2022-09-23
Reduce number of classes

fishi0x01:
We can leverage GQL fragments to reduce the number of generated classes in `service_dependencies`.

Reviewer note: the only thing that was manually changed are the `.gql` files. The rest was auto-generated


chassing (approved):
LGTM

geoberle (approved):

PR #2817, opened 2022-09-23
Use get_partition for IAM policies

cubismod:
[APPSRE-6076](https://issues.redhat.com/browse/APPSRE-6076)

Resolve an issue in FedRamp where IAM policies are resolving to the `aws` partition rather than `aws-us-gov` partition which is causing terraform-resources to fail reconcilation.


bkez322:
LGTM

bkez322 (approved):

bkez322:
Upon closer look,`${policy}` should have been `{policy}`. My bad for missing that.

PR #2820, opened 2022-09-26
Types for jumphosts

fishi0x01:
This is a great example of how to leverage GQL variables and fragments on jumphosts:

- use GQL query variables instead of python string templating
- use GQL fragment for common jumphost fields -> this can be used later for other jumphost queries


chassing (approved):
🎉 LGTM

PR #2822, opened 2022-09-27
remove SecretReader dependency from SmtpClient

chassing:
* add GQL smtp config classes
* adapt email_sender


reconcile/queries.py
```diff
@@ -3069,3 +3072,11 @@ def get_app_metadata(app_path: str) -> dict:
 def get_blackbox_exporter_monitoring_provider() -> dict:
     gqlapi = gql.get_api()
     return gqlapi.query(BLACKBOX_EXPORTER_MONITORING_PROVIDER)["providers"]
+
+
+def get_smtp_client_settings() -> SmtpSettingsV1:
+    if _settings := smtp_client_config_query(query_func=gql.get_api().query).settings:
+        if not _settings[0].smtp:
+            raise Exception("settings.smtp missing")
```
steveteahan, line 3080: Should have some general missing configuration exception?

reconcile/requests_sender.py
```diff
@@ -49,7 +49,18 @@ def get_encrypted_credentials(credentials_name, user, settings):
...
+    smtp_credentials = get_smtp_credentials(
+        secret_reader=SecretReader(settings=settings), secret=smtp_settings.credentials
+    )
+    smtp_client = SmtpClient(
+        host=smtp_credentials.server,
+        port=smtp_credentials.port,
+        username=smtp_credentials.username,
+        password=smtp_credentials.password,
+        mail_address=smtp_settings.mail_address,
+        timeout=smtp_settings.timeout or 30,
```
steveteahan, line 62: nit: `30` could become `DEFAULT_SMTP_TIMEOUT` since it's been used a couple times?

reconcile/test/test_utils_smtp_client.py
```diff
@@ -12,36 +10,23 @@ def patch_env(monkeypatch):
...
-        "reconcile.utils.secret_reader.SecretReader.read_all",
-        return_value={
-            "server": smtpd.hostname,
-            "port": smtpd.port,
-            "username": smtpd.config.login_username,
-            "password": smtpd.config.login_password,
-            "require_tls": smtpd.config.use_ssl,
-        },
-        autospec=True,
+    return SmtpClient(
```
steveteahan, line 19: Less `patch` usage is always welcome!

reconcile/utils/smtp_client.py
```diff
@@ -3,24 +3,55 @@
...
+            "path": secret.path,
+            "field": secret.field,
+            "format": secret.q_format,
+            "version": secret.version,
+        }
+    )
+    return SmtpCredentials(**data)
+
+
+class SmtpClient:
```
steveteahan, line 38: As we discussed, this is much better because it doesn't rely on the gql implementation at all. Thanks for the example!

reconcile/utils/smtp_client.py
```diff
@@ -3,24 +3,55 @@
...
+    def __init__(
+        self,
+        host: str,
+        port: int,
+        username: str,
+        password: str,
+        mail_address: str,
+        timeout: int = 30,
+    ) -> None:
+        self.host = host
```
steveteahan, line 48: nit: this could go either way ultimately. I think if we have many args (think 6+) then it does start to make sense to have dataclass(es) to combine the configs that are logically related. In this case you already have `SmtpCredentials`, so is there any reason not to just make `__init__`:

```python
def __init__(mail_address: str, credentials: SmtpCredentials, timeout: int = 30)
```

I don't think there's anything wrong with it as-is other than that we go through the trouble of building the object in all of the usage, and don't use it:

```python
        smtp_credentials = get_smtp_credentials(
            secret_reader=SecretReader(settings=queries.get_secret_reader_settings()),
            secret=smtp_settings.credentials,
        )
        smtp_client = SmtpClient(
            host=smtp_credentials.server,
            port=smtp_credentials.port,
            username=smtp_credentials.username,
            password=smtp_credentials.password,
            mail_address=smtp_settings.mail_address,
            timeout=smtp_settings.timeout or 30,
        )
```

When I used this example originally, I was thinking that we'd have a parameter per setting if we weren't going to bother adding a data class for this use case (because it really could go either way).

steveteahan (commented):
LGTM - mostly nits and minor changes to address.

reconcile/queries.py
```diff
@@ -9,13 +9,16 @@
 
 from jinja2 import Template
 
+from reconcile.gql_definitions.common.smtp_client_settings import (
+    query as smtp_client_config_query,
+    SmtpSettingsV1,
+)
```
fishi0x01, line 15: I am wondering if we should start to cut `queries.py` into smaller files, mainly because I am concerned about duplicate names of generated objects.

An example of what I mean can be seen here: https://github.com/fishi0x01/qontract-reconcile/blob/demo-solution/reconcile/queries.py#L11-L20

```python
from reconcile.gql_definitions.cluster_peering.queries.cluster_peering import (
    ClusterV1,
    DEFINITION as CLUSTER_PEERING_QUERY,
    query as cluster_peerings_query
)
from reconcile.gql_definitions.clusters.clusters import (
  ClusterV1 as OtherClusterV1,
  DEFINITION as CLUSTERS_MINIMAL_QUERY,
  query as minimal_cluster_query,
)
```

Multiple functions return lists of `ClusterV1`, but from different queries. Maybe we should make a module `typed_queries` and have smaller files in it, each dedicated to a single query, in order to have smaller scopes and avoid naming collisions? Curious of your thoughts here.

reconcile/queries.py
chassing: 🤦 I wanted to change before pushing it.

reconcile/queries.py
chassing: I like this idea!

reconcile/utils/secret_reader.py
```diff
@@ -15,6 +15,13 @@ class SecretNotFound(Exception):
     pass
 
 
+class SupportsSecret(Protocol):
```
rporres, line 18: I guess this deservers a docstring 🙏

reconcile/utils/smtp_client.py
```diff
@@ -3,24 +3,55 @@
...
+    password: str
+    # require_tls: bool - currently not in use
+
 
-        self._client: Union[smtplib.SMTP, None] = None
+def get_smtp_credentials(
+    secret_reader: SecretReader, secret: SupportsSecret
+) -> SmtpCredentials:
+    """Retrieve SMTP credentials from config or vault."""
+    # This will change later when SecretReader fully supports 'SupportsSecret'
```
rporres, line 26: At this time, the usage of `SupportsSecret` is rather obscure 😅 Can you elaborate somewhere how it is going to be used. Maybe in the protocol definition :)

rporres (commented):
This decoupling looks fantastic! <3

fishi0x01 (approved):
:heart: Amazing!

rporres (approved):

PR #2825, opened 2022-09-27
glitchtip integration

chassing:
This implements the qontract-reconcile [glitchtip](https://gitlab.com/glitchtip/glitchtip-backend) integration.

Depends on: https://github.com/app-sre/qontract-schemas/pull/274
Dev-data: https://gitlab.cee.redhat.com/app-sre/app-interface-dev-data/-/merge_requests/19
Prod-data and documentation: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/49579
Ticket: [APPSRE-6260](https://issues.redhat.com/browse/APPSRE-6260)
Design Doc: [Glitchtip](https://gitlab.cee.redhat.com/cassing/app-interface/-/blob/master/docs/app-sre/design-docs/glitchtip.md)

Notes:
* Development was against a local glitchtip server
* Glitchtip HTTP rest-client uses httpretty server with response fixtures
* Glitchtip integration uses httpretty server with response fixtures
* Integration also tested against glitchtip staging instance

Consider to use [qontract-development-cli](https://github.com/chassing/qontract-development-cli) to easily run this integration and review this PR.

```shell
$ qd profile create --app-interface YOUR_LOCAL_PATH_TO/app-interface --app-interface-pr 49579 --qontract-schemas-pr 274 --qontract-reconcile-pr 2825 --integration-name glitchtip --integration-extra-args '' glitchtip-pr-check
```


reconcile/utils/glitchtip/client.py
```diff
@@ -0,0 +1,196 @@
...
+    if links.get("next", {}).get("results", "false") == "true":
+        return links["next"]["url"]
+    return None
+
+
+class GlitchtipClient:
+    def __init__(self, host: str, token: str) -> None:
+        self.host = host
+        # todo timeout
+        self._session = requests.Session()
```
janboll, line 25: What about using a thread-local variable for the session? Could make this implementation thread-safe.

reconcile/glitchtip/README.md
```diff
@@ -0,0 +1,148 @@
+<font size=24> GlitchTip </font>
```
janboll, line 1: I assume the idea is to have a domain-specific module. I find this idea very interesting. How would this look for utils that are shared? Have you planned on restructuring the modules?

The current structure would suggest moving reconciler.py into `utils`. I would prefer to see it that way.

reconcile/utils/glitchtip/client.py
```diff
@@ -0,0 +1,196 @@
...
+        self._session.headers.update(
+            {
+                "Authorization": f"Bearer {token}",
+                "Content-Type": "application/json",
+            }
+        )
+
+    @retry()
```
janboll, line 33: Did you think about using a Custom HTTPAdatapter with a Retry setting? That way you could make the retries configurable in the client.

reconcile/glitchtip/reconciler.py
```diff
@@ -0,0 +1,278 @@
...
+    ) -> list[Project]:
+        """Reconcile organization projects."""
+        organization_projects = list(current_projects)
+        for project in set(current_projects).difference(desired_projects):
+            logging.info(
+                ["delete_project", organization_slug, project.slug, self.client.host]
+            )
+            del organization_projects[organization_projects.index(project)]
+            if not self.dry_run:
+                self.client.delete_project(
```
janboll, line 27: What if deleting a project fails, will the rest of the function run?

reconcile/glitchtip/reconciler.py
```diff
@@ -0,0 +1,278 @@
+import logging
+from typing import Iterable, Sequence
+
+from reconcile.utils.glitchtip import GlitchtipClient, Organization, Project, Team, User
+
+
+class GlitchtipReconciler:
```
janboll, line 7: Why is this a class?

reconcile/glitchtip/README.md
chassing: The idea was the following:
* `reconcile/glitchtip` is the integration and everything which belongs to the integration, e.g., `reconciler.py` should not be used from everywhere else, and therefore, it isn't in `utils`
* `reconcile/utils/glitchtip` is independent of the integration and can be used by other integrations/utils. 
* In general I like submodules instead of bloated python files :)

reconcile/utils/glitchtip/client.py
chassing: I didn't know about custom HTTPAdapter; looks promising! I'll give it a try.

reconcile/glitchtip/reconciler.py
chassing: What do you mean?

reconcile/glitchtip/reconciler.py
chassing: Do you mean to have a `try/except` around?

reconcile/glitchtip/README.md
janboll: I am 100% with you. Just afraid of introducing another standard without considering the other modules and how it would look for them, which could cause this to be a beautiful snowflake :)

reconcile/glitchtip/reconciler.py
janboll: Yes, catching errors, that are "okay" to happen. Wonder if there are more errors, like adding/deleting single users, for instance, that should not cause the integrations to crash. 
Guess there is some dependency between the data that must be solved.

chassing: I keep it as it is because otherwise, it'll hide errors. I think it's better to have a failing integration instead of error messages somewhere that nobody reads. In this case, if the project deletion fails, then most properly, all other HTTP requests will fail too.

reconcile/glitchtip/reconciler.py
chassing: Yes, you're right; it can be both. I prefer classes over methods :)

rporres (commented):
Looks great, impressing a thorough job. A lot of material but clear to follow. I've left a few comments/questions here and there.

reconcile/utils/glitchtip/models.py
```diff
@@ -0,0 +1,102 @@
...
+    pk: Optional[int] = Field(None, alias="id")
+    name: str
+    slug: str = ""
+    projects: list[Project] = []
+    teams: list[Team] = []
+    users: list[User] = []
+
+    @root_validator
+    def slugify(cls, values):  # pylint: disable=no-self-argument
+        values["slug"] = values.get("slug") or slugify(values.get("name"))
```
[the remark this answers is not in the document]
chassing: Validators are used to validate values for sure but can be also used to manipulate user inputs :) e.g. imaging you need an always uppercase string, a validator can do that. Have a look at the first validator in the [offical documentation ](https://pydantic-docs.helpmanual.io/usage/validators/)

rporres (approved):
Ship it!
```
                   ______|\
               _.-|AppSRE|:
              <___|.-----||
                     .---''---,
                      ;..__..'    _...
                    ,'/  ;|/..--''    \
                  ,'_/.-/':            :
             _..-'''/  /  |  \    \   _|/|
            \      /-./_ \;   \    \,;'   \
            ,\    / \:  `:\    \   //    `:`.
          ,'  \  /-._;   | :    : ::    ,.   .
        ,'     ::   /`-._| |    | || ' :  `.`.)
     _,'       |;._:: |  | |    | `|   :    `'
   ,'   `.     /   |`-:_ ; |    |  |  : \
   `--.   )   /|-._:    :          |   \ \
      /  /   :_|   ;`-._;   __..--';    : :
     /  (    ;|;-./_  _/.-:'o |   /     ' |
    /  , \._/_/_./--''/_|:|___|_,'        |
   :  /   `'-'--'----'---------'          |
   | :     O ._O   O_. O ._O   O_.      ; ;
   : `.      //    //    //    //     ,' /
 ~~~`.______//____//____//____//_______,'~
           //    //~   //    //
    ~~   _//   _//   _// ~ _//     ~
  ~     / /   / /   / /   / /  ~      ~~
       ~~~   ~~~   ~~~   ~~~
```

chassing:
such an awesome drawing :)

PR #2830, opened 2022-09-28
APPSRE-6255: Different way to get ServiceAccount's token

OliviaHY:
### Why:
Now that we are on Openshift 4.11( Kubernetes 1.24) which means that 
`oc serviceaccounts get-token` does not work anymore ([see this post for detail explanation](https://itnext.io/big-change-in-k8s-1-24-about-serviceaccounts-and-their-secrets-4b909a4af4e0).

Our openshift-serviceaccount-tokens integration need to use a different method to find the token.

### What:
* Filter all secrets in the namespace to find the one created with the service account, since now the secret is no longer discoverable through 'oc sa get-token' command. 
* Remove base64 encoding since the new method is getting the encoded verion.
* Remove unused variable.

### Validation:
* Created SA manually in OpenShift 4.11 cluster and verified that 1. token still created under a similar pattern but`oc sa get-token` not longer work with error msg "error: could not find a service account token for service account "xyz""
* Use the open source app interface targeting `appint-ex-01` cluster and validated the integration can run as before.

### DIFF
--- a/reconcile/openshift_serviceaccount_tokens.py
+++ b/reconcile/openshift_serviceaccount_tokens.py
@@ -87,7 +87,9 @@ def fetch_desired_state(namespaces: list[dict], ri: ResourceInventory, oc_map: O
                 )
                 if cur:
                     for token in sa_token_list:
-                        if token["data"]["token"] == cur.body.get("data", {}).get("token"):
+                        if token["data"]["token"] == cur.body.get("data", {}).get(
+                            "token"
+                        ):
                             sa_token = token["data"]["token"]
             except KeyError:
                 logging.log(
--- a/reconcile/openshift_serviceaccount_tokens.py
+++ b/reconcile/openshift_serviceaccount_tokens.py
@@ -1,4 +1,3 @@
-import base64
 import logging
 import sys
 
@@ -16,7 +15,7 @@ QONTRACT_INTEGRATION = "openshift-serviceaccount-tokens"
 QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
 
 
-def construct_sa_token_oc_resource(name, sa_name, sa_token):
+def construct_sa_token_oc_resource(name, sa_token):
     body = {
         "apiVersion": "v1",
         "kind": "Secret",
@@ -24,11 +23,23 @@ def construct_sa_token_oc_resource(name, sa_name, sa_token):
         "metadata": {
             "name": name,
         },
-        "data": {"token": base64.b64encode(sa_token).decode("utf-8")},
+        "data": {"token": sa_token},
     }
     return OR(
         body, QONTRACT_INTEGRATION, QONTRACT_INTEGRATION_VERSION, error_details=name
     )
+
+
+def get_token_for_service_account(service_account, tokens):
+    result = []
+    for token in tokens:
+        if (
+            token["metadata"]["annotations"]["kubernetes.io/service-account.name"]
+            == service_account
+            and service_account + "-token" in token["metadata"]["name"]
+        ):
+            result.append(token)
+    return result
 
 
 def fetch_desired_state(namespaces, ri, oc_map):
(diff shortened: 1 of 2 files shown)


reconcile/openshift_serviceaccount_tokens.py
```diff
@@ -42,7 +42,8 @@ def fetch_desired_state(namespaces, ri, oc_map):
             logging.log(level=oc.log_level, msg=oc.message)
             continue
         for sat in namespace_info["openshiftServiceAccountTokens"]:
-            sa_name = sat["serviceAccountName"]
+            # sa_name = sat["serviceAccountName"]
+            sa_name = "builder"
```
BumbleFeng, line 46: This is for testing, right?

reconcile/openshift_serviceaccount_tokens.py
```diff
@@ -52,13 +53,25 @@ def fetch_desired_state(namespaces, ri, oc_map):
                     ri.register_error()
                 logging.log(level=oc.log_level, msg=oc.message)
                 continue
-            sa_token = oc.sa_get_token(sa_namespace_name, sa_name)
+
+            all_tokens = oc.get_items(kind="Secret", namespace="yihuang")
```
BumbleFeng, line 57: same here

reconcile/openshift_serviceaccount_tokens.py
```diff
@@ -52,13 +53,25 @@ def fetch_desired_state(namespaces, ri, oc_map):
...
+                    lambda token: sa_name
+                    == token["metadata"]["annotations"][
+                        "kubernetes.io/service-account.name"
+                    ]
+                    and sa_name + "-token" in token["metadata"]["name"],
+                    all_tokens,
+                )
+            )
+            if sa_token_list:
+                sa_token = sa_token_list[0]["data"]["token"]
```
BumbleFeng, line 69: I was just wondering if sa_token_list is sorted in certain order so it can make sure to return the same token.

OliviaHY: That's a good point, I'll add the sorting. My thought was that {sa_name}-token being present was unique enough to pick out the default token, so theoretically this list should always only have one item, but weird thing can happen.

reconcile/openshift_serviceaccount_tokens.py
OliviaHY: Yes. Removed.

reconcile/openshift_serviceaccount_tokens.py
```diff
@@ -16,21 +15,33 @@
...
     )
 
 
+def get_token_for_service_account(service_account, tokens):
+    result = []
+    for token in tokens:
+        if (
+            token["metadata"]["annotations"]["kubernetes.io/service-account.name"]
+            == service_account
+            and service_account + "-token" in token["metadata"]["name"]
```
patjlm, line 39: it would be better to filter on the secret `type` field than on a name suffix:
```
type: kubernetes.io/service-account-token
```

reconcile/openshift_serviceaccount_tokens.py
```diff
@@ -16,21 +15,33 @@
...
-        "data": {"token": base64.b64encode(sa_token).decode("utf-8")},
+        "data": {"token": sa_token},
     }
     return OR(
         body, QONTRACT_INTEGRATION, QONTRACT_INTEGRATION_VERSION, error_details=name
     )
 
 
+def get_token_for_service_account(service_account, tokens):
+    result = []
```
patjlm, line 34: why returning a list and not the first matching token found ?

reconcile/openshift_serviceaccount_tokens.py
geoberle: +1 for what patrick said - that condition should be sufficient and we don't need to rely on name of the secret
i also wonder where a secret with the "-token" suffix comes from

reconcile/openshift_serviceaccount_tokens.py
```diff
@@ -52,13 +63,29 @@ def fetch_desired_state(namespaces, ri, oc_map):
                     ri.register_error()
                 logging.log(level=oc.log_level, msg=oc.message)
                 continue
-            sa_token = oc.sa_get_token(sa_namespace_name, sa_name)
+
+            all_tokens = oc.get_items(kind="Secret", namespace=sa_namespace_name)
+            try:
+                sa_token_list = get_token_for_service_account(sa_name, all_tokens)
+                sa_token_list.sort()
```
patjlm, line 70: the sort should be done according to a field in the token dict

patjlm:
Pushed some changes as this fix is quite urgent now

janboll (approved):

reconcile/openshift_serviceaccount_tokens.py
OliviaHY: I originally used a lambda but was getting a lint error "Cell variable sa_name defined in loop" so pulled that part out as a function. Now looking at it yes that would be a better way to write this function.

reconcile/openshift_serviceaccount_tokens.py
OliviaHY: Ah thanks for point this out I see it now.

PR #2833, opened 2022-09-29
Create trigger for openshift-saas-deploy based on new registry content

maorfr:
part of https://issues.redhat.com/browse/APPSRE-5799

depends on https://github.com/app-sre/qontract-schemas/pull/259

with this new trigger integration, it doesn't matter where, how and when an image is built. once it is built - it will be deployed.

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -281,6 +281,11 @@ class SaasHerder:
                         resource_template_name,
                         target,
                     )
+                    self._validate_image_not_used_with_commit_sha(
+                        saas_file_name,
+                        resource_template_name,
+                        target,
+                    )
                     self._validate_allowed_secret_parameter_paths(
                         saas_file_name,
                         target.get("secretParameters"),
@@ -482,6 +487,23 @@ class SaasHerder:
                 logging.error(
                     f"[{saas_file_name}/{resource_template_name}] "
                     f"upstream used with commit sha: {ref}"
+                )
+                self.valid = False
+
+    def _validate_image_not_used_with_commit_sha(
+        self,
+        saas_file_name: str,
+        resource_template_name: str,
+        target: dict,
+    ):
+        image = target.get("image")
+        if image:
+            pattern = r"^[0-9a-f]{40}$"
+            ref = target["ref"]
+            if re.search(pattern, ref):
+                logging.error(
+                    f"[{saas_file_name}/{resource_template_name}] "
+                    f"image used with commit sha: {ref}"
                 )
                 self.valid = False
 
@@ -1288,8 +1310,8 @@ class SaasHerder:
             url = rt["url"]
             for target in rt["targets"]:
                 try:
-                    # don't trigger if there is a linked upstream job
-                    if target.get("upstream"):
+                    # don't trigger if there is a linked upstream job or container image
+                    if target.get("upstream") or target.get("image"):
                         continue
                     ref = target["ref"]
                     get_commit_sha_options = {"url": url, "ref": ref, "github": github}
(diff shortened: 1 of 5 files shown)


maorfr:
example usage: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/48679

reconcile/utils/saasherder.py
```diff
@@ -1441,6 +1505,93 @@ def get_upstream_jobs_diff_saas_file(
...
+    ) -> list[TriggerSpecContainerImage]:
+        results = threaded.run(
+            self.get_container_images_diff_saas_file,
+            self.saas_files,
+            self.thread_pool_size,
+            dry_run=dry_run,
+        )
+        return list(itertools.chain.from_iterable(results))
+
+    def get_container_images_diff_saas_file(
```
rporres, line 1519: Can you write some documentation explaining what this function does?

reconcile/utils/saasherder.py
```diff
@@ -1441,6 +1505,93 @@ def get_upstream_jobs_diff_saas_file(
...
+                    desired_image_tag = self._get_commit_sha(get_commit_sha_options)
+                    # don't trigger if image doesn't exist
+                    image_registry = f"{image['org']['instance']['url']}/{image['org']['name']}/{image['name']}"
+                    image_uri = f"{image_registry}:{desired_image_tag}"
+                    image_patterns = saas_file["imagePatterns"]
+                    image_auth = self._initiate_image_auth(saas_file)
+                    error_prefix = f"[{saas_file_name}/{rt_name}] {ref}:"
+                    error = self._check_image(
+                        image_uri, image_patterns, image_auth, error_prefix
+                    )
```
rporres, line 1552: What happens if there are two pushes close enough that build for the first is not ended? If I understand correctly, this will continue until the second one finishes to be built...

reconcile/utils/saasherder.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/2833/commits/403ccec8f78666fc20665c269d390212cb4218c2

reconcile/utils/saasherder.py
maorfr: yup, that makes sense. this is consistent with the behavior of `openshift-saas-deploy-trigger-upstream-jobs`.

for the sake of the discussion, let's define the `intermediate` image as the image created by the first push, and the `final` image as the one created by the second (quick) push.

the alternative to this behavior is to:
- wait for the intermediate image to be created
- trigger a deployment
- wait for the final image to be created
- trigger a deployment

depending on the implementation of the entity building the image, there may be situations where builds will be cancelled (common in jenkins) and the intermediate image (the first push before the quick second) will not be created.

on the other hand, the final image will always be created. so i feel this behavior makes sense.

one possible concern is that a team may promote to production the intermediate image, which wasn't deployed, and hence wasn't tested.
in case the image wasn't created as described above we are covered, because the promotion MR checks that the image exists.
in case the image was created, a promotion may be feasible, but we offer and recommend teams to use gated promotions. a tool can not always provide guard rails and automation can not always cover bad practices.

/braindump

rporres: When I started thinking abut this, I had in my mind in a case in which the intermediate had a migration. I couldn't think of a case in which the behaviour could be problematic (yours is also interesting).

The migration case should be covered because migration code is always executed, usually with no effect as the state of the DB is the state in the code. We don't really care if it is executed in the intermediate or the last image, honestly.

Yours is definitely not covered, but it would be honestly weird.

If this is consistent with the other promotion code, let's keep it. I would only document it somehow, since I only learnt about this by reading a few times the code 😅

rporres (approved):
Looks good. See my above comment :)

PR #2838, opened 2022-09-30
change-owners rely on checksum if no diffs found in file

geoberle:
not all changes to datafile survive the transformation from YAML in the app-interface repo to JSON in the qontract-server bundle (newlines, comments). therefore https://github.com/app-sre/qontract-validator/pull/45 adds a checksum to each datafile during bundling so a file change is at least reflected in the checksum and can be used in `change-owners` to drive the decision process about an MR.

if the only change in a file is the checksum, then a change invisble to the integrations is assumed (newlines, comments). as such, the change itself has no impact on the integrations themselves. since it is still a change, that needs to be approved, every person who is a potential approver on such a file, can lgtm such changes.

depends lightly on https://github.com/app-sre/qontract-validator/pull/45 but keep current behaviour if the checksums are not present in the both compared bundles. so it is safe to promote in arbitrary order

### DIFF
--- a/reconcile/change_owners/change_types.py
+++ b/reconcile/change_owners/change_types.py
@@ -8,7 +8,12 @@ import jsonpath_ng
 import jsonpath_ng.ext
 import anymarkup
 
-from reconcile.change_owners.diff import Diff, DiffType, extract_diffs
+from reconcile.change_owners.diff import (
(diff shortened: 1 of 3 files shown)


patjlm (approved):
LGTM! Thanks!

PR #2844, opened 2022-10-04
Fix, environment name can be empty

janboll:
Needed to refactor this, cause the environment name might be empty, thus requiring to loop over all environments. However, the map in collect was aggregateing over namespaces. The namespace Name is not unique, so this could cause integrations to be missing.

This is a follow up fix for this: https://github.com/app-sre/qontract-reconcile/pull/2829


geoberle (approved):
thanks for the review guidance. lgtm

PR #2849, opened 2022-10-06
replace openshift-saas-deploy-wrapper with o-s-d-change-tester

geoberle:
# Why

`saas-file-owners` and `openshift-saas-deploy-wrapper` are closely related integrations. `saas-file-owners` creates an input file about saas-file that have changed that is passed to `openshift-saas-deploy-wrapper`, so it can run `openshift-saas-deploy` for each changed saas-file to verify the content.

when we switch from `saas-file-owners` to `change-owners` for self-service approvals, `saas-file-owners` will not run as part of `app-interface` PR checks anymore, so the input `openshift-saas-deploy-wrapper` expects is not there anymore.

# What changed

This MR introduces a new integration `openshift-saas-deploy-change-tester` which takes over the responsibility to run `openshift-saas-deploy` for each changed saas-file. This integration will be able to detect for itself what saas-files changed without the need of `saas-file-owners` passing information. `openshift-saas-deploy-change-tester` is using the dual-bundle `qontract-server` during app-interface PR checks to find the differences in saas-files.

`saas-file-owners` had another responsibility, namely adding a commit to the checked MR about the saas-file target `refs` that changed (`Diff: github.com/somethingsomething/old_ref....new_ref`). This functionality has been moved to `openshift-saas-deploy-change-tester` as well. This means that `saas-file-owners` and `openshift-saas-deploy-wrapper` can be safely disabled during PR checks (and removed later on) when `change-owners` and `openshift-saas-deploy-change-tester` are running instead.

# Alternatives considered

I considered changing `openshift-saas-deploy-wrapper` instead of creating a new integration but leaving both `saas-file-owner` and `openshift-saas-deploy-wrapper` unchanged brings flexibility about when and how to do the switch to the granular permission model with `change-owners`. It even allows reverting an app-interface instance back to using `saas-file-owners` in case of bugs in `change-owners` without reverting code, simply by enabling or disabling integrations for the PR checks. It also allows one app-interface instance running with `change-owners` while another one still runs `saas-file-owner`.

### DIFF
--- a/reconcile/openshift_saas_deploy_change_tester.py
+++ b/reconcile/openshift_saas_deploy_change_tester.py
@@ -83,10 +83,22 @@ def collect_state(saas_files: list[dict[str, Any]]):
                 parameters.update(saas_file_parameters)
                 parameters.update(resource_template_parameters)
                 parameters.update(target_parameters)
-                secret_parameters = []
-                secret_parameters.extend(saas_file_secret_parameters)
-                secret_parameters.extend(resource_template_secret_parameters)
-                secret_parameters.extend(target_secret_parameters)
+                secret_parameters = {}
+                secret_parameters.update(
+                    {
+                        s.get("name"): s.get("secret")
+                        for s in saas_file_secret_parameters
+                    }
+                )
+                secret_parameters.update(
+                    {
+                        s.get("name"): s.get("secret")
+                        for s in resource_template_secret_parameters
+                    }
+                )
+                secret_parameters.update(
+                    {s.get("name"): s.get("secret") for s in target_secret_parameters}
+                )
                 state.append(
                     {
                         "saas_file_path": saas_file_path,
--- a/reconcile/openshift_saas_deploy_change_tester.py
+++ b/reconcile/openshift_saas_deploy_change_tester.py
@@ -1,0 +1,110 @@
+import sys
+
+from sretoolbox.utils import threaded
+
+from reconcile import queries
+from reconcile.utils.gitlab_api import GitLabApi
+from reconcile.utils import gql
+import reconcile.openshift_saas_deploy as osd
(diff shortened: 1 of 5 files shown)


reconcile/openshift_saas_deploy_change_tester.py
```diff
@@ -0,0 +1,110 @@
...
+    return exit_code
+
+
+def init_gitlab(gitlab_project_id):
+    instance = queries.get_gitlab_instance()
+    settings = queries.get_app_interface_settings()
+    return GitLabApi(instance, project_id=gitlab_project_id, settings=settings)
+
+
+def update_mr_with_ref_diffs(
```
janboll, line 44: Would love to see a test for this function.

janboll:
Thank you for the good documentation on this Pull Request. I have one big concern with this PR and that is the reuse of functions from `saas-file-owners` integration. This creates a tight coupling between these two integrations. To make it worse, both integrations have zero test coverage.
My ask here would be to move the shared functionality into a separate module (utils?) and create tests for them.

reconcile/openshift_saas_deploy_change_tester.py
```diff
@@ -0,0 +1,110 @@
+import sys
```
janboll, line 1: Please consider adding some tests for this integration.

janboll (changes requested):

geoberle:
> My ask here would be to move the shared functionality into a separate module (utils?) and create tests for them. 

You are completely right. I will refactored and create the tests.

janboll (approved):

reconcile/openshift_saas_deploy_change_tester.py
janboll: Thank you for adding tests!

PR #2850, opened 2022-10-07
Gitlab api client instrumentation

bhushanthakur93:
Add instrumentation for GitlabApi client so we can track how many calls we make per integration.

This data will be used to potentially identify any improvements to reduce no of calls we make to gitlab.

We can follow up this MR with even better instrumentation that tells us http method, path, status code etc. This is not the scope of this MR right now.

### DIFF
--- a/reconcile/utils/gitlab_api.py
+++ b/reconcile/utils/gitlab_api.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from typing import Any, Optional, Tuple
 
 from operator import itemgetter, attrgetter
@@ -11,7 +12,7 @@ import urllib3
 
 
 from reconcile.utils.secret_reader import SecretReader
-
+from reconcile.utils.metrics import gitlab_request
 
 # The following line will suppress
 # `InsecureRequestWarning: Unverified HTTPS request is being made`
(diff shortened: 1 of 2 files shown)


reconcile/utils/gitlab_api.py
```diff
@@ -20,6 +21,9 @@
 
 MR_DESCRIPTION_COMMENT_ID = 0
 
+# The default value is there for unit test
+INTEGRATION_NAME = os.getenv("INTEGRATION_NAME", "qontract-reconcile")
```
steveteahan, line 25: If you can't get the integration name, then I'd use an empty string based on the convention used elsewhere.

reconcile/utils/gitlab_api.py
```diff
@@ -217,10 +231,12 @@ def add_group_member(self, group_name, username, access):
         if not self.check_group_exists(group_name):
             logging.error(group_name + " group not found")
         else:
+            gitlab_request.labels(integration=INTEGRATION_NAME).inc()
             group = self.gl.groups.get(group_name)
             user = self.get_user(username)
             access_level = self.get_access_level(access)
             if user is not None:
+                gitlab_request.labels(integration=INTEGRATION_NAME).inc()
```
steveteahan, line 239: It looks like there's another call to `group.members.get` below in the exception handling?

steveteahan (approved):
Couple minor fixes but otherwise LGTM! As we discussed, we can optimize in terms of the best way to handle instrumenting the client long-term once we get some initial data which is the most important thing at the moment.

reconcile/utils/gitlab_api.py
bhushanthakur93: ah yes! I missed that one.

reconcile/utils/gitlab_api.py
bhushanthakur93: Done.

PR #2851, opened 2022-10-10
[change-owners] jsonpath root ($) matches all changes in files

geoberle:
the `$` sign represents the document root in jsonpath. if this is used as a path for change coverage, every change in a document is covered by it.

### DIFF
--- a/reconcile/change_owners/change_types.py
+++ b/reconcile/change_owners/change_types.py
@@ -192,8 +192,9 @@ class BundleFileChange:
                 self.fileref, file_content
             ):
                 for dc in diffs:
-                    covered = str(dc.diff.path).startswith(allowed_path)
-                    if covered:
+                    if change_path_covered_by_allowed_path(
+                        str(dc.diff.path), allowed_path
+                    ):
                         covered_diffs[str(dc.diff.path)] = dc.diff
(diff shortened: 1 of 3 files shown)


reconcile/change_owners/change_types.py
```diff
@@ -351,3 +352,10 @@ class ChangeTypeContext:
     change_type_processor: ChangeTypeProcessor
     context: str
     approvers: list[Approver]
+
+
+JSON_PATH_ROOT = "$"
+
+
+def change_path_covered_by_allowed_path(changed_path: str, allowed_path: str) -> bool:
+    return changed_path.startswith(allowed_path) or allowed_path == JSON_PATH_ROOT
```
cubismod, line 361: I'm probably missing some context here but wouldn't this conditional always return true since every JSONPath object starts with `$`? Is this what you're intending for?

geoberle: that's what i thought as well, that is the reason this situation was handled correctly until now. it seems that is not really mandatory in JSONPath, at least the python library seems to hand out paths without the "$" prefix

cubismod: Oh, interesting. Did not know that!

cubismod (approved):

PR #2855, opened 2022-10-11
[openshift-saas-deploy-change-tester] fix too many openshift-saas-deploy calls

geoberle:
fixes a bug that started a `openshift-saas-deploy` dry-run for every target of a saas file instead of a saas-file/env-name combination.

this resulted in really poor performance.

### DIFF
--- a/reconcile/openshift_saas_deploy_change_tester.py
+++ b/reconcile/openshift_saas_deploy_change_tester.py
@@ -19,10 +19,9 @@ QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
 
 
 def osd_run_wrapper(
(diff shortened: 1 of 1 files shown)


reconcile/openshift_saas_deploy_change_tester.py
```diff
@@ -214,12 +216,12 @@ def run(
     )
 
     available_thread_pool_size = threaded.estimate_available_thread_pool_size(
-        thread_pool_size, len(saas_file_state_diffs)
+        thread_pool_size, len(changed_saas_file_envs)
```
patjlm, line 219: Should we have a `max` here ?

geoberle: lets talk about this in a followup PR

PR #2857, opened 2022-10-12
cluster-upgrade-policies md output with sortable / filterable table

patjlm:
Example output: https://gitlab.cee.redhat.com/-/snippets/5874

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -293,9 +293,9 @@ def cluster_upgrade_policies(
             for i, data in enumerate(sorted_soaking):
                 v, s = data
                 if v == upgrade_version:
-                    sorted_soaking[i] = (v, f'{s} [:dizzy:](a "{upgrade_next_run}")')
+                    sorted_soaking[i] = (v, f'{s} [💫](a "{upgrade_next_run}")')
                 elif v == upgradeable_version:
-                    sorted_soaking[i] = (v, f"{s} :tada:")
+                    sorted_soaking[i] = (v, f"{s} 🎉")
         return ", ".join([f"{v} ({s})" for v, s in sorted_soaking])
 
     for c in desired_state:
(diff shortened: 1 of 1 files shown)


tools/qontract_cli.py
```diff
@@ -305,7 +305,7 @@ def soaking_str(soaking, upgrade_policy, upgradeable_version):
         mutexes = c.get("conditions", {}).get("mutexes") or []
         item = {
             "cluster": cluster_name,
-            "version": parse_semver(version),
+            "version": version,
```
geoberle, line 308: is there a reason you don't parse anymore?

tools/qontract_cli.py
```diff
@@ -385,25 +385,42 @@ def soaking_str(soaking, upgrade_policy, upgradeable_version):
...
+            {"key": "workload"},
+            {"key": "soaking_upgrades"},
+        ]
+        print(
+            json.dumps(
+                {"fields": fields, "items": results, "filter": True, "caption": ""},
+                indent=2,
+            )
+        )
+        print("```")
```
geoberle, line 410: nit: construct a string and do one print only

geoberle (approved):
only nits. LGTM. or should i say 🎉

tools/qontract_cli.py
patjlm: there was no need.. and we were converting back to text for print out.

tools/qontract_cli.py
patjlm: why is that better ?
We also have the explanatory text above.

patjlm: done

patjlm: looks better indeed :)

geoberle: it was just a nit. personally i try to avoid multiple prints if the output should be considered as one unit. i did not want to force that style on you. thank you for addressing it ❤️

PR #2859, opened 2022-10-12
APPSRE-6433 type safety for dashdotdb_slo integration

fishi0x01:
- dashdotdb_slo integration now uses concrete types instead of untyped dicts
- also bumping qenerate version with a small bug fix for some attribute names (https://github.com/app-sre/qenerate/pull/55)

**Test**

Ran locally in --dry-run mode and checked payload with debugger


chassing (approved):
LGTM

PR #2860, opened 2022-10-13
APPSRE-6433 remove logging for dry-run

fishi0x01:
MRs in app-interface are swamped with logs because of https://github.com/app-sre/qontract-reconcile/pull/2859 -> lets fix that


chassing (approved):
LGTM

PR #2862, opened 2022-10-13
APPSRE-6433 proper type when threads join

fishi0x01:
This only affects type-hints -> so it does not make a difference in the running program. Also I checked in the debugger and there is no difference.

However, the thread pool returns a list of lists and that should be noted correctly, even if mypy doesnt catch it in this scenario.


patjlm (approved):
funny this did not get caught earlier..

PR #2866, opened 2022-10-17
OC client: only use Project kind if it is supported

janboll:


reconcile/utils/oc.py
```diff
@@ -529,9 +540,11 @@ def sa_get_token(self, namespace, name):
     def get_api_resources(self):
         # oc api-resources only has name or wide output
         # and we need to get the KIND, which is the last column
-        cmd = ["api-resources", "--no-headers"]
-        results = self._run(cmd).decode("utf-8").split("\n")
-        return [r.split()[-1] for r in results]
+        if not self.api_resources:
```
fishi0x01, line 543: I was initially wondering if there is a scenario where this could lead to stale information somehow. But I came to the conclusion that api resources should rarely change + this is only cached within one reconcile iteration, so it should be good. Not sure if I forgot sth here.

fishi0x01 (approved):

reconcile/utils/oc.py
```diff
@@ -454,7 +456,10 @@ def project_exists(self, name):
             return name in self.projects
 
         try:
-            self.get(None, "Project.project.openshift.io", name)
+            if self.is_kind_supported("Project"):
+                self.get(None, "Project.project.openshift.io", name)
```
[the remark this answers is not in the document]
janboll: Guess that would make sense. But the OCDeprecated and OCNative classes apply that pattern (direct string usage) all over the place. Perhaps we can add it to this refactoring: https://issues.redhat.com/browse/APPSRE-6438

reconcile/utils/oc.py
geoberle, line 543: this is not thread safe and i would like you to have a closer look if that could be an issue. i know that we do a lot of threaded executions when it comes to the openshift-* integrations. as long as we don't process resources from the same cluster in multiple threads, we should be fine

geoberle (approved):
LGTM

reconcile/utils/oc.py
jfchevrette: This did indeed cause issues: https://github.com/app-sre/qontract-reconcile/pull/2870

PR #2867, opened 2022-10-17
APPSRE-4905: Enable faulthandler and register miscellaneous signal.

OliviaHY:
### Why:
We occasionally has stuck integration where there isn't any logging and the state of the integration is unknown. The root cause is specific and require information to troubleshoot.

### What:
Enable faulthandler and register miscellaneous signal.

### Validation:
* Ran q-r terraform-resources locally then send SIGUSR1 was able to get desired traceback.
image

* Ran multi-thread script in the same container image and was able to get desired traceback.
image


reconcile/cli.py
```diff
@@ -376,6 +378,8 @@ def include_trigger_trace(function):
 
 
 def run_integration(func_container, ctx, *args, **kwargs):
+    faulthandler.enable(all_threads=True)
+    faulthandler.register(SIGUSR1)
```
geoberle, line 382: is this meant for manual troubleshooting by sending the event within the running pod?
or do we intend to have some tooling doing that somehow so that the desired output shows up in the pod logs?

reconcile/cli.py
```diff
@@ -376,6 +378,8 @@ def include_trigger_trace(function):
 
 
 def run_integration(func_container, ctx, *args, **kwargs):
+    faulthandler.enable(all_threads=True)
```
kwilczynski, line 381: @OliviaHY, this is nice! A small question here.

Would we need to ensure that the file descriptor the faulthanlder is using is stable?

My Python knowledge might be a bit dated. However, I remember it requiring the underlying file descriptor to be open for the lifespan of the faulthanlder (before it is disabled). Thus, if something does something silly to a standard error output, for example, we might have an issue.

For example, if something calls **dup()**/**dup2()** on the file descriptor the faulthanlder is using, it would stop working correctly.

reconcile/cli.py
OliviaHY: This is for manual troubleshooting when integrations are stuck. We can certainly build tooling around it later.

geoberle: manual is for sure sufficient for now. thanks for adding this functionality 👍

reconcile/cli.py
OliviaHY: That's a great question @kwilczynski I did come across that in their [doc](https://docs.python.org/dev/library/faulthandler.html#issue-with-file-descriptors). 
It is an unlikely situation and when it does happen, all entries writing to stderr will be writing to a different file and not just faulthandler output, which is a much bigger problem for logging and monitoring.

steveteahan:
What do we think about adding some docs about this in a troubleshooting section of this project? Maybe in `docs/troubleshooting.md`? The SOP could link to that so that the documentation and code are located together.

OliviaHY:
> What do we think about adding some docs about this in a troubleshooting section of this project? Maybe in `docs/troubleshooting.md`? The SOP could link to that so that the documentation and code are located together.
Thinking about a little more, I actually like it better in App Interface. One for consistency with other integration failure troubleshooting process, and second it is hard to make the doc/troubleshoot both general enough for this code and also as SOP for a potential pager where there need to be clear steps to follow.

reconcile/cli.py
kwilczynski: @OliviaHY, that would be not great, indeed. :smile: 

I am thinking more about the `sys.stderr`, which on some platforms (or under some circumstances) can be None (see: https://github.com/pyinstaller/pyinstaller/issues/4481), and there is also this:

- https://github.com/microsoft/vscode-python/issues/18346

This is something I stumbled upon once when I was using Visual Studio Code to debug a problem.

I was thinking more along the lines of ensuring that we have a stable file descriptor number for the standard error output, so using `sys.__stderr__`, for example:

```python
  def register_faulthandler(fileobj=sys.__stderr__):
      if fileobj is None:
          return
      try:
          import faulthandler
          faulthandler.disable()
          if (hasattr(faulthandler, 'register') and hasattr(signal, 'SIGUSR1')):
              faulthandler.register(signal.SIGUSR1, file=fileobj, all_threads=True)
      except (ImportError, RuntimeError, AttributeError):
          print('Could not enable faulthandler!')
```

The above example only registers the **SIGUSR1** signal handling. Also, I apologise for my sub-par Python! :bow:

steveteahan:
> > What do we think about adding some docs about this in a troubleshooting section of this project? Maybe in `docs/troubleshooting.md`? The SOP could link to that so that the documentation and code are located together.
> 
> Thinking about a little more, I actually like it better in App Interface. One for consistency with other integration failure troubleshooting process, and second it is hard to make the doc/troubleshoot both general enough for this code and also as SOP for a potential pager where there need to be clear steps to follow.

At a minimum, it would be good to document in the project that faulthandler has been enabled, SIGUSR1 can be used to dump the traceback, etc. The actual SOP itself can be in app-interface.

reconcile/cli.py
```diff
@@ -375,7 +377,26 @@ def include_trigger_trace(function):
     return function
 
 
+def register_faulthandler(fileobj=sys.__stderr__):
+    if fileobj:
+        if not faulthandler.is_enabled():
+            try:
+                faulthandler.enable(file=fileobj)
+                logging.info("faulthandler enabled.")
```
steveteahan, line 385: Does it make sense to move these INFO logs to DEBUG?

steveteahan (approved):
Other than changing the logging level LGTM.

PR #2868, opened 2022-10-18
APPSRE-6295 typed secret-reader for CNA

fishi0x01:
New integration for CNA should be fully typed. To achieve this, we must have a fully typed version of `SecretReader`.
This PR:

- Adds `TypedSecretReader`
- Adds `AppInterfaceSettings` query for vault settings for `TypedSecretReader`.
- Renames SMPT Settings, because a query name must be unique across our codebase.

The core idea is that `SecretReader` and `TypedSecretReader` both have the same base class. The base class is compatible with untyped `VaultClient`. The new `TypedSecretReader` abides to dependency inversion as agreed in recent discussions for new library components.

The new `SecretReader` is to be instantiated at integration `run()` level and injected into any sub-functions / classes, e.g.:

```python
from reconcile.typed_queries.app_interface_vault_settings import query as get_app_interface_vault_settings
from reconcile.utils.secret_reader import create_secret_reader

def run():
    settings = get_app_interface_vault_settings()
    secret_reader = create_secret_reader(settings=settings)
    ...
```

**Test**

- running dashdotdb-dvo locally (as an example that uses the untyped `SecretReader`)
- running new CNA integration locally with `TypedSecretReader` (new integration not part of this PR)

### DIFF
--- a/reconcile/test/test_secret_reader.py
+++ b/reconcile/test/test_secret_reader.py
@@ -1,4 +1,4 @@
-from unittest.mock import MagicMock
+from unittest.mock import create_autospec
 import pytest
 
 import reconcile.utils.secret_reader
@@ -19,13 +19,9 @@ VAULT_READ_ALL_EXPECTED = {"key2": "value2"}
 
 @pytest.fixture
 def vault_mock():
-    read_mock = MagicMock()
-    read_mock.side_effect = [VAULT_READ_EXPECTED] * 100
-    read_all_mock = MagicMock()
-    read_all_mock.side_effect = [VAULT_READ_ALL_EXPECTED] * 100
-    vault_mock = MagicMock(spec=_VaultClient)
-    vault_mock.read = read_mock
-    vault_mock.read_all = read_all_mock
+    vault_mock = create_autospec(_VaultClient)
+    vault_mock.read.side_effect = [VAULT_READ_EXPECTED] * 100
+    vault_mock.read_all.side_effect = [VAULT_READ_ALL_EXPECTED] * 100
     return vault_mock
 
 
@@ -50,7 +46,7 @@ def to_dict(secret):
 
 def test_vault_secret_reader_typed_read(vault_mock, vault_secret):
     vault_secret_reader = VaultSecretReader(vault_client=vault_mock)
-    result = vault_secret_reader.typed_read(vault_secret)
+    result = vault_secret_reader.read_secret(vault_secret)
 
     assert result == VAULT_READ_EXPECTED
     vault_mock.read.assert_called_once_with(to_dict(vault_secret))
@@ -59,7 +55,7 @@ def test_vault_secret_reader_typed_read(vault_mock, vault_secret):
 
 def test_vault_secret_reader_typed_read_all(vault_mock, vault_secret):
     vault_secret_reader = VaultSecretReader(vault_client=vault_mock)
-    result = vault_secret_reader.typed_read_all(vault_secret)
+    result = vault_secret_reader.read_all_secret(vault_secret)
 
     assert result == VAULT_READ_ALL_EXPECTED
     vault_mock.read_all.assert_called_once_with(to_dict(vault_secret))
@@ -99,7 +95,7 @@ def test_vault_secret_reader_raises(vault_mock, vault_secret, patch_sleep):
     vault_secret_reader = VaultSecretReader(vault_client=vault_mock)
 
     with pytest.raises(SecretNotFound):
-        vault_secret_reader.typed_read(vault_secret)
+        vault_secret_reader.read_secret(vault_secret)
 
     vault_mock.read.assert_called_with(to_dict(vault_secret))
     vault_mock.read_all.assert_not_called()
@@ -109,7 +105,7 @@ def test_config_secret_reader_raises(vault_secret, patch_sleep):
     config_secret_reader = ConfigSecretReader()
 
     with pytest.raises(SecretNotFound):
-        config_secret_reader.typed_read(vault_secret)
+        config_secret_reader.read_secret(vault_secret)
 
 
 def test_read_vault_raises(mocker, patch_sleep):
--- a/reconcile/test/test_secret_reader.py
+++ b/reconcile/test/test_secret_reader.py
@@ -1,9 +1,115 @@
+from unittest.mock import MagicMock
 import pytest
 
 import reconcile.utils.secret_reader
 from reconcile.utils import vault
-from reconcile.utils.secret_reader import SecretReader, SecretNotFound
+from reconcile.gql_definitions.fragments.vault_secret import VaultSecret
+from reconcile.utils.secret_reader import (
+    ConfigSecretReader,
+    SecretReader,
+    SecretNotFound,
+    VaultSecretReader,
+)
 from reconcile.utils.vault import _VaultClient
+
+
+VAULT_READ_EXPECTED = {"key": "value"}
+VAULT_READ_ALL_EXPECTED = {"key2": "value2"}
+
+
+@pytest.fixture
+def vault_mock():
+    read_mock = MagicMock()
+    read_mock.side_effect = [VAULT_READ_EXPECTED] * 100
+    read_all_mock = MagicMock()
+    read_all_mock.side_effect = [VAULT_READ_ALL_EXPECTED] * 100
+    vault_mock = MagicMock(spec=_VaultClient)
+    vault_mock.read = read_mock
+    vault_mock.read_all = read_all_mock
+    return vault_mock
+
+
+@pytest.fixture
+def vault_secret():
+    return VaultSecret(
+        path="path/test",
+        field="key",
+        format=None,
+        version=None,
+    )
+
+
+def to_dict(secret):
+    return {
+        "path": secret.path,
+        "field": secret.field,
+        "format": secret.q_format,
+        "version": secret.version,
+    }
+
+
+def test_vault_secret_reader_typed_read(vault_mock, vault_secret):
+    vault_secret_reader = VaultSecretReader(vault_client=vault_mock)
+    result = vault_secret_reader.typed_read(vault_secret)
+
+    assert result == VAULT_READ_EXPECTED
+    vault_mock.read.assert_called_once_with(to_dict(vault_secret))
+    vault_mock.read_all.assert_not_called()
+
+
+def test_vault_secret_reader_typed_read_all(vault_mock, vault_secret):
+    vault_secret_reader = VaultSecretReader(vault_client=vault_mock)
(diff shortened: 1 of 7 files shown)


reconcile/utils/secret_reader.py
```diff
@@ -106,3 +112,66 @@ def read_all(self, secret: Mapping[str, str]):
                 raise SecretNotFound(*e.args) from e
 
         return data
+
+
+class TypedSecretReader(SecretReaderBase):
```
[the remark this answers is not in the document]
fishi0x01: Just had a chat with @geoberle about this. It might make sense to inject the Reader classes you mentioned (VaultSecretReader, ConfigSecretReader) into the run() function. I.e., `create_secret_reader` actually happens at `run_integration()` time.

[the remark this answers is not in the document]
steveteahan: That was basically the motivation behind: https://gitlab.cee.redhat.com/service/dev-guidelines/-/merge_requests/140

If we use dependency injection, then we'd naturally come to the conclusion that `SecretReader` wouldn't be created in the clients themselves and instead would probably end up in `run` and used to read secrets there.

Example:

```python
# In some run() function
secret_reader = create_secret_reader(config)
some_api_token = secret_reader.read(some_secret_path)
# SomeVendorClient being the third-party client released by the vendor (ex. boto3 for AWS)
# This could also be offloaded to some create_* function
vendor_client = SomeVendorClient(some_api_token)
# SomeApiClient is the client we write as a wrapper, in utils/ (ex. AWSApi)
our_wrapper = SomeApiClient(vendor_api_client)
```
Now the client doesn't care about `settings`, `SecretReader`s, or anything else so it's not tightly coupled like what we have today. It also doesn't even need to create the underlying vendor client, so it should be easier to test overall.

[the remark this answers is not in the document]
fishi0x01: Made some changes as suggested above:

- added a `create_secret_reader(settings: Optional[SupportsVaultSettings]) -> TypedSecretReader` function. 
- `TypedSecretReader` is now the base class for `VaultSecretReader` and `ConfigSecretReader`.

@geoberle Staying backwards compatible with `SecretReader` is pretty hard, as it would require to also pass more arguments to the `__init__()` function, such as `settings: SupportsVaultSettings`. I.e., `SecretReader` would become a rather big class wrapping typed and untyped approaches + vault and config reader. For that reason I went now with @steveteahan 's suggestion. Any thoughts on this are very welcome! :bow:

reconcile/utils/secret_reader.py
```diff
@@ -27,8 +27,107 @@ class SupportsSecret(Protocol):
...
+    """
+
+    vault: bool
+
+
+class TypedSecretReader:
+    def read(self, secret: SupportsSecret) -> dict[str, str]:
+        raise NotImplementedError()
+
+    def read_all(self, secret: SupportsSecret) -> dict[str, str]:
```
steveteahan, line 43: I was thinking about this more and it seems that there were 2 things that resulted in the need to change the original `SecretReader`:

1. `__init__` relied on some unnecessary data structure (`settings`)
2. `read` and `read_all` accepted a `dict` of args that was tightly coupled to data that was returned from queries, instead of accepting the args directly

If those two things never happened, then you never would have had to make this change. So, you've already fixed 1, does it make sense to fix 2 as well?

I can see several cases in our code where we're actually creating a `dict` right now simply to conform to the required parameters of the current `SecretReader`:

* https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/openshift_resources_base.py#L439
* https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/dashdotdb_base.py#L24
* https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/dyn_traffic_director.py#L512
* https://github.com/app-sre/qontract-reconcile/blob/master/tools/app_interface_reporter.py

For those use cases, would we need to create some new class that conforms to `SupportsSecret` or start directly creating `VaultSecretV1` objects?

What if we just made this really simple:

```python
def read_all(self, path: str, field: str, version: Optional[int] = None, format: Optional[str] = None) -> dict[str, str]
```

This would support any use case where we have a `dict`, `VaultSecretV1`, or neither.

```python
# some_args is a dict like what we return from queries today
secret_reader.read_all(**some_args)
```

And as a result it's still quite easy to use with Pydantic models like:

```python
# vault_secret could be VaultSecretV1 or other pydantic models
secret_reader.read_all(**vault_secret.dict())
```

I might have missed something, so let me know if this makes sense to you or not. I'm not trying to make this more complicated than it needs to be. I'd like to make sure that these patterns that we're introducing are not going to result in similar issues that result in tight coupling and future refactors.

reconcile/utils/secret_reader.py
```diff
@@ -27,8 +27,107 @@ class SupportsSecret(Protocol):
...
+        try:
+            data = config.read_all(self._secret_to_dict(secret))
+        except config.SecretNotFound as e:
+            raise SecretNotFound(*e.args) from e
+        return data
+
+
+def create_secret_reader(
+    settings: Optional[SupportsVaultSettings],
```
steveteahan, line 115: Thinking about this even more, could this even be `def create_secret_reader(vault: Optional[bool] = None)`. As per the points above, this should support settings whether it's a `dict`, `SupportsVaultSettings`, or if someone has some other use case? It would only require that you make a minor change to the calls:

```python
create_secret_reader(vault=settings.vault)
# Instead of
create_secret_reader(settings)
```

reconcile/utils/secret_reader.py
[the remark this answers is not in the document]
steveteahan: The changes look good to me! I added a few other comments and happy to discuss further as needed.

reconcile/utils/secret_reader.py
geoberle: i am all for a broader set of read methods. SupportSecret as input parameter is nice because it integrates well with dataclasses generated from our schemas. i also support the idea of read_all with dedicated parameters. the old ones with the dict still hold value while we have not moved away from the old ways.

if we can add all those functions with different parameter types to both SecretReader and TypedSecret reader there would be no difference in the Protocol both would implement and they can be used interchangably.

reconcile/utils/secret_reader.py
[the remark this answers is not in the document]
geoberle: i like the changes as well. have a look at the other comment i wrote

reconcile/utils/secret_reader.py
steveteahan: I think it's mostly a trade-off of whether we want a smaller API that works for all things, but the integration requires maybe an additional step (calling `dict()` on pydantic, or even referencing the members directly), or if we want a larger API for the sake of convenience. The additional methods would presumably all need their own tests and could still require future modifications if the way we access data changes again? Versus the method that I proposed would only ever need to change if we fundamentally changed how secrets are accessed.

My preference in order:

1. Try to avoid any care of the data access layer in our `utils` clients (basically what I've shown in the DI examples and `read_all` above)
2. If we're really strongly against above then I'd find some method of "overloading" (Python doesn't really support it, but the methods can have different names) as acceptable, but might cost more in terms of tests, future maintainability, etc?
3. Separate clients based on whether their data and dependencies come from typed (data query classes) or untyped (dict) sources. I'm pretty strongly against this.

reconcile/utils/secret_reader.py
```diff
@@ -27,8 +27,107 @@ class SupportsSecret(Protocol):
...
+class SupportsVaultSettings(Protocol):
+    """
+    SupportsVaultSettings defines all attributes needed from app-interface-settings
+    to instantiate a SecretReader
+    """
+
+    vault: bool
+
+
+class TypedSecretReader:
```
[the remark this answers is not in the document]
fishi0x01: Absolutely no reason! Good catch, thanks! :bow:

reconcile/utils/secret_reader.py
fishi0x01: So I tried to adjust it according to the suggestions:

- SecretReader is now of type TypedSecretReader and can be used interchangeably. 
- TypedSecretReader supports more methods for different inputs: `read(SupportsSecret)`, `read(path, field, format, version)`

reconcile/utils/secret_reader.py
fishi0x01: Adjusted now as you suggest :+1:

reconcile/utils/secret_reader.py
fishi0x01: Went ahead and added some tests already :crossed_fingers:

reconcile/utils/secret_reader.py
```diff
@@ -27,8 +28,198 @@ class SupportsSecret(Protocol):
     q_format: Optional[str]
 
 
-class SecretReader:
-    """Read secrets from either Vault or a config file."""
+class TypedSecretReader(ABC):
```
steveteahan, line 31: I'd really like to avoid sprinkling the code with references to "typed". This feels like an implementation detail that nobody should care about. Likewise for the `typed_read_*` methods can we just come up with some other name that more accurately describes it?

reconcile/test/test_secret_reader.py
```diff
@@ -1,11 +1,117 @@
...
+VAULT_READ_ALL_EXPECTED = {"key2": "value2"}
+
+
+@pytest.fixture
+def vault_mock():
+    read_mock = MagicMock()
+    read_mock.side_effect = [VAULT_READ_EXPECTED] * 100
+    read_all_mock = MagicMock()
+    read_all_mock.side_effect = [VAULT_READ_ALL_EXPECTED] * 100
+    vault_mock = MagicMock(spec=_VaultClient)
```
steveteahan, line 26: Any reason not to use `create_autospec`? tldr: autospec is recursive and I don't think `spec` is? There's more in the `mock` documentation.

steveteahan (commented):
Really just a couple of minor things. Thank you for the work on this!

reconcile/utils/secret_reader.py
steveteahan: The approach seems reasonable to me. I understand the desire for backwards compatibility and while I don't love the idea of the convenience methods for handling with query data classes, if they're accompanied by more general methods like you've done, that's good enough for me. I'll resolve this thread.

reconcile/utils/secret_reader.py
fishi0x01: Renamed those now.

`TypedSecretReader` -> `SecretReaderBase`
`typed_read_*` -> `read_secret`, `read_all_secret`

reconcile/test/test_secret_reader.py
fishi0x01: Good point! Thank you :bow:

steveteahan (approved):

PR #2870, opened 2022-10-18
Revert "OC client: only use Project kind if it is supported"

jfchevrette:
Reverts app-sre/qontract-reconcile#2866

The thread locks that were introduces seem to be causing issues with the kubernetes client, used by the OCNative client. The kubernetes client creates cache files in /tmp that contains the resource kinds, verbs and versions that are available on clusters. Right when these changes were introduced we began seeing errors such as

```
[openshift-serviceaccount-tokens] load cache error: Expecting ',' delimiter: line 1 column 40958 (char 40957)
```

Upon closer inspection the json files are sometimes getting corrupted. This happens almost on every run, in at least one cache file (there is one per cluster targetted during the run)

I'm hoping reverting this MR will clear the errors. If not we can re-introduce the changes

ref.: https://coreos.slack.com/archives/GGC2A0MS8/p1666118145834499?thread_ts=1666113582.323259&cid=GGC2A0MS8

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -4,7 +4,6 @@ import logging
 import os
 import re
 import tempfile
-import threading
 import time
 from contextlib import suppress
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


suzana-nesic (approved):

jfchevrette:
I could not reproduce the issue locally so this is a tough one... But in order to further verify my suspicion I changed an integration to run with only one thread and the error went away

PR #2872, opened 2022-10-19
terraform-cloudflare-resources: pass worker script as file

jfchevrette:
The current approach passes the script as-is as a string and this causes issues with code that has `${foo}` variables which terraform try to interpolate

This changes the code such that the script is written to a temporary file which terraform reads. Terraform does not interpolate the file content


reconcile/utils/terrascript/cloudflare_resources.py
```diff
@@ -70,12 +72,15 @@ def populate(self) -> list[Union[Resource, Output]]:
                 f"for file path {gh_path} at ref {gh_ref}"
             )
         wrk_content = content.decode(encoding="utf-8")
+        tmp_file = tempfile.NamedTemporaryFile(delete=False)
```
steveteahan: Maybe leverage `prefix` to give us better hints of what caused the creation of this if resources were ever leaked?

steveteahan:
The approach seems reasonable to me. I wish we didn't have to write to the filesystem to pull this off, but the alternative of trying to escape special Terraform characters seems worse to me.

reconcile/utils/terrascript/cloudflare_resources.py
jfchevrette: I have refactored these changes to use a terraform variable instead of a temporary file. PTAL again 🙏🏻

jfchevrette:
@steveteahan I have refactored the proposed changes so that instead of using a temporary file we use a terraform input string variable which are not interpolated.

As discussed out of band the temporary file approach as drawbacks, including but not limited to possibly leaking files that are not cleaned up, exposing sensitive data for longer than needed and potential gotchas with threading.

An alternative we discussed was to escape the code similarly to [how they do it in the CDK](https://github.com/hashicorp/terraform-cdk/blob/c837a9de1306a6f3cfbd751f56a1a346386d68c5/packages/cdktf/lib/tfExpression.ts#L41) but after finding we could do it with a variable that approach seem higher maintenance and more error-prone

steveteahan:
This is an interesting approach and nice to no longer need to deal with files. LGTM!

steveteahan (approved):

PR #2873, opened 2022-10-20
APPSRE-4645: Add more RDS pending modifications for which to skip Terraform apply

kwilczynski:
The Pull Request https://github.com/app-sre/qontract-reconcile/pull/2254 brought about an ability to _skip_ an attempt to perform a Terraform **apply** when the RDS database settings had been modified, and the `apply_immediately` property had not been set to `true`.

The initial implementation took care of cases where the underlying database version (the engine) had been changed, for example, as part of routine maintenance or needing a specific version.

This change expands the current implementation beyond only detecting that there has been a change to **`engine_version`** resource attribute on the Terraform's side (or **EngineVersion** as per the AWS' API) and adds several commonly changed [settings](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.DBInstance.Modifying.html#USER_ModifyInstance.Settings) that can be deferred to be applied only during the next maintenance window.

Related: [APPSRE-4645](https://issues.redhat.com/browse/APPSRE-4645)

Signed-off-by: Krzysztof Wilczyński <kwilczynski@redhat.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@
 
 .vscode
 .tox
-.pytest_cache
+.*_cache
 .eggs
 .coverage
 build
--- a/reconcile/test/test_utils_terraform_client.py
+++ b/reconcile/test/test_utils_terraform_client.py
@@ -9,6 +9,12 @@ from reconcile.utils.external_resource_spec import (
 import reconcile.utils.terraform_client as tfclient
 from reconcile.utils.aws_api import AWSApi
 
+from botocore.errorfactory import ClientError
+
+from logging import DEBUG
+
+from operator import itemgetter
+
 
 @pytest.fixture
 def aws_api():
@@ -23,9 +29,7 @@ def tf(aws_api):
     )
 
 
-def test_no_deletion_approvals(aws_api):
-    account = {"name": "a1", "deletionApprovals": []}
-    tf = tfclient.TerraformClient("integ", "v1", "integ_pfx", [account], {}, 1, aws_api)
+def test_no_deletion_approvals(tf):
     result = tf.deletion_approved("a1", "t1", "n1")
     assert result is False
 
@@ -251,3 +255,212 @@ def test__resource_diff_changed_fields(tf):
     )
 
     assert changed == set()
+
+
+@pytest.mark.parametrize(
+    "before,after,response,expected",
+    [
+        pytest.param(
+            {},
+            {},
+            {},
+            False,
+            id="should return false with no data",
+        ),
+        pytest.param(
+            {
+                "availability_zone": "us-east-1a",
+                "allocated_storage": "100",
+            },
+            {
+                "availability_zone": "us-east-1a",
+                "allocated_storage": "100",
+            },
+            {},
+            False,
+            id="should return false with identical before and after",
+        ),
+        pytest.param(
+            {
+                "availability_zone": "us-east-1a",
+                "allocated_storage": 100,
+            },
+            {
+                "availability_zone": "us-east-1a",
+                "allocated_storage": 200,
+            },
+            {"DBInstances": [{}]},
+            False,
+            id="should return false with no database instance",
+        ),
+        pytest.param(
+            {
+                "availability_zone": "us-east-1a",
+                "allocated_storage": 100,
+            },
+            {
+                "availability_zone": "us-east-1a",
+                "allocated_storage": 200,
+            },
+            {"DBInstances": [{"PendingModifiedValues": {}}]},
+            False,
+            id="should return false for database instance without pending changes",
+        ),
+        pytest.param(
+            {
+                "availability_zone": "us-east-1a",
+                "apply_immediately": True,
+            },
+            {
+                "availability_zone": "us-east-1a",
+                "apply_immediately": False,
+            },
+            {
+                "DBInstances": [
+                    {
+                        "PendingModifiedValues": {
+                            "AllocatedStorage": 200,
+                        }
+                    }
+                ]
+            },
+            False,
+            id="should return false with data that does not match",
+        ),
+        pytest.param(
+            {
+                "availability_zone": "us-east-1a",
+                "allocated_storage": 100,
+            },
+            {
+                "availability_zone": "us-east-1a",
+                "allocated_storage": 200,
+            },
+            {
+                "DBInstances": [
+                    {
+                        "PendingModifiedValues": {
+                            "AllocatedStorage": 100,
+                        }
+                    }
+                ]
+            },
+            False,
+            id="should return false with data that is old",
+        ),
+        pytest.param(
+            {
+                "availability_zone": "us-east-1a",
+                "apply_immediately": False,
+                "allocated_storage": 100,
+            },
+            {
+                "availability_zone": "us-east-1a",
+                "apply_immediately": True,
+                "allocated_storage": 200,
+            },
+            {
+                "DBInstances": [
+                    {
+                        "PendingModifiedValues": {
+                            "AllocatedStorage": 200,
+                        }
+                    }
+                ]
+            },
+            False,
+            id="should return false with attribute that is not allowed",
+        ),
+        pytest.param(
+            {
+                "availability_zone": "us-east-1a",
+                "allocated_storage": 100,
+                "engine_version": "11.13",
+            },
+            {
+                "availability_zone": "us-east-1a",
+                "allocated_storage": 200,
+                "engine_version": "11.14",
+            },
+            {
+                "DBInstances": [
+                    {
+                        "PendingModifiedValues": {
+                            "AllocatedStorage": 200,
+                            "EngineVersion": "11.14",
+                        }
+                    }
+                ]
+            },
+            True,
+            id="should return true with valid data that matches",
+        ),
+    ],
+)
+def test__can_skip_rds_modifications(aws_api, tf, before, after, response, expected):
+    aws_api.describe_rds_db_instance.return_value = response
+
+    actual = tf._can_skip_rds_modifications(
+        account_name="a1",
+        resource_name="test-database-1",
+        resource_change={"before": before, "after": after},
+    )
+
+    assert actual == expected
+
+
+def test__can_skip_rds_modifications_with_client_error(aws_api, tf, caplog):
+    aws_api.describe_rds_db_instance.side_effect = ClientError(
+        {
+            "Error": {
+                "Code": "DBInstanceNotFound",
+                "Message": "DBInstance test-database-1 not found.",
+            }
+        },
+        "DescribeDBInstances",
+    )
+
+    caplog.set_level(DEBUG)
+
+    actual = tf._can_skip_rds_modifications(
+        account_name="a1",
+        resource_name="test-database-1",
+        resource_change={
+            "before": {
+                "availability_zone": "us-east-1a",
+                "allocated_storage": "100",
+            },
+            "after": {
+                "availability_zone": "us-east-1a",
+                "allocated_storage": "200",
+            },
+        },
+    )
+
+    assert not actual
+    assert [itemgetter(1, 2)(v) for v in caplog.record_tuples] == [
+        (DEBUG, "Resource test-database-1 changes in Terraform: ['allocated_storage']"),
+        (DEBUG, "Resource does not exist: test-database-1"),
+    ]
+
+
+def test__can_skip_rds_modifications_with_exception(aws_api, tf):
+    with pytest.raises(Exception) as error:
+        aws_api.describe_rds_db_instance.side_effect = Exception("a-test-exception")
+
+        tf._can_skip_rds_modifications(
+            account_name="a1",
+            resource_name="test-database-1",
+            resource_change={
+                "before": {
+                    "availability_zone": "us-east-1a",
+                    "allocated_storage": "100",
+                },
+                "after": {
+                    "availability_zone": "us-east-1a",
+                    "allocated_storage": "200",
+                },
+            },
+        )
+
+    assert "a-test-exception" in str(error.value)
--- a/reconcile/utils/terraform_client.py
+++ b/reconcile/utils/terraform_client.py
@@ -6,7 +6,7 @@ from datetime import datetime, timedelta
 from collections import defaultdict
 from threading import Lock
 from dataclasses import dataclass
-from typing import Iterable, Mapping, Any, Optional
+from typing import Iterable, Mapping, Any, Optional, cast
 
 from python_terraform import Terraform, IsFlagged, TerraformCommandError
 from sretoolbox.utils import retry
@@ -21,6 +21,8 @@ from reconcile.utils.external_resource_spec import (
 import reconcile.utils.lean_terraform_client as lean_tf
 
 from reconcile.utils.aws_api import AWSApi
+
+from botocore.errorfactory import ClientError
 
 ALLOWED_TF_SHOW_FORMAT_VERSION = "0.1"
 DATE_FORMAT = "%Y-%m-%d"
@@ -270,14 +272,13 @@ class TerraformClient:  # pylint: disable=too-many-public-methods
                 if (
                     action == "update"
                     and resource_type == "aws_db_instance"
-                    and self._is_ignored_rds_modification(
+                    and self._can_skip_rds_modifications(
                         name, resource_name, resource_change
                     )
                 ):
                     logging.debug(
-                        f"Not setting should_apply for {resource_name} because the "
-                        f"only change is EngineVersion and that setting is in "
-                        f"PendingModifiedValues"
+                        f"Resource {resource_name} contains pending changes that "
+                        f"can be skipped, should_apply will not be set."
                     )
                     continue
                 with self._log_lock:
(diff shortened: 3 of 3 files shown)


reconcile/utils/terraform_client.py
```diff
@@ -531,34 +531,48 @@ def cleanup(self):
...
-        changed_terraform_args = [
+        changed_resource_arguments = [
             key for key, value in before.items() if value != after.get(key)
         ]
-        if (
-            len(changed_terraform_args) == 1
-            and "engine_version" in changed_terraform_args
-            and self._aws_api is not None
-        ):
+        if len(changed_resource_arguments) > 0 and self._aws_api is not None:
```
geoberle, line 562: nit: you could rely on python truthyness
```suggestion
        if changed_resource_arguments and self._aws_api:
```

geoberle (commented):
from a high level perspective it looks good.
since i don't know the datastructures returned by AWS in detail, i would hope @steveteahan  can spare a minute to review.

reconcile/utils/terraform_client.py
```diff
@@ -531,34 +531,48 @@ def cleanup(self):
...
-            ]
-            if (
-                "EngineVersion" in pending_modified_values
-                and pending_modified_values["EngineVersion"]
-                == resource_change["after"]["engine_version"]
-            ):
-                return True
-        return False
+            pending_modified_values = cast(
+                dict[str, str], response["DBInstances"][0]["PendingModifiedValues"]
```
janboll, line 568: Is it guaranteed, there will always be one instance in the response?
And is the field PendingModifiedValues always set?

maorfr:
:heart:

reconcile/utils/terraform_client.py
```diff
@@ -531,34 +531,48 @@ def cleanup(self):
         for _, wd in self.working_dirs.items():
             shutil.rmtree(wd)
 
-    def _is_ignored_rds_modification(
+    def _get_pending_rds_modification(
```
janboll, line 534: I know there is 0 test coverage for this method, but why leave it that way?  :) 
Since you are changing it drastically, could you please add a test for this method?

reconcile/utils/terraform_client.py
kwilczynski: Done. Thank you!

reconcile/utils/terraform_client.py
kwilczynski: @janboll, the describe_rds_db_instance() is a thin wrapper around Boto's describe_db_instances() where we are passing the DBInstanceIdentifier argument with our RDS database identifier (acting as a filter) alongside the specific region where the database resides - as the database identifier has to be unique under a particular AWS account in a given region, there can be only one or zero databases for which the information will be returned.

We also assume that it has to exist (at least in the Terraform state at a given time) as we are using a resource name (the RDS database identifier) obtained from the Terraform state.

However, it's better to be safe and sorry, as they say. Thus I will add better response handling should there be empty replies and such.

Good call!

reconcile/utils/terraform_client.py
kwilczynski: @janboll, it was @geoberle who gave me a cautionary state, so to speak, when I suggested mocking AWS API response. Nonetheless, tests should be added, rightfully so.

I will refactor the code a bit to make it easier to test. Please, bear with me. :)

kwilczynski: @janboll, this turned out to be more involved to test, especially to do it properly - for instance, moto does not seem to include the **PendingModifiedValues** attribute as part of its RDS database mocked object.

This would require much refactoring, which would delay the value this change aims to bring. And we already have two RDS instances reporting different types of pending changes in our reconciliation log (which generates a lot of unwanted noise).

Additionally, @geoberle did a quick review and raised a question about why we even make an API call to AWS, given that Terraform does it too. We then parse pending information it provides - this is an excellent question. I will talk about this with @steveteahan, if he has a minute, to better understand some of the ideas behind the initial implementation. But this is also out of scope here.

reconcile/utils/terraform_client.py
```diff
@@ -531,34 +531,62 @@ def cleanup(self):
...
+            except (KeyError, IndexError):
+                raise RuntimeError(f"Expected RDS database {resource_name} to exist")
+            if pending_modified_values is None:
+                # The PendingModifiedValues attribute might not be included as
+                # part of the RDS database instance object. However, we expect
+                # it to be included here, especially since, at this point,
+                # Terraform also claims a pending change for the underlying
+                # resource. Thus if we can't agree on AWS API vs Terraform
+                # state, it's better to raise to signify a potential problem.
+                raise RuntimeError(
```
kwilczynski, line 580: **To the reviewer:**

The chances of this occurring are minuscule, albeit non-zero. However, if this does happen, it would indicate that there is something wrong with the state Terraform currently holds and the future changes that are planned to be executed.

reconcile/utils/terraform_client.py
kwilczynski: Done. Thank you!

kwilczynski:
**To the reviewer:**

This has been manually tested using **dry-run** against a tenant account that currently has two RDS databases reporting pending changes (deferred to the next maintenance window) of a different type, as per:

```
[2022-10-21 17:25:28] [WARNING] [terraform_client.py:log_plan_diff:240] - Not setting should_apply for [REDACTED] because the PendingModifiedValues contains pending modifications: DBInstanceClass
```

```
[2022-10-21 17:25:28] [WARNING] [terraform_client.py:log_plan_diff:240] - Not setting should_apply for [REDACTED] because the PendingModifiedValues contains pending modifications: AllocatedStorage
```

kwilczynski:
Re-based against master.

janboll (approved):

geoberle (approved):
❤️

reconcile/utils/terraform_client.py
```diff
@@ -531,34 +531,62 @@ def cleanup(self):
...
+            pending_values: list[str] = []
+            for argument in changed_resource_arguments:
+                value = supported_values[argument]
+                if (
+                    value in pending_modified_values
+                    and cast(dict[str, str], pending_modified_values)[value]
+                    == after[argument]
+                ):
+                    pending_values.append(value)
+        return pending_values
```
steveteahan, line 592: This changes the behavior of this method from returning `True` or `False`, indicating that all the changes in `changed_resource_arguments` have already been applied and are accounted for in `PendingModifiedValues` to only now checking if at least 1 is?

To make this more concrete, imagine that I've done the following:

1. I change `EngineVersion` and run `terraform-resources`
2. The AWS API accepted the change and now `EngineVersion` is in `PendingModifiedValues`
3. I now change `AllocatedStorage` and run `terraform-resources`

The expected behavior would be that the `apply` would run because the diff in `changed_resource_arguments` doesn't exactly match the changes queued in `PendingModifiedValues`?

Tell me if I've misunderstood this and I'll re-read the code again rather than you needing to spell it out for me :)

reconcile/utils/terraform_client.py
```diff
@@ -228,22 +228,22 @@ def log_plan_diff(
...
+                if action == "update":
+                    if resource_type == "aws_db_instance":
+                        pending_rds_modifications = self._get_pending_rds_modification(
+                            name, resource_name, resource_change
+                        )
+                        # Ignore RDS modifications that are going to occur during the next
+                        # maintenance window. This can be up to 7 days away and will cause
+                        # unnecessary Terraform state updates until they complete.
+                        if pending_rds_modifications:
+                            logging.warning(
```
steveteahan, line 240: The reconcile channel would show this on each loop unless we either ignore it, or change this to debug. I'm fine with either because the warning could be useful for debugging in the pod logs.

steveteahan:
Thank you for asking about this issue and taking the initiative to jump into it!

reconcile/utils/terraform_client.py
kwilczynski: I will change this to just emit a log line at a debug level as there might be cases where we run and the change had just been completed.

reconcile/utils/terraform_client.py
kwilczynski: @steveteahan, ah yes, doh! I forgot that we stream logs into the channel - for some reason I thought that these logs are going into the container exclusively, and then we push messages to Slack. Sorry! I will drop this change.

kwilczynski: Done. Thank you!

reconcile/utils/terraform_client.py
kwilczynski: @steveteahan, I have updated the pending changes detection logic to allow for incoming intermediate changes.

Assuming that **apply_immediately** is still disabled, any additional new change to the resource on Terraform's side that does not have a corresponding entry in the set we obtain from **PendingModifiedValues**, will enable apply to take place - this should result in the changes to be scheduled on the AWS' side, and then for us to detect a matching set of pending changes on both sides for a subsequent reconciliation run.

Let me know what you think about this approach.

.gitignore
```diff
@@ -4,7 +4,7 @@
 
 .vscode
 .tox
-.pytest_cache
+.*_cache
```
kwilczynski, line 7: **To the reviewer:**

This should cover **.mypy_cache** and **.pytest_cache**.

kwilczynski:
Squashed and re-based against master.

kwilczynski:
> Thank you for asking about this issue and taking the initiative to jump into it!

@steveteahan, thank you for adding this in the first place. :smile:

reconcile/utils/terraform_client.py
kwilczynski: I've added tests that cover the _get_pending_rds_modification(). Aside of that, I've manually tested with in staging and verified using dry-run mode.

kwilczynski:
Re-base against master.

kwilczynski:
Re-based against master.

kwilczynski:
Resolved merge conflicts and re-based against master.

reconcile/utils/terraform_client.py
```diff
@@ -575,34 +577,74 @@ def cleanup(self):
...
         """
+        # A commonly changed RDS database settings that can be applied either
+        # immediately or during the next scheduled maintenance window, see:
+        #  https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.DBInstance.Modifying.html#USER_ModifyInstance.Settings
+        #
+        # Note that changing the RDS database identifier (the database name) is
+        # a complex operation that also involves Terraform performing a resource
+        # replacement and thus it's not included here, see:
+        #   https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_RenameInstance.html
+        supported_values: dict[str, str] = {
```
janboll, line 595: Did you pick the subset of operations? And by which means? I found more operations, where applying immediately would be required on that page, i.e. `DB subnet group`

kwilczynski: @janboll, this is why we only pay attention to things that are sensible and safe to be allowed to postpone a Terraform apply run.

Everything that won't match against the list (of operations we allow to be deferred) will be a fall-through to the apply. Also, if you have pending changes i.e., **AttachedStorage** is included in the **PendingModifiedValues** property, if you decide to add something else alongside i.e., **EngineVersion**, then we will allow this to apply - to make sure that Terraform apply (provided that **apply_immediately** has not been set) can make the API call to ensure that this new change can be scheduled. However, if you then (there are two aforementioned pending changes) decide to apply changes immediately without waiting to the next maintenance window and set **apply_immediately**, then an apply will take place, and things will be applied immediately.

Does this make sense?

reconcile/utils/terraform_client.py
```diff
@@ -575,34 +577,74 @@ def cleanup(self):
...
+                    == after[argument]
+                ):
+                    pending_values.append(argument)
+            # A change to the resource on Terraform's side without being reflected
+            # on the AWS' side should not delay the apply, so that a new change to
+            # the underlying resource can be scheduled and added to the set of
+            # pending changes.
+            if set(changed_resource_arguments) - set(pending_values):
+                pending_values = []
+        return sorted([supported_values[v] for v in pending_values])
```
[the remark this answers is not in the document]
kwilczynski: @janboll, this changes "instance_class" into "DBInstanceClass" as the log message that uses this information mentions **PendingModifiedValues** (as per how AWS documentation would describe relevant properties in the API) rather than Terraform resource attributes.

reconcile/utils/terraform_client.py
janboll: Makes sense, perhaps rename it to `skippable_pending_ops`?

reconcile/utils/terraform_client.py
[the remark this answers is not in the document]
janboll: Right! overlooked that.

reconcile/test/test_utils_terraform_client.py
```diff
@@ -251,3 +251,258 @@ def test__resource_diff_changed_fields(tf):
     )
 
     assert changed == set()
+
+
+def test_get_pending_rds_modification(aws_api, tf):
```
steveteahan, line 256: Thank you for adding these tests! A couple of minor suggestions:

* You can take a look at the pytest documentation for [parametrizing test functions](https://docs.pytest.org/en/6.2.x/parametrize.html). This will run a separate test per input, which has some benefits including that a single failed assert won't mean the rest of your cases don't run, and in some cases can arguably make the tests easier to read.
* There's generally a convention of having separate functions per test case and making the names fairly descriptive. For instance, at a minimum you could break out the exception case and call it `test_get_pending_rds_modifications_raise_on_error`. With that, you'd probably reduce this all to 2 tests, one having many parameters, but very little actual code.

reconcile/utils/terraform_client.py
```diff
@@ -575,34 +577,74 @@ def cleanup(self):
...
             shutil.rmtree(wd)
 
-    def _is_ignored_rds_modification(
+    def _get_pending_rds_modification(
         self, account_name: str, resource_name: str, resource_change: Mapping[str, Any]
-    ) -> bool:
+    ) -> list[str]:
         """
-        Determine whether the RDS resource changes are cases where a terraform apply
+        Determine whether the RDS resource changes are cases where a Terraform apply
```
steveteahan, line 584: Since this method has changed significantly, it might be worth updating the docstring? It seems that the general pattern now is that a falsy value will be returned when apply can't be skipped, but a truthy value, including the names of the affected fields, will be returned when the apply can be skipped?

nit: I think the intent of all of this could possibly be made clearer by calling this `_rds_modifications_can_be_skipped`. The only thing you really lose is the list of `PendingModifiedValues`, but it's a debug log anyways. We could easily put a debug log in this function to get the same data while simplifying the other log message? Then this method just returns a `bool`.

reconcile/utils/terraform_client.py
```diff
@@ -264,22 +266,22 @@ def log_plan_diff(
...
+                    if resource_type == "aws_db_instance":
+                        pending_rds_modifications = self._get_pending_rds_modification(
+                            name, resource_name, resource_change
+                        )
+                        # Ignore RDS modifications that are going to occur during the next
+                        # maintenance window. This can be up to 7 days away and will cause
+                        # unnecessary Terraform state updates until they complete.
+                        if pending_rds_modifications:
+                            logging.debug(
+                                f"Not setting should_apply for {resource_name} because"
```
steveteahan, line 279: We're really skipping because the diff from Terraform only includes the values that have already been applied and are accounted for in `PendingModifiedValues`. Maybe we can make this log message a bit closer to that?

reconcile/test/test_utils_terraform_client.py
```diff
@@ -251,3 +251,258 @@ def test__resource_diff_changed_fields(tf):
     )
 
     assert changed == set()
+
+
+def test__get_pending_rds_modification(aws_api, tf, caplog, capsys):
+    with patch.object(aws_api, "describe_rds_db_instance"):
```
steveteahan, line 257: With `patch` it's generally always good to use autospec unless there's a good reason not to. There's more on this in our style guide: https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-testing-guide/#mock-as-little-as-possible

The short version is that setting autospec will ensure that the mock, returned from `patch`, cannot be called with the wrong number of arguments, or in the case of classes, will ensure that there isn't access to members/methods that don't exist.

reconcile/utils/terraform_client.py
kwilczynski: Done. Thank you!

reconcile/utils/terraform_client.py
kwilczynski: @steveteahan, no worries. It does make sense to make it only return an boolean and keep all the extra log messages internally. Bear with me while I update everything.

reconcile/utils/terraform_client.py
kwilczynski: @steveteahan, makes sense. I am going to update log messages a bit.

reconcile/test/test_utils_terraform_client.py
kwilczynski: @steveteahan, very good pointers! I will refactor the code accordingly.

kwilczynski: Done. Thank you!

reconcile/utils/terraform_client.py
kwilczynski: Done. Thank you!

kwilczynski:
Re-based against master.

fishi0x01 (approved):
Nice! Love the documentation :heart:

fishi0x01:
There is one redundant import making the tests fail

```
reconcile/test/test_utils_terraform_client.py:2:1: F401 'unittest.mock.patch' imported but unused
```

kwilczynski:
Hi @fishi0x01.

> There is one redundant import making the tests fail
> 
> ```
> reconcile/test/test_utils_terraform_client.py:2:1: F401 'unittest.mock.patch' imported but unused
> ```

Fixed. Thank you!

kwilczynski:
Hi @fishi0x01,

> There is one redundant import making the tests fail
> 
> ```
> reconcile/test/test_utils_terraform_client.py:2:1: F401 'unittest.mock.patch' imported but unused
> ```

Fixed. Thank you!

kwilczynski:
Hi @fishi0x01. thank you!

> Nice! Love the documentation heart

This is as per:
- [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html)

Not sure if this is anything that we would want to follow, however had some guidelines on documentation. :smile:

reconcile/test/test_utils_terraform_client.py
kwilczynski: Done. Thank you!

reconcile/utils/terraform_client.py
kwilczynski: Done. Thank you!

reconcile/utils/terraform_client.py
```diff
@@ -575,34 +576,101 @@ def cleanup(self):
         for _, wd in self.working_dirs.items():
             shutil.rmtree(wd)
 
-    def _is_ignored_rds_modification(
+    def _can_skip_rds_modifications(
         self, account_name: str, resource_name: str, resource_change: Mapping[str, Any]
     ) -> bool:
+        """Skip pending RDS modifications.
+
+        Determine whether the RDS resource has pending modifications to the
```
steveteahan, line 584: Thank you for the docstring! I'm going to approve the PR, but wanted to share our style guide on docstrings: https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#docstrings

I don't think we do a great job of making people aware that there's a style guide, so wouldn't certainly be open to your suggestions for making that better for new team members. Also, we could probably build more of these things into CI with automated checks. This one really isn't a major issue, but it's also in theory minimal effort to have our edits use the same docstring style.

steveteahan (approved):

reconcile/utils/terraform_client.py
kwilczynski: @steveteahan, updated to our preferred format.

Thank you!

kwilczynski:
Squashed and re-based. Ready to merge.

PR #2877, opened 2022-10-21
smtp-client secret_reader new function

fishi0x01:
`SecretReader` now supports `SupportsSecret` -> we can use that in smtp-client.


kwilczynski (approved):
Looks good! Let's ship!

geoberle:
nice!

PR #2878, opened 2022-10-21
Kcp compatible oc client

janboll:
Redo https://github.com/app-sre/qontract-reconcile/pull/2871

Fixed bug, where missing resources would fail saas deployments, was caused by missing field `init_api_resources` in OCNative class


reconcile/utils/oc.py
```diff
@@ -529,9 +543,12 @@ def sa_get_token(self, namespace, name):
...
         # and we need to get the KIND, which is the last column
-        cmd = ["api-resources", "--no-headers"]
-        results = self._run(cmd).decode("utf-8").split("\n")
-        return [r.split()[-1] for r in results]
+        with self.api_resources_lock:
+            if not self.api_resources:
+                cmd = ["api-resources", "--no-headers"]
+                results = self._run(cmd).decode("utf-8").split("\n")
+                self.api_resources = [r.split()[-1] for r in results]
+        return self.api_resources
```
patjlm, line 551: we could also ensure here:
`self.init_api_resources = True`

reconcile/utils/oc.py
```diff
@@ -971,21 +987,20 @@ def __init__(
         if server:
             self.client = self._get_client(server, token)
             self.api_kind_version = self.get_api_resources()
+            self.api_resources = self.api_kind_version.keys()
+            self.init_api_resources = True
```
patjlm, line 991: Would it make sense to move that within `self.get_api_resources()` ?

janboll: I would not like to do that, get_api_resources() is only resolving values for  `api_kind_version`. The naming is messed up here. I'd rather change openshiftresources then.

reconcile/utils/oc.py
```diff
@@ -940,7 +957,7 @@ def is_kind_supported(self, kind: str) -> bool:
         if "." in kind:
             # self.api_resources contains only the short kind names
             kind = kind.split(".", 1)[0]
-        return self.api_resources and kind in self.api_resources
+        return self.get_api_resources() and kind in self.get_api_resources()
```
patjlm, line 960: why not `return kind in self.get_api_resources()` ?

reconcile/utils/oc.py
```diff
@@ -971,21 +987,19 @@ def __init__(
         if server:
             self.client = self._get_client(server, token)
             self.api_kind_version = self.get_api_resources()
+            self.api_resources = self.api_kind_version.keys()
```
patjlm, line 990: is this used anymore?

patjlm (approved):
Few questions, but LGTM overall

reconcile/utils/oc.py
janboll: Of course... Let me fix that.

reconcile/utils/oc.py
janboll: Yes, used in `cluster_deployment_mapper.py`

PR #2879, opened 2022-10-21
APPSRE-5975 Log fields and diffs on updates

janboll:
This extends the log message to contain fields updated by a terraform change. Only update actions are considered. Example for a log message, backup retention was 14 and is reduced to 7:
```
[2022-10-21 14:06:08] [INFO] [terraform_client.py:log_plan_diff:362] - ['update', 'app-int-example-01', 'aws_db_instance', 'dev-jboll-rds', {'backup_retention_period': 7}]
```


cubismod (approved):

kwilczynski (approved):
Very nice! Thank you!

PR #2880, opened 2022-10-21
[change-owners] change-type inheritance

geoberle:
# What changes?

The granular permission model based on `change-types` allows fine grained control of what is self serviceable but to really allow the right amount of change for the right people, we would need to define a lot of change-types for individual aspects. Having a lot of similar `change-types` eventually leads to duplication. `change-type` inheritance is an approach to keep duplication at bay.

A `change-type` can inherit change detector configurations (jsonpaths, etc.) from other `change-types`

```yaml
$schema: "/app-interface/change-type-1.yml"

name: a-change-type

inherit:
- $ref: "/another-change-type.yml"
- ...
```

All changes along the inheritance paths will be aggregated into the `ChangeTypeProcessor` representing a `change-type`. The rest of the code is unaware this aggregation happens.

During `change-type` loading, the following conditions are checked

* there are not cycles within the `change-type` inheritance tree
* a `change-type` can only inherit from a `change-type` with the same `context-schema`, or both have no `context-schema`

If any of these conditions are not met, `change-owners` will fail. this serves as a quality gate during PR checks introducing changes to `change-types`.

# How to review

Review commit by commit. The first one does not change anything functionally but refactors the `ChangeTypeProcessor` to keep its inner state more isolated and less dependant on the actual `change-type` representation within `qontract-server`. Feel free to ignore this commit during review.

The second commit 79b84d371d64caf41a19fedd38e923817c9c6615 adapts the `ChangeTypeProcessor` loading process to aggregate inherited change detector configurations. The magic happens here. The hard part of building the graph, finding cycles and finding inheritance linages is offloaded to the amazing https://networkx.org/ library. 

All commits following from here on are tests or minor improvements.

# References

Part of: https://issues.redhat.com/browse/APPSRE-6473
Depends on schema: https://github.com/app-sre/qontract-schemas/pull/278

### DIFF
--- a/reconcile/change_owners/change_types.py
+++ b/reconcile/change_owners/change_types.py
@@ -401,8 +401,7 @@ def init_change_type_processors(
         processors[change_type.name] = build_change_type_processor(change_type)
 
     # detect cycles
-    cycles = list(networkx.simple_cycles(change_type_graph))
-    if cycles:
+    if cycles := list(networkx.simple_cycles(change_type_graph)):
         raise ChangeTypeInheritanceCycleError(
             "Cycles detected in change-type inheritance", cycles
         )
--- a/reconcile/change_owners/change_types.py
+++ b/reconcile/change_owners/change_types.py
@@ -9,7 +9,7 @@ import jsonpath_ng.ext
 import anymarkup
 import jinja2
 import jinja2.meta
-
+import networkx
 
 from reconcile.change_owners.diff import (
     SHA256SUM_FIELD_NAME,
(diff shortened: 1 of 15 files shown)


reconcile/change_owners/change_types.py
```diff
@@ -389,6 +389,51 @@ def build_change_type_processor(change_type: ChangeTypeV1) -> ChangeTypeProcesso
...
+    change_type_graph = networkx.DiGraph()
+    for change_type in change_types:
+        change_type_graph.add_node(change_type.name)
+        for i in change_type.inherit or []:
+            change_type_graph.add_edge(change_type.name, i.name)
+        processors[change_type.name] = build_change_type_processor(change_type)
+
+    # detect cycles
+    cycles = list(networkx.simple_cycles(change_type_graph))
+    if cycles:
```
chassing, line 405: nit but can be written as

```python
if cycles := list(networkx.simple_cycles(change_type_graph)):
```

chassing (approved):
LGTM

reconcile/change_owners/change_types.py
geoberle: it was such a long week that i'm not even ashamed to admit that i was not aware of this syntax :D

geoberle: changed in c28e6fbf5d5836469838817258ef238f4e0d4479

PR #2885, opened 2022-10-24
Standardize OCM integrations

lechuk47:
## Overview 

* Standardize all the OCM integrations to get clusters the same way. Clusters suitability is now checked in a `_cluster_is_compatible` method in all the integrations. Also, cluster filtering is now made within the `run` method in all cases. 

* Enable integration disablement at the cluster level. This is required to early support Hypershift clusters where a lot of OCM administration tasks are not yet in place and will be disabled in the cluster spec.  

Requires: https://github.com/app-sre/qontract-schemas/pull/282

### DIFF
--- a/reconcile/ocm/utils.py
+++ b/reconcile/ocm/utils.py
@@ -1,0 +1,9 @@
+from typing import Mapping, Any
+
+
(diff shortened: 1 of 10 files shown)


geoberle (approved):
LGTM

PR #2887, opened 2022-10-24
[change-owners] change type priority

maorfr:
part of https://issues.redhat.com/browse/APPSRE-6474

depends on https://github.com/app-sre/qontract-schemas/pull/281

with this change, `change-owners` will add a `bot/approved: <priority>` label instead of `bot/approved`. this will indicate the priority of processing of the MR (rebase/merge).


reconcile/change_owners/change_owners.py
```diff
@@ -370,7 +378,7 @@ def run(
         labels = manage_conditional_label(
             labels=labels,
             condition=self_serviceable and approved,
-            true_label=APPROVED,
+            true_label=get_approved_label_by_priority(change_type_processors),
```
geoberle, line 381: `change_type_processors` holds all change-types found in a-i, not just the ones that are active in this MR.

if will find the involved change-types in `changes[*].diff_coverage[*].coverage[*]`

geoberle, line 381: have a look at the implementation of `manage_conditional_label` and think about the following scenario:

* MR is opened with involved two change-types, one `medium` and one `critical`
* the `medium` label is applied
* the tenant changes the MR and removes the change that was covered by the `medium` change-type

reconcile/change_owners/change_types.py
```diff
@@ -380,6 +393,7 @@ def build_change_type_processor(change_type: ChangeTypeV1) -> ChangeTypeProcesso
     """
     ctp = ChangeTypeProcessor(
         name=change_type.name,
+        priority=ChangeTypePriority(change_type.priority),
```
geoberle, line 396: this way a PR would fail during PR checks if there is a priority mismatch between q-s and q-r

fishi0x01 (approved):

PR #2888, opened 2022-10-24
APPSRE-6295 add OCMBaseClient

fishi0x01:
`OCMBaseClient` is a thin layer between the OCM API and more complex OCM clients (e.g., `OCMClient` or `CNAClient`). `OCMBaseClient` can be used by more sophisticated clients to interact with the OCM API. It takes care of authentication and provides basic API operations GET, POST, PATCH, DELETE.

Using `OCMBaseClient` will also ease testing, as it can easily be mocked and passed to the sophisticated client as an input parameter.

### DIFF
--- a/reconcile/utils/ocm_base_client.py
+++ b/reconcile/utils/ocm_base_client.py
@@ -1,0 +1,104 @@
+import logging
+from typing import Any, Mapping, Optional
+from requests import Session, codes
+
+from sretoolbox.utils import retry
+
+
+REQUEST_TIMEOUT_SEC = 60
+
+
+class OCMBaseClient:
+    """
+    Thin client for OCM. This class takes care of authentication
+    and provides methods for GET, POST, PATCH, DELETE to interact with ocm API.
+    """
+
+    def __init__(
+        self,
+        url: str,
+        offline_token: str,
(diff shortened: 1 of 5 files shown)


reconcile/utils/ocm_base_client.py
```diff
@@ -0,0 +1,104 @@
+import logging
+from typing import Any, Mapping, Optional
+from requests import Session, codes
+
+from sretoolbox.utils import retry
+
+
+REQUEST_TIMEOUT_SEC = 60
```
geoberle, line 8: i love the fact, that we have timeouts, i just wonder if 60 seconds is too high considering that the `init_access_token` function uses it while having the `@retry` decorator which does 3 attempts per default.

i know that you just moved this code here so i will accept an "not the scope of this change" comment if you should give it :D

geoberle (approved):
feel free to ignore the one comment i have if you want to go forward.
otherwise LGTM and thanks for this refactor

reconcile/utils/ocm_base_client.py
geoberle: maybe you should ignore my nit. that way this change is functionally neutral! go merge

PR #2893, opened 2022-10-25
APPSRE-6295 add CNA integration

fishi0x01:
:warning: WIP - no need to review yet

Add CNA Integration with initial support for `null-asset`s. `null-asset` does not do anything and is primarily used for testing the CNA API.

As of now CNA does not support life-cycle management and dry-run functionality, i.e., we need to evaluate actual and desired state on client-side. Main part of this integration for now is to properly assemble actual and desired state and conclude additions, deletions and updates based on them. Having a state diff also provides a `dry-run` functionality.

Corresponding schema: https://github.com/app-sre/qontract-schemas/pull/275

To run this in a debugger locally you can use the following qd profile

```
qd profile create --app-interface-pr 50925 --qontract-schemas-pr 275 --qontract-reconcile-pr 2893 --integration-name cna-resources --integration-extra-args "" cna-pr-check
```

Note, this is a very early approach to interact with CNA. Further, CNA is in a very early stage itself. I.e., we expect a lot of refactors and changes here in the near future.


reconcile/cna/integration.py
```diff
@@ -0,0 +1,133 @@
...
+    pass
+
+
+class CNAIntegration:
+    def __init__(
+        self,
+        cna_clients: Mapping[str, CNAClient],
+        namespaces: Iterable[NamespaceV1],
+        desired_states: Optional[Mapping[str, State]] = None,
+        actual_states: Optional[Mapping[str, State]] = None,
```
geoberle, line 41: i like that there is a way to pass in the states. makes testing easier. why not make them mandatory? you could have the `assemble_` functions outside of the class

reconcile/cna/integration.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def assemble_desired_states(self):
+        self._desired_states = defaultdict(State)
+        for namespace in self._namespaces:
+            for provider in namespace.external_resources or []:
+                # TODO: this should probably be filtered within the query already
+                if not isinstance(provider, NamespaceCNAssetV1):
+                    continue
+                for resource in provider.resources or []:
+                    if isinstance(resource, CNANullAssetV1):
```
geoberle, line 56: we discussed this briefly. having a mapper between the types representing CNA modules and the app-interfaces types would enable us to make this generic. but since it is just this one class here, we can do that in a followup PR

reconcile/cna/integration.py
```diff
@@ -0,0 +1,133 @@
...
+            state = State()
+            state.add_raw_data(cnas)
+            self._actual_states[name] = state
+
+    def provision(self, dry_run: bool = False):
+        for provisioner_name, cna_client in self._cna_clients.items():
+            desired_state = self._desired_states[provisioner_name]
+            actual_state = self._actual_states[provisioner_name]
+
+            additions = desired_state - actual_state
```
geoberle, line 75: ❤️

reconcile/cna/assets/null.py
```diff
@@ -0,0 +1,71 @@
...
+    def api_payload(self) -> dict[str, Any]:
+        return {
+            "asset_type": "null",
+            "name": self.name,
+            "parameters": {
+                "addr_block": self.addr_block,
+            },
+        }
+
+    def __eq__(self, other: object) -> bool:
```
geoberle, line 22: would it be an option to use frozen dataclasses to get eq, ne and hash for free? from what i see, you don't update fields right now but create new instances instead, e.g. in `update_from`. so should be ok?

fishi0x01: The `__eq__` method gets overridden to ensure only `name` and `addr_block` are compared. We do not care about comparing ids, herfs, ... . This comes in handy when comparing equality of states: desired state objects (taken from app-interface) do not contain any ids, but current states objects, taken from CNA API do contain ids, still the states can be equal, regardless of their ids. 
The `__hash__` method is a bad side-effect of this implementation. It is required, because we use a custom `__eq__` methods. 

However, I actually do not like this implementation a lot and I am very happy about any suggestion here. But Im not sure how frozen dataclass can solve this.

reconcile/cna/integration.py
```diff
@@ -0,0 +1,133 @@
...
+    query_func = gql.get_api().query
+    cna_provisioners = cna_provisioners_query(query_func).cna_provisioners or []
+    namespaces = namespaces_query(query_func).namespaces or []
+
+    cna_clients = build_cna_clients(
+        secret_reader=secret_reader, cna_provisioners=cna_provisioners
+    )
+
+    integration = CNAIntegration(cna_clients=cna_clients, namespaces=namespaces)
+    integration.assemble_actual_states()
```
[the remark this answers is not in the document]
fishi0x01: Fixed

reconcile/cna/assets/null.py
geoberle: oh i understand. so you would like to compare only those fields that come from app-interface?

reconcile/cna/integration.py
fishi0x01: Fully agree. We can enhance this when we add a second kind of asset.

reconcile/cna/assets/null.py
geoberle: have you tried `dataclass.Field.compare` and `dataclass.Field.hash` ?

fishi0x01: > have you tried dataclass.Field.compare and dataclass.Field.hash ?

Interesting. Looking into it

fishi0x01: Yeah looks much better with `dataclass.Field.compare` and `dataclass.Field.hash`. Thank you! :bow:

geoberle (approved):
feel free to merge. we can iterate in parallel afterwards

PR #2897, opened 2022-10-25
Remove inline user policies

esron:
Second time trying this one.
This was the first one and it was reverted before to test sharded promotions.
https://github.com/app-sre/qontract-reconcile/pull/2656


steveteahan (approved):
Approving pending that we'll fix [APPSRE-6518](https://issues.redhat.com/browse/APPSRE-6518) first and then get this merged and promoted.

PR #2900, opened 2022-10-26
[gitlab-housekeeping] fix stale items handling

maorfr:
gitlab-housekeeping was not handling stale items properly.

this is obviously because i am dumb: #241
i mean, seriously. update_date is the latest note or `now`???

this PR simplifies the logic to not care about notes, but use the item's `updated_at` attribute.


kwilczynski (approved):
Thank you for the fix!

PR #2901, opened 2022-10-26
Add ROSA/Hypershift clusters support

lechuk47:
# Overview 
* Introduce ROSA hosted-cp clusters creation 
* OCM-clusters app-interface updates will now update values when they differ, not only when the AI values are missing. This is needed to recreate clusters automatically when they got deleted by OCM (clusters in ocm-staging are ephemeral) 
* Removed the pydantic validator to set default Rosa AWS role ARN's. Roles must be set. 
* Changed ocm-clusters updates conditions: Rosa hosted-CP clusters do not inform provision_shard nor external_id
 
Requires: https://github.com/app-sre/qontract-schemas/pull/291

### DIFF
--- a/reconcile/utils/ocm.py
+++ b/reconcile/utils/ocm.py
@@ -1,13 +1,16 @@
 from __future__ import annotations
 
 import functools
+import logging
 import random
 import re
 from abc import abstractmethod
 import string
 from typing import Any, Mapping, Optional, Tuple, Union
 
+from reconcile.utils.secret_reader import SecretReader
 import reconcile.utils.aws_helper as awsh
+
 from reconcile.ocm.types import (
     OCMClusterAutoscale,
     OCMClusterNetwork,
@@ -19,8 +22,10 @@ from reconcile.ocm.types import (
     ROSAClusterSpec,
 )
 from reconcile.utils.ocm_base_client import OCMBaseClient
-from reconcile.utils.secret_reader import SecretReader
+
+
 from sretoolbox.utils import retry
+
 
 STATUS_READY = "ready"
 STATUS_FAILED = "failed"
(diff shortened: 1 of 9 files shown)


reconcile/utils/ocm.py
```diff
@@ -414,6 +442,7 @@ def _get_create_cluster_spec(cluster_name: str, cluster: OCMSpec) -> dict[str, A
                     "rosa_creator_arn": cluster.spec.account.rosa.creator_role_arn
                 },
                 "product": {"id": "rosa"},
+                "flavour": {"id": "osd-4"},
```
geoberle, line 445: Can you elaborate on this?

reconcile/utils/ocm.py
```diff
@@ -553,7 +588,7 @@ def _ready_for_app_interface(cluster: dict[str, Any]) -> bool:
 
     def _init_clusters(self, init_provision_shards):
         api = f"{CS_API_BASE}/v1/clusters"
-        clusters = self._get_json(api)["items"]
+        clusters = self._get_json(api).get("items")
```
geoberle, line 591: Just curious. Can `items` be missing?

lechuk47: I think this was a bug since the `_get_json` method supports pagination. `items` is not missing in the API response but the `_get_json` method does not add it if it's empty in the response. 

I think It would be better to return an empty `items` list the same way the API does, but using `get("items")` is use in other methods of the class.

reconcile/utils/ocm.py
lechuk47: Flavours are used to define clusters' infrastructure options such as instance types, disk sizes, etc. 
This value is used by ROSA CLI and it's used in the payload examples I got from the OCM team. Its name is  "Openshift Dedicated 4.x" and is the default one.  

I've tried to remove it and ClustersService uses it as default, so I think it's better not to set the attribute and let CS set the default value here.

geoberle: Thank you!

geoberle (approved):
LGTM!

maorfr:
https://issues.redhat.com/browse/APPSRE-6532

PR #2904, opened 2022-10-28
qenerate update for enum support

fishi0x01:
qenerate now supports GQLEnums https://github.com/app-sre/qenerate/pull/56

This will add the following line to all generated files

```python
from enum import Enum # noqa: F401 # pylint: disable=W0611
```

This is fully backwards-compatible, as it has no effect on the logic.


chassing (approved):
LGTM

PR #2905, opened 2022-10-30
Switch OCM auth from offline token to client secret

maorfr:
part of https://issues.redhat.com/browse/APPSRE-2494

depends on https://github.com/app-sre/qontract-schemas/pull/292

docs: https://www.appsdeveloperblog.com/keycloak-client-credentials-grant-example/

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -632,6 +632,26 @@ def bot_login(ctx, cluster_name):
     print(f"oc login --server {server} --token {token}")
 
 
+@get.command()
+@click.argument("org_name")
+@click.pass_context
+def ocm_login(ctx, org_name):
+    settings = queries.get_app_interface_settings()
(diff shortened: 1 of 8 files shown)


tools/qontract_cli.py
```diff
@@ -632,6 +632,26 @@ def bot_login(ctx, cluster_name):
     print(f"oc login --server {server} --token {token}")
 
 
+@get.command()
+@click.argument("org_name")
```
rporres, line 636: Can you add documentation to this CLI method in the printed help? What does it do, what is `org_name`, etc...

rporres (approved):
LGTM, just a small request.

tools/qontract_cli.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/2905/commits/c4f6cd60f57e9a8cf13148941c48c448f2f93550

PR #2906, opened 2022-10-31
[OCM] add support for CCS clusters

maorfr:
related to https://issues.redhat.com/browse/APPSRE-6517

CCS clusters are OSD clusters created within a customer's AWS account. As such, they do not have storage/LB quota.

This PR adds a handle for a CCS cluster so we can work with such clusters.

### DIFF
--- a/reconcile/ocm/types.py
+++ b/reconcile/ocm/types.py
@@ -96,7 +96,7 @@ class ROSAClusterSpec(OCMClusterSpec):
 
 class OCMSpec(BaseModel):
     path: Optional[str]
-    spec: Union[OSDClusterSpec, ROSAClusterSpec]
+    spec: Union[OSDClusterSpec, ROSAClusterSpec, OCMClusterSpec]
     network: OCMClusterNetwork
     domain: Optional[str]
     server_url: str = Field(None, alias="serverUrl")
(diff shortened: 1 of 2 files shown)


reconcile/ocm/types.py
```diff
@@ -96,7 +96,7 @@ class Config:
 
 class OCMSpec(BaseModel):
     path: Optional[str]
-    spec: Union[OSDClusterSpec, ROSAClusterSpec]
+    spec: Union[OSDClusterSpec, ROSAClusterSpec, OCMClusterSpec]
```
fishi0x01, line 99: I think it might be good to add `Extra.forbid` to the `class Config` https://pydantic-docs.helpmanual.io/usage/model_config/

When passing raw dicts to that class like `OCMSpec(**data)`, then pydantic iterates over the types in order of declaration inside the union and the first fitting wins. I.e., we should try to always have the most specific type first (in this case, it is OSDClusterSpec, because it has more attributes). On top, `extra.Forbid` will make the matching more reliable, as only types which declare ALL given keys are considered.

maorfr: https://github.com/app-sre/qontract-reconcile/pull/2906/commits/9df32e0e83de564750922dadd923238848647163

maorfr: adding it caused failures, and this is likely outside the scope of this PR

fishi0x01 (approved):

PR #2907, opened 2022-10-31
introduce ocm-upgrade-scheduler-org

maorfr:
part of https://issues.redhat.com/browse/APPSRE-6517

depends on https://github.com/app-sre/qontract-schemas/pull/294

this PR introduces a new integration - `ocm-upgrade-scheduler-org`.

this integration managed upgrades in the exact same way as `ocm-upgrade-scheduler`, but it works for OCM orgs defined in app-interface, while the clusters themselves are not managed via app-interface.

this allows external groups to submit MRs to app-interface adding their OCM org with a list of clusters and upgrade policies, and poof! it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ Additional tools that use the libraries created by the reconciliations are also 
                                   OCM.
   ocm-machine-pools               Manage Machine Pools in OCM.
   ocm-upgrade-scheduler           Manage Upgrade Policy schedules in OCM.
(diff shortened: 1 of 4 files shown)


maorfr:
going to merge this without a review to get the capability in place. happy to get post merge reviews!

maorfr:
https://issues.redhat.com/browse/APPSRE-6526

PR #2919, opened 2022-11-02
[ocm-upgrade-scheduler] rely on spec from ocm

maorfr:
with this PR, the upgrade scheduler will not rely on the data found in app-interface for `version` and `channel`, but on the same information returned from the OCM API.

this makes the integration more resilient, in case the version has not been updated (we currently rely on an automation to update the field).

### DIFF
--- a/reconcile/ocm_upgrade_scheduler.py
+++ b/reconcile/ocm_upgrade_scheduler.py
@@ -20,7 +20,9 @@ QONTRACT_INTEGRATION = "ocm-upgrade-scheduler"
 SUPPORTED_OCM_PRODUCTS = [OCM_PRODUCT_OSD]
 
 
-def fetch_current_state(clusters, ocm_map):
+def fetch_current_state(
+    clusters: list[dict[str, Any]], ocm_map: OCMMap
+) -> list[dict[str, Any]]:
     current_state = []
     for cluster in clusters:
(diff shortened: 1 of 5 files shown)


geoberle (approved):
small and clean. nice. lgtm!

reconcile/ocm_upgrade_scheduler.py
```diff
@@ -33,14 +35,18 @@ def fetch_current_state(clusters, ocm_map):
     return current_state
 
 
-def fetch_desired_state(clusters):
+def fetch_desired_state(
+    clusters: list[dict[str, Any]], ocm_map: OCMMap
+) -> list[dict[str, Any]]:
```
geoberle, line 40: i'm just trying to figure out how this works hand in hand with `ocm-clusters`.

`ocm-clusters` creating/managing clusters and setting the `channel` based on the contents of `/openshift/clusters-1.yml` and `ocm-upgrade-policy` consuming that information from the OCM API instead from a cluster file or from the `upgradePolicyClusters[*].spec.channel` of an `/openshift/openshift-cluster-manager-1.yml`

correct?

maorfr: correct.

patjlm (approved):
Thanks!

PR #2920, opened 2022-11-02
[terraform-resources] support records for route53-zone provider

maorfr:
part of https://issues.redhat.com/browse/APPSRE-5707

depends on https://github.com/app-sre/qontract-schemas/pull/299

this PR adds support to specify records on a `route53-zone` provider. these records will be added to the zone.

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -5048,6 +5048,12 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
         zone_tf_resource = aws_route53_zone(zone_id, **values)
         tf_resources.append(zone_tf_resource)
 
+        counts = {}
+        for record in common_values.get("records") or []:
+            self.populate_route53_record(
+                account, values, record, counts, zone_tf_resource
+            )
+
         policy = {
             "Version": "2012-10-17",
             "Statement": [
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -800,9 +800,8 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
                 )
 
     def populate_route53_record(
-        self, zone, record, counts, zone_resource, default_ttl=300
+        self, account_name, zone, record, counts, zone_resource, default_ttl=300
     ):
-        acct_name = zone["account_name"]
         record_fqdn = f"{record['name']}.{zone['name']}"
         record_id = safe_resource_id(f"{record_fqdn}_{record['type'].upper()}")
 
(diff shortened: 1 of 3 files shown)


geoberle (commented):
what would happen if a records for a zone are managed via `/dependencies/dns-zone-1.yml` and the `route53-zone` provider?

steveteahan:
Related to @geoberle's comment, where does the responsibility of `dns-zone-1.yml` start and stop with this change?

maorfr:
a dns-zone-1 is meant for human management. a route53-zone is meant for service management.

that said, there are cases for a service managed zone where very few entries will be managed by humans (LE challenge for example).

steveteahan:
> a dns-zone-1 is meant for human management. a route53-zone is meant for service management.
> 
> that said, there are cases for a service managed zone where very few entries will be managed by humans (LE challenge for example).

I was originally thinking about why they don't just manage the 1-2 records in a `dns-zone-1`, but from quickly looking at the code it doesn't look like that'd be possible because both integrations would try to manage the Route 53 zone.

My only other concern is teams seeing `route53-zone` and now thinking that's how they should manage their DNS records, but I'm sure the doc updates would cover that.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -798,9 +799,46 @@ def populate_additional_providers(self, accounts):
                     assume_role={"role_arn": assume_role},
                 )
 
+    def populate_route53_record(
+        self, account_name, zone, record, counts, zone_resource, default_ttl=300
```
steveteahan, line 803: Can we add type hints?

steveteahan (commented):
LGTM pending the type hints.

maorfr:
don't go ahead and lgtm this because the implementation is wrong :sweat_smile:

steveteahan:
> don't go ahead and lgtm this because the implementation is wrong sweat_smile

I'm not sure if you meant you found something wrong, or you were asking me to not just go ahead and lgtm if I don't believe this is the right move.

Any of the solutions to this problem that I could think of quickly had several trade-offs. The solution in the PR seems perfectly reasonable. The others I was originally thinking about were:

1. Push this back of the tenant to handle. If they're already creating DNS records in their service, why can't they have a special case for "test" records? If many teams have a similar need, this could result in more effort than the pragmatic approach you've taken.
2. Move this use case to `dns-zone-1`, but then we need to figure out how to negotiate the two different integrations trying to manage a single zone. The `dns-zone-1` could get some option to indicate it's an existing zone, but that might just leave us with other patterns that aren't ideal like more "partially-managed" zones.

Really the worst-case I can imagine is that someone completely misses that `dns-zone-1` exists and adds many inline records to `route53-zone` while exposing credentials unnecessarily. Documentation and reviewing changes can help with that. The other thing I've been discussing in the context of the Cloudflare DNS integration is starting to introduce some limits for certain integrations. For instance, we may not want to support 30k DNS records for performance reasons, but 5k might be reasonable. Setting that expectation up front might users in deciding how an integration was intended to be used. In this case, the number of records managed by `route53-zone` seems like it's intended to be very small, so we could document/implement a limit to make that point even clearer.

All of that being said, I think what you have here plus some documentation updates would be perfectly reasonable.

maorfr:
thanks for this great feedback! i will add a schema validation of, say, 2 records.

what i meant above is that i found a problem with the implementation. i think i added code in the wrong place :sweat_smile:

maorfr:
redo in #3319

PR #2922, opened 2022-11-03
[ROSA Hosted clusters] Allow using multiple OCM environments per AWS Account

lechuk47:


reconcile/ocm_clusters.py
```diff
@@ -18,17 +18,50 @@
...
+def _set_rosa_ocm_attrs(cluster: Mapping[str, Any]):
+    """Cluster account (aws) attribute from App-interface differs from the OCMSpec.
+    App-interface's account includes the details for all the OCM environments
+    but the cluster only needs the target OCM environment where it belongs.
+    This method changes the cluster dicctionary to include just those.
+    """
+    cluster_ocm_env_name = cluster["ocm"]["name"]
+
+    rosa = cluster["spec"]["account"]["rosa"]
+    ocm_env_filter = filter(
```
janboll, line 30: Any particular reason you choose filter over list comprehension? I find list comprehensions easier to read and it's more common in qontract-reconcile.

lechuk47: You're right. I don't have any strong preference for one over the other.

reconcile/ocm_clusters.py
```diff
@@ -18,17 +18,50 @@
 QONTRACT_INTEGRATION = "ocm-clusters"
 
 
+def _set_rosa_ocm_attrs(cluster: Mapping[str, Any]):
+    """Cluster account (aws) attribute from App-interface differs from the OCMSpec.
+    App-interface's account includes the details for all the OCM environments
+    but the cluster only needs the target OCM environment where it belongs.
+    This method changes the cluster dicctionary to include just those.
```
rporres, line 25:
```suggestion
    This method changes the cluster dictionary to include just those.
```

reconcile/ocm_clusters.py
```diff
@@ -18,17 +18,50 @@
...
+    rosa = cluster["spec"]["account"]["rosa"]
+    ocm_env = [
+        env
+        for env in rosa["ocm_environments"]
+        if env["ocm"]["name"] == cluster["ocm"]["name"]
+    ]
+
+    if len(ocm_env) != 1:
+        logging.error(
+            "ROSA OCM attributes error: Check the account configuration for ROSA {cluster_ocm_env_name} OCM env"
```
rporres, line 37: where does this `cluster_ocm_env_name` come from? What do we have exactly to check? Can you write the error message from the point of view of somebody that may not know a lot about ROSA and OCM environments?

reconcile/ocm_clusters.py
```diff
@@ -18,17 +18,50 @@
 QONTRACT_INTEGRATION = "ocm-clusters"
 
 
+def _set_rosa_ocm_attrs(cluster: Mapping[str, Any]):
+    """Cluster account (aws) attribute from App-interface differs from the OCMSpec.
```
rporres, line 22: nit 
```suggestion
    """Cluster account (aws) attribute from app-interface differs from the OCMSpec.
```

and below :)

reconcile/ocm_clusters.py
lechuk47: applied

reconcile/ocm_clusters.py
lechuk47: Sorry, I removed the variable because it's not needed anymore. I've also improved the message.

rporres (approved):

PR #2925, opened 2022-11-03
ocm-upgrade-scheduler-org-updater

maorfr:
part of https://issues.redhat.com/browse/APPSRE-6517

depends on https://github.com/app-sre/qontract-schemas/pull/302

in order to keep `upgradePolicyClusters` up-to-date with clusters being created/deleted, we need to define conditions of clusters (ClusterDeployment labels a.k.a external configuration labels) based on which we select an upgrade policy for a discovered cluster.

this PR creates a separate integration that will submit MRs to app-interface to update `upgradePolicyClusters`, so we can keep the logic of `ocm-upgrade-scheduler-org` unchanged. the reason is that the current "static" nature of defining clusters is a no-go for management of OSD-FM fleets.

the OSD FM requirements are more complex compared to a regular customer, which usually has a single OCM org in a single OCM environment with a `statically managed set of clusters.

### DIFF
--- a/reconcile/ocm_upgrade_scheduler_org_updater.py
+++ b/reconcile/ocm_upgrade_scheduler_org_updater.py
@@ -15,7 +15,7 @@ def run(dry_run, gitlab_project_id):
     settings = queries.get_app_interface_settings()
     ocms = queries.get_openshift_cluster_managers()
     for ocm_info in ocms:
-        updates = {}
+        updates = []
         create_update_mr = False
         upgrade_policy_defaults = ocm_info.get("upgradePolicyDefaults")
         if not upgrade_policy_defaults:
@@ -29,6 +29,7 @@ def run(dry_run, gitlab_project_id):
             init_version_gates=True,
         )
         ocm_name = ocm_info["name"]
+        ocm_path = ocm_info["path"]
         ocm = ocm_map[ocm_name]
 
         for ocm_cluster_name in ocm.clusters:
@@ -47,6 +48,13 @@ def run(dry_run, gitlab_project_id):
                         logging.info(
                             ["add_cluster", ocm_name, ocm_cluster_name, default_name]
                         )
+                        item = {
+                            "action": "add",
+                            "path": ocm_path,
+                            "cluster": ocm_cluster_name,
+                            "policy": default_name,
+                        }
+                        updates.append(item)
 
         for up_cluster in upgrade_policy_clusters:
             up_cluster_name = up_cluster["name"]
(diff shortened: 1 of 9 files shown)


reconcile/ocm_upgrade_scheduler_org_updater.py
```diff
@@ -0,0 +1,78 @@
...
+                        create_update_mr = True
+                        logging.info(
+                            ["add_cluster", ocm_name, ocm_cluster_name, default_name]
+                        )
+                        item = {
+                            "action": "add",
+                            "cluster": ocm_cluster_name,
+                            "policy": default["upgradePolicy"],
+                        }
+                        updates.append(item)
```
patjlm, line 56: we should break the loop as well: the first matching default will be taken & we won't risk to have twice the same cluster in upgrade_policy_clusters if there are 2 default entries matching its labels

maorfr: good catch!

https://github.com/app-sre/qontract-reconcile/pull/2925/commits/20decd8fd297f0371486e57aa28c65ed1f53904b

reconcile/utils/mr/ocm_upgrade_scheduler_org_updates.py
```diff
@@ -0,0 +1,77 @@
...
+
+        msg = f"update {ocm_name} upgrade policy clusters"
+        gitlab_cli.update_file(
+            branch_name=self.branch,
+            file_path=ocm_path,
+            commit_message=msg,
+            content=new_content,
+        )
+
+        if not changes:
```
patjlm, line 76: shouldn't this be moved before the `yaml.dump` and `update_file` calls, with an early `return` ?

maorfr: right. i was copying the example of `clusters_updates`, where we add a commit per cluster update in a single MR (may include multiple updates). in this scenario, we are adding a commit per ocm org and submitting the MR per org, so an early cancel would make more sense.

https://github.com/app-sre/qontract-reconcile/pull/2925/commits/d3e4a202199c5cdbe6707a1e6393772d655264e3

patjlm: I think we're missing a `return` statement now also. (or i am not refreshing fast enough...)

patjlm: Ah I did not see `self.cancel` actually raises an exception

patjlm (approved):
Great, thanks!

PR #2931, opened 2022-11-06
remove SKIP_CI

maorfr:
related to https://issues.redhat.com/browse/APPSRE-6532

this PR removes the `gitlab-ci-skipper` integration. we shouldn't skip CI :)

it also removes the already unused mirrors data updater integration, together with it's unused MR spec.


geoberle (approved):
i love a good PR that removes code!

PR #2935, opened 2022-11-07
APPSRE-6539 qenerate nested fragment support

fishi0x01:
The latest qenerate update now adds support for nested fragments. I.e., within a fragment you can reference another fragment and further reduce the number of generated classes.


chassing (approved):
LGTM

PR #2946, opened 2022-11-08
Use new implementation of Image response cache

rporres:
The former implementation of the Image class had a few problems (and was broken for registries not returning `Docker-Content-Digest` header on manifest retrieval). A new implementation has been written and we have used the cache stats provided by it instead of rolling our own InstrumentedCache that was highly dependent of the previous implementation.

See the following PRs for more context on the Image change:

* https://github.com/app-sre/sretoolbox/pull/80
* https://github.com/app-sre/sretoolbox/pull/82

Signed-off-by: Rafael Porres Molina <rporres@gmail.com>


chassing (approved):
LGTM

PR #2950, opened 2022-11-11
sql-query: performance improvements by reducing http calls

chassing:
* use bulk request to fetch "/aws/rds-defaults-1.yml" resources all at once
* use state.ls instead of state.exists for query states

## old
QUERY_COUNT * "get rds default resource" + QUERY_COUNT * (state.exists + state.get)

674 + 674 * 2 = 2022 http calls

## new
"get all rds default resources" + QUERY_COUNT * state.get + state.ls

1 + 674 + 1 = 676 http calls

APPSRE-6546

### DIFF
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -189,6 +189,7 @@ class GqlApi:
         return resources[0]
 
     def get_resources_by_schema(self, schema: str) -> list[dict[str, str]]:
+        """Return all resources (resources_v1) filtered by given schema."""
         query = """
         query Resource($schema: String) {
             resources: resources_v1 (schema: $schema) {
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -188,6 +188,22 @@ class GqlApi:
 
         return resources[0]
 
+    def get_resources_by_schema(self, schema: str) -> list[dict[str, str]]:
+        query = """
+        query Resource($schema: String) {
+            resources: resources_v1 (schema: $schema) {
+                path
+                content
+                sha256sum
+            }
+        }
+        """
+
+        # Do not validate schema in resources since schema support in the
+        # resources is not complete.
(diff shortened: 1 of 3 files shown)


reconcile/utils/gql.py
```diff
@@ -188,6 +188,22 @@ def get_resource(self, path: str) -> dict[str, Any]:
...
+                path
+                content
+                sha256sum
+            }
+        }
+        """
+
+        # Do not validate schema in resources since schema support in the
+        # resources is not complete.
+        resources = self.query(query, {"schema": schema}, skip_validation=True)
```
fishi0x01, line 204: Not sure I understand this. How can we query sth with gql that is not defined in the schema?

chassing: This is a very good question! Actually, I've copied and pasted the original [get_resource](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/gql.py#L166) method and adapted it 🤪

fishi0x01: @rporres do you know hot this is possible?

chassing: Or maybe @geoberle can explain that?

rporres: Hmmm, there's no GraphQL schema for anything under `app-interface/resources`, only JSON schema. GraphQL schema only refers to objects under `data`. Resources are usually free form YAML objects and we only add validation in a few of them.

reconcile/utils/gql.py
```diff
@@ -188,6 +188,22 @@ def get_resource(self, path: str) -> dict[str, Any]:
 
         return resources[0]
 
+    def get_resources_by_schema(self, schema: str) -> list[dict[str, str]]:
```
rporres, line 191: we love docstrings and we are sad when new code doesn't add them 😢

rporres (approved):
This is cool! Please add documentation for the new public methods you're adding 🙏

PR #2956, opened 2022-11-15
Add ocm update recommended version integration

janboll:
Add an integration that updates the recommended version based on configurable weight.

Related story: [APPSRE-6543](https://issues.redhat.com/browse/APPSRE-6543)

### DIFF
--- a/reconcile/ocm_update_recommended_version.py
+++ b/reconcile/ocm_update_recommended_version.py
@@ -79,9 +79,10 @@ def get_updated_recommended_versions(
                 cluster[k].spec.version for k in cluster if k in cluster_workload
             ]
             rv_update = rv_workload[0]
-            rv_update["recommendedVersion"] = recommended_version(
-                versions, high_weight, majority_weight
-            )
+            if len(versions) > 0:
+                rv_update["recommendedVersion"] = recommended_version(
+                    versions, high_weight, majority_weight
+                )
             rv_updated.append(rv_update)
         else:
             raise ValueError("Expecting one recommended Version per workload!")
--- a/reconcile/ocm_update_recommended_version.py
+++ b/reconcile/ocm_update_recommended_version.py
@@ -4,6 +4,7 @@ from reconcile import queries, mr_client_gateway
 from reconcile.utils.mr.ocm_update_recommended_version import (
     CreateOCMUpdateRecommendedVersion,
     UpdateInfo,
+    WorkloadRecommendedVersion,
 )
 from reconcile.utils.ocm import OCMMap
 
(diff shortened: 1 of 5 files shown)


reconcile/ocm_update_recommended_version.py
```diff
@@ -0,0 +1,123 @@
...
+    for ocm_info in ocm:
+        ocm_map = OCMMap(
+            ocms=[ocm_info],
+            integration=QONTRACT_INTEGRATION,
+            settings=settings,
+            init_version_gates=True,
+        )
+
+        current, _ = ocm_map.cluster_specs()
+        if current == 0:
```
patjlm, line 105: I don't think `current` can be `== 0`, can it? Do you mean `not current` ?

reconcile/ocm_update_recommended_version.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def get_updated_recommended_versions(
+    ocm_info: dict[str, Any], cluster: dict[str, OCMSpec]
+) -> list[dict[str, str]]:
+
+    rv_current = ocm_info.get("recommendedVersions") or []
+
+    if len(rv_current) == 0:
+        return []
```
patjlm, line 65: I am surprised by this, we will never have any `recommendedVersion` set if we return when the list is empty (which is the initial state). We could return if there are no clusters though.

reconcile/ocm_update_recommended_version.py
```diff
@@ -0,0 +1,123 @@
...
+                cluster[k].spec.version for k in cluster if k in cluster_workload
+            ]
+            rv_update = rv_workload[0]
+            if len(versions) > 0:
+                rv_update["recommendedVersion"] = recommended_version(
+                    versions, high_weight, majority_weight
+                )
+            rv_updated.append(rv_update)
+        elif len(rv_workload) > 1:
+            raise ValueError("Expecting one recommended Version per workload!")
```
patjlm, line 88: what if `en(rv_workload) == 0`? Meaning we did not recommend a version yet for that workload.
If there is now at least one cluster running this workload, we should be able to recommend a version now, right ?

reconcile/ocm_update_recommended_version.py
```diff
@@ -0,0 +1,123 @@
...
+    for workload in ocm_info["upgradePolicyAllowedWorkloads"] or []:
+        rv_workload = [rv for rv in rv_current if rv["workload"] == workload]
+        if len(rv_workload) == 1:
+            cluster_workload = [
+                c["name"]
+                for c in ocm_info["upgradePolicyClusters"]
+                if workload in c["upgradePolicy"]["workloads"]
+            ]
+            versions = [
+                cluster[k].spec.version for k in cluster if k in cluster_workload
```
patjlm, line 79:
```suggestion
                cluster[k].spec.version for k in cluster_workload
```

reconcile/ocm_update_recommended_version.py
janboll: Indeed, that looks like a mistake. should be len(current)

reconcile/ocm_update_recommended_version.py
janboll: Right, my idea was, one would have to add a workload to the recommendedVersion list first, only the it would get updated

reconcile/ocm_update_recommended_version.py
janboll: The cluster list should be used to iterate, cause that's where we get the data from. I could break the integration by this suggestion

reconcile/ocm_update_recommended_version.py
janboll: Please see the comment below regarding rv_workload. Same applies here

reconcile/ocm_update_recommended_version.py
patjlm: This does not seem to match with the dynamic nature of the fleet IMHO. Can we take this scenario into consideration ?

patjlm: Only when `recommendedVersionWeight` is set ?

janboll: No, weight is optional. If not set it will be 1 for both.

patjlm: hmm i thought that default was fine initially, but it indeed presumes all orgs would have some recommended versions.  Should the default be 0 instead? and if there are no positive score, no version recommended ?

patjlm (approved):
LGTM. The "knobs" are pretty hardcoded but we can certainly start like that.
Would be nice to have @maorfr 's feedback as well.

reconcile/ocm_update_recommended_version.py
```diff
@@ -0,0 +1,128 @@
...
+    )
+    return sorted_version_set[0]
+
+
+def get_majority(version_set: set[str], versions: list[str]) -> str:
+    return max(version_set, key=versions.count)
+
+
+def recommended_version(
+    versions: list[str], high_weight: int = 0, majority_weight: int = 0
```
steveteahan, line 29: Could you accept `versions` as an `Iterable` to keep this more general?

reconcile/test/test_ocm_update_recommended_version.py
```diff
@@ -0,0 +1,112 @@
...
+
+def test_get_majority(versions, version_set):
+    majority = get_majority(version_set, versions)
+    assert majority == "1.1.0"
+
+    assert get_majority(set(["1.1.1"]), ["1.1.1"]) == "1.1.1"
+
+
+def test_recommended_version(versions, version_set):
+    assert recommended_version(versions, 10, 1) == "1.1.1"
```
steveteahan, line 42: nit: including kwarg names might make this easier to follow?

maorfr:
it seems that this integration doesn't consider the version history at all? i was expecting that the majority of the calculations would come from there...

janboll:
@patjlm and I were smashing our heads over this and couldn't come up with a use case, where the history of versions is more relevant than the current running clusters.
 
After the discussion with @patjlm, I will update the code to be a little more future-proof and allow the more dynamic addition of weighting parameters that do not require schema changes. That should prepare addition of i.e. `most_soaked`. Where `most_soaked` could be defined as “the version that has soaked the most within the currently running versions for that workload”.

maorfr:
`most_soaked` sounds like a great use case. probably the best one to use?

janboll:
I'll add most_soaked in a follow up PR.

reconcile/ocm_update_recommended_version.py
steveteahan: This one might have been missed unless I'm looking at stale data? It might be more of a nit here, but the more that we tightly couple these types everywhere, the harder it is to reuse code, so it'd be good to keep in mind for other cases.

PR #2957, opened 2022-11-15
APPSRE-6545 upgrade policy sector

patjlm:
Depends on https://github.com/app-sre/qontract-schemas/pull/314

See [APPSRE-6545](https://issues.redhat.com/browse/APPSRE-6545) and [ADR 0032](https://docs.google.com/document/d/1_aRuF_Owlryo50mycbS7E0m9I6pzE_KPNUnLp7nQaXE/edit)

This pull request introduces the notion of OCM sectors. A sector regroups a set of clusters. Sectors are enablers for progressive rollout of versions across a fleet of clusters. In the context of this PR, only cluster versions are considered, but sectors could be later used for component (applications, resources, ...) progressive rollouts.

* An OCM organization can define a set of sectors by naming them
  * A sector name is unique within the OCM organization
  * Sectors can depend on each other, creating one or multiple dependency chains.
    * loops are being detected and the integration will fail in those cases
    * it is also possible to create sector dependency trees or even graphs for more complex scenarios (e.g. "production rollout start after stage `sector3` is done, while there are still more stage sectors to be deployed")
    * a sector can depend on a sector in an other OCM organization. This allows to use an OCM organization for each environment (integration/stage/production) and ensure a version is deployed on stage before being rolled out on production for example.
    * sector dependency name can be a `*` wildcard.
      * If it is in the same organization, all other sectors will be selected as dependencies
      * If is is referring to an other organization, all sectors of that organization will be selected as dependencies.
* A cluster in this org can then be part of a sector by configuring the sector name in its `upgradePolicy.conditions`
  * This cluster can then be upgraded to a version only if previous sector clusters running the same `workload` all run at least that version
  * If there are no previous sectors, the cluster can upgrade
  * If the previous sector does not include any cluster running the workload, we will
    * lookup recursively the sector chain to find matching `workload` clusters
    * if none is found, we will allow the upgrade.

Note: this might be easier to review commit by commit

### DIFF
--- a/reconcile/ocm_upgrade_scheduler.py
+++ b/reconcile/ocm_upgrade_scheduler.py
@@ -6,10 +6,11 @@ from datetime import datetime
 from typing import Any, Mapping, Optional
 from dateutil import parser
 from croniter import croniter
+from semver import VersionInfo
 
 from reconcile import queries
 
-from reconcile.utils.ocm import OCM, OCM_PRODUCT_OSD, OCMMap
+from reconcile.utils.ocm import OCM, OCM_PRODUCT_OSD, OCMMap, Sector
 from reconcile.utils.state import State
 from reconcile.utils.data_structures import get_or_init
 from reconcile.utils.semver_helper import parse_semver, sort_versions
@@ -55,6 +56,10 @@ def fetch_desired_state(
         spec = ocm.clusters[cluster_name].spec
         upgrade_policy["current_version"] = spec.version
         upgrade_policy["channel"] = spec.channel
+        # Replace sector names by their related OCM Sector object, including dependencies
+        sector_name = upgrade_policy["conditions"].get("sector")
+        if sector_name:
+            upgrade_policy["conditions"]["sector"] = ocm.sectors[sector_name]
         desired_state.append(upgrade_policy)
 
     sorted_desired_state = sorted(desired_state, key=sort_key)
@@ -155,7 +160,42 @@ def get_version_history(dry_run, upgrade_policies, ocm_map):
     return results
 
 
-def version_conditions_met(version, history, ocm_name, workloads, upgrade_conditions):
+def workload_sector_versions(sector: Sector, workload: str) -> list[VersionInfo]:
+    """
+    get all versions of clusters running the specified workload in that sector
+    """
+    versions = []
+    for cluster_info in sector.cluster_infos:
+        # clusters within a sector always have workloads (mandatory in schema)
+        workloads = cluster_info["upgradePolicy"]["workloads"]
+        if workload in workloads:
+            versions.append(
+                parse_semver(sector.ocmspec(cluster_info["name"]).spec.version)
+            )
+    return versions
+
+
+def workload_sector_dependencies(sector: Sector, workload: str) -> set[Sector]:
+    """
+    get the list of first dependency sectors with non-empty versions for that workload in the
+    sector dependency tree. This goes down recursively through the dependency tree.
+    """
+    deps = set()
+    for dep in sector.dependencies:
+        if workload_sector_versions(dep, workload):
+            deps.add(dep)
+        else:
+            deps.update(workload_sector_dependencies(dep, workload))
+    return deps
+
+
+def version_conditions_met(
+    version: str,
+    history: Mapping[Any, Any],
+    ocm_name: str,
+    workloads: list[str],
+    upgrade_conditions: dict[str, Any],
+):
     """Check that upgrade conditions are met for a version
 
     Args:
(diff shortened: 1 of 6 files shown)


reconcile/ocm_upgrade_scheduler.py
```diff
@@ -168,7 +208,19 @@ def version_conditions_met(version, history, ocm_name, workloads, upgrade_condit
...
     """
-    conditions_met = True
+    # check if previous sectors run at least this version for that workload
+    # we will check dependencies recursively until there are versions for the given workload
+    # or no more dependencies to check
+    sector = upgrade_conditions.get("sector")
+    if sector:
+        for w in workloads:
+            for dep in workload_sector_dependencies(sector, w):
+                dep_versions = workload_sector_versions(dep, w)
```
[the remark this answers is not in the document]
patjlm: Do you  think a `lru` would make sense?

reconcile/utils/ocm.py
```diff
@@ -578,6 +638,24 @@ def __init__(
...
+                ref = SectorWeakReference(
+                    ocm_org_name=dep_ocm_org_name, sector_name=dep["name"]
+                )
+                deps.append(ref)
+            self.sectors[sector["name"]] = Sector(
+                name=sector["name"],
+                ocm=self,
+                dependencies_refs=deps,
+            )
+
```
lechuk47, line 658: Should we split our OCM definition into OCM environments and OCM organizations or accounts? Right now we don't use the concept of an organization in our schema, it is just bound to the account (user/service account) we use to authenticate with the OCM CS. 
If this is expected to grow maybe it's better to have a clear schema definition to handle multiple OCM organizations in clear path forward instead of using the current OCM schema definition.

patjlm: We could, but I am not sure what it would bring.
Our ocm schema is actually and ocm org. Its URL refers to the env.
I think it's out of scope of this PR though

lechuk47: IMO it adds clarity, our current OCM is an ENV and an Account. The ORG is set into the account. I'm not sure if an account can have multiple organizations or not, but I would rather prefer not to set the ENV URL in all the  "organization" definitions.

I agree that maybe this is out of the scope of this task. I just pointed it out because this snippet has confused me with the semantics used.

patjlm: I agree this can be confusing. I'll resolve this thread as out of scope, but we can certainly have this conversation in teamchat

reconcile/ocm_upgrade_scheduler.py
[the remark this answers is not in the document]
patjlm: I checked quickly and it does not seem to bring much value. It actually requires to rewrite some tests as it would wrongly cache some Sector args while dependencies do change in sequential tests.
I'd pass on this for now.

janboll (approved):

reconcile/utils/ocm.py
```diff
@@ -508,6 +509,64 @@ class OCMServiceAccountNotAssociatedToOrg(Exception):
...
 
 
+class SectorConfigError(Exception):
+    pass
+
+
+# A weak reference to a sector. Used to reference a sector dependency, even if this one is
+# not available yet
+@dataclass
+class SectorWeakReference:
```
geoberle, line 519: a lot of the sector deps graph related code could be streamlined by using a graph library like networkx. for change-type inheritance (https://github.com/app-sre/qontract-reconcile/blob/fa3b25276da70348e5faa2d36fec84077ad1a92b/reconcile/change_owners/change_types.py#L408) it solved the weak reference, cycle detection and path traversal with a tiny amount of code.

no need to change anything here. just an idea for future changes whenever you need to tackle more challenging scenarios with sector interdependencies.

reconcile/utils/ocm.py
```diff
@@ -1437,6 +1516,24 @@ def __init__(
         else:
             raise KeyError("expected one of clusters, namespaces or ocm.")
 
+        # link sectors across OCM orgs
+        for ocm in self.ocm_map.values():
+            for sector in ocm.sectors.values():
+                for dep_refs in sector.dependencies_refs:
+                    other_ocm = self.ocm_map[dep_refs.ocm_org_name]
+                    other_sector_names = [dep_refs.sector_name]
+                    if dep_refs.sector_name == "*":
```
geoberle, line 1525: nice!

geoberle (approved):
LGTM!

reconcile/utils/ocm.py
patjlm: I'll have a look, thanks for the review!

PR #2959, opened 2022-11-15
Revert "Allow tekton timeouts bigger than 60 minutes (#2955)"

rporres:
This reverts commit 6c5ec38de93749c2e56ef01e7ace3f2ede11921f.


rporres:
It actually made sense. I had to promote first the schema changes so that both versions in q-server that the integration needs had the updated schema

PR #2962, opened 2022-11-16
CNA bindings

fishi0x01:
Introduce the concept of `Bindings`:

- Every `Asset` has a set of `Bindings`.
- By default, every `bindable()` asset in desired state is bound to its cluster-namespace in app-interface.
- `Bindings` only happen if an asset is in `READY` state.

Further:

- Delete `TERMINATED` assets


geoberle (approved):
LGTM!

PR #2963, opened 2022-11-16
support for the usage of multiple cluster authentication backends at the same time

chassing:
In preparation for [cluster logins via RedHat SSO](https://issues.redhat.com/browse/APPSRE-6374),
the `/openshift/cluster-1.yml.auth` attribute needs to be a list of auth objects. This PR adapts all integrations and utils.

* [APPSRE-6374](https://issues.redhat.com/browse/APPSRE-6374)
* [APPSRE-6554](https://issues.redhat.com/browse/APPSRE-6554)

Depends on https://github.com/app-sre/qontract-schemas/pull/317

### DIFF
--- a/reconcile/openshift_clusterrolebindings.py
+++ b/reconcile/openshift_clusterrolebindings.py
@@ -87,7 +87,7 @@ def fetch_desired_state(ri, oc_map):
     roles: list[dict] = expiration.filter(gqlapi.query(ROLES_QUERY)["roles"])
     users_desired_state = []
     # set namespace to something indicative
-    namepsace = "cluster"
+    namespace = "cluster"
     for role in roles:
         permissions = [
             {"cluster": a["cluster"], "cluster_role": a["clusterRole"]}
@@ -124,7 +124,7 @@ def fetch_desired_state(ri, oc_map):
                     try:
                         ri.add_desired(
                             cluster,
-                            namepsace,
+                            namespace,
                             "ClusterRoleBinding",
                             resource_name,
                             oc_resource,
@@ -143,7 +143,7 @@ def fetch_desired_state(ri, oc_map):
                 try:
                     ri.add_desired(
                         cluster,
-                        namepsace,
+                        namespace,
                         "ClusterRoleBinding",
                         resource_name,
                         oc_resource,
--- a/reconcile/openshift_clusterrolebindings.py
+++ b/reconcile/openshift_clusterrolebindings.py
@@ -108,28 +108,31 @@ def fetch_desired_state(ri, oc_map):
             cluster = cluster_info["name"]
             if not oc_map.get(cluster):
                 continue
-            user_key = ob.determine_user_key_for_access(cluster_info)
-            for user in role["users"]:
-                # used by openshift-users and github integrations
-                # this is just to simplify things a bit on the their side
-                users_desired_state.append({"cluster": cluster, "user": user[user_key]})
-                if ri is None:
-                    continue
-                oc_resource, resource_name = construct_user_oc_resource(
-                    permission["cluster_role"], user[user_key]
-                )
-                try:
-                    ri.add_desired(
-                        cluster,
-                        namepsace,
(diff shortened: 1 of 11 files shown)


rporres (commented):
The changes look good to me, but I don't have a lot of experience with many of the integrations this is touching, so another pair of eyes could be definitely interesting.

I don't see openshift-users being modified. I guess it should be touched to consider users with the new identity that will be created.

If I understand correctly, this change should be a no-op and only things will start happening when a new identity provider is added to the cluster, correct?

reconcile/openshift_clusterrolebindings.py
```diff
@@ -108,28 +108,31 @@ def fetch_desired_state(ri, oc_map):
...
-                    pass
+                    if ri is None:
+                        continue
+                    oc_resource, resource_name = construct_user_oc_resource(
+                        permission["cluster_role"], user[user_key]
+                    )
+                    try:
+                        ri.add_desired(
+                            cluster,
+                            namepsace,
```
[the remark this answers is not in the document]
chassing: Of course 👍

chassing:
Yes, these changes will be a no-op with the current cluster configurations. I'll introduce any OIDC-related changes in a later PR.

chassing:
@patjlm You reviewed [the related qontract-schemas PR](https://github.com/app-sre/qontract-schemas/pull/317#pullrequestreview-1184019018). Could you also review this PR, please?

reconcile/openshift_clusterrolebindings.py
```diff
@@ -108,28 +108,31 @@ def fetch_desired_state(ri, oc_map):
...
-                )
-                try:
-                    ri.add_desired(
-                        cluster,
-                        namepsace,
-                        "ClusterRoleBinding",
-                        resource_name,
-                        oc_resource,
+            for auth in cluster_info["auth"]:
+                user_key = ob.determine_user_key_for_access(cluster, auth)
```
patjlm, line 112: Is there a risk to have the same user_key between the 2/../n providers ?

reconcile/slack_cluster_usergroups.py
```diff
@@ -66,10 +66,12 @@ def get_desired_state(slack: SlackApi) -> dict[str, Any]:
...
     )
     all_clusters = queries.get_clusters(minimal=True)
+    # this needs to be refactored - https://issues.redhat.com/browse/APPSRE-6575
     clusters = [
         c
         for c in all_clusters
-        if c.get("auth") and c["auth"].get("team") and c.get("ocm")
+        for auth in c["auth"]
+        if auth.get("team") and c.get("ocm")
     ]
```
patjlm, line 75: don't we get multiple times the same clusters here? Is that fine ?

reconcile/slack_cluster_usergroups.py
```diff
@@ -66,10 +66,12 @@ def get_desired_state(slack: SlackApi) -> dict[str, Any]:
         sendgrid=False, saas_files=False, aws=False, permissions=False
     )
     all_clusters = queries.get_clusters(minimal=True)
+    # this needs to be refactored - https://issues.redhat.com/browse/APPSRE-6575
     clusters = [
         c
         for c in all_clusters
-        if c.get("auth") and c["auth"].get("team") and c.get("ocm")
+        for auth in c["auth"]
+        if auth.get("team") and c.get("ocm")
```
patjlm, line 74: don't we get multiple times the same clusters here? Is that fine ?

chassing: Yes, but this integration needs a bigger [refactoring](https://issues.redhat.com/browse/APPSRE-6575), because without GitHub as IDP we lose the `team` attribute. I'll address this in a later PR. With the current app-interface settings everything will be the same.

patjlm: ok, so the `auth.get("team")` actually catters for that it seems. LGTM

patjlm (approved):

PR #2966, opened 2022-11-16
Require Python type hints for all new code

steveteahan:
A little while back a `STRICT MODULES` section was introduced into `setup.cfg` to ensure that new code had full coverage for type definitions. This was underutilized, probably in part because it wasn't shared well enough. I've taken a new approach here where I've permitted existing modules to ignoring missing type definitions, but anything not matching those existing modules will require type definitions to be defined.

This change will make it much simpler to enforce type hints for all new modules being added to this project.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,206 +27,206 @@ disallow_incomplete_defs = True
 ; enabling the checking of untyped defs. We can work on reducing this list by removing an item from the list and fixing
 ; the issues to make CI pass.
 
-[mypy-e2e_tests.network_policy_test_base.*]
+[mypy-e2e_tests.network_policy_test_base]
 disallow_untyped_defs = False
 disallow_incomplete_defs = False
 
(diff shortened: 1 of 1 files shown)


bhushanthakur93 (approved):
LGTM

OliviaHY:
LGTM

maorfr:
LGTM :crying_cat_face:

bkez322:
LGTM

chassing (approved):
LGTM

geoberle (approved):
LGTM

janboll (approved):

steveteahan:
There are a good number of approvals here, this was presented in the NASA sync without concerns, and there were others who gave a +1 to the idea in Slack. This is pretty easy to undo if we need to, so I'm going to merge this now.

rporres:
A late LGTM

PR #2968, opened 2022-11-17
[CNA] override and default schema represented by same schema objects

geoberle:
all CNA assets must have a `defaults` and an `overrides` section, as a lot of the terraform-resources do. the difference is, that each CNA type declares a special `XXXConfig_v1` type that is used for both fields.

```yaml
  name: CNARDSInstance_v1
  interface: CNAsset_v1
  fields:
  - { name: provider, type: string, isRequired: true }
  - { name: identifier, type: string, isRequired: true, isUnique: true }
  - { name: name, type: string }
  - { name: defaults, type: CNARDSInstanceConfig_v1 } <--
  - { name: overrides, type: CNARDSInstanceConfig_v1 } <--
```

having defaults following a strict schema and overrides being able to override all of the defaults, makes writing testable and verifiable code a lot easier.


reconcile/cna/assets/asset.py
```diff
@@ -228,6 +228,40 @@ def from_api_mapping(
...
+
+    @classmethod
+    def aggregate_config(cls, spec: AssetQueryClass) -> ConfigClass:
+        defaults = cls._get_config("defaults", spec)
+        overrides = cls._get_config("overrides", spec)
+        config_class = cls._get_config_class_type()
+        data = {
+            property: getattr(overrides, property, None)
+            or getattr(defaults, property, None)
+            for property in config_class.__annotations__.keys()
```
fishi0x01, line 261: Can a defaults file have nested parameters? Will this work properly on a nested defaults section?

fishi0x01: Lets consider this

**defaults**
```yaml
param_a:
  attr_a: something
  attr_b: else
```

**overrides**
```yaml
param_a:
  attr_a: different
```

This will remove `attr_b` from the final map, even though intention was to only override `attr_a`.
Not sure if this is a valid case or if we assume that we always override all subkeys.

geoberle: you are right. it won't work in those scenarios. e.g. 

```yaml
defaults:
  parent_field:
    nested_field_1: default_1
    nested_field_1: default_2
overrides:
  parent_field:
    nested_field_2: override_2
```

would result in 

```yaml
parent_field:
    nested_field_2: override_2
```

instead of 

```yaml
parent_field:
    nested_field_1: default_1
    nested_field_2: override_2
```

i can get that resolved properly in a followup PR if you want

fishi0x01: If this is intended then its fine. Im not even sure this is a proper case to be concerned about :sweat_smile:

fishi0x01 (approved):

PR #2970, opened 2022-11-17
tf-cloudflare: add cloudflare account 2fa enforcement

jfchevrette:
Rework of #2964 to include fields which were missing and causing issues

Ref.: [APPSRE-6571](https://issues.redhat.com/browse/APPSRE-6571)

This adds a `cloudflare_account` definition which allows setting 2FA enforcement on accounts. The plan type also needs to be defined on that resource otherwise it default to "standard"

Depends on
* https://github.com/app-sre/qontract-schemas/pull/321

### DIFF
--- a/reconcile/gql_definitions/introspection.json
+++ b/reconcile/gql_definitions/introspection.json
@@ -19261,6 +19261,30 @@
                             },
                             "isDeprecated": false,
                             "deprecationReason": null
+                        },
+                        {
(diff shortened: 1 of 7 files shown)


jfchevrette:
Will merge this w/o approvals as this is the same code as in https://github.com/app-sre/qontract-reconcile/pull/2964 with an added field

PR #2979, opened 2022-11-21
glitchtip: add missing type hints

chassing:
APPSRE-6581

### DIFF
--- a/reconcile/utils/glitchtip/models.py
+++ b/reconcile/utils/glitchtip/models.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 import re
-from typing import Any, Optional
+from typing import Any, MutableMapping, Optional
 from pydantic import BaseModel, Field, root_validator
 
 
@@ -39,8 +39,8 @@ class Team(BaseModel):
 
     @root_validator(pre=True)
     def name_xor_slug_must_be_set(  # pylint: disable=no-self-argument
-        cls, values: dict[str, Any]
-    ) -> dict[str, Any]:
+        cls, values: MutableMapping[str, Any]
+    ) -> MutableMapping[str, Any]:
         assert ("name" in values or "slug" in values) and not (
             "name" in values and "slug" in values
         ), "name xor slug must be set!"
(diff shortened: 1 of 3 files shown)


reconcile/utils/glitchtip/models.py
```diff
@@ -65,17 +74,21 @@ class Project(BaseModel):
     teams: list[Team] = []
 
     @root_validator
-    def slugify(cls, values):  # pylint: disable=no-self-argument
-        values["slug"] = values.get("slug") or slugify(values.get("name"))
+    def slugify(  # pylint: disable=no-self-argument
+        cls, values: dict[str, Any]
```
steveteahan, line 78: Can `dict` inputs become `MutableMapping` as per [dev-guidelines](https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#abstract-collection-types)?

steveteahan (approved):
One minor change related to input types, otherwise LGTM. Thank you for removing some files from that list!

reconcile/utils/glitchtip/models.py
steveteahan: I haven't had time, but I'd really like to see if there's a plugin that would remind us of this. I forget to do this for `dict` more than any other use case (compared to say `Iterable`).

reconcile/utils/glitchtip/models.py
```diff
@@ -35,25 +38,31 @@ class Team(BaseModel):
     users: list[User] = []
 
     @root_validator(pre=True)
-    def name_xor_slug_must_be_set(cls, values):  # pylint: disable=no-self-argument
+    def name_xor_slug_must_be_set(  # pylint: disable=no-self-argument
+        cls, values: MutableMapping[str, Any]
+    ) -> MutableMapping[str, Any]:
```
steveteahan, line 43: Sorry if this isn't clear in the docs, but the intent is the abstract/protocol types in inputs and the concrete types in the outputs (see [here](https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#type-hints)).

So, `Mapping` and `MutableMapping` for inputs, but `dict` is fine for the return value.

chassing: I tried to have the return type as `dict[str, Any]`, but mypy doesn't like it. 

Take this easy example:

```python
def foobar(values: Mapping[str, str]) -> dict[str, str]:
    return values
```

And mypy complains:

```
error: Incompatible return value type (got "Mapping[str, str]", expected "Dict[str, str]")
```

Sorry, my type hints kungfu is weak today :(

steveteahan: Sorry, I should have paid more attention to what was going on in these functions. This is a common error when we're modifying some mutable object and returning it. Normally we could probably skip returning a mutable object from a function because we have the original reference to it anyways, but this looks like a pydantic implementation detail, so for these sorts of cases using `Mapping` or `MutableMapping` for both the input and return makes sense.

Thanks again for the work here!

PR #2980, opened 2022-11-21
[gitlab-owners] detect and log basic OWNER file format

geoberle:
gitlab-owners expects a dictionary like structure in an OWNERS file. this fix makes sure that the integration does not fail if that is not the case and logs the repo and owner file that makes troubles.


chassing (approved):
LGTM

PR #2981, opened 2022-11-21
secret-reader: type hints

chassing:
add missing type hints and fix `read` return type

### DIFF
--- a/reconcile/utils/secret_reader.py
+++ b/reconcile/utils/secret_reader.py
@@ -32,7 +32,7 @@ class SecretReaderBase(ABC):
     @abstractmethod
     def _read(
         self, path: str, field: str, format: Optional[str], version: Optional[int]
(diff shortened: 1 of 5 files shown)


reconcile/utils/secret_reader.py
```diff
@@ -133,7 +133,7 @@ def _read_all(
         self, path: str, field: str, format: Optional[str], version: Optional[int]
     ) -> dict[str, str]:
         try:
-            data = self.vault_client.read_all(
```
fishi0x01, line 136: Interesting. Whats the mypy error code?

fishi0x01: Oh I see

```
"VaultClient" has no attribute "read_all"  [attr-defined]
```

chassing: Sure 🤦

PR #2984, opened 2022-11-21
Add triggers channel for GChat in FedRamp

cubismod:
[APPSRE-6480](https://issues.redhat.com/browse/APPSRE-6480)

In FedRamp, ship logs from *-trigger integration to a separate channel to avoid spamming the main reconcile channel with messages.

### DIFF
--- a/helm/qontract-reconcile/templates/template.yaml
+++ b/helm/qontract-reconcile/templates/template.yaml
@@ -86,11 +86,19 @@ objects:
             value: ${SLACK_ICON_EMOJI}
           {{- end }}
           {{- if $logs.googleChat }}
+          {{- if $integration.trigger }}
+          - name: GOOGLE_CHAT_WEBHOOK_URL
+            valueFrom:
+              secretKeyRef:
+                key: google_chat.triggers_url
+                name: app-interface
+          {{- else }}
           - name: GOOGLE_CHAT_WEBHOOK_URL
             valueFrom:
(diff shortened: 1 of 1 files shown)


cubismod:
I see for Slack logs, there is templating done in the Fluentd config: https://github.com/app-sre/qontract-reconcile/blob/ce6763711401ac8b52acf7db130cd85892e0475a/helm/qontract-reconcile/templates/template.yaml#L149 but I don't believe that would be necessary here since I'm just swapping out the webhook variable before it even gets to Fluentd

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -86,11 +86,19 @@ objects:
             value: ${SLACK_ICON_EMOJI}
           {{- end }}
           {{- if $logs.googleChat }}
+          {{- if $integration.trigger }}
+          - name: GOOGLE_CHAT_WEBHOOK_URL
```
TGPSKI, line 90: This will produce a naming collision

TGPSKI (approved):
On my first read of this change, I was fairly confused. Second time through it made sense. Not sure if there is a comment needed inline to help the first read through parse. Up to you.

cubismod:
> On my first read of this change, I was fairly confused. Second time through it made sense. Not sure if there is a comment needed inline to help the first read through parse. Up to you.

Comment would help so I'll add one

PR #2986, opened 2022-11-22
more details in error message about failure in upstream job

fishi0x01:
This error message will give a fast hint on which upstream was failing.


kwilczynski (approved):
@fishi0x01, nice! Thank you!

PR #2990, opened 2022-11-22
[WIP] typed externalresource spec

geoberle:
`ExternalResourceSpec` serves well as streamlined way to load external resources. sadly it operators only on `dict` data. especially for cloudflare and the upcoming CNA integratioln, it means leaving the typehinted dataclasses behind when transforming query results into `ExternalResourceSpec`.

this PR makes `ExternalResourceSpec` a base type and introduces two subtypes:

* `DictExternalResourceSpec` which represents the dict based approach to keep compatibility with the classic `terraform-resources` integrations
* `TypedExternalResourceSpec` for cloudflare and CNA, which keeps the dataclasses alive.

both types leverage python generics (typevars) to be able to store the resource and namespaces with different types and compensate their differences with customer property getters.

the loading mechanism from dataclasses to `TypedExternalResourceSpec` is done via `get_external_resource_specs_for_namespace` and leverages `Protocols` so that various namespaces, provisioners and actual resources provided by different qenerate queries can be passed.

the first commit introduces the to `ExternalResourceSpec` subtypes and adapt `terraform-resources` to use the `dict` version of it.

the second commit adapts the cloudflare integration to use `TypedExternalResourceSpec`. by that, the dataclasses survive through all layers of code into `cloudflare_resources.py`. the transition is not executed in full, because the`populate` functions have not been adapted yet to `self._spec` being a proper dataclass.

the resolver has also not been touched yet.

### DIFF
--- a/reconcile/gql_definitions/terraform_cloudflare_resources/terraform_cloudflare_resources.gql
+++ b/reconcile/gql_definitions/terraform_cloudflare_resources/terraform_cloudflare_resources.gql
@@ -2,6 +2,9 @@
 query TerraformCloudflareResources {
   namespaces: namespaces_v1 {
     name
+    cluster {
+      name
+    }
     managedExternalResources
     externalResources {
       ... on NamespaceTerraformProviderResourceCloudflare_v1 {
@@ -11,9 +14,9 @@ query TerraformCloudflareResources {
         }
         resources {
           provider
+          identifier
           ... on NamespaceTerraformResourceCloudflareWorkerScript_v1
           {
-            identifier
             name
             content_from_github {
               repo
@@ -27,7 +30,6 @@ query TerraformCloudflareResources {
           }
           ... on NamespaceTerraformResourceCloudflareZone_v1
           {
-            identifier
             zone
             plan
             type
--- a/reconcile/gql_definitions/terraform_cloudflare_resources/terraform_cloudflare_resources.py
+++ b/reconcile/gql_definitions/terraform_cloudflare_resources/terraform_cloudflare_resources.py
@@ -21,6 +21,9 @@ DEFINITION = """
 query TerraformCloudflareResources {
   namespaces: namespaces_v1 {
     name
+    cluster {
+      name
+    }
     managedExternalResources
     externalResources {
       ... on NamespaceTerraformProviderResourceCloudflare_v1 {
@@ -30,9 +33,9 @@ query TerraformCloudflareResources {
         }
         resources {
           provider
+          identifier
           ... on NamespaceTerraformResourceCloudflareWorkerScript_v1
           {
-            identifier
             name
             content_from_github {
               repo
@@ -46,7 +49,6 @@ query TerraformCloudflareResources {
           }
           ... on NamespaceTerraformResourceCloudflareZone_v1
           {
-            identifier
             zone
             plan
             type
@@ -86,6 +88,14 @@ query TerraformCloudflareResources {
 """
 
 
+class ClusterV1(BaseModel):
+    name: str = Field(..., alias="name")
+
+    class Config:
+        smart_union = True
+        extra = Extra.forbid
+
+
 class NamespaceExternalResourceV1(BaseModel):
     class Config:
         smart_union = True
@@ -102,6 +112,7 @@ class CloudflareAccountV1(BaseModel):
 
 class NamespaceTerraformResourceCloudflareV1(BaseModel):
     provider: str = Field(..., alias="provider")
+    identifier: str = Field(..., alias="identifier")
 
     class Config:
         smart_union = True
@@ -130,7 +141,6 @@ class CloudflareZoneWorkerScriptVarsV1(BaseModel):
 class NamespaceTerraformResourceCloudflareWorkerScriptV1(
     NamespaceTerraformResourceCloudflareV1
 ):
-    identifier: str = Field(..., alias="identifier")
     name: str = Field(..., alias="name")
     content_from_github: Optional[
         CloudflareZoneWorkerScriptContentFromGithubV1
@@ -191,7 +201,6 @@ class CloudflareZoneCertificateV1(BaseModel):
 class NamespaceTerraformResourceCloudflareZoneV1(
     NamespaceTerraformResourceCloudflareV1
 ):
-    identifier: str = Field(..., alias="identifier")
     zone: str = Field(..., alias="zone")
     plan: Optional[str] = Field(..., alias="plan")
     q_type: Optional[str] = Field(..., alias="type")
@@ -226,6 +235,7 @@ class NamespaceTerraformProviderResourceCloudflareV1(NamespaceExternalResourceV1
 
 class NamespaceV1(BaseModel):
     name: str = Field(..., alias="name")
+    cluster: ClusterV1 = Field(..., alias="cluster")
     managed_external_resources: Optional[bool] = Field(
         ..., alias="managedExternalResources"
     )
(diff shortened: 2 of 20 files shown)


steveteahan:
Just dropping a note to say that I looked through as a first pass. I'm going to try my best to get some high-level feedback later tonight or possibly tomorrow.

steveteahan:
Thank you in advance for working on this problem because it's non-trivial to solve. I'm just tossing out some high-level thoughts:

- My understanding is that most of the original benefits of gql data classes were that we could catch issues when loading data, we could more easily move move between typed objects, and the other general benefits of type hints (auto-completion, catching issues, etc)

- With the above in mind, it's unclear to me that the benefits are as straight-forward when the end object are Terrascript objects that are essentially `dict`s. With that being said, it's possible that could change with time (CDK).

- `NamespaceTerraformResourceCloudflareV1`, a gql data class, has found its way into the `utils` modules like `reconcile/utils/terrascript/cloudflare_client.py`. I know we never closed out [this discussion](https://gitlab.cee.redhat.com/service/dev-guidelines/-/merge_requests/134), but we mentioned in there that usage in `utils` is discouraged. We'll be working on a Cloudflare DNS integration next that would be closer to how [dns-zone-1](https://gitlab.cee.redhat.com/service/app-interface/-/tree/master#manage-dns-zones-via-app-interface-awsdns-zone-1yml-using-terraform) works. That would mean a completely different gql query and different gql data query class that would need to be added wherever `NamespaceTerraformResourceCloudflareV1` is covered now?

- Parts of this PR do introduce a decent bit of added complexity. This doesn't necessarily mean I don't think it's worth it, but I'd like to better understand what types of bugs we perceive we will be able to catch with this new approach. I know that we keep the data classes available longer, but it'd be great if there was a demonstration of actually catching a bug that we manually introduce with that.

I still feel like we're missing a layer here that I mentioned in one of the [original design docs](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/38147#note_4120540). In the case of the Cloudflare example, perhaps this would mean `CloudflareZone` and `CloudflareWorker` dataclasses that are independent of gql.

So, you'd have something like this:

1. In the integration, query the data, get back the `NamespaceTerraformResourceCloudflareV1` objects
2. In the integration, use these gql query data classes to initialize `CloudflareZone` and `CloudflareWorkers` classes that are defined in `utils.terrascript` (need to think this through more, I'd imagine `TypedExternalResourceSpec` could still be used)
3. From there, the code in `utils.terrascript` knows how to handle the data classes that it has defined manually (these are not gql data query classes)

The primary benefits here are:

1. gql data query classes never leave integration code
2. If multiple integrations use this code, we don't need multiple gql data query classes from the different queries making it to the `utils` "library" code
3. We can test the `utils` code without ever importing any `gql_definitions`
4. In theory doing this we could even replace gql with something else, and we'd never need to change anything in `utils` (not saying this is a current goal, just it's a sound approach to being open for future change)

The downside would be needing to keep those module-specific dataclasses up-to-date, but I think the reality is that when they're changing, it's probably because some logic changed as well.

Sorry for the really long response, but I think there's still some disagreement of how gql query data classes are intended to be used, which is what a lot of the comments are about.

geoberle:
> * My understanding is that most of the original benefits of gql data classes were that we could catch issues when loading data, we could more safely/easily move data between typed objects, and the other general benefits of type hints (auto-completion, catching issues, etc)

to all the benefits you mentioned i add that we can data validation in a more stable way because of actual fields being accessible and the notion of a field missing or a field just being null is a lot clearer defined.

> * With the above in mind, it's unclear to me that the benefits are as straight-forward when the end object are Terrascript objects that are essentially `dict`s. With that being said, it's possible that could change with time (Terraform CDK?).

as you can see i stopped this WIP PR at the point where the benefit of having the dataclass in `cloudflare_resources.py` would start, namely being able to validate data. but i agree with you that the value in the context of cloudflare is limited as long as terrascript plays a role. if terrascript would be typed itself, the value would be immediately apparent because the `TerrascriptResource`s in `cloudflare_resources.py` would act as a typesafe protocol translator between the a-i representation and the cloudflare one.

this PR and the idea have been kindled by trying to adopt `ExternalResourceSpec` for the CNA integration, where @fishi0x01 and I decided to have an actual typehinted dataclass representation of the CNA asset payload of a module so not loosing the dataclass is crucial for a safe and sound translation process from one representation to the other.

> * `NamespaceTerraformResourceCloudflareV1`, a gql data class, has found its way into the `utils` modules like `reconcile/utils/terrascript/cloudflare_client.py`. 

yes i was honestly not paying attention to `cloudflare_resources.py` being in utils and i wonder if it would not better fit closer to the cloudflare terraform resources integration itself, e.g. put everything cloudflare related into `/reconcile/cloudflare/`. but i have not enough insights about this.

> I still feel like we're could be missing a layer here that I mentioned in one of the [original design docs](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/38147#note_4120540). In the case of the Cloudflare example, perhaps this would mean `CloudflareZone` and `CloudflareWorker` dataclasses that are independent of gql.

yes absolutely true. for CNA we have such a layer that represents the CNA asset payload of a specific module and the dataclasses get translated into those but that layer. e.g. this https://github.com/app-sre/qontract-reconcile/blob/cna-integration/reconcile/cna/assets/aws_rds.py

> The downside would be needing to keep those module-specific dataclasses up-to-date, but I think the reality is that when they're changing, it's probably because some logic changed as well.

keeping such module specific dataclasses up to date is not really something that frightens me. strong type hints and linting will show discrepancies for most issues immediately. test show the rest. 

> Sorry for the really long response, but I think there's still some disagreement of how gql query data classes are intended to be used, which is what a lot of the comments are about.

i'm thankful for the time you spend on this topic.

fishi0x01:
My understanding now is that we all agree on the core idea behind this PR, i.e., introducing `TypedExternalResourceSpec` which is a generic for concrete types. The discussion here is mainly about which type we pass to that generic in this Cloudflare integration case right? As Gerd pointed out, this can be very different depending on the integration (mainly if `TypedExternalResourceSpec`  is used in `utils` or not).

Personally, Im fine with having a protocol layer of the types within the `utils` case. It adds some overhead, but also we gain isolation.

steveteahan:
> My understanding now is that we all agree on the core idea behind this PR, i.e., introducing TypedExternalResourceSpec which is a generic for concrete types. The discussion here is mainly about which type we pass to that generic in this Cloudflare integration case right? As Gerd pointed out, this can be very different depending on the integration (mainly if TypedExternalResourceSpec is used in utils or not).

After thinking about this a bit more, I agree that I don't really see the introduction of `TypedExternalResourceSpec` being harmful beyond some additional complexity that's introduced. In non-Terrascript use cases that additional complexity could be offset by the potential to detect type errors closer to the code where the resources are being created. For Terrascript use cases, it's unclear to me how many issues we're likely to find with the addition of the `TypedExternalResourceSpec`.

Overall, I do want to keep in the back of my mind what the overall goal is. I don't think that `dict` usage is always bad. If the cost to use types for certain use cases is high, and provides minimal benefit, then maybe we can consider other alternatives like integration testing?

This is all to say that if it's easy for us to retrofit existing Terrascript use cases it might be worth it, but if it's not then maybe the `dict` version continues to be used everywhere besides the CNA use case?

> Personally, Im fine with having a protocol layer of the types within the utils case. It adds some overhead, but also we gain isolation.

I think it comes down to either dataclasses or `Protocol`s, whichever makes sense.

geoberle:
> After thinking about this a bit more, I agree that I don't really see the introduction of `TypedExternalResourceSpec` being harmful beyond some additional complexity that's introduced.

the problem is that not being harmful is not enough. the complexity must be outvalued by the benefit, which is clearly not there as long as integrations are using terrascript. that said, i'm going to close this PR for now and we can reopen/reevaluate it when there are usecases around that fully benefit from a typed approach.

for CNA alone it is not worth the effort, especially since CNA does not use the feature of external-resource-spec like defaults resolution (CNA defaults are a datafile which a schema) and secret management (solved via CNA bindings).

PR #2992, opened 2022-11-22
Fluentd memory limit increase

cubismod:
[APPSRE-6600](https://issues.redhat.com/browse/APPSRE-6600)

With Fluentd -> GChat logging now deployed in FedRamp, we have noted the Fluentd containers consistently running over the existing memory limit of 120Mi to around 125Mi. Thus, this MR increases the limit to 130Mi to reduce the amount of SIGKILL errors in Fluentd and prevent issues such as repeating logs into Google Chat.

### DIFF
--- a/helm/qontract-reconcile/templates/template.yaml
+++ b/helm/qontract-reconcile/templates/template.yaml
@@ -362,7 +362,7 @@ objects:
               memory: 30Mi
               cpu: 15m
             limits:
-              memory: 120Mi
+              memory: 130Mi
               cpu: 25m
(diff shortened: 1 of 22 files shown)


maorfr:
have you considered using a parameter?

cubismod:
> have you considered using a parameter?

I haven't. Could do something like setting a default limit of 120Mi and then updating that limit to 130Mi in FedRamp I imagine.

cubismod:
Going to take another crack at this in a separate PR using Helm templating

PR #2996, opened 2022-11-23
Fluentd Customizable Resource Limits and Requests 

cubismod:
Take 2 at https://github.com/app-sre/qontract-reconcile/pull/2992 using Helm templating.

[APPSRE-6600](https://issues.redhat.com/browse/APPSRE-6600)

With Fluentd -> GChat logging now deployed in FedRamp, we have noted the Fluentd containers consistently running over the existing memory limit of 120Mi. This MR allows users to specify on an integration level the Fluentd resource limits/requests, similar to what we can currently do with the QR pod.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -23,30 +23,31 @@ https://github.com/helm/helm/releases
 
 ### Integration spec configuration
 
-| Parameter                   | Description                                                              | Default                            |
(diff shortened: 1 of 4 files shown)


bkez322:
Ah, I didn't realize how much logic would need to be added for the resource changes to take affect across all of the integration pods. But this makes sense to me. Looks good to me.

bkez322 (approved):

PR #2998, opened 2022-11-24
Use SQSGateway for update recommended version

janboll:
Related issues: [APPSRE-6543](https://issues.redhat.com/browse/APPSRE-6543)


maorfr:
why do we need the gitlab project id? we usually go through SQS...

janboll:
Does it work without it? Was not sure, will give it a try.

maorfr (approved):

patjlm (approved):

PR #2999, opened 2022-11-24
slack-usergroups: introdcuce GQL and typed state classes

chassing:
* GQL classes
* convert internal state dicts into typed classes
* full type hints support
* more unittests

APPSRE-6592

Hints for the reviewers: 

* I've tested `slack-usergroups` integration locally with `dryrun`
* Unittests adapted to use `pytest.fixtures` to avoid code duplications


chassing:
I need to adapt `gitlab_members` as well :(

PR #3001, opened 2022-11-24
Enforce types for dashdotdb-dvo

fishi0x01:
Tested this by running integration locally in dry-run mode.

### DIFF
--- a/reconcile/dashdotdb_dvo.py
+++ b/reconcile/dashdotdb_dvo.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
-from typing import Any, Optional, Sequence, Mapping, Generator
+from typing import Any, Optional
+from collections.abc import Generator, Sequence, Mapping
 import requests
 from requests import Response
 
--- a/reconcile/dashdotdb_dvo.py
+++ b/reconcile/dashdotdb_dvo.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass
-from typing import Any, Optional
+from typing import Any, Optional, Sequence, Mapping, Generator
 import requests
+from requests import Response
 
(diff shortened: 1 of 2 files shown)


reconcile/dashdotdb_dvo.py
```diff
@@ -1,6 +1,7 @@
 from dataclasses import dataclass
-from typing import Any, Optional
+from typing import Any, Optional, Sequence, Mapping, Generator
```
rporres, line 2: please import those abstract types for `collections.abc` as importing them from `typing` is deprecated since 3.9, e.g https://docs.python.org/3.9/library/typing.html#typing.Generator

reconcile/dashdotdb_dvo.py
```diff
@@ -31,26 +32,26 @@ class PrometheusInfo:
 
 
 class DashdotdbDVO(DashdotdbBase):
-    def __init__(self, dry_run, thread_pool_size):
+    def __init__(self, dry_run: bool, thread_pool_size: int):
```
rporres, line 35: I think you can add a return type for this if you add

```
```

reconcile/dashdotdb_dvo.py
fishi0x01: TIL! Thank you :)

reconcile/dashdotdb_dvo.py
fishi0x01: Not sure I understand. The return type of `__init__` must always be `None` right? Even importing 

```
```

doesnt change that.

rporres: nah, forget about me. My head works intermittently these days 😅

rporres (approved):

PR #3002, opened 2022-11-24
Refactor pageduty_api

chassing:
* Don't rely on SecretReader
* New `get_pageduty_map`, get_usernames_from_pagerduty` methods to be used by integrations (`slack-usergroups`, `gitlab-members`)
* common GQL classes to be used by integrations
* full type hints support
* UNITTESTS

This PR is a preparation step for [APPSRE-6592 - refactor slack-usersgroups](https://issues.redhat.com/browse/APPSRE-6592) and [APPSRE-6610 - refactor gitlab-members](https://issues.redhat.com/browse/APPSRE-6610)

Hints for the reviewers:

* `get_pageduty_map` & `get_usernames_from_pagerduty` are extracted from `slack-usergroups` and will be used later
* Local dryrun of `slack-usergroups` and `gitlab-members`
APPSRE-6596


reconcile/utils/pagerduty_api.py
```diff
@@ -1,40 +1,77 @@
...
+
+class SupportsPagerDutyInstance(Protocol):
+    name: str
+
+    @property
+    def token(self) -> SupportsSecret:
+        ...
+
+
+class SupportsUser(Protocol):
```
geoberle, line 29: naming nit: `PagerDutyUser` instead of `SupportsUser`
same for other `Supports*` Protocols. the `Supports` prefix throws me off a bit

reconcile/utils/pagerduty_api.py
```diff
@@ -1,40 +1,77 @@
...
+        now = dt.utcnow()
 
         try:
             if resource_type == "schedule":
                 users = self.get_schedule_users(resource_id, now)
             elif resource_type == "escalationPolicy":
                 users = self.get_escalation_policy_users(resource_id, now)
         except requests.exceptions.HTTPError:
-            return None
+            return []
```
geoberle, line 70: i'm aware that this kind of errorhandling was already there before the refactor and that the calling code deals with this, but this approach makes me uneasy.

no need to fix in this context because you did not change anything in behaviour

reconcile/utils/pagerduty_api.py
```diff
@@ -79,15 +118,17 @@ def get_escalation_policy_users(self, escalation_policy_id, now):
 class PagerDutyMap:
     """A collection of PagerDutyApi instances per PagerDuty instance"""
 
-    def __init__(self, instances, init_users: bool = True, settings=None):
-        self.pd_apis = {}
+    def __init__(
```
geoberle, line 121: i love it that no settings and secret readers are present in the `PagerDutyMap` 🥳

reconcile/utils/pagerduty_api.py
```diff
@@ -96,4 +137,77 @@ def get(self, name):
...
+        pager_duty_api_class=pager_duty_api_class,
+    )
+
+
+def get_pagerduty_name(user: SupportsUser) -> str:
+    return user.pagerduty_username or user.org_username
+
+
+@retry(no_retry_exceptions=PagerDutyTargetException)
+def get_usernames_from_pagerduty(
```
geoberle, line 165: this is the replacement for the one in  `slack_usergroups.py`, right?
are you going to remove the other one?

geoberle (commented):
nice. minor nits

reconcile/utils/pagerduty_api.py
chassing: I'm following just the patterns, e.g. https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/secret_reader.py#L19. Looks like it's kind of a [standard](https://mypy.readthedocs.io/en/stable/protocols.html#simple-user-defined-protocols).

reconcile/utils/pagerduty_api.py
chassing: I've added type hints, and therefore I need to decide on `list[...]` vs. `Optional[list[...]]`. The `slack-usergroups` integration, which uses `get_pagerduty_users`, will be refactored later and will benefit :)

reconcile/utils/pagerduty_api.py
chassing: Yes, this PR is a preparation step for the `slack-usergroups` refactoring

reconcile/utils/pagerduty_api.py
geoberle: afaiu `Supports*` is the standard if an object supports specific functionality. in the sense of `this objects supports to be closed`.

`SupportsUser` in the sense of `this objects supports BEING a user` sounds awkward to me. that might be just personal preference.

besides the `Supports*` and `*able` Protocols, there are others as well in core python, e.g. `Container` or `Sized`. i think i read somewhere that such names are preferable if a protocol expresses a clear concept of an object being a thing.

i admit, that this PR is not the right context to discuss this. so feel free to ignore.

geoberle: https://github.com/python/typeshed/pull/4229/files

reconcile/utils/pagerduty_api.py
geoberle: i meant that this function hides the fact that an error occured and that `None` or `[]` might not be the actual state on pagerduty

chassing: Let me add a log message

geoberle: works for me for now 👍

reconcile/utils/pagerduty_api.py
geoberle: [a616877](https://github.com/app-sre/qontract-reconcile/pull/3002/commits/a616877dc1cf01fdcbfadaa4384d87c8fb3952be) ❤️

reconcile/utils/pagerduty_api.py
geoberle: [2946115](https://github.com/app-sre/qontract-reconcile/pull/3002/commits/29461151d9053890e67ba21ecc05e1e6fb6da813) ❤️

reconcile/utils/pagerduty_api.py
chassing: I won't die on that hill :) I've renamed the classes. To be honest, I like it without `Supports`

geoberle (approved):
awesome! LGTM!

reconcile/gql_definitions/common/pagerduty_instances.gql
```diff
@@ -0,0 +1,9 @@
+# qenerate: plugin=pydantic_v1
+query PD {
```
fishi0x01, line 2: Query names must be unique globally, because all definitions share a common context. `PD` is very short and sth that might not be unique globally. I'd recommend to rely on more descriptive names here, like `PagerDutyInstances` or sth like that.

fishi0x01: I guess the reason this is short is because you'd like to have a short class name. I think a proper solution, without interfering with global gql context might be a feature flag in qenerate itself , like

```graphql
# qenerate: query_data_container_name=PD
query PagerDutyInstances { }
```

chassing: The reason for this short query name is my laziness 😈 Let me change it

reconcile/utils/pagerduty_api.py
rporres: I definitely think this deserves a reference in our style guide. FWIW, I also prefer the new names ;-)

reconcile/utils/pagerduty_api.py
```diff
@@ -1,40 +1,78 @@
-import datetime
+from datetime import datetime as dt
+import logging
+from typing import Callable
+from typing import Iterable
+from typing import Optional
+from typing import Protocol
```
rporres, line 6: nit: Callable, Iterable and Protocol should be imported from `collections.abc`

reconcile/utils/pagerduty_api.py
```diff
@@ -1,40 +1,78 @@
...
+import requests
+from pydantic import BaseModel
+from sretoolbox.utils import retry
 
 from reconcile.utils.secret_reader import SecretReader
+from reconcile.utils.secret_reader import SupportsSecret
+
+
+class PagerDutyTargetException(Exception):
+    ...
```
rporres, line 18: nit: should we use `pass` or the ellipsis here? The general style of the code is to use `pass` for this kind of exceptions, but maybe you want to do something else after...

reconcile/utils/pagerduty_api.py
```diff
@@ -1,40 +1,78 @@
...
+    org_username: str
+    pagerduty_username: Optional[str]
+
+
+class PagerDutyTarget(Protocol):
+    name: str
+    instance: PagerDutyInstance
+    escalation_policy_id: Optional[str]
+    schedule_id: Optional[str]
+
```
rporres, line 39: What's the reasoning behind this classes being protocols? That we can pass directly objects from `qenerate`? Should that deserve comments/documentation somewhere? 🙏

setup.cfg
```diff
@@ -954,6 +950,10 @@ disallow_incomplete_defs = False
 disallow_untyped_defs = False
 disallow_incomplete_defs = False
 
+[mypy-tools.test.test_utils_pagerduty_api]
+disallow_untyped_defs = True
+disallow_incomplete_defs = True
+
```
rporres, line 956: why this? shouldn't we aim for everything new being properly typed?

rporres (commented):
Looks great. Some non-very-important questions and nits

reconcile/utils/pagerduty_api.py
chassing: aaahhh, not again 🤦 I really need an IDE utility for that ;)

setup.cfg
chassing: You are right! We've a general "type hints, I don't care" setting for [all our tests](https://github.com/app-sre/qontract-reconcile/blob/master/setup.cfg#L569) and this explict enables the type checking for the pagerduty test file. `disallow = True` :)

rporres: ahhh, ok... understood 😍 

I hate properties that are not affirmative 😬

reconcile/utils/pagerduty_api.py
```diff
@@ -1,40 +1,90 @@
...
+    name: str
+    instance: PagerDutyInstance
+    escalation_policy_id: Optional[str]
+    schedule_id: Optional[str]
+
+
+class PagerDutyConfig(BaseModel):
+    """This protocol defines mandatory attributes/methods
+    which must be implemented by a class to be compatible."""
```
rporres, line 52: did you mean this class to be a Protocol then?

reconcile/utils/pagerduty_api.py
```diff
@@ -1,40 +1,90 @@
-import datetime
+import logging
+from collections.abc import Iterable
+from datetime import datetime as dt
+from typing import Callable
+from typing import Optional
+from typing import Protocol
```
rporres, line 6: Protocol and Callable should be imported from `collections.abc` too, poor things 😬

setup.cfg
chassing: Me too :)

reconcile/utils/pagerduty_api.py
chassing: `Callable` yes, `Protocol` no. Why do we have `collections.abc` and `typings`!?

reconcile/utils/pagerduty_api.py
chassing: Not at all. The `Protocol`s are a kind of interface or blueprint without any logic. A class that implements a protocol must have all methods and attributes implemented. e.g.,

```python
# library.oy
class User(Protocol):
    username: str
    email: str


def get_username(user: User) -> str:
    return u.usename


# integration-github.py
from library import get_username

class GithubUser(BaseModel): # maybe a GQL class or something else
    username: str
    email: str
    repos: list[Repo]
    settings: list[Settings]

gh = GithubUser(username="billythekid", email="billythekid@newyork.us", repos=[], settings=[])

assert get_username(user=gh) == "billythekid" 

# integration-githlab.py
from library import get_username

class GitlabUser(BaseModel): # maybe a GQL class or something else
    username: str
    email: str
    repos: list[Repo]
    settings: list[Settings]

gl = GitlabUser(username="billythekid", email="billythekid@newyork.us", repos=[], settings=[])

assert get_username(user=gl) == "billythekid" 
```

as you can see, 2 different User classes `GithubUser` and `GitlabUser` and both can be used for `get_username` because both have `username` and `email` attributes implemented.

reconcile/utils/pagerduty_api.py
rporres: yeah, you're right

The answer for that is in https://peps.python.org/pep-0585/

reconcile/utils/pagerduty_api.py
rporres: I was just meaning that the docstring suggested that `PagerDutyConfig` is a protocol whereas the class inherits from `BaseModel`.

But thx a lot for the example ❤️

rporres (approved):

reconcile/utils/pagerduty_api.py
chassing: Fixed the doc: 🤦

PR #3006, opened 2022-11-25
allow Error state

fishi0x01:
An asset can also be in state `Error`


geoberle (approved):
ah makes sense. thank you. lgtm!

PR #3007, opened 2022-11-25
Cluster upgrade shared version data

patjlm:
[APPSRE-6549](https://issues.redhat.com/browse/APPSRE-6549)

Allow to inherit cluster version data information from one org to an other. Currently, this is limited to soak days. We will add current version stats (lowest current version, ...) in a subsequent PR.

The first commit is the implementation. The second one is a refactor of the version history deep-dict into a set of dataclasses.

Depends on https://github.com/app-sre/qontract-schemas/pull/330

### DIFF
--- a/reconcile/ocm_upgrade_scheduler.py
+++ b/reconcile/ocm_upgrade_scheduler.py
@@ -8,7 +8,6 @@ from typing import (
 )
 
 from croniter import croniter
-from dateutil import parser
 from semver import VersionInfo
 
 from reconcile import queries
@@ -92,7 +91,7 @@ def update_history(version_data: VersionData, upgrade_policies: list[dict[str, A
         upgrade_policies (list): query results of clusters upgrade policies
     """
     now = datetime.utcnow()
-    check_in = parser.parse(version_data.check_in or str(now))
+    check_in = version_data.check_in or now
 
     # we iterate over clusters upgrade policies and update the version history
     for item in upgrade_policies:
@@ -114,24 +113,7 @@ def update_history(version_data: VersionData, upgrade_policies: list[dict[str, A
             else:
                 workload_history.reporting.append(cluster)
 
-    version_data.check_in = str(now)
-
-
-def aggregateVersionData(data: VersionData, added: VersionData, added_org_name: str):
-    known_workloads: set[str] = set()
-    for v in data.versions.values():
-        known_workloads.update(v.workloads.keys())
-    for version, version_data in added.versions.items():
-        for workload, workload_data in version_data.workloads.items():
-            # skip if our current version data does not contain this remote workload
-            if workload not in known_workloads:
-                continue
-            w = data.workload_history(version, workload)
-            w.soak_days += workload_data.soak_days
-            ocm_clusters = [
-                f"{added_org_name}/{cluster}" for cluster in workload_data.reporting
-            ]
-            w.reporting += ocm_clusters
+    version_data.check_in = now
 
 
 def get_version_data_map(
@@ -161,7 +143,8 @@ def get_version_data_map(
         if not dry_run:
             version_data.save(state, ocm_name)
 
-    # inherit data from other ocm orgs
+    # aggregate data from other ocm orgs
+    # this is done *after* saving the state: we do not store the other orgs data in our state.
     for ocm_name in ocm_map.instances():
         ocm = ocm_map[ocm_name]
         for other_ocm in ocm.inheritVersionData:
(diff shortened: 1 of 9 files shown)


reconcile/ocm_upgrade_scheduler.py
```diff
@@ -141,21 +131,41 @@ def get_version_history(dry_run, upgrade_policies, ocm_map):
...
+            if ocm_name == other_ocm_name:
+                raise ValueError(
+                    f"[{ocm_name}] OCM organization inherits version data from itself"
+                )
+            if ocm.name not in [
+                o["name"] for o in other_ocm.get("publishVersionData") or []
+            ]:
+                raise ValueError(
+                    f"[{ocm_name}] OCM organization inherits version data from {other_ocm_name}, but this data is not published to it: missing publishVersionData in {other_ocm_name}"
+                )
```
geoberle, line 164: ❤️

reconcile/ocm_upgrade_scheduler.py
```diff
@@ -67,72 +70,59 @@ def fetch_desired_state(
...
     """
-    default_workload_history = {
-        "soak_days": 0.0,
-        "reporting": [],
-    }
-
     now = datetime.utcnow()
-    check_in = parser.parse(get_or_init(history, "check_in", str(now)))
-    versions = get_or_init(history, "versions", {})
+    check_in = parser.parse(version_data.check_in or str(now))
```
geoberle, line 81: can you explain the wording `check_in` in this context?

reconcile/utils/cluster_version_data.py
```diff
@@ -0,0 +1,54 @@
...
+        return vh.workloads.setdefault(workload, default)
+
+    def asdict(self) -> dict[str, Any]:
+        return asdict(self)
+
+    def save(self, state: State, ocm_name: str) -> None:
+        state.add(ocm_name, asdict(self), force=True)
+
+
+def version_data_from_dict(d: dict[str, Any]) -> VersionData:
```
geoberle, line 39: this is the function that bastically restores `VersionData` from a json doc stored in `State` right?

if you use `pydantic` dataclasses over regular ones, you get object tree unmarshalling for free like

```python
from pydantic.dataclasses import dataclass

@dataclass
class VersionData:
   ...

vd = VersionData(**d)
```

reconcile/utils/ocm.py
```diff
@@ -625,6 +627,9 @@ def __init__(
 
         self._init_blocked_versions(blocked_versions)
 
+        self.inheritVersionData = inheritVersionData or []
+        self.publishVersionData = publishVersionData or []
```
geoberle, line 631: ❤️

reconcile/utils/ocm.py
```diff
@@ -602,6 +602,8 @@ def __init__(
         blocked_versions=None,
         ocm_client: Optional[OCMBaseClient] = None,
         sectors: Optional[list[dict[str, Any]]] = None,
+        inheritVersionData: Optional[list[dict[str, Any]]] = None,
+        publishVersionData: Optional[list[dict[str, Any]]] = None,
```
geoberle, line 606: i'm not a huge fan of unstructured list-dicts but i get that it would be cumbersome to come up with a dataclass just to represent this and using `qenerate` would be a bigger refactor.

idea: simplify to `list[str]` that holds the names of the OCM orgs to inherit from and publish to and verify at the place where the `OCM` object is initialized, that inherit and publish are consistent.

feel free to ignore.

reconcile/ocm_upgrade_scheduler.py
patjlm: this is from the previous implementation. It is the date/time of the last data update from the integration.

reconcile/utils/cluster_version_data.py
patjlm: great, thanks! I will have a look.

reconcile/utils/ocm.py
patjlm: I wanted to keep the whole OCM info in order to have access to other fields if needed. Though it is not needed today. Will think through this.

geoberle: ah understood. good point.

reconcile/utils/cluster_version_data.py
patjlm: I went for Pydantic `BaseModel` instead of `dataclass` in order to more easily encode/decode datetime... and then I ended up just modifying the format to use the default iso format. I still think using `BaseModel` is fine though. Let me know what you think

geoberle: nice. yes both pydantic options help with the tree unmarshalling
thanks!

geoberle (approved):
lgtm

PR #3010, opened 2022-11-27
[openshift-base] is namespace deleted

maorfr:
instead of duplicating `namespace_info.get("delete")`

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -1006,5 +1006,5 @@ def determine_user_keys_for_access(
     return user_keys
 
 
-def is_namespace_deleted(namespace_info: dict[str, Any]) -> bool:
+def is_namespace_deleted(namespace_info: Mapping) -> bool:
     return bool(namespace_info.get("delete"))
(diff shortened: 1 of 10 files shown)


reconcile/test/test_openshift_base.py
```diff
@@ -645,3 +645,18 @@ def test_determine_user_keys_for_access_not_implemented():
     auth = {"service": "not-implemented"}
     with pytest.raises(NotImplementedError):
         sut.determine_user_keys_for_access("cluster-name", [auth])
+
```
geoberle, line 648: add another one for `ns = {"delete": None}`

reconcile/openshift_base.py
```diff
@@ -1004,3 +1004,7 @@ def determine_user_keys_for_access(
                 f"[{cluster_name}] auth service not implemented: {service}"
             )
     return user_keys
+
+
+def is_namespace_deleted(namespace_info: Mapping) -> bool:
```
maorfr, line 1009: using `Mapping` for consistency with https://github.com/app-sre/qontract-reconcile/blob/8bca088d7bd0dcfc4ab437a6fc272e80be2baa05/reconcile/openshift_base.py#L72-L73

reconcile/test/test_openshift_base.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/3010/commits/d79b6a4647c415eab7a929f040a8c6d0728efe58

geoberle (approved):
lgtm

PR #3013, opened 2022-11-27
PEP 585 imports

rporres:
Fix deprecated imports following [PEP 585](https://peps.python.org/pep-0585/) recommendations.


geoberle (approved):
thank you!

chassing (approved):
awesome ❤️

PR #3017, opened 2022-11-28
glitchtip queries adjustments

maorfr:
follow up on https://github.com/app-sre/qontract-schemas/pull/334

this PR changes the way the glitchtip integrations queries for projects. instead of going through apps, we query them directly.


chassing (approved):
nice. LGTM

PR #3020, opened 2022-11-29
add delete to namespaces query for openshift-resources

maorfr:
without querying this field, openshift-resources based integrations can not ignore deleted namespaces.

another fix on top of #2942


kwilczynski (approved):
@maorfr, nice one! One more!

PR #3021, opened 2022-11-29
qenerate is now pep-585 compliant

fishi0x01:
apply recent qenerate update https://github.com/app-sre/qenerate/pull/61


rporres (approved):
awesome ❤️

PR #3022, opened 2022-11-29
make RDS work

fishi0x01:
- username is required and cannot be optional
- is_production is implicitly set
- family is implicitly calculated (postgres only)

Corresponding schema change https://github.com/app-sre/qontract-schemas/pull/338


reconcile/cna/assets/aws_rds.py
```diff
@@ -48,6 +52,14 @@ def provider() -> str:
     def asset_type() -> AssetType:
         return AssetType.AWS_RDS
 
+    @staticmethod
+    def determine_family(config: CNARDSInstanceDefaultsV1) -> str:
```
geoberle, line 56: nice!!

geoberle (approved):
LGTM!

PR #3027, opened 2022-11-29
Add test about amtool minimal version

rporres:
Or tests silently fail not giving information about the issue.

### DIFF
--- a/reconcile/utils/amtool.py
+++ b/reconcile/utils/amtool.py
@@ -11,7 +11,7 @@ from subprocess import (
 class AmtoolResult:
     """This class represents a amtool command execution result"""
 
-    def __init__(self, is_ok, message):
+    def __init__(self, is_ok: bool, message: str) -> None:
         self.is_ok = is_ok
         self.message = message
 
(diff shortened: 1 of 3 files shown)


chassing:
Please also enable full type hints check in `setup.cfg`
```
[mypy-reconcile.test.test_amtool]
check_untyped_defs = True
```

reconcile/utils/amtool.py
```diff
@@ -45,6 +46,19 @@ def config_routes_test(yaml_config: str, labels: Mapping[str, str]) -> AmtoolRes
...
 
 
+def version() -> AmtoolResult:
+    """Returns the version parsed from amtool --version"""
+    result = _run_cmd(["amtool", "--version"])
+
+    pattern = re.compile("^amtool, version (?P<version>[^ ]+) .+")
+    if m := pattern.match(result.message):
+        return AmtoolResult(True, m.group("version"))
+    else:
```
chassing, line 56: this `else` branch isn't needed

chassing (approved):
Some nits. Otherwise LGTM

reconcile/utils/amtool.py
rporres: ah, yes, return false by default. Sure that's more idiomatic, fixed in d31e81f4

PR #3028, opened 2022-11-29
[change-owners] auto approve when bot is the only approver

maorfr:
part of https://issues.redhat.com/browse/APPSRE-6638

in some cases, we want changes to be automatically accepted and merged. such cases do not require a human review.

to define such cases, we can use the bot user behind our automations as an approver. if the bot is the only approver for a change - consider the change approved.

app-interface implementation: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/52773


geoberle (approved):
❤️ ❤️‍🔥

PR #3029, opened 2022-11-30
new ocm-oidc-idp integration

chassing:
* [ocm-oidc-idp] new integration to configure OIDC IDPs for OSD clusters in OCM
* [ocm] OIDC IDP support
* Refactored `disabled.integrations` for clusters and accounts with gql classes support

Hints for the reviewer:
* Review commit by commit
* Ran `ocm-oidc-idp` locally and configured OIDC for `appint-ex-01` (See [app-interface-dev-data MR](https://gitlab.cee.redhat.com/app-sre/app-interface-dev-data/-/merge_requests/25)) 
* Feel free to run the integration by yourself with [qontract-development-cli](https://github.com/app-sre/qontract-development-cli)
  `qd profile create --app-interface ~/workspace/gitlab-cee/cassing/app-interface-dev-data --app-interface-pr 25 --qontract-schemas-pr 341 --qontract-reconcile-pr 3029 --integration-name ocm-oidc-idp --integration-extra-args '--vault-input-path app-sre/integrations-input' ocm-oidc-idp`

Depends on: 
* https://github.com/app-sre/qontract-schemas/pull/341
* https://github.com/app-sre/qontract-schemas/pull/350

APPSRE 6622

PR #3030, opened 2022-11-30
set minimum test coverage to 39%

chassing:


patjlm (approved):

rporres (approved):

steveteahan:
My only suggestion might be that we stick to at least 1% buffer. But if we want to try with less we can see how it goes. Ideally we should really get to a place where the blocker is based on the changes in your PR rather than a global blocker, but I haven't yet found the time to figure out the simplest way to achieve that.

chassing:
Let's stick with 39%. With https://github.com/app-sre/qontract-reconcile/pull/3029, the coverage will already increase to 39.74%.

PR #3032, opened 2022-11-30
[openshift-users] OIDC support

chassing:
`openshift-users` must consider all user identities based on cluster auth configuration, not just GitHub ones.

APPSRE-6646

### DIFF
--- a/reconcile/openshift_users.py
+++ b/reconcile/openshift_users.py
@@ -43,11 +43,9 @@ def get_cluster_users(cluster, oc_map, clusters):
             # ignore admins
             continue
         if any(
-            [
-                identity.startswith(identity_prefix)
-                for identity in u.get("identities", [])
-                for identity_prefix in identity_prefixes
-            ]
+            identity.startswith(identity_prefix)
+            for identity in u.get("identities", [])
+            for identity_prefix in identity_prefixes
         ):
             # the user has at least one identitiy which is managed by app-interface
             users.append(u["metadata"]["name"])
--- a/reconcile/openshift_users.py
+++ b/reconcile/openshift_users.py
@@ -1,6 +1,5 @@
 import itertools
 import logging
-import sys
 
 from sretoolbox.utils import threaded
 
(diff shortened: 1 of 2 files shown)


reconcile/openshift_users.py
```diff
@@ -1,34 +1,56 @@
...
-        and u["identities"][0].startswith("github")
-        and not u["metadata"].get("labels", {}).get("admin", "")
-    ]
+    users: list[str] = []
 
-    return [{"cluster": cluster, "user": user} for user in users or []]
+    cluster_info = None
+    for cl in clusters:
+        if cl["name"] == cluster:
+            cluster_info = cl
```
jfchevrette, line 28: nit: I like this fancy way of finding an item from a list using a generator expression. This is very subjective so feel free to just close this :)
```suggestion
cluster_info = next((cl for cl in clusters if cl["name"] == cluster), None)
```

chassing: I like your way to find an item in a list :) Thx for that

reconcile/openshift_users.py
```diff
@@ -14,21 +15,38 @@
 QONTRACT_INTEGRATION = "openshift-users"
 
 
-def get_cluster_users(cluster, oc_map):
+def get_cluster_users(cluster, oc_map, clusters):
```
geoberle, line 18: it would be cool to have a `starmap` version of `threaded.run`, then we could do something like

```
def get_cluster_users(cluster_info, ocm_cluster):
   ...


results = threaded.run(
        get_cluster_users,
        [
          (cluster_info, ocm_cluster) ... rest of the expression that generates a list of cluster_info,ocm_cluster tuples
        ]
        thread_pool_size,
    )
```

the situation that `cluster_info` would be None (or at least the mypy required None check), could be avoided right from the beginning.

reconcile/openshift_users.py
```diff
@@ -14,21 +15,38 @@
...
-        u["metadata"]["name"]
-        for u in oc.get_users()
-        if u.get("identities", None) is not None
-        and len(u["identities"]) == 1
-        and u["identities"][0].startswith("github")
-        and not u["metadata"].get("labels", {}).get("admin", "")
-    ]
+    users: list[str] = []
+
+    cluster_info = next((cl for cl in clusters if cl["name"] == cluster), None)
```
geoberle, line 25: if cluster_info can't be None (and from the rest of the code my understanding is as well that it can't be None), the following change would spare you from the `None` check (i assume it is mostly here because of mypy)
```suggestion
    cluster_info = next((cl for cl in clusters if cl["name"] == cluster))
```

geoberle (approved):
only nits

reconcile/openshift_users.py
chassing: That does the trick. Thx

reconcile/openshift_users.py
chassing: I want to keep the current solution, it is the less intrusive

bkez322:
Related: https://coreos.slack.com/archives/GGC2A0MS8/p1670374909268579

PR #3033, opened 2022-11-30
update qr base and builder images to use latest version

jfchevrette:
Promote the changes from https://github.com/app-sre/container-images/pull/61

Ref.:[APPSRE-6583](https://issues.redhat.com/browse/APPSRE-6583)


jfchevrette:
Change is minor and simply following an established process so I will merge without review

PR #3034, opened 2022-11-30
Fix issues when a cluster is deleted in OCM

patjlm:
This fixes the case where a cluster is removed from OCM but remains in the ocm schema definition.
However, there seem to be other subsequent issues in this case. I will need to dig more.


kwilczynski (approved):
@patjlm, thank you for the fix!

PR #3038, opened 2022-12-01
Terraform-users use reencryption integration

janboll:
This introduces pgp re-encryption support to terraform-users. It is added in a backward-compatible manner. A feature toggle added makes it possible to filter AWS accounts for which re-encryption should not be used. 

Related re-encryption integration:  https://github.com/app-sre/user-validator/pull/32

Is depending on schema changes: https://github.com/app-sre/qontract-schemas/pull/342

Related issue: https://issues.redhat.com/browse/APPSRE-4706


janboll:
Retest

reconcile/terraform_users.py
```diff
@@ -151,11 +198,37 @@ def run(
...
     # setup errors should skip resources that will lead
     # to terraform errors. we should still do our best
     # to reconcile all valid resources for all accounts.
+    additional_setup_args = {
+        "skip_reencrypt_accounts": None,
+        "appsre_pgp_key": None,
+    }
+    if reencrypt_settings is not None:
+        additional_setup_args["skip_reencrypt_accounts"] = skip_accounts
+        additional_setup_args["appsre_pgp_key"] = reencrypt_settings.public_gpg_key
```
suzana-nesic, line 225: I believe it should be switched to this to avoid the `Incompatible types in assignment` error

suzana-nesic (changes requested):

suzana-nesic:
I believe the gql_definitions error is coming from `pgp_reencrypt_settings_v1` not being found in `reconcile/gql_definitions/introspection.json`

reconcile/terraform_users.py
janboll: Right, removed!

suzana-nesic (approved):

geoberle (commented):
looks good.
small nit about the skip account condition in the `write_user_to_vault` and `send_email_invite`. both functions would progress with their functionality if `skip_reencrypt_accounts` would be None. right now you control that correctly from the `run` function but it would be easy to introduce an error that way. maybe make `skip_reencrypt_account` mandatory, e.g. just pass in an empty list

janboll:
Good point, I think this makes the code way simpler! I'll add a commit for this

geoberle (approved):
LGTM!

janboll:
Had to add another commit to update the qontract-cli get command that fetches the user credentials. Otherwise, this would have been broken once this change got rolled out.

PR #3042, opened 2022-12-02
Support multiple aws accounts to act in terraform resources

esron:
JIRA: [APPSRE-6586]( https://issues.redhat.com/browse/APPSRE-6586)

## Testing
With `qontract-server` configured with `app-interface-dev-data` do the following.
Run `terraform-resources` without any account filter in dry-run mode:
```
APP_INTERFACE_STATE_BUCKET="a" APP_INTERFACE_STATE_BUCKET_ACCOUNT="a" qontract-reconcile --config config.dev.toml --log-level INFO --dry-run terraform-resources
```

Run `terraform-resources` passing one account name filter in dry-run mode:
```
 APP_INTERFACE_STATE_BUCKET="a" APP_INTERFACE_STATE_BUCKET_ACCOUNT="a" qontract-reconcile --config config.dev.toml --log-level INFO --dry-run terraform-resources --account-name=app-int-example-01
```

Verify that the resources belonging to account `app-int-example-01` appears in the second output are the same that appear in the first output.
You can do the same with `app-int-example-02` account.

Run `terraform-resources` passing two account names filter in dry-run mode:
```
APP_INTERFACE_STATE_BUCKET="a" APP_INTERFACE_STATE_BUCKET_ACCOUNT="a" qontract-reconcile --config config.dev.toml --log-level INFO --dry-run terraform-resources --account-name=app-int-example-01 --account-name=app-int-example-02
```

Verify that the output is the same for running without filters, since in `app-interface-dev-data` there are only two accounts setup.

Now, to verify that we can only run with more than one `--ccount-name` in dry run:
```
APP_INTERFACE_STATE_BUCKET="a" APP_INTERFACE_STATE_BUCKET_ACCOUNT="a" qontract-reconcile --config config.dev.toml --log-level INFO terraform-resources --account-name=app-int-example-01 --account-name=app-int-example-02
```

This should give a `RuntimeError`.

To verify that we catch typos or passing a account that doesn't exists:
```
APP_INTERFACE_STATE_BUCKET="a" APP_INTERFACE_STATE_BUCKET_ACCOUNT="a" qontract-reconcile --config config.dev.toml --log-level INFO --dry-run terraform-resources --account-name=app-int-example-01 --account-name=app-int-example
```

This should give a `ValueError`.

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -344,7 +344,7 @@ def account_name_multiple(function):
         "--account-name",
         default=None,
         multiple=True,
-        help="aws account names to act on comma separated i.e.: --account-name aws-account-1 --account-name aws-account-2",
+        help="aws account names to act on i.e.: --account-name aws-account-1 --account-name aws-account-2",
     )(function)
 
     return function
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -337,6 +337,7 @@ def account_name(function):
 
     return function
 
+
 def account_name_multiple(function):
     """This option can be used when more than one account needs to be passed as argument"""
     function = click.option(
@@ -347,6 +348,7 @@ def account_name_multiple(function):
     )(function)
 
     return function
+
 
 def workspace_name(function):
     function = click.option(
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -337,6 +337,16 @@ def account_name(function):
 
     return function
 
+def account_name_multiple(function):
+    """This option can be used when more than one account needs to be passed as argument"""
+    function = click.option(
+        "--account-name",
+        default=None,
+        multiple=True,
+        help="aws account names to act on comma separated i.e.: --account-name aws-account-1 --account-name aws-account-2",
+    )(function)
+
+    return function
 
 def workspace_name(function):
     function = click.option(
@@ -1571,12 +1581,7 @@ def ldap_users(ctx, gitlab_project_id):
 @internal()
 @use_jump_host()
 @enable_deletion(default=False)
-@click.option(
-    "--account-name",
-    default=None,
-    multiple=True,
-    help="aws account names to act on comma separated i.e.: aws-account-1,aws-account-2",
-)
+@account_name_multiple
 @click.option(
     "--light/--full",
     default=False,
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1613,7 +1613,7 @@ def terraform_resources(
         use_jump_host,
         light,
         vault_output_path,
-        account_name=account_name,
+        account_name,
     )
 
 
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1571,7 +1571,11 @@ def ldap_users(ctx, gitlab_project_id):
 @internal()
 @use_jump_host()
 @enable_deletion(default=False)
-@account_name
+@click.option(
+    "--account-names",
+    default=None,
+    help="aws account names to act on comma separated i.e.: aws-account-1,aws-account-2",
+)
 @click.option(
     "--light/--full",
     default=False,
@@ -1587,7 +1591,7 @@ def terraform_resources(
     use_jump_host,
     light,
     vault_output_path,
-    account_name,
+    account_names,
 ):
     import reconcile.terraform_resources
 
@@ -1603,7 +1607,7 @@ def terraform_resources(
         use_jump_host,
         light,
         vault_output_path,
-        account_name=account_name,
+        account_names=account_names,
     )
 
 
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1572,8 +1572,9 @@ def ldap_users(ctx, gitlab_project_id):
 @use_jump_host()
 @enable_deletion(default=False)
 @click.option(
-    "--account-names",
+    "--account-name",
     default=None,
+    multiple=True,
     help="aws account names to act on comma separated i.e.: aws-account-1,aws-account-2",
 )
 @click.option(
@@ -1591,7 +1592,7 @@ def terraform_resources(
     use_jump_host,
     light,
     vault_output_path,
-    account_names,
+    account_name,
 ):
     import reconcile.terraform_resources
 
@@ -1607,7 +1608,7 @@ def terraform_resources(
         use_jump_host,
         light,
         vault_output_path,
-        account_names=account_names,
+        account_name=account_name,
     )
 
 
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -689,14 +689,14 @@ def run(
     defer=None,
 ):
     # If we are not running in dry run we don't want to run with more than one account
-    if account_name and len(account_name) > 1 and not dry_run:
+    account_names = account_name
+    if account_names and len(account_names) > 1 and not dry_run:
         logging.error(
             "Running with multiple accounts is only supported in dry-run mode"
         )
         sys.exit(1)
-    elif account_name:
-        account_names = account_name
-        account_name = account_name[0]
+    elif account_names:
+        account_name = account_names[0]
 
     ri, oc_map, tf, resource_specs = setup(
         dry_run,
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -691,10 +691,9 @@ def run(
     # If we are not running in dry run we don't want to run with more than one account
     account_names = account_name
     if account_names and len(account_names) > 1 and not dry_run:
-        logging.error(
-            "Running with multiple accounts is only supported in dry-run mode"
-        )
-        sys.exit(1)
+        message = "Running with multiple accounts is only supported in dry-run mode"
+        logging.error(message)
+        raise RuntimeError(message)
     elif account_names:
         account_name = account_names[0]
 
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -558,8 +558,16 @@ def setup(
     accounts = queries.get_aws_accounts(terraform_state=True)
     if account_names:
         accounts = [n for n in accounts if n["name"] in account_names]
-        if not accounts:
-            raise ValueError(f"aws accounts {account_names} where not found")
+        if len(accounts) < len(account_names):
+            # Some of the passed account names don't exist in app-interface
+            acc_names = tuple(
+                a
+                for a in account_names
+                if a not in tuple(account["name"] for account in accounts)
+            )
+            raise ValueError(
+                f"Accounts {acc_names} where not found in account names, check your input"
+            )
     settings = queries.get_app_interface_settings()
 
     # build a resource inventory for all the kube secrets managed by the
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -8,6 +8,7 @@ from collections.abc import (
 from textwrap import indent
 from typing import (
     Any,
+    Collection,
     Optional,
     cast,
 )
@@ -553,7 +554,7 @@ def setup(
     thread_pool_size: int,
     internal: str,
     use_jump_host: bool,
-    account_names: Optional[tuple[str]],
+    account_names: Optional[Collection[str]],
 ) -> tuple[ResourceInventory, OC_Map, Terraform, ExternalResourceSpecInventory]:
     accounts = queries.get_aws_accounts(terraform_state=True)
     if account_names:
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -694,7 +694,7 @@ def run(
     use_jump_host=True,
     light=False,
     vault_output_path="",
-    account_name: Optional[tuple[str]] = None,
+    account_name: Optional[Collection[str]] = None,
     defer=None,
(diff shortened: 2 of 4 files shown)


esron:
I know that we use --account-name for running the integration in production. So, I've been thinking about changing that to --account-names and preventing to being called with more than one account without dry-run.

esron:
> I know that we use --account-name for running the integration in production. So, I've been thinking about changing that to --account-names and preventing to being called with more than one account without dry-run.

This will affect the way `integrations-manager` deals with `terraform-resources`. So, another thing that I can do is add a new argument that can only be used with dry run that does what we intend here. But that will add some ambiguity and redundancy.

steveteahan:
> > I know that we use --account-name for running the integration in production. So, I've been thinking about changing that to --account-names and preventing to being called with more than one account without dry-run.
> 
> This will affect the way `integrations-manager` deals with `terraform-resources`. So, another thing that I can do is add a new argument that can only be used with dry run that does what we intend here. But that will add some ambiguity and redundancy.

Mentioned in Slack, but could you keep `--account-name` if it could be passed multiple times with something like [this](https://click.palletsprojects.com/en/8.1.x/options/#multiple-options). That could maintain backwards compatibility without needing to add a new CLI option?

esron:
Yes, that would suffice.

steveteahan (changes requested):
Overall the change makes sense to me and seems like it's headed in the right direction. I took a very quick look at this as a first pass for some early comments. Some open questions I have are:

* What is your plan for testing these changes manually for anything that isn't covered by unit tests? Maybe we could list out the test cases?

* I don't work on this integration all the time, but I would like to take a closer look to see if there could be any edge cases introduced by now having dry-run that supports multiple accounts, but without dry-run only a single account. For instance, any places where the methods that were changed to accept an `Iterable` are passing in a string. Keep in mind that Mypy isn't always going to catch these because we have places in the code where type hints aren't provided. The test plan above should help catching these.

Feel free to reach out if any of this doesn't make sense.

reconcile/cli.py
```diff
@@ -1586,7 +1586,12 @@ def ldap_users(ctx, gitlab_project_id):
 @internal()
 @use_jump_host()
 @enable_deletion(default=False)
-@account_name
+@click.option(
```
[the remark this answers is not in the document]
esron: Yes. That's a good idea.

reconcile/terraform_resources.py
```diff
@@ -676,14 +683,23 @@ def run(
     account_name=None,
     defer=None,
 ):
+    # If we are not running in dry run we don't want to run with more than one account
+    if account_name and len(account_name) > 1 and not dry_run:
```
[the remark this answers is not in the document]
esron: `account_name` will hold a tuple of accounts. To change that we would need to go back to something like 
`--account-names=account-1 --account-names=account-2`

reconcile/terraform_resources.py
```diff
@@ -676,14 +683,23 @@ def run(
...
+        logging.error(
+            "You can only pass two account names when running in dry-run mode"
+        )
+        sys.exit(1)
+    elif account_name:
+        account_names = account_name
```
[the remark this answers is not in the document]
esron: I've updated that. Probably one reading this would need a bit of context in the problem.

reconcile/terraform_resources.py
```diff
@@ -541,20 +548,20 @@ def setup(
     thread_pool_size: int,
     internal: str,
     use_jump_host: bool,
-    account_name: Optional[str],
+    account_names: Optional[str],
```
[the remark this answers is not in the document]
esron: I fixed, Still have no idea why it didn't caught this. Other errors where caught.

reconcile/terraform_resources.py
```diff
@@ -676,14 +683,23 @@ def run(
     account_name=None,
     defer=None,
 ):
+    # If we are not running in dry run we don't want to run with more than one account
```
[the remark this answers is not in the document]
esron: Done.

esron:
> Overall the change makes sense to me and seems like it's headed in the right direction. I took a very quick look at this as a first pass for some early comments. Some open questions I have are:
> 
> * What is your plan for testing these changes manually for anything that isn't covered by unit tests? Maybe we could list out the test cases?
> * I don't work on this integration all the time, but I would like to take a closer look to see if there could be any edge cases introduced by now having dry-run that supports multiple accounts, but without dry-run only a single account. For instance, any places where the methods that were changed to accept an `Iterable` are passing in a string. Keep in mind that Mypy isn't always going to catch these because we have places in the code where type hints aren't provided. The test plan above should help catching these.
> 
> Feel free to reach out if any of this doesn't make sense.

@steveteahan ,
There are tree main cases here.
1. No account is passed:
In this case `acc_name` will be None and the code will run as before the addition of account sharding.
2. One account is passed:
`account_names` will be a tuple with a single value.
The checks will check considering only this tuple.
`acc_name` will be the account passed.
All function that receive one account_name will receive `acc_name`.

3. More than one account is passed:
We check if we are running in dry-run mode, if not we fail and exit.
If yes, `account_names` will be a tuple with the account names passed and all the accounts will be considered for the checks.
The code exits before any function that receives only one account is called.

reconcile/terraform_resources.py
```diff
@@ -541,20 +548,20 @@ def setup(
...
+    account_names: Optional[str],
 ) -> tuple[ResourceInventory, OC_Map, Terraform, ExternalResourceSpecInventory]:
     accounts = queries.get_aws_accounts(terraform_state=True)
-    if account_name:
-        accounts = [n for n in accounts if n["name"] == account_name]
+    if account_names:
+        accounts = [n for n in accounts if n["name"] in account_names]
         if not accounts:
-            raise ValueError(f"aws account {account_name} is not found")
+            raise ValueError(f"aws accounts {account_names} where not found")
```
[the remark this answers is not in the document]
esron: I think that if one account passed is missing in `accounts` it can because of an typo or something. So we could calculate the diff and let the user know what account is missing in `accounts` and then fail safely.

reconcile/terraform_resources.py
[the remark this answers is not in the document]
steveteahan: Where is `setup` called from? Maybe this could give you a hint: https://mypy.readthedocs.io/en/stable/common_issues.html#no-errors-reported-for-obviously-wrong-code ?

reconcile/terraform_resources.py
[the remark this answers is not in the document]
steveteahan: Makes sense!

steveteahan:
> > Overall the change makes sense to me and seems like it's headed in the right direction. I took a very quick look at this as a first pass for some early comments. Some open questions I have are:
> > 
> > * What is your plan for testing these changes manually for anything that isn't covered by unit tests? Maybe we could list out the test cases?
> > * I don't work on this integration all the time, but I would like to take a closer look to see if there could be any edge cases introduced by now having dry-run that supports multiple accounts, but without dry-run only a single account. For instance, any places where the methods that were changed to accept an `Iterable` are passing in a string. Keep in mind that Mypy isn't always going to catch these because we have places in the code where type hints aren't provided. The test plan above should help catching these.
> > 
> > Feel free to reach out if any of this doesn't make sense.
> 
> @steveteahan , There are tree main cases here.
> 
> 1. No account is passed:
>    In this case `acc_name` will be None and the code will run as before the addition of account sharding.
> 2. One account is passed:
>    `account_names` will be a tuple with a single value.
>    The checks will check considering only this tuple.
>    `acc_name` will be the account passed.
>    All function that receive one account_name will receive `acc_name`.
> 3. More than one account is passed:
>    We check if we are running in dry-run mode, if not we fail and exit.
>    If yes, `account_names` will be a tuple with the account names passed and all the accounts will be considered for the checks.
>    The code exits before any function that receives only one account is called.

So, it's those tests **with and without dry-run flags**. In addition, what about detecting typos like we mentioned? What if you have multiple accounts and one is a typo? Single account and one is a typo? I'd add a "Testing" section to the PR to track that these conditions have been tested. I'm assuming these were all tested with app-interface-dev data?

esron:
@steveteahan I've updated the RP description to contain instructions for manual testing.

reconcile/terraform_resources.py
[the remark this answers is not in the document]
esron: Oh, I got it now. I've added a return type to `run` and now it is checking.

reconcile/test/test_terraform_resources.py
```diff
@@ -1,6 +1,27 @@
...
+    with pytest.raises(RuntimeError) as excinfo:
+        integ.run(False, account_name=("a", "b"))
+
+    assert "Running with multiple accounts is only supported in dry-run mode" in str(
+        excinfo.value
+    )
+
+
+def test_cannot_pass_invalid_aws_account(mocker):
+    mocker.patch("reconcile.queries.get_aws_accounts", return_value=[{"name": "a"}])
```
steveteahan, line 16: `autospec=True` in patch will ensure that you pass the correct number of arguments. It's generally a best practice for all patching.

steveteahan (commented):
Really just some nits left. Overall looks good!

reconcile/cli.py
```diff
@@ -316,6 +316,18 @@ def account_name(function):
     return function
 
 
+def account_name_multiple(function):
+    """This option can be used when more than one account needs to be passed as argument"""
+    function = click.option(
+        "--account-name",
+        default=None,
+        multiple=True,
+        help="aws account names to act on comma separated i.e.: --account-name aws-account-1 --account-name aws-account-2",
```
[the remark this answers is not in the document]
esron: Solved.

esron:
@steveteahan,  Apparently one cannot take the `len` of an `Iterable`.

steveteahan:
> @steveteahan, Apparently one cannot take the `len` of an `Iterable`.

Maybe something else here: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes ?

The motivation is to allow for general input types, but the most specific output types. `list`, `set`, `tuple`, etc. are often interchangeable for trivial use case where we're just iterating through data, or maybe checking how many items are in collections, etc.

steveteahan (approved):

reconcile/terraform_resources.py
```diff
@@ -717,16 +745,14 @@ def run(
...
-    )
+    actions = ob.realize_data(dry_run, oc_map, ri, thread_pool_size, caller=acc_name)
 
     if not light and tf.should_apply:
         disable_keys(
             dry_run,
             thread_pool_size,
             disable_service_account_keys=True,
-            account_name=account_name,
+            account_name=acc_name,
```
geoberle, line 755: account_name acts as a filter in disable_keys. by calling it this way, disable_key will only act on the first passed account

reconcile/terraform_resources.py
```diff
@@ -717,16 +745,14 @@ def run(
     # populate the resource inventory with latest output data
     populate_desired_state(ri, resource_specs)
 
-    actions = ob.realize_data(
-        dry_run, oc_map, ri, thread_pool_size, caller=account_name
-    )
+    actions = ob.realize_data(dry_run, oc_map, ri, thread_pool_size, caller=acc_name)
```
geoberle, line 748: i have not looked into the details of realize_data but it has an `all_callers` parameter and maybe that is better suited than just passing in the first account name

reconcile/terraform_resources.py
```diff
@@ -673,17 +689,29 @@ def run(
     use_jump_host=True,
     light=False,
     vault_output_path="",
-    account_name=None,
+    account_name: Optional[tuple[str]] = None,
```
geoberle, line 692: any particular reason to use tuple here instead of a generic collection type like `Sequence` or `Iterable`?

esron: I think this is safer, because AFAIK we can't change the tipe of  `click` inputs.

reconcile/terraform_resources.py
esron: The idea is that we never get here with more than one account.

reconcile/terraform_resources.py
esron: The idea is that we never get here with more than one account.

reconcile/terraform_resources.py
geoberle: Why is that?

reconcile/terraform_resources.py
geoberle: What do you mean by safer?

The type of the click Param is one thing. The type of the run function can be whatever we want.

geoberle: I just saw that you changed it to collection. Works for me.

reconcile/terraform_resources.py
geoberle: the code does not prevent that though and there might be the impression that it is a supported run model.
you could put some guarding code in place that fails the integration if it reaches this part of the code with more than one account_name

esron: There is even a test for that
https://github.com/app-sre/qontract-reconcile/pull/3042/files#diff-aa3407e2884577fb05f6ed1925376a267ff8655228672c67c899a3bafe4c7497R6-R29

geoberle (approved):

PR #3044, opened 2022-12-06
add support to manage routes in account-vpc on the aws side

maorfr:
probably the oldest tech debt around vpc peerings.

when we started out, we only had `account-vpc`. it could mostly manage the peering connection, and that's it. no route management support.

until today!

depends on https://github.com/app-sre/qontract-schemas/pull/347

`manageRoutes` currently exists on `account-vpc` connections to say "manage routes on the cluster side". we are adding a new field for backwards compatibility, and to allow a migration from routes managed elsewhere to being managed via app-interface.

once all `account-vpc` connections have both `manageRoutes` and `manageAccountRoutes`, we can refactor this PR and remove `manageAccountRoutes` altogether.

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -1005,6 +1005,7 @@ CLUSTER_PEERING_QUERY = """
             region
           }
           assumeRole
+          manageAccountRoutes
         }
         ... on ClusterPeeringConnectionAccountVPCMesh_v1 {
           account {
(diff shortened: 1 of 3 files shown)


patjlm (approved):
Did not go further in the flow, but I suppose `accepter["route_table_ids"]` is then handled correctly, as with other providers.
LGTM

PR #3052, opened 2022-12-07
[change-owners] MR with no changes is not self-serviceable + more logs

geoberle:
an MR that has no detected changes, will not be marked as self-serviceable anymore. additonally, several log statements for info and debug loglevel have been introduce to make it easier to figure out why no changes were detected on an MR

### DIFF
--- a/reconcile/change_owners/change_owners.py
+++ b/reconcile/change_owners/change_owners.py
@@ -112,30 +112,46 @@ def _parse_bundle_changes(bundle_changes) -> list[BundleFileChange]:
     """
     parses the output of the qontract-server /diff endpoint
     """
-    change_list = [
-        create_bundle_file_change(
+
+    # todo - add some logging to remove BundleFileChange file with no changes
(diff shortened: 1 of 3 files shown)


reconcile/change_owners/change_owners.py
```diff
@@ -319,20 +335,23 @@ def run(
         #   C H A N G E   C O V E R A G E
         #
         changes = fetch_bundle_changes(comparison_sha)
+        logging.info(
+            f"detected {len(changes)} changed bundle files with {sum(len(c.diff_coverage) for c in changes)} differences"
```
patjlm, line 339: nit: i'd remove `bundle` in this log message to avoid thinking "bundle files" are changed (there would be only 1?)

geoberle: good point. done in [099b0f4](https://github.com/app-sre/qontract-reconcile/pull/3052/commits/099b0f41286acf5fb843209c1e74b3ae730cac34)

PR #3055, opened 2022-12-07
Revert "[openshift-users] OIDC support"

TGPSKI:
Reverts app-sre/qontract-reconcile#3032

The OIDC related changes in qontract-reconcile and qontract-schemas break FedRAMP environment authentication. We have to revert and align on a strategy that can exist in parallel.

re: https://coreos.slack.com/archives/GGC2A0MS8/p1670432900179959


dwelch0 (approved):
Acknowledged by members of other streams. Plan to coordinate at later date.

PR #3056, opened 2022-12-09
[cna] CNA support for external-resources

geoberle:
the new `cna` integration introduces the external-resources extension for CNA

It covers the following modules

* `null` - for simple testing purposes without any prereqs on AWS accounts or clusters etc.
* `aws-assume-role` - for testing AWS assume-role access by CNA
* `aws-rds` - create an AWS RDS database

# General approach

This integration is, as many others, a translation engine from one configuration language (app-interface external resource) to another one (CNA API + modules).

A cloud native asset in CNA is represented by a module and a module has various parameters that need to be provided when creating/updating an asset. A module is e.g. an RDS database. Additionally each asset has a name (technically the name is optional but this integration uses the name to know which asset in CNA belongs to witch external resource in app-interface).

A CNA module along with their parameters are represented in qontract-schema within the `/cna/asset-1.yml` schema and the `CNAsset_v1` GQL type. The parameters are not necessarily mapped 1-1. Not all parameters in CNA might be relevant to be specified in app-interface because they might be provided by context (e.g. the region for a RDS can be inferred by the VPC) or might be provided as defaults within the code. Additionally each asset in qontract-schema has a dedicated datafile schema for defaults and provides override options for the defaults. As such the CNA external-resource schema aligns with what `terraform-resources` and `cloudflare-resources` already do (except for the fact that currently we mostly use resource files without schemas for defaults).

From a high level perspective, the CNA integration just translates a-i config objects to CNA asset API calls. Looking deeper it also handles the classical desired state (app-interface config) to current state (CNA world state) comparison and reconciliation.

Unlike the external-resource provisioners in a-i, the CNA provider does not come with a dry-run option because the CNA API does not support a dry-run. Therefore a lot of effort went into validating the data passed to CNA as good as possible by doing the following things:

* have thorough schema, even for defaults
* have the schema represented as dataclasses that are fully type hinted
* have CNA modules as dataclasses that are fully type hinted
* module based tests for the object translation process
* validate configuration for consistency in code where not possible in the schema

This is for sure not sufficient to ensure every asset can be created without issues by CNA but it is as good as we can check without a dry-run and without inspecting AWS account prereqs etc. upfront.

# How to configure app-interface for CNA

## Provisioner

`cna-experimental` serves as the provisioner, indicating the experimental nature of the integration and the `CNAExperimentalProvisioner_v1` serves as the provisioner, wrapping an OCM organization and potentially adding additional config data (we will add a API URL override to this soon so we can target CNA API instances living outside of api.openshift.com)

```yaml
$schema: /cna/experimental-provisioner-1.yml

name: app-sre
description: CNA provisioner for the app-sre OCM organization

ocm:
  $ref: /dependencies/ocm/stage.yml

<<<--- potential additional config options go here

```

## Example RDS

The defaults files for CNA are now datafile schemas as well. As such they enable us to use references, e.g. a VPC

Here is the example of a defaults file for a production DB

```yaml
$schema: /cna/aws-rds-config-1.yml
engine: postgres
username: postgres
engine_version: '14.2'
instance_class: db.m6.large
backup_retention_period: 7
db_subnet_group_name: default
allocated_storage: 200
max_allocated_storage: 1000
multi_az: true
vpc:
  $ref: /aws/some-account/vpcs/default-vpc.yml
```

and the external resource declaration that leverages it

```yaml
$schema: /openshift/namespace-1.yml
...
externalResources:
- provider: cna-experimental 
  provisioner:
    $ref: /cna/app-sre.yml
  resources:
  - provider: aws-rds
    identifier: my-db
    defaults:
      $ref: the-defaults-file.yml
    overrides:
      backup_retention_period: 14
```

design doc: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/53097
ref: https://issues.redhat.com/browse/APPSRE-6295
schema: https://github.com/app-sre/qontract-schemas/pull/355


steveteahan (commented):
Just a couple high-level comments to start. I'll pick back up with reviewing the rest tomorrow.

reconcile/cna/client.py
```diff
@@ -11,8 +20,46 @@ class CNAClient:
...
 
-    def __init__(self, ocm_client: OCMBaseClient):
+    def __init__(self, ocm_client: OCMBaseClient, init_metadata: bool = False):
         self._ocm_client = ocm_client
+        self._metadata = self._init_metadata() if init_metadata else None
+
+    def _init_metadata(self) -> dict[AssetType, AssetTypeMetadata]:
+        asset_types_metadata: dict[AssetType, AssetTypeMetadata] = {}
+        for asset_type_ref in self._ocm_client.get(
+            api_path="/api/cna-management/v1/asset_types"
```
steveteahan, line 30: nit: could we dedupe a lot of these URLs with something like `{CNA_API_V1}/asset_types`?

reconcile/cna/client.py
```diff
@@ -23,15 +70,44 @@ def list_assets(self) -> list[dict[str, Any]]:
         cnas = self._ocm_client.get(api_path="/api/cna-management/v1/cnas")
         return cnas.get("items", [])
 
+    def fetch_bindings_for_asset(self, asset: Asset) -> list[dict[str, str]]:
```
steveteahan, line 73: Generally speaking, what's the failure mode for this client? Do we raise an exception for 5xx? What's the retry strategy? Maybe overkill to document it for each method, but if the behavior is consistent then maybe added it to the `CNAClient` docstring?

If this is ultimately all driven by `OCMBaseClient`, then maybe that can be stated as well?

reconcile/cna/client.py
```diff
@@ -23,15 +70,44 @@ def list_assets(self) -> list[dict[str, Any]]:
...
+                asset.name,
+                asset.raw_asset_parameters(True),
+            )
             return
         self._ocm_client.post(
             api_path="/api/cna-management/v1/cnas",
             data=asset.api_payload(),
         )
 
+    def bind(self, asset: Asset, dry_run: bool = False):
```
steveteahan, line 96: I wonder whether the client should care about `dry_run` at all? I'm not sure how much value it is in adding the logging statement. The caller would already know that it called `bind` and could handle this further up the chain?

reconcile/cna/integration.py
```diff
@@ -57,27 +67,75 @@ def assemble_desired_states(self):
...
+                                cluster_id=binding.get("cluster_id", ""),
+                                namespace=binding.get("namespace", ""),
+                                secret_name=binding.get("secret_name", ""),
+                            )
+                        )
+                    state.add_asset(asset)
+                except UnknownAssetTypeError as e:
+                    logging.warning(e)
+                except AssetError as e:
+                    # TODO: remember this somehow in the state so we don't try to update/create this asset but skip it instead
```
steveteahan, line 123: For now should we just fail? Same goes for `UnknownAssetTypeError` because it seems unlikely to happen, but would result in trying to create the resource?

reconcile/test/cna/test_client.py
```diff
@@ -0,0 +1,45 @@
+from reconcile.cna.client import CNAClient
+
+
+def test_client_asset_type_metadata_init():
+    pass
```
steveteahan, line 5: Placeholder?

reconcile/test/cna/test_client.py
```diff
@@ -0,0 +1,45 @@
...
+            "asset_type": "null",
+            "id": "000",
+            "href": "url/000",
+            "status": "Running",
+            "creator": {"username": "another_user"},
+        },
+    ]
+
+    mocker.patch.object(CNAClient, "list_assets", return_value=listed_assets)
+    cna_client = CNAClient(None)  # type: ignore
```
steveteahan, line 41: nit: would `create_autospec(OCMBaseClient)` work here?

reconcile/test/cna/test_client.py
```diff
@@ -0,0 +1,45 @@
...
+        {
+            "asset_type": "null",
+            "id": "000",
+            "href": "url/000",
+            "status": "Running",
+            "creator": {"username": "another_user"},
+        },
+    ]
+
+    mocker.patch.object(CNAClient, "list_assets", return_value=listed_assets)
```
steveteahan, line 40: nit: reminder on `autospec` usage where it makes sense (there could be a good reason not to here).

steveteahan (commented):
Overall, it looks great. Thank you for the effort on this!

reconcile/cna/assets/aws_rds.py
```diff
@@ -0,0 +1,102 @@
...
+    backup_retention_period: Optional[int] = Field(
+        None, alias="backup_retention_period"
+    )
+    backup_window: Optional[str] = Field(None, alias="backup_window")
+    maintenance_window: Optional[str] = Field(None, alias="maintenance_window")
+    multi_az: Optional[bool] = Field(None, alias="multi_az")
+    deletion_protection: Optional[bool] = Field(None, alias="deletion_protection")
+    apply_immediately: Optional[bool] = Field(None, alias="apply_immediately")
+
+    # Those values are implicit and not set in app-interface
```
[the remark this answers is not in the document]
geoberle: i agree that `is_production` is a good thing for direct users of CNA. when we talk about app-interface, where we also have defaults and overrides, it can get messy to know where the actual settings for an RDS instance are coming from. 

i would tend to ignore the `is_production` settings, providing an arbitrary value for it (in this case `true`), while at the same time making all parameters in the defaults schema mandatory so looking at app-interface is self-sufficient to get the full picture about configuration settings. we can provide meaningful presets as referegcable defaults datafiles.

wdyt?

reconcile/test/cna/test_client.py
geoberle: @fishi0x01 and i were discussing if we should have some hard validation of the metadata provided by the CNA module metadata endpoint with what we know about the modules in code. this is the placeholder test so we don't forget :)

reconcile/cna/client.py
geoberle: good point. fixed in [a577a1c](https://github.com/app-sre/qontract-reconcile/pull/3056/commits/a577a1cf233e97dc47ea6fc2e2913092dcce737f)

reconcile/cna/client.py
geoberle: i added a comment to the class doc string - [443690b](https://github.com/app-sre/qontract-reconcile/pull/3056/commits/443690b2af95f5b2b9ee7a07e6c68596aa7cb62d)

reconcile/cna/client.py
geoberle: i've removed the log statement but kept the dry_run parameter. i still have high hopes that a `dry_run` for bind will be provided, at least some basic validation for it like the existence of the referenced cluster and maybe even the namespace

reconcile/cna/integration.py
geoberle: that would imply that the integration would fail immediately because some assets in our organization would not validate correctly against the metadata schema of the modules. the reason is the current bug in CNA that does not provide proper input parameter metadata in regards to what is mandatory and what not ([default value null](https://issues.redhat.com/browse/OSDEV-854)).

until this issue is fixed or we can have some finer tenancy in an OCM organisation (e.g. with [labels](https://issues.redhat.com/browse/OSDEV-887)), it would be ok-ish to ignore the error instead of provoking an occasional unrelated fail when someone tests CNA modules in the org outside the context of a-i

reconcile/test/cna/test_client.py
geoberle: i autospeced the patch only 

`mocker.patch.object(CNAClient, "list_assets", return_value=listed_assets, autospec=True)`

i verified that this is sufficient to ensure that `list_assets_for_creator` calls `listed_assets` correctly.

is this ok with you or do you prefer `create_autospec`?

reconcile/test/cna/test_client.py
geoberle: maybe covered in the other autospec related comment?

reconcile/cna/assets/aws_rds.py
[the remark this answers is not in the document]
steveteahan: I think that the usage of `defaults` in app-interface isn't great today. Teams tend to copy them without much thought. The values are only as good as the person who created the file. The defaults provided by `is_production` are better in my (slightly biased) opinion because they are opinionated, and allow for the ability to quickly determine where teams have deviated from these default values.

I look at it like:

* `is_production` provides your sane defaults for most use cases
* `defaults` allows teams to avoid redundant configurations across services

The issue of where values come from is going to be present whether we hide `is_production` or not.

The module documentation might help clarify things a bit as well.

reconcile/cna/client.py
steveteahan: That's fine with me.

reconcile/cna/integration.py
steveteahan: I trust your judgement here and the `TODO` makes it clear this needs to change. I'm fine with this given the experimental designation.

reconcile/test/cna/test_client.py
steveteahan: Worth a docstring or using [xfail](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail) to track?

reconcile/test/cna/test_client.py
steveteahan: I was thinking more of the fact that we're passing `None` in and need to ignore a type error? If we used `create_autospec` as a Mock factory, then I don't think you'd need to do this?

reconcile/test/cna/test_integration.py
```diff
@@ -144,12 +153,23 @@ def null_asset(name: str, addr_block: Optional[str]) -> NullAsset:
...
 def test_integration_assemble_current_states(
-    cna_clients: Mapping[str, CNAClient],
+    mocker,
     listed_assets: Iterable[Mapping[str, Any]],
     expected_state: State,
 ):
-    cna_clients["test"].list_assets.side_effect = [listed_assets]  # type: ignore
-    integration = CNAIntegration(cna_clients=cna_clients, namespaces=[])
+    mocker.patch.object(
+        CNAClient, "list_assets", create_autospec=True, return_value=listed_assets
```
steveteahan, line 161: `create_autospec` is a method whereas `autospec` should be the option to enabling autospeccing for the object created by the patch. I'm actually a bit surprised that no errors are thrown by this unless I'm missing something?

I confirmed with `patch` that `create_autospec` doesn't do what we expected, while `autospec` does:

```python
>>> from unittest.mock import patch
>>> def test(a):
...     print(a)
... 

# In this test too many arguments are passed, which should result in an error, but it works just fine
>>> with patch('__main__.test', return_value="test", create_autospec=True) as patch_test:
...     print(test('too', 'many', 'args'))
... 
test

# With autospec=True we get the expected TypeError
>>> with patch('__main__.test', return_value="test", autospec=True) as patch_test:
...     print(test('too', 'many', 'args'))
... 
  File "<stdin>", line 2, in <module>
  File "<string>", line 2, in test
  File "/usr/lib64/python3.9/unittest/mock.py", line 180, in checksig
    sig.bind(*args, **kwargs)
  File "/usr/lib64/python3.9/inspect.py", line 3043, in bind
    return self._bind(args, kwargs)
  File "/usr/lib64/python3.9/inspect.py", line 2964, in _bind
    raise TypeError('too many positional arguments') from None
TypeError: too many positional arguments
```

PR #3064, opened 2022-12-13
Add integration-specific metrics for gitlab-housekeeping

steveteahan:
These metrics should provide more data for guiding future enhancements related to speeding up gitlab-housekeeping, and improving the throughput of the integration.

I tested this manually as well by running the integration and ensuring that the metric calls returned proper data with the `collect()` method call. Any metrics that didn't run in dry-run mode were moved out of the conditional to make sure that they still work as expected.

Related: https://issues.redhat.com/browse/APPSRE-6653

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -15,6 +15,11 @@ from gitlab.v4.objects import (
     ProjectIssue,
     ProjectMergeRequest,
 )
+from prometheus_client import (
+    Counter,
+    Gauge,
+    Histogram,
+)
 from sretoolbox.utils import retry
 
 from reconcile import queries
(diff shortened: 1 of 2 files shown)


reconcile/gitlab_housekeeping.py
```diff
@@ -59,6 +64,40 @@
...
+)
+
+rebased_merge_requests = Counter(
+    name="qontract_reconcile_rebased_merge_requests",
+    documentation="Number of merge requests that have been successfully rebased in a repository",
+    labelnames=["project_id"],
+)
+
+time_to_merge = Histogram(
+    name="qontract_reconcile_time_to_merge_merge_request",
```
geoberle, line 80: Would you consider adding the unit  as a metric suffix?

steveteahan: Good catch! I've added it.

bhushanthakur93 (approved):
LGTM

PR #3066, opened 2022-12-14
Skip extra GitLab labels API calls in gitlab-housekeeping

steveteahan:
Related: https://issues.redhat.com/browse/APPSRE-6653

Also, related to draft PR #3050 which introduced multiple ideas for improving the performance of gitlab-housekeeping, that are now being addressed as separate MRs.

After a week of data, we only found 3 instances where labels were empty, but after calling the GitLab API again, labels were returned. All of these cases appear to be related to timing because the log entries matched when the first label was added to the MR, to the minute. In these cases, `gitlab-housekeeping` would have ignored these MRs for the first run, and then found them on the next run.

The performance improvements are 10-30 second reduction is run time of the integration (close to 20% reduction) and 80 fewer API calls to GitLab (>10% reduction). These performance improvements would be even clearer for cases where there are a larger number of MRs because the calls and runtime effectively grow linearly with the number of MRs.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -202,18 +202,18 @@ def is_rebased(mr, gl: GitLabApi) -> bool:
 
 
 def get_labels(mr: ProjectMergeRequest, gl: GitLabApi) -> list[str]:
-    labels = mr.attributes.get("labels")
-    if not labels:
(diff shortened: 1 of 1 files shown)


steveteahan:
Thank you, @geoberle! I'll merge this in a day or so after we get some initial data from #3064.

PR #3068, opened 2022-12-15
Remove broken time-to-merge metric in gitlab-housekeeping

steveteahan:
It looks like the exception wasn't being bubbled up properly, but data was trying to be accessed that didn't exist. I'll look to replace this shortly with some improvements to type hints to catch this in the future.


steveteahan:
Self-merging given the very limited scope here and not wanting this metric issue to affect the integration.

PR #3071, opened 2022-12-15
Add channel and initialversion

janboll:
Adds missing query fields, right now ocm-update-recommended-version creates a new MR all the time, cause the fields are not set and thus it shows a diff.


kwilczynski (approved):
Looks good!

geoberle (approved):

PR #3072, opened 2022-12-15
Improving logging in cases where gitlab-housekeeping is handling exce…

steveteahan:
…ptions

With these changes we'll start to log any Exceptions that are being caught and retried with the intent of eventually making them more specific so that we aren't retrying when it doesn't make sense to. This logging will also make it more obvious what is happening in the cases where the integration is running longer than usual.


OliviaHY:
LGTM

rporres (approved):

PR #3074, opened 2022-12-16
utils/terraform: safe_resource_id adds an underscore at the front if resource name begins with a number

jfchevrette:
Ref.: APPSRE-6585

Terraform does not allow a resource name to start with a number

This ensures that when using `utils.terraform.safe_resource_id` the resulting resource doesn't start with a number


kwilczynski (approved):
Looks good! Thank you!

PR #3077, opened 2022-12-19
use skupper to access internal integrations prometheus

patjlm:
Grafana is on a public cluster while integrations run on an internal cluster. We can use skupper to access metrics.


chassing (approved):
LGTM

PR #3078, opened 2022-12-19
Fix KeyError in ocm-update-recommended-version

janboll:


reconcile/ocm_update_recommended_version.py
```diff
@@ -75,12 +75,13 @@ def get_updated_recommended_versions(
...
         cluster_name = uc["name"]
-        for workload in uc["upgradePolicy"]["workloads"]:
-            channel_workload = (workload, cluster[cluster_name].spec.channel)
-            channel_workload_versions.setdefault(channel_workload, [])
-            channel_workload_versions[channel_workload].append(
-                cluster[cluster_name].spec.version
-            )
+        if cluster_name in cluster:
+            for workload in uc["upgradePolicy"]["workloads"]:
+                channel_workload = (workload, cluster[cluster_name].spec.channel)
```
geoberle, line 80: if i read the stack trace correctly, the `KeyError` was already on the `cluster[cluster_name]`

janboll: Let me take a look. Applying this remove the stacktrace locally though.

janboll: Now I see the problem. Line 78 was added, the rest was indented and did not change

geoberle: ah sry 🤦 
thanks for the fix

geoberle (approved):

PR #2761, opened 2022-09-04
[slack] add support to create user groups

maorfr:
https://coreos.slack.com/archives/C03KQ9M1P6C/p1662138005941939

ref: https://slack.com/help/articles/212906697-Create-a-user-group

PR #3089, opened 2022-12-22
[slack-usergroups] do not fail for non-existing usergroups

chassing:
APPSRE-6744


maorfr:
the desired behavior is that a slack usergroup will be created. see #2761

maorfr (approved):

chassing:
A non-existing slack usergroup will be created, see https://github.com/app-sre/qontract-reconcile/blob/6c1f05bdd3bf74c20c0062261efe3d5c9b286670/reconcile/slack_usergroups.py#L387. It was really just a bad `State` model attribute.

PR #3092, opened 2022-12-23
skupper qontract-reconcile integration

chassing:
[Skupper](https://skupper.io) qontract-reconcile integration

Features:
* Install skupper components (*skupper-site-controller*) into namespaces
* Create skupper config (*skupper-site* configmap)
* Create skupper sites connections via connection token exchange
* Don't allow skupper config changes after the skupper sites have been deployed. This is a current [skupper limitation](https://github.com/skupperproject/skupper/issues/58)

Qontract-schemas: https://github.com/app-sre/qontract-schemas/pull/361
Ticket: [APPSRE-6515](https://issues.redhat.com/browse/APPSRE-6515)
Design doc: [skupper-integration](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/skupper-integration.md)
Example app-interface-dev-data: https://gitlab.cee.redhat.com/app-sre/app-interface-dev-data/-/merge_requests/29

See it in action:

`qd profile create skupper-network --integration-name skupper-network --app-interface-pr 29 --qontract-schemas-pr 361 --qontract-reconcile-pr 3092`


reconcile/gql_definitions/skupper_network/skupper_networks.gql
```diff
@@ -0,0 +1,98 @@
...
+          routerCpuLimit
+          routerCpu
+          routerMemoryLimit
+          routerMemory
+          routerLogging
+          routerPodAntiaffinity
+          routerServiceAnnotations
+          routers
+          serviceController
+          serviceSync
```
fishi0x01, line 60: Would it make sense to have a fragment for `SkupperConfig` this in this folder like `skupper_network/skupper_config.gql`?

reconcile/skupper_network/models.py
```diff
@@ -0,0 +1,291 @@
...
+    ClusterPeeringConnectionClusterRequesterV1,
+    ClusterV1,
+    NamespaceV1,
+    SkupperSiteConfigDefaultsV1,
+    SkupperSiteConfigV1,
+)
+
+
+class Defaults:
+    """Define default values for a SkupperSite model."""
```
fishi0x01, line 27: Why do we query `siteConfigDefaults` from gql server if they are defined here? I see they are all merged together in `init()`, but I'm not sure I understand why we need to define defaults in 2 places?

fishi0x01 (commented):
Nice work on Skupper! I need more time to get through this - for now I just got 2 small comments/questions. :sweat_smile:

reconcile/gql_definitions/skupper_network/skupper_networks.gql
chassing: There are 2 skupper configs `SkupperSiteConfigDefaults_v1` for the skupper network, which defines the defaults for all sites, and `SkupperSiteConfig_v1`, which is basically the same, except for `skupperSiteController`, to override the defaults from the network. I don't know if that can be done via a fragment.

fishi0x01: Ah I understand. Makes sense :+1:

reconcile/skupper_network/models.py
chassing: Via `siteConfigDefaults`, you can define defaults used by all skupper sites, e.g., `routers: 1`, if not set via `siteConfigDefaults` or via skupper config in the namespace the `models.py Default` is being used.

reconcile/skupper_network/reconciler.py
```diff
@@ -0,0 +1,256 @@
...
+            # no connection token found. get the token from the connected site and import it
+            connected_site_oc = oc_map.get_cluster(connected_site.cluster.name)
+            if not connected_site_oc.project_exists(connected_site.namespace.name):
+                logging.info(
+                    f"{connected_site}: Namespace does not exist (yet)."
+                    " Skipping this skupper site for now!"
+                )
+                continue
+
+            if token := _get_token(
```
rporres, line 169: can you add a comment describing a bit what's happening here?

reconcile/skupper_network/site_controller.py
```diff
@@ -0,0 +1,179 @@
...
+                        ),
+                    ],
+                ),
+            ),
+        ),
+    )
+
+
+def site_controller_service_account(
+    site: SkupperSite,
```
rporres, line 86: do you need the `site` parameter here?

The same applies to a few methods in this module...

reconcile/test/skupper_network/test_skupper_network_models.py
```diff
@@ -0,0 +1,297 @@
...
+    network_config: SkupperSiteConfigDefaultsV1, site_config: SkupperSiteConfigV1
+) -> SkupperSiteFactory:
+    """Skupper site fixture."""
+
+    def _skupper_site_factory(
+        ns: NamespaceV1, edge: bool = False, delete: bool = False
+    ) -> SkupperSite:
+        return SkupperSite(
+            namespace=ns,
+            skupper_site_controller="just-a-image",
```
rporres, line 99: nit picking FTW 😂 

```suggestion
            skupper_site_controller="just-an-image",
```

rporres (commented):
I've made a first pass on this. It looks great. Very clear, the use of dataclasses makes things extremely understandable. Kudos!

I'll read it again tomorrow to go a bit deeper. I've left a few comments, questions. For the moment, my main worry is manifests maintenance.

reconcile/skupper_network/integration.py
```diff
@@ -0,0 +1,282 @@
...
+    return data.skupper_networks or []
+
+
+@defer
+def run(
+    dry_run: bool,
+    thread_pool_size: int = 10,
+    internal: Optional[bool] = None,
+    use_jump_host: bool = True,
+    defer: Optional[Callable] = None,
```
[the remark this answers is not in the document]
chassing: I'm using the `@defer` decorator, which injects the `defer` argument. Looking at other integration, the usage of this decorator seems like a common pattern.

reconcile/skupper_network/integration.py
```diff
@@ -0,0 +1,282 @@
...
+        for ns in skupper_network.namespaces:
+            if not ns.skupper_site:
+                # make mypy happy
+                continue
+
+            if not integration_is_enabled(QONTRACT_INTEGRATION, ns.cluster):
+                # integration is disabled for this cluster
+                continue
+
+            edge = ns.cluster.internal if ns.cluster.internal else False
```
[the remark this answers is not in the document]
chassing: That is a good point; it was always on my mind. Referring to the [design doc](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/skupper-integration.md#details), an internal cluster is behind the VPN, the question is, is it reachable by other internal clusters? Does it need a VPC peering? I still struggle with the `private` & `internal` attributes.

reconcile/skupper_network/models.py
```diff
@@ -0,0 +1,291 @@
...
+                and not other.is_edge_site
+                and not other.delete
+                and not other.is_connected_to(self)
+            ]
+
+        self.connected_sites = set(connected_sites)
+        logging.debug(f"{self} connected sites: {self.connected_sites}")
+
+    def unique_token_name(self, other: SkupperSite) -> str:
+        """Generate a unique tokan name for a site connection."""
```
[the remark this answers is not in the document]
chassing: Good catch 🙇

reconcile/test/skupper_network/test_skupper_network_models.py
chassing: 😂

reconcile/skupper_network/site_controller.py
```diff
@@ -0,0 +1,179 @@
...
+            dict(
+                kind="ServiceAccount",
+                name="skupper-site-controller",
+            ),
+        ],
+    )
+
+
+def is_usable_connection_token(secret: dict[str, Any]) -> bool:
+    """Check if secret is a finished connection token, not a token-reqeust anymore."""
```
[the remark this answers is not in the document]
chassing: Another good catch 🙇

reconcile/skupper_network/site_controller.py
chassing: You're right; it isn't needed.

reconcile/skupper_network/reconciler.py
chassing: 👍

reconcile/skupper_network/integration.py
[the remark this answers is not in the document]
rporres: hmmm, appsrep05ue1 is reachable from appsres05ue1 (and viceversa). I see no real reason to create a skupper link between them, though 😂

reconcile/skupper_network/site_controller.py
```diff
@@ -0,0 +1,179 @@
...
+        ),
+        subjects=[
+            dict(
+                kind="ServiceAccount",
+                name="skupper-site-controller",
+            ),
+        ],
+    )
+
+
```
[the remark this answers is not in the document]
chassing: It won't be hard to adapt to future manifest changes; the skupper version must be specified by the tenant for each skupper network. Based on that, the integration can decide which manifest must be deployed. I'll change the `site-controller` and prepare for that.

reconcile/skupper_network/integration.py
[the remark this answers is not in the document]
chassing: @patjlm, any comment on that? Should an `internal` cluster always be an edge site?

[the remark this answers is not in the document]
patjlm: I would say that the more links we have the better, to obtain several routing opportunities. So `private` could connect to each other.
`internal` could be more debatable for security concerns, but since the traffic is bidirectional once a link is up, I don't really see the difference. So I'd agree a site on an internal cluster is not necessarily an edge. Maybe we should compute if there are any incoming link to know if we have an edge ?
Links should be created *only* between namespaces that are configured for that skupper network. If appsrep05ue1 and appsres03ue1 namespaces are not configured to have namespaces in the same network, they will not be linked. We could still have the need though, for prometheus access for example. So it is technically possible.

reconcile/skupper_network/site_controller.py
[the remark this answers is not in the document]
rporres: I like it!

reconcile/skupper_network/integration.py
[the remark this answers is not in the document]
rporres: It is an unconsistent pattern at most 😂 I honestly don't know what it brings exactly, maybe others can chime in.

reconcile/skupper_network/integration.py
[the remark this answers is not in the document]
rporres: That sounds reasonable to me.

reconcile/skupper_network/integration.py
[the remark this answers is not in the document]
chassing: The most common use case is to call `oc_map.cleanup`:
```
$ grep -r 'defer(oc_map.cleanup)' reconcile | wc -l
22
```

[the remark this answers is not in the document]
rporres: indeed. Anyway, not a blocker, just curious. I don't know from where that `defer` argument is called, that's mainly it...

reconcile/skupper_network/integration.py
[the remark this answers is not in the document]
chassing: I've changed the connection algorithm; internal clusters aren't edge sites by default anymore.

rporres (approved):
This looks great to me. It may be interesting to have more views on it.

chassing:
@geoberle, please take a look 🙇

reconcile/skupper_network/models.py
```diff
@@ -0,0 +1,291 @@
...
+        data = {}
+        for k, v in self.dict(by_alias=True).items():
+            if isinstance(v, bool):
+                # True -> "true", False -> "false"
+                v = str(v).lower()
+            data[k] = str(v)
+        return data
+
+
+class SkupperSite(BaseModel):
```
geoberle, line 150: i love this class. it holds some powerful functions to make reading the rest of the code a breeze

reconcile/skupper_network/reconciler.py
```diff
@@ -0,0 +1,269 @@
...
+            }
+        )
+        # delete everything else that starts with 'skupper-'
+        to_delete.update(
+            {
+                f'{item["kind"]}-{item["metadata"]["name"]}': item
+                for item in oc.get_items(kind=kind, namespace=site.namespace.name)
+                if item["metadata"]["name"].startswith("skupper-")
+            }
+        )
```
geoberle, line 45: what kind of objects are these? objects created by the deployed skupper?

reconcile/skupper_network/reconciler.py
```diff
@@ -0,0 +1,269 @@
...
+    Connection sites together algorithm:
+    1. Check if all related sites (namespaces) are available
+    2. Iterate over all "to be connected sites" (connected_sites) and
+    2.1. if we are already connected (connection-token exists) -> done
+    2.2. if not connected (connection-token doesn't exist) -> token exchange alghorithm
+
+    Token exchange alghorithm:
+    1. Create a connection-token-request (secret with a unique token name) in the connected site's namespace
+    2. Wait for the site controller to process the connection-token-request
+    3. Move the connection-token (the processed connection-token-request secret) from the connected site's namespace into the site's namespace
```
geoberle, line 160: ❤️

reconcile/skupper_network/reconciler.py
chassing: The `site-controller` creates, for example, secrets (certs, keys) used by the `router`.

reconcile/skupper_network/integration.py
```diff
@@ -0,0 +1,264 @@
...
+        thread_pool_size=thread_pool_size,
+        skupper_sites=skupper_sites,
+        integration_managed_kinds=integration_managed_kinds,
+    )
+
+
+def early_exit_desired_state(*args: Any, **kwargs: Any) -> dict[str, Any]:
+    gqlapi = gql.get_api()
+    return {
+        "skupper_networks": [site.dict() for site in get_skupper_networks(gqlapi.query)]
```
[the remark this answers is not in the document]
patjlm: the private/public may be in the site dict i suppose..

reconcile/skupper_network/reconciler.py
```diff
@@ -0,0 +1,269 @@
...
+        to_delete.update(
+            {
+                f'{item["kind"]}-{item["metadata"]["name"]}': item
+                for item in oc.get_items(kind=kind, namespace=site.namespace.name)
+                if item["metadata"]["name"].startswith("skupper-")
+            }
+        )
+
+    for item in to_delete.values():
+        logging.info(
```
patjlm, line 48: should we also revoke the site tokens (outgoing to other sites) ?

reconcile/skupper_network/reconciler.py
```diff
@@ -0,0 +1,269 @@
...
+    """Delete any other connection tokens that are no longer needed."""
+    oc = oc_map.get_cluster(site.cluster.name)
+    for item in oc.get_items(
+        kind="Secret",
+        namespace=site.namespace.name,
+        labels=site.token_labels,
+    ):
+        if item["metadata"]["name"] not in [
+            site.token_name(connected_site) for connected_site in site.connected_sites
+        ]:
```
patjlm, line 218: those token should be revoked on the remote site as well, right ?

reconcile/skupper_network/integration.py
```diff
@@ -0,0 +1,264 @@
...
+    skupper_sites: Iterable[SkupperSite],
+    integration_managed_kinds: Iterable[str],
+) -> None:
+    """Realize all skupper resources and create skupper site connections."""
+    # skupper-site config map updates are not supported yet
+    # https://github.com/skupperproject/skupper/issues/58
+    # to change the skupper site configuration, the skupper site must be deleted and recreated
+    if skupper_site_config_changes(ri):
+        logging.error(
+            "skupper-site config changes detected. Please delete and recreate the affected skupper site(s)!"
```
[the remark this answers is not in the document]
patjlm: This log should also mention which network and which site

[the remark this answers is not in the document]
patjlm: > This log should also mention which network and which site
 Ah I see there is a log in the previous function, scratch this..

reconcile/skupper_network/integration.py
```diff
@@ -0,0 +1,264 @@
...
+        )
+        sys.exit(1)
```
patjlm, line 192: This means we will not reconcile other networks/sites. Can we just skip the changed sites ?

reconcile/skupper_network/reconciler.py
geoberle: it would feel a tiny bit more comfortable if you we could check that those resources don't have `qontract` annotations on them, so that we don't accidentally delete things just because of naming.

chassing: Good point! I'll add a check for that.

reconcile/skupper_network/integration.py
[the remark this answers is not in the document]
chassing: Nice catch, I'll add that.

reconcile/skupper_network/reconciler.py
chassing: Obsolete site tokens are deleted by [delete_unused_tokens](https://github.com/app-sre/qontract-reconcile/blob/ae85b58f16cf564b3c0a1bef80e9f0550593d7d3/reconcile/skupper_network/reconciler.py#L208). The `delete_skupper_site` deletes everything else that is not managed by the integration, e.g., the internal certificates and keys created by the `site-controller`.

reconcile/skupper_network/reconciler.py
chassing: This removes all tokens from sites where the current site should not be connected anymore, e.g., the site doesn't exist anymore. Such kind of token doesn't exist on the source site anymore, because it has been delete during the token transfer. See [_transfer_token](https://github.com/app-sre/qontract-reconcile/blob/ae85b58f16cf564b3c0a1bef80e9f0550593d7d3/reconcile/skupper_network/reconciler.py#L115)

reconcile/skupper_network/integration.py
chassing: Unfortunately, it's not possible. I don't have the skupper site information anymore; here I'm comparing just configmap in the `ResourceInventory`. Nevertheless, this section should be reached during a PR check only and should not affect the running integration.

reconcile/skupper_network/reconciler.py
patjlm: but does it revoke the tokens as well ?

reconcile/skupper_network/integration.py
patjlm: fair enough!

reconcile/skupper_network/reconciler.py
chassing: I don't get it :( How do I revoke a connection-token? Especially if the source site doesn't exist anymore.

reconcile/skupper_network/integration.py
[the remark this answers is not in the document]
chassing: TBH, I don't have a good solution for that at the moment. For sure, an automated delete and recreate is possible, but I don't want to do that without any further tenant notification. My hope is that the skupper site-controller can handle that in the future.

geoberle (approved):

reconcile/skupper_network/reconciler.py
patjlm: i may be misunderstanding the term "source site". Any token deletion should go with a token revocation on the site that generated it. If that site is deleted, it's fine of course, but if the deletion occurs on a connecting site, we need a revoke somehow, right ?

reconcile/skupper_network/integration.py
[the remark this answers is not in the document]
patjlm: fair enough

reconcile/skupper_network/reconciler.py
chassing: I don't think it's possible to revoke a single connection token. The `skupper` CLI has a `revoke-access` subcommand that revokes all previously issued tokens:

```
$ skupper revoke-access --help
This will invalidate all previously issued tokens and require that all
links to this site be re-established with new tokens.
```

The [code](https://github.com/skupperproject/skupper/blob/master/client/revoke_all.go#L124) behind that regenerates the site CA and all site certificates.

patjlm (approved):
Thanks a lot for all this!!!

PR #3094, opened 2023-01-02
new qenerate version - no changes in generated classes

fishi0x01:
Synch qontract-reconcile with latest qenerate release. There are no changes to generated classes, so this will not have any special effect.


andreugallofre (approved):
LGTM

PR #3095, opened 2023-01-03
APPSRE-6609 proper read_all args for secret_reader

fishi0x01:
Vault's `read_all()` only uses `path` and `version`, i.e., we can omit `format` and `field` from the function signatures.


fishi0x01:
Skipping this for now - might introduce more confusion than benefit actually, as the Protocol has field as mandatory setting

PR #3098, opened 2023-01-05
Error handling for saasherder

OliviaHY:
### Why:
The openshift-saas-deploy-trigger-moving-commits integration swallows GitHub and GitLab exception which hides the problem. We don't want to simply raise it neither because it will block the whole shard when there is problem with one or a few repo.
### What:
* Add error handling for saasherder's GitHub and GitLab related error so the integration will failed but not blocked.
* Making one existing method private to pass lint
### Validation:
Tested with local App Interface pointing to my own private repo added as target, before and after personal access token restricted.


reconcile/utils/saasherder.py
```diff
@@ -208,6 +209,12 @@ def __init__(
         if accounts:
             self._initiate_state(accounts)
 
+    def _register_error(self):
+        self.error_registered = True
+
+    def has_error_registered(self):
```
chassing, line 215: `@property` is missing

BumbleFeng (approved):

PR #3099, opened 2023-01-05
[ocm-addons-upgrade-tests-trigger] introduce new integration

maorfr:
https://issues.redhat.com/browse/APPSRE-6756

depends on https://github.com/app-sre/qontract-schemas/pull/363

this integration will trigger jobs in jenkins following an upgrade of an addon for all clusters within an OCM org.

### DIFF
--- a/reconcile/ocm_addons_upgrade_tests_trigger.py
+++ b/reconcile/ocm_addons_upgrade_tests_trigger.py
@@ -1,5 +1,6 @@
 import logging
 from typing import Optional
+
 from reconcile import queries
 from reconcile.utils.jenkins_api import JenkinsApi
 from reconcile.utils.ocm import OCMMap
--- a/reconcile/ocm_addons_upgrade_tests_trigger.py
+++ b/reconcile/ocm_addons_upgrade_tests_trigger.py
@@ -1,13 +1,24 @@
+import logging
+from typing import Optional
 from reconcile import queries
+from reconcile.utils.jenkins_api import JenkinsApi
 from reconcile.utils.ocm import OCMMap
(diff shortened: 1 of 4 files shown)


chassing (approved):
Quick & dirty :heart: LGTM

OliviaHY:
LGTM

reconcile/ocm_addons_upgrade_tests_trigger.py
```diff
@@ -0,0 +1,92 @@
...
+            init_addons=True,
+        )
+
+        ocm = ocm_map[ocm_name]
+        state_updates: dict[str, Optional[str]] = {}
+        for aut in addon_upgrade_tests:
+            addon_name = aut["addon"]["name"]
+            addon_org_version = None
+            should_trigger = True
+            state_key = f"{ocm_name}/{addon_name}"
```
rporres, line 40: you may want to take into account @patjlm comment about the state key in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/55019#note_5540116

rporres (commented):
Looks fine for the purpose. I have a doubt about the state key depending on how long we want this to live...

patjlm (approved):
LGTM. We might need to include sector info in the state, group clusters by sector and trigger job per sector (passing that sector info as a parameter). I agree it can be done later on, but we may have some time of unclear water once sectors are implemented in OCM (the job may need to be triggered manually until we sync sector info down to the state and maybe app-interface?).

maorfr:
the state key can be changed at any point in the future. this would mean that someone could run the integration locally with the trigger part commented out, thus only updating the state.

or, we could just let the jobs be triggered...

PR #3101, opened 2023-01-09
[change-owners] reusable change-types through composition

geoberle:
Add a new change provider named change-type to /app-interface/change-type-1.yml#changes.provider that can reference an existing change-type and put it into the context of changes.context.

```yaml
- provider: change-type         <-- new provider
  changeTypes:
  - $ref: namespace-owner.yml   <-- puts the existing change-type ...
  context:                      <-- ... into a new context
    selector: app.'$ref'
```

This makes the referenced change-type usable in the ownership context of the referencing change-type. It redurced ownership management toil and makes the change-types reusable. It also enables the definition of higher level concepts for ownership, e.g. the owned entity is an app and by putting existing change-types into the context of the app change-type, ownership is dynamically expanded to other entities.

Ref: https://issues.redhat.com/browse/APPSRE-6651
schema-change: https://github.com/app-sre/qontract-schemas/pull/365

### DIFF
--- a/reconcile/change_owners/bundle.py
+++ b/reconcile/change_owners/bundle.py
@@ -1,6 +1,14 @@
+from abc import abstractmethod
 from dataclasses import dataclass
 from enum import Enum
-from typing import Optional
+from typing import (
+    Any,
+    Optional,
+    Protocol,
+    Tuple,
+)
+
+from reconcile.utils.gql import get_diff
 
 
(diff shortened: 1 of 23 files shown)


geoberle:
too hard to review. i will split it into smaller commits for pure refactoring without logic change followed by commits that introduce the new feature

PR #3103, opened 2023-01-09
[github] don't examine non github usernames

chassing:
Don't try to add non-GitHub usernames if another OIDC authentication provider is configured

APPSRE-6772

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -1018,7 +1018,9 @@ def aggregate_shared_resources(namespace_info, shared_resources_type):
 
 
 def determine_user_keys_for_access(
-    cluster_name: str, auth_list: Sequence[Union[dict[str, str], HasService]]
+    cluster_name: str,
+    auth_list: Sequence[Union[dict[str, str], HasService]],
+    enforced_user_keys: Optional[list[str]] = None,
 ) -> list[str]:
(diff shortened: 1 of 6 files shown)


reconcile/openshift_base.py
```diff
@@ -1018,7 +1018,9 @@ def aggregate_shared_resources(namespace_info, shared_resources_type):
 
 
 def determine_user_keys_for_access(
-    cluster_name: str, auth_list: Sequence[Union[dict[str, str], HasService]]
+    cluster_name: str,
+    auth_list: Sequence[Union[dict[str, str], HasService]],
+    enforced_user_keys: Optional[list[str]] = None,
```
fishi0x01, line 1023: Nit: `list[str] = []` to avoid `None`. Just my personal preference - totally fine to neglect. Nothing mandatory at all :)

fishi0x01 (approved):
LGTM

reconcile/openshift_base.py
chassing: In python, you shouldn't use `variable = []` to have an empty default list. See [https://medium.com/techtofreedom/mind-the-default-arguments-in-python-functions-8b00d6f532e3](Mind the Default Arguments in Python Functions) for a deeper explanation

PR #3107, opened 2023-01-09
[glitchtip] add type hints to glitchtip tests

chassing:
APPSRE-6750

### DIFF
--- a/reconcile/test/glitchtip/test_glitchtip.py
+++ b/reconcile/test/glitchtip/test_glitchtip.py
@@ -1,5 +1,3 @@
-from typing import Optional
-
 from pytest_mock import MockerFixture
 
 from reconcile.glitchtip.integration import (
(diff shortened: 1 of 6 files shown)


reconcile/test/glitchtip/test_glitchtip.py
```diff
@@ -26,8 +31,10 @@ def sort_all(orgs: list[Organization]) -> list[Organization]:
 
 
 def test_fetch_current_state(
-    glitchtip_client: GlitchtipClient, glitchtip_server_full_api_response, fx
-):
+    glitchtip_client: GlitchtipClient,
+    glitchtip_server_full_api_response: Optional[None],
```
fishi0x01, line 35: This one looks odd :thinking:

chassing: I've updated it to `None` only, nevertheless the pytest fixture `glitchtip_server_full_api_response` returns `None`, so it's correct :)

fishi0x01 (approved):

PR #3108, opened 2023-01-09
adds insights service account oauth scope to ocm resource server

tonytheleg:
FedRAMP Insights initiative needs to leverage AMS and a new oauth scope which does not currently exist. This PR is to add the scope to Cognito in FedRAMP so a service account can be created with this new scope.

@aleccohan can address any questions for more details as the originator of the request


bkez322 (approved):
LGTM. Thanks for making this.

PR #3112, opened 2023-01-10
Remove redundant published version data

fishi0x01:
I had a closer look at ocm.py and might have spotted an issue. The OCM query looks like this:

```graphql
    ocm {
      name
      url
      accessTokenClientId
      accessTokenUrl
      accessTokenClientSecret {
        ... VaultSecret
      }
      blockedVersions
      inheritVersionData {
        name
        publishVersionData {
          name
        }
      }
      sectors {
        name
        dependencies {
          name
          ocm {
            name
          }
        }
      }
    }
```

I.e., `publishedVersionData` is part of `inheritVersionData`. Running this against latest app-interface shows that we do not have any occurrence of inheritVersionData, which might explain why this didnt produce any errors yet.

### DIFF
--- a/reconcile/utils/ocm.py
+++ b/reconcile/utils/ocm.py
@@ -1733,6 +1733,8 @@ class OCMMap:  # pylint: disable=too-many-public-methods
             name = ocm_info["name"]
             secret_reader = SecretReader(settings=self.settings)
             token = secret_reader.read(access_token_client_secret)
+            inherit_version_data = ocm_info.get("inheritVersionData")
+            publish_version_data = inherit_version_data.get("publishVersionData")
             self.ocm_map[ocm_name] = OCM(
                 name,
                 url,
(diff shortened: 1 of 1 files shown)


patjlm:
Had a second look. I don't that fix is needed. `publishedVersionData` will always be an empty list, indeed, but that's actually not used today. What we could do is remove the `publishedVersionData` field from `OCM` since it is actually never needed.
`OCM.inheritVersionData` contains extactly what graphql returns, and that's what is checked [here](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/ocm_upgrade_scheduler.py#L150-L161)

fishi0x01:
@patjlm Makes sense. I removed now `publishedVersionData` from `OCM` as you suggested.

patjlm (approved):
Thanks!

PR #3113, opened 2023-01-10
VaultReplication add policy provider

andreugallofre:
Create a new provider for VaultReplication to support replication from a Policy.
Schemas: [Schemas PR](https://github.com/app-sre/qontract-schemas/pull/366)

APPSRE-6778

### DIFF
--- a/reconcile/utils/vault.py
+++ b/reconcile/utils/vault.py
@@ -313,25 +313,32 @@ class _VaultClient:
             msg = f"permission denied accessing secret '{path}'"
             raise SecretAccessForbidden(msg)
 
-    def _list(self, path: str) -> dict:
+    def _list_kv2(self, path: str) -> dict:
         try:
-            response = self._client.list(path)
-            if response is not None:
-                return response
-            # If vault path is from a v2 engine hvac.list(path) returns None,
-            # to list secrets from a v2 engine we need to use hvac.secrets.kv.v2.list_secrets
-            secret_path = path.split("/", 1)
+            mount_point, secret_path = path.split("/", 1)
             response = self._client.secrets.kv.v2.list_secrets(
-                mount_point=secret_path[0], path=secret_path[1]
+                mount_point=mount_point, path=secret_path
             )
             return response
         except hvac.exceptions.Forbidden:
             msg = f"permission denied accessing path '{path}'"
             raise PathAccessForbidden(msg)
 
+    def _list(self, path: str) -> dict:
+        try:
+            return self._client.list(path)
+        except hvac.exceptions.Forbidden:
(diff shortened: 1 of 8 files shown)


reconcile/utils/vault.py
```diff
@@ -315,7 +315,16 @@ def _write_v1(self, path, data):
 
     def _list(self, path: str) -> dict:
         try:
-            return self._client.list(path)
+            response = self._client.list(path)
+            if response is not None:
+                return response
+            # If vault path is from a v2 engine hvac.list(path) returns None,
+            # to list secrets from a v2 engine we need to use hvac.secrets.kv.v2.list_secrets
```
rporres, line 322: I'm having a hard time finding this feature documented, so I'm not sure we can rely on it.

If you need different methods to list, why don't you first find which is the version of the secret engine and then use the relevant engine methods instead?

reconcile/utils/vault.py
```diff
@@ -315,7 +315,16 @@ def _write_v1(self, path, data):
...
+            secret_path = path.split("/", 1)
```
rporres, line 323: suggestion: I think this will be a bit clearer:

```suggestion
            mount_point, secret_path = path.split("/", 1)
```

reconcile/utils/vault.py
andreugallofre: Documentation about the v2 method is here https://hvac.readthedocs.io/en/latest/usage/secrets_engines/kv_v2.html#list-secrets 

I can try to adapt the code with two different reading methods checking!

reconcile/utils/vault.py
andreugallofre: 👍

reconcile/utils/vault.py
rporres: I meant the main `list` behaviour... I guess that if it tries to use the v1 methods against a v2 secret it returns `None` , but I'd rather try to be more explicit here.

reconcile/vault_replication.py
```diff
@@ -325,6 +344,19 @@ def replicate_paths(
             for vault_path in path_list:
                 copy_vault_secret(dry_run, source_vault, dest_vault, vault_path)
 
+        elif isinstance(path, VaultReplicationPolicyV1):
+            vault_query_data = vault_policies.query(query_func=gql.get_api().query)
+            if path.policy is None:
+                raise VaultInvalidPolicy
```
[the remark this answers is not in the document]
rporres: Thanks for the comment. I was meaning more an exception message so that it is easy to identify what's going on and even identify it in the code in case you reuse that exception.

rporres (approved):
just a nit. It looks good anyway.

PR #3117, opened 2023-01-11
cluster upgrade cross org sector dependencies

patjlm:
[APPSRE-6675](https://issues.redhat.com/browse/APPSRE-6675)

The first implementation of cluster upgrade sectors ([APPSRE-6545](https://issues.redhat.com/browse/APPSRE-6545)) allowed to reference sectors cross orgs. This implementation has several drawbacks:
- knowledge of sectors from one org to the other, which may not be the case/needed/wanted and makes changes to sector list more complex.
- complex code where a sector map needs to be global. It also forbids (or make complex) sharding of integrations that use this feature.

This PR changes this by relying on the `inheritVersionData` flag in the ocm org schema:

if `sector` is defined in the cluster upgradepolicy and the cluster org inherits from an other one, we will allow an upgrade to a version if either:
- all the cluster `workloads` exist on the inherited org, and they all run at least that version
- all `workloads` are not existing in the inherited org and all cluster in that org run at least that version

Review will be easier 1 commit at a time

### DIFF
--- a/reconcile/utils/cluster_version_data.py
+++ b/reconcile/utils/cluster_version_data.py
@@ -25,11 +25,25 @@ class VersionHistory(BaseModel):
 
 
 class Stats(BaseModel):
+    """Stats is a part of VersionData. It provides basic statistics on the OCM
+    organization current cluster versions. Currently only
+    the minimum version, globally in the org and per workload, is being stored.
+    This class also has a `inherited` field which will contain at runtime a
+    computation of the same statistics for `inheritedVersionData` organizations.
+    This field is only computed and set at runtime and not stored in state.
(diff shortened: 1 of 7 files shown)


reconcile/utils/cluster_version_data.py
```diff
@@ -23,12 +24,58 @@ class VersionHistory(BaseModel):
     workloads: dict[str, WorkloadHistory] = Field(default_factory=dict)
 
 
+class Stats(BaseModel):
```
maorfr, line 27: some documentation around this class will be super helpful

maorfr (approved):
LGTM

PR #3118, opened 2023-01-12
invalid pagerduty IDs

chassing:
pagerduty raises an exception for invalid resource_ids, slack-usergroups reconciles all user groups but exits with exit code 1 when an error occurs.

Ticket: [APPSRE-6621](https://issues.redhat.com/browse/APPSRE-6621)

### DIFF
--- a/reconcile/slack_usergroups.py
+++ b/reconcile/slack_usergroups.py
@@ -1,4 +1,5 @@
 import logging
+import sys
 from collections.abc import (
     Callable,
     Iterable,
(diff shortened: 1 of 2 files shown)


fishi0x01 (approved):
LGTM

rporres (approved):
Not a fan of globals, but I don't think there's a better way to hold that state without making bigger changes, so LGTM

chassing:
I'm also not a big fan of globals, but as you said, it would require a significant change to do it better.

geoberle (approved):
about the global. keep in mind that sharding becomes an issue as long as globals hold run state. since we only have one slack workspace right now, i don't see an imminent issue. so also LGTM from me

PR #3119, opened 2023-01-12
Support aws gov partition when assigning enhanced monitoring policy 

dwelch0:
APPSRE-6799


cubismod (approved):
Smashing that like button!

PR #3120, opened 2023-01-12
ocm-addons-upgrade-scheduler-org integration

patjlm:
https://issues.redhat.com/browse/APPSRE-6790

proposal to manage addon upgrade based on our existing upgrade policies. This is targeting hypershift only, as a short term solution

Depends on https://github.com/app-sre/qontract-schemas/pull/368

### DIFF
--- a/reconcile/ocm_upgrade_scheduler.py
+++ b/reconcile/ocm_upgrade_scheduler.py
@@ -1,9 +1,11 @@
+import copy
 import logging
 import sys
 from collections.abc import Mapping
 from datetime import datetime
 from typing import (
     Any,
+    Iterable,
     Optional,
 )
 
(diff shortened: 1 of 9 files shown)


reconcile/utils/ocm.py
```diff
@@ -1194,7 +1202,8 @@ def version_blocked(self, version):
         Returns:
             bool: is version blocked
         """
-        return any(re.search(b, version) for b in self.blocked_versions)
+        s = f"{addon_id}/{version}" if addon_id else version
```
patjlm, line 1205: Reusing `blockedVersions` with a pattern like `{addon_id}/{version-regex}` for addons

maorfr (approved):
LGTM

reconcile/ocm_upgrade_scheduler.py
```diff
@@ -69,14 +74,22 @@ def fetch_desired_state(
...
         sector_name = upgrade_policy["conditions"].get("sector")
         if sector_name:
             upgrade_policy["conditions"]["sector"] = ocm.sectors[sector_name]
-        desired_state.append(upgrade_policy)
+        if not addons:
+            spec = ocm.clusters[cluster_name].spec
+            upgrade_policy["current_version"] = spec.version
+            upgrade_policy["channel"] = spec.channel
+            desired_state.append(upgrade_policy)
+            continue
```
geoberle, line 86: forgive me that the only thing i have to say is a stupid nit: 
would you consider putting the `addons` code block into an `else` for better readability as you did it in `fetch_current_state`?

geoberle (approved):

reconcile/ocm_upgrade_scheduler.py
patjlm: Wanted to avoid longer lines :) but I'll do that

patjlm: done in [894376f](https://github.com/app-sre/qontract-reconcile/pull/3120/commits/894376fda09fa507556460399447b8ddb089f0c5)

PR #3126, opened 2023-01-16
qontract-cli output for ocm addon upgrade policies

patjlm:
https://issues.redhat.com/browse/APPSRE-6809

This is a bit more simple than clusters' output as addons can be upgraded to only one version. We can review later if more fancy display is deemed necessary (🎉 )

Example output:
https://gitlab.cee.redhat.com/-/snippets/6206

### DIFF
--- a/reconcile/ocm_addons_upgrade_scheduler_org.py
+++ b/reconcile/ocm_addons_upgrade_scheduler_org.py
@@ -5,6 +5,7 @@ import reconcile.ocm_upgrade_scheduler as ous
 from reconcile import queries
 from reconcile.utils.cluster_version_data import VersionData
 from reconcile.utils.ocm import OCMMap
+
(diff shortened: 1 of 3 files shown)


reconcile/ocm_addons_upgrade_scheduler_org.py
```diff
@@ -35,12 +36,25 @@ def calculate_diff(
...
+
+
+@dataclasses.dataclass
+class Result:
+    current_state: list[dict[str, Any]]
+    desired_state: list[dict[str, Any]]
+    diffs: list[dict[str, str]]
+
+
+def compute(
```
patjlm, line 53: The `current_state`, `desired_state` and `diffs` computation were basically extracted from the `run` function into this `compute` function. This allows to reuse it easily in the qontract-cli command.
The `compute` function returns a `Result` for each org/addon (= `ResultKey`)

maorfr (approved):
:heart:

PR #3129, opened 2023-01-17
fluentd read_from_head

patjlm:
Use [fluentd read_from_head](https://docs.fluentd.org/input/tail#read_from_head) to log all integration output from the start, even if fluentd takes more time than the integration to start.

This should avoid missing early logs on pod startup.

I will check on stage once merged if fluentd starts correctly.


rporres (approved):

cubismod (approved):
LGTM

PR #3130, opened 2023-01-17
[slack-usergroups] reset error flag at startup

chassing:
After an error occurred, the global `error_occurred` flag was still set. Reset it on each integration run to avoid that.

[APPSRE-6818](https://issues.redhat.com/browse/APPSRE-6818)

### DIFF
--- a/reconcile/slack_usergroups.py
+++ b/reconcile/slack_usergroups.py
@@ -733,6 +733,9 @@ def run(
     workspace_name: Optional[str] = None,
     usergroup_name: Optional[str] = None,
 ) -> None:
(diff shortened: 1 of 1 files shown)


geoberle (approved):
lgtm to mitigate the current error situation

can you open a ticket to get rid of globals in slack-usergroups?

cubismod (approved):

chassing:
Followup ticket https://issues.redhat.com/browse/APPSRE-6819

rporres (approved):
I agree with @geoberle 😄

PR #3131, opened 2023-01-17
[WIP] introduce jenkins-worker-fleets

BumbleFeng:


reconcile/test/test_jenkins_worker_fleets.py
```diff
@@ -0,0 +1,45 @@
+from unittest import TestCase
```
rporres, line 1: we have moved away from unittest. 

https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-testing-guide/#frameworks-and-tools

reconcile/test/test_jenkins_worker_fleets.py
```diff
@@ -0,0 +1,45 @@
...
+    act,
+    calculate_diff,
+    get_current_state,
+    get_desired_state,
+)
+from reconcile.test.fixtures import Fixtures
+from reconcile.utils.jenkins_api import JenkinsApi
+from reconcile.utils.terrascript_aws_client import TerrascriptClient as Terrascript
+
+fixture = Fixtures("jenkins_worker_fleets").get_anymarkup("api.yml")
```
rporres, line 14: I think it would be much clearer if we had different fixture files for different types of data:

* gql queries
* jcasc configs
* ...

To have one giant blob makes things more difficult to follow.

rporres (commented):
Functionality is greatly covered. Good stuff! 

Suggestions are more related to code style and a few nits here and there.

In general we're moving away from passing bare dicts across functions because that has made q-r code quite hard to maintain. Old code is harder to modify, but new code that has no dependencies does benefit greatly from avoiding using bare dicts for things that have an expected shape. That's where dataclasses (especially handled through pydantic) and qenerate come quite handy.

Let me know what you think about refactoring the code using latest best practices in q-r. I think we still have time to deliver this, so that should not be a problem.

reconcile/jenkins_worker_fleets.py
```diff
@@ -0,0 +1,223 @@
...
+                asg = {
+                    "name": identifier,
+                    "fleet": identifier,
+                    "region": region,
+                    "minSize": values.get("min_size"),
+                    "maxSize": values.get("max_size"),
+                    "computerConnector": {
+                        "sSHConnector": {
+                            "credentialsId": f["credentialsId"],
+                            "sshHostKeyVerificationStrategy": "nonVerifyingKeyVerificationStrategy",
```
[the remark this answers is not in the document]
BumbleFeng: We can add enable sshHostKeyVerification as the next iteration as long as the instance refresh. Since they all required acting on real ec2 instances.

reconcile/jenkins_worker_fleets.py
```diff
@@ -0,0 +1,223 @@
...
+                    values.get("region") or spec.provisioner["resourcesDefaultRegion"]
+                )
+                asg = {
+                    "name": identifier,
+                    "fleet": identifier,
+                    "region": region,
+                    "minSize": values.get("min_size"),
+                    "maxSize": values.get("max_size"),
+                    "computerConnector": {
+                        "sSHConnector": {
```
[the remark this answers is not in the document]
BumbleFeng: same defined by jcasc schema

reconcile/jenkins_worker_fleets.py
```diff
@@ -0,0 +1,186 @@
...
+}
+"""
+
+QONTRACT_INTEGRATION = "jenkins-worker-fleets"
+
+
+def get_current_state(jenkins: JenkinsApi) -> list[JenkinsWorkerFleet]:
+    current_state = []
+
+    jenkins_config = cast(dict[str, Any], jenkins.get_jcaas_config().get("jenkins"))
```
chassing, line 67: Is a `cast` here really needed? Try 

```
jenkins_config: dict[str, Any] = jenkins.get_jcaas_config().get("jenkins")
```

reconcile/jenkins/types.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class computerConnector(BaseModel):
+    sSHConnector: SSHConnector
+
+    def to_dict(self) -> dict[str, Any]:
+        return {"sSHConnector": self.sSHConnector.to_dict()}
+
+
+class JenkinsWorkerFleet(BaseModel):
```
chassing, line 39: You could use the [field alias](https://docs.pydantic.dev/usage/schema/#field-customization) pedantic feature to use snake_case variable names, e.g.,

```
min_spare_size: int = Field(0, alias="minSpareSize")
```

chassing (commented):
I like the dedicated `jenkins` module, instead of a plain python file ❤️ 
Please try to use *snake_case* everywhere; I hope the suggested `alias` will help.
I think a new integration should use GQL classes, instead of plain gql query strings, this makes the code so much more readable, and the type hints are helping a lot.

reconcile/jenkins_worker_fleets.py
```diff
@@ -0,0 +1,223 @@
...
+    return diffs
+
+
+def get_current_state(jenkins: JenkinsApi) -> list[dict[str, Any]]:
+    current_state = []
+
+    jenkins_config = cast(dict[str, Any], jenkins.get_jcaas_config().get("jenkins"))
+    clouds = cast(list[dict[str, Any]], jenkins_config.get("clouds", []))
+    for c in clouds:
+        fleet = c.get("eC2Fleet", None)
```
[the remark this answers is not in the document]
rporres: that's interesting. I don't understand where the difference comes, but since there's a schema, we should honour it.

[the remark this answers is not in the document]
rporres: If you can add a comment somewhere in the jenkins file where you explain this it would be great. It is a bit confusing, TBH

reconcile/jenkins_worker_fleets.py
BumbleFeng: seems not working. `reconcile/jenkins_worker_fleets.py:17: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "Dict[str, Any]")  [assignment]`

chassing: I see 👍

chassing (approved):
Thanks for changing to snake_case 🙇

reconcile/queries.py
```diff
@@ -320,15 +320,50 @@ def get_integrations(managed=False):
...
+      }
+    }
+    {% endif %}
   }
 }
 """
 
 
-def get_jenkins_instances():
+def get_jenkins_instances(work_fleets=False):
```
rporres, line 362: nit: shouldn't this be called `worker_fleets`?

```suggestion
def get_jenkins_instances(worker_fleets=False):
```

rporres (approved):
some nit picking. Looks great otherwise

PR #3134, opened 2023-01-18
integration runner refactor + sharded dry-runs

geoberle:
This PR refactors the current `cli.py` based integration startup function and enhances it with `sharded` dry-runs.

REVIEW COMMIT BY COMMIT

# Part 1 - Refactor
commit 1

The main logic for integration running has been moved to the `reconcile.utils.runtime` module. An integration and the required callback hooks `run` and `get_early_exit_desired_state` are now part of a class `QontractReconcileIntegration`. New integrations can subclass this one directly and implement the hook methods. Existing integrations are wrapped during startup into an `ModuleBasedQontractReconcileIntegration`. This way the code in `reconcile.utils.runtime.runner` deal only with `QontractReconcileIntegration` and can rely on a typed interface.

And integration can be executed by constructing a `IntegrationRunConfiguration` object which contains all the relevant generic qontract-reconcile runtime infos (`dry_run`, `validate_schema`, ...)

# Part 2 - Detect affected shards
commit 2+3

the difference in desired state state is used to find to find the affected shards for an integration. an integration can use the `get_desired_state_shard_config` function to return a `DesiredStateShardConfig` object that defines where to find the affected shards in the desired state.

the `DesiredStateShardConfig` object can contain a callback function where the integration can decide, if sharded run should be tried based on the detected shards, e.g. this can be used to not run in sharded mode if too many shards have been detected.

Sharded dry-runs are opt-int in many levels: 
* an integration must provide early-exit functionality and a `DesiredStateShardConfig` configuration
* an integration must be started with the `--check-only-affected-shards` CLI flag. This flag can be controlled via app-interface hack scripts (through data in the `/app-sre/integration-1.yml` [schema](https://github.com/app-sre/qontract-schemas/pull/373)) to activate/deactivate the functionality on demand

# Part 3 - Enable `terraform-resources` for sharded dry-runs

(!) this part will be split out into a dedicated PR - i left it here for now to show how the `desired_state_shard_config` hook works.

by implementing the `desired_state_shard_config` function, the `terraform-resources` integrations opts into the sharded integration dry-run mechanism.

the desired state returned by `early_exit_desired_state` has been enhanced by adding `IDENTIFIER_FIELD_NAME` fields to establish identifies for objects in lists, which speeds up the difference finding procedure. `IDENTIFIER_FIELD_NAME` is not something new but is populated in `bundles` by `qontract-validator` when a schema defines `uniqe` or `context-unique` fields.

this is not really necessary but helps in speeding up the diffing process so it does not run into timeouts, which would prevent sharded dry-runs.

Jira: https://issues.redhat.com/browse/APPSRE-6785
Schema change for activation flag: https://github.com/app-sre/qontract-schemas/pull/373

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -106,6 +106,20 @@ def early_exit(function):
     return function
 
 
+def check_only_affected_shards(function):
+    help_msg = (
+        "Execute a dry-run only for those integration shards where the "
+        "desired state changed. Works only when --early-exit-compare-sha is set"
+    )
+    function = click.option(
+        "--check-only-affected-shards",
+        default=False,
+        is_flag=True,
+        help=help_msg,
+    )(function)
+    return function
+
+
 def dry_run(function):
     help_msg = (
         "If `true`, it will only print the planned actions "
@@ -425,6 +439,10 @@ def register_faulthandler(fileobj=sys.__stderr__):
         )
 
 
+class UnknownIntegrationTypeError(Exception):
+    pass
+
+
 def run_integration(
     func_container: Union[ModuleType, QontractReconcileIntegration],
     ctx,
@@ -439,7 +457,9 @@ def run_integration(
         elif isinstance(func_container, ModuleType):
             integration = ModuleBasedQontractReconcileIntegration(func_container)
         else:
-            raise Exception(f"Unknown integration type {type(func_container)}")
+            raise UnknownIntegrationTypeError(
+                f"Unknown integration type {type(func_container)}"
+            )
 
         running_state = RunningState()
         running_state.integration = integration.name  # type: ignore[attr-defined]
@@ -449,12 +469,17 @@ def run_integration(
             logging.info("Integration toggle is disabled, skipping integration.")
             sys.exit(ExitCodes.SUCCESS)
 
+        check_only_affected_shards = (
+            ctx.get("check_only_affected_shards", False)
+            or os.environ.get("CHECK_ONLY_AFFECTED_SHARDS", "false") == "true"
+        )
         run_integration_cfg(
             IntegrationRunConfiguration(
                 integration=integration,
                 valdiate_schemas=ctx["validate_schemas"],
                 dry_run=ctx.get("dry_run", False),
                 early_exit_compare_sha=ctx.get("early_exit_compare_sha"),
+                check_only_affected_shards=check_only_affected_shards,
                 gql_sha_url=ctx["gql_sha_url"],
                 print_url=ctx["gql_url_print"],
                 run_args=args,
@@ -489,6 +514,7 @@ def init_log_level(log_level):
 @config_file
 @dry_run
 @early_exit
+@check_only_affected_shards
 @validate_schemas
 @dump_schemas
 @gql_sha_url
@@ -500,6 +526,7 @@ def integration(
     configfile,
     dry_run,
     early_exit_compare_sha,
+    check_only_affected_shards,
     validate_schemas,
     dump_schemas_file,
     log_level,
@@ -512,6 +539,7 @@ def integration(
     config.init_from_toml(configfile)
     ctx.obj["dry_run"] = dry_run
     ctx.obj["early_exit_compare_sha"] = early_exit_compare_sha
+    ctx.obj["check_only_affected_shards"] = check_only_affected_shards
     ctx.obj["validate_schemas"] = validate_schemas
     ctx.obj["gql_sha_url"] = gql_sha_url
     ctx.obj["gql_url_print"] = gql_url_print
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -5,7 +5,11 @@ import os
 import re
 import sys
 from signal import SIGUSR1
-from typing import Optional
+from types import ModuleType
+from typing import (
+    Optional,
+    Union,
+)
 
 import click
 import sentry_sdk
@@ -26,11 +30,15 @@ from reconcile.utils.binary import (
 from reconcile.utils.environ import environ
 from reconcile.utils.exceptions import PrintToFileInGitRepositoryError
 from reconcile.utils.git import is_file_in_git_repo
-from reconcile.utils.gql import (
-    GqlApiErrorForbiddenSchema,
-    GqlApiIntegrationNotFound,
+from reconcile.utils.runtime.integration import (
+    ModuleBasedQontractReconcileIntegration,
+    QontractReconcileIntegration,
 )
 from reconcile.utils.runtime.meta import IntegrationMeta
+from reconcile.utils.runtime.runner import (
+    IntegrationRunConfiguration,
+    run_integration_cfg,
+)
 from reconcile.utils.unleash import get_feature_toggle_state
 
 TERRAFORM_VERSION = "0.13.7"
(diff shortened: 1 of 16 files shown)


reconcile/cli.py
```diff
@@ -417,112 +425,61 @@ def register_faulthandler(fileobj=sys.__stderr__):
...
-        sys.exit(ExitCodes.SUCCESS)
-
-    dry_run = ctx.get("dry_run", False)
-    early_exit_compare_sha = ctx.get("early_exit_compare_sha")
+        if isinstance(func_container, QontractReconcileIntegration):
+            integration = func_container
+        elif isinstance(func_container, ModuleType):
+            integration = ModuleBasedQontractReconcileIntegration(func_container)
+        else:
+            raise Exception(f"Unknown integration type {type(func_container)}")
```
chassing, line 442: Please raise a dedicated exception (e.g., `UnknownIntegrationTypeError`) instead of `Exception`

reconcile/test/runtime/fixtures.py
```diff
@@ -0,0 +1,49 @@
...
+    Optional,
+)
+
+from reconcile.utils.runtime.integration import (
+    DesiredStateShardConfig,
+    QontractReconcileIntegration,
+)
+
+
+class TestIntegration(QontractReconcileIntegration):
```
chassing, line 12: Please don't name it `Test...` otherwise pytest thinks it's testcase and shows this warning

```
11:37:18 reconcile/test/runtime/fixtures.py:12
11:37:18   /package/reconcile/test/runtime/fixtures.py:12: PytestCollectionWarning: cannot collect test class 'TestIntegration' because it has a __init__ constructor (from: reconcile/test/runtime/test_utils_runtime_runner.py)
11:37:18     class TestIntegration(QontractReconcileIntegration):
```

reconcile/test/runtime/fixtures.py
```diff
@@ -0,0 +1,49 @@
+from typing import (
```
chassing, line 1: Maybe a personal taste, but I would use a `conftest.py` instead of this file and use the [pytest fixtures](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html) feature to consume `TestIntegration` or `ShardableTestIntegration` instances in my test cases

reconcile/utils/runtime/desired_state_diff.py
```diff
@@ -0,0 +1,158 @@
...
+    previous_desired_state: dict[str, Any],
+    current_desired_state: dict[str, Any],
+    return_value: dict,
+) -> None:
+    diffs = extraction_function(previous_desired_state, current_desired_state)
+    return_value["diffs"] = diffs
+
+
+class DiffDetectionTimeout(Exception):
+    pass
```
chassing, line 79: A `__doc__` would be fine.

reconcile/utils/runtime/desired_state_diff.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def extract_diffs_with_timeout(
```
chassing, line 82: A `__doc__` would be fine.

reconcile/utils/runtime/desired_state_diff.py
```diff
@@ -0,0 +1,158 @@
...
+            "no shard detection or sharded runs will be performed."
+        )
+        process.terminate()
+        process.join()
+        raise DiffDetectionTimeout()
+    else:
+        return result_value["diffs"]
+
+
+def build_desired_state_diff(
```
chassing, line 114: A `__doc__` would be fine.

reconcile/utils/runtime/integration.py
```diff
@@ -0,0 +1,85 @@
...
+from types import ModuleType
+from typing import (
+    Any,
+    Callable,
+    Optional,
+)
+
+
+@dataclass
+class ShardedRunProposal:
```
chassing, line 15: A `__doc__` would be fine.

reconcile/utils/runtime/integration.py
```diff
@@ -0,0 +1,85 @@
...
+
+    proposed_shards: set[str]
+
+
+@dataclass
+class DesiredStateShardConfig:
```
chassing, line 21: A `__doc__` would be fine.

reconcile/utils/runtime/integration.py
```diff
@@ -0,0 +1,85 @@
...
+    sharded_run_review: Callable[[ShardedRunProposal], bool]
+    shard_arg_name: str
+    shard_path_selectors: set[str]
+
+
+class QontractReconcileIntegration(ABC):
```
chassing, line 27: A `__doc__` would be fine.

reconcile/utils/runtime/integration.py
```diff
@@ -0,0 +1,85 @@
...
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        ...
+
+    @abstractmethod
+    def get_early_exit_desired_state(self, *args: Any, **kwargs: Any) -> dict[str, Any]:
+        ...
```
chassing, line 35: A `__doc__` would be fine.

reconcile/utils/runtime/integration.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @abstractmethod
+    def get_desired_state_shard_config(self) -> Optional["DesiredStateShardConfig"]:
+        ...
```
chassing, line 39: A `__doc__` would be fine.

reconcile/utils/runtime/integration.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @abstractmethod
+    def run(self, dry_run: bool, *args: Any, **kwargs: Any) -> None:
+        ...
```
chassing, line 43: A `__doc__` would be fine.

reconcile/utils/runtime/integration.py
```diff
@@ -0,0 +1,85 @@
...
+        sharding_config = self.get_desired_state_shard_config()
+        return sharding_config.shard_arg_name if sharding_config else None
+
+
+RUN_FUNCTION = "run"
+EARLY_EXIT_DESIRED_STATE_FUNCTION = "early_exit_desired_state"
+DESIRED_STATE_SHARD_CONFIG_FUNCTION = "desired_state_shard_config"
+
+
+class ModuleBasedQontractReconcileIntegration(QontractReconcileIntegration):
```
chassing, line 56: A `__doc__` would be fine.

reconcile/utils/runtime/runner.py
```diff
@@ -0,0 +1,182 @@
...
+        )
+        gql.init_from_config(
+            sha=self.early_exit_compare_sha,
+            integration=self.integration.name,
+            validate_schemas=final_validate_schemas,
+            print_url=self.print_url,
+        )
+
+
+def get_desired_state_diff(
```
chassing, line 78: A `__doc__` would be fine.

reconcile/utils/runtime/runner.py
```diff
@@ -0,0 +1,182 @@
...
+        return None
+
+    return build_desired_state_diff(
+        run_cfg.integration.get_desired_state_shard_config(),
+        previous_desired_state,
+        current_desired_state,
+    )
+
+
+def run_integration_cfg(run_cfg: IntegrationRunConfiguration) -> None:
```
chassing, line 111: A `__doc__` would be fine.

chassing (commented):
Some nits :)

reconcile/test/runtime/fixtures.py
geoberle: i see. makes sense. renamed in 0c630fb625ff1f4d83889ba63e0df7f0fb4466fb

reconcile/test/runtime/fixtures.py
geoberle: fixed in e6e3d6453535b1f29bc7615f1ee065c45fa06c64

reconcile/test/runtime/test_utils_runtime_runner.py
```diff
@@ -0,0 +1,397 @@
...
+        print_url=True,
+        run_args=(),
+        run_kwargs={},
+        main_data={"data": "a"},
+        comparison_data={"data": "b"},
+    )
+
+
+@pytest.fixture
+def wet_run_test_integration_cfg() -> IntegrationRunConfiguration:
```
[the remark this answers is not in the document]
geoberle: naming suggestions highly welcome :D

reconcile/utils/runtime/runner.py
geoberle: done

reconcile/utils/runtime/runner.py
geoberle: done

reconcile/utils/runtime/integration.py
geoberle: done

reconcile/utils/runtime/integration.py
geoberle: done

reconcile/utils/runtime/integration.py
geoberle: done

reconcile/cli.py
geoberle: done

maorfr (approved):
:rubber-stamp:

chassing (approved):
LGTM

reconcile/utils/runtime/desired_state_diff.py
```diff
@@ -0,0 +1,161 @@
...
+    def can_exit_early(self) -> bool:
+        return not self.diff_found
+
+    def can_run_sharded(self) -> bool:
+        return self.affected_shards != set()
+
+
+def find_changed_shards(
+    diffs: list[Diff],
+    previous_desired_state: dict[str, Any],
```
steveteahan, line 42: Reminder on abstracts over concrete inputs (concrete outputs are fine).

`dict` -> `Mapping`
`list` -> `Iterable`

and all that good stuff unless there's a good reason not to.

reconcile/utils/runtime/integration.py
```diff
@@ -0,0 +1,140 @@
...
+    # the name of the integrations `QontractReconcileIntegrationrun` function
+    # argument that is used to pass the shard to process
+    shard_arg_name: str
+    # some integrations offer the shard arg as a colletion of values
+    shard_arg_is_collection: bool = False
+
+
+class QontractReconcileIntegration(ABC):
+    """
+    The base class for all integrations. It defines the basic interface to interact
```
steveteahan, line 38: +1 to making this a class and `ABC`. Definitely time to start using this pattern moving forward.

reconcile/utils/runtime/runner.py
```diff
@@ -0,0 +1,200 @@
...
+from reconcile.utils import gql
+from reconcile.utils.runtime.desired_state_diff import (
+    DesiredStateDiff,
+    build_desired_state_diff,
+)
+from reconcile.utils.runtime.integration import QontractReconcileIntegration
+
+
+@dataclass
+class IntegrationRunConfiguration:
```
steveteahan, line 20: Docstring cover purpose and what each member is used for?

Our CI isn't enforcing this, so I'm not strongly opinionated, but a standard way we can doc these parameters to make them distinct from general comments: https://peps.python.org/pep-0224/

steveteahan (commented):
I did a very quick review, but the concept makes sense at a high-level and I like the approach of adding an abstract class to make this more uniform. I think that adding more docstrings and explanations for class members will make it a lot easier for the next person who has to modify the code (although it looks like the test coverage is good and well-written). This is only for public classes/functions and things that wouldn't be very obvious to someone with not a lot of previous knowledge.

No need to block on my approval. If I get time to spend more time on it before you merge I'll provide any other feedback.

reconcile/utils/runtime/runner.py
geoberle: I was not aware of attribute docstrings. TIL. I will add respective descriptions.

lechuk47:
I'm chewing this one. Do you have an example of how `DesiredStateConfig` would look to identify shards in integrations like `openshift-resources` where there is no such direct split as in `terraform-resources`?

geoberle:
> I'm chewing this one. Do you have an example of how `DesiredStateConfig` would look to identify shards in integrations like `openshift-resources` where there is no such direct split like in `terraform-resources`?

as you said, for `tf-resources`, there is a natural split of data based on aws-accounts. for `openshift-resources` where there is no balanced natural split, you need to find an artificial one. consider the following approach:

the base data that is looked at to find the affected shards is the data returned from `early_exit_desired_state`. you can change this function to add the artificial shard information to that data. e.g. `$cluster-$first_letter_for_namespace` (this is not really a proposal, just something simple to get the idea across). 

```json
{
    "namespaces":[
      {
         "shard": "cluster1-n",  <-- construct this in code in early_exit_desired_state
         "name": "my-namespace",
         ... regular namespace data ...
      }
    ],
    "resources": [
       <-- something similar here
    ],
}
```

the integration run function needs a `kwarg` to receive such a shard, e.g. `shard`. the integration then needs to process only those resources that belong to the affected artificial shard by applying the same consistent approach to resource selection that was used to build the artificial shard in `early_exit_desired_state`

then the `DesiredStateConfig` object would look like this

```python
DesiredStateShardConfig(
        shard_arg_name="shard",
        shard_path_selectors={
            "namespaces[*].shard",
            "resources[*].wherever.the.shard.can.be.found.here",
        },
        sharded_run_review=lambda proposal: len(proposal.proposed_shards) ==1, # only run sharded when 1 shard is affected
 )
```

this is just a rough sketch how it could be done with artificially constructed shards.

any thoughts?

PR #3141, opened 2023-01-23
[change-owners] auto approve when author is the only approver

maorfr:
part of https://issues.redhat.com/browse/APPSRE-6786

similar to #3028

when the author is the only approver - it's approved. tests will block merges if something is wrong. the user can only `/hold`.


geoberle (approved):
lgtm!

PR #3142, opened 2023-01-23
Define Cluster-Scoped resources in app-interface

lechuk47:
## Overview

This PR introduces a validation to ensure that cluster-scoped resources managed by App-Interface are not duplicated. 

With our current approach, multiple namespace files can define the same cluster-scoped resource using `openshiftResources`. That has happened sometimes and ends up with a clash of the reconciliation loop trying to reconcile the same resource from multiple locations. 

This change forces the cluster-scoped defined in `openshiftResoures` to be defined under `ManagedResourceNames` within the service namespace files. This allows a validation at PR-check to ensure that newly introduced cluster-scoped resources are not duplicated in the same cluster.

There are two checks: 
* CheckClusterScopedResourceNames: Ensures that all cluster-scoped resources in the desired state of a namespace are defined in the `ManagedResourceNames` list.
* CheckClusterScopedResourceDuplicates: Ensures that no cluster-scoped resource duplications exist in a cluster

**Note**: Before introducing this, all existent cluster-scoped resources should be defined in their namespace files

APPSRE-2935

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -1066,3 +1066,53 @@ def user_has_cluster_access(
     """Check user has access to cluster."""
     userkeys = determine_user_keys_for_access(cluster.name, cluster.auth)
     return any((getattr(user, userkey) in cluster_users for userkey in userkeys))
+
+
+def get_namespace_type_overrides(namespace: Mapping) -> dict[str, str]:
+    """Returns a dict with the type overrides defined in a namespace
+
+    :param namespace: the namespace
+    :return: dict with the the type overrides
+    """
+    return {
+        o["resource"]: o["override"]
+        for o in namespace.get("managedResourceTypeOverrides") or []
+    }
+
+
+def get_namespace_resource_types(
+    namespace: Mapping, type_overrides: Optional[dict[str, str]] = None
+) -> list[str]:
+    """Returns a list with the namespace ResourceTypes, with the overrides in place
+
+    :param namespace: the namespace
+    :param type_overrides: dict with the namespace type overrides
+    :return: Definitive managed resource TYPES for a namespace
+    """
+    if not type_overrides:
+        type_overrides = get_namespace_type_overrides(namespace)
+
+    return [
+        type_overrides.get(t, t) for t in namespace.get("managedResourceTypes") or []
+    ]
+
+
+def get_namespace_resource_names(
+    namespace: Mapping, type_overrides: Optional[dict[str, str]] = None
+) -> dict[str, list[str]]:
+    """Returns a list with the namespace ResourceTypeNames, with the overrides in place
+
+    :param namespace: the namespace
+    :param type_overrides: dict with the namespace type overrides
+    :return: Definitive managed resource NAMES for a namespace
+    """
+    if not type_overrides:
+        type_overrides = get_namespace_type_overrides(namespace)
+    rnames: dict[str, list[str]] = {}
+
+    for item in namespace.get("managedResourceNames") or []:
+        kind = type_overrides.get(item["resource"], item["resource"])
+        ref = rnames.setdefault(kind, [])
+        ref += item["resourceNames"]
+
+    return rnames
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -2,6 +2,7 @@ import base64
 import hashlib
 import itertools
 import hashlib
+import itertools
 import json
 import logging
 import sys
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -2,7 +2,6 @@ import base64
 import hashlib
 import itertools
 import hashlib
-import itertools
 import json
 import logging
 import sys
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -1,6 +1,7 @@
 import base64
 import hashlib
 import itertools
+import hashlib
 import json
 import logging
 import sys
@@ -1180,15 +1181,12 @@ def early_exit_desired_state(
             settings=settings,
         )
 
-    def post_process_ns(ns):
+    def add_ns_identify(ns):
         ns[IDENTIFIER_FIELD_NAME] = f"{ns['cluster']['name']}/{ns['name']}"
-        # the sharedResources have been aggreated into the openshiftResources
-        # and are no longer needed - speeds up diffing process
-        del ns["sharedResources"]
         return ns
 
     return {
-        "namespaces": [post_process_ns(ns) for ns in namespaces],
+        "namespaces": [add_ns_identify(ns) for ns in namespaces],
         "resources": resources,
     }
 
@@ -1211,7 +1209,6 @@ def _early_exit_fetch_resource(spec, settings):
         # detect changes in desired state
         c = resource["resource"].get("content")
     del resource["resource"]
-    resource[IDENTIFIER_FIELD_NAME] = id
     content_sha = hashlib.md5(c.encode("utf-8")).hexdigest()
     return {
         IDENTIFIER_FIELD_NAME: id,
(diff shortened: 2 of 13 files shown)


reconcile/openshift_base.py
```diff
@@ -1066,3 +1066,53 @@ def user_has_cluster_access(
...
+    :return: Definitive managed resource NAMES for a namespace
+    """
+    if not type_overrides:
+        type_overrides = get_namespace_type_overrides(namespace)
+    rnames: dict[str, list[str]] = {}
+
+    for item in namespace.get("managedResourceNames") or []:
+        kind = type_overrides.get(item["resource"], item["resource"])
+        ref = rnames.setdefault(kind, [])
+        ref += item["resourceNames"]
```
geoberle, line 1116: can `item["resourceNames"]` be None?

reconcile/openshift_resources_base.py
```diff
@@ -944,6 +959,209 @@ def run(
...
+) -> bool:
+
+    checks = [
+        CheckClusterScopedResourceNames(oc_map, ri, namespaces, thread_pool_size),
+        CheckClusterScopedResourceDuplicates(oc_map, all_namespaces, thread_pool_size),
+    ]
+
+    results = threaded.run(
+        lambda x: x.check(), checks, len(checks), return_exceptions=True
+    )
```
geoberle, line 1098: out of curiosity - how long do those checks take to complete?

reconcile/openshift_resources_base.py
```diff
@@ -944,6 +959,209 @@ def run(
...
+        cluster_resources_lookup = dict(results)
+        for cluster in cluster_resources_lookup:
+            for kind in cluster_resources_lookup[cluster]:
+                for name in cluster_resources_lookup[cluster][kind]:
+                    if len(cluster_resources_lookup[cluster][kind][name]) > 1:
+                        refs = cluster_resources_lookup[cluster][kind][name]
+                        errors.append(
+                            CheckError(
+                                f"Cluster resource defined in multiple namespaces. "
+                                f"cluster: {cluster}, namespaces: {refs}, "
```
geoberle, line 1055: nice!

reconcile/openshift_base.py
lechuk47: It does not make sense as these objects are meant to limit specific names but I think it's not set in the schemas. I'll update the schemas to enforce the definition.

reconcile/openshift_resources_base.py
geoberle: nice

geoberle (commented):
looks good. nothing caught my eye.
since these are changes on very central places i would love for someone else to have a look as well.

i suggest @janboll - Jan you had your fair share of oc.py and api-resources changes going on. would you have a look?

reconcile/utils/oc.py
```diff
@@ -1194,40 +1252,31 @@ def get_all(self, kind, all_namespaces=False):
             raise StatusCodeError(f"[{self.server}]: {e}")
 
     @staticmethod
-    def add_group_kind(kind, kgv, new, preferred):
+    def add_api_resource(kind, kgv, preferred, resource):
```
janboll, line 1254: Parameters:

- `kgv` Please use `kind_groupversion` instead
- `preferred` Preferred what?

lechuk47: `preferred` is part of the kubernetes client we are using, so I think it's ok to use the same.

I agree with `kgv`one. As this is now the `api_resources` dictionary, maybe it is better to use `api_resources` directly. WDYT? 

I'm also wondering if the `copy` is really needed. I tried to reuse the existing code as much as I could, but this might not be necessary.

lechuk47: I refactored this a bit using the same names as the Kubernetes dynamic client. PTAL

reconcile/openshift_resources_base.py
```diff
@@ -944,6 +959,209 @@ def run(
...
+    def check(self) -> list[Exception]:
+        pass
+
+
+@dataclass
+class CheckClusterScopedResourceNames:
+    oc_map: OC_Map
+    ri: ResourceInventory
+    namespaces: list[Mapping[str, Any]]
+    thread_pool_size: int = 10
```
janboll, line 976: Can't see usage of a threadpool, is this required?

reconcile/openshift_resources_base.py
```diff
@@ -944,6 +959,209 @@ def run(
...
+                            CheckError(
+                                f"Cluster resource defined in multiple namespaces. "
+                                f"cluster: {cluster}, namespaces: {refs}, "
+                                f"kind:{kind}, name:{name}"
+                            )
+                        )
+        return errors
+
+    def _remap_cluster_result(self, input: Tuple[str, Mapping]) -> Tuple[str, Mapping]:
+        remap: dict[str, dict[str, list[str]]] = {}
```
janboll, line 1062: please move below comment

reconcile/openshift_resources_base.py
```diff
@@ -944,6 +959,209 @@ def run(
...
+                        errors.append(
+                            CheckError(
+                                f"Cluster resource defined in multiple namespaces. "
+                                f"cluster: {cluster}, namespaces: {refs}, "
+                                f"kind:{kind}, name:{name}"
+                            )
+                        )
+        return errors
+
+    def _remap_cluster_result(self, input: Tuple[str, Mapping]) -> Tuple[str, Mapping]:
```
janboll, line 1061: Perhaps instead of remapping, you could do something like:
```
x = set()
duplicates = list[Tuple[str, str]]

for ns, resources in cluster_resources.items():
    if not ns:
        continue
    for kind, names in resources.items():
        k_ref = remap.setdefault(kind, {})
        for name in names:
            if (kind, name) in x:
                   duplicates.append( (kind, name) )
            else:
                   x.add ( (kind, name) )
return duplicates 
```

If I am not mistaken, this could make the check in CheckClusterScopedResourceDuplicates easier, as you'd have to loop over the duplicates only.

reconcile/openshift_resources_base.py
```diff
@@ -944,6 +959,209 @@ def run(
...
+    all_namespaces: Optional[list[Mapping]] = None
+    thread_pool_size: int = 10
+
+    def check(self) -> list[Exception]:
+        errors: list[Exception] = []
+        clusters = set(self.oc_map.clusters() + self.oc_map.clusters(privileged=True))
+        cluster_cs_resources = get_cluster_scoped_resources(
+            self.oc_map, clusters, self.all_namespaces
+        )
+        results = threaded.run(
```
janboll, line 1034: _remap_cluster_result is  iterating over a list, without side effects, which benefit does threading provide here?

janboll (commented):
Looking good! Thank you for the refactorings. 
Just have a couple of questions regarding the new dataclasses.

reconcile/openshift_resources_base.py
lechuk47: Good point. I'm working to "find the duplicates" instead of remapping the whole data. I want to report the involved namespaces in the logs so I'll change this snippet a bit.

reconcile/openshift_resources_base.py
lechuk47: The idea is to run the function in parallel for each cluster to do the process faster. Does that make sense?

reconcile/openshift_resources_base.py
lechuk47: PTAL

reconcile/openshift_resources_base.py
janboll: I was just wondering if it's really worth spawning threads for this given it should take only a few milliseconds.

janboll (approved):

lechuk47:
This is messed up.  It continues here: https://github.com/app-sre/qontract-reconcile/pull/3200/

PR #3145, opened 2023-01-24
Add integration to support Cloudflare DNS zone creation.

OliviaHY:
### Why:
To migrate from Dyn
### What:
New integration that can manage zones and records in Cloudflare. Based on this [design doc](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/terraform-cloudflare-dns-support.md).
### Validation:

- Locally run the integration successfully. Test cases:
- Create zone, see it shows up in my test Cloudflare account only with pending name server update.
- Delete created zones
- Create A, CNAME, NS, MX records
- Delete any created records
- Update attributes of those records
- Create NS and MX records with multiple values with same name
- Tested integration manager plan to create multiple pod with zone name passed as sharding key
- Tested Sharded dry run
- Performance test: zone with 1000 record creation took ~5min and used ~100MB RAM.
### Dependency:
[schema changes
](https://github.com/app-sre/qontract-schemas/pull/367)


reconcile/cli.py
```diff
@@ -1649,6 +1649,33 @@ def terraform_cloudflare_resources(
     )
 
 
+@integration.command(short_help="Manage Cloudflare DNS using Terraform.")
+@print_to_file
+@enable_deletion(default=False)
+@threaded(default=20)
+@binary(["terraform"])
+@binary_version("terraform", ["version"], TERRAFORM_VERSION_REGEX, TERRAFORM_VERSION)
+@account_name
```
steveteahan, line 1658: To select a subset of resources to run against, should we support account? zone? both?

reconcile/terraform_cloudflare_dns.py
```diff
@@ -0,0 +1,246 @@
...
+            aws_backend_config,
+            cf_acct.provider_version,
+        )
+
+        ts_client = TerrascriptCloudflareClient(ts_config)
+        clients.append((cf_acct.name, ts_client))
+    return clients
+
+
+def cloudflare_dns_zone_to_external_resource(
```
steveteahan, line 145: I'd definitely include a good docstring when you get around to it explaining why this is happening.

reconcile/terraform_cloudflare_dns.py
```diff
@@ -0,0 +1,246 @@
...
+    zones: list[CloudflareDnsZoneV1],
+) -> list[ExternalResourceSpec]:
+    external_resource_specs: List[ExternalResourceSpec] = []
+    default_provider = "cloudflare"
+    default_namespace = {
+        "name": "noname",
+        "cluster": {"name": "noname"},
```
steveteahan, line 152: How would I know that `noname` is a placeholder? Could this allow unexpected access of this attributes to get further than it should and fail in non-obvious ways? Are there other data types that might make this more obvious?

steveteahan (commented):
Adding a few comments for now. I'll add more as this continues to progress.

As we discussed before, I'd consider looking at #3134 once some of the other feedback is addressed. I don't think we'll need to do any profiling to know that Terraform is quite slow at planning these DNS zones and we should run only the ones that we need to.

reconcile/terraform_cloudflare_dns.py
OliviaHY: You mean a set?

reconcile/cli.py
OliviaHY: Both :)

reconcile/terraform_cloudflare_dns.py
steveteahan: Not quite, if I was writing some code to interface with this and I wanted to check whether this was dummy data or not, would I write something like:

```python
if name == 'noname':
```

or is there a better value/type that would signify that the data isn't really present without me knowing that I need to check for this special 'noname' string?

OliviaHY: Value can be set as boolean? Or even None?

steveteahan: Yeah, I was thinking `None` could be reasonable. That'd let us check if these values are even set or not without knowing that placeholder value like `noname`.

OliviaHY: I see. Let me try that.

reconcile/terraform_cloudflare_dns.py
```diff
@@ -0,0 +1,262 @@
...
+        raise ValueError(
+            f"AWS account {aws_acct.name} cannot be used for Cloudflare "
+            f"account {cf_acct.name} because it doesn't define a terraform state "
+        )
+
+    integrations = tf_state.integrations or []
+    for i in integrations or []:
+        name = i.integration
+        if name.replace("-", "_") == QONTRACT_INTEGRATION:
+            # we have to ensure the bucket key(file) is unique across
```
steveteahan, line 73: We had a bit of discussion about this: https://redhat-internal.slack.com/archives/C0354U32C9M/p1675186237722129

Can we make the comment clearer based on this disussion?

reconcile/terraform_cloudflare_dns.py
```diff
@@ -0,0 +1,262 @@
...
+    return clients
+
+
+def cloudflare_dns_zone_to_external_resource(
+    zones: list[CloudflareDnsZoneV1],
+) -> list[ExternalResourceSpec]:
+    external_resource_specs: List[ExternalResourceSpec] = []
+    default_provider = "cloudflare"
+    default_namespace = {
+        "name": False,
```
steveteahan, line 155: I'm curious why you decided on `False` vs. `None`?

reconcile/gql_definitions/terraform_cloudflare_dns/terraform_cloudflare_zones.gql
```diff
@@ -0,0 +1,64 @@
...
+      name
+      type
+      description
+      providerVersion
+      enforceTwofactor
+      apiCredentials {
+        path
+        field
+        format
+        version
```
fishi0x01, line 17: You can use `... VaultSecret` fragment here to reduce the number of generated classes.

reconcile/gql_definitions/terraform_cloudflare_dns/terraform_cloudflare_zones.gql
```diff
@@ -0,0 +1,64 @@
...
+      }
+      terraformStateAccount {
+        name
+        consoleUrl
+        terraformUsername
+        automationToken {
+          path
+          field
```
fishi0x01, line 25: Same as above: `... VaultSecret`

fishi0x01: It will also make the tests easier, as you will not have to import 2 different `VaultSecret` classes anylonger for the tests.

reconcile/cli.py
```diff
@@ -343,6 +343,16 @@ def account_name(function):
     return function
 
 
+def zone_name(function):
+    function = click.option(
+        "--zone-name",
+        help="DNS zone name to act on. Currently only applies to Cloudflare.",
```
steveteahan, line 349: If specific to Cloudflare, should this be `cloudflare_zone_name`?

reconcile/integrations_manager.py
```diff
@@ -131,6 +140,31 @@ def _aws_accounts_for_integration(self, integration: str) -> list[dict[str, Any]
         ]
 
 
+class CloudflareZoneShardManager(ShardingStrategy):
```
steveteahan, line 143: How about a nice docstring that explains what this strategy is about? Maybe the use case?

reconcile/integrations_manager.py
```diff
@@ -131,6 +140,31 @@ def _aws_accounts_for_integration(self, integration: str) -> list[dict[str, Any]
...
+    def __init__(self, cloudflare_zones: Optional[list[CloudflareDnsZoneV1]]):
```
steveteahan, line 144: Does this need to be a `list`? https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#abstract-collection-types

reconcile/integrations_manager.py
```diff
@@ -131,6 +140,31 @@ def _aws_accounts_for_integration(self, integration: str) -> list[dict[str, Any]
...
+        self.cloudflare_zones = cloudflare_zones or []
+
+    def build_integration_shards(
+        self, integration_meta: IntegrationMeta, _: Mapping[str, Any]
+    ) -> list[dict[str, Any]]:
+        if "--zone-name" in integration_meta.args:
+            cloudflare_zones = self.cloudflare_zones
```
steveteahan, line 151: nit: why does `cloudflare_zones` need to be used here instead of direct access to `self.cloudflare_zones`?

reconcile/integrations_manager.py
```diff
@@ -131,6 +140,31 @@ def _aws_accounts_for_integration(self, integration: str) -> list[dict[str, Any]
...
+                    "shard_name_suffix": f"-{zone.account.name}-{zone.identifier}"
+                    if len(cloudflare_zones) > 1
+                    else "",
+                    "extra_args": f"--zone-name {zone.identifier}",
+                }
+                for zone in cloudflare_zones
+            ]
+        else:
+            raise ValueError(
+                f"integration {integration_meta.name} does not support arg --zone-name required by the per-cloudflare-zone sharding strategy"
```
steveteahan, line 164:
```suggestion
                f"integration {integration_meta.name} does not support the provided argument. --zone-name is required by the per-cloudflare-zone sharding strategy."
```

reconcile/terraform_cloudflare_dns.py
```diff
@@ -0,0 +1,292 @@
...
+    DEFAULT_CLOUDFLARE_ACCOUNT_2FA,
+    DEFAULT_CLOUDFLARE_ACCOUNT_TYPE,
+    CloudflareAccountConfig,
+    TerraformS3BackendConfig,
+    TerrascriptCloudflareClient,
+    create_cloudflare_terrascript,
+)
+
+DEFAULT_NAMESPACE: Mapping[str, Any] = {
+    "name": False,
```
steveteahan, line 51: I mentioned this in a previous review, but why are these `False` instead of `None`? They're not boolean values, they don't have a value.

reconcile/terraform_cloudflare_dns.py
```diff
@@ -0,0 +1,292 @@
...
+    def __init__(self) -> None:
+        self.qontract_integration = "terraform_cloudflare_dns"
+        self.qontract_integration_version = make_semver(0, 1, 0)
+        self.qontract_tf_prefix = "qrtfcfdns"
+
+    @property
+    def name(self) -> str:
+        return self.qontract_integration.replace("_", "-")
+
+    def create_backend_config(
```
steveteahan, line 71: What is the purpose of the `TerraformCloudflareDNSIntegration`? What does it do? Does it need this method and others that were previously functions?

steveteahan (commented):
Providing some initial feedback. I still need to get through the other half of the review.

reconcile/terraform_cloudflare_dns.py
```diff
@@ -0,0 +1,292 @@
+import logging
+import sys
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    List,
+    Mapping,
```
fishi0x01, line 8: These should be imported `from collection.abc import (...)` for [PEP-585](https://peps.python.org/pep-0585/) compliance.

fishi0x01: Further, I think `List` and `Dict` can be omitted, as `list` and `dict` are also valid.

reconcile/terraform_cloudflare_dns.py
OliviaHY: No particular reason. I probably made a mistake. Changing it to None

reconcile/integrations_manager.py
OliviaHY: Just because it was used more than once in the code below. But I don't mind either way.

reconcile/integrations_manager.py
OliviaHY: No, changing it to Iterable

reconcile/cli.py
OliviaHY: I think this would be a trade off. `cloudflare_zone_name` just seem to be a long argument to pass in the command line and  if I understand it correctly, the scope of this argument is limited to this integration, so the cloudflare prefix can be skipped since the integration will have that piece of information. Similarly, other integrations use `account_name` instead of `cloudflare_account_name` or `aws_account_name`, although it would be more clear to people who read the code.

reconcile/terraform_cloudflare_dns.py
OliviaHY: This is part of https://github.com/app-sre/qontract-reconcile/pull/3134/files#diff-dc5ee8a177d8eecdd1e8f2adda7b5dfc0481a16aa56bf830a8e26558439bef5f 
Instead of adding integrations as module, they are now getting added as classes that implements TerraformCloudflareDNSIntegration. I'm not sure I'm following the second part of your questions. The methods are the same as they were functions.

reconcile/cli.py
steveteahan: Sorry, I meant:

```python
def cloudflare_zone_name(function):
    ... everything else the same
```

That way you can keep it `--zone-name`, but nobody else starts using this for other use cases, especially when the help message indicates it's for Cloudflare only. You could probably also just remove that bit altogether?

reconcile/integrations_manager.py
steveteahan: It's just saving a small number of characters for `self.`, but now I have to look to see where it came from rather than knowing this is the instance variable. It's not a huge point, it just seems like something we can skip.

reconcile/cli.py
OliviaHY: Ah yeah makes sense. "remove that bit altogether" that bit referring to the help message?

steveteahan: Yes, exactly, it obvious from the function name that it's for Cloudflare, and that message isn't really helpful to the person running it.

reconcile/terraform_cloudflare_dns.py
```diff
@@ -0,0 +1,300 @@
...
+            )
+
+        return backend_config
+
+    def get_cf_acct_config(
+        self,
+        cf_acct: CloudflareAccountV1,
+        secret_reader: SecretReaderBase,
+    ) -> CloudflareAccountConfig:
+        cf_acct_creds = secret_reader.read_all({"path": cf_acct.api_credentials.path})
```
fishi0x01, line 124: Sorry didn't catch the first time. 

```
secret_reader.read_all_secret(cf_acct.api_credentials)
```

reconcile/terraform_cloudflare_dns.py
steveteahan: So, the purpose of `TerraformCloudflareDNSIntegration` is to implement `QontractReconcileIntegration` and standardize some aspects of how we run integrations. My question is why this class has to know anything about creating backend configs, configuring accounts, building clients, etc? Basically what I'm getting at is the smaller the classes are (within reason), and the less that they do, the easier they will generally be to understand and test.

reconcile/terraform_cloudflare_dns.py
OliviaHY: No, I should have caught it :)

reconcile/terraform_cloudflare_dns.py
OliviaHY: Talked offline. Agree with the approach as it breaks the logic up. It will produce more specialized classes, and the functions stays in the module will facilitate later iteration that abstract those functionalities.

reconcile/terraform_cloudflare_dns.py
```diff
@@ -0,0 +1,300 @@
...
+                logging.error("Deletions detected but they are disabled")
+                sys.exit(ExitCodes.ERROR)
+
+            if dry_run:
+                sys.exit(ExitCodes.SUCCESS)
+
+            err = tf.apply()
+            if err:
+                sys.exit(ExitCodes.ERROR)
+        except Exception as e:
```
steveteahan, line 278: What is this goal of this `try`/`except` block? How is the behavior different from if it didn't exist?

reconcile/utils/terrascript/cloudflare_client.py
```diff
@@ -84,10 +87,12 @@ def create_cloudflare_terrascript(
         "enforce_twofactor": account_config.enforce_twofactor,
         "type": account_config.type,
     }
-    terrascript += cloudflare_account(
-        account_config.name,
-        **cloudflare_account_values,
-    )
+
+    if not exist:
```
steveteahan, line 91: Maybe `is_managed_account` makes more sense than `exist`? I'd document this parameter in the docstring. The rest are pretty self-explanatory, but why you'd use this or not may not be.

Does this mean that we're saying we're deferring this work and there will be a future integration to manage account-level settings? Has this been captured in a ticket?

reconcile/utils/terrascript/cloudflare_client.py
```diff
@@ -42,17 +43,21 @@ class CloudflareAccountConfig:
...
     account_config: CloudflareAccountConfig,
     backend_config: TerraformS3BackendConfig,
     provider_version: str,
+    is_managed_account: bool = True,
 ) -> Terrascript:
     """
     Configures a Terrascript class with the required provider(s) and backend
     configuration. This is offloaded to a separate function to avoid mixing additional
-    logic into TerrascriptCloudflareClient.
+    logic into TerrascriptCloudflareClient. The is_managed_account argument indicates
```
steveteahan, line 58: There's a standard way to create docstrings: https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#docstrings

params are sometimes excluded if the intent is very clear, but if we're going to document the param it should use this format.

nit: you can remove documentation of the default because it's defined above, it's unlikely to change, but if it does this doc could become stale

I'd suggest a quick explanation of why you'd enable this or not?

reconcile/utils/terrascript/cloudflare_resources.py
```diff
@@ -177,7 +177,7 @@ def populate(self) -> list[Union[Resource, Output]]:
             resources.append(cloudflare_argo(self._spec.identifier, **argo_values))
 
         for record in zone_records:
-            identifier = safe_resource_id(record.get("name"))
+            identifier = safe_resource_id(record.pop("identifier"))
```
steveteahan, line 180: This should probably be breaking a test. Any thoughts about modifications to a test in `reconcile/test/test_utils_terrascript_cloudflare_client.py` that would have caught this change?

steveteahan:
Besides my request for a performance test including a 1000 record zone, I don't think I have any other high-level feedback here.

reconcile/utils/terrascript/cloudflare_resources.py
OliviaHY: Yeah I'll update `test_create_cloudflare_resources_terraform_json`

reconcile/test/test_utils_terrascript_cloudflare_client.py
```diff
@@ -185,8 +199,23 @@ def test_terrascript_cloudflare_client_dump_existing_dir(mocker):
     mock_builtins_open.return_value.write.assert_called_once_with("some data")
 
 
-def test_create_cloudflare_terrascript(account_config, backend_config):
-    """Simple test to ensure that the Terrascript object is initialized."""
+def test_create_cloudflare_terrascript(
+    account_config, backend_config, cloudflare_account_test
+):
+    """This is to ensure that the Terrascript object is initialized and cloudflare account config gets added"""
```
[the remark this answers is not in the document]
OliviaHY: Yes, separated the tests.

reconcile/utils/terrascript/cloudflare_client.py
```diff
@@ -42,17 +43,33 @@ class CloudflareAccountConfig:
...
     backend_config: TerraformS3BackendConfig,
     provider_version: str,
+    is_managed_account: bool = True,
 ) -> Terrascript:
     """
     Configures a Terrascript class with the required provider(s) and backend
     configuration. This is offloaded to a separate function to avoid mixing additional
     logic into TerrascriptCloudflareClient.
+
+    Args:
```
steveteahan, line 60: Please switch to the style we use in the style guide and omit types since we already have type hints.

steveteahan (approved):
One minor change left, but otherwise this LGTM.

reconcile/utils/terrascript/cloudflare_client.py
OliviaHY: FYI the format I'm using right now is generated by the extension recommended in the [dev guide ](https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#docstrings)

steveteahan: Yeah, we should probably remove that suggestion. It seems outdated.

steveteahan:
Last round of changes LGTM.

PR #3146, opened 2023-01-26
[change-owners] improve diff detection when values repeat in lists

geoberle:
this might be an edge case but currently change-owners is unaware of certain changes. if a value in a list is duplicated, such duplicate values will not be reported as diff if they are next to each other in the list.

fixed in this PR

https://issues.redhat.com/browse/APPSRE-6889


geoberle:
sadly this one is not that easy. i need to investigate if this can be solved in the deepdiff lib

PR #3151, opened 2023-01-30
[terraform-users] sharded pr-checks on affected accounts

geoberle:
terraform-users is an integration that runs very long if it has to process all aws accounts. by running it only on affected aws-accounts, the dry-run runtime is a fraction of what it is when it runs in full.

like in https://github.com/app-sre/qontract-reconcile/pull/3149, the `IDENTIFIER_FIELD_NAME` field is introduced to speed up the fine grained diff extraction between the involved desired states.

https://issues.redhat.com/browse/APPSRE-6967

### DIFF
--- a/reconcile/terraform_users.py
+++ b/reconcile/terraform_users.py
@@ -212,6 +212,17 @@ def run(
 
 
 def early_exit_desired_state(*args, **kwargs) -> dict[str, Any]:
+
+    """
+    Finding diffs in deeply nested structures is time/resource consuming.
+    Having a unique known property to identify objects makes it easier to match
+    the same object in different states. This speeds up the diffing process
+    A LOT!
+
+    The IDENTIFIER_FIELD_NAME is added for that purpose. It is a well known field
+    for the DeepDiff library used in qontract-reconcile.
+    """
+
     def add_account_identity(acc):
         acc[IDENTIFIER_FIELD_NAME] = acc["path"]
         return acc
@@ -237,5 +248,6 @@ def desired_state_shard_config() -> DesiredStateShardConfig:
             "roles[*].aws_groups[*].account.name",
             "roles[*].user_policies[*].account.name",
         },
+        # Only run shard if less than 2 shards are affected. Else it is not worth the effort -> run everything.
         sharded_run_review=lambda proposal: len(proposal.proposed_shards) <= 2,
     )
--- a/reconcile/terraform_users.py
+++ b/reconcile/terraform_users.py
@@ -9,11 +9,13 @@ from reconcile import (
     queries,
     typed_queries,
 )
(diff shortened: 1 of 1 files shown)


reconcile/terraform_users.py
```diff
@@ -210,7 +212,30 @@ def run(
...
+
+def desired_state_shard_config() -> DesiredStateShardConfig:
+    return DesiredStateShardConfig(
+        shard_arg_name="account_name",
+        shard_path_selectors={
+            "accounts[*].name",
+            "roles[*].aws_groups[*].account.name",
+            "roles[*].user_policies[*].account.name",
+        },
+        sharded_run_review=lambda proposal: len(proposal.proposed_shards) <= 2,
```
fishi0x01, line 240: nit: maybe another small comment

```
# Only run shard if less than 2 shards are affected. Else it is not worth the effort -> run everything.
```

Or alternatively, name the arg to sth like `only_run_sharded_if`?

reconcile/terraform_users.py
```diff
@@ -210,7 +212,30 @@ def run(
 
 
 def early_exit_desired_state(*args, **kwargs) -> dict[str, Any]:
+    def add_account_identity(acc):
```
fishi0x01, line 215: nit: It might be good to have a comment here why this is necessary. It made sense to my now after you explained it to me :)

```
"""
Finding diffs in deeply nested structures is time/resource consuming.
Having unique known properties makes it easier to match the same object in different states.
I.e., we add a known unique property here to speed-up diff search in a later step.
"""
```

fishi0x01 (approved):
Awesome magic! :+1:

reconcile/terraform_users.py
```diff
@@ -210,7 +212,30 @@ def run(
...
+        ],
+        "roles": [add_role_identity(r) for r in get_tf_roles()],
     }
+
+
+def desired_state_shard_config() -> DesiredStateShardConfig:
+    return DesiredStateShardConfig(
+        shard_arg_name="account_name",
+        shard_path_selectors={
+            "accounts[*].name",
```
fishi0x01, line 236: Just for my understanding:

this is required to cover `resetPasswords` property right?

geoberle: yes, but also if the terraform state configuration (what bucket to use etc) changes for an account.
the assumption is that every queries field is queried for a purpose and is therefore part of the desired state.

reconcile/terraform_users.py
geoberle: fixed

reconcile/terraform_users.py
geoberle: fixed

PR #3154, opened 2023-02-01
APPSRE-6609 introduce typed OCMap

fishi0x01:
**Problem**

The untyped `OC_Map` utils class prevents us from using generated queries in many integrations.

The overall refactor is too big to fit into one PR, so this PR is solely the beginning of refactoring `OC_Map` towards a typed version. 

**Details**

Initially `OC_Map` is extracted into a dedicated file and renamed `OCMap`. I.e., this doesn't affect any integration that is not using this new class. Further, the new `OCMap` still takes the untyped dictionaries as an extra argument to keep backwards compatible with the underlying `OCDeprecated` and `OCNative` class implementations¸ e.g.,:

```python
clusters: list[ClusterV1] = get_clusters()

# convenience function that is part of this PR
connection_parameters = get_connection_parameters_from_clusters(
    secret_reader=secret_reader,
    clusters=clusters,
)
oc_map = OCMap(
        connection_parameters=connection_parameters,
        # This arg can be simply removed later once the refactor is finished.
        clusters_untyped=[cluster.dict(by_alias=True) for cluster in clusters],
        ....
 )
```

Note, that the implementation of `OCMap` is copied from `OC_Map`. The goal here is to change as less as possible.

In the next step, the extra `clusters_untyped` args will be removed once the refactor of the underlying `OCDeprecated` and `OCNative` clients is done. Specifically the jumphost settings need to be put into a proper protocol in a follow up PR.

Once all refactor PRs are done, the idea is to move the remaining integrations in follow-up PRs to use this new `OCMap` class, which in theory should be very simple. By moving integrations 1-by-1 or in small groups we reduce the blast-radius.

**Tests**

Tested this with `dashdotdb-cso` integration (which is also part of this PR). Our stream owns dashdotdb and if sth blows up here we will take care of it asap.

Further tested this with `cluster-deployment-mapper` integration, but this isn't part of this PR.


fishi0x01:
Alternatively we might consider to flatten the input a little, i.e., not passing the whole `clusters` list.

```python
@dataclass
class ClusterParameters:
  server_url: str
  jumphost_port: int
  jumphost_remote_port: int
  skip_integrations: list[str]
  ...
```

And then pass that in to `OCMap`

```python
clusters = get_clusters()  # generated dataclasses
cluster_parameters = extract_parameters(clusters)
OCMap(
  cluster_parameters=cluster_parameters,
  ...
)
```

That way we achieve higher decoupling between generated classes and `OCMap` utils class. I know this has been discussed before, but I'm wondering which way might be better here.

reconcile/utils/oc_map.py
```diff
@@ -0,0 +1,317 @@
...
+            )
+        elif namespaces and namespaces_untyped:
+            for ns_dict in namespaces_untyped:
+                cluster_d = ns_dict.get("cluster")
+                if cluster_d:
+                    self._jumphosts_dict[cluster_d.get("name")] = cluster_d.get(
+                        "jumpHost"
+                    )
+            clusters_dict = {}
+            privileged_clusters: dict[str, HasCluster] = {}
```
hemslo, line 121: Do we need to use dict here? I can only find usage of `.values()`. If not then can be simplified with list like
```python
unprivileged_clusters = [ns_info.cluster for ns_info in namespaces]
privileged_clusters = unprivileged_clusters if cluster_admin else [ns_info.cluster for ns_info in namespaces if ns_info.cluster_admin]

```

geoberle: @fishi0x01 i can relate a lot to your `class ClusterParameters` proposal. having minimal input is nice in many ways, e.g. lots of integration don't need to have access to unnecessary schemas (`allowedSchemas`) hence run less frequent during PR checks. also, it would be a lot easier to write tests.

reconcile/utils/oc_map.py
```diff
@@ -0,0 +1,317 @@
...
+    SecretReaderBase,
+)
+
+
+class HasDisable(Protocol):
+    integrations: Optional[list[str]]
+    e2e_tests: Optional[list[str]]
+
+
+class HasCluster(Protocol):
```
geoberle, line 34: i'm always having a hard time with python Protocol naming, especially with the `Has` prefix. i had a discussion with @chassing about that. in a nutshell: in a situation like this, the protocol describes an object "being" something, and therefore it would be ok to bluntly naming it `Cluster` or `ClusterConnectionConfig`. we can see this approach a lot within the collection Protocols, e.g. `Mapping` or `Sequence`.

reconcile/utils/oc_map.py
```diff
@@ -0,0 +1,317 @@
...
+                privileged=cluster_admin,
+            )
+        elif namespaces and namespaces_untyped:
+            for ns_dict in namespaces_untyped:
+                cluster_d = ns_dict.get("cluster")
+                if cluster_d:
+                    self._jumphosts_dict[cluster_d.get("name")] = cluster_d.get(
+                        "jumpHost"
+                    )
+            clusters_dict = {}
```
geoberle, line 120: since you thoroughly defined types everywhere else, please define it here as well.

reconcile/utils/oc_map.py
```diff
@@ -0,0 +1,317 @@
...
+                privileged,
+            )
+        else:
+            server_url = cluster_info.server_url
+            insecure_skip_tls_verify = cluster_info.insecure_skip_tls_verify
+
+            try:
+                if not self._secret_reader:
+                    raise Exception("No secret_reader set")
+                token = self._secret_reader.read_secret(automation_token)
```
geoberle, line 209: it would be great if we would not need the secret reader in `Oc_Map`. you have more context right now so wdyt about resolving secrets before handing clusters to `OC_Map`.

this question is clearly out of scope for this PR. carry to thought over to the followup PRs if you like

reconcile/utils/oc_map.py
```diff
@@ -0,0 +1,317 @@
...
+    In case a cluster does not have an automation token
+    the OC client will be initiated with a OCLogMessage.
+    """
+
+    def __init__(
+        self,
+        clusters: Optional[Sequence[HasCluster]] = None,
+        clusters_untyped: Optional[Sequence[MutableMapping[Any, Any]]] = None,
+        namespaces: Optional[Sequence[HasNamespace]] = None,
+        namespaces_untyped: Optional[Sequence[MutableMapping[Any, Any]]] = None,
```
geoberle, line 73: providing both typed and untyped data here irritates me a bit.

have you considered to drop support of untyped clusters and namespaces in this copy `OC_Map` instead of offering the mixed mode? since the old version still lives in `oc.py` existing usage is unaffected and integrations wanting to adopt the new version need to be touched anyways and can adhere to the typed approach during that process.

geoberle (commented):
thank you for taking care of this. this is really necessary!!

reconcile/utils/oc_map.py
fishi0x01: The untyped data is currently needed for the underlying classes `OCDeprecated`, `OCNative` and `SSHJumphost`. The idea was to not refactor them all in a single PR, but rather split this up into multiple (working) PRs.

Basically I wanted to create follow-up PRs shortly after this one in which I will refactor those underlying classes. But for this first step to be functional, I still need to pass the old nested dictionaries to work with the old classes.

reconcile/utils/oc_map.py
fishi0x01: I still had our old convention in mind. But what you suggest makes sense - TIL :) :+1: 

Anyways I'm more and more in favor of using `ClusterParameters` instead of Protocols. Now since you mentioned above that you like it too, I will introduce `ClusterParameters`.

reconcile/utils/oc_map.py
fishi0x01: Good catch. Fixed.

reconcile/utils/oc_map.py
fishi0x01: @hemslo good point. The idea was to copy/paste as much as possible from the existing logic and make it type-compatible with as less changes as possible. I.e., I didn't do any optimization on the logic itself.

reconcile/utils/oc_map.py
fishi0x01: I like the idea a lot. Once the underlying classes `OCDeprecated`, `OCNative` and `SSHJumphost` are type compatible, I will try to remove the need of secret reader. It makes sense :+1:

reconcile/utils/oc_map.py
fishi0x01: If possible, I'd like to address these optimizations in a final polishing PR, once it was confirmed that the types are working w/o breaking anything.

reconcile/utils/oc_map.py
geoberle: ah makes sense. thanks for the details!

reconcile/utils/oc_map.py
geoberle: cool. thanks!

reconcile/utils/oc_map.py
```diff
@@ -0,0 +1,317 @@
...
+    @property
+    def cluster_admin_automation_token(self) -> Optional[HasSecret]:
+        ...
+
+    @property
+    def disable(self) -> Optional[HasDisable]:
+        ...
+
+
+class HasNamespace(Protocol):
```
geoberle, line 53: <copy paste protocol naming comment>

reconcile/utils/oc_map.py
fishi0x01: Introduced now `OCConnectionParameters` class. It still has Protocols for convenience purposes, but they are named now as you suggested.

reconcile/utils/oc_map.py
fishi0x01: Same here :)
https://github.com/app-sre/qontract-reconcile/pull/3154#discussion_r1095234838

reconcile/utils/oc_map.py
fishi0x01: Due to introduction of `OCConnectionParameters` class I had to change the `OCMap` `__init__` function and applied your above mentioned suggestion.

reconcile/utils/oc_map.py
```diff
@@ -0,0 +1,253 @@
...
+        # a future PR.
+        jh["remotePort"] = 8888
+        key = f"{jh['hostname']}:{jh['remotePort']}"
+        with self._lock:
+            if key not in self._jh_ports:
+                port = JumpHostSSH.get_unique_random_port()
+                self._jh_ports[key] = port
+            jh["localPort"] = self._jh_ports[key]
+
+    def init_oc_client(
```
rporres, line 114: shouldn't this be private?

reconcile/utils/oc_map.py
```diff
@@ -0,0 +1,253 @@
...
+            privileged=False,
+        )
+        threaded.run(
+            self.init_oc_client,
+            privileged_clusters,
+            self._thread_pool_size,
+            privileged=True,
+        )
+
+    def set_jh_ports(self, jh: MutableMapping[Any, Any]) -> None:
```
rporres, line 103: shouldn't this be private?

rporres (commented):
some nit-picking about public methds but overall this is looking great.

How is the migration plan you envision? I'm just concerned in potential issues while OC_Map and OCMap coexist

fishi0x01:
@rporres 

> How is the migration plan you envision? I'm just concerned in potential issues while OC_Map and OCMap coexist

1. I will create follow-up PRs shortly after this is merged. These follow-up PRs refactor the underlying classes `OCDeprecated`, `OCNative` and `SSHJumphost` to also work with the new `OCConnectionParameters` class. 
2. Once the refactor is done, I will create PRs to shift all integrations 1-by-1 to use the new `OCMap`. 1-by-1 to keep the blast radius small if sth doesn't work.
3. I will delete the old `OC_Map` from `oc.py`.

I fully agree that `OC_Map` and `OCMap` should not co-exist for longer than necessary, but a certain overlap is needed. I do not dare refactor this in a single PR. This PR is alrdy way bigger than I wanted it to be. Further, this is a sensitive place - I want to keep the blast radius as small as possible.

reconcile/utils/oc_map.py
fishi0x01: Good catch. Fixed :+1:

reconcile/utils/oc_map.py
fishi0x01: Fixed :+1: 

Also fixed in `disabled()` function now

geoberle (approved):

rporres (approved):
migration path looks reasonable to me 👍

PR #3155, opened 2023-02-01
pylint complains about broad exceptions

fishi0x01:
Pylint recently started complaining about `broad-exception-raised` in multiple modules.

The errors seem to come from the latest pylint release https://github.com/PyCQA/pylint/releases/tag/v2.16.0

Since we explicitly do not want to pin mypy, I guess the same should hold true for pylint.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -3,7 +3,6 @@ extension-pkg-allow-list=pydantic
 
 [MESSAGES CONTROL]
 disable = attribute-defined-outside-init,
-          broad-except,
           duplicate-code,
           expression-not-assigned,
           fixme,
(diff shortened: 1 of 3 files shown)


geoberle:
can you fix this one? because this should really be fixed (shame on me):

```
reconcile/test/test_status_page_components.py:77:11: R1735: Consider using '{"name": component.display_name, "description": component.description, ... }' instead of a call to 'dict'. (use-dict-literal)
```

geoberle:
i'm in favour of fixing the `broad-exception-raised` ones over time. ignoring them for now is ok.

geoberle (approved):

PR #3156, opened 2023-02-01
[change-owners] role members can cancel /lgtm and /hold from role peers

geoberle:
Currently, a user adding a hold on an MR is the only one able to remove that hold with `/hold cancel`. this might block the MR from being merged even if another role member decides that it is now good to proceed.

this change fixes the situation while still making sure that a person from another team can't cancel another persons hold.

```
user 1 from role A: /hold
user 2 from role A: /hold cancel

result: hold is canceled
```

```
user 1 from role A: /hold
user 2 from role B: /hold cancel

result: MR is still on hold
```

https://issues.redhat.com/browse/APPSRE-6942

### DIFF
--- a/reconcile/change_owners/decision.py
+++ b/reconcile/change_owners/decision.py
@@ -1,7 +1,10 @@
 from collections import defaultdict
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any
+from typing import (
+    Any,
+    Iterable,
+)
 
 from reconcile.change_owners.bundle import FileRef
 from reconcile.change_owners.change_types import (
@@ -11,13 +14,6 @@ from reconcile.change_owners.change_types import (
 from reconcile.change_owners.diff import Diff
 
 
-@dataclass
-class Decision:
-
-    approve: bool = False
-    hold: bool = False
-
-
 class DecisionCommand(Enum):
     APPROVED = "/lgtm"
     CANCEL_APPROVED = "/lgtm cancel"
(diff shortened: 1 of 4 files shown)


reconcile/change_owners/decision.py
```diff
@@ -1,7 +1,10 @@
 from collections import defaultdict
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any
+from typing import (
+    Any,
+    Iterable,
```
fishi0x01, line 6: My understanding is we should from now on import that from `collections.abc` for [PEP-585](https://peps.python.org/pep-0585/) compliance.

reconcile/change_owners/decision.py
```diff
@@ -11,38 +14,50 @@
...
+
+    approver_name: str
+    command: DecisionCommand
+
+
 def get_approver_decisions_from_mr_comments(
-    comments: list[dict[str, Any]]
-) -> dict[str, Decision]:
-    decisions_by_users: dict[str, Decision] = defaultdict(Decision)
+    comments: Iterable[dict[str, Any]]
```
fishi0x01, line 32: nit: would `Mapping` also work here (`Iterable[Mapping[str, Any]]`)?

reconcile/test/change_owners/test_change_type_decision.py
```diff
@@ -179,16 +149,124 @@ def test_change_decision(
     ]
 
     change_decision = apply_decisions_to_changes(
-        approver_decisions={
-            yea_user: Decision(approve=True, hold=False),
-            nay_sayer: Decision(approve=False, hold=True),
-        },
+        approver_decisions=[
+            Decision(approver_name=yea_user, command=DecisionCommand.APPROVED),
+            Decision(approver_name=nay_sayer, command=DecisionCommand.HOLD),
```
fishi0x01, line 154: great variable names :laughing: :+1:

fishi0x01 (approved):
LGTM (as far as I can tell :sweat_smile: )

reconcile/change_owners/decision.py
geoberle: :facepalm: you are right. thanks!!

reconcile/change_owners/decision.py
geoberle: yes for sure. let me fix that

reconcile/change_owners/decision.py
geoberle: fixed

reconcile/change_owners/decision.py
geoberle: fixed

PR #3159, opened 2023-02-03
[gitlab-housekeeping] retest approved failed MRs

maorfr:
in case approved MRs are all rebased, and pipelines failed for an unrelated reason which was solved - they will not get merged unless something else got merged and a rebase happened.

this PR adds a functionality that adds a `/retest` comment to such "stuck" MRs.


rporres (approved):

d-hat:
I think this change is incorrect.  For one, `/retest` is [not a gitlab action](https://docs.gitlab.com/ee/user/project/quick_actions.html) and for two, the bot adds the comment repeatedly every time it runs (example MR#28 @ internal prodsec-app-sre-test repo).

PR #3160, opened 2023-02-03
[autopromoter] ignore null elements in the promotions list

geoberle:
if a null element makes it into the promotions list of an `AutoPromoter` message, the processing fails with an error. ignore those elmenets is a quick fix to resolve the issue.

a proper solution would prevent such elements from being produced. right now we don't know yet how they come into existence.

### DIFF
--- a/reconcile/utils/mr/auto_promoter.py
+++ b/reconcile/utils/mr/auto_promoter.py
@@ -33,7 +33,7 @@ class AutoPromoter(MergeRequestBase):
     name = "auto_promoter"
 
     def __init__(self, promotions):
-        self.promotions = promotions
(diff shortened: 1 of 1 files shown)


patjlm (approved):

patjlm:
FWIW, I "think" we can get a `null` promotion when we have a `delete: true` target in the same saas file, targeting the same environment (so being executed in the same saas-herder run). In that case, a `None` promotion is returned [here](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/saasherder.py#L1272-L1274) and gets added to the list of promotions being pushed for AutoPromoter

PR #3166, opened 2023-02-07
APPSRE-6609 typed jumphost (OC_Map refactor Part 2)

fishi0x01:
Add types and tests for jumphost implementation. This is required to have a typed `OCMap` class.

Follow-up of `OC_Map` refactor https://github.com/app-sre/qontract-reconcile/pull/3154

After this is merged, another PR will be opened to make `OCDeprecated` and `OCNative` fully compatible with new `OCMap` (in a backwards-compatible manner).

**Test**

- Tested with new `OCMap` class (dashdotdb-cso integration)
- Tested with old `OC_Map` class (openshift-users integration)

### DIFF
--- a/reconcile/test/test_jump_host.py
+++ b/reconcile/test/test_jump_host.py
@@ -21,7 +21,7 @@ EXPECTED_KNOWN_HOSTS_CONTENT = "known-hosts-file-content"
 
 
 @pytest.mark.parametrize(
-    "parameters,expected_port",
+    "parameters, expected_port",
     [
         (
             # Jumphost with default port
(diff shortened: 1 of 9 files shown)


reconcile/test/test_jump_host.py
```diff
@@ -0,0 +1,117 @@
...
+)
+
+EXPECTED_USER = "test-user"
+EXPECTED_HOSTNAME = "test"
+EXPECTED_KNOWN_HOSTS_PATH = "path/to/known-hosts"
+EXPECTED_KNOWN_HOSTS_CONTENT = "known-hosts-file-content"
+
+
+@pytest.mark.parametrize(
+    "parameters,expected_port",
```
chassing, line 24: Please add a space after the `,` 🙇

chassing (approved):
Great work ❤️

reconcile/test/test_jump_host.py
fishi0x01: Good catch. Fixed it here and in another test file now too.

PR #3173, opened 2023-02-07
Add exclude account parameter

esron:
Related to  [APPSRE-6586](https://issues.redhat.com/browse/APPSRE-6586) 

Things to teste here:

* Running `terraform-resources` with `--dry-run` show the changes in resources for all accounts;
* Running `terraform-resources` with `--dry-run`, `--account-name` and `--exclude-accounts` gives the user an error;
* Running `terraform-resources` with  `--exclude-accounts` and not `--dry-run` gives the user an error;
* RUnning `terraform-resources` excluding an account that doesn't exist gives the user an error; 
* Running `terraform-resources` excluding all accounts gives the user an error;
* Running `terraform-resouces` with `--exclude-account=a --exclude-account=b` show the changes in resources for all accounts except `a` and `b`.

This is testable by using `app-interface-dev-data`.

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -711,6 +711,18 @@ def populate_desired_state(
             )
 
 
+class ExcludeAccountsAndDryRunException(Exception):
+    pass
+
+
+class ExcludeAccountsAndAccountNameException(Exception):
+    pass
+
+
+class MultipleAccountNamesInDryRunException(Exception):
+    pass
+
+
 @defer
 def run(
     dry_run,
@@ -728,12 +740,12 @@ def run(
     if exclude_accounts and not dry_run:
         message = "--exclude-accounts is only supported in dry-run mode"
         logging.error(message)
-        raise RuntimeError(message)
+        raise ExcludeAccountsAndDryRunException(message)
 
     if exclude_accounts and account_name:
         message = "Using --exclude-accounts and --account-name at the same time is not allowed"
         logging.error(message)
-        raise RuntimeError(message)
+        raise ExcludeAccountsAndAccountNameException(message)
 
     # account_name is a tuple of account names for more detail go to
     # https://click.palletsprojects.com/en/8.1.x/options/#multiple-options
@@ -746,7 +758,7 @@ def run(
     if account_names and len(account_names) > 1 and not dry_run:
         message = "Running with multiple accounts is only supported in dry-run mode"
         logging.error(message)
-        raise RuntimeError(message)
+        raise MultipleAccountNamesInDryRunException(message)
 
     ri, oc_map, tf, resource_specs = setup(
         dry_run,
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -560,8 +560,11 @@ def setup(
     internal: str,
     use_jump_host: bool,
     account_names: Optional[Collection[str]],
+    exclude_accounts: Optional[Collection[str]],
 ) -> tuple[ResourceInventory, OC_Map, Terraform, ExternalResourceSpecInventory]:
     accounts = queries.get_aws_accounts(terraform_state=True)
+    if not account_names and exclude_accounts:
+        accounts = [ac for ac in accounts if ac["name"] not in exclude_accounts]
     if account_names:
         accounts = [n for n in accounts if n["name"] in account_names]
         if len(accounts) != len(account_names):
@@ -700,6 +703,7 @@ def run(
     light=False,
     vault_output_path="",
     account_name: Optional[Sequence[str]] = None,
+    exclude_accounts: Optional[Sequence[str]] = None,
     defer=None,
 ) -> None:
     # account_name is a tuple of account names for more detail go to
@@ -722,6 +726,7 @@ def run(
         internal,
         use_jump_host,
         account_names,
+        exclude_accounts,
     )
 
     if not dry_run:
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -587,26 +587,15 @@ def setup(
     accounts = queries.get_aws_accounts(terraform_state=True)
     if not include_accounts and exclude_accounts:
         excluding = filter_accounts_by_name(accounts, exclude_accounts)
-        if len(excluding) != len(exclude_accounts):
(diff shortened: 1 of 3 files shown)


reconcile/terraform_resources.py
```diff
@@ -549,17 +549,40 @@ def init_working_dirs(
...
+            raise ValueError(
+                f"Accounts {exclude_accounts} were provided as arguments, but not found in app-interface. Check your input for typos or for missing AWS account definitions."
+            )
+        accounts = exclude_accounts_by_name(accounts, exclude_accounts)
+        if len(accounts) == 0:
+            raise ValueError("You have excluded all aws accounts, verify your input")
+        account_names = tuple(ac["name"] for ac in accounts)
     if account_names:
-        accounts = [n for n in accounts if n["name"] in account_names]
+        accounts = filter_accounts_by_name(accounts, account_names)
```
hemslo, line 585: `account_names` is function parameter, should avoid mutate or reassign it, can rename it in parameter to `include_accounts` to make it consistent with `exclude_accounts`

Validate account exist by name logic is duplicated in both branches, can be extracted to a function.

So this can be like:

```python
def validate_account_names(
    accounts: Iterable[Mapping[str, Any]], names: Iterable[str]
):
    if len(accounts) != len(names):
        missing_names = set(names) - {a["name"] for a in accounts}
        raise ValueError(
            f"Accounts {missing_names} were provided as arguments, but not found in app-interface. Check your input for typos or for missing AWS account definitions."
        )

def setup(
    dry_run: bool,
    print_to_file: str,
    thread_pool_size: int,
    internal: str,
    use_jump_host: bool,
    include_accounts: Optional[Collection[str]],
    exclude_accounts: Optional[Collection[str]],
) -> tuple[ResourceInventory, OC_Map, Terraform, ExternalResourceSpecInventory]:
    if not include_accounts and exclude_accounts:
        excluding = filter_accounts_by_name(accounts, exclude_accounts)
        validate_account_names(excluding, exclude_accounts)
        accounts = exclude_accounts_by_name(accounts, exclude_accounts)
        if len(accounts) == 0:
            raise ValueError("You have excluded all aws accounts, verify your input")
    elif include_accounts:
        accounts = filter_accounts_by_name(accounts, include_accounts)
        validate_account_names(accounts, include_accounts)
    account_names = [a["name"] for a in accounts]
```
Can even extract the whole get and filter/exclude accounts logic to a function if this becomes too complex.

May need to update exception string in tests as it was using `tuple` in `ValueError` before.

esron: Thank you for the review.

esron: I've implemented your suggestions, thanks again.

reconcile/terraform_resources.py
```diff
@@ -696,8 +718,21 @@ def run(
...
     defer=None,
 ) -> None:
+    if exclude_accounts and not dry_run:
+        message = "Exclude accounts is only supported in dry-run mode"
+        logging.error(message)
+        raise RuntimeError(message)
+
+    if exclude_accounts and account_name:
+        message = (
+            "Using exclude account and account name at the same time is not allowed"
```
bhushanthakur93, line 731: nit: `--exclude-accounts` and `--account-name`

reconcile/terraform_resources.py
```diff
@@ -549,27 +549,49 @@ def init_working_dirs(
...
-            )
-            raise ValueError(
-                f"Accounts {acc_names} were provided as arguments, but not found in app-interface. Check your input for typos or for missing AWS account definitions."
-            )
+    if not include_accounts and exclude_accounts:
+        excluding = filter_accounts_by_name(accounts, exclude_accounts)
+        validate_account_names(excluding, exclude_accounts)
+        accounts = exclude_accounts_by_name(accounts, exclude_accounts)
+        if len(accounts) == 0:
+            raise ValueError("You have excluded all aws accounts, verify your input")
```
bhushanthakur93, line 589: You could also create domain specific exceptions here such as `NoAwsAccountSelectedException` or something like that. This will help in catching for that specific exception and also avoiding the need for string matching with generic exceptions like `ValueError` . You can evaluate the same for other exceptions thrown in this PR.

reconcile/test/test_terraform_resources.py
```diff
@@ -3,6 +3,51 @@
 import reconcile.terraform_resources as integ
 
 
+def test_cannot_use_exclude_accounts_if_not_dry_run():
```
bhushanthakur93, line 6: nice work on testing!

reconcile/terraform_resources.py
```diff
@@ -549,27 +549,49 @@ def init_working_dirs(
...
-            # Some of the passed account names don't exist in app-interface
-            acc_names = tuple(
-                a
-                for a in account_names
-                if a not in tuple(account["name"] for account in accounts)
-            )
-            raise ValueError(
-                f"Accounts {acc_names} were provided as arguments, but not found in app-interface. Check your input for typos or for missing AWS account definitions."
-            )
+    if not include_accounts and exclude_accounts:
```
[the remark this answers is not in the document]
esron: I'll investigate that.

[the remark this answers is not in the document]
esron: I'm failing to see how this things relate to each other.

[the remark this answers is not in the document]
esron: By reading the code, what I think is that Sharding in the runtime context is on higher level of abstraction than the Sharding in terraform-resources. Maybe moving this to there is going to make the code there too specific? 

Also, it seems like Sharding in runtime is more pointed to run without dry-run, while my changes here are specific for dry run mode.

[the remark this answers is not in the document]
OliviaHY: I apologize, this is more of the appropriate method: https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/runtime/runner.py#L172 
The main reason I mentioned this was because the new framework is trying take care of the state of the dry-run before the integration's run method gets called, which keep each integration from having to worry about it. 
I just saw Gerd's comment on your ticket, since he is the designer for this new separation of responsibility, I won't insist on this.

bhushanthakur93 (approved):
LGTM

PR #3176, opened 2023-02-08
APPSRE-6609 integrate oc with new OCMap

fishi0x01:
`OCDeprecated` and `OCNative` can now solely rely on the new `OCConnectionParameters` class. This makes `OC*` fully compliant with the new `OCMap`, which means `OCMap` now requires less args to make it backwards compatible.

This should be mostly compatible with @lechuk47 [incoming PR](https://github.com/app-sre/qontract-reconcile/pull/3142). Only adjustment should be necessary in the `__init__` method for `self.api_resources`. All class attributes keep the same name and functionality.

This is a follow-up of the ongoing `OCMap` refactor, see

- https://github.com/app-sre/qontract-reconcile/pull/3166
- https://github.com/app-sre/qontract-reconcile/pull/3154

Once this is merged, I will start shifting the integrations 1-by-1 to use the new `OCMap`.

**Test**

- Tested with new `OCMap` class (dashdotdb-cso integration)
- Tested with old `OC_Map` class (openshift-users integration)


reconcile/utils/oc.py
```diff
@@ -1016,16 +1134,24 @@ def __init__(
...
+
         if server:
             self.client = self._get_client(server, token)
             self.api_kind_version = self.get_api_resources()
             self.api_resources = self.api_kind_version.keys()
         else:
             raise Exception("A method relies on client/api_kind_version to be set")
 
-        self.object_clients = {}
+        self.object_clients: dict[Any, Any] = {}
```
esron, line 1154: Shouldn't this be a `Mapping[Any, Any]` or Mapping[str, Any]

fishi0x01: I think in this case it is ok because we instantiate a concrete object here. `Mapping`, `Iterable` and such are used in function arguments to keep them more flexible for consumers. (https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#abstract-collection-types)

esron (approved):

lechuk47 (approved):

BumbleFeng:
since we are here, could you also move jumphost_remote_port into `$schema: /openshift/jump-host-1.yml` ?

reconcile/utils/oc.py
```diff
@@ -319,6 +364,78 @@ def __init__(
...
+            oc_base_cmd.extend(["--insecure-skip-tls-verify"])
+        if self.server:
+            oc_base_cmd.extend(["--server", self.server])
+
+        token = connection_parameters.automation_token
+        if (
+            connection_parameters.is_cluster_admin
+            and connection_parameters.cluster_admin_automation_token
+        ):
+            token = connection_parameters.cluster_admin_automation_token
```
hemslo, line 387: Can use conditional assign to avoid mutation
```python
        token = connection_parameters.cluster_admin_automation_token if (
                connection_parameters.is_cluster_admin
                and connection_parameters.cluster_admin_automation_token
        ) else connection_parameters.automation_token
```

reconcile/utils/oc.py
```diff
@@ -1016,16 +1134,24 @@ def __init__(
             init_api_resources=False,
             local=local,
             insecure_skip_tls_verify=insecure_skip_tls_verify,
+            connection_parameters=connection_parameters,
         )
 
+        if connection_parameters:
+            token = connection_parameters.automation_token
+            if connection_parameters.is_cluster_admin:
+                token = connection_parameters.cluster_admin_automation_token
```
hemslo, line 1143: This logic is similar to https://github.com/app-sre/qontract-reconcile/pull/3176/files#diff-006493679cf46d27e100feeb8475e6e29987fe85cca4858a1f05608b2d536dc8R382-R387, but has different conditions, maybe extract to a method like `get_token` and reuse it?

fishi0x01: Yeah the conditions differ - I basically translated the original logic. I'd like to keep the logic as similar as possible, as I'm not 100% sure about all the use-cases or why the condition was implemented that way originally. Now the question is: which token you want to extract with `get_token()`? The admin token or the non-admin token?

fishi0x01:
@BumbleFeng I'm not fully in the loop about the `remotePort` situation. Would like to keep this out-of-scope for this PR. In order to keep this compatible with Jordi's PR, Im trying to change as less as possible here - I will have a look at further refactors once Jordi's PR is merged.

reconcile/utils/oc.py
fishi0x01: To be honest I'm not a big fan of multi-line conditional assignments. I find them harder to read than simple old-style if assignments, as they are common in other languages too - but yeah that is a taste question I guess :D
I could reduce the length of the parameter names - which would require additional tmp assignments again.

I guess the best would be like you suggest below to extract this into another method. Will keep it for now and address this in the follow-up PRs, when it is more clear to me when `automation_token` is used over `admin_token`.

fishi0x01:
@BumbleFeng the remotePort change will be handled in this PR: https://github.com/app-sre/qontract-reconcile/pull/3187

reconcile/utils/oc.py
hemslo: To simplify code for this instance, extract function is preferred over extract variable, as you mentioned tmp assign is not clean. Is it a real case that `is_cluster_admin` is `True` but `cluster_admin_automation_token` is `None`? If that case exists, then `automation_token` will be used even `is_cluster_admin`, which is different than the other logic in `OCNative`.

 `token` is only used from line 382 to line 390, can use assignment expression in `if` to make it clear, like this:


```python
def get_token(connection_parameters):
    if (
            connection_parameters.is_cluster_admin
            and connection_parameters.cluster_admin_automation_token
    ):
        return connection_parameters.cluster_admin_automation_token
    return connection_parameters.automation_token
```

```python
if token := get_token(connection_parameters):
    oc_base_cmd.extend(["--token", token])
```

if that null check for `cluster_admin_automation_token` is not needed, then it's even cleaner.

PR #3184, opened 2023-02-09
[skupper-network] internal clusters don't need a peering to reach each other

chassing:
Spawn skupper networks between sites on internal clusters because all internal clusters can reach each other without peerings.

[APPSRE-6997](https://issues.redhat.com/browse/APPSRE-6997)

### DIFF
--- a/reconcile/skupper_network/models.py
+++ b/reconcile/skupper_network/models.py
@@ -196,6 +196,11 @@ class SkupperSite(BaseModel):
         """Return True if the skupper site is not hosted on a private cluster."""
         return not self.on_private_cluster
 
+    @property
+    def on_internal_cluster(self) -> bool:
+        """Return True if the skupper site is hosted on an internal cluster."""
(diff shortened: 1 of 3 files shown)


fishi0x01 (approved):

reconcile/skupper_network/models.py
```diff
@@ -196,6 +196,11 @@ def on_public_cluster(self) -> bool:
         """Return True if the skupper site is not hosted on a private cluster."""
         return not self.on_private_cluster
 
+    @property
+    def on_internal_cluster(self) -> bool:
+        """Return True if the skupper site is hosted on an internal cluster."""
+        return self.cluster.internal if self.cluster.internal else False
```
patjlm, line 202: picky nit
```suggestion
        return self.cluster.internal or False
```

patjlm (approved):

rporres (approved):

PR #3185, opened 2023-02-09
[skupper-network] k8s labels have a 63 characters limit

chassing:
Shrink the connection-token `token-receiver` label to an upper limit of 63 characters.

[APPSRE-6993](https://issues.redhat.com/browse/APPSRE-6993)

### DIFF
--- a/reconcile/skupper_network/models.py
+++ b/reconcile/skupper_network/models.py
@@ -305,4 +305,8 @@ class SkupperSite(BaseModel):
     @property
     def token_labels(self) -> dict[str, str]:
         """Get the token labels."""
-        return {"token-receiver": self.name}
+        # This label is used to identify the site in the `skupper link status` command
+        # self.name ({skupper_network.identifier}-{ns.cluster.name}-{ns.name}) can be longer than 63 characters
+        # so use cluster.name-namespaced.name instead and trim it to 63 characters
+        # a namespace can't be in more than one skupper network, so it's safe to omit the skupper network identifier
+        return {"token-receiver": f"{self.cluster.name}-{self.namespace.name}"[0:63]}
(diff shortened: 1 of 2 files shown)


fishi0x01 (approved):

reconcile/skupper_network/models.py
```diff
@@ -288,4 +288,8 @@ def token_name(self, other: SkupperSite) -> str:
     @property
     def token_labels(self) -> dict[str, str]:
         """Get the token labels."""
-        return {"token-receiver": self.name}
+        # This label is used to identify the site in the `skupper link status` command
+        # self.name ({skupper_network.identifier}-{ns.cluster.name}-{ns.name}) can be longer than 63 characters
+        # so use cluster.name-namespaced.name instead and trim it to 63 characters
+        # a namespace can't be in more than one skupper network, so it's safe to omit the skupper network identifier
+        return {"token-receiver": f"{self.cluster.name}-{self.namespace.name}"[0:63]}
```
rporres, line 295: I know it's a bit far fetched, but we could have two namespaces that differ only in the last characters:

```
this-is-a-very-long-namespace-name-but-still-valid-under-63c-1
this-is-a-very-long-namespace-name-but-still-valid-under-63c-2
```

chassing: Yes, this can happen but:
1. I assume we don't use skupper to reach another namespace on the same cluster
2. The `token-receiver` is prefixed with the cluster name and stripped at the end

So the worst would be a `skupper link status` output like:
```
cluster-1-this-is-a-very-long-namespace-name-but-still-valid-un
cluster-2-this-is-a-very-long-namespace-name-but-still-valid-un
```

rporres: fair enough. The other solution is to md5 the namespace, but that will look awful and it really doesn't feel as needed.

rporres (approved):

PR #3188, opened 2023-02-13
APPSRE-6609 cluster-deployment-mapper uses OCMap

fishi0x01:
Integration `cluster-deployment-mapper` should use the new typed `OCMap` class.
This is part of the `OC_Map` refactor attempt. See https://github.com/app-sre/qontract-reconcile/pull/3176

This also adds convenience functions to init an `OCMap` object. Goal of the convenience function is to hide the underlying `OCConnectionParameter` class.


chassing (approved):
Nice work ❤️

PR #3189, opened 2023-02-13
extract runtime environment config

geoberle:
# What
extract the configuration and setup of the qontract-reconcile environment to a dedicated module. the `init_env` function sets up logging, initializes the config and sets up GQL connectivity to the `qontract-server`.

`init_env` can be called with or without parameters about loglevel and config. if no parameters are passed, their values are extracted from env variables.

if it is called with parameters, their values are stored in env variables. this way, if a child process is spawned, it can call `init_env` without parameters and a log + config + GQL environment compatible with the parent process is established.

the intended way to use `init_env` is with parameter from the main entry point of q-r (cli.py). this sets up env vars to allow child processes to use `init_var` without parameters to replicate the same environment.

# Why
the current way to set up the q-r env was implemented multiple times. this change unifies the process.

this is prepwork to optionally run integration shards during dry-runs with `multiprocessing.Process` instead of threading. this makes sense for certain integrations that are CPU bound and not so much I/O bound because python threading is not really allowing CPU bound tasks to benefit a lot.

### DIFF
--- a/dockerfiles/hack/run-integration.py
+++ b/dockerfiles/hack/run-integration.py
@@ -10,15 +10,15 @@ from typing import Optional
 import click
 from prometheus_client import start_http_server
 
-from reconcile.cli import (
-    LOG_DATEFMT,
-    LOG_FMT,
-)
 from reconcile.status import ExitCodes
 from reconcile.utils.metrics import (
     execution_counter,
     run_status,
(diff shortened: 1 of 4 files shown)


fishi0x01 (approved):
LGTM :+1:

PR #3190, opened 2023-02-13
APPSRE-6609 reduce generated classes

fishi0x01:
We can greatly reduce the number of generated classes on namespace queries by introducing a fragment for `ResourceValues_v1`.


geoberle (approved):
very nice!!

chassing (approved):

PR #3191, opened 2023-02-13
[WIP][integration-runtime] parameter class for class based integrations

geoberle:
# What
replace the generic `*args: Any, **kwargs: Any` parameterlist for class based integrations. to allow each integration to have their parameters passed in as verifiable as possible and typesafe as possible, each integration can define the dedicated class for the params wrapper as a generic type, e.g.

```python
class SimpleTestIntegrationParams(PydanticRunParams):
    int_arg: int
    opt_str_arg: Optional[str] = None

class SimpleTestIntegration(QontractReconcileIntegration[SimpleTestIntegrationParams]):
    ...

    def run(self, dry_run: bool, params: SimpleTestIntegrationParams) -> None:
        params.int_arg
	params.opt_str_arg
```

in cli.py, the integration is hooked up like this:

```python
@integration.command()
@click.pass_context
def simple_integration(ctx, int_param: int):
    from reconcile import simple_integration

    run_class_integration(
        integration=simple_integration.SimpleIntegration(),
        params=SimpleTestIntegrationParams(int_arg=int_param),
        ctx=ctx.obj,
    )
```

in this example, a pydantic based `RunParams` class is used to wrap the parameters. this way the entire parameter validation process can be covered by having type hints in place. pydantic drives the validation and allows for additional validation functions within the class itself.

for the module based integrations, the generic wrapper class `ArgsKwargsRunParams` is available, that makes sure the custom `run` functions in the modules do not need to change. all the wrapping happens in `reconcile.cli`.

# Why

The very generic `*args, **kwargs` does not play well with subclassing the class based integrations. Mypy complaints if we would go more specific.

# Alternatives considered

Use the init function of the class to pass accept parameters and keep the run function free of params except for maybe the `dry_run`. I will open another PR so both approaches can be compared.

https://issues.redhat.com/browse/APPSRE-7023

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -8,7 +8,7 @@ from signal import SIGUSR1
 from types import ModuleType
 from typing import (
     Optional,
-    Union,
+    TypeVar,
 )
 
 import click
@@ -31,8 +31,10 @@ from reconcile.utils.environ import environ
 from reconcile.utils.exceptions import PrintToFileInGitRepositoryError
 from reconcile.utils.git import is_file_in_git_repo
 from reconcile.utils.runtime.integration import (
+    ArgsKwargsRunParams,
     ModuleBasedQontractReconcileIntegration,
     QontractReconcileIntegration,
+    RunParams,
 )
 from reconcile.utils.runtime.meta import IntegrationMeta
 from reconcile.utils.runtime.runner import (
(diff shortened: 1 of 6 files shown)


geoberle:
closing this for now in favour of https://github.com/app-sre/qontract-reconcile/pull/3198

PR #3194, opened 2023-02-14
Add integration terraform_cloudflare_users to manage cloudflare user access

bhushanthakur93:
Add new integration `terraform-cloudflare-users` for managing user management for cloudflare. 

Ticket: APPSRE-6865

Design Doc: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/terraform-cloudflare-users-support.md

Schema dependency: https://github.com/app-sre/qontract-schemas/pull/379

Validation:

1. Assign a role to a user => cloudflare invite is sent and upon accepting the invite, the role is assigned to the user.

1. Remove a role from a user => If invite is sent but not yet accepted, invitation is cancelled. Otherwise if user is already in the account, then the user is removed.

1.  Repeat step 1 and 2 with multiple roles with different account combination and verify the result. 

The unit test also covers fair amount of possibilities for roles and account combination.


reconcile/gql_definitions/terraform_cloudflare_users/terraform_cloudflare_roles.gql
```diff
@@ -0,0 +1,40 @@
...
+        cloudflare_user
+        org_username
+      }
+    }
+    account {
+      name
+      providerVersion
+      apiCredentials {
+        path
+        field
```
fishi0x01, line 18: You can use a fragment here to reduce the amount of generated classes:

```
... VaultSecret
```

reconcile/gql_definitions/terraform_cloudflare_users/terraform_cloudflare_roles.gql
```diff
@@ -0,0 +1,40 @@
...
+      }
+      terraformStateAccount {
+        name
+        automationToken {
+          path
+          field
```
fishi0x01, line 24: Same as above

```
... VaultSecret
```

reconcile/queries.py
```diff
@@ -106,6 +106,7 @@ def get_secret_reader_settings() -> Optional[Mapping[str, Any]]:
       readTimeout
       connectTimeout
     }
+    cloudflareEmailDomainAllowList
```
fishi0x01, line 109: We recently adjusted the dev-guidelines to unify our approach for gql queries: https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#generated-graphql-data-classes

I think in this case you do not need the `queries` module any longer. You could add a new query for settings, like done for the smtp use-case: https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/gql_definitions/common/smtp_client_settings.gql

You could also add the vault-settings there, as you will require them for the secret reader too.

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def run(self, dry_run: bool, *args: Any, **kwargs: Any) -> None:
+        print_to_file = args[0]
+        account_name = args[1]
+        thread_pool_size = args[2]
+        enable_deletion = args[3]
+
+        cloudflare_roles, settings = self._get_desired_state()
+
+        secret_reader = SecretReader(settings=settings)
```
fishi0x01, line 95: We are trying to deprecate `SecretReader`. A type-safe alternative:

```python
from reconcile.utils.secret_reader import create_secret_reader
from reconcile.typed_queries.app_interface_vault_settings import (
    get_app_interface_vault_settings,
)


vault_settings = get_app_interface_vault_settings()
secret_reader = create_secret_reader(use_vault=vault_settings.vault)
```

You can also create a dedicated settings query for your integration that includes `cloudflareEmailDomainAllowList`, so you only need a single query instead of 2 different ones for the settings.

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,298 @@
...
+)
+from reconcile.gql_definitions.terraform_cloudflare_users.terraform_cloudflare_roles import (
+    CloudflareAccountRoleQueryData,
+    CloudflareAccountRoleV1,
+)
+from reconcile.status import ExitCodes
+from reconcile.utils import gql
+from reconcile.utils.external_resource_spec import ExternalResourceSpec
+from reconcile.utils.runtime.integration import QontractReconcileIntegration
+from reconcile.utils.secret_reader import SecretReader
```
fishi0x01, line 25: Lets use `SecretReaderBase`. `SecretReader` is marked for deprecation as it is untyped. For backwards-compat, `SecretReader` implements `SecretReaderBase`.

reconcile/utils/terrascript/cloudflare_client.py
```diff
@@ -146,6 +161,104 @@ def dumps(self) -> str:
...
+
+
+class TerrascriptCloudflareClientFactory:
+    @staticmethod
+    def _create_backend_config(
+        tf_state_s3: TerraformStateS3, key: str, secret_reader: SecretReader
+    ) -> TerraformS3BackendConfig:
+        aws_acct_creds = secret_reader.read_all(
+            {"path": tf_state_s3.automation_token_path}
+        )
```
fishi0x01, line 198: With the typed secret reader mentioned above (via `create_secret_reader()`) you can directly pass in a secret object and dont need to construct dictionaries any longer.

```python
aws_acct_creds = secret_reader.read_all_secret(tf_state_s3.automation_token)
```

This requires however, that the secret object is passed around instead of a path.

fishi0x01 (commented):
Just adding some comments around `SecretReader` and `queries` module. Didn't fully grasp yet the overall integration in detail - I need to read more into it :nerd_face:

geoberle (commented):
thx for this work! i highlighted only a tiny nit

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,298 @@
...
+        )
+        specs = build_external_resource_spec_from_cloudflare_users(users)
+
+        cf_clients.add_specs(specs)
+        cf_clients.populate_resources()
+
+        working_dirs = cf_clients.dump(print_to_file=print_to_file)
+
+        if print_to_file:
+            sys.exit(ExitCodes.SUCCESS)
```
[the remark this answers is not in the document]
bhushanthakur93: Yes! I'll be integrating this feedback shortly.

bhushanthakur93:
> Just adding some comments around `SecretReader` and `queries` module. Didn't fully grasp yet the overall integration in detail - I need to read more into it nerd_face

Thank you for the suggestions! I have incorporated the feedback.

reconcile/gql_definitions/terraform_cloudflare_users/terraform_cloudflare_roles.gql
```diff
@@ -0,0 +1,38 @@
+# qenerate: plugin=pydantic_v1
+
+query CloudflareAccountRole {
+  cloudflare_account_roles:	cloudflare_account_role_v1 {
+    name
+    roles
+    access_roles {
+      users {
+        cloudflare_user
+        org_username
```
fishi0x01, line 10: nit: Would it make sense to extend our `User` fragment with `cloudflare_user` and use it here?

```graphql
... User
```

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,339 @@
+from dataclasses import dataclass
+from typing import (
+    Any,
+    Iterable,
+    Mapping,
+    MutableMapping,
```
fishi0x01, line 6: We should from now on use `from collections.abc import Iterable, ...` for these for [PEP-585](https://peps.python.org/pep-0585/) compliance. Introduced by @rporres

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,339 @@
...
+    Optional,
+    Set,
+    Tuple,
```
fishi0x01, line 9: nit: For these we can use the built-ins `set` and `tuple`.

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,339 @@
...
+        account_name = self.params.account_name
+        thread_pool_size = self.params.thread_pool_size
+        enable_deletion = self.params.enable_deletion
+
+        cloudflare_roles, settings = self._get_desired_state()
+
+        secret_reader = create_secret_reader(
+            use_vault=settings.settings[0].vault
+            if settings.settings is not None
+            else True
```
fishi0x01, line 119: Not sure if we should raise an exception if settings are not found. That is currently the default behavior in other integrations, see https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/typed_queries/app_interface_vault_settings.py#L23-L30

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,339 @@
...
+        cf_clients = self._build_cloudflare_terraform_config_client_collection(
+            cloudflare_roles, secret_reader, account_name
+        )
+
+        users = get_cloudflare_users(
+            cloudflare_roles.cloudflare_account_roles,
+            account_name,
+            settings.settings[0].cloudflare_email_domain_allow_list
+            if settings.settings is not None
+            else None,
```
fishi0x01, line 131: This could be reduced if the above mentioned exception is raised.

reconcile/utils/terraform_client.py
```diff
@@ -14,6 +14,7 @@
 from threading import Lock
 from typing import (
     Any,
+    Callable,
```
fishi0x01, line 17: Same as above: `from collections.abc import Callable`

reconcile/utils/terrascript/cloudflare_resources.py
```diff
@@ -196,3 +219,44 @@ def populate(self) -> list[Union[Resource, Output]]:
...
+# class CloudflareAccountsTerrascriptData:
+#
+#     def __init__(self, spec: ExternalResourceSpec):
+#         self._spec = spec
+#
+#     def populate(self) -> list[Data]:
+#         values = ResourceValueResolver(self._spec).resolve()
+#
+#         cf_accounts = cloudflare_accounts(self._spec.identifier, **values)
+#         return [cf_accounts]
```
fishi0x01, line 233: Is this intended to be commented out?

reconcile/utils/terrascript/models.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@dataclass
+class Integration:
+    name: str
+    key: str
+
+
+@dataclass
+class VaultSecret:
```
fishi0x01, line 12: Should this be a `Protocol` to be interchangeable with `VaultSecret` fragment from queries?

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,339 @@
...
+            if account_name and role.account.name != account_name:
+                continue
+            cf_account = CloudflareAccount(
+                role.account.name,
+                VaultSecret(
+                    role.account.api_credentials.path,
+                    role.account.api_credentials.field,
+                    role.account.api_credentials.version,
+                    role.account.api_credentials.q_format,
+                ),
```
fishi0x01, line 182: If `VaultSecret` implements `Protocol`. then you do not need this conversion. I.e., 

```python
class VaultSecret(Protocol):
  path: str
  ......
```

See https://service.pages.redhat.com/dev-guidelines/docs/appsre/generated-query-classes/#how-to-use-generated-classes

By using `Protocol`, you still isolate your business logic model from the underlying query model, but you can still use the classes interchangably as long as they fulfill the specs defined in the Protocol class.

fishi0x01 (commented):
Some more smaller comments about `Protocol` usage and [PEP-585](https://peps.python.org/pep-0585/)

Apart from that it looks great. Good job!

reconcile/terraform_cloudflare_users.py
fishi0x01: Alternatively, you could use `reconcile.utils.secret_reader.HasSecret` -> it is already used by several integrations and complies to our convention of separating generated classes from utils libraries.

reconcile/utils/terrascript/models.py
fishi0x01: Alternatively, you could use `reconcile.utils.secret_reader.HasSecret` -> it is already used by several integrations and complies to our convention of separating generated classes from utils libraries.

steveteahan (commented):
Working through this a bit at a time. Providing some of the initial comments for now.

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,339 @@
...
+                pass
+
+        return cf_clients
+
+
+def get_cloudflare_users(
+    cloudflare_roles: Optional[Iterable[CloudflareAccountRoleV1]],
+    account_name: Optional[str],
+    email_domain_allow_list: Optional[Iterable[str]],
+) -> Mapping[str, Mapping[str, CloudflareUser]]:
```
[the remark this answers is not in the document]
bhushanthakur93: :+1:

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,339 @@
...
+        return QONTRACT_INTEGRATION.replace("_", "-")
+
+    def get_early_exit_desired_state(
+        self, *args: Any, **kwargs: Any
+    ) -> Optional[dict[str, Any]]:
+
+        cloudflare_roles, settings = self._get_desired_state()
+
+        return {
+            "cloudflare_roles": cloudflare_roles.dict(),
```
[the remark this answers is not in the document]
bhushanthakur93: yes, redundant but easier for readability (IMO) vs having iterating through each role and converting to dict.

reconcile/terraform_cloudflare_users.py
bhushanthakur93: using `HasSecret` now.

reconcile/utils/terrascript/models.py
bhushanthakur93: using `HasSecret` now. Thanks for pointing that out to me.

reconcile/utils/terrascript/cloudflare_resources.py
bhushanthakur93: removed.

reconcile/utils/terraform_client.py
bhushanthakur93: done.

reconcile/terraform_cloudflare_users.py
bhushanthakur93: Yeah, good point. I missed this earlier. I'm raising the exception now.

reconcile/terraform_cloudflare_users.py
bhushanthakur93: done. ty!

reconcile/terraform_cloudflare_users.py
bhushanthakur93: done. ty!

reconcile/gql_definitions/terraform_cloudflare_users/terraform_cloudflare_roles.gql
bhushanthakur93: I chose to keep it as is for now since this integration is not utilizing other fields in `User` fragment.

reconcile/utils/terraform_client.py
fishi0x01: I think the change wasn't pushed. This is still imported `from typing`

bhushanthakur93: oops! fixed this now.

fishi0x01 (approved):
LGTM :+1: 

Probably a good idea though to have someone more affiliated with cloudflare also have a look :)

README.md
```diff
@@ -171,6 +171,7 @@ Additional tools that use the libraries created by the reconciliations are also
   terraform-users                 Manage AWS users using Terraform.
   terraform-vpc-peerings          Manage VPC peerings between OSD clusters and
                                   AWS accounts or other OSD clusters.
+  terraform-cloudflare-users      Manage user access to cloudflare accounts
```
kwilczynski, line 174:
```suggestion
  terraform-cloudflare-users      Manage user access to Cloudflare accounts
```

reconcile/gql_definitions/terraform_cloudflare_users/terraform_cloudflare_roles.py
```diff
@@ -0,0 +1,168 @@
...
+    )
+
+    class Config:
+        smart_union = True
+        extra = Extra.forbid
+
+
+def query(query_func: Callable, **kwargs: Any) -> CloudflareAccountRoleQueryData:
+    """
+    This is a convenience function which queries and parses the data into
```
kwilczynski, line 154:
```suggestion
    This is a convenience function that queries and parses the data into
```

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,327 @@
...
+                role.account.enforce_twofactor,
+                role.account.q_type,
+                role.account.provider_version,
+            )
+
+            tf_state = role.account.terraform_state_account.terraform_state
+            if not tf_state:
+                raise ValueError(
+                    f"AWS account {role.account.terraform_state_account.name} cannot be used for Cloudflare "
+                    f"account {cf_account.name} because it does define a terraform state "
```
kwilczynski, line 185:
```suggestion
                    f"account {cf_account.name} because it does define a Terraform state "
```

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,327 @@
...
+
+            integration = None
+            for i in integrations:
+                if i.integration.replace("-", "_") == QONTRACT_INTEGRATION:
+                    integration = i
+                    break
+
+            if not integration:
+                raise IntegrationUndefined(
+                    "Must declare integration name under terraform state in app-interface"
```
kwilczynski, line 205:
```suggestion
                    "Must declare integration name under Terraform state in app-interface"
```

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,327 @@
...
+def get_cloudflare_users(
+    cloudflare_roles: Optional[Iterable[CloudflareAccountRoleV1]],
+    account_name: Optional[str],
+    email_domain_allow_list: Optional[Iterable[str]],
+) -> dict[str, dict[str, CloudflareUser]]:
+    """
+    Returns a two-level dictionary of users with 1st level keys mapping to cloudflare account names
+    and 2nd level keys mapping to user's email address.
+    The method also takes into consideration :param account_name: and :param email_domain_allow_list: which can be
+    used to filter users not matching these parameters
```
kwilczynski, line 240:
```suggestion
    Returns a two-level dictionary of users with 1st level keys mapping to Cloudflare account names
    and 2nd level keys mapping to the user's email address.
    The method also takes into consideration :param account_name: and :param email_domain_allow_list: which can be
    used to filter users not matching these parameters
```

reconcile/utils/terraform_client.py
```diff
@@ -683,3 +685,57 @@ def _can_skip_rds_modifications(
...
+        working_dirs,
+        thread_pool_size,
+    )
+
+    defer(tf.cleanup)
+
+    disabled_deletions_detected, err = tf.plan(enable_deletion)
+    if err:
+        raise TerraformPlanFailed(
+            f"Failed to run terraform plan for integration {QONTRACT_INTEGRATION}"
```
kwilczynski, line 716:
```suggestion
            f"Failed to run Terraform plan for integration {QONTRACT_INTEGRATION}"
```

reconcile/utils/terraform_client.py
```diff
@@ -683,3 +685,57 @@ def _can_skip_rds_modifications(
...
+        logging.warning("Deletions detected but they are disabled")
+        raise TerraformDeletionDetected("Deletions detected but they are disabled")
+
+    if dry_run:
+        return
+
+    err = tf.apply()
+    if err:
+        TerraformApplyFailed(
+            f"Failed to run terraform apply for integration {QONTRACT_INTEGRATION}"
```
kwilczynski, line 728:
```suggestion
            f"Failed to run Terraform apply for integration {QONTRACT_INTEGRATION}"
```

reconcile/gql_definitions/terraform_cloudflare_users/terraform_cloudflare_roles.py
fishi0x01: That is from the generated `qenerate` classes. This must be fixed in `qenerate` :sweat_smile:

fishi0x01: @bhushanthakur93 if you apply this change, then the pipeline will fail, because the generated classes will be re-created which results in a git diff

kwilczynski: @fishi0x01, got it. Sorry for the commotion!

bhushanthakur93: oh man! I didn't notice this :man_facepalming:

reconcile/gql_definitions/terraform_cloudflare_users/terraform_cloudflare_roles.py
```diff
@@ -0,0 +1,168 @@
...
+    )
+
+    class Config:
+        smart_union = True
+        extra = Extra.forbid
+
+
+def query(query_func: Callable, **kwargs: Any) -> CloudflareAccountRoleQueryData:
+    """
+    This is a convenience function that queries and parses the data into
```
bhushanthakur93, line 154:
```suggestion
    This is a convenience function which queries and parses the data into
```

reconcile/utils/terrascript/cloudflare_resources.py
```diff
@@ -63,6 +84,8 @@ def create_cloudflare_terrascript_resource(
         return CloudflareWorkerScriptTerrascriptResource(spec).populate()
     elif resource_type == "zone":
         return CloudflareZoneTerrascriptResource(spec).populate()
+    elif resource_type == "cloudflare_account_member":
```
steveteahan, line 87: Any reason this isn't just `account_member` to stick to the convention?

reconcile/utils/terrascript/cloudflare_resources.py
```diff
@@ -196,3 +219,32 @@ def populate(self) -> list[Union[Resource, Output]]:
         resources.extend(self._create_cloudflare_certificate_pack(zone, zone_certs))
 
         return resources
+
+
+class CloudflareAccountRolesTerrascriptData:
```
steveteahan, line 224: Is this used somewhere or was it moved into `CloudflareAccountMemberTerrascriptResource`?

reconcile/utils/terraform_client.py
```diff
@@ -683,3 +685,57 @@ def _can_skip_rds_modifications(
             return not set(changed_resource_arguments) - set(changed_values)
         else:
             return False
+
+
+@defer
+def run_terraform(
```
steveteahan, line 691: I like the idea, but I have some doubts whether it's general enough and/or if we should start with putting it into `terraform_client.py` just yet. Maybe this could move into the integration for now and then we can try to see how reusable it is in the other Cloudflare integrations in a follow-up MR? There will be significant changes to `terraform-cloudflare-resources` as part of supporting the write to Vault feature.

steveteahan (commented):
A few more comments, but overall looks good!

reconcile/terraform_cloudflare_users.py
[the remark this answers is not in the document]
bhushanthakur93: We also have another key in the dictionary that is setting email domain allow-list. So I'm keeping this as is :sweat_smile:

reconcile/utils/terrascript/cloudflare_resources.py
bhushanthakur93: done

reconcile/utils/terrascript/cloudflare_resources.py
bhushanthakur93: removed

reconcile/utils/terraform_client.py
bhushanthakur93: In light of significant changes in near future, I've moved this to the integration itself.

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,327 @@
...
+            }
+
+            cloudflare_account_member = {
+                "provider": "cloudflare_account_member",
+                "identifier": safe_resource_id(cf_user.org_username),
+                "email_address": cf_user.email_address,
+                "account_id": "${var.account_id}",
+                "role_ids": [
+                    # I know this is ugly :(
+                    f'%{{ for role in data.cloudflare_account_roles.{safe_resource_id(cf_user.account_name)}.roles ~}}  %{{if role.name == "{each}" ~}}${{role.id}}%{{ endif ~}}  %{{ endfor ~}}'
```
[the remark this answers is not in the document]
bhushanthakur93: Added a comment to explain this better. Hope this helps! :crossed_fingers: 

Regarding testing, I have `test_build_external_resource_spec_from_cloudflare_users` in `test_terraform_cloudflare_users.py` which should exercise the logic of populating `role_ids` with multiple string template for each role. 

This however doesn't exercise if the string template actually works or not, because it is something evaluated by terraform and outside of the unit test boundary. This is where manual testing is currently filling the gap for now. This would be the ideal candidate for integration tests.

reconcile/terraform_cloudflare_users.py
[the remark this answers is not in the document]
bhushanthakur93: fixed

steveteahan (approved):

OliviaHY (approved):

reconcile/terraform_cloudflare_users.py
```diff
@@ -0,0 +1,380 @@
...
+
+            if dry_run:
+                return
+
+            err = tf.apply()
+            if err:
+                TerraformApplyFailed(
+                    f"Failed to run terraform apply for integration {QONTRACT_INTEGRATION}"
+                )
+        finally:
```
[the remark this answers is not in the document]
bhushanthakur93: I saw somewhere in the teamchat that sys.exit was used because we needed some way to propagate success/error back to process running on jenkins. Exceptions or lack of exceptions also covers this.

PR #3196, opened 2023-02-15
[APPSRE-7130] [FedRAMP] add insights-auth-client for rosa-authenticator

bkez322:
At the request of Insights; Adds an insights-auth-client to the rosa-authenticator integration.


reconcile/utils/terrascript_aws_client.py
```diff
@@ -5321,16 +5321,27 @@ def populate_tf_resource_rosa_authenticator(self, spec):
...
         )
-        tf_resources.append(cognito_user_pool_client)
+        tf_resources.append(ocm_cognito_user_pool_client)
+
+        # INSIGHTS POOL CLIENT
+        insights_cognito_user_pool_client = aws_cognito_user_pool_client(
+            "ins_userpool_client",
+            name=f"insights-{identifier}-pool-client",
+            user_pool_id=f"${{{cognito_user_pool_resource.id}}}",
+            callback_urls=[f"{bucket_url}/token.html"],
```
coderbydesign, line 5340: We'll have two `callback_urls` for Insights here, per environment. These will likely need to be parameterized different from the `bucket_url`, as our redirect uri won't be an s3 bucket asset directly, but rather our domain, and API Gateway will handle the rest.

We can supply these to you, but I'm assuming there will need to be some changes to how they're consumed?

cc: @lphiri

reconcile/utils/terrascript_aws_client.py
```diff
@@ -5687,22 +5698,39 @@ def populate_tf_resource_rosa_authenticator(self, spec):
...
-                f"${{{cognito_user_pool_client.id}}}\u0026response_type=code"
+                f"${{{ocm_cognito_user_pool_client.id}}}\u0026response_type=code"
+                f"\u0026scope=openid+gateway/AccessToken\u0026redirect_uri={bucket_url}/"
+                "token.html'",
+            },
+            depends_on=["aws_api_gateway_integration.gw_integration_auth"],
+        )
+        tf_resources.append(ocm_api_gateway_integration_auth_response_resource)
+
+        # AUTH - INSIGHTS
```
coderbydesign, line 5718: We shouldn't need this here, as it will be something we'll setup in our account. Correct me if I'm wrong, @lphiri.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -5321,16 +5321,27 @@ def populate_tf_resource_rosa_authenticator(self, spec):
...
-        tf_resources.append(cognito_user_pool_client)
+        tf_resources.append(ocm_cognito_user_pool_client)
+
+        # INSIGHTS POOL CLIENT
+        insights_cognito_user_pool_client = aws_cognito_user_pool_client(
+            "ins_userpool_client",
+            name=f"insights-{identifier}-pool-client",
+            user_pool_id=f"${{{cognito_user_pool_resource.id}}}",
+            callback_urls=[f"{bucket_url}/token.html"],
+            depends_on=["aws_cognito_resource_server.userpool_gateway_resource_server"],
```
coderbydesign, line 5341: @lphiri do we need this dependency here?

lphiri: We do not.  Our API will not have any customer resources or custom roles.

reconcile/utils/terrascript_aws_client.py
lphiri: Correct,  we will have our own instance of API Gateway.

reconcile/utils/terrascript_aws_client.py
bkez322: >I'm assuming there will need to be some changes to how they're consumed?

Probably. From what you're describing, there's probably going to have to be a new "rosa-authenticator" input called "insights_callback_urls".

In production, for example, that new property would be defined here: https://gitlab.redhat-osd.local/app-sre/app-interface/-/blob/master/data/services/ocm/namespaces/uhc-production.yml#L182-193

Do you know what those URLs would be stage and prod for you guys?

reconcile/utils/terrascript_aws_client.py
bkez322: Oh, okay. I think I maybe misunderstood what you folks were looking for then. I thought you were looking for a new user-pool-client associated with the ocm-api-gateway.

But if you actually just need a new user-pool-client associated with your insights-api-gateway, would it probably be most appropriate to just define the terraform HCL for the new user-pool-client alongside the terraform HCL that's already defining your insights-api-gateway?

reconcile/utils/terrascript_aws_client.py
bkez322: What I posted here is probably null and void after having read: https://github.com/app-sre/qontract-reconcile/pull/3196/files#r1107992992

coderbydesign (approved):

PR #3198, opened 2023-02-15
[integration-runtime] replace integration args, kwargs run parameters with class

geoberle:
### What
replace the generic `*args: Any, **kwargs: Any` parameterlist on the `run` function of class based integrations with a dedicated parameter class on an integrations `__init__` function. wrapping parameters in a dataclass enables validations at runtime (e.g. pydantic) but also at static code analysis (e.g. mypy).

### How
to leverage type safety all the way, the type of the parameter class can be defined as type hint when inheriting from `QontractReconcileIntegration`. the `__init__` function will then accept that type and the `param` property of the class will have that as a verifable type as well.

```python
class SimpleTestIntegrationParams(PydanticRunParams):
    int_arg: int
    opt_str_arg: Optional[str] = None

class SimpleTestIntegration(QontractReconcileIntegration[SimpleTestIntegrationParams]):
    def __init__(self, params: SimpleTestIntegrationParams):
        super().__init__(params)

    ...

    def run(self, dry_run: bool) -> None:
        print(self.params.int_arg)
```

in this example, the `print(self.params.int_arg)` statement is completely verifiable with mypy and also VSCode yields proper autocompletion.

in cli.py, the integration is hooked up like this:

```python
@integration.command()
@click.pass_context
def simple_integration(ctx, int_param: int):
    from reconcile import simple_integration

    run_class_integration(
        integration=simple_integration.SimpleIntegration(SimpleTestIntegrationParams(int_arg=int_param)),
        ctx=ctx.obj,
    )
```

Wrapping all parameters in a single object makes it easier to pass them around. This comes in handy when sharded integration runs are spawned off into threads or processes.

for the module based integrations, the generic wrapper class `ModuleArgsKwargsRunParams` is available, that makes sure the custom `run` functions in the modules do not need to change. all the wrapping happens in `reconcile.cli` so module integrations do not need to adapt because of this change.

### Alternatives considered

Another option would have been to make the integration a dataclass and defining all parameters as properties. That would make it harder to distinguish between run parameters passed into the integration and potential object properties that are just there to hold some values for the integration run.

Yet another approach was experimented with in https://github.com/app-sre/qontract-reconcile/pull/3191 where the parameter class was passed to the run function instead to the init function.

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -6,10 +6,7 @@ import re
 import sys
 from signal import SIGUSR1
 from types import ModuleType
-from typing import (
-    Optional,
-    Union,
-)
+from typing import Optional
 
 import click
 import sentry_sdk
@@ -31,6 +28,7 @@ from reconcile.utils.environ import environ
 from reconcile.utils.exceptions import PrintToFileInGitRepositoryError
 from reconcile.utils.git import is_file_in_git_repo
 from reconcile.utils.runtime.integration import (
+    ModuleArgsKwargsRunParams,
     ModuleBasedQontractReconcileIntegration,
     QontractReconcileIntegration,
 )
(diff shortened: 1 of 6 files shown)


chassing (approved):
This is awesome ❤️

bhushanthakur93:
LGTM!

PR #3200, opened 2023-02-15
Define Cluster-Scoped resources in app-interface (2)

lechuk47:
## Overview

This PR introduces a validation to ensure that cluster-scoped resources managed by App-Interface are not duplicated. 

With our current approach, multiple namespace files can define the same cluster-scoped resource using `openshiftResources`. That has happened sometimes and ends up with a clash of the reconciliation loop trying to reconcile the same resource from multiple locations. 

This change forces the cluster-scoped defined in `openshiftResoures` to be defined under `ManagedResourceNames` within the service namespace files. This allows a validation at PR-check to ensure that newly introduced cluster-scoped resources are not duplicated in the same cluster.

There are two checks: 
* CheckClusterScopedResourceNames: Ensures that all cluster-scoped resources in the desired state of a namespace are defined in the `ManagedResourceNames` list.
* CheckClusterScopedResourceDuplicates: Ensures that no cluster-scoped resource duplications exist in a cluster

**Note**: Before introducing this, all existent cluster-scoped resources should be defined in their namespace files

APPSRE-2935

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -1066,3 +1066,53 @@ def user_has_cluster_access(
     """Check user has access to cluster."""
     userkeys = determine_user_keys_for_access(cluster.name, cluster.auth)
     return any((getattr(user, userkey) in cluster_users for userkey in userkeys))
+
+
+def get_namespace_type_overrides(namespace: Mapping) -> dict[str, str]:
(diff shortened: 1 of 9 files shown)


fishi0x01 (approved):
Thx for the rebase! Sorry for all the hustle :bow:

PR #3202, opened 2023-02-15
upgrade sretoolbox to 2.3.0 and adapt exception handling for tasks

geoberle:
sretoolbox 2.3.0 is not hiding `sys.exit` exceptions anymore from the caller of `threaded.run(return_exceptions=True)`. therefore we need to adapt our interpretation of exception results.

e.g. in the case of running integration shards threaded, a `sys.exit(0)` is still fine and not an error.

additionally, the process based parallelisation that is part of this sretoolbox release is prepwork for https://issues.redhat.com/browse/APPSRE-7040


geoberle:
@rporres follow-up PR for the sretoolbox change in exception handling

rporres (approved):

esron (approved):

PR #3207, opened 2023-02-16
APPSRE-7067 containerize pypi publish

fishi0x01:
`build_tag.sh` is executed on builder node. We want to have it containerized to isolate build environment.


chassing (approved):
LGTM

PR #3209, opened 2023-02-16
APPSRE-7067 cleanup

fishi0x01:


rporres (approved):
🤞

Makefile
```diff
@@ -79,6 +80,9 @@ clean:
 	@find . -name "__pycache__" -type d -print0 | xargs -0 rm -rf
 	@find . -name "*.pyc" -delete
 
+pypi-release:
+	@$(CONTAINER_ENGINE) run -e UID=$(shell id -u) -e TWINE_USERNAME -e TWINE_PASSWORD -v $(shell pwd):/work --rm $(PYPI_PUSH_IMAGE) ./build_tag.sh
```
BumbleFeng, line 84: I think you can use CONTAINER_UID here to replace $(shell id -u)

fishi0x01: Sadly, podman does not necessarily map uids properly back. chown UID leads to funny results on host system, even though the env var has the proper value set. Using a different approach now.

PR #3210, opened 2023-02-16
[terraform-resources] add ability to set ip_address_type to ipv4/dualstack for alb

jfchevrette:
Allow configuring ip_address_type for ALB resources

This is an optional parameter as per the [terraform aws provider docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb#ip_address_type) and neither the terraform aws provider or aws api docs document a default. The AWS console defaults to `ipv4` so we set that as a default here to be more consistent/explicit

part of https://issues.redhat.com/browse/APPSRE-7044

depends on https://github.com/app-sre/qontract-schemas/pull/381

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -4637,7 +4637,7 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
             "provider": provider,
             "name": identifier,
             "internal": False,
-            "ip_address_type": "dualstack",
+            "ip_address_type": resource.get("ip_address_type", "ipv4"),
             "load_balancer_type": "application",
             "security_groups": [f"${{{sg_tf_resource.id}}}"],
             "subnets": [s["id"] for s in vpc["subnets"]],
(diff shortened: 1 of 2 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -4649,6 +4648,10 @@ def populate_tf_resource_alb(self, spec, ocm_map=None):
         if idle_timeout:
             values["idle_timeout"] = idle_timeout
 
+        ip_address_type = resource.get("ip_address_type")
+        if ip_address_type:
+            values["ip_address_type"] = ip_address_type
```
steveteahan, line 4653: If unset I'm assuming this will just be ipv4?

steveteahan (approved):

reconcile/utils/terrascript_aws_client.py
jfchevrette: As discussed I've set the default to "ipv4" in code to match the apparent default from the AWS console

Neither the aws docs or the terraform provider docs mention a default

steveteahan:
LGTM

PR #3213, opened 2023-02-17
Add cluster_name and namespace_name args to openshift_namespaces

hemslo:
`cluster_name` and `namespace_name` args make it easy debug

example:

```
qontract-reconcile \
  --config config.debug.toml \
  --dry-run \
  --log-level DEBUG \
  openshift-namespaces \
  --cluster-name some-cluster \
  --namespace-name some-namespace
```

### DIFF
--- a/reconcile/test/test_openshift_namespaces.py
+++ b/reconcile/test/test_openshift_namespaces.py
@@ -201,3 +201,25 @@ class TestOpenshiftNamespaces(TestCase):
         with self.assertRaises(SystemExit), contextlib.redirect_stderr(f):
             openshift_namespaces.run(False, thread_pool_size=1)
             self.assertIn("SomeError", f.getvalue())
+
+    def test_run_with_cluster_name(self):
+        self.test_ns = [
+            NS(c1, n1, delete=False, exists=False),
+            NS(c2, n2, delete=False, exists=False),
+        ]
+
+        openshift_namespaces.run(False, thread_pool_size=1, cluster_name=c1)
+
+        self.oc_clients[c1].new_project.assert_called_with(n1)
+        self.assertNotIn(c2, self.oc_clients)
+
+    def test_run_with_namespace_name(self):
+        self.test_ns = [
+            NS(c1, n1, delete=False, exists=False),
+            NS(c2, n2, delete=False, exists=False),
+        ]
+
+        openshift_namespaces.run(False, thread_pool_size=1, namespace_name=n1)
+
+        self.oc_clients[c1].new_project.assert_called_with(n1)
+        self.assertNotIn(c2, self.oc_clients)
--- a/reconcile/test/test_openshift_namespaces.py
+++ b/reconcile/test/test_openshift_namespaces.py
@@ -85,22 +85,28 @@ class TestOpenshiftNamespaces(TestCase):
     def test_create_namespace(self):
         self.test_ns = [
             NS(c1, n1, delete=False, exists=False),
+            NS(c2, n2, delete=False, exists=False),
         ]
 
         openshift_namespaces.run(False, thread_pool_size=1)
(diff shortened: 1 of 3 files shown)


reconcile/test/test_openshift_namespaces.py
```diff
@@ -201,3 +201,25 @@ def test_error_handling_project_exists(self):
...
+    def test_run_with_namespace_name(self):
+        self.test_ns = [
+            NS(c1, n1, delete=False, exists=False),
+            NS(c2, n2, delete=False, exists=False),
+        ]
+
+        openshift_namespaces.run(False, thread_pool_size=1, namespace_name=n1)
+
+        self.oc_clients[c1].new_project.assert_called_with(n1)
+        self.assertNotIn(c2, self.oc_clients)
```
fishi0x01, line 225: Quickly scanned through the existing tests and Im not sure the following is already checked: Should we have one more test for what happens when neither `cluster_name` nor `namespace_name` is set? Like

```python
self.test_ns = [
    NS(c1, n1, delete=False, exists=False),
    NS(c2, n2, delete=False, exists=False),
]

openshift_namespaces.run(False, thread_pool_size=1, namespace_name=n1)

self.oc_clients[c1].new_project.assert_called_with(n1)
self.oc_clients[c1].new_project.assert_called_with(n2)
self.assertIn(c1, self.oc_clients)
self.assertIn(c2, self.oc_clients)
```

Just to be sure that this is fully backwards compatible.

fishi0x01 (commented):
Nice :+1:
My only concern is we might want to check one more test-case for backwards-compatibility.

Also out of curiosity: how did you test this locally? Did you run it in dry-run mode?

reconcile/test/test_openshift_namespaces.py
hemslo: All other test cases are called without `cluster_name` or `namespace_name` like
`test_create_namespace` or `test_delete_namespace`,

https://github.com/app-sre/qontract-reconcile/pull/3213/files#diff-f3d0b8de5c8ef880bbe90d3cffd764767cc0441158c1dc44fdb130687b9c8506L85-R103

Not sure if add another similar one will make it clearer or duplicate.

hemslo (commented):
Tested it by running dry run locally with debug config.

reconcile/test/test_openshift_namespaces.py
hemslo: All other test cases are called without `cluster_name` or `namespace_name` like
`test_create_namespace` or `test_delete_namespace`,

https://github.com/app-sre/qontract-reconcile/pull/3213/files#diff-f3d0b8de5c8ef880bbe90d3cffd764767cc0441158c1dc44fdb130687b9c8506L85-R103

Not sure if add another similar one will make it clearer or duplicate.

fishi0x01: My understanding is the other test cases do not explicitly check the resulting set? They mainly do some assert_not_called checks - not 100% sure this is equivalent. I might as well be wrong.

fishi0x01: Alternatively, would it make sense to extend these test cases you linked above? Like adding a 2nd namespace there and explicitly checking the clients are generated like you do in your new test cases?

hemslo: Good idea, maybe I can use parameterized tests to cover those cases, I'll push them later.

hemslo: Since we are not using parameterized tests, I modified first 2 happy cases for backwards-compatibility. Please have another look @fishi0x01

fishi0x01 (approved):
LGTM! Thanks you!

PR #3215, opened 2023-02-17
[skupper-network] use typed OCMap

chassing:
Refactor `skupper-network` to use the new typed `OCMap` class. 

* improve `openshift_base` to support old `OC_Map` and new `OCMap`
* improve `OCMap.get()` and `OCMap.get_cluster()` returned types
* remove unneeded `oc` checks in `openshift_users` and `dashdotdb_cso`

Ticket: [APPSRE-7033](https://issues.redhat.com/browse/APPSRE-7033)

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -104,9 +104,21 @@ class HasOrgAndGithubUsername(Protocol):
     github_username: str
 
 
+class HasOCMap(Protocol):
+    """An OCMap protocol."""
+
+    def get(
+        self, cluster: str, privileged: bool = False
+    ) -> Union[OCDeprecated, OCLogMsg]:
+        ...
+
+    def get_cluster(self, cluster: str, privileged: bool = False) -> OCDeprecated:
+        ...
+
+
 def init_specs_to_fetch(
     ri: ResourceInventory,
-    oc_map: OC_Map,
+    oc_map: HasOCMap,
     namespaces: Optional[Iterable[Mapping]] = None,
     clusters: Optional[Iterable[Mapping]] = None,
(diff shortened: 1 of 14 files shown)


chassing:
@fishi0x01, please review 🙇

reconcile/openshift_base.py
```diff
@@ -104,9 +104,21 @@ class HasOrgAndGithubUsername(Protocol):
     github_username: str
 
 
+class HasOCMap(Protocol):
```
fishi0x01, line 107: Nice idea to keep it backwards-compatible :+1: 

About the naming, what is the status there now? I guess in this case `Has` prefix is ok because we already have an `OCMap` class in this context right? Referring to this discussion https://github.com/app-sre/qontract-reconcile/pull/3154#discussion_r1094138599

reconcile/utils/oc_map.py
```diff
@@ -192,16 +192,14 @@ def _is_cluster_disabled(self, cluster_info: OCConnectionParameters) -> bool:
 
     def get(
         self, cluster: str, privileged: bool = False
-    ) -> Optional[Union[OCDeprecated, OCLogMsg]]:
+    ) -> Union[OCDeprecated, OCLogMsg]:
```
fishi0x01, line 195: Nice, this makes some of the checks redundant now. Good catch!!!

fishi0x01 (approved):
This is great! Thank you :bow:

reconcile/openshift_base.py
chassing: You are right, it shouldn't use the `Has` prefix. I renamed it to `ClusterMap`.

I want to have an AI, like ChatGPT, just for good naming suggestions 😄

PR #3216, opened 2023-02-17
APPSRE-6609 OCMap for openshift-groups

fishi0x01:
- introduce `OCMap` for `openshift-groups` integration
- enforce types for whole integration

Depends on https://github.com/app-sre/qontract-reconcile/pull/3215 to make mypy happy (fixed return types for `get()` and `get_clusters()`)


reconcile/openshift_groups.py
```diff
@@ -112,41 +101,48 @@ def fetch_current_state(thread_pool_size, internal, use_jump_host):
     return oc_map, current_state, ocm_clusters
 
 
-def fetch_desired_state(oc_map, enforced_user_keys=None):
+def fetch_desired_state(
+    oc_map: ClusterMap, enforced_user_keys: Optional[list[str]] = None
+) -> list[dict[str, str]]:
     gqlapi = gql.get_api()
-    roles = expiration.filter(gqlapi.query(ROLES_QUERY)["roles"])
```
chassing, line 117: You removed `expiration.filter`, was that on purpose?

reconcile/openshift_groups.py
```diff
@@ -112,41 +101,48 @@ def fetch_current_state(thread_pool_size, internal, use_jump_host):
...
-                a["cluster"]["auth"],
+                a.cluster.name,
+                a.cluster.auth,
                 enforced_user_keys=enforced_user_keys,
             )
-            for u in r["users"]:
-                for username in {u[user_key] for user_key in user_keys}:
+            for u in r.users:
+                for username in {
+                    u.dict(by_alias=True)[user_key] for user_key in user_keys
```
chassing, line 125: Please use `getattr(u, user_key)` instead of `u.dict(by_alias=True)[user_key]`. It would be much easier to read.

reconcile/openshift_users.py
```diff
@@ -101,6 +101,8 @@ def fetch_desired_state(
     oc_map: Optional[OCMap], enforced_user_keys: Any = None
```
chassing, line 101: Does it make sense to remove the `Optional` here?

reconcile/openshift_groups.py
fishi0x01: Oh that is a bug! `expiration.filter` is important. I didn't notice locally. Fixed now.

Great catch! Eagle eyes! :bow:

reconcile/openshift_users.py
fishi0x01: It might make sense, but Im not 100% sure. This function is also used by other integrations, like `github_org` or `slack-usergroups`. Kept the interface here stable for compatibility.

chassing (approved):
Nice work ❤️

PR #3217, opened 2023-02-17
add kustomize templating

rccrdpccl:
Add kustomize templating.
Used string templating to replace env var in the manifests (useful for IMAGE_TAG, and such - as vars support is deprecated in kustomize).

By default it fetches the `resourceTemplate.path` value for kustomize DIR, however each target can override it with `KUSTOMIZE_DIR` parameter (useful to select different overlays each env).


rccrdpccl:
Unfortunately we cannot easily tests, we're working on some way to validate. Ideas how to go about this are more than welcome

fishi0x01:
@rccrdpccl I added a comment to the above ticket. I will keep the use-case for tenants in mind. I.e., make sure no prod secrets are required and also add some documentation for (easy :crossed_fingers: ) consumption.

rccrdpccl:
It looks like this usage docs should be added in app-interface's documentation - please let us know if that's not the case and we should create something for this repository

PR #3218, opened 2023-02-19
[statuspage] qenerate + testability

geoberle:
TLDR:
this is a refactor of the status page integration to catch up with more than a year worth of improvements on writing integrations. it also prepares the code for upcoming changes in the statuspage.io python lib replacement and paves the way for more status providers (e.g. the prometheus alert one) as well as dormant work on scheduled maintenances.

the integration has been placed in a dedicated module and split into meaningful submodules that can be tested in isolation. in general, testability has been a core concern, so this refactor makes sure that nearly everything can be tested without mocking (with few exceptions).

additionally all GQL interaction has been refactored into qenerate based dataclasses.

the way pydantic dataclasses are used throughout the code will allow us to replace the statuspage python lib with a simple uplink client leveraging those dataclasses for payloads. a replacement of the current statuspage lib is necessary anyways because it is unmaintained, untyped and does no honor pagination on the statuspage.io API.

for the reviewer: the changes in structure have been significant so i would not bother too much about diffs but consider only the new code


reconcile/statuspage/atlassian.py
```diff
@@ -30,126 +35,316 @@ class AtlassianComponent(BaseModel):
...
+        self._components_by_id = {c.id: c for c in self._components}
+        self._components_by_displayname = {c.name: c for c in self._components}
+        self._group_name_to_id = {g.name: g.id for g in self._components if g.group}
+        self._group_id_to_name = {g.id: g.name for g in self._components if g.group}
+
+    def get_component_by_id(self, id: str) -> Optional[StatusComponent]:
+        raw = self.get_raw_component_by_id(id)
+        if raw:
+            return self._bound_raw_component_to_status_component(raw)
         else:
```
chassing, line 123: Unneeded `else` :)

reconcile/statuspage/atlassian.py
```diff
@@ -1,23 +1,28 @@
 import logging
-from collections.abc import Iterable
 from typing import (
     Any,
     Optional,
+    Protocol,
+    Tuple,
```
chassing, line 6: Don't use `Tuple`; you can use `tuple` instead

reconcile/statuspage/atlassian.py
```diff
@@ -30,126 +35,316 @@ class AtlassianComponent(BaseModel):
...
 
-    def component_ids(self) -> Iterable[str]:
-        return self.components_by_id.keys()
 
-    def _find_component(
-        self, component: StatusComponent
-    ) -> Optional[AtlassianComponent]:
-        if component.component_id and component.component_id in self.components_by_id:
-            return self.components_by_id.get(component.component_id)
+class LegacyLibAtlassianAPI:
```
chassing, line 54: A `__doc__` string would be nice.

reconcile/statuspage/atlassian.py
```diff
@@ -30,126 +35,316 @@ class AtlassianComponent(BaseModel):
...
+                display_name=raw_component.name,
+                description=raw_component.description,
+                group_name=group_name,
+                status_provider_configs=[
+                    ManualStatusProvider(
+                        component_status=raw_component.status,
+                    )
+                ],
+            )
+        else:
```
chassing, line 166: Unnecessary `else`

reconcile/statuspage/atlassian.py
```diff
@@ -30,126 +35,316 @@ class AtlassianComponent(BaseModel):
...
+        # shortcut execution if there is nothing to do
+        if (
+            current
+            and desired.display_name == current.name
+            and desired.description == current.description
+            and desired.group_name == current_group_name
+            and not status_update_required
+        ):
+            return False
+        else:
```
chassing, line 251: Unnecessary `else`

reconcile/statuspage/atlassian.py
```diff
@@ -30,126 +35,316 @@ class AtlassianComponent(BaseModel):
...
+        if not dry_run:
+            self._binding_state.bind_component(component_name, component_id)
 
 
-def load_provider(page: StatusPage) -> StatusPageProvider:
-    provider = AtlassianStatusPage(
-        page_id=page.page_id,
-        api_url=page.api_url,
-        token=page.credentials.get("token"),
+def init_provider_for_page(
```
chassing, line 334: TBH, I don't like the coupling to `page` and `secret_reader` here (ok maybe `page` can be argued somehow). What do you think about:

```python
def init_provider_for_page(page_id, name, api_url, token, component_binding_state)
```

reconcile/statuspage/page.py
```diff
@@ -0,0 +1,144 @@
...
+    page: StatusPageV1,
+    secret_reader: SecretReaderBase,
+    component_binding_state: ComponentBindingState,
+) -> StatusPageProvider:
+    """
+    Initialize a status page provider for a given status page.
+    """
+    if page.provider in _PROVIDERS:
+        return _PROVIDERS[page.provider](page, secret_reader, component_binding_state)
+    else:
```
chassing, line 114: Unnecessary `else`

reconcile/statuspage/status.py
```diff
@@ -0,0 +1,99 @@
...
+    component_status: str
+    """
+    The status to be used for the component if the
+    time window is active or if no time window is defined.
+    """
+
+    def get_status(self) -> Optional[str]:
+        if self._is_active():
+            return self.component_status
+        else:
```
chassing, line 64: Unnecessary `else`

reconcile/statuspage/status.py
```diff
@@ -0,0 +1,99 @@
...
+    Translates a status provider config from the desired state into
+    provider specific implementation that provides the status resolution logic.
+    """
+    if isinstance(cfg, ManualStatusProviderV1):
+        return ManualStatusProvider(
+            component_status=cfg.manual.component_status,
+            start=cfg.manual.q_from,
+            end=cfg.manual.until,
+        )
+    else:
```
chassing, line 98: Unnecessary `else`

reconcile/test/statuspage/fixtures.py
```diff
@@ -0,0 +1,178 @@
+from typing import (
+    Any,
+    Iterable,
+    Optional,
+    Tuple,
```
chassing, line 5: Please use `tuple` instead

reconcile/test/statuspage/fixtures.py
```diff
@@ -0,0 +1,178 @@
+from typing import (
+    Any,
+    Iterable,
```
chassing, line 3: `from collections.abc import Iterable`

reconcile/statuspage/integration.py
```diff
@@ -0,0 +1,107 @@
+import logging
+import sys
+from typing import Callable
```
chassing, line 3: `from collections.abc import Callable`

reconcile/statuspage/page.py
```diff
@@ -0,0 +1,144 @@
+from abc import (
+    ABC,
+    abstractmethod,
+)
+from typing import (
+    Callable,
```
chassing, line 6: `from collections.abc import Callable`

setup.py
```diff
@@ -65,6 +65,7 @@
         "deepdiff6==6.2.0",
         "jsonpath-ng~=1.5",
         "networkx~=2.8",
+        "uplink~=0.9.7",
```
chassing, line 68: It looks like you don't use `uplink`

chassing (commented):
`reconcile/test/statuspage/test_statuspage_integration.py` is an empty file

reconcile/statuspage/atlassian.py
geoberle: i agree about the secret_reader. i will replace it to make the provider ignorant about the source of the token, which is a good thing. i will keep the page object for now though.

PR #3222, opened 2023-02-21
don't print GQL SHA during dry-runs

geoberle:
the actual SHA used during PR checks is not important. hence lets not print it to keep the PR check logs easier to read

### DIFF
--- a/reconcile/utils/runtime/environment.py
+++ b/reconcile/utils/runtime/environment.py
@@ -19,7 +19,9 @@ LOG_DATEFMT = "%Y-%m-%d %H:%M:%S"
 
 
 def init_env(
-    log_level: Optional[str] = None, config_file: Optional[str] = None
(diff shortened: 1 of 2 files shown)


reconcile/utils/runtime/environment.py
```diff
@@ -19,7 +19,9 @@
 
 
 def init_env(
-    log_level: Optional[str] = None, config_file: Optional[str] = None
+    log_level: Optional[str] = None,
+    config_file: Optional[str] = None,
+    print_gql_url: Optional[bool] = True,
```
chassing, line 24: Why is `print_gql_url` optional? `True`, `False`, `None`?

geoberle: 🤦  thank you - fixed

chassing (approved):

PR #3225, opened 2023-02-21
Add new integration to support Cloudflare DNS management.

OliviaHY:
### Why:
To migrate from Dyn
### What:
New integration that can manage zones and records in Cloudflare. Based on this [design doc](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/terraform-cloudflare-dns-support.md).
### Validation:
Dry-ran terraform-cloudflare-resources against prod data
Ran terraform-cloudflare-dns against dev-data locally.
Test cases:
- Create zone, see it shows up in my test Cloudflare account only with pending name server update.
- Delete created zones
- Create A, CNAME, NS, MX records
- Delete any created records
- Update attributes of those records
- Create NS and MX records with multiple values with same name
- Tested integration manager plan to create multiple pod with zone name passed as sharding key
- Tested Sharded dry run
- Performance test: zone with 1000 record creation took ~5min and used ~100MB RAM.
### Dependency:
[schema changes
](https://github.com/app-sre/qontract-schemas/pull/389) promoted


reconcile/gql_definitions/terraform_cloudflare_resources/terraform_cloudflare_resources.py
```diff
@@ -151,11 +152,12 @@ class Config:
...
 
 
-class CloudflareZoneRecordV1(BaseModel):
+class CloudflareDnsRecordV1(BaseModel):
+    identifier: str = Field(..., alias="identifier")
     name: str = Field(..., alias="name")
     q_type: str = Field(..., alias="type")
     ttl: int = Field(..., alias="ttl")
-    value: str = Field(..., alias="value")
+    value: Optional[str] = Field(..., alias="value")
```
OliviaHY, line 160: This was because value and data field are mutually exclusive.  https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/record#data

reconcile/gql_definitions/common/namespaces.py
```diff
@@ -274,7 +274,7 @@ class Config:
         extra = Extra.forbid
 
 
-class NamespaceManagedResourceNamesV1(BaseModel):
+class ManagedResourceNamesV1(BaseModel):
```
OliviaHY, line 277: This was due to https://github.com/app-sre/qontract-schemas/pull/382/files renamed the schema and was merged before the schema change for this.

steveteahan (approved):
Approving with the assumption that the only changes since the rollback were the two new commits.

PR #3226, opened 2023-02-21
APPSRE-6609 fix ocm-groups

fishi0x01:
Bug introduced here https://github.com/app-sre/qontract-reconcile/pull/3216

Apparently `ocm-groups` also uses one of these functions -,-

This makes it backwards compatible again.


lechuk47 (approved):
LGTM

PR #3227, opened 2023-02-22
Fix combining owners

d-hat:
when scanning for applicable `OWNERS` files, the test `path.startswith` is used to identify parent directories.  This can go wrong when there are two adjacent directories `foo/OWNERS` and `foobar/OWNERS`, as `foo` is considered to be a parent of `foobar`.

This incorrect behaviour was verified in `utils.repo_owners.RepoOwners`.  To fix, the test is changed from:

```py
child_path.startswith(parent_path)
```

to:

```py
os.path.commonpath([child_path, parent_path]) == parent_path
```

also found a couple of other locations which look like having the same test applied incorrectly.

**NOTE** no tests have been added - sorry, I'm not familiar enough with the code base or test suite to see how to connect any effective tests.  If you can point me in a direction of doing so, I'm happy to try and contribute.

Interactive demo:

```py
In [1]: '/foo/bar'.startswith('/foo/b')
Out[1]: True

In [2]: '/foo/bar'.startswith('/foo')
Out[2]: True

In [3]: import os

In [4]: os.path.commonpath(['/foo/bar', '/foo/b']) == '/foo/b'
Out[4]: False

In [5]: os.path.commonpath(['/foo/bar', '/foo']) == '/foo'
Out[5]: True
```

https://issues.redhat.com/browse/APPSRE-7164

### DIFF
--- a/reconcile/change_owners/change_types.py
+++ b/reconcile/change_owners/change_types.py
@@ -19,7 +19,6 @@ from typing import (
     Tuple,
 )
 
-import os
 import anymarkup
 import jinja2
 import jinja2.meta
@@ -887,7 +886,7 @@ JSON_PATH_ROOT = "$"
 
 
 def change_path_covered_by_allowed_path(changed_path: str, allowed_path: str) -> bool:
-    return (allowed_path == JSON_ROOT) or (os.path.commonpath([changed_path, allowed_path]) == allowed_path)
+    return changed_path.startswith(allowed_path) or allowed_path == JSON_PATH_ROOT
 
 
 def get_priority_for_changes(
--- a/reconcile/change_owners/change_types.py
+++ b/reconcile/change_owners/change_types.py
@@ -19,6 +19,7 @@ from typing import (
     Tuple,
 )
 
+import os
 import anymarkup
 import jinja2
 import jinja2.meta
@@ -886,7 +887,7 @@ JSON_PATH_ROOT = "$"
 
 
 def change_path_covered_by_allowed_path(changed_path: str, allowed_path: str) -> bool:
-    return changed_path.startswith(allowed_path) or allowed_path == JSON_PATH_ROOT
+    return (allowed_path == JSON_ROOT) or (os.path.commonpath([changed_path, allowed_path]) == allowed_path)
 
 
 def get_priority_for_changes(
--- a/reconcile/test/test_repo_owners.py
+++ b/reconcile/test/test_repo_owners.py
@@ -1,44 +1,47 @@
 from reconcile.utils import repo_owners
 
+
 def test_repo_owners_subpath() -> None:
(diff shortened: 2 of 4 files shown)


geoberle:
hi @d-hat. can you create a ticket on https://issues.redhat.com/projects/APPSRE and link it in the PR description?

i will try to guide you through this PR. do you have an example repo where this can be observed? if it is a private one, mention it on the ticket.

geoberle:
thanks @d-hat. you identified the `repo_owners.py` file as the correct one to fix this. the other two files are unrelated and deal with other types of ownerships, so please revert the changes in these files.

your fix in `repo_owners.py` is spot on.

about testing: the `RepoOwners` class is not tested in full right now so there is no existing test file i can point you to so i suggest you create a new one. if you want to see how you can mock some things, e.g. the `get_owners_map` method on the `RepoOwners` class, you can find a test that does similar things [here](https://github.com/app-sre/qontract-reconcile/blob/353ef03027ea67c76e2a49c1968805eb47c80f55/reconcile/test/test_slack_usergroups.py#L216)).

d-hat:
ack.  I'll revert the 2nd commit - though from looking at the surrounding code I think one of those changes should be considered.

The change I made in `reconcile/change_owners/change_types.py` looks to be incorrect, since iiuc these are JSON paths & not filesystem paths.  I think `.startswith()` is still risky there for similar reasons, but getting it right is not as simple as handing over to `os.path`.

The change in `reconcile/utils/saasherder.py` I'm more confident was going in the right direction.  `allowedSecretParameterPaths` is indeed a list of filesystem paths so it would seem to want this kind of comparison.  But I'm not at all familiar with that functionality, so I won't wade in with attempting to test & stage a fix.

d-hat:
Added some tests in de78670.  I'm not familiar with using `unittest` to build mocks and my first efforts in that direction failed so I went more brute-force, monkey-patching an instantiated `RepoOwners` object initialised with `git_cli = None`.  These tests do fail without 43003d2 and pass with it, so at least they're hitting the right code.

geoberle:
> I think one of those changes should be considered

that would be https://github.com/app-sre/qontract-reconcile/pull/3242 right? thanks again, i will have a look at it as well

geoberle:
about the PR check failure: you can get rid of all import/format related issues by running `make format`

d-hat:
okay if that passes it passes, if it doesn't I'm out of steam.  Unable to see the jenkins results and running the test locally takes forever.

kwilczynski:
> okay if that passes it passes, if it doesn't I'm out of steam. Unable to see the jenkins results and running the test locally takes forever.
[...]

Having a quick look at the job in Jenkins, everything looks OK. @geoberle, when he comes on-line, will have a final look.

Thank you a lot for your help here!

PR #3228, opened 2023-02-22
APPSRE-6609 OCMap for openshift-upgrade-watcher

fishi0x01:
- `OCMap` for `openshift-upgrade-watcher`
- enforce types for `openshift-upgrade-watcher`

### DIFF
--- a/reconcile/openshift_upgrade_watcher.py
+++ b/reconcile/openshift_upgrade_watcher.py
@@ -1,11 +1,25 @@
 import logging
+from collections.abc import (
+    Callable,
+    Iterable,
+)
 from datetime import datetime
 from typing import Optional
 
 from reconcile import queries
+from reconcile.gql_definitions.common.clusters import ClusterV1
 from reconcile.slack_base import slackapi_from_queries
+from reconcile.typed_queries.app_interface_vault_settings import (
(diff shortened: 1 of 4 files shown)


reconcile/openshift_upgrade_watcher.py
```diff
@@ -72,13 +86,19 @@ def notify_upgrades_start(
...
+def notify_upgrades_done(
+    clusters: Iterable[ClusterV1], state: State, slack: Optional[SlackApi]
+) -> None:
     for cluster in clusters:
-        cluster_name = cluster["name"]
-        version = cluster["spec"]["version"]
+        cluster_name = cluster.name
+        if not cluster.spec:
+            raise RuntimeError(f"Cluster '{cluster_name}' does not have any spec.")
+        version = cluster.spec.version
```
chassing, line 101: Remove `cluster_name` and use 'cluster.name' instead (same for `version`). This will be much easier to read :)

chassing (approved):
Just a small nit, feel free to ignore it :)

reconcile/openshift_upgrade_watcher.py
fishi0x01: Good point. Fixed :)

PR #3229, opened 2023-02-22
Add a new exception to hande the case of a secret with no versions

andreugallofre:
We don't want to cause a failure of the integration as this cause a critical page to the team, but we want to be sure that we are aware of this cases.


chassing (approved):
LGTM

cubismod (approved):

PR #3233, opened 2023-02-23
allow disable per cluster for route53

fishi0x01:
elbFQDNs are used in a zone in our `app-sre` account. Having a bad FQDN in a single cluster can block this integration for other clusters too, so we should have a way to disable it on a per cluster base.

Schema: https://github.com/app-sre/qontract-schemas/pull/393

### DIFF
--- a/reconcile/terraform_aws_route53.py
+++ b/reconcile/terraform_aws_route53.py
@@ -78,7 +78,7 @@ def build_desired_state(
             if target_cluster:
                 if not integration_is_enabled(
                     integration=QONTRACT_INTEGRATION.replace("_", "-"),
-                    disable_obj=target_cluster["disable"],
+                    disable_obj=target_cluster,
                 ):
                     continue
                 target_cluster_elb = target_cluster["elbFQDN"]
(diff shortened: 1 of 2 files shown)


chassing:
You can use the utils method [integration_is_enabled](https://github.com/app-sre/qontract-reconcile/blob/e8fd4690cbb06916f3a9dcfd80c1bc3fd94cb49a/reconcile/utils/disabled_integrations.py#L39) instead

reconcile/terraform_aws_route53.py
```diff
@@ -75,6 +76,11 @@ def build_desired_state(
             # Process '_target_cluster'
             target_cluster = record.pop("_target_cluster", None)
             if target_cluster:
+                if not integration_is_enabled(
+                    integration=QONTRACT_INTEGRATION.replace("_", "-"),
+                    disable_obj=target_cluster,
+                ):
+                    continue
```
chassing, line 83: Please add log.info message, otherwise, it would be hard to see why the integration doesn't do anything for a specific cluster

andreugallofre (approved):
LGTM

reconcile/terraform_aws_route53.py
fishi0x01: Fixed

chassing (approved):
LGTM

PR #3235, opened 2023-02-23
APPSRE-7138: Support default Server Side Encryption (SSE) setting for S3 buckets

kwilczynski:
AWS has been rolling out their new approach to enabling a default S3 Bucket encryption policy.

Many of the old S3 Buckets did not have the SSE (Server Side Encryption) encryption enabled in the past, and there was also no default, or rather the default was no encryption of the objects stored at rest. This rollout of new settings and policies will result in the Terraform detecting drift perpetually where the SSE wasn't previously enabled.

Thus, add a new default SEE setting that will be applied where the configuration was not explicitly added to stop the problem with the state drift and align with the new recommendation from AWS per:

- [Setting default server-side encryption behavior for Amazon S3 buckets](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucket-encryption.html)
- [Amazon S3 now automatically encrypts all new objects](https://docs.aws.amazon.com/AmazonS3/latest/userguide/default-encryption-faq.html)

Related: [APPSRE-7138](https://issues.redhat.com/browse/APPSRE-7138)

Signed-off-by: Krzysztof Wilczyński <kwilczynski@redhat.com>

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -1666,8 +1666,9 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
             values.setdefault("lifecycle", {}).setdefault("ignore_changes", []).append(
                 "grant"
             )
-        server_side_encryption_configuration = common_values.get(
-            "server_side_encryption_configuration", DEFAULT_S3_SSE_CONFIGURATION
+        server_side_encryption_configuration = (
+            common_values.get("server_side_encryption_configuration")
+            or DEFAULT_S3_SSE_CONFIGURATION
         )
         values[
             "server_side_encryption_configuration"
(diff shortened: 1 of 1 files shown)


kwilczynski:
Re-based against master.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1663,12 +1667,11 @@ def populate_tf_resource_s3(self, spec):
                 "grant"
             )
         server_side_encryption_configuration = common_values.get(
-            "server_side_encryption_configuration"
+            "server_side_encryption_configuration", DEFAULT_S3_SSE_CONFIGURATION
```
geoberle, line 1670: would you consider using
`common_values.get("server_side_encryption_configuration") or DEFAULT_S3_SSE_CONFIGURATION`

just in case someone would ever have a `server_side_encryption_configuration: null` in their s3 defaults file.

definitely a nit kind-of-thing though

geoberle (approved):

reconcile/utils/terrascript_aws_client.py
kwilczynski: Done. Thank you.

PR #3236, opened 2023-02-23
display cross-org version data inheritance in fleet upgrade policy outputs

patjlm:
APPSRE-7162

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -517,13 +517,13 @@ def get_available_upgrades(ocm: OCM, version: str, channel: str) -> list[str]:
     return [u for u in upgrades if not ocm.version_blocked(u)]
 
 
-def inheritVersionDataText(ocm_org: str, ocm_specs: list[dict]) -> str:
+def inherit_version_data_text(ocm_org: str, ocm_specs: list[dict]) -> str:
     ocm_spec = [o for o in ocm_specs if o["name"] == ocm_org][0]
-    inheritVersionData = ocm_spec["inheritVersionData"]
-    if not inheritVersionData:
+    inherit_version_data = ocm_spec["inheritVersionData"]
+    if not inherit_version_data:
         return ""
-    inheritedOrgs = [f"[{o['name']}](#{o['name']})" for o in inheritVersionData]
-    return f"inheriting version data from {', '.join(inheritedOrgs)}"
+    inherited_orgs = [f"[{o['name']}](#{o['name']})" for o in inherit_version_data]
+    return f"inheriting version data from {', '.join(inherited_orgs)}"
 
 
 @get.command()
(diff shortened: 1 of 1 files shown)


tools/qontract_cli.py
```diff
@@ -517,6 +517,15 @@ def get_available_upgrades(ocm: OCM, version: str, channel: str) -> list[str]:
     return [u for u in upgrades if not ocm.version_blocked(u)]
 
 
+def inheritVersionDataText(ocm_org: str, ocm_specs: list[dict]) -> str:
```
geoberle, line 520: would you consider snake case as the pythonic way of naming things?

tools/qontract_cli.py
```diff
@@ -517,6 +517,15 @@ def get_available_upgrades(ocm: OCM, version: str, channel: str) -> list[str]:
...
+    ocm_spec = [o for o in ocm_specs if o["name"] == ocm_org][0]
```
geoberle, line 521: this PR is nice!

so allow me some nits-picking :D 

is there a situation where `org_name` is not available in the `ocm_specs`? in the current call contexts for sure not because you control that well. but if this method would be called from somewhere else, the access to [0] would fail.

you could use next with an iterator, resolve to `None` if not available and explicitly check for it.
```
    ocm_spec = next(o for o in ocm_specs if o["name"] == ocm_org], None)
    if ocm_spec:
       ...
```
the nice thing about `next` is, that it stops the moment you find an element. for sure not relevant in this context but hey, this is about nit-picking :)

tools/qontract_cli.py
```diff
@@ -517,6 +517,15 @@ def get_available_upgrades(ocm: OCM, version: str, channel: str) -> list[str]:
...
+    inheritVersionData = ocm_spec["inheritVersionData"]
+    if not inheritVersionData:
+        return ""
+    inheritedOrgs = [f"[{o['name']}](#{o['name']})" for o in inheritVersionData]
```
geoberle, line 525: is repeating `o['name']` twice intended? the output would be something like
```
inheriting version data from [org-1](#org-1), [org-2](#org-2)
```

patjlm: yes, each org has its own section, and we're creating links to them.

tools/qontract_cli.py
patjlm: oh, sure

tools/qontract_cli.py
patjlm: TIL!
I am not against it, but I am wondering if that's better though. Shouldn't we always call that function knowing the org with that name exists ? Isn't it better to fail in that case?

tools/qontract_cli.py
patjlm: Fixed in eb4f8e51

tools/qontract_cli.py
geoberle: failing is a good thing. but it should be something nicer than an `IndexError`. so it is absolutely fine to not use the next/iterator pattern. checking for len and failing with a generic `ValueError` with a readable error message works as well.

patjlm: fixed in [ad2f6db](https://github.com/app-sre/qontract-reconcile/pull/3236/commits/ad2f6dbb707af5b396c302549ad0e36e9e6695af)

geoberle (approved):
LGTM!

PR #3239, opened 2023-02-27
Output to s3

janboll:
Vault is just too slow to list all the secrets. Read encrypted passwords from s3 instead.

[APPSRE-7193](https://issues.redhat.com/browse/APPSRE-7193)


geoberle:
TIL: there is no bulk secret read functionality in vault 😱 

what does "too slow" mean? too slow for what? can you share some numbers?

i see that for the s3 implementation you used threading to parallelise, which is a good thing considering that most time is spent on waiting for I/O. have you tried to do the same for the vault?

can you create/link a ticket for this?

janboll:
My concern with parallel requests towards Vault was that I did not want to cause any issues with it, especially, given its significance to us. Hence the switch to S3 (which was also simple to implement).

geoberle:
> My concern with parallel requests towards Vault was that I did not want to cause any issues with it, especially, given its significance to us. Hence the switch to S3 (which was also simple to implement).

valid concerns. i don't have a better idea. also the secrets are already there in the bucket in the tf states file so we would not really add new secret data to the bucket so to say, also they are encrypted, right?. try to get consensus for this with the team anyways.

tools/qontract_cli.py
```diff
@@ -940,24 +948,19 @@ def user_credentials_migrate_output(ctx, account_name, output_path) -> None:
     )
     credentials = []
     for account, output in tf.outputs.items():
-        if account in skip_accounts:
```
fishi0x01, line 943: Why is `skip_accounts` removed here?

fishi0x01 (commented):
Just one quick question

tools/qontract_cli.py
janboll: Good catch! I initially copy and pasted that migration CLI from the get user credentials one. However, I did not remove this line. But it should be fine to delete, cause we are only migrating one account from the parameter (see setup in line 935). 
Further: my intention is to remove this code once the rollout is done.

fishi0x01 (approved):
Since creds are encrypted LGTM :+1:

geoberle (approved):

PR #3242, opened 2023-02-28
allowedSecretParameterPaths prefix but not parent of secretParameters

d-hat:
trying to make a test for another example of #3227.  In this test, we have `foo` in `allowedSecretParameterPaths` but the `secretParameters` has `path = "foobar/baz"`.

I think this means validation should fail in `_validate_allowed_secret_parameter_paths` with the log message `secret parameter path 'foobar/baz' does not match any of allowedSecretParameterPaths`.

But this is based on a naive reading of the code + test cases, and I could be wrong.  Somebody who uses this function should review.

https://issues.redhat.com/browse/APPSRE-7196

### DIFF
--- a/reconcile/test/test_saasherder_allowed_secret_paths.py
+++ b/reconcile/test/test_saasherder_allowed_secret_paths.py
@@ -1,0 +1,57 @@
+from reconcile.utils.saasherder import SaasHerder
+
+def test_saasherder_allowed_secret_paths():
+    saas_files = [
+            {
+                "path": "path1",
+                "name": "a1",
+                "managedResourceTypes": [],
+                "allowedSecretParameterPaths": [
+                    "foo",
+                ],
+                "resourceTemplates": [
+                    {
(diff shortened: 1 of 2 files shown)


geoberle:
your fix to `saas_herder.py` from https://github.com/app-sre/qontract-reconcile/pull/3227/commits/bc052942aef993bddd315de08e15ba6e977ee644 looks good. also the test looks good. can you add a couple of more testcases?

`make format` will get rid of the format related PR check failure.

thanks again!

geoberle:
can you create another APPSRE ticket for this? ty

geoberle:
format and pylint (`tox -e lint`) still fail, can you check?

geoberle:
also, `make test-app` runs the same checks as the PR check

geoberle (approved):

geoberle:
thanks for your contribution @d-hat !

PR #3247, opened 2023-02-28
[change-owners] fix jsonpath sorting

geoberle:
for certain operations, jsonpaths need to be sorted. the currently used naive approach simply compared the jsonpath strings, which starts to fail when double digit indices are involved.

e.g. `a.b.[10] comes before a.b.[9]` in a pure string comparison

this PR introduces a function that creates a sortable string representation of jsonpaths that takes care of those index numbers by padding with zeros.

https://issues.redhat.com/browse/APPSRE-7197

### DIFF
--- a/reconcile/utils/jsonpath.py
+++ b/reconcile/utils/jsonpath.py
@@ -45,6 +45,29 @@ def narrow_jsonpath_node(
     return None
 
 
+def sortable_jsonpath_string_repr(
+    path: jsonpath_ng.JSONPath, index_padding: int = 5
+) -> str:
+    """
+    Return a string representation of the JSONPath that can be used for sorting.
+    The relevant thing is the reprsentation of an Index, which needs to be left
+    padded with zeros to ensure comparability of the string representation.
+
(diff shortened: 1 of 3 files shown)


reconcile/utils/jsonpath.py
```diff
@@ -45,6 +45,29 @@ def narrow_jsonpath_node(
     return None
 
 
+def sortable_jsonpath_string_repr(
+    path: jsonpath_ng.JSONPath, index_padding: int = 5
+) -> str:
+    """
+    Return a string representation of the JSONPath that can be used for sorting.
+    The relevant thing is the reprsentation of an Index, which needs to be left
```
patjlm, line 53: unacceptable:
```suggestion
    The relevant thing is the representation of an Index, which needs to be left
```

patjlm (approved):

reconcile/utils/jsonpath.py
geoberle: Mon Dieu!!

PR #3248, opened 2023-02-28
APPSRE-6609 fix namespace connections bug

fishi0x01:
`OCMap` initialized with namespaces currently does at least one vault query per namespace per namespace. However, we must only query once per cluster.

**Test**

- Tested locally with `skupper-network` integration, as it is the only one consuming this functionality.
- Added unit test that catches the above issue


fishi0x01:
Closing in favor of https://github.com/app-sre/qontract-reconcile/pull/3249

PR #3249, opened 2023-02-28
APPSRE-6609 fix OCMap connections

fishi0x01:
The current `OCMap` has a crucial bug: it does not create an unprivileged client if there is a privileged client available. This does not comply with how the original `OC_Map` behaves and leads to issues with namespace initialized `OCMap`s.

This overall aligns `OCMap` closer to the behavior of the old `OC_Map`.

This PR:

- fixes the bug: OCMap must hold a non-privileged client for **EVERY** cluster, also the ones that have a privileged connection established.
- reduces the number of vault requests on namespace queries by grouping clusters from namespaces.
- shifts all logic about how to group OC connections away from `OCMap`. `OCMap` is only concerned about creating and holding clients based on the given connection parameters.

**Test**

- ran `skupper-network` locally
- ran `openshift-groups` locally
- added unit tests which cover the above bugs

### DIFF
--- a/reconcile/utils/oc_map.py
+++ b/reconcile/utils/oc_map.py
@@ -60,24 +60,10 @@ class OCMap:
         self._lock = Lock()
         self._jh_ports: dict[str, int] = {}
 
-        unprivileged_connections = (
-            c for c in connection_parameters if not c.is_cluster_admin
-        )
-        privileged_connections = (
-            c for c in connection_parameters if c.is_cluster_admin
-        )
-
         threaded.run(
             self._init_oc_client,
-            unprivileged_connections,
+            connection_parameters,
             self._thread_pool_size,
-            privileged=False,
-        )
-        threaded.run(
-            self._init_oc_client,
-            privileged_connections,
-            self._thread_pool_size,
-            privileged=True,
         )
 
     def _set_jumphost_tunnel_ports(
@@ -91,9 +77,11 @@ class OCMap:
             connection_parameters.jumphost_local_port = self._jh_ports[key]
 
     def _init_oc_client(
-        self, connection_parameters: OCConnectionParameters, privileged: bool
+        self,
+        connection_parameters: OCConnectionParameters,
     ) -> None:
         cluster = connection_parameters.cluster_name
+        privileged = connection_parameters.is_cluster_admin
         if not privileged and self._oc_map.get(cluster):
             return None
         if privileged and self._privileged_oc_map.get(cluster):
--- a/reconcile/utils/oc_map.py
+++ b/reconcile/utils/oc_map.py
@@ -67,7 +67,9 @@ class OCMap:
         # happy with regular dedicated-admin and will request a cluster
         # with oc_map.get(cluster) without specifying privileged access
         # specifically
-        def filter_unique_connection_parameters(connection_parameters: Iterable[OCConnectionParameters]) -> list[OCConnectionParameters]:
+        def filter_unique_connection_parameters(
+            connection_parameters: Iterable[OCConnectionParameters],
+        ) -> list[OCConnectionParameters]:
             unique_by_cluster_name = {c.cluster_name: c for c in connection_parameters}
             return list(unique_by_cluster_name.values())
 
--- a/reconcile/utils/oc_map.py
+++ b/reconcile/utils/oc_map.py
@@ -67,24 +67,22 @@ class OCMap:
         # happy with regular dedicated-admin and will request a cluster
         # with oc_map.get(cluster) without specifying privileged access
         # specifically
-        privileged_clusters: list[OCConnectionParameters] = [
-            c for c in connection_parameters if (c.is_cluster_admin or cluster_admin)
-        ]
-        unprivileged_clusters: list[OCConnectionParameters] = [
-            c
+        all_clusters = {c.cluster_name: c for c in connection_parameters}
+        privileged_clusters = {
+            c.cluster_name: c
             for c in connection_parameters
-            if not (c.is_cluster_admin or cluster_admin)
-        ]
+            if (c.is_cluster_admin or cluster_admin)
+        }
 
         threaded.run(
             self._init_oc_client,
-            unprivileged_clusters,
+            list(all_clusters.values()),
             self._thread_pool_size,
             privileged=False,
         )
         threaded.run(
             self._init_oc_client,
-            privileged_clusters,
+            list(privileged_clusters.values()),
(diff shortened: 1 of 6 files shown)


reconcile/utils/oc_map.py
```diff
@@ -67,24 +67,22 @@ def __init__(
         # happy with regular dedicated-admin and will request a cluster
         # with oc_map.get(cluster) without specifying privileged access
         # specifically
-        privileged_clusters: list[OCConnectionParameters] = [
-            c for c in connection_parameters if (c.is_cluster_admin or cluster_admin)
-        ]
-        unprivileged_clusters: list[OCConnectionParameters] = [
-            c
+        all_clusters = {c.cluster_name: c for c in connection_parameters}
+        privileged_clusters = {
```
hemslo, line 71: Only `.values()` are used for these 2, can we use list instead of dict here? Or are we using dict to remove duplicate values by `cluster_name` (last one wins)?

fishi0x01: Good observation :+1: 

Yeah by using dicts Im trying to reduce the thread overhead.

The current behavior is random one wins with extra threading overhead, see [here](https://github.com/app-sre/qontract-reconcile/blob/ee0f52bf7f2551d00833d6858861ba16d1dd2357/reconcile/utils/oc_map.py#L106-L109).

Now, the original behavior was last one wins, see [here](https://github.com/app-sre/qontract-reconcile/blob/ee0f52bf7f2551d00833d6858861ba16d1dd2357/reconcile/utils/oc.py#L1586-L1590).

Whether last one wins or random one wins doesn't matter, but I would like to reduce the overhead of creating threads unnecessarily and align closer with behavior from original `OC_Map` class.

hemslo: Not sure if these are real cases:
1. Given there are 2 items in `connection_parameters` with same `cluster_name`, one `is_cluster_admin` True, the other one `is_cluster_admin` False
    When create `all_clusters`
    Then last one in `connection_parameters` will be the only one in `all_clusters`
    Will this cause trouble?
2. Given there are 2 items in `connection_parameters` with same `cluster_name`, both `is_cluster_admin` True
    When create `privileged_clusters`
    Then last one in `connection_parameters` will be the only one in `privileged_clusters`
    Will this cause trouble?

If deduplication is happening here, and it doesn't matter which connection parameter is picked, I suggest we make it clear like this:


```python
def get_unique_connection_parameters(connection_parameters):
    unique_by_cluster_name = {c.cluster_name: c for c in connection_parameters}
    return list(unique_by_cluster_name.values())
```

```python
        all_clusters = get_unique_connection_parameters(connection_parameters)
        privileged_clusters = get_unique_connection_parameters(
            (c for c in connection_parameters if (c.is_cluster_admin or cluster_admin))
        )

        threaded.run(
            self._init_oc_client,
            all_clusters,
            self._thread_pool_size,
            privileged=False,
        )
        threaded.run(
            self._init_oc_client,
            privileged_clusters,
            self._thread_pool_size,
            privileged=True,
        )
```

fishi0x01: Case 1: shouldnt be a problem - the cluster will be the same with `is_cluster_admin` and w/o, i.e., it doesn't matter which one wins (note, currently it is random anyways).

Case 2: the last one wins is fine here.

Basically this already happens in the original implementation mentioned [here](https://github.com/app-sre/qontract-reconcile/blob/ee0f52bf7f2551d00833d6858861ba16d1dd2357/reconcile/utils/oc.py#L1586-L1590). I am just porting that behavior, which is working fine for several months now in production - so Im very confident this will be ok.

Will add a function as you suggest to make it more clear - good idea!

fishi0x01: About the `is_cluster_admin` attribute: it is not considered in the rest of the code. This is the only part where it is used, i.e., to distinguish whether the cluster is initialized with `privileged` flag in the `threaded` call or w/o.

fishi0x01: I will highlight the intent through the variable naming instead then.

fishi0x01: I renamed the variables now and added a comment on why it is important to iterate over all parameters.

hemslo: `get_unique_connection_parameters` didn't change the execution order, should keep the same order as raw `connection_parameters`, `(c for c in connection_parameters if (c.is_cluster_admin or cluster_admin))` is iterator, so it didn't change order either.

fishi0x01: Ohhh you are right! I oversaw the extra brackets. Yeah that should work :)
Added your suggestion.

fishi0x01: I believe that ideally all this shoud be done in [oc_connection_parameters.py](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/oc_connection_parameters.py). For now it will happen in `OCMap` to keep it as close to the original implementation as possible. Once the migration away from old `OC_Map` is finished, I will take a look to reduce the logic needed here.

fishi0x01:
The more I think about it the more I am convinced that [this PR](https://github.com/app-sre/qontract-reconcile/pull/3248) can also be solved here. Basically shifting this whole logic away from `OCMap`, towards `oc_connection_parameters`. Will close the other PR and put this in draft for now.

reconcile/utils/oc_map.py
```diff
@@ -61,30 +60,22 @@ def __init__(
...
-            c
-            for c in connection_parameters
-            if not (c.is_cluster_admin or cluster_admin)
-        ]
+        unprivileged_connections = (
+            c for c in connection_parameters if not c.is_cluster_admin
+        )
+        privileged_connections = (
+            c for c in connection_parameters if c.is_cluster_admin
+        )
```
lechuk47, line 68: Is not this behaving the same way as before? Where the non-privileged connection is created when `c.is_cluster_admin` is set?

fishi0x01: Yeah the logic that fixes the bug is moved to oc_connection_parameters: https://github.com/app-sre/qontract-reconcile/pull/3249/files#diff-d532c0b963df575b717e917a3238abec2f0e7a38ee9aff9efb717221bb533f0bR226-R236

fishi0x01: As discussed offline, I removed redundant filtering block. Thanks for the suggestion :bow:

lechuk47 (approved):
LGTM

PR #3250, opened 2023-02-28
[integration-runtime] detect new shard when moving config from shard to shard

geoberle:
when a configuration items is moved from one shard to another one, the integration runtime only detects the old shard as the affected one, instead of detecting both.

https://issues.redhat.com/browse/APPSRE-7204

### DIFF
--- a/reconcile/test/runtime/test_utils_desired_state_diff.py
+++ b/reconcile/test/runtime/test_utils_desired_state_diff.py
@@ -220,7 +220,10 @@ def test_extract_diffs_with_exception():
 #
 
 
-def test_find_changed_shards_removed_diff():
+def test_config_removed_from_shard():
+    """
+    when something is removed from a shard, that shard is affected
+    """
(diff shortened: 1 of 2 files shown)


reconcile/test/runtime/test_utils_desired_state_diff.py
```diff
@@ -271,19 +280,47 @@ def test_find_changed_shards_change_diff():
...
+        current_desired_state={
+            "data": [
+                {"shard": "a", "value": "a", IDENTIFIER_FIELD_NAME: "a"},
+                {"shard": "c", "value": "b", IDENTIFIER_FIELD_NAME: "b"},
+            ]
+        },
+    ).affected_shards == {
+        "b",  # the old one
+        "c",  # the new one
+    }
```
patjlm, line 321: ❤️

patjlm (approved):
Thanks!

PR #3255, opened 2023-03-01
Bugfix for gitlab for compliance

janboll:
This fixes a bug in gitlab fork compliance that will falsely report devtools-bot missing in cases where the user was not added to the repository directly but got permission by adding the grup instead. Prominent case for this is when devtools bot itself is creating MRs directly on the repository in question, this is common for automated MRs.

Required for: https://issues.redhat.com/browse/APPSRE-6369

### DIFF
--- a/reconcile/gitlab_fork_compliance.py
+++ b/reconcile/gitlab_fork_compliance.py
@@ -2,7 +2,6 @@ import logging
 import sys
 
 from gitlab import MAINTAINER_ACCESS
-from gitlab.exceptions import GitlabGetError
 
 from reconcile import queries
 from reconcile.utils.gitlab_api import GitLabApi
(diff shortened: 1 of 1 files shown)


reconcile/gitlab_fork_compliance.py
```diff
@@ -84,20 +83,12 @@ def check_branch(self):
...
     def check_bot_access(self):
-        # The bot needs access to the fork project
-        try:
-            self.src = GitLabApi(
-                self.instance,
-                project_id=self.mr.source_project_id,
-                settings=self.settings,
-            )
-            project_bot = self.src.project.members.get(self.gl_cli.user.id)
-        except GitlabGetError:
```
fishi0x01, line 95: I assume this is not an issue, but could this still be thrown on `self.src.get_project_maintainers()`?

fishi0x01 (approved):
Deletes more than it adds. Great!!!! :)
Just one small question that is likely not an issue.

reconcile/gitlab_fork_compliance.py
janboll: Good catch! That makes the change even smaller....

PR #3256, opened 2023-03-02
Jira + Gitlab Client Improvements for change-control-automation

TGPSKI:
This MR improves GitLab and Jira Clients in qontract-reconcile to support features needed by change-control-automation.

Jira client:

* adds custom JQL query support to get_issues
* adds ability to specify assignee in create_issue
* add delete_issue method (use case is deleting tickets for MRs that no longer exist)

GitLab client:

* adds get_merge_request_first_commit_hash method

https://issues.redhat.com/browse/APPSRE-6934

### DIFF
--- a/reconcile/utils/jira_client.py
+++ b/reconcile/utils/jira_client.py
@@ -39,12 +39,16 @@ class JiraClient:
             self.server, token_auth=token_auth, timeout=(read_timeout, connect_timeout)
         )
 
-    def get_issues(self, fields: Optional[Mapping] = None) -> list[Issue]:
+    def get_issues(self, fields: Optional[Mapping] = None, custom_jql: Optional[str, Any] = None) -> list[Issue]:
         block_size = 100
         block_num = 0
 
         all_issues: list[Issue] = []
-        jql = "project={}".format(self.project)
+
+        jql = f"project={self.project}"
+        if custom_jql:
+            jql = f"{jql} and {custom_jql}"
+        
         kwargs: dict[str, Any] = {}
         if fields:
             kwargs["fields"] = ",".join(fields)
--- a/reconcile/utils/jira_client.py
+++ b/reconcile/utils/jira_client.py
@@ -40,7 +40,7 @@ class JiraClient:
         )
 
     def get_issues(
-        self, fields: Optional[Mapping] = None, custom_jql: Optional[str, Any] = None
+        self, fields: Optional[Mapping] = None, custom_jql: Optional[str] = None
     ) -> list[Issue]:
         block_size = 100
         block_num = 0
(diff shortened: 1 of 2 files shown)


reconcile/utils/jira_client.py
```diff
@@ -74,14 +80,19 @@ def create_issue(
...
+        if not issueType:
+            issueType = {"name": "Task"}
         issue = self.jira.create_issue(
             project=self.project,
             summary=summary,
             description=body,
             labels=labels,
-            issuetype={"name": "Task"},
+            issuetype=issueType,
+            assignee=assignee,
```
bkez322, line 95: I'm not sure if this line will work as expected.

bkez322: https://atlassian-python-api.readthedocs.io/jira.html#manage-issues

```
# Assign issue to user
jira.assign_issue(issue_key, account_id)
```

TGPSKI: Based on what I've read, project,summary,description,and labels are all required attributes for create_issue. All other fields are supported, including specifying an assignee.

https://community.atlassian.com/t5/Jira-Software-questions/Python-API-Create-issue-Assignee-field-is-not-populated/qaq-p/2014457

bkez322: I'm having trouble finding documentation for the `create_issue` function to understand if `assignee` is a valid parameter.

Assuming it is, my assumption from how I see "accounts" referenced in other function in the docs, you'll need to use a user's account_id, rather than a user's username.

Is that going to be okay with how this function is intended to be consumed?

TGPSKI: I would call this a "thin" client, as it doesn't concern itself too much with the arguments into the methods itself, it just passes them in to the API calls assuming you know what is supported and what is not. 

https://jira.readthedocs.io/examples.html#issues

I'm looking at the dict input examples as reference. 

And this note:

```
Project, summary, description and issue type are always required when creating issues. Your Jira may require additional fields for creating issues; see the jira.createmeta method for getting access to that information.
```

Off topic, but is that what a thin client really is? Or is there a definition?

bkez322: These are just FYI's to make sure you're aware of, in case this alters the way you want this particular MR to look.

Otherwise, there's nothing inherently wrong about this MR, and LGTM.

TGPSKI: Assigning based on username is supported from what I can tell. This support thread from Atlassian is not the best, but I used it as a data point giving me confidence that we can assign in "create_issue".

bkez322: This is pretty confused.

I'm reading directions like this that supposedly how to check your "account ID", but I don't actually see an account ID.

https://community.atlassian.com/t5/Jira-questions/how-to-find-accountid/qaq-p/1111436

---

Then, according to answers like this, Jira-on-prem doesn't use an account-id?

https://community.atlassian.com/t5/Jira-questions/How-to-find-account-id-on-jira-on-premise/qaq-p/1168652

So I guess that means we're using Jira-on-prem and we don't have account-ids?

In which case, I have no idea what you're supposed to actually use with this library for assignee. Hopefully using a "username" works after all because I guess that's all we can use lol.

bkez322 (approved):

TGPSKI:
Closed in lieu of https://github.com/app-sre/qontract-reconcile/pull/3392

PR #3258, opened 2023-03-02
Fix race condition in tox

fishi0x01:
- `py39` and `generate` test envs have a race condition when executed in parallel
- `set -e` is required for CI to fail


chassing (approved):
❤️

PR #3260, opened 2023-03-02
Add cloudflare record to enable deletion whitelist.

OliviaHY:
**Changes:**
Add cloudflare_record to enable deletion whitelist so that we don't need deletionApproval for each record.
**Validation:**
Locally delete zone in dev-data with 
```
deletionApprovals:
- type: cloudflare_zone 
  name: devshift1
  expiration: '2023-12-31'
- type: cloudflare_zone_settings_override 
  name: devshift1
  expiration: '2023-12-31'
  ```

### CODE REVIEW

File: reconcile/utils/terraform_client.py [hunk 1]
```diff
@@ -267,6 +267,7 @@ def log_plan_diff(
         always_enabled_deletions = {
             "random_id",
             "aws_lb_target_group_attachment",
+            "cloudflare_record",
```
steveteahan, line 270: Maybe just worth a quick comment that zones can contain hundreds or more records, so it's not practical to require adding record to deletionApprovals?

reconcile/utils/terraform_client.py
```diff
@@ -267,6 +267,7 @@ def log_plan_diff(
         always_enabled_deletions = {
             "random_id",
             "aws_lb_target_group_attachment",
+            "cloudflare_record",  # This is because a zone can contain up to one thousand records and it's not ot practical to require adding record to deletionApprovals
```
steveteahan, line 270:
```suggestion
            "cloudflare_record",  # This is because a zone can contain up to one thousand records and it's not practical to require adding each record to deletionApprovals
```

steveteahan (approved):

PR #3261, opened 2023-03-02
enable always_reconnect by default

BumbleFeng:
auto try reconnecting failed node


rporres (approved):

BumbleFeng:
replace by https://github.com/app-sre/qontract-reconcile/pull/3283

PR #3262, opened 2023-03-03
Add delete for removed ldap users from app-sre/infra

janboll:
Adding a MergeRequest template, that deletes Users deleted from app-sre/infra when tey get deleted from ldap. 

Related issue: https://issues.redhat.com/browse/APPSRE-6369


reconcile/ldap_users.py
```diff
@@ -75,5 +81,8 @@ def run(dry_run, gitlab_project_id=None):
         logging.info(["delete_user", username])
 
         if not dry_run:
-            mr = CreateDeleteUser(username, paths)
-            mr.submit(cli=mr_cli)
+            mr = CreateDeleteUserAppInterface(username, paths)
+            mr.submit(cli=mr_cli_app_interface)
+
+            mr_infra = CreateDeleteUserInfra(username)
+            mr_infra.submit(cli=mr_cli_infra)
```
fishi0x01, line 88: This writes to the same file, i.e., wouldn't it create a lot of merge conflicts if a list of users gets deleted? Every user will issue a new MR.

fishi0x01: Basically Im wondering if it would make sense to pass a list of users here, to keep the number of potential merge conflicts lower.

fishi0x01: Not sure if its feasible, but Im wondering if there is a nice way to have a dedicated file in infra per user (similar to how we do in app-interface). The user file can have a `is_deleted` flag. That way there won't be any merge conflicts.

janboll: Oh, excellent point! I'll change it to delete all users in one MR.

janboll: Good point. I think it would only create one MR per integration run.... I'll test it, but nethertheless, one MR for all users will be a small change, let me take a look

janboll: @fishi0x01 all users are now handled in one MR: https://gitlab.cee.redhat.com/app-sre/infra/-/merge_requests/633

fishi0x01 (approved):
LGTM :+1:

cubismod (approved):

PR #3266, opened 2023-03-05
[app-interface state] initialize via builder function

geoberle:
change all initialization of the `reconcile.utils.state.State` class to be handled via the `init_state` function. this is a small minimal preparational step to make additional improvements in upcoming PRs easier where we will fully inject all dependencies to the `init_state` function (e.g. `SecretReader`) while getting  rid of other dependencies (e.g. `accounts`) and add other means of providing configuration data for state bucket initialization.

This change could have been bigger, touching `SecretReader` initializiation in various places, but the scope of this PR has been kept small on purpose.

https://issues.redhat.com/browse/APPSRE-7226

### DIFF
--- a/reconcile/utils/saasherder.py
+++ b/reconcile/utils/saasherder.py
@@ -49,7 +49,7 @@ from reconcile.utils.openshift_resource import (
     fully_qualified_kind,
 )
 from reconcile.utils.secret_reader import SecretReader
-from reconcile.utils.state import State
+from reconcile.utils.state import init_state
 
 TARGET_CONFIG_HASH = "target_config_hash"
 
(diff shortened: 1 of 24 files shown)


reconcile/utils/saasherder.py
```diff
@@ -174,7 +174,7 @@ def __init__(
         integration_version,
         settings,
         jenkins_map=None,
-        accounts=None,
+        initialise_state=False,
```
chassing, line 177: I would prefer a `state: Optional[State]` instead of `initialise_state: bool`

chassing (commented):
Very nice refactoring ❤️

reconcile/utils/saasherder.py
geoberle: me too. this PR is minimal on purpose so things like injecting`SecretReader` and even the`State` itself will happen in the followup MR if that is ok with you

chassing (approved):

PR #3268, opened 2023-03-06
[Draft] Add support for vault auth using kubernetes

dwelch0:
APPSRE-7074
SDE-2558

**Summary:** adds support for kubernetes auth option within vault client and explicitly defines service account volume projection in order to overwrite default expiration.

**Changes:**
* Adds option for kubernetes auth to be utilized by the Vault client
  * Defaults to approle when specific kubernetes auth attributes are not set within toml or passed as arguments when initializing `VaultClient`
* Adds environment variable and explicit service account volume projection
  * this is enabled by default but set to expiry of `3607`. This equates to `1 year`.


esron (approved):

dwelch0:
Breaking this into separate MRs to isolate logic change from template updates

PR #3269, opened 2023-03-06
Update vault query to use fragment so it can get all the fields, and …

OliviaHY:
#### What:
* Update vault query to use fragment so it can get all the fields
* Use typed new method to get vault secret.
#### Why:
Fixing the [error](https://ci.int.devshift.net/job/service-app-interface-gl-pr-check/211799/artifact/temp/reports/index.html) from this [MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/60450/diffs).

#### Validation:
Local dry run with a new Cloudflare account that referencing a new versioned vault secret as apiCredentials.

### DIFF
--- a/reconcile/gql_definitions/introspection.json
+++ b/reconcile/gql_definitions/introspection.json
@@ -21983,6 +21983,18 @@
                             },
                             "isDeprecated": false,
                             "deprecationReason": null
+                        },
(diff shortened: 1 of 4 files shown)


geoberle (approved):
LGTM

PR #3270, opened 2023-03-07
APPSRE-6609 OCmap for openshift-namespaces integration

fishi0x01:
- `openshift-namespaces` integration now fully typed
- new `OCMap` used
- add `oc_filters.py` -> start moving logic away from `openshift_base.py` and adding tests for it. Follow-up PRs will bisect more `openshift_base.py` functions to make it work with `OCMap`


chassing (approved):
Great!

PR #3272, opened 2023-03-07
Draft: fix mypy issues

fishi0x01:
**DRAFT**

New mypy release doesnt like too many positional unnamed args. Or too less :D


fishi0x01:
For now we stick with mypy 1.0.1 due to bugs with pydantic https://github.com/app-sre/qontract-reconcile/pull/3272

fishi0x01:
Close in favor of https://github.com/app-sre/qontract-reconcile/pull/3288

PR #3274, opened 2023-03-07
[statuspage] explicit datetime conversion

geoberle:
pydantic is able to implicitly convert datetime strings to proper datetimes, but explicit is safer so we can control what format we accept.


kwilczynski (approved):
@geoberle, looks good!

PR #3277, opened 2023-03-08
ocm-clusters: update version when not set

patjlm:


maorfr (approved):

maorfr:
seems to have been lost here: https://github.com/app-sre/qontract-reconcile/pull/2425/files#diff-c6864bdf1ca61fad743e15096a363fa2ca01e50383721bef22ceb2a711179419L110

PR #3278, opened 2023-03-08
Publish all logs

janboll:
This was added under assumption there is only one container. This can cause issues if there are more containers. Export logs for all containers as default.

https://issues.redhat.com/browse/APPSRE-7037


patjlm (approved):
According to comments in https://issues.redhat.com/browse/APPSRE-7037 this option is available in oc 4.10 and above.

janboll:
Right, the oc version provided by the base image supports this:
```
(venv) ➜  ~ oc version
Client Version: 4.10.15
error: You must be logged in to the server (Unauthorized)
(venv) ➜  ~ oc logs --help |grep all-containers
      --all-containers=false: Get all containers' logs in the pod(s).
```

rporres (approved):

BumbleFeng (approved):

PR #3284, opened 2023-03-08
Set Cloudflare provider rps based on zone

OliviaHY:
#### What
*  Adjust Cloudflare terraform provider's rps based on the number of records in the zone.
* Catch the 'referenced before assignment' error with a better error message.
#### Validation:
Ran the terraform-cloudflare-dns locally against dev-data and was able to see the provider setting in the state file, and was able to `terraform apply`
Dry ran terraform-cloudflare-resources locally against dev-data.

### DIFF
--- a/reconcile/terraform_cloudflare_dns.py
+++ b/reconcile/terraform_cloudflare_dns.py
@@ -7,6 +7,7 @@ from collections.abc import (
 from typing import (
     Any,
     Optional,
+    Sequence,
 )
 
 from reconcile.gql_definitions.terraform_cloudflare_dns import (
@@ -15,6 +16,7 @@ from reconcile.gql_definitions.terraform_cloudflare_dns import (
 from reconcile.gql_definitions.terraform_cloudflare_dns.terraform_cloudflare_zones import (
     AWSAccountV1,
     CloudflareAccountV1,
+    CloudflareDnsRecordV1,
     CloudflareDnsZoneQueryData,
     CloudflareDnsZoneV1,
 )
@@ -41,6 +43,7 @@ from reconcile.utils.terraform_client import TerraformClient
 from reconcile.utils.terrascript.cloudflare_client import (
     DEFAULT_CLOUDFLARE_ACCOUNT_2FA,
     DEFAULT_CLOUDFLARE_ACCOUNT_TYPE,
+    DEFAULT_PROVIDER_RPS,
     CloudflareAccountConfig,
     TerraformS3BackendConfig,
     TerrascriptCloudflareClient,
@@ -176,6 +179,27 @@ class TerraformCloudflareDNSIntegration(
         )
 
 
+def get_cloudflare_provider_rps(
+    records: Optional[Sequence[CloudflareDnsRecordV1]],
+) -> int:
+    """
+    Setting Cloudlare Terraform provider's RPS based on the size of the zone to improve performance of MR checks.
+    Specifically it was observed that 1000 records zone will result in around 250 seconds build time, and it become
+    problematic for MR merge throughput when exceeding 5 minutes. Therefore setting rps to 2 for smaller zone to
+    save throttle quota, and 6 for the large zones so MR checks won't take more than 250 seconds.
+    """
+
+    if not records:
+        return DEFAULT_PROVIDER_RPS
+    size = len(records)
+    if size <= 50:
+        return 2
+    elif size <= 1000:
+        return DEFAULT_PROVIDER_RPS
+    else:
+        return 6
+
+
 def create_backend_config(
     secret_reader: SecretReaderBase,
     aws_acct: AWSAccountV1,
@@ -196,6 +220,8 @@ def create_backend_config(
     integrations = tf_state.integrations or []
     for i in integrations or []:
         name = i.integration
+
+        bucket_key = bucket_name = bucket_region = None
         if name.replace("-", "_") == integration_name:
             # Currently terraform-state-1.yml can only have one bucket
             # but multiple integrations, which means without schema changes
@@ -271,10 +297,13 @@ def build_clients(
             integration_name=integration_name,
         )
 
+        rps = get_cloudflare_provider_rps(zone.records)
+
         ts_config = create_cloudflare_terrascript(
-            cf_acct_config,
-            aws_backend_config,
-            cf_acct.provider_version,
+            account_config=cf_acct_config,
+            backend_config=aws_backend_config,
+            provider_version=cf_acct.provider_version,
+            provider_rps=rps,
             is_managed_account=False,
         )
 
--- a/reconcile/utils/terrascript/cloudflare_client.py
+++ b/reconcile/utils/terrascript/cloudflare_client.py
@@ -44,6 +44,7 @@ TMP_DIR_PREFIX = "terrascript-cloudflare-"
 DEFAULT_CLOUDFLARE_ACCOUNT_TYPE = "standard"
 DEFAULT_CLOUDFLARE_ACCOUNT_2FA = False
 DEFAULT_IS_MANAGED_CLOUDFLARE_ACCOUNT = True
+DEFAULT_PROVIDER_RPS = 4
 
 
 @dataclass
@@ -62,6 +63,7 @@ def create_cloudflare_terrascript(
     account_config: CloudflareAccountConfig,
     backend_config: TerraformS3BackendConfig,
     provider_version: str,
+    provider_rps: int = DEFAULT_PROVIDER_RPS,
     is_managed_account: bool = True,
 ) -> Terrascript:
     """
@@ -98,6 +100,7 @@ def create_cloudflare_terrascript(
         "cloudflare": {
             "source": "cloudflare/cloudflare",
             "version": provider_version,
+            "rps": provider_rps,
         }
     }
 
@@ -259,7 +262,10 @@ class TerrascriptCloudflareClientFactory:
         backend_config = cls._create_backend_config(tf_state_s3, key, secret_reader)
         cf_acct_config = cls._create_cloudflare_account_config(cf_acct, secret_reader)
         ts_config = create_cloudflare_terrascript(
-            cf_acct_config, backend_config, cf_acct.provider_version, is_managed_account
+            cf_acct_config,
+            backend_config,
+            cf_acct.provider_version,
+            is_managed_account=is_managed_account,
         )
         client = TerrascriptCloudflareClient(ts_config)
         return client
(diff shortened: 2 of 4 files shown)


reconcile/terraform_cloudflare_dns.py
```diff
@@ -176,6 +176,16 @@ def get_desired_state_shard_config(self) -> DesiredStateShardConfig:
         )
 
 
+def get_cloudflare_provider_rps(zone_size: int) -> int:
+    if zone_size <= 50:
+        rps = 2
+    elif zone_size <= 1000:
+        rps = 4
```
kwilczynski, line 183: The 4 r/s is the default.

I wonder if it only made sense to limit for when a zone has over 1000 resource records?

Hopefully, Terraform parallelism (or an attempt at when resources are independent) does not stand in the way here. I suppose the entire provider has a single client.

reconcile/terraform_cloudflare_dns.py
```diff
@@ -196,6 +206,10 @@ def create_backend_config(
     integrations = tf_state.integrations or []
     for i in integrations or []:
         name = i.integration
+        # bucket_key = f"{integration_name}-{cf_acct.name}-{zone}.tfstate"
```
kwilczynski, line 209: Any use for the commented-out block?

reconcile/terraform_cloudflare_dns.py
```diff
@@ -271,10 +286,16 @@ def build_clients(
             integration_name=integration_name,
         )
 
+        if zone.records is not None:
```
kwilczynski, line 289: Would this work?

```python
rps = 4
if zone.records:
    rps = get_cloudflare_provider_rps(len(zone.records))
```

You could also make it so that the default Cloudflare's SDK limit applies if there aren't any zones. Or, make the helper return default (4 r/s) if the list if None or empty.

reconcile/terraform_cloudflare_dns.py
```diff
@@ -206,17 +220,18 @@ def create_backend_config(
...
         backend_config = TerraformS3BackendConfig(
             aws_acct_creds["aws_access_key_id"],
             aws_acct_creds["aws_secret_access_key"],
             bucket_name,
             bucket_key,
             bucket_region,
         )
-    else:
-        raise ValueError(f"No state bucket config found for account {aws_acct.name}")
+    except UnboundLocalError:
```
kwilczynski, line 231: Something like:

```python
if all([bucket_name, bucket_key, bucket_region]):
```

Wouldn't work? I am not sure about `UnboundLocalError` usage.

reconcile/utils/terrascript/cloudflare_client.py
```diff
@@ -62,6 +62,7 @@ def create_cloudflare_terrascript(
     account_config: CloudflareAccountConfig,
     backend_config: TerraformS3BackendConfig,
     provider_version: str,
+    provider_rps: int = 4,
```
kwilczynski, line 65: There are a few places where this default is set at the moment.

reconcile/terraform_cloudflare_dns.py
OliviaHY: Hm I'm not understanding your question. Do you mean for records <= 1000 so no setting of rps = 2?

reconcile/utils/terrascript/cloudflare_client.py
OliviaHY: I was thinking about leaving the exacting of it as constant later since there is refactor pending, but you are right, it's too scattered.

reconcile/terraform_cloudflare_dns.py
OliviaHY: I like the idea of keeping the check in helper

reconcile/terraform_cloudflare_dns.py
OliviaHY: No. Removing.

reconcile/terraform_cloudflare_dns.py
OliviaHY: No, the problem this is trying to solve is a "UnboundLocalError: local variable 'bucket_name' referenced before assignment" error. 

Is your concern that this is catching a very specific error? I thought about it too but considering `TerraformS3BackendConfig ` is just a data class, and any exception should be raised anyway.

steveteahan:
Were you able to do any performance testing with the rps setting for larger zones to confirm the operations finished more quickly than with the default?

reconcile/terraform_cloudflare_dns.py
```diff
@@ -176,6 +178,18 @@ def get_desired_state_shard_config(self) -> DesiredStateShardConfig:
         )
 
 
+def get_cloudflare_provider_rps(records: Optional[list[CloudflareDnsRecordV1]]) -> int:
```
steveteahan, line 181: I'd suggest a docstring that explains why this is being done. Also, could this be `Iterable`?

reconcile/terraform_cloudflare_dns.py
steveteahan: I agree that starting to catch `UnboundLocalError` feels a little messy to me. What if you assumed this was about `bucket_name`, but it was that `aws_acct_creds` wasn't set?

Could you just assign the variables before this? Or anything else that wouldn't involve only assigning variables under a certain condition, but then trying to reference them after?

```python
bucket_key = bucket_name = bucket_region = None
```

And then leave the original check, or do what @kwilczynski suggested?

OliviaHY: Updated. Can't use 
```
if all([bucket_name, bucket_key, bucket_region]):
```
because type threw error about Optional[str]

OliviaHY:
> Were you able to do any performance testing with the rps setting for larger zones to confirm the operations finished more quickly than with the default?

I did, didn't see any improvement during when creating 1000 records with 4 and 6 rps. My account is free tier that has the 1200 per 5 minutes rate limits, which is probably why.

reconcile/terraform_cloudflare_dns.py
OliviaHY: FYI. We have the 5000 per 5 minutes rate limit for that account which is about 16 rps so from 4 to 6 at provider level should be having a performance relief for this.

reconcile/terraform_cloudflare_dns.py
hemslo: Note if `records` is empty list, the rps will be `DEFAULT_PROVIDER_RPS`, is this expected?
Btw to make branch condition easier to understand, can do early return without mutations, like this:

```python
    if not records:
        return DEFAULT_PROVIDER_RPS
    size = len(records)
    if size <= 50:
        return 2
    elif size <= 1000:
        return DEFAULT_PROVIDER_RPS
    else:
        return 6
```

To make it more pythonic, instead of multiple if conditions, can use a config like this:

```python
RPS_TIERS = [
    {"max": 50, "rps": 2},
    {"max": 1000, "rps": DEFAULT_PROVIDER_RPS},
    {"max": math.inf, "rps": 6},
]


def get_cloudflare_provider_rps(records: Optional[list[CloudflareDnsRecordV1]]) -> int:
    """
    Setting Cloudlare Terraform provider's RPS based on the size of the zone to improve performance.
    """
    if not records:
        return DEFAULT_PROVIDER_RPS
    size = len(records)
    return next(t["rps"] for t in RPS_TIERS if size <= t["max"])
```

Since there are just 3 different branches, the first approach is good enough.

kwilczynski: @hemslo, that is nice.

However, a bit too much just for three choices and a default. This code does not need to be smart in any way.

reconcile/terraform_cloudflare_dns.py
OliviaHY: The reason we are doing this is because the amount of records in a zone increase the amount of request making to Cloudflare provider APIs, therefore to complete a build. This affect not only when people change that large zone, it also affect pr-check when promoting qontract-reconcile for example. And previous we were seeing around 250 seconds for a 1000 records zone, and that was where we drew the line. But there was a strong request for us to manage a zone with up to 1300 records now, we are trying to accommodate to that while not sacrifice the performance too much. I.e not to let a integration run more than 250 seconds. And that's how we decided on 1500 records upper limit with 6 rps. 

>  if we have an issue with 4 r/s when a zone has 50 RRs, then wouldn't a limit of 4 be an issue for a zone with 999 RRs?
So the problem really wasn't that 50 records were taking too long, it was 1300 will take a long time. And 2 rps for 50 records was just to save our throttle quota and those small zone (which will probably be most of them )are still gonna finish within a short amount time.

> Also, if there are no RRs or a zone has over 1000 RRs, then we set the limit to 6 r/s, but why? Does Cloudflare throttle all of the API requests or just for specific endpoints? So, for example, anything else but zone-related operations can achieve more r/s (or have higher parallelism).
We will have zone with more than 1000 records and reasons are like what i mentioned. I was under the impression that this rate limit is global based on https://developers.cloudflare.com/fundamentals/api/reference/limits/ but I don't have a perfect answer to how specifically Cloudflare throttle API requests. 

> With the default of 4 for everything, are we having problems with extensive zones only? That would be my uneducated assumption.
Hope the background info I gave above answers this.

reconcile/terraform_cloudflare_dns.py
OliviaHY: Good point about the early return @hemslo. I too think the data structure is a bit overkill. Very smart though.

I couldn't use `Iterable` because the use of len(), since `Iterable` is not "Sized". I could use `Sequence` if that's something you feeling very strongly about @steveteahan .

reconcile/terraform_cloudflare_dns.py
kwilczynski: [...]
> Hope the background info I gave above answers this.

I appreciate the background and additional details. The "how" was obvious (per the code), but the "why" wasn't.

Thank you!

reconcile/terraform_cloudflare_dns.py
```diff
@@ -176,6 +178,22 @@ def get_desired_state_shard_config(self) -> DesiredStateShardConfig:
...
+    """
+    Setting Cloudlare Terraform provider's RPS based on the size of the zone to improve performance.
```
steveteahan, line 183: If someone came to this code in 2 years with no context, would they understand why 2 was chosen for small zones, or why large zones need 6? You can keep it pretty general in terms of the performance of MR checks, exceeding 5 minutes being problematic for MR merge throughput, etc.

steveteahan (approved):

reconcile/terraform_cloudflare_dns.py
steveteahan: Sorry, it's kind of my canned feedback and I didn't see the `len()`, but yes `Collection` or `Sequence` would be perfectly fine as well. The idea is just that the type is general enough now that if it weren't a `list`, your code can still support it.

PR #3285, opened 2023-03-09
Bump pyOpenSSL to 23.0 to reduce heap memory

hemslo:
High memory consumption is caused by old openssl packages, bumping `pyOpenSSL` to `23.0` and removing `cryptography` pinned version can greatly reduce heap size.

This change reverted https://github.com/app-sre/qontract-reconcile/pull/2369, need to keep an eye on cert issue after deployment.

APPSRE-4141

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
         "pypd>=1.1.0,<1.2.0",
         "Jinja2>=2.10.1,<3.2.0",
         "jira~=3.1",
-        "pyOpenSSL~=21.0",
+        "pyOpenSSL~=23.0",
         "ruamel.yaml>=0.16.5,<0.18.0",
         "terrascript==0.9.0",
         "tabulate>=0.8.6,<0.9.0",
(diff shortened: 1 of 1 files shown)


kwilczynski (approved):
@hemslo, looks good!

Thank you for the fix!

janboll (approved):

setup.py
```diff
@@ -60,8 +60,6 @@
         # this is really needed only in lint and type validations.
         # Is there any better place to put this in?
         "packaging~=21.3",
-        # Needed to fix issues with router's certificates
-        "cryptography==36.0.2",
```
patjlm, line 64: Don't we want to keep at least a light pinning, avoiding unexpected major version bump?

eg `cryptography~=39.0`

patjlm: This is also important for idempotent builds on a commit ref.

hemslo: Since this is not our direct dependency, and `pyOpenSSL` is [managing this](https://github.com/pyca/pyopenssl/blob/main/setup.py#L101), do we still need to pin it?

patjlm: Great,  did not see that, sorry. Good for me then.

patjlm (approved):

setup.py
hemslo: To achieve idempotent builds on a commit ref, we either need a freeze requirement.txt or use something like `poetry`

rporres:
This closes https://issues.redhat.com/browse/APPSRE-5475

PR #3289, opened 2023-03-09
introduce ocm orgId as mandatory field

geoberle:
the new `ocmId` field will be used for cluster filtering within an OCM org instead of relying on the `ocm whoami` detected org id of the used SA. this enables us to use OCM SAs that are not bound to an org (which is a very common thing anyways) and that can act within multiple orgs.

schema change: https://github.com/app-sre/qontract-schemas/pull/404
part of https://issues.redhat.com/browse/APPSRE-7252


fishi0x01 (approved):
LGTM

chassing (approved):

patjlm (approved):

PR #3292, opened 2023-03-09
Overwrite default service account volume projection

dwelch0:
APPSRE-7074
SDE-2558

[related design doc](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/59519)

**Changes:**
* Adds environment variable and explicit service account volume projection
  * This is enabled by default but set to expiry of 3607. This equates to 1 year. See [3607 behavior](https://stackoverflow.com/questions/69375195/kubernetes-projected-service-account-token-expiry-time-issue)

### DIFF
--- a/reconcile/jenkins/types.py
+++ b/reconcile/jenkins/types.py
@@ -49,9 +49,10 @@ class JenkinsWorkerFleet(BaseModel):
     num_executors: int = Field(2, alias="numExecutors")
     idle_minutes: int = Field(30, alias="idleMinutes")
     min_spare_size: int = Field(0, alias="minSpareSize")
+    max_total_uses: int = Field(-1, alias="maxTotalUses")
     no_delay_provision: bool = Field(False, alias="noDelayProvision")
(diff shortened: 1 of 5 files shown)


dwelch0:
Input appreciated on where `KUBE_SA_TOKEN_PATH` should be defined within template as this will determine which helm fixtures require updating

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -591,6 +607,8 @@ parameters:
   value: "0"
 - name: USER_ID
   value: dummy
+- name: KUBE_SA_TOKEN_PATH
+  value: '/var/run/secrets/kubernetes.io/serviceaccount'
```
cubismod, line 611: So just to clarify, this value should always be the same?

cubismod (approved):
LGTM!

helm/qontract-reconcile/templates/template.yaml
dwelch0: Correct. This is the current path utilized by default projection

PR #3293, opened 2023-03-09
Fix awsAccount -> awsAccounts as it is a list

esron:
Relates to https://github.com/app-sre/qontract-schemas/pull/415


cubismod:
Are there any other code level changes you'll need to make to ensure that integrations manager can handle multiple AWS accounts instead of just one?

esron:
> Are there any other code level changes you'll need to make to ensure that integrations manager can handle multiple AWS accounts instead of just one?

Oh, I see now. Yeah, I'm gonna need to update that.

geoberle (approved):

PR #3295, opened 2023-03-10
[qontract-cli] introduce selectorsyncset-managed-resources

maorfr:
with this qontract-cli command (which can be added to app-interface-output), we can get a mapping of resources that are managed on a customer cluster via a Hive SelectorSyncSet.

this command will provide output that will be helpful in answering the questions:
- is this resource created by a SelectorSyncSet?
- what is the name of that SelectorSyncSet?
- what SaaS file is deploying it?

since SaaS file names are unique, this information will lead to the OpenShift template that contains the SelectorSyncSet.

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -33,6 +33,10 @@ from reconcile.cli import config_file
 from reconcile.jenkins_job_builder import init_jjb
 from reconcile.prometheus_rules_tester import get_data_from_jinja_test_template
 from reconcile.slack_base import slackapi_from_queries
+from reconcile.typed_queries.app_interface_vault_settings import (
+    get_app_interface_vault_settings,
+)
+from reconcile.typed_queries.clusters import get_clusters
 from reconcile.utils import (
     amtool,
     config,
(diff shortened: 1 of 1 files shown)


tools/qontract_cli.py
```diff
@@ -1760,6 +1765,64 @@ def app_interface_merge_history(ctx):
...
+        "SelectorSyncSet_name",
+        "SaaSFile_name",
+        "kind",
+        "namespace",
+        "name",
+    ]
+    data = []
+    for c in clusters:
+        c_name = c.name
+        if c_name != "hivei01ue1":
```
geoberle, line 1795: why exclude this hive shard and not all the others as well?

patjlm: I think some selectorsyncsets apply on hive shards... inception kind of thing

tools/qontract_cli.py
```diff
@@ -1760,6 +1765,62 @@ def app_interface_merge_history(ctx):
...
+            continue
+        selectorsyncsets = oc.get_all("SelectorSyncSet")["items"]
+        for sss in selectorsyncsets:
+            try:
+                for resource in sss["spec"]["resources"]:
+                    kind = resource["kind"]
+                    namespace = resource["metadata"].get("namespace")
+                    name = resource["metadata"]["name"]
+                    item = {
+                        "cluster": c_name,
```
patjlm, line 1808: so that's the hive cluster, right, not the customer cluster. Would it make sense to add the clusterdeployment label selector from the SelectorSyncSet, so we can see (or guess) which customer clusters this applies to ?

maorfr: it probably would. i'm just here to "make it work". happy to see further contributions to "make it work right" and "make it beautiful" :heart:

maorfr: the reason i didn't add it is because it is usually not required when looking at a resource.

patjlm (approved):

PR #3297, opened 2023-03-10
manage route53 vpc associations on tgw peerings

patjlm:
https://issues.redhat.com/browse/APPSRE-7273

On TGW attachments to non-hive transit accounts, we may want to associate Private Hostedzones to our cluster VPC in order to resolve hostnames.

This is especially the case of the hypershift "service-transit-account", which contains private hosted zones for hypershift service clusters VPCE interfaces (privatelink endpoints).

Depends on https://github.com/app-sre/qontract-schemas/pull/408

### DIFF
--- a/reconcile/utils/aws_api.py
+++ b/reconcile/utils/aws_api.py
@@ -1098,13 +1098,7 @@ class AWSApi:  # pylint: disable=too-many-public-methods
                 zones = []
                 for page in paginator.paginate():
                     for zone in page["HostedZones"]:
-                        if zone["Config"]["PrivateZone"]:  # TODO: add filter on tags?
-                            zone_name = zone["Name"]
-                            # ignore high-level zone names, like p1.openshiftapps.com.
-                            # since zone names end with a dot, we get a length of 4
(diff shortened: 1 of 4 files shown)


reconcile/utils/aws_api.py
```diff
@@ -1091,6 +1092,23 @@ def get_tgws_details(
                 "region": region_name,
             }
 
+            if route53_associations:
+                route53 = self._account_route53_client(account["name"], region_name)
+                paginator = route53.get_paginator('list_hosted_zones')
+                zones = []
+                for page in paginator.paginate():
+                    for zone in page["HostedZones"]:
+                        if zone["Config"]["PrivateZone"]:  # TODO: add filter on tags?
```
patjlm, line 1101: We'll need an agreement on this tag filtering with @dustman9000

patjlm: no need for tag filtering. AVO will be able to delete the zone even it is associated to our VPCs and there should be no other hosted zones in the service transit account

PR #3298, opened 2023-03-10
PoC: Make Integrations Manager reusable

janboll:
Making integrations manager reusable:

- add upstream as parameter to filter out integrations and fetch ref from repo
- add caller to populate_current_state to filter resource inventory on caller. makes it possible to deploy multiple upstream in one namespace

Missing: setting image location, either pass in environment variable or add it to the integration schema

@geoberle & @maorfr

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -2400,6 +2400,12 @@ def vault_replication(ctx):
     help="git ref to use as IMAGE_TAG for given environment. example: '--image-tag-from-ref app-interface-dev=master'.",
     multiple=True,
     callback=parse_image_tag_from_ref,
+    envvar="INTEGRATIONS_MANAGER_IMAGE_TAG_FROM_REF"
+)
+@click.option(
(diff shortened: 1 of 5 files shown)


reconcile/cli.py
```diff
@@ -2400,6 +2400,12 @@ def vault_replication(ctx):
     help="git ref to use as IMAGE_TAG for given environment. example: '--image-tag-from-ref app-interface-dev=master'.",
     multiple=True,
     callback=parse_image_tag_from_ref,
+    envvar="INTEGRATIONS_MANAGER_IMAGE_TAG_FROM_REF"
+)
+@click.option(
+    "--upstream",
+    "-u",
+    help="specify upstream of managed integrations",
```
maorfr, line 2408: shouldn't this be read from integration files?

maorfr: is this related to why you need to use `caller`?

PR #3304, opened 2023-03-13
Close VaultClient after replication to fix memory leak

hemslo:
`vault-replication` integration is leaking memory, every time it runs it will leak fd (sockets to vault).
`requests.Session()` need to be closed after the client is no longer needed.
In this change, I implemented context manager for `_VaultClient`, so it can be used with `with` to proper close session.

APPSRE-7274


patjlm (approved):
LGTM

PR #3305, opened 2023-03-13
reenable pylint tests

chassing:
Enabled tests:

* import-error
* invalid-overridden-method
* implicit-str-concat
* no-else-continue
* no-else-raise
* no-else-return
* pointless-statement
* pointless-string-statement
* pointless-exception-statement
* raising-non-exception
* expression-not-assigned

APPSRE-7280


fishi0x01 (approved):
Nice cleanup! :bow:

PR #3309, opened 2023-03-13
add ability to remove resetPassword entries for users not in LDAP.

jonmosco:
Part of APPSRE-6367, requires https://github.com/app-sre/qontract-schemas/pull/411.

Extends the `ldap_users` integration to include removal of `resetPasswords` entries that no longer have a user associated in LDAP.  Schema changes are in https://github.com/app-sre/qontract-schemas/pull/411 to provide the paths to `user_maintenance.py`.  

Testing was done with qontract-reconcile-cli and qd.

### DIFF
--- a/reconcile/ldap_users.py
+++ b/reconcile/ldap_users.py
@@ -16,7 +16,7 @@ from reconcile.utils.mr.user_maintenance import PathTypes
 QONTRACT_INTEGRATION = "ldap-users"
 
 
-def init_users() -> dict:
+def init_users() -> list[dict[str, list]]:
     app_int_users = queries.get_users(refs=True)
 
     users = defaultdict(list)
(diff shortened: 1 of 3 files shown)


reconcile/utils/mr/user_maintenance.py
```diff
@@ -53,6 +54,7 @@ def process(self, gitlab_cli):
                     commit_message=self.title,
                     content=new_content,
                 )
+<<<<<<< HEAD
```
BumbleFeng, line 57: please fix this line and append your code change here

reconcile/ldap_users.py
```diff
@@ -16,7 +16,7 @@
 QONTRACT_INTEGRATION = "ldap-users"
 
 
-def init_users():
+def init_users() -> dict:
```
chassing, line 19: The return type hint should be 

```
list[dict[str, Union[str, list]]
```

PR #3312, opened 2023-03-14
Implement thread safe close for vault client

hemslo:
This change solves 2 issues continue from https://github.com/app-sre/qontract-reconcile/pull/3304:

1. `_auto_refresh_client_auth` will bring back closed connections due to it's running in another thread
2. `VaultClient` singleton is not thread safe, old client is not closed before new one assigned

APPSRE-7274


kwilczynski (approved):
@hemslo, nice catch! Re: lack of thread safety.

Thank you for the fix!

maorfr:
related to #1481

PR #3314, opened 2023-03-14
Speed up user-credentials output

janboll:
Speed up user-credentials by disabling tf setup if no accounts are supposed to be skipped

First line is the new implementation, second line the old:
```
python debugging_cli.py --config config.debug.toml get   9.44s user 0.38s system 52% cpu 18.767 total
python debugging_cli.py --config config.debug.toml get   143.24s user 17.03s system 56% cpu 4:45.50 total

```


fishi0x01 (approved):
LGTM :+1:

PR #3296, opened 2023-03-10
[saasherder] GQL classes and type hints

chassing:
Refactor `saasherder` to support generated GQL classes and add type hints

Ticket: [APPSRE-7152](https://issues.redhat.com/browse/APPSRE-7152)

PR #3315, opened 2023-03-14
[saasherder] GQL classes and type hints

chassing:
2nd try after reverting https://github.com/app-sre/qontract-reconcile/pull/3296

This PR fixes the `AutoPromoter` class. The `promotions` attribute must be json serializable because it's sent to SQS.


reconcile/utils/mr/auto_promoter.py
```diff
@@ -32,10 +36,20 @@ class ParentSaasConfigPromotion:
...
-    def __init__(self, promotions):
-        self.promotions = [p for p in promotions if p]
+    def __init__(
+        self, promotions: Union[Sequence[Promotion], Sequence[dict[str, Any]]]
+    ):
+        # !!! Attention !!!
+        # AutoPromoter is also initialized with promitions as dict by 'gitlab_mr_sqs_consumer'
+        # loaded from SQS message body, therefore self.promotions must be json serializable
+        self.promotions = [
+            p.dict() if isinstance(p, Promotion) else p for p in promotions
```
fishi0x01, line 46: Not sure, but would it be safer here to do `dict(by_alias=True)`?

fishi0x01 (approved):

reconcile/utils/mr/auto_promoter.py
chassing: `Promotion` isn't a qenerate generated GQL class, and it doesn't use aliases.

fishi0x01: Was wondering because it has some Protocol class as one attribute. Can't that potentially be a problem?

chassing: Good point! Thanks for that! 🙇 The `json.dump` was working, but not the `json.load` :(
I reworked this stuff and introduced a `SaasPromotion` Protocol, set `by_alias`, and extended the test fixtures.

PR #3316, opened 2023-03-14
[APPSRE-6626] Add Cloudflare Logpush related resources

bhushanthakur93:
Dependent on https://github.com/app-sre/qontract-schemas/pull/412

Example usage

```
    - provider: logpush_ownership_challenge
      identifier: test-logpush-challenge-zone
      zone: test-zone
      destination_conf: s3://bhthakur/logs?region=us-east-1&sse=AES256
    - provider: logpush_ownership_challenge
      identifier: test-logpush-challenge-account
      destination_conf: s3://bhthakur/logs?region=us-east-1&sse=AES256
    - provider: logpush_job
      identifier: test-logpush-job-zone
      enabled: true
      zone: test-zone
      logpull_options: fields=RayID,ClientIP,EdgeStartTimestamp&timestamps=rfc3339
      destination_conf: s3://bhthakur/logs?region=us-east-1&sse=AES256
      ownership_challenge: some-challenge
      dataset: http_requests
      frequency: high
    - provider: logpush_job
      identifier: test-logpush-job-account
      enabled: true
      logpull_options: fields=RayID,ClientIP,EdgeStartTimestamp&timestamps=rfc3339
      destination_conf: s3://bhthakur/logs?region=us-east-1&sse=AES256
      ownership_challenge: some-challenge
      dataset: audit_logs
      frequency: high
    - provider: logpull_retention
      identifier: test-logpull-retention
      zone: test-zone
      enabled: true
```


steveteahan:
It's worth noting the hyphens vs. spaces for provider names might be something we want to consider now. See more here: https://github.com/app-sre/qontract-reconcile/pull/3335/files#r1140471024

With that feedback we might decide it's worth changing now, but let's hold off on it for this second. Either way, it's probably worth actually enforcing this in code if we decide underscores in provider names are more trouble than they're worth.

steveteahan (commented):
Looks good overall! Let's just figure out the provider naming.

bhushanthakur93:
> It's worth noting the hyphens vs. spaces for provider names might be something we want to consider now. See more here: https://github.com/app-sre/qontract-reconcile/pull/3335/files#r1140471024
> 
> With that feedback we might decide it's worth changing now, but let's hold off on it for this second. Either way, it's probably worth actually enforcing this in code if we decide underscores in provider names are more trouble than they're worth.

Keeping the provider names as is based on conversation in other MR.

reconcile/test/test_utils_terrascript_cloudflare_client.py
```diff
@@ -232,6 +303,42 @@ def test_create_cloudflare_resources_terraform_json(account_config, backend_conf
                     "depends_on": ["cloudflare_zone.domain-com"],
                 }
             },
+            "cloudflare_logpush_ownership_challenge": {
+                "logpush_ownership_challenge_zone": {
+                    "destination_conf": "s3://bucket/logs?region=us-east-1&sse=AES256",
```
OliviaHY, line 308: Nit: Would it be good to extract specific value like this as constant and referenced it in the mock data and expected result? Also to `logpull_options`?

bhushanthakur93: In this case both `destination_conf` and `logpull_options` are not exactly constant, they can have different values depending on Logpush destination and dataset fields. It looks like they are constant because I copy pasted them in other places :sweat_smile: 

Using constant would have made more sense if this field was actually a constant and used in business logic when creating logpush terraform resource through `CloudflareLogpushOwnershipChallengeResource`.  

If we look at other fields within `test_create_cloudflare_resources_terraform_json` , we will see lot of string values present in the spec being copy-pasted in the expected results. Is this necessarily a bad thing? IMO no. There's always a trade-off, we could very well end up with whole bunch of constants which will make test readability a bit poor IMO.

steveteahan (approved):

PR #3317, opened 2023-03-14
add use-jump-host to get selectorsyncset

BumbleFeng:


tools/qontract_cli.py
```diff
@@ -114,6 +114,16 @@ def to_string(function):
     return function
 
 
+def use_jump_host(function):
```
maorfr, line 117: can this be imported from `reconcile.cli`?

maorfr (approved):

PR #3325, opened 2023-03-15
[saasherder] GQL classes and type hints

chassing:
Refactor `saasherder` to support generated GQL classes and add type hints

Ticket: [APPSRE-7152](https://issues.redhat.com/browse/APPSRE-7152)

Another try after reverting this PR twice.


fishi0x01 (approved):
3rd times a charm :)

PR #3328, opened 2023-03-16
fail on bad qenerate version

fishi0x01:
qenerate version mismatch potentially fails CI. However, the error is not very obvious and leads to confusion.

This introduces an early exit with a clear error message if there is a version mismatch.


geoberle (approved):
Ah nice. This is valuable. LGTM

PR #3330, opened 2023-03-17
Refactor defer with ExitStack

hemslo:
Refactor defer with [contextlib.ExitStack.callback](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack.callback), use build in functions to achieve the same effect as before.


fishi0x01 (approved):
Nice! TIL :+1:

PR #3331, opened 2023-03-17
print CI tooling versions

fishi0x01:
We do not use lockfiles (yet). So for debugging purposes it would be nice to see which tool versions where run during a jenkins check.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,18 +9,23 @@ envlist =
 
 [testenv]
 commands =
+    pytest --version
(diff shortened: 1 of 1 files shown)


chassing:
What about a general `pip freeze` somewhere? This would help to debug dependency issues as well.

fishi0x01:
My understanding is tox creates a separate venv for each test block. Further some test blocks have more deps declared than others. I think for now this is the easiest way - but Im also happy if there is a simple way to achieve a general `pip freeze`.

chassing (approved):

PR #3333, opened 2023-03-17
Ensure Cloudflare DNS record identifier unique within zone.

OliviaHY:
**What:**
Ensure Cloudflare DNS record identifier unique within zone.
**Why:**
We introduced stable ID for reliability when record with same name get's removed
**Validation:**
```
qontract-reconcile --dry-run --config config.dev.toml --log-level INFO terraform-cloudflare-dns 
[2023-03-17 10:24:45] [WARNING] [terraform_cloudflare_dns.py:are_record_identifiers_unique_within_zone:196] - test1 already exists in zone olivia-large-zone-devshift-net
[2023-03-17 10:24:45] [ERROR] [terraform_cloudflare_dns.py:run_with_defer:96] - Duplicated identifier detected.
```


reconcile/terraform_cloudflare_dns.py
```diff
@@ -92,6 +92,10 @@ def run_with_defer(self, dry_run: bool, defer: Callable) -> None:
 
         query_zones = self._get_cloudflare_desired_state()
 
+        if not are_record_identifiers_unique_within_zone(query_zones):
+            logging.error("Duplicated identifier detected.")
```
steveteahan, line 96:
```suggestion
            logging.error("Duplicate DNS record identifier(s) detected.")
```

reconcile/terraform_cloudflare_dns.py
```diff
@@ -179,6 +183,21 @@ def get_desired_state_shard_config(self) -> DesiredStateShardConfig:
...
+) -> bool:
+    for zone in zone_query_data.zones or []:
+        existing_records = set()
+        for record in zone.records or []:
+            record_id = record.identifier
+            if record_id not in existing_records:
+                existing_records.add(record_id)
+            else:
+                logging.warning(f"{record_id} already exists in zone {zone.identifier}")
+                return False
```
steveteahan, line 197: Should we continue processing all records in case there is >1 duplicate record? If there are say 10 duplicates, it'd take several iterations through MRs to fix.

steveteahan (commented):
A couple minor changes, but otherwise LGTM.

geoberle (commented):
depending on what the goal is you can achieve the same during bundling by adding a hint to the schema

```yaml
 - name: CloudflareDnsRecord_v1
   fields:
-  - { name: identifier, type: string, isRequired: true}
+  - { name: identifier, type: string, isRequired: true, isContextUnique: true }
   - { name: name, type: string, isRequired: true }
   - { name: type, type: string, isRequired: true}
   - { name: ttl, type: int, isRequired: true}
```

if the identifier is not unique within a zone you get an error like this during bundling (i manually changed the identifier of a record to be non-unique):

```
ERROR: /cloudflare/app-sre/dns/olivia-large-zone-1000.devshift.net.yml - records.[*] - the context identifier ['identifier'] = test1 is not unique within this context
make: *** [bundle] Error 1
```

this prevents non unique identifiers within a zone making it to an a-i repo

more about the `isContextUnique` property can be found here: https://github.com/app-sre/qontract-server/blob/master/SCHEMA.md

reconcile/terraform_cloudflare_dns.py
OliviaHY: Do you mean find all the dups in all the zones or just within that same zone? 
It's a tradeoff right, performance vs user friendliness. I don't expect MR checks to catch all the syntax mistake or other inconsistency I made at once, why is this different? Also once we make this self serviceable, it's up to the tenant to get a hint from this first error and find all the duplications.

OliviaHY: FYI, Gerd's suggestion about schema also don't give all the duplications, which I was hoping it did. But I still think that schema change is better.

geoberle:
```
ERROR: context uniqueness error - file: /cloudflare/app-sre/dns/olivia-large-zone-1000.devshift.net.yml, path: records.[1], identifier = ['identifier'], value = test1
ERROR: context uniqueness error - file: /cloudflare/app-sre/dns/olivia-large-zone-1000.devshift.net.yml, path: records.[3], identifier = ['identifier'], value = test3
make: *** [bundle] Error 1
```

steveteahan (approved):
LGTM -- I'll assume that you're moving forward with this for now and we can follow-up with a ticket for the schema change that could replace this in the future?

OliviaHY:
> LGTM -- I'll assume that you're moving forward with this for now and we can follow-up with a ticket for the schema change that could replace this in the future?

Thanks for the approval! Yeah that's plan

PR #3335, opened 2023-03-17
[terraform-cloudflare-resources] Add support for writing outputs to Vault

steveteahan:
`terraform-cloudflare-resources` didn't have a use case for creating Kubernetes `Secrets` or publishing Vault secrets before. With the working on automating Cloudflare ACM certificate renewal, we'll need to periodically update the ACME DCV records for Let's Encrypt certificates. There is more on this in the ticket [here](https://issues.redhat.com/browse/APPSRE-6948).

## Testing

1. Tested that the `validation_records` output from a zone using `certificates` to provision an ACM certificate writes a `Secret` the configured namespace, and writes the output to Vault
2. Tested that a `worker_script`, which doesn't have any outputs, doesn't create a `Secret` or write an entry to Vault
3. Run `terraform-resources` in dry-run mode to ensure the `ExternalResourceSpec` changes didn't have any effect on output names


reconcile/utils/external_resource_spec.py
```diff
@@ -113,7 +113,15 @@ def cluster_name(self) -> str:
...
-        return f"{self.identifier}-{self.provider}"
+        # Adhere to DNS-1123 subdomain names spec. It's reasonable to have provider
+        # names that have underscores, but without replacing them with hyphens we run
+        # into issues. Alternatively, we could change Cloudflare worker_script to
+        # worker-script and prevent the use of underscores going forward.
+        #
+        # More info can be found at:
+        # https://kubernetes.io/docs/concepts/overview/working-with-objects/names/.
+        provider = self.provider.replace("_", "-")
+        return f"{self.identifier}-{provider}"
```
steveteahan, line 124: @geoberle wouldn't mind a second set of eyes here to make sure this isn't going to cause problems. We've never enforced that providers needed to use hyphens, and as a result `worker_script` in Cloudflare ended up with an underscore. I'd suspect if there were going to be an issues it'd be that output names changed and we'd see a diff when running terraform-resources, which I don't.

The alternative would be to change `worker_script` to `worker-script`. In theory, not a huge deal, but it is a breaking change and something we'd need to update in a few places. If we were going to do that I think it'd also make sense to add some enforcement related to provider naming.

reconcile/gql_definitions/terraform_cloudflare_resources/terraform_cloudflare_resources.gql
```diff
@@ -2,6 +2,43 @@
...
+          field
+          version
+          format
+        }
+      }
+      automationToken {
+        path
+        field
+        version
+        format
```
fishi0x01, line 26:
```gql
... VaultSecret
```

reconcile/gql_definitions/terraform_cloudflare_resources/terraform_cloudflare_resources.gql
```diff
@@ -2,6 +2,43 @@
...
+      }
+      clusterAdminAutomationToken {
+        path
+        field
+        version
+        format
```
fishi0x01, line 32:
```gql
... VaultSecret
```

reconcile/gql_definitions/terraform_cloudflare_resources/terraform_cloudflare_resources.gql
```diff
@@ -2,6 +2,43 @@
...
+        hostname
+        knownHosts
+        user
+        port
+        identity {
+          path
+          field
+          version
+          format
+        }
```
fishi0x01, line 20: We could use a fragment here

```gql
... CommonJumphostFields
```

reconcile/gql_definitions/terraform_cloudflare_resources/terraform_cloudflare_resources.gql
```diff
@@ -2,6 +2,43 @@
...
+        field
+        version
+        format
+      }
+      spec {
+        region
+      }
+      internal
+      disable {
+        integrations
```
fishi0x01, line 39: add `e2eTests` -> This will make the protocol happy.

reconcile/terraform_cloudflare_resources.py
```diff
@@ -127,20 +152,201 @@ def build_clients(
...
+    use_jump_host: bool,
+    thread_pool_size: int,
+    internal: Optional[bool] = None,
+    account_names: Optional[Iterable[str]] = None,
+) -> tuple[ResourceInventory, OCMap]:
+    ri = ResourceInventory()
+
+    oc_map = init_oc_map_from_namespaces(
+        # The Namespace Protocol isn't happy with NamespaceV1, but since it's a
+        # generated class there isn't a lot that can be done here.
```
fishi0x01, line 167: I think there is `e2eTests` missing in disable to make protocol happy. See above

fishi0x01 (commented):
Just some smaller comments.

reconcile/terraform_cloudflare_resources.py
steveteahan: Ah, different `Cluster`... completely missed that.

reconcile/gql_definitions/terraform_cloudflare_resources/terraform_cloudflare_resources.gql
steveteahan: Fixed and as you mentioned the type error resolved. Thanks!

reconcile/gql_definitions/terraform_cloudflare_resources/terraform_cloudflare_resources.gql
steveteahan: Fixed

reconcile/gql_definitions/terraform_cloudflare_resources/terraform_cloudflare_resources.gql
steveteahan: Fixed

reconcile/gql_definitions/terraform_cloudflare_resources/terraform_cloudflare_resources.gql
steveteahan: Fixed

reconcile/terraform_cloudflare_resources.py
steveteahan: Fixed

steveteahan:
> Just some smaller comments.

Thanks for the feedback and reminders!

reconcile/test/test_terraform_cloudflare_resources.py
```diff
@@ -20,6 +21,18 @@ def query_data(external_resources):
...
+                    name="test-cluster",
+                    serverUrl="http://localhost",
+                    insecureSkipTLSVerify=None,
+                    jumpHost=None,
+                    automationToken=None,
+                    clusterAdminAutomationToken=None,
+                    spec=None,
+                    internal=None,
+                    disable=None,
+                ),
```
fishi0x01, line 35: More of an fyi: Chris added a nice way to create generated classes from sparse dicts https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/conftest.py#L102

Specifically designed to keep boilerplate in tests low. In this case it is not really important, but I thought it might be worth advertising Chris' awesome new feature :)

bhushanthakur93:
Looks good overall!

reconcile/utils/external_resource_spec.py
steveteahan: Thanks for the sanity check. As Gerd also pointed out, if anything was using an underscore it would already be failing `oc apply`. I also compared the tf.json outputs of `terraform-resources` and `terraform-cloudflare-resources` and there aren't any changes.

bhushanthakur93 (approved):
LGTM

PR #3336, opened 2023-03-17
fix jjb filter configs output

BumbleFeng:


jonmosco (approved):
LGTM.  Thanks for the fast fix @BumbleFeng

PR #3338, opened 2023-03-20
Check the server url before returning the client

janboll:
Makes it harder to extract tokens from app-interface

https://issues.redhat.com/browse/APPSRE-7312


reconcile/utils/oc.py
```diff
@@ -1655,7 +1655,7 @@ def init_oc_client(self, cluster_info, privileged: bool):
             secret_reader = SecretReader(settings=self.settings)
 
             try:
-                token = secret_reader.read(automation_token)
+                token_secret = secret_reader.read_all(automation_token)
```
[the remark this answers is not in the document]
janboll: Thanks, will update that one as well.

reconcile/test/oc/test_oc_connection_parameters.py
```diff
@@ -121,17 +122,39 @@ def test_from_cluster(
...
         secret_reader=secret_reader,
         cluster=test_cluster,
         cluster_admin=False,
         use_jump_host=use_jump_host,
     )
-
     assert parameters == expected_parameters
 
 
+def test_wrong_server_url():
```
fishi0x01, line 139: Nice!!! :heart:

reconcile/utils/oc_connection_parameters.py
```diff
@@ -105,13 +125,19 @@ def from_cluster(
...
                 except SecretNotFound:
                     logging.error(
                         f"[{cluster.name}] admin token {cluster.cluster_admin_automation_token} not found"
                     )
+                else:
+                    cluster_admin_automation_token = (
+                        OCConnectionParameters._get_token_verify_server_url(
+                            ClusterSecret(**secret_raw), cluster
+                        )
+                    )
```
fishi0x01, line 140: nit: wondering if we require a separate else block here. My understanding is it cannot raise a `SecretNotFound` error, so it cannot accidentally be caught. Not 100% sure, but I think in this case it is safe to put this into the try block too?

reconcile/utils/oc_connection_parameters.py
```diff
@@ -121,13 +147,17 @@ def from_cluster(
...
                 except SecretNotFound:
                     logging.error(
                         f"[{cluster.name}] automation token {cluster.automation_token} not found"
                     )
+                else:
+                    automation_token = (
+                        OCConnectionParameters._get_token_verify_server_url(
+                            ClusterSecret(**secret_raw), cluster
+                        )
+                    )
```
fishi0x01, line 160: nit: same as above

reconcile/utils/oc.py
```diff
@@ -1666,6 +1666,20 @@ def init_oc_client(self, cluster_info, privileged: bool):
...
+            if server_url != known_server_url:
+                self.set_oc(
+                    cluster,
+                    OCLogMsg(
+                        log_level=logging.ERROR,
+                        message=f"[{cluster}] server URL mismatch",
+                    ),
+                    privileged,
+                )
+                return
```
fishi0x01, line 1681: This behaves now different from new `OCMap`. Should we also raise an exception here to stay consistent, i.e., fail hard? Alternatively, I will open a follow-up PR to adjust `OCConnectionParameters` to also transport an `OCLogMsg` - might be good either way to have that option. Wdyt?

(Very sorry for all the hustle :bow: I will remove `OC_Map` latest by 20.04 and get rid of this duality.)

fishi0x01 (commented):
Looks great! My only concern is the exception raising.

reconcile/utils/oc.py
janboll: I think it works slightly similar in `OC_Map`, if you look at `get_cluster` it will raise if the object is of type `OCLogMsg`. So it would fail in this instance

janboll: Just tried locally: 
```
[2023-03-21 10:01:49] [ERROR] [openshift_base.py:init_specs_to_fetch:153] - [app-sre-stage-01] server URL mismatch
[2023-03-21 10:01:49] [ERROR] [openshift_base.py:init_specs_to_fetch:153] - [app-sre-stage-01] server URL mismatch

Process finished with exit code 1
```

reconcile/utils/oc_connection_parameters.py
janboll: Just trucked along with the existing pattern. Found it confusing as well. Can do a refactoring here

reconcile/utils/oc.py
fishi0x01: Awesome! Thanks for checking! :bow:

fishi0x01 (approved):
LGTM!

reconcile/utils/oc_connection_parameters.py
```diff
@@ -105,8 +134,12 @@ def from_cluster(
...
             if cluster.cluster_admin_automation_token:
                 try:
-                    cluster_admin_automation_token = secret_reader.read_secret(
-                        cluster.cluster_admin_automation_token
+                    cluster_admin_automation_token = (
+                        OCConnectionParameters._get_automation_token(
+                            secret_reader,
+                            cluster.cluster_admin_automation_token,
+                            cluster,
+                        )
```
fishi0x01, line 142: I think the outer brackets are a typo?

janboll: they are added by make format

fishi0x01 (approved):
LGTM

PR #3340, opened 2023-03-20
APPSRE-6609 OCMap for tf-resources

fishi0x01:
`terraform-resources` integration using the new OCMap.

This also introduces a [new qenerate version](https://github.com/app-sre/qenerate/pull/76) which allows custom mapping of GQL scalars to types -> We do not want pydantic's `Json` type here. To stay backwards compatible, we must map `JSON` -> `str` for now.

**Test**

```
qd profile run prod terraform-resources
qd profile run prod ocm-aws-infrastructure-access
```

terraform-resources executed against `app-sre` account.


chassing (approved):
This is awesome ❤️ 

Do you mind and remove the `mypy-reconcile.terraform_resources` section in `setup.cfg`?

PS: I love that you are mentioning `qd` in the PR description 🙇

PR #3341, opened 2023-03-20
Support _records_from_vault in terraform-aws-route53 integration

steveteahan:
Related: https://issues.redhat.com/browse/APPSRE-6948

This feature will allow us to consume Vault records from other integrations and use them as a source for creating DNS records in the `terraform-aws-route53` integration. The primary use case is domain control validation (DCV) records for certificates created in Cloudflare through the ACM feature: https://blog.cloudflare.com/advanced-certificate-manager/

## Example

```yaml
---
$schema: /dependencies/dns-zone-1.yml

...

allowed_vault_secret_paths:
- some-path/to-a/secret

records:
- name: _acme-challenge.subdomain
  type: TXT
  _records_from_vault:
    - path: some-path/to-a/secret
      field: validation_records
      key: _acme-challenge.subdomain.domain.com
```

`allowed_vault_secret_paths` must contain the approved paths from which a secret can be pulled. This will not be self-serviceable. This prevents a user from picking any path from which they could expose secrets.

`validation_records` is an output of `terraform-cloudflare-resources` where it is a JSON mapping of Cloudflare ACM DNS records values to the DCV values needed to prove ownership of the domain.

## Testing

1. Happy path of creating a record with 1 or more secrets sourced from Vault
2. Create a record that has a secret that doesn't exist
3. Create a record that uses a `key` in the secret that doesn't exist
4. Create a record that has a secret that cannot be decoded as JSON

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -2686,6 +2686,12 @@ DNS_ZONES_QUERY = """
         name
         elbFQDN
       }
+      _records_from_vault {
+        path
+        field
+        key
+        version
+      }
       _target_namespace_zone {
         namespace {
           managedExternalResources
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -2671,6 +2671,7 @@ DNS_ZONES_QUERY = """
       vpc_id
       region
     }
+    allowed_vault_secret_paths
     records {
       %s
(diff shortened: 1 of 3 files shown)


jfchevrette (approved):
Looks good!

PR #3342, opened 2023-03-20
APPSRE-7115: remove dyn-traffic-director integration

jfchevrette:
Dyn is retiring their Managed DNS services, as such we are removing support for the Dyn Traffic Director integration

Related: removal of schemas: https://github.com/app-sre/qontract-schemas/pull/416

Ref.: https://issues.redhat.com/browse/APPSRE-7115


fishi0x01 (approved):
Best!!

PR #3345, opened 2023-03-21
Add trigger reason to pipeline notifications

andreugallofre:
Add trigger reason on the message sent to slack

Ticket: [APPSRE-6647](https://issues.redhat.com/browse/APPSRE-6647)


lahavyuv86:
@andreugallofre This is a great addition! thanks!

PR #3346, opened 2023-03-21
tf-cloudflare-resources: add initial handling of cache_reserve value from schema

jfchevrette:
This adds handling of the cache_reserve schema field added in https://github.com/app-sre/qontract-schemas/pull/417

Here we pop the `cache_reserve` value out of the spec as it is currently not support by the cloudflare terraform provider

Once upstream adds supports we can handle the value and pass it along to terrascript

Ref.: https://issues.redhat.com/browse/APPSRE-7319


fishi0x01 (approved):
LGTM

PR #3351, opened 2023-03-22
[OCNative] remove get_api_resources

maorfr:
related to https://issues.redhat.com/browse/APPSRE-6438

OCNative should only handle GET requests. This removed function is long running and because OCDeprecated (should be OCCli) is doing everything else, there is little to no sense to get api-resources this way (especially when it takes a very long time).


chassing (approved):
RIP `get_api_resources` 💀

BumbleFeng:
depends on https://github.com/app-sre/qontract-reconcile/pull/3367

reconcile/test/test_utils_oc_native.py
BumbleFeng, line 1: please also remove https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/fixtures/oc_native/api.yml

maorfr: https://github.com/app-sre/qontract-reconcile/pull/3351/commits/b898933f3d90b89c8294692e00ca5b7fe15c1c01

reconcile/utils/oc.py
BumbleFeng, line 1: need to remove unused import https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/oc.py#L33

maorfr: https://github.com/app-sre/qontract-reconcile/pull/3351/commits/dc1a6007a44d53cdb8228a2f002e2f04a649959d

hemslo:
This one failed because we still need to get `(kind, group_version)` through [_parse_kind](https://github.com/app-sre/qontract-reconcile/blob/a538da70ccd63ccdb98fcc075f3858d1c0b4c06b/reconcile/utils/oc.py#L1128), which is rely on `api_resources`, it's being used in multiple methods in `OCNative`, like [get_items](https://github.com/app-sre/qontract-reconcile/blob/a538da70ccd63ccdb98fcc075f3858d1c0b4c06b/reconcile/utils/oc.py#L1341).

`is_kind_supported` is a public method, if we remove it, it will fallback to the parent method [is_kind_supported](https://github.com/app-sre/qontract-reconcile/blob/a538da70ccd63ccdb98fcc075f3858d1c0b4c06b/reconcile/utils/oc.py#L1162), which is also relying on `api_resources`.

PR #3352, opened 2023-03-22
[saasherder] add information to recover from configuration mismatch

maorfr:
related to https://issues.redhat.com/browse/APPSRE-7269

```
[ERROR] [saasherder.py:validate_promotions:1837] - Parent saas target has run with a newer configuration and the same commit (ref). Check if other MR exists for this target
[ERROR] [openshift_saas_deploy.py:run:172] - invalid promotions 
```

the workaround to this error is to remove the `promotion_data` section in a separate MR.
while we are working on a better solution, we can add information to the error message that will allow a user to update the same MR with the updated config hash in order to get a passing pr-check.


kwilczynski (approved):
Thank you!

PR #3354, opened 2023-03-22
Add check for server url fixed

janboll:
Redoing https://github.com/app-sre/qontract-reconcile/pull/3338

No need to revert, issue was based on server URL mismatch.

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1022,7 +1022,7 @@ def openshift_resources(
 @environ(["APP_INTERFACE_STATE_BUCKET", "APP_INTERFACE_STATE_BUCKET_ACCOUNT"])
 @environ(["gitlab_pr_submitter_queue_url"])
 @gitlab_project_id
-@threaded(default=20)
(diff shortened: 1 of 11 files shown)


chassing (approved):

reconcile/utils/oc_connection_parameters.py
```diff
@@ -105,6 +105,9 @@ class OCConnectionParameters:
 
     @staticmethod
     def _get_token_verify_server_url(secret: ClusterSecret, cluster: Cluster) -> str:
+        # Edge case, new cluster
+        if cluster.server_url == "":
```
chassing, line 109: Please use `if not cluster.server_url:`

janboll: Can do, but why?

chassing: Because it's more [pythonic](https://letmegooglethat.com/?q=what+is+pythonic+code)

PR #3357, opened 2023-03-23
Update thread size for openshift_saas_deploy to default 10

hemslo:
`openshift_saas_deploy` can spawn 20 `oc` processes at peak time when thread pool size is 20, each `oc` process consumes 50+ MB, reduce thread pool size to limit memory usage.


kwilczynski (approved):
@hemslo, looks good!

PR #3361, opened 2023-03-23
[terraform-cloudflare-resources] write validation_records output as JSON

steveteahan:
The current output isn't in JSON, so when we later try to decode it, when consumed by other integrations, it doesn't work.

## Testing

This output is currently unused, so minimal testing is necessary here. I tested to ensure that `terraform plan` is working and that the integration runs in dry-run mode. I didn't test it end-to-end given this output isn't used anywhere and it requires Cloudflare ACM (Enterprise-only feature) for end-to-end testing.


bhushanthakur93 (approved):
LGTM

PR #3363, opened 2023-03-24
Add resources config to internal-certificates

hemslo:
Add missing resource config for `internal-certificates`


kwilczynski (approved):
@hemslo, looks good!

PR #3364, opened 2023-03-24
APPSRE-7367 get-commit-sha for clients

fishi0x01:
Unify approach to fetch commit-shas from refs.

Github and Gitlab client should implement `get_commit_sha(ref: str) -> str`. This is needed for SAPM (and potentially other integrations such as `moving-commits`). This introduces:

- `github_api` is now fully typed
- more tests for `github_api`
- gitlab and github clients now both support `get_commit_sha`

**Local Tests**

```
qd profile run prod slack-usergroups
qd profile run prod terraform-cloudflare-resources
```

### DIFF
--- a/reconcile/utils/gitlab_api.py
+++ b/reconcile/utils/gitlab_api.py
@@ -665,5 +665,5 @@ class GitLabApi:  # pylint: disable=too-many-public-methods
 
     def get_commit_sha(self, ref: str, repo_url: str) -> str:
         project = self.gl.get_project(repo_url)
-        commits = project.commits.list(ref_name=ref)
+        commits = project.commits.list(ref_name=ref, per_page=1)
         return commits[0].id


reconcile/utils/gitlab_api.py
```diff
@@ -662,3 +662,8 @@ def last_comment(
                 continue
             return comment
         return None
+
+    def get_commit_sha(self, ref: str, repo_url: str) -> str:
+        project = self.gl.get_project(repo_url)
+        commits = project.commits.list(ref_name=ref)
```
hemslo, line 668: since here we just care about the first one, can limit `list` to only return 1 commit using [per_page](https://python-gitlab.readthedocs.io/en/stable/api-usage.html#pagination).

```python
commits = project.commits.list(ref_name=ref, per_page=1)
```

fishi0x01: Good idea. Fixed

PR #3366, opened 2023-03-24
[terraform-cloudflare-dns] Correct where rps is set.

OliviaHY:
#### What & Why:
It need to be set for performance concern of creating large zones and was set at the wrong place.

#### Validation:
Screenshot at Mar 24 16-41-50


jfchevrette (approved):
LGTM

esron (approved):

kwilczynski (approved):
@OliviaHY, looks good!

Thank you for fixing it! Also, I am sorry! We should have caught this during reviews! :bow:

steveteahan (approved):

PR #3367, opened 2023-03-25
decoupling gabi-authorized-users tests from OCNative

BumbleFeng:
and migrate the tests to pytest

it also requires change test_is_kind_namespaced from OCNative to OCDeprecated since we would remove OCNative's get_api_resources in https://github.com/app-sre/qontract-reconcile/pull/3351 and init OCNative requires a mock request of get version https://github.com/kubernetes-client/python-base/blob/master/dynamic/discovery.py#L141 which mocked in old gabi tests. check demo in https://github.com/app-sre/qontract-reconcile/pull/3375


setup.cfg
```diff
@@ -488,9 +488,6 @@ check_untyped_defs = False
 [mypy-reconcile.test.test_checkpoint]
 check_untyped_defs = False
 
-[mypy-reconcile.test.test_gabi_authorized_users]
-check_untyped_defs = False
-
```
chassing, line 493: Please add 

```
[mypy-reconcile.test.test_gabi_authorized_users]
disallow_untyped_defs = True
```

to enforce mypy testing on your tests. There is a [default](https://github.com/app-sre/qontract-reconcile/blob/master/setup.cfg#L470) to skip it for tests.

chassing (commented):
Please add missing type hints to `test_gabi_authorized_users`

maorfr (approved):
LGTM

pending above changes of course (does that goes without saying at this point?)

chassing (approved):

PR #3370, opened 2023-03-27
Cleanup request session when client is not used

hemslo:
Session is not closed when some clients are not used anymore, this change adds context manager for `oc` and `gitlab` clients to do cleanup to release resources.

### DIFF
--- a/reconcile/utils/github_api.py
+++ b/reconcile/utils/github_api.py
@@ -1,6 +1,10 @@
 import os
 from pathlib import Path
-from typing import Optional
+from types import TracebackType
+from typing import (
+    Optional,
+    Type,
+)
 from urllib.parse import urlparse
 
 from github import (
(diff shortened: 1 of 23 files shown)


hemslo:
There are some tests mock setup need to be updated, I'll fix them later.

reconcile/utils/github_api.py
```diff
@@ -38,6 +42,22 @@ def __init__(
...
+        exc_value: Optional[BaseException],
+        traceback: Optional[TracebackType],
+    ) -> None:
+        self.cleanup()
+
+    def cleanup(self) -> None:
+        """
+        Align with GitLabApi
+        """
+
```
cubismod, line 60: So just to confirm, the GitHub API doesn't allow us to manage sessions like the GitLab api?

hemslo: No, it's no exposed to user, all managed by PyGitHub internally.

PR #3380, opened 2023-03-29
rename OCDeprecated to OCCli

maorfr:
related to https://issues.redhat.com/browse/APPSRE-6438

this PR is the answer to the question: "can we deprecate `OCDeprecated`"
the answer is: "no" :sweat_smile: 
why: because `OCNative` only implements GET related operations (#1400)

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -31,8 +31,8 @@ from reconcile.utils.oc import (
     MayNotChangeOnceSetError,
     MetaDataAnnotationsTooLongApplyError,
     OC_Map,
+    OCCli,
     OCClient,
-    OCDeprecated,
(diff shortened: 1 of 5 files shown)


kwilczynski (approved):
@maorfr, looks good!

The only answer here is _yes_, I suppose :smile:

rporres (commented):
Looks good and it clarifies things. Since we don't have a deprecation mechanism (and I don't think this PR should wait for it), I would just add a note in the docstring clarifying that read methods from `OCCli` should not be used.

Another approach would be to make read methods private since they're used internally in the `OCCli` class. Choose your own adventure... ;-)

rporres:
ok, so then the second approach is not possible or we would break things 😂

PR #3385, opened 2023-03-29
prevent openshift integrations from fighting over resources

maorfr:
related to https://issues.redhat.com/browse/APPSRE-7401

this PR adds logic to check if a current resource was applied by the same integration that is currently trying to manage it (desired).
in case of conflict - error.

this will prevent integrations from fighting over resources - taking it from one another again and again.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -590,11 +590,23 @@ def _realize_resource_data(
             else:
                 # If resource doesn't have annotations, annotate and apply
                 if not c_item.has_qontract_annotations():
-                    msg = (
-                        "[{}/{}] resource '{}/{}' present "
-                        "w/o annotations, annotating and applying"
-                    ).format(cluster, namespace, resource_type, name)
-                    logging.info(msg)
+                    # if resource has annotations from a different integration, error
+                    c_int = c_item.qontract_integration_annotation()
+                    if c_int and c_int != d_item.integration:
+                        ri.register_error()
+                        msg = (
+                            f"[{cluster}/{namespace}] resource '{resource_type}/{name}' "
+                            "present with qontract annotations, conflict with integration: "
+                            f"{c_item.qontract_integration_annotation()}"
+                        )
+                        logging.error(msg)
+                        continue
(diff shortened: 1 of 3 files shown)


reconcile/openshift_base.py
```diff
@@ -590,11 +590,23 @@ def _realize_resource_data(
             else:
                 # If resource doesn't have annotations, annotate and apply
                 if not c_item.has_qontract_annotations():
-                    msg = (
-                        "[{}/{}] resource '{}/{}' present "
-                        "w/o annotations, annotating and applying"
-                    ).format(cluster, namespace, resource_type, name)
-                    logging.info(msg)
+                    # if resource has annotations from a different integration, error
```
patjlm, line 593: how will we be able to move resource from openshiftResources to saas files after that ?

patjlm: or the other way around...

patjlm: can we use `takeover` there?

hemslo: I want to understand the `integration` check here. I'm following saas deploy logic.

* `d_itm.integration` is populated by [saasherder](https://github.com/app-sre/qontract-reconcile/blob/b1f7a0e2d43a30993474f8f7dddd49aed1f0fa36/reconcile/utils/saasherder/saasherder.py#L1251), so it's value will always be the current integration's `QONTRACT_INTEGRATION`.
* `c_item.integration` is populated by [populate_desired_state_saas_file](https://github.com/app-sre/qontract-reconcile/blob/b1f7a0e2d43a30993474f8f7dddd49aed1f0fa36/reconcile/utils/saasherder/saasherder.py#L1251), so it's value is also the current integration's `QONTRACT_INTEGRATION`
* `c_int` is the actual `qontract.integration` annotation in k8s.

Does this mean `c_int != d_item.integration` is equivalent to the check in [has_qontract_annotations](has_qontract_annotations)? If yes, then we just need to let `has_qontract_annotations` to return additional info so we can act on different cases rather than just `True` or `False`.

PR #3388, opened 2023-03-30
Fix memory leak caused by lru_cache on instance methods

hemslo:
Use `@lru_cache` on instance methods will cause memory leak because `self` is captured.

Same fix as https://github.com/app-sre/qontract-reconcile/pull/1947, but for `vault` and `sentry_client`.

### DIFF
--- a/reconcile/utils/sentry_client.py
+++ b/reconcile/utils/sentry_client.py
@@ -1,5 +1,5 @@
-import functools
 import json
+from functools import lru_cache
 
 import parse
 import requests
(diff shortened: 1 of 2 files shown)


reconcile/utils/sentry_client.py
```diff
@@ -12,6 +12,9 @@ class SentryClient:  # pylint: disable=too-many-public-methods
     def __init__(self, host, token):
         self.host = host
         self.auth_token = token
+        self.get_organizations = lru_cache()(self._get_organizations)
+        self.get_projects = lru_cache()(self._get_projects)
+        self.get_users = lru_cache()(self._get_users)
```
patjlm, line 17: don't you want to keep the `maxsize=128` ?

patjlm (approved):
LGTM, just wondering about the `maxsize` missing attributes in `reconcile/utils/sentry_client.py`. I am honnestly unsure how `128` had been chosen, nor if it is a good value for the instance level cache..

rporres (approved):

hemslo:
Since `128` was the default `maxsize` in class level before, use the same for instance level should be enough. If we want to find a good value, then we can use [cache_info](https://docs.python.org/3/library/functools.html#functools.lru_cache) to get some metrics.

PR #3390, opened 2023-03-30
Going back to original Insights Cognito Settings

xxlhacker:
After further review need to remove the changes and revert back to the original.


TGPSKI (approved):
LGTM

PR #3391, opened 2023-03-30
Adjust rps to reduce rate limiting.

OliviaHY:


reconcile/terraform_cloudflare_dns.py
```diff
@@ -213,10 +214,10 @@ def get_cloudflare_provider_rps(
         return DEFAULT_PROVIDER_RPS
     size = len(records)
     if size <= 50:
-        return 2
+        return SMALL_ZONE_RPS
```
steveteahan, line 217: What about something that would step from 1 -> 4 rps in increments of 50 records?

```python
min(-(-size // 50), 4)
```

reconcile/terraform_cloudflare_dns.py
```diff
@@ -213,10 +214,10 @@ def get_cloudflare_provider_rps(
         return DEFAULT_PROVIDER_RPS
     size = len(records)
     if size <= 50:
-        return 2
+        return min(-(-size // 50), DEFAULT_PROVIDER_RPS)
```
steveteahan, line 217: This takes care of everything? if it's >= 200 it sets `DEFAULT_PROVIDER_RPS`, and `return LARGE_ZONE_RPS` is also then redundant? It replaces the need for if statements.

hemslo: `-(-size // 50)` is always `1` when `0 < size <= 50`, do you mean to use this formula under other conditions?

.coverage.Olivias-MacBook-Pro.12824.895100
hemslo, line 1: This file shouldn't be committed

steveteahan:
I was going to suggest this in a separate fix, but to save yourself time in promoting changes, what are your thoughts about also changing the number of zones that run in parallel?

reconcile/terraform_cloudflare_dns.py
OliviaHY: This is really cool. I didn't realize the minus is executed before the floor division is.

steveteahan (approved):

PR #3393, opened 2023-03-31
[ocm-utils] testability and test refactor

geoberle:
this PR is a first step to harmonize the httpretty mocks and fixtures for OCM testability.
`reconcile.test.ocm.conftest` defines pluggable fixtures, e.g. for `OCMBaseClient` and `OCM` that enables concise test cases and fixture setup.

not all OCM related tests have been migrated to linted/mypy-verified/lean mocked tests so far, so this is only to be considered a first step.

### DIFF
--- a/reconcile/test/ocm/test_utils_ocm_oidc_idp.py
+++ b/reconcile/test/ocm/test_utils_ocm_oidc_idp.py
@@ -173,6 +173,7 @@ def test_ocm_delete_idp(
     cluster_name: str,
     cluster_id: str,
     register_ocm_url_callback: Callable[[str, str, Callable], None],
+    find_ocm_http_request: Callable[[str, str], Optional[HTTPrettyRequest]],
 ) -> None:
     url = f"/api/clusters_mgmt/v1/clusters/{cluster_id}/identity_providers/idp-id-1"
 
@@ -183,3 +184,4 @@ def test_ocm_delete_idp(
 
     register_ocm_url_callback("DELETE", url, request_callback)
     ocm.delete_idp(cluster_name, "idp-id-1")
+    assert find_ocm_http_request("DELETE", url)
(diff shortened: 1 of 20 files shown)


reconcile/utils/ocm_base_client.py
```diff
@@ -161,8 +162,15 @@ def access_token_client_secret(self) -> HasSecret:
         ...
 
 
+class OCMAPIClientConfiguration(BaseModel, arbitrary_types_allowed=True):
```
fishi0x01, line 165: why do we allow arbitrary types here?

reconcile/test/ocm/test_utils_ocm_oidc_idp.py
```diff
@@ -0,0 +1,185 @@
...
+) -> None:
+    url = f"/api/clusters_mgmt/v1/clusters/{cluster_id}/identity_providers/idp-id-1"
+
+    def request_callback(
+        request: HTTPrettyRequest, uri: str, response_headers: dict[str, str]
+    ) -> tuple[int, dict, str]:
+        return (201, response_headers, json.dumps({}))
+
+    register_ocm_url_callback("DELETE", url, request_callback)
+    ocm.delete_idp(cluster_name, "idp-id-1")
```
fishi0x01, line 185: should there be some asserts for this test?

fishi0x01 (approved):
Nice testing structure! :heart:

reconcile/utils/ocm_base_client.py
geoberle: hmm. good point. i think at one point mypy was complaining and this fixed it but now it works without as well. going to remove it

geoberle: ah now i see. i get runtime errors without this when i provide a `VaultSecret` for the `HasSecret`

PR #3394, opened 2023-04-02
qontract-cli get ocm-login environment url fix

maorfr:
minor fix


geoberle (approved):
Thanks for cleaning up after me ❤️

PR #3395, opened 2023-04-03
[ocm-upgrade-scheduler] fix ocm env sharding/filtering

geoberle:
the ocm env filter was not handled so all clusters/orgs/upgrade policies from all orgs were provided as state.


patjlm (approved):

patjlm:
Thanks!

PR #3384, opened 2023-03-29
[dev] Refactor dev/run.sh

chassing:
Refactor dev container start script to support running arbitrary
qontract reconcile commands (tools, qontract-cli, integrations)
and getting the command arguments from the environment.

Ticket: [APPSRE-7400](https://issues.redhat.com/browse/APPSRE-7400)

PR #3397, opened 2023-04-04
Improve doc and experience to run integration in container

hemslo:
* quote `run` parameters to avoid `INTEGRATION_EXTRA_ARGS` break command syntax
* use `exec` in `run.sh` to allow container process to handle signal
* add `qontract-development-cli` to README


chassing:
Do you mind and keep your PR in sync with https://github.com/app-sre/qontract-reconcile/pull/3384?

hemslo:
@chassing nice, you are already using `exec` at the end in #3384, so my changes regarding `run.sh` can be discarded once yours merged.

chassing (approved):

PR #3399, opened 2023-04-05
Fix Pylint consider-using-with

hemslo:
This change fixed lint `consider-using-with` to ensure external resources are properly handled and release when not used.

* Replace `_log_lock.acquire()` and `_log_lock. release()` with `with _log_lock`
* Fix typo `teardown_mehtod` to `teardown_method`
* Replace `tempfile.mkdtemp()` and `shutil.rmtree(wd)` with `with tempfile.TemporaryDirectory() as wd`
* Replace `Popen` with high level `subprocess.run`
* Ensure `open` called using context manager `with open(...) as f:`
* Specify `pylint` version to `>=2.16` so all options in `.pylintrc` are supported

### DIFF
--- a/reconcile/test/test_quay_mirror.py
+++ b/reconcile/test/test_quay_mirror.py
@@ -48,7 +48,7 @@ class TestIsCompareTags:
         with open(os.path.join(self.tmp_dir.name, CONTROL_FILE_NAME), "w") as fh:
             fh.write(str(NOW - 100.0))
 
-    def teardown_mehtod(self):
+    def teardown_method(self):
         self.tmp_dir.cleanup()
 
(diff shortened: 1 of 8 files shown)


reconcile/test/test_quay_mirror.py
```diff
@@ -48,7 +48,7 @@ def setup_method(self):
         with open(os.path.join(self.tmp_dir.name, CONTROL_FILE_NAME), "w") as fh:
             fh.write(str(NOW - 100.0))
 
-    def teardown_mehtod(self):
```
chassing, line 51: nice catch

chassing (approved):

PR #3408, opened 2023-04-06
Remove inline user policies

esron:
The last step on https://issues.redhat.com/browse/APPSRE-4543

Removing inline user policies will remove duplicated resources. This will also be a final test on shardSpecOverride for pr-checks.


fishi0x01:
Was this tested locally against prod accounts? I.e., are we sure we do not have any inline policies defined somewhere in app-interface?

esron:
> Was this tested locally against prod accounts? I.e., are we sure we do not have any inline policies defined somewhere in app-interface?

Well, I don't see the use of `aws_iam_user_policy` in other parts of `qontract-reconcile`. Either way this code is removing "duplicated" resources.

This was even merged once and reverted for other reasons. 
https://github.com/app-sre/qontract-reconcile/pull/2656

fishi0x01:
According to the linked ticket, this should happen on 18th correct? LGTM, but maybe @steveteahan has more insights here?

PR #3410, opened 2023-04-06
adding a fix for the bug present in APPSRE-7028

jonmosco:
The jjb_client.py does not work when a `/` is in branch names.   The error thrown:
```
too many values to unpack (expected 3)
INTEGRATION FAILED: jenkins-job-builder 
```

On line 188,  the split is performed on `/`, and is expecting 3 values to be unpacked, but when a branch has a `/` in its name, that produces an extra argument, resulting in the error shown above.

This fix will allow the branch name to contain a `/` and replace it with an underscore `_`.

Testing was performed locally with `qd` by creating Jenkins jobs with repos containing one or more `/`'s in their name. By using the `*` operator, the unpacking issue is solved by doing a catch-all on the `item` value, and then joining with an underscore if necessary.   

```
INFO: ['create', 'job', 'ci-int', 'testbuild-with-slash-FINDME_BRANCH']
```


reconcile/utils/jjb_client.py
```diff
@@ -185,7 +185,9 @@ def print_diff(self, files, replace_path, action):
                 if equal:
                     continue
 
-            instance, item, _ = f.replace(replace_path + "/", "").split("/")
+            instance, *items, _ = f.replace(replace_path + "/", "").split("/")
+            item = "_".join(items)
```
fishi0x01, line 189: Is there a specific reason to replace the original app-interface defined value with `_`? My understanding is `item` is only used for logging right? 

```
instance, item, _ = f.replace(replace_path + "/", "").split("/", 2)
```

That way you could still log the original branch name defined in app-interface.

fishi0x01 (commented):
Thanks for looking into this bug! :bow: 
Quick question for my understanding.

reconcile/utils/jjb_client.py
jonmosco: It actually creates a directory structure names after the branch, so it will create a nested directory in the case above without replacing the value:
```
ls
testbuild-with-slash-FINDME/BRANCH/config.xml
```

jonmosco: jjb_client.py uses a wrapper class around the [jenkins-job-builder project](https://jenkins-job-builder.readthedocs.io/en/latest/index.html), so `item` is used in `builder.py` from this project, and inside the `JenkinsManager` class that is imported: `from jenkins_jobs.builder import JenkinsManager`.

Debug output from the same repo configuration containing a `/`:
```
_setup_output: testbuild-with-slash-FINDME/BRANCH
```

BumbleFeng:
The root cause of this issue is we use branch as part of the job name https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/resources/jenkins/global/templates.yaml#L20. I think Jenkins doesn't allow a job name contains '/'. If you try to create one, it says ‘/’ is an unsafe character.  In order to fix it all at once, I think we only have two options here:

1. replace '/' as '-' when formatting the job name. 
    e.g `name: '{gh_org}-{gh_repo}-gh-build-{branch|replace("/","-")}{job_suffix}'`
2. override the job name when branch needs to include '/'. 
```
- 'gh-build-master':
    branch: feature/test
    name: build-test
```

I am not sure if option 1 is possible for jjb, in the worst case we can just notify the tenant and update our docs for option 2.

jonmosco:
Option 1 still fails with the original error:
```
    instance, item, _ = f.replace(replace_path + "/", "").split("/")
    ^^^^^^^^^^^^^^^^^
ValueError: too many values to unpack (expected 3)
```

mshriver:
If I can get my dev env working properly I can help test this, its currently impacting me when using a `/` character for the gitlab project in a JJB template.

mshriver:
@BumbleFeng In my app-interface branch, I still hit the `ValueError` even when I override the template `name` field to not include any slashes.

BumbleFeng:
> @BumbleFeng In my app-interface branch, I still hit the `ValueError` even when I override the template `name` field to not include any slashes.

Do you have the MR link? I will take a look into it.

PR #3411, opened 2023-04-06
APPSRE-7367 gh cleanup

fishi0x01:
The gh client has a cleanup function. It doesnt do anything yet, but lets call it in case that changes in the future.


kwilczynski (approved):
@fishi0x01, looks good!

Nice to close a TODO item! :rocket:

reconcile/saas_auto_promotions_manager/utils/vcs.py
```diff
@@ -128,8 +128,7 @@ def open_app_interface_merge_request(self, mr: MergeRequestBase) -> None:
             mr.submit_to_gitlab(gitlab_cli=self._app_interface_api)
 
     def cleanup(self) -> None:
-        for _ in self._gh_per_repo_url.values():
-            # TODO: gh_client.cleanup()
-            pass
+        for gh_client in self._gh_per_repo_url.values():
+            gh_client.cleanup()
```
OliviaHY, line 132: Nit: I see that the cleanup() was recently added? Would it be nice to add a comment here saying it's doing anything? If the plan is to implement it very soon then don't worry about.

OliviaHY (commented):
LGTM. Just one non blocking nit.

reconcile/saas_auto_promotions_manager/utils/vcs.py
fishi0x01: Good point. However, if I add a comment here about it not doing anything right now, then this might be obsolete once someone adds this functionality in the gh client. Most likely, the person adding that functionality will overlook to change the comment in this consumer then. I'd say to keep the comment about this close to the cleanup function itself - and there seems to be a comment already.

PR #3412, opened 2023-04-06
APPSRE-7367 turning on SAPM w/o auto-merge

fishi0x01:
Next step of gradual rollout of SAPM. It will create and manage promotion MRs in Draft mode **without auto-merge**. This behavior can be turned on/off via unleash flags.

**Background**

SAPM has been running successfully  in dry-run mode for some time now. It is time now to allow SAPM to actually open and manage promotion MRs without any of them being merged, i.e., without doing any actual auto-promotion.

**Details**

In this PR the auto-merge tag is removed and draft mode is active, i.e., these MRs will not get merged automatically. Once they go stale (merge-conflict or newer promotion detected), SAPM will close them on its own, i.e., the MRs will clean-up on their own over time.

Doing this we can have SAPM run in parallel with the current auto-promotion mechanism and observe what it does w/o creating too much havoc.

All MRs are created with the `SAPM` label, i.e., worst-case it is fairly simple to find and close them all manually/scripted.

Since this opens draft MRs, none will trigger a MR check or ping any tenants to review.

Further, some minor changes to the housekeeping logic are introduced here.

### DIFF
--- a/reconcile/test/saas_auto_promotions_manager/merge_request_manager/merge_request_manager/test_housekeeping.py
+++ b/reconcile/test/saas_auto_promotions_manager/merge_request_manager/merge_request_manager/test_housekeeping.py
@@ -4,14 +4,14 @@ from collections.abc import (
 )
 
 from reconcile.saas_auto_promotions_manager.merge_request_manager.merge_request_manager import (
-    CONTENT_HASH,
-    NAMESPACE_REF,
-    SAPM_LABEL,
-    TARGET_FILE_PATH,
     MergeRequestManager,
 )
 from reconcile.saas_auto_promotions_manager.merge_request_manager.renderer import (
+    CONTENT_HASH,
+    FILE_PATH,
+    NAMESPACE_REF,
     PROMOTION_DATA_SEPARATOR,
+    SAPM_LABEL,
     Renderer,
 )
 from reconcile.saas_auto_promotions_manager.utils.vcs import VCS
@@ -57,7 +57,7 @@ def test_valid_description(vcs_builder: Callable[[Mapping], VCS], renderer: Rend
                     Blabla
                     {PROMOTION_DATA_SEPARATOR}
                     {NAMESPACE_REF}: some_ref
-                    {TARGET_FILE_PATH}: some_target
+                    {FILE_PATH}: some_target
                     {CONTENT_HASH}: some_hash
                 """,
                 }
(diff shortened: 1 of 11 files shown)


OliviaHY:
Just want to make sure I understand it correctly. You have already implemented the actual way of creating MRs with the auto-merge label and without Draft mode, and this is to revert that for the purpose of gradual roll out? I saw that you mentioned SAPM is running nicely in Production in dry-run mode 🎉 , which I'm assuming you observed the logs indicating the potential MRs are being "created" as expected?

fishi0x01:
@OliviaHY yes you are right. This is the next step in the gradual rollout. No dry-run. MRs are opened in Draft mode and not marked for auto-merge. That way, no CI job will be triggered for these MRs. SAPM will also manage these draft MRs and _should_ clean them up over time automatically and can run in parallel to the current auto-promoter.

I adjusted the PR description to highlight this more - it was a bit flaky and short :see_no_evil: 
Thanks a lot for pointing this out! :bow:

reconcile/test/saas_auto_promotions_manager/merge_request_manager/merge_request_manager/test_housekeeping.py
```diff
@@ -152,3 +152,40 @@ def test_bad_mrs(vcs_builder: Callable[[Mapping], VCS], renderer: Renderer):
...
+            ]
+        }
+    )
+    merge_request_manager = MergeRequestManager(
+        vcs=vcs,
+        renderer=renderer,
+    )
+    merge_request_manager.fetch_sapm_managed_open_merge_requests()
+    merge_request_manager.housekeeping()
+    vcs.close_app_interface_mr.assert_called_once()  # type: ignore[attr-defined]
```
OliviaHY, line 190: Curious why this type error is being ignored?

OliviaHY (approved):

reconcile/test/saas_auto_promotions_manager/merge_request_manager/merge_request_manager/test_housekeeping.py
fishi0x01: Good question. My understanding is that mypy thinks this is a `VCS` object. This object was built by a fixture and injected by the pytest framework. Basically mypy has no way of knowing, that this is an auto-speced object, which has `assert_called*` methods. You will see these ignores in a couple of tests that inject auto-speced objects via fixtures.

PR #3414, opened 2023-04-07
Add error handling for already decoded strings in get_api_resources

cubismod:
A recent change modified the behavior of `OCDeprecated._run()` to sometimes return undecoded strings and other times return utf-8 decoded strings. Thus in FedRamp where we only utilize `OCDeprecated`, we were seeing the following error in PR checks:

```py
File "/usr/local/lib/python3.9/site-packages/reconcile/utils/oc.py", line 371, in _init_old_without_types
  self.api_resources = self.get_api_resources()
File "/usr/local/lib/python3.9/site-packages/reconcile/utils/oc.py", line 718, in get_api_resources
  results = self._run(cmd).decode("utf-8").split("\n")
AttributeError: 'str' object has no attribute 'decode'
```

This adds error handling so that if a string is already utf-8 decoded, `decode()` isn't called again.

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -711,27 +711,33 @@ class OCDeprecated:  # pylint: disable=too-many-public-methods
         cmd = ["sa", "-n", namespace, "get-token", name]
         return self._run(cmd)
 
+    def _process_api_resources(self, results):
+        for line in results:
+            r = line.split()
+            kind = r[-1]
+            namespaced = r[-2].lower() == "true"
+            group_version = r[-3].split("/", 1)
+            # Core group (v1)
+            group = ""
+            api_version = group_version
+            if len(group_version) > 1:
+                # group/version
+                group = group_version[0]
+                api_version = group_version[1]
+            obj = OCDeprecatedApiResource(kind, group, api_version, namespaced)
+            d = self.api_resources.setdefault(kind, [])
+            d.append(obj)
+
     def get_api_resources(self):
         with self.api_resources_lock:
             if not self.api_resources:
                 cmd = ["api-resources", "--no-headers"]
-                results = self._run(cmd).decode("utf-8").split("\n")
-                for line in results:
-                    r = line.split()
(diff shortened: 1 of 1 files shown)


bkez322:
This makes sense and looks good to me.

bkez322 (approved):

reconcile/utils/oc.py
```diff
@@ -711,27 +711,33 @@ def sa_get_token(self, namespace, name):
...
+                api_version = group_version[1]
+            obj = OCDeprecatedApiResource(kind, group, api_version, namespaced)
+            d = self.api_resources.setdefault(kind, [])
+            d.append(obj)
+
     def get_api_resources(self):
         with self.api_resources_lock:
             if not self.api_resources:
                 cmd = ["api-resources", "--no-headers"]
-                results = self._run(cmd).decode("utf-8").split("\n")
```
kwilczynski, line 718: @cubismod, I wonder if the way of `self._run(cmd).encode().decode("utf-8").split("\n")` would work here?

Example:

```python
>>> s = "Hello"
>>> str(s, "utf-8")
TypeError: decoding str is not supported
>>> s.decode("utf-8")
AttributeError: 'str' object has no attribute 'decode'. Did you mean: 'encode'?
>>> s.encode().decode("utf-8")
'Hello'
```

cubismod: Huh yeah I think that would work too

hemslo:
This is related to this [change](https://github.com/app-sre/qontract-reconcile/pull/3399/files#diff-006493679cf46d27e100feeb8475e6e29987fe85cca4858a1f05608b2d536dc8R1116), I saw that we were using `strip` before `decode`, thought it maybe better to convert `bytes` to `str` first, but didn't update all usages. `get_api_resources` is the only method affected by `bytes` or `str`, so we can change return of `_run` back to `result.stdout.strip()`, then the current pr can be reverted to avoid confusing. Maybe specify the return of `_run` to be `bytes` to make it clear.

PR #3415, opened 2023-04-09
APPSRE-7414 promotion-state for utils

fishi0x01:
This is an attempt to make SAPM's DeploymentState usable by multiple integrations. In a next step, `openshift-saas-deploy` will use `DeploymentState` to store information about promotion. Afterwards we will make changes to this class to support multiple publishers per channel. 

In a gist what happens:

- Rename SAPM's `DeploymentState` to `PromotionState`
- Move `PromotionState` to general utlis module

### DIFF
--- a/reconcile/utils/promotion_state.py
+++ b/reconcile/utils/promotion_state.py
@@ -38,10 +38,11 @@ class PromotionState:
     def _fetch_promotions_list(self) -> None:
         all_keys = self._state.ls()
         for commit in all_keys:
+            # Format: /promotions/{channel}/{commit-sha}
             if not commit.startswith("/promotions/"):
                 continue
-            parts = commit.split("/")
-            self._commits_by_channel[parts[2]].add(parts[3])
+            _, _, channel_name, commit_sha = commit.split("/")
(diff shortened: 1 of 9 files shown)


reconcile/utils/promotion_state.py
```diff
@@ -23,29 +23,29 @@ class Config:
...
+        self._fetch_promotions_list()
 
-    def _fetch_deployment_list(self) -> None:
+    def _fetch_promotions_list(self) -> None:
         all_keys = self._state.ls()
         for commit in all_keys:
             if not commit.startswith("/promotions/"):
                 continue
             parts = commit.split("/")
             self._commits_by_channel[parts[2]].add(parts[3])
```
janboll, line 44: I know this is existing code, and not part of this PR. Nevertheless, could you add a comment, or make the pars named variables?

janboll (approved):

reconcile/utils/promotion_state.py
fishi0x01: Good point! Names the parts now and added a comment about the expected format

PR #3417, opened 2023-04-10
[APPSRE-7463] OCConnectionParameters should support flexible vault token field

cubismod:
Fix for #3354 so that `automationToken` no longer assumes that `token` is always the field specified for a cluster file. Also adds a test for it.

https://issues.redhat.com/browse/APPSRE-7463


fishi0x01 (approved):
LGTM! Thanks for the fix :bow:

PR #3420, opened 2023-04-11
APPSRE-7367 SAPM one mr per channel combo

fishi0x01:
We must reduce the number of auto-promotion MRs.

We create one MR per channel combination, i.e., we aggregate all commits for subscribers of the same channel(s) into one MR.

Further, this introduces a version flag on the MR description, which SAPM uses to close MRs from older versions (if there are still any open).


fishi0x01:
This PR only concerns SAPM. It has no further effects.

SAPM runs in prod and manages draft MRs to give us a better understanding of whats happening over time in-field - something that is hard to test locally. As a result, we have a couple of bugifx PRs coming up.

These PRs are quite big and very domain specific. 
I am going ahead and merging this w/o review, as it will only take away time from the reviewer and unnecessarily increases the debugging feedback loop on this one. The worst thing that can happen is SAPM failing - nobody relies on it working now, so this is not an issue.

PR #3421, opened 2023-04-11
[app-interface-state] init state bucket from various sources

geoberle:
implement state bucket initialization from `app-interface-settings` if the env var overrides are not present.

https://issues.redhat.com/browse/APPSRE-7226


reconcile/utils/state.py
```diff
@@ -53,16 +59,65 @@ def init_state(
...
+            secret_reader=secret_reader,
+            state_settings=state_settings,
+        )
+
+    raise StateInaccessibleException(
+        f"app-interface-settings-1.yml state provider {state_settings.provider} is not supported."
+    )
+
+
+def init_state_from_settings(
```
fishi0x01, line 98: nit: my understanding is this shouldn't be called from outside right? We should always use `init_state()`. In that case, would it make sense to prefix with `_` to highlight that we should use `init_state` instead?

fishi0x01 (approved):
Awesome!!! One step closer :D

reconcile/utils/state.py
geoberle: good point. done

PR #3423, opened 2023-04-11
APPSRE-7367 unleash flag for auto-promoter

fishi0x01:
This unleash flag allows us to turn on/off SQS events for auto-promotions. This flag will be used to smoothly shift towards SAPM and roll-back if necessary.


hemslo (approved):
LGTM!

PR #3433, opened 2023-04-12
[change-owners] resourcesfiles with schemas

geoberle:
resources files with schemas were not self-serviceable if they are used with generic change-types like `resource-owners` that does not enforce a schema.

with this change, resourcesfiles with schemas can now be made self-serviceable with change-types that don't enforce a schema.

### DIFF
--- a/reconcile/change_owners/change_types.py
+++ b/reconcile/change_owners/change_types.py
@@ -725,21 +725,16 @@ class ChangeTypeProcessor:
         ChangeTypeV1. the paths are represented as jsonpath expressions pinpointing
         the root element that can be changed
         """
-        paths = []
-        paths.extend(
-            self._allowed_changed_paths_for_file_type_and_schema(
-                file_ref.file_type, file_ref.schema, file_content, ctx
-            )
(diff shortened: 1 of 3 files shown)


reconcile/change_owners/change_types.py
```diff
@@ -725,13 +725,37 @@ def allowed_changed_paths(
         ChangeTypeV1. the paths are represented as jsonpath expressions pinpointing
         the root element that can be changed
         """
+        paths = []
+        paths.extend(
+            self._allowed_changed_paths_for_file_type_and_schema(
+                file_ref.file_type, file_ref.schema, file_content, ctx
+            )
+        )
+        if file_ref.schema is not None:
```
patjlm, line 734: don't we need to check the change type context schema there as well ?

geoberle: in this case not. if a file_ref with a schema comes in, we need to also check path expressions of the change-type that are not bound to any schema (which would be the change-schema anyways).

geoberle: i simplified the code a bit

patjlm (approved):

PR #3379, opened 2023-03-28
APPSRE-7258: Add soft limit for Cloudflare DNS zones.

OliviaHY:
#### Why & what:
Add logic to check if the total number of records exceed the default max items. If so, check the max_items in the zone file and ask for it to be set and approved.
Depend on https://github.com/app-sre/qontract-schemas/pull/424 schema change
Ticket Reference: https://issues.redhat.com/browse/APPSRE-7258
#### Tested cases:
* App Interface default max record set to 500 and zone max_records set to 1300 with a zone large than 1000,  integration ran successfully.
*  Both App Interface default max record and zone max_records unset, seeing the sane value 500 was set and zone has more than 500 record threw error
*  App Interface default max record set to 500 and zone max_records set to 1000 with a zone large than 1000,  integration threw "exceeds the configured max_items 1000" error.
*  App Interface default max record set to 500 and zone max_records set to 200 with a zone large than 1000,  integration threw "exceeds the configured max_items 200" error.

PR #3434, opened 2023-04-12
APPSRE-7258:  Add soft limit for Cloudflare DNS zones. 

OliviaHY:
#### Why & what:
Add logic to check if the total number of records exceed the default max items. If so, check the max_items in the zone file and ask for it to be set and approved.
Depend on https://github.com/app-sre/qontract-schemas/pull/424 schema change
Ticket Reference: https://issues.redhat.com/browse/APPSRE-7258
#### Note:
This was reverted by https://github.com/app-sre/qontract-reconcile/pull/3427 (Original MR: https://github.com/app-sre/qontract-reconcile/pull/3379) due to a bug where `max_records` were passed to terraform as part of zone spec. Added it to the excluded keys.
#### Tested cases:
* App Interface default max record set to 500 and zone max_records set to 1300 with a zone large than 1000,  integration ran successfully. (This was the test case that wasn't run thoroughly to catch previous bug). I most likely didn't wait for the integration to run all the way to the end. Double checked this time.
*  Both App Interface default max record and zone max_records unset, seeing the sane value 500 was set and zone has more than 500 record threw error
*  App Interface default max record set to 500 and zone max_records set to 1000 with a zone large than 1000,  integration threw "exceeds the configured max_items 1000" error.
*  App Interface default max record set to 500 and zone max_records set to 200 with a zone large than 1000,  integration threw "exceeds the configured max_items 200" error.


reconcile/terraform_cloudflare_dns.py
```diff
@@ -87,11 +92,34 @@ def run(self, dry_run: bool) -> None:
 
     @defer
     def run_with_defer(self, dry_run: bool, defer: Callable) -> None:
-        vault_settings = get_app_interface_vault_settings()
-        secret_reader = create_secret_reader(use_vault=vault_settings.vault)
+
+        settings = self._get_app_interface_settings()
+
+        if not settings.settings:
```
esron, line 98: Just a nit. I don't see the use `settings` alone, it's possible to instead of using `settings.settings[0]` we return `settings[0]` in `_get_app_interface_settings` and use here as `settings`

esron (approved):
LGTM, aside from some nits that I have commented.

PR #3435, opened 2023-04-13
Add tests for terraform tgw attachments

hemslo:
### <samp>🤖 Generated by Copilot at bb0b39f</samp>

Add pytest fixtures for testing terraform_tgw_attachments. The fixtures mock various AWS resources and configurations that are needed to verify the logic of the reconciliation. The file `reconcile/test/test_terraform_tgw_attachments.py` is updated with the new fixtures.
Add simple end to end cases of dry run, non dry run, with and without tgw connections.

### <samp>🤖 Generated by Copilot at bb0b39f</samp>

*  Add pytest fixtures for mock data ([link](https://github.com/app-sre/qontract-reconcile/pull/3435/files?diff=unified&w=0#diff-b5ab464582f4a599b9e05939387fe44570d983cb1a58d777a7af990bfb3124d4R1-R104))
* Add test cases ([link](https://github.com/app-sre/qontract-reconcile/pull/3435/files?diff=unified&w=0#diff-b5ab464582f4a599b9e05939387fe44570d983cb1a58d777a7af990bfb3124d4R107-R308))

### DIFF
--- a/reconcile/test/test_terraform_tgw_attachments.py
+++ b/reconcile/test/test_terraform_tgw_attachments.py
@@ -1,15 +1,132 @@
-def _setup_mocks(mocker, clusters=None, accounts=None):
+import pytest
+
+
+@pytest.fixture
+def account_tgw_connection():
+    return {
+        "name": "account_tgw_connection",
+        "provider": "account-tgw",
+        "manageRoutes": True,
+        "account": {
+            "name": "tgw_account",
+            "uid": "a-uid",
+            "terraformUsername": "tf-user",
+        },
+        "assumeRole": None,
+        "cidrBlock": "172.16.0.0/16",
+        "deleted": False,
+    }
+
+
+@pytest.fixture
+def account_vpc_connection():
+    return {
+        "name": "account_vpc_connection",
+        "provider": "account-vpc",
+    }
+
+
+@pytest.fixture
+def cluster_with_tgw_connection(account_tgw_connection):
+    return {
+        "name": "cluster_with_tgw_connection",
+        "ocm": {"name": "cluster_with_tgw_connection-ocm"},
+        "spec": {
+            "region": "us-east-1",
+        },
+        "network": {"vpc": "10.0.0.0/16"},
+        "peering": {
+            "connections": [
+                account_tgw_connection,
+            ]
+        },
+    }
+
+
+@pytest.fixture
+def cluster_with_vpc_connection(account_vpc_connection):
+    return {
+        "name": "cluster_with_vpc_connection",
+        "ocm": {"name": "cluster_with_vpc_connection-ocm"},
+        "spec": {
+            "region": "us-east-1",
+        },
+        "network": {"vpc": "10.0.0.1/16"},
+        "peering": {
+            "connections": [
+                account_vpc_connection,
+            ]
+        },
+    }
+
+
+@pytest.fixture
+def cluster_with_mixed_connections(account_tgw_connection, account_vpc_connection):
+    return {
+        "name": "cluster_with_mixed_connections",
+        "ocm": {"name": "cluster_with_mixed_connections-ocm"},
+        "spec": {
+            "region": "us-east-1",
+        },
+        "network": {"vpc": "10.0.0.2/16"},
+        "peering": {
+            "connections": [
+                account_tgw_connection,
+                account_vpc_connection,
+            ]
+        },
+    }
+
+
+@pytest.fixture
+def tgw():
+    return {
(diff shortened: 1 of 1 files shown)


reconcile/test/test_terraform_tgw_attachments.py
```diff
@@ -0,0 +1,308 @@
...
+        "terraformUsername": account_tgw_connection["account"]["terraformUsername"],
+        "assume_role": assume_role,
+        "assume_region": cluster_with_mixed_connections["spec"]["region"],
+        "assume_cidr": cluster_with_mixed_connections["network"]["vpc"],
+    }
+
+    mocks["ts"].populate_additional_providers.assert_called_once_with(
+        [expected_tgw_account, expected_tgw_account]
+    )
+    mocks["ts"].populate_tgw_attachments.assert_called_once_with(
```
patjlm, line 265: can we validate the the returns as well ?

hemslo: `populate_tgw_attachments` has no `return`

reconcile/test/test_terraform_tgw_attachments.py
```diff
@@ -0,0 +1,302 @@
...
+        "peering": {
+            "connections": [
+                account_tgw_connection,
+            ]
+        },
+    }
+
+
+@pytest.fixture
+def cluster_with_vpc_connection(account_vpc_connection):
```
fishi0x01, line 49: nit: would it make sense to extract these fixtures into a builder fixture instead, e.g., https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/conftest.py#L45-L75 or https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/saas_auto_promotions_manager/merge_request_manager/merge_request_manager/conftest.py#L69-L80?

This could then later be easily adjusted to use the [gql_class_factory](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/conftest.py#L172) once we use typed queries for this integration.

reconcile/test/test_terraform_tgw_attachments.py
```diff
@@ -0,0 +1,302 @@
...
+    integ.run(True, enable_deletion=False)
+
+    mocks["tf"].plan.assert_called_once_with(False)
+    mocks["tf"].apply.assert_not_called()
+
+
+def test_non_dry_run(mocker):
+    mocks = _setup_mocks(mocker)
+
+    integ.run(False, enable_deletion=False)
```
fishi0x01, line 158: Im wondering if instead of testing the whole `run()` method it might make sense to test single functions to reduce the whole `patch()` madness. `build_desired_state_tgw_attachments()` might be an ideal candidate for testing as it covers a lot of logic and we could pass the deps in there as clean mocks. In a next step for further tests, it might make sense to refactor the `run()` method and extract essential logic into functions with dep inversion in mind. (Personally, I really dislike python's patch approach and feel it is a sign of bad code structure in the integration itself. I know we do this in a lot of places, but if there is a way to avoid it then Im always voting for that :sweat_smile: ) Wdyt?

fishi0x01 (approved):
Thank you for adding tests here!! :bow: :heart: 
This is one of those untested integrations that really need it :see_no_evil: 

Only concern is around the whole patching approach, but I dont want to block you with my dislike for pytest patching.

reconcile/test/test_terraform_tgw_attachments.py
hemslo: cool, I'll try this today

reconcile/test/test_terraform_tgw_attachments.py
hemslo: I'm testing `run()` because that's the public method used eventually. I'm going to add sharding logic to it, so want to skip testing inner methods for now as they may change a lot during the process. Unless those are also public methods and we see them as interfaces. I'd like to treat them as internal sociable and external solitary. [Solitary or Sociable?](https://martinfowler.com/bliki/UnitTest.html)

I also don't like mocker patch this way, it's coupling with how I import a lib, if I change the import alias, it will break, or even work in a surprise way.

Change the existing structure to DI may not be easy, as some dependencies require results from previous step to build. If we want to have some clean tests, then we can extract a layer of pure functions, they can use clean mocks. Testing control logic still require current mocker patch as we are not using any DI framework.

reconcile/test/test_terraform_tgw_attachments.py
hemslo: I'll merge this pr then create a new one for refactor

PR #3442, opened 2023-04-17
APPSRE-7367 refactor diff in subscriber

fishi0x01:
- move diff logic inside the subscriber
- add tests for diffs
- unify all tests to use common `ConfigHash`


reconcile/saas_auto_promotions_manager/subscriber.py
```diff
@@ -45,6 +45,16 @@ def __init__(
         self.namespace_file_path = namespace_file_path
         self._content_hash = ""
 
+    def has_diff(self) -> bool:
+        current_hashes: list[ConfigHash] = []
```
hemslo, line 49: since `current_hashes` only used as `set`, can define it as `set` directly. Can use set comprehension or flatten by itertools

```python
current_hashes = {el for s in self.config_hashes_by_channel_name.values() for el in s}
```

or 

```python
current_hashes = set(itertools.chain.from_iterable(self.config_hashes_by_channel_name.values()))
```

fishi0x01: Done :+1:

hemslo (approved):
LGTM! Thanks!

PR #3447, opened 2023-04-18
ssl_verify to True everywhere

maorfr:
our certificate management is quite good at this point in time. all our CI instances have valid certificates, and so do prometheus instances of private clusters.

let's validate certificates.


fishi0x01 (approved):
LGTM :rocket:

PR #3449, opened 2023-04-18
APPSRE-7517 promote target_config_hash only on publishJobLogs flag

fishi0x01:
- subscriber only cares about target_config_hashes if the publishJobLogs flag is set
- diff logic only checks if desired hashes are a subset of current hashes -> we do not care about any additional hashes - this is caught in validation phase https://github.com/app-sre/qontract-reconcile/pull/3448
- adding more tests


reconcile/saas_auto_promotions_manager/subscriber.py
```diff
@@ -44,14 +45,17 @@ def __init__(
...
         self._content_hash = ""
+        self._use_target_config_hash = use_target_config_hash
 
     def has_diff(self) -> bool:
         current_hashes = {
             el for s in self.config_hashes_by_channel_name.values() for el in s
         }
+        desired_hashes_are_in_current_hashes = len(
+            current_hashes & set(self.desired_hashes)
+        ) == len(self.desired_hashes)
```
hemslo, line 56: We are checking if `set(self.desired_hashes)` is subset of `current_hashes`, right?

It can be simplified by [issubset](https://docs.python.org/3.9/library/stdtypes.html#frozenset.issubset):

```python
desired_hashes_are_in_current_hashes = set(self.desired_hashes).issubset(current_hashes)
```

or

```python
desired_hashes_are_in_current_hashes = set(self.desired_hashes) <= current_hashes
```

Btw, complete diff 2 sets is usually mean to find [symmetric difference](https://docs.python.org/3.9/library/stdtypes.html#frozenset.symmetric_difference) (`a ^ b`), please check if subset is good enough here.

fishi0x01: Adjusted :+1: 

> Btw, complete diff 2 sets is usually mean to find [symmetric difference](https://docs.python.org/3.9/library/stdtypes.html#frozenset.symmetric_difference) (a ^ b), please check if subset is good enough here.

I explicitly do not want complete diff -> please see test https://github.com/app-sre/qontract-reconcile/pull/3449/files#diff-0dfd4aae8e4dbed3b3f6a2523517a347efb3dadffcdc5cf67eb35a47432aa96fR96

Only if desired hash is subset.

PR #3450, opened 2023-04-18
allow recycled object to be not found

maorfr:
in this [MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/65169) there was a failure in the deployment where it tried to delete a STS and also recycle it.

this is likely due to an immutable change and also a change in a configmap.

if when trying to recycle, the recycled object is not there - there is nothing left to do?


maorfr:
more discussion in the rollback [MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/65183)

PR #3451, opened 2023-04-18
Refactor ALB rule conditions to support multiple condition types

jfchevrette:
Also adds support for additional condition types

See the https://github.com/app-sre/qontract-schemas/pull/431 for an example of the new schema/data format

Ref.: https://issues.redhat.com/browse/APPSRE-7483


reconcile/utils/terrascript_aws_client.py
```diff
@@ -4865,15 +4865,34 @@ def populate_tf_resource_alb(self, spec, ocm_map=None):
...
+                elif cond["type"] == "path-pattern":
+                    condition_values.append(
+                        {"path_pattern": {"values": cond["path_pattern"]}}
+                    )
+                elif cond["type"] == "source-ip":
+                    condition_values.append(
+                        {"source_ip": {"values": cond["source_ip"]}}
+                    )
+                else:
+                    raise KeyError(f"unknown alb rule condition type {cond['type']}")
```
hemslo, line 4891: Can extract supported types to a constant like
```python
SUPPORTED_LB_LISTENER_RULE_CONDITION_TYPES = {
    "host-header",
    "http-request-method",
    "path-pattern",
    "source-ip",
}
```

Then this can be simplified to:

```python
    @staticmethod
    def _get_condition_value(condition):
        condition_type = condition["type"]
        if condition_type not in SUPPORTED_LB_LISTENER_RULE_CONDITION_TYPES:
            raise KeyError(f"unknown alb rule condition type {condition_type}")
        return {condition_type: {"values": condition[condition_type]}}
```

```python
condition_values = [self._get_condition_value(c) for c in conditions]
```

Btw, we don't support `http_header` and `query_string`, right?

jfchevrette: I chose to do it like that because it is very clear what is going on. IMO a repeating `if/elif` block like that may not be pretty but it makes the code very clear and readable. Another  reason I did it like this is because the `type` value has dashes where the corresponding dict key names have underscores (`host-header`/`host_header`) so your proposal would need to be adjusted to replace `-` with `_` for the key name. This may seem silly but I chose to go with a direct mapping to the terraform implementation

Lastly,

> Btw, we don't support http_header and query_string, right?

not in this implementation but if we eventually do, your proposal would not work as those have different kinds of values

Would you be OK if we leave it as it?

steveteahan (approved):

reconcile/utils/terrascript_aws_client.py
hemslo: The reason I dislike repeating `if/elif` is not only code duplication, but also taking time to understand the intention of that code block, have to go through all branches to check if any branch has some special handling like side effect. Plus the new logic is already in an outer for loop. So I prefer declarative to imperative, in this case at least extract a function `_get_condition_value` with repeating `if/elif` will help understanding the logic.


```python
    @staticmethod
    def _get_condition_value(condition):
        if condition["type"] == "host-header":
            return {"host_header": {"values": condition["host_header"]}}
        elif condition["type"] == "http-request-method":
            return {"http_request_method": {"values": condition["http_request_method"]}}
        elif condition["type"] == "path-pattern":
            return {"path_pattern": {"values": condition["path_pattern"]}}
        elif condition["type"] == "source-ip":
            return {"source_ip": {"values": condition["source_ip"]}}
        else:
            raise KeyError(f"unknown alb rule condition type {condition['type']}")
```

If we upgrade to python 3.10, then we can use pattern match to make it cleaner.

Sorry I didn't notice that `-` `_` thing, this should work:

```python
SUPPORTED_LB_LISTENER_RULE_CONDITION_TYPE_MAPPING = {
    "host-header": "host_header",
    "http-request-method": "http_request_method",
    "path-pattern": "path_pattern",
    "source-ip": "source_ip",
}
```

```python
    @staticmethod
    def _get_condition_value(condition):
        condition_type = condition["type"]
        condition_type_key = SUPPORTED_LB_LISTENER_RULE_CONDITION_TYPE_MAPPING.get(
            condition_type, None
        )
        if condition_type_key is None:
            raise KeyError(f"unknown alb rule condition type {condition_type}")
        return {condition_type_key: {"values": condition[condition_type_key]}}
```

To support different mapping schemas, can change mapping to `str -> function`, so each schema can have it's own, but it maybe too much for a simple transformer like this.

I rarely write python functions longer than 10 statements, find it's very hard to see the flow in this file, it will be great if we can make it clearer when introducing new logic.

So I'm OK as long as we can keep the flow clear in the main function.

jfchevrette: I have implemented your suggestion. Thank you!

PR #3452, opened 2023-04-18
delete previous diff comment

BumbleFeng:
keep MR comments history clean


hemslo:
Since we are reusing the same logic, can we extract it to a method in `gitlab_api`? A simple extraction is like this:

```python
    def delete_merge_request_comments(self, mr_id, startswith):
        comments = self.get_merge_request_comments(mr_id)
        for c in sorted(comments, key=lambda k: k["created_at"]):
            body = c["body"] or ""
            if c["username"] == self.user.username and body.startswith(startswith):
                self.delete_gitlab_comment(mr_id, c["id"])
```

2 issues:

* Is `sorted` needed? It seems we are deleting all matching comments, order seems not related.
* `merge_request` and `note` fetched twice, one in `get_merge_request_comments`, the other in `delete_gitlab_comment`, to avoid that, we can have a new delete method accepting `note` object to do `note.delete()` directly, this can save `2 * N (N = comments to delete)` requests

reconcile/utils/gitlab_api.py
```diff
@@ -369,7 +370,7 @@ def get_merge_request_author_username(self, mr_id: int) -> str:
         return merge_request.author["username"]
 
     def get_merge_request_comments(
-        self, mr_id: int, include_description: bool = False
+        self, mr_id: int, include_description: bool = False, include_note: bool = False
```
hemslo, line 373: The result of `get_merge_request_comments` contains a list of non system notes with an optional description item, can we always have `note` key to simplify this call, so we don't need to have `include_note`?

The value can be `Optional[ProjectMergeRequestNote]`, `None` for description item, note object for real note.

Since it's optional we need to have a check before use it, can do a filter in https://github.com/app-sre/qontract-reconcile/pull/3452/files#diff-1d5ab1abb6ecb1d8c243a9e8c04d4cf6505e6be2c8932df654ea9a11c6c1b64bR409

hemslo (approved):
LGTM! Thank you for the refactor, I left a small optional improvement comment.

geoberle (approved):
thanks for the refactor @BumbleFeng !

reconcile/utils/gitlab_api.py
BumbleFeng: `include_description` is only used when need to `get_approver_decisions_from_mr_comments`, for `delete_merge_request_comments` include_description would keep as false as default, we don't need to add a filter here.

hemslo: Cool, it's OK for now, only needed when we apply typing to this file, like convert `dict` to `dataclass`, then `note` field become `Optional[ProjectMergeRequestNote]`, it requires a none check before use.

rporres (approved):
nice!

PR #3455, opened 2023-04-19
[OCNative] remove get_api_resources

hemslo:
Yet another attempt following

* https://github.com/app-sre/qontract-reconcile/pull/3351
* https://github.com/app-sre/qontract-reconcile/pull/3382

The root cause of error is a wrong value type in `api_resources` of `OCDeprecated`, expecte `api_version` to be `str`, but actually it's assigned a list value, fixed in 48d7b6c58796db756a0eddfe4fa2c8537b58d530

Note after this change, `OCNative` will use `get_api_resources` from `OCDeprecated`, it's faster this way because `OCDeprecated` is calling `oc` cli to get `api-resources`, it can send 100+ concurrent requests to get result under 2s, while `OCNative` one was doing that sequentially for about 1 minute.


maorfr (approved):
LGTM

PR #3462, opened 2023-04-20
Set Cloudflare user status to 'accepted' to skip accepting invites

steveteahan:
Cloudflare invites expire quickly, so there'd either be a need for additional automation to resend expired invites. The current assumption is that the user already has a Cloudflare user account, so we can skip this process entirely by setting the status to accepted.

## Testing

1. For an existing Cloudflare user in the "accepted" state, applied this change and verified that while Terraform detected a state change, nothing changed in the user account, and there weren't new notifications
2. If we revert this change, the user doesn't go back to a "pending" state even though that's the default in the provider


bhushanthakur93 (approved):
LGTM

PR #3464, opened 2023-04-20
Terraform Repo Qontract Reconcile Integration

cubismod:
https://issues.redhat.com/browse/APPSRE-7473

Terraform Repo is a new App Interface integration laid out in this [design doc](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/terraform-repo.md). The integration allows tenants to define unique and bespoke pieces of infrastructure that is not easily represented with our current [terraform-resources integration](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/terraform_resources.py) which uses TerraScript while Terraform Repo allows tenants to define their own HCL Terraform repositories.

Terraform Repo consists of the following components:

- [Schemas](https://github.com/app-sre/qontract-schemas/pull/425)
- QR integration (this part)
- [Executor](https://github.com/app-sre/terraform-repo-executor)

The QR integration is responsible for collecting each Terraform repo definition from App Interface and coordinating a YAML plan that is sent to the executor which performs `terraform plan` & `terraform apply`.

Each terraform repo is represented with a file in App Interface like so:

```yaml
---
$schema: /aws/terraform-repo-1.yml

account:
    $ref: /aws/app-int-example-01/account.yml

name: plan_a
repository: https://gitlab.example.com/terraform/tf-repo-example
ref: a390f5cb20322c90861d6d80e9b70c6a579be1d0
projectPath: plan_a
```

This integration takes that file and then produces an action plan which looks like this:

```yaml
---
delete: false
dry_run: true
name: my_cool_repo
project_path: plan_a
ref: a390f5cb20322c90861d6d80e9b70c6a579be1d0
repository: https://gitlab.example.com/terraform/tf-repo-example
secret:
  path: vault/creds/terraform/app-int-example-01/config
  version: null
```
which will be fed to the [executor](https://github.com/app-sre/terraform-repo-executor) as a mounted file during PR checks and build master jobs to app interface.

For multiple repositories, multiple individual YAML files will be created and fed to separate runs of the executor. The location of these outputs can be controlled with the `--output-dir` CLI option.

Please note that `account, name, repository, and projectPath` cannot be modified once a repo has been added to App Interface and the integration itself checks for these. Additionally, each repository target is validated to ensure that the project exists in the existing App Interface GitLab instance (configured separately).

One important aspect about this integration is that it will not run on OpenShift but rather as a PR check job and build master job running in Jenkins exclusively. [Reasoning is discussed a bit in the design doc](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/terraform-repo.md?ref_type=heads#timer-based-reconciliation-vs-triggers). Also this integration will only be deployed to FedRamp initially.

### <samp>🤖 Generated by Copilot at d2dc6ed</samp>

*  Add a new subcommand `terraform-repo` to the `reconcile` CLI ([link](https://github.com/app-sre/qontract-reconcile/pull/3464/files?diff=unified&w=0#diff-2ac404c3a4f4d289c1811816c94253a70cc6559b8de427388de06546141f1f43R1549-R1568))
* Add a new GraphQL query definition file `terraform_repo.gql` ([link](https://github.com/app-sre/qontract-reconcile/pull/3464/files?diff=unified&w=0#diff-07306e78e084b1597025733d9c9787506f617028ba6ebe1cb240937c39ed8f34R1-R18))
* Add a new Python module `terraform_repo.py` in the `reconcile/gql_definitions/terraform_repo` directory ([link](https://github.com/app-sre/qontract-reconcile/pull/3464/files?diff=unified&w=0#diff-3bcae5acc37bc36a49c3434c915a8d6ea887a80353e72105e7c199a581b51023R1-R91))
* Add a new Python module `terraform_repo.py` in the `reconcile` directory ([link](https://github.com/app-sre/qontract-reconcile/pull/3464/files?diff=unified&w=0#diff-05e8f500c44ae2dc8a3b8087020cc1d2946506df84ea9a118e110c8d8ad82ba2R1-R302))
* Modify the `s3_state_builder` fixture in the `reconcile/test/conftest.py` module ([link](https://github.com/app-sre/qontract-reconcile/pull/3464/files?diff=unified&w=0#diff-d95b64d0d71775e869a296652b83fb00da405610a4793eedc7cc22d0bdb3977fL67-R73))
* Add a new Python module `test_terraform_repo.py` in the `reconcile/test` directory ([link](https://github.com/app-sre/qontract-reconcile/pull/3464/files?diff=unified&w=0#diff-f578fef16df2ffb09f95711b988012130afa7eea66f67d80c3168c0effe4c178R1-R215))

Dependent on https://github.com/app-sre/qontract-schemas/pull/425


cubismod:
Screenshot_20230425_113035
my inspect element hack of what PR check output might look like

bkez322 (approved):
On the qontract-reconcile project, a pull request has been approved, enabling a valuable new feature called "the terraform-repo integration". This enhancement allows tenants (system users) to efficiently manage their own terraform configuration files for AWS accounts in accordance with Infrastructure-as-Code (IaC) and Continuous Delivery (CD) best practices. By leveraging this integration, users can streamline their cloud infrastructure management while ensuring consistency and compliance with established industry standards.

I have thoroughly examined the proposed changes, and after careful consideration, I am pleased to express my approval and satisfaction with the quality of the work presented. The modifications align with the project's objectives and adhere to best practices, thus earning my wholehearted endorsement.

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,304 @@
...
+                    a_repo.name
+                )
+            )
+        if (a_repo.ref != b_repo.ref) or (a_repo.delete != b_repo.delete):
+            changed[a_key] = b_repo
+
+    return changed
+
+
+def gql_to_tf_repo(repo: TerraformRepoV1) -> TFRepo:
```
hemslo, line 169: Is this just for serialization/deserialization? pydantic models can be exported via `model.dict(by_alias=True)` [doc](https://docs.pydantic.dev/usage/exporting_models/), so we can use it directly instead of creating similar classes manually.

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,304 @@
...
+    account: AWSAccount
+    delete: Optional[bool]
+
+
+class ActionPlan(BaseModel):
+    dry_run: bool
+    repos: list[TFRepo]
+
+
+def get_repos(query_func: Callable) -> list[TFRepo]:
```
hemslo, line 61: I'm not sure if we should move this to [typed_queries](https://github.com/app-sre/qontract-reconcile/tree/master/reconcile/typed_queries), then integration code only call typed query functions.

Since this is the only usage now, maybe OK to not extract, just not sure about the pattern we should follow.

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+def map_repos(repos: list[TFRepo]) -> MutableMapping[str, TFRepo]:
+    """Generate keys for each repo to more easily compare"""
+    repo_map: MutableMapping[str, TFRepo] = dict()
+    for repo in repos:
+        # repo names are unique per app interface instance as enforced in the schema
+        repo_map[repo.name] = repo
+
+    return repo_map
```
hemslo, line 90: Would it be clearer using dict comprehension like this?

```python
{repo.name: repo for repo in repos}
```

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,304 @@
...
+    b: Mapping[str, TFRepo],
+    requireDelete: bool = False,
+) -> MutableMapping[str, TFRepo]:
+    """
+    Returns a mapping of repos that are present in mapping a but missing in mapping b
+    When requireDelete = True, each repo from a is checked to ensure that the delete flag is set"""
+    missing: MutableMapping[str, TFRepo] = dict()
+    for a_key, a_repo in a.items():
+        b_repo = b.get(a_key)
+        if b_repo is None:
```
hemslo, line 128: Is there a case that `a_key` exist in `b` but the value is `None`? If not, then `b_repo` is only used in exist check, can use key check directly like

```python
if a_key not in b:
```

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,304 @@
...
+            if requireDelete is True and a_repo.delete is not True:
```
hemslo, line 129: `is True` and `is not True` are not necessary, can directly use

```python
if requireDelete and not a_reop.delete
```

[reference](https://peps.python.org/pep-0008/#programming-recommendations):

```
Don’t compare boolean values to True or False using ==:
# Correct:
if greeting:

# Wrong:
if greeting == True:

Worse:
# Wrong:
if greeting is True:
```

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,304 @@
...
+) -> MutableMapping[str, TFRepo]:
+    """
+    Returns a mapping of repos that have changed between a and b in the form of returning their b values
+    In order for Terraform to work as expected, the tenant can only update the ref between MRs
+    Updating account, repo, or project_path can lead to unexpected behavior so we error
+    on that
+    """
+    changed: MutableMapping[str, TFRepo] = dict()
+    for a_key, a_repo in a.items():
+        b_repo = b.get(a_key, a_repo)
```
hemslo, line 151: if `b` has no `a_key` then this will return `a_repo` as value, is this expected? It will pass diff implicitly, maybe worth do a simple get query and None check to make it explicit.

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,304 @@
...
+
+    to_be_created = diff_missing(desired_map, existing_map)
+    to_be_updated = diff_changed(existing_map, desired_map)
+
+    # indicates repos which have had their definitions deleted from App Interface
+    # a pre-requisite to this step is setting the delete flag in the repo definition
+    to_be_deleted = diff_missing(existing_map, desired_map, True)
+
+    if not dry_run and state:
+        update_state(to_be_created, to_be_deleted, to_be_updated, state)
```
chassing, line 255: Instead of having these dictionary comparisons, you may use `set` comparison. e.g https://docs.python.org/3/library/stdtypes.html#frozenset.difference

pseudo code:
```
def calculate ...
existing_state: list[TFRepo],
desired_state: list[TFRepo],

existing = set(existing_state)
desired = set(desired_state)

to_be_created = desired - existing
to_be_deleted = existing - desired

you need to implement a `__eq__` method, e.g. https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/skupper_network/models.py#L39
```

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,304 @@
...
+                path=repo.account.automation_token.path,
+                version=repo.account.automation_token.version,
+            ),
+        ),
+        delete=repo.delete,
+    )
+
+
+def merge_results(
+    created: Mapping[str, TFRepo], updated: Mapping[str, TFRepo]
```
chassing, line 192: I would use `created: Iterable[TFRepo], updated: Iterable[TFRepo]` instead, then the methods would be much easier to read.

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,304 @@
...
+    """
+    Merges results into a RepoOutput dict which will be transformed to outputted YAML.
+    This includes checking modified values for a delete flag
+    """
+    output: list[TFRepo] = []
+    for c_value in created.values():
+        logging.info(["create_repo", c_value.account.name, c_value.name])
+        output.append(c_value)
+    for u_value in updated.values():
+        if u_value.delete is True:
```
chassing, line 203: Please rewrite as `if u_value.delete:`

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,304 @@
...
+            logging.info(["delete_repo", u_value.account.name, u_value.name])
+            output.append(u_value)
+        else:
+            logging.info(["update_repo", u_value.account.name, u_value.name])
+            output.append(u_value)
```
chassing, line 208: Why is delete and not delete the same? If so you can write it as:

```
if u_value.delete:
    logging.info(["delete_repo", u_value.account.name, u_value.name])
else:
    logging.info(["update_repo", u_value.account.name, u_value.name])
output.append(u_value)
```

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,304 @@
...
+    """
+    State represents TFRepo data structures similar to their GQL representations.
+    In regards to deleting a Terraform Repo, when the delete flag is set to True, then
+    the state representation of this repo is also deleted even though the definition may
+    still exist in App Interface
+    """
+    created_and_updated = {**created, **updated}
+    try:
+        for cu_key, cu_val in created_and_updated.items():
+            if cu_val.delete is True:
```
chassing, line 227: Please write it as `if cu_val.delete:`

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,304 @@
...
+                    print_to_file
+                )
+            )
+    else:
+        print(yaml.safe_dump(data=action_plan.dict(), explicit_start=True))
+
+
+def early_exit_desired_state(*args: Any, **kwargs: Any) -> dict[str, Any]:
+    gqlapi = gql.get_api()
+    return {"repos": get_repos(query_func=gqlapi.query)}
```
chassing, line 304: Please convert the `list[TFRepo]` to a plain list of dictionaries; this will avoid unexpected side effects in the future. E.g. https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/skupper_network/integration.py#L307

reconcile/test/test_terraform_repo.py
chassing, line 1: You're basically just testing the `calculate_diff` method. Please add some tests for the other important methods too. e.g. `update_state`, `get_existing_state`. Karl introduced a nice `s3_state_builder` pytest fixture you may want to use. see it in [action](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/utils/test_promotion_state.py#L13)

reconcile/terraform_repo.py
hemslo: @chassing I noticed we are using this diff pattern a lot, also suggested something similar before, would it worth extracting a generic util function to diff? Given two mappings, return a symmetric difference result like

```
   only_in_a: [A]
   only_in_b: [B]
   both: [(A, B)]
```

reconcile/test/test_terraform_repo.py
cubismod: Gotcha, will do :)
I'll take a look at that fixture

reconcile/terraform_repo.py
cubismod: I was running into issues trying to serialize from JSON to the Pydantic model around https://github.com/app-sre/qontract-reconcile/pull/3464/files#diff-05e8f500c44ae2dc8a3b8087020cc1d2946506df84ea9a118e110c8d8ad82ba2R74

reconcile/terraform_repo.py
cubismod: Good question. This is the only integration for the forseeable future that will use this query but later on down the line others may use it.

reconcile/terraform_repo.py
cubismod: No there shouldn't be a case that the value is `None`

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,304 @@
...
+def check_ref(repo_url: str, ref: str, path: str) -> None:
+    """
+    Checks whether a Git ref is valid
+    and whether config.tf exists in the project path"""
+    instance = queries.get_gitlab_instance()
+    with GitLabApi(
+        instance, settings=queries.get_secret_reader_settings(), project_url=repo_url
+    ) as gl:
+        try:
+            gl.get_commit_sha(ref=ref, repo_url=repo_url)
```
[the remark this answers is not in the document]
cubismod: This is to validate that the SHA ref passed in actually exists

reconcile/terraform_repo.py
cubismod: I was also wondering about a generic diff function when writing this heh. Poking around it seems like most integrations take their own approach at performing diffs when it could be consolidated.

reconcile/terraform_repo.py
cubismod: However, I will review the docs and take another crack at just using the `TerraformRepoV1` class everywhere instead of `TFRepo` as it is just duplicated classes.

cubismod: Okay I figured that issue out. It was due to the `TerraformRepoV1` class using aliases for some of its fields

cubismod: Now using `TerraformRepoV1` rather than `TFRepo` custom class

reconcile/terraform_repo.py
hemslo: Cool, we can leave it as is for now, extract when reuse is needed.

reconcile/terraform_repo.py
hemslo: I created one in https://github.com/app-sre/qontract-reconcile/pull/3475, please see if it's easy to use. If we can try it in this new integration successfully, then we can apply it to others.

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,244 @@
...
+
+def get_existing_state(state: State) -> list[TerraformRepoV1]:
+    """Get the existing state of terraform repos from S3"""
+    repo_list: list[TerraformRepoV1] = []
+    keys = state.ls()
+    for key in keys:
+        if value := state.get(key.lstrip("/"), None):
+            repo = TerraformRepoV1.parse_raw(value)
+
+            if repo is not None:
```
chassing, line 55: I think `repo` can't be `None` here. This check isn't needed

reconcile/terraform_repo.py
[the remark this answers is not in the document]
cubismod: Tbh had no idea of this class! I will take a look at that

reconcile/terraform_repo.py
cubismod: Works for me!

reconcile/terraform_repo.py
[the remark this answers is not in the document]
cubismod: I subclassed the integration, just need to figure out the tests. Thanks for the suggestion.

reconcile/terraform_repo.py
cubismod: You're right. The `parse_raw` function does not return `None`, it returns a `ValidationError` exception instead

cubismod:
Resolved a bunch of comments as much of the boilerplate was able to be removed with Di's differ function. Working on improving test cases now.

reconcile/test/test_terraform_repo.py
cubismod: I've added some tests for updating state and getting existing state

reconcile/cli.py
```diff
@@ -1546,6 +1546,33 @@ def ldap_users(ctx, infra_project_id, app_interface_project_id):
     )
 
 
+@integration.command(short_help="Manages raw HCL Terraform from a separate repository.")
+@click.option(
+    "-d",
+    "--output-dir",
+    help="Specify a directory to individually output each repo plan to for the executor",
+)
+@click.option(
```
TGPSKI, line 1555: Curious when would we not validate git?

cubismod: Good point. I added that option to avoid calling the check_ref function https://github.com/app-sre/qontract-reconcile/pull/3464/files#diff-05e8f500c44ae2dc8a3b8087020cc1d2946506df84ea9a118e110c8d8ad82ba2R169-R180 during testing, although I don't think that's the correct pattern and I should probably just mock the GitLab API and query calls. If the integration can't reach and validate Git, then the executor also wouldn't be able to

TGPSKI: You could keep the parameter input but not expose it to the CLI?

cubismod: That works too

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,302 @@
...
+    validate_git: bool
+
+
+class TerraformRepoIntegration(
+    QontractReconcileIntegration[TerraformRepoIntegrationParams]
+):
+    def __init__(self, params: TerraformRepoIntegrationParams) -> None:
+        super().__init__(params)
+        self.qontract_integration = "terraform_repo"
+        self.qontract_integration_version = make_semver(0, 1, 0)
```
TGPSKI, line 69: Am I imagining that this is handled in different locations depending on the integration? Is this our current best practice location? 

Alternative and off topic, is anyone using these for anything? Or is this just sounds-good code cruft?

reconcile/terraform_repo.py
```diff
@@ -0,0 +1,302 @@
...
+        :type query_func: Callable
+        :return: list of Terraform repos or empty list if none are defined in A-I
+        :rtype: list[TerraformRepoV1]
+        """
+        query_results = query(query_func=query_func).repos
+        if query_results:
+            return query_results
+        return []
+
+    def get_existing_state(self, state: State) -> list[TerraformRepoV1]:
```
TGPSKI, line 138: General curiosity - as used, is State a terraform concept or a QR concept?

TGPSKI: It's the QR state. proceed

TGPSKI: Reading more below and also looking through the State utility. Confusing nomenclature but makes complete sense to me and its a great implementation strategy IMHO

TGPSKI (approved):
LGTM, :fire:, :rocket:

reconcile/terraform_repo.py
cubismod: The boilerplate setup for integrations isn't standardized and varies a lot depending on the integration. This new `QontractReconcileIntegration` class is being used in a few newer integrations and was suggested by Olivia in https://github.com/app-sre/qontract-reconcile/pull/3464#discussion_r1179208282

PR #3467, opened 2023-04-21
Add clusters_with_peering typed query

hemslo:
This is going to replace [queries.get_clusters_with_peering_settings](https://github.com/app-sre/qontract-reconcile/blob/80ea1a4d6e53a3cc4fec57cb8f6c14c19538a92f/reconcile/queries.py#L1160).

### <samp>🤖 Generated by Copilot at d76fb13</samp>

This pull request adds a new GraphQL query and related Python modules to query and access cluster peering information from the Qontract server. It also adds a test module to verify the functionality of the query and the data classes. The new query and modules are located in `reconcile/gql_definitions/common/clusters_with_peering.gql`, `reconcile/gql_definitions/common/clusters_with_peering.py`, `reconcile/typed_queries/clusters_with_peering.py`, and `reconcile/test/test_typed_queries/test_clusters_with_peering.py`.

### <samp>🤖 Generated by Copilot at d76fb13</samp>

*  Add a new GraphQL query definition file `clusters_with_peering.gql` to fetch clusters and their peering connections from the Qontract server ([link](https://github.com/app-sre/qontract-reconcile/pull/3467/files?diff=unified&w=0#diff-8a5cb60bd58734ab171f8daeea4e7a780b581f276e0bfdda5df0fc286b624202R1-R121))
* Generate a Pydantic data class module `clusters_with_peering.py` based on the query definition using the `qenerate` tool and the `pydantic_v1` plugin ([link](https://github.com/app-sre/qontract-reconcile/pull/3467/files?diff=unified&w=0#diff-ee7cd090300b01c05024fc10dbab5ca6c58da81ad2ad7f2e88306692a63a447dR1-R393))
* Provide a convenience function `query` in the generated module to query and parse the data into the data classes ([link](https://github.com/app-sre/qontract-reconcile/pull/3467/files?diff=unified&w=0#diff-ee7cd090300b01c05024fc10dbab5ca6c58da81ad2ad7f2e88306692a63a447dR1-R393))
* Define a function `get_clusters_with_peering` in a new module `reconcile/typed_queries/clusters_with_peering.py` that uses the `query` function to filter out the clusters without peering connections ([link](https://github.com/app-sre/qontract-reconcile/pull/3467/files?diff=unified&w=0#diff-325083ae3b256e2712c1e02ffd984a4fd6de2f48dbba3fc22004f0b9534d2187R1-R12))
* Add a test module `test_clusters_with_peering.py` to test the `get_clusters_with_peering` function using pytest and pytest-mock ([link](https://github.com/app-sre/qontract-reconcile/pull/3467/files?diff=unified&w=0#diff-439efa3cf8b97825c09ca95381012b6350fff5b0c700d1bf72a9b879ecaa7458R1-R83))
* Use fixtures to mock the data and the query function in the test module ([link](https://github.com/app-sre/qontract-reconcile/pull/3467/files?diff=unified&w=0#diff-439efa3cf8b97825c09ca95381012b6350fff5b0c700d1bf72a9b879ecaa7458R1-R83))
* Assert that the function returns the expected list of clusters with peering connections in the test module ([link](https://github.com/app-sre/qontract-reconcile/pull/3467/files?diff=unified&w=0#diff-439efa3cf8b97825c09ca95381012b6350fff5b0c700d1bf72a9b879ecaa7458R1-R83))


reconcile/typed_queries/clusters_with_peering.py
```diff
@@ -0,0 +1,12 @@
+from reconcile.gql_definitions.common.clusters_with_peering import (
+    ClusterV1,
+    query,
+)
+from reconcile.utils import gql
+
+
+def get_clusters_with_peering() -> list[ClusterV1]:
```
fishi0x01, line 8: Lets add `query_func` as Callable input arg -> then we can avoid `mocker.patch`.

fishi0x01: E.g., https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/typed_queries/app_interface_repo_url.py#L9

fishi0x01: `query_func` is also already available for these tests: https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/test_typed_queries/conftest.py

hemslo: I didn't add `query_func` to that function is because it will lead to the branch tested in test is not the one actually used in real call. For example these [2 lines](https://github.com/app-sre/qontract-reconcile/blob/80ea1a4d6e53a3cc4fec57cb8f6c14c19538a92f/reconcile/typed_queries/app_interface_repo_url.py#L11-L12) are not test covered.
So for those functions with `query_func`, there should be a test case using `macker.patch` to test `query_func` is `None` branch, feels not worth it.

fishi0x01: Shouldnt these 2 lines be tested somewhere else? I dont think that `gql.get_api()` should be responsibility of the typed_query module.

fishi0x01: But I agree - a test for this query is not really worth it - maybe we should add more tests for the gql module instead.

hemslo: All lines inside `get_clusters_with_peering` need to be covered by `get_clusters_with_peering` unit tests, the logic of `gql.get_api().query` can be extracted to a function in `gql` like `gql.get_query_func`, then in `get_cluster_with_peering` test just need to verify `gql.get_query_func` called.

hemslo: Actually in this function, there are 2 dependencies `reconcile.gql_definitions.common.clusters_with_peering.query` and `reconcile.utils.gql`, to do it in DI way, both of them need to be injected, just not very common in python.

fishi0x01: You are right. Nice idea! I like the idea of `gql.get_query_func()` a lot - very clean! :heart: :+1:

fishi0x01 (approved):
Thanks for adding a new query! :bow:

reconcile/typed_queries/clusters_with_peering.py
hemslo: @fishi0x01 I tried to inject `query_func` to make it using DI, but mypy is not happy about it, and I feel tests are harder to understand. For example:

```python
# gql.py

P = ParamSpec("P")
R = TypeVar("R")


def inject_query_func(func: Callable[Concatenate[Callable, P], R]) -> Callable[P, R]:
    @wraps(func)
    def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
        query_func = get_query_func()
        return func(query_func, *args, **kwargs)

    return wrapper
```

```python
# clusters_with_peering.py

@gql.inject_query_func
def get_clusters_with_peering(query_func: Callable) -> list[ClusterV1]:
    data = query(query_func)
    clusters = data.clusters or []
    return [c for c in clusters if c.peering is not None]
```

```python
# test_clusters_with_peering.py

def test_get_clusters_with_peering(
    query_func: Callable[..., Callable],
    clusters_with_peering: ClustersWithPeeringQueryData,
) -> None:
    qf = query_func(clusters_with_peering.dict(by_alias=True))

    clusters = get_clusters_with_peering.__wrapped__(qf)

    assert clusters == clusters_with_peering.clusters
```

It works but `mypy` doesn't know `__wrapped__` (`error: "Callable[[], List[ClusterV1]]" has no attribute "__wrapped__"  [attr-defined]`), and we still need to write another test case to verify `get_clusters_with_peering` has proper `query_func` injection by `@gql.inject_query_func`.

Without a DI framework, `mocker.patch` is the only way to ensure full test coverage, I don't like to have one condition branch just used in tests, I had cases where all tests pass but still fail in real run, even that function looks like covered. What do you think about it?

hemslo:
@fishi0x01 I added a new [refactor](https://github.com/app-sre/qontract-reconcile/pull/3467/commits/5d574db737fc9e53a648970170cba17bbad95a87) based on https://github.com/app-sre/qontract-reconcile/pull/3467#discussion_r1173631409, please have a look, thanks!

reconcile/typed_queries/clusters_with_peering.py
fishi0x01: Yeah mypy and pytest are not really working together well. You will see a lot of `ignore [attr-defined]` comments in tests :sweat_smile: 

If you mock the whole gql api as you suggested above, then the following wouldnt work?

**conftest.py**
```python
@pytest.fixture
def gql() -> Callable[..., GqlApi]:
    def builder(data: Mapping) -> GqlApi:
        def query_func():
            return data
        mock = create_autospec(spec=GqlApi)
        mock.get_query_func.side_effect = [query_func]
        return mock
    return builder
```

**test.py**
```python
def test(gql, gql_class_factory):
    mock = gql(gql_class_factory(CLS, {...}).dict(by_alias=True))
    resp = my_typed_query(gql=mock)
    assert resp == ...
```

I.e., all complexity is in the `conftest.py` and the tests only use the builder function to get a `GqlApi` mock? Not 100% sure, but my understanding was that this mock can then be injected, so you still test all the lines in typed_queries module.

hemslo: Refactored with `gql_api` as dependency https://github.com/app-sre/qontract-reconcile/pull/3467/commits/a25cba1c0872677770d0e697922ee0433a8dcfcc

reconcile/typed_queries/clusters_with_peering.py
```diff
@@ -0,0 +1,11 @@
+from reconcile.gql_definitions.common.clusters_with_peering import (
+    ClusterV1,
+    query,
+)
+from reconcile.utils.gql import GqlApi
+
+
+def get_clusters_with_peering(gql_api: GqlApi) -> list[ClusterV1]:
```
fishi0x01, line 8: This is great :heart:

PR #3469, opened 2023-04-21
[change-owners] empty self-service-role must render a change non-self-serviceable

geoberle:
# Motivation

If a change is covered by a self-service role without members, nobody can approve the change while at the same time the MR is still marked as self-serviceable. This way it does not show up in the review queue.

But even if IC gets aware of the MR, they can't /lgtm or add an lgtm label. The only way out is a manual force merge by IC.

# What changed?

Refine the term self-serviceable. If a change is only covered by a self-service role without members, the change is not considered self-serviceable. This way AppSRE IC will get aware of the change via the review queue and can approve the regular way.

The change coverage report in the MR should still show that the change would be self-serviceable but is not because of the empty role.

part of APPSRE-7535

### DIFF
--- a/reconcile/change_owners/change_owners.py
+++ b/reconcile/change_owners/change_owners.py
@@ -3,7 +3,10 @@ import sys
 import traceback
 
 from reconcile import queries
-from reconcile.change_owners.approver import GqlApproverResolver
+from reconcile.change_owners.approver import (
+    Approver,
+    GqlApproverResolver,
+)
 from reconcile.change_owners.bundle import (
     BundleFileType,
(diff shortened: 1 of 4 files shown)


maorfr:
another alternative is to not allow orphaned self-serviceable changes:
https://github.com/app-sre/qontract-reconcile/blob/ae918c6f2f201230d661b84aa3b3c6d6d5400dbe/reconcile/utils/saasherder/saasherder.py#L246-L255

rporres (approved):
looks great, thx

geoberle:
closing this in favour of maor idea

PR #3473, opened 2023-04-25
[openshift saas deploy] multiple publishers

chassing:
The deployment information (promotion data) is stored in an S3 bucket. Make the S3 key unique to avoid overwrites if multiple publishers are using the same channel name.

`promotions/{channel}/{commit_sha}`  -> `/deployments/{saas_target_uid}/{commit_sha}`

`saas_target_uid` is an unique blake2s hash composed of `{saas_name}/{resource_template_name}/{target_name|"default"}/{cluster_name}/namespace_name}`

Ticket: [APPSRE-7414](https://issues.redhat.com/browse/APPSRE-7414)

### DIFF
--- a/reconcile/saas_auto_promotions_manager/utils/saas_files_inventory.py
+++ b/reconcile/saas_auto_promotions_manager/utils/saas_files_inventory.py
@@ -30,6 +30,7 @@ class SaasFilesInventory:
         self._channels_by_name: dict[str, Channel] = {}
         self.subscribers: list[Subscriber] = []
         self.publishers: list[Publisher] = []
+        self._assemble_channels()
         self._assemble_subscribers_with_auto_promotions()
         self._assemble_publishers()
         self._remove_unsupported()
@@ -66,6 +67,7 @@ class SaasFilesInventory:
                     )
                     publisher = Publisher(
                         ref=target.ref,
+                        uid=target.uid(saas_file.name, resource_template.name),
                         repo_url=resource_template.url,
                         auth_code=auth_code,
                     )
@@ -149,13 +151,6 @@ class SaasFilesInventory:
         for subscriber in self.subscribers:
             is_supported = True
             for channel in subscriber.channels:
-                if len(channel.publishers) > 1:
-                    logging.error(
-                        "[%s] We do not support multiple publishers for a single channel - blocked by https://issues.redhat.com/browse/APPSRE-7414",
-                        channel.name,
-                    )
-                    is_supported = False
-                    break
                 if not channel.publishers:
                     logging.error(
                         "[%s] There must be at least one publisher per channel.",
@@ -163,18 +158,7 @@ class SaasFilesInventory:
                     )
                     is_supported = False
                     break
-                if (
-                    len(subscriber.config_hashes_by_channel_name.get(channel.name, []))
-                    > 1
-                ):
-                    logging.error(
-                        "[%s] We do not support multiple publishers for a single channel - blocked by https://issues.redhat.com/browse/APPSRE-7414",
-                        channel.name,
-                    )
-                    is_supported = False
-                    break
+
             if is_supported:
                 supported_subscribers.append(subscriber)
         self.subscribers = supported_subscribers
-        # Ideally we also remove the publishers that are left w/o subscriber.
-        # But lets solve APPSRE-7414 - then it wont be necessary in the first place.
--- a/reconcile/saas_auto_promotions_manager/utils/saas_files_inventory.py
+++ b/reconcile/saas_auto_promotions_manager/utils/saas_files_inventory.py
@@ -30,29 +30,9 @@ class SaasFilesInventory:
         self._channels_by_name: dict[str, Channel] = {}
         self.subscribers: list[Subscriber] = []
         self.publishers: list[Publisher] = []
-        self._assemble_channels()
         self._assemble_subscribers_with_auto_promotions()
         self._assemble_publishers()
         self._remove_unsupported()
-
-    def _assemble_channels(self) -> None:
-        for saas_file in self._saas_files:
-            for resource_template in saas_file.resource_templates:
-                for target in resource_template.targets:
-                    if not target.promotion:
-                        continue
-                    for publish_channel in target.promotion.publish or []:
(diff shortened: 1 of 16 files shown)


reconcile/saas_auto_promotions_manager/utils/saas_files_inventory.py
```diff
@@ -30,6 +30,7 @@ def __init__(self, saas_files: Iterable[SaasFile]):
         self._channels_by_name: dict[str, Channel] = {}
         self.subscribers: list[Subscriber] = []
         self.publishers: list[Publisher] = []
+        self._assemble_channels()
```
chassing, line 33: @fishi0x01, adding the initialization of all channels; without that, only channels with active subscribers are honored. Am I right?

lechuk47:
There is validation to prevent duplicate channel names when saas files are validated if I recall correctly: https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/saasherder/saasherder.py#L394

chassing:
> There is validation to prevent duplicate channel names when saas files are validated if I recall correctly: https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/saasherder/saasherder.py#L394

Thanks for highlighting that. `openshift-saas-deploy` doesn't reach this validation check, but I created another jira ticket([APPSRE-7563](https://issues.redhat.com/browse/APPSRE-7563)) to adapt `saasherder`

reconcile/typed_queries/saas_files.py
```diff
@@ -75,6 +76,13 @@ class Config:
         # ignore `namespaceSelector` and 'provider' fields from the GQL schema
         extra = Extra.ignore
 
+    def uid(self, saas_file_name: str, resource_template_name: str) -> str:
+        """Returns a unique identifier for a target."""
+        return hashlib.blake2s(
+            f"{saas_file_name}:{resource_template_name}:{self.name if self.name else 'default'}:{self.namespace.cluster.name}:{self.namespace.name}".encode(),
```
patjlm, line 82: should we use the saas-file `name` field instead of `saas_file_name`?

chassing: I don't know what you mean. `saas_file_name` here is the `name` attribute of the saas-file. The context here (`self`) is the target.

patjlm: ok, i thought it was the actual file name... sorry!

reconcile/typed_queries/saas_files.py
```diff
@@ -75,6 +76,13 @@ class Config:
         # ignore `namespaceSelector` and 'provider' fields from the GQL schema
         extra = Extra.ignore
 
+    def uid(self, saas_file_name: str, resource_template_name: str) -> str:
```
fishi0x01, line 79: Does it make sense to have these args as attributes instead? I.e., when creating `SaasResourceTemplateTarget` also add the parents `SaasResourceTemplate.name` and `SaasFile.name`?

chassing: Unfortunately, this isn't easy to achieve. See https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/typed_queries/saas_files.py#L248 how the `SaasFile` gets composed.

fishi0x01: Oh I see.

In that case, does it make sense to extend the `export_model` function? It is only used for the purpose of converting gql model data to our rendered saas file data right?

```python
class SaasResourceTemplateTarget(ConfiguredBaseModel):
    ...
    # By not specifying the field parameter, we should be fine if those are missing in dict data
    parent_saas_name: Optional[str]
    parent_template_name: Optional[str]


def export_model(model: SaasFileV2) -> SaasFile:
    saas_file = SaasFile(**convert_parameters_to_json_string(model.dict(by_alias=True)))
    for tmpl in saas_file.templates:
        for target in tmpl.targets:
            target.parent_saas_name = saas_file.name
            target.parent_template_name = tmpl.name
    return saas_file
```

Not 100% sure it is worth the effort - might as well be Im completely missing sth here. My naive feeling here is that it would be nice for the consumer to not care about saas_file name etc anylonger and just be able to call `.uid()`.

chassing: Please take a look at https://github.com/app-sre/qontract-reconcile/pull/3473/commits/cd8bc893ee72f8ec3b90c57e08a37fd2577ea4d5

reconcile/saas_auto_promotions_manager/utils/saas_files_inventory.py
```diff
@@ -34,25 +34,6 @@ def __init__(self, saas_files: Iterable[SaasFile]):
...
-                            self._channels_by_name[publish_channel] = Channel(
-                                name=publish_channel,
-                                publishers=[],
-                            )
-                    for subscribe_channel in target.promotion.subscribe or []:
-                        if subscribe_channel not in self._channels_by_name:
-                            self._channels_by_name[subscribe_channel] = Channel(
-                                name=subscribe_channel,
-                                publishers=[],
-                            )
```
fishi0x01, line 54: Thanks for catching this and cleanup! :bow:

reconcile/utils/promotion_state.py
```diff
@@ -47,28 +47,51 @@ def cache_commit_shas_from_s3(self) -> None:
...
-            # Format: /promotions/{channel}/{commit-sha}
-            if not commit.startswith("/promotions/"):
+            # for backwards compatibility - remove this after a while
+            if commit.startswith("/promotions/"):
+                # Format: /promotions/{channel}/{commit-sha}
+                _, _, channel_name, commit_sha = commit.split("/")
+                self._commits_by_channel[channel_name].add(commit_sha)
+            # / for backwards compatibility - remove this after a while
+
+            # Format: /deployments/{saas-target-uid}/{channel}/{commit-sha}
```
fishi0x01, line 57: nit: I wonder if it would make sense to put the channel first:

```
/deployments/{channel}/{saas-target-uid}/{commit-sha}
```

That way, it is easier to navigate through the bucket when debugging sth, because usually you are looking for the channel first.

PR #3475, opened 2023-04-27
Add a generic differ function to get diff result of 2 mappings/lists

hemslo:
This is a generic differ function can be used to simplify repeated logic in multiple places.

It supports diff 2 mappings, 2 simple lists, or 2 lists with key selectors.

Based on comments from https://github.com/app-sre/qontract-reconcile/pull/3464#discussion_r1177584112.

### <samp>🤖 Generated by Copilot at ac0202a</samp>

This pull request introduces the `differ` module, which provides functions and data classes to compare mappings and iterables. It also adds unit tests for the `differ` module in the `reconcile/test/utils` directory. The `differ` module is used to compare the current and desired states of various resources managed by the qontract-reconcile tool.

### <samp>🤖 Generated by Copilot at ac0202a</samp>

*  Implement a generic and reusable way to compare mappings and iterables ([link](https://github.com/app-sre/qontract-reconcile/pull/3475/files?diff=unified&w=0#diff-92d30bcad821fdb092179a4ba46aa1d77b5236d0e96c94856e607ca5e4c1e648R1-R205))
* Add unit tests for the `differ` module ([link](https://github.com/app-sre/qontract-reconcile/pull/3475/files?diff=unified&w=0#diff-aeaafcb70467dc4315824cf6e4cc3eaf3508822fd4b9eb7ea5040164454cf963R1-R217))

### DIFF
--- a/reconcile/utils/differ.py
+++ b/reconcile/utils/differ.py
@@ -5,10 +5,9 @@ from collections.abc import (
 )
 from dataclasses import dataclass
 from typing import (
+    Any,
     Generic,
-    Optional,
     TypeVar,
-    cast,
 )
 
 T = TypeVar("T")
@@ -35,18 +34,14 @@ def _default_equal(current: Current, desired: Desired) -> bool:
     return current == desired
 
 
-def _default_current_key(current: Current) -> Key:
-    return cast(Key, current)
-
-
-def _default_desired_key(desired: Desired) -> Key:
-    return cast(Key, desired)
+def _default_key(item: Any) -> Any:
+    return item
 
 
 def diff_mappings(
     current: Mapping[Key, Current],
     desired: Mapping[Key, Desired],
-    equal: Optional[Callable[[Current, Desired], bool]] = None,
+    equal: Callable[[Current, Desired], bool] = _default_equal,
 ) -> DiffResult[Current, Desired, Key]:
     """
     Compare two mappings and return a `DiffResult` instance containing the differences between them.
@@ -55,9 +50,9 @@ def diff_mappings(
     :type current: Mapping[Key, Current]
     :param desired: The desired mapping to compare.
     :type desired: Mapping[Key, Desired]
-    :param equal: An optional function that compares two elements in the mappings and returns True if they are equal.
-        If not provided, the default behavior is to use the `==` operator.
-    :type equal: Optional[Callable[[Current, Desired], bool]]
+    :param equal: A function that compares two elements in the mappings and returns True if they are equal.
+        The default behavior is to use the `==` operator.
+    :type equal: Callable[[Current, Desired], bool]
     :return: A `DiffResult` instance containing the differences between the `current` and `desired` mappings,
         including elements that were added, deleted, changed or identical.
     :rtype: DiffResult[Current, Desired, Key]
@@ -74,8 +69,6 @@ def diff_mappings(
             identical={'i': DiffPair(current=1, desired=1)},
         )
     """
-    if equal is None:
-        equal = _default_equal
     add = {k: desired[k] for k in desired.keys() - current.keys()}
     delete = {k: current[k] for k in current.keys() - desired.keys()}
     change, identical = {}, {}
@@ -96,9 +89,9 @@ def diff_mappings(
 def diff_any_iterables(
     current: Iterable[Current],
     desired: Iterable[Desired],
-    current_key: Optional[Callable[[Current], Key]] = None,
-    desired_key: Optional[Callable[[Desired], Key]] = None,
-    equal: Optional[Callable[[Current, Desired], bool]] = None,
+    current_key: Callable[[Current], Key] = _default_key,
+    desired_key: Callable[[Desired], Key] = _default_key,
+    equal: Callable[[Current, Desired], bool] = _default_equal,
 ) -> DiffResult[Current, Desired, Key]:
     """
     Compare two iterables and return a `DiffResult` instance containing the differences between them.
@@ -107,15 +100,15 @@ def diff_any_iterables(
     :type current: Iterable[Current]
     :param desired: The desired iterable to compare.
     :type desired: Iterable[Desired]
-    :param current_key: An optional function that returns the key for an element in the `current` iterable.
+    :param current_key: A function that returns the key for an element in the `current` iterable.
         If not provided, the default behavior is to use the element itself as the key.
-    :type current_key: Optional[Callable[[Current], Key]]
-    :param desired_key: An optaionl function that returns the key for an element in the `desired` iterable.
+    :type current_key: Callable[[Current], Key]
+    :param desired_key: A function that returns the key for an element in the `desired` iterable.
         If not provided, the default behavior is to use the element itself as the key.
-    :type desired_key: Optional[Callable[[Desired], Key]]
-    :param equal: An optional function that compares two elements in the iterables and returns True if they are equal.
+    :type desired_key: Callable[[Desired], Key]
+    :param equal: A function that compares two elements in the iterables and returns True if they are equal.
         If not provided, the default behavior is to use the `==` operator.
-    :type equal: Optional[Callable[[Current, Desired], bool]]
+    :type equal: Callable[[Current, Desired], bool]
     :return: A `DiffResult` instance containing the differences between the `current` and `desired` iterables,
         including elements that were added, deleted, changed or identical.
     :rtype: DiffResult[Current, Desired, Key]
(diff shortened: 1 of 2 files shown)


reconcile/utils/differ.py
```diff
@@ -0,0 +1,52 @@
...
+from dataclasses import dataclass
+from typing import (
+    Generic,
+    Optional,
+    TypeVar,
+)
+
+C = TypeVar("C")
+D = TypeVar("D")
+K = TypeVar("K")
```
chassing, line 14: `C, D, K`? Please use descriptive names 🙇

reconcile/utils/differ.py
```diff
@@ -0,0 +1,52 @@
...
+@dataclass(frozen=True, eq=True)
+class DiffResult(Generic[C, D, K]):
+    add: dict[K, D]
+    delete: dict[K, C]
+    change: dict[K, tuple[C, D]]
+
+
+def diff(
+    current: Mapping[K, C],
+    desired: Mapping[K, D],
```
chassing, line 26: Why are you using `Mapping`s as parameter? Most of the time, we've different `list` to compare to.

hemslo: This is based on https://github.com/app-sre/qontract-reconcile/blob/8257677076a73eb79443559747edef376e872c93/reconcile/openshift_base.py#L585

which I worked on some parts before, where are we using `list` diff?

hemslo: @chassing How about 2 functions, one for diff dict, another for diff lists?

chassing: Here are some examples:

* https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/ocm_oidc_idp.py#L48
* https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/aus/base.py#L108
* https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/openshift_users.py#L102
* https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/openshift_groups.py#L114

chassing: > @chassing How about 2 functions, one for diff dict, another for diff lists?

Sounds good to me!

reconcile/utils/differ.py
hemslo: I was thinking just `T1`, `T2`, `TR` like most generic type templates, I can give them full names, just function signature will become very long...

reconcile/utils/differ.py
hemslo: Now there are 3 functions for different usages:
* `diff_mappings` for 2 dicts
* `diff_by_key` for 2 iterables, this is handy when no `__eq__` and `__hash__` overrides
* `diff_lists` for 2 lists, can be simple elements or with `__eq__` and `__hash__` overrides

chassing: I think you should return a standard `DiffResult` class instead of differentiating between mappings and lists. I think the most common use-case will be something like the following. Imaging `Foo` is class from `fetch_desired_state`:

```python
@dataclass
class Foo:
    name: str
    value: int

    def __eq__(self, other: object) -> bool:
        if not isinstance(other, Foo):
            return False

        return self.name == other.name and self.value == other.value

    def __hash__(self) -> int:
        return hash(self.name)


a_foos = [Foo(name="foo", value=1), Foo(name="bar", value=2)]
b_foos = [Foo(name="foo", value=2), Foo(name="lala", value=2)]

print(
    diff(
        a_foos,
        b_foos,
        key=lambda x: x.name,
    )
)
# DiffResult(add={'lala': Foo(name='lala', value=2)}, delete={'bar': Foo(name='bar', value=2)}, change={'foo': Foo(name='foo', value=2)})

```

and 
```python
def diff(current, desired, key, equal: Callable = None) ...
    if not equal:
        equal = lambda x, y: x == y
    ...
```

And having just the desired item in `change`. What do you think?

hemslo: * How to use one `DiffResult` to support both dict and list? If diff 2 lists without key func like https://github.com/app-sre/qontract-reconcile/blob/8257677076a73eb79443559747edef376e872c93/reconcile/openshift_users.py#L131 

* I usually like to write immutable code, once a variable is assigned, never assign it to other value, reassign function parameter sometimes cause confusion, maybe OK in this case as it's very common in python to set default value this way, just a habit from eslint rule [no-param-reassign](https://eslint.org/docs/latest/rules/no-param-reassign)

* in some cases we want to zip `current` with `desired` to do a compare, for example https://github.com/app-sre/qontract-reconcile/blob/8257677076a73eb79443559747edef376e872c93/reconcile/ocm_oidc_idp.py#L141, it will be very handy to return a tuple in `change` than retrieve them later like current code using `O(N^2)` time complexity.

hemslo: and sometimes current and desired are not using the same type, that's why I split them and provide two key func as parameters, the signature is very similar to `Join` in [Linq](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.join).

chassing: > How to use one DiffResult to support both dict and list? ...
I mean having the same attributes: add, delete, change. In your `DiffListResult` you have identical instead of change.

>  If diff 2 lists without key func like

`key=lamda x: x`?

> ... maybe OK in this case as it's very common in python to set default value this way ...

Yes, that's is very common pattern in python.

> ... it will be very handy to return a tuple in change than retrieve them ...

But this isn't very readable:

```python
changes = diff(
    a_foos,
    b_foos,
    key=lambda x: x.name,
)

delete_foos(changes.delete)
add_foos(changes.add)
update_foos(changes.change[1]) <- I don't like that
```

hemslo: I use `identical` instead of `change` for list because that's the result of intersect, name it `change` seems strange, they are `__eq__` but maybe changed... To keep structure consistent, I think rename it to `change` is OK. I'll see if I can use a single `DIffResult` class using generic now that they have same fields.

> key=lamda x: x

Good idea, then maybe we can always use the old `DiffResult` structure. 

Regarding `change`, how about use named tuple or another dataclass with `current` and `desired` as fields?

chassing: > Regarding change, how about use named tuple or another dataclass with current and desired as fields?

Yes, sure, sounds good. 🙇

cubismod (approved):
Thank you for doing this Di! I tested in my code by using the `diff_by_key` function and it cuts down a lot of boilerplate so this will work well for my purposes. See https://github.com/cubismod/qontract-reconcile/commit/29d6fc8cc6198fea741995a2aec57e3e424c9681#diff-05e8f500c44ae2dc8a3b8087020cc1d2946506df84ea9a118e110c8d8ad82ba2R251

reconcile/utils/differ.py
```diff
@@ -0,0 +1,215 @@
...
+            add={'a': {'name': 'a', 'value': 30}},
+            delete={'d': {'name': 'd', 'value': 3}},
+            change={'c': DiffPair(current={'name': 'c', 'value': 2}, desired={'name': 'c', 'value': 20})),
+            identical={'i': DiffPair(current={'name': 'i', 'value': 1}, desired={'name': 'i', 'value': 1})},
+        )
+    """
+    if current_key is None:
+        current_key = _default_current_key
+    if desired_key is None:
+        desired_key = _default_desired_key
```
chassing, line 152: You can set both defaults in the method definition instead. 

```python
def diff_any_iterables(
     current: Iterable[Current],
     desired: Iterable[Desired],
     current_key: Optional[Callable[[Current], Key]] = _default_current_key,
     desired_key: Optional[Callable[[Desired], Key]] = _default_current_key,
     equal: Optional[Callable[[Current, Desired], bool]] = None,
 ) -> DiffResult[Current, Desired, Key]:
```

chassing (approved):
Just another tiny nit. Thanks for your work!

reconcile/utils/differ.py
hemslo: Oh, thanks! I'm struggling making mypy happy, let me try this.

hemslo: It works!

hemslo:
@cubismod I added a new simpler function `diff_iterables` to diff lists with the same type, and same key func, this can save one argument.

```python
        result = diff_iterables(
            existing_state,
            desired_state,
            lambda x: x.name,
        )
```

rporres (approved):
So cool! Nice read of the code and the discussions leading to the final result! ❤️

PR #3476, opened 2023-04-27
fix pypi release

BumbleFeng:
latest urllib3 version [2.0.0](https://pypi.org/project/urllib3/2.0.0/) does not work with requests. https://github.com/psf/requests/issues/6432


hemslo (approved):
LGTM! Thanks! I wish we have a better dependency manager with containerised run for these tasks.

rporres (approved):

PR #3477, opened 2023-04-30
[jenkins-webhooks] allow no push hook

maorfr:
part of https://issues.redhat.com/browse/APPSRE-7536

the `jenkins-webhooks` integration creates webhooks from gitlab to jenkins based on the job - either a hook for a merge request or a hook for a push.

in this [MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/66494) we added an option for a build job to not be triggered by a push. in such a case, there should be no webhook created:
```
        triggers:
        - gitlab:
            trigger-push: false
```
ref: https://jenkins-job-builder.readthedocs.io/en/latest/triggers.html#triggers.gitlab

### DIFF
--- a/reconcile/utils/jjb_client.py
+++ b/reconcile/utils/jjb_client.py
@@ -293,7 +293,10 @@ class JJB:  # pylint: disable=too-many-public-methods
                     project_url = project_url_raw.strip("/").replace(".git", "")
                     gitlab_triggers = job["triggers"][0]["gitlab"]
                     mr_trigger = gitlab_triggers["trigger-merge-request"]
-                    trigger = "mr" if mr_trigger else "push"
+                    push_trigger = gitlab_triggers["trigger-push"]
+                    trigger = "mr" if mr_trigger else "push" if push_trigger else None
+                    if trigger is None:
+                        continue
                     hook = {
                         "job_url": job_url,
                         "trigger": trigger,


reconcile/utils/jjb_client.py
```diff
@@ -293,7 +293,10 @@ def get_job_webhooks_data(self):
                     project_url = project_url_raw.strip("/").replace(".git", "")
                     gitlab_triggers = job["triggers"][0]["gitlab"]
                     mr_trigger = gitlab_triggers["trigger-merge-request"]
-                    trigger = "mr" if mr_trigger else "push"
+                    push_trigger = gitlab_triggers["trigger-push"]
+                    trigger = "mr" if mr_trigger else "push" if push_trigger else None
```
hemslo, line 297: nit: this logic is a little nested, can we flatten it like this:

```python
mr_trigger = "mr" if gitlab_triggers["trigger-merge-request"] else None
push_trigger = "push" if gitlab_triggers["trigger-push"] else None
trigger = mr_trigger or push_trigger
```

hemslo (approved):
LGTM!

reconcile/utils/jjb_client.py
maorfr: a nit indeed

maorfr: i'll leave things the way they are for now if you don't mind. i personally find it more readable.

hemslo: @maorfr it wasn't very clear to me, I was wondering which one has higher priority, it turned out `trigger-merge-request` is higher than `trigger-push`, the nested one line `if else` hide it from me because I wasn't sure about the evaluation order.

For example, `A if B else C if D else E`, will `D` be evaluated before or after `B`, in this case `B` will run first, but it is not always first expression first run in other python expressions like comprehensions. I don't like nested expressions in general, quote zen of python `Flat is better than nested.`

To me, the best way to describe this logic is to extract a function like:

```python
def get_trigger(gitlab_triggers):
  if gitlab_triggers["trigger-merge-request"]:
    return "mr"
  if gitlab_triggers["trigger-push"]:
    return "push"
  return None
```

maorfr: i challenge you to find the real problem with the code :wink:

maorfr: also notice this integration changes so rarely, so there is a very low ROI here

hemslo: There are 2 ways of coding, do it complex so there is no obvious problem, or do it simple so there is obviously no problem. I prefer the later.

hemslo: you see, a cleaner approach will reveal this easily as `trigger = mr_trigger or push_trigger` will indicate there can be at most one of them, that's the assumption I made from reading original code

hemslo: If we want to create different triggers in different conditions, I suggest we extract a function

```python
def get_trigger(gitlab_triggers):
    triggers = []
    if gitlab_triggers["trigger-merge-request"]:
      triggers.append("mr")
    if gitlab_triggers["trigger-push"]:
      triggers.append("push")
    return triggers
```

then we can create iterate that to create triggers, it can handle no trigger case as well.

hemslo: I checked gitlab [add-project-hook api](https://docs.gitlab.com/ee/api/projects.html#add-project-hook), one webhook can support multiple triggers, so one webhook is enough here.

Instead of iterate on `triggers` to create multiple webhooks, we can check it like `mr in triggers` in https://github.com/app-sre/qontract-reconcile/blob/47761a04b50a07df8464a36dc34710102c23ba18/reconcile/utils/gitlab_api.py#L537-L551.

Btw it seems we only have `create_project_hook` but no `update` or `delete`, also create only in [jenkins_webhooks](https://github.com/app-sre/qontract-reconcile/blob/47761a04b50a07df8464a36dc34710102c23ba18/reconcile/jenkins_webhooks.py). So existing webhooks won't be synced.

maorfr: this is by design, to allow users to define their own webhooks. there is another integration to cleanup, but it's mostly for migrations.

PR #3480, opened 2023-05-02
fix early-exit method signature of tf-tgw-attachments

geoberle:
fix early-exit method signature of tf-tgw-attachments

 ```
TypeError: early_exit_desired_state() takes from 0 to 3 positional arguments but 4 were given
```


chassing:
Your failing PR checks will be fixed with #3483

janboll (approved):

PR #3486, opened 2023-05-03
[saas-file-validator] validate targets in app

maorfr:
depends on https://github.com/app-sre/qontract-schemas/pull/438

this PR adds a validation that when `validateTargetsInApp` is set to `true` all targets are validated to be in the same app as the saas file itself.

with such an attribute, we can grant saas file owner full permissions to `targets`.


patjlm (approved):

maorfr:
https://issues.redhat.com/browse/APPSRE-7587

PR #3488, opened 2023-05-04
Improve runtime of aws-iam-password-reset

janboll:
Remove saas files from roles query
Initialize AWS client only if it is required

https://issues.redhat.com/browse/APPSRE-7583

### DIFF
--- a/reconcile/aws_iam_password_reset.py
+++ b/reconcile/aws_iam_password_reset.py
@@ -38,7 +38,7 @@ def account_in_roles(roles: Iterable[Mapping[str, Any]], aws_account: str) -> bo
 def run(dry_run, defer=None):
     accounts = queries.get_aws_accounts(reset_passwords=True)
     settings = queries.get_app_interface_settings()
-    roles = queries.get_roles(aws=True)
+    roles = queries.get_roles(aws=True, saas_files=False)
     state = init_state(integration=QONTRACT_INTEGRATION)
     defer(state.cleanup)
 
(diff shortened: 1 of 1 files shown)


reconcile/aws_iam_password_reset.py
```diff
@@ -47,27 +47,35 @@ def run(dry_run, defer=None):
...
+
+            if aws_api is None:
+                aws_api = AWSApi(1, [a], settings=settings)
+
+            aws_api.reset_password(account_name, user_name)
+            aws_api.reset_mfa(account_name, user_name)
+            state.add(state_key)
+
+        if aws_api:
+            aws_api.cleanup()
```
geoberle, line 81: this cleanup might not happen if the integration fails for whatever reason before it reaches this step

janboll: Right, this should be fine, the cleanup is removing sessions. 
Nevertheless, I'll change it slightly to use context-managers again

janboll:
Tested locally by resetting my account in dev

reconcile/aws_iam_password_reset.py
```diff
@@ -34,40 +36,66 @@ def account_in_roles(roles: Iterable[Mapping[str, Any]], aws_account: str) -> bo
...
+                logging.error(f"User {user_name} is not in account {account_name}")
+                sys.exit(1)
+
+            accounts_to_reset.append(
+                AwsProfileToReset(
+                    **{
+                        "account": a,
+                        "user_name": user_name,
+                        "state_key": state_key,
+                    }
```
geoberle, line 83: can you pass the arguments directly?

reconcile/aws_iam_password_reset.py
```diff
@@ -34,40 +36,66 @@ def account_in_roles(roles: Iterable[Mapping[str, Any]], aws_account: str) -> bo
...
+                )
+            )
+
+    for a in accounts_to_reset:
+        with AWSApi(1, [a.account], settings=settings) as aws_api:
```
geoberle, line 88: wouldn't this establish a new AWS client for each user password to reset, even if all are in the same AWS account?

janboll: Do you reset accounts for others? :smiling_imp:

geoberle: thanks!

geoberle (approved):
LGTM

PR #3489, opened 2023-05-04
[change-owners] make file moves self-serviceable

geoberle:
Moving files without changing their content is not necessarily self-serviceable via change-types. Since a-i is path-agnostic, we don't really care too much about path changes.

this PR introduces the ability to aggreate an delete/add change that represents a file move, into one change that only has the path as the changing attribute.

https://issues.redhat.com/browse/APPSRE-7590

### DIFF
--- a/reconcile/change_owners/change_owners.py
+++ b/reconcile/change_owners/change_owners.py
@@ -1,6 +1,7 @@
 import logging
 import sys
 import traceback
+from collections import defaultdict
 
 from reconcile import queries
 from reconcile.change_owners.approver import GqlApproverResolver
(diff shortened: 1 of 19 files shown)


reconcile/change_owners/change_owners.py
```diff
@@ -120,7 +121,71 @@ def fetch_bundle_changes(comparison_sha: str) -> list[BundleFileChange]:
...
+            # if there are two candidates, they could represent a file move.
+            # lets check if that is the case
+            deletion = (
+                candidates[0] if candidates[0].is_file_deletion() else candidates[1]
+            )
+            addition = (
+                candidates[0] if candidates[0].is_file_creation() else candidates[1]
+            )
+            if (
+                deletion != addition
```
hemslo, line 166: Is there a case that `deletion` and `addition` can be the same one?

reconcile/change_owners/change_owners.py
```diff
@@ -120,7 +121,71 @@ def fetch_bundle_changes(comparison_sha: str) -> list[BundleFileChange]:
...
+    for candidates in move_candidates.values():
+        if len(candidates) == 2:
+            # if there are two candidates, they could represent a file move.
+            # lets check if that is the case
+            deletion = (
+                candidates[0] if candidates[0].is_file_deletion() else candidates[1]
+            )
+            addition = (
+                candidates[0] if candidates[0].is_file_creation() else candidates[1]
+            )
```
hemslo, line 164: Nit: these checks have duplicate logic with init step, to make it simpler and less nested, we can extract a dataclass

```python
@dataclass
class MoveCandidates:
    creations: list[BundleFileChange] = field(default_factory=list)
    deletions: list[BundleFileChange] = field(default_factory=list)

    def changes(self) -> list[BundleFileChange]:
        if len(self.creations) == 1 and len(self.deletions) == 1:
            creation = self.creations[0]
            deletion = self.deletions[0]
            if (
                creation != deletion
                and deletion.fileref.file_type == creation.fileref.file_type
                and deletion.fileref.path != creation.fileref.path
            ):
                move_change = create_bundle_file_change(
                    deletion.fileref.path,
                    deletion.fileref.schema,
                    deletion.fileref.file_type,
                    deletion.old,
                    creation.new,
                )
                if move_change:
                    return [move_change]

        return self.creations + self.deletions
```

Then use it

```python
    move_candidates: dict[str, MoveCandidates] = defaultdict(MoveCandidates)
    new_bundle_changes = []
    for c in bundle_changes:
        if c.is_file_creation():
            move_candidates[c.new_content_sha()].creations.append(c)
        elif c.is_file_deletion():
            move_candidates[c.old_content_sha()].deletions.append(c)
        else:
            new_bundle_changes.append(c)
    move_candidate_changes = itertools.chain.from_iterable(
        c.changes() for c in move_candidates.values()
    )
    new_bundle_changes.extend(move_candidate_changes)
    return new_bundle_changes
```

reconcile/test/change_owners/test_change_type_bundle_change_post_processing.py
```diff
@@ -0,0 +1,89 @@
+import pytest
+
+from reconcile.change_owners.change_owners import _aggregate_file_moves
```
hemslo, line 3: test private method `_aggregate_file_moves` seems not right, maybe worth move it to a seperate module then make it public as `aggregate_file_moves`

reconcile/change_owners/change_owners.py
```diff
@@ -120,7 +121,71 @@ def fetch_bundle_changes(comparison_sha: str) -> list[BundleFileChange]:
     explicitely passed comparision bundle - usually the state of the master branch).
     """
     changes = gql.get_diff(comparison_sha)
-    return _parse_bundle_changes(changes)
+    bundle_changes = _parse_bundle_changes(changes)
+    try:
+        return _aggregate_file_moves(bundle_changes)
+    except Exception as e:
```
hemslo, line 127: What exception are we expecting here?

reconcile/test/change_owners/test_change_type_bundle_change_post_processing.py
geoberle: you are absolutely right. i will move some parts around `BundleFileChange`, their initialization and aggregation into a dedicated module.

reconcile/change_owners/change_owners.py
geoberle: this tries to recover from a situation where the metadata retrieved from qontract-server would not be valid, e.g. sha checksum missing (`InvalidBundleFileMetadataError`)

i reconsidered though and do not catch the exception anymore. if qontract-server would really not return valid metadata (or any other error situation) it is best to fail fast.

reconcile/change_owners/change_owners.py
hemslo: cool, then this check is no longer needed https://github.com/app-sre/qontract-reconcile/pull/3489/files#diff-901662fc5b9f7c76e9baf26b19ee2998ebad095634cd8bd1c6013285a0f366b8R282

hemslo (approved):
LGTM! Thanks!

reconcile/change_owners/change_owners.py
geoberle: removed it. thank you

PR #3490, opened 2023-05-04
remove e2e-tests

maorfr:
part of https://issues.redhat.com/browse/APPSRE-7589


OliviaHY (approved):

reconcile/gql_definitions/common/clusters.gql
```diff
@@ -13,7 +13,7 @@ query Clusters($name: String) {
     managedClusterRoles
     insecureSkipTLSVerify
     jumpHost {
-      ... CommonJumphostFields
+      ...CommonJumphostFields
```
rporres, line 16: Please avoid linting changes 🙏

rporres (commented):
Can you take a look to the amount of linting changes that doesn't belong to this PR? They are certainly weird, as if it was a GH artifact

reconcile/gql_definitions/common/clusters.gql
maorfr: that wasn't intentional :sweat_smile:

maorfr: it was "prettier": prettier.io

PR #3494, opened 2023-05-05
Add Control Plane upgrade

janboll:
Adding code to allow upgrades on Control Planes. 
Also includes some refactorings to GateAgreements, moving it to PolicyHandler, since it relates to multiple types of upgradable components.

https://issues.redhat.com/browse/APPSRE-7243


reconcile/aus/base.py
```diff
@@ -214,6 +232,8 @@ def act(self, dry_run: bool, ocm: OCM) -> None:
         elif self.action == "delete":
             self.policy.delete(ocm)
         elif self.action == "create":
+            if self.gates_to_agree:
```
patjlm, line 235: nit: this `if` does not seem needed

reconcile/aus/base.py
```diff
@@ -684,6 +708,32 @@ def calculate_diff(
                         ),
                     )
                 )
+            elif (
+                cluster.spec.hypershift is not None and cluster.spec.hypershift is True
```
patjlm, line 711: This could just be 
```suggestion
                cluster.spec.hypershift
```
But I think we may have a `None` `spec` for imported clusters, so maybe a check like `cluster.spec is not None` is needed ?

reconcile/aus/base.py
```diff
@@ -665,6 +685,10 @@ def calculate_diff(
         if verify_lock_should_skip(p, locked, ocm, p.cluster):
             continue
 
+        cluster = ocm.clusters.get(p.cluster)
+        if not cluster:
+            continue
```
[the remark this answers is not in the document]
janboll: That's to satisfy mypy checks. The get() returns `Optional[OCMSpec]`

reconcile/aus/base.py
janboll: I'll revisit this, your proposal was breaking tests. Perhaps I can find another way to fix them

janboll:
Creating a policy:

`[2023-05-09 09:37:35] [DEBUG] [connectionpool.py:_make_request:433] - [https://api.openshift.com:443](https://api.openshift.com/) "POST /api/clusters_mgmt/v1/clusters/<clusterid>/control_plane/upgrade_policies HTTP/1.1" 201 353`

Running integration again:
`[2023-05-09 09:38:38] [DEBUG] [base.py:verify_current_should_skip:575] - [ocm-production/hst01ue1] skipping cluster with existing upgrade policy`

patjlm:
Do we need an update in openshift_upgrade_watcher, and in any other integration or qontract-cli that reads the current version from OCM ?

patjlm:
Do we need an update in openshift_upgrade_watcher, and in any other integration or qontract-cli that reads the current version from OCM ?

janboll:
Only if they not use base.fetch_current_state. Let me check

janboll:
Following from usage of `ocm.get_upgrade_policies`, there are no other use cases outside these code changes.

janboll:
tested successfully with a test cluster in ocm production. Requires https://github.com/app-sre/qontract-reconcile/pull/3498 to be merged

patjlm (approved):

PR #3495, opened 2023-05-08
[ocm-addons-upgrade-scheduler-org] Fix delete addon upgrade policy

hemslo:
`AddonUpgradePolicy` should use `delete_addon_upgrade_policy` instead of `delete_upgrade_policy`.


geoberle (approved):
LGTM

PR #3501, opened 2023-05-12
[change-owners] refactor diff loading from qontract server and metadata handling

geoberle:
the main point of this PR is the handling of metadata changes, e.g. if a file gets moved or if just a SHA changes without any content changes (e.g. YAML comments not being present in the qs bundle). previously metadata and data were mixed in `BundleFileChange` objects. now metadata fields are wiped from the actual file contents and are stored seperately. this simplified a lot of code revolving around self-service for metadata changes.

additionally, the processing of the qontract-server `/diff` endpoint has been refactored to use pydantic dataclasses for validation.

because of this change, metadata changes like file moves are now self-serviceable.

part of https://issues.redhat.com/browse/APPSRE-7590


reconcile/test/change_owners/test_change_type_diff.py
```diff
@@ -568,16 +560,20 @@ def test_bundle_change_diff_item_reorder():
         },
     )
 
-    assert not bundle_change
+    # there is a bundle change ...
+    assert bundle_change
+    # ... but it has no diffs ...
+    assert not bundle_change.diffs
+    # ... but it has different SHAs
+    assert bundle_change.old_content_sha != bundle_change.new_content_sha
```
patjlm, line 568: so we consider list order is never important, right ?

reconcile/change_owners/changes.py
```diff
@@ -106,6 +95,13 @@ def cover_changes(self, change_type_context: ChangeTypeContext) -> list[Diff]:
         The change-type contexts that cover a change, are registered in the
         `DiffCoverage.coverage` list right next to the Diff that is covered.
         """
+
+        # if this is a metadata only change, register the change type context
+        # as a source of approvers
+        if self.metadata_only_change and not self.diffs:
+            self._metadata_only_diff_coverage().coverage.append(change_type_context)
+            return []
```
patjlm, line 103: this `return` means there is no coverage for this change, right ?

reconcile/change_owners/changes.py
```diff
@@ -236,8 +234,31 @@ def create_bundle_file_change(
...
+            new_content_sha=new_content_sha,
             diffs=diffs,
         )
+
+    if old_content_sha != new_content_sha or old_path != new_path:
+        # there were no diffs but the content of the file has changed, e.g.
+        # - the SHA of the file: the SHA can change even when no diffs are detected.
+        #   this can be happen in case something undetectable has changed, e.g. comments
+        #   in a YAML file are invisible to YAML parsers
+        # - the path of a file has changed a.k.a. the file was moved. in such a case
```
patjlm, line 247: is list reordering also part of these 'invisible' change ? If so, it may be worth adding in this comment.

reconcile/test/change_owners/test_change_type_diff.py
geoberle: this was an initial agreement and in most cases it is not important.
but it might be important in some cases and there was an idea lingering around to have schema metadata available to state if the order is important or not.

reconcile/change_owners/changes.py
geoberle: the return value of this function is only used in tests to assert about what happened. i never liked this hack.
let me fix this.

reconcile/change_owners/changes.py
geoberle: i added a comment

reconcile/change_owners/changes.py
```diff
@@ -244,6 +244,8 @@ def _create_bundle_file_change(
         # - the SHA of the file: the SHA can change even when no diffs are detected.
         #   this can be happen in case something undetectable has changed, e.g. comments
         #   in a YAML file are invisible to YAML parsers
+        # - item reodering in lists: we ignore reordering of elements in lists right now
```
patjlm, line 247: nit: typo on the first `reorrrrrrdering`

reconcile/change_owners/changes.py
geoberle: and it is gone

reconcile/change_owners/changes.py
geoberle: it is time for a vscode spell-checking plugin

PR #3504, opened 2023-05-12
[AUS] more details in action logging

geoberle:
instead of

```
[ocm-addons-upgrade-scheduler-org] ['create', 'my-org', 'my-cluster', '1.2.3']
```

action logging will look like this

```
[ocm-addons-upgrade-scheduler-org] create addon upgrade policy - {'cluster': 'my-cluster', 'ocm-org': 'my-org', 'addon-id': 'my-addon', 'version': '1.2.3', 'next-run': 'soon'}
```

this change also adds back the addon-id to the action log

### DIFF
--- a/reconcile/aus/base.py
+++ b/reconcile/aus/base.py
@@ -170,7 +170,7 @@ class AddonUpgradePolicy(AbstractUpgradePolicy):
     def summarize(self, ocm_org_name: str) -> str:
         details = {
             "cluster": self.cluster,
-            "ocm-org": ocm_org_name,
+            "ocm_org": ocm_org_name,
             "version": self.version,
             "next_run": self.next_run,
             "addon_id": self.addon_id,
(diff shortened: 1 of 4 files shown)


reconcile/aus/base.py
```diff
@@ -165,6 +167,16 @@ def delete(self, ocm: OCM) -> None:
         }
         ocm.delete_addon_upgrade_policy(self.cluster, item)
 
+    def summarize(self, ocm_org_name: str) -> str:
+        details = {
+            "cluster": self.cluster,
+            "ocm-org": ocm_org_name,
```
patjlm, line 173: maybe use underscore as for the other entries ?

patjlm: same in the other `summarize`

patjlm (approved):

PR #3513, opened 2023-05-18
Make alert-to-receiver command use alertname

rporres:
Prometheus adds automatically alertname label. Adding it to the considered labels in alert-to-receiver command so that we can easily test routing based on alertname.

Structures in the command have been modified to avoid duplication.


andreugallofre (approved):
lgtm

PR #2000, opened 2021-11-16
[terraform-vpc-peerings] add support to provide role to assume for account-vpc

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4092

depends on https://github.com/app-sre/qontract-schemas/pull/13

to be able to create vpc peerings without OCM using the account-vpc provider.

PR #3515, opened 2023-05-21
[terraform-vpc-peerings] add support for cluster2cluster vpc peerings with assume role

maorfr:
part of https://issues.redhat.com/browse/APPSRE-7652

this PR adds support to provide `assumeRole` to a cluster to cluster vpc peering, to be used instead of OCM infrastructure access.

similar to #2000


patjlm (approved):
Looks good

geoberle (approved):

PR #3522, opened 2023-05-22
[terraform-aws-route53] implement early-exit using state

maorfr:
related to https://issues.redhat.com/browse/APPSRE-7655

this will reduce the number of calls to Route53 API significantly, as we will only run the integration when there is an actual change that needs to happen due to graphql changes.

in other words - manual actions will now be ignored by the integration, until the next iteration that has changes will act.

### DIFF
--- a/reconcile/terraform_aws_route53.py
+++ b/reconcile/terraform_aws_route53.py
@@ -19,6 +19,7 @@ from reconcile.utils.external_resources import (
     get_external_resource_specs,
 )
 from reconcile.utils.semver_helper import make_semver
+from reconcile.utils.state import State, init_state
 from reconcile.utils.terraform_client import TerraformClient as Terraform
 from reconcile.utils.terrascript_aws_client import TerrascriptClient as Terrascript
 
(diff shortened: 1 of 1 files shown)


patjlm (approved):

rporres:
The title of this MR is quite misleading. This is not using the early-exit mechanism (being early-exit mechanism the one defined by calling `early_exit_desired_state`), this is creating a third state to be considered by the integration (being app-interface one and terraform state the other two). That may be potentially complicated... I wonder if there's a solution that doesn't imply keeping data in three different places...

patjlm:
From a discussion in https://issues.redhat.com/browse/APPSRE-7655, would it make sense to still force run eg once per hour? I am thinking this could be achieved by storing the current hour in each zone state and compare it with the current hour in `should_run`

rporres:
If the state is just used to force the hourly run, I don't see a problem.

patjlm:
> If the state is just used to force the hourly run, I don't see a problem.

It would be used for both
* run only if needed according to the state: "compare desired vs stored state"
* run at least once an hour (keep track of the last time/hour it ran)

PR #3523, opened 2023-05-22
[AUS] fix cluster discovery

geoberle:
discover clusters for AUS based only on subscription labels. the organization labels don't hold any cluster configuration, only org-wide configuration for blocked versions and sectors.


kwilczynski (approved):
@geoberle, looks good!

Thank you for a quick fix! :rocket:

PR #3527, opened 2023-05-23
[aus] fix sector processing from labels

geoberle:
removed magic code that processed data from org labels and instead pass it in manually like in good old times so that mypy can tell me that i made a mistake


kwilczynski (approved):
@geoberle, looks good!

Thank you for the prompt fix!

For the good ol' times :beers:

PR #3528, opened 2023-05-23
Add new integration to delete old unused AMIs

rporres:
part of [APPSRE-7272](https://issues.redhat.com/browse/APPSRE-7272)

design-doc in: [docs/app-sre/design-docs/aws-ami-cleanup.md](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/aws-ami-cleanup.md)

schema change: https://github.com/app-sre/qontract-schemas/pull/446


rporres:
this depends on https://github.com/app-sre/qontract-reconcile/pull/3529

setup.py
```diff
@@ -30,9 +30,8 @@
         "python-gitlab>=1.11.0,<1.12.0",
         "semver~=2.13",
         "python-terraform>=0.10.0,<0.11.0",
-        # memory leak after boto3 1.24.72, https://github.com/boto/boto3/issues/3614
-        "boto3==1.24.71",
-        "botocore==1.27.71",
+        "boto3==1.26.134",
+        "botocore==1.29.134",
```
rporres, line 34: This will be removed once https://github.com/app-sre/qontract-reconcile/pull/3529 is merged.

reconcile/queries.py
```diff
@@ -547,6 +557,7 @@ def get_aws_accounts(
     sharing=False,
     terraform_state=False,
     ecrs=True,
+    cleanup=False,
```
hemslo, line 560: Now there is a new way to use typed query to get aws account instead of adding more params to this function, an example is like this https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/gql_definitions/terraform_tgw_attachments/aws_accounts.gql.

You can create a new one for new integration, and reuse some fragments.

reconcile/aws_ami_cleanup/integration.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+def check_aws_ami_in_use(
+    aws_ami: AWSAmi, app_interface_amis: list[AIAmi]
+) -> Optional[str]:
+    """Verifies if the given AWS ami is in use in a defined app-interface ASG."""
+    for ai_ami in app_interface_amis:
+        # This can happen if the ASG init template has changed over the time. We don't have a way
+        # to properly delete these automatically since we cannot assure they are not in use.
+        if len(ai_ami.tags) > len(aws_ami.tags):
```
hemslo, line 174: Is `ai_ami.tags` defined in app interface? What if we add new tags but old AMI were built without them?

reconcile/aws_ami_cleanup/__init__.py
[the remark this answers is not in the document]
rporres: I've learnt from the best ;-)

reconcile/queries.py
chassing: Good point Rafa! In this case, keep using the old dict-based queries.

chassing (approved):
Awesome work! Ship it 🚀

hemslo (approved):
Thanks for the cleanup!

reconcile/queries.py
rporres: added a comment on 47b58c56 around this.

PR #3531, opened 2023-05-25
ocm machine pools add hypershift support

janboll:
Adding Node Pool support to ocm_machine_pools

- Refactoring query, using data class-based Query
- Introduce data classes for handling existing machine/node pools 
- Refactoring calculate_diff to use introduced data classes
- Adding tests 

https://issues.redhat.com/browse/APPSRE-7661

Related schema change: https://github.com/app-sre/qontract-schemas/pull/475

### DIFF
--- a/reconcile/ocm_machine_pools.py
+++ b/reconcile/ocm_machine_pools.py
@@ -277,7 +277,7 @@ def calculate_diff(
     diffs: list[PoolHandler] = []
     errors: list[InvalidUpdateError] = []
 
-    all_desired_pools: list[tuple[str, str]] = []
+    all_desired_pools: set[tuple[str, str]] = set()
     for desired in desired_state.cluster_pools:
 
         for desired_machine_pool in desired.pools:
@@ -286,7 +286,7 @@ def calculate_diff(
                 for p in current_state.get(desired.cluster_name, [])
                 if p.id == desired_machine_pool.q_id
             ]
-            all_desired_pools.append((desired_machine_pool.q_id, desired.cluster_name))
+            all_desired_pools.add((desired_machine_pool.q_id, desired.cluster_name))
             if not current_machine_pool:
                 if desired.hypershift:
                     diffs.append(
--- a/reconcile/ocm_machine_pools.py
+++ b/reconcile/ocm_machine_pools.py
@@ -213,41 +213,36 @@ def fetch_current_state(
     ocm_map: OCMMap,
     clusters: Iterable[ClusterV1],
 ) -> Mapping[str, list[AbstractPool]]:
-    current_state: MutableMapping[str, list[AbstractPool]] = {}
-    for cluster in clusters:
-        current_state[cluster.name] = []
-        ocm = ocm_map.get(cluster.name)
-        if cluster.spec and cluster.spec.hypershift:
-            machine_pools = ocm.get_node_pools(cluster.name)
-            for machine_pool in machine_pools:
-                current_state[cluster.name].append(
-                    NodePool(
-                        id=machine_pool["id"],
-                        replicas=machine_pool["replicas"],
-                        aws_node_pool=AWSNodePool(
-                            instance_type=machine_pool["aws_node_pool"]["instance_type"]
-                        ),
-                        taints=machine_pool.get("taints"),
-                        labels=machine_pool.get("labels"),
-                        subnet=machine_pool.get("subnet"),
-                        cluster=cluster.name,
-                    )
-                )
-        else:
-            machine_pools = ocm.get_machine_pools(cluster.name)
-            for machine_pool in machine_pools:
-                current_state[cluster.name].append(
-                    MachinePool(
-                        id=machine_pool["id"],
-                        replicas=machine_pool["replicas"],
-                        instance_type=machine_pool["instance_type"],
-                        taints=machine_pool.get("taints"),
-                        labels=machine_pool.get("labels"),
-                        cluster=cluster.name,
-                    )
-                )
+    return {c.name: fetch_current_state_for_cluster(c, ocm_map.get(c.name)) for c in clusters}
 
-    return current_state
+
+def fetch_current_state_for_cluster(cluster, ocm):
+    if cluster.spec and cluster.spec.hypershift:
+        return [
+            NodePool(
+                id=machine_pool["id"],
+                replicas=machine_pool["replicas"],
+                aws_node_pool=AWSNodePool(
+                    instance_type=machine_pool["aws_node_pool"]["instance_type"]
+                ),
+                taints=machine_pool.get("taints"),
+                labels=machine_pool.get("labels"),
+                subnet=machine_pool.get("subnet"),
+                cluster=cluster.name,
+            )
+            for machine_pool in ocm.get_node_pools(cluster.name)
+        ]
+    return [
+        MachinePool(
+            id=machine_pool["id"],
+            replicas=machine_pool["replicas"],
+            instance_type=machine_pool["instance_type"],
+            taints=machine_pool.get("taints"),
+            labels=machine_pool.get("labels"),
+            cluster=cluster.name,
+        )
+        for machine_pool in ocm.get_machine_pools(cluster.name)
+    ]
 
 
 def create_desired_state_from_gql(
@@ -381,7 +376,9 @@ def run(dry_run: bool):
         sys.exit(0)
 
     settings = queries.get_app_interface_settings()
-    cluster_like_objects = [cluster.dict(by_alias=True) for cluster in filtered_clusters]
+    cluster_like_objects = [
+        cluster.dict(by_alias=True) for cluster in filtered_clusters
+    ]
     ocm_map = OCMMap(
         clusters=cluster_like_objects,
         integration=QONTRACT_INTEGRATION,
--- a/reconcile/ocm_machine_pools.py
+++ b/reconcile/ocm_machine_pools.py
@@ -2,29 +2,60 @@ import json
 import logging
 import sys
 from collections.abc import Mapping
-from typing import Any
+from typing import (
+    Any,
+    Iterable,
+)
+
+from pydantic import BaseModel
 
 from reconcile import queries
+from reconcile.gql_definitions.common.clusters import ClusterV1
+from reconcile.gql_definitions.common.clusters import query as clusters_query
+from reconcile.utils import gql
 from reconcile.utils.disabled_integrations import integration_is_enabled
 from reconcile.utils.ocm import OCMMap
 
 QONTRACT_INTEGRATION = "ocm-machine-pools"
 
 
-def fetch_current_state(clusters):
+class CurrentMachinePool(BaseModel):
+    # Abstract class for machine pools, abstracts OSD/Hypershift
+    id: str
+    replicas: int
+    instance_type: str
+    taints: list[Mapping[str, str]]
+    labels: Mapping[str, str]
+    cluster: str
+
+
+def fetch_current_state(
+    clusters: Iterable[ClusterV1],
+) -> tuple[OCMMap, list[CurrentMachinePool]]:
     settings = queries.get_app_interface_settings()
+    cluster_like_objects = [cluster.dict(by_alias=True) for cluster in clusters]
     ocm_map = OCMMap(
-        clusters=clusters, integration=QONTRACT_INTEGRATION, settings=settings
+        clusters=cluster_like_objects,
+        integration=QONTRACT_INTEGRATION,
+        settings=settings,
     )
 
     current_state = []
     for cluster in clusters:
-        cluster_name = cluster["name"]
-        ocm = ocm_map.get(cluster_name)
-        machine_pools = ocm.get_machine_pools(cluster_name)
+        ocm = ocm_map.get(cluster.name)
+        machine_pools = ocm.get_machine_pools(cluster.name)
         for machine_pool in machine_pools:
-            machine_pool["cluster"] = cluster_name
-            current_state.append(machine_pool)
+            machine_pool["cluster"] = cluster.name
+            current_state.append(
+                CurrentMachinePool(
+                    id=machine_pool["id"],
+                    replicas=machine_pool["replicas"],
+                    instance_type=machine_pool["instance_type"],
+                    taints=machine_pool["taints"],
+                    labels=machine_pool["labels"],
+                    cluster=machine_pool["cluster"],
+                )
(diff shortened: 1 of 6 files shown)


reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,332 @@
...
+        ocm.create_machine_pool(self.cluster, self.dict(by_alias=True))
+
+    def update(self, ocm: OCM) -> None:
+        ocm.update_machine_pool(self.cluster, self.dict(by_alias=True))
+
+    def has_diff(self, pool: ClusterMachinePoolV1) -> bool:
+        return (
+            self.replicas != pool.replicas
+            or self.taints != pool.taints
+            or self.labels != pool.labels
```
patjlm, line 81: should other field be taken into account, like machine type? or is that forbidden ?

patjlm: ah, i should have read just below i suppose :)

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,332 @@
...
-            machine_pool["cluster"] = cluster_name
-            current_state.append(machine_pool)
+        current_state[cluster.name] = {
+            "machine_pools": [],
+        }
+        ocm = ocm_map.get(cluster.name)
+        if cluster.spec.hypershift:
+            machine_pools = ocm.get_node_pools(cluster.name)
+            for machine_pool in machine_pools:
+                current_state[cluster.name]["machine_pools"].append(
```
patjlm, line 175: so we store `node_pools` in our `machine_pools` list for our current state?

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,332 @@
...
-            diffs.append(c)
+    all_desired_pools: list[str] = []
+    for cluster_name, desired_data in desired_state.items():
+        for desired_machine_pool in desired_data["machine_pools"]:
+            current_machine_pool = [
+                p
+                for p in current_state.get(cluster_name, {}).get("machine_pools", [])
+                if p.id == desired_machine_pool.q_id
+            ]
+            all_desired_pools.append(f"{desired_machine_pool.q_id}-{cluster_name}")
```
patjlm, line 234: maybe nicer to use a tuple instead of a string here

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,332 @@
...
+            ocm = ocm_map.get(diff.pool.cluster.name)
+            diff.act(dry_run, ocm)
+
 
+def _cluster_is_compatible(cluster: ClusterV1) -> bool:
+    return (
+        cluster.ocm is not None
+        and cluster.machine_pools is not None
+        # and not cluster.spec.hypershift
+    )
```
patjlm, line 300: don't you need to check if `spec` is set?

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,332 @@
...
+                        PoolHandler(
+                            action="create",
+                            pool=MachinePool.create_from_gql(
+                                pool=desired_machine_pool,
+                                cluster=cluster_name,
+                            ),
+                        )
+                    )
+                    continue
+            if current_machine_pool[0].has_diff(desired_machine_pool):
```
patjlm, line 257: `elif`

patjlm: or `continue` in the `if ` above

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,332 @@
...
+    @abstractmethod
+    def update(self, ocm: OCM) -> None:
+        pass
+
+    @abstractmethod
+    def has_diff(self, pool: ClusterMachinePoolV1) -> bool:
+        pass
+
+    @abstractmethod
+    def invalid_diff(self, pool: ClusterMachinePoolV1) -> Optional[str]:
```
geoberle, line 61: can you add doc strings in general? especially the purpose of this function is not immediately apparent without reading all of the code

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,332 @@
...
+    current_state = {}
     for cluster in clusters:
-        cluster_name = cluster["name"]
-        ocm = ocm_map.get(cluster_name)
-        machine_pools = ocm.get_machine_pools(cluster_name)
-        for machine_pool in machine_pools:
-            machine_pool["cluster"] = cluster_name
-            current_state.append(machine_pool)
+        current_state[cluster.name] = {
+            "machine_pools": [],
```
geoberle, line 169: the naming indicates that this list will only hold `MachinePool`objects

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,332 @@
...
-            err = True
-            continue
-        d.pop("instance_type")
-        for key in ["labels", "taints"]:
-            if c.get(key, None) != d.get(key, None):
-                # https://github.com/openshift/machine-api-operator/blob/master/FAQ.md
-                logging.warning(
-                    f"update {key} for machine pool {d['id']} "
-                    f"will only be applied to new Nodes"
-                )
```
geoberle, line 84: is this warning worth bringing over to the new code?

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,332 @@
...
 
-    return ocm_map, current_state
+    return current_state
 
 
-def fetch_desired_state(clusters):
-    desired_state = []
+def fetch_desired_state(
+    clusters: Iterable[ClusterV1],
+):
```
geoberle, line 207: nit: return type hint

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,332 @@
...
+                    )
+                )
 
-    return ocm_map, current_state
+    return current_state
 
 
-def fetch_desired_state(clusters):
-    desired_state = []
+def fetch_desired_state(
```
geoberle, line 205: good opportunity to rename this function since it is not fetching anything, just transforming the given input

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,332 @@
...
-            if c.get(key, None) != d.get(key, None):
-                # https://github.com/openshift/machine-api-operator/blob/master/FAQ.md
-                logging.warning(
-                    f"update {key} for machine pool {d['id']} "
-                    f"will only be applied to new Nodes"
-                )
-        d["action"] = "update"
-        diffs.append(d)
+def calculate_diff(
+    current_state: Mapping[str, Mapping[str, list[AbstractPool]]],
```
geoberle, line 220: can you add a docstring what the keys and nested keys of `current_state` are?

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,332 @@
...
+    desired_state,
+) -> Tuple[Iterable[PoolHandler], Iterable[InvalidUpdateError]]:
```
geoberle, line 222: nit: `tuple` lowercase `t` - no import needed

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

geoberle (commented):
i like the approach with the handler so far. same as with cluster upgrades. consistent patterns FTW!

reconcile/ocm_machine_pools.py
janboll: Good point, but that line was for testing only.

reconcile/ocm_machine_pools.py
janboll: Actually not required, removing the submapping

reconcile/ocm_machine_pools.py
janboll: Good point, renamed to `create_desired_state_from_gql`

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,332 @@
...
+    desired_state = {}
     for cluster in clusters:
-        cluster_name = cluster["name"]
-        machine_pools = cluster["machinePools"]
-        for machine_pool in machine_pools:
-            machine_pool["cluster"] = cluster_name
-            labels = machine_pool.pop("labels")
-            if labels:
-                machine_pool["labels"] = json.loads(labels)
-            taints = machine_pool.pop("taints")
```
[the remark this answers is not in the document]
janboll: I'll check this, I'll consider adding this to MachinePool.update/create methods

reconcile/ocm_machine_pools.py
janboll: Removed the nested keys,

reconcile/ocm_machine_pools.py
janboll: added

reconcile/ocm_machine_pools.py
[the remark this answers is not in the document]
janboll: added

janboll:
Manually tested:

- Create Machine Pools/Node Pools 
- Update these pools: scale, labels, taint
- Test, errors occur up on updating subnet/instance type
- Delete these pools

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,400 @@
...
-            if taints:
-                machine_pool["taints"] = taints
-            desired_state.append(machine_pool)
+        if cluster.machine_pools:
+            is_hypershift = False
+            if cluster.spec:
+                is_hypershift = True if cluster.spec.hypershift else False
```
hemslo, line 261: Can be simplified by

```python
is_hypershift = cluster.spec is not None and bool(cluster.spec.hypershift)
```

Not sure if `cluster.spec` can be `None`, if it's not nullable in schema, just `bool(cluster.spec.hypershift` will be enough.

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,400 @@
...
+                        replicas=machine_pool["replicas"],
+                        instance_type=machine_pool["instance_type"],
+                        taints=machine_pool.get("taints"),
+                        labels=machine_pool.get("labels"),
+                        cluster=cluster.name,
+                    )
+                )
 
-    return ocm_map, current_state
+    return current_state
```
hemslo, line 250: This is quite long and nested, can be extracted like this to avoid mutation:

```python
def fetch_current_state(
    ocm_map: OCMMap,
    clusters: Iterable[ClusterV1],
) -> Mapping[str, list[AbstractPool]]:
    return {c.name: fetch_current_state_for_cluster(c, ocm_map.get(c.name)) for c in clusters}


def fetch_current_state_for_cluster(cluster, ocm):
    if cluster.spec and cluster.spec.hypershift:
        return [
            NodePool(
                id=machine_pool["id"],
                replicas=machine_pool["replicas"],
                aws_node_pool=AWSNodePool(
                    instance_type=machine_pool["aws_node_pool"]["instance_type"]
                ),
                taints=machine_pool.get("taints"),
                labels=machine_pool.get("labels"),
                subnet=machine_pool.get("subnet"),
                cluster=cluster.name,
            )
            for machine_pool in ocm.get_node_pools(cluster.name)
        ]
    return [
        MachinePool(
            id=machine_pool["id"],
            replicas=machine_pool["replicas"],
            instance_type=machine_pool["instance_type"],
            taints=machine_pool.get("taints"),
            labels=machine_pool.get("labels"),
            cluster=cluster.name,
        )
        for machine_pool in ocm.get_machine_pools(cluster.name)
    ]
```

reconcile/ocm_machine_pools.py
```diff
@@ -1,137 +1,400 @@
...
-        d["action"] = "update"
-        diffs.append(d)
+def calculate_diff(
+    current_state: Mapping[str, list[AbstractPool]],
+    desired_state: DesiredStateList,
+) -> tuple[list[PoolHandler], list[InvalidUpdateError]]:
+    diffs: list[PoolHandler] = []
+    errors: list[InvalidUpdateError] = []
+
+    all_desired_pools: list[tuple[str, str]] = []
```
hemslo, line 280: Since `all_desired_pools` is only used in contains check, can change to `set[tuple[str, str]]` to improve performance for large dataset.

If you are interested, can also use new generic [differ](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/differ.py) to simplify diff logic.

geoberle (approved):

reconcile/ocm_machine_pools.py
janboll: Hm, I actually prefer the more verbose version :slightly_smiling_face:

reconcile/ocm_machine_pools.py
janboll: Great refactoring, added!

reconcile/ocm_machine_pools.py
janboll: Changed to a set, good spot!

PR #3485, opened 2023-05-03
[change-owners] avoid orphaned self-service roles

geoberle:
self-service roles (roles with a self-service section) must be assigned to at least one approver (user or bot). this avoids the situation that an MR is marked as self-serviceable but no approvers are available.

https://issues.redhat.com/browse/APPSRE-7535

PR #3534, opened 2023-05-26
[change-owners] fix empty self-service-role check

geoberle:
placed the validation of self-service roles into the loading mechanism so that roles from current and desired state are both validated accordingly.

follows up on https://github.com/app-sre/qontract-reconcile/pull/3485

part of https://issues.redhat.com/browse/APPSRE-7535


fishi0x01 (approved):
LGTM

PR #3536, opened 2023-05-26
tf-resources: allow enabling access_logs on aws_lb

jfchevrette:
This adds an opinionated way to enable access logging on an ALB. Instead of exposing the`access_logs` arguments, it is defined as a `boolean` and the bucket creation is opinionated (or hardcoded) 

Related schema change: https://github.com/app-sre/qontract-schemas/pull/452

This benefits the work in https://issues.redhat.com/browse/SDE-2715


bhushanthakur93 (approved):
LGTM!

PR #3543, opened 2023-06-01
[cluster user mgmt] OCM cluster user mgmt based standalone service

geoberle:
Implementation of the [Red Hat User Management SRE capability](https://issues.redhat.com/browse/SDE-2628)

This capability can be used by RH internal teams to attach their LDAP groups (e.g. Rover) to the `cluster-admins` or `dedicated-admins` roles of a cluster. Users can declare their desire for group syncs via OCM organization or subscription labels. Group syncs defined on the organization are applied on all clusters of the organization. Group syncs defined on subscriptions will only be applied to the individual clusters. Configuration on the organization and on the subscriptions are cummulative.

If a cluster role is either managed on the organisation or subscription, its current members will be replaced by the ones from the defined LDAP groups.

The following label keys are supported:

* `sre-capabilities.user-mgmt.rover.authz.dedicated-admins` contains a CSV list of rover group CNs (see LDAP Common Name on the rover group page) for the `dedicated-admins` role
* `sre-capabilities.user-mgmt.rover.authz.cluster-admins` contains a CSV list of rover group CNs for the `cluster-admins` role

The `cluster-admins` role can only be managed if the cluster has the proper capability in OCM.

This standalone capability does not fail on errors but reports them differently based on the kind of issue, e.g. configuration validation errors the user is responsible for (e.g. non-existing LDAP groups) are reported to the cluster owners as OCM service logs. Additionally several prometheus metrics are exposed that can be used for fine grained alerting:

```text
ocm_user_management_organization_validation_errors{ocm_env="ocm-production",org_id="xxx"} 0.0
ocm_user_management_organization_reconciled_total{ocm_env="ocm-production",org_id="xxx"} 1.0
ocm_user_management_organization_actions_total{action="add-user",ocm_env="ocm-production",org_id="xxx"} 24.0
ocm_user_management_organization_actions_total{action="remove-user",ocm_env="ocm-production",org_id="xxx"} 7.0
ocm_user_management_organization_reconcile_errors_total{ocm_env="ocm-production",org_id="xxx"} 0.0
```

part of <https://issues.redhat.com/browse/APPSRE-7709>

### DIFF
--- a/reconcile/oum/base.py
+++ b/reconcile/oum/base.py
@@ -1,0 +1,385 @@
+import logging
+from abc import (
+    ABC,
+    abstractmethod,
+)
+from collections import defaultdict
+from typing import Optional
+
+from reconcile.gql_definitions.common.ocm_environments import (
+    query as ocm_environment_query,
+)
+from reconcile.gql_definitions.fragments.ocm_environment import OCMEnvironment
+from reconcile.oum.metrics import (
+    OCMUserManagementOrganizationActionCounter as ReconcileActionCounter,
+)
+from reconcile.oum.metrics import (
+    OCMUserManagementOrganizationReconcileCounter as ReconcileCounter,
+)
+from reconcile.oum.metrics import (
+    OCMUserManagementOrganizationReconcileErrorCounter as ReconcileErrorCounter,
+)
+from reconcile.oum.metrics import (
+    OCMUserManagementOrganizationValidationErrorsGauge as ValidationErrorsGauge,
+)
+from reconcile.oum.models import (
+    ClusterError,
+    ClusterRoleReconcileResult,
+    ClusterUserManagementSpec,
+    OrganizationUserManagementConfiguration,
+)
+from reconcile.oum.providers import (
+    GroupMemberProvider,
+    init_ldap_group_member_provider,
+)
+from reconcile.utils import (
+    gql,
+    metrics,
+)
+from reconcile.utils.ocm.cluster_groups import (
+    OCMClusterGroupId,
+    add_user_to_cluster_group,
+    delete_user_from_cluster_group,
+    get_cluster_groups,
+)
+from reconcile.utils.ocm.clusters import CAPABILITY_MANAGE_CLUSTER_ADMIN
+from reconcile.utils.ocm_base_client import (
+    OCMBaseClient,
+    init_ocm_base_client,
+)
+from reconcile.utils.runtime.integration import (
+    PydanticRunParams,
+    QontractReconcileIntegration,
+)
+
+
+class OCMUserManagementIntegrationParams(PydanticRunParams):
+    ocm_environment: Optional[str] = None
+    ocm_organization_ids: Optional[set[str]] = None
+    group_provider_specs: list[str]
+
+
+class OCMUserManagementIntegration(
+    QontractReconcileIntegration[OCMUserManagementIntegrationParams], ABC
+):
+    def run(self, dry_run: bool) -> None:
+        with metrics.transactional_metrics(self.name):
+            # init group providers
+            self.group_member_providers = get_group_providers(
+                self.params.group_provider_specs
+            )
+
+            for ocm_env in self.get_ocm_environments():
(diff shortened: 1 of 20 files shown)


reconcile/oum/base.py
```diff
@@ -0,0 +1,373 @@
...
+            if group.users
+        }
+        # only process roles present in the desired state and ignore the ones that are not
+        # this way a role can still be managed manually while another one is managed by this integration
+        for group in desired_groups.keys():
+            desired_members = desired_groups.get(group, set())
+            current_members = current_groups.get(group, set())
+
+            # add missing users
+            for missing_user in desired_members - current_members:
```
hemslo, line 269: nit: Do you want to try the generic [differ](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/differ.py) here?

to do group level diff
```python
diff_result = diff_mappings(current_groups, desired_groups)
```

To do member level diff

```python
diff_result = diff_iterables(current_members, desired_members)
```

then process different parts

```python
diff_result.add
diff_result.delete
```

Since no need to delete current groups, it's not that useful, just save some mind burden of calculating diff.

reconcile/oum/standalone.py
```diff
@@ -0,0 +1,197 @@
...
+                try:
+                    role_groups = build_cluster_config_from_labels(
+                        provider=p,
+                        org_labels=provider_org_labels,
+                        subscription_labels=provider_subs_labels,
+                    )
+                    for role_id, group_refs in role_groups.items():
+                        cluster_config.roles[role_id].extend(group_refs)
+
+                except Exception as e:
```
hemslo, line 195: What exception are we expecting here?

reconcile/oum/base.py
geoberle: oh i forgott about the generif differ. wanted to try that one.
i will update the PR

reconcile/oum/standalone.py
geoberle: pydantic validation errors. the `build_cluster_config_from_labels` processes OCM labels, which are basically free form labels. a pydantic dataclass takes care of validating some formal constraints and fails if they are not met.

i can for sure be more specific here by using the concrete pydantic error class. on the other hand i don't want any potential error from within this block to leak out ever.

hemslo: Right, that pydantic validation, I thought there were some external calls, but didn't find any, good to catch them!

reconcile/oum/base.py
geoberle: after having a closer look i don't think it helps a lot here because diffing is really just a `-` operation beetween to sets of strings. i will keep the differ in mind though for other scenarios where diffs are more cumbersome to calculate.

ok with you?

hemslo: Yes, diff here is very simple, generic differ is more suitable where we need to find things to add, delete, change and not changed.

reconcile/utils/ldap_client.py
```diff
@@ -34,14 +36,43 @@ def get_users(self, uids: Iterable[str]) -> set[str]:
...
+        _, _, users, _ = self.connection.search(
+            self.base_dn,
+            filter,
+            attributes=["uid", "memberOf"],
+        )
+        groups_and_members: dict[str, set[str]] = defaultdict(set[str])
+        for u in users:
+            uid = u["attributes"]["uid"][0]
+            for group in set(u["attributes"]["memberOf"]).intersection(rover_group_dns):
+                groups_and_members[rover_group_dn_mapping[group]].add(uid)
```
hemslo, line 57: nit: use both `rover_group_dns` and `rover_group_dn_mapping` together plus nested access is a little hard understand. An easier to understand version can be:

```python
            member_groups = (
                rover_group_dn_mapping[g]
                for g in u["attributes"]["memberOf"]
                if g in rover_group_dn_mapping
            )
            for group in member_groups:
                groups_and_members[group].add(uid)
```

hemslo (approved):
LGTM! But may require another review as I'm not familiar with this feature.

PR #3546, opened 2023-06-02
[addons-upgrade-scheduler] don't fail when clusters get deprovisioned

geoberle:
fix a bug that fails the addons-upgrade-scheduler integration if a cluster mentioned in an a-i OCM org file is deprovisioned. ignoring such declarations is a valid solution because such clusters get cleaned up automatically after a while anyways.


kwilczynski (approved):
@geoberle, looks good!

Thank you did the fix!

PR #3547, opened 2023-06-02
[AUS] expose current cluster version in metrics

geoberle:
the current cluster version information can be used in reporting. the metrics now looks as follows:

```
aus_cluster_version_remaining_soak_days{cluster_uuid="xxx", soaking_version="4.13.1", current_version="4.13.0"}
```

i considered putting the `current_version` into another existing metric, but none would quite fit. introducing a new one just for this piece of information would have been an overkill.

i'm more than open for a discussion :)

### DIFF
--- a/reconcile/aus/metrics.py
+++ b/reconcile/aus/metrics.py
@@ -19,6 +19,7 @@ class AUSClusterVersionRemainingSoakDaysGauge(AUSBaseMetric, GaugeMetric):
 
     cluster_uuid: str
     soaking_version: str
+    current_version: str
 
     @classmethod
     def name(cls) -> str:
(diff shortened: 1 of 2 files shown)


kwilczynski (approved):
@geoberle, looks good!

Why no! Metrics for all the things!

patjlm:
What happens if there are no version to be upgraded to, if we run the latest and greatest version from the cluster channel, so no soak days metrics? We won't be able to retrieve the current version I suppose?
I am not sure this metric is the best here. I'd consider cluster info, or a new one.. but since you thought of those already, I'd be happy to discuss why you discarded these options.

geoberle:
Ah you are right. So maybe a new metric. But then we might think more of an explicit exporter. Or we find an existing internal metric that exposes this an more information already?

PR #3548, opened 2023-06-04
add support for assume action and Federated assume principal

maorfr:
part of https://issues.redhat.com/browse/APPSRE-7734

depends on https://github.com/app-sre/qontract-schemas/pull/460

this PR adds support to allow creating roles with:
- a different assume action (instead of AssumeRole)
- a Federated assume Principal

this is required to attach IAM roles to OpenShift ServiceAccounts.

this fits the backplane usage, but is a good pattern instead of using IAM credentials, so this possibly opens up a door for https://issues.redhat.com/browse/APPSRE-4201

### DIFF
--- a/reconcile/gql_definitions/terraform_resources/terraform_resources_namespaces.py
+++ b/reconcile/gql_definitions/terraform_resources/terraform_resources_namespaces.py
@@ -139,8 +139,10 @@ query TerraformResourcesNamespaces {
                 assume_role {
                     AWS
                     Service
+                    Federated
                 }
                 assume_condition
+                assume_action
                 inline_policy
                 output_resource_name
                 annotations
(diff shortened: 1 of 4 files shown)


reconcile/gql_definitions/terraform_resources/terraform_resources_namespaces.py
```diff
@@ -590,12 +592,14 @@ class NamespaceTerraformResourceSecretsManagerServiceAccountV1(
 class AssumeRoleV1(ConfiguredBaseModel):
     aws: Optional[list[str]] = Field(..., alias="AWS")
     service: Optional[list[str]] = Field(..., alias="Service")
+    federated: Optional[str] = Field(..., alias="Federated")
```
patjlm, line 595: I don't understand where this new field is being used.

maorfr: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/70140/diffs#1ab2116cf4200d3ee1d441a014b545e06e6abf46_36_47

PR #3554, opened 2023-06-05
Check typing in dockerfiles/hack directory

rporres:
and make run-integration.py pass.


rporres:
The ignored mypy error

```
dockerfiles/hack/run-integration.py:175: error: Argument 1 to "int" has incompatible type "Union[str, int, None]"; expected "Union[str, bytes, bytearray, memoryview, array[Any], mmap, _CData, PickleBuffer, SupportsInt, SupportsIndex, SupportsTrunc]"  [arg-type]
```

It is certainly beyond this patch to fix this :)

PR #3559, opened 2023-06-07
Fix long saas_name in console_url on slack notification

hemslo:
Currently long `saas_name` produces wrong `console_url` in slack notification.

For example, expect `Runs?name=saas-openshift-cert-manager-routes-app-sre-product` but got `Runs?name=saas-openshift-cert-manager-routes-App-SRE-production`.

This change extracts same build `pipeline_run` name logic, reuse in both trigger step and notification step.

### <samp>🤖 Generated by Copilot at ad9e048</samp>

This pull request refactors and improves the naming and console url generation logic for tekton pipelineruns triggered by saas file deployments. It also adds a new test module to verify the expected console url format for different saas file scenarios. The changes affect the `reconcile.openshift_saas_deploy_trigger_base`, `reconcile.openshift_saas_deploy`, `reconcile.utils.saasherder.saasherder`, and `reconcile.test.test_openshift_saas_deploy` modules.

### <samp>🤖 Generated by Copilot at ad9e048</samp>

*  Refactor saas file data model to use pydantic instead of dataclasses
  * Replace import of `SaasFile` class from `reconcile.utils.saasherder.interfaces` with `reconcile.typed_queries.saas_files` in `reconcile/openshift_saas_deploy.py` ([link](https://github.com/app-sre/qontract-reconcile/pull/3559/files?diff=unified&w=0#diff-304fd1216ea8201e3242ba749fcc1bebbe71e61fd70851f7ed266e7318ed40b2R20), [link](https://github.com/app-sre/qontract-reconcile/pull/3559/files?diff=unified&w=0#diff-304fd1216ea8201e3242ba749fcc1bebbe71e61fd70851f7ed266e7318ed40b2L27))
  * Add unit tests for `compose_console_url` function using mock saas file objects from `reconcile.typed_queries.saas_files` in `reconcile/test/test_openshift_saas_deploy.py` ([link](https://github.com/app-sre/qontract-reconcile/pull/3559/files?diff=unified&w=0#diff-0f6dbc40d4600c4527755db5370c8c15906daa9c1269fe71e0de1add0243849eR1-R87))
  * Add helper method `build_saas_file_env_combo` to `SaasHerder` class in `reconcile/utils/saasherder/saasherder.py` that returns a tuple of short and long names for a saas file and environment combo, and checks for length limits ([link](https://github.com/app-sre/qontract-reconcile/pull/3559/files?diff=unified&w=0#diff-52a815feb4bcd8276212403b5f0ae22e3a3d9508acb5b7a75a5134dcf2657bbcR21), [link](https://github.com/app-sre/qontract-reconcile/pull/3559/files?diff=unified&w=0#diff-52a815feb4bcd8276212403b5f0ae22e3a3d9508acb5b7a75a5134dcf2657bbcR472-R489))
  * Use `build_saas_file_env_combo` method in `compose_console_url` function in `reconcile/openshift_saas_deploy.py` to generate tekton pipelinerun name and console url query parameter, and handle long saas file name exception ([link](https://github.com/app-sre/qontract-reconcile/pull/3559/files?diff=unified&w=0#diff-304fd1216ea8201e3242ba749fcc1bebbe71e61fd70851f7ed266e7318ed40b2L50-R55))
  * Use `build_saas_file_env_combo` method in `_construct_tekton_trigger_resource` function in `reconcile/openshift_saas_deploy_trigger_base.py` to generate tekton pipelinerun name and long name, and return both values ([link](https://github.com/app-sre/qontract-reconcile/pull/3559/files?diff=unified&w=0#diff-4135c05c1738ae40c0a45092f3748e2d3768f6777db8a2399f7e6afc50f2ccebL36), [link](https://github.com/app-sre/qontract-reconcile/pull/3559/files?diff=unified&w=0#diff-4135c05c1738ae40c0a45092f3748e2d3768f6777db8a2399f7e6afc50f2ccebL353-R359), [link](https://github.com/app-sre/qontract-reconcile/pull/3559/files?diff=unified&w=0#diff-4135c05c1738ae40c0a45092f3748e2d3768f6777db8a2399f7e6afc50f2ccebL398-R401))
  * Use `build_saas_file_env_combo` method in `_check_saas_file_env_combo_unique` method in `SaasHerder` class in `reconcile/utils/saasherder/saasherder.py` to check for uniqueness of tekton pipelinerun names and long names ([link](https://github.com/app-sre/qontract-reconcile/pull/3559/files?diff=unified&w=0#diff-52a815feb4bcd8276212403b5f0ae22e3a3d9508acb5b7a75a5134dcf2657bbcL477-R498))

### DIFF
--- a/reconcile/openshift_saas_deploy_trigger_base.py
+++ b/reconcile/openshift_saas_deploy_trigger_base.py
@@ -356,7 +356,7 @@ def _construct_tekton_trigger_resource(
     # we may want to revisit traceability, but this is compatible
     # with what we currently have in Jenkins.
     ts = datetime.datetime.utcnow().strftime("%Y%m%d%H%M")  # len 12
-    name = f"{tkn_name}-{ts}"
+    name = f"{tkn_name.lower()}-{ts}"
 
     parameters = [
         {"name": "saas_file_name", "value": saas_file_name},
@@ -395,7 +395,10 @@ def _construct_tekton_trigger_resource(
 
         body["spec"]["timeout"] = timeout
 
-    return OR(body, integration, integration_version, error_details=name), tkn_long_name
+    return (
+        OR(body, integration, integration_version, error_details=name),
+        tkn_long_name.lower(),
+    )
 
 
 def _register_trigger(name: str, already_triggered: set[str]) -> bool:
--- a/reconcile/openshift_saas_deploy_trigger_base.py
+++ b/reconcile/openshift_saas_deploy_trigger_base.py
@@ -321,6 +321,22 @@ def _pipeline_exists(
     return False
 
 
+def _build_pipeline_run_long_name(
+    saas_file_name: str,
+    env_name: str,
+) -> str:
+    return f"{saas_file_name}-{env_name}".lower()
+
+
+def build_pipeline_run_name_prefix(
+    saas_file_name: str,
+    env_name: str,
+) -> str:
+    return _build_pipeline_run_long_name(saas_file_name, env_name)[
+        :UNIQUE_SAAS_FILE_ENV_COMBO_LEN
+    ]
+
+
 def _construct_tekton_trigger_resource(
     saas_file_name: str,
     env_name: str,
@@ -350,13 +366,13 @@ def _construct_tekton_trigger_resource(
     Returns:
         OpenshiftResource: OpenShift resource to be applied
     """
-    long_name = f"{saas_file_name}-{env_name}".lower()
+    name_prefix = build_pipeline_run_name_prefix(saas_file_name, env_name)
     # using a timestamp to make the resource name unique.
     # we may want to revisit traceability, but this is compatible
     # with what we currently have in Jenkins.
     ts = datetime.datetime.utcnow().strftime("%Y%m%d%H%M")  # len 12
     # max name length can be 63. leaving 12 for the timestamp - 51
-    name = f"{long_name[:UNIQUE_SAAS_FILE_ENV_COMBO_LEN]}-{ts}"
+    name = f"{name_prefix}-{ts}"
 
     parameters = [
         {"name": "saas_file_name", "value": saas_file_name},
@@ -395,6 +411,7 @@ def _construct_tekton_trigger_resource(
 
         body["spec"]["timeout"] = timeout
 
+    long_name = _build_pipeline_run_long_name(saas_file_name, env_name)
     return OR(body, integration, integration_version, error_details=name), long_name
 
 
--- a/reconcile/openshift_saas_deploy_trigger_base.py
+++ b/reconcile/openshift_saas_deploy_trigger_base.py
@@ -33,7 +33,6 @@ from reconcile.utils.oc_map import (
 from reconcile.utils.openshift_resource import OpenshiftResource as OR
 from reconcile.utils.parse_dhms_duration import dhms_to_seconds
 from reconcile.utils.saasherder import (
-    UNIQUE_SAAS_FILE_ENV_COMBO_LEN,
     Providers,
     SaasHerder,
     TriggerSpecUnion,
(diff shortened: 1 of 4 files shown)


reconcile/openshift_saas_deploy_trigger_base.py
```diff
@@ -321,6 +321,22 @@ def _pipeline_exists(
...
+    return f"{saas_file_name}-{env_name}".lower()
+
+
+def build_pipeline_run_name_prefix(
+    saas_file_name: str,
+    env_name: str,
+) -> str:
+    return _build_pipeline_run_long_name(saas_file_name, env_name)[
+        :UNIQUE_SAAS_FILE_ENV_COMBO_LEN
+    ]
```
fishi0x01, line 337: nit: would it make sense to have a small comment here that uniqueness for `UNIQUE_SAAS_FILE_ENV_COMBO_LEN` is already ensured by saasherder? It might also be redundant to add that comment, as the constant naming already might suggest it. But still I was checking now if uniqueness is actually guaranteed here :sweat_smile:

reconcile/test/test_openshift_saas_deploy.py
```diff
@@ -0,0 +1,87 @@
...
+def test_compose_console_url_with_medium_saas_name(
+    saas_file_builder: Callable[..., SaasFile],
+) -> None:
+    saas_name = "saas-openshift-cert-manager-routes"
+    saas_file = saas_file_builder(saas_name)
+    env_name = "app-sre-production"
+
+    url = compose_console_url(saas_file, env_name)
+
+    expected_run_name = f"{saas_name}-{env_name}"[:50]
```
fishi0x01, line 66: nit: would it make sense to use `[:UNIQUE_SAAS_FILE_ENV_COMBO_LEN]` here?

fishi0x01 (approved):
First tests for openshift-saas-deploy! :heart:

reconcile/openshift_saas_deploy_trigger_base.py
hemslo: I actually have no idea what this constant means, just extracting existing logic to this func...

But you pointed out another usage of `UNIQUE_SAAS_FILE_ENV_COMBO_LEN`, let me see if I can reuse this logic in one more place, so this `UNIQUE_SAAS_FILE_ENV_COMBO_LEN` is not needed here anymore.

reconcile/test/test_openshift_saas_deploy.py
hemslo: I'm trying to reference implementation as few as possible, so accidentally change in implementation should lead to test fail.

chassing (approved):
first tests for openshift-saas-deploy!! ❤️

reconcile/openshift_saas_deploy_trigger_base.py
hemslo: @fishi0x01 I extracted another function from [saasherder.py](https://github.com/app-sre/qontract-reconcile/pull/3559/files#diff-52a815feb4bcd8276212403b5f0ae22e3a3d9508acb5b7a75a5134dcf2657bbcR473), more strict than before (case insensitive by `lower()`), should be ok, right?

rporres: I honestly find no reason that it shouldn't work, but saasherder.py is full of surprises. I would refrain from changing stuff beyond the nice refactor and I would remove the lower 😅

hemslo: @rporres that's a good point! It's already not consistent, in saas herder, the unique check is using case sensitive (no lower call), in actual saas deploy trigger run, it's using case insensitive (call lower). As it's not related to the issue I'm trying to fix, I updated the logic to keep the original behaviour, only reuse it to build console url, which case doesn't matter in UI search. Please have another look, thanks!

rporres: It looks great. Thx!

PR #3564, opened 2023-06-08
[gitlab-members] allow to run without a pagerduty instance

maorfr:
part of https://issues.redhat.com/browse/APPSRE-7755

the integration seems to be fine with this, as long as there are no permissions with a pagerduty target.
than again, if there are permissions with a pagerduty target, there is a pagerduty instance.

tl;dr this should be fine


janboll (approved):

cubismod:
Easier than I expected, ty Maor!

PR #3569, opened 2023-06-09
Refactor tf-repo for single repo enforcement

cubismod:
To prevent tenants from modifying multiple repos in one PR, checks are now in place to throw an exception if this happens. The CLI has also been refactored to only output to one repo.yaml file and tests updated to reflect this change.

Part of https://issues.redhat.com/browse/APPSRE-7672

## Why?
This is to limit complexity particularly on the `terraform apply` stage. Spawning multiple executor containers would probably require us to have a pipeline spawning other pipelines. Additionally, this makes tf-repo changes easier to audit as each modification to a repo's infrastructure is associated with just one MR.

### <samp>🤖 Generated by Copilot at 55530f6</samp>

* Change the `terraform_repo` command option from `-d`/`--output-dir` to `-o`/`--output-file` to reflect the new behavior of the integration ([link](https://github.com/app-sre/qontract-reconcile/pull/3569/files?diff=unified&w=0#diff-2ac404c3a4f4d289c1811816c94253a70cc6559b8de427388de06546141f1f43L1560-R1565))
* Update the `TerraformRepoIntegration` constructor and the `run` method to use the `output_file` parameter instead of the `output_dir` parameter ([link](https://github.com/app-sre/qontract-reconcile/pull/3569/files?diff=unified&w=0#diff-2ac404c3a4f4d289c1811816c94253a70cc6559b8de427388de06546141f1f43L1571-R1571), [link](https://github.com/app-sre/qontract-reconcile/pull/3569/files?diff=unified&w=0#diff-05e8f500c44ae2dc8a3b8087020cc1d2946506df84ea9a118e110c8d8ad82ba2L59-R59), [link](https://github.com/app-sre/qontract-reconcile/pull/3569/files?diff=unified&w=0#diff-05e8f500c44ae2dc8a3b8087020cc1d2946506df84ea9a118e110c8d8ad82ba2L96-R113), [link](https://github.com/app-sre/qontract-reconcile/pull/3569/files?diff=unified&w=0#diff-05e8f500c44ae2dc8a3b8087020cc1d2946506df84ea9a118e110c8d8ad82ba2L121-R122))
* Modify the `calculate_diff` method of the `TerraformRepoIntegration` class to return an optional `TerraformRepoV1` object instead of a list of `TerraformRepoV1` objects, and to raise an exception if more than one repo is modified in the merge request ([link](https://github.com/app-sre/qontract-reconcile/pull/3569/files?diff=unified&w=0#diff-05e8f500c44ae2dc8a3b8087020cc1d2946506df84ea9a118e110c8d8ad82ba2L247-R248), [link](https://github.com/app-sre/qontract-reconcile/pull/3569/files?diff=unified&w=0#diff-05e8f500c44ae2dc8a3b8087020cc1d2946506df84ea9a118e110c8d8ad82ba2L261-R275), [link](https://github.com/app-sre/qontract-reconcile/pull/3569/files?diff=unified&w=0#diff-05e8f500c44ae2dc8a3b8087020cc1d2946506df84ea9a118e110c8d8ad82ba2L293-R308))
* Update the test functions in `test_terraform_repo.py` to assert that the diff is equal to the expected repo object instead of a list containing the repo object, and to add two new test functions to cover the new logic in the `calculate_diff` method ([link](https://github.com/app-sre/qontract-reconcile/pull/3569/files?diff=unified&w=0#diff-f578fef16df2ffb09f95711b988012130afa7eea66f67d80c3168c0effe4c178L106-R106), [link](https://github.com/app-sre/qontract-reconcile/pull/3569/files?diff=unified&w=0#diff-f578fef16df2ffb09f95711b988012130afa7eea66f67d80c3168c0effe4c178L127-R127), [link](https://github.com/app-sre/qontract-reconcile/pull/3569/files?diff=unified&w=0#diff-f578fef16df2ffb09f95711b988012130afa7eea66f67d80c3168c0effe4c178L168-R168), [link](https://github.com/app-sre/qontract-reconcile/pull/3569/files?diff=unified&w=0#diff-f578fef16df2ffb09f95711b988012130afa7eea66f67d80c3168c0effe4c178R216-R238))

### DIFF
--- a/reconcile/terraform_repo.py
+++ b/reconcile/terraform_repo.py
@@ -264,6 +264,15 @@ class TerraformRepoIntegration(
         """
         diff = diff_iterables(existing_state, desired_state, lambda x: x.name)
 
+        merged = self.merge_results(diff)
+
+        # validate that only one repo is being modified in each MR
+        # this lets us fail early and avoid multiple GL requests we don't need to make
+        if len(merged) > 1:
+            raise Exception(
+                "Only one repository can be modified per merge request, please split your change out into multiple MRs"
+            )
+
         # added repos: do standard validation that SHA is valid
         if self.params.validate_git:
             for add_repo in diff.add.values():
@@ -291,13 +300,6 @@ class TerraformRepoIntegration(
             if self.params.validate_git:
                 self.check_ref(d.repository, d.ref)
 
-        merged = self.merge_results(diff)
-
-        # validate that only one repo is being modified in each MR
-        if len(merged) > 1:
-            raise Exception(
-                "Only one repository can be modified per merge request, please split your change out into multiple MRs"
-            )
         if len(merged) != 0:
             if not dry_run and state:
                 self.update_state(diff, state)
--- a/reconcile/terraform_repo.py
+++ b/reconcile/terraform_repo.py
@@ -56,7 +56,7 @@ class RepoOutput(BaseModel):
 
 
 class TerraformRepoIntegrationParams(PydanticRunParams):
-    output_dir: Optional[str]
+    output_file: Optional[str]
     validate_git: bool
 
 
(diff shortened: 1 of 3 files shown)


reconcile/terraform_repo.py
```diff
@@ -290,6 +287,14 @@ def calculate_diff(
...
                 self.check_ref(d.repository, d.ref)
 
+        # validate that only one repo is being modified in each MR
+        total_modifications = len(diff.add) + len(diff.change) + len(diff.delete)
+
+        if total_modifications > 1:
+            raise Exception(
+                "Only one repository can be modified per merge request, please split your change out into multiple MRs"
+            )
+
```
dwelch0, line 297: move to top of function (after `L261`)? Avoid unnecessary gitlab requests

cubismod: Because...it broke my tests. On second thought, this is kind of a naive approach to determining how many repos are being acted on so I'm going to take another crack at it

cubismod: Updated to make use of the `merge_results` function to determine how many repos are being updated because the data from the diff doesn't correspond 1:1 with how many repos are being updated. Also moved the check up top to fail early and avoid those extra requests. Tradeoff is that the user may see output like `update_repo repo_a` in their pr check output but the integration will fail so it will be pretty clear that there was an error.

dwelch0 (approved):
LGTM

PR #3570, opened 2023-06-11
[ocm-aws-infrastructure-access] ignore IAM SAs with assume_role

maorfr:
related to https://issues.redhat.com/browse/APPSRE-7735

when `assume_role` is provided on a namespace aws external resource for an `aws-iam-service-account` provider - it means ocm-aws-infrastructure-access is not the entity to grant the access.

in such case, the integration should skip this resource.

the `assume_role` field is widely used in fedramp, where `ocm-aws-infrastructure-access` is not used.


patjlm:
❤️ Thanks!

PR #3571, opened 2023-06-12
Limit the size of Tekton Pipeline names

rporres:
Pipeline names are built using the following convention:

```
o-<one per saas file pipeline name>-<saas file name>
```

The max size of `<saas file name>` was chosen to be 39 characters taking
into account that pipeline name was `openshift-saas-file`, so that we
can have pipelines with names up to 63 characters. This also worked well
with our PipelineRuns, which appended a 14 characeters date but trimmed
the Pipeline size so that, again, objects were right in size.

This didn't take into account that Tekton retries create PipelineRun names
that are named appending a 7 character random string to the corresponing
Pipeline name, which make certain Pipeline names not possible.

This patch makes sure that we have Pipeline names that will be suitable
for Tekton retries. It also changes many fixtures that use the old and
not longer allowed long pipeline name so that they are more realistic.

The code protects beyond what the schema will allow, since pipeline
names [are also being limited in their length](https://github.com/app-sre/qontract-schemas/pull/472)

part of [APPSRE-6031](https://issues.redhat.com/browse/APPSRE-6031)

### DIFF
--- a/reconcile/openshift_saas_deploy.py
+++ b/reconcile/openshift_saas_deploy.py
@@ -10,7 +10,9 @@ from reconcile import (
     queries,
 )
 from reconcile.gql_definitions.common.saas_files import PipelinesProviderTektonV1
-from reconcile.openshift_tekton_resources import build_one_per_saas_file_tkn_object_name
+from reconcile.openshift_tekton_resources import (
+    build_one_per_saas_file_tkn_pipeline_name,
+)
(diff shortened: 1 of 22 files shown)


chassing (approved):

hemslo (approved):
LGTM! Thank you for keeping test fixtures up to date!

PR #3572, opened 2023-06-12
[user-mgmt] allow cluster-admin mgmt for rosa clusters

geoberle:
currently, the user-mgmt capability refuses to manage cluster-admins groups on clusters without the respective AMS capability. that is the right behaviour for OSD but not for ROSA. for ROSA clusters, the AMS capability is not required.

fix for https://issues.redhat.com/browse/APPSRE-7709

### DIFF
--- a/reconcile/test/ocm/oum/test_oum_base.py
+++ b/reconcile/test/ocm/oum/test_oum_base.py
@@ -317,13 +317,22 @@ def test_build_spec_from_config_missing_group(
 
 
 @pytest.mark.parametrize(
-    "cluster_product, errors", [(PRODUCT_ID_OSD, True), (PRODUCT_ID_ROSA, False)]
+    "cluster_product, errors, expected_groups",
+    [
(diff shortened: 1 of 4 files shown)


maorfr (approved):

reconcile/test/ocm/oum/test_oum_base.py
```diff
@@ -340,9 +349,12 @@ def test_build_spec_from_config_cluster_admin_without_capability(
         },
     )
     assert len(specs) == 1
-    assert specs[0].roles == {}
-    assert specs[0].cluster == cluster
-    assert len(specs[0].errors) == 1
+    if errors:
```
hemslo, line 352: use parametrize tests and conditional assert together seems not serving the purpose of parametrize well, I suggest to extract more assertion into params then make test body consistent. For example, to have `expected_roles` in params, then

```python
assert len(specs[0].errors) == errors
assert specs[0].roles == expected_roles
```

patjlm (approved):

PR #3574, opened 2023-06-13
[saas-file-validator] validate targets in app may include parent app

maorfr:
with this PR, `validateTargetsInApp` now allows saas file targets (namespaces) to reference a parent app of it's own app.

### DIFF
--- a/reconcile/utils/saasherder/interfaces.py
+++ b/reconcile/utils/saasherder/interfaces.py
@@ -47,11 +47,15 @@ class HasSecretParameters(Protocol):
         ...
 
 
+class SaasParentApp(Protocol):
+    name: str
+
+
 class SaasApp(Protocol):
     name: str
 
     @property
-    def parent_app(self) -> Optional[SaasApp]:
+    def parent_app(self) -> Optional[SaasParentApp]:
         ...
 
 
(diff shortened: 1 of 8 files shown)


maorfr:
ref: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/71009

reconcile/utils/saasherder/interfaces.py
```diff
@@ -50,6 +50,10 @@ def secret_parameters(self) -> SaasSecretParameters:
 class SaasApp(Protocol):
     name: str
 
+    @property
+    def parent_app(self) -> Optional[SaasApp]:
+        ...
+
```
geoberle, line 56: i think the mypy errors you get are related to this protocol. it would need to match `AppV1`and `AppV1_AppV1`. while `AppV1` provides all the things the protocol needs, `AppV1_AppV1` does not (lacks the `parent_app`).

geoberle: maybe something like this could work? (disclaimer: i did not test this)

```
class SaasApp(Protocol):
    @property
    def name(self) -> str:
        ...

    @property
    def parent_app(self) -> Optional[Union[SaasApp, SaasAppWithoutParent]]:
        ...


class SaasAppWithoutParent(Protocol):
    @property
    def name(self) -> str:
        ...

geoberle: ah i see you solved it already

maorfr: @chassing is telling me what to do behind the scenes while i am ranting

geoberle (approved):
LGTM

PR #3588, opened 2023-06-14
[tf-repo] Fix for loading state

cubismod:
Fix `get_existing_state` for terraform repo. I was under the impression that `state.get()` would return a JSON string but instead it returns a Python object. Therefore, I'm switching to `parse_obj()`. See https://docs.pydantic.dev/latest/usage/models/#helper-functions for reference.


dwelch0 (approved):
LGTM

PR #3590, opened 2023-06-14
APPSRE-7561 add jinja2 hash filter for lists

fishi0x01:
We need a way to generate a hash from a list in jinja2. To my knowledge there is not such thing builtin in the jinja2 renderer.

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -311,9 +311,11 @@ def hash_list(input: Iterable) -> str:
     str_lst = []
     for el in lst:
         if isinstance(el, list) or isinstance(el, dict):
-            raise RuntimeError(f"jinja2 hash_list function received non-primitive value {el}. All values received {lst}")
+            raise RuntimeError(
+                f"jinja2 hash_list function received non-primitive value {el}. All values received {lst}"
+            )
         str_lst.append(str(el))
-    msg = "a" # keep non-empty for hashing empty list
+    msg = "a"  # keep non-empty for hashing empty list
     msg += "".join(sorted(str_lst))
     m = hashlib.sha256()
     m.update(msg.encode("utf-8"))
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -310,7 +310,7 @@ def hash_list(input: Iterable) -> str:
     lst = list(input)
     str_lst = []
     for el in lst:
-        if isinstance(el, list) or isinstance(el, dict):
+        if isinstance(el, (list, dict)):
             raise RuntimeError(
                 f"jinja2 hash_list function received non-primitive value {el}. All values received {lst}"
             )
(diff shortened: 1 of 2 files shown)


reconcile/openshift_resources_base.py
```diff
@@ -349,6 +349,17 @@ def urlunescape(string: str, encoding: Optional[str] = None) -> str:
     return parse.unquote(string, encoding=encoding)
 
 
+def hash_list(input):
+    lst = list(input)
+    sorted_list = sorted(lst, key=lambda x: x["name"])
```
geoberle, line 354: this code expects the list elements to be dicts that have at least a name key. isn't that restricting the reusability of the hash_list filter?

fishi0x01: You are absolutely right! I need to add more args to the filter to make it reusable. Also the char length of the hash.
This is still very early draft mode - will adjust the PR description accordingly. Thanks though for the fast input :) :bow:

fishi0x01: Probably a filter is not the best idea in the first place here - might be better to do what we do for `query()` too. That way we have a function with arguments we can pass, like

```jinja2
{% hash_list(namespaces, length=6, sort_by=lambda x: x["name"]) %}
```

fishi0x01: I came to the conclusion that lambda wont cut it. It would force us to do some kind of `eval()` on strings defined in jinja2 templates. I dont like that at all for security reasons. So I expect now a flat list of non-dicts here. A use-case can be found here: https://gitlab.cee.redhat.com/service/app-interface/-/blob/0800c7be8f25dfff09e9c22acd79c7480bc7f673/resources/setup/dynatrace/log-config-job.yaml.j2#L8

reconcile/openshift_resources_base.py
```diff
@@ -300,6 +300,19 @@ def lookup_graphql_query_results(query: str, **kwargs) -> list[Any]:
     return results
 
 
+def hash_list(input: Iterable) -> str:
+    """
+    Deterministic hash of a list for jinja2 templates.
+    The order of the list doesnt matter as it is sorted
```
rporres, line 306:
```suggestion
    The order of the list doesn't matter as it is sorted
```

reconcile/openshift_resources_base.py
```diff
@@ -300,6 +300,19 @@ def lookup_graphql_query_results(query: str, **kwargs) -> list[Any]:
...
+    before hashing. Note, that the list elements
+    must be flat (no dicts).
```
rporres, line 308: What happens if you pass a dict? Shouldn't it raise a well-known exception?

reconcile/openshift_resources_base.py
fishi0x01: Fixed :+1:

reconcile/openshift_resources_base.py
fishi0x01: Very good point. Added a check that we only work with primitives. Added some tests for the function too.

rporres (approved):

PR #3591, opened 2023-06-14
Remove data processing from report from Jenkins CD

rporres:
The promotions section is always empty so the data processing makes no sense.

The post_deploy_jobs appears to have more information, e.g.

```
post_deploy_jobs:
- cluster: insights-perf
  namespace: platform-mq-perf
  post_deploy_job: false
```

but there are never elements with post_deploy_job set to true since post_deploy_jobs are handled differently in Tekton world.

Since they don't longer show anything meaningful and they build on top of soon to be removed app-interface settings such as saasDeployJobTemplate it is good to remove them.

It's important to note that this patch does not remove the sections in the final report as that would imply a change in the content format version.

### DIFF
--- a/reconcile/jenkins_job_builder.py
+++ b/reconcile/jenkins_job_builder.py
@@ -44,11 +44,6 @@ QONTRACT_INTEGRATION = "jenkins-job-builder"
 GENERATE_TYPE = ["jobs", "views"]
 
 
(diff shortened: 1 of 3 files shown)


fishi0x01 (approved):
Great cleanup!! :heart: :bow:

maorfr (approved):

PR #3592, opened 2023-06-15
[terraform msk] multi user support

chassing:
Support the creation of multiple MSK SASL/SCRAM users.

Depends on: https://github.com/app-sre/qontract-schemas/pull/467
Ticket: [APPSRE-7534](https://issues.redhat.com/browse/APPSRE-7534)


hemslo (approved):
LGTM!
Just one thing, there is no test for this change...

PR #3593, opened 2023-06-15
[tf-repo] update to allow multi repo execution in non dry run

cubismod:
In #3569, I refactored the QR integration to enforce a strict limit of one repo being acted on at a time. This was to cut down on executor complexity and enforce smaller and easier to parse merge requests. However, while we can and should gate tenants to only update one repo at once, there isn't a guarantee that tf-repo will only ever work on one repo at a time.

There could possibly be edge cases due to:
- state loss/network issues causing int to be unable to get existing desired state
- pipeline execution delays causing multiple repos to be acted on in one pipeline

Therefore, this merge request keeps the enforcement of one repo per MR, but allows multiple repos to be acted on when in production. This coincides with a logic change for: https://github.com/app-sre/terraform-repo-executor/pull/9

### DIFF
--- a/reconcile/terraform_repo.py
+++ b/reconcile/terraform_repo.py
@@ -41,18 +41,22 @@ class RepoSecret(BaseModel):
 
 
 class RepoOutput(BaseModel):
-    """
-    Output of the QR terraform-repo integration and input to the executor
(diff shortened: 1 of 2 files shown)


dwelch0 (approved):
LGTM

PR #3596, opened 2023-06-17
[change-owners] resourcefile backrefs for ownership expansion

geoberle:
this PR introduces ownership expansions for resource files based on their usage (backref). if a resourcefile is used within datafile, ownership of the datafile can expand to ownership of the resourcefile with the following configuration, showcasing it for namespaces and RDS default files.

```yaml
$schema: /app-interface/change-type-1.yml
name: namespace-owner
contextType: datafile
contextSchema: /openshift/namespace-1.yml

changes:
- provider: change-type
  changeTypes:
  - $ref: /app-interface/changetype/rds-defaults-maintainer.yaml
  context:
    where: backrefs (1)
    selector: externalResources[?(@.provider=="aws")].resources[?(@.provider=="rds")].defaults (2)
```

(1) `where: backrefs` defines that the backrefs of a resourcefile need to be inspected in order to relate it back to the owned namespace file
(2) ... but to be more restrictive, the RDS resource file must be used in a very specific location within the namespace file

this way, certain resourcefiles don't need to be owned explicitely anymoe via the `resource-owner` change-type but can be automatically owned based on context

disclaimer: if a resourcefile is used in multiple datafiles (e.g. shared RDS default files), ownership is expaned to all datafiles using the resourcefile, which ultimatly ends up granting owners of ALL datafiles approval permissions.

depends on schema https://github.com/app-sre/qontract-schemas/pull/470
part of https://issues.redhat.com/browse/APPSRE-7812

### DIFF
--- a/reconcile/change_owners/bundle.py
+++ b/reconcile/change_owners/bundle.py
@@ -101,6 +101,15 @@ def _clean_datafile_content(data: Optional[dict[str, Any]]) -> Optional[dict[str
     return {k: v for k, v in data.items() if k not in DATAFILE_CONTENT_CLEANUP_FIELDS}
 
 
+class QontractServerResourcefileBackref(BaseModel):
+    """
+    Represents a backref from a resourcefile to a datafile
+    """
+
+    path: str
+    datafileschema: str = Field(..., alias="datafileSchema")
+
+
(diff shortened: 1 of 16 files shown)


maorfr (approved):
LGTM!

PR #3598, opened 2023-06-19
[redhat idp] OIDC configuration for App-Interface and SRE Capability clusters

chassing:
This is part one of the RedHat IDP Capability and covers the OIDC configuration.

This PR introduces the base RHIDP methods and refactors the `ocm_oidc_idp` integration to run on App-Interface and SRE Capability clusters.

This integration comes in two flavors:
* The App-Interface one queries clusters from App-Interface 
* And the SRE Capability (standalone) fetches cluster with the `sre-capabilities.rhidp` subscription label

Then, OIDC will be configured via OCM on those clusters.

Design doc: TODO
Ticket: [SDE-2620](https://issues.redhat.com/browse/SDE-2620)

For the reviewers:

* run the app-interface flavor
  ```shell
  $ qd profile create --integration-extra-args "--vault-input-path app-sre/integrations-input/rhidp" ocm-oidc-idp
  $ qd profile run dev ocm-oidc-idp
  ```
* run the OCM flavor
  ```shell
  $ qd profile create --integration-extra-args "--vault-input-path app-sre/integrations-input/rhidp" ocm-oidc-idp-standalone
  $ qd profile run dev ocm-oidc-idp-standalone
  ```


reconcile/rhidp/ocm_oidc_idp/integration.py
```diff
@@ -0,0 +1,58 @@
...
+
+        run(
+            integration_name=self.name,
+            clusters=clusters,
+            secret_reader=self.secret_reader,
+            vault_input_path=self.params.vault_input_path,
+            dry_run=dry_run,
+        )
+
+    def get_clusters(self, query_func: Callable) -> list[ClusterV1]:
```
hemslo, line 53: If here pass `gqlapi` instead of `query_func`, it will be easier to setup mock in test, what do you think?

hemslo (approved):
LGTM! But may need another reviewer who's more familiar with this feature.

reconcile/rhidp/ocm_oidc_idp/base.py
```diff
@@ -1,33 +1,63 @@
...
 
 
+def run(
+    integration_name: str,
+    clusters: Iterable[ClusterV1],
+    secret_reader: SecretReaderBase,
+    vault_input_path: str,
+    dry_run: bool,
+) -> None:
+    with metrics.transactional_metrics(integration_name) as metrics_container:
```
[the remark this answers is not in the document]
chassing: The kudos goes to @geoberle; he created the `metrics` package

reconcile/rhidp/ocm_oidc_idp/integration.py
chassing: TBH, I don't like mocking. Passing a `query_func` makes testing easier without mocking. This is a common pattern, used in other test cases too.

maorfr (approved):

reconcile/utils/ocm/sre_capability_labels.py
```diff
@@ -17,11 +17,15 @@
...
+def sre_capability_label_key(
+    sre_capability: str, config_atom: Optional[str] = None
+) -> str:
     """
-    Generates label keys for aus, compliant with the naming schema defined in
+    Generates label keys compliant with the naming schema defined in
     https://service.pages.redhat.com/dev-guidelines/docs/sre-capabilities/framework/ocm-labels/
     """
+    if config_atom is None:
+        return f"sre-capabilities.{sre_capability}"
```
geoberle, line 28: ❤️

geoberle:
LVGTM

geoberle (approved):

reconcile/rhidp/ocm_oidc_idp/integration.py
hemslo: Mocking is needed anyway, but in different layers, it can be a manual one like [this](https://github.com/app-sre/qontract-reconcile/pull/3598/files#diff-551e1d77b15ca2654a18f6b7b8e257ec10fcdccc963dd9f6a21d8678312a74b9R29), but it will miss some handy assertions like `assert_called_once_with`. I also don't like mocking, I like classical TDD and sociable unit tests, [ref](https://martinfowler.com/articles/practical-test-pyramid.html#SociableAndSolitary).

Just clarify what I mean `gqlapi` is easier than `query_func`, if it's

```python
    def get_clusters(self, gqlapi: gql.GqlApi) -> list[ClusterV1]:
        return get_clusters(self.name, gqlapi.query)
```

Then unit test can be done like this

```python
from unittest.mock import create_autospec

mock_gqlapi = create_autospec(spec=gqlapi.GqlApi)
mock_get_clusters = mocker.patch('reconcile.rhidp.ocm_oidc_idp.integration.py.get_clusters')
intg = OCMOidcIdpIntegration()

intg.get_clusters(mock_gqlapi)

mock_get_clusters.assert_called_once_with("ocm-oidc-idp", mock_gqlapi.query)
```

Another part makes this method hard to test is the free var `get_clusters` from `from reconcile.rhidp.common import get_clusters`, that requires a `mocker.patch` to test this method in solitary.

Your tests call through `get_clusters`, only the last layer is mocked, so there is no problem with it.

I tried to use dependency injection way of passing `query_func`, but mypy is not happy about it, that's why in the end I switched to passing `GqlApi` object, https://github.com/app-sre/qontract-reconcile/pull/3467#discussion_r1173829008

PR #3599, opened 2023-06-21
[AUS] handle reconcile errors and continue reconciling

geoberle:
reconcile errors in one organization prevents reconciling to continue for other organizations

this PR catches such unhandled exceptions and summarizes them at the end

https://issues.redhat.com/browse/APPSRE-7835

### DIFF
--- a/reconcile/aus/advanced_upgrade_service.py
+++ b/reconcile/aus/advanced_upgrade_service.py
@@ -116,15 +116,18 @@ class AdvancedUpgradeServiceIntegration(OCMClusterUpgradeSchedulerOrgIntegration
         dry_run: bool,
         org_upgrade_spec: OrganizationUpgradeSpec,
         exception: Exception,
-    ) -> None:
+    ) -> bool:
         """
         AUS will not fail on a reconcile issue. If issues should be noticed by an SRE team,
         alerts based on the metrics in the `reconcile.aus.metrics` module should be set up.
+
+        The function returns true to indicte that the exception was properly handled in the current context.
         """
         logging.error(
             f"Failed to reconcile cluster upgrades in OCM organization {org_upgrade_spec.org.org_id}",
             exc_info=exception,
         )
+        return True
 
 
 def discover_clusters(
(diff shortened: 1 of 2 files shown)


reconcile/aus/advanced_upgrade_service.py
```diff
@@ -116,15 +116,18 @@ def signal_reconcile_issues(
         dry_run: bool,
         org_upgrade_spec: OrganizationUpgradeSpec,
         exception: Exception,
-    ) -> None:
+    ) -> bool:
         """
         AUS will not fail on a reconcile issue. If issues should be noticed by an SRE team,
         alerts based on the metrics in the `reconcile.aus.metrics` module should be set up.
+
+        The function returns true to indicte that the exception was properly handled in the current context.
```
patjlm, line 124: nit:
```suggestion
        The function returns true to indicate that the exception was properly handled in the current context.
```

patjlm (approved):

reconcile/aus/advanced_upgrade_service.py
```diff
@@ -116,15 +116,18 @@ def signal_reconcile_issues(
...
         """
         logging.error(
             f"Failed to reconcile cluster upgrades in OCM organization {org_upgrade_spec.org.org_id}",
             exc_info=exception,
         )
+        return True
```
janboll, line 130: Do you have something planned? There is no False returned, why use a bool here?

janboll: Ah, just saw `AdvancedUpgradeSchedulerBaseIntegration.signal_reconcile_issues`

janboll: You add an attribute with a handler method, that is executed by `signal_reconcile_issues`. That method is registered in init. by that, you could avoid changing the implementation of `signal_reconcile_issues`.
Could make reading this code less of a surprise.
Example `signal_reconcile_issues`

```
if self.handler:
    return self.handler(dry_run, org_upgrade_spec, e)
return False
```

geoberle: i like the idea of extension points like handlers. in this case i would rather add some comments in the base function and overriden one to give the reader some hints about what is going on.

reconcile/aus/advanced_upgrade_service.py
geoberle: fixed. ty

PR #3604, opened 2023-06-22
Add the assumed role name into each terraform provider name.

patjlm:
This allows to not share providers when 2 clusters are running on the same account. This is especially important when different assume role and/or region need to be used.

Note: this will rename all providers and add some. We checked with @maorfr and terraform accepts that without wanting to reapply everything.

The same function `get_alias_name_from_assume_role` is used to set the provider name and reference that provider from resources (as far as we could see).


andreugallofre (approved):
lgtm

maorfr (approved):

PR #3606, opened 2023-06-23
[AUS] cluster upgrade overview dashboard improvments

geoberle:
* show recently upgraded clusters
* show stats for recently upgraded clusters
* show version distribution

### DIFF
--- a/grafana-dashboards/sre-capability-aus.configmap.yaml
+++ b/grafana-dashboards/sre-capability-aus.configmap.yaml
@@ -202,6 +202,18 @@ data:
                     "id": "custom.filterable"
                   }
                 ]
+              },
+              {
+                "matcher": {
+                  "id": "byName",
+                  "options": "Cluster"
+                },
+                "properties": [
+                  {
+                    "id": "custom.minWidth",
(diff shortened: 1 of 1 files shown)


patjlm (approved):

grafana-dashboards/sre-capability-aus-ops.configmap.yaml
```diff
@@ -336,9 +997,9 @@ data:
       },
       "timepicker": {},
       "timezone": "",
-      "title": "AUS Ops Dashboard",
-      "uid": "bqA9EFwVk",
-      "version": 11,
+      "title": "AUS Cluster Upgrade Overview",
```
patjlm, line 1000: I wonder if you updated the good file there 🤔

patjlm (changes requested):
need to check the correct dashboard is being updated

grafana-dashboards/sre-capability-aus-ops.configmap.yaml
geoberle: no i did not :) thank you 🙇

PR #3607, opened 2023-06-23
[redhat-idp] SSO client management

chassing:
This is the second part of the RHIDP capability; it introduces a new SSO client management integration in two flavors (app-interface & OCM).
The integration creates and deletes SSO clients in given Keycloak instances.

Design doc: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/redhat-idp.md
Ticket: SDE-2620
Follow up to: #3598

For the reviewers:
* This PR includes some `ocm-oidc-idp` refactoring due to the nature of both integrations


reconcile/rhidp/ocm_oidc_idp/standalone.py
```diff
@@ -41,10 +42,19 @@ def name(self) -> str:
...
+            # data query
+            # clusters with enabled RHIDP
             clusters = self.get_clusters(
-                ocm_env=ocm_env, org_ids=self.params.ocm_organization_ids
+                ocm_env=ocm_env,
+                org_ids=self.params.ocm_organization_ids,
+                label_value=RhidpLabelValue.ENABLED,
+            )
+            # with disabled RHIDP
+            clusters += self.get_clusters(
```
hemslo, line 53: Since we need to get both enabled and disabled clusters back, would it make sense to let `get_clusters` support optional `label_value` as filter, if not provided, then return all.

Or if we need to filter out just `enabled` or `disabled` value for that key, we can use `is_in` https://github.com/app-sre/qontract-reconcile/blob/51502def6e6a20f3e8a51dba8b19342112355bcc/reconcile/utils/ocm/search_filters.py#L281

chassing: I need to get either the enabled (or not labeled) or the disabled clusters to set or not to set the `auth` cluster attribute. The `ClusterDetails` model [doesn't have](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/ocm/clusters.py#L84) the cluster labels included, so it's much easier to separate it at the beginning, instead of comparing `enabled` vs `disabled` in the `get_clusters` method.

hemslo: Can you get not labeled clusters back in current code? https://github.com/chassing/qontract-reconcile/blob/aa91a8843e36a61f5dcabb5157a11a3e3e230002/reconcile/rhidp/common.py#L42C26-L42C26
indicate the label filter is always applied.

chassing: Sorry, I meant, I need to distinguish between `enabled`, `disabled`, and not labeled clusters. The `rhidp-sso-client` integration has a state of all clusters in Vault and can delete obsolete clusters (not `enabled` == `disable` or not labeled) but the `oidc` integration doesn't have a state for all clusters, therefore I need to introduce a `disabled` state.

chassing: For visibility, here are two flowcharts for both integrations.
 
For `SSOClientStandalone` the logic is the following:

```mermaid
flowchart TD
  OCM -- OCM_CLUSTERS := $RHIDP_LABEL_KEY : enabled --> B[Compare both cluster list]
  Vault -- VAULT_CLUSTERS := SSO client data --> B[Compare both cluster list]
  B -->|cluster in OCM_CLUSTERS and not in VAULT_CLUSTERS | C[create SSO client]
  B -->|cluster not in OCM_CLUSTERS and in VAULT_CLUSTERS| D[delete SSO client]
  B -->|cluster in OCM_CLUSTERS in VAULT_CLUSTERS| E[do nothing]
```
As you can see, to delete an SSO client, it's sufficient if the cluster isn't labeled anymore or labeled with something different to `enabled`.

For `OCMOidcIdpStandalone`, the logic is slightly different because there is no cluster state in Vault.

```mermaid
flowchart TD
  OCM -- $RHIDP_LABEL_KEY : enabled --> C[add a ClusterAuthOIDCV1 to cluster.auth]
  OCM -- $RHIDP_LABEL_KEY : disabled --> D[set an empty cluster.auth attribute]

  CLUSTER_OCM --> OIDC{OIDC configured?}
  C --> B[ClusterAuthOIDCV1?]
  D --> B

  OIDC -->|Yes| F[delete OIDC config]
  B -->|No| F

  OIDC -->|No| G[create OIDC config]
  B -->|No| G
```
As you can see, if the cluster isn't labeled, the `OCMOidcIdpStandalone` doesn't "know" the cluster at all and can't handle it.

You are right; the implicit usage of `auth_name` as a flag is confusing, let me change that.

chassing: Fixed in 507b00a3297bd72ba049e46faf8e12e7237876a7

hemslo: Wow, nice mermaid charts! Thank you so much!

in https://github.com/app-sre/qontract-reconcile/commit/507b00a3297bd72ba049e46faf8e12e7237876a7#diff-8ad175c675b0abe1f96c4abbe39dcbdd5d82d7f21ee16f6ca720b89d27c1dfc2R92-R97, we have to iterate clusters and patch `auth` field again, I find sometimes it's easy to forget to set fields after init, how about assign it when it's constructed like this:

```python
def build_cluster_obj(
    ocm_env: OCMEnvironment,
    cluster: ClusterDetails,
    auth,
):
return ClusterV1(
  ...
  auth=auth,
)
```

```python
# reconcile/rhidp/ocm_oidc_idp/standalone.py

def _build_cluster(
  self,
  ocm_env,
  cluster,
  label_value,
):
  auths = build_cluster_auths(
    name=self.params.auth_name,
    issuer_url=self.params.auth_issuer_url,
    ) if label_value == RhidpLabelValue.ENABLED else []
  return build_cluster_obj(ocm_env=ocm_env, cluster=c, auth=auths)

...
# def get_clusters(
 clusters = [
            self._build_cluster(ocm_env=ocm_env, cluster=c, label_value=label_value)
            for ocm_clusters in clusters_by_org.values()
            for c in ocm_clusters
        ]
```

similar here, 
https://github.com/app-sre/qontract-reconcile/commit/507b00a3297bd72ba049e46faf8e12e7237876a7#diff-e80f4985ebd4a036531423d08771de15099880dbd2b89f8a8faeeb2bbc7837f1R73

```python
# reconcile/rhidp/sso_client/standalone.py
def _build_cluster(
  self,
  ocm_env,
  cluster,
):
  auths = build_cluster_auths(
    name=self.params.auth_name,
    issuer_url=self.params.auth_issuer_url,
    )
  return build_cluster_obj(ocm_env=ocm_env, cluster=c, auth=auths)

..
        clusters = [
            self._build_cluster(
              ocm_env=ocm_env,
              cluster=c,
            )
            for ocm_clusters in clusters_by_org.values()
            for c in ocm_clusters
        ]
```

hemslo: In the last flowchart, Both `B` -> `F` & `B` -> `G` are `No`, one of them should be `Yes`, right?

```mermaid
flowchart TD
  OCM -- $RHIDP_LABEL_KEY : enabled --> C[add a ClusterAuthOIDCV1 to cluster.auth]
  OCM -- $RHIDP_LABEL_KEY : disabled --> D[set an empty cluster.auth attribute]

  CLUSTER_OCM --> OIDC{OIDC configured?}
  C --> B[ClusterAuthOIDCV1?]
  D --> B

  OIDC -->|Yes| F[delete OIDC config]
  OIDC -->|No| G[create OIDC config]
  
  B -->|Yes| G
  B -->|No| F
```

hemslo: Nice! Thanks!

hemslo (approved):

PR #3608, opened 2023-06-26
bump black version

chassing:
Use the latest black 23.3 version.


fishi0x01 (approved):
LGTM. Might make sense to announce it so ppl update locally, else it might raise questions.

PR #3609, opened 2023-06-26
fix vpc-peerings-validator return

BumbleFeng:
and fix some format issue


reconcile/vpc_peerings_validator.py
```diff
@@ -68,15 +68,13 @@ def validate_no_cidr_overlap(
...
                     }
                     peerings_entries.append(vpc_peering_info)
-        overlaps_peering_entries_dict = find_cidr_duplicates_and_overlap(
-            cluster.name, peerings_entries
-        )
-        if not overlaps_peering_entries_dict:
-            return False
+            find_overlap = find_cidr_overlap(cluster.name, peerings_entries)
+            if find_overlap:
+                return False
```
BumbleFeng, line 73: here is the trick

PR #3612, opened 2023-06-27
[gitlab-housekeeping] add self-service label to qontract_reconcile_merged_merge_requests counter

maorfr:
with this new label, we can distinguish between self-serviceable and not-self-serviceable MRs.
this will allow us to visualize our self-service frequency, and also to establish internal SLOs for percentage of self-serviceable MRs (if we want).

example of the current metric: https://prometheus.appsrep05ue1.devshift.net/graph?g0.expr=qontract_reconcile_merged_merge_requests_total&g0.tab=1&g0.stacked=0&g0.show_exemplars=0&g0.range_input=1h


geoberle (approved):
❤️

PR #3613, opened 2023-06-27
[AUS] reference version history of an org around ocm-env + org-id

geoberle:
change the way the version history of an OCM organization is referenced. right now it is referenced by OCM organization name. this is problematic when

1) the name of an organization changes (rarely)
2) the name of an organization is not unique (quite common in AMS)

especially organization name duplicates are problem in environments where the name is not controlled by us, e.g. AMS organization names.

the path of the state will change as follows

* old: `$integration/$org-name`
* new: `$integration/$ocm-env/$org-id`

please note that this PR does not contain any migration logic for the state. state keys need to be renamed before these integrations run. a migration script is prepared to handle this.

[APPSRE-7846](https://issues.redhat.com/browse/APPSRE-7846)


reconcile/aus/base.py
```diff
@@ -470,6 +471,10 @@ def update_history(
     version_data.check_in = now
 
 
+def version_data_state_key(ocm_env: str, org_id: str, addon_id: Optional[str]) -> str:
+    return f"{ocm_env}/{org_id}/{addon_id}" if addon_id else f"{ocm_env}/{org_id}"
```
patjlm, line 475: ❤️

patjlm (approved):
Thanks!

PR #3614, opened 2023-06-27
Add alerts report from sd-app-sre-channel

rporres:
Create a report parsing the output of the conversations.list API
endpoint for our #sd-app-sre-alert channel that can be filtered by days
from the moment that report is run.

We also have explicit timestamp to be able to recreate a certain report
from the past.

This is a `get` qontract-cli command as the rest of the similar reports
we have.

It generates a table with the following colums:

* Alert name
* Triggered
* Resolved
* Median time to resolve

A few comments about the potential scenarios:

* Triggered doesn't need to be equal to resolved, as triggering message
  can be outside the considered time window.
* Median time to resolve sometimes cannot be determined as we use alert
  messages for that that can contain dynamic parts.

part of APPSRE-7762

### DIFF
--- a/reconcile/test/fixtures/slack_api/conversations_history_messages.yaml
+++ b/reconcile/test/fixtures/slack_api/conversations_history_messages.yaml
@@ -1,0 +1,549 @@
+# PrometheusTargetFlapping alerts
+# Cases to test multiple entries to calculate medians
+## Target pushgateway-nginx-gate-1/10.131.0.19:9091
+## elapsed: 900s
+- attachments:
+  - actions:
+    - id: '1'
+      style: ''
+      text: 'Runbook :green_book:'
+      type: button
+      url: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/sop/prometheus/prometheus-target-flapping.md
+    - id: '2'
+      style: ''
+      text: 'Query :mag:'
+      type: button
+      url: https://prometheus.app-sre-prod-01.devshift.net/graph?g0.expr=changes%28up%7Bnamespace%3D%22app-sre-observability-production%22%7D%5B15m%5D%29+%3E+4&g0.tab=1
+    - id: '3'
+      style: ''
+      text: 'Alert Definition :git:'
+      type: button
+      url: https://gitlab.cee.redhat.com/service/app-interface/blob/master/resources/observability/prometheusrules/prometheus-app-sre.prometheusrules.yaml.j2
+    - id: '4'
+      style: ''
+      text: 'Silence :no_bell:'
+      type: button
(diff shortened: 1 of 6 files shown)


rporres:
```
qontract-cli --config config.development.toml get sd-app-sre-alert-report  --days 2
ALERT NAME                                                        TRIGGERED    RESOLVED  MEDIAN TIME TO RESOLVE
--------------------------------------------------------------  -----------  ----------  ------------------------
PatchmanAlertEvalDelay                                                    6           6
KubePodCrashLooping                                                       4           4  299.9933099746704
RDSStorageLow                                                             3           1
SLOMetricAbsent                                                           3           2  2700.006809949875
App-cost-download-queue-overload-In-hccm                                  2           2  2400.0005301237106
ThanosReceiveHttpRequestErrorRateHigh                                     1           1
AMS RHIT certificate expiring in less than 7 days - production            1           1  73200.14591002464
QontractReconcileError                                                    1           1  299.989129781723
ContainerSecurityOperatorPodCount                                         0           1
```

patjlm:
Which unit is used for `Median time to resolve` ? That could be worth providing in the output.

maorfr (approved):

maorfr:
so all in all O(1) calls to slack api?

rporres:
> Which unit is used for Median time to resolve ? That could be worth providing in the output.

Seconds. Definitely worth clarifying and now that I think about it, formatting it more nicely.

rporres:
> so all in all O(1) calls to slack api?

I'm not sure I follow what you have in mind. The slack calls to that endpoint are paginated, so if you want to get the alerts in the last month you will issue more calls than if you just want the alerts of the last day... Does this help your question?

maorfr:
yes, it does!
a bit out of scope for the PR itself, but have you thought about rate limits?

rporres:
@patjlm I've formatted the time to resolve in d88b930c. Let me know what you think.

```
ALERT NAME                                               TRIGGERED    RESOLVED  MEDIAN TIME TO RESOLVE (H:MM:SS)
-----------------------------------------------------  -----------  ----------  ----------------------------------
QontractReconcileError                                           4           2  0:25:00
KubePodCrashLooping                                              3           3  0:05:00
SRSServiceRegistryAPIWriteLatency30mto6hP99BudgetBurn            3           3
UHC account manager Too High CPU usage - production              2           2  0:05:00
PatchmanAlertEvalDelay                                           2           2
RDSStorageLow                                                    2           0
ThanosReceiveHttpRequestErrorRateHigh                            1           1
App-cost-download-queue-overload-In-hccm                         1           1  0:45:00
QontractReconcileIntegrationStuck2hours                          1           1  0:05:00
```

rporres:
> a bit out of scope for the PR itself, but have you thought about rate limits?

Getting the last day's alert is usually one call to the conversations.history endpoint with our alert. I haven't had any issue while testing.

I was thinking in running it twice a day (once at the start of every IC shift) with a one day period and once a month (the beginning of each month), printing the output in the app-interface-output. I don't think this is going to cause rate limiting issues.

As for the code, the library handles retries when rate limited.

patjlm (approved):
Looks great, thanks!

PR #3615, opened 2023-06-28
[qontract-cli] fix NotRenderableError on saas-targets subcommand

chassing:
A `saas.resource_template.target.parameter` can be a JSON object which can't be rendered in a rich table. Cast all parameters to string.


hemslo (approved):
LGTM!

patjlm (approved):

PR #3618, opened 2023-06-28
[AUS] frozen condition for cluster upgrade policies

geoberle:
Sometimes an upgrade for a cluster needs to prevented. The current ways of setting high soakdays or removing the policy are not nice because for both the fact that the cluster is just temporary frozen is not apparent.

Adding a new `frozen` condition to upgrade policies allows to define explicit upgrade freeze.

depends on https://github.com/app-sre/qontract-schemas/pull/479
part of https://issues.redhat.com/browse/APPSRE-7870

### DIFF
--- a/reconcile/aus/advanced_upgrade_service.py
+++ b/reconcile/aus/advanced_upgrade_service.py
@@ -306,6 +306,7 @@ class ClusterUpgradePolicyLabelSet(BaseModel):
     schedule: str = Field(alias=aus_label_key("schedule"))
     mutexes: Optional[CSV] = Field(alias=aus_label_key("mutexes"))
     sector: Optional[str] = Field(alias=aus_label_key("sector"))
+    frozen: Optional[bool] = Field(alias=aus_label_key("frozen"))
     _schedule_validator = validator("schedule", allow_reuse=True)(cron_validator)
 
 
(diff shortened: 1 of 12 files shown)


reconcile/aus/advanced_upgrade_service.py
```diff
@@ -306,6 +306,7 @@ class ClusterUpgradePolicyLabelSet(BaseModel):
     schedule: str = Field(alias=aus_label_key("schedule"))
     mutexes: Optional[CSV] = Field(alias=aus_label_key("mutexes"))
     sector: Optional[str] = Field(alias=aus_label_key("sector"))
+    frozen: Optional[bool] = Field(alias=aus_label_key("frozen"))
```
janboll, line 309: Did you consider using `disable_scheduling` or just `disabled` instead of frozen? Frozen leaves some room for interpretation IMHO.

janboll: I.e. will soakdays be updated if frozen? What about inheritance in that case?

geoberle: soak days will be updated. this condition is evaluated the same as the others. condition checking happens within the context of `calculate_diff` while soak days are updated before when the version history is prepared.

i've not considered other naming. can you elaborate on other interpretations?

janboll: Just the immutability of the policy. I.e. would frozen make the policy immutable?

geoberle: ah understood. @patjlm wdyt about `disabled`?

patjlm: I think the same mis-interpretations can occur (upgrades are disabled, not the policy, nor soakdays reports), but `disabled`  is definitely fine, probably better than `frozen`.

PR #3620, opened 2023-06-29
[rhidp] fix multiple IDPs with same username error

chassing:
This PR introduces a bugfix for the **Could not create user** cluster login error ([APPSRE-7869](https://issues.redhat.com/browse/APPSRE-7869)) for users where the GitHub and the RedHat usernames are the same.

This PR changes the `mapping_method` OIDC configuration from `claim` to `add`. 

This will change all new and existing OIDC IDPs but only for new GitHub IDPs (no update of the existing config).

**claim**
The default value. Provisions a user with the identity’s preferred user name. **Fails** if a user with that user name is already mapped to another identity.

**add**

Provisions a user with the identity’s preferred user name. If a user with that user name > already exists, the identity is mapped to the existing user, adding to any existing identity mappings for the user. Required when multiple identity providers are configured that identify the same set of users and map to the same user names. 

Ticket: [APPSRE-7872](https://issues.redhat.com/browse/APPSRE-7872)


geoberle (commented):
looks good. from the looks of it there is currently no way to define the mapping method for the label based configuration approach, right?

chassing:
Yes, currently, there is no way to specify it. Can be introduced later, if needed.

geoberle (approved):

PR #3622, opened 2023-06-29
[terraform-resources] more robust vault secrets update

chassing:
Make the update of secrets in Vault more robust. 

The `disable_keys` method could raise an error, and the Vault secret updates are lost. Especially if the integration takes longer, e.g., creating an MSK cluster, the cached bundle version is outdated, and `disable_keys` raises a 404 error.

Write the outputs to Vault as soon as possible (and retry it) to avoid such cases.

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -14,7 +14,10 @@ from typing import (
     cast,
 )
 
-from sretoolbox.utils import threaded
+from sretoolbox.utils import (
+    retry,
+    threaded,
+)
 
 import reconcile.openshift_base as ob
 from reconcile import queries
@@ -296,6 +299,7 @@ def cleanup_and_exit(
     sys.exit(status)
 
 
+@retry()
 def write_outputs_to_vault(
     vault_path: str, resource_specs: ExternalResourceSpecInventory
 ) -> None:
@@ -415,6 +419,15 @@ def run(
         if err:
             cleanup_and_exit(tf, err)
 
+        if defer:
+            defer(
+                disable_keys,
+                dry_run,
+                thread_pool_size,
+                disable_service_account_keys=True,
+                account_name=acc_name,
+            )
+
     # refresh output data after terraform apply
     tf.populate_terraform_output_secrets(
         resource_specs=resource_specs, init_rds_replica_source=True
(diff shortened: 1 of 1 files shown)


janboll (approved):

reconcile/terraform_resources.py
```diff
@@ -428,6 +432,11 @@ def run(
             dry_run, oc_map, ri, thread_pool_size, caller=acc_name
         )
 
+    # update secrets in vault immediately after applying them to the clusters
+    # so that nothing can raise an error before the secrets are updated
+    if actions and vault_output_path:
+        write_outputs_to_vault(vault_output_path, resource_specs)
```
maorfr, line 438: i would use `defer` here to avoid so that nothing can raise an error before disabling keys

hemslo: Here we only moved `write_outputs_to_vault` up, how about other steps after `disable_keys`, those cleanup steps can be skipped too.

chassing: As @hemslo said, it's about making `write_outputs_to_vault` more bulletproof. In the worst, worst, worst case, `disabled_keys` would run at the next `terraform-resources` reconciliation loop but it runs in `aws-iam-keys` anyways. It also doesn't depend on the current run, but `write_outputs_to_vault` does.

@hemslo, there isn't anything after `disabled_keys`

maorfr: the runs of aws-iam-keys is different if it runs from terraform-resources.

chassing: Ah, I see 👍 , but anyway, a missed `write_outputs_to_vault` is worse than a missed `disabled_keys`.

maorfr: if only we had something that will allow us not to miss anything, like `defer`.

hemslo: if `disabled_keys` throw error, then `cleanup_and_exit` won't run

hemslo: if we expect `disable_keys` to throw error, how about catch it and process it similar to `ri.has_error_registered`?

reconcile/terraform_resources.py
```diff
@@ -415,6 +419,15 @@ def run(
         if err:
             cleanup_and_exit(tf, err)
 
+        if defer:
+            defer(
```
chassing, line 423: @maorfr, what do you think?

maorfr: i think you removed `if not light and tf.should_apply:`

chassing: No, take a look at [line 417](https://github.com/app-sre/qontract-reconcile/blob/288ffab049924f1275455a7989b18e78b4348957/reconcile/terraform_resources.py#L417)

maorfr (approved):

PR #3624, opened 2023-06-30
Fluentd error log filtering for Google Chat

cubismod:
https://issues.redhat.com/browse/APPSRE-7881

We make use of Fluentd for shipping logs to Cloudwatch, Slack, and Google Chat (within FedRamp). Unfortunately, Google Chat isn't well suited for displaying tons of log messages and the UI performance becomes atrocious if we're shipping every single info message to gchat. So instead, let's just send non-INFO (warning, error) to gchat within FedRamp.

## What's going on here?
- A new environment variable `PREFIX_LOG_LEVEL`, is set to true which triggers QR to format log messages prefixed with their log level `INFO`, `ERROR`, etc.
- The fluentd config is updated for FedRamp such that [rewrite_tag_filter](https://docs.fluentd.org/output/rewrite_tag_filter) is used to rewrite error and warning log events to `gchat`
- A new `match` statement is added for `gchat`, meaning that these newly tagged events get sent to Google Chat
- The filtering is done in such a way that `INFO` level logs are still sent to AWS CloudWatch as expected

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -35,6 +35,7 @@ https://github.com/helm/helm/releases
 | logs                      | ship logs to providers                                                   | cloudwatch is enabled by default                                        |
 | logs.slack                | ship logs to slack                                                       | false                                                                   |
 | logs.googleChat           | ship logs to google chat                                                 | false                                                                   |
+| logs.errorsOnly           | filter shipped logs to only include errors (anything not INFO level)     | false                                                                   |
 | resources                 | CPU/Memory resource requests/limits                                      |                                                                         |
 | fluentdResources          | CPU/Memory resource requests/limits for Fluentd                          | {requests: {memory: 30Mi, cpu: 15m}, limits: {memory: 120Mi, cpu: 25m}} |
 | shards                    | number of shards to run integration with                                 | 1                                                                       |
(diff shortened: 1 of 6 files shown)


helm/README.md
```diff
@@ -34,7 +34,7 @@ https://github.com/helm/helm/releases
 | internalCertificates      | integration requires internal certificates to execute                    | false                                                                   |
 | logs                      | ship logs to providers                                                   | cloudwatch is enabled by default                                        |
 | logs.slack                | ship logs to slack                                                       | false                                                                   |
-| logs.googleChat           | ship logs to google chat                                                 | false                                                                   |
+| logs.googleChat           | ship any error logs to google chat (anything not INFO level)             | false                                                                   |
```
bkez322, line 37:
```suggestion
| logs.googleChat           | ship logs to google chat (note: if the 'PREFIX_LOG_LEVEL' environment variable is set to 'TRUE', then logs with the prefix '[INFO]' are excluded from shipment)             | false                                                                   |
```

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -175,34 +175,47 @@ objects:
...
                 log_stream_name {{ $integration.name }}
                 auto_create_stream true
               </store>
+              {{- if $logs.googleChat }}
+              # only ship errors and warnings to Google Chat
+              <store>
+                @type rewrite_tag_filter
+                <rule>
+                  key message
+                  pattern /\[INFO\]/
```
bkez322, line 190:
```suggestion
                  pattern /^\[INFO\]/
```

^ Makes sure `[INFO]` is at the exact beginning of the string.

helm/README.md
bkez322: Actually I'm stupid and cannot read code. Disregard this suggestion.

bkez322 (approved):
I made a minor suggestion for updating the regex used. Otherwise this LGTM!

cubismod:
Gotta fix the tests now

PR #3627, opened 2023-07-03
[gitlab-housekeeping] add auto_merge as a label to merge metric

maorfr:
following up on #3612, we want to distinct non self-serviceable MRs from such that are submitted and merged by the bot (`bot/automerge`) in dashboard.

this PR adds an `auto_merge` label to the metric, so we can exclude such MRs.


reconcile/rhidp/ocm_oidc_idp/base.py
```diff
@@ -34,6 +34,7 @@ def run(
     vault_input_path: str,
     dry_run: bool,
 ) -> None:
+    print('hello!')
```
patjlm, line 37: 👋

geoberle (approved):

PR #3521, opened 2023-05-22
[openshift saas deploy] multiple publishers

chassing:
2nd try (#3473)!

The deployment information (promotion data) is stored in an S3 bucket. Make the S3 key unique to avoid overwrites if multiple publishers are using the same channel name.

promotions/{channel}/{commit_sha} -> /deployments/{saas_target_uid}/{commit_sha}

saas_target_uid is an unique blake2s hash composed of {saas_name}/{resource_template_name}/{target_name|"default"}/{cluster_name}/namespace_name}

Ticket: [APPSRE-7414](https://issues.redhat.com/browse/APPSRE-7414)

PR #3628, opened 2023-07-03
APPSRE-7563 cleanup/remove old auto-promoter

fishi0x01:
**What**
Remove old auto-promoter and related logic.

**Why**
Auto-promotions are being handled since >2 months by SAPM. We are planning further changes to support multi-publisher channels in saas files. These upcoming changes will introduce changes to the saas-file model (see https://github.com/app-sre/qontract-reconcile/pull/3521) and removing dead-weight code will make this effort easier. Since SAPM is fully reconcile based, we should not require a SQS fallback anylonger and maintaining both in parallel seems redundant.


fishi0x01:
cc @maorfr I believe this is fine to cleanup. Please let me know if Im forgetting something important here.

maorfr (approved):
i believe so as well

PR #3630, opened 2023-07-03
[jira-permissions-validator] introduce integration

maorfr:
part of https://issues.redhat.com/browse/APPSRE-7890

this integration validates that it can create tickets on all jira boards defined in app-interface. since the integration is using the same credentials as our jiralert instance, this actually validates that jiralert can create tickets.


patjlm (approved):

reconcile/jira_permissions_validator.py
```diff
@@ -0,0 +1,21 @@
...
+
+from reconcile import queries
+from reconcile.status import ExitCodes
+from reconcile.utils.jira_client import JiraClient
+
+QONTRACT_INTEGRATION = "jira-permissions-validator"
+
+
+def run(dry_run: bool) -> None:
+    settings = queries.get_app_interface_settings()
```
fishi0x01, line 12: Since this is a new integration, can we use typed queries instead of queries.py?

reconcile/jira_permissions_validator.py
```diff
@@ -0,0 +1,21 @@
...
+    error = False
+    for jira_board in queries.get_jira_boards(with_slack=False):
```
fishi0x01, line 14: Same: can we use typed queries? (Though here i understand why youd want to reuse some)

maorfr:
reconcile/utils/jira_client.py is not ready for typed queries unfortunately

PR #3631, opened 2023-07-03
allow typed queries for jira client

fishi0x01:
Make `JiraClient` compatible with typed queries. This is fully backwards compatible.

**Test**

Run current integration that relies on backwards compatibility.

```
qd profile run prod jira-watcher
```


chassing (approved):
❤️

reconcile/utils/jira_client.py
```diff
@@ -17,13 +18,25 @@
...
+        jira_board: Mapping[str, Any],
+        settings: Optional[Mapping] = None,
+        jira_watcher_settings: Optional[JiraWatcherSettings] = None,
+        secret_reader: Optional[SecretReader] = None,
     ):
-        self.secret_reader = SecretReader(settings=settings)
+        if not (secret_reader or settings):
+            raise RuntimeError("JiraClient needs secret_reader or settings.")
+        if not secret_reader:
+            self.secret_reader = SecretReader(settings=settings)
```
hemslo, line 39: Do you mean by assign `secret_reader` to `self.secret_reader` if passed else new.

```python
self.secret_reader = secret_reader if secret_reader is not None else SecretReader(settings=settings)
```

reconcile/utils/jira_client.py
```diff
@@ -35,6 +48,10 @@ def __init__(
             read_timeout = settings["jiraWatcher"]["readTimeout"]
             connect_timeout = settings["jiraWatcher"]["connectTimeout"]
 
+        if jira_watcher_settings:
```
hemslo, line 51: It seems we need a priority assign here, `jira_watcher_settings -> settings -> default`, but the order is reversed in mutation style. To make it clearer, can do it in an extracted function:

```
DEFAULT_READ_TIMEOUT = 60
DEFAULT_CONNECT_TIMEOUT = 60

def get_jira_timeout(jira_watcher_settings, settings):
  if jira_watcher_settings:
    return (jira_watcher_settings.read_timeout, jira_watcher_settings.connect_timeout)
  if settings and settings["jiraWatcher"]:
    return (settings["jiraWatcher"]["readTimeout"], settings["jiraWatcher"]["connectTimeout"])
  return (DEFAULT_READ_TIMEOUT, DEFAULT_CONNECT_TIMEOUT)
```

hemslo (commented):
We may need to cover new changes in `JiraClient` as new test cases.

reconcile/utils/jira_client.py
fishi0x01: This was actually a bug because I forgot to set the secret_reader if its given as parameter. Fixed now. Good catch!

reconcile/utils/jira_client.py
fishi0x01: The settings are actually mutual exclusive. I added a RuntimeError now to highlight this.

fishi0x01:
@hemslo haha I was lazy with the test. You got me :)
Added some tests now for the init function.

reconcile/utils/jira_client.py
hemslo: But that check is for `secret_reader and settings`, this timeout is from `jira_watcher_settings`

fishi0x01: Good point. Added another test case :+1:

hemslo: Why `jira_api` added as a new optional parameter to `JiraClient` and used only in tests? And we shouldn't add private attributes just for tests, should test public behaviour instead of implementation details.

To see why test private attributes is bad, you can remove `self.jira = (...` in current code and test `test_create_defaults` will still pass.


One test example:

```
Given jira_board and secret_reader
When create JiraClient with them
Then it should return a default JiraClient object with attribute jira
    and jira is created by JIRA(expected_server, expected_token_auth, (expected_read_timeout, expected_connect_timeout))
```

=>

```python
def test_create_defaults(
    secret_reader: Mock,
    jira_board: dict,
    mocker: MockerFixture,
) -> None:
    mocked_jira = mocker.patch("reconcile.utils.jira_client.JIRA")
    expected_server = jira_board["server"]["serverUrl"]
    expected_token_auth = secret_reader.read.return_value
    expected_read_timeout = 60
    expected_connect_timeout = 60

    jira_client = JiraClient(jira_board=jira_board, secret_reader=secret_reader)

    assert jira_client.jira == mocked_jira.return_value
    mocked_jira.assert_called_once_with(
        expected_server,
        token_auth=expected_token_auth,
        timeout=(expected_read_timeout, expected_connect_timeout),
    )

```

fishi0x01: Im not a big fan of mocker.patch stuff. Thats why I added inversion control for the JiraApi. This will undergo more changes as we deprecate the old settings - it is just a first step.

hemslo: https://github.com/app-sre/qontract-reconcile/pull/3631/files#diff-79ed764eef7f4eb6ccd177b3ba8236b3ffe7263f233e4b1035e185241837493dR72-R76 is not covered without mocker.patch, it's essential to know what's being tested and what should be mocked, never mock the thing being tested, in this instance constructing JIRA instance is the core logic of init, if the result is mocked as `create_autospec(spec=JIRA)`, then it's same as no test. `mocker.patch` is unavoidable as long as there are free variables (`JIRA` here).

A key mindset is implementation is driven by test, and test is driven by behaviour, tests should bring value instead of burden.

hemslo: If you want to merge it, I'd suggest to remove those private attributes and tests too.

fishi0x01: > never mock the thing being tested

Most of the discussion here and hence test cases are simply about whether the input should raise or not. I think that is achieved regardless of the JiraApi mock, i.e., I would like to keep the raise tests at least.

I dont mind removing the JiraApi mock + test case that relies on it - will send a patch soon :+1:

hemslo (approved):
LGTM! Thank you for tests improvements!

reconcile/utils/jira_client.py
fishi0x01: @hemslo Thinking about this Im still not 100% happy. I think you are right and we should stay pragmatic here and use the patching method. I dont like python's monkey patching voodoo, but for this case it is reasonable to apply. 

At least future tests for the other functions of JiraClient will not need mock patching.

I have re-added the 2 removed test cases with mock patching as you suggested :+1:

fishi0x01: One alternative might be changing the scope of that function or better create another one that is responsible of fetching/assembling the parameters w/o creating the client itself. E.g.,

```python
class JiraClient:
    def __init__(self, params):
        self.jira = JIRA(params)

    @staticmethod
    def get_params(settings, secret_reader) -> params:
        ...
        return params


JiraClient(params=JiraClient.get_params(...))
```

I.e., we do not need to test the `__init__` function at all since the core logic was moved and is now about params instead of JIRA. Testing `get_params` would be straight forward w/o mock.patching. Then again, all that stuff to avoid mock.patch might not be worth it here.

fishi0x01: > At least future tests for the other functions of JiraClient will not need mock patching.

I take that back. We will need to inject JIRA somehow or else we will need to mock.patch for every other function's test too :disappointed:

fishi0x01: I've refactored JiraClient for dependency injection now. There is a glue function `create_jira_client` (we use a similar pattern with `create_secret_reader`). If we want to test that glue function, then we can use mock.patch. However, any other function within JiraClient itself can be tested with clean dep mocks.

Wdyt?

hemslo: Oh, you want to switch to DI style, that's good! Then we will have factory method to create `JiraClient`, and simple init method for `JiraClient`.

Normally factory method need to be tested via mocker patch if it's creating some dependencies.

Init doesn't need too much testing as it's just simple assign, and some backwards-compatibility like you already did.

I feel the new `parameters` as a public method with tests is not that clean, as a user of `JiraClient`, I'll wonder what's that method mean, it's actually only used in factory method, should be kept as private. The logic is not that complex, since factory method requires mocker patch anyway, it can be tested in one pass.


```python
    @staticmethod
    def create(
        jira_board: Mapping[str, Any],
        jira_watcher_settings: Optional[JiraWatcherSettings],
    ):
        server = jira_board["server"]["serverUrl"]
        secret = jira_board["server"]["token"]
        token = secret_reader.read(secret)
        timeout = (
            (jira_watcher_settings.read_timeout, jira_watcher_settings.connect_timeout)
            if jira_watcher_settings
            else (JiraClient.DEFAULT_READ_TIMEOUT, JiraClient.DEFAULT_CONNECT_TIMEOUT)
        )
        jira_api = JIRA(
            server,
            token_auth=token,
            timeout=timeout,
        )

        return JiraClient(
            project=jira_board["name"],
            jira_api=jira_api,
        )
```

```python
def test_create_with_jira_watcher_settings(
    mocker: MockerFixture,
    secret_reader: Mock,
    jira_board: dict,
):
    mocked_jira = mocker.patch("reconcile.utils.jira_client.JIRA", autospec=True)
    jira_watcher_settings = JiraWatcherSettingsV1(readTimeout=42, connectTimeout=43)

    jira_client = JiraClient.create(
        jira_board=jira_board,
        secret_reader=secret_reader,
        jira_watcher_settings=jira_watcher_settings,
    )

    assert jira_client.project == jira_board["name"]
    assert jira_client.jira == mocked_jira.return_value
    expected_server = jira_board["server"]["serverUrl"]
    expected_token = secret_reader.read.return_value
    expected_connect_timeout = jira_watcher_settings.connect_timeout
    expected_read_timeout = jira_watcher_settings.read_timeout
    mocked_jira.assert_called_once_with(
        expected_server,
        token_auth=expected_token,
        timeout=(expected_read_timeout, expected_connect_timeout),
    )
```

The bad of `mocker.patch` is it depends on how you import things in implementation, that's the most tricky part, but Python didn't come with any other built-in way to bypass it, even with some DI frameworks like [this](https://python-dependency-injector.ets-labs.org/introduction/di_in_python.html), test DI container construction requires patch. So either use it for better coverage, or don't use it and leave code untested. I'd prefer to use it when necessary to cover all lines in implementation at least once.

If you feel the `create` is too complex, can extract the creation of `JIRA`, or it's dependencies into a new place, if it's requiring seperate tests apart from `create`, then worth creating a new `JIRA` related util class to keep `JiraClient` clean.

fishi0x01: I like the `create()` method. Will add it.

fishi0x01: @hemslo added `create()` method and mock.patch tests as you suggested. Further tests on other functions for that class will not require mock.patch any longer. Does that look good to you?

hemslo: Nice! You are planning to use new `create` in another pr, right?

fishi0x01: Yeah will open a small follow-up PR for this https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/jira_permissions_validator.py

fishi0x01: Probably also for [jira-watcher](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/jira_watcher.py) - I think then we can also remove the deprecated stuff :)

PR #3634, opened 2023-07-03
Consider "needs-rebase" a hold label

cubismod:
If an App-SRE member adds the "needs-rebase" label to an MR, then that should be considered a holding tag as it requires actions on the tenant's side. This ensures that these MRs don't block the merge queue.


kwilczynski (approved):
@cubismod, looks good!

Good idea! Thank you for adding this! :+1:

PR #3602, opened 2023-06-22
Log warning for empty run in terraform_tgw_attachments

hemslo:
Log warning for empty run in terraform_tgw_attachments, so we can disable the integration to save resource. [APPSRE-7624](https://issues.redhat.com/browse/APPSRE-7624)

### <samp>🤖 Generated by Copilot at efa6673</samp>

This pull request improves the test coverage and performance of the `reconcile/terraform_tgw_attachments.py` integration. It adds more tests and mocks to the `reconcile/test/test_terraform_tgw_attachments.py` file and adds a check to skip the integration if there are no participating AWS accounts.

### <samp>🤖 Generated by Copilot at efa6673</samp>

*  Add early return and warning log when no AWS accounts are participating in the integration ([link](https://github.com/app-sre/qontract-reconcile/pull/3602/files?diff=unified&w=0#diff-b2a48a56ab3feb0903c9b33c5eec954718a01524b0dd49a99d882637bb2cc226R372-R377))
*  Add mock for logging module and verify expected log messages in different scenarios ([link](https://github.com/app-sre/qontract-reconcile/pull/3602/files?diff=unified&w=0#diff-b5ab464582f4a599b9e05939387fe44570d983cb1a58d777a7af990bfb3124d4R564-R566))
*  Refactor test_dry_run function into test_empty_run and test_dry_run, and use parameters to customize mocked data for different test cases ([link](https://github.com/app-sre/qontract-reconcile/pull/3602/files?diff=unified&w=0#diff-b5ab464582f4a599b9e05939387fe44570d983cb1a58d777a7af990bfb3124d4L574-R618), [link](https://github.com/app-sre/qontract-reconcile/pull/3602/files?diff=unified&w=0#diff-b5ab464582f4a599b9e05939387fe44570d983cb1a58d777a7af990bfb3124d4L598-R646))
*  Assert no AWS API, OCM, or Terraform calls are made when cluster only has VPC connection ([link](https://github.com/app-sre/qontract-reconcile/pull/3602/files?diff=unified&w=0#diff-b5ab464582f4a599b9e05939387fe44570d983cb1a58d777a7af990bfb3124d4L737-R787))
[...]

PR #3635, opened 2023-07-04
Log warning for terraform_resources empty run

hemslo:
Log warning for terraform_resources empty run, so we can disable the integration to save resource.

Similar to https://github.com/app-sre/qontract-reconcile/pull/3602

[APPSRE-7624](https://issues.redhat.com/browse/APPSRE-7624)

### <samp>🤖 Generated by Copilot at a986ed1</samp>

Add logging and testing for empty terraform namespaces. This pull request adds a warning log message to `reconcile/terraform_resources.py` and a corresponding test case to `reconcile/test/test_terraform_resources.py` to handle the case when no terraform namespaces are found for the given account names.

### <samp>🤖 Generated by Copilot at a986ed1</samp>

* Add a warning log message when no terraform namespaces are found ([link](https://github.com/app-sre/qontract-reconcile/pull/3635/files?diff=unified&w=0#diff-0e9f20c0944ab8088807c726d964c2df24b5c33b873c4c4ef1a4a78c5fe86517R223-R227))
* Import pytest_mock fixture for mocking objects and functions in test cases ([link](https://github.com/app-sre/qontract-reconcile/pull/3635/files?diff=unified&w=0#diff-aa3407e2884577fb05f6ed1925376a267ff8655228672c67c899a3bafe4c7497R9))

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -220,6 +220,11 @@ def setup(
     # build a resource inventory for all the kube secrets managed by the
     # app-interface managed terraform resources
     tf_namespaces = get_tf_namespaces(account_names)
+    if not tf_namespaces:
+        logging.warning(
+            "No terraform namespaces found, consider disabling this integration, account names: "
+            f"{', '.join(account_names)}"
+        )
     ri, oc_map = fetch_current_state(
         dry_run, tf_namespaces, thread_pool_size, internal, use_jump_host, account_names
     )
--- a/reconcile/test/test_terraform_resources.py
+++ b/reconcile/test/test_terraform_resources.py
@@ -6,6 +6,7 @@ from collections.abc import (
 from typing import Any
 
 import pytest
+from pytest_mock import MockerFixture
 
 import reconcile.terraform_resources as integ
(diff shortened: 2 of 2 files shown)


reconcile/terraform_resources.py
```diff
@@ -220,6 +220,11 @@ def setup(
     # build a resource inventory for all the kube secrets managed by the
     # app-interface managed terraform resources
     tf_namespaces = get_tf_namespaces(account_names)
+    if not tf_namespaces:
+        logging.warning(
+            "No terraform namespaces found, consider disabling this integration, account names: "
+            f"{', '.join(account_names)}"
+        )
```
patjlm, line 227: should we also return/exit in this case?

hemslo: It's hard to do a return here as this is in an extracted setup function, and based on what I know, we'd like to remove usages of `sys.exit`.

hemslo: Normally this warning will be addressed soon after deploy and those accounts will be disabled, so even we don't stop here, it won't run anything in the end.

patjlm (approved):

PR #3638, opened 2023-07-04
Clusters-network for CCS/ROSA clusters

patjlm:
Include CCS/Rosa clusters in `qontract-cli clusters-network`

https://issues.redhat.com/browse/APPSRE-7848

Those clusters don't rely on OCM infrastructure access to retrieve their network info. Instead, we'll use the direct access to their account (no `assume_role`).

### DIFF
--- a/reconcile/utils/aws_api.py
+++ b/reconcile/utils/aws_api.py
@@ -846,6 +846,8 @@ class AWSApi:  # pylint: disable=too-many-public-methods
         self, account_name: str, assume_role: str, assume_region: str, client_type="ec2"
     ) -> EC2Client:
         session = self.get_session(account_name)
+        if assume_role is None:
+            return self.get_session_client(session, client_type)
         sts = self.get_session_client(session, "sts")
         assumed_session = self._get_assume_role_session(
             sts, account_name, assume_role, assume_region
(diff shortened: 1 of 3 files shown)


reconcile/utils/aws_api.py
```diff
@@ -846,6 +846,8 @@ def _get_assumed_role_client(
         self, account_name: str, assume_role: str, assume_region: str, client_type="ec2"
     ) -> EC2Client:
         session = self.get_session(account_name)
+        if assume_role is None:
```
patjlm, line 849: this would also be triggered in the case of a missing `awsInfrastructureManagementAccess` section. I am wondering if we could use an empty string instead of `None` to show that the config is valid, but there are no role to assume.
WDYT @maorfr ?

patjlm: Proposed using an empty string in 1eab41f

tools/qontract_cli.py
```diff
@@ -1083,41 +1085,6 @@ def ocm_aws_infrastructure_access_switch_role_links(ctx):
         print_output(ctx.obj["options"], by_user[user], columns)
 
 
-@get.command()
-@click.pass_context
-def clusters_egress_ips(ctx):
```
patjlm, line 1073: @TGPSKI @bkez322 , can you confirm you do not use that qontract-cli command on your side?

The same info (and other) is available in `clusters-network`

maorfr (approved):

tools/qontract_cli.py
patjlm: I'll go ahead and merge. We can always reinstate that if needed (adding the vpcid lookup that is now needed), but I believe  you'd get covered with `clusters-network`

maorfr: i don't think there is an app-interface-output is fedramp (yet?)

PR #3639, opened 2023-07-04
APPSRE-7563 start publishing promotions_v2

fishi0x01:
We need a way to support multiple publishers for a single channel.

We already had multiple attempts on this (https://github.com/app-sre/qontract-reconcile/pull/3521). For that reason we chunk this effort into smaller backwards-compatible PRs. You can regard this PR as a subset of https://github.com/app-sre/qontract-reconcile/pull/3521

This PR introduces a new additional publisher path for promotions in the s3 state: `f"promotions_v2/{channel}/{target_uid}/{sha}"`.

The `target_uid` is a uid based on a hash from unique combination of `(saas_file_name, resource_template_name, namespace, cluster)`. That uid allows us to distinguish between multiple publishers for the same channel.

This PR only handles the publishing aspect and is fully backwards compatible. Follow-up PRs will deal with the consumption from the new `promotions_v2` path.

This PR will initially add overhead as we publish promotions to 2 paths now. Future PRs will remove that overhead as we transition towards `promotions_v2`.

**Test**

```
qd profile run prod openshift-saas-deploy-trigger-moving-commits
```

```
# run for saas file uhc-integration after changing a ref locally
qd profile run prod openshift-saas-deploy
```

Note, that testing happened in dry-run. This is very limited, as the publish promotion step only happens in non-dry-run. However, we do not really have a proper way to test this.

### DIFF
--- a/reconcile/test/test_saasherder.py
+++ b/reconcile/test/test_saasherder.py
@@ -28,10 +28,11 @@ from reconcile.gql_definitions.common.saas_files import (
     SaasResourceTemplateTargetV2_SaasSecretParametersV1,
     SaasResourceTemplateV2,
 )
+from reconcile.typed_queries.saas_files import SaasFile
 from reconcile.utils.jjb_client import JJB
 from reconcile.utils.openshift_resource import ResourceInventory
 from reconcile.utils.saasherder import SaasHerder
-from reconcile.utils.saasherder.interfaces import SaasFile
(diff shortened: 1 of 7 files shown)


fishi0x01:
@chassing I think it makes most sense if you also have a look at this PR :bow:

chassing (approved):
First tiny step. LGTM! Thank you.

PR #3644, opened 2023-07-05
[terraform-resources] availability zone validation

maorfr:
with this PR, whenever we create/update an elasticache instance, we will validate the subnet group availability zones.

example for a fix that would have been detected: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/73198

terraform error:
```
InvalidVPCNetworkStateFault: No subnet is present in the cache subnet group for availability zone
```

### DIFF
--- a/reconcile/utils/aws_api.py
+++ b/reconcile/utils/aws_api.py
@@ -41,6 +41,7 @@ if TYPE_CHECKING:
         TransitGatewayVpcAttachmentTypeDef,
         VpcTypeDef,
     )
+    from mypy_boto3_elasticache import ElastiCacheClient
     from mypy_boto3_iam import IAMClient
     from mypy_boto3_iam.type_defs import AccessKeyMetadataTypeDef
     from mypy_boto3_rds import RDSClient
@@ -79,6 +80,8 @@ else:
         LaunchPermissionModificationsTypeDef
     ) = (
         FilterTypeDef
+    ) = (
+        ElastiCacheClient
     ) = (
         Route53Client
     ) = (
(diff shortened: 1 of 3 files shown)


pablomh:
Very nice and useful :)

reconcile/utils/terraform_client.py
```diff
@@ -741,6 +748,19 @@ def validate_db_upgrade(
...
+        self, account_name: str, resource_name: str, resource_change: Mapping[str, Any]
+    ):
+        if self._aws_api is not None:
+            resource_azs = resource_change["after"]["availability_zones"]
+            available_azs = self._aws_api.get_elasticache_availability_zones(
+                account_name
+            )
+            if not all(az in available_azs for az in resource_azs):
+                raise ValueError(
+                    f"elasticache availability zones invalid for {resource_name}, should be within {available_azs}"
```
kwilczynski, line 761:
```suggestion
                    f"ElastiCache availability zones are invalid for {resource_name}, should be within {available_azs}"
```

reconcile/utils/aws_api.py
```diff
@@ -941,6 +950,16 @@ def get_cluster_nat_gateways_egress_ips(self, account):
 
         return egress_ips
 
+    def get_elasticache_availability_zones(self, account_name: str) -> set[str]:
+        elasticache = self._account_elasticache_client(account_name)
+        subnet_groups = elasticache.describe_cache_subnet_groups()["CacheSubnetGroups"]
```
kwilczynski, line 955: I wouldn't trust AWS API to guarantee this key will always exist. Terraform learned the hard way. :smile:

hemslo: Note `describe_cache_subnet_groups` response is paginated https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elasticache/client/describe_cache_subnet_groups.html, to ensure get all azs from subnets, we need to implement fetch all using `Marker`.

kwilczynski: @hemslo, good point! Even though the default records count is set to 100, which probably would cover us for a while, we should do it to future-proof things, indeed. :+1:

PR #3645, opened 2023-07-05
Run ocm.update_cluster in dry-run

esron:
Ticket: [APPSRE-2938](https://issues.redhat.com/browse/APPSRE-2938)


maorfr (changes requested):
the idea would be to call the OCM update cluster endpoint (PATCH /clusters/{id}) with the dryRun query parameter.

OliviaHY:
LGTM. @maorfr Could you please take another look?

maorfr (approved):
beautiful. LGTM

PR #3647, opened 2023-07-06
clusters-network for CCS/ROSA clusters #2

patjlm:
https://issues.redhat.com/browse/APPSRE-7848

New try for #3638, reverted by #3646, adding a fix in the last commit.

Include CCS/Rosa clusters in qontract-cli clusters-network.

Those clusters don't rely on OCM infrastructure access to retrieve their network info. Instead, we'll use the direct access to their account (no assume_role).


geoberle (approved):
lgtm

PR #3651, opened 2023-07-06
APPSRE-7307 promotion state get_v2

fishi0x01:
`promotion_state` now tries to fetch data from `promotions_v2` path. This is fully backwards-compatible, as saasherder and SAPM are still not passing any `target_uid` to the get call.

This is a follow-up on https://github.com/app-sre/qontract-reconcile/pull/3639

A follow-up PR will adjust saasherder to pass the target_uid and fetch from v2.
Another follow-up PR will adjust SAPM to fetch from v2.
A third final follow-up PR will adjust SAPMs logic to calculate refs/hashes.

### DIFF
--- a/reconcile/utils/promotion_state.py
+++ b/reconcile/utils/promotion_state.py
@@ -39,6 +39,9 @@ class PromotionState:
         self._state = state
         self._commits_by_channel: dict[str, set[str]] = defaultdict(set)
 
+    def _target_key(self, channel: str, target_uid: str) -> str:
+        return f"{channel}/{target_uid}"
+
     def cache_commit_shas_from_s3(self) -> None:
         """
         Caching commit shas locally - this is used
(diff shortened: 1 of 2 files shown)


fishi0x01:
cc @chassing does that look ok?

reconcile/utils/promotion_state.py
```diff
@@ -47,24 +50,56 @@ def cache_commit_shas_from_s3(self) -> None:
...
         all_keys = self._state.ls()
         for commit in all_keys:
+            # BACKWARDS-COMPAT BLOCK
             # Format: /promotions/{channel}/{commit-sha}
-            if not commit.startswith("/promotions/"):
+            if commit.startswith("/promotions/"):
+                # Keep this for backwards-compatibility with v1 promotions
+                # We wait a couple of months to be sure all pipelines have
+                # been triggered since.
+                # This can be deleted after 30.09.2023
```
chassing, line 59: A birthday present to me 🎁

chassing (approved):
LGTM!

PR #3655, opened 2023-07-07
[AUS] refactor all AUS flavors

geoberle:
## Purpose

the goal of this refactor was to concentrate most of the required queries to determine the state into the `get_ocm_env_upgrade_specs` method of the integration class. it was hard to understand what data was loaded where and the same data was loaded multiple times due to that. the heavy use of the `OCM/OCMMap` classes in various places of the code were witness of that.

the refactoring will enable new AUS features to be a lot easier to be integrated. all kinds of nap-kin math about OCM API usage will be easier to conduct.

## What happened

* extracted version-gate, version-gate agreement, policy mgmt functions from the `OCM` class into reconcile.utils.ocm since they are not organization related. this allows to have lightweight context during these operations and simplifies testing
* ClusterUpgradeSpec is used throughout the code as desired state instead of the Configured* classes. this allows us to keep way more context that is not lost during model translation and hence we don't need to query missing data again via `OCM/OCMMap`
* translated all related `unittest` tests to `pytest`
* reuse the `get_ocm_env_upgrade_specs` functionality in `qontract-cli` to get rid of custom state compilation (extensive testing was done here. the resulting reports are still the same as previously to the letter)

## Tests conducted

Besides heavily reviewing our unittests, additonal manual testing was conducted to ensure correctness

To test the correctness of the construction of the desired state, the `qontract-cli` reports were generated (`ocm-upgrade-scheduler`, `ocm-upgrade-scheduler-org`, `advanced-upgrade-service`, `ocm-addons-upgrade-scheduler-org`) and compared to the existing reports in `app-interface-output`. They were identical to the letter.

Each condition was dry-runned in debug mode to see if the right decision is taken per-condition. This was supported with local app-interface modifications as well, e.g. to see a schedule behaving correctly, soak-days to be sufficient, soak-day inheritance to kick in.


reconcile/rhidp/ocm_oidc_idp/standalone.py
```diff
@@ -99,6 +99,7 @@ def get_clusters(
             )
             for ocm_clusters in clusters_by_org.values()
             for c in ocm_clusters
+            if c.ocm_cluster.console
```
geoberle, line 102: @chassing i needed to change the `console` field in the `OCMCluster` to `Optional` because a HCP reports as ready while the `console` might still be missing for a while.

janboll (commented):
This looks good to me, though I focused my review on understanding the general direction of the change. 
To increase assurance into this change, can you please document the test cases you ran locally to ensure the code still works as expected?

reconcile/rhidp/ocm_oidc_idp/standalone.py
chassing: It is ok for new clusters, but hopefully, we don't see it for existing ones!

reconcile/aus/base.py
```diff
@@ -207,15 +216,13 @@ def expose_org_upgrade_spec_metrics(
 class GateAgreement(BaseModel):
     id: str
 
-    def create(self, ocm: OCM, cluster_name: str) -> None:
-        logging.info(
-            f"create agreement for gate {self.id} on cluster {cluster_name} in OCM org {ocm.name}"
-        )
-        agreement = ocm.create_version_agreement(self.id, cluster_name)
+    def create(self, ocm_api: OCMBaseClient, cluster_id: str) -> None:
+        logging.info(f"create agreement for gate {self.id} on cluster {cluster_id}")
```
patjlm, line 220: I am not sure this log will be very easy to parse. Cluster names are easier than id for that matter

reconcile/aus/base.py
```diff
@@ -615,112 +583,118 @@ def version_conditions_met(
...
-        gate["id"]
-        for gate in ocm.get_version_gates(version_prefix)
-        if gate["id"] not in agreements and semver_cluster.match(f"<{version_prefix}.0")
+    semver_cluster = parse_semver(f"{cluster.version.raw_id}")
+
+    relevant_gates = [
+        g
+        for g in gates
+        if g.version_raw_id_prefix == version_prefix
+        and g.sts_only == cluster.is_sts()
```
patjlm, line 610: ❤️

reconcile/queries.py
```diff
@@ -787,6 +787,7 @@ def get_jumphosts(hostname: Optional[str] = None) -> JumphostsQueryData:
         soakDays
         mutexes
         sector
+        blockedVersions
```
patjlm, line 790: this (and equiv) depends on https://github.com/app-sre/qontract-schemas/pull/479

reconcile/openshift_upgrade_watcher.py
```diff
@@ -84,7 +85,8 @@ def _get_start_hypershift(
     ocm_map: OCMMap, cluster_name: str
 ) -> tuple[Optional[str], Optional[str]]:
     ocm_map_cluster = ocm_map.get(cluster_name)
-    schedules = ocm_map_cluster.get_control_plan_upgrade_policies(cluster_name)
+    cluster_id = ocm_map_cluster.cluster_ids[cluster_name]
+    schedules = get_control_plan_upgrade_policies(ocm_map_cluster._ocm_client, cluster_id)
```
patjlm: nit: typo notice `get_control_plan_upgrade_policies` -> `get_control_plane_upgrade_policies`

reconcile/aus/base.py
patjlm: hmm i am actually not sure if we should only retrieve `sts_only` gates for STS clusters, or if all gates should be checked int that case

reconcile/aus/base.py
geoberle: i will add the cluster_name but the ID gets more important when we act in lots of orgs where the cluster name is not unique anymore

reconcile/queries.py
geoberle: forgot to remove that. snuck in from another PR. thx

reconcile/aus/base.py
geoberle: [66733c0](https://github.com/app-sre/qontract-reconcile/pull/3655/commits/66733c087c4661b31d324c4b563602c7e3315a6d)

patjlm: maybe add the org id (or name once we can ensure we have a meaningful one..)

reconcile/aus/base.py
patjlm: that's only a guess. I think `only_sts` refers to the gate, meaning that gate applies only to STS clusters. But other gates may apply to all, including STS. We should confirm with OCM folks I think.

reconcile/openshift_upgrade_watcher.py
geoberle: [66f9ed7](https://github.com/app-sre/qontract-reconcile/pull/3655/commits/66f9ed74c3254839f7d54f21d59649de087afae5)

reconcile/aus/base.py
geoberle: your guess is right.
given the bigger risk of version gates for STS clusters, i opened a dedicated ticket to discuss a potential change for the version gate agreement flow - https://issues.redhat.com/browse/APPSRE-7949

geoberle: with [9c865a5](https://github.com/app-sre/qontract-reconcile/pull/3655/commits/9c865a5fee29d5c782b95ab6a8447b4d93223d22) sts-only gates are ignored and not approved. this is compatible with how it worked before the refactor.

but if we would start to do y-stream upgrades that usually involve an sts-version gate, scheduling upgrades for sts-clusters will fail. this is where https://issues.redhat.com/browse/APPSRE-7949 needs to be implemented to handle it safely

reconcile/aus/base.py
```diff
@@ -305,26 +313,25 @@ def summarize(self, ocm_org_name: str) -> str:
...
+        create_control_plane_upgrade_policy(ocm_api, self.cluster.id, policy)
 
-    def delete(self, ocm: OCM) -> None:
-        item = {
-            "id": self.id,
-        }
-        ocm.delete_control_plane_upgrade_policy(self.cluster, item)
+    def delete(self, ocm_api: OCMBaseClient) -> None:
+        if not self.id:
+            return
```
janboll, line 328: isn't this a bug condition? Shouldn't this raise some alert? Like warn-level logging

janboll: Applies for other deletes as well

janboll (approved):
LGTM, as far as I can comprehend :-)

reconcile/aus/base.py
geoberle: ack. i'm raising an exception now because a call to delete can not happen for an unsaved policy

PR #3653, opened 2023-07-06
APPSRE-7307 multi-publisher support for SAPM

fishi0x01:
Introduce multi-publisher support for moving refs.

This is a follow-up on a range of PRs for this effort:
- https://github.com/app-sre/qontract-reconcile/pull/3651
- https://github.com/app-sre/qontract-reconcile/pull/3650
- https://github.com/app-sre/qontract-reconcile/pull/3639

Note, this does not support target-config-hashes yet. They require introduction of publisher_uid into the hash object. Target-config-hashes are subject to further discussion https://issues.redhat.com/browse/APPSRE-7516

Currently only a small set of saas files rely on target-config-hash (saas files dedicated to tests), i.e., this PR will provide multi-publisher support for majority of saas files.

PR #3656, opened 2023-07-07
APPSRE-7307 remove publish promotion_v1

fishi0x01:
Our promotions do not rely on the v1 path any longer:

- https://github.com/app-sre/qontract-reconcile/pull/3650
- https://github.com/app-sre/qontract-reconcile/pull/3653

This PR stops publishing promotions to our v1 path.

### DIFF
--- a/reconcile/test/utils/test_promotion_state.py
+++ b/reconcile/test/utils/test_promotion_state.py
@@ -116,9 +116,6 @@ def test_publish_info(s3_state_builder: Callable[[Mapping], State]):
         target_uid="uid",
         data=promotion_info,
     )
-    deployment_state._state.add.assert_any_call(  # type: ignore[attr-defined]
-        "promotions/channel/sha", promotion_info.dict(), True
-    )
-    deployment_state._state.add.assert_any_call(  # type: ignore[attr-defined]
+    deployment_state._state.add.assert_called_once_with(  # type: ignore[attr-defined]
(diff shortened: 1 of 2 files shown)


patjlm:
Have all app-interface publish been through the v2 path? Even if there was no rollout since v2 has been put in place

fishi0x01:
@patjlm this should be fully backwards-compatible. This will only remove publishing. Reading will still happen from both paths: https://github.com/app-sre/qontract-reconcile/blob/1fc7afa1c8dd8d18ec6d50371309e81b904630ab/reconcile/utils/promotion_state.py#L84-L88

I.e., any saas that didnt have a v2 deployment yet will still find the v1 path. We simply will not publish any new v1 path.

patjlm:
> https://github.com/app-sre/qontract-reconcile/blob/1fc7afa1c8dd8d18ec6d50371309e81b904630ab/reconcile/utils/promotion_state.py#L84-L102

Great, thanks. I am a bit surprised by the order of this backward compatible code. I would have thought we'd first check v2 and then fallback to checking v1. But that's out of scope here I suppose.

patjlm (approved):

fishi0x01:
> I am a bit surprised by the order of this backward compatible code. I would have thought we'd first check v2 and then fallback to checking v1. 

You are absolutely right. Makes more sense to check v2 first to reduce calls to s3. Adjusted it accordingly now.

PR #3657, opened 2023-07-07
typed queries for jira-permissions-validator

fishi0x01:
My Friday afternoon project:

Since `jira-permissions-validator` is a new integration, it should use typed queries. This PR introduces:

- typed query for `JiraBoards`
- adjust `JiraClient` factory to use plain primitives
- remove usage of `queries.py` from `jira-permissions-validator`

`JiraClient` supports typed queries now: https://github.com/app-sre/qontract-reconcile/pull/3631

**Test**

```
qd profile run prod jira-permissions-validator
```

cc @hemslo


hemslo (approved):
Nice refactor! It would be good if we have tests for `reconcile/jira_permissions_validator.py`.

fishi0x01:
You are right. Tests would be nice here indeed. Since this is no high prio and the integration didnt have any tests before, Im saving adding tests for Friday afternoon. We will get there step by step :)

PR #3658, opened 2023-07-07
Fix pagination

janboll:
max_pages needs to be evaluated for non value and only then it need so be cheked, that current_pages is smaller max_pages

Additionally, the params_copy needs to be used for self.get to start from the next page

### DIFF
--- a/reconcile/utils/ocm_base_client.py
+++ b/reconcile/utils/ocm_base_client.py
@@ -92,7 +92,7 @@ class OCMBaseClient:
         # fetch pages
         records_seen = 0
         while True:
-            rs = self.get(api_path, params=params)
+            rs = self.get(api_path, params=params_copy)
             for item in rs.get("items", []):
                 yield item
             total_records = rs.get("total", 0)
@@ -100,7 +100,7 @@ class OCMBaseClient:
             records_on_page = rs.get("size", len(rs.get("items", [])))
             records_seen += records_on_page
             if total_records > records_seen and (
-                max_pages and max_pages < current_page
+                not max_pages or max_pages > current_page
             ):
                 # more page available
                 params_copy["page"] = current_page + 1


geoberle (approved):

reconcile/utils/ocm_base_client.py
```diff
@@ -92,15 +92,15 @@ def get_paginated(
...
+            rs = self.get(api_path, params=params_copy)
             for item in rs.get("items", []):
                 yield item
             total_records = rs.get("total", 0)
             current_page = rs.get("page", 0)
             records_on_page = rs.get("size", len(rs.get("items", [])))
             records_seen += records_on_page
             if total_records > records_seen and (
-                max_pages and max_pages < current_page
+                not max_pages or max_pages > current_page
```
patjlm, line 103: when would `not max_pages` be true?

janboll: You are right, the expression is still wrong. Guess a nested if/else or tenery is required.

hemslo: Is `max_pages` parameter used anywhere? In which case we want to early terminate pagination even not get full results?

Btw to keep pagination logic consistent with ocm cli https://github.com/openshift-online/ocm-cli/blob/bb3d95aee24b91436da8b5c2e2a46a681af5ff4a/cmd/ocm/list/cluster/cmd.go#L188C2-L215C1, can use if `records_on_page` is less than `max_page_size` to check if it's the last page. An edge case of that logic is it can send an additional request if the last page is a full page, but it's safer than check `total` as `total` value can change during pagination which can cause wrong termination point.

janboll: Hm, how could the total value change? The old pagination logic worked as you proposed.

janboll: @hemslo, aligned implementation. Please take a look

reconcile/utils/ocm_base_client.py
```diff
@@ -89,18 +89,17 @@ def get_paginated(
...
+            rs = self.get(api_path, params=params_copy)
             for item in rs.get("items", []):
                 yield item
-            total_records = rs.get("total", 0)
             current_page = rs.get("page", 0)
             records_on_page = rs.get("size", len(rs.get("items", [])))
-            records_seen += records_on_page
-            if total_records > records_seen and (
-                max_pages and max_pages < current_page
+            if records_on_page >= max_page_size and (
```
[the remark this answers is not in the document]
janboll: Great idea!

reconcile/utils/ocm_base_client.py
```diff
@@ -89,23 +89,17 @@ def get_paginated(
             params_copy = params.copy()
         params_copy["size"] = max_page_size
 
-        # fetch pages
-        records_seen = 0
         while True:
```
janboll, line 92: max pages could also have a default of, say 1000 and the condition could be added to the loop? :thinking:

hemslo (approved):
LGTM!

PR #3659, opened 2023-07-07
[change-owners] detect field type changes

geoberle:
field type changes are not reported as proper change, instead they are surfaced as metadata change only (change in SHA for the datafile).

now, type changes are reported as regular changes which can be owned and self-serviced.

https://issues.redhat.com/browse/APPSRE-7934


fishi0x01 (approved):
:runner: :rocket:

patjlm (approved):

PR #3662, opened 2023-07-09
[gitlab-housekeeping] add priority label to time_to_merge histogram

maorfr:
this is an attempt to be able to visualize the average time it takes to merge a MR per priority.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -343,7 +343,7 @@ def get_merge_requests(
         if not is_good_to_merge(labels):
             continue
 
-        label_priotiry = min(
+        label_priority = min(
             MERGE_LABELS_PRIORITY.index(merge_label)
             for merge_label in MERGE_LABELS_PRIORITY
             if merge_label in labels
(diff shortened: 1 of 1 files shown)


reconcile/gitlab_housekeeping.py
```diff
@@ -353,6 +353,7 @@ def get_merge_requests(
             "mr": mr,
             "labels": labels,
             "label_priority": label_priotiry,
+            "priority": f"{label_priotiry} - {MERGE_LABELS_PRIORITY[label_priotiry]}",
```
patjlm, line 356: just noticed that typo: `label_priotiry`. Not new though...

maorfr: https://github.com/app-sre/qontract-reconcile/pull/3662/commits/07091194d3a2db9a46d34bf9ab861ab2d39e4feb

patjlm (approved):

reconcile/gitlab_housekeeping.py
```diff
@@ -352,7 +352,8 @@ def get_merge_requests(
         item = {
             "mr": mr,
             "labels": labels,
-            "label_priority": label_priotiry,
+            "label_priority": label_priority,
+            "priority": f"{label_priority} - {MERGE_LABELS_PRIORITY[label_priority]}",
```
geoberle, line 356: so the label value would be something like `1 - approved: critical`, `2 - approved: high` etc? this is nice. in case the we add new priorities because, e.g. `1 - approved: ultra-high` because we could then see that critical would only be 2nd anymore, e.g. `2 - approved: critical`.

same could be achieved by having the absolute numberic priority and the humand readable priorty as a two dedicated labels. depends on what we want to do with these values in dashboards or in alerts.

just mumbling through my first coffee, ignore me if it does not make any sense.

maorfr: side note: i think the label would be `1 - critical`, no?

to the point, i chose a single label so it can be a single column in a table, but i'm open to improving in a future iteration.

PR #3663, opened 2023-07-09
[change-owners] make context conditions available for self-referencing context

geoberle:
Implementing change-types for the explicit removal of files is not easily possible because specifying a context to self and defining a when-condition removal yields an error. This PR enables self-referencing contexts by specifying `context.selector: $.path` or by omitting the `selector` alltogether. Together with the `context.when` condition, file removals can now be self-serviced.

This was previously not possible because only data from the changed file was selectable via `context.selector` and metadata such as `path` is not part of that data. 

This change also harmoinzes the use of direct and indirect context detection within the code.

e.g. the following change-types grants permissions to self-service the removal

```yaml
---
$schema: /app-interface/change-type-1.yml

name: jira-board-auto-remover

priority: low

contextType: datafile
contextSchema: /dependencies/jira-board-1.yml

changes:
- provider: jsonPath
  jsonPathSelectors:
  - $
  context:
    [selector: $.path]  <-- optional because it defaults to this
    when: removed
```

[unrelated change i'm sneaking in with this one: the filediffresolver now caches results and even has the changed files of an MR prepopulated in its cache. for speed]

https://issues.redhat.com/browse/APPSRE-7940


maorfr (approved):
:heart:

patjlm:
The schemas still makes `selector` required: https://github.com/app-sre/qontract-schemas/blob/main/graphql-schemas/schema.yml#L3891. If you're fine, I'll set it as optional

geoberle:
> The schemas still makes `selector` required: https://github.com/app-sre/qontract-schemas/blob/main/graphql-schemas/schema.yml#L3891. If you're fine, I'll set it as optional

right now it is still required and the code would fail if it is not there

patjlm:
> right now it is still required and the code would fail if it is not there
ok, so i'll update https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/76184 to use `selector: $.path`

geoberle:
> > right now it is still required and the code would fail if it is not there
> 
> ok, so i'll update https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/76184 to use `selector: $.path`

that will also fail during runtime. this PR here was actually trying to solve that issue :) 
i will give it another shot soon. sory this was dormant for so long

PR #3668, opened 2023-07-10
sendgrid API proper paging for invites and teammates

geoberle:
use the proper `limit/offset` combination to page until no new items are returned

### DIFF
--- a/reconcile/sendgrid_teammates.py
+++ b/reconcile/sendgrid_teammates.py
@@ -44,27 +44,37 @@ def fetch_desired_state(users):
 @retry()
 def fetch_current_state(sg_client):
     state = []
+    limit = 100
 
     # pending invites
-    limit = 300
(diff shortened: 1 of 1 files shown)


chassing (approved):

reconcile/sendgrid_teammates.py
```diff
@@ -44,27 +44,37 @@ def fetch_desired_state(users):
...
+    while True:
+        invites = sg_client.teammates.pending.get(
+            query_params={"limit": limit, "offset": offset}
+        ).to_dict["result"]
+        if not invites:
+            break
+        for invite in invites:
+            t = Teammate(invite["email"], pending_token=invite["token"])
+            state.append(t)
+        offset += limit
```
patjlm, line 60: shouldn't that be ?
```
offset += len(invites)
```

PR #3670, opened 2023-07-11
[namespaceSelector] additional attributes

chassing:
Add these additional attributes:

* namespace.cluster.spec.region
* namespace.cluster.labels
* namespace.cluster.externalConfiguration.labels

Ticket: [APPSRE-7604](https://issues.redhat.com/browse/APPSRE-7604)


reconcile/gql_definitions/common/saas_files.py
```diff
@@ -84,7 +84,29 @@
     }
   }
   cluster {
-    ...OcConnectionCluster
```
fishi0x01, line 87: nit: Why not adjust the fragment with the additional parameters instead?

fishi0x01 (approved):
LGTM

reconcile/gql_definitions/common/saas_files.py
chassing: I don't like to misuse fragments for other use cases. It looks like the `OcConnectionCluster` fragment defines just the needed attributes for `OcMap`.

PR #3674, opened 2023-07-12
[rhidp] fix idp callback uri for rosa clusters

chassing:
Rosa clusters use a different IDP callback uri pattern.


geoberle (approved):

geoberle:
lgtm for now. lets consider switching from the `ClusterV1` to the `OCMCluster` model during reconciling because it includes more metadata

PR #3675, opened 2023-07-12
update filtering to only filter ERROR messages

cubismod:
Update to https://github.com/app-sre/qontract-reconcile/pull/3624

Filter out only `ERROR` messages as we get a lot of repeat `WARNING`s


dwelch0 (approved):
LGTM

PR #3676, opened 2023-07-13
[rhidp] fix idp callback uri for rosa clusters

chassing:
Rosa clusters use `:433` in the redirect URI which was forgotten in the last bug fix #3674. The CRC is also not displaying the OAuth callback URL correctly 🤦


hemslo (approved):
LGTM! Nice tests.

PR #3678, opened 2023-07-14
Fix: ShardSpecOverrides do not pick the imageRef

lechuk47:
This fix addresses a bug where the `shardSpecOverride.imageRef` is not picked up to as the image for a shard. The main problem is that the helm template uses AWSShardSpecOverride.image_ref attribute instead of the alias `imageRef`. 

This change also includes a refactor over all the helm tests using the generated pedantic classes instead of dictionaries. This issue was not caught in the tests because of that. 

APPSRE-7982


fishi0x01 (approved):
Nice test refactor :rocket:

PR #3684, opened 2023-07-18
[jenkins-job-builder] generate app_name field for each job

maorfr:
part of https://issues.redhat.com/browse/APPSRE-7990

this PR adds an automatic generation of an `app_name` field that will be added to each job definition. this means that job templates can use the `app_name` to template content.


hemslo (approved):
LGTM!

maorfr:
promoted in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/74350

PR #3688, opened 2023-07-19
add terraform and oc version in ASDF .tool-versions, and set a correct python 3.11 version

patjlm:


fishi0x01 (approved):
:heart:

patjlm:
I will not promote this commit in app-interface since this is only developer side

chassing:
❤️

PR #3690, opened 2023-07-19
Add filter option to get_cluster_addons

janboll:
Make it possible to filter returned addons by state. Add filter to aus addon upgrade scheduler

Fixes bug: https://issues.redhat.com/browse/APPSRE-8011

### DIFF
--- a/reconcile/aus/base.py
+++ b/reconcile/aus/base.py
@@ -411,7 +411,9 @@ def fetch_upgrade_policies(
             sector = ocm.sectors[cluster.upgrade_policy.conditions.sector]
 
         if addons:
-            cluster_addons = ocm.get_cluster_addons(cluster_name, with_version=True)
+            cluster_addons = ocm.get_cluster_addons(
+                cluster_name, with_version=True, required_state="ready"
+            )
             for addon in cluster_addons:
                 ccaup = ConfiguredAddonUpgradePolicy.from_cluster_upgrade_spec(
                     cluster,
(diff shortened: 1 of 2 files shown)


reconcile/aus/base.py
```diff
@@ -411,7 +411,9 @@ def fetch_upgrade_policies(
             sector = ocm.sectors[cluster.upgrade_policy.conditions.sector]
 
         if addons:
-            cluster_addons = ocm.get_cluster_addons(cluster_name, with_version=True)
+            cluster_addons = ocm.get_cluster_addons(
+                cluster_name, with_version=True, required_state="ready"
+            )
```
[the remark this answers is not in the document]
janboll: Yeah, I saw this pattern with other OCM methods and chose it, cause it does not change the output of the methods. I.e. reduced side-effects of the change.

reconcile/utils/ocm/ocm.py
```diff
@@ -1673,6 +1676,9 @@ def get_cluster_addons(
             return results
 
         for item in items:
+            if required_state:
+                if item["state"] != required_state:
+                    continue
```
hemslo, line 1681: Can we use `search` params on https://api.openshift.com/#/default/get_api_clusters_mgmt_v1_clusters__cluster_id__addons to filter by `state`? So the filtering logic is in server side, we don't need to do another post filter on client side.

janboll: Getting an unrelated API error locally, while trying that. Will revisit later. Want to fix integration first.

PR #3691, opened 2023-07-19
Temporarily remove insist feature

fishi0x01:
Waiting for one pipeline requires ~1 minute. Potentially any pipeline in any repo can add to that. 
While this is considered a feature, we might want to disable this behavior while we run in degraded state in order to process app-interface MRs faster.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -549,7 +549,8 @@ def run(dry_run, wait_for_pipeline):
                     limit,
                     rebase,
                     pipeline_timeout,
-                    insist=True,
+                    # TODO: set insist=True after gitlab issues cleared
+                    insist=False,
                     wait_for_pipeline=wait_for_pipeline,
                     gl_instance=instance,
                     gl_settings=settings,


OliviaHY (approved):
Not very familiar with this, but no harm to try under current situation especially if this saves us 1 minute per loop.

fishi0x01:
I decided to manually `hold` affected/stuck MRs for now. Better logging will help here https://github.com/app-sre/qontract-reconcile/pull/3692

reconcile/gitlab_housekeeping.py
```diff
@@ -549,7 +549,8 @@ def run(dry_run, wait_for_pipeline):
                     limit,
                     rebase,
                     pipeline_timeout,
-                    insist=True,
+                    # TODO: set insist=True after gitlab issues cleared
+                    insist=False,
```
hemslo, line 553: if `insist` is `False` here, then lower priority mrs could more likely to be merged before high priority ones.

fishi0x01:
Actually this can be configured directly in the integration.yaml
will close this.

PR #3694, opened 2023-07-20
add retry for multiple method in gitlab api

maorfr:
it's better to (re)try and fail than not (re)try at all


patjlm:
I suppose that's a balance between us wanting to get our calls answered and gitlab hoping us to limit our calls... Don't you think that if a retry is needed, there is a chance that gitlab has some trouble keeping pace already, and adding retries would make things worse?

hemslo:
Almost all these methods will be changed in my final cleanup, currently retry won't help much when gitlab server is under heavy pressure, I'd prefer to let it crash and wait for the next round.

maorfr:
the problem i am trying to fix is: change-owners failed silently: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/74577

PR #3695, opened 2023-07-20
[tf-repo] update error message to be more descriptive

cubismod:
Through usage in FedRamp, this error message can appear if a tenant hasn't rebased their merge request, leading their MR to appear as if multiple repos are going to be modified when in fact only one is.


jonmosco:
LGTM

jonmosco (approved):
LGTM - good idea on the hint

PR #3697, opened 2023-07-21
[ocm] basic support for rosa non-sts clusters

chassing:
This PR adds "read" support for non-sts ROSA clusters (clusters w/o an `aws.sts` attribute). No support for the creation of non-sts rosa clusters! Basically, it makes all arns in the `sts` attribute optional.

Ticket: [APPSRE-8040](https://issues.redhat.com/browse/APPSRE-8040)

### DIFF
--- a/reconcile/utils/ocm/ocm.py
+++ b/reconcile/utils/ocm/ocm.py
@@ -32,6 +32,7 @@ from reconcile.ocm.types import (
     ROSAClusterAWSAccount,
     ROSAClusterSpec,
     ROSAOcmAwsAttrs,
+    ROSAOcmAwsStsAttrs,
 )
 from reconcile.utils.exceptions import ParameterError
 from reconcile.utils.ocm_base_client import (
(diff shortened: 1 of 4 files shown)


reconcile/utils/ocm/ocm.py
```diff
@@ -502,20 +511,20 @@ def _get_create_cluster_spec(cluster_name: str, cluster: OCMSpec) -> dict[str, A
...
-                            "worker_role_arn": cluster.spec.account.rosa.worker_role_arn,
+                            "worker_role_arn": cluster.spec.account.rosa.sts.worker_role_arn,
                         },
                         "operator_role_prefix": f"{cluster_name}-{operator_roles_prefix}",
                     },
                 },
             }
 
-            if cluster.spec.account.rosa.controlplane_role_arn:
+            if cluster.spec.account.rosa.sts.controlplane_role_arn:
```
patjlm, line 524: will that fail is there is no `sts`?

patjlm: ah just saw `if not cluster.spec.account.rosa.sts:` above

patjlm (approved):

reconcile/utils/ocm/ocm.py
chassing: See line 503, an error will be raised before.

PR #3698, opened 2023-07-24
Revert "Log warn and error from terraform run"

hemslo:
`os.environ` mutation is not thread safe, can fail with

```
KeyError: 'TF_LOG'
```


kwilczynski:
@hemslo, ah bummer!

kwilczynski (approved):
@hemslo, looks good!

Thank you for a quick fix!

kwilczynski:
> `os.environ` mutation is not thread safe, can fail with
> 
> ```
> KeyError: 'TF_LOG'
> ```

@hemslo, ah bummer!

PR #3703, opened 2023-07-25
Fix status board logging

janboll:


reconcile/status_board.py
```diff
@@ -163,51 +174,70 @@ def get_diff(
...
         return return_list
 
+    @staticmethod
+    def apply_diff(
+        dry_run: bool, ocm_api: OCMBaseClient, diff: list[StatusBoardHandler]
+    ) -> None:
+        def _sort_func(x: StatusBoardHandler) -> int:
+            if x.status_board_object.__class__.__name__ == Product.__name__:
+                return 0
+            return 1
```
hemslo, line 236: Why sort here? Is that because we need to always process `Product` first, then `Application`?

Instead of using class name reflection, how about add a new order field like `priority` to `StatusBoardhandler`, so the sort can just be
```python
sorted(diff, key=lambda h: h.priority)
```

janboll: Considered this option, but did not want to pollute the data classes with metadata.

hemslo: @janboll type or class checking in polymorphism is an anti pattern, it `Product` is always higher than `Application`, we could give them a method:

```python
class Product(AbstractStatusBoard):
  def get_priority(self) -> int:
    return 0
```

```python
class Application(AbstractStatusBoard):
  def get_priority(self) -> int:
    return 1
```

```python
sorted(diff, key=lambda h: h. status_board_object.get_priority())
```

chassing (approved):

hemslo (approved):
LGTM! Thank you for the refactoring effort!

PR #3704, opened 2023-07-25
Machine pool autoscaling

janboll:
This should enable configuring autoscaling for machine pools.

Related schema change: https://github.com/app-sre/qontract-schemas/pull/489

[APPSRE-7899](https://issues.redhat.com/browse/APPSRE-7899)


reconcile/ocm_machine_pools.py
```diff
@@ -123,14 +156,27 @@ class NodePool(AbstractPool):
     aws_node_pool: AWSNodePool
     subnet: Optional[str]
 
+    @staticmethod
+    def _patch_autoscaling(spec: Mapping[str, Any]):
```
hemslo, line 160: Instead of always patch it, can we do conditional patch for hypershift cluster only. So this can support machine pool operations in both cases. Why `hypershift` is not using the same `MachinePoolAutoscaling` schema in https://api.openshift.com/?urls.primaryName=Clusters%20management%20service#/default/post_api_clusters_mgmt_v1_clusters__cluster_id__machine_pools

reconcile/ocm_machine_pools.py
```diff
@@ -94,6 +159,7 @@ def has_diff(self, pool: ClusterMachinePoolV1) -> bool:
             or self.taints != pool.taints
             or self.labels != pool.labels
             or self.instance_type != pool.instance_type
+            or (self.autoscaling is not None and self.autoscaling.has_diff(pool))
```
hemslo, line 162: if `self.autoscaling` is `None` and `pool.autoscale` is not `None`, this will return `False`, but should return `True`. Optional equal is a little tricky, may worth adding a new test case.

esron (approved):

reconcile/ocm_machine_pools.py
janboll: good spot!
However coming from autoscaling non means, that replicas had a value before. So in the scenario, you mentioned replicas would have changed.

hemslo: One more edge case, if `pool.autoscale` is `None`, https://github.com/app-sre/qontract-reconcile/pull/3704/files#diff-eeceffbceb8fdf008edf0e87f095e168a636b2c2c9081be6eb6c4fe766a071cbR41 will return `False`, but it should return `True`, also in real case `replicas` should also change.

hemslo: It maybe easier if we do all `autoscale` None check here, then only pass not None `autoscale` object to `has_diff`

hemslo: like this
```python
def has_diff_autoscale(self, pool):
  if self.autoscale is None and pool.autoscale is None:
    return False
  if self.autoscale is None or pool.autoscale is None:
    return True
  return self.autoscale.has_diff(pool.autoscale)
```

hemslo: or a fancy one:
```python
        match (self.autoscale is None, pool.autoscale is None):
            case (True, True):
                return False
            case (False, False):
                return self.autoscale.has_diff(pool.autoscale)
            case _:
                return True

```

hemslo: just realized a even better one:

```python
def has_diff_autoscale(self, pool):
  match (self.autoscale, pool.autoscale):
    case (None, None):
        return False
    case (None, _) | (_, None):
        return True
    case _:
        return self.autoscale.has_diff(pool.autoscale)
```

hemslo (approved):
LGTM!

PR #3706, opened 2023-07-26
Set skip_region_validation to work with not supported regions

hemslo:
Our current AWS terraform provider doesn't support `ap-southeast-4` yet, use `skip_region_validation` to bypass validation.

https://github.com/hashicorp/terraform-provider-aws/issues/29231

[APPSRE-8071](https://issues.redhat.com/browse/APPSRE-8071)


kwilczynski (approved):
@hemslo, looks good!

PR #3709, opened 2023-07-27
Pin botocore to same version as in setup.py

janboll:
Found, that pip will download every minor version to determine fitting version. This drasticaly slows down builds. Pin version to same as in setup.py to speed up builds


kwilczynski (approved):
@janboll, looks good!

Nice find! Thank you for fixing it!

PR #3712, opened 2023-08-01
Fix deletion of a single app

janboll:
Fix deletion of a single app. Right now single apps can not be deleted, cause ID is not set. This fixes this by reusing existing intances instead of creating a new one


reconcile/status_board.py
```diff
@@ -214,15 +214,22 @@ def create_app(app_name: str, product: Product) -> Application:
...
+                found_apps = [
+                    application
+                    for application in current_products[product_name].applications or []
+                    if application.name == app
+                ]
+                if len(found_apps) != 1:
+                    logging.error(
+                        f'Application "{app}" not found in product "{product_name}"'
+                    )
+                    continue
```
hemslo, line 228: Which case can cause app not found, this is from diff result, so it should always exist.

Can be like this:

```python
                application = next(
                    a
                    for a in current_products[product_name].applications
                    if a.name == app_name
                )
```

hemslo: on second thought, this is already in a loop for names, to avoid O(N^2) nested look up, can do this:

```python
        # existing product, only add/remove applications
        for product_name, apps in diff_result.change.items():
            product = current_products[product_name]
            for app_name in apps.desired - apps.current:
                return_list.append(
                    StatusBoardHandler(
                        action="create",
                        status_board_object=create_app(app_name, product),
                    )
                )
            to_delete = apps.current - apps.desired
            for application in product.applications or []:
                if application.name in to_delete:
                    return_list.append(
                        StatusBoardHandler(
                            action="delete",
                            status_board_object=application,
                        )
                    )
```

janboll: Don't see how second option is different.
I see the point with using next though!

hemslo: @janboll second option is faster as first option is iterate `applications` inside another loop, can be O(N^2) in the worst case.

janboll: I see your point now! thanks for explaining it.

hemslo (approved):
LGTM!

PR #3718, opened 2023-08-02
[AUS] fix addon spec duplication error + use correct SA

geoberle:
due to a bug, the current state for addon upgrade policies had duplicates, which then failed duplicate check during reconciling.

also use the correct service account for addon related work - the one on the org if it exists and only if not the one from the env


patjlm (approved):
Thanks a lot @geoberle

PR #3720, opened 2023-08-03
[AUS] cluster level blocked versions

geoberle:
add support to block versions on cluster level

part of https://issues.redhat.com/browse/APPSRE-7870
depends on https://github.com/app-sre/qontract-schemas/pull/479

### DIFF
--- a/reconcile/aus/models.py
+++ b/reconcile/aus/models.py
@@ -45,7 +45,9 @@ class ClusterUpgradeSpec(BaseModel):
 
     @property
     def blocked_versions(self) -> set[str]:
-        return set(self.org.blocked_versions or [])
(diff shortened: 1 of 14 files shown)


reconcile/aus/models.py
```diff
@@ -45,7 +45,9 @@ def current_version(self) -> str:
 
     @property
     def blocked_versions(self) -> set[str]:
-        return set(self.org.blocked_versions or [])
+        return set(self.org.blocked_versions or []) | set(
+            self.upgrade_policy.conditions.blocked_versions or []
+        )
```
patjlm, line 50: ❤️

patjlm (approved):
Great, thanks!

janboll (approved):

PR #3587, opened 2023-06-14
Implement log retention cloudwatch logic

suzana-nesic:
Log groups are created with infinite log retention. All log groups will have a retention period set within app-interface in order to keep storage costs in check.

Qontract-schemas PR: https://github.com/app-sre/qontract-schemas/pull/456

Part of [APPSRE-7249](https://issues.redhat.com/browse/APPSRE-7249)

PR #3726, opened 2023-08-07
APPSRE-8127 fix ami cleaner

fishi0x01:
https://github.com/app-sre/qontract-reconcile/pull/3587 introduces a change to the AWS query:

```
 cleanup {
      provider
     # NEW
      ... on AWSAccountCleanupOptionCloudWatch_v1 {
        regex
        retention_in_days
      }
      ... on AWSAccountCleanupOptionAMI_v1 {
        regex
        age
        region
      }
}
```

The AMI cleaner only checks for any cleanup entry, thus also taking cloudwatch entries into account. We need to properly distinguish between the 2.

**Test**

Ran locally

```
qd profile run prod aws-ami-cleanup
```

### DIFF
--- a/reconcile/aws_ami_cleanup/integration.py
+++ b/reconcile/aws_ami_cleanup/integration.py
@@ -211,11 +211,15 @@ def run(dry_run: bool, thread_pool_size: int, defer: Optional[Callable] = None) 
     # cannot still be properly monitored (see https://issues.redhat.com/browse/APPSRE-7674),
     # it's easy that it breaks without being noticed. Once it is properly monitored, this should
     # be moved to a typed query.
-    cleanup_accounts = [
-        a
-        for a in queries.get_aws_accounts(terraform_state=True, cleanup=True)
-        if a.get("cleanup")
-    ]
+    query_data = queries.get_aws_accounts(terraform_state=True, cleanup=True)
+    cleanup_accounts = []
+    for data in query_data:
+        cleanup = data.get("cleanup")
+        if not cleanup:
+            continue
+        if not (cleanup.get("regex") or cleanup.get("age") or cleanup.get("region")):
+            continue
+        cleanup_accounts.append(data)
 
     vault_settings = get_app_interface_vault_settings()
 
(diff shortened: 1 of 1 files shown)


kwilczynski:
Adding @suzana-nesic for visibility. :smile:

reconcile/aws_ami_cleanup/integration.py
```diff
@@ -211,11 +211,18 @@ def run(dry_run: bool, thread_pool_size: int, defer: Optional[Callable] = None)
...
+    for data in query_data:
+        cleanups = data.get("cleanup")
+        if not cleanups:
+            continue
+        is_ami_related = False
+        for cleanup in cleanups:
+            is_ami_related |= cleanup.get("provider") == "ami"
+        if not is_ami_related:
+            continue
+        cleanup_accounts.append(data)
```
hemslo, line 225: Since the filter logic become complex, I suggest to extract it.


```python
def _is_cleanup_account(account: Mapping[str, Any]) -> bool:
    cleanups = account.get("cleanup")
    if not cleanups:
        return False
    return any(cleanup.get("provider") == "ami" for cleanup in cleanups)
```

Then it can be used like

```python
    cleanup_accounts = list(
        filter(
            _is_cleanup_account,
            queries.get_aws_accounts(terraform_state=True, cleanup=True),
        )
    )
```

Plus we need a new test for this

chassing (approved):

reconcile/aws_ami_cleanup/integration.py
fishi0x01: Just trying to quick fix the open alert for now - but I fully agree we should add tests here :+1:

hemslo: Then we should add a proper fix after this, a quick fix without tests is a new bug.

rporres: @fishi0x01 Can you create a JIRA card to track the refactor so that we don't lose track of it?

PR #3728, opened 2023-08-08
remove ocp-release-mirror

maorfr:
part of https://issues.redhat.com/browse/APPSRE-7730

we are no longer mirroring releases.


hemslo (approved):
Nice! One less stuck integration!

PR #3730, opened 2023-08-09
[glitchtip] support updating project name and platform

chassing:
Glitchtip project update support.

Depends on: https://github.com/app-sre/qontract-schemas/pull/491
Ticket: [APPSRE-8141](https://issues.redhat.com/browse/APPSRE-8141)


reconcile/glitchtip/reconciler.py
```diff
@@ -27,9 +28,16 @@ def _reconcile_projects(
     ) -> list[Project]:
         """Reconcile organization projects."""
         organization_projects = list(current_projects)
```
hemslo, line 30: `organization_projects` return value is not used, is there a reason to calculate and maintain it?

reconcile/glitchtip/reconciler.py
```diff
@@ -54,6 +62,21 @@ def _reconcile_projects(
                 new_project = Project(name=project.name, platform=project.platform)
             organization_projects.append(new_project)
 
+        for project in [p.desired for p in project_diff.change.values()]:
```
hemslo, line 65: can use iterator to avoid creating a new list

```python
for project in (p.desired for p in project_diff.change.values()):
```

reconcile/glitchtip/reconciler.py
```diff
@@ -54,6 +62,21 @@ def _reconcile_projects(
...
+                    name=project.name,
+                    platform=project.platform,
+                )
+                organization_projects[
+                    organization_projects.index(project)
+                ] = updated_project
+
         for desired_project in desired_projects:
             current_project = organization_projects[
                 organization_projects.index(desired_project)
```
hemslo, line 82: Now that we have `project_diff`, we can directly iterate on current desired pairs:

```python
        for diff_pair in itertools.chain(
            project_diff.change.values(), project_diff.identical.values()
        ):
            current_project = diff_pair.current
            desired_project = diff_pair.desired
```

reconcile/glitchtip/reconciler.py
```diff
@@ -27,9 +28,16 @@ def _reconcile_projects(
...
         """Reconcile organization projects."""
         organization_projects = list(current_projects)
-        for project in set(current_projects).difference(desired_projects):
+        project_diff = diff_any_iterables(
+            current=current_projects,
+            desired=desired_projects,
+            current_key=lambda p: p.slug,
+            desired_key=lambda p: p.slug,
+            equal=lambda p1, p2: p1.diff(p2),
+        )
```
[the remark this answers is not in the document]
chassing: Thanks for the suggestions. I applied them in [e139c55](https://github.com/app-sre/qontract-reconcile/pull/3730/commits/e139c551b9549f7ed58378662ff77e4b91d43407)

reconcile/glitchtip/reconciler.py
chassing: Unfortunately, this won't work. `organization_projects` holds the latest state of all projects, including already applied updates in this reconciliation run. In `project_diff`, we have the differences at the beginning of the `_reconcile_projects` run.

reconcile/glitchtip/reconciler.py
chassing: See my comment below about the usage of `project_diff`. The return of `organization_projects` makes [testing](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/glitchtip/test_glitchtip_reconciler.py#L155) easier.

reconcile/glitchtip/reconciler.py
chassing: Good spot! Applied in f7aeaa5

reconcile/glitchtip/reconciler.py
hemslo: In `_reconcile_projects`, changes for `add` and `delete` should not affect this, the only related state change is in `changes` that create new project object.

If we extract the logic of reconcile project teams into a new method:

```python
    def _reconcile_project_teams(self, current_project, desired_project, organization_slug, organization_teams):
        for team in set(current_project.teams).difference(desired_project.teams):
            logging.info(
                [
                    "remove_project_from_team",
                    organization_slug,
                    team.slug,
                    desired_project.slug,
                    self.client.host,
                ]
            )
            del current_project.teams[current_project.teams.index(team)]
            if not self.dry_run:
                self.client.remove_project_from_team(
                    organization_slug=organization_slug,
                    team_slug=team.slug,
                    slug=desired_project.slug,
                )
        for team in set(desired_project.teams).difference(current_project.teams):
            try:
                org_team = organization_teams[organization_teams.index(team)]
            except ValueError:
                logging.error(
                    f"Cannot find team {team.slug} in organization. This shouldn't happen!"
                )
                continue

            logging.info(
                [
                    "add_project_to_team",
                    organization_slug,
                    org_team.slug,
                    current_project.slug,
                    self.client.host,
                ]
            )
            current_project.teams.append(team)
            if not self.dry_run:
                self.client.add_project_to_team(
                    organization_slug=organization_slug,
                    team_slug=org_team.slug,
                    slug=current_project.slug,
                )
```

Then it can be used like:

```python
        for diff_pair in project_diff.change.values():
            project = diff_pair.desired
            logging.info(
                ["update_project", organization_slug, project.name, self.client.host]
            )
            if not self.dry_run:
                updated_project = self.client.update_project(
                    organization_slug=organization_slug,
                    slug=project.slug,
                    name=project.name,
                    platform=project.platform,
                )
                organization_projects[
                    organization_projects.index(project)
                ] = updated_project
                self._reconcile_project_teams(
                    updated_project,
                    project,
                    organization_slug,
                    organization_teams,
                )

        for diff_pair in project_diff.identical.values():
            self._reconcile_project_teams(
                diff_pair.current,
                diff_pair.desired,
                organization_slug,
                organization_teams,
            )
        return organization_projects
```

In this way, `organization_projects` is not used in new method, if we want to keep it for testing purpose, can be simplified by start with an empty list, add projects to it when processing instead of 

```python
            del organization_projects[organization_projects.index(project)]
```

or

```python
                organization_projects[
                    organization_projects.index(project)
                ] = updated_project
```

reconcile/glitchtip/reconciler.py
hemslo: can be discussed in https://github.com/app-sre/qontract-reconcile/pull/3730#discussion_r1288230385 since they are related.

reconcile/glitchtip/reconciler.py
chassing: I tried to implement your suggestion, but this would lead to a huge PR. I struggle slightly with the `dry-run` mode and getting the `organization_projects` right. For now, I want to keep it as it is and open a new Jira ticket for the refactoring.

chassing: The jira ticket: https://issues.redhat.com/browse/APPSRE-8155

PR #3732, opened 2023-08-09
First throw at adding OCM metrics

janboll:
Add metrics to OCM to pin down HTTP Rate limiting issues observed in production


geoberle (approved):
LGTM
midterm we should find a way to expose a bit more details like the rough resourcetype to help narrow down the culprit. but for now this is great ROI

PR #3665, opened 2023-07-10
Mark if dry-run is active in the logs

rporres:
It makes running integrations and reading logs a much safer and relaxed experience.

PR #3734, opened 2023-08-09
[dynatrace-token-provider] Add a new integration to distribute dynatrace access token

OliviaHY:
### Why:
We need a way to deliver the necessary dynatrace token to Hypershift clusters.
This implement [this](https://gitlab.cee.redhat.com/service/app-interface/-/blob/0fe180e35df1473d2fcc8e5d5e66d63eb5005dc1/docs/app-sre/design-docs/dynatrace-token-provder.md) design doc.
### What
The integration should be able to retrieve clusters with "sre-capabilies.dtp*" labels, compare them against an allow list of OCM org IDs passed through parameters. Then create two tokens with specific scope in Dynatrace for each cluster and make sure the toknes are distributed to the cluster through SyncSet.

We are also assuming that this integration is the only thing that will touch this specific syncset.
### Dependency:
https://github.com/app-sre/qontract-schemas/pull/495/files 
We need this schema change to be able to dynamically pull all the Dynatrace tenant from App Interface and pull its apiToken from Vault.
### Validation:
I created the label "'sre-capabilities.dtp.tenant: {dynatrace_tenant_id}" for`app-int-ex-01` cluster based on [dev guideline](https://service.pages.redhat.com/dev-guidelines/docs/sre-capabilities/framework/ocm-labels/) then ran integration through following use cases:
1. No syncset with defined name exist, was able to see syncset with following structure created
```
{
  "kind": "Syncset",
  "href": "/api/clusters_mgmt/v1/clusters/cluster_id/external_configuration/syncsets/ext-dynatrace-tokens",
  "id": "ext-dynatrace-tokens",
  "resources": [
    {
      "data": {
        "id": "dt0c01.id1",
        "token": "token1"
      },
      "kind": "Secret",
      "metadata": {
        "name": "dynatrace-ingestion-token"
      },
      "apiVersion": "v1"
    },
    {
      "data": {
        "id": "dt0c01.id2",
        "token": "token2"
      },
      "kind": "Secret",
      "metadata": {
        "name": "dynatrace-operator-token"
      },
      "namespace": "dynatrace",
      "apiVersion": "v1"
    }
  ]
}
```

2. Deleting one or both token in Dynatrace, was able to see token gets created and patched in syncset
3. No Ops when both token still present in Dynatrace.


maorfr:
cool progress!

reconcile/dynatrace_token_provider.py
```diff
@@ -0,0 +1,342 @@
...
+                                                            if not dry_run:
+                                                                operator_token = self.create_dynatrace_operator_token(
+                                                                    dt_client,
+                                                                    cluster.ocm_cluster.external_id,
+                                                                )
+                                                                logging.info(
+                                                                    f"Token: {operator_token.id} created in Dynatrace environment: {dt_tenant_id}"
+                                                                )
+                                                            else:
+                                                                logging.info(
```
patjlm, line 230: We usually print the same log (prefixed with DRY-RUN, since https://github.com/app-sre/qontract-reconcile/pull/3665/files) in dry-run and wet-run modes. So there could be just one log before the `if not dry_run`

reconcile/dynatrace_token_provider.py
```diff
@@ -0,0 +1,342 @@
...
+                                                                    "Would create operator token in Dynatrace"
+                                                                )
+                                                    unhandled_exceptions.append(
+                                                        f"{env}/{cluster.organization_id}: {e}"
+                                                    )
+                                        break
+
+                            if ingestion_token and operator_token:
+                                if not dry_run:
```
patjlm, line 239: why is the construct of `syncset_to_be_created` within a dry-run check?

reconcile/dynatrace_token_provider.py
```diff
@@ -0,0 +1,404 @@
...
+        if unhandled_exceptions:
+            raise ReconcileErrorSummary(unhandled_exceptions)
+        sys.exit(0)
+
+    def get_ocm_environments(self) -> list[OCMEnvironment]:
+        return ocm_environment_query(gql.get_api().query).environments
+
+    def get_all_dynatrace_tenant_ids(self) -> list[str]:
+        # TODO: This is pending on the schema changes that register all the dynatrace bootstrap token with App Interface:
+        # https://github.com/app-sre/qontract-schemas/pull/495/files is merged, need to update here to graphQL query
```
geoberle, line 182: 👍

reconcile/dynatrace_token_provider.py
```diff
@@ -0,0 +1,404 @@
...
+            need_patching = False
+            for token_name, token in tokens.items():
+                if not self.token_exist_in_dynatrace(dt_client, token["id"]):
+                    need_patching = True
+                    logging.info(f"{token_name} missing in Dynatrace.")
+                    if token_name == DYNATRACE_INGESTION_TOKEN_NAME:
+                        if not dry_run:
+                            ingestion_token = self.create_dynatrace_ingestion_token(
+                                dt_client, cluster.ocm_cluster.external_id
+                            )
```
geoberle, line 247: what happens with orphaned tokens (token generated here but for whatever reason they did not make it into a syncset, e.g. because of a process crash, network issue, OCM issue, ...). would that be problematic? would we have a way to identify them clean them up later on?

i'm not suggesting that this needs to be solved right away. but maybe we can think this through, e.g. having a way to identify all tokens created through this integration so we can crosscheck later on which ones are used in syncsets and which ones are not.

reconcile/dynatrace_token_provider.py
```diff
@@ -0,0 +1,404 @@
...
+        except Exception as e:
+            if "Not Found" in e.args[0]:
+                syncset = None
+            else:
+                raise e
+        return syncset
+
+    def token_exist_in_dynatrace(self, dt_client: Dynatrace, token_id: str) -> bool:
+        try:
+            result = dt_client.tokens.get(token_id)
```
geoberle, line 301: is there a way to fetch all tokens with one API call?

reconcile/dynatrace_token_provider.py
OliviaHY: Ah yeah good point. The very least we can prefix those tokens with `dtp-` so they are easily recognizable. What do you think?

reconcile/dynatrace_token_provider.py
OliviaHY: For both ingestion and operator token? The only GET APIs are list all token or get a specific token: https://www.dynatrace.com/support/help/dynatrace-api/environment-api/tokens-v2/api-tokens

reconcile/dynatrace_token_provider.py
geoberle: good idea!

reconcile/dynatrace_token_provider.py
OliviaHY: Ah I see. We'd still need to make one API call to each dynatrace environment, but potentially we can get all the tokens from that environment and filter through by prefix finding the relevant ones

geoberle: awesome!

geoberle (approved):
LGTM. thanks for the hard work.
i'm eager to see it in action.

PR #1629, opened 2021-06-11
Remove timeout via tekton

rporres:
The tekton implementation remove running pods associated with the
TaskRun, which is not acceptable for us

https://issues.redhat.com/browse/APPSRE-3352

Signed-off-by: Rafa Porres Molina <rporresm@redhat.com>

PR #1586, opened 2021-05-27
[openshift-saas-deploy-triggers] add support for timeout

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3187

this PR adds the ability to set a timeout for a deployment (for saas-file-1 - supported through the Jenkins timeout wrapper).

docs: https://tekton.dev/docs/pipelines/pipelineruns/#configuring-a-failure-timeout

PR #3738, opened 2023-08-11
APPSRE-8164 set tekton taskrun timeouts according to saas files settings

patjlm:
[APPSRE-8164](https://issues.redhat.com/browse/APPSRE-8164)

PipelineRun `.spec.timeout` is [deprecated](https://tekton.dev/docs/pipelines/pipelineruns/#configuring-a-failure-timeout) and replaced by `.spec.timeouts`: 
> ⚠️ ** timeout is deprecated and will be removed in future versions. Consider using timeouts instead.

In this new field:
* `.spec.timeouts.tasks` sets the timeout globally for all tasks in the pipeline.
* `.spec.timeouts.pipeline` needs to be greater or equal to `.spec.timeouts.tasks + .spec.timeouts.finally`.

So we set `.spec.timeouts.pipeline = 0` and `.spec.timeouts.tasks = ${saas-file-timeout}`, allowing tasks to benefit from that timeout, and all finally tasks to run successfuly, even in case of timeout.

However, that is not sufficient. If we only set values in PipelineRun, each individual TaskRun gets the default timeout of `1h`.
So we also need to set each task timeout.
That is done in Pipeline, adding a `timeout` field in each `.spec.tasks[*]` and `.spec.finally[*]` definition.

### DIFF
--- a/reconcile/openshift_saas_deploy_trigger_base.py
+++ b/reconcile/openshift_saas_deploy_trigger_base.py
@@ -397,7 +397,10 @@ def _construct_tekton_trigger_resource(
                 f"timeout {timeout} is smaller than 60 minutes"
             )
 
-        body["spec"]["timeout"] = timeout
+        body["spec"]["timeouts"] = {
+            "pipeline": "0",
+            "tasks": timeout,
+        }
 
     return (
         OR(body, integration, integration_version, error_details=name),
--- a/reconcile/openshift_tekton_resources.py
+++ b/reconcile/openshift_tekton_resources.py
@@ -48,6 +48,7 @@ SAAS_FILES_QUERY = """
       name
       provider
     }
+    timeout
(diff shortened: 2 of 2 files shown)


rporres:
> However, that is not sufficient. If we only set values in PipelineRun, each individual TaskRun gets the default timeout of 1h. So we also need to set each task timeout.

The [TaskRun documentation](https://tekton.dev/docs/pipelines/taskruns/#configuring-the-failure-timeout) suggests otherwise but it is certainly vague, so I like the idea of being explicit.

rporres (approved):

maorfr:
related to #1586

maorfr:
related to #1629

patjlm:
related to https://github.com/app-sre/qontract-reconcile/pull/2955

PR #3739, opened 2023-08-11
Creating logic for default log retention set up on all log groups

suzana-nesic:
Part of [APPSRE-8172](https://issues.redhat.com/browse/APPSRE-8172) and a follow up to [APPSRE-7249](https://issues.redhat.com/browse/APPSRE-7249)

Currently, the integration is written to make the log retention value configurable. This PR will track the work to include the logic in order to add a default retention value to all log groups in all aws accounts.

### DIFF
--- a/reconcile/aws_cloudwatch_log_retention/integration.py
+++ b/reconcile/aws_cloudwatch_log_retention/integration.py
@@ -1,11 +1,7 @@
 import logging
 import re
 import typing
-from collections.abc import Callable
-from typing import (
-    TYPE_CHECKING,
-    Optional,
-)
+from typing import TYPE_CHECKING
 
 from botocore.exceptions import ClientError
 from pydantic import BaseModel
--- a/reconcile/aws_cloudwatch_log_retention/integration.py
+++ b/reconcile/aws_cloudwatch_log_retention/integration.py
@@ -65,16 +65,20 @@ def check_cloudwatch_log_group_tag(
     return log_group_list
 
 
-def create_aws_log_client(aws_acct: dict) -> tuple[list, AWSApi]:
+def create_awsapi_client(aws_acct: dict) -> AWSApi:
     settings = queries.get_secret_reader_settings()
     accounts = queries.get_aws_accounts(uid=aws_acct.get("uid"))
     awsapi = AWSApi(1, accounts, settings=settings, init_users=False)
+    return awsapi
+    
+    
+def create_log_group_list(awsapi: AWSApi, aws_acct: dict) -> list:
     log_groups = awsapi.get_cloudwatch_logs(aws_acct)
     session = awsapi.get_session(aws_acct["name"])
     region = aws_acct["resourcesDefaultRegion"]
     log_client = awsapi.get_session_client(session, "logs", region)
     log_group_list = check_cloudwatch_log_group_tag(log_groups, log_client)
-    return log_group_list, awsapi
+    return log_group_list
 
 
 def run(dry_run: bool, thread_pool_size: int, defer: Optional[Callable] = None) -> None:
--- a/reconcile/aws_cloudwatch_log_retention/integration.py
+++ b/reconcile/aws_cloudwatch_log_retention/integration.py
@@ -70,8 +70,8 @@ def create_awsapi_client(aws_acct: dict) -> AWSApi:
     accounts = queries.get_aws_accounts(uid=aws_acct.get("uid"))
     awsapi = AWSApi(1, accounts, settings=settings, init_users=False)
     return awsapi
-    
-    
+
+
 def create_log_group_list(awsapi: AWSApi, aws_acct: dict) -> list:
     log_groups = awsapi.get_cloudwatch_logs(aws_acct)
     session = awsapi.get_session(aws_acct["name"])
@@ -90,7 +90,7 @@ def run(dry_run: bool, thread_pool_size: int, defer: Optional[Callable] = None) 
                 aws_acct
             )
             awsapi = create_awsapi_client(aws_acct)
-            log_group_list= create_log_group_list(awsapi, aws_acct)
+            log_group_list = create_log_group_list(awsapi, aws_acct)
 
             for cloudwatch_cleanup_entry in cloudwatch_cleanup_list:
                 for log_group in log_group_list:
@@ -126,7 +126,7 @@ def run(dry_run: bool, thread_pool_size: int, defer: Optional[Callable] = None) 
                                 )
         try:
             awsapi = create_awsapi_client(aws_acct)
-            log_group_list= create_log_group_list(awsapi, aws_acct)
+            log_group_list = create_log_group_list(awsapi, aws_acct)
         except ClientError as e:
             if e.response["Error"]["Code"] == "AccessDeniedException":
                 logging.info(f" Access denied for {aws_act_name}. Skipping...")
--- a/reconcile/aws_cloudwatch_log_retention/integration.py
+++ b/reconcile/aws_cloudwatch_log_retention/integration.py
@@ -1,6 +1,7 @@
 import logging
 import re
 from collections.abc import Callable
+from botocore.exceptions import ClientError
 from typing import (
     TYPE_CHECKING,
     Optional,
@@ -64,6 +65,18 @@ def check_cloudwatch_log_group_tag(
     return log_group_list
 
 
+def create_aws_log_client(aws_acct):
+    settings = queries.get_secret_reader_settings()
+    accounts = queries.get_aws_accounts(uid=aws_acct.get("uid"))
+    awsapi = AWSApi(1, accounts, settings=settings, init_users=False)
+    log_groups = awsapi.get_cloudwatch_logs(aws_acct)
+    session = awsapi.get_session(aws_acct["name"])
+    region = aws_acct["resourcesDefaultRegion"]
+    log_client = awsapi.get_session_client(session, "logs", region)
+    log_group_list = check_cloudwatch_log_group_tag(log_groups, log_client)
+    return log_group_list, awsapi
+
+
 def run(dry_run: bool, thread_pool_size: int, defer: Optional[Callable] = None) -> None:
     aws_accounts = get_aws_accounts(cleanup=True)
     for aws_acct in aws_accounts:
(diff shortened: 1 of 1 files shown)


reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -64,21 +65,26 @@ def check_cloudwatch_log_group_tag(
     return log_group_list
 
 
+def create_aws_log_client(aws_acct: dict) -> tuple[list, AWSApi]:
+    settings = queries.get_secret_reader_settings()
+    accounts = queries.get_aws_accounts(uid=aws_acct.get("uid"))
+    awsapi = AWSApi(1, accounts, settings=settings, init_users=False)
```
janboll, line 71: Please add a return statement here

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -64,21 +65,26 @@ def check_cloudwatch_log_group_tag(
...
+    log_groups = awsapi.get_cloudwatch_logs(aws_acct)
+    session = awsapi.get_session(aws_acct["name"])
+    region = aws_acct["resourcesDefaultRegion"]
+    log_client = awsapi.get_session_client(session, "logs", region)
+    log_group_list = check_cloudwatch_log_group_tag(log_groups, log_client)
+    return log_group_list, awsapi
```
janboll, line 77: Please make this part a new method

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -64,21 +65,26 @@ def check_cloudwatch_log_group_tag(
     return log_group_list
 
 
+def create_aws_log_client(aws_acct: dict) -> tuple[list, AWSApi]:
```
janboll, line 68: This name is misleading, it does actually many more things :sweat_smile:

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -112,3 +118,17 @@ def run(dry_run: bool, thread_pool_size: int, defer: Optional[Callable] = None)
                                         cloudwatch_cleanup_entry.log_retention_day_length
                                     ),
                                 )
+        else:
+            aws_act_name = aws_acct["name"]
+            try:
+                log_group_list, awsapi = create_aws_log_client(aws_acct)
+            except ClientError as e:
+                if e.response["Error"]["Code"] == "AccessDeniedException":
+                    logging.info(f" Access denied for {aws_act_name}. Skipping...")
```
janboll, line 127: Is this missing a continue statement?

janboll: Could also be written like
```python
try:
    log_group_list, awsapi = create_aws_log_client(aws_acct)
except ClientError as e:
    if e.response["Error"]["Code"] == "AccessDeniedException":
        logging.info(f" Access denied for {aws_act_name}. Skipping...")
else:
    for log_group in log_group_list:
        group_name = log_group["logGroupName"]
        retention_days = log_group.get("retentionInDays")
        if retention_days is None:
            logging.info(f" Setting {group_name} retention days to 90")
            if not dry_run:
                awsapi.set_cloudwatch_log_retention(aws_acct, group_name, 90)
```

reconcile/aws_cloudwatch_log_retention/integration.py
hemslo: There is another issue, `create_awsapi_client` is called within a loop, so `queries.get_secret_reader_settings()` and `queries.get_aws_accounts(uid=aws_acct.get("uid"))` (not needed as we already have account object) will called per iteration, and `AWSApi` is not cleaned up afterwards.

I suggest to change the structure to create one `AWSApi` to use:

```python
def create_awsapi_client(accounts: list, thread_pool_size: int) -> AWSApi:
    settings = queries.get_secret_reader_settings()
    return AWSApi(thread_pool_size, accounts, settings=settings, init_users=False)
```

```python
def run(dry_run: bool, thread_pool_size: int) -> None:
    aws_accounts = get_aws_accounts(cleanup=True)
    with create_awsapi_client(aws_accounts) as awsapi:
        for aws_acct in aws_accounts:
           ...
```

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -112,3 +124,18 @@ def run(dry_run: bool, thread_pool_size: int, defer: Optional[Callable] = None)
...
+                logging.info(f" Access denied for {aws_act_name}. Skipping...")
+                continue
+        else:
+            for log_group in log_group_list:
+                group_name = log_group["logGroupName"]
+                retention_days = log_group.get("retentionInDays")
+                if retention_days is None:
+                    logging.info(f" Setting {group_name} retention days to 90")
+                    if not dry_run:
+                        awsapi.set_cloudwatch_log_retention(aws_acct, group_name, 90)
```
hemslo, line 141: There are duplicate logic in this block with `cleanup` block above, it seems what you want is

* for each aws account
  * get all log groups `create_log_group_list` (should rename to `get_log_group_list` as we are not creating new resources)
  * for each log group
    * find matching `cloudwatch_cleanup_entry` 
    * if find
      * `create_cloudwatch_tag` and `set_cloudwatch_log_retention` accordingly
    * else
      * `set_cloudwatch_log_retention` with default value

If you follow above logic and extract methods for each section, it should be easier to understand.

A few tips:
* `log_regex` in `AWSCloudwatchLogRetention` can be `typing.Pattern` directly (`log_regex=re.compile(aws_acct_field["regex"]),`) to avoid recompile multiple times
* to find a matching `cloudwatch_cleanup_entry`, can use this (`cloudwatch_cleanup_entry` will be `None` if not found)
  ```python
                       cloudwatch_cleanup_entry = next(
                            (
                                c
                                for c in cloudwatch_cleanup_list
                                if c.log_regex.match(group_name)
                            ),
                            None,
                        )
  ```

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -64,51 +65,75 @@ def check_cloudwatch_log_group_tag(
...
-                            if not dry_run:
-                                awsapi.set_cloudwatch_log_retention(
-                                    aws_acct,
-                                    group_name,
-                                    int(
-                                        cloudwatch_cleanup_entry.log_retention_day_length
-                                    ),
-                                )
+    app_interface_aws_accounts = get_aws_accounts(cleanup=True)
+    with create_awsapi_client(app_interface_aws_accounts, 1) as awsapi:
```
hemslo, line 84: Now that we only create in `awsapi` in `run`, can pass `thread_pool_size` to it to speed up init

```python
with create_awsapi_client(app_interface_aws_accounts, thread_pool_size) as awsapi:
```

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -64,51 +65,75 @@ def check_cloudwatch_log_group_tag(
...
+def get_log_group_list(awsapi: AWSApi, aws_acct: dict) -> list:
+    log_groups = awsapi.get_cloudwatch_logs(aws_acct)
+    session = awsapi.get_session(aws_acct["name"])
+    region = aws_acct["resourcesDefaultRegion"]
+    log_client = awsapi.get_session_client(session, "logs", region)
+    log_group_list = check_cloudwatch_log_group_tag(log_groups, log_client)
+    return log_group_list
+
+
 def run(dry_run: bool, thread_pool_size: int, defer: Optional[Callable] = None) -> None:
```
hemslo, line 82: `defer: Optional[Callable] = None` this is unused, can be removed.

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -24,7 +25,7 @@
 class AWSCloudwatchLogRetention(BaseModel):
     name: str
     acct_uid: str
-    log_regex: str
+    log_regex: typing.Pattern
```
hemslo, line 28: Also need to change init to use `re.compile` in https://github.com/app-sre/qontract-reconcile/pull/3739/files#diff-773495110cdde5ed044f7f1bbf0eb68bf7ba5d387c6e3985cedaea64ec613760R43

```python
log_regex=re.compile(aws_acct_field["regex"]),
```

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -64,51 +65,75 @@ def check_cloudwatch_log_group_tag(
...
+                        c
+                        for c in cloudwatch_cleanup_list
+                        if c.log_regex.match(group_name)
+                    ),
+                    None,
+                )
+                log_group_tags = aws_log_group["tags"]
+                if not all(
+                    item in log_group_tags.items() for item in MANAGED_TAG.items()
+                ):
```
hemslo, line 114: Do you mean `if log_group_tags.get("managed_by_integration") != QONTRACT_INTEGRATION:` here?

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -64,51 +65,75 @@ def check_cloudwatch_log_group_tag(
...
+                    logging.info(
+                        f" Setting {group_name} retention days to {cloudwatch_cleanup_entry.log_retention_day_length}"
+                    )
+                    if not dry_run:
+                        awsapi.set_cloudwatch_log_retention(
+                            app_interface_aws_acct,
+                            group_name,
+                            int(cloudwatch_cleanup_entry.log_retention_day_length),
+                        )
```
hemslo, line 139: Since there is a check for current `retention_days` match desired in `elif`, maybe good to apply it to default ones too. Can extract the logic to avoid duplication like this

```python
                desired_retention_days = (
                    90
                    if cloudwatch_cleanup_entry is None
                    else int(cloudwatch_cleanup_entry.log_retention_day_length)
                )
                if retention_days != desired_retention_days:
                    logging.info(
                        f" Setting {group_name} retention days to {desired_retention_days}"
                    )
                    if not dry_run:
                        awsapi.set_cloudwatch_log_retention(
                            app_interface_aws_acct, group_name, desired_retention_days
                        )

```

hemslo (approved):
LGTM! Thank you for the effort!

PR #3741, opened 2023-08-14
Use a sticky session for GQL requests

janboll:
Adding a session to the GQL transport adapter. Allows reusing HTTP sessions. Adding a session cache to clean up sessions after an integration run.

Tested locally against qontract-server on appsrep05 showed around 20% improvement for a shared openshift-resources run.

[APPSRE-5616](https://issues.redhat.com/browse/APPSRE-5616)

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -24,6 +24,7 @@ from reconcile.utils.binary import (
 )
 from reconcile.utils.exceptions import PrintToFileInGitRepositoryError
 from reconcile.utils.git import is_file_in_git_repo
+from reconcile.utils.gql import GqlApiSingleton
 from reconcile.utils.runtime.environment import init_env
 from reconcile.utils.runtime.integration import (
     ModuleArgsKwargsRunParams,
@@ -582,6 +583,22 @@ def integration(
     ctx.obj["gql_sha_url"] = gql_sha_url
     ctx.obj["gql_url_print"] = not dry_run and bool(gql_url_print)
     ctx.obj["dump_schemas_file"] = dump_schemas_file
+
+
+@integration.result_callback()
+def exit_integration(
+    ctx,
+    configfile,
+    dry_run,
+    early_exit_compare_sha,
+    check_only_affected_shards,
+    validate_schemas,
+    dump_schemas_file,
+    log_level,
+    gql_sha_url,
+    gql_url_print,
+):
+    GqlApiSingleton.close()
 
 
 @integration.command(short_help="Manage AWS Route53 resources using Terraform.")
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -234,15 +234,20 @@ class GqlApi:
 
 class GqlApiSingleton:
     gql_api: Optional[GqlApi] = None
-    gqlapi_lock = threading.RLock()
+    gqlapi_lock = threading.Lock()
 
-    def __new__(cls, *args, **kwargs):
+    @classmethod
+    def create(cls, *args, **kwargs) -> GqlApi:
         with cls.gqlapi_lock:
             if cls.gql_api:
                 logging.debug("Resestting GqlApi instance")
-                cls.close()
+                cls.close_gqlapi()
             cls.gql_api = GqlApi(*args, **kwargs)
-        return cls
+        return cls.gql_api
+
+    @classmethod
+    def close_gqlapi(cls):
+        cls.gql_api.close()
 
     @classmethod
     def instance(cls) -> GqlApi:
@@ -254,7 +259,7 @@ class GqlApiSingleton:
     def close(cls) -> None:
         with cls.gqlapi_lock:
             if cls.gql_api:
-                cls.gql_api.close()
+                cls.close_gqlapi()
                 cls.gql_api = None
 
 
@@ -266,7 +271,7 @@ def init(
     commit: Optional[str] = None,
     commit_timestamp: Optional[str] = None,
 ):
-    singleton = GqlApiSingleton(
+    return GqlApiSingleton.create(
         url,
         token,
         integration,
@@ -274,7 +279,6 @@ def init(
         commit=commit,
         commit_timestamp=commit_timestamp,
     )
-    return singleton.gql_api
 
 
 def get_resource(path: str) -> dict[str, Any]:
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -234,7 +234,7 @@ class GqlApi:
 
 class GqlApiSingleton:
     gql_api: Optional[GqlApi] = None
-    gqlapi_lock = threading.Lock()
+    gqlapi_lock = threading.RLock()
 
     def __new__(cls, *args, **kwargs):
         with cls.gqlapi_lock:
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -237,10 +237,10 @@ class GqlApiSingleton:
     gqlapi_lock = threading.Lock()
 
     def __new__(cls, *args, **kwargs):
-        if cls.gql_api:
-            logging.debug("Resestting GqlApi instance")
-            cls.close()
         with cls.gqlapi_lock:
+            if cls.gql_api:
+                logging.debug("Resestting GqlApi instance")
+                cls.close()
             cls.gql_api = GqlApi(*args, **kwargs)
         return cls
 
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -1,12 +1,15 @@
 import logging
 import textwrap
+import threading
 from datetime import (
     datetime,
     timezone,
 )
 from typing import (
     Any,
+    Dict,
     Optional,
+    Union,
 )
 from urllib.parse import urlparse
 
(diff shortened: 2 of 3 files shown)


reconcile/utils/gql.py
```diff
@@ -245,13 +255,32 @@ def get_resource(path: str) -> dict[str, Any]:
...
 
 
-def _init_gql_client(url: str, token: Optional[str]) -> Client:
-    req_headers = None
-    if token:
-        # The token stored in vault is already in the format 'Basic ...'
-        req_headers = {"Authorization": token}
-    # Here we are explicitly using sync strategy
-    return Client(transport=RequestsHTTPTransport(url, headers=req_headers, timeout=30))
+class PersistentRequestsHTTPTransport(RequestsHTTPTransport):
```
hemslo, line 258: I had a quick look at [RequestsHTTPTransport](https://github.com/graphql-python/gql/blob/013fa6aceb066be843849e83cc72747f51c07d0e/gql/transport/requests.py), that design is already using session, just can't change `pool_maxsize` (default 10), I'm thinking to change all `@threaded` to max 10.

According to [ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor)

>Changed in version 3.8: Default value of max_workers is changed to min(32, os.cpu_count() + 4). This default value preserves at least 5 workers for I/O bound tasks. It utilizes at most 32 CPU cores for CPU bound tasks which release the GIL. And it avoids using very large resources implicitly on many-core machines.

We are using 4-8 cpu cores, 10 is a proper value, and it can save us a lot of effort to customize parameters in many libs.

We just need to ensure only one Client is created in a single run, `GqlApi` init should happen in main thread, already a Singleton.

To close session after run, can use `self.client.transport.close()`, maybe worth unset `global _gqlapi` after close to do a full cleanup, or refactor Singleton like [VaultClient](https://github.com/app-sre/qontract-reconcile/blob/42bf5140ea9ae61f20d35c1ebfcabf8ec80af014/reconcile/utils/vault.py#L433) to be thread safe, can add a new static method to reset `_instance` to `None`.

In this way, we don't need to have a `global_session_cache`, just use local session and singleton instance.

hemslo: Oh, just found the session will be closed after each [execute_sync](https://github.com/graphql-python/gql/blob/013fa6aceb066be843849e83cc72747f51c07d0e/gql/client.py#L228), so we still need to customize it a little bit.

hemslo: so we can handle session inside our `GqlApi`, then inject it into `PersistentRequestsHTTPTransport`

```python
class GqlApi:
    def __init__(self):
        self.session = requests.session()
    def _init_gql_client(self) -> Client:
        transport = PersistentRequestsHTTPTransport(self.session, self.url, headers=req_headers, timeout=30)
        ...
    def close(self):
        self.session.close()

class PersistentRequestsHTTPTransport(RequestsHTTPTransport):
    def __init__(
        self,
        session: requests.Session,
        url: str,
        headers: Optional[Dict[str, Any]] = None,
        cookies: Optional[Union[Dict[str, Any], RequestsCookieJar]] = None,
        auth: Optional[AuthBase] = None,
        use_json: bool = True,
        timeout: Optional[int] = None,
        verify: Union[bool, str] = True,
        retries: int = 0,
        method: str = "POST",
        **kwargs: Any,
    ):
        super().__init__(
            url,
            headers,
            cookies,
            auth,
            use_json,
            timeout,
            verify,
            retries,
            method,
            **kwargs,
        )
        self.session = session

    def connect(self) -> None:
        pass

    def close(self) -> None:
        pass
```

janboll: Like the idea about closing the client, I will take a look.

reconcile/cli.py
```diff
@@ -584,6 +585,22 @@ def integration(
...
+    dry_run,
+    early_exit_compare_sha,
+    check_only_affected_shards,
+    validate_schemas,
+    dump_schemas_file,
+    log_level,
+    gql_sha_url,
+    gql_url_print,
+):
+    GqlApiSingleton().close()
```
hemslo, line 601: can directly call `close` since it's a class method
```python
GqlApiSingleton.close()
```

reconcile/utils/gql.py
```diff
@@ -218,7 +231,28 @@ def commit_timestamp_utc(self) -> Optional[str]:
...
-_gqlapi: Optional[GqlApi] = None
+class GqlApiSingleton:
+    gql_api: Optional[GqlApi] = None
+    gqlapi_lock = threading.Lock()
+
+    def __new__(cls, *args, **kwargs):
+        with cls.gqlapi_lock:
+            if not cls.gql_api:
+                cls.gql_api = GqlApi(*args, **kwargs)
+        return cls
```
hemslo, line 242: usually `__new__` returns instance object

reconcile/utils/gql.py
```diff
@@ -218,7 +231,28 @@ def commit_timestamp_utc(self) -> Optional[str]:
...
+
+    @classmethod
+    def instance(cls):
+        if not cls.gql_api:
+            raise GqlApiError("gql module has not been initialized.")
+        return cls
```
hemslo, line 248: here we can return instance `cls.gql_api` to match method name

reconcile/utils/gql.py
```diff
@@ -229,29 +263,61 @@ def init(
     commit: Optional[str] = None,
     commit_timestamp: Optional[str] = None,
 ):
-    global _gqlapi
-    _gqlapi = GqlApi(
+    return GqlApiSingleton(
```
hemslo, line 266: This is not exactly match previous logic, previously it will always create a new `GqlApi`, while `__new__` is a conditional create, we need a different force reinit method.

reconcile/utils/gql.py
```diff
@@ -229,29 +263,61 @@ def init(
...
+            auth,
+            use_json,
+            timeout,
+            verify,
+            retries,
+            method,
+            **kwargs,
+        )
+        # can't directly assign, due to mypy type checking
+        self._session = session
```
hemslo, line 314: if we don't assign it in `__init__`, then there will be an edge case that, when `close` is called, `self.session` is None. https://github.com/app-sre/qontract-reconcile/pull/3741/files#diff-ebb669f2f6dc69ceb880efd75f362c14513914fb8737d013dd98f82351964672R140 will break.

janboll: Close just has `pass`. What should close the session here?

hemslo: not that close
```
self.client.transport.session
```
will be None

reconcile/utils/gql.py
janboll: True, but why do we want to reinit? I removed cause I found it not required. The client will be initialized on every integration run (since GqlApi.close() is called)

reconcile/utils/gql.py
janboll: Will adjust

reconcile/utils/gql.py
janboll: Will adjust

reconcile/cli.py
janboll: Will adjust

reconcile/utils/gql.py
janboll: will add 
```
# type: ignore
```

janboll: Concluded, will change the implementation again to always reset the singleton, cause some uses cases like early exist rely on recreating the API instance with changed parameters

reconcile/utils/gql.py
janboll: __new__ should return the instance of the created class. returning another type might work, but is very counterintuitive. Also mypy complains :-)

reconcile/utils/gql.py
```diff
@@ -218,7 +232,30 @@ def commit_timestamp_utc(self) -> Optional[str]:
...
+class GqlApiSingleton:
+    gql_api: Optional[GqlApi] = None
+    gqlapi_lock = threading.Lock()
+
+    def __new__(cls, *args, **kwargs):
+        if cls.gql_api:
+            logging.debug("Resestting GqlApi instance")
+            cls.close()
+        with cls.gqlapi_lock:
+            cls.gql_api = GqlApi(*args, **kwargs)
```
hemslo, line 244: There is an edge case that another thread can acquire the lock and create a new `GqlApi`, it will be override without close, should guard the whole process with lock:

```python
    def __new__(cls, *args, **kwargs):
        with cls.gqlapi_lock:
            if cls.gql_api:
                logging.debug("Resestting GqlApi instance")
                cls.gql_api.close()
            cls.gql_api = GqlApi(*args, **kwargs)
            return cls.gql_api # <- return instance object
```

reconcile/utils/gql.py
hemslo: How about not using magic methods, just plain old classmethod to make it explicit

```
    @classmethod
    def init(cls, *args, **kwargs) -> GqlApi:
        with cls.gqlapi_lock:
            if cls.gql_api:
                logging.debug("Resestting GqlApi instance")
                cls.close()
            cls.gql_api = GqlApi(*args, **kwargs)
            return cls.gql_api
```

reconcile/utils/gql.py
```diff
@@ -218,7 +232,30 @@ def commit_timestamp_utc(self) -> Optional[str]:
         return None
 
 
-_gqlapi: Optional[GqlApi] = None
+class GqlApiSingleton:
+    gql_api: Optional[GqlApi] = None
+    gqlapi_lock = threading.RLock()
```
hemslo, line 237: When are we reentering the lock?

hemslo (approved):
LGTM!

PR #3748, opened 2023-08-21
[ldap-groups] fix group comparision

chassing:
This addresses a reconcile loop. e.g. https://redhat-internal.slack.com/archives/CS0E65QCV/p1692616370687719

### DIFF
--- a/reconcile/utils/internal_groups/models.py
+++ b/reconcile/utils/internal_groups/models.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from enum import Enum
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


fishi0x01 (approved):

reconcile/utils/internal_groups/models.py
```diff
@@ -21,6 +23,9 @@ def __eq__(self, other: object) -> bool:
             raise NotImplementedError("Cannot compare to non Entity objects.")
         return self.type == other.type and self.id == other.id
 
+    def __lt__(self, other: Entity) -> bool:
+        return self.id < other.id
```
hemslo, line 27: @chassing If we use `frozen = True` as `Config` like https://docs.pydantic.dev/1.10/usage/model_config/, then we can save `__eq__` and `__lt__`, let it auto generate `__hash__`. So we can directly compare them like `set(self.members) == set(other.members)`

chassing: Thank you. Please take a look at https://github.com/app-sre/qontract-reconcile/pull/3749

PR #3752, opened 2023-08-22
APPSRE-8218 fix addon upgrade

patjlm:
We are currently considering all addons to compute addon-specific version data stats.
This means we end up with cert-manager version as a minimum version for acm version data.

### DIFF
--- a/reconcile/aus/ocm_addons_upgrade_scheduler_org.py
+++ b/reconcile/aus/ocm_addons_upgrade_scheduler_org.py
@@ -71,9 +71,18 @@ class OCMAddonsUpgradeSchedulerOrgIntegration(
             if isinstance(spec, ClusterAddonUpgradeSpec)
         }
         for addon_id in addons:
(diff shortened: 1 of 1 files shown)


reconcile/aus/ocm_addons_upgrade_scheduler_org.py
```diff
@@ -71,9 +71,18 @@ def process_upgrade_policies_in_org(
...
+                specs=[
+                    spec
+                    for spec in org_upgrade_spec.specs
+                    if isinstance(spec, ClusterAddonUpgradeSpec)
+                    and spec.addon.id == addon_id
+                ],
+            )
             version_data = aus.get_version_data_map(
                 dry_run=dry_run,
-                org_upgrade_spec=org_upgrade_spec,
```
geoberle, line 76: ah i see. thanks for spotting and fixing this!

geoberle (approved):

PR #3753, opened 2023-08-22
APPSRE-8219 normalize tekton pipeline timeout format

patjlm:
https://issues.redhat.com/browse/APPSRE-8219

We can define a pipeline timeout with any Golang accepted duration format.
However tekton will update the Pipeline specs with timeouts following the default format XhYmZs. For example, 90m gets updated as 1h30m0s.
To avoid continuous reconciliation due to this discrepency, we need to set the timeout value with the default Golang format

### DIFF
--- a/reconcile/utils/parse_dhms_duration.py
+++ b/reconcile/utils/parse_dhms_duration.py
@@ -31,6 +31,20 @@ HANDLE_UNIT_MAP = {
 }
 
 
+def seconds_to_hms(seconds: int) -> str:
+    s = seconds % 60
+
+    minutes = seconds // 60
+    if minutes == 0:
+        return f"{s}s"
+
+    h, m = divmod(minutes, 60)
+    if h == 0:
+        return f"{m}m{s}s"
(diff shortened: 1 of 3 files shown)


reconcile/utils/parse_dhms_duration.py
```diff
@@ -31,6 +31,22 @@ def _seconds_to_seconds(seconds: int) -> int:
...
 
+def seconds_to_hms(seconds: int) -> str:
+    s = seconds % 60
+
+    minutes = seconds // 60
+    if minutes == 0:
+        return f"{s}s"
+
+    m = minutes % 60
+    h = minutes // 60
```
hemslo, line 42: Can use `divmod` to save some lines:

```python
    minutes, s = divmod(seconds, 60)
    if minutes == 0:
        return f"{s}s"

    h, m = divmod(minutes, 60)
```

Do we need to support days (`d`) in it?

rporres (approved):

PR #3755, opened 2023-08-23
Make SLO Reporter RHOBS compatible

janboll:
Update DashdotSLO, and allow it to use an external reference for getting SLO data from Prometheus.

Related schema change: https://github.com/app-sre/qontract-schemas/pull/505

https://issues.redhat.com/browse/APPSRE-8197


hemslo (approved):
LGTM!

PR #3756, opened 2023-08-23
use divmod for minutes as well

patjlm:
missing refactor from https://github.com/app-sre/qontract-reconcile/pull/3753


hemslo (approved):
LGTM!

PR #3758, opened 2023-08-23
fix qontract-cli upgrade-cluster-addon 

patjlm:
We need to use the cluster id instead of the cluster name


kwilczynski (approved):
@patjlm, looks good!

Thank you for the fix!

PR #3759, opened 2023-08-23
Fix clusters-network report for ROSA clusters

rporres:
clusters-network report is broken for ROSA clusters that define `awsInfrastructureManagementAccounts`. This patch fixes it while APPSRE-8224 is addressed.


geoberle:
The ROSA condition is absolutely OK since we don't deal with CCS OSD. LGTM and thanks!

PR #3767, opened 2023-08-31
Fix missing child apps

janboll:
Child Apps where are missing as of now. Adding Child Apps with this PR.

https://issues.redhat.com/browse/APPSRE-8227


chassing (approved):

jmelis (approved):
LGTM

PR #3768, opened 2023-08-31
Introduce 3way Merge diff for openshift resources

lechuk47:
# Overview  
This PR changes the way Openshift objects are compared to know if they need to be applied or not. This change affects multiple integrations, like all openshift-related integrations and terraform-resources. 

Currently, QR uses a custom method to compare objects which sometimes is not accurate. For example, removing an annotation from an object is not detected, non-expected attributes in desired objects that are removed by the API server are not detected, etc. 

This PR introduces a 3-way merge strategy to compare objects. 

## How it works

image

### Notes 
* The Original object is represented by the QR hash. It's the last Desired state that was applied 
* The last comparison is made for the desired object attributes only. Changes introduced by operators or other stuff are not considered.

## Rollout 
* A feature toggle to use this strategy has been set. Clusters need to be added to the `openshift-resources-3way-diff-strategy` feature flag in unleash to enable this strategy. 
* The plan is to test this locally before enabling the feature flag in the QR runtime.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -1,5 +1,4 @@
 import base64
-
 import itertools
 import logging
 from collections.abc import (
@@ -43,9 +42,9 @@ from reconcile.utils.oc import (
     StatusCodeError,
     UnsupportedMediaTypeError,
 )
+from reconcile.utils.openshift_resource import QONTRACT_ANNOTATIONS
 from reconcile.utils.openshift_resource import OpenshiftResource as OR
 from reconcile.utils.openshift_resource import ResourceInventory
-from reconcile.utils.openshift_resource import QONTRACT_ANNOTATIONS
 
 ACTION_APPLIED = "applied"
 ACTION_DELETED = "deleted"
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -17,7 +17,12 @@ from typing import (
     runtime_checkable,
 )
 
+import json
+import copy
+import base64
 import yaml
+import jsonpatch
+
 from sretoolbox.utils import (
     retry,
     threaded,
@@ -556,6 +561,123 @@ def check_unused_resource_types(ri):
             logging.warning(msg)
 
 
+def _normalize_secret(secret: OR):
+    body = secret.body
+    string_data = body.pop("stringData", None)
+    if string_data:
+        body.setdefault("data", {})
+        for k, v in string_data.items():
+            v = base64.b64encode(str(v).encode()).decode("utf-8")
+            body["data"][k] = v
+
+
+def _normalize_noop(item: OR):
+    pass
+
+
+def normalize_object(item: OR, deep_copy=True) -> OR:
+    if deep_copy:
+        n = copy.deepcopy(item)
+    else:
+        n = item
+
+    n.body["metadata"].setdefault("annotations", {})
+    n.remove_qontract_annotations()
+
+    # Remove K8s managed attributes not neede to compare objects
+    n.body["metadata"].pop("creationTimestamp", None)
+    n.body["metadata"].pop("resourceVersion", None)
+    n.body["metadata"].pop("generation", None)
+    n.body["metadata"].pop("selfLink", None)
+    n.body["metadata"].pop("uid", None)
+    n.body["metadata"].pop("fieldRef", None)
+    n.body["metadata"].pop("managedFields", None)
+    n.body["metadata"].pop("namespace", None)
+
+    # Run normalizers on Kinds with special needs
+    normalizers = {"Secret": _normalize_secret}
+    normalizers.get(n.body["kind"], _normalize_noop)(n)
+    return n
+
+
+def three_way_merge_patch_diff(c_item: OR, d_item: OR) -> bool:
+    # If the current item does not have the last applied configuratiob annotation
+    # Apply it. This integration is based in client-side applies, so this annotation
+    # must be present
+    if not c_item.last_applied_configuration:
+        logging.debug(
+            "Current object does not have last applied configuration annotation"
+        )
+        return True
+
+    try:
+        original = OR(json.loads(c_item.last_applied_configuration), "", "")
+    except Exception:
+        logging.debug("Error parsing last applied configuration configuration")
+        return True
+
+    current = normalize_object(c_item)
+    original = normalize_object(original, deep_copy=False)
+    desired = normalize_object(d_item, deep_copy=False)
+
+    # If differences between original and desired object -> Apply
+    patch = jsonpatch.JsonPatch.from_diff(desired.body, original.body)
+    if len(patch.patch) > 0:
+        logging.debug("Original and Desired object differs -> Apply")
+        logging.debug(patch.patch)
+        return True
+
+    # If there are differences between current and desired -> Apply
+    # The patch only detects changes with attributes defined in the desired state.
+    # Values in the current state added by operators or other actors are not taken
+    # into account
+    patch = jsonpatch.JsonPatch.from_diff(current.body, desired.body)
+    for item in patch.patch:
+        if item["op"] == "add" or item["op"] == "replace":
+            logging.debug("Current and Desired object differs -> Apply")
+            logging.debug(patch.patch)
+            return True
+
+    return False
+
+
+def three_way_merge_patch_diff_using_hash(c_item: OR, d_item: OR) -> bool:
+    # Get the ORIGINAL object hash
+    # This needs to be improved in OR, by now is just a PoC
+    c_item_sha256 = ""
+    try:
+        annotations = c_item.body["metadata"]["annotations"]
+        c_item_sha256 = annotations["qontract.sha256sum"]
+    except KeyError:
+        # IF no HASH -> Apply
+        return True
+
+    # Original object does not match Desired -> Apply
+    # Current object is not recalculated!
+    # d_item_sha256 = OR.calculate_sha256sum(OR.serialize(d_item.body))
+    # if c_item_sha256 != d_item_sha256:
+    if c_item_sha256 != d_item.sha256sum():
+        logging.info("Original and Desired object differs -> Apply")
+        return True
+
+    # If there are differences between current and desired -> Apply
+    # The patch only detects changes with attributes defined in the desired state.
+    # Values in the current state added by operators or other actors are not taken
+    # into account
+    current = normalize_object(c_item)
+    desired = normalize_object(d_item, deep_copy=False)
+
+    patch = jsonpatch.JsonPatch.from_diff(current.body, desired.body)
+    for item in patch.patch:
+        if item["op"] == "add" or item["op"] == "replace":
+            # Add or Replace from DESIRED Over CURRENT
+            logging.info("Desired and Current objects differ -> Apply")
+            logging.info(patch.patch)
+            return True
+
+    return False
+
+
 def _realize_resource_data(
     unpacked_ri_item,
     dry_run,
@@ -592,18 +714,10 @@ def _realize_resource_data(
                 )
                 logging.debug(msg)
             else:
-                # If resource doesn't have annotations, annotate and apply
-                if not c_item.has_qontract_annotations():
-                    msg = (
-                        "[{}/{}] resource '{}/{}' present "
-                        "w/o annotations, annotating and applying"
-                    ).format(cluster, namespace, resource_type, name)
-                    logging.info(msg)
-
                 # don't apply if there is a caller (saas file)
                 # and this is not a take over
                 # and current item caller is different from the current caller
-                elif caller and not take_over and c_item.caller != caller:
+                if caller and not take_over and c_item.caller != caller:
                     # if the current item is owned by a caller that no longer exists,
                     # do nothing. the condition is nested so we fall into this condition
                     # so we end up either applying to take ownership, or we error if the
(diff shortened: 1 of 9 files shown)


reconcile/utils/unleash.py
```diff
@@ -51,6 +51,20 @@ def apply(self, context: Optional[dict] = None) -> bool:
         return enable
 
 
+class EnableClusterStrategy(Strategy):
+    def load_provisioning(self) -> list:
```
janboll, line 55: Could create a ClusterStrategy class, that implements this only once.

reconcile/utils/openshift_resource.py
```diff
@@ -193,6 +197,14 @@ def api_version_mutation(val1, val2):
             return True
         return val1 == val2
 
+    @property
+    def last_applied_configuration(self):
```
hemslo, line 201: Where is the usage of this?

reconcile/openshift_base.py
```diff
@@ -568,6 +647,177 @@ def _realize_resource_data(
...
+                "namespace": namespace,
+                "kind": resource_type,
+                "name": name,
+                "privileged": privileged,
+            }
+            actions.append(action)
+        except StatusCodeError as e:
+            ri.register_error()
+            msg = "[{}/{}] {}".format(cluster, namespace, str(e))
+            logging.error(msg)
```
hemslo, line 804: Now that we are creating a new method, can we make the logic flow less deep nested?
There is a generic differ can be used to diff current and desired dict, https://github.com/app-sre/qontract-reconcile/blob/2fa08e3f01b2d0aabde8ce8bf3247b41d6d320ef/reconcile/utils/differ.py#L41C5-L41C18.

```python
    diff_result = differ.diff_mappings(
        data["currrent"], data["desired"], equal=three_way_merge_patch_diff_using_hash
    )
```

Then can process each part:

```python
    diff_result.add
    diff_result.delete
    diff_result.change
    diff_result.identical
```

It will be good we don't have any `continue` in this method to make it a flat flow.

reconcile/openshift_base.py
```diff
@@ -556,6 +559,82 @@ def check_unused_resource_types(ri):
...
+            body["data"][k] = v
+
+
+def _normalize_noop(item: OR):
+    pass
+
+
+def normalize_object(item: OR, deep_copy=True) -> OR:
+    if deep_copy:
+        n = copy.deepcopy(item)
```
hemslo, line 578: Can we avoid deepcopy by cherry pick fields to create new object, instead of mutate in place?

reconcile/openshift_base.py
```diff
@@ -556,6 +559,82 @@ def check_unused_resource_types(ri):
...
+
+    # Remove K8s managed attributes not neede to compare objects
+    n.body["metadata"].pop("creationTimestamp", None)
+    n.body["metadata"].pop("resourceVersion", None)
+    n.body["metadata"].pop("generation", None)
+    n.body["metadata"].pop("selfLink", None)
+    n.body["metadata"].pop("uid", None)
+    n.body["metadata"].pop("fieldRef", None)
+    n.body["metadata"].pop("managedFields", None)
+    n.body["metadata"].pop("namespace", None)
```
hemslo, line 593: To avoid duplicate logic, can extract a `K8S_MANAGED_ATTRIBUTES` then return a new dict:

```python
    K8S_MANAGED_ATTRIBUTES = {"creationTimestamp", "resourceVersion", "generation", ...}
    {k: v for k, v in n.body["metadata"].items() if k not in K8S_MANAGED_ATTRIBUTES}
```

reconcile/openshift_base.py
```diff
@@ -568,6 +647,177 @@ def _realize_resource_data(
...
+            actions.append(action)
+        except StatusCodeError as e:
+            ri.register_error()
+            msg = "[{}/{}] {}".format(cluster, namespace, str(e))
+            logging.error(msg)
+
+    return actions
+
+
+def _realize_resource_data_old(
```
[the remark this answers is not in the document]
lechuk47: I renamed it to `_qr`.  Once the new strategy is rolled out to new clusters this will be removed. So not a big deal.

reconcile/openshift_base.py
```diff
@@ -556,6 +559,82 @@ def check_unused_resource_types(ri):
             logging.warning(msg)
 
 
+def _normalize_secret(secret: OR):
```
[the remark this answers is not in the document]
lechuk47: fixed.

reconcile/openshift_base.py
```diff
@@ -556,6 +559,82 @@ def check_unused_resource_types(ri):
...
+    body = secret.body
+    string_data = body.pop("stringData", None)
+    if string_data:
+        body.setdefault("data", {})
+        for k, v in string_data.items():
+            v = base64.b64encode(str(v).encode()).decode("utf-8")
+            body["data"][k] = v
+
+
+def _normalize_noop(item: OR):
```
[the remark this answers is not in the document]
lechuk47: fixed.

reconcile/openshift_base.py
```diff
@@ -556,6 +559,82 @@ def check_unused_resource_types(ri):
...
+    n.body["metadata"].pop("resourceVersion", None)
+    n.body["metadata"].pop("generation", None)
+    n.body["metadata"].pop("selfLink", None)
+    n.body["metadata"].pop("uid", None)
+    n.body["metadata"].pop("fieldRef", None)
+    n.body["metadata"].pop("managedFields", None)
+    n.body["metadata"].pop("namespace", None)
+
+    # Run normalizers on Kinds with special needs
+    normalizers = {"Secret": _normalize_secret}
```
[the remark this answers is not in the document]
lechuk47: it does. fixed.

reconcile/utils/unleash.py
```diff
@@ -51,6 +51,20 @@ def apply(self, context: Optional[dict] = None) -> bool:
...
+    def apply(self, context: Optional[dict] = None) -> bool:
+        enable = False
+
+        if context and "cluster_name" in context.keys():
+            # if cluster in context is in clusters sent from server, enable
+            enable = context["cluster_name"] in self.parsed_provisioning
+
+        return enable
+
+
```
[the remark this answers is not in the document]
lechuk47: I grouped the cluster strategies with a common base class and added a test. PTAL.

reconcile/utils/unleash.py
lechuk47: done. PTAL

reconcile/utils/openshift_resource.py
lechuk47: This was used when I tested the same thing but using the last applied annotation instead of the hash. Removed.

reconcile/openshift_base.py
lechuk47: I wasn't aware of this differ class. I refactored the method to make use of it and added tests as well. Thanks for doing it and letting me know. PTAL.

reconcile/openshift_base.py
lechuk47: Yes, changed the method to create a new object instead of copying it.

reconcile/openshift_base.py
lechuk47: done

reconcile/openshift_base.py
```diff
@@ -568,6 +647,177 @@ def _realize_resource_data(
     no_dry_run_skip_compare,
     override_enable_deletion,
     recycle_pods,
+):
+    args = locals()
```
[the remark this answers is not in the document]
chassing: It isn't about personal preference, it's more about readability and mypy type hints. Currently, the whole file isn't type-safe and things like `locals()` make everything not easier to read.

reconcile/openshift_base.py
```diff
@@ -556,6 +559,82 @@ def check_unused_resource_types(ri):
...
+        for k, v in string_data.items():
+            v = base64.b64encode(str(v).encode()).decode("utf-8")
+            body["data"][k] = v
+
+
+def _normalize_noop(item: OR):
+    pass
+
+
+def normalize_object(item: OR, deep_copy=True) -> OR:
```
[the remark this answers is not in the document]
lechuk47: fixed

reconcile/openshift_base.py
```diff
@@ -568,6 +647,177 @@ def _realize_resource_data(
...
+    oc_map: ClusterMap,
+    ri: ResourceInventory,
+    take_over,
+    caller,
+    all_callers,
+    wait_for_namespace,
+    no_dry_run_skip_compare,
+    override_enable_deletion,
+    recycle_pods,
 ):
```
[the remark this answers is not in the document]
lechuk47: fixed

reconcile/openshift_base.py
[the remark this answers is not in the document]
lechuk47: Ok, I changed the code to not use locals().

reconcile/openshift_base.py
hemslo: There is one issue with that new `OR` object, `body=item.body` so `item.body` will pass as ref, then any mutation to `n.body` will also affect original body, is this expected?

An easy way to use without worry about mutation is always return a new object.
For example:

To add or override a dict entry, instead of `n.body["metadata"] = metadata`, can use `n.body = n.body | {"metadata": metadata}`.

To remove a dict entry, instead of `annotations.pop("qontract.integration", None)`, can use `new_annotations = {k: v for k, v in body.items() if k != "qontract.integration"}`

reconcile/openshift_base.py
hemslo: small typo `neede` -> `needed`

reconcile/openshift_base.py
```diff
@@ -556,18 +570,441 @@ def check_unused_resource_types(ri):
...
+    cluster: str,
+    namespace: str,
+    resource_type: str,
+    data: Mapping[Any, Any],
+    options: ApplyOptions,
+) -> list[dict[str, Any]]:
+    actions: list[dict[str, Any]] = []
+
+    for name, dp in modified_resources.items():
+        b_apply = True
```
hemslo, line 717: what's `b_apply`? Is that more like should apply? Maybe extract filter logic out to make it clear:

```python
def should_apply(current, options, ri, cluster, name, namespace, resource_type):
    if not options.dry_run and options.no_dry_run_skip_compare:
        msg = (
            f"[{cluster}/{namespace}] skipping compare of resource"
            f"'{resource_type}/{name}'"
        )
        logging.debug(msg)
        return True
    if (
        not options.take_over
        and options.caller
        and current.caller != options.caller
        and options.all_callers
        and current.caller in options.all_callers
    ):
        ri.register_error()
        logging.error(
            f"[{cluster}/{namespace}] resource '{resource_type}/{name}' present and managed by another caller: {current.caller}"
        )
        return False
    logging.debug("CURRENT: " + OR.serialize(OR.canonicalize(current.body)))
    return True
```

Then it can be used like

```python
        if should_apply(current, options, ri, cluster, name, namespace, resource_type):
```

some params just for logging purpose, can also define that method inside `handle_modified_resources` to use closure to reduce some params.
```

File: reconcile/openshift_base.py [hunk 15]
```diff
@@ -556,18 +570,441 @@ def check_unused_resource_types(ri):
...
+    cluster: str,
+    namespace: str,
+    resource_type: str,
+    data: Mapping[Any, Any],
+    options: ApplyOptions,
+) -> list[dict[str, Any]]:
+    actions: list[dict[str, Any]] = []
+
+    for name, current in deleted_resources.items():
+        b_delete = True
```
hemslo, line 779: similar here, can extract a method for `should_delete`

reconcile/openshift_base.py
hemslo: Nice!

reconcile/openshift_base.py
lechuk47: fixed.

reconcile/openshift_base.py
lechuk47: makes sense. PTAL

reconcile/openshift_base.py
lechuk47: done.

reconcile/openshift_base.py
hemslo: that only solved one layer of copy, nested dict object is still a reference to original one, so it still mutate original object, to make it fully decoupled, there are a few changes needed:

```python
def _normalize_secret(secret: OR) -> None:
    body = secret.body
    string_data = body.get("stringData")
    if string_data:
        data = body.get("data") or {}
        body["data"] = data | {
            k: base64.b64encode(str(v).encode()).decode("utf-8")
            for k, v in string_data.items()
        }
        secret.body = {k: v for k, v in body.items() if k != "stringData"}

QONTRACT_ANNOTATIONS = {
    "qontract.integration",
    "qontract.integration_version",
    "qontract.sha256sum",
    "qontract.update",
    "qontract.caller_name",
}

def normalize_object(item: OR) -> OR:
    # Remove K8s managed attributes not needed to compare objects
    metadata = {
        k: v
        for k, v in item.body["metadata"].items()
        if k not in NORMALIZE_COMPARE_EXCLUDED_ATTRS
    }

    annotations = metadata.get("annotations") or {}

    metadata["annotations"] = {
        k: v for k, v in annotations.items() if k not in QONTRACT_ANNOTATIONS
    }

    n = OR(
        body=item.body | {"metadata": metadata},
        integration=item.integration,
        integration_version=item.integration_version,
        error_details=item.error_details,
        caller_name=item.caller_name,
        validate_k8s_object=False,
    )

    # Run normalizers on Kinds with special needs
    NORMALIZERS.get(n.body["kind"], _normalize_noop)(n)
    return n
```

In this way, for every dict operation, only root level field assign value, without any side effect of original dict.

lechuk47: Fixed. PTAL

hemslo: Nice! https://github.com/app-sre/qontract-reconcile/pull/3768/files#diff-e1ae1861e6058b02d61980281bb71def45745bdfc82c269bfcf2a52f6be4fabdR295 just left for future usage, right? Not used noe.

lechuk47: Method Removed.

hemslo (approved):
LGTM! A big step forward!

PR #3770, opened 2023-08-31
Revert "Remove print statement"

janboll:
Reverts app-sre/qontract-reconcile#3769


jmelis (approved):

patjlm:
for  context, why reverting that?

janboll:
It will create duplicate entries, cause of not checking for uniqueness

PR #3774, opened 2023-09-01
Capture errors from secrets in saasherder

rporres:
We currently silently fail when saasherder cannot populate desired state due to secret management issues.

This patch also contain a safeguard to make sure that populate desired state does log something (in DEBUG, to avoid duplicates in prod) so that debugging is feasible. Not the cleanest, but a better solution would require a whole refactoring of _process_template.

### DIFF
--- a/reconcile/utils/saasherder/saasherder.py
+++ b/reconcile/utils/saasherder/saasherder.py
@@ -881,11 +881,16 @@ class SaasHerder:  # pylint: disable=too-many-public-methods
             environment_parameters = self._collect_parameters(
                 target.namespace.environment
             )
-            environment_secret_parameters = self._collect_secret_parameters(
-                target.namespace.environment
-            )
             target_parameters = self._collect_parameters(target)
-            target_secret_parameters = self._collect_secret_parameters(target)
+
+            try:
+                environment_secret_parameters = self._collect_secret_parameters(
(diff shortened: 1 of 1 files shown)


chassing (approved):

reconcile/utils/saasherder/saasherder.py
```diff
@@ -1267,8 +1272,11 @@ def populate_desired_state_saas_file(
                 github=spec.github,
                 target_config_hash=spec.target_config_hash,
             )
-        except Exception:
-            # log message send in _process_template
+        except Exception as e:
+            # error log message send in _process_template. We log here debug to have a
+            # safeguard in case something breaks there unexpectedly. We cannot just
+            # register an error without logging as inventory errors don't have details.
+            logging.debug(f"Error in populate_desired_state_saas_file: {e}")
```
hemslo, line 1279: If `re.register_error` can collect actual error object and print details somewhere, that will be best, but also requires bigger change. The exception is eaten here, won't appear  in APM or glitchtip.

hemslo (approved):
LGTM!

reconcile/utils/saasherder/saasherder.py
rporres: yeah, the way things are managed are not ideal.

PR #3777, opened 2023-09-01
 terraform-state-1.yml support for Terraform Repo #3766 

cubismod:
[APPSRE-8242](https://issues.redhat.com/browse/APPSRE-8242)

Allows Terraform Repo to work with [terraform-state-1.yml](https://github.com/app-sre/qontract-schemas/blob/main/schemas/dependencies/terraform-state-1.yml) files. terraform-state files are separate from the actual Terraform state and allow you to specify where state files for specific AppSRE integrations are stored within AWS rather than relying on hard-coded paths and Vault secret values. This is a key step to getting terraform-repo in commercial.

Here's what a terraform state file would look like for Terraform Repo:

```
$schema: /dependencies/terraform-state-1.yml
provider: s3
bucket: tf-state
region: us-east-1 
integrations:
- integration: terraform-repo
  key: tf-repo
```

So then, if you created a terraform repo named "my-cool-repo" in an AWS account with this state attached, it would be stored in the following location on S3:

`s3://tf-state/tf-repo/my-cool-repo-tf-repo.tfstate`

Additionally adds new tests to validate outputs are correct for the executor. Executor PR: https://github.com/app-sre/terraform-repo-executor/pull/14

Example of expected output from terraform-repo QR integration now:
```yaml
dry_run: true
repos:
- repository: https://git-example/tf-repo-example
  name: a_repo
  ref: a390f5cb20322c90861d6d80e9b70c6a579be1d0
  project_path: tf
  delete: false
  secret:
    path: aws-secrets/terraform/foo
    version: 1
  bucket: app-sre
  region: us-east-1
  bucket_path: tf-repo
```

### DIFF
--- a/reconcile/terraform_repo.py
+++ b/reconcile/terraform_repo.py
@@ -102,7 +102,7 @@ class TerraformRepoIntegration(
         if repo_diff_result:
             self.print_output(repo_diff_result, dry_run)
 
-    def print_output(self, diff: list[TerraformRepoV1], dry_run: bool):
+    def print_output(self, diff: list[TerraformRepoV1], dry_run: bool) -> None:
         """Parses and prints the output of a Terraform Repo diff for the executor
 
         :param diff: list of terraform repos to be acted on
--- a/reconcile/terraform_repo.py
+++ b/reconcile/terraform_repo.py
@@ -47,6 +47,9 @@ class RepoOutput(BaseModel):
     project_path: str
     delete: bool
     secret: RepoSecret
+    bucket: Optional[str]
+    region: Optional[str]
+    bucket_path: Optional[str]
 
 
 class OutputFile(BaseModel):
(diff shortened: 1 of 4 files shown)


esron (approved):
LGTM

reconcile/test/test_terraform_repo.py
```diff
@@ -209,6 +301,62 @@ def test_update_repo_state(int_params, existing_repo, state_mock):
...
+
+    diff = integration.calculate_diff(
+        existing_state=existing_state,
+        desired_state=desired_state,
+        dry_run=True,
+        state=state_mock,
+    )
+
+    assert diff
+    if diff:
```
hemslo, line 324: `diff` is already `assert`, why need to `if` check again?

reconcile/test/test_terraform_repo.py
```diff
@@ -209,6 +301,62 @@ def test_update_repo_state(int_params, existing_repo, state_mock):
...
```
hemslo, line 351: same here

reconcile/test/test_terraform_repo.py
cubismod: mypy was complaining about `if` possibly being None however doesn't seem to be giving me issues now

cubismod:
cc @OliviaHY following up on our chat last week

reconcile/terraform_repo.py
```diff
@@ -97,40 +100,57 @@ def run(
...
+                delete=repo.delete or False,
+                secret=RepoSecret(
+                    path=repo.account.automation_token.path,
+                    version=repo.account.automation_token.version,
+                ),
+            )
+            # terraform-repo will store its statefiles in a specified directory if there is a
+            # terraform-state yaml file associated with the AWS account and a configuration is
+            # listed for terraform-repo, otherwise it will default to loading this information
+            # from the automation_token secret in Vault
```
OliviaHY, line 130: Under this default situation, the only difference is the statefile key will be different? Would the actual repo name still show up in the file name?

OliviaHY (commented):
Thank you so much for this change! Exciting to see terraform-repo in action!

reconcile/gql_definitions/ocm_oidc_idp/__init__.py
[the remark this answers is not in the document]
cubismod: Running `make generate` added that for some reason

reconcile/terraform_repo.py
cubismod: Yes so each repo has its own TF statefile. In the default situation, each repo statefile would just be stored at the root directory of the S3 bucket rather than under a subdirectory

OliviaHY (approved):

PR #3784, opened 2023-09-08
[sharded dry-runs] speed up shard detection

geoberle:
shard detection turned out to be slow on bigger desired states like on prometheus-rules tester and openshift-resources.

in this MR the approach has been simplified for these integrations by returning a desired state that encodes the shards as keys in the state and uses a hash for all the configs that belong to that shard.

this makes the desired state to operate on consistently small and all actions are way faster. for hash calculation, deepdiff.Deephash is used, which produces consistent hashes in a fast way.

before this change, prometheus rules tester took > 15s to find the affected shards, now it does the same in ~2s.

also, i believe we should restructure the early exit desired state of all integrations to have their shards encoded as top level keys. this way, shard detection could be achieved without defining jsonpaths to the places where shards can be found in the desired state because it is found in the same place for all integrations. harmonization ftw.


reconcile/openshift_resources_base.py
```diff
@@ -1357,9 +1367,6 @@ def desired_state_shard_config() -> DesiredStateShardConfig:
     return DesiredStateShardConfig(
         shard_arg_name="cluster_name",
         shard_arg_is_collection=True,
-        shard_path_selectors={
-            "namespaces[*].cluster.name",
-            "resources[*].cluster",
-        },
+        shard_path_selectors={"state.*.shard"},
```
patjlm, line 1370: Can this be even `{"state.*"}` given the state key is the shard name/id? Then we would not even need teh `"shard"` key in the value. The value could  even be directly the sha?
```
return {
        "state": {
            cluster: DeepHash(state).get(state)
            for cluster, state in state_for_clusters.items()
        }
    }
```
Or would the diff mechanism not work with that?

geoberle: I tried that but with jsonpath there is no way to match and extract on a dict key, only values. 

But in the end it does not matter. I see a simplification refactor upcoming where we can get rid of jsonpath altogether by coming up with a standard to structure the desired state.

patjlm (approved):

rporres:
I arrive late to the party, but this looks great. Thx!

PR #3785, opened 2023-09-08
[APPSRE-8289] update tekton template for new version

BumbleFeng:


patjlm (approved):
Please ensure this will be compatible on all clusters, including those not yet upgraded with the latest openshift pipelines, and fedramp

BumbleFeng:
fedramp has updated openshift pipelines to 1.10 now

PR #3786, opened 2023-09-11
[glitchtip-project-alerts] add urlSecret attribute

chassing:
A webhook URL may contain credentials, e.g. slack webhooks. Add an `urlSecret` attribute to retrieve a URL via Vault as well.

Depends on: https://github.com/app-sre/qontract-schemas/pull/514
Ticket: [[glitchtip project-alerts] Webhook URL from vault](https://issues.redhat.com/browse/APPSRE-8288)

### DIFF
--- a/reconcile/glitchtip_project_alerts/integration.py
+++ b/reconcile/glitchtip_project_alerts/integration.py
@@ -14,6 +14,7 @@ from reconcile.gql_definitions.glitchtip.glitchtip_instance import (
 )
 from reconcile.gql_definitions.glitchtip_project_alerts.glitchtip_project import (
     GlitchtipProjectAlertRecipientEmailV1,
+    GlitchtipProjectAlertRecipientV1,
     GlitchtipProjectAlertRecipientWebhookV1,
     GlitchtipProjectsV1,
 )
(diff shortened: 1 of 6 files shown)


reconcile/glitchtip_project_alerts/integration.py
```diff
@@ -69,35 +69,44 @@ def fetch_desired_state(
...
+                        if not url:
+                            raise ValueError(
+                                "url or urlSecret must be set for webhook recipient"
+                            )
+                        project_alert_recipients.append(
+                            ProjectAlertRecipient(
+                                recipient_type=RecipientType.WEBHOOK,
+                                url=url,
+                            )
+                        )
```
hemslo, line 93: Since this become a little complex now, how about extract a method for it?

```python
    def build_project_alert_recipient(self, recipient):
        match recipient:
            case GlitchtipProjectAlertRecipientEmailV1():
                return ProjectAlertRecipient(recipient_type=RecipientType.EMAIL)
            case GlitchtipProjectAlertRecipientWebhookV1(
                url=url,
                url_secret=url_secret,
            ):
                if not url and url_secret:
                    url = self.secret_reader.read_secret(url_secret)
                if not url:
                    raise ValueError(
                        "url or urlSecret must be set for webhook recipient"
                    )
                return ProjectAlertRecipient(
                    recipient_type=RecipientType.WEBHOOK,
                    url=url,
                )
```

Then it can be used similar as before:

```python
            alerts = [
                ProjectAlert(
                    name=alert.name,
                    timespan_minutes=alert.timespan_minutes,
                    quantity=alert.quantity,
                    recipients=list(
                        map(self.build_project_alert_recipient, alert.recipients)
                    ),
                )
                for alert in glitchtip_project.alerts or []
            ]
```

rporres (approved):
@hemslo suggestion looks reasonable. LGTM otherwise.

reconcile/glitchtip_project_alerts/integration.py
chassing: Thanks for your suggestion. I applied it with [f5c5559](https://github.com/app-sre/qontract-reconcile/pull/3786/commits/f5c5559af4c0c45173ee253605aa7d90c4d479d7)

hemslo (approved):
LGTM!

PR #3790, opened 2023-09-13
add all_callers to tf-r

BumbleFeng:


reconcile/terraform_resources.py
```diff
@@ -201,16 +201,20 @@ def setup(
...
 ) -> tuple[
-    ResourceInventory, Optional[OCMap], Terraform, ExternalResourceSpecInventory
+    ResourceInventory,
+    Optional[OCMap],
+    Terraform,
+    ExternalResourceSpecInventory,
+    Iterable[str],
 ]:
     accounts = queries.get_aws_accounts(terraform_state=True)
+    all_accounts = tuple(a["name"] for a in accounts)
```
hemslo, line 211: `account_names` is used in caller check in https://github.com/app-sre/qontract-reconcile/pull/3790/files#diff-0e9f20c0944ab8088807c726d964c2df24b5c33b873c4c4ef1a4a78c5fe86517R232-R234, bug or expected?

reconcile/terraform_resources.py
```diff
@@ -201,16 +201,20 @@ def setup(
...
     if not include_accounts and exclude_accounts:
         excluding = filter_accounts_by_name(accounts, exclude_accounts)
         validate_account_names(excluding, exclude_accounts)
         accounts = exclude_accounts_by_name(accounts, exclude_accounts)
         if len(accounts) == 0:
             raise ValueError("You have excluded all aws accounts, verify your input")
-        account_names = tuple(ac["name"] for ac in accounts)
```
BumbleFeng, line 213: redundant with https://github.com/app-sre/qontract-reconcile/pull/3790/files#diff-0e9f20c0944ab8088807c726d964c2df24b5c33b873c4c4ef1a4a78c5fe86517R221

reconcile/terraform_resources.py
BumbleFeng: I think you mean https://github.com/app-sre/qontract-reconcile/pull/3790#discussion_r1323817962

hemslo: not only that, but here https://github.com/app-sre/qontract-reconcile/pull/3790/files#diff-0e9f20c0944ab8088807c726d964c2df24b5c33b873c4c4ef1a4a78c5fe86517L124, not sure if it just want to filter current account names, or check all account names

BumbleFeng: ah, I see. Then this pr would not actually work as expected. If the current resources do not match with the caller, they will never be added to ResourceInventory. https://github.com/BumbleFeng/qontract-reconcile/blob/234229993e1a433a2b27d40681e4f62dfdbfba86/reconcile/terraform_resources.py#L96-L99

BumbleFeng: IIRC, the original idea was to reduce the current state items. Unlike openshift-saas-deploy deploys resources in a single namespace, tf-r applies secrets across different namespaces. the secrets may also come for other integrations. Filter secrets by the caller(account names) can keep the current state items as little as necessary.

PR #3793, opened 2023-09-13
DORA LTTC collector

jmelis:
As part of the Services Dashboard effort, the DORA LTTC (lead time to change metrics) must be collected.

Definitions:
- `app_name` is the value of `.name` in the saas-file.
- `env_name` is the value of a given `.resourceTemplates.targets.namespace.$ref -> ns file -> environment file -> .name`.
- `trigger_reason` is one of the fields registered in DeploymentDB that identifies the promotion event. It looks like: `https://gitlab.../service/app-interface/commit/<sha>`.
- `promotion_sha` is the `sha` contained in the `trigger_reason`.

The purpose of this MR is to add a new DashdotDB collector that does the following:

* Get a list of saasfiles that contain the `dora` field in its `labels`. The value is a comma-separated list of target `env_name`s. Relevant [MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/79454).
* Query DashdotDB (over API) to obtain latest registered entry for that `app_name` and `env_name` pair.
* Query DeploymentDB (over SQL) to obtain the deployments associated to each `app_name` and `env_name` pair from the latest registered entry in DashdotDB or from the last 90 days if there are none.
* From the DeploymentDB entry `trigger_reason`, extract the changes: url of the repository, previous commit, promotion commit. This is done by fetching the saas-file from GitLab's API at at the `promotion_sha` commit, and comparing it with the `promotion_sha^` commit (note the `^`).
* Using GitHub or GitLab's API, obtain the list of commits between `promotion_sha^` and `promotion_sha`.
* Build the payload for DashdotDB and submit it as a `POST`.

Ticket: [OSDEV-1255](https://issues.redhat.com/browse/OSDEV-1255)

### DIFF
--- a/reconcile/dashdotdb_dora.py
+++ b/reconcile/dashdotdb_dora.py
@@ -9,7 +9,10 @@ from datetime import (
 )
 from typing import (
     Any,
+    Iterable,
+    Mapping,
     Optional,
+    Self,
 )
 
 import requests
@@ -99,16 +102,14 @@ class DeploymentDB:
         password: str,
     ):
         self.conn = connect(
-            "host={} port={} dbname={} user={} password={}".format(
-                host, port, name, user, password
-            ),
+            f"host={host} port={port} dbname={name} user={user} password={password}"
         )
         LOG.info("Connected to DeploymentDB")
 
     def deployments(
         self,
         trigger_reason_pattern: str,
-        app_env_since_list: list[tuple[AppEnv, datetime]],
+        app_env_since_list: Iterable[tuple[AppEnv, datetime]],
     ) -> list[tuple[AppEnv, Deployment]]:
         deployments = []
         with self.conn.cursor() as cur:
@@ -253,7 +254,7 @@ class DashdotdbDORA(DashdotdbBase):
         self.gh_token = secret_reader.read(gh_instance["token"])
         self._gh_apis: dict[str, GithubRepositoryApi] = {}
 
-    def __enter__(self) -> "DashdotdbDORA":
+    def __enter__(self) -> Self:
         return self
 
     def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
@@ -471,7 +472,7 @@ class DashdotdbDORA(DashdotdbBase):
             for commit in self.gh_api(repo).compare(rc.ref_from, rc.ref_to)
         ]
 
-    def post(self, data: dict[str, Any]) -> None:
+    def post(self, data: Mapping[str, Any]) -> None:
         endpoint = f"{self.dashdotdb_url}/api/v1/dora"
         response = self._do_post(endpoint, data)
         try:
@@ -488,24 +489,6 @@ class DashdotdbDORA(DashdotdbBase):
                 "Successfully posted to DashdotDB %s deployments.",
                 len(data["deployments"]),
             )
-
-    @staticmethod
-    def get_saas_file_summary(
-        saas_file: dict[str, Any]
-    ) -> dict[SummaryKey, SummaryEntry]:
-        summary: dict[SummaryKey, SummaryEntry] = {}
-
-        for rt in saas_file["resourceTemplates"]:
-            repo_url = rt["url"]
-            for target in rt["targets"]:
-                rt_name = rt["name"]
-                target_ns = target["namespace"]["$ref"]
-                target_ref = target["ref"]
-
-                summary[SummaryKey(rt_name, target_ns)] = SummaryEntry(
-                    repo_url, target_ref
-                )
-        return summary
 
     @staticmethod
     def build_deployment_post(
--- a/reconcile/dashdotdb_dora.py
+++ b/reconcile/dashdotdb_dora.py
@@ -102,7 +102,7 @@ class DeploymentDB:
         password: str,
     ):
         self.conn = connect(
-            f"host={host} port={port} dbname={name} user={user} password={password}"
+            host=host, port=port, dbname=name, user=user, password=password
         )
         LOG.info("Connected to DeploymentDB")
 
--- a/reconcile/dashdotdb_dora.py
+++ b/reconcile/dashdotdb_dora.py
@@ -1,0 +1,538 @@
+import functools
+import os
+from collections import defaultdict
+from dataclasses import dataclass
+from datetime import (
+    datetime,
+    timedelta,
+    timezone,
+)
+from typing import (
+    Any,
+    Optional,
+)
+
+import requests
+import yaml
+from psycopg2 import (
+    connect,
+    sql,
+)
+from sretoolbox.utils import threaded
+
+from reconcile import queries
+from reconcile.dashdotdb_base import (
+    LOG,
+    DashdotdbBase,
+)
+from reconcile.typed_queries.app_interface_vault_settings import (
+    get_app_interface_vault_settings,
+)
+from reconcile.typed_queries.saas_files import get_saas_files
+from reconcile.utils.github_api import GithubRepositoryApi
+from reconcile.utils.gitlab_api import GitLabApi
+from reconcile.utils.secret_reader import create_secret_reader
+
+QONTRACT_INTEGRATION = "dashdotdb-dora"
+
+
+@dataclass
+class DeploymentDBSecret:
+    path: str
+    field: str
+    q_format: Optional[str]
+    version: Optional[int]
+
+
+@dataclass(eq=True, frozen=True)
+class AppEnv:
+    app_name: str
+    env_name: str
+
+
+@dataclass
+class Deployment:
+    trigger_reason: str
+    deployment_time: datetime
+
+
+@dataclass(eq=True, frozen=True)
+class SaasTarget:
+    app_name: str
+    env_name: str
+    path: str
+    resource_template: str
+    namespace: str
+    pipeline: str
+
+    @property
+    def app_env(self) -> AppEnv:
+        return AppEnv(self.app_name, self.env_name)
+
+
+@dataclass(eq=True, frozen=True)
+class RepoChanges:
+    repo_url: Optional[str]
+    ref_from: Optional[str]
+    ref_to: Optional[str]
+
+
+@dataclass(eq=True, frozen=True)
+class SummaryKey:
+    rt_name: str
+    target_ns: str
+
+
+@dataclass(eq=True, frozen=True)
+class SummaryEntry:
+    repo_url: str
+    target_ref: str
+
+
+class DeploymentDB:
+    def __init__(
+        self,
+        host: str,
+        port: str,
+        name: str,
+        user: str,
+        password: str,
+    ):
+        self.conn = connect(
+            "host={} port={} dbname={} user={} password={}".format(
+                host, port, name, user, password
+            ),
+        )
+        LOG.info("Connected to DeploymentDB")
+
+    def deployments(
+        self,
+        trigger_reason_pattern: str,
+        app_env_since_list: list[tuple[AppEnv, datetime]],
+    ) -> list[tuple[AppEnv, Deployment]]:
+        deployments = []
+        with self.conn.cursor() as cur:
+            subq = [
+                sql.SQL(
+                    "(app_name = {} AND env_name = {} AND deployment_time > {})"
+                ).format(
+                    sql.Literal(app_env.app_name),
+                    sql.Literal(app_env.env_name),
+                    sql.Literal(since),
+                )
+                for (app_env, since) in app_env_since_list
+            ]
+
+            query = sql.SQL(
+                """
+                    SELECT app_name, env_name, trigger_reason, deployment_time
+                    FROM deployments
+                    WHERE
+                        succeeded = True
+                        AND trigger_reason LIKE {}
+                        AND ({})
+                    ORDER BY deployment_time ASC;
+                """
+            ).format(sql.Literal(trigger_reason_pattern), sql.SQL(" OR ").join(subq))
+
+            cur.execute(query)
+
+            for record in cur:
+                deployments.append(
+                    (AppEnv(record[0], record[1]), Deployment(record[2], record[3]))
+                )
+
+        if not deployments:
+            LOG.info("No deployments found")
+
+        return deployments
+
+
+@dataclass
+class Commit:
+    repo: str
+    sha: str
+    date: datetime
+
+    def lttc(self, finish_timestamp: datetime) -> int:
+        commit_date_tzaware = self.date
+        finish_timestamp_tzaware = finish_timestamp
+
+        if commit_date_tzaware.tzinfo is None:
+            commit_date_tzaware = commit_date_tzaware.replace(tzinfo=timezone.utc)
+
+        if finish_timestamp_tzaware.tzinfo is None:
+            finish_timestamp_tzaware = finish_timestamp_tzaware.replace(
+                tzinfo=timezone.utc
+            )
+
+        return int((finish_timestamp_tzaware - commit_date_tzaware).total_seconds())
+
+
+class DashdotdbDORA(DashdotdbBase):
+    """DashdotDB DORA collector.
+
+    Definitions:
+
+    * `app_name` is the value of `.name` in the saas-file.
+    * `env_name` is the value of a given
+      `.resourceTemplates.targets.namespace.$ref -> ns file -> environment file
+      -> .name`.
+    * `trigger_reason` is one of the fields registered in DeploymentDB that
+      identifies the promotion event. It looks like:
+      https://gitlab.../service/app-interface/commit/<sha>.
+    * `promotion_sha` is the sha contained in the trigger_reason.
+
+    This collector performs the followig tasks:
+
+    * Get a list of saasfiles that contain the dora field in its labels. The
+      value is a comma-separated list of target env_names. Relevant MR.
+    * Query DashdotDB (over API) to obtain latest registered entry for that
+      app_name and env_name pair.
+    * Query DeploymentDB (over SQL) to obtain the deployments associated to each
+      app_name and env_name pair from the latest registered entry in DashdotDB
+      or from the last 90 days if there are none.
+    * From the DeploymentDB entry trigger_reason, extract the changes: url of
+      the repository, previous commit, promotion commit. This is done by
+      fetching the saas-file from GitLab's API at at the promotion_sha commit,
+      and comparing it with the promotion_sha^ commit (note the ^).
+    * Using GitHub or GitLab's API, obtain the list of commits between
+      promotion_sha^ and promotion_sha.
+    * Build the payload for DashdotDB and submit it as a POST.
+
+    Caveats:
+
+    * This class has been designed to run as a cronjob, so the cleanup code
+      should be reviewed if this changes its execution pattern to a service.
+    """
+
+    def __init__(
+        self, dry_run: bool, gitlab_project_id: str, thread_pool_size: int = 5
+    ) -> None:
+        self.gitlab_project_id = gitlab_project_id
+        self.settings = queries.get_app_interface_settings()
+        vault_settings = get_app_interface_vault_settings()
+        secret_reader = create_secret_reader(use_vault=vault_settings.vault)
+
+        # init dashdotdb
+        super().__init__(
+            dry_run=dry_run,
+            thread_pool_size=thread_pool_size,
+            marker="DDDB_DORA:",
+            scope="dora",
+            secret_reader=secret_reader,
+        )
+
+        # init dep db
+        depdb_secret = secret_reader.read_all_secret(
+            DeploymentDBSecret(
+                field="",
+                path=os.environ["DEPLOYMENT_DB_SECRET"],
+                q_format=None,
+                version=None,
+            )
+        )
+
+        self.dep_db = DeploymentDB(
+            depdb_secret["db.host"],
+            depdb_secret["db.port"],
+            depdb_secret["db.name"],
+            depdb_secret["db.user"],
+            depdb_secret["db.password"],
+        )
+
+        # init GitLab API
+        gl_instance = queries.get_gitlab_instance()
+        self.gl = GitLabApi(
+            gl_instance, project_id=self.gitlab_project_id, settings=self.settings
+        )
+        self.gl_app_interface_get_file = functools.cache(self.gl.get_file)
+
+        # init GitHub API
+        gh_instance = queries.get_github_instance()
+        self.gh_token = secret_reader.read(gh_instance["token"])
+        self._gh_apis: dict[str, GithubRepositoryApi] = {}
+
+    def __enter__(self) -> "DashdotdbDORA":
+        return self
+
+    def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
+        self.gl.cleanup()
+        for gh in self._gh_apis.values():
+            gh.cleanup()
+
+    def gh_api(self, repo: str) -> GithubRepositoryApi:
+        if repo not in self._gh_apis:
+            self._gh_apis[repo] = GithubRepositoryApi(repo, self.gh_token)
+        return self._gh_apis[repo]
+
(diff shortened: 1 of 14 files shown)


reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,480 @@
...
+        gh_instance = queries.get_github_instance()
+        self.gh_token = secret_reader.read(gh_instance["token"])
+
+        # init app_interface_get_file
+        gl_instance = queries.get_gitlab_instance()
+        gl = GitLabApi(
+            gl_instance, project_id=13582, settings=settings
+        )  # 13582 is app-interface
+
+        self.app_interface_get_file = functools.cache(gl.get_file)
```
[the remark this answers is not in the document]
jmelis: I think this is not applicable because these DashdotDB reconcilers run on a cron expression.

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,480 @@
...
+
+        # this is the RAW GitLab API, not our wrapper GitLabApi
+        self.gl = gl.gl
```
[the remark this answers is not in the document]
jmelis: I think this is also not applicable for integrations like this one that run on a cron.

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,480 @@
...
+        elif rc.repo_url.startswith("https://gitlab.cee.redhat.com"):
+            commits = self._gitlab_compare_commits(rc)
+        else:
+            raise Exception(f"Unknown git hosting {rc.repo_url}")
+
+        return rc, commits
+
+    def get_latest(self, app_env: AppEnv) -> Optional[datetime]:
+        endpoint = f"{self.dashdotdb_url}/api/v1/dora/latest"
+        response = self._do_get(
```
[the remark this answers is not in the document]
jmelis: ack, resolving.

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,480 @@
...
+            "env_name": saastarget.env_name,
+            "pipeline": saastarget.pipeline,
+            "trigger_reason": deployment.trigger_reason,
+            "finish_timestamp": str(deployment.deployment_time),
+            "commits": [
+                {
+                    "repo": c.repo,
+                    "revision": c.sha,
+                    "timestamp": str(c.date),
+                    "lttc": c.lttc(deployment.deployment_time),
```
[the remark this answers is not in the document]
jmelis: this is in seconds, which is correctly returned by the `lttc` method ([this processes it](https://github.com/app-sre/dashdotdb/blob/master/dashdotdb/services/dora.py#L69)). But I will fix the `finish_timestamp` on top.

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,480 @@
...
+                sql = """
+                        SELECT trigger_reason, deployment_time
+                        FROM deployments
+                        WHERE
+                            app_name = %s
+                            AND env_name = %s
+                            AND succeeded = True
+                            AND trigger_reason LIKE 'https://gitlab.cee.redhat.com/service/app-interface/commit/%%'
+                            AND deployment_time > %s::timestamp
+                        ORDER BY deployment_time ASC LIMIT %s OFFSET %s;
```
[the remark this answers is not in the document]
geoberle: if the concern is a big query result you could go for https://www.psycopg.org/docs/usage.html#server-side-cursors

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,480 @@
...
+    def run(self) -> None:
+        saastargets = self.get_saastargets()
+
+        # Build unique combination of app_name and env_names
+        # so we can fetch the deployments associated with each one.
+        # We are doing this so we don't fetch the deployments
+        # from the DB for a unique (app_name, env_name) multiple times.
+        app_envs = {s.app_env for s in saastargets}
+
+        app_env_deployments_list: list[tuple[AppEnv, list[Deployment]]] = threaded.run(
```
geoberle, line 212: would it be an option to fetch all app/env deployments in one query? did you experience some perf issues to go for a threaded approach?

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,480 @@
...
+            for commit in result["commits"]
+        ]
+
+    def _github_compare_commits(self, rc: RepoChanges) -> list[Commit]:
+        if not rc.repo_url:
+            return []
+
+        prefix = "https://github.com/"
+        repo = rc.repo_url[rc.repo_url.startswith(prefix) and len(prefix) :]
+        gh = GithubRepositoryApi(repo, self.gh_token)
```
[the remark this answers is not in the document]
jmelis: Note that the `repo` argument is variable. I would need to build a dict per `repo`. Is it worth it considering this integration runs on a cron?

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,480 @@
...
+
+        saastarget_dep_repo_changes: list[
+            tuple[SaasTarget, Deployment, RepoChanges]
+        ] = threaded.run(
+            func=self.get_repo_changes,
+            iterable=repo_changes_iterable,
+            thread_pool_size=self.thread_pool_size,
+        )
+
+        # filter invalid ones
```
geoberle, line 234: can you add some detail what an invalid one is? i can read the condition but would still be nice for future readers/maintainers

reconcile/dashdotdb_dora.py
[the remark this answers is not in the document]
jmelis: It's true that the initialization is not passive. I will fix this.

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,480 @@
...
+    def get_saastargets(self) -> list[SaasTarget]:
+        targets = []
+        for saas_file in get_saas_files():
+            if not saas_file.labels:
+                continue
+
+            if "dora" not in saas_file.labels:
+                continue
+
+            dora_target_env_names = saas_file.labels["dora"].split(",")
```
geoberle, line 275: nit: potentially get rid of leading, trailing whitespaces after the split

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,480 @@
...
+        return targets
+
+    def get_app_env_deployments(
+        self, app_env: AppEnv
+    ) -> tuple[AppEnv, list[Deployment]]:
+        since = self.get_latest(app_env)
+
+        if not since:
+            # grab last 90 days
+            since = self.start_timestamp - timedelta(days=90)
```
geoberle, line 306: reading deployments from the last 90 days means that deployments are reprocessed for quite some time. what was the thought process for this value?

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,480 @@
...
+
+    def compare(self, rc: RepoChanges) -> tuple[RepoChanges, list[Commit]]:
+        if not rc.repo_url:
+            return rc, []
+
+        LOG.info("Fetching commits %s", rc)
+        if rc.repo_url.startswith("https://github.com"):
+            commits = self._github_compare_commits(rc)
+        elif rc.repo_url.startswith("https://gitlab.cee.redhat.com"):
+            commits = self._gitlab_compare_commits(rc)
```
geoberle, line 361: i'm not entirely sold on mentioning gitlab.cee explicitely here but i have no ideas that are considerably better :(

reconcile/dashdotdb_dora.py
geoberle: sry i misunderstood. this is basically just for an initial import and 90d is for sure fine with that

setup.py
```diff
@@ -56,6 +56,7 @@
         "pydantic~=1.10.6",
         "MarkupSafe==2.1.1",
         "filetype~=1.2.0",
+        "psycopg2-binary~=2.9",
```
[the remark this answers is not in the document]
geoberle: https://github.com/app-sre/dashdotdb/pull/79

reconcile/dashdotdb_dora.py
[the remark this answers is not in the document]
hemslo: An easy fix is to use instance level cache, as the runtime may change from a cronjob to a long running deployment.

```python
self.app_interface_get_file = functools.cache()(gl.get_file)
```

reconcile/dashdotdb_dora.py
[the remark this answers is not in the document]
hemslo: Similar concern to https://github.com/app-sre/qontract-reconcile/pull/3793#discussion_r1329876129, it's an easy fix to avoid leak then we don't need to worry about change it from cronjob to a deployment later.

reconcile/dashdotdb_dora.py
jmelis: good point, I think it has an easy fix

reconcile/dashdotdb_dora.py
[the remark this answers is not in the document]
jmelis: I don't understand `functools.cache()(gl.get_file)`, it also seems to fail:

```
>>> import functools
>>> functools.cache()
TypeError: cache() missing 1 required positional argument: 'user_function'
>>> def a(s):
...     return s+"!"
...
>>> functools.cache()(a)
TypeError: cache() missing 1 required positional argument: 'user_function'
```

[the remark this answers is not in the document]
hemslo: oh, sorry, the signature is a little different than `lru_cache`, it' OK, your usage is already good!

source: https://github.com/python/cpython/blob/afa7b0d743d9b7172d58d1e4c28da8324b9be2eb/Lib/functools.py#L652

```python
def cache(user_function, /):
    'Simple lightweight unbounded cache.  Sometimes called "memoize".'
    return lru_cache(maxsize=None)(user_function)
```

So it's already a instance level cache! Use `@cache` on instance method will cause memory leak because the scope is different, use normal function call is good here.

reconcile/dashdotdb_dora.py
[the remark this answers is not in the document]
jmelis: implemented with server side cursor

reconcile/dashdotdb_dora.py
jmelis: done! I just didn't think of it, it makes no sense as I wrote it. Thank you for raising this :)

reconcile/dashdotdb_dora.py
geoberle: thank you for `rc.repo_url.startswith(self.gl.server)`!

reconcile/dashdotdb_dora.py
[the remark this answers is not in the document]
hemslo: @jmelis it seems we can reuse the same `Github` object to fetch different repos, `GithubRepositoryApi` just for single repo usage, if we design `GithubRepositoryApi` to accept `Github` object instead of creating it inside `__init__`, then it's very easy to support your use case here. Or you can simply init one `Github` instance, use it to fetch repos.

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,503 @@
...
+                """
+                    SELECT app_name, env_name, trigger_reason, deployment_time
+                    FROM deployments
+                    WHERE
+                        succeeded = True
+                        AND trigger_reason LIKE {}
+                        AND ({})
+                    ORDER BY deployment_time ASC;
+                """
+            ).format(sql.Literal(trigger_reason_pattern), sql.SQL(" OR ").join(subq))
```
hemslo, line 112: This query is a little complex, and I'm not sure if this can work with large dataset as this query has too many conditions, hard to use index.

I'm feeling we are missing a domain concept here, the logic looks like 

```
Return all app-sre managed unregistered deployment info
```

I'm not sure if deployments table contains any non app-sre managed deployment info, if none, then we don't need that filter. If yes, then maybe a new column `managed_by` value `app-sre` can simplify this query.

There is another filter about `deployment_time`, if we can introduce another column `registered_time`, populate it after register in dashdotdb, then the whole query will be like this:


```python
                    SELECT app_name, env_name, trigger_reason, deployment_time
                    FROM deployments
                    WHERE
                        succeeded = True
                        AND trigger_reason LIKE {}
                        AND managed_by = {}
                        AND registered_time is NULL
                    ORDER BY deployment_time ASC;
```

Index on `managed_by, registered_time` will speed up this query.

For first time migration, if we want to skip old deployments, can populate `registered_time` to a value or not set `managed_by`.

This is based on we control deployments database scheme, if it's out of our control, then the complex query is needed.

jmelis: This is probably premature optimization. The query takes `0:00:00.592230` on the first run, and `0:00:00.252267` in subsequent ones, so I think this is reasonable for now. If it's OK with you I'd prefer not to solve this problem at this stage.

reconcile/dashdotdb_dora.py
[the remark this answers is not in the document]
jmelis: Is it OK if I add a note in the code and not solve this? I don't care about performance that much, and the memory usage will not skyrocket.

reconcile/dashdotdb_dora.py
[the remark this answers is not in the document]
jmelis: Fixed

reconcile/dashdotdb_dora.py
hemslo: 250ms is a very slow query in db, if this feature is not designed to scale, then I'm ok with this.

reconcile/dashdotdb_dora.py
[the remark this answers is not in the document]
hemslo: sure, need to note the runtime of this integration is a cronjob, resource cleanup is not done via code, but the termination of the process. I may include this as part of github cleanup once [pygithub 2.0.2rc1](https://github.com/PyGithub/PyGithub/milestone/44) released.

reconcile/dashdotdb_dora.py
jmelis: Thanks. I agree 250ms is not ideal, but the runtime here is not important. We are going to run this once per day.

reconcile/dashdotdb_dora.py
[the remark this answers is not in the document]
jmelis: I added the info as a class documentation. Resolving this. Please reopen if you disagree.

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,480 @@
...
+    target_ns: str
+
+
+@dataclass(eq=True, frozen=True)
+class SummaryEntry:
+    repo_url: str
+    target_ref: str
+
+
+class DashdotdbDORA(DashdotdbBase):
```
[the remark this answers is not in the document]
jmelis: I added the tests now.

setup.py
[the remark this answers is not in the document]
jmelis: @hemslo what do I need to do with this in the end? should I simply `s/psycopg2-binary/psycopg2/' and it will work?

reconcile/dashdotdb_dora.py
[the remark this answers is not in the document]
hemslo: Thanks!

setup.py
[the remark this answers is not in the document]
hemslo: yes, the base images are updated, just need to rebase and update it to `psycopg2`.

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,538 @@
...
+    def __init__(
+        self,
+        host: str,
+        port: str,
+        name: str,
+        user: str,
+        password: str,
+    ):
+        self.conn = connect(
+            "host={} port={} dbname={} user={} password={}".format(
```
chassing, line 102: Please use the f-string syntax. 🙇

hemslo (approved):
LGTM!

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,538 @@
...
+                host, port, name, user, password
+            ),
+        )
+        LOG.info("Connected to DeploymentDB")
+
+    def deployments(
+        self,
+        trigger_reason_pattern: str,
+        app_env_since_list: list[tuple[AppEnv, datetime]],
```
chassing, line 111: Please use the generic `Iterbale` type. 
```
app_env_since_list: Iterbale[tuple[AppEnv, datetime]]
```

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,538 @@
...
+    ) -> list[tuple[AppEnv, Deployment]]:
+        deployments = []
+        with self.conn.cursor() as cur:
+            subq = [
+                sql.SQL(
+                    "(app_name = {} AND env_name = {} AND deployment_time > {})"
+                ).format(
```
chassing, line 118: The usage of an f-string would increase the readability :bow:

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,538 @@
...
+                """
+                    SELECT app_name, env_name, trigger_reason, deployment_time
+                    FROM deployments
+                    WHERE
+                        succeeded = True
+                        AND trigger_reason LIKE {}
+                        AND ({})
+                    ORDER BY deployment_time ASC;
+                """
+            ).format(sql.Literal(trigger_reason_pattern), sql.SQL(" OR ").join(subq))
```
chassing, line 136: The usage of an f-string would increase the readability :bow:

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,538 @@
...
+            gl_instance, project_id=self.gitlab_project_id, settings=self.settings
+        )
+        self.gl_app_interface_get_file = functools.cache(self.gl.get_file)
+
+        # init GitHub API
+        gh_instance = queries.get_github_instance()
+        self.gh_token = secret_reader.read(gh_instance["token"])
+        self._gh_apis: dict[str, GithubRepositoryApi] = {}
+
+    def __enter__(self) -> "DashdotdbDORA":
```
chassing, line 256: We are on Python 3.11. You can use `Self` :)

```
def __enter__(self) -> Self:
```

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,538 @@
...
+      promotion_sha^ and promotion_sha.
+    * Build the payload for DashdotDB and submit it as a POST.
+
+    Caveats:
+
+    * This class has been designed to run as a cronjob, so the cleanup code
+      should be reviewed if this changes its execution pattern to a service.
+    """
+
+    def __init__(
```
chassing, line 209: Now I know why you must mock your `DashdotdbDORA.__init__` in your test cases. Using the dependency injection pattern to decouple your class for easier testing would be best.

```python
    def __init__(
        self,
        dry_run: bool,
        gitlab_api: GitLabApi,
        github_token: str,
        thread_pool_size: int = 5,
    ) -> None:
```

The best is you don't need to query app-interface in your `__init__` method.

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,538 @@
...
+
+        prefix = "https://github.com/"
+        repo = rc.repo_url[rc.repo_url.startswith(prefix) and len(prefix) :]
+
+        return [
+            Commit(rc.repo_url, commit.sha, commit.commit.committer.date)
+            for commit in self.gh_api(repo).compare(rc.ref_from, rc.ref_to)
+        ]
+
+    def post(self, data: dict[str, Any]) -> None:
```
chassing, line 474: Please use the generic `Mapping` type:

```
data: Mapping[str, Any]
```

reconcile/dashdotdb_dora.py
```diff
@@ -0,0 +1,538 @@
...
+                response.text,
+            )
+        else:
+            LOG.info(
+                "Successfully posted to DashdotDB %s deployments.",
+                len(data["deployments"]),
+            )
+
+    @staticmethod
+    def get_saas_file_summary(
```
chassing, line 493: Is this method in use?

reconcile/test/test_dashdotdb_dora.py
```diff
@@ -0,0 +1,292 @@
...
+    Deployment,
+    RepoChanges,
+    SaasTarget,
+    SummaryEntry,
+    SummaryKey,
+)
+
+
+def test_get_repo_ref_for_sha(mocker: MockerFixture):
+    mocker.patch("reconcile.dashdotdb_dora.DashdotdbDORA.__init__").return_value = None
```
chassing, line 20: After refactoring your `DashdotdbDORA` class, the `__init__` mocks should no longer be needed. You can use `pytest.fixture`'s to get your dependencies required, e.g., GitLabApi. I think you'll find examples in our existing tests.

reconcile/utils/github_api.py
```diff
@@ -2,12 +2,14 @@
 from pathlib import Path
 from types import TracebackType
 from typing import (
+    List,
```
chassing, line 5: Please use `list` instead

reconcile/utils/saasherder/interfaces.py
```diff
@@ -389,6 +391,7 @@ class ManagedResourceName(Protocol):
 class SaasFile(HasParameters, HasSecretParameters, Protocol):
     path: str
     name: str
+    labels: Optional[Json]
```
chassing, line 394: Please use `label: Optional[dict[str, Any]]`

reconcile/dashdotdb_dora.py
jmelis: Note that we shouldn't use f-strings here, as the format is applied on a `sql.SQL` object, which works different to `str`. It handles correct SQL escaping etc.

https://www.psycopg.org/docs/sql.html

reconcile/dashdotdb_dora.py
jmelis: same as above

reconcile/dashdotdb_dora.py
hemslo: or use kwargs if no special requirement, according to doc here https://www.psycopg.org/docs/module.html#psycopg2.connect

> The connection parameters can be specified as a [libpq connection string](https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING) using the dsn parameter:
conn = psycopg2.connect("dbname=test user=postgres password=secret")
or using a set of keyword arguments:
conn = psycopg2.connect(dbname="test", user="postgres", password="secret")

reconcile/dashdotdb_dora.py
jmelis: good catch :)

reconcile/dashdotdb_dora.py
chassing: Ah ok, for sure you can write it as an f-string with:
```
sql.SQL(
    f"(app_name = {sql.Literal(app_env.app_name)} AND env_name = {sql.Literal(app_env.env_name)} AND deployment_time > {sql.Literal(since)})"
)
```
but this isn't not readable anymore.

reconcile/dashdotdb_dora.py
chassing: :this:

setup.py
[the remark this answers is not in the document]
jmelis: done, thanks!

reconcile/dashdotdb_dora.py
jmelis: done with kwargs, I don't know why I didn't see this before! thanks!

reconcile/dashdotdb_dora.py
jmelis: Agreed, but going to leave this for the following iteration.

reconcile/dashdotdb_dora.py
jmelis: fixed

PR #3794, opened 2023-09-14
Skupper-network: Use full resource names

lechuk47:
The integration is getting `Role.authorization.openshift.io` Roles as current resources (default roles in Openshift 4.x) but it expects `Role.rbac.authorization.k8s.io`. Using full resource names ensures the integration is using the same resources


patjlm (approved):

chassing (approved):
LGTM

PR #3795, opened 2023-09-14
Add node pool upgrades

janboll:
1. Slight refactoring: Extracting upgrade policy creation, make the method shorter and easier to read
2. Move get_node_pools into cluster module
3. Adding node pool upgrade implementation

Implements [APPSRE-7617](https://issues.redhat.com/browse/APPSRE-7617)


reconcile/aus/base.py
```diff
@@ -346,6 +353,35 @@ def summarize(self) -> str:
...
+            "next_run": self.next_run,
+        }
+        create_node_pool_upgrade_policy(
+            ocm_api, self.cluster.id, self.node_pool, policy
+        )
+
+    def delete(self, ocm_api: OCMBaseClient) -> None:
+        raise NotImplementedError("NodePoolUpgradePolicy.delete() not implemented")
+
+    def summarize(self) -> str:
```
patjlm, line 375: do we need the org id in there ?

reconcile/aus/base.py
```diff
@@ -739,6 +775,55 @@ def verify_lock_should_skip(
...
+    node_pools = get_node_pools(ocm_api, spec.cluster.id)
+    if node_pools:
+        for pool in sorted(node_pools, key=lambda x: x["id"]):
+            current = get_node_pool_upgrade_policies(
+                ocm_api, spec.cluster.id, pool["id"]
+            )
+            if current:
+                continue
+            pool_version_id = pool.get("version", {}).get("id")
+            pool_version = get_version(ocm_api, pool_version_id)["raw_id"]
```
patjlm, line 808: can we avoid polling OCM continuously for this one? I guess we don't have the version `raw_id` directly in the node pool version, do we? Any caching possible?

reconcile/aus/base.py
```diff
@@ -739,6 +775,55 @@ def verify_lock_should_skip(
...
+            if semver.match(pool_version, f"<{spec.current_version}"):
+                next_schedule = verify_schedule_should_skip(spec, now)
```
patjlm, line 810: I wonder if we should really check the schedule here. this may end up having nodepools upgraded 1 day (or even more) after the control plane

reconcile/utils/ocm/clusters.py
```diff
@@ -193,3 +213,21 @@ def cluster_ready_for_app_interface() -> Filter:
...
+    api = f"/api/clusters_mgmt/v1/clusters/{cluster_id}/node_pools"
+
+    for item in ocm_api.get_paginated(api):
+        result = {k: v for k, v in item.items() if k in NODE_POOL_DESIRED_KEYS}
+        results.append(result)
+
+    return results
+
+
+def get_version(ocm_api: OCMBaseClient, version: str) -> dict[str, Any]:
```
patjlm, line 229: a lru cache would be nice here

reconcile/aus/base.py
```diff
@@ -771,6 +856,16 @@ def calculate_diff(
     now = datetime.utcnow()
     gates = get_version_gates(ocm_api)
     for spec in desired_state.specs:
+        # Upgrading node pools, only required for Hypershift clusters
```
patjlm, line 859: on the previous block, shouldn't we set mutexes when there are ongoing nodepool upgrades as well?

reconcile/aus/base.py
```diff
@@ -771,6 +856,16 @@ def calculate_diff(
...
+        if spec.cluster.is_rosa_hypershift():
```
patjlm, line 860: i think this nodepools upgrades block might need to be extracted in its own for loop: when a control plane upgrade happens, we should keep the mutex (so not allow an other control plane to take it) and run nodepool upgrades for this cluster.

reconcile/aus/base.py
janboll: Very good point. I think I have to change quite some bits in the current PR to meet this.

reconcile/aus/base.py
janboll: Makes sense, once node pool upgrades add mutexes this should be straightforward! Will refactor this

reconcile/aus/base.py
janboll: Yes, I think we should honor the maintenance window here, cause node pool upgrades might cause pod restarts. WDYT?

patjlm: I am not sure. With a weekly schedule, we'll end up having control plane upgrade one week, and 1 worker pool upgrade the following week, and repeat with the number of worker pool.

Whereas if we do not check the schedule for worker pools, we get the control plane upgrade on schedule, followed right after by each worker pool upgrades (one after the other). This looks much more like a single upgrade, on the configured schedule. WDYT?

janboll: Agreed!

reconcile/aus/base.py
```diff
@@ -394,6 +431,14 @@ def fetch_current_state(
...
                 upgrade_policy["cluster"] = spec.cluster
                 current_state.append(ControlPlaneUpgradePolicy(**upgrade_policy))
+            for node_pool in get_node_pools(ocm_api, spec.cluster.id):
+                node_upgrade_policies = get_node_pool_upgrade_policies(
+                    ocm_api, spec.cluster.id, node_pool["id"]
+                )
+                for upgrade_policy in node_upgrade_policies:
+                    upgrade_policy["cluster"] = spec.cluster
+                    upgrade_policy["node_pool"] = node_pool["id"]
+                    current_state.append(NodePoolUpgradePolicy(**upgrade_policy))
```
geoberle, line 441: * can multiple nodepool upgrades exist at the same time for a cluster?
* can a controlplane upgrade exist at the same time as nodepoolupgrades? (e.g. controlplane upgrade to X+1 while nodepools upgrade to X)

i'm asking this because i think there are some places in our code that assume there is only one AbstractUpgradePolicy object per cluster. if multiple can exist we need to revisite (not a huge problem i guess)

janboll: The diff method sets the mutex for node pool upgrades. Thus our code will not create an upgrade policy if an upgrade was scheduled already. 


Will test the following just for understanding:
- create multiple node pool upgrade policies
- schedule control plane upgrade and schedule a node pool upgrade

janboll: A cluster can have both, a control plane upgrade and a node pool upgrade scheduled

janboll: A node pool can only have one upgrade policy: 
```
  "reason": "Upgrade policy with upgrade type 'NodePool' already exists. Only a single policy of the same type is supported",
```

geoberle:
food for thought: how about we manage only control plane upgrades via our code and create an automatic upgrade policy for all node pools. it would need some investigation what CS would do in such a situation but i would assume they implemented it in a way that the automatic policy on node pools would do the right thing.

janboll:
Automatic requires a schedule. So we'd have to provide a cron expression in the API call, for cluster service to determine the next schedule window. I prefer manual scheduling for this

reconcile/aus/base.py
geoberle: do i understand correctly that our code will make sure that only one thing upgrades for a cluster at a single point in time? so either the control plane, or nodepool 1, or nodepool 2, ... ?

janboll: Indeed, node pool policies rely on the same mutex logic as control plane upgrades

reconcile/aus/base.py
```diff
@@ -739,6 +784,49 @@ def verify_lock_should_skip(
...
+        next_run=next_schedule,
+    )
+
+
+def _calulate_node_pool_diffs(
+    ocm_api: OCMBaseClient, spec: ClusterUpgradeSpec, now: datetime
+) -> Optional[UpgradePolicyHandler]:
+    node_pools = get_node_pools(ocm_api, spec.cluster.id)
+    if node_pools:
+        for pool in sorted(node_pools, key=lambda x: x["id"]):
```
hemslo, line 810: Why sort here? It seems we only want to return at most one upgrade policy, is that expected? Why not upgrade all mismatch ones?

reconcile/aus/base.py
```diff
@@ -739,6 +784,49 @@ def verify_lock_should_skip(
...
+            pool_version_id = pool.get("version", {}).get("id")
+            pool_version = get_version(ocm_api, pool_version_id)["raw_id"]
+            if semver.match(pool_version, f"<{spec.current_version}"):
+                next_schedule = (now + timedelta(minutes=6)).strftime(
```
hemslo, line 814: 6 mins seems a magic number, any special requirement for it?

janboll: Yes, ocm api requires at least 5 minutes in the future, adding an extra minute just for safety.

reconcile/aus/base.py
janboll: Right, this can actually go away now, this was from when I was not using mutexes. Will remove it.

reconcile/aus/base.py
```diff
@@ -739,6 +784,49 @@ def verify_lock_should_skip(
...
+        )
+    return ClusterUpgradePolicy(
+        cluster=spec.cluster,
+        version=version,
+        schedule_type="manual",
+        next_run=next_schedule,
+    )
+
+
+def _calulate_node_pool_diffs(
```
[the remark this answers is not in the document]
janboll: fixed

reconcile/aus/base.py
hemslo: oh, then maybe extract a const like `MIN_DELTA_MINUTES = 6` to make it explicit.

janboll: Good idea, done!

reconcile/aus/base.py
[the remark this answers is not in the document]
hemslo: @janboll you might missed commit rename changes in https://github.com/app-sre/qontract-reconcile/pull/3795/commits/c4a72ab4155583d15191276873cf70ecf3a902c9

geoberle (approved):
LGTM

reconcile/aus/base.py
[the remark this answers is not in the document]
janboll: done in 
72c07bb

PR #3796, opened 2023-09-14
[change-owners] place change-owners role labels from self-service roles on MRs

geoberle:
self-serviceable MRs intentionally do not show up in the review queue. but if an MR is self-serviceable and the approvers are AppSRE, we would still love to see such MRs in the review queue, so IC is aware and can review and approve.

to fix this we will:
* include MRs with the label "change-owner/show-self-serviceable-in-review-queue" in the review-queue, regardless of its self-serviceable state
* change-owners will apply additional labels to MRs based on the "change-owners-label" label of the involved self-service roles (prefixing them with `change-owners` to ensure uniqueness and avoid conflict)

E.g. if the following role is approver on an MR, the `change-owner/show-self-serviceable-in-review-queue` label will be added the the MR.

```yaml
---
$schema: /access/role-1.yml
labels:
  change-owners-labels: show-self-serviceable-in-review-queue
```

the qontract-cli app-interface-review-queue command will react to this label and show respective MRs in the review queue report regardless of their self-service state

part of https://issues.redhat.com/browse/APPSRE-8322

### DIFF
--- a/reconcile/change_owners/change_owners.py
+++ b/reconcile/change_owners/change_owners.py
@@ -40,6 +40,7 @@ from reconcile.utils.mr.labels import (
     HOLD,
     NOT_SELF_SERVICEABLE,
     SELF_SERVICEABLE,
+    change_owner_label,
     prioritized_approval_label,
 )
 from reconcile.utils.output import format_table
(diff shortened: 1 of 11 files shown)


geoberle:
> * can this be used by tenants to set additional labels for themselves? I am thinking, maybe for insights..

yes that is a possible usecase for this feature as well

> * is there a risk of conflict with other regular labels. I suppose no, that to the `change-owner` prefix, but checking

yes the `change-owner/` prefix is meant to avoid conflicts with other existing labels that would affect our workflows, e.g.

```yaml
---
$schema: /access/role-1.yml
labels:
  change-owners-labels: lgtm
```

rporres (approved):
So cool ❤️

patjlm (approved):

PR #3798, opened 2023-09-15
Introduce mutation controls to the 3way-diff strategy

lechuk47:
* Code reorg: 3-way diff strategy code + tests have been moved to their own files 
* Added an extensible logic to control object mutations. 
  * For now, only cpu mutations are checked. K8s mutates cpu values. e.g: "1000m" becomes "1"

### DIFF
--- a/reconcile/utils/three_way_diff_strategy.py
+++ b/reconcile/utils/three_way_diff_strategy.py
@@ -70,6 +70,7 @@ def normalize_object(item: OR) -> OR:
 
 
 CPU_REGEX = re.compile(r"/.*/(requests|limits)/cpu$")
+EMPTY_ENV_VALUE = re.compile(r"/.*/env/[0-9]+/value$")
 
 
 def is_cpu_mutation(current: OR, desired: OR, patch: Mapping[str, Any]) -> bool:
@@ -82,6 +83,26 @@ def is_cpu_mutation(current: OR, desired: OR, patch: Mapping[str, Any]) -> bool:
     return False
 
 
+def is_empty_env_value(current: OR, desired: OR, patch: Mapping[str, Any]) -> bool:
+    """Check if the patch is an empty env value. Empty values are removed in the
+    current object
+
+    :param current: Current object
+    :param desired: Desired object
+    :param patch: JSON patch
+    :return: True if the change is not needed, False otherwise
+    """
+    pointer = patch["path"]
+    if (
+        patch["op"] == "add"
+        and patch["value"] == ""
+        and re.match(EMPTY_ENV_VALUE, pointer)
+    ):
+        return True
+
+    return False
+
+
 def is_valid_change(current: OR, desired: OR, patch: Mapping[str, Any]) -> bool:
     # Only consider added or replaced values on the Desired object
     if patch["op"] not in ["add", "replace"]:
@@ -90,6 +111,10 @@ def is_valid_change(current: OR, desired: OR, patch: Mapping[str, Any]) -> bool:
     # Check known mutations. Replaced values can happen if values have been
     # mutated by the API server
     if is_cpu_mutation(current, desired, patch):
+        return False
+
+    # Other cases
+    if is_empty_env_value(current, desired, patch):
         return False
 
     return True
--- a/reconcile/utils/three_way_diff_strategy.py
+++ b/reconcile/utils/three_way_diff_strategy.py
@@ -1,0 +1,134 @@
+import base64
+import logging
+import re
+from collections.abc import Mapping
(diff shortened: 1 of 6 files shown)


reconcile/utils/three_way_diff_strategy.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def is_a_valid_mutation(current: OR, desired: OR, patch: Mapping[str, Any]) -> bool:
+    pointer = patch["path"]
+    if re.match(r"/.*/(requests|limits)/cpu$", pointer):
+        current_value = resolve_pointer(current.body, pointer)
+        desired_value = patch["value"]
+        return check_cpu_mutation(current_value, desired_value)
+    else:
+        return False
```
lechuk47, line 83: This is what has been introduced to control the mutations

reconcile/utils/three_way_diff_strategy.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def is_a_valid_mutation(current: OR, desired: OR, patch: Mapping[str, Any]) -> bool:
+    pointer = patch["path"]
+    if re.match(r"/.*/(requests|limits)/cpu$", pointer):
+        current_value = resolve_pointer(current.body, pointer)
+        desired_value = patch["value"]
+        return check_cpu_mutation(current_value, desired_value)
+
+    return False
```
hemslo, line 83: Is this mean all non cpu changes are invalid?

lechuk47: No. This means that the change in the patch has not been detected as a valid mutation and the value in the patch is not equal to the value pointed by the pointer in the current object. 

I added a method docstring to clarify this. If you have a batter naming or any idea I can consider it.

hemslo: Oh, I see, it's a custom equal check, since it's only used in valid changes filter, how about extract them all into one method?

```python
def is_valid_change(current: OR, desired: OR, patch: Mapping[str, Any]) -> bool:
    if patch["op"] not in ["add", "replace"]:
        return False

    pointer = patch["path"]
    if re.match(r"/.*/(requests|limits)/cpu$", pointer):
        current_value = resolve_pointer(current.body, pointer)
        desired_value = patch["value"]
        return not OR.cpu_equal(current_value, desired_value)
    return True
```

```python
    valid_changes = [
        item for item in patch.pathc if is_valid_change(current, desired, item)
    ]
```

btw that re expression can be extracted into a compiled modulo const to avoid recompile on runtime


```python
CPU_REGEX = re.compile(r"/.*/(requests|limits)/cpu$")
```

lechuk47: Yup, I think this way is more clear. I maintained the concept of mutations though as I expect to add more cases and think that being explicit in the code about them is more clear.

hemslo: Oh, I finally understand `mutation` you mean they are the same value but different forms, I thought it means changed, maybe need a better word.

hemslo: `variant`?

lechuk47: The CPU is just a case. AFAIK K8s use "mutation" for this. This code will manage all the mutations made by Mutating admission controllers (https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook)  and other controllers as well.

One case that I have in mind (and that's why I pass the desired object to the method) is related to the Skupper-managed objects. Services with Skupper annotations are deeply modified by the Skupper controller, even changing attributes in the "desired" service with Skupper-related attributes. 

For now, we are not using annotated services anywhere, but the option is there. With this approach, we can easily add more methods to control these behaviors if needed.

lechuk47: TBH, I think this is splitting hairs a little bit. I think `is_cpu_mutation` is clear enough and there is also a comment in the code saying this is going to check mutations. 

As per the old code and checking a bit the current pods, memory does not get mutated the same way (there are pods with requests/memory = 1000Mi and are not transformed into 1Gi for example).

hemslo: I also feel it's clear enough, but maybe because of the context in this thread, not sure if someone read the code the first time will get this.

I remember I saw a loop once `1.5Gi` can't compare with `1536Mi` in memory, I have to specify it in `Mi` most time, and only integer value of `Gi` can work without loop.

hemslo (approved):
LGTM!

PR #3799, opened 2023-09-15
[terraform-repo] option to ignore statefile loading errors

cubismod:
When trying to update terraform-repo in FedRamp where there are a number of different statefiles already saved in S3, I ran into many validation errors:

```
[2023-09-15 14:37:43] [ERROR] [DRY-RUN] [terraform_repo.py:get_existing_state:184] - 1 validation error for TerraformRepoV1
account -> terraformState
  field required (type=value_error.missing)
Unable to parse existing state for repo: '/app-sre-base-networking', skipping
```

This is due to the updates to the GQL syntax updates done in a previous MR: https://github.com/app-sre/qontract-reconcile/pull/3777/files#diff-07306e78e084b1597025733d9c9787506f617028ba6ebe1cb240937c39ed8f34R11

It seems that Pydantic expects the terraformState attribute to be defined, just as `null` so it's unable to parse. To fix this, I've added a CLI flag to the command instructing terraform repo to ignore statefile errors and proceed with statefile recreation. Note that this is only recreating the statefiles stored in AWS S3 for the QR integration, not the actual terraform .tfstate files. This will result in some extra `terraform apply` calls if the flag is enabled.

I'll add an SOP detailing this into App Interface. Shouldn't be an option we need to use often, just if there are breaking GQL schema changes.

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1604,14 +1604,21 @@ def ldap_groups(ctx):
     "--output-file",
     help="Specify where to place the output of the integration",
 )
+@click.option(
+    "--ignore-state-errors",
+    is_flag=True,
+    help="Instructs terraform-repo to ignore state load errors and re-create repo states",
+)
 @click.pass_context
-def terraform_repo(ctx, output_file):
+def terraform_repo(ctx, output_file, ignore_state_errors):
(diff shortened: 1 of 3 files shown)


jonmosco (approved):
LGTM

PR #3800, opened 2023-09-17
expose cluster upgrade state via metrics

geoberle:
we lack upcoming and ongoing upgrade information on the AUS dashboards.

this patch encodes such data as negative values in the `aus_cluster_version_remaining_soak_days` metric. this is safe since the metric currently has a min value of 0.

we will use
* -1 for scheduled upgrades (upgrade policy created in OCM)
* -2 for started/ongoing upgrades.

the dashboard tile for "currently soaking" has been adapted to deal with negative numbers (ignore them) and the matrix now shows the well known icons for ready, scheduled and ongoing.

part of https://issues.redhat.com/browse/APPSRE-7878

additionally the channel group of a cluster is exposed in the AUS info metric for dashboarding purposes.

part of https://issues.redhat.com/browse/APPSRE-7877

### DIFF
--- a/reconcile/utils/ocm/upgrades.py
+++ b/reconcile/utils/ocm/upgrades.py
@@ -37,11 +37,27 @@ def get_addon_upgrade_policies(
     ):
         if addon_id and policy["addon_id"] != addon_id:
             continue
-        results.append(
-            {k: v for k, v in policy.items() if k in ADDON_UPGRADE_POLICY_DESIRED_KEYS}
+        policy_data = {
+            k: v for k, v in policy.items() if k in ADDON_UPGRADE_POLICY_DESIRED_KEYS
+        }
+        policy_data["state"] = get_addon_upgrade_policy_state(
+            ocm_api, cluster_id, policy["id"]
         )
+        results.append(policy_data)
 
     return results
+
+
+def get_addon_upgrade_policy_state(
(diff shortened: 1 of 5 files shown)


reconcile/utils/ocm/upgrades.py
```diff
@@ -37,13 +37,29 @@ def get_addon_upgrade_policies(
     ):
         if addon_id and policy["addon_id"] != addon_id:
             continue
-        results.append(
-            {k: v for k, v in policy.items() if k in ADDON_UPGRADE_POLICY_DESIRED_KEYS}
+        policy_data = {
+            k: v for k, v in policy.items() if k in ADDON_UPGRADE_POLICY_DESIRED_KEYS
+        }
+        policy_data["state"] = get_addon_upgrade_policy_state(
```
patjlm, line 43: won't that do too many calls to the OCM API?

geoberle: there is no other way to get the state. at least for HCP the state is embedded in the upgradepolicy.

i ack the problem but in the end it does not matter too much for now, because not too many upgrade policies exist at the same time.

patjlm (approved):
LGTM,
just wondering if that will not end up with too many more calls to OCM API.
Also a little description text on the grafana panel would be "ausome"!

PR #3801, opened 2023-09-18
extend gabi expiration max days to 365 days

maorfr:
part of https://issues.redhat.com/browse/APPSRE-8298

the origin of the 90 days limitation is unclear.

design doc: https://docs.google.com/document/d/1rQBCGFy2rc6A9RdeqOkTEXsr0kZlRwXTer5qarabjbA

extending this to 1 year, as even compliance training are only done once per year.

we can argue over making this number parameterized, but i don't think it is worth the effort.


jmelis:
LGTM

patjlm (approved):

PR #3805, opened 2023-09-19
[AUS] highlight blocked versions on dashboard

geoberle:
if a version is blocked, the metric for remaining soak days will expose -4 instead.

additionally the remaining soak days metric will not be exposed for versions if another followup (non-blocked reachable version) with 0 soak days is already available. this way we don't clutter the dashboard with version columns that bear no real value because such versions will not be considered for upgrade. important events like currently scheduled or running upgrades, or blocks will still be exposed.

APPSRE-7878


reconcile/aus/ocm_upgrade_scheduler.py
```diff
@@ -5,6 +5,7 @@
 from reconcile.aus import base as aus
 from reconcile.aus.cluster_version_data import VersionData
 from reconcile.aus.metrics import (
+    UPGRADE_BLOCKED,
     UPGRADE_LONG_RUNNING,
```
janboll, line 9: naming consistency:
- UPGRADE_BLOCKED_METRIC_VALUE,
-  UPGRADE_LONG_RUNNING_METRIC_VALUE,

janboll (approved):

reconcile/aus/ocm_upgrade_scheduler.py
geoberle: thx. fixed

PR #3814, opened 2023-09-20
Allow restricting resources by name in saas files

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4355

in this PR we leverage the `managedResourceNames` field in saas files (currently unused) to create an experience similar to using that field in namespace files. we override that field of the namespace with the same field coming from the saas file (like we do for `managedResourceTypes`) and let the core openshift_base logic take over.

this is basically sufficient and will create an identical experience to namespace files - an exception and failure in case a desired item is not defined in `managedResourceNames`.

in saas files, differently, we want to allow the openshift template to contain other resources, and simply filter them. for that, we except `ResourceNotManagedError`. the approach of excepting and handling the error out of the core logic is following lessons learned in #1622.

### DIFF
--- a/reconcile/utils/saasherder/interfaces.py
+++ b/reconcile/utils/saasherder/interfaces.py
@@ -380,6 +380,12 @@ class SaasRole(Protocol):
 SaasPipelinesProviders = Union[SaasPipelinesProviderTekton, SaasPipelinesProvider]
 
 
+@runtime_checkable
+class ManagedResourceName(Protocol):
+    resource: str
+    resource_names: list[str]
+
+
 class SaasFile(HasParameters, HasSecretParameters, Protocol):
     path: str
     name: str
@@ -421,4 +427,8 @@ class SaasFile(HasParameters, HasSecretParameters, Protocol):
 
     @property
     def self_service_roles(self) -> Optional[Sequence[SaasRole]]:
+        ...
+
+    @property
+    def managed_resource_names(self) -> Optional[Sequence[ManagedResourceName]]:
         ...
--- a/reconcile/utils/saasherder/models.py
+++ b/reconcile/utils/saasherder/models.py
@@ -4,6 +4,7 @@ from enum import Enum
 from typing import (
     Any,
     Optional,
+    Sequence,
     Union,
 )
 
@@ -15,6 +16,7 @@ from pydantic import (
 
 from reconcile.utils.oc_connection_parameters import Cluster
 from reconcile.utils.saasherder.interfaces import (
+    ManagedResourceName,
     SaasApp,
     SaasEnvironment,
     SaasPipelinesProviders,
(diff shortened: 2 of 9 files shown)


reconcile/utils/saasherder/saasherder.py
```diff
@@ -688,9 +688,10 @@ def _collect_namespaces(self) -> list[Namespace]:
                             environment=target.namespace.environment,
                             app=target.namespace.app,
                             cluster=target.namespace.cluster,
-                            # managedResourceTypes is defined per saas_file
-                            # add it to each namespace in the current saas_file
+                            # managedResourceTypes and managedResourceNames are defined per saas_file
+                            # add them to each namespace in the current saas_file
                             managed_resource_types=saas_file.managed_resource_types,
+                            managed_resource_names=saas_file.managed_resource_names,
```
maorfr, line 694: this is the most important line of this PR. we are essentially overriding the namespace's `managedResourceNames` with what is in the saas file. this is the same approach as `managedResourceTypes`, as can be seen one line above.

reconcile/utils/saasherder/interfaces.py
```diff
@@ -380,10 +380,17 @@ class SaasRole(Protocol):
...
+class ManagedResourceName(Protocol):
+    resource: str
+    resource_names: list[str]
+
+
 class SaasFile(HasParameters, HasSecretParameters, Protocol):
     path: str
     name: str
     managed_resource_types: list[str]
+    managed_resource_names: Optional[list[ManagedResourceName]]
```
chassing, line 393: this must be written as

```
    @property
    def managed_resource_names(self) -> Optional[list[ManagedResourceName]]:
        ...
```

Yes, it's stupid and I hate it too.

reconcile/utils/saasherder/models.py
```diff
@@ -191,6 +195,7 @@ class TargetSpec:
     cluster: str
     namespace: str
     managed_resource_types: Iterable[str]
+    managed_resource_names: Optional[list[ManagedResourceName]]
```
chassing, line 198: And convert this class to a pydantic model:

```
class TargetSpec(BaseModel):
```

for a very strange reason you can't assign a `Sequence` to a `list` with a `dataclass` :(

reconcile/utils/saasherder/interfaces.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/3814/commits/ae8ecec604dc05b1544f73f920a29598df19260d

reconcile/utils/saasherder/models.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/3814/commits/ae8ecec604dc05b1544f73f920a29598df19260d

patjlm (approved):

chassing (approved):

PR #3816, opened 2023-09-20
Fix S3 -> SQS Notifications

cubismod:
Ticket: https://issues.redhat.com/browse/APPSRE-8351

A GQL change broke the `event_notifications` functionality of S3 buckets due to an alias being used in GQL. This MR removes that alias and adds a test for this functionality specifically for sending notifications to SQS from S3.

I also ran QR locally against App Interface production and noted that the output indicates that the bucket notification objects are being created now.

```
[2023-09-20 09:27:11] [INFO] [DRY-RUN] [terraform_client.py:log_plan_diff:332] - ['create', 'app-sre-stage', 'aws_s3_bucket_notification', 'a-stage-event-notifications', set()]
[2023-09-20 09:27:11] [INFO] [DRY-RUN] [terraform_client.py:log_plan_diff:332] - ['create', 'app-sre-stage', 'aws_s3_bucket_notification', 'b-stage-event-notifications', set()]
```

Slack discussion here: https://redhat-internal.slack.com/archives/GGC2A0MS8/p1695144541954949

### DIFF
--- a/reconcile/gql_definitions/terraform_resources/terraform_resources_namespaces.gql
+++ b/reconcile/gql_definitions/terraform_resources/terraform_resources_namespaces.gql
@@ -56,7 +56,7 @@ query TerraformResourcesNamespaces {
                 overrides
                 sqs_identifier
                 s3_events
-                sns_event_notifications: event_notifications {
+                event_notifications {
                     destination_type
                     destination
                     event_type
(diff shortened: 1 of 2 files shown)


OliviaHY (approved):
LGTM. Pretty sure this will fix the bug. I don't know why this was renamed in the first place but didn't see any reference of `sns_event_notifications`. Maybe @fishi0x01 can shed some light on the whys?

PR #3824, opened 2023-09-22
[APPSRE-8072] get upstream repo info for sast scan

BumbleFeng:


patjlm:
@BumbleFeng I thought SAST checking would be fully on the CI side. Isn't it the case?

If we do that during deploy as well, could we do it fully async and ignore failure (just print a warning log), maybe adding a parallel Task in the pipeline? We could also maybe put that in the trigger rather than deploy code?

BumbleFeng:
My original idea is also to do this in CI. However, [OpenScanHub](https://cov01.lab.eng.brq2.redhat.com/osh) can only access inside the VPN, so ci.ext can not access it. 
In order to trigger a scan. We need to authenticate with [Kerberos](https://source.redhat.com/personal_blogs/configure_idm_and_automate_kerberos_tickets~2) service account and use [osh-cli](https://copr.devel.redhat.com/coprs/kdudka/covscan/). So we'd better run it in a separate [container](https://gitlab.cee.redhat.com/app-sre/container-images/-/blob/master/osh-cli/Dockerfile). or we may have too much redhat internal config inside q-r.

patjlm (approved):
Trying to get this out of the way. The actual benefit of these email SAST reports is not clear to me, but we need to proceed

PR #3826, opened 2023-09-22
Update installation doc

hemslo:
Update README for new dependencies for `psycopg2` setup, also includes existing runtime tools needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,15 @@ source venv/bin/activate
 pip install --upgrade pip setuptools
 ```
 
+Install build prerequisites for [psycopg2](https://www.psycopg.org/docs/install.html#build-prerequisites):
+
+* A C compiler
+* The Python header files. They are usually installed in a package such as `python-dev` or `python3-dev`.
+* `libpq` and `pg_config`
+  * The `libpq` header files. They are usually installed in a package such as `libpq-dev`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -231,6 +231,15 @@ source venv/bin/activate
...
 ```
 
+Install build prerequisites for [psycopg2](https://www.psycopg.org/docs/install.html#build-prerequisites):
+
+* A C compiler
+* The Python header files. They are usually installed in a package such as `python-dev` or `python3-dev`.
+* `libpq` and `pg_config`
+  * The `libpq` header files. They are usually installed in a package such as `libpq-dev`.
+  * The `pg_config` program: it is usually installed by the `libpq-dev` package but sometimes it is not in a `PATH` directory.
+  * On macOS, can be installed via `brew install postgresql@15`
```
jmelis, line 241:
```suggestion
  * On macOS, can be installed via `brew install libq`. Make sure the installation path is added to your PATH, otherwise `pg_config` will not be available.
  
Note:
In macOS with M1/M2 `pip` will fail to install unless LDFLAGS reference the openssl library path. It can be fixed with `export LDFLAGS="-I/opt/homebrew/opt/openssl/include -L/opt/homebrew/opt/openssl/lib"`.
```

jmelis (approved):
LGTM (the psycopg2 part). Thanks for adding this.

patjlm (approved):
Thanks Di!

PR #3827, opened 2023-09-25
use a-i state via STS (and more ways to connect to a-i state)

geoberle:
until now, state initialization was suppored based on
* AWS account stored in app-interface
* state in app-interface-settings

with this MR, the following ways are supported

* env vars APP_INTERFACE_STATE_BUCKET, APP_INTERFACE_STATE_BUCKET_REGION, APP_INTERFACE_STATE_AWS_PROFILE and an AWS config or creds file is available
* env vars APP_INTERFACE_STATE_BUCKET, APP_INTERFACE_STATE_BUCKET_REGION, APP_INTERFACE_STATE_VAULT_SECRET (and optionally APP_INTERFACE_STATE_VAULT_SECRET_VERSION)
* env vars APP_INTERFACE_STATE_BUCKET, APP_INTERFACE_STATE_BUCKET_REGION, APP_INTERFACE_STATE_BUCKET_ACCESS_KEY_ID, APP_INTERFACE_STATE_BUCKET_ACCESS_KEY_ID
* env vars APP_INTERFACE_STATE_BUCKET, APP_INTERFACE_STATE_BUCKET_REGION and APP_INTERFACE_STATE_BUCKET_ACCOUNT if the mentioned AWS account is present in app-interface
* state settings in app-interface-settings-1.yml

especially the mechanism via `APP_INTERFACE_STATE_AWS_PROFILE` enables us to leverage STS for state access.

important: the new way to initialize state is fully compatible to the previous way. all existing env variables are honored and the new mounts and secret based env vars are set to optional so that existing q-r deployments still work without the need for change.


reconcile/test/fixtures/helm/state.yml
```diff
@@ -126,6 +126,32 @@ objects:
...
+                key: aws.secret.access.key
+                optional: true
+          - name: AWS_CONFIG_FILE
+            value: /aws/credentials
+          - name: APP_INTERFACE_STATE_AWS_PROFILE
+            valueFrom:
+              secretKeyRef:
+                name: app-interface
+                key: aws.profile.state
+                optional: true
```
patjlm, line 154: when the secret is absent, this will not set the env var, right?

maorfr (approved):

reconcile/utils/state.py
```diff
@@ -59,99 +60,179 @@ def init_state(
...
+        "APP_INTERFACE_STATE_BUCKET_SECRET_ACCESS_KEY"
+    )
+    state_bucket_vault_secret = os.environ.get("APP_INTERFACE_STATE_VAULT_SECRET")
+    state_bucket_vault_secret_version = os.environ.get(
+        "APP_INTERFACE_STATE_VAULT_SECRET_VERSION"
     )
 
+    state_bucket_aws_profile = os.environ.get("APP_INTERFACE_STATE_AWS_PROFILE")
+
+    # if an AWS config file can be found an a profile for state usage is set ...
```
hemslo, line 130: `an a profile` typo?

reconcile/utils/state.py
```diff
@@ -59,99 +60,179 @@ def init_state(
...
-    integration: str,
-    secret_reader: SecretReaderBase,
-    state_settings: AppInterfaceStateConfigurationS3V1,
-) -> "State":
+    def build_client(self) -> S3Client:
+        session = boto3.Session(profile_name=self.profile, region_name=self.region)
+        return session.client("s3")
+
+
+def aquire_state_settings(secret_reader: SecretReaderBase) -> S3StateConfiguration:
```
hemslo, line 102: `aquire` -> `acquire`

reconcile/utils/state.py
```diff
@@ -59,99 +60,179 @@ def init_state(
...
 
-    :raises StateInaccessibleException: if the bucket is missing
-    or not accessible
+    * env variables pointing to a bucket and an AWS profile
+    * env variables pointing to static credentials
+    * env variables pointing to a bucket and a vault secret for creds
+    * env variables pointing to a bucket and an AWS account from app-interface for creds
+    * state settings in app-interface-settings-1.yml
+
+    If no settings can be found, a StateInaccessibleException is raised.
```
hemslo, line 112: Can we use default boto3 credentials behaviour to let it auto load? https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html

Since this config is a global one, we can inject secrets from vault to env vars, then let boto3 auto loader do the work.

reconcile/utils/state.py
```diff
@@ -59,99 +60,179 @@ def init_state(
...
-    return State(
-        integration=integration,
-        bucket=bucket_name,
-        client=aws_api.get_session_client(session, "s3"),
+    raise StateInaccessibleException(
+        "app-interface state must be configured to use stateful integrations. "
+        "use one of the following options to provide state config: "
+        "* env vars APP_INTERFACE_STATE_BUCKET, APP_INTERFACE_STATE_BUCKET_REGION, APP_INTERFACE_STATE_AWS_PROFILE and AWS_CONFIG (hosting the requested profile) \n"
+        "* env vars APP_INTERFACE_STATE_BUCKET, APP_INTERFACE_STATE_BUCKET_REGION, APP_INTERFACE_STATE_VAULT_SECRET (and optionally APP_INTERFACE_STATE_VAULT_SECRET_VERSION) \n"
+        "* env vars APP_INTERFACE_STATE_BUCKET, APP_INTERFACE_STATE_BUCKET_REGION, APP_INTERFACE_STATE_BUCKET_ACCESS_KEY_ID, APP_INTERFACE_STATE_BUCKET_ACCESS_KEY_ID \n"
```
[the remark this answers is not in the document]
geoberle: goot catch! ty!

reconcile/utils/state.py
geoberle: fixed

reconcile/utils/state.py
geoberle: i would like us to stay in full control for now how we authenticate.
but i'm sure we need to reconsider anyways when we adopt more pieces of q-r to profile, STS etc. 
lets come back to this then.

reconcile/utils/state.py
geoberle: fixed

hemslo (approved):
LGTM!

PR #3829, opened 2023-09-26
Vault-secrets empty values are not populated

lechuk47:
As there are cases where empty data entries in Secrets are needed, let's handle all of the options the same way. 

Context:
* https://issues.redhat.com/browse/APPSRE-1448 
* https://github.com/app-sre/qontract-reconcile/pull/1144

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -60,6 +60,7 @@ from reconcile.utils.openshift_resource import (
     ResourceInventory,
     ResourceKeyExistsError,
     ResourceNotManagedError,
+    base64_encode_secret_field_value,
 )
 from reconcile.utils.runtime.integration import DesiredStateShardConfig
 from reconcile.utils.secret_reader import SecretReader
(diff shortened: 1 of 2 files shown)


reconcile/utils/openshift_resource.py
```diff
@@ -708,5 +708,5 @@ def build_secret(
 
 def base64_encode_secret_field_value(value: str) -> Optional[str]:
     if value == "":
-        return None
+        return ""
```
patjlm, line 711: why make a specific case of this? Wouldn't the base64 call below output an empty string as well?

lechuk47: Just to avoid the encode/decode for an empty string, which results in an empty string

patjlm: I think I'd just remove the test case, but it's probably a matter of personal taste :)

reconcile/openshift_resources_base.py
```diff
@@ -577,12 +577,10 @@ def fetch_provider_vault_secret(
 
     # populate data
     for k, v in raw_data.items():
-        if v == "":
-            continue
         if k.lower().endswith(QONTRACT_BASE64_SUFFIX):
             k = k[: -len(QONTRACT_BASE64_SUFFIX)]
             v = v.replace("\n", "")
-        elif v is not None:
+        elif v != "":
```
hemslo, line 583: `base64_encode_secret_field_value` function can be reused here so no duplicate logic.

same [here](https://github.com/app-sre/qontract-reconcile/pull/3829/files#diff-e1ae1861e6058b02d61980281bb71def45745bdfc82c269bfcf2a52f6be4fabdL424)

hemslo (approved):
LGTM!

PR #3830, opened 2023-09-26
[oc] try to get all managed resource names at once

maorfr:
when `managedResourceNames` is used, it is assumed we operate within a restricted namespace. this means the logic tries to get one resource at a time.

but what if we are not in a restricted namespace?

this PR introduces an attempt to get all the resources by names once, and only go one by one in case the attempt it forbidden.

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -487,8 +487,7 @@ class OCCli:  # pylint: disable=too-many-public-methods
         if resource_names:
             items = []
             try:
-                resource_cmd = cmd + resource_names
-                items_list = self._run_json(resource_cmd, allow_not_found=True)
+                items_list = self._run_json(cmd + resource_names, allow_not_found=True)
             except ForbiddenError:
                 for resource_name in resource_names:
                     resource_cmd = cmd + [resource_name]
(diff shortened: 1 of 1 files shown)


geoberle (approved):
makes sense

patjlm:
LGTM, but I don't understand when the access would be `forbidden`

reconcile/utils/oc.py
```diff
@@ -486,12 +486,15 @@ def get_items(self, kind, **kwargs):
...
-                item = self._run_json(resource_cmd, allow_not_found=True)
-                if item:
-                    items.append(item)
-            items_list = {"items": items}
+            try:
+                items_list = self._run_json(cmd + resource_names, allow_not_found=True)
+            except ForbiddenError:
+                for resource_name in resource_names:
+                    resource_cmd = cmd + [resource_name]
+                    item = self._run_json(resource_cmd, allow_not_found=True)
```
hemslo, line 494: if trace the requests like 

```
oc get pods -o json -n some-namespace a b --v=8
```

Then it shows multiple `GET` requests to individual resources, is there a case that some requests are forbidden but some pass?

maorfr: here is an example we used to have: https://github.com/openshift/managed-cluster-config/commit/bf2fc7679e4975a31b43aa5660d6115461784300

hemslo: When there is no permission to get some resource, will it also throw `ForbiddenError` when get that individual one back?
Can we add a warning log here so we know there can be a case that enter `except ForbiddenError` then processed normally?

PR #3831, opened 2023-09-26
Fix openshift-namespaces cluster and namespace options

rporres:
They're currently broken because cli.py passes a tuple from `--cluster-name` and `--namespace-name` options. Since this is something that is taken into consideration by other integrations, this patch changes the interface of `filter_namespaces_by_cluster_and_namespace` as it is only currently used by `openshift-namespaces`.


fishi0x01 (approved):
LGTM!

PR #3833, opened 2023-09-26
Revert "Revert "[dynatrace-token-provider]Correct the Secret content …

OliviaHY:
…in SyncSet to be consistent with what's expected from the operator. (#3820)" (#3832)"

This reverts commit 9c6e589d260a1316c90a0e6a8c726996cebc7fdc.

I have found out that the failing reason is that the old syncset need to be deleted for the new format to run though successfully. Will deployed this, then manually delete the syncset, after which the new run will go through based on my test.


fishi0x01 (approved):
LGTM :rocket:

PR #3837, opened 2023-09-27
[openshift-saas-deploy-change-tester] re-use get_saas_files

maorfr:
related to https://issues.redhat.com/browse/APPSRE-8352

it was observed that runs of the integration were taking a long time. by re-using the results of the first `get_saas_files` call we avoid multiple processing of dynamic namespace selectors.

### DIFF
--- a/reconcile/openshift_saas_deploy.py
+++ b/reconcile/openshift_saas_deploy.py
@@ -99,6 +99,7 @@ def run(
     env_name: Optional[str] = None,
     trigger_integration: Optional[str] = None,
     trigger_reason: Optional[str] = None,
+    all_saas_files: Optional[list[SaasFile]] = None,
     defer: Optional[Callable] = None,
 ) -> None:
     vault_settings = get_app_interface_vault_settings()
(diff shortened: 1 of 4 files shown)


geoberle (approved):

reconcile/openshift_saas_deploy.py
```diff
@@ -99,6 +99,7 @@ def run(
     env_name: Optional[str] = None,
     trigger_integration: Optional[str] = None,
     trigger_reason: Optional[str] = None,
+    all_saas_files: Optional[list[SaasFile]] = None,
```
patjlm, line 102: shouldn't there be a `if` on line 108 below?

```
if all_saas_file is None:
  all_saas_files = get_saas_files()

patjlm: Also, could `get_saas_files(saas_file_name, env_name)` accept `all_saas_files` as a parameter?

maorfr: did the first in https://github.com/app-sre/qontract-reconcile/pull/3839

will avoid the second today :pray:

PR #3847, opened 2023-10-02
[terraform-resources] publish external resource invetory metrics

maorfr:
part of https://issues.redhat.com/browse/APPSRE-8184

this PR attempts to add a metric to describe the inventory of external resources managed by `terraform-resources`.
nice to have: make it re-usable between other integrations managing external resources.

PR #3848, opened 2023-10-02
[terraform-cloudflare-resources] publish external resource invetory metics

maorfr:
same as #3847


geoberle (approved):
i love the minimalism ❤️

PR #3740, opened 2023-08-14
APPSRE-8111 map namespace to jsonpath in saasfile renderer

fishi0x01:

PR #3850, opened 2023-10-02
APPSRE-8111 allow namespace selector in SAPM

fishi0x01:
Allow SAPM to render promotion MRs for namespace selectors. 

Our saas file logic in app-interface does not use namespace selectors at all. At query level we convert namespace selectors into concrete namespace targets. SAPM must be able to match these targets to determine if a namespace selector is eligible for a given namespace.

The core idea is to re-use `reconcile.typed_queries.saas_files.get_namespaces_by_selector` to determine if a rendered namespace falls into the targets namespace selector as defined in raw yaml.

Alternative to https://github.com/app-sre/qontract-reconcile/pull/3740


chassing (approved):
Great! ❤️

PR #3853, opened 2023-10-03
order labels and clusters when fetching with pagination

geoberle:
overall fetch results for OCM list operations can be inconsistent when paging is involved, e.g. an item that was seen on page 1 is also present on page 2 while another item might be missing.

giving the fetch result an order clause mitigates that. using ordering by creation time helps with consistency when new items are added while pagination is going on. the pagination logic is compatible with new items showing up mid pagination.


patjlm (approved):
very weird that param names are different ..

geoberle:
> very weird that param names are different ..

it is implemented individually on all endpoints

PR #3855, opened 2023-10-03
Add `eval` filter to jinja2

jmelis:
We have a use case where we want to render the slo documents into PrometheusAlerts.

The fields in the slo-documents datafile do contain Jinja2 parameters, for example:

```
expr: |
    sum(increase(aggregation:quay_request_duration_seconds_count:rate1m:sum{route="api.repository",status!~"5[0-9][0-9]",method="GET"}[{{window}}]))
    /
    sum(increase(aggregation:quay_request_duration_seconds_count:rate1m:sum{route="api.repository",method="GET"}[{{window}}]))
```

The `{{window}}` parameter must be evaluated inside the jinja2 expression:

```
{{% set expr = slo.expr | eval(window=slo.SLOParameters.window) | trim + ' <= ' + slo.SLOTarget | string %}}
```

In order for the `eval` part to work, this filter is required.

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -329,6 +329,10 @@ def hash_list(input: Iterable) -> str:
     return m.hexdigest()
 
 
+def eval_filter(input, **kwargs) -> str:
+    return jinja2.Template(input).render(**kwargs)
+
+
 def json_to_dict(input):
     """Jinja2 filter to parse JSON strings into dictionaries.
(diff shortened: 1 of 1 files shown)


geoberle (approved):
LGTM. can you add a docstring to the eval_filter func to describe what this can be used for?

jmelis:
@maorfr done: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/82855

But only because you are doing positive reinforcement

PR #3857, opened 2023-10-03
Refactor repo filtering in quay-mirror

rporres:
The previous option (--image) was not very intuitive to understand. It is clearer now with --repository-url what we mean when we want to filter repositories. This patch also includes an exclude option

part of APPSRE-8401

### DIFF
--- a/reconcile/quay_mirror.py
+++ b/reconcile/quay_mirror.py
@@ -67,7 +67,8 @@ class QuayMirror:
         control_file_dir: Optional[str] = None,
         compare_tags: Optional[bool] = None,
         compare_tags_interval: int = 86400,
-        images: Optional[Iterable[str]] = None,
+        repository_urls: Optional[Iterable[str]] = None,
+        exclude_repository_urls: Optional[Iterable[str]] = None,
     ) -> None:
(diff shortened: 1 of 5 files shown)


reconcile/quay_mirror.py
```diff
@@ -147,15 +151,21 @@ def process_repos_query(
...
 
                 for item in quay_repo["items"]:
-                    if images and item["name"] not in images:
+                    if item["mirror"] is None:
                         continue
 
-                    if item["mirror"] is None:
+                    mirror_url = item["mirror"]["url"]
+
+                    if repository_urls and mirror_url not in repository_urls:
```
geoberle, line 159: nit: depending how often we run the `for item in ...` loop, it might be faster to have `repository_urls` defined as a `set` - just a nit, ignore if you want

geoberle (approved):
LGTM

chassing (approved):

reconcile/quay_mirror.py
rporres: makes sense. Added in ce7404cd6f0ae484d514d69ad67d452259618cdc

PR #3858, opened 2023-10-03
openshift resource inventory metrics

maorfr:
this PR adds a new `qontract_reconcile_openshift_resource_inventory` metric, a guage which describes the number of resources managed per cluster, namespace, kind and state (current/desired).


patjlm (approved):
Nice. Can we get rid of an app-interface-output file after that?

PR #3859, opened 2023-10-04
Fix wrong params assign of ocm_standalone_user_management

hemslo:
Field name of `OCMUserManagementIntegrationParams` is `ocm_organization_ids` instead of `ocm_org_ids`, wrong params assign caused `ocm_org_ids` not being used as filter, lead to reconcile error for not managed ocm org.


geoberle (approved):
thank you!!

PR #3862, opened 2023-10-05
Remove Feature flag usage for enabling 3 way diff strategy

lechuk47:
* Use the 3-way diff strategy as default, without the feature flag. 
* Remove the old logic 
* Turn the logging messages of the diff method to DEBUG

### DIFF
--- a/reconcile/utils/three_way_diff_strategy.py
+++ b/reconcile/utils/three_way_diff_strategy.py
@@ -121,25 +121,20 @@ def is_valid_change(current: OR, desired: OR, patch: Mapping[str, Any]) -> bool:
 
 
 def three_way_diff_using_hash(c_item: OR, d_item: OR) -> bool:
-    # Get the ORIGINAL object hash
-    # This needs to be improved in OR, by now is just a PoC
     c_item_sha256 = ""
     try:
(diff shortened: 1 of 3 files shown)


janboll (approved):

reconcile/utils/three_way_diff_strategy.py
```diff
@@ -152,8 +147,7 @@ def three_way_diff_using_hash(c_item: OR, d_item: OR) -> bool:
         item for item in patch.patch if is_valid_change(current, desired, item)
     ]
     if len(valid_changes) > 0:
-        logging.info("Desired and Current objects differ -> Apply")
-        logging.info(valid_changes)
+        logging.debug(f"Desired and Current objects differ -> Apply: {valid_changes}")
```
hemslo, line 150: for logging method, best to use const string:

```
logging.debug("Desired and Current objects differ -> Apply: %s", valid_changes)
```

see https://docs.python.org/3/howto/logging.html#optimization and https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-format-interpolation.html

hemslo (approved):
LGTM!

PR #3863, opened 2023-10-05
add openshift resource inventory metrics to all openshift integrations

maorfr:
follow up on #3858


janboll (approved):

reconcile/terraform_resources.py
```diff
@@ -442,6 +442,7 @@ def run(
     # populate the resource inventory with latest output data
     populate_desired_state(ri, resource_specs)
 
+    ob.publish_metrics(ri, QONTRACT_INTEGRATION.replace("_", "-"))
```
hemslo, line 445: Why this one has a special replace, is it a requirment that `_` should be `-` in metric name? Should this be handled inside `db.publish_metrics` method?

maorfr: https://github.com/app-sre/qontract-reconcile/pull/3863/commits/7f125862723bf9e057e4b2d11cfddbc78032152f

hemslo (approved):
LGTM!

geoberle (approved):
❤️

PR #3864, opened 2023-10-05
change 3 way diff log from info to debug

patjlm:


patjlm:
Closing as fixed in https://github.com/app-sre/qontract-reconcile/pull/3862

PR #3866, opened 2023-10-06
qontract-cli get cidr-blocks

patjlm:
[APPSRE-8415](https://issues.redhat.com/browse/APPSRE-8415)

New qontract-cli command to list all CIDR blocks registered in app-interface (from clusters or AWS VPCs). The aim is to help selecting a free CIDR block when creating a new cluster or defining a new VPC.

Example output in Markdown: https://gitlab.cee.redhat.com/-/snippets/7236

Corresponding schema change: https://github.com/app-sre/qontract-schemas/pull/531

### <samp>🤖 Generated by Copilot at 3102a7a</samp>

This pull request refactors the GraphQL queries and data classes for clusters, VPCs, and peering, using fragments to reduce duplication and improve maintainability. It also adds a new query file and a new CLI subcommand for getting the CIDR blocks of the network resources.


tools/qontract_cli.py
```diff
@@ -1052,6 +1052,62 @@ def clusters_network(ctx, name):
...
+    )
+
+    for index, cidr in enumerate(cidrs):
+        network = ipaddress.ip_network(cidr["cidr"])
+        overlaps = [
+            f"{c['type']}/{c['name']}"
+            for i, c in enumerate(cidrs)
+            if i != index and network.overlaps(ipaddress.ip_network(c["cidr"]))
+        ]
+        cidr["overlaps"] = ", ".join(overlaps)
```
geoberle, line 1103: nit: for X cidrs it is X*X-1 comparisons. considering our low numbe of CIRDs i think this is absolutely ok for now and the report is of great value.

with the rise of AWS account numbers, we need to keep this in mind though

geoberle (approved):
LGTM ❤️

tools/qontract_cli.py
patjlm: I did not put a lot of thoughts into this. is there an easy way to get more efficient?

We could maybe slice the second `enumerate(cidrs)` as `enumerate(cidrs[index+1:])` but would need to register overlaps both ways somehow

reconcile/gql_definitions/introspection.json
```diff
@@ -397,6 +397,37 @@
                             "isDeprecated": false,
                             "deprecationReason": null
                         },
+                        {
+                            "name": "awsvpcs_v1",
```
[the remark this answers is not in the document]
patjlm: ah yes... need to push it. thanks!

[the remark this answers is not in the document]
patjlm: https://github.com/app-sre/qontract-schemas/pull/531

tools/qontract_cli.py
```diff
@@ -1052,6 +1052,62 @@ def clusters_network(ctx, name):
...
+                "account": vpc.account.name,
+                "cidr": vpc.cidr_block,
+                "from": str(ipaddress.ip_network(vpc.cidr_block)[0]),
+                "to": str(ipaddress.ip_network(vpc.cidr_block)[-1]),
+                "hosts": ipaddress.ip_network(vpc.cidr_block).num_addresses,
+                "description": vpc.description,
+            }
+            for vpc in vpcs
+        ]
+    )
```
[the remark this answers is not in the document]
patjlm: thanks, fixed in f06d794f

tools/qontract_cli.py
patjlm: current run time:
```
real	0m4.123s
user	0m2.544s
sys	0m0.614s
```

hemslo: can use [itertools.combinations](https://docs.python.org/3/library/itertools.html#itertools.combinations) to get all combos, then collect overlaps, then iterate cidrs once to set them.

```python
    overlaps = defaultdict(list)
    for a, b in itertools.combinations(cidrs, 2):
        if ipaddress.ip_network(a["cidr"]).overlaps(ipaddress.ip_network(b["cidr"])):
            key_a = f"{a['type']}/{a['name']}"
            key_b = f"{b['type']}/{b['name']}"
            overlaps[key_a].append(key_b)
            overlaps[key_b].append(key_a)
    for cidr in cidrs:
        cidr["overlaps"] = ", ".join(overlaps[f"{cidr['type']}/{cidr['name']}"])
```

In terms of time complexity, it's still O(N^2), further improvement requires a different way to find overlap, not worth it now.

hemslo (approved):
LGTM!

tools/qontract_cli.py
patjlm: yeah, i am not sure it makes it more readable nor more efficient. So if you're both fine, I'll keep it as is for now

hemslo: yes, keep it as is for now, no big differences considering our data size.

hemslo: Just note here in case later we need optimize this or similar features, a more efficient way is using [Interval Tree](https://en.wikipedia.org/wiki/Interval_tree), O(NlogN).

PR #3867, opened 2023-10-06
APPSRE-7307 validate promotion via publisher uid

fishi0x01:
This is a prerequisite to switch to promotions_v2 in saasherder, which is needed to allow multi-publisher channels.

A subscriber needs a relation to all publisher target uids in order to validate promotions from multiple different publishers towards the same publisher channel.

This PR mainly introduces a method to introduce a channel class to the saas file class. The channel object holds all publisher uids.

Saasherder gets a full list of saas files from openshift-saas-deploy. This list is required to resolve publisher -> channel -> subscriber relations. Currently only openshift-saas-deploy uses validate_promotions, so this should be fine.

Note, that saasherder is still primarily using promotions_v1 to validate promotions.
If this change does not lead to issues, we will switch to promotions_v2 in a follow-up PR and cross fingers.

**Test**

Ran openshift-saas-deploy locally

### DIFF
--- a/reconcile/utils/saasherder/saasherder.py
+++ b/reconcile/utils/saasherder/saasherder.py
@@ -142,7 +142,7 @@ class SaasHerder:  # pylint: disable=too-many-public-methods
         self.include_trigger_trace = include_trigger_trace
         self.state = state
         self._promotion_state = PromotionState(state=state) if state else None
-        self._all_saas_files = all_saas_files
+        self._channel_map = self._assemble_channels(saas_files=all_saas_files)
 
         # each namespace is in fact a target,
         # so we can use it to calculate.
@@ -1039,7 +1039,9 @@ class SaasHerder:  # pylint: disable=too-many-public-methods
 
         target_promotion = None
         if target.promotion:
-            channels = self._assemble_channels(channel_names=target.promotion.subscribe)
+            channels = [
+                self._channel_map[sub] for sub in target.promotion.subscribe or []
+            ]
             target_promotion = Promotion(
                 auto=target.promotion.auto,
                 publish=target.promotion.publish,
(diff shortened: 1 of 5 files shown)


reconcile/utils/saasherder/saasherder.py
```diff
@@ -1051,6 +1055,37 @@ def _process_template(
...
+                            parent_saas_file_name=saas_file.name,
+                            parent_resource_template_name=tmpl.name,
+                        )
+                        if publish not in channel_map:
+                            channel_map[publish] = Channel(
+                                name=publish,
+                                publisher_uids=[],
+                            )
+                        channel_map[publish].publisher_uids.append(publisher_uid)
+        return list(channel_map.values())
```
hemslo, line 1087: Note this is will iterate on all saas files for every `_process_template`, I recommend to build a graph `publisher -> channel -> subscriber` once before hand, then just lookup info when needed.

Btw, to simplify exist check, can use a `defaultdict`:

```python
        channel_map = defaultdict(list)
        for saas_file in self._all_saas_files:
            for tmpl in saas_file.resource_templates:
                for target in tmpl.targets:
                    if not target.promotion:
                        continue
                    for publish in target.promotion.publish or []:
                        if publish not in channel_names_set:
                            continue
                        publisher_uid = target.uid(
                            parent_saas_file_name=saas_file.name,
                            parent_resource_template_name=tmpl.name,
                        )
                        channel_map[publish].append(publisher_uid)
        return [Channel(name=k, publisher_uids=v) for k, v in channel_map.items()]
```

fishi0x01: Very nice catch! Thank you!
Running assemble_channel now only once at init :+1:

chassing (approved):

PR #3868, opened 2023-10-06
ignore ASG instances not InService for a-i output reporting

geoberle:
instances that are not InService don't have enough metadata to be shown in the report. every attempt to get more data fails.


tools/qontract_cli.py
```diff
@@ -2204,6 +2205,12 @@ def ec2_jenkins_workers(ctx, aws_access_key_id, aws_secret_access_key, aws_regio
     for a in auto_scaling_groups:
         for i in a["Instances"]:
             instance = ec2.Instance(i["InstanceId"])
+            lifecycle_state = i["LifecycleState"]
```
hemslo, line 2208: Will `i["InstanceId"]` have value when lifecycle state is not `InService`, should we skip `ec2.Instance` too?

geoberle: that is the strange thing, it does. but when boto3 tries to access more data for the instance it fails.

i moved the call to ec2.Instance() down so we don't call it when the lifecycle state results in a skip anyways.

hemslo (approved):
LGTM!

PR #3871, opened 2023-10-09
[sql-query] Publish sql-query metrics for qontract_reconcile_openshift_resource_inventory

hemslo:
Follow up on #3858, `sql-query` integration `qontract_reconcile_openshift_resource_inventory` metric is a special one.

This integration is using state bucket to do conditional operations without populating all current and desired states. This change handles metrics in the following logic:

* For new query, populate `desired` from constructed openshift resources
* For existing query:
  * If state is `DONE`, publish current and desired with 0 as all resources are already deleted
  * If pending deletion (explicit delete or TTL expired), publish current openshift resources to be deleted
  * If still active, reconstruct desired openshift resources and use them as both desired and current

### <samp>🤖 Generated by Copilot at b6676e4</samp>

Added and updated tests for `sql_query` integration. The tests cover various scenarios of querying data from different sources and handling errors. The test file also has updated imports to reflect the changes in the integration code.

### <samp>🤖 Generated by Copilot at b6676e4</samp>

*  Add imports for modules and classes used in new tests ([link](https://github.com/app-sre/qontract-reconcile/pull/3871/files?diff=unified&w=0#diff-7c15e1482046349edcc507e8edf0a26d896111ff958613cf3ff79c75d06dc1b4L1-R12))
*  Add tests for `run` function of `sql_query` integration ([link](https://github.com/app-sre/qontract-reconcile/pull/3871/files?diff=unified&w=0#diff-7c15e1482046349edcc507e8edf0a26d896111ff958613cf3ff79c75d06dc1b4R29-R316))

### DIFF
--- a/reconcile/sql_query.py
+++ b/reconcile/sql_query.py
@@ -5,6 +5,7 @@ from collections.abc import (
     Iterable,
     Mapping,
 )
+from enum import Enum
 from textwrap import indent
 from typing import (
     Any,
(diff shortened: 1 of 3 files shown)


chassing (approved):
Nice refactoring. LGTM

PR #3872, opened 2023-10-09
Database Access Manager integration

janboll:
Adding Database Access Manager integration

image

Required Schema Change: https://github.com/app-sre/qontract-schemas/pull/509

This is only the first step of the implementation. Some missing features will be implemented in later versions:

- User deletion/Access revocation
- Partial permissions update

https://issues.redhat.com/browse/APPSRE-8131


reconcile/database_access_manager.py
```diff
@@ -0,0 +1,506 @@
...
+                            "image": f"{job_data.image_repository}/{job_data.engine}:{job_data.engine_version}",
+                            "command": [
+                                command,
+                            ],
+                            "args": [
+                                "-ae",
+                                "--host=$(db.host)",
+                                "--port=$(db.port)",
+                                "--username=$(db.user)",
+                                "--dbname=$(db.name)",
```
geoberle, line 183: i tend to forget about the $() notation supported in `args`. nice

reconcile/database_access_manager.py
```diff
@@ -0,0 +1,506 @@
...
+        resource: NamespaceTerraformResourceRDSV1,
+        resource_name: str,
+        settings: dict[any, any],
+    ) -> list[DBAMResource]:
+        managed_resources: list[DBAMResource] = []
+
+        # create service account
+        managed_resources.append(
+            DBAMResource(
+                resource=get_service_account_spec(resource_name),
```
geoberle, line 337: do you want to prefix the SA name with something to prevent potential conflicts with other SAs that exist in the namespace already?

reconcile/database_access_manager.py
```diff
@@ -0,0 +1,506 @@
...
+        encounteredErrors = False
+
+        for namespace in get_database_access_namespaces():
+            with OC_Map(
+                namespaces=[namespace.dict(by_alias=True)],
+                integration=QONTRACT_INTEGRATION,
+                settings=settings,
+            ) as oc_map:
+                for external_resource in namespace.external_resources:
+                    for resource in external_resource.resources:
```
geoberle, line 417: is there no filtering here to only look at RDS resources?

reconcile/database_access_manager.py
```diff
@@ -0,0 +1,506 @@
...
+                                    dry_run=dry_run,
+                                    oc_map=oc_map,
+                                    cluster=namespace.cluster.name,
+                                    namespace=namespace.name,
+                                    resource_type=r.resource.kind,
+                                    resource=r.resource,
+                                    wait_for_namespace=False,
+                                )
+
+                            while True:
```
geoberle, line 473: so the integration creates a job and waits for the results. what happens if the integration gets killed in between (various reasons from cluster upgrades to qr promotion).

reconcile/database_access_manager.py
```diff
@@ -0,0 +1,506 @@
...
+
+
+def generate_user_secret_spec(
+    name: str, db_access: DatabaseAccessV1
+) -> OpenshiftResource:
+    secret = secret_head(name)
+    secret["data"] = {
+        "db.host": "",
+        "db.name": base64.b64encode(db_access.database.encode("utf-8")).decode("utf-8"),
+        "db.password": "",
```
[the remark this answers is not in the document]
janboll: correct, fixed now

reconcile/database_access_manager.py
```diff
@@ -0,0 +1,506 @@
...
+    WHERE NOT EXISTS (SELECT FROM pg_database WHERE datname = '{self._get_db()}');\gexec
+    
+    /* revoke only required for databases lower version 15
+       this makes the behaviour compliant with postgres 15, writes are only allowed
+       in the schema created for the role/user */
+    REVOKE ALL ON DATABASE {self._get_db()} FROM public;
+    
+    \c {self._get_db()}
+    
+    select 'CREATE ROLE {self._get_user()}  WITH LOGIN PASSWORD ''foo'' VALID UNTIL ''infinity'''
```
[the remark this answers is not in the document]
janboll: correct, fixed now

reconcile/database_access_manager.py
geoberle: sgtm. but why the while true then? you can pick up where you left of during the next reconcile run, no?

reconcile/database_access_manager.py
```diff
@@ -0,0 +1,506 @@
...
+        body=secret,
+        integration=QONTRACT_INTEGRATION,
+        integration_version=QONTRACT_INTEGRATION_VERSION,
+    )
+
+
+def generate_script_secret_spec(name: str, script: str) -> OpenshiftResource:
+    secret = secret_head(name)
+    secret["data"] = {
+        "script.sql": base64.b64encode(script.encode("utf-8")).decode("utf-8"),
```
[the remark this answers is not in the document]
janboll: Max size is around 1MByte, that should suffice :sweat_smile:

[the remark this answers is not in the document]
janboll: https://kubernetes.io/docs/concepts/configuration/secret/#restriction-data-size

reconcile/database_access_manager.py
janboll: The resource_name is "{dbam-dbam_resource.name}". That should make it unique

reconcile/database_access_manager.py
janboll: No, filtering for nonUnique attributes is not possible as of now

reconcile/database_access_manager.py
```diff
@@ -0,0 +1,620 @@
...
+
+class JobStatusCondition(BaseModel):
+    type: str
+
+
+class JobStatus(BaseModel):
+    conditions: list[JobStatusCondition]
+
+    def is_complete(self) -> bool:
+        return True if self.conditions else False
```
geoberle, line 333: the purpose of this function is to figure out if the job is not running anymore (at least i understand that from the usage), regardless out outcome right?

a successful job also has a condition, namely `type=Complete`

geoberle: nevermind. i checked and it seems that's the case. a running job has no conditions and only gets one when succeeding or failing.

geoberle: there are some job types where there might be conditions in between (index runs) but we are not using them here so should be ok

geoberle (approved):
LGTM

reconcile/database_access_manager.py
janboll: I think that does not apply here, since this only checks if the job is still running. Not caring about the exit status

reconcile/database_access_manager.py
```diff
@@ -0,0 +1,621 @@
...
+                for external_resource in [
+                    er
+                    for er in namespace.external_resources or []
+                    if isinstance(er, NamespaceTerraformProviderResourceAWSV1)
+                ]:
+                    for resource in [
+                        r
+                        for r in external_resource.resources or []
+                        if isinstance(r, NamespaceTerraformResourceRDSV1)
+                    ]:
```
hemslo, line 596: The filter logic is too deep nested, may worth extracting out like this:

```python
    @staticmethod
    def _resources(namespace: NamespaceV1) -> Iterator[NamespaceTerraformResourceRDSV1]:
        for external_resource in namespace.external_resources or []:
            if isinstance(external_resource, NamespaceTerraformProviderResourceAWSV1):
                for resource in external_resource.resources or []:
                    if isinstance(resource, NamespaceTerraformResourceRDSV1):
                        yield resource
```

Then it will be just 

```python
for resource in self._resources(namespace):
```

You can even extract out all loop params needed for `_process_db_access` out, and all iterate logic out, then it can be used like

```python
            for params in self._db_accesses(namespaces):
                _process_db_access(
                    ...
                )
```

reconcile/database_access_manager.py
```diff
@@ -0,0 +1,621 @@
...
+                                        db_access,
+                                        oc_map,
+                                        namespace.cluster.name,
+                                        namespace.name,
+                                        admin_secret_name,
+                                        get_db_engine(resource),
+                                        sql_query_settings,
+                                    )
+                                except JobFailedError:
+                                    encounteredErrors = True
```
hemslo, line 618: May worth adding a log here

reconcile/gql_definitions/terraform_resources/database_access_manager.gql
```diff
@@ -0,0 +1,47 @@
...
+                  dbschema
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
hemslo, line 47: Add an EOL here to avoid warning

reconcile/database_access_manager.py
```diff
@@ -0,0 +1,621 @@
...
+
+        sql_query_settings = settings.get("sqlQuery")
+        if not sql_query_settings:
+            raise KeyError("sqlQuery settings are required")
+
+        state = init_state(
+            integration=QONTRACT_INTEGRATION, secret_reader=self.secret_reader
+        )
+
+        encounteredErrors = False
```
hemslo, line 578: `encounteredErrors` -> `encountered_errors`, instead of a bool, can collect all error objects then raise a `ExceptionGroup` to contain error details.

hemslo (commented):
Since this integration is managing several openshift resources, it should publish metrics `qontract_reconcile_openshift_resource_inventory`, like https://github.com/app-sre/qontract-reconcile/pull/3871

PR #3874, opened 2023-10-09
autodetect default branch name for automated MRs

geoberle:
the a-i default branch name can't be assumed to be "master" for all a-i repos but can be "main" or anything else as well.

this PR introduces main branch name detection.

(also added some typehints along the way, thought it is not comprehensive since the gitlab api does not really expose good typehints on its own 😢 )

https://issues.redhat.com/browse/APPSRE-8420

### DIFF
--- a/reconcile/utils/gitlab_api.py
+++ b/reconcile/utils/gitlab_api.py
@@ -100,6 +100,22 @@ class GitLabApi:  # pylint: disable=too-many-public-methods
             gitlab_request.labels(integration=INTEGRATION_NAME).inc()
             self.project = self.gl.projects.get(project_id)
         self.saas_files = saas_files
+        self._main_branch = None
+
+    @property
+    def main_branch(self) -> str:
+        if self._main_branch is None:
+            self._main_branch = self._determine_main_branch()
+        return self._main_branch or "master"
+
+    def _determine_main_branch(self) -> Optional[str]:
+        if self.project:
+            main_branch_obj = next(
+                (b for b in self.project.branches.list() if b.default), None
+            )
+            if main_branch_obj:
+                return main_branch_obj.name
+        return None
 
     def __enter__(self):
         return self
--- a/reconcile/utils/gitlab_api.py
+++ b/reconcile/utils/gitlab_api.py
@@ -4,6 +4,7 @@ from collections.abc import (
     Iterable,
     Set,
 )
+from functools import cached_property
 from operator import (
     attrgetter,
     itemgetter,
@@ -36,6 +37,7 @@ MR_DESCRIPTION_COMMENT_ID = 0
 
 # The default value is there for unit test
 INTEGRATION_NAME = os.getenv("INTEGRATION_NAME", "")
+DEFAULT_MAIN_BRANCH = "master"
 
 
 class MRState:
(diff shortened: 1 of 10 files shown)


reconcile/utils/gitlab_api.py
```diff
@@ -100,6 +100,22 @@ def __init__(
...
+        return self._main_branch or "master"
+
+    def _determine_main_branch(self) -> Optional[str]:
+        if self.project:
+            main_branch_obj = next(
+                (b for b in self.project.branches.list() if b.default), None
+            )
+            if main_branch_obj:
+                return main_branch_obj.name
+        return None
```
hemslo, line 118: It seems we want to cache `main_branch` computation, right?
How about using `cached_property`?

```python
    DEFAULT_MAIN_BRANCH = "master"

    @cached_property
    def main_branch(self) -> str:
        if self.project:
            return next(
                (b.name for b in self.project.branches.list() if b.default),
                DEFAULT_MAIN_BRANCH,
            )
        return DEFAULT_MAIN_BRANCH
```

hemslo (commented):
To support more types, we need to update `PyGithub` version.

reconcile/utils/mr/base.py
```diff
@@ -206,17 +206,17 @@ def delete_tmp_branch(self, gitlab_cli):
...
-    def submit(self, cli: MRClient):
+    def submit(self, cli: MRClient) -> Union[Any, None]:
         if isinstance(cli, GitLabApi):
             return self.submit_to_gitlab(gitlab_cli=cli)
 
         if isinstance(cli, SQSGateway):
-            return self.submit_to_sqs(sqs_cli=cli)
+            self.submit_to_sqs(sqs_cli=cli)
 
         raise AttributeError(f"client {cli} not supported")
```
chassing, line 221: Why did you remove `return`?

reconcile/utils/gitlab_api.py
geoberle: there are some code paths where the projects becomes available only later. so i want to avoid caching a default branch name when there might be a better one on the project

reconcile/utils/mr/base.py
geoberle: `submit_to_sqs` does not return anything and mypy was starting to complain when i added the type hints

reconcile/utils/gitlab_api.py
hemslo: then how about this:

```python
    @cached_property
    def project_main_branch(self) -> str:
        return next(
            (b.name for b in self.project.branches.list() if b.default),
            DEFAULT_MAIN_BRANCH,
        )

    @property
    def main_branch(self) -> str:
        return this.project_main_branch if self.project else DEFAULT_MAIN_BRANCH
```

reconcile/utils/mr/base.py
chassing: But now, if someone runs it with `SQSGateway`, it will always raise an `AttributeError`.

reconcile/utils/gitlab_api.py
geoberle: i like this, i will use it.

in general i think the gitlab_api.py code needs some refactoring, because it mixes pure API things with projects things. being able to switch out the project behind the scene makes me kind of nervous.

in the meantime, your code will work nicely. thank you

chassing (approved):

hemslo (approved):
LGTM!

PR #3875, opened 2023-10-10
APPSRE-7307 remove promotions_v1 fetch

fishi0x01:
Deprecate promotions_v1. Deployment validation logic will not use any data from old promotions storage prefix.

Note, in the meantime we still keep publishing to promotions_v1 path, so that we can easily revert without having lost potential promotion data. Once this PR is confirmed to have worked, we will remove publishing to promotions_v1 in a follow-up PR.


chassing (approved):
LGTM. Let's hope the best 💣

PR #3876, opened 2023-10-10
q-cli get slo-document-services

jmelis:
CLI that will be used by the ops-dev-team (@mzazrivec) to obtain the list of Status Board services that must be created.


reconcile/typed_queries/slo_documents.py
```diff
@@ -0,0 +1,14 @@
...
+    query,
+)
+from reconcile.utils import gql
+from reconcile.utils.gql import GqlApi
+
+
+def get_slo_documents(gql_api: Optional[GqlApi] = None) -> list[SLODocumentV1]:
+    api = gql_api if gql_api else gql.get_api()
+    data = query(query_func=api.query)
+    return list(data.slo_document_v1 or [])
```
chassing, line 14: Why the additional `list` and not just `return data.slo_document_v1 or []`? Copy and paste from somewhere else?

chassing (approved):
Besides the nit mentioned above. LGTM

PR #3880, opened 2023-10-11
[glitchtip-project-alerts] Jira integration

chassing:
Create alert webhooks for the Jira integration ([glitchtip-jira-bridge](https://github.com/app-sre/glitchtip-jira-bridge)) automatically if the `glitchtip_project.jira` attribute is set. Also refactor `get_glitchtip_settings` slightly.

Ticket: https://issues.redhat.com/browse/APPSRE-8238
Design doc: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/glitchtip-jira.md
Depends on: https://github.com/app-sre/qontract-schemas/pull/535


reconcile/glitchtip_project_alerts/integration.py
```diff
@@ -103,8 +103,42 @@ def fetch_desired_state(
...
+                        "Either jira.project or jira.board must be set for Jira integration"
+                    )
+
+                url = f"{gjb_alert_url}/{jira_project_key}"
+                if gjb_token:
+                    url += f"?token={quote(gjb_token)}"
+
+                alerts.append(
+                    ProjectAlert(
+                        name="Jira-integration",
```
hemslo, line 124: Since this value is hard coded, will duplicate `name` causing issues?

reconcile/typed_queries/glitchtip_settings.py
```diff
@@ -1,18 +1,33 @@
...
+        if ai_gs.read_timeout:
+            gs.read_timeout = ai_gs.read_timeout
+        if ai_gs.max_retries:
+            gs.max_retries = ai_gs.max_retries
+        if ai_gs.mail_domain:
+            gs.mail_domain = ai_gs.mail_domain
+        if ai_gs.glitchtip_jira_bridge_alert_url:
+            gs.glitchtip_jira_bridge_alert_url = ai_gs.glitchtip_jira_bridge_alert_url
+        if ai_gs.glitchtip_jira_bridge_token:
+            gs.glitchtip_jira_bridge_token = ai_gs.glitchtip_jira_bridge_token
```
hemslo, line 32: Do you mean something like this?

```python
GlitchtipSettings(**ai_gs.dict(exclude_none=True))
```

chassing: Unfortunately, this doesn't work because of the aliases in the `VaultSecret` GQL class :( and I don't want to copy that class.

reconcile/glitchtip_project_alerts/integration.py
chassing: Good point. The integration will raise an error if someone uses the reserved alert name. See [eb3091e](https://github.com/app-sre/qontract-reconcile/pull/3880/commits/eb3091ed23b470e5ee50873a4089080e89be2efc)

reconcile/typed_queries/glitchtip_settings.py
chassing: @geoberle [suggested](https://github.com/app-sre/qontract-schemas/pull/535#discussion_r1360853341) to move the settings to the instance object. I did that in [7f56daa](https://github.com/app-sre/qontract-reconcile/pull/3880/commits/7f56daa3784b36873998dcfae9e9ccf346b2e5a6), and now it's much cleaner :)

hemslo: Nice, one issue with that or https://github.com/app-sre/qontract-reconcile/pull/3880/commits/7f56daa3784b36873998dcfae9e9ccf346b2e5a6#diff-4ce1a595f5fdda9d51fb6cd5777e6c849c81c115924afc45030f00af91d42143R42-R43 is `0` will be override to `3` or `30`, is that expected?

chassing: You have 🦅 👀. Changed in [31f3d38](https://github.com/app-sre/qontract-reconcile/pull/3880/commits/31f3d389306f2ce5c5a570ae2b7b698c42fca526)

hemslo (approved):
LGTM!

PR #3885, opened 2023-10-13
[OCM service logs] switch to cluster specific service log list endpoint

geoberle:
the service log LIST endpoint `/api/service_logs/v1/cluster_logs` is being deprecated and we need to switch to the `/api/service_logs/v1/clusters/cluster_logs` one.

the new endpoint is cluster specific and not org specific anymore, but since we were fetching service logs for specific cluster anyways (filtering server side via `search`) the change has low impact to the overall usage of service logs in qontract-reconcile.

part of https://issues.redhat.com/browse/APPSRE-8422

### DIFF
--- a/reconcile/utils/ocm/service_log.py
+++ b/reconcile/utils/ocm/service_log.py
@@ -12,16 +12,22 @@ from reconcile.utils.ocm.base import (
 from reconcile.utils.ocm.search_filters import Filter
 from reconcile.utils.ocm_base_client import OCMBaseClient
 
+CLUSTER_SERVICE_LOGS_LIST_ENDPOINT = "/api/service_logs/v1/clusters/cluster_logs"
+CLUSTER_SERVICE_LOGS_CREATE_ENDPOINT = "/api/service_logs/v1/cluster_logs"
 
-def get_service_logs(
-    ocm_api: OCMBaseClient, filter: Filter
+
+def get_service_logs_for_cluster_uuid(
+    ocm_api: OCMBaseClient, cluster_uuid: str, filter: Optional[Filter] = None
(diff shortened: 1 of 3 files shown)


reconcile/utils/ocm/service_log.py
```diff
@@ -12,16 +12,22 @@
...
 from reconcile.utils.ocm_base_client import OCMBaseClient
 
+CLUSTER_SERVICE_LOGS_LIST_ENDPOINT = "/api/service_logs/v1/clusters/cluster_logs"
+CLUSTER_SERVICE_LOGS_CREATE_ENDPOINT = "/api/service_logs/v1/cluster_logs"
 
-def get_service_logs(
-    ocm_api: OCMBaseClient, filter: Filter
+
+def get_service_logs_for_cluster_uuid(
+    ocm_api: OCMBaseClient, cluster_uuid: str, filter: Optional[Filter] = None
```
hemslo, line 20: Is there a use case that `filter` is `None`?

Another case is what if an empty filter (`Filter()`) passed in, should `params["search"]` be populated as empty string or that key shouldn't be in the payload at all?

geoberle: this basically just represents the functionality of the endpoint which supports unfiltered requests since the `search` parameter is optional (therefore i'm not adding the param if there is no filter). there will be other usecases for this endpoint in upcoming work and we will see then how the interface of this function can evolve. i think it is ok to not have a filter but we could add some protection against too big responses.

empty filters are not supported. they raise an error on rendering. i added a test do show the behaviour.

hemslo (approved):
LGTM!

PR #3886, opened 2023-10-13
Update Status board

janboll:
Change application naming to be consistent, always prefix with parent name

### DIFF
--- a/reconcile/typed_queries/status_board.py
+++ b/reconcile/typed_queries/status_board.py
@@ -40,10 +40,11 @@ def get_selected_app_names(
         apps["apps"].append(app)
 
         for child in namespace.app.children_apps or []:
-            if child.name not in selected_app_names:
+            name = f"{namespace.app.name}-{child.name}"
+            if name not in selected_app_names:
                 selected_app_names.add(f"{namespace.app.name}-{child.name}")
                 child_dict = child.dict(by_alias=True)
-                child_dict["name"] = f"{namespace.app.name}-{child.name}"
+                child_dict["name"] = name
                 apps["apps"].append(child_dict)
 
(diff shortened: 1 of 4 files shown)


reconcile/typed_queries/status_board.py
```diff
@@ -27,19 +27,24 @@ def get_selected_app_names(
...
     apps: dict[str, Any] = {"apps": []}
     for namespace in product.product_environment.namespaces or []:
-        apps["apps"].append(namespace.app.dict(by_alias=True))
+        prefix = ""
+        if namespace.app.parent_app:
+            prefix = f"{namespace.app.parent_app.name}-"
+        selected_app_names.add(f"{prefix}{namespace.app.name}")
+        app = namespace.app.dict(by_alias=True)
+        app["name"] = f"{prefix}{namespace.app.name}"
+        apps["apps"].append(app)
```
hemslo, line 40: This will change `name` in `apps`, will this affect the result of selector usage https://github.com/app-sre/qontract-reconcile/pull/3886/files#diff-a660d41a807b3693327e4b178284222146e6a42507601df0e8bbcf6facb2c89aR56?

jmelis (approved):
LGTM

reconcile/typed_queries/status_board.py
janboll: I added another test case for this

hemslo: Nice! that `f"{prefix}{namespace.app.name}"` can be extracted to a `name` so we can see clearly the one added to `selected_app_names` is the same as `app["name"]`

hemslo (approved):
LGTM!

PR #3889, opened 2023-10-15
[mypy/requests] code fixes because of types-requests changes

geoberle:
`types-requests` now allows `requests.exceptions.HTTPError.response` to be `None` and in order to make mypy happy, we need to check for it

see https://github.com/python/typeshed/pull/10875


hemslo (approved):
LGTM!

PR #3892, opened 2023-10-16
Revert "Upgrade package that has higher level vulnerabilities and rem…

OliviaHY:
…ove irrelevant type ignore. (#3891)"

This reverts commit 1261f543efa97aad0ea503198e2f0bdc7326549c.

Because there is some incompatibility https://ci.int.devshift.net/job/service-app-interface-gl-pr-check/295236/artifact/temp/reports/index.html


hemslo:
@OliviaHY that error is not related to this change, see a later build https://ci.int.devshift.net/job/service-app-interface-gl-pr-check/295316/, it's working, no need to revert.

PR #3893, opened 2023-10-17
[openshift-saas-deploy] publish metrics

maorfr:
The [AppSRE Overview dashboard](https://grafana.app-sre.devshift.net/d/xNTPSl-Vk/appsre-overview) currently shows OpenShift resources that do not include resources managed via saas files.

With this PR, we add an option for `openshift-saas-deploy` to run as a metrics exporter, acting just like each `openshift-saas-deploy` deployment (tekton), but instead of deploying - publishing metrics.

The intention is to run this integration as an exporter just like any other integration.


patjlm (approved):
❤️

PR #3846, opened 2023-09-29
Speed up saas-file parsing

patjlm:
Further improvement on top of https://github.com/app-sre/qontract-reconcile/pull/3837

Linked to https://issues.redhat.com/browse/APPSRE-8352

- cache namespaceselectors result
- convert pydantic namespaces to dict only once
- avoid calling multiple time gql and parsing namespaceselectors by reusing all_saas_files when filtering by name. Note we now need to deep copy individual `saas_file` to not mutate `all_saas_files`

This PR will be easier to review 1 commit at a time.

On my machine, retrieving saas-files in openshift-saas-deploy (without param) goes from 24s to 7s

### <samp>🤖 Generated by Copilot at 19bba5e</samp>

Refactored the code for querying and filtering saas files using a new `SaasFileList` class. This improves the performance, readability, and error handling of the code. Updated the tests and the `openshift_saas_deploy` and `openshift_saas_deploy_change_tester` integrations to use the new class.

### <samp>🤖 Generated by Copilot at 19bba5e</samp>

[...]

PR #3895, opened 2023-10-17
APPSRE-8410 simplify selector matcher

fishi0x01:
The namespace selector matcher for the SAPM renderer has been copy/pasted from `typed_queries.saas_files.py`. That function had a recent make over in `SaasFileList` (https://github.com/app-sre/qontract-reconcile/pull/3846).

In this PR SAPM's renderer gets a simplified version to reverse match a given namespace to a selector and stay more decoupled from `SaasFileList`.

### DIFF
--- a/reconcile/saas_auto_promotions_manager/merge_request_manager/renderer.py
+++ b/reconcile/saas_auto_promotions_manager/merge_request_manager/renderer.py
@@ -56,14 +56,9 @@ class Renderer:
                 **namespace_selector
             )
             # Check if the target namespace is addressed by the selector
-            return (
-                len(
-                    get_namespaces_by_selector(
-                        namespaces=[subscriber.target_namespace],
-                        namespace_selector=selector,
-                    )
-                )
-                == 1
+            return is_namespace_addressed_by_selector(
+                namespace=subscriber.target_namespace,
+                namespace_selector=selector,
             )
         return target["namespace"]["$ref"] == subscriber.target_namespace.path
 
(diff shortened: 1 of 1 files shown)


reconcile/saas_auto_promotions_manager/merge_request_manager/renderer.py
```diff
@@ -158,56 +153,38 @@ def render_title(self, channels: str) -> str:
...
-            for match in parser.parse(include).find(namespaces_as_dict):
-                cluster_name = match.value["cluster"]["name"]
-                ns_name = match.value["name"]
-                filtered_namespaces[
-                    f"{cluster_name}-{ns_name}"
-                ] = _get_namespace_by_cluster_and_name(cluster_name, ns_name)
+            if len(parser.parse(include).find(namespace_as_dict)):
+                # We found a match, i.e., the given namespace does
+                # fit the selector expression
+                do_include |= True
```
hemslo, line 172: Another way to early return is

```python
        do_include = any(
            parser.parse(include).find(namespace_as_dict)
            for include in namespace_selector.json_path_selectors.include
        )
```

If `parser.parse` need error handling, better to extract it to a method

```python
def _parse(path):
    try:
        return parser.parse(path)
    except JsonPathParserError as e:
        raise RuntimeError(f"Invalid jsonpath expression '{path}' :{e}")
```

reconcile/saas_auto_promotions_manager/merge_request_manager/renderer.py
```diff
@@ -158,56 +153,38 @@ def render_title(self, channels: str) -> str:
...
+        do_exclude = False
         for exclude in namespace_selector.json_path_selectors.exclude or []:
-            for match in parser.parse(exclude).find(namespaces_as_dict):
-                cluster_name = match.value["cluster"]["name"]
-                ns_name = match.value["name"]
-                filtered_namespaces.pop(f"{cluster_name}-{ns_name}", None)
+            if parser.parse(exclude).find(namespace_as_dict):
+                # We have a match, i.e., the given namespace fits
+                # the exclude expression
+                do_exclude |= True
```
hemslo, line 184: Another way to early return is

```python
        do_exclude = any(
            parser.parse(exclude).find(namespace_as_dict)
            for exclude in namespace_selector.json_path_selectors.exclude or []
        )
```

reconcile/saas_auto_promotions_manager/merge_request_manager/renderer.py
fishi0x01: Nice suggestions! Definitely helps with read-ability :+1:

PR #3899, opened 2023-10-23
Add integration for RHACS rbac configuration

dwelch0:
https://issues.redhat.com/browse/APPSRE-8467
https://issues.redhat.com/browse/APPSRE-7607

Dependencies:
*  https://github.com/app-sre/qontract-schemas/pull/539

This integration will be responsible for reconciling [RH ACS rbac resources](https://docs.openshift.com/acs/4.2/operating/manage-user-access/manage-role-based-access-control-3630.html) . The integration will derive necessary rbac resources via oidc-permissions (see [expansion to oidc-permission-1 schema](https://github.com/app-sre/qontract-schemas/pull/539/files#diff-7a842619b2728c197543a91c51fc7eb0b944b2cb85b43900c981ce6522073faaR71-R96)) associated with App Interface roles that users reference.

`acs` variants of the oidc-permission are defined like so:
```yaml
---
$schema: /access/oidc-permission-1.yml

name: app-sre-cluster-scope

description: example of assigning permission at cluster scope

service: acs

permission_set: admin

clusters: 
- $ref: /clusters/appsres05ue1/cluster.yml
- $ref: /clusters/appint-ex-01/cluster.yml
```

These permissions are then referenced in the existing `oidc_permissions` attribute of the `role-1` schema:
```yaml
---
$schema: /access/role-1.yml

name: team-member-role

permissions:
- $ref: /teams/example-team/permissions/team-slack-permissions.yml

oidc_permissions:
- $ref: /dependencies/acs/permissions/app-sre-cluster-scope.yml
```

After defining an acs oidc-permission and referencing it within a role used by App Interface users, the integration becomes responsible for reconciling the ACS RBAC resources necessary to grant the desired access.

The `acs` oidc-permission defined within App Interface corresponds to two ACS rbac api resources: `role` and `access scope`. These two resources will share the same name (derived from name of the oidc-permission).
* note: `role` is derived from the oidc-permission definition instead of the App Interface role(s) the permission is referenced by due ability for A-I roles to reference multiple `oidc_permissions`. There would be potential for multiple ACS roles with same name.

`permission_set` is a enum with values that directly correspond to system default `permission sets` within ACS. ACS provides ample `permission set` options by default that cover our use cases. The enum list can be expanded to cover additional system default options if deemed necessary.

Finally, the link between an App Interface user that references a role with an `acs` oidc-permission attached corresponds to an ACS api `group`. Each `group` resource creates a rule within the ACS auth provider for assignment of permission to an authenticated user.
Example of one `group` represented in list of auth rules:
image

### DIFF
--- a/reconcile/acs_rbac.py
+++ b/reconcile/acs_rbac.py
@@ -1,4 +1,5 @@
 import logging
+from collections import defaultdict
 from collections.abc import Callable
 from typing import Optional
 
@@ -6,7 +7,7 @@ from pydantic import BaseModel
 
 from reconcile.gql_definitions.acs.acs_instances import AcsInstanceV1
 from reconcile.gql_definitions.acs.acs_instances import query as acs_instances_query
-from reconcile.gql_definitions.acs.acs_rbac import OidcPermissionAcsV1
+from reconcile.gql_definitions.acs.acs_rbac import OidcPermissionAcsV1, UserV1
 from reconcile.gql_definitions.acs.acs_rbac import query as acs_rbac_query
 from reconcile.typed_queries.app_interface_vault_settings import (
     get_app_interface_vault_settings,
@@ -78,6 +79,55 @@ class AcsRole(BaseModel):
     access_scope: AcsAccessScope
     system_default: Optional[bool]
 
+    class PermissionWithUsers(BaseModel):
+        permission: OidcPermissionAcsV1
+        usernames: list[str]
+
+    @classmethod
+    def build(cls, pu: PermissionWithUsers):
+        assignments = [
+            AssignmentPair(
+                # https://github.com/app-sre/qontract-schemas/blob/main/schemas/access/user-1.yml#L16
+                key="org_username",
+                value=u,
+            )
+            for u in pu.usernames
+        ]
+
+        is_unrestricted_scope = (
+            pu.permission.clusters is None or len(pu.permission.clusters) == 0
+        ) and (pu.permission.namespaces is None or len(pu.permission.namespaces) == 0)
+
+        return cls(
+            name=pu.permission.name,
+            description=pu.permission.description,
+            assignments=assignments,
+            permission_set_name=PERMISSION_SET_NAMES[pu.permission.permission_set],
+            access_scope=AcsAccessScope(
+                # Due to api restriction, additional Unrestricted scopes
+                # cannot be made.
+                # Therefore, desired scopes that meet unrestricted condition
+                # are treated as the system default 'Unrestricted'
+                name=DEFAULT_ADMIN_SCOPE_NAME
+                if is_unrestricted_scope
+                else pu.permission.name,
+                description=DEFAULT_ADMIN_SCOPE_DESC
+                if is_unrestricted_scope
+                else pu.permission.description,
+                # second arg is returned even if first arg == False
+                clusters=[cluster.name for cluster in (pu.permission.clusters or [])],
+                # mirroring format of 'rules.includedNamespaces' in /v1/simpleaccessscopes response
+                namespaces=[
+                    {
+                        "clusterName": n.cluster.name,
+                        "namespaceName": n.name,
+                    }
+                    for n in (pu.permission.namespaces or [])
+                ],
+            ),
+            system_default=False,
+        )
+
 
 class AcsRbacIntegrationParams(PydanticRunParams):
     thread_pool_size: int
@@ -88,7 +138,6 @@ class AcsRbacIntegration(QontractReconcileIntegration[AcsRbacIntegrationParams])
         super().__init__(params)
         self.qontract_integration = "acs_rbac"
         self.qontract_integration_version = make_semver(0, 1, 0)
-        self.qontract_tf_prefix = "qracsrbac"
 
     @property
     def name(self) -> str:
@@ -120,73 +169,22 @@ class AcsRbacIntegration(QontractReconcileIntegration[AcsRbacIntegrationParams])
         if query_results is None:
             return []
 
-        desired_roles: dict[str, AcsRole] = {}
+        permission_with_users: dict[str, AcsRole.PermissionWithUsers] = {}
         for user in query_results:
             for role in user.roles or []:
                 for permission in role.oidc_permissions or []:
                     if isinstance(permission, OidcPermissionAcsV1):
-                        # first encounter of specific permission
-                        # derive the Acs role specifics and add initial user
-                        if permission.name not in desired_roles:
-                            # empty cluster and namespaces attributes in oidc-permission
-                            # signifies unrestricted scope.
-                            is_unrestricted_scope = (
-                                permission.clusters is None
-                                or len(permission.clusters) == 0
-                            ) and (
-                                permission.namespaces is None
-                                or len(permission.namespaces) == 0
-                            )
-
-                            desired_roles[permission.name] = AcsRole(
-                                name=permission.name,
-                                description=permission.description,
-                                assignments=[
-                                    AssignmentPair(
-                                        # https://github.com/app-sre/qontract-schemas/blob/main/schemas/access/user-1.yml#L16
-                                        key="org_username",
-                                        value=user.org_username,
-                                    )
-                                ],
-                                permission_set_name=PERMISSION_SET_NAMES[
-                                    permission.permission_set
-                                ],
-                                access_scope=AcsAccessScope(
-                                    # Due to api restriction, additional Unrestricted scopes
-                                    # cannot be made.
-                                    # Therefore, desired scopes that meet unrestricted condition
-                                    # are treated as the system default 'Unrestricted'
-                                    name=DEFAULT_ADMIN_SCOPE_NAME
-                                    if is_unrestricted_scope
-                                    else permission.name,
-                                    description=DEFAULT_ADMIN_SCOPE_DESC
-                                    if is_unrestricted_scope
-                                    else permission.description,
-                                    # second arg is returned even if first arg == False
-                                    clusters=[
-                                        cluster.name
-                                        for cluster in (permission.clusters or [])
-                                    ],
-                                    # mirroring format of 'rules.includedNamespaces' in /v1/simpleaccessscopes response
-                                    namespaces=[
-                                        {
-                                            "clusterName": n.cluster.name,
-                                            "namespaceName": n.name,
-                                        }
-                                        for n in (permission.namespaces or [])
-                                    ],
-                                ),
-                                system_default=False,
+                        if permission.name not in permission_with_users:
+                            permission_with_users[
+                                permission.name
+                            ] = AcsRole.PermissionWithUsers(
+                                permission=permission, usernames=[user.org_username]
                             )
                         else:
-                            # role accounted for by prior user ref. Append additional desired user
-                            desired_roles[permission.name].assignments.append(
-                                AssignmentPair(
-                                    key="org_username", value=user.org_username
-                                )
+                            permission_with_users[permission.name].usernames.append(
+                                user.org_username
                             )
-
-        return list(desired_roles.values())
+        return [AcsRole.build(pus) for pus in permission_with_users.values()]
 
     def get_current_state(self, acs: AcsApi, auth_provider_id: str) -> list[AcsRole]:
         """
--- a/reconcile/acs_rbac.py
+++ b/reconcile/acs_rbac.py
@@ -168,8 +168,6 @@ class AcsRbacIntegration(QontractReconcileIntegration[AcsRbacIntegrationParams])
         """
         Get current ACS roles and associated users from ACS api
 
-        :param acs: ACS api client
-        :param auth_id: id of auth provider within ACS instance to target for reconciliation
         :return: list of current AcsRole associated with specified auth provider
         """
         current_roles: dict[str, AcsRole] = {}
@@ -256,9 +254,6 @@ class AcsRbacIntegration(QontractReconcileIntegration[AcsRbacIntegrationParams])
         Creates desired ACS roles as well as associated access scopes and rules
 
         :param to_add: result of 'diff_iterables(current, desired).add' for ACS roles
-        :param acs: ACS api client
-        :param auth_id: id of auth provider within ACS instance to target for reconciliation
-        :param dry_run: run in dry-run mode
         """
         access_scope_id_map = {s.name: s.id for s in self.acs.get_access_scopes()}
         permission_sets_id_map = {
@@ -336,9 +331,6 @@ class AcsRbacIntegration(QontractReconcileIntegration[AcsRbacIntegrationParams])
         Deletes desired ACS roles as well as associated access scopes and rules
 
         :param to_delete: result of 'diff_iterables(current, desired).delete' for ACS roles
-        :param acs: ACS api client
-        :param auth_id: id of auth provider within ACS instance to target for reconciliation
-        :param dry_run: run in dry-run mode
         """
         access_scope_id_map = {s.name: s.id for s in self.acs.get_access_scopes()}
         role_group_mappings: dict[str, list[str]] = {}
@@ -387,9 +379,6 @@ class AcsRbacIntegration(QontractReconcileIntegration[AcsRbacIntegrationParams])
         Updates desired ACS roles as well as associated access scopes and rules
 
         :param to_update: result of 'diff_iterables(current, desired).change' for ACS roles
-        :param acs: ACS api client
-        :param auth_id: id of auth provider within ACS instance to target for reconciliation
-        :param dry_run: run in dry-run mode
         """
         access_scope_id_map = {s.name: s.id for s in self.acs.get_access_scopes()}
         permission_sets_id_map = {
--- a/reconcile/acs_rbac.py
+++ b/reconcile/acs_rbac.py
@@ -168,7 +168,7 @@ class AcsRbacIntegration(QontractReconcileIntegration[AcsRbacIntegrationParams])
         role_assignments: RoleAssignments = self.build_role_assignments(auth_id, groups)
 
         for role in roles:
-            # process roles that are not system default 
+            # process roles that are not system default
             # OR
             # system default roles referenced in auth rules
             # however, do not reconcile the auth provider minimum access rule associated with 'None' system default
@@ -304,10 +304,9 @@ class AcsRbacIntegration(QontractReconcileIntegration[AcsRbacIntegrationParams])
                         f"Failed to create group(s) for role: {role.name}\t\n{e}"
                     )
                     continue
-            for a in role.assignments:
-                logging.info(
-                    f"Created rule assigning '{a.value}' to role '{role.name}'"
-                )
+            logging.info(
+                f"Added users to role '{role.name}': {[a.value for a in role.assignments]}"
+            )
 
     def delete_rbac(self, to_delete: dict[str, AcsRole], acs: AcsApi, dry_run: bool):
         access_scope_id_map = {s.name: s.id for s in acs.get_access_scopes()}
@@ -327,10 +326,9 @@ class AcsRbacIntegration(QontractReconcileIntegration[AcsRbacIntegrationParams])
                         f"Failed to delete group(s) for role: {role.name}\t\n{e}"
                     )
                     continue
-            for a in role.assignments:
-                logging.info(
-                    f"Deleted rule assigning '{a.value}' to role '{role.name}'"
-                )
+            logging.info(
+                f"Deleted users from role '{role.name}': {[a.value for a in role.assignments]}"
+            )
             # only reconcile rules associated with a system default role
             # do not delete the role and associated access scope
             if role.system_default:
@@ -356,23 +354,47 @@ class AcsRbacIntegration(QontractReconcileIntegration[AcsRbacIntegrationParams])
         self,
         to_update: dict[str, DiffPair[AcsRole, AcsRole]],
         acs: AcsApi,
+        auth_id: str,
         dry_run: bool,
     ):
-        # role_group_mappings = {}
-        # for group in acs.get_groups():
-        #    if group.role_name not in role_group_mappings:
-        #        role_group_mappings[group.role_name] = []
-        #    role_group_mappings[group.role_name].append(group)
+        role_group_mappings: dict[str[dict[str, str]]] = {}
+        for group in acs.get_groups():
+            if group.role_name not in role_group_mappings:
+                role_group_mappings[group.role_name] = {}
+            role_group_mappings[group.role_name][group.value] = group
 
         for role_diff_pair in to_update.values():
-            # check rules
             diff = diff_iterables(
                 role_diff_pair.current.assignments,
                 role_diff_pair.desired.assignments,
                 lambda x: x.value,
             )
-            print("RULE DIFFS")
-            print(diff)
+            old = [
+                role_group_mappings[role_diff_pair.current.name][d.value]
+                for d in diff.delete.values()
+            ]
+            new = [
+                AcsApi.GroupAdd(
+                    role_name=role_diff_pair.desired.name,
+                    key=a.key,
+                    value=a.value,
+                    auth_provider_id=auth_id,
+                )
+                for a in diff.add.values()
+            ]
+            if not dry_run:
+                try:
+                    acs.patch_group_batch(old, new)
+                except Exception as e:
+                    logging.error(
+                        f"Failed to update rules for role: {role_diff_pair.current.name}\t\n{e}"
+                    )
+                    continue
+            logging.info(
+                f"Updated rules for role '{role_diff_pair.desired.name}':\n\t"
+                + f"Added: {[n.value for n in new]}\n\t"
+                + f"Deleted: {[o.value for o in old]}"
+            )
 
     def run(
         self,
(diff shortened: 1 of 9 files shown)


reconcile/acs_rbac.py
```diff
@@ -0,0 +1,332 @@
...
+        vault_settings = get_app_interface_vault_settings()
+        secret_reader = create_secret_reader(use_vault=vault_settings.vault)
+        token = secret_reader.read_all_secret(instance.credentials)
+
+        acs = AcsApi(
+            instance={"url": instance.url, "token": token[instance.credentials.field]}
+        )
+
+        desired = self.get_desired_state(gqlapi.query)
+        existing = self.get_existing_state(acs, auth_id=instance.auth_provider.q_id)
```
patjlm, line 325: nit: we usually name this the `current` state, right?

reconcile/utils/acs_api.py
```diff
@@ -0,0 +1,384 @@
...
+        return AccessScope(response.json())
+
+    def get_access_scopes(self) -> list[AccessScope]:
+        response = self.generic_get_request("/v1/simpleaccessscopes")
+
+        access_scopes = []
+        for scope in response.json()["accessScopes"]:
+            access_scopes.append(AccessScope(scope))
+
+        return access_scopes
```
hemslo, line 327: same list comprehension here

reconcile/acs_rbac.py
```diff
@@ -0,0 +1,568 @@
...
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, AcsAccessScope):
+            return (
+                self.name == other.name
+                and self.description == other.description
+                and self.clusters == other.clusters
+                and self.namespaces == other.namespaces
+            )
+        return False
```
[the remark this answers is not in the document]
dwelch0: my understanding is that python's default behavior for `==` with custom classes is only comparison of the references and not data of attributes?

reconcile/acs_rbac.py
```diff
@@ -0,0 +1,568 @@
...
+            return []
+
+        desired_roles: dict[str, AcsRole] = {}
+        for user in query_results:
+            for role in user.roles or []:
+                for permission in role.oidc_permissions or []:
+                    if isinstance(permission, OidcPermissionAcsV1):
+                        # first encounter of specific permission
+                        # derive the Acs role specifics and add initial user
+                        if permission.name not in desired_roles:
```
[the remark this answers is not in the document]
dwelch0: Good point! `permission.name` in this context corresponds to [name within oidc-permission-1 file](https://github.com/app-sre/qontract-schemas/blob/d5b8f7c501f9411f2d799bd02c4520167d723718/schemas/access/oidc-permission-1.yml)

I think a decent solution would be to [enforce uniqueness for the name attribute in this schema](https://github.com/app-sre/qontract-schemas/pull/539/commits/d5b8f7c501f9411f2d799bd02c4520167d723718)

reconcile/utils/acs_api.py
```diff
@@ -0,0 +1,384 @@
...
+            url=f"{self.url}{path}",
+            headers={"Authorization": f"Bearer {self.token}"},
+            timeout=self.timeout,
+        )
+        try:
+            response.raise_for_status()
+        except requests.exceptions.RequestException as details:
+            raise requests.exceptions.RequestException(
+                f"Failed to perform GET request:\n\t{details}\n\t{response.text}"
+            )
```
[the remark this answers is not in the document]
cubismod: How so, Di? Potential of flooding logs or leaking sensitive details to the logs?

reconcile/utils/acs_api.py
```diff
@@ -0,0 +1,384 @@
...
+            response.raise_for_status()
+        except requests.exceptions.RequestException as details:
+            raise requests.exceptions.RequestException(
+                f"Failed to perform POST request with body:\n\t{json}\n\t{details}\n\t{response.text}"
+            )
+
+        return response
+
+    def generic_put_request(self, path: str, json: Any) -> requests.Response:
+        response = requests.put(
```
cubismod, line 161: Perhaps you could consolidate some of these request functions into a single function? Maybe like:
`def generic_request(self, path, request_type, json as an optional any)`

To cut down on duplicated code.

reconcile/utils/acs_api.py
```diff
@@ -0,0 +1,384 @@
+from typing import Any
```
cubismod, line 1: Is there any reference for the ACS "api" (in very loose terms), that you could add as a comment? Just for clarity to anyone trying to understand the API

cubismod (commented):
First level review from me

reconcile/utils/acs_api.py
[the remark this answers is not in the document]
hemslo: Main concern is sensitive data, and this is a generic method, very easy to dump unwanted data into log, then forwarded to multiple places.

reconcile/acs_rbac.py
```diff
@@ -0,0 +1,568 @@
...
+
+        :param query_func: function which queries GQL server
+        :return: list of AcsRole derived from oidc-permission-1 definitions
+        """
+
+        query_results = acs_rbac_query(query_func=query_func).acs_rbacs
+        if query_results is None:
+            return []
+
+        desired_roles: dict[str, AcsRole] = {}
```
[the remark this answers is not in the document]
dwelch0: Great point. I modeled this slightly different from original suggestion to avoid duplicating the permission object. This made the `build` code more intuitive in my opinion because it now only receives an object containing one `permission` and a list of users associated with the permission.

reconcile/acs_rbac.py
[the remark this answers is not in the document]
hemslo: It depends on how `__eq__` is implemented in base class, `BaseModel` has a `__eq__` defined [here](https://github.com/pydantic/pydantic/blob/2e939dc3bfb88f54efb66f8f1a031ff22e4f9865/pydantic/main.py#L907)

reconcile/acs_rbac.py
```diff
@@ -0,0 +1,568 @@
...
+                            role.access_scope.description,
+                            role.access_scope.clusters,
+                            role.access_scope.namespaces,
+                        )
+                    except Exception as e:
+                        logging.error(
+                            f"Failed to create access scope: {role.access_scope.name} for role: {role.name}\t\n{e}"
+                        )
+                        continue
+                logging.info(f"Created access_scope '{role.access_scope.name}'")
```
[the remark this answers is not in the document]
dwelch0: should this refactor be applied to the `logging.error` statements as well or due to the higher log level is that unnecessary?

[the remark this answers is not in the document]
hemslo: To make them consistent, better to use one style in the file.

reconcile/acs_rbac.py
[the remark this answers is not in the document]
hemslo: If `permission` uniq is ensured, then ideally it's a frozen dict that hash is auto generated, can be used like `data[permission].append(user)`, but our auto generated model doesn't have that flag set, so we can inherit it to make it work:

```python

class Permission(OidcPermissionAcsV1):
    def __hash__(self):
        return hash(self.name)

...

        permission_with_users = defaultdict(list)
        for user in query_results:
            for role in user.roles or []:
                for permission in role.oidc_permissions or []:
                    if isinstance(permission, OidcPermissionAcsV1):
                        permission_with_usernames[
                            Permission(**permission.dict(by_alias=True))
                        ].append(user.org_username)
        return [
            AcsRole.build(permission, usernames)
            for permission, usernames in permission_with_usernames.items()
        ]
```

btw, you can directly use `-> Self` instead of `-> "AcsRole"`.

reconcile/acs_rbac.py
```diff
@@ -0,0 +1,568 @@
...
+            if group.role_name not in role_group_mappings:
+                role_group_mappings[group.role_name] = {}
+            role_group_mappings[group.role_name][group.value] = group
+
+        for role_diff_pair in to_update.values():
+            # auth rule (groups) portion
+            diff = diff_iterables(
+                role_diff_pair.current.assignments,
+                role_diff_pair.desired.assignments,
+                lambda x: x.value,
```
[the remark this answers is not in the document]
dwelch0: Yes. In this case `key` corresponds to the attribute to specify for an auth rule (exs: `roleid`, `email`, `org_username`, etc. This integration will create rules using `org_username` as the key and value will be the unique username values.

reconcile/acs_rbac.py
```diff
@@ -0,0 +1,568 @@
...
+            )
+            # due to usage of 'value' in auth rules for the key, if a single rule requires a change
+            # it will appear as one entry to delete and one entry to add
+            # ex: desired value = foo. current value = bar
+            # output will be an entry to delete bar and an entry to add foo
+            if any(len(lst) > 0 for lst in [diff.add, diff.delete, diff.change]):  # type: ignore
```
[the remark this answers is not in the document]
dwelch0: Thanks for catching. `diff.change` should not be necessary here given usage of `lambda x: x.value` for iterables of `AssignmentPair`, which only contains `key` and `value` attributes. `key` will always be `org_username` for current use case so `diff.change` will always be empty.

reconcile/acs_rbac.py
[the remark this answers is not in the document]
hemslo: So you mean in `assignments`, `key` is always `org_username`, but `value` can be different username values, right? Then it looks good as long as `value` is unique across all items inside `assignments`. And based on this data structure, it can only be `add`, delete` and `identical`, no `change`.

reconcile/utils/acs_api.py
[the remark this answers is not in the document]
dwelch0: `response.text` was helpful during initial dev debugging but yes agree with concern and removal.

reconcile/acs_rbac.py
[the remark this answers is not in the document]
dwelch0: Correct. `org_username` corresponds to https://github.com/app-sre/qontract-schemas/blob/main/schemas/access/user-1.yml#L16

reconcile/acs_rbac.py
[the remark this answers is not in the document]
dwelch0: Very nice! Thanks

reconcile/utils/acs_api.py
dwelch0: I will double-check with ACS time but afaik there's no publicly accessible docs. I had to make an authenticated request to `/api/docs/swagger` to my test ACS instance to retrieve docs.

reconcile/utils/acs_api.py
dwelch0: Yep, good call. Kept appending request functions as I progressed through the reconcile operations and completely forgot to circle back on the sprawl of redundant code :sweat_smile: 

Pushed an update to consolidate

reconcile/utils/acs_api.py
cubismod: lol that's fun!

reconcile/utils/acs_api.py
```diff
@@ -0,0 +1,384 @@
...
+        self,
+        instance: Any,
+        timeout: int = 30,
+    ) -> None:
+        self.url = instance["url"]
+        self.token = instance["token"]
+        self.timeout = timeout
+
+    def generic_get_request(self, path: str) -> requests.Response:
+        response = requests.get(
```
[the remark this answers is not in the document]
dwelch0: agree with inclusion of a `self.session = requests.Session()` within the init for `AcsApi` but confused by the suggested usage:

```python
with AcsApi(...) as acs_api:
  ...
```
would this not create a new session each time?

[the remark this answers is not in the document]
cubismod: A session would be created/destroyed for each reconcile of the acs rbac integration which is much less than for each http request the acs api makes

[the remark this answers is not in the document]
cubismod: that with statement could then be added to the run function: https://github.com/app-sre/qontract-reconcile/pull/3899/files#diff-9de8e729bc5a20136ab30c1beda6f0f512d64c80d551234d25d34e1ebb41874bR554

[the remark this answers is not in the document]
dwelch0: thanks for clarifying. Misinterpretation on my end regarding usage of the `with`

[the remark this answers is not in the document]
dwelch0: update pushed. lmk if I misinterpreted

reconcile/acs_rbac.py
```diff
@@ -0,0 +1,558 @@
...
+            or self.description != b.description
+        )
+
+    def is_unrestricted_scope(self) -> bool:
+        # empty cluster and namespaces attributes signifies unrestricted scope
+        return not self.access_scope.clusters and not self.access_scope.namespaces
+
+
+class AcsRbacIntegrationParams(PydanticRunParams):
+    thread_pool_size: int
```
hemslo, line 135: Is this field used anywhere?

reconcile/acs_rbac.py
```diff
@@ -0,0 +1,558 @@
...
+                except Exception as e:
+                    logging.error(
+                        "Failed to retrieve current permission set: %s for role: %s\n\t%s",
+                        role.permission_set_id,
+                        role.name,
+                        e,
+                    )
+                    continue
+
+                current_roles[role.name] = AcsRole(
```
hemslo, line 235: `current_roles` is a `dict` and only values are returned, are we expecting there are duplicate `role.name` in `roles`, follow the current logic that last one wins, is this expected?

reconcile/utils/acs_api.py
```diff
@@ -0,0 +1,328 @@
...
+        elif verb == "POST":
+            response = self.session.put(
+                url, headers=headers, json=json, timeout=self.timeout
+            )
+        elif verb == "PUT":
+            response = self.session.post(
+                url, headers=headers, json=json, timeout=self.timeout
+            )
+        else:
+            raise ValueError(f"Unsupported request type: {verb}")
```
hemslo, line 158: `session` has a `request` method doing the same thing

```python
        response = self.session.request(
            verb,
            url,
            headers=headers,
            json=json,
            timeout=self.timeout,
        )
```

since this method is used internally in this class, skip verb validation is ok, or use typing to do the checking:

```python
Verb: TypeAlias = Literal["GET", "POST", "PUT", "DELETE"]

    def generic_request(
        self, path: str, verb: Verb, json: Optional[Any] = None
    ) -> requests.Response:
    ...
```

reconcile/utils/acs_api.py
```diff
@@ -0,0 +1,328 @@
...
+        # attributes defined within stackrox(ACS) API for GET /v1/groups
+        check_len_attributes(["roleName", "props"], api_data)
+        try:
+            check_len_attributes(
+                ["id", "authProviderId", "key", "value"], api_data["props"]
+            )
+        except ValueError as e:
+            # it is valid for the default None group to contain empty key/value
+            if api_data["roleName"] != "None":
+                raise e
```
hemslo, line 56: Instead of using error handling to change control flow, how about skip check explicitly?

```python
        if api_data["roleName"] != "None":
            check_len_attributes(
                ["id", "authProviderId", "key", "value"], api_data["props"]
            )
```

reconcile/utils/acs_api.py
```diff
@@ -0,0 +1,328 @@
...
+    def __init__(self, api_data: Any) -> None:
+        # attributes defined within stackrox(ACS) API for GET /v1/simpleaccessscopes/{id}
+        unrestricted = False
+        try:
+            check_len_attributes(["id", "name", "rules"], api_data)
+        except ValueError as e:
+            # it is valid for the default Unrestricted access scope to have null `rules`
+            if api_data.get("name") != "Unrestricted":
+                raise e
+            unrestricted = True
```
hemslo, line 83: Similar here, skip check when it's `Unrestricted`. What's the relationship between `name` is `Unrestricted` and failed check? When `name` is `Unrestricted` but check pass, do you expect `unrestricted` to be `True` or `False`?

If you just want to do a nullable access, you can use
`api_data.get("rules", {}).get("includedClusters", [])`

hemslo (commented):
Regarding error handling, it seems all exceptions are caught and continue, which means this integration will never fail, is this expected? Otherwise need to collect errors and report, need an error test case. Other parts looking good!

dwelch0:
Upon further testing after receiving an updated oidc client that returned custom `org_username` attribute within JWTs, I've uncovered that although the ACS `GroupService` api allows creation of groups (auth rules) with custom key values (ex: `org_username`), these rules don't appear to be enforced during auth process.

I've inquired to team about this behavior but in mean time I've pushed update to create rules using `email` as the key. Additionally, [update pushed to qontrat-schema](https://github.com/app-sre/qontract-schemas/pull/539/commits/50ed5030d1b2c65e4c07ef791aa2f8e72b8e96e3) to support this. Less than ideal to require user files to now include an additional attribute but this can be reverted in future once enforcement of custom auth rule keys is understood/enabled.

reconcile/acs_rbac.py
dwelch0: Thanks for catching. Initial implementation of CRUD operations for rules utilized this but later found support for batch and dropped threading for the calls.

dwelch0: [addressed here](https://github.com/app-sre/qontract-reconcile/pull/3899/commits/657c3818a33f5342ce10506c131f61ab3fdf7426)

reconcile/utils/acs_api.py
dwelch0: `Unrestricted` is returned as:
```json
    {
      "id": "ffffffff-ffff-fff4-f5ff-ffffffffffff",
      "name": "Unrestricted",
      "description": "Access to all clusters and namespaces",
      "rules": null,
      "traits": {
        "mutabilityMode": "ALLOW_MUTATE",
        "visibility": "VISIBLE",
        "origin": "DEFAULT"
      }
    }
```

whereas all other access scopes contain following attributes within `rules`:
```json
{
      "id": "1b10220a-5fbc-4ad8-afcd-cd9ea3f828fe",
      "name": "ns-exclusives",
      "description": "example of assigning permission at ns scope",
      "rules": {
        "includedClusters": [],
        "includedNamespaces": [
          {
            "clusterName": "some-cluster",
            "namespaceName": "some-ns"
          }
        ],
        "clusterLabelSelectors": [],
        "namespaceLabelSelectors": []
      },
      "traits": null
    }
```
`When name is Unrestricted but check pass` given the api response behavior above, this should always fail when attempting to check `rules`

reconcile/acs_rbac.py
dwelch0: Not expecting duplicates. Role name uniquess is enforced in ACS:
image

reconcile/utils/acs_api.py
dwelch0: control flow update [here](https://github.com/app-sre/qontract-reconcile/pull/3899/commits/46e19b701c0a7159d6e36acec751026c60c589ce)

reconcile/utils/acs_api.py
dwelch0: [updated](https://github.com/app-sre/qontract-reconcile/pull/3899/commits/26f6e065ac9c73596206f8b7e63a95cc0bf77dc0)

reconcile/utils/acs_api.py
dwelch0: [updated](https://github.com/app-sre/qontract-reconcile/pull/3899/commits/4546a1fd9cb8560d436edea945dc40fe4207b584)

dwelch0:
> Regarding error handling, it seems all exceptions are caught and continue, which means this integration will never fail, is this expected? Otherwise need to collect errors and report, need an error test case. Other parts looking good!

Current usages of `continue`s was deliberate attempt to allow progression of reconciliation to additional roles if issue is encountered during operations on one. But upon review of existing prom rules around qr integrations, I believe I misinterpreted and these errors will not increment a metric for alerting. 

I will refactor and add an error test

reconcile/utils/acs_api.py
hemslo: to avoid mutate `unrestricted`, can be like this:

```python
        unrestricted = api_data["name"] == "Unrestricted"
        if not unrestricted:
            check_len_attributes(["rules"], api_data)
```

reconcile/acs_rbac.py
hemslo: Then you can use `current_roles` as `list[AcsRole]`, do `current_roles.append(...` and return it directly.

reconcile/test/test_acs_rbac.py
```diff
@@ -0,0 +1,714 @@
...
+        acs_api.PermissionSet(
+            api_data={
+                "id": "3",
+                "name": "Vulnerability Management Admin",
+            }
+        ),
+    ]
+
+
+def test_get_desired_state(mocker, query_data_desired_state, modeled_acs_roles):
```
hemslo, line 360: Since this is a new test file, it would be good to have types in all methods.

dwelch0: [updated](https://github.com/app-sre/qontract-reconcile/pull/3899/commits/30c10c560d2dab2db4a5ee098b2c3715ab5c8448)

reconcile/utils/acs_api.py
dwelch0: [update](https://github.com/app-sre/qontract-reconcile/pull/3899/commits/ae91b7eff0e453ef0861bf9a84ea2b3e0c9d1586)

reconcile/acs_rbac.py
dwelch0: hmm. Yeah unsure why I thought usage of a dict was necessary here. Thanks
[updated](https://github.com/app-sre/qontract-reconcile/pull/3899/commits/aba611a97e850e5b8b93d32b91922c671cc62443)

dwelch0:
@hemslo have a feeling you're going to have some requested changes to [error handling refactor](https://github.com/app-sre/qontract-reconcile/pull/3899/commits/3a753713c8599a25ed38f5166ee2610bb655bae5) but this is roughly the pattern I'd like to go implement.

imo its desirable for the reconcile run to continue with additional `roles` if an issue is encountered with one. Intended behavior here is for the bad role to be skipped, allow progression to reconciliation of any remaining roles, but still exit with an error.

reconcile/acs_rbac.py
```diff
@@ -0,0 +1,562 @@
...
+        with AcsApi(
+            instance={"url": instance.url, "token": token[instance.credentials.field]}
+        ) as acs_api:
+            current = self.get_current_state(acs_api, instance.auth_provider.q_id)
+            errors_occurred = self.reconcile(
+                desired, current, acs_api, instance.auth_provider.q_id, dry_run
+            )
+
+        if errors_occurred:
+            sys.exit(ExitCodes.ERROR)
```
hemslo, line 562: For new integrations, we don't use `sys.exit` anymore, just throw an exception like `RuntimeError` or any other custom exception. You can also include the actual number of errors in the error message too.

That `errors_occurred = 0` is redundant as the value is always the return of `self.reconcile`. 

`err_count += 1` always have a following `continue`, if you want to remove that duplication, can extract process single role to a function, directly throw the exception, catch it like this:

```python
        for role in to_add.values():
            try:
                add_rbac_for_role(role, ...)
            except Exception as e:
                logging.error(...)
                err_count += 1
```

If you want to try the new feature of `ExceptionGroup` to collect all exceptions, you can use it to replace custom `error_count`, like this

```python
        errors = []
        for role in to_add.values():
            try:
                add_rbac_for_role(role, ...)
            except Exception as e:
                errors.append(e)
        if errors:
            raise ExceptionGroup("Add role errors", errors)
```

Or you can choose throw it at a higher level in `reconcile`

```python
        errors = []
        diff = diff_iterables(current, desired, lambda x: x.name)
        if diff.add:
            errors.extend(self.add_rbac(diff.add, acs, auth_provider_id, dry_run))
        if diff.delete:
            errors.extend(self.delete_rbac(diff.delete, acs, auth_provider_id, dry_run))
        if diff.change:
            errors.extend(self.update_rbac(diff.change, acs, auth_provider_id, dry_run))
        if errors:
            raise ExceptionGroup("Reconcile errors occurred", errors)
```

That will make `run` logic clean as it doesn't need to care about error handling anymore.

If the error message in each function call like `acs.create_access_scope` is good enough, then you don't need to catch and rethrow, just write success path handling. You may find the new extracted function can further improve readability as it only need related value passed in like `access_scope_id_map[DEFAULT_ADMIN_SCOPE_NAME]` and `permission_sets_id_map[role.permission_set_name]`.


Btw I just noticed that `acs.get_access_scopes()`, `acs.get_permission_sets()` and `get_groups()` called multiple times, you may want to call it once and pass to later usages. I don't know the relationship between `get_access_scope_by_id` and `get_access_scopes`, will `get_access_scopes` return all data without pagination? `get_access_scope_by_id` is called in a loop to fetch one by one, can it be covered by `get_access_scopes`?

dwelch0: Regarding the usage of `AcsApi` calls to obtain existing state, you're right, this could be reduced significantly.

[refactored](https://github.com/app-sre/qontract-reconcile/pull/3899/commits/634292c48c1788dd0a51cceb8aa350a5cda94c05)

dwelch0: Thanks for the suggestions for error handling. Much cleaner approach.

 [error handling refactor](https://github.com/app-sre/qontract-reconcile/pull/3899/commits/93d7313e6c8135b75c6af2c5f431de941abcfe7a)

hemslo (approved):
LGTM! Thank you for the effort!

dwelch0:
Thanks for all the great feedback and patience. It has been a good dose of medicine finally contributing more than simple  patch to QR :sweat_smile:

dwelch0:
Note: was able to [revert to using org_username](https://github.com/app-sre/qontract-reconcile/pull/3899/commits/14d4771858b2c67658442bbf96b97b65b5f338ff) and avoid addition of `acs_user` attribute to user files. A change to jwt claim mappings for the auth provider within our ACS instance was made to enable usage of `org_username` mappings to `userid`

PR #3902, opened 2023-10-31
Use FIPS endpoints with AWS for GovCloud Accounts

cubismod:
[APPSRE-8486](https://issues.redhat.com/browse/APPSRE-8486)

Updates the AWS API wrapper to configure any client for a GovCloud account to use FIPS endpoints in order to comply with FedRamp compliance requirements. GovCloud accounts are determined by checking the `partition` value in the `account.yml` file for each. The reason being is that we have some non GovCloud accounts in FedRamp so we cannot enable FIPS across the board for all accounts.

Tested locally against FedRamp App Interface with a variety of AWS based integrations (terraform-resources, users, etc.). Should not affect commercial at all since we have no us-gov partition accounts there but I will do a rollout to stage before promoting to prod in commercial.

Also adds a note to the README about libpq and Fedora


dwelch0 (approved):
LGTM

PR #3906, opened 2023-11-03
Terraform cloudflare resources custom ssl certificate

esron:
Related to: https://issues.redhat.com/browse/APPSRE-8315
Needs to be tested with: https://github.com/app-sre/qontract-schemas/pull/532

This adds support to create [custom_ssl_certificate](https://registry.terraform.io/providers/cloudflare/cloudflare/3.32.0/docs/resources/custom_ssl) for Cloudflare.

### DIFF
--- a/reconcile/terraform_cloudflare_resources.py
+++ b/reconcile/terraform_cloudflare_resources.py
@@ -20,7 +20,6 @@ from reconcile.gql_definitions.terraform_cloudflare_resources.terraform_cloudfla
     TerraformCloudflareAccountsQueryData,
 )
 from reconcile.gql_definitions.terraform_cloudflare_resources.terraform_cloudflare_resources import (
-    CloudflareCustomSSLCertificateV1,
     NamespaceTerraformProviderResourceCloudflareV1,
     NamespaceTerraformResourceCloudflareZoneV1,
     NamespaceV1,
@@ -309,11 +308,10 @@ def _filter_cloudflare_namespaces(
     return cloudflare_namespaces
 
 
-def _filter_custom_ssl_certificate(
+def _populate_custom_ssl_certificate(
     namespaces: Iterable[NamespaceV1],
-) -> list[CloudflareCustomSSLCertificateV1]:
-    certificates: list[CloudflareCustomSSLCertificateV1] = []
-
+    secret_reader: SecretReaderBase,
+) -> None:
     for ns in namespaces:
         if ns.external_resources:
             for ext_resource in ns.external_resources:
@@ -328,8 +326,13 @@ def _filter_custom_ssl_certificate(
                             isinstance(zone, NamespaceTerraformResourceCloudflareZoneV1)
                             and zone.custom_ssl_certificates
                         ):
-                            certificates += zone.custom_ssl_certificates
-    return certificates
+                            for cert in zone.custom_ssl_certificates:
+                                cert_data = secret_reader.read_secret(
+                                    cert.certificate_secret.certificate
+                                )
+                                key_data = secret_reader.read_secret(
+                                    cert.certificate_secret.key
+                                )
 
 
 @defer
@@ -370,7 +373,7 @@ def run(
         logging.info("No cloudflare namespaces were detected, nothing to do.")
         sys.exit(ExitCodes.SUCCESS)
 
-    custom_ssl_certificates = _filter_custom_ssl_certificate(cloudflare_namespaces)
+    # _filter_custom_ssl_certificate(cloudflare_namespaces, secret_reader)
 
     # Build Cloudflare clients
     cf_clients = TerraformConfigClientCollection()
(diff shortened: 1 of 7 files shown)


reconcile/test/test_terraform_cloudflare_resources.py
```diff
@@ -269,3 +358,51 @@ def test_cloudflare_namespace_validation(
...
+    if {
+        "path": "certificate/secret/cert/path",
+        "field": "certificate.key",
+        "version": 1,
+        "q_format": "plain",
+    } == args[0]:
+        return "----- KEY -----"
+
+
+def test_terraform_cloudflare_resources_dry_run(
```
bhushanthakur93, line 382: From the test implementation it looks like the purpose of this test is to verify we don't run terraform apply. In that case, you don't need any specific implementation for custom ssl resource in `custom_ssl_secret_reader_side_effect`. 

Also take a look at `test_terrascript_cloudflare_client.py` , this is where we actually test the terraform resource generation. You can add a test there for `cloudflare_custom_ssl` resource and use mocking for secret reader there with the side effect logic you have  in `custom_ssl_secret_reader_side_effect`

reconcile/test/test_terraform_cloudflare_resources.py
```diff
@@ -269,3 +358,51 @@ def test_cloudflare_namespace_validation(
...
+        "reconcile.utils.terrascript.cloudflare_resources.create_secret_reader",
+        autospec=True,
+    )
+    create_secret_reader.return_value = secret_reader
+    with pytest.raises(SystemExit) as sample:
+        integ.run(True, None, False, 10)
+    assert sample.value.code == ExitCodes.SUCCESS
+    assert mock_terraform_client.called is True
+    assert call().apply() not in mock_terraform_client.method_calls
+    assert True
```
bhushanthakur93, line 408: You can remove this

bhushanthakur93 (changes requested):
Took a first pass at this, let me know if this makes sense.

reconcile/test/test_terraform_cloudflare_resources.py
bhushanthakur93: just found the similar example here https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/utils/test_terrascript_cloudflare_client.py#L433

reconcile/test/test_terraform_cloudflare_resources.py
```diff
@@ -72,7 +79,7 @@ def external_resources(provisioner_config):
         provisioner=provisioner_config,
         resources=[
             NamespaceTerraformResourceCloudflareZoneV1(
-                provider="cloudflare_zone",
+                provider="zone",
```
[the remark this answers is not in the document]
esron: `provider="cloudflare_zone"` raises an `UnsupportedCloudflareResourceError` error in https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/terrascript/cloudflare_resources.py#L99C15-L99C15

reconcile/utils/terrascript/cloudflare_resources.py
```diff
@@ -202,6 +241,7 @@ def populate(self) -> list[Union[Resource, Output]]:
         zone_records = values.pop("records", [])
         zone_workers = values.pop("workers", [])
         zone_certs = values.pop("certificates", [])
+        zone_custom_sll = values.pop("custom_ssl_certificates", [])
```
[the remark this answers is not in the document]
esron: Fixed, thanks.

reconcile/terraform_cloudflare_resources.py
```diff
@@ -345,6 +345,8 @@ def run(
         logging.info("No cloudflare namespaces were detected, nothing to do.")
         sys.exit(ExitCodes.SUCCESS)
 
+    # _filter_custom_ssl_certificate(cloudflare_namespaces, secret_reader)
```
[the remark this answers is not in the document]
esron: Fixed, thanks.

reconcile/test/test_terraform_cloudflare_resources.py
esron: I've copied this pattern from https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/test_terraform_cloudflare_users.py#L465

which is similar in use and behavior. I was to lazy to create the `expected_args` because the resource definition for this one was way bigger.

Also to remove the `custom_ssl_secret_reader_side_effect` I would need to remove the `custom_ssl` from the resources mocked as well.

reconcile/test/test_terraform_cloudflare_resources.py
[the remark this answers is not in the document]
esron: Before the test that runs the full integration, that error never appeared.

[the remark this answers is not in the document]
bhushanthakur93: ah, ok. We had the wrong fixture that never got tested but with your test we caught it. Nice!

reconcile/test/test_terraform_cloudflare_resources.py
bhushanthakur93: ah right, makes sense!

IMO, you can also add a test in `test_terrascript_cloudflare_client` to actually verify terraform resource structure. That will give us more confident that we generate the right resource.

esron: Done.

bhushanthakur93 (approved):
LGTM

PR #3910, opened 2023-11-07
[AUS] expose version data metrics

maorfr:
with these metrics we will be able to show the experience we have gained with different versions. this will help demonstrate the value we provide as customer zero.

### DIFF
--- a/reconcile/aus/metrics.py
+++ b/reconcile/aus/metrics.py
@@ -108,3 +108,15 @@ class AUSOrganizationErrorRate(ErrorRateMetricSet):
                 org_id=org_id,
             ),
         )
+
+
+class AUSOrganizationVersionDataGauge(AUSBaseMetric, GaugeMetric):
+    """Gauge for the version data for an OCM organization"""
+
+    org_id: str
+    version: str
+    workload: str
+
+    @classmethod
+    def name(cls) -> str:
+        return "aus_organization_version_data"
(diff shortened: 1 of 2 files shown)


geoberle (approved):
thanks. LGTM

reconcile/aus/ocm_upgrade_scheduler.py
```diff
@@ -79,6 +83,25 @@ def get_ocm_env_upgrade_specs(
...
         )
 
+    def expose_version_data_metrics(
+        self,
+        ocm_env: str,
+        org_id: str,
+        version_data: VersionData,
+    ) -> None:
+        for version, version_history in version_data.versions.items():
+            for workload, workload_histry in version_history.workloads.items():
```
patjlm, line 93: nit: typo `workload_history`

I am feeling so useful there..

reconcile/aus/metrics.py
```diff
@@ -108,3 +108,15 @@ def __init__(self, integration: str, org_id: str, ocm_env: str) -> None:
                 org_id=org_id,
             ),
         )
+
+
+class AUSOrganizationVersionDataGauge(AUSBaseMetric, GaugeMetric):
```
patjlm, line 113: Why is this a Gauge and not a Counter? Do we expect soak days can go down?

maorfr: It's a value that comes from the state, so it's also not "counted".

reconcile/aus/ocm_upgrade_scheduler.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/3912

PR #3911, opened 2023-11-08
skip namespaces with targetNS in dashdotdb-slo

jmelis:
Renames `namespaceTarget` to `SLONamespace` according to:
https://github.com/app-sre/qontract-schemas/pull/548

In addition it adds a temporary workaround that skips namespaces with `SLONamespace` defined until https://issues.redhat.com/browse/APPSRE-8513 is implemented.

### DIFF
--- a/reconcile/dashdotdb_slo.py
+++ b/reconcile/dashdotdb_slo.py
@@ -99,10 +99,9 @@ class DashdotdbSLO(DashdotdbBase):
         LOG.debug("SLO: processing %s", slo_document.name)
         result: list[ServiceSLO] = []
         for namespace_access in slo_document.namespaces:
-            # skip if there is a `targetNamespace` defined. in this case we
-            # don't know which is the prometheus that must evaluate the query.
-            # This functionlity is implemented here instead:
-            # app-interface/docs/status-board/statusboard-alertmanager-receiver.md
+            # TODO: APPSRE-8513 Dashdotdb SLO collector should deal with
+            # namespaceTargets This `if` is a temporary workaround until
+            # APPSRE-8513 is implemented.
             if namespace_access.target_namespace:
                 continue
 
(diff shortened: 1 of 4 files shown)


reconcile/dashdotdb_slo.py
```diff
@@ -99,6 +99,13 @@ def _get_service_slo(self, slo_document: SLODocumentV1) -> list[ServiceSLO]:
         LOG.debug("SLO: processing %s", slo_document.name)
         result: list[ServiceSLO] = []
         for namespace_access in slo_document.namespaces:
+            # skip if there is a `targetNamespace` defined. in this case we
+            # don't know which is the prometheus that must evaluate the query.
+            # This functionlity is implemented here instead:
```
rporres, line 104:
```suggestion
            # This functionality is implemented here instead:
```

reconcile/dashdotdb_slo.py
```diff
@@ -99,6 +99,13 @@ def _get_service_slo(self, slo_document: SLODocumentV1) -> list[ServiceSLO]:
         LOG.debug("SLO: processing %s", slo_document.name)
         result: list[ServiceSLO] = []
         for namespace_access in slo_document.namespaces:
+            # skip if there is a `targetNamespace` defined. in this case we
+            # don't know which is the prometheus that must evaluate the query.
```
rporres, line 103:
```suggestion
            # don't know which is the Prometheus that must evaluate the query.
```

reconcile/dashdotdb_slo.py
rporres: I don't get this, this functionality is not implemented anywhere. We're not collecting for this kind of namespaces...

I honestly don't like the `targetNamespace` denomination for something that looks so RHOBS specific and that appears to be bound to how we process rules in that scenario...

Anyway, at least I'd just add a TODO here and a ticket...

rporres (approved):

PR #3913, opened 2023-11-08
APPSRE-8300: Optionally silence successful deploy notices

lisa:
Sometimes teams might want to be notified only on successful notifications. This allows for that.

There is a companion schemas PR to come shorlty.

### DIFF
--- a/reconcile/openshift_saas_deploy.py
+++ b/reconcile/openshift_saas_deploy.py
@@ -67,8 +67,16 @@ def slack_notify(
     in_progress: bool,
     trigger_integration: Optional[str] = None,
     trigger_reason: Optional[str] = None,
+    skip_successful_notifications: Optional[bool] = False,
 ) -> None:
     success = not ri.has_error_registered()
+    # if the deployment doesn't want any notifications for successful
+    # deployments, then we should grant the wish. However, there's a user
+    # expereince concern where the deployment owners will receive a "in
+    # progress" notice but no subsequent notice. We handle this case by
+    # including an "fyi" message for in progress deployments down below.
+    if success and skip_successful_notifications and not in_progress:
+        return
     if in_progress:
         icon = ":yellow_jenkins_circle:"
         description = "In Progress"
@@ -87,6 +95,8 @@ def slack_notify(
         message += f". Reason: {trigger_reason}"
     if trigger_integration:
         message += f" triggered by _{trigger_integration}_"
+    if in_progress and skip_successful_notifications:
+        message += f". There will not be a notice for success."
     slack.chat_post_message(message)
 
 
@@ -120,6 +130,9 @@ def run(
     # - this is not a dry run
     # - there is a single saas file deployed
     notify = not dry_run and len(saas_files) == 1
+    skip_successful_deploy_notifications = False
+    if len(saas_files) > 0:
+        skip_successful_deploy_notifications = saas_files[0].skip_successful_deploy_notifications
     slack = None
     if notify:
         saas_file = saas_files[0]
@@ -151,6 +164,7 @@ def run(
                         in_progress=False,
                         trigger_integration=trigger_integration,
                         trigger_reason=trigger_reason,
+                        skip_successful_notifications=skip_successful_deploy_notifications,
                     )
                 )
             # deployment start notification
@@ -164,6 +178,7 @@ def run(
                     in_progress=True,
                     trigger_integration=trigger_integration,
                     trigger_reason=trigger_reason,
+                    skip_successful_notifications=skip_successful_deploy_notifications,
                 )
 
     jenkins_map = jenkins_base.get_jenkins_map()
--- a/reconcile/typed_queries/saas_files.py
+++ b/reconcile/typed_queries/saas_files.py
@@ -109,6 +109,9 @@ class SaasFile(ConfiguredBaseModel):
     pipelines_provider: Union[PipelinesProviderTektonV1, PipelinesProviderV1] = Field(
         ..., alias="pipelinesProvider"
     )
+    skip_successful_deploy_notifications: Optional[bool] = Field(
+        ..., alias="skipSuccessfulDeployNotifications"
+    )
     deploy_resources: Optional[DeployResourcesV1] = Field(..., alias="deployResources")
     slack: Optional[SlackOutputV1] = Field(..., alias="slack")
     managed_resource_types: list[str] = Field(..., alias="managedResourceTypes")


geoberle:
i think you need to modify the query in `reconcile/gql_definitions/common/saas_files.gql` as well and run `make qenerate` which will modify `saas_files.py` for you

lisa:
> i think you need to modify the query in `reconcile/gql_definitions/common/saas_files.gql` as well and run `make qenerate` which will modify `saas_files.py` for you

How do you think this PR looks in this state (with its companion schemas change; linked in the original comment now)?

reconcile/openshift_saas_deploy.py
```diff
@@ -67,8 +67,17 @@ def slack_notify(
...
+    skip_successful_notifications: Optional[bool] = False,
 ) -> None:
     success = not ri.has_error_registered()
+    # if the deployment doesn't want any notifications for successful
+    # deployments, then we should grant the wish. However, there's a user
+    # expereince concern where the deployment owners will receive a "in
+    # progress" notice but no subsequent notice. We handle this case by
+    # including an "fyi" message for in progress deployments down below.
+    if success and skip_successful_notifications and not in_progress:
+        logging.info(f"Skipping Slack notification for {saas_file_name} to {env_name} because deploy was successful.")
```
chassing, line 79: I would use `logging.debug` instead. Those info logs will be shown in our #sd-app-sre-econcile slack channel.

reconcile/openshift_saas_deploy.py
```diff
@@ -120,6 +131,9 @@ def run(
     # - this is not a dry run
     # - there is a single saas file deployed
     notify = not dry_run and len(saas_files) == 1
+    skip_successful_deploy_notifications = False
+    if len(saas_files) > 0:
+        skip_successful_deploy_notifications = saas_files[0].skip_successful_deploy_notifications
```
chassing, line 136: Can be written as

```python
skip_successful_deploy_notifications = saas_files[0].skip_successful_deploy_notifications if saas_files else False
```

reconcile/openshift_saas_deploy.py
lisa: @chassing Ooh, thanks. I didn't realize those make it into Slack. Is `debug` log level the preferred way to put these messages into the pods' logs (or elsewhere so they are generally available if one went to look for them)?

bhushanthakur93:
Could you add unit tests for the new code in openshift_saas_deploy ?

reconcile/openshift_saas_deploy.py
OliviaHY, line 136: What would happen when `skip_successful_deploy_notifications` is not set in the saas file?

hemslo: `saas_files` already check in above code and raised exception, and `notify` will only be set to `True` if there is only one saas_file, so  `skip_successful_deploy_notifications` is not needed here, can directly use `saas_file.skip_successful_deploy_notifications` in `slack_notify` calls.

reconcile/openshift_saas_deploy.py
hemslo: It maybe ok here to use info log, as this integration is run tekton, the log message won't show in slack channel.

chassing: Ah yes, you're right. I missed that point 🙇

reconcile/openshift_saas_deploy.py
lisa: The codebase, here and elsewhere, is expected to behave as before this PR. Semantically, it means that integrations need to opt into the skip notification.

lisa:
> > Could you add unit tests for the new code in openshift_saas_deploy ?
> 
> I will figure out how to do this and add them.

OK, it took forever, but there are tests with c6c9561.

reconcile/test/test_openshift_saas_deploy.py
```diff
@@ -86,3 +93,130 @@ def test_compose_console_url_with_long_saas_name(
...
+    api = slack_api.SlackApi(
+        workspace_name="test",
+        token="test-token",
+        init_usergroups=False,
+    )
+
+    def intercepted_slack_chat_post_message(message: str) -> str:
+        # Refer to the test_slack_notify_skipped_success method for a
+        # description of message_type
+        global message_type
```
hemslo, line 108: Don't use global var in test, a better way to write tests for `slack_notify` is like this:

first clarity what to test, example `test_slack_notify_skipped_success`

* Given `ri` has no errors
* When call `slack_notify` with `test_slack_notify_skipped_success: True`
* Then `slack.chat_post_message` should not be called

Translate that to code is this:

```python
def test_slack_notify_skipped_success():
    api = create_autospec(slack_api.SlackApi)

    slack_notify(
        saas_file_name="test-slack_notify--skipped-success.yaml",
        env_name="test",
        slack=api,
        ri=openshift_resource.ResourceInventory(),
        console_url="https://test.local/console",
        in_progress=False,
        skip_successful_notifications=True,
    )

    api.chat_post_message.assert_not_called()
```

Similar example for `test_slack_notify_unskipped_success`:

```python
def test_slack_notify_unskipped_success():
    api = create_autospec(slack_api.SlackApi)

    slack_notify(
        saas_file_name="test-slack_notify--unskipped-success.yaml",
        env_name="test",
        slack=api,
        ri=openshift_resource.ResourceInventory(),
        console_url="https://test.local/console",
        in_progress=False,
        skip_successful_notifications=False,
    )

    api.chat_post_message.assert_called_once_with(
        ":green_jenkins_circle: SaaS file *test-slack_notify--unskipped-success.yaml* "
        "deployment to environment *test*: Success "
        "(<https://test.local/console|Open>)"
    )
```

Tests should verify the full expected behaviour, use full raw input and output(side effect) to cover all cases.

hemslo (commented):
regarding https://github.com/app-sre/qontract-reconcile/pull/3913#issuecomment-1812825897, instructions are in [README](https://github.com/app-sre/qontract-reconcile#query-classes)

note `Generated by qenerate plugin=pydantic_v1. DO NOT MODIFY MANUALLY!` in the file

reconcile/test/test_openshift_saas_deploy.py
lisa: Thanks for that. It felt really gross to use the `global` but Python is not my first (second, third, etc) language. This really gives me something to work with.

lisa: @hemslo How does it look now?

hemslo: Nice! just need to cleanup commented code https://github.com/app-sre/qontract-reconcile/pull/3913/files/cf13bacf93706ba9107fc19dee89e874a83837c9..644a2d89f2da4a77d22d8d548f801bb78ccdd17f#diff-0f6dbc40d4600c4527755db5370c8c15906daa9c1269fe71e0de1add0243849eR174

reconcile/openshift_saas_deploy.py
hemslo: code has changed, to clarify what I mean:

* https://github.com/app-sre/qontract-reconcile/pull/3913/files#diff-304fd1216ea8201e3242ba749fcc1bebbe71e61fd70851f7ed266e7318ed40b2R136-R138 is not needed
* https://github.com/app-sre/qontract-reconcile/pull/3913/files#diff-304fd1216ea8201e3242ba749fcc1bebbe71e61fd70851f7ed266e7318ed40b2R170 and https://github.com/app-sre/qontract-reconcile/pull/3913/files#diff-304fd1216ea8201e3242ba749fcc1bebbe71e61fd70851f7ed266e7318ed40b2R184 can directly use `saas_file.skip_successful_deploy_notifications`

reconcile/test/test_openshift_saas_deploy.py
lisa: Oh! Well spotted.

OliviaHY:
APPSRE-8300

reconcile/openshift_saas_deploy.py
OliviaHY: But `skip_successful_deploy_notifications ` is added by the schema file right? Have you try dry run the integration locally against a saas file that doesn't have this field set?

In genearl, I encourage adding the manual tests you have done locally to validate your change in the MR description.

chassing (approved):
Great enhancement. Thank you

PR #3914, opened 2023-11-08
[qontract-cli] get network-connections

maorfr:
this command outputs a [mermaid](https://mermaid.js.org/) output with a list of network connections managed via app-interface. this includes everything managed by `terraform-tgw-attachments` and `terraform-vpc-peerings`.

the output can be found here: https://gitlab.cee.redhat.com/service/app-interface-output/-/snippets/7332

why `mermaid`? to possibly be used with a Grafana [diagram](https://grafana.com/grafana/plugins/jdbranham-diagram-panel/) plugin.

a demo dashboard can be found here: https://grafana.dev.devshift.net/d/8QA350y4k/new-dashboard

while this PR is quite raw, it is immediately usable. you can also call it a PoC. as long as this gets thoughts out of you, success.

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -2299,8 +2299,7 @@ def network_connections(ctx):
             continue
         r = item.requester
         source = f"{r.account.name}/{r.region}/{r.tgw_id}"
-        a = item.accepter
-        target = f"{a.account.name}/{a.region}/{a.vpc_id}"
+        target = item.accepter.cluster_name
         print(f"    {source} --> {target}")
 
     # terraform-vpc-peerings
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -2295,6 +2295,8 @@ def network_connections(ctx):
 
     print("graph LR")
     for item in desired_state:
+        if item.deleted:
+            continue
         r = item.requester
         source = f"{r.account.name}/{r.region}/{r.tgw_id}"
         a = item.accepter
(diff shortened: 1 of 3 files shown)


tools/qontract_cli.py
```diff
@@ -2269,6 +2270,75 @@ def slo_document_services(ctx, status_board_instance):
...
+
+    tftgw._validate_tgw_connection_names(desired_state)
+
+    print("graph LR")
+    for item in desired_state:
+        r = item.requester
+        source = f"{r.account.name}/{r.region}/{r.tgw_id}"
+        a = item.accepter
+        target = f"{a.account.name}/{a.region}/{a.vpc_id}"
+        print(f"    {source} --> {target}")
```
hemslo, line 2302: There is a flag field `deleted` in `item`, real integration run will skip it https://github.com/app-sre/qontract-reconcile/blob/393123ac83be16c432faa5c92832c0e5d98f6b2d/reconcile/utils/terrascript_aws_client.py#L1090

Can either skip it here or render as

```mermaid
graph LR
A --> |deleted| B
```

similarly for vpc peerings.

maorfr: thanks! https://github.com/app-sre/qontract-reconcile/pull/3914/commits/dd6e4c7376b0d36077383cbb446ba7d1ee3efe76

hemslo (approved):
LGTM! When we extract function for get desired state, this cli function will be simplified.

PR #3915, opened 2023-11-10
[ocm-ext-config-labels] allowlist for labels

geoberle:
Right now a-i is the source of truth for external configuration labels on clusters. We use them to mark hive shards and backplane shards. But it seems that other parties also put labels on clusters, e.g. cluster service adds "legacy-ingress-support" when a cluster gets upgraded to 4.14. our current integrations are deleting such labels because they are not listed in cluster files.

This PR introduces an allow list for labels in an OCM organizations. clusters of that organization can only manage labels from that list and will ignore all others.

depends on https://github.com/app-sre/qontract-schemas/pull/549

part of https://issues.redhat.com/browse/APPSRE-8521


patjlm (approved):
Thanks!

PR #3921, opened 2023-11-15
add logging of DESIRED resource when debugging

jfchevrette:
When troubleshooting resource apply, it is useful to also have the DESIRED resource as part of debug logging

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -581,6 +581,7 @@ class ApplyOptions:
 
 def should_apply(
     current: OR,
+    desired: OR,
     ri: ResourceInventory,
     options: ApplyOptions,
     cluster: str,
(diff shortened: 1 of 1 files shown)


reconcile/openshift_base.py
```diff
@@ -609,6 +610,7 @@ def should_apply(
         return False
 
     logging.debug("CURRENT: " + OR.serialize(OR.canonicalize(current.body)))
+    logging.debug("DESIRED: " + OR.serialize(OR.canonicalize(desired.body)))
```
geoberle, line 613: nit: i wonder if we should check if debug is enabled on the logger to not seriable and canonicalize when unnecessary

jfchevrette: I have added a condition to check if debug level logging is enabled

PR #3924, opened 2023-11-17
[AUS] support disable integrations on org files

geoberle:
With the latest schema change in org files, we can now disable integrations per OCM organization. See https://github.com/app-sre/qontract-schemas/pull/541

* AUS will execute on every organization that is present as an OCM org file
* ... except if it is disabled for it

This complements the current approach of mentioning all org IDs to act on as command line parameters and will replace it in an upcoming PR.

Additionally the organization name from the org file is used for `org_name` label on exposed metrics. This resolves the issue that many organizations are named `Red Hat` within OCM that makes them hard to distinguish (only based on organization ID).


chassing (approved):
Nice!!! ❤️

PR #3927, opened 2023-11-21
DBAM Fixing settings

janboll:
Need to pass in entire settings dictionary into _process_db_access and split usage there

### DIFF
--- a/reconcile/database_access_manager.py
+++ b/reconcile/database_access_manager.py
@@ -508,11 +508,15 @@ def _process_db_access(
         resource_prefix,
     )
 
+    sql_query_settings = settings.get("sqlQuery")
+    if not sql_query_settings:
+        raise KeyError("sqlQuery settings are required")
+
     managed_resources = _populate_resources(
         db_access,
         engine,
-        settings["imageRepository"],
-        settings["pullSecret"],
+        sql_query_settings["imageRepository"],
+        sql_query_settings["pullSecret"],
         admin_secret_name,
         resource_prefix,
         settings,
@@ -576,10 +580,6 @@ class DatabaseAccessManagerIntegration(QontractReconcileIntegration):
     def run(self, dry_run: bool) -> None:
         settings = queries.get_app_interface_settings()
 
-        sql_query_settings = settings.get("sqlQuery")
-        if not sql_query_settings:
-            raise KeyError("sqlQuery settings are required")
-
         state = init_state(
             integration=QONTRACT_INTEGRATION, secret_reader=self.secret_reader
         )
@@ -607,7 +607,7 @@ class DatabaseAccessManagerIntegration(QontractReconcileIntegration):
                     for db_access in resource.database_access or []:
                         try:
                             with OC_Map(
-                                clusters=namespace.cluster.dict(by_alias=True),
+                                clusters=[namespace.cluster.dict(by_alias=True)],
                                 integration=QONTRACT_INTEGRATION,
                                 settings=settings,
                             ) as oc_map:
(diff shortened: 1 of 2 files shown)


reconcile/database_access_manager.py
```diff
@@ -607,7 +607,7 @@ def run(self, dry_run: bool) -> None:
                     for db_access in resource.database_access or []:
                         try:
                             with OC_Map(
-                                clusters=namespace.cluster.dict(by_alias=True),
+                                clusters=[namespace.cluster.dict(by_alias=True)],
```
hemslo, line 610: If there are multiple namespaces in the same cluster, the same `OC_Map` will be created multiple times, consider extract params needed for `_process_db_access` out, then construct one `OC_Map` object.

janboll: Hm, that's right. I'll move the state check up here , so the OC_Map is only created if really required

janboll: Did the opposite :sweat:

hemslo: To achieve create oc map on needed and only once per cluster, can do this:

Extract all params required to a data class:

```python
class DatabaseAccessModel(BaseModel):
    database_access: DatabaseAccessV1
    namespace: NamespaceV1
    admin_secret_name: str
    engine: str
```

Extract iterate logic into one method:

```python
    @staticmethod
    def _get_db_access_to_process(namespaces) -> Iterator[DatabaseAccessModel]:
        for namespace in namespaces:
            for external_resource in namespace.external_resources or []:
                if not isinstance(
                    external_resource, NamespaceTerraformProviderResourceAWSV1
                ):
                    continue
                for resource in external_resource.resources or []:
                    if not isinstance(resource, NamespaceTerraformResourceRDSV1):
                        continue
                    admin_secret_name = (
                        resource.output_resource_name
                        or f"{resource.identifier}-{resource.provider}"
                    )
                    for database_access in resource.database_access or []:
                        yield DatabaseAccessModel(
                            database_access=database_access,
                            namespace=namespace,
                            admin_secret_name=admin_secret_name,
                            engine=get_db_engine(resource),
                        )
```

Then gather all params to create one oc map for all namespaces required:

```python
        database_access_models = list(self._get_db_access_to_process(namespaces))
        with OC_Map(
            namespaces=[
                db_access.namespace.dict(by_alias=True)
                for db_access in database_access_models
            ],
            integration=QONTRACT_INTEGRATION,
            settings=settings,
        ) as oc_map:
```

Inject it to `_process_db_access` and use `ExceptionGroup` to handle errors:


```python
            errors = []
            for db_access_model in database_access_models:
                try:
                    _process_db_access(
                        dry_run,
                        state,
                        db_access_model.database_access,
                        db_access_model.namespace,
                        db_access_model.admin_secret_name,
                        db_access_model.engine,
                        settings,
                        oc_map,
                    )
                except JobFailedError as e:
                    errors.append(e)
            if errors:
                raise ExceptionGroup("One or more jobs failed to complete", errors)
```
```

janboll: I think this will be slower, since OC_Map is only required for new/changed db access (hence putting it after the state check). This will always create an OC_Map where a db access is configured.

janboll: I like the generator idea for the run method though. Will add it in another PR :+1:

hemslo: it won't be slower than current code, only required clusters related to new/changed db access will be included in the single `oc_map`, you can local debug to verify it.

hemslo: we can merge this fix first then do the `oc_map` related change in the new PR

hemslo (approved):

reconcile/database_access_manager.py
janboll: Think I really have to read up on generators in Python :thinking:

PR #3929, opened 2023-11-23
Pass in Admin connection to Script generator

janboll:
Need to pass in admin permission to Script Generator. Terraform-resources allows it to change the username of the admin connection. Need to pass the admin connection in to be able to create a managed user

### DIFF
--- a/reconcile/database_access_manager.py
+++ b/reconcile/database_access_manager.py
@@ -9,6 +9,7 @@ from typing import (
     Any,
     Callable,
     Optional,
+    TypedDict,
 )
 
 from pydantic import BaseModel
@@ -436,13 +437,18 @@ def _generate_password() -> str:
     return "".join(choices(ascii_letters + digits, k=32))
 
 
+class _DBDonnections(TypedDict):
+    user: DatabaseConnectionParameters
+    admin: DatabaseConnectionParameters
+
+
 def _create_database_connection_parameter(
     db_access: DatabaseAccessV1,
     namespace_name: str,
     oc: OCClient,
     admin_secret_name: str,
     user_secret_name: str,
-) -> dict[str, DatabaseConnectionParameters]:
+) -> _DBDonnections:
     def _decode_secret_value(value: str) -> str:
         return base64.b64decode(value).decode("utf-8")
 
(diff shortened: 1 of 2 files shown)


reconcile/database_access_manager.py
```diff
@@ -439,42 +442,52 @@ def _create_database_connection_parameter(
...
         user_secret_name,
         allow_not_found=True,
     )
+    admin_secret = oc.get(
+        namespace_name,
+        "Secret",
+        admin_secret_name,
+        allow_not_found=False,
+    )
+    print(f"admin_secret: {admin_secret}")
```
hemslo, line 462: debug?

reconcile/database_access_manager.py
```diff
@@ -439,42 +442,52 @@ def _create_database_connection_parameter(
     oc: OCClient,
     admin_secret_name: str,
     user_secret_name: str,
-) -> DatabaseConnectionParameters:
+) -> dict[str, DatabaseConnectionParameters]:
     def _decode_secret_value(value: str) -> str:
         return base64.b64decode(value).decode("utf-8")
 
+    print("creating database connection parameter")
```
rporres, line 449: print? shouldn't this be done via `logging`?

reconcile/database_access_manager.py
```diff
@@ -439,42 +442,52 @@ def _create_database_connection_parameter(
     oc: OCClient,
     admin_secret_name: str,
     user_secret_name: str,
-) -> DatabaseConnectionParameters:
+) -> dict[str, DatabaseConnectionParameters]:
```
hemslo, line 445: Another way is to create a new dataclass with 2 fields, or  typed dict to limit keys.

reconcile/database_access_manager.py
janboll: :broom:

reconcile/database_access_manager.py
janboll: :broom:

reconcile/database_access_manager.py
janboll: Considered it, but since it's internal and just used right below it I omited that thought

hemslo (approved):
LGTM!

reconcile/database_access_manager.py
hemslo: just for type check, still can be used as normal dict https://docs.python.org/3.8/library/typing.html#typing.TypedDict

PR #3930, opened 2023-11-23
Add user deletion to DBAM

janboll:
This enables the deletion of provisioned users.
Content created by these users will not be deleted, but ownership transferred to the admin user. Schemas are retained as well.

### DIFF
--- a/reconcile/database_access_manager.py
+++ b/reconcile/database_access_manager.py
@@ -77,6 +77,9 @@ class PSQLScriptGenerator(BaseModel):
     def _get_user(self) -> str:
         return self.db_access.username
 
+    def _get_admin_user(self) -> str:
+        return self.admin_connection_parameter.user
+
     def _generate_create_user(self) -> str:
         return f"""
 \\set ON_ERROR_STOP on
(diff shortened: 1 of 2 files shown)


reconcile/database_access_manager.py
```diff
@@ -77,6 +77,9 @@ def _get_db(self) -> str:
     def _get_user(self) -> str:
         return self.db_access.username
 
+    def _get_admin_user(self) -> str:
+        return self.admin_connection_parameter.user
```
hemslo, line 81: readonly property is more common:

```python
    @property
    def admin_user(self) -> str:
        return self.admin_connection_parameter.user
```

reconcile/database_access_manager.py
```diff
@@ -105,8 +122,13 @@ def _generate_db_access(self) -> str:
...
     def generate_script(self) -> str:
-        x = self._generate_create_user() + "\n" + self._generate_db_access()
-        return x
+        if not self.db_access.delete:
+            script = self._generate_create_user() + "\n" + self._generate_db_access()
+        else:
+            script = (
+                self._generate_revoke_db_access() + "\n" + self._generate_delete_user()
+            )
+        return script
```
hemslo, line 131: Now this method has 2 responsibilities `create` and `delete`, maybe make it explicit by extracting a provision script and a deprovision script like:

```python
    def provision_script(self) -> str:
        return self._generate_create_user() + "\n" + self._generate_db_access()

    def deprovision_script(self) -> str:
        return self._generate_revoke_db_access() + "\n" + self._generate_delete_user()

    def generate_script(self) -> str:
        return (
            self.deprovision_script()
            if self.db_access.delete
            else self.provision_script()
        )
```

reconcile/database_access_manager.py
janboll: Hm, these are not properties of `PSQLScriptGenerator`, but helpers to make accessing related data easier.

reconcile/database_access_manager.py
```diff
@@ -94,9 +97,23 @@ def _generate_create_user(self) -> str:
...
+\\c "{self._get_db()}"
+REASSIGN OWNED BY "{self._get_user()}" TO "{self._get_admin_user()}";
+DROP ROLE IF EXISTS "{self._get_user()}";\\gexec"""
+
+    def _generate_revoke_db_access(self) -> str:
+        statements: list[str] = ["\n"]
+        for access in self.db_access.access or []:
+            statement = f'REVOKE ALL PRIVILEGES ON ALL TABLES IN SCHEMA "{access.target.dbschema}" FROM "{self._get_user()}";\n'
+            statements.append(statement)
+        return "".join(statements)
```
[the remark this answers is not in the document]
janboll: :+1:

reconcile/database_access_manager.py
hemslo: It's rare in python to define getter functions, https://stackoverflow.com/questions/2627002/whats-the-pythonic-way-to-use-getters-and-setters.

Such internal helpers can be like this

```python
@property
def _admin_user(self) -> str:
    return self.admin_connection_parameter.user
```

reconcile/database_access_manager.py
janboll: :+1:

PR #3933, opened 2023-11-24
[ocm-clusters] make controlplane role optional

geoberle:
HCP clusters don't have a controlplane account role

part of [APPSRE-8561
](https://issues.redhat.com/browse/APPSRE-8561)

depends on https://github.com/app-sre/qontract-schemas/pull/552/files


hemslo (approved):
LGTM!

It would be good to update test files too https://github.com/app-sre/qontract-reconcile/blob/e7d92c397b4189308dc84d66639d15685f7944a0/reconcile/test/test_ocm_clusters.py#L262

PR #3937, opened 2023-11-28
[aws-version-sync] show not auto-merge MRs in AppSRE review queue

chassing:
... otherwise, we don't see such MRs because the `engine_version` field is `self-serviceable`.


hemslo (approved):
LGTM!

PR #3940, opened 2023-11-29
[aws-version-sync] no RDS downgrades

chassing:
Convert the `engine_version` strings to `semver.VersionInfo` objects to have a proper version comparison to avoid RDS downgrade MRs. 

Additionally, enhance our custom `parse_semver` function with `optional_minor_and_patch` to mimik semver3 [Version.parse](https://python-semver.readthedocs.io/en/latest/api.html#semver.version.Version.parse)

### DIFF
--- a/reconcile/aws_version_sync/integration.py
+++ b/reconcile/aws_version_sync/integration.py
@@ -87,7 +87,7 @@ class ExternalResource(BaseModel):
             self.resource_engine,
         )
 
-    @validator("resource_engine_version")
+    @validator("resource_engine_version", pre=True)
     def parse_resource_engine_version(  # pylint: disable=no-self-argument
         cls, v: str | semver.VersionInfo
     ) -> semver.VersionInfo:
--- a/reconcile/aws_version_sync/integration.py
+++ b/reconcile/aws_version_sync/integration.py
@@ -250,6 +250,12 @@ class AVSIntegration(QontractReconcileIntegration[AVSIntegrationParams]):
         for diff_pair in diff.change.values():
             aws_resource = diff_pair.desired
             app_interface_resource = diff_pair.current
+            if (
+                aws_resource.resource_engine_version
+                <= app_interface_resource.resource_engine_version
+            ):
+                # do not downgrade the version
+                continue
             # make mypy happy
             assert app_interface_resource.namespace_file
             assert app_interface_resource.provisioner.path
--- a/reconcile/aws_version_sync/integration.py
+++ b/reconcile/aws_version_sync/integration.py
@@ -5,7 +5,11 @@ from collections.abc import (
 )
 from typing import Any
 
-from pydantic import BaseModel
+import semver
+from pydantic import (
+    BaseModel,
+    validator,
+)
 
 from reconcile.aws_version_sync.merge_request_manager.merge_request import (
     Parser,
@@ -42,6 +46,7 @@ from reconcile.utils.runtime.integration import (
     PydanticRunParams,
     QontractReconcileIntegration,
 )
+from reconcile.utils.semver_helper import parse_semver
 from reconcile.utils.unleash import get_feature_toggle_state
 from reconcile.utils.vcs import VCS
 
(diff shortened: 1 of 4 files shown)


reconcile/aws_version_sync/integration.py
```diff
@@ -67,7 +72,10 @@ class ExternalResource(BaseModel):
     resource_provider: str
     resource_identifier: str
     resource_engine: str
-    resource_engine_version: str
+    resource_engine_version: semver.VersionInfo | str
```
hemslo, line 75: We expect `resource_engine_version` to always be converted to `VersionInfo`, right?

can directly define

```python
    resource_engine_version: semver.VersionInfo
```

just need to use pre validator to convert it

```python
    @validator("resource_engine_version", pre=True)
```

reconcile/aws_version_sync/integration.py
```diff
@@ -228,8 +244,8 @@ def reconcile(
             current=external_resources_app_interface,
             desired=external_resources_aws,
             key=lambda r: r.key,
-            equal=lambda r1, r2: r1.resource_engine_version
-            == r2.resource_engine_version,
+            equal=lambda external_resources_app_interface, external_resources_aws: external_resources_app_interface.resource_engine_version
+            >= external_resources_aws.resource_engine_version,
```
hemslo, line 248: I didn't notice the `>=` check at the first glance, later realized it is a smart check here to ignore version bump in app interface. I think it's more explicit if do the check while iterating `diff_pair`:

```python
            if aws_resource.resource_engine_version > app_interface_resource.resource_engine_version:
```

so we make it clear that this integration only create mr to upgrade engine version.

reconcile/utils/semver_helper.py
```diff
@@ -7,7 +7,16 @@ def make_semver(major: int, minor: int, patch: int) -> str:
     return str(semver.VersionInfo(major=major, minor=minor, patch=patch))
 
 
-def parse_semver(version: str) -> semver.VersionInfo:
+def parse_semver(
+    version: str, optional_minor_and_patch: bool = False
+) -> semver.VersionInfo:
+    if optional_minor_and_patch:
+        # semver3 supports optional minor and patch.
+        # until we upgrade to semver3, we support this by adding a default minor and/or patch
```
hemslo, line 15: Any blocker to bump version? https://python-semver.readthedocs.io/en/latest/migration/migratetosemver3.html looks ok.

chassing:
```
$ pip install -U semver
unleashclient 5.7.1 requires semver<3.0.0, but you have semver 3.0.2 which is incompatible.
sretoolbox 2.4.2 requires semver~=2.13, but you have semver 3.0.2 which is incompatible.
```
Let's do the semver upgrade later :)

reconcile/aws_version_sync/integration.py
chassing: Yes, explicit is better than implicit. Applied in [42a62bd](https://github.com/app-sre/qontract-reconcile/pull/3940/commits/42a62bd6e0f86569e454dfc8f7b7775f8ae921b5)

reconcile/aws_version_sync/integration.py
chassing: Strange, I had some mypy error w/o `str`. Nevertheless, now it's working [3c87702](https://github.com/app-sre/qontract-reconcile/pull/3940/commits/3c87702e2e8474947e132913363595cec090ba8c)

reconcile/aws_version_sync/integration.py
hemslo: Do you want to revert back to equal https://github.com/app-sre/qontract-reconcile/pull/3940/files#diff-cbc1c6cbfb9a7d7a7394fa5e47bcfdcbf95f17e250cfe4aa4bfdffacd35cfd5aR247-R248?

reconcile/aws_version_sync/integration.py
chassing: 🤦 fixed in f370bcb

reconcile/aws_version_sync/integration.py
chassing: Applied in c1bb8bf

hemslo (approved):
LGTM!

PR #3942, opened 2023-11-30
Revert "APPSRE-8300: Optionally silence successful deploy notices"

chassing:
Reverts app-sre/qontract-reconcile#3913


chassing:
FYI @lisa https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/sop/app-interface-integrations-promotions.md

PR #3945, opened 2023-12-04
[APPSRE-8396] allow s3 to send lambda_notifications

BumbleFeng:


OliviaHY (commented):
Can you please add some context in the description to explain why this is related to APPSRE-8396?

hemslo:
Another common flow is `s3 -> sqs -> lambda` for better error handling. But for this use case, it seems you want to apply same ACL on every objects in the bucket, that's not what ACL designed for, if you need different access for different objects in one bucket, use ACL, otherwise use policy.

* [Controlling ownership of objects and disabling ACLs for your bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html)
* [Example 2: Bucket owner granting cross-account bucket permissions](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-walkthroughs-managing-access-example2.html)

BumbleFeng:
> Another common flow is `s3 -> sqs -> lambda` for better error handling. But for this use case, it seems you want to apply same ACL on every objects in the bucket, that's not what ACL designed for, if you need different access for different objects in one bucket, use ACL, otherwise use policy.

The problem here is that ACL is required by cloud-front log delivery system. https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html#access-logs-choosing-s3-bucket.

hemslo:
Can you try the last example in [Example walkthroughs: Managing access to your Amazon S3 resources](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-walkthroughs-managing-access.html), [Example 4: Bucket owner granting cross-account permission to objects it does not own](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-walkthroughs-managing-access-example4.html).

For cloudfront logs, grant full control to the bucket owner is already done, should be less steps.

BumbleFeng:
> Can you try the last example in [Example walkthroughs: Managing access to your Amazon S3 resources](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-walkthroughs-managing-access.html), [Example 4: Bucket owner granting cross-account permission to objects it does not own](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-walkthroughs-managing-access-example4.html).
> 
> For cloudfront logs, grant full control to the bucket owner is already done, should be less steps.

That example required more than granting full control to the bucket owner. It needs Bucket owner enforced(aka: ACLs disabled) which allows the bucket owner to override the object owner to itself and apply bucket policy on it. Check the differences here https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html. Meanwhile, for cloudfront logs `Don't choose an Amazon S3 bucket with [S3 Object Ownership](https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html) set to bucket owner enforced. That setting disables ACLs for the bucket and the objects in it, which prevents CloudFront from delivering log files to the bucket.` In short, only one of ACL or bucket policy can take effect, and aws has chosen to use ACL to deliver cloudfront logs.

BumbleFeng:
> Can you try the last example in [Example walkthroughs: Managing access to your Amazon S3 resources](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-walkthroughs-managing-access.html), [Example 4: Bucket owner granting cross-account permission to objects it does not own](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-walkthroughs-managing-access-example4.html).
> 
> For cloudfront logs, grant full control to the bucket owner is already done, should be less steps.

They need to use Athena to run cross-account queries, and it does not support assuming role https://docs.aws.amazon.com/athena/latest/ug/security-iam-cross-account-glue-catalog-access.html

PR #3948, opened 2023-12-07
[ocm-machine-pools] Update deletable check for OSD and ROSA HCP

hemslo:
OSD NON CCS clusters can't delete default `worker` machine pool, ROSA HCP clusters can delete default `workers*` node pools.

This change updates deletable check to fail for OSD delete default machine pool, pass for ROSA HCP delete default node pools.

[APPSRE-8556](https://issues.redhat.com/browse/APPSRE-8556)

### DIFF
--- a/reconcile/ocm_machine_pools.py
+++ b/reconcile/ocm_machine_pools.py
@@ -11,6 +11,7 @@ from typing import (
 
 from pydantic import (
     BaseModel,
+    Field,
     root_validator,
 )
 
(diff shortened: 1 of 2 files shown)


reconcile/ocm_machine_pools.py
```diff
@@ -330,12 +357,24 @@ def fetch_current_state(
     }
 
 
-def _is_hypershift(cluster: ClusterV1) -> bool:
-    return bool(cluster.spec and cluster.spec.hypershift)
+def _classify_cluster_type(cluster: ClusterV1) -> ClusterType:
+    assert cluster.spec  # cluster.spec is required
```
janboll, line 361: Ist that assert called? Wouldn't it be nicer to throw a ValueError with some description?

hemslo: right, somehow this field is not required in schema, I'll do a explicit check here

PR #3954, opened 2023-12-13
import OCM cluster OIDC endpoint URL into a-i cluster files

geoberle:
Import the OIDC endpoint URL of ROSA clusters into app-interface cluster files, similar to the ID and external ID of a cluster. It is supposed to be read only information that can be used by people and automation to define trust policies in IAM policies without the need to do OCM lookups. Since this URL is static, "caching" it like this in the cluster file is a legit approach.

part of https://issues.redhat.com/browse/APPSRE-8656

depends on https://github.com/app-sre/qontract-schemas/pull/560

### DIFF
--- a/reconcile/gql_definitions/introspection.json
+++ b/reconcile/gql_definitions/introspection.json
@@ -16566,6 +16566,18 @@
                             "deprecationReason": null
                         },
                         {
+                            "name": "issueSecurityId",
+                            "description": null,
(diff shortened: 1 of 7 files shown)


chassing (approved):

patjlm:
I was wondering recently if we should do the same for egress IPs as well ...

PR #3959, opened 2023-12-15
minimize pylint ignore rules

chassing:
Remove/Move pylint ignore rules:
- remove PLC0415: import should be at the top-level of a file
- remove PLC1901: ... can be simplified to `...` as an empty string is falsey
- remove PLR1704: Redefining argument with the local name
- remove PLR5501: Use `elif` instead of `else` then `if`, to reduce indentation
- remove PLR6201: Use a `set` literal when testing for membership
- remove PLW0602: Using global for `...` but no assignment is done
- move PLW0603 (usage of global) to occurences instead of having it global
- remove PLW1510: `subprocess.run` without explicit `check` argument
- remove PLW1514: `open` in text mode without explicit `encoding` argument
- move PLW3201 (Bad or misspelled dunder method name) to occurrence instead of having it globally

Please review it commit by commit.

Ticket: [APPSRE-8624](https://issues.redhat.com/browse/APPSRE-8624)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,6 @@ ignore = [
     "PLR0915", # Too many statements
     "PLR0917", # Too many positional arguments
     "PLR2004", # Magic value used in comparison
-    "PLR5501", # Use `elif` instead of `else` then `if`, to reduce indentation
     "PLR6201", # Use a `set` literal when testing for membership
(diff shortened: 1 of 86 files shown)


fishi0x01 (approved):

rporres:
I couldn't review this on time but this is great stuff. Thx <3

PR #3960, opened 2023-12-15
Change admission implementation

janboll:
This PR introduces change admission, by which change types can be declared as restrictive. Any changes to files covered by restrictive change types need to be done by approvers only. Changes could be allowed by adding a `/good-to-test` command though.
 
[APPSRE-8625](https://issues.redhat.com/browse/APPSRE-8625)


reconcile/change_owners/change_owners.py
```diff
@@ -231,6 +231,20 @@ def init_gitlab(gitlab_project_id: str) -> GitLabApi:
...
+        for dc in change.diff_coverage:
+            for c in dc.coverage:
+                if c.change_type_processor.restrictive:
+                    approvers = {a.org_username for a in c.approvers}
+                    if user not in approvers:
+                        logging.error(
+                            f"change type {c.change_type_processor.name} is restrictive"
+                            f"user  {user} is not an approver"
+                        )
+                        sys.exit(1)
```
patjlm, line 245: should we allow anyone to submit, but need a `/good-to-test` handler of some sort?

janboll: Yeah, I'd add that in a new iteration

reconcile/change_owners/change_owners.py
```diff
@@ -231,6 +232,31 @@ def init_gitlab(gitlab_project_id: str) -> GitLabApi:
...
+                                f"Restrictive change approved by {gttapprover}"
+                            )
+                            return
+
+                    if user not in approvers:
+                        logging.error(
+                            f"change type {c.change_type_processor.name} is restrictive"
+                            f"user  {user} is not an approver"
+                        )
+                        sys.exit(1)
```
geoberle, line 257: i think it would be better to signal failed assertion back to the caller and handle the exit there.
do we want to leave a message on the MR to help the author to figure out what is going on or is the output in the PR check logs enough?

at least we would need to mention who are the approvers and who would need to give their `/good-to-test`. can we add that somehow to the existing MR comment with all the approvers listed?

geoberle: also, for MRs where AppSRE is the approriate approver i would love that the MR to shows up in the review queue

reconcile/gql_definitions/common/clusters.py
```diff
@@ -645,7 +645,7 @@ class ClusterV1(ConfiguredBaseModel):
     name: str = Field(..., alias="name")
     server_url: str = Field(..., alias="serverUrl")
     console_url: str = Field(..., alias="consoleUrl")
-    kibana_url: str = Field(..., alias="kibanaUrl")
+    kibana_url: Optional[str] = Field(..., alias="kibanaUrl")
```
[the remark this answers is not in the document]
janboll: Might have had bad timing with generating the gql definitions. Let me rebase and do it again

reconcile/change_owners/change_owners.py
janboll: Great feedback. I'll try to make it easier to digest for the user

janboll:
Good to test label implemented/tested, reference: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/92588

janboll:
Review queue will be updated in another PR

reconcile/change_owners/change_owners.py
```diff
@@ -231,6 +242,31 @@ def init_gitlab(gitlab_project_id: str) -> GitLabApi:
...
+                                f"Restrictive change approved by {gttapprover}"
+                            )
+                            return True
+
+                    if user not in approvers:
+                        logging.error(
+                            f"change type {c.change_type_processor.name} is restrictive"
+                            f"user {user} is not an approver"
+                        )
+                        return False
```
hemslo, line 266: `user` here is the author, why check whether a change is admitted requires checking author? I thought the logic is just all restrictive changes have good to test approved. Like this:

```python
    return all(
        {a.org_username for a in c.approvers} & good_to_test_approvers
        for change in changes
        for dc in change.diff_coverage
        for c in dc.coverage
        if c.change_type_processor.restrictive
    )

```

janboll: No it's more complex: Users with the restrictive role assigned are allowed to do these restrictive changes and add the looks good to test comment.

janboll: A change is either admitted by schema or by a comment of an authorized person. I think the name is fine.

rporres: What do you mean by "admitted by schema"? I don't really see where in this function only the schema check takes place (most likely is my newbiness to this integration).

I share a bit the feeling of not understanding well this function. The relationship between the `user`, the `approvers` and the `good_to_test_approvers` is complex.

reconcile/change_owners/change_owners.py
```diff
@@ -390,5 +439,11 @@ def run(
                 # skips MRs with this label
                 gl.remove_label(merge_request, SELF_SERVICEABLE)
 
+            if not change_admitted:
+                raise NotAdmittedError("Change not admitted")
+
+    except NotAdmittedError as e:
+        logging.error(e)
+        sys.exit(1)
```
rporres, line 447: We want to fail here because we assume this is the first integration that is going to run and we don't want to move on with the rest of the integrations, correct?

geoberle: exactly

reconcile/change_owners/change_owners.py
hemslo: I may understand what you mean. Unlike `lgtm` flow, even the author is part of approvers, still need a `lgtm` to mark it as approved. But for this admission check, if author is part of approvers, then it's admitted implicitly, right?

Btw this code is not covering all cases due to use of early `return`.

If the change has 2 different coverages, each one has a different list of approvers, one is `[A]`, the other one is `[B]`, only `A` commented good to test, then this change should be considered not admitted, require `B` to comment good to test.

To describe the condition:

* changes are admitted when all restrictive coverages are admitted
* a coverage is admitted when the author is part of approvers OR some approvers commented good to test

Translate to code:

```python

def is_coverage_admitted(coverage, user, good_to_test_approvers):
    return any(
        a.org_username == user or a.org_username in good_to_test_approvers
        for a in coverage.approvers
    )


def is_change_admitted(
    changes: list[BundleFileChange], user: str, good_to_test_approvers: set[str]
) -> bool:
    return all(
        is_coverage_admitted(c, user, good_to_test_approvers)
        for change in changes
        for dc in change.diff_coverage
        for c in dc.coverage
        if c.change_type_processor.restrictive
    )

```

Maybe should rename `user` to `author` to make it explicit.

reconcile/change_owners/change_owners.py
rporres: Nice, thanks. A comment explaining this would be very helpful, especially because it express an expectation on how integrations are being run via hack scripts.

janboll: added

reconcile/change_owners/change_owners.py
janboll: updated

hemslo (approved):
LGTM!

rporres (approved):
nice!

PR #3961, opened 2023-12-17
[openshift-serviceaccount-tokens] ignore deleted namespace

maorfr:
follow up on #2942 and #3010

openshift-serviceaccount-tokens currently does not handle deleted namespaces. with this PR, we add the `delete` field to the query, allowing the proper handle of this scenario.


chassing (approved):

patjlm:
:heart:

PR #3967, opened 2023-12-18
APPSRE-8160: Handle vault keys with whitespace

lisa:
APPSRE-8160: Handle vault keys with whitespace issues.


esron:
`make format` can get rid of most formatting errors.

reconcile/utils/vault.py
```diff
@@ -189,6 +193,13 @@ def read_all_with_version(self, secret: Mapping) -> tuple[Mapping, Optional[str]
         if data is None:
             raise SecretNotFound
 
+        # Check the keys to ensure that they do not have any whitespace around them.
+        for k in data:
+            if k.strip() != k:
+                raise SecretFormatProblem(
```
bhushanthakur93, line 199: We should add this exception in `retry` decorator through `no_retry_exceptions` to avoid unnecessary extra calls.

lisa:
> `make format` can get rid of most formatting errors.

make one simple change and the linter goes nuts.

bhushanthakur93 (changes requested):

reconcile/utils/vault.py
lisa: Done.

TGPSKI:
LGTM

+ suggest `make format` as prereq for git push (saves time)

After re-tracing the read_all -> read_all_with_version -> _read_all_v1 / read_all_v2, I forgot how complex this read chain is. I'm confident you covered every case, however.

reconcile/test/test_vault_utils.py
```diff
@@ -45,3 +45,65 @@ def test_sleep_is_called(sleep):
...
             client._auto_refresh_client_auth()
+
+
+def test_key_has_leading_space():
+    # The natural thing to do to gain access to the SecretFormatProblem custom
+    # Exception would be to rehome its import to the top of this test file.
+    # However, the TestVaultUtils class will reimport the vault library and
+    # "undo" the import. Thus, the import will be done in each test method that
+    # uses the exception.
+    from reconcile.utils.vault import SecretFormatProblem
```
chassing, line 56: Please do this import at the top of the file

reconcile/test/test_vault_utils.py
```diff
@@ -45,3 +45,65 @@ def test_sleep_is_called(sleep):
...
+
+
+def test_key_has_trailing_space():
+    from reconcile.utils.vault import SecretFormatProblem
+
+    with pytest.raises(
+        SecretFormatProblem,
+        match="Secret key has whitespace. Expected 'trailing_space' but got 'trailing_space '",
+    ):
+        vc = testVaultClient()
```
chassing, line 78: Do you know https://docs.pytest.org/en/7.3.x/how-to/parametrize.html#pytest-mark-parametrize? with `pytest-mark-parametrize` you could reuse your test function.

chassing (approved):
Just a few nits. LGTM

reconcile/utils/vault.py
```diff
@@ -27,6 +27,10 @@ class SecretNotFound(Exception):
     pass
 
 
+class SecretFormatProblem(Exception):
```
lechuk47, line 30:
```suggestion
class SecretKeyFormatError(Exception):
```
Just a NIT. I think it's more clear

lisa:
/hold

will relocate this to a better (perhaps) spot

reconcile/utils/vault.py
lisa: 👍 will adopt this as part of the code relocation.

reconcile/test/test_vault_utils.py
lisa: For what purpose? It will have no effect for the reason stated in the comment directly preceeding line 56.

reconcile/openshift_resources_base.py
```diff
@@ -600,6 +606,15 @@ def fetch_provider_vault_secret(
         raise FetchResourceError(str(e))
 
 
+# check to ensure that all of the keys are valid by looking to see if there are any white space issues. If any issues are uncovered, an exception will be raised.
+def assert_valid_secret_keys(items):
```
chassing, line 610: Please add type hints.

chassing (commented):
Why did you move the validation into `openshift_resources_base` instead of having it in 'utils.vault'?

reconcile/openshift_resources_base.py
bhushanthakur93, line 610: A better check would be to abide by the spec. https://kubernetes.io/docs/concepts/configuration/secret/#restriction-names-data 

> The keys of data and stringData must consist of alphanumeric characters, -, _ or .

I also suggest updating the error message so it's more user friendly. The error message should point out vault path and key, in addition to the spec requirements. Often times I see our tenants struggling with the lack of good error messaging, so small things like this definitely adds up.

reconcile/test/test_openshift_resources_base.py
```diff
@@ -482,3 +482,21 @@ def test_cluster_params():
...
+    "test_keyname, exception_expected",
+    [
+        (" leading_space", orb.SecretKeyFormatError),
+        (" space_padding ", orb.SecretKeyFormatError),
+        ("trailing_space ", orb.SecretKeyFormatError),
+        ("no space issues", None),  # but should there be...
+        ("no_spacing", None),
+    ],
+)
+def test_secret_keys(test_keyname, exception_expected):
```
bhushanthakur93, line 497: I suggest to split up the tests so that they can verify one behavior at a time. In this case, I suggest you create a separate test that expects the error and another one without error.

reconcile/openshift_resources_base.py
lisa: that's a good point @bhushanthakur93

reconcile/test/test_openshift_resources_base.py
lisa: I've decided not to do that for two reasons. First, I want to avoid future comments similar to https://github.com/app-sre/qontract-reconcile/pull/3967#discussion_r1431064114. Second, I think the test table approach here is a best fit because the tests all follow the same structure and flow.

bhushanthakur93: I highly suggest reading https://testing.googleblog.com/2014/04/testing-on-toilet-test-behaviors-not.html.

You can still utilize parametrize tests for different inputs, just isolate the two different behavior of the same function in different tests :) 

That said, I leave this up to you if you still want to keep the same test.

bhushanthakur93 (approved):

reconcile/openshift_resources_base.py
```diff
@@ -600,6 +608,19 @@ def fetch_provider_vault_secret(
...
+# check to ensure that all of the keys are valid by looking to see if there are
+# any white space issues. If any issues are uncovered, an exception will be
+# raised.
+# we're receiving the full key: value information, not simply a list of keys.
+def assert_valid_secret_keys(secrets_data: dict[str, str]):
+    for k in secrets_data:
+        matches = re.search(KUBERNETES_SECRET_DATA_KEY_RE, k)
+        if not matches:
+            raise SecretKeyFormatError(
+                f"Secret key has whitespace. Expected '{k.strip()}' but got '{k}'"
```
chassing, line 620: Please adapt the error message too.

chassing (approved):
Just a nit, otherwise LGTM. Thanks for your work

reconcile/openshift_resources_base.py
lisa: Whoops. Copy/paste error. ;)

lisa: @chassing Have a look now.

chassing: Perfect. Thank you

PR #3969, opened 2023-12-19
[tf-vpc-peerings] allow access to private HCP API VPCE via peering

geoberle:
the HCP API is offered via a VPC endpoint in the clusters AWS account. access to this VPCE is per default only granted from within the VPC. when accessing the API via VPC peering, the security group of the VPCE is preventing communication.

this patch will add new rules to the security group of the VPC endpoint. sadly we can't create a dedicated SG and add it to the VPCE because that is only possible with the AWS TF provider > 4 (and we are at 3.7.x). so our only way is to add more rules to the existing SG.

i checked the code of the hypershift operator that reconciling this SG. it is not removing any excessive rules, only adding missing ones. once we upgrade to a newer TF provider, we can add new security groups to existing VPC endpoints. but for now we need to accept this approach.

for the our first HCP private cluster, the output of the integration would be

```
[2023-12-19 02:42:55] [INFO] [DRY-RUN] [terraform_client.py:log_plan_diff:335] - ['create', 'app-sre', 'aws_security_group_rule', 'api-access-from-peering-backplanei03ue1_app-sre-ci-int', set()]
[2023-12-19 02:42:55] [INFO] [DRY-RUN] [terraform_client.py:log_plan_diff:335] - ['create', 'app-sre', 'aws_security_group_rule', 'api-access-from-peering-backplanei03ue1_appsrep05ue1', set()]
[2023-12-19 02:42:55] [INFO] [DRY-RUN] [terraform_client.py:log_plan_diff:335] - ['create', 'app-sre', 'aws_security_group_rule', 'api-access-from-peering-backplanei03ue1_appsres05ue1', set()]
```

part of https://issues.redhat.com/browse/APPSRE-8666

### DIFF
--- a/reconcile/terraform_vpc_peerings.py
+++ b/reconcile/terraform_vpc_peerings.py
@@ -199,8 +199,15 @@ def build_desired_state_single_cluster(
                 "are not suppored"
             )
 
-        requester_vpc_id, requester_route_table_ids, _ = awsapi.get_cluster_vpc_details(
-            req_aws, route_tables=requester_manage_routes
+        (
+            requester_vpc_id,
+            requester_route_table_ids,
+            _,
+            api_security_group_id,
+        ) = awsapi.get_cluster_vpc_details(
+            req_aws,
+            route_tables=requester_manage_routes,
+            hcp_vpc_endpoint_sg=_private_hosted_control_plane(cluster_info),
         )
         if requester_vpc_id is None:
             raise BadTerraformPeeringState(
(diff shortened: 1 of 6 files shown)


reconcile/terraform_vpc_peerings.py
```diff
@@ -456,6 +488,12 @@ def build_desired_state_vpc_single_cluster(
     return desired_state
 
 
+def _private_hosted_control_plane(cluster_info: dict[str, Any]) -> bool:
+    return cluster_info["spec"].get("hypershift", False) and cluster_info["spec"].get(
```
patjlm, line 492: I think `, False` are useless, right?

patjlm (approved):

reconcile/terraform_vpc_peerings.py
geoberle: fixed. thanks

maorfr (approved):
LGTM!

PR #3976, opened 2023-12-22
[AUS] expect at least one version gate for a minor version

geoberle:
change: if a minor version has no version gates, upgrades to that minor versions are skipped.

reasoning: usually every minor version has at least one version gate in place. this is not technically necessary but practically speaking this has been the case for every minor version of the last couple of years. we want to skip scheduling an upgrade in such situations because missing version gates might be an indicator that the gap analysis has not been conducted yet from the platform team. such upgrades bear many unknowns and are not considered safe. this mostly happens early during pre-GA, e.g. for RCs.

https://issues.redhat.com/browse/APPSRE-8694

### DIFF
--- a/reconcile/aus/base.py
+++ b/reconcile/aus/base.py
@@ -1025,7 +1025,7 @@ def calculate_diff(
                             GateAgreement(gate=g)
                             for g in gates_to_agree(
                                 minor_version_gates,
-                                spec.cluster,
+                                spec.cluster.id,
                                 ocm_api,
                             )
                         ],
--- a/reconcile/aus/base.py
+++ b/reconcile/aus/base.py
@@ -735,12 +735,12 @@ def gates_to_agree(
     """Check via OCM if a version is agreed
 
     Args:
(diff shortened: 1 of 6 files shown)


reconcile/aus/base.py
```diff
@@ -1002,16 +1001,31 @@ def set_mutex(
...
+                    gates, target_version_prefix
+                )
+                # skipping upgrades when there are no version gates is a safety
+                # precaution to prevent cluster upgrades being scheduled.
+                # missing version gates are an indicator that the version has not yet gone
+                # through SREP gap analysis and is not yet ready for upgrades.
+                #
+                # this might change in the future - revisite for 4.16
+                if not minor_version_gates:
+                    logging.warning(
```
patjlm, line 1015: I wonder if warning is the best log level here. Will it continuously write in the reconcile channel?

patjlm (approved):

reconcile/aus/base.py
geoberle: you are right, it will continously write there. lets go for debug for the time being.

PR #3977, opened 2023-12-28
[qontract-cli] add get rosa-create-cluster-command

maorfr:
this command will allow adding a cluster file into app-interface and generating a `rosa create cluster` command.

this is a handy shortcut that will remove some manual error prone operations in provisioning rosa clusters.


geoberle (approved):

chassing:
❤️

PR #3979, opened 2023-12-31
[qontract-cli] create new cmd and env commands

maorfr:
the `cmd` command gathers commands that output a cli command to run:
- qontract-cli cmd oc login
- qontract-cli cmd ocm login
- qontract-cli cmd rosa create cluster

the `env` command gathers commands that output environment variables to set:
- qontract-cli env aws

we can continue to iterate if we like this structure :pray:


patjlm (approved):
I like it

PR #3980, opened 2024-01-02
[glitchtip-project-dsn] reduce project listing API calls

chassing:
Don't retrieve an organization project list, just rely on the key existence. This reduces the Glitchtip API calls by the number of projects.

Ticket: [APPSRE-8675](https://issues.redhat.com/browse/APPSRE-8675)


hemslo (approved):
LGTM!

Another improvement can be reuse session across threads, so connections fd can be reused. As this error is happening for other glitchtip integrations too. Described in [APPSRE-8179](https://issues.redhat.com/browse/APPSRE-8179).

PR #3981, opened 2024-01-03
trigger sast scan for image build

BumbleFeng:
This is a missing part of [APPSRE-8072](https://issues.redhat.com/browse/APPSRE-8072). Deploy services built by RHTAP would be triggered by image build, We should also allow triggering sast scan in this case. I revisit this feature as I would like to use the same logic for triggering malware detection scan.


patjlm:
Can you add a short description and ticket number for this change please?

PR #3983, opened 2024-01-04
Review queue, escape pipes in titles

janboll:
Pipes in content can break the review queue, see: https://gitlab.cee.redhat.com/service/app-interface-output/-/blob/249d7417ff9e9552589ba65614f35aab26b810bb/app-interface-review-queue.md


fishi0x01 (approved):
Nice :+1:

PR #3985, opened 2024-01-04
APPSRE-8280: added code to fix multiple sns subscription for terraform

mehfuz:
This consists of code to fix the issue - [APPSRE-8280](https://issues.redhat.com/browse/APPSRE-8280)

Problem - The parsing logic created the same identifier for `"aws_sns_topic_subscription`. Hence  the terrascipt only used to render the resource considering the last subscription added

Now with the new code every subscription for an sns topic will have a unique  identifier. As a result all the subscription for a particular topic will be created.

Local Testing
For the following configuration
image
The terrascript created the following plan
 
image

With single subscription the output is like
image

the output we get is as follows
Screenshot 2024-01-08 at 2 38 40 PM

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -2620,7 +2620,7 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
 
         if "subscriptions" in common_values.keys():
             subscriptions = common_values.get("subscriptions")
-            for index,sub in enumerate(subscriptions):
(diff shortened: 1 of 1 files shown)


janboll:
Did you run the integration locally?

hemslo (approved):
LGTM!

mehfuz:
> Did you run the integration locally? Does it create the correct resources in AWS?



> Did you run the integration locally? Does it create the correct resources in AWS?

@janboll  the aws resources are created as follows
image

janboll (approved):

PR #3988, opened 2024-01-08
[AUS] add org name to log output

geoberle:
add the org name of a cluster to the log output because the org ID is not providing immediate help to identify an org without a human looking it up.


patjlm (approved):
Thanks!

PR #3993, opened 2024-01-10
[openshift-resources] Fix for html_url in Prometheus alerts

cubismod:
A second attempt at #3990 

Some colleagues pointed out that the earlier attempt resulted in a doubling of qontract-server requests which means more $$$. This take removes that additional GQL call as the function already has a settings dict passed to it which includes the `repoUrl` value I was previously querying for in GQL.

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -6,32 +6,55 @@ import logging
 import re
 import sys
 from collections import defaultdict
-from collections.abc import Iterable, Mapping, Sequence
+from collections.abc import (
+    Iterable,
+    Mapping,
+    Sequence,
+)
 from contextlib import contextmanager
 from dataclasses import dataclass
 from functools import cache
 from textwrap import indent
 from threading import Lock
-from typing import Any, Optional, Protocol, Tuple
+from typing import (
+    Any,
+    Optional,
+    Protocol,
+    Tuple,
+)
 from urllib import parse
 
 import anymarkup
 import jinja2
 import jinja2.sandbox
 from deepdiff import DeepHash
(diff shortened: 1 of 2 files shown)


reconcile/openshift_resources_base.py
```diff
@@ -44,7 +67,10 @@
 from reconcile.utils.secret_reader import SecretReader
 from reconcile.utils.semver_helper import make_semver
 from reconcile.utils.sharding import is_in_shard
-from reconcile.utils.vault import SecretVersionIsNone, SecretVersionNotFound
+from reconcile.utils.vault import (
+    SecretVersionIsNone,
+    SecretVersionNotFound,
+)
```
rporres, line 73: why this lint changes? where do they come from?

cubismod: I think in my previous PR I had ruff setup incorrectly with the vscode extension. When I ran `make format` this time, it outputted these changes

rporres (commented):
Looks good except for those linting changes that I don't know where they come from...

reconcile/openshift_resources_base.py
rporres: weird, I run `make format` locally and I don't get any change...

cubismod: User error on my part :rofl:

rporres (approved):
anyway, it's fine... The current import syntax looks the most used in the code.

reconcile/openshift_resources_base.py
rporres: it's fine, it got me a bit confused. But the new syntax is the most widely used in the code.

patjlm (approved):

PR #3994, opened 2024-01-10
[rosa-authenticator] redact sensitive field in CloudWatch

cubismod:
Ensure that we aren't logging sensitive info into CloudWatch when a user interacts with rosa-authenticator.


jonmosco (approved):
LGTM.

PR #3995, opened 2024-01-12
Extended early exit cache management

hemslo:
Extended early exit cache management, with `qontract-cli` methods to test.

Support 3 methods:

* `head` check cache status
* `get` get cache value
* `set` set cache value

Local test via `localstack`

```shell
# start localstack
$ make localstack

# set envs to point to localstack s3

# check cache status
$ qontract-cli --config config.toml early-exit-cache head -i a -v b --dry-run -c {}
CacheStatus.MISS

# set cache with 30s ttl
$ qontract-cli --config config.debug.toml early-exit-cache set -i a -v b --dry-run -c {} -d {} --ttl 30

# check cache status
$ qontract-cli --config config.toml early-exit-cache head -i a -v b --dry-run -c {}
CacheStatus.HIT

# get cache value
$ qontract-cli --config config.toml early-exit-cache get -i a -v b --dry-run -c {}
desired_state={} log_output='' applied_count=0

# wait for 30s then check cache status
$ qontract-cli --config config.toml early-exit-cache head -i a -v b --dry-run -c {}
CacheStatus.EXPIRED
```

[APPSRE-8725](https://issues.redhat.com/browse/APPSRE-8725)

### DIFF
--- a/reconcile/test/utils/test_early_exit_cache.py
+++ b/reconcile/test/utils/test_early_exit_cache.py
@@ -12,18 +12,12 @@ from reconcile.utils.early_exit_cache import (
     CacheValue,
     EarlyExitCache,
 )
-from reconcile.utils.secret_reader import SecretReaderBase
 from reconcile.utils.state import State
 
 
 @pytest.fixture
 def state() -> Any:
     return create_autospec(State)
-
-
-@pytest.fixture
-def secret_reader() -> Any:
-    return create_autospec(SecretReaderBase)
 
 
 def test_early_exit_cache_build(
--- a/reconcile/test/utils/test_early_exit_cache.py
+++ b/reconcile/test/utils/test_early_exit_cache.py
@@ -1,0 +1,185 @@
+import datetime
+from typing import Any
+from unittest.mock import create_autospec
+
+import pytest
+from deepdiff import DeepHash
+from pytest_mock import MockerFixture
+
+from reconcile.utils.early_exit_cache import (
+    CacheKey,
+    CacheStatus,
+    CacheValue,
+    EarlyExitCache,
+)
+from reconcile.utils.secret_reader import SecretReaderBase
+from reconcile.utils.state import State
+
+
+@pytest.fixture
+def state() -> Any:
+    return create_autospec(State)
+
+
+@pytest.fixture
+def secret_reader() -> Any:
+    return create_autospec(SecretReaderBase)
+
+
+def test_early_exit_cache_build(
+    mocker: MockerFixture,
+    state: Any,
+    secret_reader: Any,
+) -> None:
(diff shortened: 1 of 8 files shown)


reconcile/test/utils/test_early_exit_cache.py
```diff
@@ -0,0 +1,185 @@
...
+    CacheStatus,
+    CacheValue,
+    EarlyExitCache,
+)
+from reconcile.utils.secret_reader import SecretReaderBase
+from reconcile.utils.state import State
+
+
+@pytest.fixture
+def state() -> Any:
```
chassing, line 20: NIT: You can use the `mocker` fixture here:
```python
from pytest_mock import MockerFixture

@pytest.fixture
def state(mocker: MockerFixture) -> Any:
    return mocker.create_autospec(State)
```

reconcile/test/utils/test_early_exit_cache.py
```diff
@@ -0,0 +1,185 @@
...
+    return create_autospec(State)
+
+
+@pytest.fixture
+def secret_reader() -> Any:
```
chassing, line 25: Do you mind and use the existing [secret_reader fixture](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/conftest.py#L40) instead?

chassing (approved):
Nice work!

reconcile/test/utils/test_early_exit_cache.py
hemslo: Is `mocker.create_autospec` just calling `unittest.mock.create_autospec`? Any advantage using `mocker` fixture?

reconcile/test/utils/test_early_exit_cache.py
hemslo: Nice! Reused in https://github.com/app-sre/qontract-reconcile/pull/3995/commits/ef28c2980b7c8b83726791998f4c25370ad3dcec

reconcile/test/utils/test_early_exit_cache.py
chassing: TBH, there is no real advantage. I find the `mocker` fixture just handy.

hemslo: Oh, my preference is only use `mocker.patch` to deal with dependencies need patching, for creating simple mocks/stubs I use `unitest.mock` builtin methods.

tools/qontract_cli.py
```diff
@@ -2411,6 +2412,176 @@ def rm(ctx, integration, key):
     state.rm(key)
 
 
+@root.group()
```
janboll, line 2415: I think that code should go to `reconcile/cli.py` ?

PR #3996, opened 2024-01-14
introduce ocm-internal-notifications

maorfr:
part of https://issues.redhat.com/browse/OSD-14164

this PR adds a new integration (capability) to notify users of clusters stuck in uninstalling state.

this is intended for internal users only, for non-production environments only. the intention is cost savings.

this capability is intended to run as a CronJob, once every... 24 hours? and is intended to run in the sre-capabilities instance.


reconcile/ocm_internal_notifications/integration.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def run(self, dry_run: bool) -> None:
+        gqlapi = gql.get_api()
+        environments = self.get_environments(gqlapi.query)
+
+        self.ocm_apis = self.init_ocm_apis(environments, init_ocm_base_client)
+
+        for env_name, ocm in self.ocm_apis.items():
+            if env_name == "ocm-production":
+                continue
```
maorfr, line 71: temporary, will be replaced with labels

reconcile/ocm_internal_notifications/integration.py
```diff
@@ -0,0 +1,93 @@
...
+
+            clusters = ocm.get(
+                api_path="/api/clusters_mgmt/v1/clusters",
+                params={"search": "state like 'uninstalling' and managed='true'"},
+            ).get("items")
```
geoberle, line 76: you can use the functionality from the `reconcile.utils.ocm` module collection to get the clusters and subscriptions you want in 2 OCM API calls.

```python
from reconcile.utils.ocm.search_filters import Filter
from reconcile.utils.subscriptions import get_subscriptions
from reconcile.utils.clusters import get_ocm_clusters
from reconcile.utils.base import OCMSubscription

subscription_ids = [cluster.subscription.id for cluster in get_ocm_clusters(ocm, Filter().eq("managed", true).eq("state", "uninstalling")]

subscriptions: dict[str, OCMSubscription] = get_subscriptions(ocm, Filter().in("id", subscription_ids))
```

reconcile/ocm_internal_notifications/integration.py
```diff
@@ -0,0 +1,93 @@
...
+
+            slack_user_ids = set()
+            for cluster in clusters:
+                subscription = ocm.get(api_path=cluster["subscription"]["href"])
+                creator = ocm.get(api_path=subscription["creator"]["href"])
+                email = creator["email"]
+                user, mail_address = email.split("@")
+                user_name = user.split("+")[0]
```
geoberle, line 84: `user` might already contain what you are looking for. not everyone uses gmail aliases for their OCM account. this might be a quick win to increase the hit rate for finding an adressable slack user.

some optimizations for a later stage
* RH associates can request arbitrary email aliases. maybe they are stored in LDAP and can help to find the slack handle
* cluster creators can be service accounts as well. in such cases we could look into the organization and find their members/administrators and find a something that resolves to a slack handle there
* /api/accounts_mgmt/v1/subscriptions/{subId}/notification_contacts
* for well known big organizations with high throughput we could place some contacts in org labels (demo.redhat.com)

reconcile/ocm_internal_notifications/integration.py
maorfr: this will require some bigger changes to the existing classes since an uninstalling cluster may not have an api url (pydantic complains) and since subscriptions do not currently support `creator`.

happy to see future refactors, but i'm interested in a quick(er) win.

i did follow up on making a single api call to get subscriptions, thank you! also used `Filter`. cool stuff.

reconcile/ocm_internal_notifications/integration.py
maorfr: since this is internal environments only, we are expecting mostly internal users not to RH, but to SD.

the current code covers all the existing clusters in the searched state, and i'm happy to fail forward.

reconcile/ocm_internal_notifications/label_sources.py
```diff
@@ -0,0 +1,76 @@
+from abc import (
```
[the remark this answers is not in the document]
maorfr: removed it

reconcile/ocm_internal_notifications/integration.py
geoberle: ah understood. ack

reconcile/ocm_internal_notifications/integration.py
geoberle: completely understood. the win is the reduction in API calls anyways, not pydantic.

geoberle (approved):

PR #3998, opened 2024-01-15
ROSA TGW attachments and peerings without osd-experience

patjlm:
Trying to remove the need for osd-experience/network-mgmt role for TGW attachments and peerings. This is done by using directly the cluster account terraform user instead of relying on a remote terraform user and assuming role onto the network-mgmt role.

The review should be easier one commit at a time.

The basic idea is to add an `infra_account_name` field in desired state items. This field represents the account against which we run terraform (basically the one holding the state). This allows to have requester/accepter cluster account which do not use this central account + assume_role for their provider, but can directly be targeted with their own terraform user.

The aim here is that running the integration outputs a TF script which is fully identical if we keep the `assumeRole` fields in peering connections.
When we remove those `assumeRole`:
- renamed aliases for ROSA cluster accounts where `assumeRole` is not set, and use of direct terraform user in there
- use of those aliases in TGW/Peering related resources

### DIFF
--- a/reconcile/gql_definitions/common/clusters_with_peering.gql
+++ b/reconcile/gql_definitions/common/clusters_with_peering.gql
@@ -24,6 +24,12 @@ query ClustersWithPeering {
 
     spec {
       region
+      ... on ClusterSpecROSA_v1 {
+        account {
+          name
+          uid
+        }
+      }
     }
     network {
       vpc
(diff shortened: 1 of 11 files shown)


OliviaHY:
APPSRE-8761

geoberle (approved):
LGTM. TY!!

PR #3999, opened 2024-01-15
APPSRE-8657: Log a message when Slack can't find the channel

lisa:
In some situations Slack can't send a message to a Slack channel, maybe it doesn't exist, is misspelled, is private, or maybe some API timeout. In any of those cases (whenever a `SlackApiError` is caught within `slack_notify` method), log an error message.


maorfr:
related: https://github.com/app-sre/qontract-reconcile/blob/59c00eb23f40ec94e1f8383b62e9f9211cfa890a/reconcile/utils/slack_api.py#L238-L239

reconcile/openshift_saas_deploy.py
```diff
@@ -100,7 +102,13 @@ def slack_notify(
...
     if in_progress and skip_successful_notifications:
         message += ". There will not be a notice for success."
-    slack.chat_post_message(message)
+    try:
+        slack.chat_post_message(message)
+    except SlackApiError as e:
+        if e.response["error"] == "channel_not_found":
+            logging.error(f"Slack API says can't find channel {slack.channel}")
+            # raise up to let someone else handle the exception
+            raise e
```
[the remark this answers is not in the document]
lisa: Python tabs strike again!

reconcile/openshift_saas_deploy.py
```diff
@@ -100,7 +102,13 @@ def slack_notify(
...
     if in_progress and skip_successful_notifications:
         message += ". There will not be a notice for success."
-    slack.chat_post_message(message)
+    try:
+        slack.chat_post_message(message)
+    except SlackApiError as e:
+        if e.response["error"] == "channel_not_found":
+            logging.error(f"Slack API says can't find channel {slack.channel}")
+        # raise up to let someone else handle the exception
+        raise e
```
maorfr, line 111: this already happens within `chat_post_message`, no?

lisa: Judging from the output in the Jira card I am not so sure that it is.

reconcile/openshift_saas_deploy.py
```diff
@@ -100,7 +102,13 @@ def slack_notify(
         message += f" triggered by _{trigger_integration}_"
     if in_progress and skip_successful_notifications:
         message += ". There will not be a notice for success."
-    slack.chat_post_message(message)
+    try:
+        slack.chat_post_message(message)
+    except SlackApiError as e:
```
bhushanthakur93, line 107: I suggest adding a small comment to explain why we are doing this i.e limitation of slack API to provide useful error message with the channel identifier and exposing user friendly log message to tenant.

lisa: No problem at all.

reconcile/openshift_saas_deploy.py
hemslo: There are some error handings in https://github.com/app-sre/qontract-reconcile/blob/59c00eb23f40ec94e1f8383b62e9f9211cfa890a/reconcile/utils/slack_api.py#L237, can move this case there too.

Also need to update test case to reflect the correct error type https://github.com/app-sre/qontract-reconcile/blob/59c00eb23f40ec94e1f8383b62e9f9211cfa890a/reconcile/test/utils/test_slack_api.py#L197

Btw to rethrow the exception, should just use `raise` instead of `raise e` to keep the original stack trace.

lisa: > There are some error handings in
> 
> https://github.com/app-sre/qontract-reconcile/blob/59c00eb23f40ec94e1f8383b62e9f9211cfa890a/reconcile/utils/slack_api.py#L237
> 
> , can move this case there too.

Why would `slack_api.py` care about sending a warning about a channel not existing? It seems like that should be the domain of the user of the API.

> Btw to rethrow the exception, should just use `raise` instead of `raise e` to keep the original stack trace.

I'll make this change.

reconcile/openshift_saas_deploy.py
```diff
@@ -100,6 +100,10 @@ def slack_notify(
         message += f" triggered by _{trigger_integration}_"
     if in_progress and skip_successful_notifications:
         message += ". There will not be a notice for success."
+        # When a message is sent to #someChannel and the Slack API can't find
```
bhushanthakur93, line 103: Move the comment to `slack_api`

bhushanthakur93 (changes requested):

bhushanthakur93 (approved):
LGTM

PR #4000, opened 2024-01-17
APPSRE8701 retain last successed builds for CI

skryzhny:
If last build is cleaned then SCM that pools git consider as it newer built last commit and initiates new (duplicated) build.

We should preserve last SUCCESSFUL built even if cleanup rules matches it.

### DIFF
--- a/reconcile/jenkins_job_builds_cleaner.py
+++ b/reconcile/jenkins_job_builds_cleaner.py
@@ -33,9 +33,10 @@ def delete_builds(jenkins, builds_todel, dry_run=True):
                 msg = f"failed to delete {job_name}/{build_id}"
                 logging.exception(msg)
 
+
 def get_last_build_ids(builds):
     builds_to_keep = []
-    #FIXME need to explicitly sort and not rely on order API returns
+    # FIXME need to explicitly sort and not rely on order API returns
 
     good_build_timestamp = 0
     for build in builds:
@@ -51,6 +52,7 @@ def get_last_build_ids(builds):
                 good_build_timestamp = build["timestamp"]
     return builds_to_keep
 
+
 def find_builds(jenkins, job_names, rules):
     # Current time in ms
     time_ms = time.time() * 1000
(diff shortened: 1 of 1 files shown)


reconcile/jenkins_job_builds_cleaner.py
```diff
@@ -43,7 +43,17 @@ def find_builds(jenkins, job_names, rules):
...
                 builds = jenkins.get_builds(job_name)
+                # We need to keep last successful build (https://issues.redhat.com/browse/APPSRE-8701)
+                good_build_timestamp = 0
                 for build in builds:
+                    if build["result"] == "SUCCESS":
+                        if build["timestamp"] > good_build_timestamp:
+                            build_to_keep = build["id"]
+                            good_build_timestamp = build["timestamp"]
+                for build in builds:
+                    if build["id"] == build_to_keep:
```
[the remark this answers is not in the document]
skryzhny: moved to dedicated function but not sorted yet

reconcile/jenkins_job_builds_cleaner.py
```diff
@@ -34,6 +34,19 @@ def delete_builds(jenkins, builds_todel, dry_run=True):
                 logging.exception(msg)
 
 
+def get_last_build_ids(builds):
+    builds_to_keep = []
+    if builds != []:
```
chassing, line 39: Can be written as `if builds:`

reconcile/jenkins_job_builds_cleaner.py
```diff
@@ -34,6 +34,19 @@ def delete_builds(jenkins, builds_todel, dry_run=True):
...
+        last_build = sorted(builds, key=lambda b: b["timestamp"], reverse=True)[0]
```
chassing, line 40: Please don't sort `builds` twice!

reconcile/jenkins_job_builds_cleaner.py
```diff
@@ -36,11 +36,11 @@ def delete_builds(jenkins, builds_todel, dry_run=True):
 
 def get_last_build_ids(builds):
     builds_to_keep = []
-    if builds != []:
+    if builds:
         last_build = sorted(builds, key=lambda b: b["timestamp"], reverse=True)[0]
         builds_to_keep.append(last_build["id"])
 
-    for build in sorted(builds, key=lambda b: b["timestamp"], reverse=True):
+    for build in builds:
```
chassing, line 43: But `builds` is not sorted anymore. I would suggest:

```python
def ...
    sorted_builds = sorted(builds, key=lambda b: b["timestamp"], reverse=True)
    if sorted_builds:
        last_build = sorted_builds[0]
    ...
    for build in sorted_builds:
    ....
```

skryzhny: was not my best day, fixed

chassing (approved):
LGTM

PR #4003, opened 2024-01-18
tgw-attachments: fix early-exit desiredstate

patjlm:
Reducing the list of accounts taken into account for the early exit datasource. This list had been wrongly extended in https://github.com/app-sre/qontract-reconcile/pull/3998

### DIFF
--- a/reconcile/terraform_tgw_attachments.py
+++ b/reconcile/terraform_tgw_attachments.py
@@ -471,7 +471,9 @@ def run(
 
 
 def early_exit_desired_state(*args: Any, **kwargs: Any) -> dict[str, Any]:
-    return _fetch_desired_state_data_source().dict(by_alias=True)
+    datasource = _fetch_desired_state_data_source()
+    datasource.accounts = _filter_tgw_accounts(datasource.accounts, datasource.clusters)
+    return datasource.dict(by_alias=True)
 
 
 def desired_state_shard_config() -> DesiredStateShardConfig:
--- a/reconcile/test/test_terraform_tgw_attachments.py
+++ b/reconcile/test/test_terraform_tgw_attachments.py
@@ -1133,7 +1133,7 @@ def test_early_exit_desired_state(
 
     expected_early_exit_desired_state = {
         "clusters": [cluster_with_tgw_connection.dict(by_alias=True)],
-        "accounts": [tgw_account.dict(by_alias=True), vpc_account.dict(by_alias=True)],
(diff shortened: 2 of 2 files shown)


reconcile/terraform_tgw_attachments.py
```diff
@@ -471,7 +471,9 @@ def run(
 
 
 def early_exit_desired_state(*args: Any, **kwargs: Any) -> dict[str, Any]:
-    return _fetch_desired_state_data_source().dict(by_alias=True)
+    datasource = _fetch_desired_state_data_source()
+    datasource.accounts = _filter_tgw_accounts(datasource.accounts, datasource.clusters)
```
hemslo, line 475: can we do this inside `_fetch_desired_state_data_source`, are non filtered accounts used anywhere?

hemslo: or split `accounts` into `all_accounts` and `tgw_accounts` so the same `DesiredStateDataSource` can be used directly.

patjlm: > are non filtered accounts used anywhere?
yes they are now. They're used to get cluster accounts (non-tgw).

> split accounts into all_accounts and tgw_accounts so the same DesiredStateDataSource can be used directly.
But if `all_account` is in the desired state `datasource.dict(by_alias=True)` will also return all accounts in the early exit. Or do you then propose to only return `clusters` and `tgw_accounts` ? It looks pretty much the same than what is proposed here IMHO.

hemslo: If `all_account` is used, should it be included in early exit state anyway? Otherwise it may wrongly exit.

patjlm: I think it should not. From `all_accounts` we end up using only the ones with TGW and the ones corresponding to ROSA/CCS clusters

hemslo: if ehe ones corresponding to ROSA/CCS clusters changed, will desired state change?

hemslo: So what we need is `tgw_rosa_cluster_accounts` and `tgw_accounts`, right? Are we using any other accounts inside `all_accounts`?

patjlm: > if the ones corresponding to ROSA/CCS clusters changed, will desired state change?

No

> So what we need is tgw_rosa_cluster_accounts and tgw_accounts, right? 

I don't understand what you want to gain there.

PR #4004, opened 2024-01-18
Extended early exit for terraform resources

hemslo:
Extended early exit for terraform resources.

* Refactored terraform resources for easy split desired state and apply result
* Add `extended_early_exit_run` to use cache to do early exit
* `extended_early_exit_run` applied in terraform resources to achieve early exit
* export metrics `qontract_reconcile_extended_early_exit_count`

Test locally:

```shell
# start localstack
$ make localstack

# setup envs to use local s3

# run multiple times to verify early exit
$ qontract-reconcile --config config.toml \
    --log-level DEBUG \
    --dry-run \
    terraform-resources \
    --enable-extended-early-exit \
    --extended-early-exit-cache-ttl-seconds 3600 \
    --log-cached-log-output \
    --account-name YOUR_ACCOUNT_NAME
```

[APPSRE-8725](https://issues.redhat.com/browse/APPSRE-8725)

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -453,11 +453,13 @@ def runner(
     populate_desired_state(ri, ts.resource_spec_inventory)
 
     ob.publish_metrics(ri, QONTRACT_INTEGRATION)
-    actions = []
-    if oc_map:
-        actions = ob.realize_data(
-            dry_run, oc_map, ri, thread_pool_size, caller=acc_name
-        )
+    actions = ob.realize_data(
+        dry_run,
+        oc_map,
+        ri,
+        thread_pool_size,
+        caller=acc_name,
+    )
 
     if actions and vault_output_path:
         write_outputs_to_vault(vault_output_path, ts.resource_spec_inventory)
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -181,8 +181,7 @@ def exclude_accounts_by_name(
 def validate_account_names(
     accounts: Collection[Mapping[str, Any]], names: Collection[str]
 ) -> None:
-    missing_names = set(names) - {a["name"] for a in accounts}
-    if missing_names:
+    if missing_names := set(names) - {a["name"] for a in accounts}:
         raise ValueError(
             f"Accounts {missing_names} were provided as arguments, but not found in app-interface. Check your input for typos or for missing AWS account definitions."
         )
@@ -214,7 +213,7 @@ def get_aws_accounts(
     if exclude_accounts:
         validate_account_names(accounts, exclude_accounts)
         filtered_accounts = exclude_accounts_by_name(accounts, exclude_accounts)
-        if len(filtered_accounts) == 0:
+        if not filtered_accounts:
             raise ValueError("You have excluded all aws accounts, verify your input")
         return filtered_accounts
 
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -32,7 +32,10 @@ from reconcile.typed_queries.terraform_namespaces import get_namespaces
 from reconcile.utils import gql
 from reconcile.utils.aws_api import AWSApi
 from reconcile.utils.defer import defer
-from reconcile.utils.extended_early_exit import extended_early_exit_run
+from reconcile.utils.extended_early_exit import (
+    ExtendedEarlyExitRunnerResult,
+    extended_early_exit_run,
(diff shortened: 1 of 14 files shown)


reconcile/terraform_resources.py
```diff
@@ -172,64 +167,72 @@ def init_working_dirs(
...
+) -> list[dict[str, Any]]:
+    return [ac for ac in accounts if ac["name"] not in names]
 
 
 def validate_account_names(
     accounts: Collection[Mapping[str, Any]], names: Collection[str]
 ) -> None:
-    if len(accounts) != len(names):
-        missing_names = set(names) - {a["name"] for a in accounts}
+    missing_names = set(names) - {a["name"] for a in accounts}
```
chassing, line 184: nit: you can use `if missing_names := set ...` there

reconcile/terraform_resources.py
```diff
@@ -172,64 +167,72 @@ def init_working_dirs(
...
-    if not include_accounts and exclude_accounts:
-        excluding = filter_accounts_by_name(accounts, exclude_accounts)
-        validate_account_names(excluding, exclude_accounts)
-        accounts = exclude_accounts_by_name(accounts, exclude_accounts)
-        if len(accounts) == 0:
+
+    if exclude_accounts:
+        validate_account_names(accounts, exclude_accounts)
+        filtered_accounts = exclude_accounts_by_name(accounts, exclude_accounts)
+        if len(filtered_accounts) == 0:
```
chassing, line 217: nit: please use `if not filtered_accounts:` instead

reconcile/terraform_resources.py
```diff
@@ -368,63 +356,79 @@ def run(
...
     if dry_run:
-        cleanup_and_exit(tf)
+        return
 
+    acc_name = accounts[0]["name"] if accounts else None
     if not light and tf.should_apply:
         err = tf.apply()
         if err:
-            cleanup_and_exit(tf, err)
+            raise RuntimeError("Terraform apply has errors")
```
chassing, line 431: Do you really want to display a python stacktrace here? A `logging.error` would be sufficient, or what do you think?
The same applies to the other `RuntimeErrors` above and below.

hemslo: previously it's using `sys.exit(1)` instead of exceptions, even `logging.error` here, I still need to let it exit in error state, and I remember `sys.exit(1)` is no longer recommended in this code base, right?

reconcile/terraform_resources.py
hemslo: nice, fixed in https://github.com/app-sre/qontract-reconcile/pull/4004/commits/7d22564b05b0b740b750c5d29291177ce7fdf5f8

reconcile/terraform_resources.py
hemslo: nice, fixed in https://github.com/app-sre/qontract-reconcile/pull/4004/commits/7d22564b05b0b740b750c5d29291177ce7fdf5f8

reconcile/terraform_resources.py
chassing: I think the usage of `sys.exit(1)` is valid here.

hemslo: For current usage `sys.exit(1)` is ok, but this logic is not in root level functions, and `sys.exit(1)` means to raise `SystemExit`, which can't be caught by `expect Exception` (need `expect SystemExit`), may surprise people if any other error handing logic added later.

Any concerns regarding print additional stacktrace here? This error shouldn't happen often, and it's just more lines in log, right?

chassing: Ok 👍

lechuk47:
What if we need to retry a dry-run run that has been cached? 
A Terraform run can fail due to external causes (timeouts, etc).  From what I see, that is going to be cached as well right?

hemslo:
> What if we need to retry a dry-run run that has been cached? A Terraform run can fail due to external causes (timeouts, etc). From what I see, that is going to be cached as well right?

If it's for local test, then run it without `--enable-extended-early-exit` would run without cache.

Only success run will be cached, see test case `test_extended_early_exit_run_when_error` https://github.com/app-sre/qontract-reconcile/pull/4004/files#diff-a3548109883b8dc5d0488be99cf31732c68bfa2031d5b4e806d8f3be84124874R294.

To retry a success run when cache is still valid, need to delete the cache or update it's metadata to let it expire.

janboll (approved):

PR #4007, opened 2024-01-19
Fix change admission

janboll:
Each changetype needs only one approval.

- Filter out change types that are restricted.
- Ensure at least one approval is set for each changetypecontext

Fixes: [APPSRE-8776](https://issues.redhat.com/browse/APPSRE-8776)


reconcile/change_owners/change_owners.py
```diff
@@ -244,12 +244,16 @@ def init_gitlab(gitlab_project_id: str) -> GitLabApi:
 
 
 def is_coverage_admitted(
-    coverage: ChangeTypeContext, mr_author: str, good_to_test_approvers: set[str]
+    contexts: list[ChangeTypeContext], mr_author: str, good_to_test_approvers: set[str]
```
fishi0x01, line 247: nit: `Iterable`

fishi0x01 (approved):
Phew change type magic :tada: :)
LGTM FWIW - maybe @geoberle has better insights though

reconcile/change_owners/change_owners.py
```diff
@@ -260,12 +264,18 @@ def is_change_admitted(
...
+        c
         for change in changes
         for dc in change.diff_coverage
         for c in dc.coverage
         if c.change_type_processor.restrictive
+    ]
+
+    return all(
+        is_coverage_admitted(to_check, mr_author, good_to_test_approvers)
+        for dc in to_check
```
hemslo, line 278: `dc` is not used in `is_coverage_admitted(to_check, mr_author, good_to_test_approvers)`.

It seems the new logic is the changes are admitted when all restrictive coverage origins are admitted, right?

Not sure what's `origin`, but this maybe clear, keep original `is_coverage_admitted` not changed.

```python
    restrictive_coverages = [
        c
        for change in changes
        for dc in change.diff_coverage
        for c in dc.coverage
        if c.change_type_processor.restrictive
    ]

    change_types_to_approve = {c.origin for c in restrictive_coverages}
    change_types_approved = {c.origin for c in restrictive_coverages if is_coverage_admitted(c, mr_author, good_to_test_approvers)}
    return change_types_to_approve == change_types_approved
```

May also need to update function doc.

janboll: Great spot! Totally missed that... will update it

hemslo (approved):
LGTM!

PR #4008, opened 2024-01-19
APPSRE-8505 unbatch failed MRs

fishi0x01:
Incremental step 3 for batching SAPM MRs.

Adding logic for closing batched MRs that do not pass MR checks. Through unbatching, we prevent unrelated MRs blocking each other.

The batching logic itself will be added in a later step.


reconcile/saas_auto_promotions_manager/merge_request_manager/merge_request_manager.py
```diff
@@ -214,8 +215,24 @@ def _unbatch_failed_mrs(self) -> None:
...
+                # The MR is already marked as not batchable which means we do not need to unbatch
+                open_mrs_after_unbatching.append(mr)
+                continue
+            if not mr.failed_mr_check:
+                # The MR passed check and doesnt need unbatching
+                open_mrs_after_unbatching.append(mr)
+                continue
+            self._vcs.close_app_interface_mr(
+                mr,
+                "Closing this MR because it failed MR check and isnt marked un-batchable yet.",
```
hemslo, line 230: `isnt` -> `is not` / `isn't`

reconcile/saas_auto_promotions_manager/merge_request_manager/merge_request_manager.py
```diff
@@ -214,8 +215,24 @@ def _unbatch_failed_mrs(self) -> None:
...
+            if not mr.is_batchable:
+                # The MR is already marked as not batchable which means we do not need to unbatch
+                open_mrs_after_unbatching.append(mr)
+                continue
+            if not mr.failed_mr_check:
+                # The MR passed check and doesnt need unbatching
+                open_mrs_after_unbatching.append(mr)
+                continue
+            self._vcs.close_app_interface_mr(
+                mr,
```
hemslo, line 229: type mismatch, `close_app_interface_mr` is accepting `ProjectMergeRequest` but passed `OpenMergeRequest`

reconcile/saas_auto_promotions_manager/merge_request_manager/merge_request_manager.py
```diff
@@ -214,8 +215,24 @@ def _unbatch_failed_mrs(self) -> None:
         and close the old batched MR. By doing so, we ensure that unrelated MRs are not blocking each other.
         Unbatched MRs are marked and will never be batched again.
         """
-        # TODO: implemented in follow-up MR
-        pass
+        open_mrs_after_unbatching: list[OpenMergeRequest] = []
+        for mr in self._open_mrs:
```
hemslo, line 219: The code in `_unbatch_failed_mrs` has some duplicates and complex loop/condition logic, better to extract a method to keep collection operation and single item operation seperate. Like this:

```python
 def _unbatch_mr(self, mr: OpenMergeRequest) -> bool:
    if not mr.is_batchable:
        return False
    if not mr.failed_mr_check:
        return False
    self._vcs.close_app_interface_mr(...)
    ...
    return True
```

```python
def _unbatch_failed_mrs(self) -> None:
    self._open_mrs = [mr for mr in self._open_mrs if not self._unbatch_mr(mr)]
````
```

```

File: reconcile/test/saas_auto_promotions_manager/merge_request_manager/merge_request_manager/test_mr_batching.py [hunk 4]
```diff
@@ -0,0 +1,111 @@
...
+        ],
+        # All MR checks fail
+        PIPELINE_RESULTS: [MRCheckStatus.FAILED, MRCheckStatus.FAILED],
+    })
+    merge_request_manager = MergeRequestManager(
+        vcs=vcs,
+        renderer=renderer,
+    )
+    merge_request_manager.housekeeping()
+    assert vcs.close_app_interface_mr.call_count == 2  # type: ignore[attr-defined]
```
hemslo, line 67: better to verify all calls using [assert_has_calls](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_has_calls), not just call count, but also args.

reconcile/saas_auto_promotions_manager/merge_request_manager/merge_request_manager.py
fishi0x01: Interesting mypy didnt catch that :thinking:

reconcile/saas_auto_promotions_manager/merge_request_manager/merge_request_manager.py
fishi0x01: Nice idea! But would like to keep this in a single function for now. I think the cognitive load of the function is acceptable.

esron (approved):
LGTM

reconcile/saas_auto_promotions_manager/merge_request_manager/merge_request_manager.py
hemslo: OK, but try not to use multiple `continue` in a large loop, it's like `GOTO`, hard to track logic flow. Since this logic is quite simple, maybe a positive filter is cleaner:

```python
        if mr.is_batchable and mr.failed_mr_check:
            self._vcs.close_app_interface_mr(
                mr.raw,
                "Closing this MR because it failed MR check and isn't marked un-batchable yet.",
            )
            # Remember these hashes as unbatchable
            content_hashes = mr.content_hashes.split(ITEM_SEPARATOR)
            self._unbatchable_hashes.update(content_hashes)
        else:
            open_mrs_after_unbatching.append(mr)
```

hemslo (approved):
LGTM!

reconcile/saas_auto_promotions_manager/merge_request_manager/merge_request_manager.py
fishi0x01: Very good point. Fixed :+1:

PR #4010, opened 2024-01-23
gitlab-housekeeping add app_sre_bot label to qontract_reconcile_merged_merge_requests

maorfr:
with this label we can differentiate between MRs submitted by the bot or not.

this will allow us to measure increase in AppSRE productivity by seeing an increase in bot MRs and a decrease in team members MRs.


maorfr:
do you really mind that much? i'll implement what you are proposing but it someone will say "why are we hardcoding app_sre and not something else".

this PR is straight forward.

rporres:
It's way nicer to say what we want with just a label.

Cardinality wise I was wrong, we would have the same increase: three time series instead of one in each case.

Do I care that much so as to block this unless changed? No.

PR #4011, opened 2024-01-23
[ocm-internal-notifications] forgive slack user not found

maorfr:
related to https://issues.redhat.com/browse/OSD-14164

instead of failing, log a warning and continue to provide a heads up at least to users that are found.


patjlm (approved):
Thanks!

PR #4012, opened 2024-01-23
PoC: External Resoures V2

lechuk47:
External Resources V2 Integration. 
This PR adds a working integration to manage External Resources using Jobs (JobController). Since the amount of code is getting bigger and bigger, I think it's worth starting with this working code and add improvements in smaller pieces of code. This way the reviews will be easier to manage and more concise.


reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,423 @@
...
+                                "image": image,
+                                "imagePullPolicy": "Always",
+                                "env": [
+                                    {
+                                        "name": "DRY_RUN",
+                                        "value": "False",
+                                    },
+                                    {
+                                        "name": "INPUT",
+                                        "value": input,
```
janboll, line 106: Is this terraform config? Any size limitations here?

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,423 @@
...
+                                        "subPath": "credentials",
+                                    },
+                                    {
+                                        "name": "workdir",
+                                        "mountPath": "/work",
+                                    },
+                                ],
+                            }
+                        ],
+                        "containers": [
```
chassing, line 126: Nice idea to use an init-container for the actual work and a std container for outputs.

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,423 @@
...
+                break
+
+        return resources
+
+    def resource_needs_reconciliation(
+        self,
+        resource: ExternalResource,
+        state: ExternalResourceState,
+    ) -> bool:
+        if state.status == ResourceStatus.NOT_EXISTS:
```
chassing, line 286: you can use [match ... case](https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching) here.

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,423 @@
...
+                            state.status = ResourceStatus.ERROR
+                        logging.info("Resource Status updated to " + state.status)
+            except StateNotFoundError:
+                state = ExternalResourceState(
+                    key=key,
+                    ts=datetime.now(timezone.utc),
+                    status=ResourceStatus.NOT_EXISTS,
+                )
+            except Exception:
+                raise
```
hemslo, line 344: try block should be limit to only related to expected error `StateNotFoundError`, it looks like only `self.state_mgr.get_external_resource_state(key)` can throw it, but I can't find the location throwing it. And if no handling for generic `Exception`, no need to except it.

Worth extract method to simplify it:

```python
    def _get_state(self, key: ExternalResourceKey) -> ExternalResourceState:
        try:
            state = self.state_mgr.get_external_resource_state(key)
        except StateNotFoundError:
            return ExternalResourceState(
                key=key,
                ts=datetime.now(timezone.utc),
                status=ResourceStatus.NOT_EXISTS,
            )

        if state.status != ResourceStatus.IN_PROGRESS:
            return state

        match self.impl.get_resource_reconcile_status(key):
            case ReconcileStatus.SUCCESS:
                state.status = ResourceStatus.CREATED
            case ReconcileStatus.ERROR:
                state.status = ResourceStatus.ERROR
            case _:
                return state

        state.ts = datetime.now(timezone.utc)
        logging.info("Resource Status updated to " + state.status)
        return state
```

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,423 @@
...
+            state = self.state_mgr.get_external_resource_state(key)
+            if state.status == ResourceStatus.DELETE_IN_PROGRESS:
+                reconcile_status = self.impl.get_resource_reconcile_status(key)
+                if reconcile_status != ReconcileStatus.IN_PROGRESS:
+                    state.ts = datetime.now(timezone.utc)
+                    if reconcile_status == ReconcileStatus.SUCCESS:
+                        state.status = ResourceStatus.DELETED
+                    elif reconcile_status == ReconcileStatus.ERROR:
+                        state.status = ResourceStatus.ERROR
+                    logging.info("Updating resource status to " + state.status)
```
hemslo, line 387: similar here, worth extracting method, btw I'm thinking `handle_desired_resources` and `handle_deleted_resources` are called together, can we simplify it so one method handle iterating and pair desired state and current state.

```python
    def handle_resources(
        self,
        specs: Iterable[ExternalResourceSpec],
    ) -> None:
        spec_by_key = {ExternalResourceKey.from_spec(spec): spec for spec in specs}
        diff_result = diff_mappings(
            self.state_mgr.index,
            spec_by_key,
            lambda _c, _d: True,
        )
        
        for key, spec in diff_result.add.items():
            handle_add_resource(key, spec)
        
        for key, pair in diff_result.identical.items():
            handle_identical_resource(key, pair.current, pair.desired)
        
        for key, current in diff_result.delete.items():
            handle_delete_resource(key, current)
```

```python

```

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,423 @@
...
+        return f"{module.image}:{module.default_version}"
+
+    def build_external_resource(self, ers: ExternalResourceSpec) -> ExternalResource:
+        resource = self.providers.get_factory(
+            ers.provision_provider
+        ).create_external_resource(ers)
+        return resource
+
+    def serialize_resource_input(self, resource: ExternalResource) -> str:
+        return base64.b64encode(json.dumps(resource.dict()).encode()).decode()
```
hemslo, line 423: we have one similar function in https://github.com/app-sre/qontract-reconcile/blob/8b4d20b528fa91849022d4726e1c5563afd99e9a/reconcile/utils/openshift_resource.py#L715, worth extract and reuse.

reconcile/external_resources/state.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def del_external_resource_state(self, key: ExternalResourceKey) -> None:
+        del self.index[key]
+
+    def get_all_resource_keys(self) -> list[ExternalResourceKey]:
+        return list[ExternalResourceKey](self.index.keys())
+
+    def save_state(self) -> None:
+        self._write_index_file()
+        self._write_external_resource_states()
```
hemslo, line 131: What's the relationship between these 2 files? Now I can see some transaction issues if one success the other fail.

lechuk47: Each index item has an external_resource_state (ERS) stored in a separate file. The attribute has a value if the item has been reconciled.
write_external_resource_states iterates over all elements in the index and stores existent external_resource_states in their separate file. 

I chose this way to reduce the index file size; if no reconciliation is needed, there is no need to get the external_resource_state data.  The external_resource_state data is only used if a resource needs to be deleted (to get the Job image and input in case the spec is not present anymore).

So the only possible issue I foresee if the second call fails is that the deletion might get a different input data.... Let me sleep a bit on this to think about a better solution.

hemslo: So index file `external_resources_index.json` content is a list of all `ExternalResourceState` without `resource_state` field (where is this field defined?).

Both `ExternalResourceReconcileData` and `ExternalResourceState` have a same key field, do you want to use a one-to-one mapping between them and store in 2 different objects?

Let me try to grab the requirement here:

* Have a way to query all `ExternalResourceState`
* Have a way to query one `ExternalResourceReconcileData`

Put all `ExternalResourceState` in one index file may lead to concurrent issues, other storage such as dynamodb can use a different schema and index to better achieve it.

If we stick to S3 and use it as pure kv store, can use object path as a way embed info and to do query, so we can keep `ExternalResourceState` and `ExternalResourceReconcileData` in the same object, use list objects with path filter like `in-progress/key`, `success/key`. Just status change requires putting a new object in a different path and delete the old one.

lechuk47: Oh sorry, `resource_state === reconcile_data`.  The reconcile_data is already part of the object but is stored in a separate file. 

```
class ExternalResourceState(BaseModel):
    key: ExternalResourceKey
    ts: datetime
    status: ResourceStatus
    resource_hash: Optional[str] = ""
    reconcile_data: Optional[ExternalResourceReconcileData] = None 
```
There is other data in de index entries like the object_hash, used to determine if a resource has changed. I fell like this is an early optimization. Maybe is a better idea (for now) to use one file per resources and lets make the optimization later.

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,578 @@
...
+    ) -> None:
+        self.state_mgr = state_manager
+        self.oc = oc
+        self.cluster = cluster
+        self.namespace = namespace
+        self.settings = settings
+        self.modules = modules
+        self.impl = impl
+        self.dry_run = dry_run
+        self.providers = ObjectFactory[ExternalResourceFactory]()
```
janboll, line 303: A great use case for factory pattern!
So, extending for i.e. GCP would simply mean registering a new factory?

reconcile/external_resources/state.py
```diff
@@ -0,0 +1,246 @@
...
+    KEY_PROVISIONER_NAME = "key.provisioner_name"
+    KEY_PROVIDER = "key.provider"
+    KEY_IDENTIFIER = "key.identifier"
+    RECONCILIATION_RESOURCE_DIGEST = "reconcilitation.resource_digest"
+    RECONCILIATION_IMAGE = "reconciliation.image"
+    RECONCILIATION_INPUT = "reconciliation.input"
+    RECONCILIATION_ACTION = "reconciliation.action"
+    RESOURCE_KEY = "resource_key"
+    RESOURCE_STATUS = "resource_status"
+    RESOURCE_DIGEST = "resource_digest"
```
janboll, line 120: nit: In other places we do not use _digest_ but _hash_ for similar concepts. Perhaps stick to it?

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,578 @@
...
+            if _obj:
+                obj = OpenshiftResource(
+                    _obj, QONTRACT_INTEGRATION, QONTRACT_INTEGRATION_VERSION
+                )
+
+        if obj is None:
+            return ReconcileStatus.NOT_EXISTS
+
+        status = obj.body["status"]
+        backofflimit = obj.body["spec"].get("backoffLimit", 6)
```
janboll, line 236: Perhaps worth a global or a setting?

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,578 @@
...
+            )
+            module = self.modules[m_key]
+            self.impl.reconcile_resource(
+                reconciliation=r,
+                module=module,
+                dry_run=self.dry_run,
+            )
+            running_jobs.add(r)
+
+        success_jobs, error_jobs = self._wait_for_running_jobs(running_jobs)
```
janboll, line 553: Why is dry-run handling job results? Wouldn't that block MRs?

reconcile/external_resources/manager.py
lechuk47: Done, This changed recently but I managed to use match-case.

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,423 @@
...
+        if obj is not None:
+            resource_status = obj.body["status"]
+            backofflimit = obj.body["spec"].get("backoffLimit", 6)
+            if resource_status.get("succeeded", 0) > 0:
+                status = ReconcileStatus.SUCCESS
+            elif resource_status.get("failed", 0) >= backofflimit:
+                status = ReconcileStatus.ERROR
+            else:
+                status = ReconcileStatus.IN_PROGRESS
+        return status
```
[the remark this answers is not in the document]
lechuk47: This code no longer exists in this PR.

reconcile/external_resources/manager.py
lechuk47: This also changed. I close this comment, PTAL to the new code.

reconcile/external_resources/manager.py
lechuk47: This also changed, I can not use your suggestions as to get the specs for the deleted resources I need to do a table scan.

reconcile/external_resources/state.py
lechuk47: This changed a lot. PTAL again.

reconcile/external_resources/manager.py
lechuk47: This changed a bit. There 2 levels of factories, 1 to use tech stacks and another one for providers. 

1.- CDKTF/Terraform, others
2.- AWS-IAM-ROLE, CLOUDFLARE-..., GCP-... 

But yes, the idea is to isolate different use-cases/tech-stacks.

reconcile/external_resources/manager.py
lechuk47: This code no longer exists here.

reconcile/external_resources/manager.py
lechuk47: MRs need to report the status of the JOBS (TF PLAN) back to the PR to feed the information back to users. Each resource change runs in a JOB in a dry-run mode (Tf plan) and gets the results.

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,578 @@
...
+                        provisioner=erp.provisioner.dict(),
+                        resource=er.dict(),
+                        namespace=ns.dict(),
+                    )
+                    resources.append(ers)
+                break
+
+        return resources
+
+    def resource_needs_reconciliation(
```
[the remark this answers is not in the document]
lechuk47: changed.

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,578 @@
...
+        self.settings = settings
+        self.modules = modules
+        self.impl = impl
+        self.dry_run = dry_run
+        self.providers = ObjectFactory[ExternalResourceFactory]()
+        self.providers.register_factory(
+            "aws", AWSExternalResourceFactory(settings=settings, modules=modules)
+        )
+
+    def get_external_resources(
```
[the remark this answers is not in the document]
lechuk47: External resources only live in namespace files, and the function Gets them using a gql query, like every other integration does. I'm not sure I understand you suggestion but if if you have a better name I will consider it.

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,330 @@
...
+                    reason = "Resource status ERROR"
+        if reconcile:
+            logging.info(
+                "Reconciling: Status: [%s], Action: [%s], reason: [%s], key:[%s]",
+                state.resource_status.value,
+                reconciliation.action.value,
+                reason,
+                reconciliation.key,
+            )
+        return reconcile
```
hemslo, line 139: there are many `reconcile = True` and set `reason`, how about extract all reason string to enum, like `ReconcileStatus`, then it can be something like this:

```python
class ReconcileStatus(Enum):
    Reconciled = "Reconciled"
    NotExists = "NotExists"
    ...

    def _get_reconcile_status(
        self,
        reconciliation: Reconciliation,
        state: ExternalResourceState,
    ) -> ReconcileStatus:
        if reconciliation.action == Action.APPLY:
            match state.resource_status:
                case ResourceStatus.NOT_EXISTS:
                    return ReconcileStatus.NotExists
            ...
        if reconciliation.action == Action.DESTROY:
            ...

    def _resource_needs_reconciliation(
        self,
        reconciliation: Reconciliation,
        state: ExternalResourceState,
    ) -> bool:
        reconcile_status = self._get_reconcile_status(reconciliation, state)
        reconcile = reconcile_status != ReconcileStatus.Reconciled
        if reconcile:
            logging...
        return reconcile
```

reconcile/external_resources/integration.py
```diff
@@ -0,0 +1,93 @@
...
+            cluster, namespace, secret_reader, vault_path="app-sre"
+        ),
+    )
+
+    if dry_run:
+        er_mgr.handle_dry_run_resources()
+        if er_mgr.errors:
+            logging.error("Validation Errors:")
+            for k, e in er_mgr.errors.items():
+                logging.error("ExternalResourceKey: %s, Error: %s" % (k, e))
```
hemslo, line 91: should integration fail when there are errors? like raise `ExceptionGroup` here

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,330 @@
...
+            state = self.state_mgr.get_external_resource_state(key)
+            reconciliation = Reconciliation(
+                key=key,
+                resource_hash=state.resource_hash,
+                module_configuration=state.reconciliation.module_configuration,
+                input=state.reconciliation.input,
+                action=Action.DESTROY,
+            )
+            r.add(reconciliation)
+        return r
```
hemslo, line 180: can use set comprehension here

```python
        return {
            Reconciliation(
                key=key,
                resource_hash=state.resource_hash,
                module_configuration=state.reconciliation.module_configuration,
                input=state.reconciliation.input,
                action=Action.DESTROY,
            )
            for key in deleted_keys
            if (state := self.state_mgr.get_external_resource_state(key)) is not None
        }
```

if `state` always not None, maybe a Reconciliation builder can make it clear:


```python
        return {
            Reconciliation.build_destroy_action(
                key=key,
                state=self.state_mgr.get_external_resource_state(key),
            )
            for key in deleted_keys
        }
```

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,330 @@
...
+        :param r: Reconciliation object
+        :param state: State object
+        :return: True/False if target Secret needs to be reconciled.
+        """
+        need_secret_sync = False
+
+        if state.resource_status not in set([
+            ResourceStatus.DELETE_IN_PROGRESS,
+            ResourceStatus.IN_PROGRESS,
+        ]):
```
hemslo, line 198: can use set literal to avoid create list first

```python
{ResourceStatus.DELETE_IN_PROGRESS, ResourceStatus.IN_PROGRESS}
```

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,330 @@
...
+                    r.action.value,
+                    r.key,
+                )
+                error = True
+        if error:
+            state.resource_status = ResourceStatus.ERROR
+            state.reconciliation_errors = state.reconciliation_errors + 1
+            self.state_mgr.set_external_resource_state(state)
+
+        return need_secret_sync
```
hemslo, line 244: `error` and `need_secret_sync` live too long in this method, make it harder to track changes, how about unify status handling and do early return?

```python
        match self.reconciler.get_resource_reconcile_status(state.reconciliation):
            case ReconcileStatus.SUCCESS:
                logging.info(
                    "Reconciliation ended SUCCESSFULLY. Action: %s, key:%s",
                    r.action.value,
                    r.key,
                )
                if r.action == Action.APPLY:
                    state.resource_status = ResourceStatus.CREATED
                    state.reconciliation_errors = 0
                    self.state_mgr.set_external_resource_state(state)
                    return True
                elif r.action == Action.DESTROY:
                    state.resource_status = ResourceStatus.DELETED
                    self.state_mgr.del_external_resource_state(r.key)
                    return False
            case (ReconcileStatus.ERROR | ReconcileStatus.NOT_EXISTS) as error_status:
                logging.info(
                    "Reconciliation ended with error status %s: Action:%s, Key:%s",
                    error_status,
                    r.action.value,
                    r.key,
                )
                state.resource_status = ResourceStatus.ERROR
                state.reconciliation_errors = state.reconciliation_errors + 1
                self.state_mgr.set_external_resource_state(state)
                return False
            case _:
                return False
```

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,330 @@
...
+        state.resource_hash = r.resource_hash
+        state.reconciliation = r
+        self.state_mgr.set_external_resource_state(state)
+
+    def _need_secret_sync(
+        self, r: Reconciliation, state: ExternalResourceState
+    ) -> bool:
+        if r.action == Action.APPLY and state.resource_status == ResourceStatus.CREATED:
+            return True
+        return False
```
hemslo, line 261: avoid unnecessary wrapping of if true return true by directly return

```python
        return r.action == Action.APPLY and state.resource_status == ResourceStatus.CREATED
```

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,330 @@
...
+                ),
+                float(state.reconciliation_errors),
+            )
+
+            if self._resource_needs_reconciliation(reconciliation=r, state=state):
+                self.reconciler.reconcile_resource(reconciliation=r)
+                self._update_state(r, state)
+
+            if to_sync:
+                self._sync_secrets(keys=to_sync)
```
hemslo, line 305: I think you mean to do this after all iterations, right? Wrong indentation?

reconcile/external_resources/manager.py
```diff
@@ -0,0 +1,330 @@
...
+        triggered = set[Reconciliation]()
+
+        for r in reconciliations:
+            state = self.state_mgr.get_external_resource_state(key=r.key)
+            if (
+                r.action == Action.APPLY and state.resource_hash != r.resource_hash
+            ) or r.action == Action.DESTROY:
+                triggered.add(r)
+
+        threaded.run(self.reconciler.reconcile_resource, triggered, thread_pool_size=5)
```
hemslo, line 320: `thread_pool_size` hard code to 5?

reconcile/external_resources/model.py
```diff
@@ -0,0 +1,221 @@
...
+    tf_state_key: str
+
+
+class ExternalResourceProvision(BaseModel):
+    """External resource app-interface attributes. They are not part of the resource but are needed
+    for annotating secrets or other stuff"""
+
+    provision_provider: str  # aws
+    provisioner: str  # ter-int-dev
+    provider: str  # aws-iam-role
```
hemslo, line 203: these comment values require cleanup?

reconcile/external_resources/state.py
```diff
@@ -0,0 +1,209 @@
...
+                    state.reconciliation.module_configuration.reconcile_timeout_minutes
+                )
+            },
+        }
+
+
+class ExternalResourcesStateDynamoDB:
+    def __init__(self, table_name: str, index_name: str) -> None:
+        self.adapter = DynamoDBStateAdapater()
+        self.client = boto3.client("dynamodb", region_name="us-east-1")
```
hemslo, line 147: `region_name` hard code?

reconcile/external_resources/model.py
```diff
@@ -0,0 +1,221 @@
...
+        return hashlib.md5(
+            json.dumps(self.dict(), sort_keys=True).encode("utf-8")
+        ).hexdigest()
+
+    @property
+    def state_path(self) -> str:
+        return f"{self.provision_provider}/{self.provisioner_name}/{self.provider}/{self.identifier}"
+
+
+class ExternalResourcesInventory:
```
janboll, line 59: Looks like a Mapping, perhaps inherit from MutatableMapping class ?

reconcile/external_resources/state.py
lechuk47: I created it as a Setting. PTAL

reconcile/external_resources/model.py
lechuk47: cleaned up.

reconcile/external_resources/manager.py
lechuk47: I don't think we need more than that for now. We can implement a better pool size handling in the future, but as a starting point I think we are good with 5.

reconcile/external_resources/manager.py
lechuk47: Good catch. Fixed.

reconcile/external_resources/manager.py
lechuk47: I changed the logic to something along the lines of what you have described. Thanks for the suggestion. PTAL

reconcile/external_resources/aws.py
```diff
@@ -0,0 +1,133 @@
...
+        #     except SecretNotFound:
+        #         errors.append(
+        #             "Secret referenced in ca_cert attribute not found in the Secrets Engine: "
+        #             + str(resource.data["ca_cert"])
+        #         )
+        #     except Exception as e:
+        #         errors.append("Error finding Secret %s" + str(e))
+
+        # if errors:
+        #     raise ExternalResourceValidationError(errors)
```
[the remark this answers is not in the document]
lechuk47: cleaned up. I leave the methods for future improvements on the Data validation.

reconcile/external_resources/model.py
lechuk47: Yup, that's right. I moved this convenience method to the `ExternalResourcesInventory`class. PTAL.

reconcile/external_resources/aws.py
```diff
@@ -0,0 +1,133 @@
...
+            provision_provider=provision_provider,
+            provisioner_name=provisioner,
+            provider=provider,
+            identifier=identifier,
+        )
+        spec = self.er_inventory.get(key)
+        if not spec:
+            msg = f"Resource spec not found: Provider {provider}, Id: {identifier}"
+            raise FetchResourceError(msg)
+        return spec
```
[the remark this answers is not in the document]
lechuk47: Yup, that's right. I moved this convenience method to the ExternalResourcesInventoryclass. PTAL.

reconcile/external_resources/aws.py
```diff
@@ -0,0 +1,133 @@
...
+#     @staticmethod
+#     def region_from_az(az: str = "") -> str | None:
+#         if len(az) > 1:
+#             return az[:-1]
+
+#         raise ValueError(f"Invalid AZ {az}")
+
+#     @staticmethod
+#     def rds_instance_arn(region: str, account: str, identifier: str) -> str:
+#         return f"arn:aws:rds:{region}:{account}:instance:{identifier}"
```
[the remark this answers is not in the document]
lechuk47: cleaned up

reconcile/external_resources/aws.py
```diff
@@ -0,0 +1,133 @@
...
+    @abstractmethod
+    def validate(self, resource: ExternalResource) -> None: ...
+
+
+class AWSDefaultResourceFactory(AWSResourceFactory):
+    def __init__(
+        self, er_inventory: ExternalResourcesInventory, secrets_reader: SecretReaderBase
+    ):
+        self.er_inventory = er_inventory
+        self.secrets_reader = secrets_reader
```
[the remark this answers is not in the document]
lechuk47: not needed actually. I just removed the __init__ method.

reconcile/external_resources/manager.py
hemslo: we have `@threaded()` in other integrations to make it customizable as cli arg, can be `@threaded(default=5)`.

reconcile/external_resources/state.py
```diff
@@ -0,0 +1,209 @@
...
+    RECONC_MODCONF_VERSION = "reconc.modconf.version"
+    RECONC_MODCONF_DRIFT_MINS = "reconc.modconf.drift_mins"
+    RECONC_MODCONF_TIMEOUT_MINS = "reconc.modconf.timeout_mins"
+    RECONC_INPUT = "reconc.input"
+    RECONC_ACTION = "reconc.action"
+    RECONC_ERRORS = "reconciliation_errors"
+    RESOURCE_KEY = "resource_key"
+    RESOURCE_STATUS = "resource_status"
+    RESOURCE_HASH = "resource_hash"
+    TIMESTAMP = "ts"
```
hemslo, line 65: I'm not clear about the dynamodb schema and index, I checked in dev account, the PK is `resource_key`, and there is a GSI `index_resources` also using the same PK `resource_key`, seems not needed. Can you clarify dynamodb schema and usages?

hemslo: btw, typo `Adapater` -> `Adapter`

lechuk47: You are right. 
I wanted to get some fields of each record to make the initial Desired-Current comparison and I thought an index was required for that, but it is not. I changed the code to use the table with the required fields as projected attributes instead of an index. 

I also changed the structure to use document types and simplified it a bit. PTAL.

reconcile/external_resources/manager.py
lechuk47: Implemented the thread pool size with threaded.

reconcile/external_resources/state.py
```diff
@@ -0,0 +1,247 @@
...
+        DynamoDBStateAdapter.ER_KEY,
+        DynamoDBStateAdapter.TIMESTAMP,
+        DynamoDBStateAdapter.RESOURCE_STATUS,
+        DynamoDBStateAdapter.RECONCILIATION_ERRORS,
+        f"{DynamoDBStateAdapter.RECONC}.{DynamoDBStateAdapter.RECONC_RESOURCE_HASH}",
+    ])
+
+    def __init__(self, table_name: str, index_name: str, region_name: str) -> None:
+        self.adapter = DynamoDBStateAdapter()
+        self.client = boto3.client("dynamodb", region_name=region_name)
```
hemslo, line 183: not sure if this client will create a new session that need to be closed after use, can add the cleanup later once we find it's leaking.

reconcile/external_resources/state.py
```diff
@@ -0,0 +1,247 @@
...
+        partials = {}
+        for item in self.client.scan(
+            TableName=self._table, ProjectionExpression=self.PARTIALS_PROJECTED_VALUES
+        ).get("Items", []):
+            s = self.adapter.deserialize(item)
+            partials[s.key] = s
+        return partials
+
+    def get_all_resource_keys(self) -> list[ExternalResourceKey]:
+        return [k for k in self.partial_resources.keys()]
```
hemslo, line 236: seems the only usage is to use the result as `set`, how about directly return `set(self.partial_resources.keys())`

reconcile/external_resources/state.py
```diff
@@ -0,0 +1,247 @@
...
+    def _get_partial_resources(
+        self,
+    ) -> dict[ExternalResourceKey, ExternalResourceState]:
+        """A Partial Resoure is the minimum resource data reguired
+        to check if a resource has been removed from the configuration.
+        Getting less data from DynamoDb saves money and the logic does not need it.
+        """
+        logging.info("Getting Managed resources from DynamoDb")
+        partials = {}
+        for item in self.client.scan(
```
hemslo, line 228: I think this is the paginator https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb/paginator/Scan.html

hemslo (approved):
LGTM!

reconcile/external_resources/state.py
lechuk47: added

PR #4017, opened 2024-01-24
APPSRE-8660 align with schema

fishi0x01:
Schema changes for dt token provider https://github.com/app-sre/qontract-schemas/pull/576


fishi0x01:
Changed schema scope

PR #4024, opened 2024-01-25
Rename template test integration

janboll:
Rename it to resource template test, cause it's testing resource templates. Template Test would collide with future addition of templating

[APPSRE-8710](https://issues.redhat.com/browse/APPSRE-8710)

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -2628,10 +2628,10 @@ def prometheus_rules_tester(ctx, thread_pool_size, cluster_name):
 
 @integration.command(short_help="Tests templating of resources.")
 @click.pass_context
-def template_tester(ctx):
-    import reconcile.template_tester
+def resource_template_tester(ctx):
+    import reconcile.resource_template_tester
 
-    run_integration(reconcile.template_tester, ctx.obj)
+    run_integration(reconcile.resource_template_tester, ctx.obj)
 
 
 @integration.command(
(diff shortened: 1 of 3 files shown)


reconcile/cli.py
```diff
@@ -2628,10 +2628,10 @@ def prometheus_rules_tester(ctx, thread_pool_size, cluster_name):
 
 @integration.command(short_help="Tests templating of resources.")
 @click.pass_context
-def template_tester(ctx):
-    import reconcile.template_tester
+def resource_template_tester(ctx):
```
hemslo, line 2631: Rename here is a breaking change, need to have additional promotion notes.

reconcile/resource_template_tester.py
```diff
@@ -11,13 +11,13 @@
...
 
-QONTRACT_INTEGRATION = "template-tester"
+QONTRACT_INTEGRATION = "resource-template-tester"
 QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
 
 
 TEMPLATE_TESTS_QUERY = """
 {
-  tests: template_tests_v1 {
+  tests: resource_template_tests_v1 {
```
hemslo, line 20: missing gql schema sync?

janboll: Yes, it's an old-school integration.

reconcile/cli.py
janboll: For fedramp? Will create one

reconcile/resource_template_tester.py
hemslo: I mean there is no `make gql-introspection` change for new schema?

janboll: Oh, there is. it's not required since that validation is not used here. I can add it.

hemslo (approved):
LGTM!

PR #4026, opened 2024-01-25
Add geo_restrictions to cloudflare resources custom ssl certificate

esron:
I'm trying to avoid errors like this one on a terraform apply run:
```
Error: failed to create custom ssl cert: Could not complete the requested operation. Invalid or missing parameters. (2001)
```

I'm not aware of a way to test this with an apply run.


esron:
The bug was in another castle.

PR #4028, opened 2024-01-25
skip vpc-peerings when cluster VPC cannot be discovered

patjlm:
[APPSRE-8792](https://issues.redhat.com/browse/APPSRE-8792)

This allows to define the peerings at cluster creation. We rely here on the fact that `awsapi.get_cluster_vpc_details` returns None when the cluster VPC cannot be discovered. It raises an Exception if something goes wrong with the AWS API call, so we should not end up removing a peering if the discovery fails for infra reasons.

### DIFF
--- a/reconcile/terraform_vpc_peerings.py
+++ b/reconcile/terraform_vpc_peerings.py
@@ -210,9 +210,11 @@ def build_desired_state_single_cluster(
             )
         )
         if requester_vpc_id is None:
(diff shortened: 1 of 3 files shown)


OliviaHY:
LGTM

OliviaHY (approved):

geoberle (approved):

PR #4030, opened 2024-01-26
grant cluster-admin clusterrole to the cluster-admin bot

patjlm:
[APPSRE-8784](https://issues.redhat.com/browse/APPSRE-8784)

That's a miss from https://github.com/app-sre/qontract-reconcile/pull/4016

### DIFF
--- a/reconcile/openshift_cluster_bots.py
+++ b/reconcile/openshift_cluster_bots.py
@@ -124,7 +124,11 @@ def retrieve_token(kubeconfig: str, namespace: str, sa: str) -> str:
 
 
 def create_sa(
-    kubeconfig: str, namespace: str, sa: str, create_namespace: bool = False
+    kubeconfig: str,
+    namespace: str,
+    sa: str,
+    create_namespace: bool = False,
+    cluster_admin: bool = False,
 ) -> str:
     items: list[dict] = []
     if create_namespace:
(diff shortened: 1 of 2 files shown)


reconcile/openshift_cluster_bots.py
```diff
@@ -159,6 +163,26 @@ def create_sa(
             "type": "kubernetes.io/service-account-token",
         },
     ])
+    if cluster_admin:
+        items.append({
+            "apiVersion": "rbac.authorization.k8s.io/v1",
+            "kind": "ClusterRoleBinding",
+            "metadata": {
+                "name": f"{namespace}-{sa}",
```
patjlm, line 171: happy to change this if someone has a better idea...

reconcile/test/test_openshift_cluster_bots.py
```diff
@@ -234,6 +234,6 @@ def test_run_cluster_admin(
     )
     mocks.oc.return_value = {"data": {"token": "mytoken"}}
     ocb.run(**integ_params)
-    assert mocks.oc.call_count == 7
+    assert mocks.oc.call_count == 8
```
geoberle, line 237: :)

reconcile/openshift_cluster_bots.py
geoberle: it's backwards compatible with what we have right now, so should be ok

geoberle (approved):

PR #4032, opened 2024-01-26
Status-board: Handle creations before deletions

lechuk47:
Deletions need to be handled in a reverse priority order.

### DIFF
--- a/reconcile/status_board.py
+++ b/reconcile/status_board.py
@@ -240,7 +240,23 @@ class StatusBoardExporterIntegration(QontractReconcileIntegration):
     def apply_diff(
         dry_run: bool, ocm_api: OCMBaseClient, diff: list[StatusBoardHandler]
     ) -> None:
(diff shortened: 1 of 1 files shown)


reconcile/status_board.py
```diff
@@ -240,7 +240,24 @@ def create_app(app_name: str, product: Product) -> Application:
...
+
+        creations:list[StatusBoardHandler] = []
+        deletions:list[StatusBoardHandler] = []
+
+        for o in diff:
+            match o.action:
+                case "create":
+                    creations.append(o)
+                case "delete":
+                    deletions.append(o)
```
janboll, line 252: That looks unnecessary, since you create the list via comprehension a little lower again

lechuk47: oh sry, I changed to this to just iterate one time over the diffs

janboll (approved):

PR #4035, opened 2024-01-26
[acs-policies] utilize cluster id for scope assignment

dwelch0:
Assigns a cluster's internal ACS id to a scope instead of the cluster name. Utilizing name is causing issue in enforcement of policies against scopes. This was not detected during development due to absence of valid cluster targets within dev acs instance


reconcile/utils/acs/policies.py
```diff
@@ -151,3 +151,13 @@ def list_notifiers(self) -> list[NotifierIdentifiers]:
...
+
+    class ClusterIdentifiers(BaseModel):
+        id: str
+        name: str
+
+    def list_clusters(self) -> list[ClusterIdentifiers]:
+        return [
+            self.ClusterIdentifiers(id=c["id"], name=c["name"])
+            for c in self.generic_request("/v1/clusters", "GET").json()["clusters"]
+        ]
```
hemslo, line 163: 1. need to check if list clusters endpoint returns all clusters in one response or in paginated ones.

2. Is `name` unique cross all clusters? `cluster_name_to_id` will map to wrong id if there are duplicate names.

dwelch0: 1. With our current number of onboarded clusters, the api returns all clusters in one response. The swagger doc is sparse for the `/v1/clusters` endpoint and does not detail any pagination-related behavior.
2. The ACS integrations rely on the [cluster_v1 schema](https://github.com/app-sre/qontract-schemas/blob/main/schemas/openshift/cluster-1.yml) . The [`name` attribute of the cluster schema enforces uniqueness](https://github.com/app-sre/qontract-schemas/blob/main/graphql-schemas/schema.yml#L1019).

hemslo (approved):
LGTM!

PR #4038, opened 2024-01-29
[terraform-resources] add support for aws-iam-role role policy

maorfr:
part of https://issues.redhat.com/browse/APPSRE-8794

depends on https://github.com/app-sre/qontract-schemas/pull/580

this PR adds support to define a policy to be created and attached to a role, instead or in addition to an inline policy.

this change is quite similar to https://github.com/app-sre/qontract-reconcile/blob/94db30819eceadcb38e99d107eb2c1f1f163b321/reconcile/utils/terrascript_aws_client.py#L2328-L2342

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -2469,6 +2469,7 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
                 name=identifier,
                 policy=role_policy,
                 depends_on=self.get_dependencies([role_tf_resource]),
+                tags=common_values["tags"],
             )
             tf_resources.append(tf_aws_iam_policy)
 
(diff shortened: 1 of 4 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -2461,6 +2462,25 @@ def populate_tf_resource_role(self, spec):
         role_tf_resource = aws_iam_role(identifier, **values)
         tf_resources.append(role_tf_resource)
 
+        role_policy = common_values.get("role_policy")
+        if role_policy:
```
hemslo, line 2466: since `role_policy` only used in `if` block, can be

```python
if role_policy := common_values.get("role_policy")
```

hemslo (approved):
LGTM!

reconcile/utils/terrascript_aws_client.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/4038/commits/ebd6a58459e85fe19289b8bdffdedb7e742b2545

PR #4042, opened 2024-01-29
[terraform-resources] aws-iam-role add support for lifecycle

maorfr:
part of https://issues.redhat.com/browse/APPSRE-8799

depends on https://github.com/app-sre/qontract-schemas/pull/583

this PR adds logic to consider a `lifecycle` attribute on an `aws-iam-role`.

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -238,6 +238,7 @@ VARIABLE_KEYS = [
     "subscriptions",
     "records",
     "extra_tags",
+    "lifecycle",
 ]
 
 EMAIL_REGEX = re.compile(r"^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$")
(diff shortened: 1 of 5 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -870,6 +871,22 @@ def get_provider_alias(account: Mapping[str, str]) -> str:
         role_name = awsh.get_id_from_arn(account["assume_role"])
         return f"account-{uid}-{role_name}"
 
+    @staticmethod
+    def get_resource_lifecycle(
+        common_values: dict[str, Any],
+    ) -> Optional[dict[str, Any]]:
+        if lifecycle := common_values.get("lifecycle"):
```
fishi0x01, line 878: would it make sense to use the generated type here to gain some type safety? Like

```
lifecycle := NamespaceTerraformResourceLifecycleV1(common_values.get("lifecycle")**)
...
return lifecycle.dict(by_alias=True)
```

fishi0x01 (approved):

reconcile/utils/terrascript_aws_client.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/4042/commits/471e476a878f168da8ee66ba0556db569a105bb7

PR #4049, opened 2024-02-01
APPSRE-8804 better SAPM metrics

fishi0x01:
- properly calculate parallel open MRs
- add batch size as dimension for open MR counter -> batch sizes are limited, so cardinality will stay low

### DIFF
--- a/reconcile/saas_auto_promotions_manager/merge_request_manager/metrics.py
+++ b/reconcile/saas_auto_promotions_manager/merge_request_manager/metrics.py
@@ -16,6 +16,8 @@ class SAPMOpenedMRsCounter(SAPMBaseMetric, CounterMetric):
     "Counter for the number of opened auto-promotion MRs"
 
     is_batchable: bool
+    # We do not expect batches >10, i.e., cardinality will stay in check here.
+    batch_size: int
 
     @classmethod
     def name(cls) -> str:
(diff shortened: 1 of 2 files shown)


reconcile/saas_auto_promotions_manager/merge_request_manager/metrics.py
```diff
@@ -16,6 +16,8 @@ class SAPMOpenedMRsCounter(SAPMBaseMetric, CounterMetric):
     "Counter for the number of opened auto-promotion MRs"
 
     is_batchable: bool
+    # We do not expect batches >10, i.e., cardinality will stay in check here.
+    batch_size: int
```
geoberle, line 20: do you still need the `is_batchable` label after introducing the `batch_size` one?

fishi0x01: Very good question. I do need it though, because a MR with batch size 1 can be both, batchable and non-batchable. 

The batchable attribute describes whether a specific hash has failed MR checks previously (i.e., was previously removed from a batched MR).

geoberle: ah i see. thanks for the insights

geoberle (approved):

PR #4057, opened 2024-02-05
Add shard to early exit cache and avoid stale cache

hemslo:
Current early exit cache can't handle stale cache, for example in `terraform-resources`:

* MR1: modify resource A
* MR2: revert MR1

Expect MR2 to show some diff, but due to cache hit for MR2 same as state before MR1, pr check will show nothing, prod run will skip until cache expire.

From terraform config point of view, the config after MR2 is the same as before MR1, but the real current state already changed due to MR1, we need to let cache aware of it.

New design is to let `no-dry-run` mode have one cache item per shard, bump cache hash when set cache so old cache won't be reused. `dry-run` mode will additionally check `no-dry-run` latest cache to determine if it's stale.

This change redesign the cache key path:

* dry-run path: `/<integration>/<integration_version>/dry-run(/<shard>)/<cache_source_digest>`
* no-dry-run path: ` /<integration>/<integration_version>/no-dry-run(/<shard>)/latest`

cache item will have 2 new metadata:

* `cache-source-digest`, value is the consistent hash of `key.cache_source` (terraform config for `terraform-resources`).
* `latest-cache-source-digest`, value is the latest `cache-source-digest` in `no-dry-run` during cache lookup time.

`no-dry-run` will use the `cache-source-digest` to determine if the cache is HIT, each integration shard will have at most 1 `latest` cache in `no-dry-run` mode.

`dry-run` mode will additional locate the matching `no-dry-run` latest cache, if `cache-source-digest` is not same as `latest-cache-source-digest` saved in `dry-run` cache, then something changed in current state, we shouldn't use this cache anymore, mark it as stale.

Test this change locally:

```shell
# start localstack
$ make localstack

# set envs

# check head dry run cache miss
$ qontract-cli --config config.toml early-exit-cache head -i a -v b --dry-run -c '{"k": "v"}' -s shard-1
status=<CacheStatus.MISS: 'MISS'> latest_cache_source_digest=''

# check head no dry run cache miss
$ qontract-cli --config config.toml early-exit-cache head -i a -v b --no-dry-run -c '{"k": "v"}' -s shard-1
status=<CacheStatus.MISS: 'MISS'> latest_cache_source_digest=''

# set dry-run cache
$ qontract-cli --config config.toml early-exit-cache set -i a -v b --dry-run -c '{"k": "v"}' -s shard-1 -p {} --ttl 3600

# check dry-run HIT
$ qontract-cli --config config.toml early-exit-cache head -i a -v b --dry-run -c '{"k": "v"}' -s shard-1
status=<CacheStatus.HIT: 'HIT'> latest_cache_source_digest=''

# set no-dry-run cache
$ qontract-cli --config config.toml early-exit-cache set -i a -v b --no-dry-run -c '{"k": "v"}' -s shard-1 -p {} --ttl 3600

# check no-dry-run HIT
$ qontract-cli --config config.toml early-exit-cache head -i a -v b --no-dry-run -c '{"k": "v"}' -s shard-1
status=<CacheStatus.HIT: 'HIT'> latest_cache_source_digest='d3f1f34d36332ded8721468ff57cdaf988f91a68837ef59a3a16aaf2b03e0443'

# now dry-run cache should be stale since new no-dry-run cache updated
$ qontract-cli --config config.toml early-exit-cache head -i a -v b --dry-run -c '{"k": "v"}' -s shard-1
status=<CacheStatus.STALE: 'STALE'> latest_cache_source_digest='d3f1f34d36332ded8721468ff57cdaf988f91a68837ef59a3a16aaf2b03e0443'

# simulate dry-run run again to set cache
$ qontract-cli --config config.toml early-exit-cache set -i a -v b --dry-run -c '{"k": "v"}' -s shard-1 -p {} --ttl 3600

# check again, should HIT
$ qontract-cli --config config.toml early-exit-cache head -i a -v b --dry-run -c '{"k": "v"}' -s shard-1
status=<CacheStatus.HIT: 'HIT'> latest_cache_source_digest='d3f1f34d36332ded8721468ff57cdaf988f91a68837ef59a3a16aaf2b03e0443'

# set a new no-dry-run cache to simulate changes in MR1
$ qontract-cli --config config.toml early-exit-cache set -i a -v b --no-dry-run -c '{"k": "v2"}' -s shard-1 -p {} --ttl 3600

# now check both old caches again
$ qontract-cli --config config.toml early-exit-cache head -i a -v b --dry-run -c '{"k": "v"}' -s shard-1
status=<CacheStatus.STALE: 'STALE'> latest_cache_source_digest='29ec39a1e5d350154aa53702f1b79240a5dbeece159a67ecae4fbd4c25510412'

$ qontract-cli --config config.toml early-exit-cache head -i a -v b --no-dry-run -c '{"k": "v"}' -s shard-1
status=<CacheStatus.MISS: 'MISS'> latest_cache_source_digest='29ec39a1e5d350154aa53702f1b79240a5dbeece159a67ecae4fbd4c25510412'
```

### DIFF
--- a/reconcile/utils/early_exit_cache.py
+++ b/reconcile/utils/early_exit_cache.py
@@ -13,6 +13,7 @@ from reconcile.utils.state import State, init_state
 STATE_INTEGRATION = "early-exit-cache"
 EXPIRE_AT_METADATA_KEY = "expire-at"
 CACHE_SOURCE_DIGEST_METADATA_KEY = "cache-source-digest"
+LATEST_CACHE_SOURCE_DIGEST_METADATA_KEY = "latest-cache-source-digest"
 
 
 class CacheKey(BaseModel):
@@ -84,6 +85,11 @@ class CacheStatus(Enum):
     STALE = "STALE"  # latest cache source digest is different, which means there are new applied changes
 
 
+class CacheHeadResult(BaseModel):
+    status: CacheStatus
+    latest_cache_source_digest: str
+
+
 class EarlyExitCache:
     def __init__(self, state: State):
         self.state = state
@@ -109,11 +115,27 @@ class EarlyExitCache:
         value = self.state.get(str(key))
         return CacheValue.parse_obj(value)
 
-    def set(self, key: CacheKey, value: CacheValue, ttl_seconds: int) -> None:
+    def set(
+        self,
+        key: CacheKey,
+        value: CacheValue,
+        ttl_seconds: int,
+        latest_cache_source_digest: str,
+    ) -> None:
+        """
+        Set the cache value for the given key.
+
+        :param key: cache key
+        :param value: cache value
+        :param ttl_seconds: time to live in seconds
+        :param latest_cache_source_digest: latest cache source digest, used to check stale for dry run cache
+        :return: None
+        """
         expire_at = datetime.now(tz=UTC) + timedelta(seconds=ttl_seconds)
         metadata = {
             EXPIRE_AT_METADATA_KEY: str(int(expire_at.timestamp())),
             CACHE_SOURCE_DIGEST_METADATA_KEY: key.cache_source_digest,
+            LATEST_CACHE_SOURCE_DIGEST_METADATA_KEY: latest_cache_source_digest,
         }
         self.state.add(
             str(key),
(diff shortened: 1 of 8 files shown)


reconcile/utils/early_exit_cache.py
```diff
@@ -64,24 +115,112 @@ def get(self, key: CacheKey) -> CacheValue:
...
-        return CacheStatus.HIT if now < expire_at else CacheStatus.EXPIRED
+        return now >= expire_at
+
+    def _head_dry_run_status(
+        self,
+        key: CacheKey,
+        latest_cache_source_digest: str,
+    ) -> CacheStatus:
+        exists, metadata = self.state.head(key.dry_run_path())
+        if not exists:
```
janboll, line 182: looks like match case?

reconcile/utils/early_exit_cache.py
```diff
@@ -64,24 +115,112 @@ def get(self, key: CacheKey) -> CacheValue:
...
+            return CacheStatus.MISS
+        if self._is_expired(metadata):
+            return CacheStatus.EXPIRED
+        if self._is_stale(metadata, latest_cache_source_digest):
+            return CacheStatus.STALE
+        return CacheStatus.HIT
+
+    def _head_dry_run(self, key: CacheKey) -> CacheHeadResult:
+        _, latest_metadata = self.state.head(key.no_dry_run_path())
```
janboll, line 191: Which shard will `key.no_dry_run_path()` use in dry-run?

reconcile/utils/early_exit_cache.py
janboll: TIL, thanks

reconcile/utils/early_exit_cache.py
janboll: But, how does that work? In MR checks we do not ruse sharding, or are you going to change this?

hemslo: I plan to ask why it's disabled, it was running with shard before.

When it's not used, which means there is no matching latest cache in no dry run, so stale check will always pass (not stale).

When both pr check and real run are running with the same shard model, the stale check can work as expected.

janboll: Wouldn't it be safer in this case to not do an early exit?

hemslo: Then it means the cache is not usable for pr check at all. So far the stale case is an edge case, most changes just forward changes, do revert and hit an old cache within 1 hour is rare.

janboll (approved):

reconcile/utils/early_exit_cache.py
hemslo: Then it will run without shard, means no matching `no-dry-run` cache, stale check is disabled. If we can override TTL (extra-args) in shard mode to be a long one, no shard mode to be a small one, the risk of showing staled output in PR check should be reduced.

hemslo: I'll merge this to test run, set pr check cache TTL to 30 mins to avoid edge cases, and can cover most rebase cases.

PR #4062, opened 2024-02-06
Draft: Add templating tester integration

janboll:
This adds an integration to test templates for data files.


reconcile/templating/test.py
```diff
@@ -0,0 +1,92 @@
...
+                    )
+                if test.expected_to_render is not None:
+                    if test.expected_to_render != r.should_render():
+                        self.diffs.append(
+                            TemplateDiff(
+                                template=template.name,
+                                test=test.name,
+                                diff="Template should not render, but it did",
+                            )
+                        )
```
hemslo, line 74: diff description is `Template should not render, but it did`, if condition `test.expected_to_render` is `False`, and `r.should_render()` is `True`, it will also meet.

I guess `expected_to_render` is a nullable bool, None means don't check, right? Then the exact condition is

```python
                if (
                    test.expected_to_render is not None
                    and not test.expected_to_render
                    and r.should_render()
                ):
```

and to avoid call `r.should_render()` in later check, can extract var for it.

reconcile/test/fixtures/templating/full.yaml
```diff
@@ -0,0 +1,11 @@
...
+  name: full
+
+  targetPath: /{{bar}}/foo.yml
+
+  template: |
+    foo: {{ bar }}
+
+  templateTest: []
+expected: |
+  foo: bar
```
hemslo, line 11: Missing new line at `EOF`

reconcile/test/fixtures/templating/patch_overwrite_nested.yaml
```diff
@@ -0,0 +1,28 @@
...
+      buz: baz
+      r: bar
+    oof: rab
+
+expected: |
+  version:
+    bar: baz
+    foo:
+      buz: baz
+    oof: rab
```
hemslo, line 28: Missing new line at `EOF`

reconcile/templating/rendering.py
```diff
@@ -0,0 +1,106 @@
...
+
+            updated = False
+            for data in matched_value:
+                if data.get(self.template.patch.identifier) == dta_identifier:
+                    data.update(data_to_add)
+                    updated = True
+                    continue
+
+            if not updated:
+                matched_value.append(data_to_add)
```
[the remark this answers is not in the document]
janboll: That is a very neat expression. thanks!

reconcile/templating/test.py
janboll: The description is wrong. It should be: `f"Expected to render mismatch, got: {should_render}, expected: {test.expected_to_render}"`

reconcile/templating/test.py
chassing, line 1: Please rename the file, `test.py` is very misleading. `integration.py` would fit

reconcile/templating/rendering.py
```diff
@@ -0,0 +1,106 @@
...
+    def _jinja2_render_kwargs(self) -> dict[str, Any]:
+        return {**self._get_variables(), "current": self._get_current()}
+
+    def _render_template(self, template: str) -> str:
+        return self.jinja_env.from_string(template).render(
+            **self._jinja2_render_kwargs()
+        )
+
+    @abstractmethod
+    def get_output(self) -> str:
```
chassing, line 36: All methods in this class could be properties.

reconcile/templating/rendering.py
```diff
@@ -0,0 +1,106 @@
...
+
+class TemplateData(BaseModel):
+    variables: dict[str, Any]
+    current: Optional[dict[str, Any]]
+
+
+def create_renderer(template: TemplateV1, data: TemplateData) -> Renderer:
+    if template.patch:
+        return PatchRenderer(template=template, data=data)
+    else:
```
chassing, line 105: Unnecessary `else`

reconcile/templating/test.py
```diff
@@ -0,0 +1,97 @@
...
+                    ),
+                )
+                if test.expected_target_path:
+                    self.diff_result(
+                        template.name,
+                        test.name,
+                        r.get_target_path().strip(),
+                        test.expected_target_path.strip(),
+                    )
+                should_render = r.should_render()
```
chassing, line 68: See above; it would be easier to read if `should_render` would be a property.

```
if (
    test.expected_to_render is not None
     and test.expected_to_render != r.should_render
):
```

reconcile/templating/test.py
```diff
@@ -0,0 +1,97 @@
...
+                if (
```
chassing, line 69: can be written as
```
if test.expected_to_render != should_render:
```

reconcile/templating/test.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class TemplateDiff(BaseModel):
+    template: str
+    test: str
+    diff: str
+
+
+class TemplatingTestIntegration(QontractReconcileIntegration):
+    diffs: list[TemplateDiff]
```
chassing, line 31: No need to declare this variable here.
```
self.diffs: list[TemplateDiff] = []
```
in line 47 would be sufficient

reconcile/templating/rendering.py
```diff
@@ -0,0 +1,106 @@
...
+from pydantic import BaseModel
+from ruamel import yaml
+
+from reconcile.gql_definitions.templating.templates import TemplateV1
+from reconcile.utils.jsonpath import parse_jsonpath
+
+
+class Renderer(ABC):
+    template: TemplateV1
+    data: "TemplateData"
```
chassing, line 14: No need to declare the variable here.

reconcile/templating/rendering.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def __init__(self, template: TemplateV1, data: "TemplateData"):
```
chassing, line 16: Please move the `TemplateData` class above this class and you can specify `data: TemplateData` without quotes

reconcile/templating/rendering.py
```diff
@@ -0,0 +1,106 @@
...
+        return self._render_template(self.template.condition or "True") == "True"
+
+
+class FullRenderer(Renderer):
+    def get_output(self) -> str:
+        return self._render_template(self.template.template)
+
+
+class PatchRenderer(Renderer):
+    def get_output(self) -> str:
```
chassing, line 52: What does this method? There are so many operations, but at the end you don't return any of the variables? Please also add a `__doc__` string.

reconcile/test/templating/__init__.py
chassing, line 1: the test folder doesn't need a `__init__.py`

reconcile/test/templating/conftest.py
```diff
@@ -0,0 +1,24 @@
+from typing import Any, Callable
```
chassing, line 1: please import `Callable` from `collections.abc`

reconcile/test/templating/test_full_rendering.py
```diff
@@ -0,0 +1,19 @@
+from typing import Callable
```
chassing, line 1: please import `Callable` from `collections.abc`

reconcile/test/fixtures/templating/__init__.py
chassing, line 1: this file isn't needed at all

reconcile/templating/test.py
janboll: Hm, it's the test integration, `integration.py` will be the actual integration reconciling templates. What about `validator`?

chassing: Sounds good. 🙇

reconcile/templating/test.py
janboll: No, it will raise an exception. I moved it to `__init__`

reconcile/templating/test.py
janboll: Hm, none means do not run this test (check should render). 
If `test.expected_to_render` is none, the test does not specify it, then I do not want to test for it, cause the test might not be specific for checking the condition.

reconcile/templating/rendering.py
janboll: Right, `get_` is a bad name here. fixed

chassing (approved):
Few nits. LGTM

reconcile/test/templating/test_patch_rendering.py
```diff
@@ -0,0 +1,47 @@
+from typing import Callable
```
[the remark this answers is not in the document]
janboll: :facepalm:

PR #4063, opened 2024-02-06
terrascript secret_reader support

fishi0x01:
Lets remove settings as much as possible. Vast majority of cases we want a secret_reader


chassing (approved):
nice!

PR #4064, opened 2024-02-06
added code to supress SecretNotFound via allow_not_found flag

mehfuz:


reconcile/openshift_resources_base.py
```diff
@@ -295,7 +295,8 @@ def lookup_secret(path, key, version=None, tvars=None, settings=None):
         secret_reader = SecretReader(settings)
         return secret_reader.read(secret)
     except Exception as e:
-        raise FetchSecretError(e)
+        if not allow_not_found:
+            raise FetchSecretError(e)
```
patjlm, line 299: what do we return if `allow_not_found` ?

PR #4071, opened 2024-02-08
added new flag to lookup_secret to manage secret not found exception

mehfuz:
Currently in order to migrate aws rds resources having `grafana_datasource: true` we need a 2 step process:-
1. Comment `grafana_datasource: true` in first MR while moving data to new namespace file and merge.
2. Uncomment `grafana_datasource: true` in next MRs.
 The reason for this is because integration tries to find secrets which are not created until MR is merged and fails with `SecretNotFound` exception.

This PR introduces a flag `allow_not_found` to manage  `SecretNotFound` exception to tackle situations like above.

Tested on local by creating new namespace file on a new cluster for status-board service.
Updated the `observability/grafana/grafana-datasources.secret.yaml` as follows

```jinja
{{%- set db_name = vault(vault_path, 'db.name',allow_not_found=True) %}}
            {{%- set db_host = vault(vault_path, 'db.host',allow_not_found=True) %}}
            {{%- set db_port = vault(vault_path, 'db.port',allow_not_found=True) %}}
            {{%- set db_passowrd = vault(vault_path, 'db.password',allow_not_found=True) %}}
            {{%- if db_name is not none and db_host is not none and db_port is not none and db_password is not none %}}
            {
                "name": "{{{ tfr.identifier }}}",
                "orgId": 1,
                "type": "postgres",
                "database": "{{db_name}}",
                "url": "{{db_host}}:{{db_port}}",
                "version": 1,
                "jsonData": {
                    "sslmode": "disable"
                },
                "access": "proxy",
                "editable": false,
                "user": "{{db_user}}",
                "secureJsonData": {
                    "password": "{{db_password}}"
                }
            },
             {{%- endif %}}
``` 
The code works without raising any exception


reconcile/openshift_resources_base.py
```diff
@@ -308,6 +314,8 @@ def lookup_secret(
             secret_reader = SecretReader(settings)
         return secret_reader.read(secret)
     except Exception as e:
+        if allow_not_found:
+            return "SECRET_NOT_FOUND"
```
patjlm, line 318: This means this exact value needs to be tested in the template to add or not the datasource?

mehfuz: @patjlm I am testing out this approach..

patjlm: I guess returning `None` would be good, right ?

reconcile/openshift_resources_base.py
```diff
@@ -308,6 +314,8 @@ def lookup_secret(
             secret_reader = SecretReader(settings)
         return secret_reader.read(secret)
     except Exception as e:
```
janboll, line 316: Please instead catch SecretNotFound before catch generic exception and then do the if match/raise forward thing.

patjlm: Indeed. For example if the connection to vault fails for whatever reason, you do not want to return a value, but you need to raise the exception.

reconcile/openshift_resources_base.py
```diff
@@ -289,7 +289,13 @@ def __init__(self, msg):
 
 @retry()
 def lookup_secret(
-    path, key, version=None, tvars=None, settings=None, secret_reader=None
+    path,
+    key,
+    allow_not_found=False,
```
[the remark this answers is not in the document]
patjlm: the order in the colling code looks good :wink:

janboll (approved):

PR #4075, opened 2024-02-08
APPSRE-9796: allow enable s3 logging on bucket

jfchevrette:
This adds the ability to set the following in an s3 bucket's (and s3-related resources) defaults file

```yaml
s3_bucket_logging:
  target_bucket_name: "foo" # Optional
  target_prefix: "logs/" # Optional
```

If `s3_bucket_logging` is defined the code does the following:
- Looks up the target S3 bucket using terraform's data resource
- Creates an s3_bucket_logging resource

It is left to the user to create an S3 bucket with appropriate permissions and lifecycle rules to log into. This allows turning off logging while retaining the bucket because it is defined separately

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -1931,63 +1931,18 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
             # common_values['cors_rules'] is a list of cors_rules
             values["cors_rule"] = cors_rules
         # S3 Bucket Logging
-        s3_bucket_logging = common_values.get("s3_bucket_logging", {})
-        if s3_bucket_logging.get("enabled", False):
+        s3_bucket_logging = common_values.get("s3_bucket_logging")
+        if s3_bucket_logging:
             target_bucket_name = s3_bucket_logging.get("target_bucket_name")
-            target_bucket_identifier = f"{identifier}-logging"
-            if not target_bucket_name:
-                target_bucket_name = f"{identifier}-logs"
+            logging_identifier = f"{identifier}-logging"
 
-            # Logging target bucket
-            target_bucket_values = {
-                "bucket": target_bucket_name,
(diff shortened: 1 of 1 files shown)


janboll (approved):

rporres:
I'm not very familiar to how s3 logging works so I may be completely wrong here, but here's my doubt:  would be interesting to set lifecycle rules to the bucket that you create? Because if we want delete the logging bucket once we don't use it anymore, the deletion will fail as the bucket is not empty...

We can do it in a subsequent MR if this is urgent.

jfchevrette:
I have updated and simplified the PR such that this only creates the logging config pointing to an existing S3 bucket. The config of the target bucket is left to the user via the existing s3 bucket resource/provider configs

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1926,6 +1930,27 @@ def populate_tf_resource_s3(self, spec):
...
+        s3_bucket_logging = common_values.get("s3_bucket_logging")
+        if s3_bucket_logging:
+            target_bucket_name = s3_bucket_logging.get("target_bucket_name")
+            logging_identifier = f"{identifier}-logging"
+
+            # Lookup target S3 bucket by name
+            self.add_resource(
+                account,
+                data.aws_s3_bucket(target_bucket_name, bucket=target_bucket_name),
+            )
```
rporres, line 1943: why do we need to add this resource here if it has been defined elsewhere? Is it to check that it exists?

jfchevrette: It is a recommended practices that I picked up over time which is to use a data source to lookup and use a resource which is not part of the terraform config. In this case the s3_bucket_logging resource will fail if provided with a bucket that does not exist so this does not provide a lot of value at the increase cost of additional AWS API calls. This was used more in code which I since removed so I'll remove the data source as well

PR #4076, opened 2024-02-08
[APPSRE-7832] change saas images to set

BumbleFeng:


patjlm:
Can you explain why this was needed please?

PR #4078, opened 2024-02-08
[openshift-rolebindings] add kind to roleRef

maorfr:
another fix on top of #4067
```
unsupported role reference kind: ""
```

since we use admin/edit/view and other ClusterRoles, it should be ok to hard code `ClusterRole`, at least for now.


jfchevrette:
LGTM

jfchevrette (approved):

PR #4080, opened 2024-02-09
native deepdiff 6.7.x path parsing and tf-res sharded pr check fix

geoberle:
the deepdiff 6.7.x library yields different path formats for the found diffs. our custom parse logic was not able to deal with that but deepdiff now also offers a native way to parse paths.

this PR uses the deepdiff path parse logic instead of our own. additionally a bug in the produced early exit desired state was fixed. the state can only contain dicts and not dataclasses.

### DIFF
--- a/reconcile/change_owners/diff.py
+++ b/reconcile/change_owners/diff.py
@@ -1,6 +1,5 @@
 import copy
 import json
-import re
 from dataclasses import dataclass
 from enum import Enum
 from functools import reduce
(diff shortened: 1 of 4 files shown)


reconcile/change_owners/diff.py
```diff
@@ -252,17 +254,16 @@ def deepdiff_path_to_jsonpath(deep_diff_path: str) -> jsonpath_ng.JSONPath:
...
+                return jsonpath_ng.Fields(element)
 
-        if "." in element:
-            return jsonpath_ng.Fields(f"'{element}'")
-        return jsonpath_ng.Fields(element)
-
-    path_parts = [
-        build_jsonpath_part(p) for p in DEEP_DIFF_RE.findall(deep_diff_path[4:])
-    ]
+    path_parts = [build_jsonpath_part(p) for p in parse_path(deep_diff_path)]
```
hemslo, line 266: Is `DEEP_DIFF_RE` still used anywhere else?

hemslo (approved):
LGTM!

reconcile/change_owners/diff.py
geoberle: good point, cleaned up

PR #4081, opened 2024-02-09
Add validation setps

janboll:
- This adds yamllint to the template validation integration. This requires moving the yamllint configuration file into the resources directory and also updating the hack scripts.
- This also adds validation using `qontract-valdation`, for this the following is required: https://github.com/app-sre/qontract-validator/pull/59

[APPSRE-8796](https://issues.redhat.com/browse/APPSRE-8796)

### DIFF
--- a/reconcile/templating/validator.py
+++ b/reconcile/templating/validator.py
@@ -1,9 +1,12 @@
+import json
 import logging
 from difflib import context_diff
 from typing import Callable, Optional
 
 from pydantic import BaseModel
 from ruamel import yaml
+from validator import validate_bundle # type: ignore
+from validator.bundle import load_bundle # type: ignore
 from yamllint import linter  # type: ignore
 from yamllint.config import YamlLintConfig  # type: ignore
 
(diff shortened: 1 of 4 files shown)


reconcile/templating/validator.py
```diff
@@ -30,10 +32,30 @@ class TemplateDiff(BaseModel):
...
 
+class TemplateError(BaseModel):
+    template: str
+    error: str
+    lineno: Optional[int] = None
+
+
+def lint_yaml(template_name: str, yaml_to_lint: str) -> list[TemplateError]:
+    # Possibly move this path to the templating configuration schema in the future
+    resource = gql.get_api().get_resource("/yamllint/yamllint.yml")
```
patjlm, line 43: i think this should be an integration option indeed. And if not set, it would disable the linting check. This will ease FedRamp setup, not requiring a file somewhere in resources. WDYT?

janboll: Indeed, I found this the hardest part of this change :smile: 

My idea was to:

- move `.yamllint` from app-interface root into resources folder
- update the hack script to read the file from there
- then read configuration using GQL

In the design doc, there is a specification about templating configuration. I'd move this setting there in the future. WDYT

janboll:
On hold. will be refactored to a CLI command

PR #4086, opened 2024-02-12
[terraform-resources] restructure desired state for early exit

geoberle:
same as we did for openshift-resources in https://github.com/app-sre/qontract-reconcile/pull/3784

this simplifies the changed shard detection by having the shard as top level element in the desired state dict.

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -544,26 +544,22 @@ def early_exit_desired_state(*args: Any, **kwargs: Any) -> dict[str, Any]:
         for spec in get_external_resource_specs(
             ns_info.dict(by_alias=True), provision_provider=PROVIDER_AWS
         ):
-            spec_resources: dict[str, str] = {}
+            resource_paths = [
+                spec.resource.get("defaults"),
+                spec.resource.get("parameter_group"),
+            ] + [
+                spec_item.get("defaults")
+                for spec_item in spec.resource.get("specs") or []
+            ]
+            resources = {
+                resource["path"]: resource["sha256sum"]
+                for path in resource_paths
+                if path and (resource := gqlapi.get_resource(path))
+            }
             spec_state = {
                 "spec": asdict(spec),
-                "resources": spec_resources,
+                "resources": resources,
             }
-
-            def register_resource(resource: Optional[dict[str, Any]]) -> None:
-                if resource:
-                    spec_resources[resource["path"]] = resource["sha256sum"]
-
-            defaults = spec.resource.get("defaults")
-            if defaults:
-                register_resource(gqlapi.get_resource(defaults))
-            parameter_group = spec.resource.get("parameter_group")
-            if parameter_group:
-                register_resource(gqlapi.get_resource(parameter_group))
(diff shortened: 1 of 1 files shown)


reconcile/terraform_resources.py
```diff
@@ -533,49 +533,52 @@ def runner(
...
+    }
+    for ns_info in get_tf_namespaces():
         for spec in get_external_resource_specs(
             ns_info.dict(by_alias=True), provision_provider=PROVIDER_AWS
         ):
+            spec_resources: dict[str, str] = {}
+            spec_state = {
+                "spec": asdict(spec),
+                "resources": spec_resources,
+            }
```
hemslo, line 551: Since we are refactoring here, how about do it using comprehensions to avoid mutation? like this:

```python
            resource_paths = [
                spec.resource.get("defaults"),
                spec.resource.get("parameter_group"),
            ] + [
                spec_item.get("defaults")
                for spec_item in spec.resource.get("specs") or []
            ]
            resources = {
                resource["path"]: resource["sha256sum"]
                for path in resource_paths
                if path and (resource := gqlapi.get_resource(path))
            }
            spec_state = {
                "spec": asdict(spec),
                "resources": resources,
            }
            spec_id = f"{spec.cluster_name}/{spec.namespace_name}/{spec.provisioner_name}/{spec.provider}/{spec.identifier}"
            state_for_accounts[spec.provisioner_name][spec_id] = spec_state
```

reconcile/terraform_resources.py
```diff
@@ -533,49 +533,52 @@ def runner(
...
             if parameter_group:
-                register_resource(spec, gqlapi.get_resource(parameter_group))
+                register_resource(gqlapi.get_resource(parameter_group))
             for spec_item in spec.resource.get("specs") or []:
                 defaults = spec_item.get("defaults")
                 if defaults:
-                    register_resource(spec, gqlapi.get_resource(defaults))
+                    register_resource(gqlapi.get_resource(defaults))
+            spec_id = f"{spec.cluster_name}/{spec.namespace_name}/{spec.provisioner_name}/{spec.provider}/{spec.identifier}"
+            state_for_accounts[spec.provisioner_name][spec_id] = spec_state
```
hemslo, line 568: maybe need to add a test here, I think you mean to populate `state_for_accounts[spec.provisioner_name]["specs"][spec_id] = spec_state`, here extract function and use in comprehension would make it cleaner too.

reconcile/terraform_resources.py
geoberle: fixed

reconcile/terraform_resources.py
geoberle: fixed by your comprehension proposal

chassing (approved):

reconcile/terraform_resources.py
hemslo: @geoberle it's not fixed, `specs` field in `state_for_accounts` is always empty, not a big issue as no conflict between `spec_id` and `meta`. Either remove `specs` from `state_for_accounts` or put specs under `specs`.

PR #4090, opened 2024-02-12
Add logging in case ClowdApp and Deployment are used together

janboll:
If Clowdapp is used, it will manage resources of type Deployment. This integration should not manage these resources then, hence users should remove Deployment from managed resource typpes

[APPSRE-9794](https://issues.redhat.com/browse/APPSRE-9794)


cubismod (approved):
LGTM

maorfr (approved):

PR #4093, opened 2024-02-13
WIP [version gate approver] creates version gate agreements when conditions are met

geoberle:
not all version gates can be agreed to just like that, e.g. the STS gates require some actions on the clusters account.

this new integration checks for different version gate conditions and conditionally creates version gate agreements.

some version gate agreements were created by AUS integrations previously, but this has now been handed over to the version gate approver integration. AUS will not start an upgrade when it finds unacked gates for a cluster.


reconcile/utils/rosa/rosa_cli.py
```diff
@@ -0,0 +1,193 @@
...
+            ttl_seconds_after_finished=3600,
+            template=V1PodTemplateSpec(
+                metadata=V1ObjectMeta(
+                    annotations=self.annotations(), labels=self.labels()
+                ),
+                spec=V1PodSpec(
+                    containers=[
+                        V1Container(
+                            name="outputs",
+                            image="registry.ci.openshift.org/ci/rosa-aws-cli:latest",
```
jfchevrette, line 147: should we have the flexibility of specifying a version here?

reconcile/utils/rosa/rosa_cli.py
```diff
@@ -0,0 +1,193 @@
...
+        return {
+            "qontract.rosa.account_name": self.account_name,
+            "qontract.rosa.cluster_name": self.cluster_name,
+            "qontract.rosa.org_id": self.org_id,
+        }
+
+    def job_spec(self) -> V1JobSpec:
+        return V1JobSpec(
+            backoff_limit=1,
+            ttl_seconds_after_finished=3600,
```
jfchevrette, line 138: we may want to keep the job longer so we have ample time to inspect logs in case of failures

reconcile/utils/rosa/rosa_cli.py
```diff
@@ -0,0 +1,193 @@
...
+                        V1Container(
+                            name="outputs",
+                            image="registry.ci.openshift.org/ci/rosa-aws-cli:latest",
+                            command=["/bin/bash", "-c"],
+                            args=[self.cmd],
+                            image_pull_policy="Always",
+                            env=[
+                                V1EnvVar(
+                                    name="AWS_ACCESS_KEY_ID",
+                                    value=self.aws_credentials.access_key_id,
```
jfchevrette, line 154: can we have the creds as a secret and change these to a secretRef? As it is a user with view privs on the namespace would be able to get those secrets from the job/pod definitions

jfchevrette: for automated cleanup we can set an `ownerReference` on the secret such that it gets garbage collected when the job is deleted after `ttl_seconds_after_finished`

geoberle: the general idea was to use temporary credentials exclusively, so these AWS creds and also the OCM creds are invalid after 15 minutes (15 min is the minimum valididity for AWS STS tmp creds and 15 min is also the validity of an OCM access token).

i considered secrets but it implied managing the secret resource as well, which makes the job controller more complex again. the benefit of having only a job resource revolves around the `ttl_seconds_after_finished` property and its ability to clean up "old" jobs. 

i'm not saying managing a secret is unreasonable efforts. but i tend to believe that temporary credentials that expire after 15 minutes combined with a namespace where only we have access should be sufficient protection.

i'm not going to die on that hill though. if you have serious concerns security wise, we can discuss a design that revolves around secrets.

reconcile/utils/rosa/rosa_cli.py
geoberle: the job logs are transfered to stdout of the integration after either a timeout, a success or an error. but i wonder if the job and its pods would be interesting to keep for longer for any other debuging steps else but reading logs 🤔

reconcile/utils/rosa/rosa_cli.py
geoberle: absolutely

reconcile/utils/rosa/rosa_cli.py
OliviaHY: That was what I wanted to do too, using secretRef, however considering the STS temporary creds are only valid for short period of time which largely limited the risk. I think it's okay to use env.

reconcile/utils/rosa/session.py
```diff
@@ -0,0 +1,182 @@
...
+from reconcile.utils.rosa.rosa_cli import (
+    LogHandle,
+    RosaCliException,
+    RosaCliResult,
+    RosaJob,
+)
+from reconcile.utils.secret_reader import SecretReaderBase
+
+
+class RosaSession:
```
OliviaHY, line 27: Is this class intend to be the starting point of all ROSA cli run going forward? Including my use case, the longer script that involves:
1. creation of ocm-role, account-roles, user-role
2.cluster creation, 
3. operator-roles and oidc config creation

They will be in a new method in this class?

reconcile/utils/rosa/rosa_cli.py
jfchevrette: I still think the idea of using `ownerReferences` is interesting and would solve this security concern but I understand this adds complexity where it may not be needed due to the ephemeral nature of the credentials. I'm not going to die on that hill either :)

reconcile/utils/rosa/rosa_cli.py
jfchevrette: SGTM. Lets leave it as is and we can optimize later if needed

reconcile/utils/rosa/session.py
geoberle: yes that is the idea. such new methods would then call the `cli_execute` function and pass in their command. as the doc strings state, multiple commands would need to be wrapped into  `bash -c "cmd 1 ; cmd2 ; cmd3"`

reconcile/utils/rosa/rosa_cli.py
geoberle: hmm. would `ownerReference` work?  what would be the order of resource creation?

a) create the job that references the secret that does not exist yet, then create the secret with the job in the `ownerReference`. would the job pod wait for the secret to appear?

b) create the secret, without an owner reference, create the job, patch the owner ref in the secret (does that work)?

geoberle: i gave it a try and a) seems to work. i will implement it like this. thanks for your feedback on this aspect

reconcile/utils/rosa/rosa_cli.py
geoberle: done

geoberle:
since a lot of refactoring and pulling out individual PRs happened, it is better to close this one in favour of this one - https://github.com/app-sre/qontract-reconcile/pull/4103

PR #4095, opened 2024-02-14
botocore library conflict

geoberle:
the botocore library dependency is declared twice: in setup.py (with a newer version) and in requirements-test.txt (with an older version).

this results in a lenghty venv build time

```
INFO: pip is looking at multiple versions of boto3 to determine which version is compatible with other requirements. This could take a while.
Collecting boto3>=1.9.201 (from moto~=2.2->-r requirements/requirements-test.txt (line 11))
  Downloading boto3-1.34.40-py3-none-any.whl.metadata (6.6 kB)
  Downloading boto3-1.34.39-py3-none-any.whl.metadata (6.6 kB)
  Downloading boto3-1.34.38-py3-none-any.whl.metadata (6.6 kB)
  Downloading boto3-1.34.37-py3-none-any.whl.metadata (6.6 kB)
  Downloading boto3-1.34.36-py3-none-any.whl.metadata (6.6 kB)
  Downloading boto3-1.34.35-py3-none-any.whl.metadata (6.6 kB)
  Downloading boto3-1.34.34-py3-none-any.whl.metadata (6.6 kB)
...
```

using the same versions solves the issue.


setup.py
```diff
@@ -30,6 +30,7 @@
         "semver~=2.13",
         "boto3==1.34.15",
         "botocore==1.34.15",
+        "mypy-boto3-s3==1.34.14",
```
geoberle, line 33: sadly there is no mypy-boto3-s3 version for 1.34.15 yet

jfchevrette (approved):

patjlm:
:heart: Thanks!

PR #4098, opened 2024-02-14
[job-controller] support secret data for jobs

geoberle:
the `K8sJob` class can define secret data it relies on and the job controller will manage the lifecycle of a kubernetes Secret based on that data.

```python

class MyJob(K8sJob):

   ...

   def secret_data(self) -> dict[str: str]:
     return {
       "super": "secret"
     }

```

the resulting secret has an `ownerReference` towards the job, which implies that the secret gets cleaned up when the job gets cleaned up.

### DIFF
--- a/reconcile/utils/jobcontroller/controller.py
+++ b/reconcile/utils/jobcontroller/controller.py
@@ -2,7 +2,13 @@ import logging
 import time
 from typing import Optional, Protocol
 
-from kubernetes.client import ApiClient, V1Job  # type: ignore[attr-defined]
+from kubernetes.client import (  # type: ignore[attr-defined]
+    ApiClient,
+    V1Job,
+    V1ObjectMeta,
+    V1OwnerReference,
+    V1Secret,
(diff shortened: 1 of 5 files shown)


reconcile/utils/jobcontroller/controller.py
```diff
@@ -242,19 +248,69 @@ def enqueue_job(
...
+    def _lookup_job_uid(self, job_name: str) -> Optional[str]:
+        job_resource = self.oc.get(
+            self.namespace, "Job", job_name, allow_not_found=True
+        )
+        if not job_resource:
+            return None
+        return job_resource.get("metadata", {}).get("uid")
+
+    def build_secret(self, job: K8sJob) -> Optional[V1Secret]:
+        secret_data = job.secret_data()
```
OliviaHY, line 260: Should we label those Job owned Secret with something so they can be easily looked up?

OliviaHY (approved):

PR #4099, opened 2024-02-15
introduce custom messages (+use in ocm-internal-notifications)

maorfr:
part of https://issues.redhat.com/browse/APPSRE-9820

depends on https://github.com/app-sre/qontract-schemas/pull/591

this change will cause ocm-internal-notifications to print custom instructions based on app-interface-settings.


patjlm (approved):

reconcile/ocm_internal_notifications/integration.py
```diff
@@ -106,6 +109,9 @@ def run(self, dry_run: bool) -> None:
 
             if not dry_run and slack_user_ids:
                 users = " ".join([f"<@{uid}>" for uid in slack_user_ids])
+                instructions = get_app_interface_custom_message(
+                    "ocm-internal-notifications-instructions"
+                )
```
patjlm, line 114: It might be nice to have a default value, even `""` in this case, if this is not defined (maybe along with a warning?)

PR #4100, opened 2024-02-15
Add templating integration

janboll:
This implements the template renderer integration. It includes rendering to local files. MR management will not be included yet.

Requires: https://github.com/app-sre/qontract-schemas/pull/592
Solves: [APPSRE-8796](https://issues.redhat.com/browse/APPSRE-8796)


reconcile/templating/renderer.py
```diff
@@ -0,0 +1,172 @@
...
+            return f.read()
+
+
+def unpack_static_variables(
+    collection_variables: TemplateCollectionVariablesV1,
+) -> dict:
+    variables = {}
+    if collection_variables.static:
+        variables = collection_variables.static
+    return variables
```
hemslo, line 80: avoid var reassign by `return collection_variables.static or {}`

reconcile/templating/renderer.py
```diff
@@ -0,0 +1,172 @@
...
+) -> dict:
+    variables: dict[str, Any] = {}
+    if collection_variables.dynamic:
+        gql_no_validation = init_from_config(validate_schemas=False)
+        for dv in collection_variables.dynamic or []:
+            variables[dv.name] = []
+            result = gql_no_validation.query(dv.query)
+            for key in result or {}:
+                variables[dv.name].append(result[key])
+    return variables
```
hemslo, line 94: There is no duplicate `dv.name` for one collection, right? Then it can be simplified with guard exit and comprehension

```python
    if not collection_variables.dynamic:
        return {}

    gql_no_validation = init_from_config(validate_schemas=False)

    return {
        dv.name: list((gql_no_validation.query(dv.query) or {}).values())
        for dv in collection_variables.dynamic or []
    }
```

reconcile/templating/rendering.py
```diff
@@ -14,8 +13,28 @@ class TemplateData(BaseModel):
...
 
 
+class HasTemplatePatch(Protocol):
+    path: str
+    identifier: Optional[str]
+
+    def dict(self) -> dict[str, str]: ...
+
+
+class HasTemplate(Protocol):
```
hemslo, line 23: Any special meaning of naming in `Has`?

reconcile/templating/renderer.py
```diff
@@ -0,0 +1,172 @@
...
+        target_path = r.render_target_path()
+        current_str: Optional[str] = None
+        try:
+            current_str = persistence.read(
+                target_path,
+            )
+            y = yaml.YAML()
+
+            r.data.current = y.load(current_str)
+        except FileNotFoundError:
```
hemslo, line 133: `FileNotFoundError` is only for `persistence.read`, right? Should limit try block only include expected exceptions.

Then no need to specify `current_str` outside, as it will early raise, won't be None when used later.

reconcile/cli.py
```diff
@@ -1869,6 +1869,30 @@ def template_validator(ctx):
...
 
+@integration.command(short_help="Render datafile templates in app-interface.")
+@click.option(
+    "--app-interface-data-path",
+    help="Path to app-interface dictory, used to write output and calculate diff in dry-run.",
+    required=False,
+    envvar="APP_INTERFACE_DATA_PATH",
+)
+@click.pass_context
+def template_renderer(ctx, app_interface_data_path):
```
hemslo, line 1880: Is this a long running integration? I thought it's a cli command in `qontract_cli.py`, this `--app-interface-data-path` won't work in real integration run container, right?

reconcile/templating/rendering.py
janboll: Yes: https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#naming-conventions

reconcile/cli.py
janboll: It will be one in the next iteration. I wanted to make the PRs shorter and easier to review. Think MR management (which will make this a long-running integration) would have made this way more complicated.

reconcile/templating/renderer.py
janboll: Right, makes sense especially, since persistence would also be remote gitlab in the future. I'll change `read` to return `Optional[str]` and catch that exception.

reconcile/templating/rendering.py
hemslo: My understanding of that convention means `HasTemplate` has a field called `template`.

reconcile/templating/renderer.py
```diff
@@ -0,0 +1,179 @@
...
+
+class TemplateRendererIntegrationParams(PydanticRunParams):
+    app_interface_data_path: Optional[str]
+
+
+def join_path(base: str, sub: str) -> str:
+    # not using os.path.sep, since app-interface relies on unix paths
+    if sub.startswith(APP_INTERFACE_PATH_SEPERATOR):
+        return os.path.join(base, sub[1:])
+    return os.path.join(base, sub)
```
chassing, line 106: Because of your comment above, `os.path.join` is using `os.path.sep` too :)

reconcile/templating/renderer.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class TemplateRendererIntegrationParams(PydanticRunParams):
+    app_interface_data_path: Optional[str]
+
+
+def join_path(base: str, sub: str) -> str:
+    # not using os.path.sep, since app-interface relies on unix paths
+    if sub.startswith(APP_INTERFACE_PATH_SEPERATOR):
+        return os.path.join(base, sub[1:])
```
chassing, line 105: `sub.lstrip(APP_INTERFACE_PATH_SEPERATOR)` should work as well

reconcile/templating/renderer.py
```diff
@@ -0,0 +1,179 @@
...
+    @property
+    def name(self) -> str:
+        return QONTRACT_INTEGRATION
+
+    def run(self, dry_run: bool) -> None:
+        outputs: list[TemplateOutput] = []
+        gql_no_validation = init_from_config(validate_schemas=False)
+        persistence = LocalFilePersistence(self.params.app_interface_data_path)
+
+        ruaml_instance = yaml.YAML()
```
chassing, line 157: Please do me a favor and move this part into a utils method. There are already 3 different places where we initialize `rumal` like this.

reconcile/templating/rendering.py
janboll: Not sure, found other examples. My first draft did not have that prefix

janboll: I'll just remove it if you agree

chassing: Please don't name it with the `Has` prefix. It's more than just a `template` protocol. `Template` would be fine.

reconcile/templating/renderer.py
janboll: Thought about that. But, what if someone runs it on Windows? Our paths are POSIX, I even thought about raising an exception if this was run on windows.

reconcile/templating/renderer.py
janboll: Yes, that's a good idea. I'll do it in a new PR

reconcile/templating/renderer.py
chassing: I think no one can run qontract-reconcile on Windows, especially as we rely on other binaries like terraform, git, ... Even on Windows, you can run the qontract-reconcile container.

reconcile/templating/renderer.py
```diff
@@ -0,0 +1,180 @@
...
+
+class TemplateRendererIntegrationParams(PydanticRunParams):
+    app_interface_data_path: Optional[str]
+
+
+def join_path(base: str, sub: str) -> str:
+    # not using os.path.sep, since app-interface relies on unix paths
+    if sub.startswith(APP_INTERFACE_PATH_SEPERATOR):
+        return os.path.join(base, sub.lstrip(APP_INTERFACE_PATH_SEPERATOR))
+    return os.path.join(base, sub)
```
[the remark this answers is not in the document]
chassing: 😸

chassing (approved):

reconcile/templating/renderer.py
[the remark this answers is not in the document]
janboll: Of course... Happy monday :sweat_smile:

hemslo (approved):
LGTM!

PR #4105, opened 2024-02-19
Create ruaml instance from utils

janboll:
Create ruaml objects in a single place. Makes configuration simpler and standardizes it for various use cases.


chassing (approved):
❤️

hemslo (approved):
Nice extraction!

PR #4107, opened 2024-02-19
Add validation cli

janboll:
This adds a CLI which helps develop templates locally. It can be invoked as such:
```bash
template_validation.py --template-path /templating/templates/cert-manager/cert-manager.yml --ai-path $HOME/app-interface
```

The cli will do:

- Evaluate the provided template by parsing it and checking it against configured tests
- Ensuring the YAML is valid by running `yamllint`

It is a more convenient way to develop new templates, focused on creating tests first,

### DIFF
--- a/reconcile/templating/validator.py
+++ b/reconcile/templating/validator.py
@@ -42,7 +42,9 @@ class TemplateValidatorIntegration(QontractReconcileIntegration):
 
     @staticmethod
     def _create_renderer(
-        template: TemplateV1, template_test: TemplateTestV1, secret_reader: SecretReaderBase
+        template: TemplateV1,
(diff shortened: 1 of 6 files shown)


reconcile/templating/validator.py
```diff
@@ -31,70 +37,98 @@ class TemplateDiff(BaseModel):
 
 
 class TemplateValidatorIntegration(QontractReconcileIntegration):
-    def __init__(self, params: RunParamsTypeVar) -> None:
+    def __init__(self, params: PydanticRunParams) -> None:
         super().__init__(params)
```
geoberle, line 41: i don't think you need this initializer

geoberle (approved):

PR #4112, opened 2024-02-20
Enhance template rendering

janboll:
Add the openshift_namespace_exists method, which can be used by templates to check if a namespace exists on a cluster.

Make it possible to mock these lambdas by passing in `_template_mock` variables.

### DIFF
--- a/reconcile/utils/jinja2/utils.py
+++ b/reconcile/utils/jinja2/utils.py
@@ -20,6 +20,7 @@ from reconcile.utils.jinja2.filters import (
     urlescape,
     urlunescape,
 )
+from reconcile.utils.oc import OCCli
 from reconcile.utils.oc_map import init_oc_map_from_clusters
 from reconcile.utils.secret_reader import SecretNotFound, SecretReader, SecretReaderBase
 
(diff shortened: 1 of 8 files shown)


reconcile/utils/jinja2/utils.py
```diff
@@ -156,7 +177,13 @@ def process_jinja2_template(
         "hash_list": hash_list,
         "query": lookup_graphql_query_results,
         "url": url_makes_sense,
+        "openshift_namespace_exists": lambda c, n: openshift_namespace_exists(
+            cluster_name=c, namespace=n, secret_reader=secret_reader
+        ),
```
hemslo, line 182: Make `openshift_namespace_exists` avaialbe in all jinja2 template will cause desired state rely on current cluster state, can't cache easily or do early exit. Can we just implement it as a check in desired state from gql result only?

janboll: I agree that state is an issue, check other functions, there are more that have this issue.

hemslo: yes, but they are more like data fetcher, less likely to conflict with reconcile condition, but fetch info from cluster can have some unexpected result.

For example, the namespace deleted or some API issues, the expected render content will flap.

We had similar issues for some openshift tokens, use current data to determine desired data lead to unstable result.

janboll: Understand your point better now thanks. I added this with cert-manager setup in mind. the namespace is created by the operator, and only after creation added by app-interface.

I'll remove the method for now, perhaps we can find a way around.

hemslo (approved):
LGTM!

PR #4113, opened 2024-02-20
Build a new fedramp image including a different oc binary version

cubismod:
[APPSRE-9849](https://issues.redhat.com/browse/APPSRE-9849)

Recently QR was updated to include a newer version of `oc`. This broke things in FedRamp due to some FIPs related issues. Therefore, this MR creates a new image located at `quay.io/app-sre/qontract-reconcile-fedramp`. The only difference between the two is that a different version of `oc` is pre-bundled into the image.

I chose this approach of switching the version in the QR Dockerfile as the alternative of updating the [container-images Dockerfile](https://github.com/app-sre/container-images/blob/master/qontract-reconcile-base/Dockerfile) would have required 4 images (QR base image, QR base image + different version of oc, QR, QR fedramp). This approach only requires three images.

A-I MR to create the new Quay repo: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/97565


dockerfiles/Dockerfile
```diff
@@ -64,3 +64,14 @@ RUN microdnf upgrade -y && \
...
+FROM prod-image as fedramp-prod-image
+ENV FEDRAMP_OC_VERSION=4.12.46
+
+# FedRamp is typically on older versions of OpenShift clusters than commercial so we build a new version of
+# the prod image just replacing oc and kubectl with our own version
+RUN curl -sfL https://mirror.openshift.com/pub/openshift-v4/clients/ocp/${FEDRAMP_OC_VERSION}/openshift-client-linux.tar.gz \
+    -o oc.tar.gz && \
+    tar -zvxf oc.tar.gz && \
+    mv oc kubectl /usr/local/bin && \
+    rm oc.tar.gz README.md
```
hemslo, line 77: Due to overlay2 fs, 2 versions of `oc` will exist in different layers, `4.12.46` will be the file used eventually, but total image size will include binaries for both versions.

cubismod: Darn yeah that's right I forgot about that.

cubismod:
Closing this as it got a bit too complex. Going to just revert the MR updating the OC version.

PR #4118, opened 2024-02-21
Fixing dynamic variables

janboll:
Fix for templating renderer integration

Query results are now unpacked correctly, before a unneccessary array was return, this is now removed, by matching the name of the query and the name of the query in query itself.


reconcile/templating/renderer.py
```diff
@@ -87,13 +87,17 @@ def unpack_static_variables(
...
-    return {
-        dv.name: list(gql.query(dv.query).values())
-        for dv in collection_variables.dynamic or []
+    return_dict: dict[str, list] = {
+        dv.name: [] for dv in collection_variables.dynamic or []
     }
+    for dv in collection_variables.dynamic or []:
+        for rname, rdata in list(gql.query(dv.query).items()):
+            if rname == dv.name:
+                return_dict[dv.name] = rdata
```
hemslo, line 99: I'm confused about this logic, will this override the same key?

hemslo: and allow a different dynamic query result to override others?

janboll: It's possible to write a query like this:
```
{
  cluster_v1 {}
  users_v1 {}
}
```
The result will have maps in "data", containing results for both objects. My intent here is to only map only one. To make it more explicit.

janboll: My first idea was to simply map all results.

hemslo: oh, seems a complex usage when one query return a dict with multiple keys, how about like this, we don't do any mapping, just put the result under the `name`, keep the raw result data as dict?

```python
    {
        dv.name: gql.query(dv.query) or {} for dv in collection_variables.dynamic or []
    }
```

janboll: That's where it came from :sweat_smile:  It's very unintuitive to use.
I propose the following, a query like this:
```
{
  cluster_v1 {}
  users_v1 {}
}
```
Will create a mapping like this:
variables:
   cluster_v1: ...
   users_v1: ....

hemslo: what if I declare 2 queries, both contains

```gql
{
  cluster_v1 {}
  users_v1 {}
}
```

with different conditions, what should be the final `variables`?

janboll: It will be the content of the last query. The initial one would need to be changed to include both contents. I think that's a good limitation to reduce number of queries.

hemslo: a little confusing to use, as if I only declare one `name` called `cluster_info`, then I will get `cluster_info: []` as the result because no matching names between `cluster_info` and gql query result.

hemslo: Or you mean we always only cherry pick the gql result based on `name`:

```python
    {
        dv.name: (gql.query(dv.query) or {}).get(dv.name, [])
        for dv in collection_variables.dynamic or []
    }
```

janboll: image

janboll: You assign the query a name in the collection:
```
  dynamic:
  - name: hosting_cluster
    query: |
      {
        hosting_cluster: clusters_v1(filter: {name: "abc"}) { }
      }
  - name: rosa_clusters
    query: |
      {
        rosa_clusters: clusters_v1(filter: {name: "foo"}) {
      }

```

hemslo: I mean `name` is not used, like this:

```yaml
  dynamic:
  - name: abc
    query: |
      {
        hosting_cluster: clusters_v1(filter: {name: "abc"}) { }
      }
  - name: def
    query: |
      {
        rosa_clusters: clusters_v1(filter: {name: "foo"}) {
      }
  static:
    hosting_cluster: [xxx]
    rosa_clusters: [yyy]
```

what do you expect final variables to be?

janboll: Yeah, the name is redundant now. Think it can be safely removed.

janboll: forgot to answer your question, this should be the result:
```yaml
variables:
   hosting_cluster: [xxx]
   rosa_clusters: [yyy]
```

hemslo: oh, then you need to clarify precedence about different variable names, like static over dynamic, last one wins. I feel to keep the same structure as config maybe easier to understand, example:

```yaml
  dynamic:
  - name: abc
    query: |
      {
        clusters: clusters_v1(filter: {name: "abc"}) { }
      }
  - name: def
    query: |
      {
        clusters: clusters_v1(filter: {name: "foo"}) {
      }
  static:
    clusters: [xxx]
```

resolve to 

```yaml
variables:
  dynamic:
    abc:
      clusters: [abc]
    def:
      clusters: [def]
  static:
    clusters: [xxx]
```

so we don't need to worry about name conflict, can just use `dynamic.abc.clusters` inside the template.

janboll: Hm, well, I initially found this too complicated.
But perhaps it makes templates a little easier to understand. I'll add it like that!

hemslo (approved):
LGTM!

PR #4134, opened 2024-02-28
[prometheus-rules-tester] reduce resources to prometheus-rule only

maorfr:
slack thread: https://redhat-internal.slack.com/archives/CCRND57FW/p1709123601855539

with this change, we not only reduce the run time of the integration, but also eliminate runs with `early_exit` in case no prometheus tests have changed.


geoberle (approved):

rporres:
late to the party, but this makes total sense, indeed <3

PR #4137, opened 2024-02-29
increase timeout for rosa cluster creation job

geoberle:
increase the timeout for the ROSA HCP cluster creation job to 5 minutes up from 60 seconds. this is defined statically for now to solve an issue. an upcoming PR will make this configurable

### DIFF
--- a/reconcile/utils/rosa/session.py
+++ b/reconcile/utils/rosa/session.py
@@ -88,6 +88,8 @@ class RosaSession:
         cmd: str,
         annotations: Optional[dict[str, str]] = None,
         image: Optional[str] = None,
+        check_interval_seconds: Optional[int] = 5,
+        timeout_seconds: int = 60,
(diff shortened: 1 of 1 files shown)


ninabauer (approved):

patjlm:
if the execution can last longer there is more risk the pod gets deleted for whatever reason. How do we ensure idempotency, restart on failure/reschedule, ... ?

geoberle:
> if the execution can last longer there is more risk the pod gets deleted for whatever reason. How do we ensure idempotency, restart on failure/reschedule, ... ?

if a pod vanishes while the controller is waiting for it, the execution will fail raise an error. usually that means the integration fails and retries next time. for cluster creation that is something that is supported (also for the other rosa cli jobs we have).

a pod vanishing can happen after 2 seconds already. so writing jobs in a way that they are idempotent is crucial regardless of the timeout we set

PR #4140, opened 2024-03-01
[ocm-machine-pools] nodepool subnet specdrift

geoberle:
specifying the subnet IDs of HCP nodepools in the clusters file upfront the cluster installation does not work well with how the rosa cli creates nodepools during cluster creation. there is no way to influence nodepool subnet placement at this stage. so it makes sense to not specify the subnets of HCP nodepools in app-interface during cluster creation

BUT

we can patch them in the cluster file once they are known!

this PR opens an MR to add the subnet IDs to the cluster file if it is not defined yet.

bonus: this PR will also ignore clusters that are not ready yet for machine pool management (`spec.id` not set yet). this avoids`ocm-machine-pool` reconcile errors during cluster provisioning.

reminder:
* when a nodepool does not exist yet in OCM, app-interface is the source of truth for the subnet
* once a nodepool exists in OCM, the subnet ID can't be changed anymore and OCM acts as the source of truth (the existing code reflects that by not allowing a subnet ID change on existing nodepools)

demo MR: https://gitlab.cee.redhat.com/goberlec/app-interface/-/merge_requests/12/diffs

part of https://issues.redhat.com/browse/APPSRE-9883

### DIFF
--- a/reconcile/ocm_machine_pools.py
+++ b/reconcile/ocm_machine_pools.py
@@ -247,7 +247,7 @@ class NodePool(AbstractPool):
     subnet: Optional[str]
 
     def delete(self, ocm: OCM) -> None:
-        ocm.delete_node_pool(self.cluster, self.dict(by_alias=True))
+        ocm.delete_node_pool(self.cluster, self.id)
 
     def create(self, ocm: OCM) -> None:
         spec = self.dict(by_alias=True)
--- a/reconcile/ocm_machine_pools.py
+++ b/reconcile/ocm_machine_pools.py
@@ -262,7 +262,9 @@ class NodePool(AbstractPool):
             or self.taints != pool.taints
             or self.labels != pool.labels
             or self.aws_node_pool.instance_type != pool.instance_type
-            or self.subnet != pool.subnet
+            # if the nodepool in app-interface does not define a subnet explicitely
+            # we don't consider it a diff. we don't manage it
+            or (pool.subnet is not None and self.subnet != pool.subnet)
             or self._has_diff_autoscale(pool)
         )
 
@@ -373,7 +375,7 @@ def _classify_cluster_type(cluster: ClusterV1) -> ClusterType:
             raise ValueError(f"unknown cluster type for cluster {cluster.name}")
 
 
-def fetch_current_state_for_cluster(cluster, ocm):
+def fetch_current_state_for_cluster(cluster: ClusterV1, ocm: OCM) -> list[AbstractPool]:
     cluster_type = _classify_cluster_type(cluster)
     if cluster_type == ClusterType.ROSA_HCP:
         return [
--- a/reconcile/ocm_machine_pools.py
+++ b/reconcile/ocm_machine_pools.py
@@ -547,34 +547,31 @@ def calculate_diff(
     return diffs, errors
 
 
+def _drift_updates(
+    current_pools: list[AbstractPool],
+    desired: DesiredMachinePool,
+) -> list[ClusterMachinePoolV1]:
+    current_machine_pools = {pool.id: pool for pool in current_pools}
+    return [
+        spec
+        for pool in desired.pools
+        if (current_pool := current_machine_pools.get(pool.q_id))
+        and (spec := current_pool.get_pool_spec_to_update(pool)) is not None
+    ]
+
+
 def calculate_spec_drift(
     current_state: Mapping[str, list[AbstractPool]],
     desired_state: Mapping[str, DesiredMachinePool],
-) -> list[tuple[str, ClusterMachinePoolV1]]:
+) -> dict[str, list[ClusterMachinePoolV1]]:
     """
     Finds spec drifts between OCM and app-interface and returns a list of them.
     """
-    current_machine_pools = {
-        (cluster_name, machine_pool.id): machine_pool
-        for cluster_name, machine_pools in current_state.items()
-        for machine_pool in machine_pools
+    return {
+        desired_state[k].cluster_path: updates
+        for k in current_state.keys() & desired_state.keys()
+        if (updates := _drift_updates(current_state[k], desired_state[k]))
     }
-
-    desired_machine_pools = {
-        (desired.cluster_name, desired_machine_pool.q_id): desired_machine_pool
-        for desired in desired_state.values()
-        for desired_machine_pool in desired.pools
-    }
-    spec_drift_updates: list[tuple[str, ClusterMachinePoolV1]] = []
-    for pool_key, current_pool in current_machine_pools.items():
-        desired_pool = desired_machine_pools.get(pool_key)
-        if not desired_pool:
-            continue
-        update = current_pool.get_pool_spec_to_update(desired_pool)
-        if update:
-            cluster_path = desired_state[pool_key[0]].cluster_path
-            spec_drift_updates.append((cluster_path, update))
-    return spec_drift_updates
 
 
 def update_ocm(dry_run: bool, diffs: Iterable[PoolHandler], ocm_map: OCMMap) -> None:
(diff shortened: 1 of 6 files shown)


reconcile/ocm_machine_pools.py
```diff
@@ -262,7 +262,9 @@ def has_diff(self, pool: ClusterMachinePoolV1) -> bool:
             or self.taints != pool.taints
             or self.labels != pool.labels
             or self.aws_node_pool.instance_type != pool.instance_type
-            or self.subnet != pool.subnet
+            # if the nodepool in app-interface does not define a subnet explicitely
+            # we don't consider it a diff. we don't manage it
+            or (pool.subnet is not None and self.subnet != pool.subnet)
```
patjlm, line 267: so this means we'll have non matching data between app-interface and the reality ?

geoberle: it was not really non-matching data. not specifying them in app-interface means "i don't manage this".

i added the promised phase 2 to this MR in the meantime to open an app-interface MR to patch the missing subnets.

maorfr (approved):

reconcile/ocm_machine_pools.py
```diff
@@ -504,19 +546,73 @@ def calculate_diff(
...
+    spec_drift_updates: list[tuple[str, ClusterMachinePoolV1]] = []
+    for pool_key, current_pool in current_machine_pools.items():
+        desired_pool = desired_machine_pools.get(pool_key)
+        if not desired_pool:
+            continue
+        update = current_pool.get_pool_spec_to_update(desired_pool)
+        if update:
+            cluster_path = desired_state[pool_key[0]].cluster_path
+            spec_drift_updates.append((cluster_path, update))
+    return spec_drift_updates
```
hemslo, line 576: Since the return type `list[tuple[str, ClusterMachinePoolV1]]` requires additional transform to `dict[str, list[ClusterMachinePoolV1]]` anyway, how about create it directly here to simplify? Like this:

```python
def _drift_updates(
    current_pools: list[AbstractPool],
    desired: DesiredMachinePool,
) -> list[ClusterMachinePoolV1]:
    current_machine_pools = {pool.id: pool for pool in current_pools}
    return [
        spec
        for pool in desired.pools
        if (current_pool := current_machine_pools.get(pool.q_id))
        and (spec := current_pool.get_pool_spec_to_update(pool)) is not None
    ]

def calculate_spec_drift(
    current_state: Mapping[str, list[AbstractPool]],
    desired_state: Mapping[str, DesiredMachinePool],
) -> list[tuple[str, ClusterMachinePoolV1]]:
    return {
        desired_state[k].cluster_name: updates
        for k in current_state.keys() & desired_state.keys()
        if (updates := _drift_updates(current_state[k], desired_state[k]))
    }
```

reconcile/utils/mr/cluster_machine_pool_updates.py
```diff
@@ -0,0 +1,74 @@
...
+from ruamel.yaml import YAML
+
+from reconcile.change_owners.decision import DecisionCommand
+from reconcile.utils.gitlab_api import GitLabApi
+from reconcile.utils.mr.base import MergeRequestBase
+
+yaml = YAML()
+yaml.explicit_start = True
+# Lets prevent line wraps
+yaml.width = 4096
```
hemslo, line 13: can reuse `create_ruamel_instance` for duplicate config

reconcile/utils/mr/cluster_machine_pool_updates.py
```diff
@@ -0,0 +1,74 @@
...
+            cluster_fs_path = f"data{cluster_path}"
+            if not machine_pool_updates:
+                continue
+
+            raw_file = gitlab_cli.project.files.get(
+                file_path=cluster_fs_path, ref=gitlab_cli.main_branch
+            )
+            content = yaml.load(raw_file.decode())
+            if "machinePools" not in content:
+                self.cancel("machinePools missing. Nothing to do.")
```
hemslo, line 53: better to include `cluster_fs_path` for easy debug

reconcile/utils/mr/cluster_machine_pool_updates.py
```diff
@@ -0,0 +1,74 @@
...
+                yaml.dump(content, stream)
+                new_content = stream.getvalue()
+
+            msg = f"update cluster {content['name']} spec fields"
+            gitlab_cli.update_file(
+                branch_name=self.branch,
+                file_path=cluster_fs_path,
+                commit_message=msg,
+                content=new_content,
+            )
```
hemslo, line 71: `gitlab_cli.update_file` can still be called even there are no changes for the current file, maybe better to split the whole `process` into 3 steps:

* preprocess to get list of changed files to update
* if no changed files, call `self.cancel`
* otherwise call `gitlab_cli.update_file` for each changed file

reconcile/ocm_machine_pools.py
geoberle: we can and we should specify subnets for day2 node pool operations. but at installtime we can't influence it and we end up with a set of pools where ROSA decides how to spread them accross the available private subnets.

we can improve on this behaviour once there is the ability to create HCP clusters without node pools at install time. becaues then we wait until the cluster is ready and `ocm-machine-pools` can create the pools in the explicitely defined subnets.

reconcile/utils/mr/cluster_machine_pool_updates.py
geoberle: done

reconcile/ocm_machine_pools.py
```diff
@@ -273,6 +292,26 @@ def invalid_diff(self, pool: ClusterMachinePoolV1) -> Optional[str]:
...
+        in such a case this method returns a clone of the current pool definition
+        in app-interface with the subnet updated to the one in OCM. if no update
+        is required, None is returned
+
+        how can this happen?
+        * cluster gets created without specified subnets and ROSA CLI picks the subnet
+          assignment for the nodepools
+        * nodepools have been recreated on OCM side without app-interface involvement
+        """
+        if pool.subnet is None and self.subnet != pool.subnet:
```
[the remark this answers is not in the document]
geoberle: it is `or`
good spot. ty

reconcile/utils/mr/cluster_machine_pool_updates.py
geoberle: ack. i will add some tests first

reconcile/utils/mr/cluster_machine_pool_updates.py
geoberle: done

reconcile/ocm_machine_pools.py
geoberle: done. thank you

reconcile/utils/mr/cluster_machine_pool_updates.py
hemslo: oh, so we still create branch and update file, abort when there is no change in final submit. Then why we still detect `changes` in `process` and `cancel` it?

reconcile/ocm_machine_pools.py
hemslo: > > Does that mean we should never apply subnet from app-interface to ocm? When app-interface pool subnet is not None and we have a different subnet value in current pool, this condition will be true, and show as diff, is this expected?
> 
> once a node pool has been created, its subnet ID can't be changed anymore. at this point OCM becomes the source of truth for it. we implemented the notion of an `invalid_diff` to highlight the problem, so it is expected that `has_diff` shows a diff in subnet between a-i and OCM.

So for the use case of `nodepools have been recreated on OCM side without app-interface involvement`, this diff will show in `has_diff` but will fail `invalid_diff` check, the integration will fail until patch mr merged, is this expected?

geoberle: yes, it still makes us aware and we need to intervene. we should not do things manually in OCM.

reconcile/utils/mr/cluster_machine_pool_updates.py
geoberle: the branch is created before we enter the `process` function.

hemslo: yes, I mean since the responsibility of branch `create`/`cancel` are managed by base method, should we just log and return in `process` instead of calling `cancel`?

hemslo: oh, found other subclass also call `cancel` during `process`, ok for the current design then.

hemslo (approved):
LGTM!

reconcile/utils/mr/cluster_machine_pool_updates.py
geoberle: this entire set of classes needs a refactor for sure

PR #4148, opened 2024-03-04
[openshift-namespaces] fail on openshift namespaces on dry-run

maorfr:
we currently do not fail in dry-run: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/97765
leading to reverts: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/98529

this is known - we can not create `openshift-` namespaces. let's fail in dry-run.

the logic to raise an exception is consistent and will be handled in https://github.com/app-sre/qontract-reconcile/pull/4148/files#diff-3faaaf915e42878fb84d5d26fff3791258ff2e290185c31da412e4ab0f5104beR129-R141


chassing (approved):
❤️

PR #4152, opened 2024-03-05
[terraform-tgw-attachments] allow traffic from the TGW on an HCPs VPC endpoint

geoberle:
similar to regular peerings, we want traffic via TGW attachments to reach the API of a private HCP via its VPC endpoint

follows up on https://github.com/app-sre/qontract-reconcile/pull/3969

part of https://issues.redhat.com/browse/APPSRE-9883

depends on https://github.com/app-sre/qontract-schemas/pull/598

### DIFF
--- a/reconcile/gql_definitions/common/clusters_with_peering.gql
+++ b/reconcile/gql_definitions/common/clusters_with_peering.gql
@@ -24,6 +24,8 @@ query ClustersWithPeering {
 
     spec {
       region
+      hypershift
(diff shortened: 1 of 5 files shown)


rporres (approved):

geoberle:
for transparency: i added a schema extension to control this behaviour after the LGTMs https://github.com/app-sre/qontract-schemas/pull/598

@rporres @maorfr

PR #4154, opened 2024-03-05
Terrafor vpc creation appsre 8743

esron:
Closes [APPSRE-8743](https://issues.redhat.com/browse/APPSRE-8743)

To test this you may need this changes in your local `app-interface` repo:
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/103083/diffs

Feel free to suggest ways to better test this code that doesn't involve creating MRs in `app-interface`.


reconcile/test/test_terraform_vpc_resources.py
```diff
@@ -0,0 +1,127 @@
...
+from reconcile.status import ExitCodes
+from reconcile.terraform_vpc_resources import (
+    TerraformVpcResources,
+    TerraformVpcResourcesParams,
+)
+
+
+@pytest.fixture
+def query_data():
+    return VPCResourcesAWSAccountsQueryData(
```
fishi0x01, line 23: `gql_class_factory` might come in handy https://github.com/app-sre/qontract-reconcile/blob/b14feca7affff061773cf6cf712da2de3fec488a/reconcile/test/conftest.py#L124

It helps you creating (nested) data classes and fills the gaps with default values. I.e., you only need to set the values that you are interested in for your test, greatly reducing the boilerplate. 

Example on how this can be used https://github.com/app-sre/qontract-reconcile/blob/b14feca7affff061773cf6cf712da2de3fec488a/reconcile/test/test_terraform_resources.py#L123-L135

Note, that `gql_class_factory` is in our global `conftest.py`, i.e., it is already available for every test you are writing.

reconcile/test/test_terraform_vpc_resources.py
```diff
@@ -0,0 +1,127 @@
...
+                        ),
+                    ]
+                ),
+            ),
+        ]
+    )
+
+
+@pytest.fixture
+def mock_gql(mocker):
```
fishi0x01, line 52: We have a global fixture for the gql api - might come in handy https://github.com/app-sre/qontract-reconcile/blob/b14feca7affff061773cf6cf712da2de3fec488a/reconcile/test/conftest.py#L146

Examples of how this can be used https://github.com/app-sre/qontract-reconcile/blob/b14feca7affff061773cf6cf712da2de3fec488a/reconcile/test/test_typed_queries/test_jira_settings.py#L37-L49

reconcile/test/test_terraform_vpc_resources.py
```diff
@@ -0,0 +1,127 @@
...
+        vault=True
+    )
+    return mocked_app_interfafe_vault_settings
+
+
+@pytest.fixture
+def mock_create_secret_reader(mocker):
+    return mocker.patch(
+        "reconcile.terraform_vpc_resources.create_secret_reader", autospec=True
+    )
```
fishi0x01, line 75: nit: Personal recommendation is to use dependency injection pattern instead of mock patching. I.e., 

```python
from unittest.mock import create_autospec

mocked_secret_reader = create_autospec(spec=SecretReaderBase)
integration = Integration(secret_reader=mocked_secret_reader)

mocked_secret_reader.read_secret.assert_called_once_with(...)
```

This gives you a clean mock object that will throw errors if it is used in a non-supported way. You can find this pattern in many integrations. That being said, we must resort to mock patching for older components that were not built with dependency injection in mind.

reconcile/test/test_terraform_vpc_resources.py
esron: I'm only seeing this being used in `typed_queries` and in `terraform-cloudflare-users`. And in `terraform-cloudflare-users` is created just to be used as the return of a `mocker.patch`.

I can see the value if we use api as dependency of the integration like you commented below. But that would be something new codebase wise, as far as I know.

reconcile/test/test_terraform_vpc_resources.py
esron: I like dependency injection but I not finding any integration that uses this in the integration constructor. I see many examples of the secret reader being created on the `run` method and being passed as parameter to other methods.

reconcile/test/test_terraform_vpc_resources.py
esron: I'm failing to see how this makes the code better. Seems like you are typing less but dictionaries are harder to read for me than class constructors. Dictionaries also don't have autocomplete.

That said I want to be proved wrong.

esron: I think I get now. I needed to add more fields to the query and the more I add test cases more places where I need to update appear. So using `gql_class_factory` and a `something_dict` function can make that easier.

esron: But I still could just use fixtures under fixtures.

fishi0x01: The main advantage is maintainability. `gql_class_factory` adds sane defaults to fields - even if you do not declare them. If you maintain your own fixtures, you will need to adjust them each time sth changes in the data structure.

reconcile/test/test_terraform_vpc_resources.py
fishi0x01: At least SAPM uses that pattern :)

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1177,6 +1179,19 @@ def populate_vpc_peerings(self, desired_state):
                     tf_resource = aws_route(route_identifier, **values)
                     self.add_resource(infra_account_name, tf_resource)
 
+    def populate_vpc_requests(self, vpc_requests: list[VPCRequest]) -> None:
```
fishi0x01, line 1182: Lets use `Iterable[VPCRequest]` to stay flexible, see https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#abstract-collection-types

reconcile/gql_definitions/fragments/aws_vpc_request.gql
```diff
@@ -0,0 +1,33 @@
...
+    resourcesDefaultRegion
+    providerVersion
+    terraformState {
+      provider
+      bucket
+      region
+      integrations {
+        integration
+        key
+      }
```
fishi0x01, line 22: We can use an existing fragment here instead: `...TerraformState`

reconcile/terraform_vpc_resources.py
```diff
@@ -0,0 +1,106 @@
...
+                    sys.exit(ExitCodes.ERROR)
+        else:
+            logging.warning("No VPC requests found, nothing to do.")
+            sys.exit(ExitCodes.SUCCESS)
+
+        accounts_untyped: list[dict] = [acc.dict(by_alias=True) for acc in accounts]
+        with TerrascriptClient(
+            integration=QONTRACT_INTEGRATION,
+            integration_prefix=QONTRACT_TF_PREFIX,
+            thread_pool_size=1,
```
fishi0x01, line 72: Should this be set to `thread_pool_size` from the given params?

reconcile/terraform_vpc_resources.py
```diff
@@ -0,0 +1,106 @@
...
+        if self.params.print_to_file:
+            sys.exit(ExitCodes.SUCCESS)
+
+        tf_client = TerraformClient(
+            integration=QONTRACT_INTEGRATION,
+            integration_version=QONTRACT_INTEGRATION_VERSION,
+            integration_prefix=QONTRACT_TF_PREFIX,
+            accounts=accounts_untyped,
+            working_dirs=working_dirs,
+            thread_pool_size=1,
```
fishi0x01, line 89: Same as above: `thread_pool_size` from given params

reconcile/terraform_vpc_resources.py
```diff
@@ -0,0 +1,106 @@
...
+        secret_reader = create_secret_reader(use_vault=vault_settings.vault)
+
+        data = get_aws_vpc_requests(gql_api=gql.get_api())
+
+        if data:
+            accounts = self._filter_accounts(data)
+            if account_name:
+                accounts = [
+                    account for account in accounts if account.name == account_name
+                ]
```
fishi0x01, line 58: nit: would it make sense to put this into `_filter_accounts`? It fits with the semantics and would make this section less nested.

```python
def _filter_accounts(data: Iterable[VPCRequest], account_name: str) -> list[AWSAccountV1]
```

reconcile/terraform_vpc_resources.py
```diff
@@ -0,0 +1,106 @@
...
+        tf_client = TerraformClient(
+            integration=QONTRACT_INTEGRATION,
+            integration_version=QONTRACT_INTEGRATION_VERSION,
+            integration_prefix=QONTRACT_TF_PREFIX,
+            accounts=accounts_untyped,
+            working_dirs=working_dirs,
+            thread_pool_size=1,
+        )
+
+        tf_client.plan(enable_deletion=False)
```
fishi0x01, line 92: Im not sure, but is there a scenario in which we want to delete a VPCRequest? Should this be a param?

fishi0x01 (commented):
First of all: great work with a clean structure! Was a joy to read it :heart: 

Added some smaller comments.

Another concern is that there is no test actually testing the filter logic, i.e., is the data passed to terrascript client what we actually expect? Currently we check only for a returned success or error code.

reconcile/test/test_terraform_vpc_resources.py
fishi0x01: Yes this makes sense if you use DI. If you use mock.patch instead, then you dont need that fixture.

reconcile/terraform_vpc_resources.py
```diff
@@ -0,0 +1,106 @@
...
+                    account for account in accounts if account.name == account_name
+                ]
+                if not accounts:
+                    logging.warning(
+                        f"The account {account_name} doesn't have any managed vpc. Verify your input"
+                    )
+                    sys.exit(ExitCodes.ERROR)
+        else:
+            logging.warning("No VPC requests found, nothing to do.")
+            sys.exit(ExitCodes.SUCCESS)
```
fishi0x01, line 66: Another thing that came up around the sys.exit pattern https://gitlab.cee.redhat.com/service/dev-guidelines/-/merge_requests/177

Essentially, there are benefits on using raise instead of sys.exit

reconcile/terraform_vpc_resources.py
```diff
@@ -0,0 +1,106 @@
...
+            accounts = self._filter_accounts(data)
+            if account_name:
+                accounts = [
+                    account for account in accounts if account.name == account_name
+                ]
+                if not accounts:
+                    logging.warning(
+                        f"The account {account_name} doesn't have any managed vpc. Verify your input"
+                    )
+                    sys.exit(ExitCodes.ERROR)
```
fishi0x01, line 63: Another thing that came up around the sys.exit pattern https://gitlab.cee.redhat.com/service/dev-guidelines/-/merge_requests/177

Essentially, there are benefits on using raise instead of sys.exit

reconcile/test/test_terraform_vpc_resources.py
esron: I'm going to take a look into that.

reconcile/terraform_vpc_resources.py
esron: Agreed. Fixed.

reconcile/utils/terrascript_aws_client.py
esron: Solved.

reconcile/terraform_vpc_resources/integration.py
```diff
@@ -0,0 +1,216 @@
...
+            dry_run=dry_run,
+            allow_deleting_mrs=False,
+            allow_opening_mrs=True,
+        )
+
+        mr_manager = MergeRequestManager(
+            vcs=vcs,
+            renderer=Renderer(),
+            parser=create_parser(),
+            auto_merge_enabled=True,
```
OliviaHY, line 189: Nice usage of the new `MergeRequestManager`!

OliviaHY (commented):
Looks great overall! Have a request about the provider version and a clarifying question.

reconcile/terraform_vpc_resources/integration.py
```diff
@@ -0,0 +1,216 @@
...
+        enable_deletion = self.params.enable_deletion
+
+        vault_settings = get_app_interface_vault_settings()
+        secret_reader = create_secret_reader(use_vault=vault_settings.vault)
+
+        gql_api = gql.get_api()
+        data = get_aws_vpc_requests(gql_api=gql_api)
+
+        if data:
+            accounts = self._filter_accounts(data, account_name)
```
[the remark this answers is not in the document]
esron: `accounts` here are only used to create the `TerrascriptClient`. The actual logic that polulate the `config.tf.json` files are in the `populate_vpc_requests` in `terrascript_aws_client`. There it iterates through the requests adding resources of the same account in the same `config.tf.json`.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1181,6 +1185,64 @@ def populate_vpc_peerings(self, desired_state):
                     tf_resource = aws_route(route_identifier, **values)
                     self.add_resource(infra_account_name, tf_resource)
 
+    def populate_vpc_requests(self, vpc_requests: Iterable[VPCRequest]) -> None:
+        for request in vpc_requests:
+            # The default values here come from infra repo's module configuration
+            values = {
+                "source": "terraform-aws-modules/vpc/aws",
+                "version": "5.7.1",
```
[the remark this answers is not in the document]
OliviaHY: Okay. What was the bug? Can we make this a variable that get's passed in from integration with a comment linked to the bug explaining the hard coded value?

reconcile/terraform_vpc_resources/merge_request_manager.py
```diff
@@ -0,0 +1,107 @@
+import logging
```
jfchevrette, line 1: This code is duplicated almost verbatim in at least 3 places that I can find. The code is very generic too which to me makes it a good candidate for a utils library. Do you want to attempt to extract it to a library and instead use the library in integration.py?

We can have jira cards to make other integrations use the utility library, no need to do that as part of this PR

reconcile/utils/terrascript_aws_client.py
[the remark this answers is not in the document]
jfchevrette: +1 to Olivia's suggestion. Also this is the module's version, not the same as the provider version

```yaml
$schema: /aws/account-1.yml

name: some-account
# ...
providerVersion: 5.39.1
providerModulesVersions:
  terraform-aws-modules/vpc/aws: 5.7.1
```

I think it improves maintainability, for instance this would enable us to test a new version in a stage account first

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1181,6 +1185,64 @@ def populate_vpc_peerings(self, desired_state):
...
+                "name": request.identifier,
+                "cidr": request.cidr_block.network_address,
+                "private_subnet_tags": {"kubernetes.io/role/internal-elb": "1"},
+                "public_subnet_tags": {"kubernetes.io/role/elb": "1"},
+                "create_database_subnet_group": False,
+                "enable_dns_hostnames": True,
```
jfchevrette, line 1199: I suppose we have no reason to believe we might want to tweak those in the future and hardcoding them is fine and will fit all of our needs? We can also parametrize them later as needed. Just double-checking...

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1181,6 +1185,64 @@ def populate_vpc_peerings(self, desired_state):
...
+                "tags": {
+                    "Terraform": True,
```
jfchevrette, line 1201: can this be set tothe usual tags?
- DEFAULT_TAGS (maybe not?)
- managed_by_integration can be very useful to tell that a resource is managed

reconcile/terraform_vpc_resources/merge_request_manager.py
esron: That is a great idea.

reconcile/utils/terrascript_aws_client.py
esron: Yeah, trying to prevent early abstractions.

reconcile/utils/terrascript_aws_client.py
esron: Yes. Do you happen to know the source for the usual tags?

reconcile/terraform_vpc_resources/merge_request_manager.py
esron: One that notice trying to move this to a utils library is that most of the classes are already a subclass of something commom and the code has some specifics from the VPC merge.
I'm not that confident that there is value in abstracting this further.

`MrData` is singular to this integration and it what is used in `create_merge_request` to check for the existence of an MR.

The shared part is `house_keeping` and the constructor for the `MergeRequestManagerBase` class

reconcile/utils/terrascript_aws_client.py
jfchevrette: No definite source other than the definition of DEFAULT_TAGS in this very file and some managed_by_integration definitions scattered around. There is no standard as far as I can tell

esron: I see that DEFAULT_TAGS is used in the constructor for the class and get applied to resources without tags. Other methods here set `managed_by_integration` to `self.integration`. That worked for this one.

Does that solves the issue here?

jfchevrette: SGTM!

reconcile/utils/terrascript_aws_client.py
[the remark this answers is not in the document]
OliviaHY: So I'm okay with extract this as a variable first then create a ticket to track necessary changes for  `providerModulesVersions:`. What do we think?

OliviaHY (approved):

PR #4155, opened 2024-03-06
[aws-saml-roles] apply user policies

chassing:
`aws-saml-roles` must also consider user-specific AWS policies (`/aws/role-1.user_policies`). The integration creates an AWS policy for each user policy (saml-$GROUP_NAME-$ROLE_NAME-$POLICY_NAME) with an extra `Condition` section to allow it only to the role.users. 

Additionally, it removes the `${aws.username}` and `${aws.accountid}` string replacement in favor of properly escaped variables in the policy. e.g. `$${aws.username}` instead of `${aws.username}`

Ticket: [APPSRE-9900](https://issues.redhat.com/browse/APPSRE-9900)

### DIFF
--- a/reconcile/aws_saml_roles/integration.py
+++ b/reconcile/aws_saml_roles/integration.py
@@ -90,16 +90,26 @@ class AwsSamlRolesIntegration(
     ) -> list[AWSGroupV1]:
         """Get all AWS groups with SSO enabled."""
         data = aws_groups_query(query_func)
-        return [
-            group
-            for group in data.aws_groups or []
-            if integration_is_enabled(self.name, group.account)
-            and (not account_name or group.account.name == account_name)
-            and group.account.sso
-            and group.roles
-            and group.policies
-            and any(role.users for role in group.roles)
-        ]
+        groups = []
+        for group in data.aws_groups or []:
+            if (
+                integration_is_enabled(self.name, group.account)  # noqa: PLR0916
+                and (not account_name or group.account.name == account_name)
+                and group.account.sso
+                and group.roles
+                and group.policies
+                and any(role.users for role in group.roles)
+            ):
+                # remove user policies from other AWS accounts
+                for role in group.roles:
+                    role.user_policies = [
+                        user_policy
+                        for user_policy in role.user_policies or []
+                        if user_policy.account.uid == group.account.uid
+                    ]
+                groups.append(group)
+
+        return groups
 
     def populate_saml_iam_roles(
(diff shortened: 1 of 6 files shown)


rporres:
I have a doubt here. A lot of those `user_policies` are related to Managing own MFAs. Do we want to keep those or we just don't care if they are there or not since they're not going to be used?

chassing:
> I have a doubt here. A lot of those `user_policies` are related to Managing own MFAs. Do we want to keep those or we just don't care if they are there or not since they're not going to be used?

I'll [remove](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/98869) most of them because they aren't needed anymore.

rporres (approved):

reconcile/aws_saml_roles/integration.py
```diff
@@ -90,16 +90,26 @@ def get_aws_groups(
...
+                and group.policies
+                and any(role.users for role in group.roles)
+            ):
+                # remove user policies from other AWS accounts
+                for role in group.roles:
+                    role.user_policies = [
+                        user_policy
+                        for user_policy in role.user_policies or []
+                        if user_policy.account.uid == group.account.uid
+                    ]
```
patjlm, line 109: should we warn on policies not matching the account UID?

reconcile/aws_saml_roles/integration.py
```diff
@@ -113,12 +123,36 @@ def populate_saml_iam_roles(
...
+            additional_role_policies = []
+            for role in group.roles or []:
+                for user_policy in role.user_policies or []:
+                    # patch policy statements and add a aws:userid condition
+                    for statement in user_policy.policy["Statement"]:
+                        condition = statement.setdefault("Condition", {})
+                        string_like = condition.setdefault("StringLike", {})
+                        string_like["aws:userid"] = [
+                            f"*:{user.org_username}" for user in role.users
+                        ]
```
patjlm, line 135: this will override this field if it exists already. Is there a risk  users reference the same policy and we keep on patching it?

patjlm: ah i guess the policy name just below makes this unique, right?

chassing:
With this approach, we'll easily exceed some AWS [IAM limits](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html). For example, a policy document is limited to 6144 characters. We need another approach based an app-interface roles, not on groups.

PR #4157, opened 2024-03-06
[terraform-resources] terraform v1 poc

maorfr:
part of https://issues.redhat.com/browse/APPSRE-9906

depends on https://github.com/app-sre/container-images/pull/97

note: this PR is into `tfv1` and not into `master`.

this PR is an attempt to work on a long lived feature branch, and perhaps explore promoting it with `shardSpecOverrides`.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,5 +1,5 @@
 # this file is used by asdf to automatically select the correct tool versions
 # but it also serves as a reference for non-asdf users
 python 3.11.4
-terraform 0.13.7
+terraform 1.5.5
 oc 4.12.46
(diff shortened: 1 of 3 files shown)


esron (approved):
Nice start

.tool-versions
```diff
@@ -1,5 +1,5 @@
 # this file is used by asdf to automatically select the correct tool versions
 # but it also serves as a reference for non-asdf users
 python 3.11.4
-terraform 0.13.7
+terraform 1.7.4
```
rporres, line 4: For the time being it may be better to use something in 1.5.x for the license stuff, just in case. Unless there's a very compelling reason to have 1.7.4

maorfr: https://github.com/app-sre/container-images/pull/98

PR #4158, opened 2024-03-06
APPSRE-9898 export SAPM publisher data to S3

fishi0x01:
Export SAPM publisher data to S3

```json
{
  "{saas-name}/{resource-template}/{target-name}/{cluster}/{namespace}":
  {
    "commit_sha": "1234",
    "deployment_state": "success|failed|missing"
  },
  ...
}
```

### DIFF
--- a/reconcile/saas_auto_promotions_manager/integration.py
+++ b/reconcile/saas_auto_promotions_manager/integration.py
@@ -19,6 +19,7 @@ from reconcile.saas_auto_promotions_manager.merge_request_manager.renderer impor
     Renderer,
 )
 from reconcile.saas_auto_promotions_manager.publisher import Publisher
+from reconcile.saas_auto_promotions_manager.s3_exporter import S3Exporter
 from reconcile.saas_auto_promotions_manager.subscriber import Subscriber
 from reconcile.saas_auto_promotions_manager.utils.saas_files_inventory import (
     SaasFilesInventory,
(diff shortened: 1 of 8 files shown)


rporres:
Who's the intended user of this data? Because it looks like we're going to have a good amount of json documents difficult to navigate in.

I may be completely off here but what I'm thinking of is: would it make sense for the destination (S3, Postgres, Mongo, whatever) to be something configurable and make the exporter something more generic that calls the different exporters based on the configuration?

fishi0x01:
@rporres afaik we currently have a team around @jmelis that is interested in that data. cc @maorfr 

Current idea is to have a single json file containing all the information. 

I like your idea of configurable exporters - it would be a much better design. Still, due to time constraints, Im currently leaning more towards supporting that at a later point once there is a need.

Another intent of this effort is to provide an entrypoint for SAPM data export that can be easily adjusted while Im on PTO.

rporres:
fair enough!

rporres (approved):

reconcile/saas_auto_promotions_manager/integration.py
```diff
@@ -133,11 +139,16 @@ def init_external_dependencies(
     deployment_state = PromotionState(
         state=init_state(integration=OPENSHIFT_SAAS_DEPLOY, secret_reader=secret_reader)
     )
+    s3_exporter = S3Exporter(
+        state=init_state(integration=QONTRACT_INTEGRATION, secret_reader=secret_reader),
```
hemslo, line 143: `state` requires call `state.cleanup()` after use to close aws client.

fishi0x01: Good catch! Fixing here https://github.com/app-sre/qontract-reconcile/pull/4162

fishi0x01:
@patjlm yes adding a "first_seen" timestamp is a good idea. My remaining question is whether that timestamp should be updated on re-deploys (e.g., test saas file with changed params) - but that can be adjusted easily as we need it.

PR #4159, opened 2024-03-07
[terraform-resources] sensitive outputs

maorfr:
discovered as part of the exploration in https://issues.redhat.com/browse/APPSRE-9906

https://developer.hashicorp.com/terraform/language/values/outputs#sensitive-suppressing-values-in-cli-output

this needs to be tested to see what is the impact to the integration, since it relies on the state to later populate output secrets.


geoberle:
> this needs to be tested to see what is the impact to the integration, since it relies on the state to later populate output secrets

can you elaborate? is anything else but these secrets showing up in stdout expected?

hemslo (approved):
We always use json output, and sensitive filter not applied to json, shouldn't affect our usage.

PR #4162, opened 2024-03-07
APPSRE-9898 cleanup connections

fishi0x01:
Follow up of https://github.com/app-sre/qontract-reconcile/pull/4158


hemslo (approved):
LGTM!

PR #4169, opened 2024-03-10
[openshift-resources] support templating from s3 object

maorfr:
part of https://issues.redhat.com/browse/APPSRE-9898

following up on #4158, we want to be able to dump these results from s3 to a ConfigMap.

this PR introduces templating from an s3 object, which will allow us to:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: publisher-data
data:
  publisher-data.json:
    {{ s3('account-name', 'bucket-name', 'path/to/object') | tojson }}
```
example: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/99057

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -42,6 +42,7 @@ from reconcile.utils.exceptions import FetchResourceError
 from reconcile.utils.jinja2.utils import (
     FetchSecretError,
     lookup_github_file_content,
+    lookup_s3_object,
     lookup_secret,
     process_extracurlyjinja2_template,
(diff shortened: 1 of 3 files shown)


geoberle (commented):
Nice. Do you also need to register a PR-check mock for this new jinja function for early exit purposes?

maorfr:
@geoberle done in https://github.com/app-sre/qontract-reconcile/pull/4169/commits/2c75c2ae077d9f9b840e4b826fd24f14d3e96706

geoberle (approved):

PR #4178, opened 2024-03-11
[terraform-resources] prevent same account from being included and excluded

maorfr:
we currently fail if `--account-name` is used with `--exclude-accounts`.

this failure is introduced by the `select-integrations.py` enhancement to only check affected shards (the implementation is the usage of `--account-name`), while also supporting `shardSpecOverrides`.

with this adjustment, we only fail if the same account is both included and excluded.

### DIFF
--- a/reconcile/terraform_resources.py
+++ b/reconcile/terraform_resources.py
@@ -203,8 +203,8 @@ def get_aws_accounts(
         logging.error(message)
         raise ExcludeAccountsAndDryRunException(message)
 
-    if exclude_accounts and include_accounts:
-        message = "Using --exclude-accounts and --account-name at the same time is not allowed"
+    if any(a in exclude_accounts for a in include_accounts):
+        message = "Using --exclude-accounts and --account-name with the same account is not allowed"
         logging.error(message)
         raise ExcludeAccountsAndAccountNameException(message)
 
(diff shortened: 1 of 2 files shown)


reconcile/terraform_resources.py
```diff
@@ -203,8 +203,10 @@ def get_aws_accounts(
         logging.error(message)
         raise ExcludeAccountsAndDryRunException(message)
 
-    if exclude_accounts and include_accounts:
-        message = "Using --exclude-accounts and --account-name at the same time is not allowed"
+    if (exclude_accounts and include_accounts) and any(
+        a in exclude_accounts for a in include_accounts
```
hemslo, line 207: could use `set`

```python
any(set(exclude_accounts) & set(include_accounts))
```

hemslo (approved):
LGTM!

reconcile/terraform_resources.py
maorfr: i find this less readable, apologies :pray:

hemslo: It's useful when we want to put duplicate result in exception logging like `duplicates := (set(exclude_accounts) & set(include_accounts))`

Not a big issue here as length of both `exclude_accounts` and `include_accounts` should be relative small lists (less than 10), time complexity of `a in exclude_accounts for a in include_accounts` is O(NM) (N = `len(exclude_accounts)`, M = `len(include_accounts`) while using set is O(N+M), in general any N^2 should be avoided in code. Preprocess both to `set` would simplify other checks too, as they all converted to set eventually.

PR #4181, opened 2024-03-11
terraform required version

maorfr:
terraform integrations validate the binary version.

there are cases that require manual intervention, such as importing resources. in such cases, we use the `--print-to-file` option and execute `terraform` locally.

this PR adds usage of `required_version` to avoid mistakes of binary version: https://developer.hashicorp.com/terraform/language/settings#specifying-a-required-terraform-version

### DIFF
--- a/reconcile/utils/terrascript/cloudflare_client.py
+++ b/reconcile/utils/terrascript/cloudflare_client.py
@@ -107,7 +107,7 @@ def create_cloudflare_terrascript(
     terrascript += Terraform(
         backend=backend,
         required_providers=required_providers,
-        required_version=TERRAFORM_VERSION,
+        required_version=TERRAFORM_VERSION[0],
     )
 
     cloudflare_provider_values = {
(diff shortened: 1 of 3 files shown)


reconcile/utils/terrascript/cloudflare_client.py
```diff
@@ -21,6 +21,7 @@
     provider,
 )
 
+from reconcile.cli import TERRAFORM_VERSION
```
patjlm, line 24: why is `TERRAFORM_VERSION` a list? If there are multiple version, should we always use the first `TERRAFORM_VERSION[0]`?

maorfr: this did not start out as a list: #1361

maorfr: and changed to a list here: #4126

geoberle (approved):

reconcile/utils/terrascript/cloudflare_client.py
geoberle: ah so it is not really used as a list for the TF case but was made a list because of OC

PR #4182, opened 2024-03-11
APPSRE-8713-01: added code for deadmanssnitch integration

mehfuz:
Reference: https://issues.redhat.com/browse/APPSRE-8713


reconcile/cli.py
```diff
@@ -3333,6 +3333,16 @@ def acs_policies(ctx):
         ctx=ctx.obj,
     )
 
+@integration.command(short_help="Automate Deadmanssnitch Creation/Deletion")
+@click.pass_context
+def deadmanssnitch_automation(ctx):
```
patjlm, line 3338: NIT: can we just name this `deadmansnitch`? Same for the file `reconcile/deadmanssnitch_automation.py`. All this is about automation, no need to mention it ;-)

reconcile/utils/deadmanssnitch_api.py
```diff
@@ -0,0 +1,59 @@
...
+
+class DeadMansSnitchApi:
+    def __init__(self, token: str, url: str = BASE_URL) -> None:
+        self.token = token
+        self.url = url
+
+    def get_snitches(self, tags: list[str]) -> list[Snitch]:
+        full_url = f"{self.url}?tags={','.join(tags)}"
+        logging.info("Getting snitches for tags:%s", tags)
+        response = requests.get(url=full_url, auth=(self.token, ""))
```
hemslo, line 40: since methods in this class can be called multiple times in one run, `requests.get/post/delete` will create a new session every time, should init a session during init, can do cleanup after usage, checkout `requests.Session()` usage, example [acs](https://github.com/app-sre/qontract-reconcile/blob/a804113c6645d60394086ccf7583ef24042fb6db/reconcile/utils/acs/base.py#L24)

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def create_diff_data(self, cluster_name: str, action: str, current_state: Optional[dict[str, Any]] = None) -> dict[str, str]:
+        data: dict[str, str] = {}
+        data["action"] = action
+        data["cluster_name"] = cluster_name
+        if action == self.UPDATE_VAULT:
+            data["snitch_url"] = current_state["check_in_url"]
+        elif action == self.DELETE_SNITCH:
+            data["token"] = current_state["token"]
+        return data
```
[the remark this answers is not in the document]
mehfuz: @hemslo dynamic dict was chosen because each operation had its own different fields....for example to create snitch the diff object like u can see here....https://github.com/mehfuz/qontract-reconcile/blob/upstream/reconcile/test/test_deadmanssnitch.py#L172

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,187 @@
...
+                if snitch.get_cluster_name() == cluster.name:
+                    snitch_map = snitch.dict(by_alias=True)
+                    snitch_map["cluster_name"] = cluster.name
+                    try:
+                        full_secret_path = {"path": snitch_secret_path, "field": f"deadmanssnitch-{snitch_map['cluster_name']}-url"}
+                        # match the exact url value
+                        snitch_map["vault_snitch_value"] = self.secret_reader.read(full_secret_path).strip()
+                    except SecretNotFound:
+                        # need to create a new vault secret for this record.
+                        snitch_map["vault_snitch_value"] = SECRET_NOT_FOUND
```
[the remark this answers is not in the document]
mehfuz: @hemslo  we set a value only for a specfic form of exception which is SecretNotFound.. A None value will cause ambiguity like is it because of secret not found? or connection error? or something else.. Settings a value confirms that this is due to absence of secret and hence should be created

reconcile/deadmanssnitch.py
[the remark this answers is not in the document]
hemslo: each action can have their own type, a base type plus 3 subtypes.

reconcile/test/test_deadmanssnitch.py
```diff
@@ -0,0 +1,256 @@
...
+from reconcile.gql_definitions.common.clusters_with_dms import (
+    ClusterV1,
+)
+from reconcile.utils.deadmanssnitch_api import DeadMansSnitchApi
+
+TOKEN = "test_token"
+FAKE_URL = "https://fake.deadmanssnitch.com/v1/snitches"
+@pytest.fixture
+def deadmanssnitch_api() -> DeadMansSnitchApi:
+    return DeadMansSnitchApi(token=TOKEN, url=FAKE_URL)
```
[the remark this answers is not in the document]
mehfuz: @hemslo u mean something like this right https://github.com/mehfuz/qontract-reconcile/blob/upstream/reconcile/test/test_deadmanssnitch.py#L191toL220 ?

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,187 @@
...
+    CREATE_SNITCH = "create_snitch"
+    DELETE_SNITCH = "delete_snitch"
+
+    def __init__(self, deadmanssnitch_api: DeadMansSnitchApi, settings: DeadMansSnitchSettingsV1, vault_client: VaultClient = None) -> None:
+        self.deadmanssnitch_api = deadmanssnitch_api
+        self.settings = settings
+        if vault_client is not None:
+            self.vault_client = vault_client
+        else:
+            self.vault_client = cast(_VaultClient, VaultClient())
```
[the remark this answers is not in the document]
mehfuz: initially VaultCilent was supposed to be intialzed as a part of DiffHandler. Had to inject this for better testing.

reconcile/test/test_deadmanssnitch.py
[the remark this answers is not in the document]
hemslo: I mean methods like `create_snitch` already tested, here use a mock object `create_autospec(DeadMansSnitchApi)`, then just verify `.create_snitch.assert_called_once_with...`

reconcile/deadmanssnitch.py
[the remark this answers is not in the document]
hemslo: then can also use the same in real run, make the behaviour same as in test.

reconcile/deadmanssnitch.py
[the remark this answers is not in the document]
hemslo: if you make `vault_snitch_value` as a field, value as `str | None`, then it's clearer than always string. Or have another explicit field `valut_snitch_exist` as `bool`.

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,226 @@
...
+            DiffData(
+                cluster_name=cluster,
+                action=Action.update_vault,
+                data=current_state[cluster].check_in_url,
+            )
+            for cluster in desired_cluster_enabled_true & current_state.keys()
+            # add to update set only if they exists, flag is set to true but vault and snitch url mismatch
+            if current_state[cluster].vault_data
+            and current_state[cluster].vault_data != current_state[cluster].check_in_url
+        ]
```
hemslo, line 184: `current_state[cluster]` is used multiple times, can assign a var to it for reuse

```python
        [
            DiffData(
                cluster_name=cluster_name,
                action=Action.update_vault,
                data=cluster.check_in_url,
            )
            for cluster_name in desired_cluster_enabled_true & current_state.keys()
            if (cluster := current_state[cluster_name])
            and cluster.vault_data !=  cluster.check_in_url
        ]
```

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,187 @@
...
+                    desired_states.remove(cluster)
+        for desired_state in desired_states:
+            # we are left with only cluster which needs create_snitch if `enableDeadMansSnitch: True`
+            if desired_state.enable_dead_mans_snitch:
+                diffs.append(diff_handler.create_diff_data(action=DiffHandler.CREATE_SNITCH, cluster_name=desired_state.name))
+        return diffs
+
+
+    def apply_diffs(self, dry_run: bool, diffs: dict[str, Any], diff_handler: DiffHandler) -> None:
+        logging.info(diff_handler.summarize(diffs=diffs))
```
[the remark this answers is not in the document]
hemslo: use `if not diffs` instead of `len` check

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,226 @@
...
+        ]
+        # current state - get snitches for tag app-sre
+        current_state = self.get_current_state(
+            deadmanssnitch_api, clusters, settings.snitches_path
+        )
+        if len(current_state) > 0 and len(clusters) > 0:
+            diff = self.get_diff(current_state=current_state, desired_state=clusters)
+            self.apply_diffs(dry_run, diff, diff_handler)
+        # close session
+        deadmanssnitch_api.close_session()
```
hemslo, line 226: `close_session` will be skipped when any exceptions happen before it, use `with` as context manager for `deadmanssnitch_api` to ensure it will be closed.

reconcile/typed_queries/clusters_with_dms.py
```diff
@@ -0,0 +1,16 @@
...
+from reconcile.utils import gql
+from reconcile.utils.gql import GqlApi
+
+
+def get_clusters_with_dms(
+    gql_api: Optional[GqlApi] = None,
+) -> list[ClusterV1]:
+    api = gql_api if gql_api else gql.get_api()
+    data = query(query_func=api.query)
+    return list(data.clusters or [])
```
hemslo, line 16: is `list` here needed?

reconcile/utils/deadmanssnitch_api.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class DeadMansSnitchApi:
+    def __init__(self, token: str, url: str = BASE_URL) -> None:
+        self.token = token
+        self.url = url
+        self.session = requests.Session()
+
+    def close_session(self) -> None:
+        self.session.close()
```
hemslo, line 44: for easy context management, use `__enter__` and `__exit__` as context manager to close session, can call `self.close_session` in `__exit__`

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,187 @@
...
+    def apply_diffs(self, dry_run: bool, diffs: dict[str, Any], diff_handler: DiffHandler) -> None:
+        logging.info(diff_handler.summarize(diffs=diffs))
+        if dry_run:
+            return
+        for diff in diffs:
+            try:
+                diff_handler.apply_diff(diff)
+            except Exception as e:
+                logging.error(str(e))
+                continue
```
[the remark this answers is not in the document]
mehfuz: @hemslo if we do not continue here..the other diff objects won't run.even if they are safe to run..

reconcile/gql_definitions/common/clusters_with_dms.gql
```diff
@@ -0,0 +1,13 @@
+# qenerate: plugin=pydantic_v1
+
+query ClustersWithMonitoring($name: String) {
+  clusters: clusters_v1(name: $name) {
```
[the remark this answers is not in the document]
mehfuz: @hemslo removed this dependency

reconcile/deadmanssnitch.py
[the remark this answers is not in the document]
hemslo: checkout how to use `ExceptionGroup` gather exceptions and throw after iteration, when any of them fail, the final integration should fail.

reconcile/deadmanssnitch.py
hemslo: note `cluster` can be used in `data=cluster.check_in_url` too

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,226 @@
...
+        })
+        vault_client = cast(_VaultClient, VaultClient())
+        with DeadMansSnitchApi(token=token) as deadmanssnitch_api:
+            diff_handler = DiffHandler(deadmanssnitch_api, settings, vault_client)
+            # desired state - filter cluster having enableDeadMansSnitch field
+            clusters = [
+                cluster
+                for cluster in get_clusters_with_dms()
+                if cluster.enable_dead_mans_snitch is not None
+            ]
```
hemslo, line 217: `get_clusters_with_dms` this name sounds like the result only contains clusters with dms, ideally the filter can be done in server side to only return clusters with `enable_dead_mans_snitch` not None, using server side filter https://github.com/app-sre/qontract-server?tab=readme-ov-file#field-not-equal-predicate

```
"filter": {
    "enable_dead_mans_snitch: {
        "ne": null
    }
}
```

This can reduce size of the response and simplify client side logic.

reconcile/test/test_deadmanssnitch.py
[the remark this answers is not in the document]
mehfuz: @hemslo  added this here (https://github.com/mehfuz/qontract-reconcile/blob/upstream/reconcile/test/test_deadmanssnitch.py#L256to263) and (https://github.com/mehfuz/qontract-reconcile/blob/upstream/reconcile/test/test_deadmanssnitch.py#L308to311)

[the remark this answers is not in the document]
hemslo: now you can replace this fixture with `create_autospec(DeadMansSnitchApi)` to simplify this test file.

reconcile/gql_definitions/common/clusters_with_dms.gql
```diff
@@ -0,0 +1,13 @@
+# qenerate: plugin=pydantic_v1
+
+query ClustersWithMonitoring($filter: JSON){
+  clusters: clusters_v1(filter:$filter){
```
hemslo, line 4: better to put some spaces for readability
```
clusters: clusters_v1(filter: $filter) {
```

reconcile/cli.py
mehfuz: @patjlm updated the integratiion name to `deadmanssnitch`

reconcile/utils/deadmanssnitch_api.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def to_dict(self) -> dict[str, Any]:
+        return self.dict(by_alias=True)
+
+
+class DeadMansSnitchApi:
+    def __init__(self, token: str, url: str = BASE_URL) -> None:
+        self.token = token
+        self.url = url
+        self.session = requests.Session()
```
janboll, line 42: Think you have to specify some time out, AFAIR requests does not specify a default one :thinking: https://requests.readthedocs.io/en/latest/user/advanced/#timeouts

reconcile/utils/deadmanssnitch_api.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def get_snitches(self, tags: list[str]) -> list[Snitch]:
+        full_url = f"{self.url}?tags={','.join(tags)}"
+        logging.debug("Getting snitches for tags:%s", tags)
+        response = self.session.get(url=full_url, auth=(self.token, ""))
+        response.raise_for_status()
+        snitches = [Snitch(**item) for item in response.json()]
+        return snitches
+
+    def create_snitch(self, payload: dict) -> Snitch:
```
janboll, line 58: Can you use a BaseModel for the payload?

reconcile/utils/deadmanssnitch_api.py
mehfuz: @janboll updated the code likewise

reconcile/utils/deadmanssnitch_api.py
mehfuz: @janboll add baseModel would be enforcing all the fields here...https://github.com/app-sre/qontract-reconcile/blob/f0c369d5c391d1db90ef56a97588b076ca7bae3b/reconcile/deadmanssnitch.py#L90to99 ..the fact is deadmanssnith does not require all these fields...this file is supposed to be close to the actual api.deadmanssnitch.com in its behavior hence a dict was chosen

reconcile/test/test_deadmanssnitch.py
```diff
@@ -0,0 +1,333 @@
...
+        DiffData(
+            cluster_name="test_cluster_5",
+            data="testURL",
+            action=Action.update_vault,
+        ),
+    ]
+
+    expected_output_map = [data.__dict__ for data in expected_output]
+    diff_data_map = [data.__dict__ for data in diff_data]
+    assert diff_data_map == expected_output_map
```
hemslo, line 220: why require `__dict__`? If `DiffData` is a `BaseModel`, then eq compare should work for value check

mehfuz: @hemslo `DiffData` is not a BaseModel so had to use `__dict__` for testcase

hemslo: Can we make it a BaseModel to simplify this, since it's a dataclass, either use `@dataclass` or BaseModel to make it explicit.

hemslo: Now that it becomes a `BaseModel`, it reveals a type check error, `data` can be optional, but when action is delete, `data` can't be None. Can describe this kind of typing strictly like this:

```python
class DiffData(BaseModel, ABC):
    cluster_name: str
    action: Action

class CreateSnitchDiffData(DiffData):
    action = Action.create_snitch

class DeleteSnitchDiffData(DiffData):
    action = Action.delete_snitch
    data: str

class UpdateVaultDiffData(DiffData):
    action = Action.update_vault
    data: str
```

You can rename `data` to better suit for the meaning of that class. Then you can simply use the subclass directly and pattern match on it:

```python
        create_snitches = [
            CreateSnitchDiffData(cluster_name=cluster)
            # add cluster to create set only in case they do not exist in current state
            for cluster in desired_cluster_enabled_true - current_state.keys()
        ]
```

```python
    def apply_diff(self, diff: DiffData) -> None:
        match diff:
            case CreateSnitchDiffData(cluster_name=cluster_name):
                self.create_snitch(cluster_name)
```

hemslo (approved):
LGTM!

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,244 @@
...
+                self.vault_client.write(
+                    {
+                        "path": self.settings.snitches_path,
+                        "data": {f"deadmanssnitch-{cluster_name}-url": check_in_url},
+                    },
+                    decode_base64=False,
+                )
+
+    def create_snitch(self, cluster_name: str) -> None:
+        tags = ["app-sre"]
```
maorfr, line 110: `"app-sre"` should not be hard coded. perhaps from settings?

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,244 @@
...
+        return snitch
+
+    def get_current_state(
+        self,
+        deadmanssnitch_api: DeadMansSnitchApi,
+        clusters: list[ClusterV1],
+        snitch_secret_path: str,
+    ) -> dict[str, Snitch]:
+        # current state includes for deadmanssnithch response and associated secret in vault
+        snitches = deadmanssnitch_api.get_snitches(tags=["app-sre"])
```
maorfr, line 160: same as above, `"app-sre"` should not be hard coded

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,244 @@
...
+                        "data": {f"deadmanssnitch-{cluster_name}-url": check_in_url},
+                    },
+                    decode_base64=False,
+                )
+
+    def create_snitch(self, cluster_name: str) -> None:
+        tags = ["app-sre"]
+        alert_email = [self.settings.alert_email]
+        payload = {
+            "name": f"prometheus.{cluster_name}.devshift.net",
```
maorfr, line 113: can this come from the cluster `prometheusUrl` field instead?

reconcile/utils/deadmanssnitch_api.py
```diff
@@ -0,0 +1,85 @@
...
+    notes: str
+    status: str
+    check_in_url: str
+    interval: str
+    alert_type: str
+    alert_email: list[str]
+    vault_data: Optional[str]
+
+    def get_cluster_name(self) -> str:
+        return self.name.split(".")[1]
```
maorfr, line 33: can use a mapping between cluster names and their `prometheusUrl` field to avoid string splitting? probably worth adding a `cluster_name` field somehow?

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,244 @@
...
+        current_state = {
+            cluster.name: self.add_vault_data(snitch, snitch_secret_path)
+            for cluster in clusters
+            if (snitch := snitches_with_cluster_mapping.get(cluster.name))
+        }
+        return current_state
+
+    @staticmethod
+    def get_diff(
+        current_state: dict[str, Snitch], desired_state: list[ClusterV1]
```
maorfr, line 174: i find it confusing that `current_state` and `desired_state` are of different types.

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,244 @@
...
+            case UpdateVaultDiffData(
+                cluster_name=cluster_name, check_in_url=check_in_url
+            ):
+                self.vault_client.write(
+                    {
+                        "path": self.settings.snitches_path,
+                        "data": {f"deadmanssnitch-{cluster_name}-url": check_in_url},
+                    },
+                    decode_base64=False,
+                )
```
maorfr, line 107: just my opinion, but i feel that this class is only causing us to move data from place to place, and the actual functionality can be significantly reduced and added into `DeadMansSnitchIntegration`.

maorfr (approved):
first off, kudos. this PR will work exactly as intended.
the handling of of True/False/None to allow snitches to be deleted is great.
i like it the simplicity of utils/deadmanssnitch_api :+1:

this is why i'm approving this PR. please feel free to ignore the rest of this comment!

having said that, i find reconcile/deadmanssnitch too complicated. imagine the following pseudo code:
```
def run():
  clusters = get_clusters_with_dms()
  current_state = get_current_state(clusters)
  for cluster in clusters:
    if cluster.dms_enabled and cluster not in current_state:
      // create and update vault
    if cluster.dms_disabled and cluster in current_state:
      // delete
```
i'll leave it to you to decide if you want to simplify.

thank you!

reconcile/deadmanssnitch.py
chassing: Or introduce a command line `--tags` option

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,244 @@
...
+class DeadMansSnitchIntegration(QontractReconcileIntegration[NoParams]):
+    """Integration to automate deadmanssnitch snitch api during cluster dressup."""
+
+    @property
+    def name(self) -> str:
+        return QONTRACT_INTEGRATION
+
+    def __init__(self) -> None:
+        super().__init__(NoParams())
+        self.qontract_integration_version = make_semver(0, 1, 0)
```
chassing, line 140: There is no need to set this variable or override `__init__`

chassing (approved):
LGTM

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,159 @@
...
+        )
+
+    def reconcile(
+        self,
+        dry_run: bool,
+        deadmanssnitch_api: DeadMansSnitchApi,
+        cluster: ClusterV1,
+        snitch: Optional[Snitch] = None,
+    ) -> None:
+        match (cluster.enable_dead_mans_snitch, snitch):
```
maorfr, line 109: i personally find this switch case confusing. consider a simple if-else?

maorfr:
big like on the refactor commit. things are much simpler! still room to simplify :pray:

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,159 @@
...
+    def write_snitch_to_vault(self, cluster_name: str, snitch_url: str) -> None:
+        self.vault_client.write(
+            {
+                "path": self.settings.snitches_path,
+                "data": {f"deadmanssnitch-{cluster_name}-url": snitch_url},
+            },
+            decode_base64=False,
+        )
+
+    def add_vault_data(self, snitch: Snitch, snitch_secret_path: str) -> Snitch:
```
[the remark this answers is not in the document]
mehfuz: @maorfr even if I change it there we would still need `snitch.get_cluster_name` for creating mapping between clustername and snitch object like here https://github.com/app-sre/qontract-reconcile/blob/cdc7253c9feec868ead6c4386b98304a29817f30/reconcile/deadmanssnitch.py#L76

reconcile/deadmanssnitch.py
mehfuz: Original idea was the same..Can be done but involves lot of condition, like the following 
`
if cluster.enable_dead_mans_snitch and snitch is None:
   //create_snitch
if not cluster.enable_dead_mans_snitch and snitch:
//delete_snitch
if cluster.enable_dead_mans_snitch and snitch:
     if snitch.check_in_url != snitch.vault_value:
           //update in vault
`

reconcile/deadmanssnitch.py
mehfuz: So actually the main reson they both are different is because of the `check_in_url`. having same type only suits in case of delete and update as we have a valid value for check_in_url` in that case. But in case of create we do not have a value for `check_in_url`. So I kept it different.

reconcile/deadmanssnitch.py
mehfuz: added `tags` in schema along with some other fields which were previously hardcoded here.

reconcile/deadmanssnitch.py
maorfr: much more readable compared to `case (True, None)`

reconcile/deadmanssnitch.py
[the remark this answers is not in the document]
maorfr: https://github.com/app-sre/qontract-reconcile/pull/4182/files#r1531593408

reconcile/deadmanssnitch.py
mehfuz: will update it accordingly

reconcile/deadmanssnitch.py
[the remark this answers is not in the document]
mehfuz: will add the change

reconcile/deadmanssnitch.py
hemslo: the condition is too nested, avoid multiple indented conditional check via extracting methods to give them meaningful names, add comments to explain each create/delete/update is a bad smell. Should be easy to read like this:


```python
if should_create_snitch(cluster, snitch):
   # self.create_snitch(
if should_delete_snitch(cluster, snitch):
   # self.delete_snitch(
if should_update_in_vault(cluster, snitch):
   # self.update_in_vault(
```

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,159 @@
...
+                    self.reconcile(
+                        dry_run,
+                        deadmanssnitch_api,
+                        cluster,
+                        snitch=current_state.get(cluster.name),
+                    )
+                except Exception as e:
+                    errors.append(e)
+                if errors:
+                    raise ExceptionGroup("Errors occurred while reconcile", errors)
```
hemslo, line 159: this `if` is in wrong indent, should do this after all clusters processed

reconcile/utils/deadmanssnitch_api.py
mehfuz: added mapping between prometheus and cluster here https://github.com/app-sre/qontract-reconcile/blob/ff5294cb3123414309323dcd2b3916a5c5da49dd/reconcile/deadmanssnitch.py#L157

maorfr (approved):
LGTM

some more comments remain, but can be improved at a later point.

nice work!

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,182 @@
...
+        snitches = deadmanssnitch_api.get_snitches(tags=["app-sre"])
+        # create snitch_map only for  the desired clusters
+        snitches_with_cluster_mapping = {
+            cluster.name: snitch
+            for snitch in snitches
+            for cluster in clusters
+            if (cluster_to_prometheus_mapping.get(cluster.name) == snitch.name)
+        }
+        current_state = {
+            cluster.name: self.add_vault_data(cluster.name, snitch, snitch_secret_path)
```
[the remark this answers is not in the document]
hemslo: I saw that too, but later found that method cached secret internally, so should be only one request to vault

reconcile/deadmanssnitch.py
hemslo: saw the new change `snitch.needs_vault_update()`, it is good since it reduced 1 software cyclonic complexity, but if we step back to see the bigger scope here, we can find we always need to check `cluster.enable_dead_mans_snitch` and `snitch` together to do next action, which indicate we are missing a domain here.

There should be a domain that contains knowledge about the condition of each action and how to evaluate them, there are also cases that you need additional property such as clusters' prometheus but it's not available in gql generated model, `cluster_to_prometheus_mapping` is not suitable in current method because other args are all limited to a single object already, here it should also narrow down to a single prometheus.

Suppose you create a new domain model `ClusterSnitch`, contains all fields (cluster, snitch, prometheus...) you need, this method would be natural like this:

```python
  def reconcile(
        self,
        dry_run: bool,
        deadmanssnitch_api: DeadMansSnitchApi,
        cluster_snitch: ClusterSnitch,
  ) -> None:
    if cluster_snitch.need_create():
      # create
    if cluster_snitch.need_delete():
      # delete
    if cluster_snitch.need_update():
      # update
```

reconcile/deadmanssnitch.py
```diff
@@ -0,0 +1,182 @@
...
+        self,
+        deadmanssnitch_api: DeadMansSnitchApi,
+        clusters: list[ClusterV1],
+        snitch_secret_path: str,
+        cluster_to_prometheus_mapping: dict[str, str],
+    ) -> dict[str, Snitch]:
+        # current state includes for deadmanssnithch response and associated secret in vault
+        snitches = deadmanssnitch_api.get_snitches(tags=["app-sre"])
+        # create snitch_map only for  the desired clusters
+        snitches_with_cluster_mapping = {
```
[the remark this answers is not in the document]
mehfuz: @maorfr tried to come up with your suggestion

PR #4184, opened 2024-03-11
:necktie: aws-saml-roles: AWS SSO based on app-interface roles

chassing:
Adapt `aws-saml-roles` to use app-interface roles (`/access/role-1.yml`) instead of AWS groups (`/aws/group-1.yml`)

:page_facing_up: Ticket: [APPSRE-9900](https://issues.redhat.com/browse/APPSRE-9900)
:memo: Design doc: [aws-sso: role-based access - design doc](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/99203)


reconcile/aws_saml_roles/integration.py
```diff
@@ -85,40 +177,75 @@ def get_aws_accounts(
...
         """Populate the SAML IAM roles."""
-        for group in aws_groups:
-            # aws groups without policies are filtered out by the query
-            # duplicated policies aren't allowed in the same role
-            policies = group.policies or []
-            if len(policies) != len(set(policies)):
-                raise ValueError(
-                    f"Group {group.name} has duplicated policies: {policies}"
+        for role in aws_roles:
+            seen_custom_policies = []
```
hemslo, line 231: should use set instead of list for `in` check. btw is duplicate policy already validated before?

reconcile/utils/terrascript_aws_client.py
```diff
@@ -2573,18 +2573,30 @@ def populate_tf_resource_role(self, spec):
...
     ) -> None:
         """Manage the an IAM role needed for SAML authentication."""
         managed_policy_arns = [
             f"arn:{self._get_partition(account)}:" + f"iam::aws:policy/{policy}"
-            for policy in policies
+            for policy in aws_managed_policies
+        ]
+        managed_policy_arns += [
+            f"arn:{self._get_partition(account)}:"
+            + f"iam::{self.uids[account]}:policy/{policy}"
```
hemslo, line 2598: no need for `+`, can directly put it in one string, just 2 lines in code if one line is too long.

```python
            f"arn:{self._get_partition(account)}:"
            f"iam::{self.uids[account]}:policy/{policy}"
```

same for the `aws_managed_policies` part.

chassing: Fixed in [03d472e](https://github.com/app-sre/qontract-reconcile/pull/4184/commits/03d472e3315e211202cea88e10d1ce4b602dc1a2)

reconcile/aws_saml_roles/integration.py
chassing: Fixed in [076aefe](https://github.com/app-sre/qontract-reconcile/pull/4184/commits/076aefeafa9d107cfec95216cc301b023d23f7f5). What do you 

> btw is duplicate policy already validated before?

what do you mean?

hemslo: I saw some validation logic checking duplicate policies, not sure if it's already checked that all policies passed here already deduplicated.

hemslo (approved):
LGTM!

reconcile/aws_saml_roles/integration.py
chassing: The main logic is below. The integration creates a policy once per account, regardless of how often a policy is referenced in other roles.

hemslo: I see, same policy can be refered in multiple roles, but we should only create it once, this is to get uniq policies, this is role level deduplicate, right? The code will reinit `seen_custom_policies` for every role, can 2 different roles for the same account? And it seems to deduplicate by policy name of one role? Is it same as this logic?

```python
        for role in aws_roles:
            unique_policies = {
                custom_policy.name: custom_policy
                for custom_policy in role.custom_policies
            }

            for custom_policy in unique_policies.values():
                ts.populate_iam_policy(
                    account=role.account,
                    name=custom_policy.name,
                    policy=custom_policy.policy,
                )
```

hemslo: Or you mean account level deduplicate

```python
        unique_policies = {
            (role.account, custom_policy.name): custom_policy
            for role in aws_roles
            for custom_policy in role.custom_policies
        }
```

chassing: Very good catch Di ❤️ I hate [Mock.assert_has_calls](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_has_calls) because it doesn't validate the call count :(
Fixed in the last commit.

hemslo: plus one, I only use `assert_has_calls` when `assert_called_once_with` can't be used, strict check in test is better.

PR #4188, opened 2024-03-12
[terraform-resources] aws provider 5.x enhancements

maorfr:
part of https://issues.redhat.com/browse/APPSRE-8577

these changes are required to be backwards compatible with data in app-interface(s), and also with terraform aws provider version 5.x.

we do this by updating `values` to match changes in the terraform provider.

testing release candidate here: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/99480

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -4050,9 +4050,9 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
             if val is not None:
                 values[key] = val
 
-        if self.versions and not self.versions.get(spec.provisioner_name).startswith(
-            "3"
-        ):
+        if self.versions and not self.versions.get(
+            spec.provisioner_name, ""
+        ).startswith("3"):
             if spec.provider == "rds":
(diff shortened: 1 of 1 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -4047,6 +4050,24 @@ def init_values(self, spec: ExternalResourceSpec, init_tags: bool = True) -> dic
             if val is not None:
                 values[key] = val
 
+        if self.versions and not self.versions.get(
+            spec.provisioner_name, ""
+        ).startswith("3"):
+            if spec.provider == "rds":
+                if db_name := values.pop("name", None):
+                    values["db_name"] = db_name
```
patjlm, line 4058: shouldn't we use `values["name"]` for provider v3*

maorfr: this block is conditional only for non v3*, right?

patjlm: ah missed the `not` sorry

patjlm (approved):

PR #4190, opened 2024-03-12
add logging for image error

BumbleFeng:
this error would be swallowed by https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/saasherder/saasherder.py#L1351-L1356


jonmosco (approved):
LGTM

PR #4197, opened 2024-03-14
[terraform-resources] 5.x provider asg updates (backwards compatible)

maorfr:
a better version of #4196

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -5472,7 +5472,7 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
         tags = [
             {"key": k, "value": v, "propagate_at_launch": True} for k, v in tags.items()
         ]
-        if self.versions.get(account).startswith("3"):
+        if self.versions.get(account, "").startswith("3"):
             asg_value["tags"] = tags
         else:
             asg_value["tag"] = tags


reconcile/utils/terrascript_aws_client.py
```diff
@@ -5469,9 +5469,13 @@ def populate_tf_resource_asg(self, spec) -> None:
...
             )
 
-        asg_value["tag"] = [
+        tags = [
             {"key": k, "value": v, "propagate_at_launch": True} for k, v in tags.items()
         ]
+        if self.versions.get(account).startswith("3"):
+            asg_value["tags"] = tags
+        else:
+            asg_value["tag"] = tags
```
hemslo, line 5478: I thought both versions support `tag` in the same format, no?

maorfr: both are supported, but i'm trying to perform a migration that has `No changes. Your infrastructure matches the configuration.` in every step of the way

reconcile/utils/terrascript_aws_client.py
```diff
@@ -5469,9 +5469,13 @@ def populate_tf_resource_asg(self, spec) -> None:
                 override
             )
 
-        asg_value["tag"] = [
+        tags = [
             {"key": k, "value": v, "propagate_at_launch": True} for k, v in tags.items()
         ]
+        if self.versions.get(account, "").startswith("3"):
+            asg_value["tags"] = tags
```
geoberle, line 5476: it seems asg 3.x also support `tag` and `tags` but it seems `tags` is something they call interpolated tags 

https://registry.terraform.io/providers/hashicorp/aws/3.73.0/docs/resources/autoscaling_group#interpolated-tags

geoberle: ah just saw the `No changes` idea. makes sense

geoberle (approved):

hemslo (approved):
LGTM!

PR #4208, opened 2024-03-18
Run terraform init in sequence

hemslo:
Avoid issues when multiple terraform init try to populate plugin cache at the same time.

https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache

> Note: The plugin cache directory is not guaranteed to be concurrency safe. The provider installer's behavior in environments with multiple terraform init calls is undefined.

### DIFF
--- a/reconcile/utils/terraform_client.py
+++ b/reconcile/utils/terraform_client.py
@@ -142,7 +142,8 @@ class TerraformClient:  # pylint: disable=too-many-public-methods
             TerraformSpec(name=name, working_dir=wd)
             for name, wd in self.working_dirs.items()
         ]
-        threaded.run(self.terraform_init, self.specs, self.thread_pool_size)
+        for spec in self.specs:
+            self.terraform_init(spec)
 
     @contextmanager
     def _terraform_log_file(


reconcile/utils/terraform_client.py
```diff
@@ -142,7 +142,8 @@ def init_specs(self):
             TerraformSpec(name=name, working_dir=wd)
             for name, wd in self.working_dirs.items()
         ]
-        threaded.run(self.terraform_init, self.specs, self.thread_pool_size)
+        for spec in self.specs:
+            self.terraform_init(spec)
```
patjlm, line 146: would it make sense to have a dedicated working dir for each `spec`, avoiding any conflict?

hemslo: currently each working dir should already be uniq, generated here https://github.com/app-sre/qontract-reconcile/blob/1b57e81edae9e5d8dd1add675218b744e4fa3bcd/reconcile/utils/terrascript_aws_client.py#L4009

jfchevrette (approved):
LGTM

PR #4209, opened 2024-03-18
[terraform-tgw-attachments] ignore deletion approvals in early exit

maorfr:
to avoid running this integration in pr-check or changes to, for example, deletion of external resources.

if nothing changes except for deletion approvals, the integration has nothing to do.


maorfr:
the deletion would fail without deletion approvals, so such MR would not get merged. right?

patjlm:
would it fail or just skip deletion with a gentle warning?

maorfr:
fail

patjlm (approved):

PR #4211, opened 2024-03-19
:zap: sentry: set default event level to CRITICAL

chassing:
Send only critical events to Glichtip to avoid unnecessary traffic.

Ticket: [APPSRE-9911](https://issues.redhat.com/browse/APPSRE-9911)

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -15,6 +15,7 @@ from typing import (
 
 import click
 import sentry_sdk
+from sentry_sdk.integrations.logging import LoggingIntegration
 
 from reconcile.status import (
     ExitCodes,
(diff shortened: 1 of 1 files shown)


reconcile/cli.py
```diff
@@ -66,7 +67,26 @@ def before_breadcrumb(crumb, hint):
...
+            raise ValueError(
+                "Invalid value for SENTRY_EVENT_LEVEL. Must be CRITICAL or ERROR."
+            )
+
+    sentry_sdk.init(
+        os.environ["SENTRY_DSN"],
+        before_breadcrumb=before_breadcrumb,
+        integrations=[
+            LoggingIntegration(
+                # send critical errors as events
```
hemslo, line 85: note the comment says `critical` but can be `error` depends on env.

chassing: Removed the misleading comment

PR #4214, opened 2024-03-19
[AUS] enable ingress gate to be auto approvable

geoberle:
add the `gate-ingress` version gate to the list of valid gates to be auto acknowledable by the AUS version-gate-approver

the implementation is mostly about letting the version-gate-approver integration know that such a gate exists and how to handle it if users decide that they want to auto ack it.

depends on https://github.com/app-sre/qontract-schemas/pull/609

part of https://issues.redhat.com/browse/APPSRE-7949


maorfr (approved):
so simple even i can review it

LGTM

PR #4219, opened 2024-03-20
[AUS] memory leaks with requests sessions

geoberle:
this PR makes sure several requests sessions in the AUS context are closed properly when they are not required anymore. 
* for the `OCMBaseClient` class, a context manager was implemented which takes care of that
* for prometheus queries, an explicit session is used as well in a context manager

local testing showed a constant FD count even after hours of running the integration.

part of https://issues.redhat.com/browse/APPSRE-9971

### DIFF
--- a/reconcile/utils/ocm_base_client.py
+++ b/reconcile/utils/ocm_base_client.py
@@ -3,10 +3,12 @@ from collections.abc import (
     Generator,
     Mapping,
 )
+from types import TracebackType
 from typing import (
     Any,
     Optional,
     Protocol,
+    Type,
 )
 
 from pydantic import BaseModel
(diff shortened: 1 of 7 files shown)


reconcile/utils/prometheus.py
```diff
@@ -61,13 +61,14 @@ class PrometheusHttpQuerier(BaseModel):
...
-            },
-        )
+        with requests.Session() as session:
+            response = session.get(
+                self.query_url,
+                params={"query": query},
+                headers={
+                    "Authorization": f"Bearer {self.auth_token}",
+                },
+            )
```
hemslo, line 71: `requests.get` is using `with sessions.Session() as session` internally, so this won't cause memory leak.

But if we want to reuse the same connection, we should manage session in a field or pass in as an argument, then if this request called multiple times in one run, we should see same fd reuse instead keep creating and close, that's a performance improvement task.

geoberle: reuse is and not worth the effort here. we talk about one request per reconcile run and OCM environment, so three for the time being.

i reverted this particular change

hemslo (approved):
LGTM!

reconcile/utils/ocm_base_client.py
```diff
@@ -148,9 +150,20 @@ def delete(self, api_path: str):
         r = self._session.delete(f"{self._url}{api_path}", timeout=REQUEST_TIMEOUT_SEC)
         r.raise_for_status()
 
-    def close(self):
+    def close(self) -> None:
         self._session.close()
 
+    def __enter__(self) -> "OCMBaseClient":
```
hemslo, line 156: can use new `Self` directly here

PR #4221, opened 2024-03-20
APPSRE-9922: Add visualization for reserved networks

lisa:
This depends on https://github.com/app-sre/qontract-schemas/pull/610 which has content that will change, and force this PR to change. However, the new schema has yet to be created, and so this PR will stand with https://github.com/app-sre/qontract-schemas/pull/610 (at the time of this PR's opening) until changed.


esron:
Can you add instructions on how to test this locally?

lisa:
> Can you add instructions on how to test this locally?

There are two options:

1. `qontract-cli --config /path/to/config.toml  get network-reservations` to display in a terminal-appropriate format
2. `qontract-cli --config /path/to/config.toml  get --output md network-reservations` to display in Markdown format

You will need to ensure the toml file is correct (including pointing to a valid graphql server with appropriate data).

esron (approved):

PR #4222, opened 2024-03-21
Enable writting Job logs to stdout

lechuk47:
Requirement for ERv2. It introduces the option to write job logs into stdout.

### DIFF
--- a/reconcile/utils/oc.py
+++ b/reconcile/utils/oc.py
@@ -14,13 +14,10 @@ from contextlib import suppress
 from dataclasses import dataclass
 from datetime import datetime
 from functools import wraps
+from io import TextIOWrapper
 from subprocess import Popen
 from threading import Lock
-from typing import (
-    Any,
-    Optional,
-    Union,
-)
+from typing import Any, Optional, Union
 
 import urllib3
 from kubernetes.client import (  # type: ignore[attr-defined]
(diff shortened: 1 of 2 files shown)


reconcile/utils/oc.py
```diff
@@ -725,15 +722,30 @@ def wait_for_job_running(self, namespace, name):
...
-        output_file = open(os.path.join(output, name), "w", encoding="locale")
+        if isinstance(output, TextIOWrapper):
+            output_file = output
+        else:
+            output_file = open(os.path.join(output, name), "w", encoding="locale")
         # collect logs to file async
-        Popen(self.oc_base_cmd + cmd, stdout=output_file)
+        p = Popen(self.oc_base_cmd + cmd, stdout=output_file)
+        if wait_for_logs_process:
+            p.wait()
```
hemslo, line 748: process is not captured if `wait_for_logs_process` is `False`, main thread may finish before subprocess, maybe unexpected leak. Better to collect them using `subprocess.run`. To avoid block main thread, can start this in a new thread/pool. See [APPSRE-7403](https://issues.redhat.com/browse/APPSRE-7403)

lechuk47: I'm just adding the option to get the logs to the stdout. I was unaware of that ticket but would prefer not to wait before this is fixed to use this feature. This can not be deprecated as is used by the JobController. 

Is it ok for you to add this change and tackle the refactor in APPSRE-7403?

hemslo: Then can we leave old code not changed but use `subprocess.run` for the new feature?

```python
args = self.oc_base_cmd + cmd
if wait_for_logs_process:
  subprocess.run(args, stdout=output_file)
else:
  # collect logs to file async
  Popen(args, stdout=output_file)
```

lechuk47: Done.

hemslo (approved):
LGTM!

PR #4224, opened 2024-03-22
[AUS] support new addon service

geoberle:
certain addon related OCM API endpoints have been moved to their own service. this PR introduces support for the new endpoints while keeping support for the old ones during the time of migration.

the most involved change is the rename of `upgrade_policies` to `upgrade_plans` with some minor nuances in field names.

a label on the OCM environment schema file controls which version of the API is to be used for that specific OCM environment.

```yaml
---
$schema: /openshift/openshift-cluster-manager-environment-1.yml
labels:
  feature_flag_addon_service_version: "$version"
```

use `v1` for the old APIs under `cluster_mgmt` and `v2` for the new API under `addons_mgmt`.

new API: https://gitlab.cee.redhat.com/ocm/ocm-addons-service/-/blob/main/openapi/ocm-addons-service.yaml

part of: https://issues.redhat.com/browse/APPSRE-9931

### DIFF
--- a/reconcile/utils/ocm/addons.py
+++ b/reconcile/utils/ocm/addons.py
@@ -6,46 +6,209 @@ from typing import (
 from reconcile.utils.ocm.base import OCMAddonInstallation
 from reconcile.utils.ocm_base_client import OCMBaseClient
 
+ADDON_UPGRADE_POLICY_DESIRED_KEYS = {
+    "id",
+    "addon_id",
+    "schedule_type",
+    "schedule",
+    "next_run",
+    "version",
+}
 
-def get_addons_for_cluster(
-    ocm_api: OCMBaseClient,
-    cluster_id: str,
-    addon_latest_versions: dict[str, str],
-    required_state: Optional[str],
-) -> list[OCMAddonInstallation]:
-    """
-    Returns a list of Addons installed on a cluster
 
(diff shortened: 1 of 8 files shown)


reconcile/utils/ocm/addons.py
```diff
@@ -6,46 +6,209 @@
...
-    for addon in ocm_api.get_paginated(
-        api_path=f"/api/clusters_mgmt/v1/clusters/{cluster_id}/addons",
-        params=params,
-    ):
-        current_version = addon["addon_version"]["id"]
-        latest_version = addon_latest_versions.get(addon["id"])
-        addon["addon_version"]["available_upgrades"] = (
-            [latest_version] if latest_version != current_version else []
+
+class AddonService:
```
janboll, line 19: Why not make `ocm_api` an instance field? It's used in a lot of places.

patjlm (approved):

reconcile/utils/ocm/addons.py
patjlm: if we have access to `ocm_api` during class instanciation, it would make sense indeed

geoberle: i'm going to delete all this code after all OCM environments are on v2. so i wanted to keep the method signatures the same as they were before.

PR #4225, opened 2024-03-22
[AUS] skip cluster upgrade scheduling on unacked gates

geoberle:
if a cluster has unacked version gates, the OCM API to create an upgrade policy will fail. this PR prevents that failure by not scheduling an upgrade in such situations.

follow up to https://github.com/app-sre/qontract-reconcile/pull/4103 which was incomplete

part of APPSRE-7949


patjlm (approved):
Tx

PR #4226, opened 2024-03-22
[AUS] bugfix in gate handling

geoberle:
the variable `gates` representing all the version gates that exist in an OCM environment was accidentally overwritten in a subsequent code branch with a shorter and potentially even empty list.

the condition that prevents cluster upgrades from happening when there are no gates defined started to act arbitrarily

### DIFF
--- a/reconcile/aus/base.py
+++ b/reconcile/aus/base.py
@@ -1097,7 +1097,7 @@ def calculate_diff(
                         gate.id for gate in gates_with_missing_agreements
                     ]
                     logging.info(
-                        f"[{spec.org.org_id}/{spec.org.name}/{spec.cluster.name}] found gates for {target_version_prefix} - {missing_gate_ids} "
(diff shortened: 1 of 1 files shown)


reconcile/aus/base.py
```diff
@@ -1078,23 +1079,25 @@ def set_mutex(
...
                 )
-                if gates:
-                    gate_ids = [gate.id for gate in gates]
+                if gates_with_missing_agreements:
+                    missing_gate_ids = [
+                        gate.id for gate in gates_with_missing_agreements
+                    ]
                     logging.info(
-                        f"[{spec.org.org_id}/{spec.org.name}/{spec.cluster.name}] found gates for {target_version_prefix} - {gate_ids} "
+                        f"[{spec.org.org_id}/{spec.org.name}/{spec.cluster.name}] found gates for {target_version_prefix} - {missing_gate_ids} "
```
patjlm, line 1100:
```suggestion
                        f"[{spec.org.org_id}/{spec.org.name}/{spec.cluster.name}] found gates with missing agreements for {target_version_prefix} - {missing_gate_ids} "
```

patjlm (approved):

PR #4229, opened 2024-03-26
Refactor MergeRequestManager, create base class in utils package

janboll:
[APPSRE-9976](https://issues.redhat.com/browse/APPSRE-9976)


reconcile/utils/merge_request_manager/merge_request_manager.py
```diff
@@ -0,0 +1,95 @@
...
+                logging.info(
+                    "Bad MR description format. Closing %s",
+                    mr.attributes.get("web_url", "NO_WEBURL"),
+                )
+                self._vcs.close_app_interface_mr(
+                    mr, "Closing this MR because of bad description format."
+                )
+                continue
+            self._open_mrs.append(OpenMergeRequest(raw=mr, mr_info=mr_info))
+        self._housekeeping_ran = True
```
janboll, line 95: Need to add check in AVS child as well

reconcile/utils/merge_request_manager/merge_request_manager.py
```diff
@@ -0,0 +1,95 @@
...
+                mr_info_dict.get(k) == expected_data.get(k)
+                for k in expected_data.keys()
+            ):
+                return mr
+
+        return None
+
+    def _fetch_avs_managed_open_merge_requests(self) -> list[ProjectMergeRequest]:
+        all_open_mrs = self._vcs.get_open_app_interface_merge_requests()
+        return [mr for mr in all_open_mrs if self._mr_label in mr.labels]
```
hemslo, line 49: method name shouldn't contain subclass name

reconcile/aws_version_sync/integration.py
```diff
@@ -338,7 +338,7 @@ def reconcile(
         external_resources_app_interface: AppInterfaceExternalResources,
     ) -> None:
         # initialize the merge request manager
-        merge_request_manager.fetch_avs_managed_open_merge_requests()
+        merge_request_manager._fetch_avs_managed_open_merge_requests()
```
hemslo, line 341: shouldn't call private method

reconcile/aws_version_sync/merge_request_manager/merge_request_manager.py
```diff
@@ -54,7 +48,7 @@ def process(self, gitlab_cli: GitLabApi) -> None:
         )
 
 
-class MergeRequestManager:
+class MergeRequestManager(MergeRequestManagerBase):
```
hemslo, line 51: The usage of `MergeRequestManager` not match inheritance. If use inheritance, `MergeRequestManagerBase` should define all interface method that be used in integration. Like this:

```python
class MergeRequestManagerBase:
    def housekeeping(self): ...

    @abstractmethod
    def expected_data(self):
        pass

    def _merge_request_already_exists(self, expected_data): ...

    def create_merge_request(self):
        data = self.expected_data()
        if self._merge_request_already_exists(data):
            ...
        ...
```

each subclass implement abstract method, integration should only use public interface of base class, not methods in subclass.

If there are too many different creation logic in subclass, the whole `create_merge_request` can be abstract, but in that case we should use composition over inheritance, change `is a` to `has a`, each `MergeRequestManager` has a `MergeRequestManagerUtil` do common tasks like check mr exist used in it's own create method like `create_avs_merge_request`.

reconcile/aws_version_sync/merge_request_manager/merge_request_manager.py
```diff
@@ -167,14 +80,15 @@ def create_avs_merge_request(
...
-        ):
-            if mr.avs_info.resource_engine_version == resource_engine_version:
+        if mr := self._merge_request_already_exists({
+            "provider": provider,
+            "account_id": provisioner_uid,
+            "resource_provider": resource_provider,
+            "resource_identifier": resource_identifier,
+            "resource_engine": resource_engine,
+        }):
+            assert isinstance(mr.mr_info, AVSInfo)
```
hemslo, line 90: can also use generic to avoid type cast

reconcile/aws_version_sync/integration.py
janboll: Good find, that line can go away

reconcile/aws_version_sync/merge_request_manager/merge_request_manager.py
janboll: made create merge request an abstract method

reconcile/templating/lib/merge_request_manager.py
```diff
@@ -211,8 +137,9 @@ def create_tr_merge_request(self, output: list[TemplateOutput]) -> None:
         collection_hash = collection_hashes.pop()
 
         """Create a new MR with the rendered template."""
-        if mr := self._merge_request_already_exists(collection):
-            if mr.template_info.collection_hash == collection_hash:
+        if mr := self._merge_request_already_exists({"collection": collection}):
+            assert isinstance(mr.mr_info, TemplateInfo)
```
chassing, line 141: I also don't like those type cast asserts :(

reconcile/aws_version_sync/merge_request_manager/merge_request_manager.py
janboll: will look at it

reconcile/templating/lib/merge_request_manager.py
janboll: will look at it

reconcile/utils/merge_request_manager/merge_request_manager.py
chassing: What does ^this mean?

chassing (approved):

reconcile/utils/merge_request_manager/merge_request_manager.py
```diff
@@ -0,0 +1,104 @@
...
+            except ParserError:
+                logging.info(
+                    "Bad MR description format. Closing %s",
+                    mr.attributes.get("web_url", "NO_WEBURL"),
+                )
+                self._vcs.close_app_interface_mr(
+                    mr, "Closing this MR because of bad description format."
+                )
+                continue
+            self._open_mrs.append(self._omr_klass(raw=mr, mr_info=mr_info))
```
hemslo, line 103: you don't need to pass `omr_klass` if use generic all the way down. Just use one template type var for the different part.

```python
T = TypeVar("T", bound=BaseModel)

@dataclass
class OpenMergeRequest(Generic[T]):
    raw: ProjectMergeRequest
    mr_info: T

class MergeRequestManagerBase(Generic[T]):
    """ """

    def __init__(self, vcs: VCS, parser: Parser[T], mr_label: str):
        self._vcs = vcs
        self._parser = parser
        self._mr_label = mr_label
        self._open_mrs: list[OpenMergeRequest[T]] = []
        self._open_mrs_with_problems: list[OpenMergeRequest[T]] = []
        self._housekeeping_ran = False
```

Then this line can be

```python
self._open_mrs.append(OpenMergeRequest[T](raw=mr, mr_info=mr_info))
```

no need to have another layer of

```python
class MRInfo(OpenMergeRequest):
    mr_info: AVSInfo
```

directly use 
```python
class MergeRequestManager(MergeRequestManagerBase[AVSInfo]):
```

janboll: Sure can do as well. I find this btw way more complicated than a single-line type assertion :-D

reconcile/utils/merge_request_manager/merge_request_manager.py
janboll: Please see:  f63f9b6

reconcile/utils/merge_request_manager/merge_request_manager.py
hemslo: it's all for mypy check, runtime won't have generic type info at all...

hemslo (approved):
LGTM!

PR #4232, opened 2024-03-28
:bug: ignore AWS accounts without a terraformState

chassing:
A newly created AWS account (`/aws/account-1.yml` via `aws-account-manager`) doesn't have a `terraformState` set; this will be done via another additional integration `terraform-init` later based on the account file (`/aws/account-1.yml`). All terraform related integrations should ignore such an account until a `terraformState` attribute is present.

Context: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/101355

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -583,7 +583,13 @@ def get_aws_accounts(
         ecrs=ecrs,
         cleanup=cleanup,
     )
-    return gqlapi.query(query)["accounts"]
+    accounts = gqlapi.query(query)["accounts"]
+    if terraform_state:
+        # a new account does not have a terraform state yet, ignore it until terraform-init does its job
+        return [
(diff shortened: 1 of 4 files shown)


reconcile/queries.py
```diff
@@ -583,7 +583,13 @@ def get_aws_accounts(
         ecrs=ecrs,
         cleanup=cleanup,
     )
-    return gqlapi.query(query)["accounts"]
+    accounts = gqlapi.query(query)["accounts"]
+    if terraform_state:
```
esron, line 587: What about integrations that use this query and don't care about tf state? I don't have any examples but was that considered?

maorfr (approved):

reconcile/queries.py
chassing: If the integration doesn't care about the `terraformState` then there isn't a problem :)

PR #4237, opened 2024-04-01
[CLI]Add cli option to get a cidr block for a new cluster.

OliviaHY:
Building on top of `qontract-cli get cidr-blocks`, with this we can get the latest available CIDR block for a new cluster. This can server as a building block for later automation of one yaml cluster provisioning.  Example: `qontract-cli --config config.debug.toml get cidr-blocks --for-cluster True --mask 24` Output:
```
INFO: Latest available network address: 10.170.71.0
INFO: You are reserving 256 network addresses.

You can use: 10.170.71.0/24
```

Note that this is only for cluster but not AWS resource VPC for now.

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -1025,19 +1025,19 @@ def network_reservations(ctx) -> None:
     "--managed",
     help="Only consider managed network, i.e. network reserved with a network-1 yaml in App Interface",
     type=bool,
-    default=False
+    default=False,
 )
 @click.option(
     "--for-cluster",
     help="Get the latest available CIDR block for new cluster. A decimal number between 1~32.",
     type=bool,
-    default=False
+    default=False,
 )
 @click.option(
     "--mask",
     help="Mask for the latest available CIDR block for AWS resources. A decimal number between 1~32.",
     type=int,
-    default=24
+    default=24,
 )
 @click.pass_context
 def cidr_blocks(ctx, managed: bool, for_cluster: int, mask: int) -> None:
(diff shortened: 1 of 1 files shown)


tools/qontract_cli.py
```diff
@@ -1070,8 +1088,25 @@ def cidr_blocks(ctx) -> None:
...
+        try:
+            result_cidr_block = str(ipaddress.ip_network((avail_addr, mask)))
+        except ValueError:
+            print(f"ERROR: Invalid CIDR Mask {mask} Provided.")
+            sys.exit(1)
+        print(f"INFO: You are reserving {str(2 ** (32 - mask))} network addresses.")
+        print(f"\nYou can use: {str(result_cidr_block)}")
+    if not for_cluster:
+        ctx.obj["options"]["sort"] = False
+        print_output(ctx.obj["options"], cidrs, columns)
```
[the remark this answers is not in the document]
OliviaHY: Good point. I removed the `managed` option for now. We can add it after migration.

tools/qontract_cli.py
```diff
@@ -1070,8 +1082,22 @@ def cidr_blocks(ctx) -> None:
 
     cidrs.sort(key=lambda item: ipaddress.ip_network(item["cidr"]))
 
-    ctx.obj["options"]["sort"] = False
-    print_output(ctx.obj["options"], cidrs, columns)
+    if for_cluster:
+        latest_declared_addr = ipaddress.ip_address(
+            [item for item in cidrs if item["type"] == "cluster"][-1]["to"]
+        )
```
hemslo, line 1088: this will throw error when there is no cluster cidr, better to find it explicitly then do None handling

```python
        latest_cluster_cidr = next(
            (item for item in reversed(cidrs) if item["type"] == "cluster"),
            None,
        )
```

OliviaHY: Updated and added None check

hemslo (approved):
LGTM!

tools/qontract_cli.py
```diff
@@ -1070,8 +1082,29 @@ def cidr_blocks(ctx) -> None:
...
+
+        if not latest_cluster_cidr:
+            print("ERROR: Unable to find any existing cluster CIDR block.")
+            sys.exit(1)
+
+        avail_addr = ipaddress.ip_address(latest_cluster_cidr["to"]) + 1
+
+        print(f"INFO: Latest available network address: {str(avail_addr)}")
+        try:
+            result_cidr_block = str(ipaddress.ip_network((avail_addr, mask)))
```
patjlm, line 1099: I think this should check if there is any overlap with AWS account VPCs as well. Since we use peerings we must not have overlaps with those as well

PR #4238, opened 2024-04-02
Appsre 8713 02: change desired state object

mehfuz:
related to https://issues.redhat.com/browse/APPSRE-8713

### DIFF
--- a/reconcile/deadmanssnitch.py
+++ b/reconcile/deadmanssnitch.py
@@ -130,12 +130,12 @@ class DeadMansSnitchIntegration(QontractReconcileIntegration[NoParams]):
                 except Exception as e:
                     errors.append(e)
         for cluster_name, diff_pair in diffs.identical.items():
-            if not dry_run:
-                if diff_pair.current.needs_vault_update():
-                    try:
-                        logging.info(
+            if diff_pair.current.needs_vault_update():
+                logging.info(
                             "[cluster_name:%s] [Action:update_vault]", cluster_name
                         )
+                if not dry_run:
+                    try:
                         self.write_snitch_to_vault(
                             cluster_name=cluster_name,
                             snitch_url=diff_pair.current.check_in_url,
--- a/reconcile/deadmanssnitch.py
+++ b/reconcile/deadmanssnitch.py
@@ -4,6 +4,8 @@ from typing import (
     cast,
 )
 
+from pydantic import BaseModel
+
 from reconcile.gql_definitions.common.clusters_with_dms import ClusterV1
 from reconcile.typed_queries.app_interface_deadmanssnitch_settings import (
     get_deadmanssnitch_settings,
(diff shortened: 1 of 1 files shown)


reconcile/deadmanssnitch.py
```diff
@@ -32,6 +34,17 @@
 SECRET_NOT_FOUND = "SECRET_NOT_FOUND"
 
 
+class ClusterFields(BaseModel):
+    """Class to hold values from cluster file and settings"""
+
+    prometheus_url: str
```
hemslo, line 40: Could use `SnitchSpec` to indicate it's desired, and the value of `prometheus_url` is no longer a url but a name already, right? Can use `name`.

reconcile/deadmanssnitch.py
```diff
@@ -147,41 +169,22 @@ def get_current_state(
...
         clusters: list[ClusterV1],
-        current_state: dict[str, Snitch],
         cluster_to_prometheus_mapping: dict[str, str],
-    ) -> dict[str, Snitch]:
+    ) -> dict[str, ClusterFields]:
         desired_state = {
-            cluster.name: self.map_desired_snitch_value(
-                cluster.name, current_state, cluster_to_prometheus_mapping
+            cluster.name: ClusterFields(
+                prometheus_url=cluster_to_prometheus_mapping.get(cluster.name),
```
hemslo, line 176: `cluster_to_prometheus_mapping` seems not needed here, as the url is from `cluster.prometheus_url.replace("https://", "")`, since we already have `cluster` here, we don't need to use `cluster_to_prometheus_mapping`. Can use a simple selector function `get_snitch_name(cluster)`.

And if you check the usage of `cluster_to_prometheus_mapping` in `get_current_state`, it's using reverse lookup, if create the reverse mapping directly

```python
snitch_name_to_cluster_name_mapping = {get_snitch_name(cluster): cluster.name for cluster in clusters}
```

Then you'll find it's easier to construct current state, no need to do O(N^2) iterate on snitches and clusters, can directly construct

```python
current_state = {
    cluster_name: self.add_vault_data(cluster.name, snitch, snitch_secret_path)
    for snitch in snitches
    if (cluster_name := snitch_name_to_cluster_name_mapping.get(snitch.name))
}

hemslo: https://github.com/app-sre/qontract-reconcile/pull/4238/files/d7c7fbc0e3669589f076bee03e7c769137c7bd9b..c1c19294fa3b76f8bd31cfe02ec5dd0e9f545016#diff-27ae2e4c7d059f24158bac19d459e906eb2890c078c3d79eb0205423f5ae5e6cR162-R171

here no need to build `snitches_with_cluster_mapping` first then build `current_state`, given the list of snitches and `snitch_name_to_cluster_name_mapping`, can build `current_state` directly.

hemslo (approved):
LGTM!

PR #4241, opened 2024-04-04
APPSRE-8713: fix vault writes with deadmanssnitch integration

mehfuz:
While testing it was found out that while writing data to vault...the integration is not upserting the data rather writing. This is not desired for the integration and lead to data loss.

Added changes to integration for reading/writing to vault only once if there is change in vault secret object

### DIFF
--- a/reconcile/deadmanssnitch.py
+++ b/reconcile/deadmanssnitch.py
@@ -19,12 +19,8 @@ from reconcile.utils.runtime.integration import (
     NoParams,
     QontractReconcileIntegration,
 )
-from reconcile.utils.secret_reader import (
-    SecretNotFound,
-)
 from reconcile.utils.semver_helper import make_semver
 from reconcile.utils.vault import (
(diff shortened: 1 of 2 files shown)


janboll (approved):

reconcile/deadmanssnitch.py
```diff
@@ -102,28 +84,31 @@ def create_snitch(
...
         deadmanssnitch_api: DeadMansSnitchApi,
+        vault_snitch_map: dict[str, str],
     ) -> None:
         diffs = diff_mappings(
             current=current_state,
             desired=desired_state,
             equal=lambda current, desired: current.name == desired.name,
         )
         errors = []
+        vault_snitch_map_copy = vault_snitch_map.copy()
```
hemslo, line 103: maybe give it a meaningful name like `desired_vault_snitch_map`

hemslo (approved):
LGTM!

PR #4243, opened 2024-04-05
Introduce persistence context

janboll:
Context holds the value of a file managed by a template collection.

This fixes a bug, where a collection updating a file multiple times using more then one template, could lead to missing updates. Reason was, that previously everytime the file was read from persistence, i.e. gitlab. Now the content is held in the cache of a context manager.

Fixes: [APPSRE-10033](https://issues.redhat.com/browse/APPSRE-10033)

### DIFF
--- a/reconcile/templating/renderer.py
+++ b/reconcile/templating/renderer.py
@@ -5,7 +5,7 @@ import os
 from abc import ABC, abstractmethod
 from collections.abc import Callable
 from datetime import datetime, timedelta
-from typing import Any, Optional
+from typing import Any, Optional, Self
 
 import gitlab
 from ruamel import yaml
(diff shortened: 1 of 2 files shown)


reconcile/templating/renderer.py
```diff
@@ -40,8 +40,7 @@
 
 QONTRACT_INTEGRATION = "template-renderer"
 
-# Renders templates again to detect drift after one day
-CACHE_TTL_MINUTES = 24 * 60
+CACHE_TTL_MINUTES = 12
```
hemslo, line 43: Should this be a setting config?

reconcile/templating/renderer.py
```diff
@@ -107,6 +106,31 @@ def read(self, path: str) -> Optional[str]:
         return current
 
 
+class PersistenceContext(FilePersistence):
+    def __init__(self, persistence: FilePersistence, dry_run: bool) -> None:
```
hemslo, line 110: A little strange to both inherit `FilePersistence` and pass in another `FilePersistence`, maybe call it something like `TransactionPersistence` and add some docstring?

janboll: Yeah, `TransactionablePersistence` was my first take as well. I can rename it.

janboll:
Going to iterate over this again, to replace gitlab persistence with local git clone persistence. Should speed up and fix the need for caching as well and remove the need for state management.

PR #4245, opened 2024-04-07
[qc/get/osd-component-versions] add more information

maorfr:
- replace ref with commit sha from SAPM
- add deployment state


chassing (approved):

patjlm:
This is pretty cool, thanks. LGTM.
One thing only: we'll get info only for publisher targets, right? Any target that is not publishing on any channel will show `ref` and `unknown`.

PR #4249, opened 2024-04-08
Template Renderer use git clone

janboll:
This introduces ClonedRepoGitlabPersistence, which uses local directory for read and gitlab api for writes. 

This drastically reduces API calls and speeds up the process overall. It also fixes a bug by introducing the persistence context.

### DIFF
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1973,13 +1973,12 @@ def template_validator(ctx):
     envvar="APP_INTERFACE_DATA_PATH",
 )
 @click.option(
-    "--git-clone-path",
+    "--clone-repo",
     help="Path to a folder app-interface repo should be cloned to. Use this for regular integration run.",
-    required=False,
-    envvar="GIT_CLONE_PATH",
+    default=False,
 )
 @click.pass_context
-def template_renderer(ctx, app_interface_data_path, git_clone_path):
+def template_renderer(ctx, app_interface_data_path, clone_repo):
     from reconcile.templating.renderer import (
         TemplateRendererIntegration,
         TemplateRendererIntegrationParams,
@@ -1989,7 +1988,7 @@ def template_renderer(ctx, app_interface_data_path, git_clone_path):
         integration=TemplateRendererIntegration(
             TemplateRendererIntegrationParams(
                 app_interface_data_path=app_interface_data_path,
-                git_clone_path=git_clone_path,
+                clone_repo=clone_repo,
             )
         ),
         ctx=ctx.obj,
--- a/reconcile/cli.py
+++ b/reconcile/cli.py
@@ -1968,12 +1968,18 @@ def template_validator(ctx):
 @integration.command(short_help="Render datafile templates in app-interface.")
 @click.option(
     "--app-interface-data-path",
-    help="Path to app-interface dictory, used to write output and calculate diff in dry-run.",
+    help="Path to data dir in app-interface repo. Use this for local rendering or in MR checks.",
     required=False,
     envvar="APP_INTERFACE_DATA_PATH",
 )
+@click.option(
+    "--git-clone-path",
+    help="Path to a folder app-interface repo should be cloned to. Use this for regular integration run.",
+    required=False,
+    envvar="GIT_CLONE_PATH",
+)
 @click.pass_context
-def template_renderer(ctx, app_interface_data_path):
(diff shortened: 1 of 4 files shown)


reconcile/utils/git.py
```diff
@@ -6,9 +6,13 @@ class GitError(Exception):
     pass
 
 
-def clone(repo_url, wd):
+def clone(repo_url, wd, depth=None):
```
chassing, line 9:
```suggestion
def clone(repo_url, wd, depth=None):
    cmd = ["git", "clone"]
    if depth:
        cmd += ["--depth", str(depth)]
    cmd += [repo_url, wd]
```
something like this, would be easier to read

reconcile/templating/renderer.py
```diff
@@ -256,10 +272,17 @@ def run(self, dry_run: bool) -> None:
                 vcs=vcs,
                 parser=create_parser(),
             )
-            persistence = GitlabFilePersistence(vcs, merge_request_manager)
+            url = get_app_interface_repo_url()
 
-            state = init_state(QONTRACT_INTEGRATION, self.secret_reader)
+            with tempfile.TemporaryDirectory(
+                prefix=join_path(self.params.git_clone_path, "clone"),
```
chassing, line 278: `prefix` doesn't change the destination directory, it'll be still `$TMP`. The correct parameter is `dir`

reconcile/templating/renderer.py
```diff
@@ -101,10 +125,15 @@ def write(self, outputs: list[TemplateOutput]) -> None:
 
     def read(self, path: str) -> Optional[str]:
         try:
-            current = self.vcs.get_file_content_from_app_interface_master(path)
-        except gitlab.exceptions.GitlabGetError:
-            return None
-        return current
+            with open(
```
chassing, line 128: Why not use `pathlib.Path` here?

reconcile/templating/renderer.py
hemslo: since we only need one tmp dir, can use integration name as prefix, like `QONTRACT_INTEGRATION`

reconcile/utils/git.py
```diff
@@ -6,9 +6,13 @@ class GitError(Exception):
...
+    base_cmd = ["git", "clone"]
+    if depth:
+        base_cmd.extend(["--depth", str(depth)])
+    base_cmd.append(repo_url)
+    cmd = base_cmd + [wd]
     # pylint: disable=subprocess-run-check
-    cmd = ["git", "clone", repo_url, wd]
     result = subprocess.run(
         cmd, cwd=wd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False
     )
```
hemslo, line 18: since there is `check=False` already, no need to disable pylint. And since we don't check output, we can use high level args

```python
result = subprocess.run(cmd, cwd=wd, capture_output=False, check=False)
```

reconcile/templating/renderer.py
```diff
@@ -90,8 +88,34 @@ def read(self, path: str) -> Optional[str]:
         return None
 
 
-class GitlabFilePersistence(FilePersistence):
-    def __init__(self, vcs: VCS, mr_manager: MergeRequestManager) -> None:
+class PersistenceContext(FilePersistence):
```
hemslo, line 91: same as previous pr, use something like `transaction` or docstring to indicate this will persistent at the end.

reconcile/cli.py
```diff
@@ -1968,12 +1968,18 @@ def template_validator(ctx):
 @integration.command(short_help="Render datafile templates in app-interface.")
 @click.option(
     "--app-interface-data-path",
-    help="Path to app-interface dictory, used to write output and calculate diff in dry-run.",
+    help="Path to data dir in app-interface repo. Use this for local rendering or in MR checks.",
     required=False,
     envvar="APP_INTERFACE_DATA_PATH",
 )
+@click.option(
+    "--git-clone-path",
```
hemslo, line 1976: I thought this is the git url of the repo, but later I found that is read from settings, then why do we need a path here since we always use a new temp dir? Maybe change it to a flag to indicate whether use local or do a git clone.

janboll: Yes, that was about the last reason I had for this parameter. You are right, I'll change it

hemslo (approved):
LGTM!

PR #4250, opened 2024-04-09
Fix bug in prometheus rules tester

janboll:
Passed in cluster_name is already an iterable, do not convert it to list anymore, so the namespace filter can use it correctly

[APPSRE-10066](https://issues.redhat.com/browse/APPSRE-10066)

### DIFF
--- a/reconcile/prometheus_rules_tester/integration.py
+++ b/reconcile/prometheus_rules_tester/integration.py
@@ -122,12 +122,12 @@ def fetch_rule_and_tests(
 def get_rules_and_tests(
     vault_settings: AppInterfaceSettingsV1,
     thread_pool_size: int,
-    cluster_name: Optional[tuple[str]] = None,
+    cluster_names: Optional[tuple[str]] = None,
 ) -> list[Test]:
     """Iterates through all namespaces and returns a list of tests to run"""
     namespace_with_prom_rules, _ = orb.get_namespaces(
         PROVIDERS,
-        cluster_names=cluster_name if cluster_name else [],
+        cluster_names=cluster_names if cluster_names else [],
         namespace_name=NAMESPACE_NAME,
     )
 
@@ -225,14 +225,14 @@ def check_rules_and_tests(
     vault_settings: AppInterfaceSettingsV1,
     alerting_services: Iterable[str],
     thread_pool_size: int,
-    cluster_name: Optional[tuple[str]] = None,
+    cluster_names: Optional[tuple[str]] = None,
 ) -> list[Test]:
     """Fetch rules and associated tests, run checks on rules and tests if they exist
     and return a list of failed checks/tests"""
     tests = get_rules_and_tests(
         vault_settings=vault_settings,
         thread_pool_size=thread_pool_size,
-        cluster_name=cluster_name,
+        cluster_names=cluster_names,
     )
     threaded.run(
         func=run_test,
@@ -254,7 +254,7 @@ def run(
     orb.QONTRACT_INTEGRATION_VERSION = QONTRACT_INTEGRATION_VERSION
 
     failed_tests = check_rules_and_tests(
-        cluster_name=cluster_names,
+        cluster_names=cluster_names,
         vault_settings=get_app_interface_vault_settings(),
         alerting_services=get_alerting_services(),
         thread_pool_size=thread_pool_size,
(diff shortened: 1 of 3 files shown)


reconcile/prometheus_rules_tester/integration.py
```diff
@@ -247,7 +247,7 @@ def check_rules_and_tests(
 
 
 def run(
-    dry_run: bool, thread_pool_size: int, cluster_name: Optional[str] = None
+    dry_run: bool, thread_pool_size: int, cluster_name: Optional[tuple[str]] = None
```
hemslo, line 250: should we also rename the parameter from `cluster_name` to `cluster_names` since it's not a single str anymore?

reconcile/prometheus_rules_tester/integration.py
```diff
@@ -136,6 +137,15 @@ def get_rules_and_tests(
...
         ]
+        prom_rules += [
+            r
+            for r in list(
+                itertools.chain.from_iterable([
+                    s["openshiftResources"] for s in namespace["sharedResources"]
+                ])
+            )
+            if r["provider"] in PROVIDERS
+        ]
```
rporres, line 148: I don't understand why this is needed... `orb.get_namespaces` should take care of shared resources, right?

hemslo: same question, it should be done in https://github.com/app-sre/qontract-reconcile/blob/d26fb63665718895913b6e4ad5ce0067155ece95/reconcile/openshift_resources_base.py#L805 already, or any edge cases missing for this?

reconcile/prometheus_rules_tester/integration.py
hemslo: also need to rename all later usages like https://github.com/app-sre/qontract-reconcile/pull/4250/files#diff-edd71a28997929cdbab0eda4cd50dc7c10880c200a93035d621c241dc8a4b18aR238

reconcile/prometheus_rules_tester/integration.py
janboll: You are right. perhaps I was to deep into it and got confused :see_no_evil: . Let me check again

janboll: Yes, it's not needed

reconcile/prometheus_rules_tester/integration.py
janboll: done

reconcile/prometheus_rules_tester/integration.py
```diff
@@ -122,12 +122,12 @@ def fetch_rule_and_tests(
 def get_rules_and_tests(
     vault_settings: AppInterfaceSettingsV1,
     thread_pool_size: int,
-    cluster_name: Optional[str] = None,
+    cluster_names: Optional[tuple[str]] = None,
```
chassing, line 125: Please use `Iterable` instead of `tuple` here.

reconcile/prometheus_rules_tester/integration.py
```diff
@@ -225,14 +225,14 @@ def check_rules_and_tests(
     vault_settings: AppInterfaceSettingsV1,
     alerting_services: Iterable[str],
     thread_pool_size: int,
-    cluster_name: Optional[str] = None,
+    cluster_names: Optional[tuple[str]] = None,
```
chassing, line 228: Same as above

reconcile/prometheus_rules_tester/integration.py
```diff
@@ -247,14 +247,14 @@ def check_rules_and_tests(
 
 
 def run(
-    dry_run: bool, thread_pool_size: int, cluster_name: Optional[str] = None
+    dry_run: bool, thread_pool_size: int, cluster_names: Optional[tuple[str]] = None
```
chassing, line 250: `Iterable` again

reconcile/prometheus_rules_tester/integration.py
janboll: Hm, but I think click will always create a tuple :thinking:

chassing: `tuple` is iterable :) Using `Iterable` instead of `tuple` would mean that you can use other iterable types as well, e.g. `list.´`

chassing (approved):
LGTM

PR #4251, opened 2024-04-09
Make tmp an emptydir

janboll:
`/tmp` is part of overlay fs right now. However it is used by i.e. terraform-resources and will be used by template-renderer soon for writig temporary data. Using an OverlayFS for writes can cause significant performance issues. Thus, create a separate mount for it using emptyDir

### DIFF
--- a/helm/qontract-reconcile/templates/template.yaml
+++ b/helm/qontract-reconcile/templates/template.yaml
@@ -416,6 +416,8 @@ objects:
             {{- end }}
           {{- end }}
           volumeMounts:
+          - name: tmp-dir
+            mountPath: /tmp
           - name: qontract-reconcile-toml
             mountPath: /config
           {{- if $integration.cache }}
@@ -485,8 +487,6 @@ objects:
               memory: 120Mi
           {{- end }}
           volumeMounts:
-          - name: tmp-dir
-            mountPath: /tmp
           - name: logs
             mountPath: /fluentd/log/
           - name: fluentd-config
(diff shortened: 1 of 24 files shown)


helm/qontract-reconcile/templates/template.yaml
```diff
@@ -521,6 +523,8 @@ objects:
           emptyDir: {}
         - name: fluentd-config
           emptyDir: {}
+        - name: tmp-dir
+          emptyDir: { }
```
rporres, line 527: how about adding a sizeLimit? Just in case

hemslo: we can expand `resources` section above to let each integration config `ephemeral-storage`, https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#setting-requests-and-limits-for-local-ephemeral-storage

janboll: Right now it's unrestricted as well. Why not add it when it gets a problem.

hemslo: I think that can be done in a new change as we usually don't set it even using emptyDir in many places.

hemslo (approved):
LGTM!

helm/qontract-reconcile/templates/template.yaml
rporres: why keeping something that we know it's wrong... 😂 

I agree it's much better to write in emptyDir than in the overlay, so this change is good. Since we have the possibility of restricting the size of the emptyDir, it is good to restrict it too in case something gets crazy.

PR #4253, opened 2024-04-10
terraform-init integration

chassing:
This PR introduces a new integration (`terraform-init`) that prepares a new AWS account for Terraform usage. Basically, it creates an S3 bucket and all needed app-interface settings in `/aws/account-1.yml` via template collection. This is the 2nd and last part of the new AWS account management.

Ticket: [APPSRE-9982](https://issues.redhat.com/browse/APPSRE-9982)
Design doc: [AppSRE AWS account management](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/appsre-aws-account-management.md)

Hints for the reviewers:

* The best would be to review this PR commit by commit.


reconcile/terraform_init/merge_request_manager.py
```diff
@@ -0,0 +1,102 @@
...
+            logging.info("MR already exists for %s", data.account)
+            return None
+
+        try:
+            self._vcs.get_file_content_from_app_interface_master(file_path=data.path)
+            # the file exists, nothing to do
+            return None
+        except GitlabGetError as e:
+            if e.response_code != 404:
+                raise e
```
hemslo, line 86: can just use `raise` to preserve the full stack trace.

reconcile/terraform_init/integration.py
```diff
@@ -0,0 +1,165 @@
...
+        state_collection: str,
+        bucket_name: str,
+        account: AWSAccountV1,
+    ) -> None:
+        logging.info("Creating bucket '%s' for account '%s'", bucket_name, account.name)
+        if not dry_run:
+            # the creation of the bucket is idempotent
+            account_aws_api.s3.create_bucket(
+                name=bucket_name, region=account.resources_default_region
+            )
```
hemslo, line 103: in previous setup, there is a bucket policy attached, is that no needed or we do it in another step?

chassing: We don't need this extra policy. The new bucket is owned by the creator, which is the future terraform user. In the previous setup, the root user owns the bucket, therefore an extra policy is needed.

reconcile/terraform_init/merge_request_manager.py
chassing: done

reconcile/terraform_init/integration.py
hemslo: I think that extra policy is used to limit access to that bucket to terraform user only, like even I use SSO login to that aws account, I can't do anything to that bucket including list objects.

chassing: Do we need this type of protection? Why should we limit the admin users?

hemslo: I don't know, quote the [doc](https://gitlab.cee.redhat.com/service/app-interface/-/tree/master/docs/aws/terraform)

>The s3 backend is used to share state and is set up with a bucket policy allowing strict access as the state files have highly sensitive content including keys.

chassing: Ok, I'll leave it as it is. Otherwise, we would need to use the root user (which we may not have access to) to operate on those buckets.

reconcile/utils/aws_api_typed/s3.py
```diff
@@ -0,0 +1,25 @@
...
+
+class AWSApiS3:
+    def __init__(self, client: S3Client) -> None:
+        self.client = client
+
+    def create_bucket(self, name: str, region: str) -> str:
+        """Create an S3 bucket without any ACLs therefore the creator will be the owner and returns the ARN."""
+        bucket_kwargs = {}
+        if region != "us-east-1":
+            # you can't specify the location if it's us-east-1 :(
```
hemslo, line 18: maybe attach the doc link here https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateBucketConfiguration.html, as this is an unusual logic.

hemslo (approved):
LGTM!

reconcile/utils/aws_api_typed/s3.py
chassing: done in [b3c93bf](https://github.com/app-sre/qontract-reconcile/pull/4253/commits/b3c93bf03e7ae4eb97516f2cc236574a37b49985)

reconcile/utils/aws_api_typed/s3.py
```diff
@@ -0,0 +1,26 @@
...
+            # see valid values "LocationConstraint" here: https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateBucketConfiguration.html
+            bucket_kwargs = {
+                "CreateBucketConfiguration": {
+                    "LocationConstraint": region,
+                },
+            }
+        self.client.create_bucket(Bucket=name, **bucket_kwargs)  # type: ignore
```
patjlm, line 25: so here we'll fail silently if the previous reconciliation loop created the bucket and sent the merge request, but it's not merged yet. Is that correct? Or do we raise an exception until the MR is merged?

What if the bucket creation fails for an other reason? (eg name used by any other AWS customer..)?

chassing: Unfortunately or fortunately, boto3 `create_bucket` doesn't raise an exception if the bucket already exists, so it'll silently "fail" which is ok. If the bucket already exists, but for another AWS customer ((which is unlikely), we'll hit that error with the first usage of `terraform`. I think we can live with that.

PR #4254, opened 2024-04-10
Improvements to the SAPM S3 publisher

jmelis:
This PR brings improves several things in the SAPM S3 exporter:

- `app_name` is provided as part of the node. This is necessary because integrations using the publisher data will filter by `app_name`
- Information about whether or not a job is a test job is codified via `publish_job_logs`.
- If target is not defined, it defaults to `None` instead of default. This is more in line with other integrations we have.
- And more importantly, now all nodes that publish will be in the output, which is a changed compared with the current state, where only those that publish AND are being subscribed to by other nodes do appear.

### DIFF
--- a/reconcile/saas_auto_promotions_manager/utils/saas_files_inventory.py
+++ b/reconcile/saas_auto_promotions_manager/utils/saas_files_inventory.py
@@ -34,6 +34,11 @@ class SaasFilesInventory:
         self._assemble_publishers()
         self._remove_unsupported()
 
+
+    @property
+    def publishers_with_subscribers(self) -> list[Publisher]:
+        return [p for p in self.publishers if p.has_subscriber]
+
     def _assemble_publishers(self) -> None:
         for saas_file in self._saas_files:
             for resource_template in saas_file.resource_templates:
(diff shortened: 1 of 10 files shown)


jmelis:
@fishi0x01 let me know if you prefer the PR to be split into separate ones

jmelis:
To further understand why these changes are needed, it may be interesting to look at https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/102423 (which depends on this MR)

reconcile/saas_auto_promotions_manager/utils/saas_files_inventory.py
```diff
@@ -34,6 +34,11 @@ def __init__(self, saas_files: Iterable[SaasFile]):
         self._assemble_publishers()
         self._remove_unsupported()
 
+
+    @property
+    def publishers_with_subscribers(self) -> list[Publisher]:
```
jmelis, line 39: @fishi0x01 note that I added this to keep tests adequate.

This changes things then, and I'm not sure I understand the consequences. Before this PR `publishers` report nodes that have subscribers. After this PR is `publishers` will report all nodes that publish, and you need to specifically access `publishers_with_subscribers` if you want to get to the previous behaviour.

Alternatively we could reverse it, i.e. keep `publishers` as it was, add `publishers_without_subscribers`, and leverage that one in the S3 exporter.

What do you prefer?

fishi0x01 (approved):
LGTM

PR #4258, opened 2024-04-10
Fix prom test cli, iterate over sharedResources as well

janboll:
Some alerts/rules were moved to shared resources. CLI needs to check these as well. Hence adding them to the list of prometheus rules to consider.


cubismod (approved):
LGTM

rporres:
I don't understand why this is needed... `orb.get_namespaces` should take care of shared resources, correct?

PR #4262, opened 2024-04-11
 template-validator: log diffs in PR checks

chassing:
Show the diffs in MR/PR checks. This is much more handy and convenient for regular users.

### DIFF
--- a/reconcile/templating/validator.py
+++ b/reconcile/templating/validator.py
@@ -127,7 +127,7 @@ class TemplateValidatorIntegration(QontractReconcileIntegration):
         if diffs:
             for diff in diffs:
                 logging.error(f"template: {diff.template}, test: {diff.test}")
-                logging.debug(diff.diff)
+                logging.error(diff.diff)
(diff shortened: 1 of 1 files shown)


reconcile/templating/validator.py
```diff
@@ -127,7 +127,7 @@ def run(self, dry_run: bool) -> None:
         if diffs:
             for diff in diffs:
                 logging.error(f"template: {diff.template}, test: {diff.test}")
-                logging.debug(diff.diff)
+                logging.error(diff.diff)
```
janboll, line 130: I removed this, cause you can easily leak secrets with this. (usage of vault reader function in jinja)

janboll (changes requested):

reconcile/templating/validator.py
janboll: I'll add a comment, to explain this in the code section.

chassing: I see

PR #4264, opened 2024-04-11
unleash-feature-toggles integration

chassing:
This PR introduces the new `unleash-toggles` integration that manages Unleash feature toggles.

The integration creates, updates, and deletes feature toggles but doesn't manage their actual state!

Depends: [schemas](https://github.com/app-sre/qontract-schemas/pull/650)
Ticket: [APPSRE-10072](https://issues.redhat.com/browse/APPSRE-10072)
EPIC: [SDE-3869](https://issues.redhat.com/browse/SDE-3869)
Design-doc: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/unleash-managed-toggles.md

Hints for the reviewers:

* Please review commit by commit


reconcile/unleash_feature_toggles/integration.py
```diff
@@ -0,0 +1,310 @@
...
+        current_projects: Iterable[Project],
+        desired_projects: Iterable[UnleashProjectV1],
+    ) -> None:
+        """Validate that all projects referenced in the desired state are actually exist."""
+        current_project_names = {p.name.lower() for p in current_projects}
+        desired_project_names = {p.name.lower() for p in desired_projects}
+        if missing_projects := desired_project_names - current_project_names:
+            for p in missing_projects:
+                logging.error(f"[{instance_name}] Project '{p}' does not exist!")
+            raise ValueError(f"Non-existing projects: {missing_projects}")
```
hemslo, line 119: here we are actually validating 2 cases, desired project missing and current project missing, how about do both in one validation, or just diff them and use later like this:

```python
                diff = diff_any_iterables(
                    current=projects,
                    desired=instance.projects or [],
                    current_key=lambda c: c.name.lower(),
                    desired_key=lambda d: d.name.lower(),
                    equal=lambda c, d: c.name.lower() == d.name.lower(),
                )
                if diff.add:
                    raise ValueError(f"missing projects: {','.join(diff.add.keys())}")
                if diff.delete:
                    raise ValueError(
                        f"not found projects: {','.join(diff.delete.keys())}"
                    )
                for pair in diff.identical.values():
                    self.reconcile(
                        client=client,
                        instance=instance,
                        project_id=pair.current.pk,
                        dry_run=dry_run,
                        current_state=pair.current.feature_toggles,
                        desired_state=pair.desired.feature_toggles or [],
                    )
```

reconcile/unleash_feature_toggles/integration.py
```diff
@@ -0,0 +1,310 @@
...
+            desired_key=lambda d: d.name,
+            equal=feature_toggle_equal,
+        )
+        for add in diff.add.values():
+            logging.info(
+                f"[{instance.name}/{project_id}] Adding feature toggle {add.name}"
+            )
+            try:
+                assert add.unleash.q_type  # make mypy happy
+                feature_type = FeatureToggleType[add.unleash.q_type]
```
hemslo, line 144: for kill switch type, `add.unleash.q_type` is `kill-switch`, this will fail as only `kill_switch` is accepted. `FeatureToggleType["kill-switch"]` will throw `KeyError: 'kill-switch'`

reconcile/unleash_feature_toggles/integration.py
```diff
@@ -0,0 +1,310 @@
...
+            except KeyError:
+                raise ValueError(
+                    f"[{instance.name}/{project_id}/{add.name}] Invalid feature toggle type '{add.unleash.q_type}', Possible values are: {', '.join(FeatureToggleType.__members__)}"
```
hemslo, line 147: this will print values `experiment, kill_switch, release, operational, permission`, while `kill_switch` should be `kill-switch`

reconcile/unleash_feature_toggles/integration.py
```diff
@@ -0,0 +1,310 @@
...
+                )
+                if not dry_run:
+                    client.delete_feature_toggle(
+                        project_id=project_id,
+                        name=delete.name,
+                    )
+            elif not instance.allow_unmanaged_feature_toggles:
+                raise UnleashFeatureToggleDeleteError(
+                    f"[{instance.name}/{project_id}] Found unmanaged feature toggles '{[d.name for d in diff.delete.values()]}'"
+                )
```
hemslo, line 194: this will show all `diff.delete.values()`, some of them may already included in the previous `if desired_toggle` branch.

reconcile/unleash_feature_toggles/integration.py
```diff
@@ -0,0 +1,310 @@
...
+        client: UnleashServer,
+        instance: UnleashInstanceV1,
+        project_id: str,
+        dry_run: bool,
+        current_state: Sequence[FeatureToggle],
+        desired_state: Iterable[FeatureToggleUnleashV1],
+        available_environments: Iterable[Environment],
+    ) -> None:
+        """Reconcile the feature toggle states."""
+        # Manage the actual feature toggle states
```
hemslo, line 207: I thought we don't manage toggle state, or this is something else?

reconcile/unleash_feature_toggles/integration.py
chassing: Wow, nice catch! 🙇 Changed in https://github.com/app-sre/qontract-schemas/pull/650/commits/3a03ce9a4ad8cccaf58cae378abea00ab7130b11

reconcile/unleash_feature_toggles/integration.py
chassing: Changed in https://github.com/app-sre/qontract-schemas/pull/650/commits/3a03ce9a4ad8cccaf58cae378abea00ab7130b11

reconcile/unleash_feature_toggles/integration.py
chassing: You can now manage the state as well, but it's optional.

reconcile/unleash_feature_toggles/integration.py
chassing: We'll hit that in a PR check, and only if someone deletes a managed toggle and there are unmanaged toggles and unmanaged toggles aren't allowed. This is an edge case and can be ignored.

reconcile/unleash_feature_toggles/integration.py
chassing: Changed in the last commit.

reconcile/unleash_feature_toggles/integration.py
hemslo: I haven't tested it in api call, does unleash api accept `kill-switch` or `kill_switch`

chassing: The Unleash API takes `kill-switch` and in app-interface we use `kill_switch`. I've tested it.

reconcile/unleash_feature_toggles/integration.py
hemslo: so when it's defined like 

```yaml
 environments:
    default: true
```

the integration will reset it to `on` if toggle off manually in unleash.

Another question is whether enable or disable project `environment` itself can affect toggle state?

hemslo: Screenshot 2024-05-16 at 4 16 23 PM
will this affect fetched toggle state value?

hemslo: btw to simplify state management, we can flatten all toggles and envs to make diff easier, like this:

```python
        current_toggle_states = {
            (state.name, env.name): env.enabled
            for state in current_state
            for env in state.environments
        }
        desired_toggle_states = {
            (state.name, env_name): enabled
            for state in desired_state
            for env_name, enabled in (state.unleash.environments or {}).items()
        }

        diff_result = diff_mappings(current=current_toggle_states, desired=desired_toggle_states)
        if diff_result.add:
            raise ValueError("missing xxxx")
        for (name, env), pair in diff_result.change.items():
            if not dry_run:
                client.set_feature_toggle_state(
                    project_id=project_id,
                    name=name,
                    environment=env,
                    enabled=pair.desired,
                )
```

chassing: No, disabled projects are ignored

hemslo: how is that ignored work? when fetch `features`, returned `environments` will auto exclude disabled ones?

reconcile/utils/unleash/server.py
```diff
@@ -0,0 +1,140 @@
...
+        projects = []
+        for project_data in self._list("/api/admin/projects", attribute="projects"):
+            project = Project(**project_data)
+            if include_feature_toggles:
+                project.feature_toggles = self.feature_toggles(project.pk)
+            projects.append(project)
+        return projects
+
+    def feature_toggles(self, project_id: str) -> list[FeatureToggle]:
+        """List all feature toggles for an project."""
```
hemslo, line 67: `an` -> `a`

reconcile/unleash_feature_toggles/integration.py
chassing: > how is that ignored work? when fetch features, returned environments will auto exclude disabled ones?

Exactly, disabled environments are not listed in the `/api/admin/projects/{project_id}/features` results.

chassing: > btw to simplify state management, ....

change in the latest commit

reconcile/utils/unleash/server.py
chassing: fixed

hemslo (approved):
LGTM!

PR #4265, opened 2024-04-11
:recycle: refactor api client

chassing:
Move parts of the `GlitchtipClient` into a separate base class. The future `UnleashServer` API client will reuse those base parts.


reconcile/utils/rest_api_base.py
```diff
@@ -0,0 +1,92 @@
...
+        host: str,
+        token: str,
+        max_retries: int | None = None,
+        read_timeout: float | None = None,
+    ) -> None:
+        self.host = host
+        self.token = token
+        self.max_retries = max_retries if max_retries is not None else 3
+        self.read_timeout = read_timeout if read_timeout is not None else 30
+        self._thread_local = threading.local()
```
hemslo, line 34: now we are making this as a base, let's fix the session issue to make it use the same session object and do auto close with context manager.

reconcile/utils/rest_api_base.py
```diff
@@ -0,0 +1,92 @@
...
+    * https://gitlab.com/glitchtip/glitchtip-backend/-/blob/master/glitchtip/pagination.py#L34
+    * https://requests.readthedocs.io/en/latest/api/?highlight=links#requests.Response.links
+    """
+    if links.get("next", {}).get("results", "false") == "true":
+        return links["next"]["url"]
+    return None
+
+
+class ApiBase:
+    """This class provides a common standard for REST API clients."""
```
hemslo, line 21: Do you want to make this as the base for all api clients we use, I have a similar one can benefit from the base, https://github.com/app-sre/qontract-reconcile/blob/master/tools/cli_commands/cost_report/cost_management_api.py, but it requires oauth token handling and customized retry (backoff and status)

chassing: I've added `session` and `auth` attributes to `__init__` to customize those things easily.

reconcile/utils/rest_api_base.py
chassing: Yes, you're right. Changed that in the last commit.

reconcile/utils/rest_api_base.py
hemslo: Thank you! I'll try to use the new base for cost  management api!

hemslo (approved):
LGTM!

PR #4273, opened 2024-04-16
:lock: ocm-aws-infrastructure-access: use federated users for accounts with SSO enabled

chassing:
Use federated user arns for accounts with SSO enabled.

Ticket: [APPSRE-9980](https://issues.redhat.com/browse/APPSRE-9980)


chassing:
Unfortunately, this doesn't work because

```
[ocm_base_client.py:post:123] - {"kind":"Error","id":"400","href":"/api/clusters_mgmt/v1/errors/400","code":"CLUSTERS-MGMT-400","reason":"Attribute 'user_arn' must be of format: 'arn:aws:iam::123456789012:user|group/name'","operation_id":"585cb993-db36-4612-89d2-54385452e8c4"}
```

PR #4274, opened 2024-04-16
APPSRE-10053 extend vcs to get diff

fishi0x01:
We need a way to calculate the number of commits between 2 given commits. We extend the VCS util to calculate that diff. This is vcs agnostic, i.e., the consumer doesn't concern itself on whether the underlying repo url is gitlab or github.

### DIFF
--- a/reconcile/utils/vcs.py
+++ b/reconcile/utils/vcs.py
@@ -31,7 +31,7 @@ class MRCheckStatus(Enum):
     RUNNING = 3
 
 
-@dataclass
+@dataclass(order=True)
 class Commit:
     repo: str
     sha: str
@@ -163,7 +163,7 @@ class VCS:
     ) -> str:
         if bool(self._is_commit_sha_regex.search(ref)):
             return ref
-        if "github.com" in repo_url:
+        if repo_url.startswith("https://github.com/"):
             github = self._init_github(repo_url=repo_url, auth_code=auth_code)
             return github.get_commit_sha(ref=ref)
         # assume gitlab by default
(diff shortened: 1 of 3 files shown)


reconcile/utils/vcs.py
```diff
@@ -157,6 +169,44 @@ def get_commit_sha(
...
+        commit_from: str,
+        commit_to: str,
+        auth_code: Optional[HasSecret],
+    ) -> list[Commit]:
+        """
+        Return a list of commits between two commits.
+        Note, that the commit_to is included in the result list, whereas commit_from is not included.
+        """
+        commits: list[Commit] = []
+        if "github.com" in repo_url:
```
hemslo, line 184: `in` check is a little tricky, what if my gitlab has something like `https://gitlab.com/xxx/test.github.com`, is our github repo url always starts with `https://github.com/`? Then we could just check

```python
        if repo_url.startswith("https://github.com/"):
```

reconcile/utils/vcs.py
```diff
@@ -157,6 +169,44 @@ def get_commit_sha(
...
+            github = self._init_github(repo_url=repo_url, auth_code=auth_code)
+            data = github.compare(commit_from=commit_from, commit_to=commit_to)
+            for gh_commit in data:
+                commits.append(
+                    Commit(
+                        repo=repo_url,
+                        sha=gh_commit.sha,
+                        date=gh_commit.commit.committer.date,
+                    )
+                )
```
hemslo, line 194: since `commits` can either be in `gitlab` branch or `github` branch, not both, to make it clear, we can skip outer `commits`, do early return here using list comprehension directly.

```python
            return [
                Commit(
                    repo=repo_url,
                    sha=gh_commit.sha,
                    date=gh_commit.commit.committer.date,
                )
                for gh_commit in data
            ]
```

same as `gitlab` section below.

reconcile/utils/vcs.py
fishi0x01: Good point! Fixed :+1:

reconcile/utils/vcs.py
fishi0x01: Adjusted :+1:

hemslo (approved):
LGTM!

PR #4275, opened 2024-04-16
[acs-policies] notifiers based on integrations.notifiers

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10043

design document: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/102231

this PR updates the `acs-policies` integration to use notifiers created by `acs-notifiers`.

### DIFF
--- a/reconcile/acs_policies.py
+++ b/reconcile/acs_policies.py
@@ -64,11 +64,11 @@ class AcsPoliciesIntegration(QontractReconcileIntegration[NoParams]):
             pc for c in gql_policy.conditions if (pc := self._build_policy_condition(c))
         ]
         jira_notifier = (
-            notifier_name_to_id[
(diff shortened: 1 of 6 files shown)


reconcile/acs_policies.py
```diff
@@ -3,6 +3,7 @@
 from typing import cast
 
 import reconcile.gql_definitions.acs.acs_policies as gql_acs_policies
+from reconcile.acs_notifiers import AcsNotifiersIntegration
```
chassing, line 6: TBH, I don't like it if one integration uses another integration. This is always hard to refactor. Do you mind and move the needed parts into a utils file? Especially, `AcsNotifiersIntegration._build_jira_notifier`, as a "private" method :(

chassing (approved):

PR #4276, opened 2024-04-16
Output must be written after each template

janboll:
Writes must be done after each template iteration.

Will look into improving tests for this.

### DIFF
--- a/reconcile/templating/renderer.py
+++ b/reconcile/templating/renderer.py
@@ -239,7 +239,6 @@ class TemplateRendererIntegration(QontractReconcileIntegration):
         gql_no_validation = init_from_config(validate_schemas=False)
 
         for c in get_template_collections():
-            outputs: list[TemplateOutput] = []
             variables = {}
             if c.variables:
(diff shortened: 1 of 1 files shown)


reconcile/templating/renderer.py
```diff
@@ -267,8 +267,8 @@ def reconcile(
                     if output:
                         outputs.append(output)
 
-                if not dry_run:
-                    p.write(outputs)
+                    if not dry_run:
+                        p.write(outputs)
```
hemslo, line 271: `outputs` is defined outside the loop, so this will write it multiple times like `[1]`, `[1,2]`, `[1,2,3]`, any issues with write the whole `outputs` once?

janboll: It doesn't matter, cause it's iterating. That's why it worked in manual test anyhow.
But great catch, the list can go away and `output` can be written directly.

I will add tests later today :+1:

hemslo (approved):
LGTM!

PR #4282, opened 2024-04-17
jira permissions validator: extended early exit

chassing:
Enhance `jira-permissions-validator` with [extended early exit feature](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/runbook/extended-early-exit.md) to avoid hitting Jira API rate limits too often. Additionally, cache some API replies in `jiraclient`.

Tickets:
* [APPSRE-10091](https://issues.redhat.com/browse/APPSRE-10091)
* [APPSRE-10096](https://issues.redhat.com/browse/APPSRE-10096)

### DIFF
--- a/reconcile/jira_permissions_validator.py
+++ b/reconcile/jira_permissions_validator.py
@@ -2,14 +2,11 @@ import logging
 import sys
 from collections.abc import Callable, Iterable
 from enum import IntFlag, auto
-from typing import Any
+from typing import Any, TypedDict
 
 from jira import JIRAError
 from pydantic import BaseModel
 
-from reconcile.gql_definitions.jira_permissions_validator.jira_boards_for_permissions_validator import (
-    DEFINITION as JIRA_BOARDS_DEFINITION,
-)
 from reconcile.gql_definitions.jira_permissions_validator.jira_boards_for_permissions_validator import (
     JiraBoardV1,
 )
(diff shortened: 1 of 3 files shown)


reconcile/utils/jira_client.py
```diff
@@ -82,6 +83,11 @@ def __init__(
         self.project = project
         self.jira = jira_api
 
+        # some caches
+        self.priorities = functools.lru_cache(maxsize=None)(self._priorities)
+        self.public_projects = functools.lru_cache(maxsize=None)(self._public_projects)
+        self.my_permissions = functools.lru_cache(maxsize=None)(self._my_permissions)
```
patjlm, line 89: :heart:

reconcile/jira_permissions_validator.py
```diff
@@ -256,13 +264,59 @@ def get_jira_boards(query_func: Callable) -> list[JiraBoardV1]:
...
+        secret_reader = create_secret_reader(use_vault=vault_settings.vault)
+
+        cache_source = CacheSource(
+            boards=export_boards(boards),
+        )
+        extended_early_exit_run(
+            integration=QONTRACT_INTEGRATION,
+            integration_version=QONTRACT_INTEGRATION_VERSION,
+            # don't use `dry_run` in the cache key because this is a read-only integration
+            dry_run=False,
```
hemslo, line 298: This means prod run and pr check run will share the same cache, no dry run cache pattern is `/<integration>/<integration_version>/no-dry-run(/<shard>)/latest`, is it ok to let pr check run set cache and cause prod run early exit?

chassing: That was exactly the idea. I want dry-runs and non-dry-runs using the same cached state

hemslo: I never tested this use case, let's see if this can cause any issues after deploy

hemslo (approved):
LGTM! Another better integration!

PR #4289, opened 2024-04-19
[tf-cloudflare-dns] fix shard detection in desired state

geoberle:
similar to tf-resources shard detection issue, the desired state of tf-cloudflare-dns has been restructured to dicts for safe shard detection

see also https://github.com/app-sre/qontract-reconcile/pull/4086


jfchevrette (approved):
LGTM

PR #4290, opened 2024-04-19
[APPSRE-10198] Use generateName instead of timestamp to create a unique PipelineRun name

BumbleFeng:
https://issues.redhat.com/browse/APPSRE-10198

Use generateName instead of timestamp to create a unique PipelineRun name. https://ibm.github.io/tekton-tutorial-openshift/lab1/6_create-pipeline-run/. And fix PipelineRun name too long to retrigger issue.

`
admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: Invalid resource name: length must be no more than 63 characters: metadata.name
`

### DIFF
--- a/reconcile/utils/openshift_resource.py
+++ b/reconcile/utils/openshift_resource.py
@@ -214,7 +214,11 @@ class OpenshiftResource:
 
     @property
     def name(self):
-        return self.body["metadata"]["name"]
+        # PipelineRun name can be empty when creating
+        if self.kind == "PipelineRun" and "name" not in self.body["metadata"]:
+            return self.body["metadata"]["generateName"][:-1]
+        else:
+            return self.body["metadata"]["name"]
 
     @property
     def kind(self):
--- a/reconcile/utils/openshift_resource.py
+++ b/reconcile/utils/openshift_resource.py
@@ -225,6 +225,10 @@ class OpenshiftResource:
         return self.body["kind"]
 
     @property
+    def annotations(self):
+        return self.body["metadata"].get("annotations", {})
+
+    @property
     def kind_and_group(self):
         return fully_qualified_kind(self.kind, self.body["apiVersion"])
 
--- a/reconcile/openshift_saas_deploy_trigger_base.py
+++ b/reconcile/openshift_saas_deploy_trigger_base.py
@@ -1,4 +1,3 @@
-import datetime
 import logging
 from collections.abc import Callable
 from threading import Lock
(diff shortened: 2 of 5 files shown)


geoberle:
i would love for this to work, since i had my fair share of name-too-long issues when retriggering. i'm a tiny bit worried though that saas-file name + env name might not be unique anymore when we trim more characters. is that a problem?

BumbleFeng:
> i would love for this to work, since i had my fair share of name-too-long issues when retriggering. i'm a tiny bit worried though that saas-file name + env name might not be unique anymore when we trim more characters. is that a problem?

Yes, it is a potential issue. It would fail PR-check when promoting, so we can fix it there. Another workaround is to shorten the timestamp length to buy some room.

BumbleFeng:
We do have a lot existing names need to fix https://ci.int.devshift.net/job/service-app-interface-gl-pr-check/362945/artifact/temp/reports/reconcile_reports_fail/reconcile-saas-file-validator.txt.

cubismod:
> We do have a lot existing names need to fix https://ci.int.devshift.net/job/service-app-interface-gl-pr-check/362945/artifact/temp/reports/reconcile_reports_fail/reconcile-saas-file-validator.txt.

Fixing those would require changing saas file `name` properties, right?

BumbleFeng:
> > We do have a lot existing names need to fix https://ci.int.devshift.net/job/service-app-interface-gl-pr-check/362945/artifact/temp/reports/reconcile_reports_fail/reconcile-saas-file-validator.txt.
> 
> Fixing those would require changing saas file `name` properties, right?

Yeah, I will try shorten timestamp length first to reduce the impact.

cubismod (approved):
LGTM!

cubismod (approved):
LGTMing the new changes. I use `generateName` for tf-repo and it works perfectly so I like this approach here

reconcile/utils/openshift_resource.py
```diff
@@ -214,7 +214,11 @@ def api_version_mutation(val1, val2):
 
     @property
     def name(self):
-        return self.body["metadata"]["name"]
+        # PipelineRun name can be empty when creating
+        if self.kind == "PipelineRun" and "name" not in self.body["metadata"]:
+            return self.body["metadata"]["generateName"][:-1]
+        else:
+            return self.body["metadata"]["name"]
```
hemslo, line 221: since this logic is same as https://github.com/app-sre/qontract-reconcile/pull/4290/files#diff-006493679cf46d27e100feeb8475e6e29987fe85cca4858a1f05608b2d536dc8R166-R173, should extract an util function to be used for both to avoid later inconsistent

geoberle:
nice!

geoberle (approved):

reconcile/utils/openshift_resource.py
BumbleFeng: PTAL https://github.com/app-sre/qontract-reconcile/pull/4290/commits/04491507f1e9e6feec020a0aa6446a4804d3fdf0

hemslo (approved):
LGTM!

rporres (approved):
nice!

patjlm:
:heart:

PR #4293, opened 2024-04-22
Template Renderer: Adding more tests to reconcile method

janboll:
Sligt refactoring, moving compute template hash out, makes it easier to test


reconcile/templating/renderer.py
```diff
@@ -175,6 +175,14 @@ def unpack_dynamic_variables(
     }
 
 
+def calc_template_hash(c: TemplateCollectionV1, variables: dict[str, Any]) -> str:
+    return hashlib.sha256(
+        "".join(sorted([str(t) for t in c.templates] + [json.dumps(variables)])).encode(
```
hemslo, line 180: to ensure `json.dumps` produce stable result, can use `json.dumps(variables, sort_keys=True)`, or use `DeepHash`.

janboll: Why do the json dumps sort, The entire array is sorted in the end? 

Isn't DeepHash creating a hash for all items?

hemslo: If you use `DeepHash`, it can be simplified to use those variables as is like this

```python
    hashable = {
        "templates": {t.name: t.dict(by_alias=True) for t in c.templates},
        "variables": variables,
    }
    return DeepHash(hashable)[hashable]
```

janboll: This will do deep hash on all values, but that not required in this use case, right?

hemslo: what's `str(t)` before? Is that also dump all fields as string?

janboll: what about:
```Python
hashable = {
        "templates": c.templates,
        "variables": json.dumps(variables, sort_keys=True),
    }
hashable_string = DeepHash(hashable, apply_hash=False)[hashable]
return DeepHash(hashable_string)[hashable_string]
```

hemslo (approved):
LGTM!

PR #4295, opened 2024-04-23
fix test warnings

chassing:
**:white_check_mark: rename wrongly named test classes**

`pytest` uses `Test` as a pattern to discover tests. <- @janboll, @geoberle 

**:coffin: sql-query: fix deprecation warning**

### DIFF
--- a/reconcile/test/test_ocm_machine_pools.py
+++ b/reconcile/test/test_ocm_machine_pools.py
@@ -32,7 +32,7 @@ from reconcile.ocm_machine_pools import (
 from reconcile.utils.ocm import OCM
 
 
-class TestPool(AbstractPool):
+class TstPool(AbstractPool):
     created = False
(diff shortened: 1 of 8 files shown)


reconcile/test/test_ocm_machine_pools.py
```diff
@@ -32,7 +32,7 @@
 from reconcile.utils.ocm import OCM
 
 
-class TestPool(AbstractPool):
+class TstPool(AbstractPool):
```
hemslo, line 35: Since this is only used in test as stubs, a more suitable name would be `PoolStub`

chassing: `...Stub` is much better. Changed in the last commit. Thanks

hemslo (approved):
LGTM!

PR #4299, opened 2024-04-24
fix aws_api mypy

BumbleFeng:
suit for mypy 1.10.0


chassing (changes requested):
Please don't merge. #4300 does it more easily.

PR #4306, opened 2024-04-26
APPSRE-10117 schedule for SAPM MRs

fishi0x01:
:warning: **THIS IS STILL WIP**

Introduce the `schedule` attribute in SAPM MRs.

A `schedule` is used to determine whether an MR should be auto-merged and/or batched. A `schedule` in the future will simply open an unbatched promotion without auto-merge. The tenant can still lgtm the promotion before the `schedule` is reached. Further, the tenant has visibility on planned promotions.

Once a scheduled non-auto-merge MR reached its due-date, it will be closed in favor of a new auto-mergable/batchable MR. 

Note, that we keep `schedule` state within the SAPM MRs itself. I.e., a MR `content_hash` is tied to the `schedule` in already opened MRs.

`soakDays` will be the first saas attribute to influence a `schedule`.

Related schema: https://github.com/app-sre/qontract-schemas/pull/633


fishi0x01:
We choose a different direction https://github.com/app-sre/qontract-reconcile/pull/4320

PR #4307, opened 2024-04-26
aws_msk: set configuration name based on version

jfchevrette:
This is a proposal to fix an issue where the terraform provider forces a resource re-creation when the kafka_version field is changed (see [here](https://github.com/hashicorp/terraform-provider-aws/blob/9e20b5ae29ff3f6dc4f62a368930973dca08f274/internal/service/kafka/configuration.go#L56))

The problem for us lies in the fact that terraform defaults to removing a resource before is it re-created (replaced). For MSK Configurations that is not allowed by AWS because that config is assigned to a cluster

One frequent solution to these issues is to use terraform's create_before_destroy lifecycle argument, which will force a new resource first, then assign it to the parent resource, and finally destroy the old resource. This does not work in our case because the msk configuration resource name is static and based on the mks cluster name

This PR aims to solve this by making the msk_configuration resource name dynamic based on the version so we can use the create_before_destroy lifecycle argument

See https://issues.redhat.com/browse/APPSRE-10204


BumbleFeng (approved):

chassing (approved):
Thank you

PR #4310, opened 2024-04-29
Use AWSApi to get AWS credentials + Add workers cluster and ns as settings

lechuk47:
- **Use AWSApi to get AWS credentials + Add workers cluster and ns as settings**

Schemas: https://github.com/app-sre/qontract-schemas/pull/640


reconcile/external_resources/integration.py
```diff
@@ -23,14 +25,15 @@
     get_namespaces,
     get_settings,
 )
+from reconcile.utils.aws_api import AWSApi
```
chassing, line 28: You may want to use `reconcile.utils.aws_api_typed` and implement `dynamodb`, wdyt?

lechuk47: I implemented the basic structure for DynamodDB in the `aws_api_typed` and now the code uses the AWSStaticCredentials classes, etc. 

However, I don't see much value in re-implementing the CRUD methods in this layer as it's like duplicating code and types. I set a property to return the raw DynamoDB client instead. 

I left two commented methods in the DynamoDB implementation class for you to understand what I mean. We can discuss this approach if you want but I think, at least in this case, it's not worth adding put_item, del_tiem, get_item, etc. in the aws_api_typed. 

PTAL

reconcile/external_resources/integration.py
```diff
@@ -65,23 +96,28 @@ def run(
         er_inventory=er_inventory,
         module_inventory=m_inventory,
         state_manager=ExternalResourcesStateDynamoDB(
+            aws_api=get_aws_api(
+                query_func=gql.get_api().query,
+                account_name=er_settings.state_dynamodb_account.name,
+                region=er_settings.state_dynamodb_region,
+                secret_reader=secret_reader,
+            ),
```
hemslo, line 104: `aws_api` need to close session after use, can either use `with` or `defer` to call `aws_api.close`

reconcile/external_resources/integration.py
```diff
@@ -41,12 +47,32 @@ def fetch_current_state(
...
+    query_func: Callable,
+    account_name: str,
+    region: str,
+    secret_reader: SecretReaderBase,
+) -> AWSApi:
+    account = [
+        acc
+        for acc in aws_accounts_query(query_func).accounts or []
+        if acc.name == account_name
+    ][0]
```
hemslo, line 60: since this is a new gql query always find by account name, we can use server side filter to pass account name https://github.com/app-sre/qontract-server?tab=readme-ov-file#field-equality-predicate

reconcile/external_resources/integration.py
lechuk47: Added a context manager. PTAL

reconcile/external_resources/integration.py
lechuk47: Added. Thanks for the suggestion. PTAL

reconcile/external_resources/integration.py
```diff
@@ -41,12 +47,39 @@ def fetch_current_state(
...
+            query_func, variables={"filter": {"name": account_name}}
+        ).accounts
+        or []
+    )
+    try:
+        account = accounts[0]
+    except IndexError:
+        raise Exception(
+            "External Resources configured AWS account does not exist or can not be found"
+        )
```
hemslo, line 67: instead of try except, how about do the check directly like

```python
    if not accounts:
      raise ...
```

hemslo (approved):
LGTM!

reconcile/external_resources/integration.py
chassing: Yeah, it looks good, but please remove the commented/dead code from `reconcile/utils/aws_api_typed/dynamodb.py` 🙇

chassing (approved):

reconcile/external_resources/integration.py
lechuk47: removed.

PR #4311, opened 2024-04-29
:sparkles: ldap-groups: manage notes and owners

chassing:
`ldap-groups` manage Rover notes and owners.

The goal is to enable our LDAP groups to be authZ groups for Bitwarden collections. Rover group requirements are:

* Rover group notes must contain the string `@BW@` somewhere
* At least 2 owners are set

Depends on: https://github.com/app-sre/qontract-schemas/pull/639

Tickets
* [APPSRE-10201](https://issues.redhat.com/browse/APPSRE-10201)
* [APPSRE-8399](https://issues.redhat.com/browse/APPSRE-8399)

### DIFF
--- a/reconcile/ldap_groups/integration.py
+++ b/reconcile/ldap_groups/integration.py
@@ -93,12 +93,12 @@ class LdapGroupsIntegration(QontractReconcileIntegration[LdapGroupsIntegrationPa
         desired_groups_for_roles = self.get_desired_groups_for_roles(
             roles,
             contact_list=self.settings.contact_list,
-            owners=[owner],
+            default_owners=[owner],
         )
         desired_groups_for_aws_roles = self.get_desired_groups_for_aws_roles(
             roles,
             contact_list=self.settings.contact_list,
-            owners=[owner],
+            default_owners=[owner],
         )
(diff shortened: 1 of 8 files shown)


reconcile/ldap_groups/integration.py
```diff
@@ -147,38 +147,47 @@ def get_roles(self, query_func: Callable) -> list[RoleV1]:
...
+            groups.append(
+                Group(
+                    name=role.ldap_group.name,
+                    description="Persisted App-Interface role. Managed by qontract-reconcile",
+                    notes=role.ldap_group.notes,
+                    display_name=f"{role.ldap_group.name} (App-Interface))",
+                    members=members,
+                    owners=default_owners
+                    if not role.ldap_group.members_are_owners
+                    else default_owners + members,
```
fishi0x01, line 181: I dont think its a problem, but can it be an issue if a `member` is already part of the `default_owners` list, i.e., a `member` entity appears twice in the `owners` list?

chassing: The `default_owners` are just our service accounts; therefore, it isn't a problem at all.

fishi0x01 (approved):

PR #4314, opened 2024-04-29
Refactor slack usergroups

janboll:
The purpose of this PR is to make slack-usergroups read for extended early exit. 

 * Remove Slack API calls from get_desired_state methods. This makes it possible to profit from extended early exit, to reduce Slack API calls.
 * Update tests
 * Add Slack Request metrics to slack_api, allows to track number of request made to slack api and check improvements made


reconcile/utils/metrics.py
```diff
@@ -115,6 +115,12 @@
     labelnames=["verb", "client_id"],
 )
 
+slack_request = Counter(
```
chassing, line 118: ❤️

chassing (approved):

reconcile/slack_usergroups.py
```diff
@@ -492,7 +478,7 @@ def get_desired_state_cluster_usergroups(
             for u in openshift_users_desired_state
             if u["cluster"] == cluster.name
         ]
-        cluster_usernames = list({
+        cluster_usernames = set({
```
hemslo, line 481: `set` can be skipped here as it's already using set comprehension

reconcile/utils/slack_api.py
```diff
@@ -300,17 +308,26 @@ def _initiate_usergroups(self) -> None:
...
         if len(usergroup) != 1:
             raise UsergroupNotFoundException(handle)
         return usergroup[0]
 
     def create_usergroup(self, handle: str) -> str:
+        slack_request.labels("usergroups.create", "POST").inc()
+
         response = self._sc.usergroups_create(name=handle, handle=handle)
+        # Invalidate the usergroups list cache
+        self._user_groups_initialized = False
```
hemslo, line 330: is this response containing all info about the user group? can we add it to `self.usergroups` instead of invalidate cache to save future api calls to list usergroups in one reconcile run?

reconcile/slack_usergroups.py
```diff
@@ -689,24 +663,40 @@ def act(
                 usergroup, State()
             )
 
+            slack_client = slack_map[workspace].slack
+
+            desired_ug_state.users = {
```
hemslo, line 668: change desired state in `act` is a little strange, since `desired_ug_state.users` only used in `_update_usergroup_users_from_state`, can we construct `users` inside it instead of update desired state?

reconcile/slack_usergroups.py
```diff
@@ -689,24 +663,40 @@ def act(
...
+                SlackObject(pk=pk, name=name)
+                for pk, name in slack_client.get_users_by_names(
+                    sorted(desired_ug_state.user_names)
+                ).items()
+            }
+
+            desired_ug_state.channels = {
```
hemslo, line 675: similar here, `desired_ug_state.channels` only used in `_update_usergroup_from_state`.

PR #4315, opened 2024-04-29
update entrypoint for tini

cubismod:
Removing the `/run-integration.py` argument in the FIPS image entrypoint. The reason being that this will break Jenkins builds as they attempt to perform a docker/podman run.

`podman run --rm ... quay.io/app-sre/qontract-reconcile-fips:9a81949 qontract-reconcile --config /config/config.toml ...` will evaluate to actually running the following command in the container:

`/tini -- /run-integration.py qontract-reconcile --config /config/config.toml ...` when it should just run `/tini -- qontract-reconcile --config /config/config.toml ...`


jonmosco (approved):
LGTM!

PR #4316, opened 2024-04-29
[dtp] Create new token for Route Monitoring Operator

OliviaHY:
Create new token for Route Monitoring Operator as a new Secret in SyncSet for a different namespace.

This was tested e2e in integration environment.

### DIFF
--- a/reconcile/dynatrace_token_provider.py
+++ b/reconcile/dynatrace_token_provider.py
@@ -61,8 +61,13 @@ QONTRACT_INTEGRATION = "dynatrace-token-provider"
 SYNCSET_ID = "ext-dynatrace-tokens-dtp"
 SECRET_NAME = "dynatrace-token-dtp"
 SECRET_NAMESPACE = "dynatrace"
+RMO_SECRET_NAME = "dynatrace-token"  # Name of the Secret that contains the API token for Route Monitor Operator
+RMO_SECRET_NAMESPACE = "openshift-route-monitor-operator"  # Namespace of the Secret that contains the API token for Route Monitor Operator
 DYNATRACE_INGESTION_TOKEN_NAME = "dynatrace-ingestion-token"
 DYNATRACE_OPERATOR_TOKEN_NAME = "dynatrace-operator-token"
+DYNATRACE_RMO_TOKEN_NAME = (
+    "dynatrace-rmo-token"  # Name of the API token for Route Monitor Operator
+)
 
 
 class DynatraceTokenProviderIntegrationParams(PydanticRunParams):
@@ -146,6 +151,8 @@ class DynatraceTokenProviderIntegration(
                 existing_dtp_tokens = {}
 
                 for cluster in clusters:
+                    if cluster.ocm_cluster.display_name != "appint-ex-01":
+                        continue
                     try:
                         with DTPOrganizationErrorRate(
                             integration=self.name,
@@ -238,14 +245,16 @@ class DynatraceTokenProviderIntegration(
         if not existing_syncset:
             if not dry_run:
                 try:
-                    (ingestion_token, operator_token) = self.create_dynatrace_tokens(
-                        dt_client, cluster.ocm_cluster.external_id
+                    (ingestion_token, operator_token, rmo_token) = (
+                        self.create_dynatrace_tokens(
+                            dt_client, cluster.ocm_cluster.external_id
+                        )
                     )
                     create_syncset(
                         ocm_client,
                         cluster.ocm_cluster.id,
                         self.construct_syncset(
-                            ingestion_token, operator_token, dt_api_url
+                            ingestion_token, operator_token, rmo_token, dt_api_url
                         ),
                     )
                 except Exception as e:
@@ -287,15 +296,28 @@ class DynatraceTokenProviderIntegration(
                         logging.info(
                             f"Operator token created in Dynatrace for cluster {cluster.ocm_cluster.external_id}."
                         )
+                    elif token_name == DYNATRACE_RMO_TOKEN_NAME:
+                        if not dry_run:
+                            rmo_token = self.create_dynatrace_rmo_token(
+                                dt_client, cluster.ocm_cluster.external_id
+                            )
+                            token["id"] = rmo_token.id
+                            token["token"] = rmo_token.token
+                        logging.info(
+                            f"RMO token created in Dynatrace for cluster {cluster.ocm_cluster.external_id}."
+                        )
                 elif token_name == DYNATRACE_INGESTION_TOKEN_NAME:
                     ingestion_token = ApiTokenCreated(raw_element=token)
                 elif token_name == DYNATRACE_OPERATOR_TOKEN_NAME:
                     operator_token = ApiTokenCreated(raw_element=token)
+                elif token_name == DYNATRACE_RMO_TOKEN_NAME:
+                    rmo_token = ApiTokenCreated(raw_element=token)
             if need_patching:
                 if not dry_run:
                     patch_syncset_payload = self.construct_base_syncset(
                         ingestion_token=ingestion_token,
                         operator_token=operator_token,
+                        rmo_token=rmo_token,
                         dt_api_url=dt_api_url,
                     )
                     try:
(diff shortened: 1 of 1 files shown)


reconcile/dynatrace_token_provider.py
```diff
@@ -328,14 +348,22 @@ def get_syncset(
 
     def get_tokens_from_syncset(self, syncset: Mapping) -> Mapping:
         tokens = {}
+        operator_token_id = operator_token = ingest_token_id = ingest_token = (
+            rmo_token_id
+        ) = rmo_token = "DoNotExist"
```
hemslo, line 353: `DoNotExist` is a little confusing as according to the context, all these vars should be populated in the end, otherwise a new token called `DoNotExist` will be created later.

Can directly populate `tokens` in each branch, then raise exception if any of them missing.

OliviaHY: It won't be created because the id is only used to check if there is a token in Dynatrace with such ID exist and they won't have a value as `DoNotExist`.

hemslo: What I mean is this default value changed the behaviour of this method, previously when no matching secrets found, it will raise exception as vars like `operator_token_id` is undefined. New code have them a new default value set, so even no matching secrets found, there will always have `tokens` dict fully populated, which lead to all 3 tokens eventually created.

I'm trying to get what's the expected requirement here:

AC1:

```
Given syncset not exist in cluster
When process cluster
Then a syncset will all 3 tokens created
```

AC2:

```
Given syncset exist with all 3 tokens
When process cluster
Then nothing changed
```

AC3:

```
Given syncset exist with 2 tokens
When process cluster
Then create missing token and patch syncset with 3 tokens
```

If above requirements are true, to make the code clear and get rid of duplicate similar logic, like this:


```python
    def get_tokens_from_syncset(self, syncset: Mapping) -> dict[str, dict[str, str]]:
        tokens = {
            DYNATRACE_INGESTION_TOKEN_NAME: {},
            DYNATRACE_OPERATOR_TOKEN_NAME: {},
            DYNATRACE_RMO_TOKEN_NAME: {},
        }
        for resource in syncset["resources"]:
            if resource["kind"] != "Secret":
                continue
            if resource["metadata"]["namespace"] == SECRET_NAMESPACE:
                tokens[DYNATRACE_OPERATOR_TOKEN_NAME] = {
                    "id": self.base64_decode(resource["data"]["apiTokenId"]),
                    "token": self.base64_decode(resource["data"]["apiToken"]),
                }
                tokens[DYNATRACE_INGESTION_TOKEN_NAME] = {
                    "id": self.base64_decode(resource["data"]["dataIngestTokenId"]),
                    "token": self.base64_decode(resource["data"]["dataIngestToken"]),
                }
            elif resource["metadata"]["namespace"] == RMO_SECRET_NAMESPACE:
                tokens[DYNATRACE_RMO_TOKEN_NAME] = {
                    "id": self.base64_decode(resource["data"]["apiTokenId"]),
                    "token": self.base64_decode(resource["data"]["apiToken"]),
                }
        return tokens
```

```python
    def process_token(
        self,
        token_name: str,
        token: dict[str, Any],
        existing_dtp_tokens: Iterable[str],
        dry_run: bool,
        dt_client: Dynatrace,
        cluster_external_id: str,
    ) -> ApiTokenCreated | None:
        if token.get("id") in existing_dtp_tokens:
            return ApiTokenCreated(raw_element=token)
        logging.info(
            f"Create new token {token_name} in Dynatrace for cluster {cluster_external_id}."
        )
        if not dry_run:
            if token_name == DYNATRACE_INGESTION_TOKEN_NAME:
                return self.create_dynatrace_ingestion_token(
                    dt_client,
                    cluster_external_id,
                )
            if token_name == DYNATRACE_OPERATOR_TOKEN_NAME:
                return self.create_dynatrace_operator_token(
                    dt_client,
                    cluster_external_id,
                )
            if token_name == DYNATRACE_RMO_TOKEN_NAME:
                return self.create_dynatrace_rmo_token(
                    dt_client,
                    cluster_external_id,
                )
        return None
```

```python
    def process_cluster(
        self,
        dry_run: bool,
        cluster: ClusterDetails,
        dt_client: Dynatrace,
        ocm_client: OCMBaseClient,
        existing_dtp_tokens: list[str],
        tenant_id: str,
    ) -> None:
        ...
        else:
            tokens = self.get_tokens_from_syncset(existing_syncset)
            need_patching = any(
                token.get("id") not in existing_dtp_tokens for token in tokens.values()
            )
            if need_patching:
                processed_tokens = {
                    token_name: self.process_token(
                        token_name,
                        token,
                        existing_dtp_tokens,
                        dry_run,
                        dt_client,
                        cluster.ocm_cluster.external_id,
                    )
                    for token_name, token in tokens.items()
                }
                if not dry_run:
                    patch_syncset_payload = self.construct_base_syncset(
                        ingestion_token=processed_tokens[
                            DYNATRACE_INGESTION_TOKEN_NAME
                        ],
                        operator_token=processed_tokens[DYNATRACE_OPERATOR_TOKEN_NAME],
                        rmo_token=processed_tokens[DYNATRACE_RMO_TOKEN_NAME],
                        dt_api_url=dt_api_url,
                    )
                    try:
                        logging.info(f"Patching syncset {SYNCSET_ID}.")
                        patch_syncset(
                            ocm_client,
                            cluster_id=cluster.ocm_cluster.id,
                            syncset_id=SYNCSET_ID,
                            syncset_map=patch_syncset_payload,
                        )
                    except Exception as e:
                        _expose_errors_as_service_log(
                            ocm_client,
                            cluster.ocm_cluster.external_id,
                            f"DTP can't patch Syncset {SYNCSET_ID} due to {str(e.args)}",
                        )
                logging.info(f"Syncset {SYNCSET_ID} patched.")
```

May need to adjust logging to make it suitable for both dry run and non dry run.

OliviaHY: > which lead to all 3 tokens eventually created

This is not true. The way this method is being used is to compare the token_id in a list of existing ids. And `DoNotExist` is never going to match any DT API token ID.

hemslo: that's why all 3 tokens (`DYNATRACE_INGESTION_TOKEN_NAME`, `DYNATRACE_OPERATOR_TOKEN_NAME`, `DYNATRACE_RMO_TOKEN_NAME`) will be created, `DoNotExist` will cause `token["id"] not in existing_dtp_tokens` always return True, then go to create token branch.

OliviaHY: Why all 3 will be created? `token["id"] not in existing_dtp_tokens` only return true for the specific one that has the value of `DoNotExist`. The tokens that's already in the SyncSet will be retrieved by following block and overwrite the original value of `DoNotExist`
```
for resource in syncset["resources"]:
            if resource["kind"] != "Secret":
                continue
            if resource["metadata"]["namespace"] == SECRET_NAMESPACE:
                operator_token_id = self.base64_decode(resource["data"]["apiTokenId"])
                operator_token = self.base64_decode(resource["data"]["apiToken"])
                ingest_token_id = self.base64_decode(
                    resource["data"]["dataIngestTokenId"]
                )
                ingest_token = self.base64_decode(resource["data"]["dataIngestToken"])
            elif resource["metadata"]["namespace"] == RMO_SECRET_NAMESPACE:
                rmo_token_id = self.base64_decode(resource["data"]["apiTokenId"])
                rmo_token = self.base64_decode(resource["data"]["apiToken"])
```

hemslo: I mean eventually 3 tokens will be created, from the logic you posted above, even the syncset has no secret, `tokens` will have all 3 populated as `DoNotExist`, then all 3 tokens will be created. Using `DoNotExist` and overwrite later is very confusing, check immutable version at https://github.com/app-sre/qontract-reconcile/pull/4316#discussion_r1587096921

OliviaHY: If SyncSet is empty, then the token should be created, that's the goal. To create tokens when they are missing in SyncSet.

hemslo: if syncset exist but somehow `resources` is empty, previous code will throw exception, new code will populate `tokens` with all keys and then create tokens, that's a breaking change/bugfix require unit tests.

There are also warnings like `ingestion_token local variable referenced before assignment` in this file because of too many conditional branches, if these 3 always need to be used together, instead of put them in a `dict`, can create one `BaseModel` to hold them all to simplify params passing.

OliviaHY:
It was decided to go with a different direction to support multiple tokens so this MR now is obsolete

PR #4334, opened 2024-05-01
[fips image] add CMD

cubismod:
Followup to #4315

According to [this issue](https://github.com/moby/moby/issues/5147) & [stackoverflow](https://stackoverflow.com/questions/49028644/is-cmd-in-parent-docker-overriden-by-cmd-entrypoint-in-child-docker-image) along with running a `docker inspect` on the FIPS image:

image

when using a multistage build, if you define an entrypoint in the child image, the `CMD` is set to `null` which breaks functionality on OpenShift. I've updated the Dockerfile to match [Tini's recommend method](https://github.com/krallin/tini?tab=readme-ov-file#using-tini) in light of this.


jonmosco (approved):
LGTM

PR #4335, opened 2024-05-02
[status-page-components] deprecate usage of transity-statuspageio

maorfr:
* the underlying library is not maintained anymore
* the library has no type annotated
* the library does not support pagination which becomes important for this API


geoberle (approved):
thanks for cleaning up ❤️

patjlm (approved):

PR #4339, opened 2024-05-03
:coffin: qontract-cli: remove outdated terraform_users_credentials output

chassing:
Remove the `qontract-cli terraform-users-credentials` command because it's entirely [outdated](https://gitlab.cee.redhat.com/service/app-interface-output/-/blob/master/terraform-users-credentials.md) and shows wrong information. No integration or job removes old IAM user secrets from Vault or the S3 state. SSO is enabled for most of our AWS accounts, so it isn't worth investing time fixing that.

Ticket: [APPSRE-10216](https://issues.redhat.com/browse/APPSRE-10216)

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -32,7 +32,6 @@ from rich.console import (
 )
 from rich.table import Table
 from rich.tree import Tree
-from sretoolbox.utils import threaded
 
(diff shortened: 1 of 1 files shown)


fishi0x01 (approved):
LGTM. We must remove it from the jenkins job in app-interface then too.

PR #4344, opened 2024-05-06
[status-page-maintenances] introduce integration

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10225

design document: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/maintenances.md

depends on #4343

PR #4349, opened 2024-05-08
[status-page-maintenances] component support

maorfr:

design document: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/maintenances.md

follows up on #4344

in this PR we add support to add affected components to an incident.

### DIFF
--- a/reconcile/statuspage/page.py
+++ b/reconcile/statuspage/page.py
@@ -47,12 +47,14 @@ class StatusComponent(BaseModel):
         arbitrary_types_allowed = True
 
     @classmethod
-    def init_from_page_component(cls, component: StatusPageComponentV1) -> Self:
+    def init_from_page_component(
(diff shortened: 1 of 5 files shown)


reconcile/statuspage/page.py
```diff
@@ -103,12 +105,24 @@ class StatusMaintenance(BaseModel):
...
+        cls,
+        maintenance: MaintenanceV1,
+        page_components: list[StatusPageComponentV1],
+    ) -> Self:
+        affected_services = [a.name for a in maintenance.affected_services]
+        affected_components = [
+            StatusComponent.init_from_page_component(c, name_override=c.display_name)
+            for c in page_components
+            if c.app.name in affected_services
+        ]
```
maorfr, line 121: this is the important part of this PR, where we find the affected components based on the affected services.

PR #4348, opened 2024-05-07
[qontract-cli] get maintenances

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10234

this qontract-cli command will list maintenances

PR #4350, opened 2024-05-08
[status-page-maintenances] only manage scheduled maintenances

maorfr:

design document: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/maintenances.md

follows up on https://github.com/app-sre/qontract-reconcile/pull/4344 and #4348

in this PR we add a condition to only manage scheduled maintenances.

### DIFF
--- a/reconcile/statuspage/integrations/maintenances.py
+++ b/reconcile/statuspage/integrations/maintenances.py
@@ -65,7 +65,7 @@ class StatusPageMaintenancesIntegration(QontractReconcileIntegration[NoParams]):
                 self.reconcile(
                     dry_run=dry_run,
                     desired_state=desired_state,
-                    current_state=page_provider.maintenances,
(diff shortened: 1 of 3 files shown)


reconcile/statuspage/integrations/maintenances.py
```diff
@@ -46,13 +47,15 @@ def reconcile(
...
         pages = get_status_pages()
+        now = datetime.now(timezone.utc)
 
         error = False
         for p in pages:
             try:
                 desired_state = [
                     StatusMaintenance.init_from_maintenance(m)
                     for m in p.maintenances or []
+                    if datetime.fromisoformat(m.scheduled_start) > now
```
maorfr, line 58: this is the important part of the PR, where we only manage maintenances that have not yet started (and matching qontract-cli with the same behavior)

PR #4355, opened 2024-05-09
[qontract-cli/app-interface-review-queue] post to slack

maorfr:
while it's nice having the queue in app-interface-output, let's have it in slack as well :pray:


reconcile/utils/output.py
```diff
@@ -13,7 +13,7 @@ def print_output(
     options: Mapping[str, Union[str, bool]],
     content: list[dict],
     columns: Iterable[str] = (),
-):
+) -> Any:
```
hemslo, line 16: from the code below, this function can return `str | None`, any other possible types?

hemslo (approved):
LGTM!

reconcile/utils/output.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/4355/commits/efed676f5138760f17d7ffa1b97b7c205b260eaf

patjlm (approved):

PR #4359, opened 2024-05-10
Add extended early exit

janboll:
This implements extended early exit for slack usergroups integration. 

- Payload is not set in Result, since the state data structure is not JSON hashable (using sets)
- This does not reduce the runtime by much (around. 1/5th). It drastically reduces API requests towards slack, which caused API rate liming errors in the past. More caching for computing desired state would be required to further reduce runtime.

### DIFF
--- a/reconcile/slack_usergroups.py
+++ b/reconcile/slack_usergroups.py
@@ -9,10 +9,15 @@ from typing import (
     Any,
     Optional,
     Sequence,
+    TypedDict,
     Union,
 )
 from urllib.parse import urlparse
 
+from build.lib.reconcile.utils.extended_early_exit import (
+    ExtendedEarlyExitRunnerResult,
+    extended_early_exit_run,
+)
 from github.GithubException import UnknownObjectException
 from pydantic import BaseModel
 from pydantic.utils import deep_update
@@ -67,6 +72,8 @@ from reconcile.utils.slack_api import (
 
 DATE_FORMAT = "%Y-%m-%d %H:%M"
 QONTRACT_INTEGRATION = "slack-usergroups"
+INTEGRATION_VERSION = "0.1.0"
+
 error_occurred = False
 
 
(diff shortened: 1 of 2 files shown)


reconcile/slack_usergroups.py
```diff
@@ -793,9 +851,7 @@ def run(
         slack_map=slack_map,
         dry_run=dry_run,
     )
-    if error_occurred:
-        logging.error("Error(s) occurred.")
-        sys.exit(1)
+    return ExtendedEarlyExitRunnerResult(payload={}, applied_count=0)
```
mehfuz, line 854: why do we use applied_count as 0?

janboll: The blunt answer is: it is a required field and I did not care about this debugging info at the moment. Perhaps add it later?
Would require additional refactorings...

hemslo: `applied_count` is not debugging info, if it's not 0, next run won't use cache, until a run with nothing applied (applied_count = 0), this is the way to detect reconcile loop.

janboll: My bad, I thought it was informational only.

reconcile/slack_usergroups.py
```diff
@@ -787,15 +850,13 @@ def run(
...
+    apply_count = act(
         current_state=current_state,
         desired_state=desired_state,
         slack_map=slack_map,
         dry_run=dry_run,
     )
-    if error_occurred:
-        logging.error("Error(s) occurred.")
-        sys.exit(1)
+    return ExtendedEarlyExitRunnerResult(payload={}, applied_count=apply_count)
```
hemslo, line 859: Now we have a use case that no payload needed, we can update `ExtendedEarlyExitRunnerResult` to make it default `None` to simplify usage.

```python
    payload: object = None
```

hemslo (approved):
LGTM! Another faster integration!

patjlm:
:heart:

PR #4360, opened 2024-05-10
adding Unleash log filters for FR

suzana-nesic:
Editing fluentd configuration to filter out more warning messages coming from Unleash in the FedRAMP environment

### DIFF
--- a/helm/qontract-reconcile/templates/template.yaml
+++ b/helm/qontract-reconcile/templates/template.yaml
@@ -157,6 +157,16 @@ objects:
               </exclude>
             </filter>
 
+            {{- if $integration.unleashConnectionPool}}
+            <filter integration>
+              @type grep
+              <exclude>
+                key message
+                pattern /Max retries exceeded with url: \/api\/client\/features/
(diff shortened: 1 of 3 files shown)


cubismod (approved):
LGTM

patjlm:
I wonder if some dedupe filter would be possible to not completely remove those messages (just print the number of successive ocurrences).
maybe something like https://github.com/sonots/fluent-plugin-filter_dup (never used it, no idea how good/bad it is..)

cubismod:
> I wonder if some dedupe filter would be possible to not completely remove those messages (just print the number of successive ocurrences). maybe something like https://github.com/sonots/fluent-plugin-filter_dup (never used it, no idea how good/bad it is..)

Oh jeez just looked at that plugin and it hasn't been updated in 10 years. I would argue it's not the most worthwhile use of time to add in more plugins to the fluentd image as Lisa is already spiking Vector as an alternative to Fluentd in this ticket: https://issues.redhat.com/browse/APPSRE-8094

PR #4362, opened 2024-05-13
hashable distancekey

fishi0x01:
Fixing error in prod

```
TypeError: unhashable type: 'VaultSecret'
```

### DIFF
--- a/tools/saas_metrics_exporter/commit_distance/commit_distance.py
+++ b/tools/saas_metrics_exporter/commit_distance/commit_distance.py
@@ -26,6 +26,9 @@ class DistanceKey:
     ref_from: str
     ref_to: str
 
+    def __hash__(self) -> int:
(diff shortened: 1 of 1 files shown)


tools/saas_metrics_exporter/commit_distance/commit_distance.py
```diff
@@ -26,6 +26,9 @@ class DistanceKey:
     ref_from: str
     ref_to: str
 
+    def __hash__(self) -> int:
+        return hash((self.repo_url, self.ref_from, self.ref_to))
```
hemslo, line 30: how about declare `auth_code` field not used for hash or compare? like `auth_code: HasSecret | None = field(hash=False, compare=False)`

fishi0x01: Good idea. Fixed

hemslo (approved):
LGTM!

PR #4364, opened 2024-05-14
[cst] add a missing error return

cubismod:
When working on https://github.com/app-sre/terraform-repo-executor/pull/22, I noticed that one error case didn't return an error on failure so fixing that here.


rporres (approved):
indeed, thx for fixing it.

PR #4365, opened 2024-05-14
APPSRE-10229 common typed query for OCMEnv

fishi0x01:
`get_ocm_environment()` functions are scattered across multiple places. Lets have a single typed query -> DTP will be the first to use it

### DIFF
--- a/reconcile/test/test_typed_queries/test_ocm_environments.py
+++ b/reconcile/test/test_typed_queries/test_ocm_environments.py
@@ -1,0 +1,70 @@
+from collections.abc import (
+    Callable,
+    Mapping,
+)
+from typing import Optional
+
+import pytest
+
+from reconcile.gql_definitions.common.ocm_environments import (
+    DEFINITION,
+    OCMEnvironmentsQueryData,
+)
+from reconcile.typed_queries.ocm_environments import (
+    NoOCMEnvironmentsFoundError,
+    get_ocm_environments,
+)
+from reconcile.utils.gql import GqlApi
+
+
+def test_no_ocm_environments(
+    gql_api_builder: Callable[[Optional[Mapping]], GqlApi],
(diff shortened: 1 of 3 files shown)


reconcile/test/test_typed_queries/test_ocm_environments.py
```diff
@@ -0,0 +1,70 @@
...
+                    "name": "env1",
+                    "accessTokenClientSecret": {},
+                },
+            ]
+        },
+    )
+    api = gql_api_builder(data.dict(by_alias=True))
+    environments = get_ocm_environments(gql_api=api, env_name="env1")
+    api.query.assert_called_once_with(DEFINITION, variables={"name": "env1"})
+    assert len(environments) == 1
```
hemslo, line 70: It would be good to assert on the whole object instead of just length, like extract

```python
                {
                    "name": "env1",
                    "accessTokenClientSecret": {},
                }
```
into a expected result, then verify them.
Otherwise I can make the tests pass to just return dummy `OCMEnvironment` objects.

reconcile/typed_queries/ocm_environments.py
```diff
@@ -0,0 +1,21 @@
...
+
+def get_ocm_environments(
+    gql_api: GqlApi,
+    env_name: str | None = None,
+) -> list[OCMEnvironment]:
+    """Returns OCM Environments and raises err if none are found"""
+    variables = {"name": env_name} if env_name else None
+    data = query(query_func=gql_api.query, variables=variables)
+    if not data.environments:
+        raise NoOCMEnvironmentsFoundError(f"No OCM Environments found for {env_name=}")
```
hemslo, line 20: previously this will return empty list, now it throws exception, this is a breaking change, or we have data validation to ensure it always not empty?

maorfr:
consider consolidating into https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/typed_queries/ocm.py

PR #4366, opened 2024-05-15
Remove stacked retry() calls in secrets_reader

lechuk47:
Stacked `@retry` definitions multiply the retry attempts for a function call. 

```python 
@retry()  
def func1():
  func2()
 
@retry() 
def func2():
  do_stuff
 ```

`retry` has a default `max_attemps=3`. In case of failure in `func2`, `func1` will be retried 9 times.

### DIFF
--- a/reconcile/utils/secret_reader.py
+++ b/reconcile/utils/secret_reader.py
@@ -12,7 +12,6 @@ from typing import (
 )
 
 from hvac.exceptions import Forbidden
-from sretoolbox.utils import retry
 
 from reconcile.utils import (
     config,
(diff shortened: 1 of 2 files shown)


fishi0x01 (approved):
My only concern is that Im not sure all consumers are currently adding their own retries. But I agree that this is sth that should either way be defined on consumer side and that this should be the common approach.

fishi0x01:
Still wondering if we should get some more opinions on this one

hemslo (approved):
LGTM! If we just want to retry on request error, can use `HTTPAdapter(max_retries=x)` to replace `@retry` for `_VaultClient` too.

lechuk47:
> My only concern is that Im not sure all consumers are currently adding their own retries. But I agree that this is sth that should either way be defined on consumer side and that this should be the common approach.

All the `retry()` calls I have removed have a `retry()` in the call to vault. 

I also think the retry calls should be in the top calls (this way, the retry parameters can be changed in every specific case) but I'm not sure every caller have a retry defined.

PR #4370, opened 2024-05-17
refactor gitlab-permissions integration

mehfuz:
Part of - https://issues.redhat.com/browse/APPSRE-10272

The PR is an effort to reduce the execution time and api call made to gitlab. Changes are made keeping the flow of the integration as is.This is done by requesting only specfic members of project using `user_ids` query so we only get the app-sre group users from gitlab.
Local Testing:-
By keeping threadsize as one the integration took - `0:06:08.068157` . By adding the modification it takes `0:03:13.167428`


patjlm:
Nice. Can you tell the number of gl queries without and with this change?

reconcile/gitlab_permissions.py
```diff
@@ -26,6 +27,25 @@ def get_members_to_add(repo, gl, app_sre):
...
+    chunks = [
+        app_sre_user_ids[i : i + PAGE_SIZE]
+        for i in range(0, len(app_sre_user_ids), PAGE_SIZE)
+    ]
+    for chunk in chunks:
+        maintainers = gl.get_project_maintainers(
+            repo, query=create_user_ids_query(chunk)
+        )
+        app_sre_maintainers.extend(maintainers)
+    return app_sre_maintainers
```
hemslo, line 42: in python 3.12, there is [itertools.batched](https://docs.python.org/3.12/library/itertools.html#itertools.batched) can do chunk, can add it to util and doc can be replaced in 3.12


```python
def batched(iterable, n):
    # batched('ABCDEFG', 3) → ABC DEF G
    if n < 1:
        raise ValueError('n must be at least one')
    it = iter(iterable)
    while batch := tuple(islice(it, n)):
        yield batch
```

Then this logic can be simplified like

```python
    app_sre_user_ids = [user.id for user in app_sre]
    chunks = batched(app_sre_user_ids, PAGE_SIZE)
    app_sre_maintainers = (
        gl.get_project_maintainers(repo, query=create_user_ids_query(chunk))
        for chunk in chunks
    )
    return list(itertools.chain.from_iterable(app_sre_maintainers))
```

reconcile/gitlab_permissions.py
```diff
@@ -26,6 +27,25 @@ def get_members_to_add(repo, gl, app_sre):
...
+
+
+def create_user_ids_query(ids):
+    return {"user_ids": ",".join(str(id) for id in ids)}
```
hemslo, line 46: according to [api doc](https://docs.gitlab.com/ee/api/members.html#list-all-members-of-a-group-or-project-including-inherited-and-invited-members), `user_ids` is array of ints, have you tried to pass it directly like `{"user_ids": ids}`

mehfuz: @hemslo  by passing it directly it is taking the value as multilpe `user_ids` in api like, 
`?user_ids=xxx&user_ids=xx1&user_ids=xx2` which results the API to return only the user having the last user_id field in url query. But if we convert it to above format it creates a coma seperated query 
`?user_ids=xxx,xx1,xx2` and returns all the users matching the ids mentioned

mehfuz:
> Nice. Can you tell the number of gl queries without and with this change?

@patjlm  i think the number of gql queries will still remain same.. The gql query part can be optimised upto some extent by adding extended early exit.

patjlm:
> > Nice. Can you tell the number of gl queries without and with this change?
> 
> @patjlm  i think the number of gql queries will still remain same.. The gql query part can be optimised upto some extent by adding extended early exit.

I meant gitlab, not gql, sorry

mehfuz:
> > > Nice. Can you tell the number of gl queries without and with this change?
> > 
> > 
> > @patjlm  i think the number of gql queries will still remain same.. The gql query part can be optimised upto some extent by adding extended early exit.
> 
> I meant gitlab, not gql, sorry

@patjlm  local testing found out:-
Without using `user_ids` no. of requests is `367` 
With  `user_ids` no. of requests is `189`

reconcile/gitlab_permissions.py
mehfuz: @hemlso updated the code

hemslo (approved):
LGTM!

PR #4372, opened 2024-05-17
increase timeout for dashdot slo

BumbleFeng:


patjlm:
Can you explain the rationale a bit more?

PR #4375, opened 2024-05-21
Introduce ERv2 Outputs sync integration

lechuk47:
This PR introduces an integration to sync External resource outputs from Vault into the target clusters.

As discussed in the design, ERv2 will continuously sync the outputs from Vault instead of running Terraform for all the resources at each iteration.

The full Picture: 
ERv2 Resource ACTION (Create or Update) -> It will update Vault and Sync the Outputs into the target cluster. 
ERv2 Secrets Sync Integration -> Will loop over all Outputs and sync them into clusters to ensure the Outputs desired state.

Requires: 
* https://github.com/app-sre/qontract-schemas/pull/654/files

### DIFF
--- a/reconcile/external_resources/integration_secrets_sync.py
+++ b/reconcile/external_resources/integration_secrets_sync.py
@@ -1,0 +1,32 @@
+from reconcile.external_resources.model import ExternalResourcesInventory
+from reconcile.external_resources.secrets_sync import VaultSecretsReconciler
+from reconcile.typed_queries.app_interface_vault_settings import (
+    get_app_interface_vault_settings,
+)
+from reconcile.typed_queries.external_resources import get_namespaces, get_settings
+from reconcile.utils.openshift_resource import ResourceInventory
+from reconcile.utils.secret_reader import create_secret_reader
+from reconcile.utils.semver_helper import make_semver
+
+QONTRACT_INTEGRATION = "external_resources_secrets_sync"
+QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
+
+
+def run(dry_run: bool, thread_pool_size: int) -> None:
+    """Integration that syncs External Resources Outputs Secrets from Vault into
+    the target clusters
+    """
+    vault_settings = get_app_interface_vault_settings()
+    secrets_reader = create_secret_reader(use_vault=vault_settings.vault)
+    er_settings = get_settings()[0]
+    namespaces = [ns for ns in get_namespaces() if ns.external_resources]
+    er_inventory = ExternalResourcesInventory(namespaces)
+
+    reconciler = VaultSecretsReconciler(
+        ri=ResourceInventory(),
+        secrets_reader=secrets_reader,
+        vault_path=er_settings.vault_secrets_path,
+        thread_pool_size=thread_pool_size,
+        dry_run=dry_run,
+    )
+    reconciler.sync_secrets(er_inventory.values())
--- a/reconcile/external_resources/integration_secrets_sync.py
+++ b/reconcile/external_resources/integration_secrets_sync.py
@@ -1,9 +1,16 @@
-from reconcile.external_resources.model import ExternalResourcesInventory
+from reconcile.external_resources.model import (
+    ExternalResourcesInventory,
+    load_module_inventory,
+)
 from reconcile.external_resources.secrets_sync import VaultSecretsReconciler
 from reconcile.typed_queries.app_interface_vault_settings import (
     get_app_interface_vault_settings,
 )
-from reconcile.typed_queries.external_resources import get_namespaces, get_settings
+from reconcile.typed_queries.external_resources import (
+    get_modules,
+    get_namespaces,
+    get_settings,
+)
 from reconcile.utils.openshift_resource import ResourceInventory
 from reconcile.utils.secret_reader import create_secret_reader
 from reconcile.utils.semver_helper import make_semver
@@ -17,16 +24,24 @@ def run(dry_run: bool, thread_pool_size: int) -> None:
     the target clusters
     """
     vault_settings = get_app_interface_vault_settings()
-    secrets_reader = create_secret_reader(use_vault=vault_settings.vault)
+    secret_reader = create_secret_reader(use_vault=vault_settings.vault)
     er_settings = get_settings()[0]
+
     namespaces = [ns for ns in get_namespaces() if ns.external_resources]
     er_inventory = ExternalResourcesInventory(namespaces)
+    m_inventory = load_module_inventory(get_modules())
+
+    to_sync_specs = []
+    for key, spec in er_inventory.items():
+        module = m_inventory.get_from_external_resource_key(key)
+        if module.outputs_secret_sync:
+            to_sync_specs.append(spec)
 
     reconciler = VaultSecretsReconciler(
         ri=ResourceInventory(),
-        secrets_reader=secrets_reader,
+        secrets_reader=secret_reader,
         vault_path=er_settings.vault_secrets_path,
         thread_pool_size=thread_pool_size,
         dry_run=dry_run,
     )
-    reconciler.sync_secrets(er_inventory.values())
+    reconciler.sync_secrets(to_sync_specs)
--- a/reconcile/external_resources/integration_secrets_sync.py
+++ b/reconcile/external_resources/integration_secrets_sync.py
@@ -31,11 +31,11 @@ def run(dry_run: bool, thread_pool_size: int) -> None:
     er_inventory = ExternalResourcesInventory(namespaces)
     m_inventory = load_module_inventory(get_modules())
 
-    to_sync_specs = []
-    for key, spec in er_inventory.items():
-        module = m_inventory.get_from_external_resource_key(key)
-        if module.outputs_secret_sync:
-            to_sync_specs.append(spec)
+    to_sync_specs = [
+        spec
+        for key, spec in er_inventory.items()
+        if m_inventory.get_from_external_resource_key(key).outputs_secret_sync
+    ]
 
     reconciler = VaultSecretsReconciler(
         ri=ResourceInventory(),
--- a/reconcile/external_resources/secrets_sync.py
+++ b/reconcile/external_resources/secrets_sync.py
@@ -1,12 +1,11 @@
 import base64
-import dataclasses
 import json
 import logging
 from abc import abstractmethod
 from collections.abc import Iterable, Mapping
-from datetime import datetime
+from datetime import datetime, timezone
 from hashlib import shake_128
-from typing import Any, Optional, cast
+from typing import Any, Optional
 
 from pydantic import BaseModel
 from sretoolbox.utils import threaded
@@ -14,10 +13,17 @@ from sretoolbox.utils import threaded
 from reconcile.external_resources.meta import (
     QONTRACT_INTEGRATION,
     QONTRACT_INTEGRATION_VERSION,
+    SECRET_ANN_IDENTIFIER,
+    SECRET_ANN_PROVIDER,
+    SECRET_ANN_PROVISION_PROVIDER,
+    SECRET_ANN_PROVISIONER,
+    SECRET_UPDATED_AT,
+    SECRET_UPDATED_AT_TIMEFORMAT,
 )
 from reconcile.external_resources.model import ExternalResourceKey
+from reconcile.openshift_base import ApplyOptions, apply_action
 from reconcile.typed_queries.clusters_minimal import get_clusters_minimal
-from reconcile.utils.differ import DiffPair, diff_mappings
+from reconcile.utils.differ import diff_mappings
 from reconcile.utils.external_resource_spec import (
     ExternalResourceSpec,
 )
@@ -33,14 +39,6 @@ from reconcile.utils.vault import (
     _VaultClient,  # noqa
 )
 
-SECRET_ANN_PREFIX = "external-resources"
-SECRET_ANN_PROVISION_PROVIDER = SECRET_ANN_PREFIX + "/provision_provider"
-SECRET_ANN_PROVISIONER = SECRET_ANN_PREFIX + "/provisioner_name"
-SECRET_ANN_PROVIDER = SECRET_ANN_PREFIX + "/provider"
-SECRET_ANN_IDENTIFIER = SECRET_ANN_PREFIX + "/identifier"
-SECRET_UPDATED_AT = SECRET_ANN_PREFIX + "/updated_at"
-SECRET_UPDATED_AT_TIMEFORMAT = "%d/%m/%Y %H:%M:%S"
-
 
 class VaultSecret(BaseModel):
     """Generic class to use the Protocol with Dicts"""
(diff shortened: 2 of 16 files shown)


reconcile/external_resources/secrets_sync.py
```diff
@@ -217,13 +244,41 @@ def build_incluster_secrets_reconciler(
...
+
+    def _read_secret(self, spec: ExternalResourceSpec) -> None:
+        secret_path = f"{self.vault_path}/{spec.cluster_name}/{spec.namespace_name}/{spec.identifier}"
+        try:
+            logging.debug("Reading Secret %s", secret_path)
+            data = self.secrets_reader.read_all({"path": secret_path})
+            spec.secret = data
+        except Exception:
+            msg = f"Error getting secret from vault, skipping. [{secret_path}]"
+            logging.debug(msg)
```
hemslo, line 284: This exception only log debug, how do we get notified if some secret is always missing in vault?

reconcile/external_resources/secrets_sync.py
```diff
@@ -159,56 +165,77 @@ def __init__(
...
+        for spec in [s for s in specs if s.secret]:
+            secret_path = f"{self.vault_path}/{spec.cluster_name}/{spec.namespace_name}/{spec.identifier}"
+            stringified_secret = {k: str(v) for k, v in spec.secret.items()}
+            desired_secret = {"path": secret_path, "data": stringified_secret}
+            self.vault_client.write(desired_secret, decode_base64=False)  # type: ignore[attr-defined]
+
     def sync_secrets(self, specs: Iterable[ExternalResourceSpec]) -> None:
         super().sync_secrets(specs)
-        self._delete_source_secrets()
+        self._write_secrets_to_vault(specs)
```
hemslo, line 228: this call is new, is it missed in previous logic?

reconcile/external_resources/secrets_sync.py
```diff
@@ -99,40 +96,42 @@ def _init_ocmap(self, specs: Iterable[ExternalResourceSpec]) -> OCMap:
...
-
     def sync_secrets(self, specs: Iterable[ExternalResourceSpec]) -> None:
         self._populate_secret_data(specs)
-        ri = ResourceInventory()
-        self._initialize_ri(ri, specs)
-        ocmap = self._init_ocmap(specs)
-        for item in ri:
-            self.reconcile_data(item, ri, ocmap)
+
+        to_sync_specs = [spec for spec in specs if spec.secret]
```
hemslo, line 102: this filter logic duplicated 3 times in this file, worth extracting a method to avoid errors when adding new logic.

reconcile/external_resources/integration_secrets_sync.py
```diff
@@ -0,0 +1,32 @@
...
+    er_inventory = ExternalResourcesInventory(namespaces)
+
+    reconciler = VaultSecretsReconciler(
+        ri=ResourceInventory(),
+        secrets_reader=secrets_reader,
+        vault_path=er_settings.vault_secrets_path,
+        thread_pool_size=thread_pool_size,
+        dry_run=dry_run,
+    )
+    reconciler.sync_secrets(er_inventory.values())
```
hemslo, line 32: Could there be any conflict operations caused by concurrent run of this integration and the external resources integration?

reconcile/external_resources/secrets_sync.py
lechuk47: Vault interaction was put off until now, this PR introduces all vault interactions.

reconcile/external_resources/secrets_sync.py
lechuk47: done.

reconcile/external_resources/integration_secrets_sync.py
lechuk47: It can happen (rarely) that a secret is written by this integration right after the external resources integration writes the secret to the cluster and before writing it to Vault. The last secret will eventually land in the cluster with this integration. 

If this becomes a problem we can think of some optimistic locking approach or other mechanism, but I rather wait to see if this becomes an issue.

hemslo: my understanding of current flow of external resources integration is to sync secrets to cluster first, then save to vault, is it correct?

consider this flow:

* external resources integration sync secret v1 to cluster
* external resources integration save secret v1 to vault
* secrets sync integration sync secret v1 (no op)

Then new changes come in:

* external resources integration sync secret v2 to cluster
* secrets sync integration sync secret v1 to cluster (because vault not updated yet)
* external resources integration save secret v2 to vault
* next secrets sync integration sync secret v2 to cluster

So eventually the secret is correct for this case.

What if some errors happen when the integration write to vault failed after successful sync secrets to cluster?

lechuk47: Yup, I see where are you going. 
I think the easier solution for this is to introduce an intermediate state (like PENDING_SECRET_SYNC) before marking a resource as complete. Then if the secret is written successfully, mark it as CREATED. 
This way the main integration would retry the sync logic until it finishes successfully. WDYT?

reconcile/external_resources/secrets_sync.py
lechuk47: Some modules might not have outputs, hence no secrets in Vault. New specs might have missing Secrets until they are reconciled for the first time. 
I don't see any reason why a Secret can be always missing but you might be right. Would you prefer this as info level? 

* I'm thinking of adding a module configuration attribute to indicate if a module needs to reconcile secrets or not and avoid those in the loop (that solves a part of the problem).

hemslo: +1 to have config so we can tell if a spec has output/secret, until then we can explicit check secret not exist exception instead of all `Exception`, do info log so we can see if it's always missing.

reconcile/external_resources/integration_secrets_sync.py
hemslo: How about save to vault first then sync to cluster? Then cluster secrets should always update after vault value update, reduce the chance of sync old secrets to cluster.

lechuk47: But the main problem still persists. 
If a resource is updated (Terraform) then the Secret needs to land into the cluster ASAP (RDS password for example). 

* If the Vault call fails, the secret will eventually be synced with a wrong value. 
* If the cluster OC fails, the secret can take too much time to land in a cluster 

So to improve this I propose to: 
* Use an additional STATE (e:g PENDING_SECRET_SYNC) that will be set in case the secret sync fails (either OC or VAULT) 
* Secrets Sync integration will query the STATE and only sync resources marked as CREATED. 
* The main integration will retry to sync PENDING_SECRET_SYNC resources at each iteration. Since the intermediate Secret will exist there is no need to run terraform on resources with this state.

hemslo: I'm thinking can we do it without introducing a shared state between 2 integrations.

The requirements are:

1. sync secrets to cluster ASAP
2. eventually write secrets to vault
3. old secrets shouldn't overwrite new secrets

The last action of in cluster reconcile is to delete source secret, can we use it to indicate `PENDING_SECRET_SYNC`? Then requirement 1 and 2 are met.

To avoid old secrets overwrite new secrets, we can use optimistic lock to cover most cases, add a new version field (can be timestamp) to secret, before apply secret to cluster, check if desired version is greater than current version

lechuk47: I introduced the optimistic locking using a timestamp + the PENDING_SECRET_SYNC state in case the secret sync operation fails for a spec. PTAL.

reconcile/external_resources/secrets_sync.py
lechuk47: Added the configuration + changed the exception. PTAL.

reconcile/external_resources/integration_secrets_sync.py
hemslo: the new state `PENDING_SECRET_SYNC` is treated as an error state added later, `CREATED` is the final success state, the current flow is state saved as `CREATED` after resource applied, mark to `PENDING_SECRET_SYNC` if failed, then finally change back to `CREATED` if sync success, this makes the code a little complex as need to do different state transition in https://github.com/app-sre/qontract-reconcile/pull/4375/files#diff-a7fbe2fedf2aacf47fbbe401c068e86e355ad64676b7ca07367c7dfcc4ce800bR278.

How about change the state flow to after resource applied, if it requires secret sync, save state as `PENDING_SECRET_SYNC`, then change it to `CREATED` after sync?

This way you can just pass all keys require sync to `_sync_secrets` as single param, mark them to `CREATED` if no errors.

reconcile/external_resources/secrets_sync.py
```diff
@@ -99,54 +118,95 @@ def _init_ocmap(self, specs: Iterable[ExternalResourceSpec]) -> OCMap:
...
+
+        # Updated_at attribute must be removed before syncing the secret
+        # but is needed afterwards. All specs are copied to preserve the originals.
+        to_sync_specs = [
+            self._copy_spec(spec) for spec in self._specs_with_secret(specs)
+        ]
+
+        for spec in to_sync_specs:
+            self._annotate(spec)
+            self._del_secret_updated_at(spec)
```
hemslo, line 139: this is needed because we embed update time into secret data, would it make sense to add a new root field to spec so we don't need to do add and remove?

reconcile/external_resources/integration_secrets_sync.py
lechuk47: I wanted to avoid an additional DDB update for each spec reconciliation. Since PENDING_SECRET_SYNC is an error state it should not be happening assiduously. 
But I agree that the flow is better understood this way. I changed it to your suggestion.

reconcile/external_resources/integration_secrets_sync.py
```diff
@@ -17,16 +24,24 @@ def run(dry_run: bool, thread_pool_size: int) -> None:
...
+
     namespaces = [ns for ns in get_namespaces() if ns.external_resources]
     er_inventory = ExternalResourcesInventory(namespaces)
+    m_inventory = load_module_inventory(get_modules())
+
+    to_sync_specs = []
+    for key, spec in er_inventory.items():
+        module = m_inventory.get_from_external_resource_key(key)
+        if module.outputs_secret_sync:
+            to_sync_specs.append(spec)
```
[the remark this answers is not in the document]
lechuk47: To me, it's harder to read. I don't see any advantage of writing it this way honestly.

reconcile/external_resources/secrets_sync.py
```diff
@@ -99,54 +118,95 @@ def _init_ocmap(self, specs: Iterable[ExternalResourceSpec]) -> OCMap:
...
-    def _write_secrets_to_vault(self, spec: ExternalResourceSpec) -> None:
-        if spec.secret:
-            secret_path = f"{self.vault_path}/{QONTRACT_INTEGRATION}/{spec.cluster_name}/{spec.namespace_name}/{spec.output_resource_name}"
-            stringified_secret = {k: str(v) for k, v in spec.secret.items()}
-            desired_secret = {"path": secret_path, "data": stringified_secret}
-            self.vault_client.write(desired_secret, decode_base64=False)
+    def _copy_spec(self, spec: ExternalResourceSpec) -> ExternalResourceSpec:
+        copy = dataclasses.replace(spec)
+        copy.secret = dict(spec.secret)
+        return copy
```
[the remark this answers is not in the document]
lechuk47: It does not work because `spec.secret` has an `Init=False` and the attribute can not be used in the constructor, hence not with `replace`.
This code no longer exists so not a problem anymore.

reconcile/external_resources/secrets_sync.py
```diff
@@ -159,56 +219,100 @@ def __init__(
...
                 if decoded.startswith("__vault__:"):
                     _secret_ref = json.loads(decoded.replace("__vault__:", ""))
                     secret_ref = VaultSecret(**_secret_ref)
                     data[k] = self.secrets_reader.read_secret(secret_ref)
                 else:
                     data[k] = decoded
+
+            data[SECRET_UPDATED_AT] = datetime.now().strftime(
+                SECRET_UPDATED_AT_TIMEFORMAT
+            )
```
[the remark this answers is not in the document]
lechuk47: Done.

reconcile/external_resources/secrets_sync.py
```diff
@@ -159,56 +219,100 @@ def __init__(
...
+        self, specs: Iterable[ExternalResourceSpec]
+    ) -> list[ExternalResourceSpec]:
+        try:
+            specs_with_error = super().sync_secrets(specs)
+        except Exception as e:
+            # There is no an easy way to map which secrets have not been reconciled with the specs. If the sync
+            # fails at this stage all the involved specs will be retried in the next iteration
+            logging.error(
+                "Error syncing Secrets to clusters. "
+                "All specs reconciled in this iteration are marked as pending secret syncronization\n%s",
```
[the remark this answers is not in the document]
lechuk47: Fixed.

reconcile/external_resources/secrets_sync.py
```diff
@@ -159,56 +219,100 @@ def __init__(
...
-    def sync_secrets(self, specs: Iterable[ExternalResourceSpec]) -> None:
-        super().sync_secrets(specs)
-        self._delete_source_secrets()
+        for spec in self._specs_with_secret(specs):
+            try:
+                self._write_secret_to_vault(spec)
+                self._delete_source_secret(spec)
+            except Exception as e:
+                key = ExternalResourceKey.from_spec(spec)
+                logging.error("Error writting Secret to Vault. Key: %s.\n%s", key, e)
```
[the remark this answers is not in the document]
lechuk47: Fixed.

reconcile/external_resources/secrets_sync.py
```diff
@@ -159,56 +219,100 @@ def __init__(
...
+                specs_with_error.append(spec)
+                continue
```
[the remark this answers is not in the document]
lechuk47: Fixed.

reconcile/external_resources/integration_secrets_sync.py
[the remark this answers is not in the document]
hemslo: oh, sorry, I thought `module` is appended, then just

```python
    to_sync_specs = [
        spec
        for key, spec in er_inventory.items()
        if m_inventory.get_from_external_resource_key(key).outputs_secret_sync
    ]
```

if get `to_sync_specs` is extracted in a method with some name like `build_to_sync_specs`, then it doesn't matter which syntax to use as we can tell the intention from method name. When inline implementation, it will take a while to understand what these `for` loop and `if` condition are doing, what's the final thing needed later. Using list comprehension makes it easier to understand, as in some other languages, writing `xxx = []` then mutating it in loop will show warnings and auto convert to functional `map` `filter` equivalent code.

reconcile/external_resources/secrets_sync.py
```diff
@@ -22,13 +26,21 @@
...
     _VaultClient,  # noqa
 )
 
+SECRET_ANN_PREFIX = "external-resources"
+SECRET_ANN_PROVISION_PROVIDER = SECRET_ANN_PREFIX + "/provision_provider"
+SECRET_ANN_PROVISIONER = SECRET_ANN_PREFIX + "/provisioner_name"
+SECRET_ANN_PROVIDER = SECRET_ANN_PREFIX + "/provider"
+SECRET_ANN_IDENTIFIER = SECRET_ANN_PREFIX + "/identifier"
+SECRET_UPDATED_AT = SECRET_ANN_PREFIX + "/updated_at"
+SECRET_UPDATED_AT_TIMEFORMAT = "%d/%m/%Y %H:%M:%S"
```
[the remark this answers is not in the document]
hemslo: btw, if you want to keep it same as our json schema `date-time` or k8s timestamp format, then this has to be a timezone aware string, `.isoformat()` will produce `20240101T00:00:00+00:00`, you can also specify `Z` explicitly as we only use UTC `%Y-%m-%dT%H:%M:%SZ`.

And because utc iso time string is comparable, we don't even need to parse it back to `datetime`, can directly do string compare in time check.

reconcile/test/external_resources/test_secrets_sync.py
```diff
@@ -0,0 +1,107 @@
...
+
+# @fixture
+# def vault_secrets_reconciler() -> VaultSecretsReconciler:
+#     return VaultSecretsReconciler(
+#         ri=Mock(),
+#         secrets_reader=Mock(),
+#         vault_path="/path",
+#         thread_pool_size=1,
+#         dry_run=False,
+#     )
```
hemslo, line 70: WIP?

reconcile/utils/external_resource_spec.py
```diff
@@ -87,6 +87,11 @@ class ExternalResourceSpec:
     resource: MutableMapping[str, Any]
     namespace: Mapping[str, Any]
     secret: Mapping[str, str] = field(init=False, default_factory=lambda: {})
+    # Metadata is used for processing data that shuold not be included in the secret data
+    # e.g: ERV2 adds a created_at attribute that acts as optimistic lock.
```
hemslo, line 91: `created_at` in comment, `updated_at` in code

reconcile/external_resources/integration_secrets_sync.py
hemslo: We can improve update performance later if needed, like use batch update instead of update one by one to reduce requests

reconcile/external_resources/secrets_sync.py
hemslo: I thought optimistic locking only needed in vault to cluster, in code it's reused for both reconciler, but shouldn't be any issues

hemslo (approved):
LGTM!

reconcile/external_resources/integration_secrets_sync.py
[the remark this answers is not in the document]
lechuk47: Added.

reconcile/test/external_resources/test_secrets_sync.py
lechuk47: Removed. It was an old fixture.

reconcile/external_resources/secrets_sync.py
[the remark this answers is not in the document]
lechuk47: TIL. Added.

PR #4380, opened 2024-05-22
requirements-test: botocore match version from setup.py

jfchevrette:
This also addresses the same issue which was fixed some time ago in #4095


jfchevrette:
addressed by #4383

PR #4384, opened 2024-05-23
[qontract-cli] get systems-and-tools

maorfr:
related to https://issues.redhat.com/browse/SDE-3947

with this cli command we will create an app-interface-output report for all systems and tools that are managed via app-interface.

example output: https://gitlab.cee.redhat.com/service/app-interface-output/-/merge_requests/42


esron (approved):
LGTM

PR #4386, opened 2024-05-23
✨ openshift-resources: bugfix

chassing:
This PR fixes the mocking of some jinja2 template functions during the PR checks. The goal is to prevent any external requests to speedup them. 

Initially, the plan was to add an extended early exit, but the integration needs a lot of time to compute the desired state, which would also be used in the extended early exit; therefore it doesn't make any sense.

Ticket: [APPSRE-8726](https://issues.redhat.com/browse/APPSRE-8726)

Review commit by commit:

1. :recycle: openshift-resources-base: add type hints
2. fix early exit

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -7,8 +7,11 @@ import re
 import sys
 from collections import defaultdict
 from collections.abc import (
+    Callable,
+    Generator,
     Iterable,
     Mapping,
+    MutableMapping,
     Sequence,
 )
 from contextlib import contextmanager
@@ -237,23 +240,23 @@ KUBERNETES_SECRET_DATA_KEY_RE = "^[-._a-zA-Z0-9]+$"
 _log_lock = Lock()
 
 
-def _locked_info_log(msg: str):
+def _locked_info_log(msg: str) -> None:
     with _log_lock:
         logging.info(msg)
 
 
-def _locked_debug_log(msg: str):
+def _locked_debug_log(msg: str) -> None:
     with _log_lock:
         logging.debug(msg)
 
 
-def _locked_error_log(msg: str):
+def _locked_error_log(msg: str) -> None:
     with _log_lock:
         logging.error(msg)
 
 
 class FetchRouteError(Exception):
-    def __init__(self, msg):
+    def __init__(self, msg: Any):
         super().__init__("error fetching route: " + str(msg))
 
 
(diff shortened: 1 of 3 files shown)


reconcile/openshift_resources_base.py
```diff
@@ -1198,57 +1196,22 @@ def _early_exit_fetch_resource(spec: Sequence, settings: Mapping) -> dict[str, s
...
+        jinja2_utils,
+        lookup_secret=DEFAULT,
+        lookup_github_file_content=DEFAULT,
+        url_makes_sense=DEFAULT,
+        lookup_s3_object=DEFAULT,
+    ) as mocks:
+        mocks["lookup_secret"].return_value = "secret"
+        mocks["lookup_github_file_content"].return_value = "file-content"
+        mocks["url_makes_sense"].return_value = False
+        mocks["lookup_s3_object"].return_value = "s3-object"
```
hemslo, line 1209: previously we return value based on called args, so even not do real fetch, still return some identifier back, do we need that?

chassing: I think we don't need that.

patjlm: I think it's interesting to keep a return value that relates to arguments. If a template is updated with a parameter change of one of these calls (eg a vault secret version), we want the change to be detected and no early exit. WDYT?

chassing: Good point!

chassing: Changed in [30dd5a2](https://github.com/app-sre/qontract-reconcile/pull/4386/commits/30dd5a214b04cebd78d10f7081e4ad3f9f474afc)

patjlm: so in the end, what do we fix in early exit?
I agree this refactor is nicer than teh initial code, but I don't think it fixes anything, right?

chassing: It just fixes the mocking of the template functions (`vault,` `github,`...), and this should now speed up the calculation of the early exit state.

hemslo (approved):
LGTM!

reconcile/openshift_resources_base.py
patjlm: I don't see how we fix it. It used to return a f-string, and it still does.

But fine, I like the use of mocks much more ;-)

patjlm: aaaaaah cool, thanks!

PR #4387, opened 2024-05-23
[status-page-maintenances] do not set reminders if there are no components

maorfr:

reminders are for components. if there are no affected components - there can be no notifications.


esron (approved):
LGTM

PR #4392, opened 2024-05-27
support group permission to codeComponents in gitlab_permission

mehfuz:
PR consists modification to `gitlab-permissions` integration.
Instead of adding every member from `app-sre`, the integration will direct share the project with group with `maintainer` role.  This can be done by enabling the feature flag.
ticket : - (https://issues.redhat.com/browse/APPSRE-10300)
Context :-  https://redhat-internal.slack.com/archives/GGC2A0MS8/p1716387049100519?thread_ts=1716383243.038439&cid=GGC2A0MS8


reconcile/utils/gitlab_api.py
```diff
@@ -232,6 +232,7 @@ def get_project_maintainers(
             project = self.get_project(repo_url)
         if project is None:
             return None
+        gitlab_request.labels(integration=INTEGRATION_NAME).inc()
```
hemslo, line 235: what's this count for? `get_items` has this label already

reconcile/utils/gitlab_api.py
```diff
@@ -250,6 +251,46 @@ def get_group_if_exists(self, group_name: str) -> Group | None:
         except gitlab.exceptions.GitlabGetError:
             return None
 
+    def share_project_with_group(
+        self, repo_url: str, group_id: int, dry_run: bool, access: str = "maintainer"
+    ) -> None:
+        gitlab_request.labels(integration=INTEGRATION_NAME).inc()
```
hemslo, line 257: same here, `get_project` already inc counter

reconcile/utils/gitlab_api.py
```diff
@@ -250,6 +251,46 @@ def get_group_if_exists(self, group_name: str) -> Group | None:
...
+            return None
+        access_level = self.get_access_level(access)
+        gitlab_request.labels(integration=INTEGRATION_NAME).inc()
+        # check if we have 'access_level' access so we can  add the group with same role.
+        members = self.get_items(
+            project.members.all, query_parameters={"user_ids": self.user.id}
+        )
+        if self.user.id not in [
+            member.id for member in members if member.access_level >= access_level
+        ]:
```
hemslo, line 269: can be checked using `any`

```python
if not any(self.user.id == member.id and member.access_level >= access_level for member in members):
```

reconcile/utils/gitlab_api.py
```diff
@@ -250,6 +251,46 @@ def get_group_if_exists(self, group_name: str) -> Group | None:
...
+            logging.error(
+                "'{}' is not shared with {} as 'Maintainer'".format(
+                    repo_url, self.user.username
+                )
+            )
```
hemslo, line 274: shouldn't hard code `Maintainer` as it's can be other `access`, and no need to call `format` in `logging.error`

```python
            logging.error(
                "%s is not shared with %s as %s",
                repo_url,
                self.user.username,
                access,
            )
```

reconcile/utils/gitlab_api.py
```diff
@@ -250,6 +251,46 @@ def get_group_if_exists(self, group_name: str) -> Group | None:
...
+        self, group_name: str
+    ) -> tuple[int, list[dict]]:
+        gitlab_request.labels(integration=INTEGRATION_NAME).inc()
+        group = self.gl.groups.get(group_name)
+        return group.id, [
+            project
+            for project in group.shared_projects
+            for shared_group in project["shared_with_groups"]
+            if shared_group["group_id"] == group.id
+            and shared_group["group_access_level"] >= 40
```
hemslo, line 291: magic number `40` to `gitlab.MAINTAINER_ACCESS`

reconcile/utils/gitlab_api.py
```diff
@@ -250,6 +251,46 @@ def get_group_if_exists(self, group_name: str) -> Group | None:
         except gitlab.exceptions.GitlabGetError:
             return None
 
+    def share_project_with_group(
```
hemslo, line 254: new api methods require tests

reconcile/utils/gitlab_api.py
```diff
@@ -250,6 +251,46 @@ def get_group_if_exists(self, group_name: str) -> Group | None:
...
+                    repo_url, self.user.username
+                )
+            )
+            return None
+        logging.info(["add_group_as_maintainer", repo_url, group_id])
+        if not dry_run:
+            gitlab_request.labels(integration=INTEGRATION_NAME).inc()
+            project.share(group_id, access_level)
+
+    def get_group_id_and_shared_projects(
```
hemslo, line 281: new api methods require tests

reconcile/gitlab_permissions.py
```diff
@@ -1,64 +1,26 @@
...
-        get_members_to_add, repos, thread_pool_size, gl=gl, app_sre=app_sre
-    )
-    members_to_add = list(itertools.chain.from_iterable(results))
-    for m in members_to_add:
-        logging.info(["add_maintainer", m["repo"], m["user"].username])
-        if not dry_run:
-            gl.add_project_member(m["repo"], m["user"])
+    group_id, shared_projects = gl.get_group_id_and_shared_projects(APP_SRE_GROUP_NAME)
+    shared_project_repos = [project["web_url"] for project in shared_projects]
+    repos_to_share = [repo_url for repo_url in set(repos) - set(shared_project_repos)]
```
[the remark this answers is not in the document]
hemslo: since `shared_project_repos` is only used to calculate `repos_to_share`, no need to create list first then convert to set, can directly construct set `{project["web_url"] for project in shared_projects}`

reconcile/test/utils/test_gitlab_api.py
```diff
@@ -468,3 +470,70 @@ def test_get_group_members(
...
+    mocker: MockerFixture, mocked_gitlab_api: GitLabApi
+):
+    project = create_autospec(Project)
+    project.members = create_autospec(ProjectMemberManager)
+    project.members.all.return_value = [
+        create_autospec(ProjectMember, id=mocked_gitlab_api.user.id, access_level=40)
+    ]
+    mocker.patch(
+        "reconcile.utils.gitlab_api.GitLabApi.get_project"
+    ).return_value = project
```
hemslo, line 485: can set `mocked_gitlab_api.project = project`

reconcile/test/utils/test_gitlab_api.py
```diff
@@ -468,3 +470,70 @@ def test_get_group_members(
...
+    mocker.patch(
+        "reconcile.utils.gitlab_api.GitLabApi.get_access_level"
+    ).return_value = 40
```
hemslo, line 488: is this patch needed? same for later test cases

mehfuz: if this patch is not added it treats the value of `get_access_level` as a `MagicMock` and not as `int`. As a result our comaprison at https://github.com/app-sre/qontract-reconcile/blob/64ea350e9a104f41c32458db9a4f43fdab072342/reconcile/utils/gitlab_api.py#L265 and 291 Fails. So i mocked this one seperately to return int

reconcile/test/utils/test_gitlab_api.py
mehfuz: the value for `project` comes from `get_project` so needs to mock this . I think `mocked_gitlab_api.project = project` is valid in case we need `self.project`. In the methods we use `self.get_project()`.

hemslo: the naming is a little confusing, `mocked_gitlab_api` itself is not mock, it's the real `GitLabApi` object we want to test, we shouldn't patch another method belongs to `GitLabApi`, should setup `mocked_gl`

reconcile/test/utils/test_gitlab_api.py
hemslo: that's because `gitlab` is being patched, here we shouldn't patch another `GitLabApi` method, since all those values just const, we can update import in `gitlab_api.py` to

```python
from gitlab.const import MAINTAINER_ACCESS
```

so this logic doesn't require patch

reconcile/gitlab_permissions.py
```diff
@@ -50,6 +52,19 @@ def run(dry_run, thread_pool_size=10, defer=None):
...
         defer(gl.cleanup)
     repos = queries.get_repos(server=gl.server, exclude_manage_permissions=True)
+    toggle_state = get_feature_toggle_state(
+        "gitlab-permissions-share-with-group",
+        default=False,
+    )
+    if not toggle_state:
+        share_project_with_group(gl, repos, dry_run)
+    else:
+        share_project_with_group_members(gl, repos, thread_pool_size, dry_run)
```
hemslo, line 62: When `toggle_state` is `True`, it should mean `share_project_with_group`, but `if` condition is in the opposite, better to rename it to something meaningful like `should_share_project_with_group` or just inline condition check

hemslo: @mehfuz what I mean is variable name is not meaningful, by looking at `toggle_state`, can't tell what does it represent, feature toggle name is ok

hemslo (approved):
LGTM!

reconcile/utils/gitlab_api.py
```diff
@@ -250,6 +251,45 @@ def get_group_if_exists(self, group_name: str) -> Group | None:
...
+            return None
+        logging.info(["add_group_as_maintainer", repo_url,group_id])
+        if not dry_run:
+            gitlab_request.labels(integration=INTEGRATION_NAME).inc()
+            project.share(group_id, access_level)
+
+    def get_group_id_and_shared_projects(
+        self, group_name: str
+    ) -> tuple[int, list[dict]]:
+        group = self.gl.groups.get(group_name)
```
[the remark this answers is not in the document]
mehfuz: @patjlm will try to come up with something about this in upcoming MRs..

PR #4396, opened 2024-05-28
APPSRE-10117 soakDays for saasherder

fishi0x01:
- saasherder gating promotions with accumulated soakDays
- fixing a bug in which certain promotion criteria can skip validation of following promotions.

### DIFF
--- a/reconcile/utils/saasherder/saasherder.py
+++ b/reconcile/utils/saasherder/saasherder.py
@@ -1881,13 +1881,10 @@ class SaasHerder:  # pylint: disable=too-many-public-methods
             # TODO: add environment.parameters to the include list!?!?
         )
 
-    def _validate_promotion(self, promotion: Promotion | None) -> bool:
+    def _validate_promotion(self, promotion: Promotion) -> bool:
         # Placing this check here to make mypy happy
         if not (self.state and self._promotion_state):
             raise Exception("state is not initialized")
-
-        if not promotion:
-            return True
 
         if not promotion.subscribe:
             return True
--- a/reconcile/utils/saasherder/saasherder.py
+++ b/reconcile/utils/saasherder/saasherder.py
@@ -1915,13 +1915,8 @@ class SaasHerder:  # pylint: disable=too-many-public-methods
                             )
                             has_errors = True
                             continue
-                        check_in = (
-                            datetime.fromisoformat(deployment.check_in)
-                            if deployment.check_in
-                            else None
-                        )
-                        if check_in:
-                            passed_soak_days += now - check_in
+                        if check_in := deployment.check_in:
+                            passed_soak_days += now - datetime.fromisoformat(check_in)
                         if deployment.target_config_hash:
                             config_hashes.add(deployment.target_config_hash)
 
--- a/reconcile/utils/saasherder/saasherder.py
+++ b/reconcile/utils/saasherder/saasherder.py
@@ -1881,7 +1881,11 @@ class SaasHerder:  # pylint: disable=too-many-public-methods
             # TODO: add environment.parameters to the include list!?!?
         )
 
-    def _validate_promotion(self, promotion: Promotion) -> bool:
+    def _validate_promotion(self, promotion: Promotion | None) -> bool:
+        # Placing this check here to make mypy happy
+        if not (self.state and self._promotion_state):
+            raise Exception("state is not initialized")
+
         if not promotion:
             return True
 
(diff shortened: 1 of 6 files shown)


reconcile/utils/saasherder/saasherder.py
```diff
@@ -1884,12 +1885,17 @@ def validate_promotions(self) -> bool:
         if not (self.state and self._promotion_state):
             raise Exception("state is not initialized")
 
+        has_errors = False
```
hemslo, line 1888: save errors to `has_errors` and mutate in each iteration, with too many `continue` in nested loop is a little confusing. consider extract a method `validate_promotion(self, promotion) -> bool`, then you can do 

```python
return any(self.validate_promotion(promotion) for promotion in self.promotions)
```

if you don't want to early terminate validations, then collect results to list first

```python
        errors = [self.validate_promotion(promotion) for promotion in self.promotions]
        return any(errors)
```

reconcile/utils/saasherder/saasherder.py
```diff
@@ -1904,7 +1910,15 @@ def validate_promotions(self) -> bool:
...
-                            return False
+                            has_errors = True
+                            continue
+                        check_in = (
+                            datetime.fromisoformat(deployment.check_in)
+                            if deployment.check_in
+                            else None
+                        )
+                        if check_in:
+                            passed_soak_days += now - check_in
```
hemslo, line 1921: can simplify with one condition check:

```python
                        if check_in := deployment.check_in:
                            passed_soak_days += now - datetime.fromisoformat(check_in)
```

reconcile/utils/saasherder/saasherder.py
fishi0x01: Would like to avoid bigger refactors in the validate function at this point - considering this now more of a styling issue.

reconcile/utils/saasherder/saasherder.py
fishi0x01: Applied

reconcile/utils/saasherder/saasherder.py
hemslo: and if we let `validate_*` method return `True` means valid, `False` means invalid, then it's even more readable


```python
        return all(self.validate_promotion(promotion)
                   for promotion in self.promotions
                   if promotion is not None and promotion.subscribe)
```

and both `validate_promotions` and `validate_promotion` are easier to read and test.

fishi0x01: I hear you and I agree. But I would really love to get unblocked on this :bow: without adding much more stuff around saasherder. This is not a functional problem. Im happy to open a ticket to invest more time into saasherder complexity in general.

fishi0x01: > is it covered by any tests?
All saasherder tests are currently single publisher based, so this scenario is not covered.
Ive added multiple publisher scenario in the new tests, but they focus on soakDays. Will now see if I can add a generalized test for multi publisher.

hemslo: a new pr just touch this code block after current pr merged is good, no need for a big refactor, whenever adding new logic, should maintain complexity instead of letting it grow. similar to test coverage, if we have code quality report, the goal is not to make it worse for new changes.

hemslo (approved):
LGTM!

fishi0x01:
@hemslo Ive added the extra function as you suggested. Also extended existing hash config tests with a multi-channel saas file.

Thank you for all the input! :bow: :heart:

reconcile/utils/saasherder/saasherder.py
fishi0x01: Test adjusted with multi-channel examples.

reconcile/utils/saasherder/saasherder.py
```diff
@@ -1877,82 +1878,101 @@ def sanitize_namespace(
...
-                    # The parent at the same ref has succeed if this code
-                    # is reached though.
-                    if not parent_saas_config:
-                        logging.info(
-                            "Parent Saas config missing on target "
-                            "rely on the success state only"
-                        )
-                        return True
+        if not promotion:
+            return True
```
hemslo, line 1887: note you can mark `promotion` as `Promotion` to only support not None as we already filtered out None values when use.

fishi0x01: Good catch! Fixed

PR #4398, opened 2024-05-28
expose flatten to openshift resource templating

maorfr:
allows converting structures from data files into flat dicts.

a possible use case is to take a cluster upgradePolicy and use it as labels.


mehfuz (approved):

patjlm:
note that ansible exposes a `flatten` jinja filter that does something different:

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_filters.html#managing-list-variables

PR #4404, opened 2024-05-29
AUS dashboard: fix the failure count

patjlm:
Noticed a non-null value in this panel while the one next to it did not show any failed cluster upgrade.
Using an instant query instead of a range seems to fix this


geoberle (approved):
*grafana transcendence*

PR #4405, opened 2024-05-30
do not raise on no pagerduty instances

maorfr:
part of https://issues.redhat.com/browse/SDE-3947

in an attempt to run `qc get systems-and-tools` on sre-capabilities, there is no pagerduty in that repository.
generally, the code should accept multiple situations. slack-usergroups should be able to run for repos without pagerduty.

this PR makes the behavior more forgiving.


hemslo (approved):
LGTM!

patjlm (approved):

PR #4406, opened 2024-05-30
[qontract-cli] get systems-and-tools collect clusters from ocm organizations

maorfr:
part of https://issues.redhat.com/browse/SDE-3947

depends on https://github.com/app-sre/qontract-schemas/pull/660

with this addition we can run the report on sre-capabilities and expose OSD-FM managed clusters.

output: https://gitlab.cee.redhat.com/service/sre-capabilities-output/-/blob/main/systems-and-tools.md

PR #4407, opened 2024-05-30
fix create_ocm_upgrade_scheduler_org_updates_mr

maorfr:
fixes #4406

not all messages have the `id` and `url` fields, like `delete` actions.

only use these keys when it's an `add` action to avoid `KeyError`.


cubismod (approved):
LGTM

PR #4410, opened 2024-05-31
Fix terraform vpc resources exit conditions

esron:
Related to: [APPSRE-8743](https://issues.redhat.com/browse/APPSRE-8743)

After the deployment I've found some issues in the integration code:

Due to the way sharding is done in production the integration floods the reconcile channel in Slack. That is because most of the accounts don't have vpc-requests and if you pass one account (what is done in the shards) the integration raises an error. That was done to warn the user that they are not creating anything and do an early exit.

Also, if the integration is run without shards but there is no vpc-request defined, i.e. all the previously defined vpc-requests are deleted in app-interface, the integration skips the terraform apply step. That was also done to early exit.

### DIFF
--- a/reconcile/gql_definitions/common/aws_vpc_requests.py
+++ b/reconcile/gql_definitions/common/aws_vpc_requests.py
@@ -33,6 +33,7 @@ fragment TerraformState on TerraformStateAWS_v1 {
 
 fragment VPCRequest on VPCRequest_v1 {
   identifier
+  delete
   account {
     name
     uid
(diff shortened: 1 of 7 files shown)


OliviaHY (approved):

OliviaHY:
Want to point out that with this implementation, the deletion of an VPC is done via deleting a VPC request. It's fine but it's different from how App Interface users would normally deleting a TF resource. So we will need to emphasize that in documentation.

PR #4416, opened 2024-06-03
ignore missing integration file when not validating schemas

maorfr:
for usage with `qontract-reconcile --no-validate-schemas`

### DIFF
--- a/reconcile/utils/gql.py
+++ b/reconcile/utils/gql.py
@@ -122,7 +122,7 @@ class GqlApi:
                     self._valid_schemas = integration["schemas"]
                     break
 
-            if not self._valid_schemas:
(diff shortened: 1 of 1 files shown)


rporres:
What's the use case of not validating schemas? 🤔

maorfr:
either "development" or running integrations without an app-interface instance, which is only in my imagination right now.

nonetheless, this option already exists, so might as well do what it is expected?

rporres:
>  nonetheless, this option already exists, so might as well do what it is expected?

or we can delete it 😂 

It is fine, though. I don't want to put barriers to your imagination

rporres (approved):

PR #4418, opened 2024-06-04
fix parent group exception for gitlab-permissions

mehfuz:
fixes below issue while sharing project with app-sre group
```
gitlab.exceptions.GitlabCreateError: 409: Project cannot be shared with the group it is in or one of its ancestors.
```

### DIFF
--- a/reconcile/gitlab_permissions.py
+++ b/reconcile/gitlab_permissions.py
@@ -77,9 +77,11 @@ def share_project_with_group_members(
 
 
 def share_project_with_group(gl: GitLabApi, repos: list[str], dry_run: bool) -> None:
+    # get repos not owned by app-sre
+    non_app_sre_projects = {repo for repo in repos if "/app-sre/" not in repo}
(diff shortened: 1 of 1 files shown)


maorfr (approved):
LGTM

after we stabilize the integration, i'd like us to think how we can generalize it so it can be used by our customers.

reconcile/gitlab_permissions.py
```diff
@@ -77,9 +77,11 @@ def share_project_with_group_members(
 
 
 def share_project_with_group(gl: GitLabApi, repos: list[str], dry_run: bool) -> None:
+    # get repos not owned by app-sre
+    non_app_sre_projects = {repo for repo in repos if "/app-sre/" not in repo}
```
hemslo, line 81: can we get this via gitlab api instead of check url string?

mehfuz: As discussed...Will be taken care in generic implementation...

PR #4420, opened 2024-06-04
APPSRE-8094: Test Vector for fluentd replacement

lisa:
For this test, we will run vector and fluentd side-by-side (each in their own sidecar). This test will not be for a long period of time - just long enough to understand if the Vector config (and helm/deployment config) is working correctly.

Vector maintains the same, high level functionality as fluentd in that it will still deliver events to Slack, Google Chat and Cloudwatch logs. In this effort, the message format has been unified between gchat and slack (per team consensus).

https://github.com/app-sre/qontract-reconcile/pull/4420/files#r1691743772 is an annotated description of the nasty looking Vector config in the transform block.


helm/qontract-reconcile/templates/template.yaml
```diff
@@ -228,6 +322,8 @@ objects:
           volumeMounts:
           - name: fluentd-config
             mountPath: /fluentd/etc/
+          - name: vector-config
+            mountPath: /etc/vector/
```
patjlm, line 326: i think you are missing the corresponding emptyDir volume definition

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -773,6 +902,14 @@ parameters:
   value: ''
 - name: PUSHGATEWAY_SECRET_NAME
   value: push-gateway-auth-url
+### APPSRE-8096: Vector Section Begins
+- name: VECTOR_IMAGE
+  value: timberio/vector
```
cubismod, line 907: Please mirror the Vector image to Quay since we can't use Dockerhub images in App Interface.

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -127,8 +127,102 @@ objects:
...
+                  source: -"using gql endpoint" AND -"Certificate did not match expected hostname"
+
+              ### Craft a JSON document that has escaped double quotes and changes the event from the 
+              ### Vector-default `message` to `text`, which is require by Slack and Google Chat.
+              rename_message_webhooks:
+                type: remap
+                inputs:
+                  - despam_events
+                source: |-
+                  .message = join!(
```
cubismod, line 161: took me a minute to realize what the `!` meant, for future reviewers: https://vector.dev/docs/reference/vrl/expressions/#function_call-grammar

helm/qontract-reconcile/templates/template.yaml
TGPSKI: In addition, we need to build this image from ubi9 or eus base images for these changes to be accepted in FedRAMP

TGPSKI: It could be a great opportunity to build vector with konflux, fwiw

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -127,8 +127,102 @@ objects:
...
+                  [
+                    "{\"text\":\"",
+                    format_timestamp!(
+                      now(),
+                      "%Y-%m-%d %H:%M:%S %z"
+                    ),
+                    " ",
```
cubismod, line 168: why the extra space here?

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -127,8 +127,102 @@ objects:
...
+                    " [${POD_NAME}] ",
+                    replace!(
+                      del(.message),
```
cubismod, line 171: Okay this one is kinda throwing me through a loop as I'm understanding VRL. So this basically means that the `.message` field is removed from this event and returned from the `del` function. And the end result is that `.message` will equal the result of this join? Right?

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -127,8 +127,102 @@ objects:
...
+                encoding:
+                  codec: text
+                  only_fields:
+                    - message
+                method: post
+                uri: ${SLACK_WEBHOOK_URL}
+            {{- end }}
+
+            {{- if $logs.googleChat }}
+              ################
```
cubismod, line 208: Would it be possible to add filtering for Google Chat messages so that only ERROR level messages get sent? See https://github.com/app-sre/qontract-reconcile/pull/4420/files#diff-1b43156729c75dded2e7c94742402e1f8b0de449917632dccbe032d231396bd0R289 for how we do it with Vector.

Reason for this is that GChat really sucks at displaying log messages in real-time and will crash your browser lol

cubismod: Additionally, have you considered buffering to disk? We do so currently with GChat and Fluentd since GChat has some pretty strict rate limits. I worry a bit about the default of buffering to memory with a 120Mi limit.

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -773,6 +902,14 @@ parameters:
   value: ''
 - name: PUSHGATEWAY_SECRET_NAME
   value: push-gateway-auth-url
+### APPSRE-8096: Vector Section Begins
+- name: VECTOR_IMAGE
+  value: timberio/vector
+- name: VECTOR_IMAGE_TAG
+  value: 0.38.0-debian
+- name: VECTOR_IMAGE_PULL_POLICY
+  value: Always
```
TGPSKI, line 911: Do we use `IfNotPresent` elsewhere?

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -127,8 +127,102 @@ objects:
...
+                encoding:
+                  codec: text
+                  only_fields:
+                    - message
+                method: post
+                uri: https://${GOOGLE_CHAT_WEBHOOK_URL}&threadKey=${POD_NAME}
+            {{- end}}
+            EOF
+            ### APPSRE-8096: Vector Section Ends
+            echo "Creating /fluentd/etc/fluent.conf"
```
TGPSKI, line 225: Is this line important? It wasn't present before?

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -127,8 +127,102 @@ objects:
                 key: log_group_name
           command: ["/bin/sh", "-c"]
           args:
+          # generate /etc/vector/vector.yaml
+          # and generate /fluentd/etc/fluent.conf
```
hemslo, line 131: why we generate these config files in init container instead of put it in a configmap and mount it?

cubismod: Probably one of those choices that was made years and years ago and we didn't really think about it until now

helm/qontract-reconcile/templates/template.yaml
lisa: Up on line 161 there is a `join!` command which is joining an array of strings into one single string. The `" ",` includes a literal "` `" into the array so that there's a space after the timestamp and before the `[PODNAMEHERE]` string.

helm/qontract-reconcile/templates/template.yaml
lisa: To summarize:

In VRL, Vector considers the `join` method to have the potential to be "failable." (I'm not quite sure _how_ it can fail, since it joins multiple strings into one, but there you go.) VRL also supports appending an `!` to certain methods to discard any possible errors ("failures"). It's equivalent to Golang's `result, _ := methodWithError()`.

helm/qontract-reconcile/templates/template.yaml
lisa: Not quite. Let me annotate the entirety of the `.message = join!(....)`:

```
161 .message = join!(
162 [
163   "{\"text\":\"",
164   format_timestamp!(
165     now(),
166     "%Y-%m-%d %H:%M:%S %z"
167   ),
168   " ",
169   " [${POD_NAME}] ",
170   replace!(
171     del(.message),
172     "\"",
173     "\\\""),
174   "\"}"
175 ])
```

Example log message from /var/log/system.log
```
Jul 25 11:36:39 rhodium syslogd[143]: ASL Sender Statistics
```

Line 161: By default, when Vector emits a logging event, the pathway from the "source" to the "sink(s)" involves creating a sort of object on which is a `message` attribute. This `message` attribute is the content of the event, such as a log line. In the example, `message` would be equal to `ASL Sender Statistics`.

This poses a problem because when using the webhooks of Slack and Google Chat to transmit the log message, each endpoint wants the payload to be a JSON document in the format of `{"text": "ASL Sender Statistics"}`. Additionally, the sink wants the source of that payload to be the contents of the `message` attribute. That means that we must "transform" the `message` to some other value.

The way to construct the JSON payload is to join a bunch of strings together into one final payload document, and that begins on this line 161 by reassigning the value of the `message` attribute to the result of a `join`. The `!` is required to discard any potential errors that _could_ occur (as far as Vector is concerned) in the `join` process.

Line 162: Declare the start of the array that will be joined.

Line 163: Escape double quotes to set the first element of the array to `{"text":"}`. Note the trailing `"` which declares the start of the JSON value string for `text` JSON key.

Line 164-167: Format the current timestamp with the standard `strftime(3)` expansions. Here, again, the `!` tells Vector to dicard any possible errors from the `format_timestamp` method. The reulting string is the second element in the array-to-be-joined, and represents the first thing in the `text` payload string.

Line 168: Insert a space immediately after the timestamp as the third array element.

Line 169: Reference and interpolate the environment variable, `POD_NAME`, (which is set in the `PodSpec`) and include that in the `text` payload.

Line 170-173: If the `message` contained a doublequote (`"`) character, it would break the construction of the JSON document we're making. That means we must globally `replace` every `"` in the `message` with an escaped version of it.

Line 170: The `replace!` method takes the format of `replace!(haystack, needle, replacement)` where all needles in the haystack are replaced. As before, the `!` discards any potential errors from the `replace` method.

Line 171: `del(.message)` deletes _and returns_ from the object the `message`. This return value serves as an input into the `replace!` method on line 170.

Line 172: Search for literal `"`

Line 173: Replace them with a literal `\"`

Line 174: The final element in the join element is `"}`, thus closing the opened string on line 163.

Line 175: Terminate the array, and close the arguments into the `join!` call from line 161.

reconcile/test/fixtures/helm/aws_account_shard_disabled.yml
```diff
@@ -47,8 +47,68 @@ objects:
                 key: log_group_name
           command: ["/bin/sh", "-c"]
           args:
+          # generate /etc/vector/vector.yaml
```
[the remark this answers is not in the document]
lisa: It may be possible to come up with a way to auto-generate all 28 test fixtures, but that is very likely to be out of scope for this spike.

helm/qontract-reconcile/templates/template.yaml
lisa: Possibly to reference environment variables such as `POD_NAME`? Does Fluentd support the referencing of those from inside its normal rules?

helm/qontract-reconcile/templates/template.yaml
lisa: They're here only for the case we find that we must excise Vector during or immediately after the evaluation period. It makes it very clear where the section begins and ends. It probably made more sense before needing to adjust 28 individual fixture files.

helm/qontract-reconcile/templates/template.yaml
cubismod: Gotcha, appreciate the explanation!

helm/qontract-reconcile/templates/template.yaml
cubismod: Thank you for taking the time to explain all of this. Makes sense to me now

helm/qontract-reconcile/templates/template.yaml
cubismod: Oh that's a good point. `POD_NAME` is defined as an environment variable for the init container rather than an OpenShift Template parameter so I don't think it would work as just a `ConfigMap`: https://github.com/app-sre/qontract-reconcile/blob/52d03a93cfe8fbc4f99934ad45c498425166a821/helm/qontract-reconcile/templates/template.yaml#L118

helm/qontract-reconcile/templates/template.yaml
cubismod: Gotcha, makes sense to me

helm/qontract-reconcile/templates/template.yaml
lisa: I'll look into these as they're both clearly blockers. Not sure about Konflux, but we'll see. I have no preference either way.

helm/qontract-reconcile/templates/template.yaml
hemslo: both fluentd and vector support env vars in config

* https://docs.fluentd.org/quickstart/faq#how-can-i-use-environment-variables-to-configure-parameters-dynamically
* https://vector.dev/docs/reference/configuration/#environment-variables

I think it's easier to read and debug for config file stored in configmap instead of shell script. If we can try it for vector, then all fluented related components can be left unchanged. But there maybe other unknown issues with that, up to you to decide.

helm/qontract-reconcile/templates/template.yaml
lisa: 👍 I'll resolve the convo, but link this comment thread in the main PR text.

helm/qontract-reconcile/templates/template.yaml
lisa: @hemslo There is an issue with using a `ConfigMap` because the config for Vector and Fluentd differ based on what the helm template values are. For instance, it isn't always true that logs will be sent to Slack, and so that config should be omitted. There's not a real good way to do that with a mounted ConfigMap.

helm/qontract-reconcile/templates/template.yaml
lisa: I'll look at the Error filtering.

I did look briefly at buffering to disk in local testing and for some reason it didn't quite work out? Maybe that was simply due to the small scale of local testing. I can add it in and we can see how it goes.

helm/qontract-reconcile/templates/template.yaml
lisa: I'll make this change as well.

helm/qontract-reconcile/templates/template.yaml
lisa: I'll be using https://catalog.redhat.com/software/containers/openshift-logging/vector-rhel9/6447991f84aeefde2c9f2bcf?container-tabs=overview going forward.

helm/qontract-reconcile/templates/template.yaml
lisa: @TGPSKI I had a look again while I was going to change this (with `sed`) and we use `Always` for pull policies in many places. I'm sort of inclined to leave this as `Always`. Do you feel strongly to change this one to `IfNotPresent`?

TGPSKI: It's just about saving data transfer costs. To me, no value in always pulling if we have a cached image available locally. 

What is the value in always pulling?

reconcile/test/fixtures/helm/aws_account_shard_disabled.yml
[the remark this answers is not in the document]
TGPSKI: This is just about your speed and efficiency. If I was doing this, I would already have ChatGPT write me a script to do this automatically so that I don't waste time on it.

helm/qontract-reconcile/templates/template.yaml
lisa: The value is in maintianing consistency across all instances in this config. They should either all be `IfNotPresent` or none. I have no opinion on which is chosen.

helm/qontract-reconcile/templates/template.yaml
lisa: @hemslo That would make a great Jira card to follow up and reduce complexity as it is outside the scope of this card. This card's scope is already quite expansive and the final shape of the config will not have fluentd. If the spike isn't successful (ie, no suitable alternative for fluentd can be found), this will all be rolled back.

hemslo: https://issues.redhat.com/browse/APPSRE-10690

helm/qontract-reconcile/templates/template.yaml
cubismod: LGTM

helm/qontract-reconcile/templates/template.yaml
cubismod: Appreciate you doing the error filtering. We can see how buffering works out in FedRamp and then I'll let you know if we need to revisit logging to disk vs memory.

cubismod (approved):
LGTM!

helm/qontract-reconcile/templates/template.yaml
lisa: @cubismod Sounds like a plan. 👍

helm/qontract-reconcile/templates/template.yaml
```diff
@@ -449,6 +557,37 @@ objects:
...
+            valueFrom:
+              secretKeyRef:
+                name: ${CLOUDWATCH_SECRET}
+                key: aws_secret_access_key
+          resources:
+            requests:
+              memory: 30Mi
+              cpu: 15m
+            limits:
+              memory: 120Mi
```
hemslo, line 585: `resources` may need to be extracted as parameter

PR #4421, opened 2024-06-04
cli command to track Jenkins RHEL8 migration

rporres:
Related to APPSRE-7757

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -2712,6 +2712,76 @@ def systems_and_tools(ctx):
     print_output(ctx.obj["options"], inventory.data, inventory.columns)
 
 
+@get.command
+@click.pass_context
+def jenkins_jobs(ctx):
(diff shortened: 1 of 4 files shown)


tools/qontract_cli.py
```diff
@@ -2712,6 +2712,75 @@ def systems_and_tools(ctx):
...
+            if "project" not in c:
+                continue
+
+            project = c["project"]
+            root_node = project.get("node") or ""
+            if "jobs" not in project:
+                continue
+
+            for pj in project["jobs"]:
+                for job_name, job in pj.items():
```
hemslo, line 2740: `job_name` is unused

hemslo (approved):

tools/qontract_cli.py
rporres: fixed in 92235bff

patjlm (approved):

PR #4423, opened 2024-06-04
:bug: Add enableDeletions and deletionApprovals to vpc_requests query

esron:
This will enable us to delete the resources created by the integration by using the `deletionApprovals` object in app-interface.

Related to: [APPSRE-8743](https://issues.redhat.com/browse/APPSRE-8743)

### DIFF
--- a/reconcile/gql_definitions/common/aws_vpc_requests.py
+++ b/reconcile/gql_definitions/common/aws_vpc_requests.py
@@ -47,6 +47,12 @@ fragment VPCRequest on VPCRequest_v1 {
     terraformState {
       ...TerraformState
     }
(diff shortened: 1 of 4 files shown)


OliviaHY:
How would the deletionApprovals look like when user trying to delete an VPC? Is the `type` going to be `aws_vpc_request`?

esron:
> How would the deletionApprovals look like when user trying to delete an VPC? Is the `type` going to be `aws_vpc_request`?

No, I'm preparing an example MR, it will involve a extensive list in the `deletionApprovals` for the related account.

OliviaHY (approved):

PR #4424, opened 2024-06-04
[APPSRE-9928] Add validation code to ensure output_resource_name is unique in a namespace

bhushanthakur93:
To prevent error described in jira.

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -7,6 +7,7 @@ import random
 import re
 import string
 import tempfile
+from collections import Counter
 from collections.abc import (
     Iterable,
     Mapping,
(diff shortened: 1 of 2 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -1500,6 +1515,20 @@ def init_populate_specs(
...
+                    spec.namespace_name,
+                    spec.output_resource_name,
+                )
+                if (
+                    unique_constraint
+                    not in self.unique_constraint_for_external_resource
+                ):
+                    self.unique_constraint_for_external_resource.add(unique_constraint)
+                else:
+                    raise OutputResourceNameNotUniqueException(unique_constraint)
```
hemslo, line 1530: this is already inside one namespace, any reason to keep uniq check in instance level `self.unique_constraint_for_external_resource`? This will make it stateful and can't call `init_populate_specs` twice.

For simple local uniq validation, can directly get all duplicates inside one namespace via

```python
            name_counter = Counter(spec.output_resource_name for spec in specs)
            duplicates = [name for name, count in name_counter.items() if count > 1]
            if duplicates:
                raise
```

bhushanthakur93: > this is already inside one namespace

Do you mean the partition per namespace?

> This will make it stateful and can't call init_populate_specs twice.

Good point! This is something I had not considered.

hemslo: https://github.com/app-sre/qontract-reconcile/pull/4424/files#diff-4db367e141fd1992ae5a98872e6101dde4bf61a460c1bd0097176c6fbc849ac2R1508

here each namespace should be unique already, so cluster name and namespace name are unique, we just need to validate output resource name, local unique check per namespace is enough.

bhushanthakur93: You are right, I don't know what I was thinking :sweat_smile: 

Made the change you suggested.

hemslo (approved):
LGTM!

PR #4426, opened 2024-06-06
Add AWS tags to input data

lechuk47:
This PR adds the same resource tags logic as terraform-resources.

### DIFF
--- a/reconcile/external_resources/factories.py
+++ b/reconcile/external_resources/factories.py
@@ -9,6 +9,7 @@ from reconcile.external_resources.aws import (
     AWSRdsFactory,
     AWSResourceFactory,
 )
+from reconcile.external_resources.meta import QONTRACT_INTEGRATION
 from reconcile.external_resources.model import (
     ExternalResource,
     ExternalResourceKey,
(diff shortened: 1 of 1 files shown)


reconcile/external_resources/factories.py
```diff
@@ -28,6 +29,14 @@
 
 T = TypeVar("T")
 
+AWS_DEFAULT_TAGS = [
+    {
+        "tags": {
+            "app": "app-sre-infra",
```
hemslo, line 35: a similar one was added in https://github.com/app-sre/qontract-reconcile/pull/3836 for tracking cost, should we make this a config instead of hard coded `app-sre-infra`?

lechuk47: If we need this to be configurable we can add the setting afterwards if needed. 
.- I expect to have all resources tagged with the new modules. 
.- This is just to avoid changes in the migration process to make it smooth.

hemslo (approved):
LGTM!

reconcile/external_resources/factories.py
hemslo: since now it's the second time we use this default value, better to make it configurable later

PR #4427, opened 2024-06-06
template-renderer: allow using variables in dynamic variables

patjlm:
This allows this kind of variable settings in template collections:

```
variables:
  static:
    parentApp: app-sre
  dynamic:
  - name: parentApp
    query: |
      {
        apps: apps_v1
        (name: "{{ static.parentApp }}")
        {
          path
        }
      }
```

### DIFF
--- a/reconcile/templating/renderer.py
+++ b/reconcile/templating/renderer.py
@@ -171,7 +171,7 @@ def unpack_dynamic_variables(
         query = process_jinja2_template(
             body=dv.query, vars={"static": static, "dynamic": dynamic}
         )
-        dynamic[dv.name] = gql.query(query)
+        dynamic[dv.name] = gql.query(query) or {}
     return dynamic
 
 
--- a/reconcile/templating/renderer.py
+++ b/reconcile/templating/renderer.py
@@ -31,6 +31,7 @@ from reconcile.typed_queries.gitlab_instances import get_gitlab_instances
 from reconcile.utils import gql
 from reconcile.utils.git import clone
 from reconcile.utils.gql import init_from_config
+from reconcile.utils.jinja2.utils import process_jinja2_template
 from reconcile.utils.ruamel import create_ruamel_instance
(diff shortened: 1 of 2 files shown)


maorfr (approved):

chassing (approved):
❤️

reconcile/templating/renderer.py
```diff
@@ -166,12 +165,14 @@ def unpack_static_variables(
...
-    return {
-        dv.name: gql.query(dv.query) or {} for dv in collection_variables.dynamic or []
-    }
+    static = collection_variables.static or {}
+    dynamic: dict[str, dict[str, Any]] = {}
+    for dv in collection_variables.dynamic or []:
+        query = process_jinja2_template(
+            body=dv.query, vars={"static": static, "dynamic": dynamic}
+        )
+        dynamic[dv.name] = gql.query(query)
```
hemslo, line 174: pass `dynamic` into `vars` may have some unexpected result depending on processing order, and there was a default value `{}` if `query` returns no response, not sure if used anywhere, better to add test cases for this change at https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/templating/test_renderer.py

patjlm: I expect the order to be kept (is that correct?), and was doing that to be able to reference data from one dynamic variable to an other.

I'll add the missed default and check tests

hemslo: and the default `{}` was used in final result `gql.query(dv.query) or {}` instead of `dv.query`

patjlm: done in https://github.com/app-sre/qontract-reconcile/pull/4427/commits/4492a9a266cb093880b9ef6b8d259859aa4f4eb9 and https://github.com/app-sre/qontract-reconcile/pull/4427/commits/d407418349495e207ee6ed82bd6eb478b5ddba12

hemslo (approved):
LGTM!

PR #4433, opened 2024-06-07
:bug: Don't create outputs for deleted requests

esron:
If we create the outputs this can lead to a vpc data file being created with an empty `vpc_id`.


mehfuz (approved):
lgtm

PR #4434, opened 2024-06-09
saasherder validate promotion of hotfix versions

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10364

depends on https://github.com/app-sre/qontract-schemas/pull/669

with this PR, commit sha which will be defined under `hotfixVersions` for a `codeComponent` will be able to be promoted even being published with success to any subscribed channels (automated promotion).

### DIFF
--- a/reconcile/utils/saasherder/interfaces.py
+++ b/reconcile/utils/saasherder/interfaces.py
@@ -361,7 +361,7 @@ class SaasServiceOwner(Protocol):
 
 class AppCodeComponent(Protocol):
     url: str
-    hotfix_versions: list[str]
+    hotfix_versions: Optional[list[str]]
 
 
 SaasPipelinesProviders = Union[SaasPipelinesProviderTekton, SaasPipelinesProvider]
--- a/reconcile/utils/saasherder/interfaces.py
+++ b/reconcile/utils/saasherder/interfaces.py
@@ -63,6 +63,9 @@ class SaasApp(Protocol):
     @property
     def service_owners(self) -> Optional[Sequence[SaasServiceOwner]]: ...
 
+    @property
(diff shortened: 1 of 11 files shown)


reconcile/utils/saasherder/interfaces.py
```diff
@@ -356,6 +359,11 @@ class SaasServiceOwner(Protocol):
     email: str
 
 
+class AppCodeComponent(Protocol):
+    url: str
+    hotfix_versions: list[str]
```
chassing, line 364:
```suggestion
    hotfix_versions: Optional[list[str]]
```

maorfr: AppCodeComponent(Protocol):

reconcile/utils/saasherder/saasherder.py
```diff
@@ -1889,6 +1899,9 @@ def _validate_promotion(self, promotion: Promotion) -> bool:
         if not promotion.subscribe:
             return True
 
+        if promotion.commit_sha in self.hotfix_versions.get(promotion.url, set()):
```
fishi0x01, line 1902: nit: can we add a small comment or dedicated func to highlight that this is a general gate switch, i.e., it must always run before any other gate is evaluated? Im afraid I might accidentally miss that in the future.

fishi0x01 (approved):
LGTM :rocket:

reconcile/utils/saasherder/saasherder.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/4434/commits/6f6aef7c3981af1d3c8b8d1ed335500d06ae8f42

PR #4435, opened 2024-06-10
saasherder validate promotion of blocked versions

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10279

depends on #670

with this PR, commit sha which will be defined under `blockedVersions` for a `codeComponent` will not be able to be promoted.


patjlm (approved):

fishi0x01:
Note, that SAPM will still open auto-promotion merge requests. Will add this logic to SAPM asap to avoid opening.

PR #4436, opened 2024-06-10
:sparkles: glitchtip-project-alerts: support Jira components and issue type

chassing:
Add support for Jira components and custom issue type.

Ticket: [APPSRE-10356](https://issues.redhat.com/browse/APPSRE-10356)

Depends on:
* https://github.com/app-sre/qontract-schemas/pull/671
* https://github.com/app-sre/glitchtip-jira-bridge/pull/16

### DIFF
--- a/reconcile/glitchtip_project_alerts/integration.py
+++ b/reconcile/glitchtip_project_alerts/integration.py
@@ -124,13 +124,14 @@ class GlitchtipProjectAlertsIntegration(
                 )
             if glitchtip_project.jira and gjb_alert_url:
                 params: dict[str, str | list[str]] = {}
-                if gjb_token:
-                    params["token"] = gjb_token
+                token_params = {"token": gjb_token} if gjb_token else {}
                 alert_labels = glitchtip_project.jira.labels or []
 
                 if glitchtip_project.jira.project:
-                    params["components"] = glitchtip_project.jira.components or []
-                    params["labels"] = alert_labels
+                    params = {
+                        "labels": alert_labels,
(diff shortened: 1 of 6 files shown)


reconcile/glitchtip_project_alerts/integration.py
```diff
@@ -160,6 +146,44 @@ def fetch_desired_state(
...
+                            QONTRACT_INTEGRATION, board
+                        ) or not integration_is_enabled(
+                            jira_permissions_validator.QONTRACT_INTEGRATION, board
+                        ):
+                            continue
+                        params["labels"] = alert_labels + (channels.jira_labels or [])
+                        if channels.jira_component:
+                            params["components"] = [channels.jira_component]
+                        if board.issue_type:
+                            params["issue_type"] = board.issue_type
```
hemslo, line 167: `params` init outside of `for` loop, if there are multiple boards, later board will use previous board `params` when `if` conditions not met.

hemslo: Now the default value `"Bug"` is always needed for `issue_type`, is it used anywhere else in this repo? If you still need it to be optional in final params, and since the `token` is the only additional field, we can make it shareable like:


```python
                token_params = {"token": gjb_token} if gjb_token else {}
...
                        params = {
                            "labels": alert_labels + (channels.jira_labels or []),
                            "components": [channels.jira_component] if channels.jira_component else [],
                        } | token_params
                        if board.issue_type:
                            params["issue_type"] = board.issue_type
```

hemslo (approved):
LGTM!

PR #4437, opened 2024-06-10
oc apply handle request entity too large

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10378

Because of several escaping and the last-applied-configuration holds an older version of the enormous grafana JSON.

There is a potential solution that could work-removing the existing last-applied-configuration annotation before oc apply.


chassing (approved):
Nice. I wasn't aware of `oc.remove_last_applied_configuration`

PR #4438, opened 2024-06-10
:bug: Raises exception when plan errors are detected

esron:
This prevent users approving a failed run of `terraform plan`.

### DIFF
--- a/reconcile/terraform_vpc_resources/integration.py
+++ b/reconcile/terraform_vpc_resources/integration.py
@@ -168,10 +168,7 @@ class TerraformVpcResources(QontractReconcileIntegration[TerraformVpcResourcesPa
             thread_pool_size=thread_pool_size,
         )
 
-        disable_deletions, errors = tf_client.plan(enable_deletion=enable_deletion)
-
-        if disable_deletions or errors:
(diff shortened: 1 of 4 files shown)


reconcile/terraform_vpc_resources/integration.py
```diff
@@ -164,7 +168,10 @@ def run(self, dry_run: bool) -> None:
             thread_pool_size=thread_pool_size,
         )
 
-        tf_client.plan(enable_deletion=enable_deletion)
+        disable_deletions, errors = tf_client.plan(enable_deletion=enable_deletion)
+
+        if disable_deletions or errors:
+            raise PlanStepError("Errors in terraform plan step, please verify output.")
```
hemslo, line 174: similar error handling in terraform resources https://github.com/app-sre/qontract-reconcile/blob/836f3bbb0a07648c2049d198a52dc4afaea0ada3/reconcile/terraform_resources.py#L479-L483 and several other places, maybe worth extracting and use the same flow

esron: I've moved inside `terraform_client`.

OliviaHY (approved):

PR #4439, opened 2024-06-11
APPSRE-10279 blockversions for SAPM

fishi0x01:
- SAPM must not open MRs for blocked versions.
- more logging for visibility which gate blocked a promotion

### DIFF
--- a/reconcile/saas_auto_promotions_manager/subscriber.py
+++ b/reconcile/saas_auto_promotions_manager/subscriber.py
@@ -44,6 +44,7 @@ class Subscriber:
         use_target_config_hash: bool,
         uid: str,
         soak_days: int,
+        blocked_versions: set[str],
     ):
         self.saas_name = saas_name
         self.template_name = template_name
@@ -58,6 +59,7 @@ class Subscriber:
         self.soak_days = soak_days
         self._content_hash = ""
         self._use_target_config_hash = use_target_config_hash
+        self._blocked_versions = blocked_versions
 
     def has_diff(self) -> bool:
         current_hashes = {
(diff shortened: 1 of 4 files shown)


reconcile/saas_auto_promotions_manager/subscriber.py
```diff
@@ -142,22 +144,42 @@ def _compute_desired_ref(self) -> None:
...
-        ):
-            # We keep current state
-            self.desired_ref = self.ref
-        else:
-            # We have a common single publisher ref w/o any deployment issues
-            self.desired_ref = next(iter(publisher_refs))
+            return
+
+        if not self._passed_accumulated_soak_days():
+            logging.info(
```
patjlm, line 159: debug

fishi0x01: Good point. Added debug level for soak days

reconcile/saas_auto_promotions_manager/subscriber.py
```diff
@@ -142,22 +144,42 @@ def _compute_desired_ref(self) -> None:
...
+            return
+
+        if any_bad_deployment:
+            logging.info(
+                "Subscriber at path %s promotion stopped because of bad publisher deployment",
+                self.target_file_path,
+            )
+            return
+
+        desired_ref = next(iter(publisher_refs))
```
hemslo, line 172: in old code, we have `len(publihser_refs) != 1` check, we checked `> 1` case above, can there be any case `publisher_refs` is empty?

fishi0x01: Good catch! Added the check :+1:

hemslo (approved):
LGTM!

PR #4440, opened 2024-06-11
Add a external-resources get-input cli command to get er spec input data

lechuk47:
This command is to facilitate resources migration. It prints the resource input data that can be used directly by a module to import existing assets to the new terraform state.

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -56,6 +56,14 @@ from reconcile.cli import (
     config_file,
     use_jump_host,
 )
+from reconcile.external_resources.manager import (
+    FLAG_RESOURCE_MANAGED_BY_ERV2,
+    setup_factories,
+)
+from reconcile.external_resources.model import (
+    ExternalResourcesInventory,
+    load_module_inventory,
+)
 from reconcile.gql_definitions.advanced_upgrade_service.aus_clusters import (
     query as aus_clusters_query,
 )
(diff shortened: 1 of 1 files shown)


tools/qontract_cli.py
```diff
@@ -3770,5 +3783,40 @@ def remove(ctx, sso_client_vault_secret_path: str):
...
+    )
+    vault_settings = get_app_interface_vault_settings()
+    secret_reader = create_secret_reader(use_vault=vault_settings.vault)
+    er_settings = get_settings()[0]
+    m_inventory = load_module_inventory(get_modules())
+    factories = setup_factories(er_settings, m_inventory, er_inventory, secret_reader)
+    f = factories.get_factory(spec.provision_provider)
+    resource = f.create_external_resource(spec)
+    f.validate_external_resource(resource)
+    print(json.dumps(resource.dict(exclude={"data": {FLAG_RESOURCE_MANAGED_BY_ERV2}})))
```
hemslo, line 3818: any requirement to use `json.dumps` instead of directly `resource.json()`? And how to verify this command? [unit tests](https://github.com/app-sre/qontract-reconcile/blob/836f3bbb0a07648c2049d198a52dc4afaea0ada3/tools/test/test_qontract_cli.py) or some examples in doc?

lechuk47: Changed to resource.json() and added an example in the docs.

hemslo: I have this error in output

```
reconcile.utils.exceptions.FetchResourceError: error fetching resource: Resource spec not found: Provider rds, Id: dashdotdb-stage
```

Any pre requirement to use this command?

lechuk47: This resource is not yet managed by ErV2. It's just an example to show how the command works. If you want a working example: 

`qontract-cli --config=config.debug.toml external-resources get-input aws app-sre-stage rds erv2-tests` 

But this resource will be removed so it's not worth using it as an example.

This command will be used in the SOP to migrate resources along with all the required steps. It will make more sense once the full SOP is written.

hemslo: Is this for human usage or machine usage? Can use `json(indent=2` to pretty print result

hemslo (approved):
LGTM!

PR #4441, opened 2024-06-11
APPSRE-10351 fetch saas state via cli

fishi0x01:
qontract-cli command to help debugging saas deploy state. The command lists all publisher states from S3 for a given channel/sha

```
qontract-cli --config config.cli.toml get-promotion-state --channel my-channel --sha my-sha

Last state for publisher with ID 398f48dcaa4bbcac0a95f7ad0d94211d7c212bd0:
success=True target_config_hash='ada2226de1208856' saas_file='saas-file-name' check_in='2024-06-10 08:34:58.595463+00:00'
```

### DIFF
--- a/tools/saas_promotion_state/saas_promotion_state.py
+++ b/tools/saas_promotion_state/saas_promotion_state.py
@@ -9,7 +9,7 @@ from reconcile.typed_queries.app_interface_vault_settings import (
     get_app_interface_vault_settings,
 )
 from reconcile.typed_queries.saas_files import SaasFile, get_saas_files
-from reconcile.utils.promotion_state import PromotionState
+from reconcile.utils.promotion_state import PromotionData, PromotionState
 from reconcile.utils.secret_reader import create_secret_reader
 from reconcile.utils.state import init_state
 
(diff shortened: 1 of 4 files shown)


tools/saas_promotion_state/saas_promotion_state.py
```diff
@@ -0,0 +1,81 @@
...
+                    if not target.promotion:
+                        continue
+                    for publish_channel in target.promotion.publish or []:
+                        if publish_channel == channel:
+                            publisher_uids.append(
+                                target.uid(
+                                    parent_saas_file_name=saas_file.name,
+                                    parent_resource_template_name=resource_template.name,
+                                )
+                            )
```
hemslo, line 40: can there be duplicate values in `publisher_uids`?

tools/test/test_saas_promotion_state.py
```diff
@@ -0,0 +1,78 @@
...
+            ],
+        },
+    ])
+
+    promotion_state = create_autospec(spec=PromotionState)
+    promotion_state.get_promotion_data.return_value = ["data"]
+    saas_promotion_state = SaasPromotionState.create(
+        promotion_state=promotion_state, saas_files=saas_files
+    )
+    saas_promotion_state.get(channel="channel-a", sha="main")
```
hemslo, line 71: Do you want to verify expected output? May need to let `get` just return string and do actual print in cli method

fishi0x01: Good idea. Fixed :+1:

tools/saas_promotion_state/saas_promotion_state.py
fishi0x01: There shouldn't be duplicates:
- Every saas target has a unique id (theoretical hash collisions possible, but Id neglect this possibility).
- We focus on fetching publishers for a single channel. I.e., the same publisher cannot be in that channel multiple times.

hemslo (approved):
LGTM!

PR #4442, opened 2024-06-11
:art: Upgrade ruff and enable pyupgrade checks

chassing:
Upgrade to the latest `ruff` version and enable `pyupgrade` checks. `pyupgrade` takes care of obsolete things, like deprecated imports or new Python features like `foobar : str | None ...` instead of `foobar: Optional[str] ...`

Please review commit by commit!


release/version.py
```diff
@@ -47,7 +48,7 @@ def commit(length: int = 7) -> str:
     return p.stdout.decode("utf-8").strip()
 
 
-def semver(git_version: str | None = None) -> str:
+def semver(git_version: Optional[str] = None) -> str:  # noqa: UP007 - RHEL8 has python 3.8
```
hemslo, line 51: we should create a card to make release run in container so no dependencies on ci host

hemslo (approved):
Nice upgrade and cleanup!

release/version.py
chassing: Done https://issues.redhat.com/browse/APPSRE-10403

patjlm:
❤️

PR #4443, opened 2024-06-11
enable render specific template collection

BumbleFeng:
make it easier to check output of new template collection.

### DIFF
--- a/reconcile/templating/renderer.py
+++ b/reconcile/templating/renderer.py
@@ -46,11 +46,14 @@ APP_INTERFACE_PATH_SEPERATOR = "/"
 
 
 def get_template_collections(
-    query_func: Callable | None = None,
+    query_func: Callable | None = None, template_collection_name: str | None = None
 ) -> list[TemplateCollectionV1]:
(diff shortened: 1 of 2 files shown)


patjlm (approved):

reconcile/templating/renderer.py
```diff
@@ -287,6 +288,11 @@ def reconcile(
         gql_no_validation = init_from_config(validate_schemas=False)
 
         for c in get_template_collections():
+            if (
+                self.params.template_collection_name
+                and c.name != self.params.template_collection_name
+            ):
+                continue
```
hemslo, line 295: Can we put this logic in gql? To use [$filter](https://github.com/app-sre/qontract-server?tab=readme-ov-file#gql-query-filtering) in gql query, then we can expand filter fields later.

```gql
query TemplateCollection_v1($filter: JSON) {
  template_collection_v1(filter: $filter) {
```

Then pass `variables` to `query` function

```python
    variables = {"filter": {}}
    if template_collection_name:
        variables["filter"]["name"] = template_collection_name
    query(query_func, variables=variables)
```

PR #4429, opened 2024-06-06
template-renderer: foreach loop

patjlm:
https://issues.redhat.com/browse/APPSRE-10379

depends on https://github.com/app-sre/qontract-schemas/pull/667

Example in a template-collection:

```yaml
forEach:
  items:
  - cluster_name: app-sre-prod-01
    value: value1
  - cluster_name: app-sre-prod-04
    value: value2

variables:
  static:
    foo: FOO
    bar: "This is for value {{ foreach_item.value }}"
  dynamic:
  - name: cluster
    query: |
      {
      clusters: clusters_v1(filter: {
          name: "{{ foreach_item.cluster_name }}"
        })
        {
          name
          consoleUrl
        }
      }
```

PR #4444, opened 2024-06-11
fix template foreach loop

BumbleFeng:
fix https://github.com/app-sre/qontract-reconcile/pull/4429

according to https://github.com/app-sre/qontract-reconcile/blob/e58b957b523d0dd46f0197eb41dd5fcd73569439/reconcile/templating/lib/merge_request_manager.py#L237-L250 for a single template collection, it can only open one MR at the same time. So all outputs generated by one collection should be written together to be included in one MR. (Failed example:   https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/108782)

Meanwhile, template_hash should include all unpack variables to detect collection changes.

### DIFF
--- a/reconcile/templating/lib/merge_request_manager.py
+++ b/reconcile/templating/lib/merge_request_manager.py
@@ -136,7 +136,7 @@ class MergeRequestManager(MergeRequestManagerBase[TemplateInfo]):
 
         output = data.data
         collections = {o.input.collection for o in output}
-        collection_hashes = {o.input.collection_hash for o in output}
+        collection_hashes = {o.input.calc_template_hash() for o in output}
         additional_labels = {label for o in output for label in o.input.labels}
         # From the way the code is written, we can assert that there is only one collection and one template hash
         assert len(collections) == 1
--- a/reconcile/templating/lib/model.py
+++ b/reconcile/templating/lib/model.py
@@ -1,11 +1,29 @@
+from typing import Any
+
+from deepdiff import DeepHash
 from pydantic import BaseModel
+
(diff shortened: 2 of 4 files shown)


reconcile/templating/lib/merge_request_manager.py
```diff
@@ -136,7 +136,7 @@ def create_merge_request(self, data: MrData) -> None:
 
         output = data.data
         collections = {o.input.collection for o in output}
-        collection_hashes = {o.input.collection_hash for o in output}
+        collection_hashes = {o.input.calc_template_hash() for o in output}
```
hemslo, line 139: the hash is only used to determine keep or close mr, right? Then why not directly use all generated content and path as hash source?

BumbleFeng: I don't have a preference, but I think @janboll might be better at answering this question.

hemslo: I think it's better to use output, there are cases that input changed but output stay same, like changed a variable but conditional render result still stay unchanged, then no need to open new mr. To make it easier to compute hash, we can restructure `TemplateOutput` to be like

```python
class TemplateResult(BaseModel):
    input
    output
```

pick `output` field to calculate hash.

BumbleFeng: The refactoring is bigger than I expected. I need to separate it in another PR.

hemslo: Then let's merge this first

hemslo (approved):

PR #4446, opened 2024-06-12
APPSRE-10229 dynatrace client

fishi0x01:
A client wrapper around the Dynatrace API.

The functions are fairly simple, but test layer is easily extensible for future functions which might be more complex.

### DIFF
--- a/reconcile/utils/dynatrace/client.py
+++ b/reconcile/utils/dynatrace/client.py
@@ -1,0 +1,32 @@
+from __future__ import annotations
+
+from collections.abc import Iterable
+
+from dynatrace import Dynatrace
+
+
+class DynatraceTokenCreationError(Exception):
+    pass
+
+
+class DynatraceClient:
+    def __init__(self, environment_url: str, api: Dynatrace) -> None:
+        self._environment_url = environment_url
+        self._api = api
+
+    def create_api_token(self, name: str, scopes: Iterable[str]) -> str:
+        try:
+            token = self._api.tokens.create(name=name, scopes=scopes)
+        except Exception as e:
+            raise DynatraceTokenCreationError(
+                f"{self._environment_url=} Failed to create token for {name=}", e
+            ) from e
+        return token.token
+
+    @staticmethod
+    def create(
+        environment_url: str, token: str | None, api: Dynatrace | None
+    ) -> DynatraceClient:
+        if not api:
+            api = Dynatrace(base_url=environment_url, token=token)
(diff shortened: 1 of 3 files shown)


reconcile/utils/dynatrace/client.py
```diff
@@ -0,0 +1,32 @@
...
+            token = self._api.tokens.create(name=name, scopes=scopes)
+        except Exception as e:
+            raise DynatraceTokenCreationError(
+                f"{self._environment_url=} Failed to create token for {name=}", e
+            ) from e
+        return token.token
+
+    @staticmethod
+    def create(
+        environment_url: str, token: str | None, api: Dynatrace | None
```
mehfuz, line 28: if we are certain about token...why is it optional here?

fishi0x01: This is a dependency injection pattern (mainly used for inserting mocks in tests)

fishi0x01: Essentially, I personally dislike pythons monkey patching. I prefer building objects where I can insert deps. For tests, we do not have any token - but a fully mocked API object that we pass in.

fishi0x01: > The reason i am asking is because __init__() of Dynatrace does not have token param as opitional.
`create` is a static factory method. It creates the API from token if token is given - but it also accepts a full API object.
Its a common pattern.

hemslo: @fishi0x01 this is an anti pattern, conditional path for test usage only. in test `api` always pass in, then `not api` will always be `False`, if I change `api = Dynatrace(base_url=environment_url, token=token)` to something else, all tests still pass, and the real run will fail. No verification of `Dynatrace` should have been called with `base_url=environment_url, token=token`.

And the signature of `create` becomes ambiguity, what if I call `create("xxx", None, None)`, it's valid for the wrapper method, but type mismatch for `Dynatrace` as `token` should be `str`.

The correct split is to make `__init__` accept injectable objects (already right in code), and `create` only accept needed input to construct them.
You can make it a `classmethod` to make it explicit as a builder.

```python
    @classmethod
    def create(cls, environment_url: str, token: str) -> Self:
        api = Dynatrace(base_url=environment_url, token=token)
        return cls(environment_url=environment_url, api=api)
```

In tests, `create` method has to be tested with patch as `Dynatrace` is an external dependency, but all other methods can be tested using mocked `Dynatrace` in `__init__`.

PR #4450, opened 2024-06-13
APPSRE-10351 print state url

fishi0x01:
Print quick access link to the saas deploy state file in S3 for further debugging.

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -3681,6 +3681,7 @@ def get_promotion_state(channel: str, sha: str):
     )
 
     bucket = os.environ.get("APP_INTERFACE_STATE_BUCKET")
+    region = os.environ.get("APP_INTERFACE_STATE_BUCKET_REGION", "us-east-1")
     promotion_state = SaasPromotionState.create(promotion_state=None, saas_files=None)
     for publisher_id, state in promotion_state.get(channel=channel, sha=sha).items():
         print()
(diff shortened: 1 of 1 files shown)


tools/qontract_cli.py
```diff
@@ -3680,14 +3680,18 @@ def get_promotion_state(channel: str, sha: str):
...
     for publisher_id, state in promotion_state.get(channel=channel, sha=sha).items():
         print()
         if not state:
             print(f"No state found for {publisher_id=}")
         else:
-            print(f"State for {publisher_id=}:")
-            print(state)
+            print(f"{publisher_id=}")
+            print(
+                f"State link: https://us-east-1.console.aws.amazon.com/s3/object/{bucket}?region=us-east-1&bucketType=general&prefix=state/openshift-saas-deploy/promotions_v2/{channel}/{publisher_id}/{sha}"
```
hemslo, line 3692: Can we get region from env var? `us-east-1` can be default if not provided

hemslo (approved):
LGTM!

tools/qontract_cli.py
fishi0x01: Good idea. Added from env var

hemslo: @fishi0x01 missed one replace in string `?region=us-east-1`

fishi0x01: Good catch - adding here https://github.com/app-sre/qontract-reconcile/pull/4455

PR #4456, opened 2024-06-14
:sparkles: aws-account-manager: support takeover of existing accounts

chassing:
If `account-request.uid` is specified, `aws-account-manager` will take over an existing account instead of creating a new one.

Ticket: [APPSRE-10430](https://issues.redhat.com/browse/APPSRE-10430)
Depends on: https://github.com/app-sre/qontract-schemas/pull/672

### DIFF
--- a/reconcile/aws_account_manager/integration.py
+++ b/reconcile/aws_account_manager/integration.py
@@ -159,7 +159,8 @@ class AwsAccountMgmtIntegration(
     ) -> None:
         """Create new AWS accounts."""
         for account_request in account_requests:
-            if not (
+            uid = account_request.uid
+            if not uid and not (
                 uid := reconciler.create_organization_account(
                     aws_api=aws_api,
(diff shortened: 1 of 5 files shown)


cubismod (approved):
LGTM

reconcile/aws_account_manager/integration.py
```diff
@@ -159,7 +159,8 @@ def create_accounts(
     ) -> None:
         """Create new AWS accounts."""
         for account_request in account_requests:
-            if not (
+            uid = account_request.uid
+            if not uid and not (
```
hemslo, line 163: to simplify check condition, we can

```python
uid = account_request.uid or reconciler.create_organization_account(...)
if not uid:
    continue
```

chassing: For sure. Applied in last commit

PR #4459, opened 2024-06-16
[gitlab-housekeeping] publish personal access token expiration metrics

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10433

this PR attempts to add a metric for remaining days of personal access tokens until expiration. we can use these metrics for alerting.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -5,6 +5,7 @@ from collections.abc import (
 )
 from dataclasses import dataclass
 from datetime import (
+    UTC,
     datetime,
     timedelta,
 )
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -26,7 +26,7 @@
         "hvac>=0.7.0,<0.8.0",
         "ldap3>=2.9.1,<2.10.0",
         "anymarkup>=0.7.0,<0.9.0",
-        "python-gitlab>=1.11.0,<1.12.0",
+        "python-gitlab~=4.6",
```
hemslo, line 29: we may need to fix some breaking changes for this 5 years version upgrade https://python-gitlab.readthedocs.io/en/stable/changelog.html

maorfr: #4467

PR #4471, opened 2024-06-18
:sparkles: template-renderer: make the jinja2 whitespace control behavior configurable

chassing:
Make the jinja [whitespace control](https://jinja.palletsprojects.com/en/3.1.x/templates/#whitespace-control) behavior configurable for a template by introducing `trim_blocks`, `lstrip_blocks`, and `keep_trailing_newline` as `templateRenderOptions`. The default will still be `False` for all three options.

Depends on: https://github.com/app-sre/qontract-schemas/pull/675

### DIFF
--- a/reconcile/templating/lib/rendering.py
+++ b/reconcile/templating/lib/rendering.py
@@ -5,7 +5,11 @@ from typing import Any, Protocol
 
 from pydantic import BaseModel
 
-from reconcile.utils.jinja2.utils import Jinja2TemplateError, process_jinja2_template
+from reconcile.utils.jinja2.utils import (
+    Jinja2TemplateError,
+    TemplateRenderOptions,
+    process_jinja2_template,
+)
 from reconcile.utils.jsonpath import parse_jsonpath
 from reconcile.utils.ruamel import create_ruamel_instance
 from reconcile.utils.secret_reader import SecretReaderBase
@@ -43,17 +47,13 @@ class Renderer(ABC):
         template: Template,
         data: TemplateData,
         secret_reader: SecretReaderBase | None = None,
-        trim_blocks: bool = False,
-        lstrip_blocks: bool = False,
-        keep_trailing_newline: bool = False,
+        template_render_options: TemplateRenderOptions | None = None,
     ):
         self.template = template
         self.data = data
         self.secret_reader = secret_reader
         self.ruamel_instance = create_ruamel_instance(explicit_start=True)
-        self.trim_blocks = trim_blocks
-        self.lstrip_blocks = lstrip_blocks
-        self.keep_trailing_newline = keep_trailing_newline
+        self.template_render_options = template_render_options
 
     def _jinja2_render_kwargs(self) -> dict[str, Any]:
         return {**self.data.variables, "current": self.data.current}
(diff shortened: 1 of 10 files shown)


reconcile/templating/lib/rendering.py
```diff
@@ -43,11 +43,17 @@ def __init__(
         template: Template,
         data: TemplateData,
         secret_reader: SecretReaderBase | None = None,
+        trim_blocks: bool = False,
+        lstrip_blocks: bool = False,
+        keep_trailing_newline: bool = False,
```
hemslo, line 48: since these 3 params always bundle together, should we extract a class like `TemplateRenderOptionsV1` but not optional fields?

reconcile/utils/jinja2/utils.py
```diff
@@ -32,8 +32,18 @@ def __init__(self, msg: Any):
...
+    extra_curly: bool = False,
+    trim_blocks: bool = False,
+    lstrip_blocks: bool = False,
+    keep_trailing_newline: bool = False,
+) -> Any:
+    env: dict[str, Any] = {
+        "trim_blocks": trim_blocks,
+        "lstrip_blocks": lstrip_blocks,
+        "keep_trailing_newline": keep_trailing_newline,
+    }
```
hemslo, line 46: env will be reset if `extra_curly` is `True`, these params not needed for extra curly?

chassing: Fixed in [21cbd8c](https://github.com/app-sre/qontract-reconcile/pull/4471/commits/21cbd8c42d919b392e62e952df472148abd838da)

reconcile/templating/lib/rendering.py
chassing: good idea. Applied in [fc5c62f](https://github.com/app-sre/qontract-reconcile/pull/4471/commits/fc5c62fed3dcf42244673e740cad381e6324d395)

hemslo: `TemplateRenderOptions` is frozen, how about create a builder method and reuse it like

```python
    @classmethod
    def create(cls, options: TemplateRenderOptionsV1):
        return cls(
            trim_blocks=options.trim_blocks or False,
            lstrip_blocks=options.lstrip_blocks or False,
            keep_trailing_newline=options.keep_trailing_newline or False,
        )
```

and simplify usage of `env` using `template_render_options.dict()`?

chassing: Also a good idea. Take a look at the last 2 commits 🙇

hemslo: now I really miss object mapper to auto handle nullable fields convert in other langs like

```
template.template_render_options?.to_model() ?? TemplateRenderOptions()
```

instead of 

```python
TemplateRenderOptions.create(
                trim_blocks=template.template_render_options.trim_blocks
                if template.template_render_options
                else None,
                lstrip_blocks=template.template_render_options.lstrip_blocks
                if template.template_render_options
                else None,
                keep_trailing_newline=template.template_render_options.keep_trailing_newline
                if template.template_render_options
                else None,
            )
```

hemslo (approved):
LGTM!

PR #4487, opened 2024-06-21
:sparkles: template-rendering: support jsonpath patch.identifier

chassing:
Support `patch.identifier` as jsonpath, e.g.

```
patch:
  path: "$.openshiftResources"
  identifier: $.variables.name

template: |
  provider: resource-template
  type: jinja2
  path: /observability/servicemonitors/aws-resource-exporter.servicemonitor.yaml
  variables:
    namespace: app-sre-observability-production
    name: aws-resource-exporter-{{ static.name }}
```

to update/add an `openshiftResources` based on a variable.

Ticket: [APPSRE-10384](https://issues.redhat.com/browse/APPSRE-10384)


patjlm:
so here is would add an item in the `openshiftResources` list, if there is no item that matches the `variables.name` field, right?

reconcile/templating/lib/rendering.py
```diff
@@ -123,11 +123,31 @@ def render_output(self) -> str:
...
-                )
-            dta_identifier = data_to_add.get(self.template.patch.identifier)
+
+            def get_identifier(data: dict[str, Any] | str) -> Any:
+                assert self.template.patch is not None  # mypy
+                if not self.template.patch.identifier:
+                    raise ValueError(
+                        f"Expected identifier in patch for list at {self.template}"
+                    )
+                if isinstance(data, str):
```
patjlm, line 133: should you check `identifier == "."` ?

chassing: `patch.identifier` is not used for lists, but it's mandatory. I think `patch.identifier: .` is just convenient. Added the check in the last commit

patjlm:
I think your template and test current/expected do not match in the PR description, but the tests are looking good

patjlm (approved):

reconcile/templating/lib/rendering.py
```diff
@@ -123,11 +123,35 @@ def render_output(self) -> str:
...
+                    # jsonpath support
+                    if matches := [
+                        match.value
+                        for match in parse_jsonpath(
+                            self.template.patch.identifier
+                        ).find(data)
+                    ]:
+                        return matches[0]
+                    return None
+                return data.get(self.template.patch.identifier)
```
hemslo, line 152: can we align both to use json path? like if identifier is `$`, it can handle both list of string and normal json path.

chassing: I don't get it. For lists, do you want `identifier: $.` instead of `identifier: .`?

hemslo: just `identifier: $`, then let jsonpath support branch handle it.

chassing: But `identifier` isn't used for lists.

hemslo: it can be used, update test fixture https://github.com/app-sre/qontract-reconcile/pull/4487/files#diff-9c849344aa377c6ae1b6b04967c3d28e203c7334bb57bd65c526d7ff3ce0576cR10 to `identifier: $`, then remove `str` check https://github.com/app-sre/qontract-reconcile/pull/4487/files#diff-29d1a22f3899b8251e82525a326a71f6be4c31d519ef497d3ca233f557a55630R133-R140, update json path condition to `.startswith("$")`, test pass.

chassing: Awesome! 🙇 Applied in the last commit.

hemslo (approved):
LGTM! Please also update PR description to match new implementation.

PR #4488, opened 2024-06-21
[Template] use generated outputs as hash source

BumbleFeng:
restructure TemplateInput to be TemplateResult

### DIFF
--- a/reconcile/templating/lib/merge_request_manager.py
+++ b/reconcile/templating/lib/merge_request_manager.py
@@ -4,7 +4,7 @@ import string
 
 from pydantic import BaseModel
 
-from reconcile.templating.lib.model import TemplateOutput
+from reconcile.templating.lib.model import TemplateOutput, TemplateResult
 from reconcile.utils.gitlab_api import GitLabApi
 from reconcile.utils.merge_request_manager.merge_request_manager import (
     MergeRequestManagerBase,
@@ -24,14 +24,14 @@ TR_LABEL = "template-output"
 
 VERSION_REF = "tr_version"
 COLLECTION_REF = "collection"
-TEMPLATE_COLLECTION_HASH_REF = "collection_hash"
+TEMPLATE_RESULT_HASH_REF = "result_hash"
 
 COMPILED_REGEXES = {
     i: re.compile(rf".*{i}: (.*)$", re.MULTILINE)
     for i in [
         VERSION_REF,
         COLLECTION_REF,
-        TEMPLATE_COLLECTION_HASH_REF,
+        TEMPLATE_RESULT_HASH_REF,
     ]
 }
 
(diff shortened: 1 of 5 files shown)


reconcile/templating/renderer.py
```diff
@@ -269,9 +269,9 @@ def reconcile_template_collection(
         gql_api: GqlApi,
         persistence: FilePersistence,
         ruamel_instance: yaml.YAML,
-        input: TemplateInput,
+        result: TemplateResult,
         each: dict[str, Any],
-    ) -> list[TemplateOutput]:
+    ) -> None:
```
hemslo, line 274: the old signature is easier to understand, can we remove the `result` from params, and still return `list[TemplateOutput]`?

reconcile/templating/lib/model.py
```diff
@@ -1,34 +1,23 @@
...
-class TemplateOutput(BaseModel):
-    input: TemplateInput
-    is_new: bool = False
-    path: str
-    content: str
-    auto_approved: bool = False
+        if not self.result_hash:
+            hashable = {"outputs": sorted(self.outputs, key=lambda x: x.path)}
+            self.result_hash = DeepHash(hashable)[hashable]
+        return self.result_hash
```
hemslo, line 23: This method is to calc hash only once, right? Can use [cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) to simplify it, example to use with `BaseModel` at https://github.com/app-sre/qontract-reconcile/blob/1b80a586850a1145aff072f7837b71bb716d223d/reconcile/utils/early_exit_cache.py#L76.

To make hash safe and stable, should also mark `TemplateOutput` as frozen. If `path` is unique, can make `hashable` a dict `{o.path: o for o in self.outputs}`

BumbleFeng: I found out calc hash would only run once when submitting MR, so there is no need to add cache.

reconcile/templating/lib/merge_request_manager.py
```diff
@@ -121,34 +121,26 @@ def process(self, gitlab_cli: GitLabApi) -> None:
...
-
-
 class MergeRequestManager(MergeRequestManagerBase[TemplateInfo]):
     def __init__(self, vcs: VCS, parser: Parser):
         super().__init__(vcs, parser, TR_LABEL)
 
-    def create_merge_request(self, data: MrData) -> None:
+    def create_merge_request(
+        self, data: TemplateResult, auto_approved: bool = False
+    ) -> None:
```
hemslo, line 130: in `MergeRequestManagerBase`, `create_merge_request` is an abstract method, better to keep params (`data`) unchanged.

hemslo (approved):
LGTM!

PR #4490, opened 2024-06-22
[APPSRE-10455] add createOnly to template

BumbleFeng:
depends on https://github.com/app-sre/qontract-schemas/pull/678

Allow the template only to create the target file once.

### DIFF
--- a/reconcile/gql_definitions/introspection.json
+++ b/reconcile/gql_definitions/introspection.json
@@ -32659,6 +32659,18 @@
                             "deprecationReason": null
                         },
                         {
+                            "name": "createOnly",
+                            "description": null,
+                            "args": [],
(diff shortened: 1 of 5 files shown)


OliviaHY (approved):

OliviaHY:
@patjlm pointed out that we can use condition for this: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/templating/templates/aws-account/account.yml#L8-9

patjlm:
> @patjlm pointed out that we can use condition for this: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/templating/templates/aws-account/account.yml#L8-9

This also allows to update the target file multiple times before stopping completely on the global condition

PR #4492, opened 2024-06-23
[qontract-cli] get logs for integration

maorfr:
convenience function to get a command to get logs from an integration


tools/qontract_cli.py
```diff
@@ -2720,6 +2721,46 @@ def systems_and_tools(ctx):
     print_output(ctx.obj["options"], inventory.data, inventory.columns)
 
 
+@get.command()
```
rporres, line 2724: Can you add a `help` description about what this command does? It would be good to specify that it gets **production** logs from the cluster where it is running.

mehfuz (approved):
lgtm

PR #4495, opened 2024-06-24
APPSRE-10229 dt client list token ids

fishi0x01:
Dynatrace Client needs to fetch token ids for tokens with given prefix

### DIFF
--- a/reconcile/test/utils/dynatrace/conftest.py
+++ b/reconcile/test/utils/dynatrace/conftest.py
@@ -20,7 +20,21 @@ def dynatrace_api_builder() -> Callable[[Mapping], Dynatrace]:
             token.token = token_created_result
             token_service.create.return_value = token
         else:
+            # For raising exceptions
             token_service.create.side_effect = token_created_result
+
+        list_token_result = data.get("LIST_TOKEN_RESULT", [])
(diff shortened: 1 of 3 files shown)


reconcile/test/utils/dynatrace/conftest.py
```diff
@@ -20,7 +20,21 @@ def builder(data: Mapping[str, Any]) -> Dynatrace:
...
+            # For raising exceptions
             token_service.create.side_effect = token_created_result
+
+        list_token_result = data.get("LIST_TOKEN_RESULT", [])
+        if isinstance(list_token_result, list):
+            listed_tokens = []
+            for token_tup in list_token_result:
+                token = create_autospec(spec=ApiTokenCreated)
+                token.id = token_tup[1]
+                token.name = token_tup[0]
```
hemslo, line 32: is there a field called `name` in `ApiTokenCreated`? `list` should return `ApiToken`

hemslo (approved):
LGTM!

reconcile/test/utils/dynatrace/conftest.py
fishi0x01: Oh copy pasta. Fixed now. Nice catch! :+1:

PR #4497, opened 2024-06-24
APPSRE-10229 DTP dep inversion + tests

fishi0x01:
DTP will undergo some changes to support more cluster and token types. As a preparation:

- introducing tests for DTP
- adding dependency inversion for tests
- rely on more utils functions (e.g., `DynatraceClient`, shared gql queries)

**Dependencies**

- https://github.com/app-sre/qontract-reconcile/pull/4495
- https://github.com/app-sre/qontract-reconcile/pull/4502

**Tests**

Apart from the added unit tests, I also ran the integration in dry-run locally.


reconcile/dynatrace_token_provider/ocm.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+        return Cluster(
+            id=cluster.ocm_cluster.id,
+            external_id=cluster.ocm_cluster.external_id,
+            organization_id=cluster.organization_id,
+            dt_tenant=dt_tenant,
+        )
+
+
+class OCMClient:
```
OliviaHY, line 52: Should this be in utils/ instead so folks can use this pattern to make testing easier going forward?

OliviaHY (approved):
This is great! Love what you did for testing.

reconcile/dynatrace_token_provider/ocm.py
fishi0x01: Good question :thinking: 
For now I would like to keep it here, as it also contains DTP specific things (like the Cluster information dedicated for DTP).

PR #4498, opened 2024-06-24
[ocm-labels] enable disabling integration per ocm org

maorfr:
query for `disable.integrations` and consider it for ocm orgs for ocm-labels


patjlm (approved):

chassing (approved):
LGTM

PR #4500, opened 2024-06-25
[ocm-clusters] do not enforce one HCP cluster per AWS account

maorfr:
partially revert #4136 

supporting a single cluster per aws account is an AppSRE decision for managing our own clusters.
app-interface, as a tool, should be able to support multiple clusters in an aws account. open source ftw :tada: 

we can add a safety net in app-interface to enforce that every cluster in the AppSRE OCM org is in a unique aws account. this should not be a code limitation.


rporres:
I thought we had added that limitation due to the bugs on cluster deletion in a shared AWS account.

PR #4502, opened 2024-06-25
APPSRE-10229 return token id too

fishi0x01:
When a token is created, it has the secret and an id. We should return both to the caller.


hemslo (approved):
LGTM!

PR #4511, opened 2024-06-28
APPSRE-10229 catch exception

fishi0x01:
Catching error in label discovery call. We want to process all ocm envs, even if one is failing.

### DIFF
--- a/reconcile/dynatrace_token_provider/integration.py
+++ b/reconcile/dynatrace_token_provider/integration.py
@@ -66,11 +66,15 @@ class DynatraceTokenProviderIntegration(
         with metrics.transactional_metrics(self.name):
             unhandled_exceptions = []
             for ocm_env_name, ocm_client in dependencies.ocm_client_by_env_name.items():
-                clusters = ocm_client.discover_clusters_by_labels(
-                    label_filter=subscription_label_filter().like(
-                        "key", dtp_label_key("%")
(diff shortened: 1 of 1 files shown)


reconcile/dynatrace_token_provider/integration.py
```diff
@@ -66,11 +66,15 @@ def reconcile(self, dry_run: bool, dependencies: Dependencies) -> None:
...
-                )
+                clusters: list[Cluster] = []
+                try:
+                    clusters = ocm_client.discover_clusters_by_labels(
+                        label_filter=subscription_label_filter().like(
+                            "key", dtp_label_key("%")
+                        ),
+                    )
+                except Exception as e:
+                    unhandled_exceptions.append(f"{ocm_env_name}: {e}")
```
hemslo, line 77: If we need full stacktrace of excpetions, can do error logging and aggregate them together using [ExceptionGroup](https://docs.python.org/3/library/exceptions.html#ExceptionGroup) and raise at the end.

hemslo (approved):

reconcile/dynatrace_token_provider/integration.py
fishi0x01: Good idea! Wll for now keep the current mechanism, but will add that when I start making changes for DTPv2

PR #4513, opened 2024-06-28
allow to patch multiple object in patch path by a single template

BumbleFeng:


patjlm:
Can you provide more context and maybe an example of what this enables?

PR #4514, opened 2024-06-30
extract qontract-reconcile-toml secret name to parameter

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10491

this change makes it optional to define a different Secret to use, instead of updating the secret in place in the namespace.


hemslo (approved):
LGTM!

PR #4518, opened 2024-07-01
fix cluster admin usage in oc delete

maorfr:
should be set in options to be considered

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -808,14 +808,14 @@ def handle_deleted_resources(
             resource_type=resource_type,
             options=options,
         ):
-            privileged = data["use_admin_token"].get(name, False)
+            options.privileged = data["use_admin_token"].get(name, False)
             action = {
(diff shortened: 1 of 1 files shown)


reconcile/openshift_base.py
```diff
@@ -808,14 +808,14 @@ def handle_deleted_resources(
             resource_type=resource_type,
             options=options,
         ):
-            privileged = data["use_admin_token"].get(name, False)
+            options.privileged = data["use_admin_token"].get(name, False)
```
patjlm, line 811:
```suggestion
            options.privileged = data["use_admin_token"].get(name) or False
```

maorfr: i'm consistent with the rest of the file. for example: https://github.com/app-sre/qontract-reconcile/pull/4518/files#diff-a14858c214c6abf739ca222b3450b66851c93a7a407d27cb9dff88979dfc043fR768

patjlm (approved):

PR #4519, opened 2024-07-01
request helm 3.11.1 to be installed

patjlm:
helm used in integrations deployment but is also now supported in saas-deploy


mehfuz (approved):
lgtm...

PR #4521, opened 2024-07-02
Add delete flag to external-resources

lechuk47:
External resources deletion will work with the `delete: true` attribute at the resource level. 

```yaml 
externalResources:
- provider: aws
  provisioner:
    $ref: /aws/app-sre-stage/account.yml
  resources:
  - provider: rds
    managed_by_erv2: true
    identifier: erv2-tests
    defaults: /terraform/resources/erv2-tests/rds/pg14-defaults.yml
    parameter_group: /terraform/resources/erv2-tests/rds/pg14-default-pg.yml
    delete: true # <-- This resource will get deleted.
```

Why: 
* This is a well-established pattern to delete resources in app-interface (saas targets, namespaces, etc).
* TF-Resources `deletion_approvals` are meant to avoid non-wanted resource deletions in case resource specs are removed from app-interface. Requiring a `delete: true`  attribute accomplishes the same objective. 
* Tenants don't need to care about secondary terraform objects to delete, they just want to remove the resource.
* This change introduces the "feature" of abandoning resources. 

Schema update: https://github.com/app-sre/qontract-schemas/pull/683


esron (approved):

reconcile/external_resources/manager.py
```diff
@@ -179,21 +180,22 @@ def _get_desired_objects_reconciliations(self) -> set[Reconciliation]:
         return r
 
     def _get_deleted_objects_reconciliations(self) -> set[Reconciliation]:
-        desired_keys = set(self.er_inventory.keys())
-        state_resource_keys = self.state_mgr.get_all_resource_keys()
-        deleted_keys = state_resource_keys - desired_keys
-        r: set[Reconciliation] = set()
-        for key in deleted_keys:
+        to_reconcile: set[Reconciliation] = set()
+        for key in [k for k, v in self.er_inventory.items() if v.marked_to_delete]:
```
hemslo, line 184: multiple conditions `for` and `if` in one statement, worth extract a var like before 

```python
deleted_keys = (k for k, v in self.er_inventory.items() if v.marked_to_delete)
for key in deleted_keys:
    ...
```

And please doc if the item is removed instead of mark to delete, it will be orphan
```

File: reconcile/external_resources/manager.py [hunk 2]
```diff
@@ -311,7 +313,9 @@ def _build_external_resource(
 
     def _serialize_resource_input(self, resource: ExternalResource) -> str:
         return json.dumps(
-            resource.dict(exclude={"data": {FLAG_RESOURCE_MANAGED_BY_ERV2}})
+            resource.dict(
+                exclude={"data": {FLAG_RESOURCE_MANAGED_BY_ERV2, FLAG_DELETE_RESOURCE}}
+            )
```
hemslo, line 318: why do we need to exclude those fields? There is a similar method `serialize_input` in `ExternalResource`, and to serialize json, can directly call `resource.json(` instead of `resource.dict(` then `json.dumps`

lechuk47: - These fields are not expected in the module input. Since the `overrides` parameter doesn't have a schema, all the overridden parameters are passed to the terraform object directly (the same way as terraform-resources). So all parameters not expected in the input shall be removed. 

- I changed the call to `resource.json()` and removed the method in ExternalResource (this was used in an old implementation when the INPUT was passed as an env var)

reconcile/external_resources/manager.py
lechuk47: `multiple conditions for and if in one statement, worth extract a var like before`  I think we should define guidelines on these things because sometimes I get the contrary message (avoid defining a variable to just use it one time). 

Yup, abandoning resources will be documented.

hemslo: it depends on context, extract variable and inline variable are common refactoring pair, if it's hard to see the intention of a expression, then extract, if it's already clear from context like method name, then inline:

some examples:
* https://refactoring.guru/extract-variable
* https://refactoring.guru/inline-temp

reconcile/external_resources/manager.py
hemslo: Is there a way to ignore unexpected input or define a clear mapping? Otherwise we have to remember edit this part whenever we put new fields in `data`

hemslo (approved):
LGTM!

PR #4523, opened 2024-07-03
APPSRE-10229 add ocm manifests API

fishi0x01:
This is  a copy of the existing tests and implementation of Syncset. No changes apart from naming have been made. There is no intent of optimizing the current syncset tests/implementation.

### DIFF
--- a/reconcile/utils/ocm/manifests.py
+++ b/reconcile/utils/ocm/manifests.py
@@ -1,0 +1,39 @@
+from collections.abc import Generator, Mapping
+from dataclasses import dataclass
+from typing import Any
+
+from reconcile.utils.ocm_base_client import OCMBaseClient
+
+
+@dataclass
+class Manifest:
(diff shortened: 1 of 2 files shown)


reconcile/utils/ocm/manifests.py
```diff
@@ -0,0 +1,39 @@
...
+from dataclasses import dataclass
+from typing import Any
+
+from reconcile.utils.ocm_base_client import OCMBaseClient
+
+
+@dataclass
+class Manifest:
+    def __init__(self, cluster_id: str):
+        self.href = f"/api/clusters_mgmt/v1/clusters/{cluster_id}/external_configuration/manifests"
```
hemslo, line 11: Are all API operations under `external_configuration` follow the same convention? If yes, please extract common CRUD methods when adding another one.

hemslo (approved):

PR #4526, opened 2024-07-04
yesterday

maorfr:
all my troubles seemed so far away

use this function to template tickets that relate to cost-savings, directly into the Due Date field in jira.

:wink:

### DIFF
--- a/reconcile/utils/jinja2/utils.py
+++ b/reconcile/utils/jinja2/utils.py
@@ -213,7 +213,7 @@ def process_jinja2_template(
         "url": url_makes_sense,
         "s3": lookup_s3_object,
         "flatten_dict": flatten,
-        "yesterday": lambda: (datetime.now() - datetime.timedelta(1)).strftime('%Y-%m-%d'),
+        "yesterday": lambda: (datetime.datetime.now() - datetime.timedelta(1)).strftime('%Y-%m-%d'),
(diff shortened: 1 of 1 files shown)


patjlm (approved):
:laughing:

reconcile/utils/jinja2/utils.py
```diff
@@ -212,6 +213,7 @@ def process_jinja2_template(
         "url": url_makes_sense,
         "s3": lookup_s3_object,
         "flatten_dict": flatten,
+        "yesterday": lambda: (datetime.now() - datetime.timedelta(1)).strftime('%Y-%m-%d'),
```
patjlm, line 216:
```suggestion
        "yesterday": lambda: (datetime.datetime.now() - datetime.timedelta(1)).strftime('%Y-%m-%d'),
```

patjlm, line 216: or even fancier:
```suggestion
        "date_in_n_days": lambda n, format='%Y-%m-%d': (datetime.datetime.now() + datetime.timedelta(n)).strftime(format),
```
which works with negative numbers as well :wink:

maorfr: no it has to be yesterday

rporres (approved):

PR #4530, opened 2024-07-08
:recycle: openshift-serviceaccount-tokens: refactoring

chassing:
Refactor `openshift-serviceaccount-tokens`:

* GQL support
* Type hints
* Unittests

Preparing for work on [APPSRE-10518](https://issues.redhat.com/browse/APPSRE-10518)


reconcile/openshift_serviceaccount_tokens.py
```diff
@@ -47,117 +53,147 @@ def get_tokens_for_service_account(
...
+                        **sat.namespace.dict(by_alias=True),
+                        sharedResources=None,
+                        openshiftServiceAccountTokens=None,
+                    )
+                )
 
     return canonicalized_namespaces
 
 
+def get_serviceaccount_tokens(query_func: Callable) -> list[NamespaceV1]:
```
hemslo, line 153: naming is a little confusing, as return is a list of namespace instead of tokens

reconcile/openshift_base.py
```diff
@@ -1306,6 +1306,79 @@ def aggregate_shared_resources(namespace_info, shared_resources_type):
...
+            shared_type_resources_items = []
+            for or_shared_resources_item in namespace.shared_resources:
+                if shared_type_resources := getattr(
+                    or_shared_resources_item, "openshift_resources"
+                ):
+                    shared_type_resources_items.extend(shared_type_resources)
+            if namespace.openshift_resources:
+                namespace.openshift_resources.extend(shared_type_resources_items)
+            else:
+                namespace.openshift_resources = shared_type_resources_items
```
hemslo, line 1379: is this method safe to run multiple times? it seems changed fields like `namespace.openshift_service_account_tokens` will keep growing.

reconcile/openshift_serviceaccount_tokens.py
```diff
@@ -47,117 +53,147 @@ def get_tokens_for_service_account(
...
+    for namespace in namespaces:
+        ob.aggregate_shared_resources_typed(namespace)
+        if namespace.openshift_service_account_tokens:
+            canonicalized_namespaces.append(namespace)
+            for sat in namespace.openshift_service_account_tokens:
+                if f"{sat.namespace.cluster.name}/{sat.namespace.name}" in [
+                    f"{ns.cluster.name}/{ns.name}" for ns in canonicalized_namespaces
+                ]:
+                    # do not add duplicate namespaces to speed up fetch_current_state
+                    continue
```
hemslo, line 140: how about restructure `canonicalized_namespaces` to `dict` to do auto deduplicate like 

```python
    canonicalized_namespaces = {}
    for namespace in namespaces:
        ob.aggregate_shared_resources_typed(namespace)
        if namespace.openshift_service_account_tokens:
            canonicalized_namespaces[f"{namespace.cluster.name}/{namespace.name}"] = (
                namespace
            )
            for sat in namespace.openshift_service_account_tokens:
                canonicalized_namespaces[
                    f"{sat.namespace.cluster.name}/{sat.namespace.name}"
                ] = NamespaceV1(
                    **sat.namespace.dict(by_alias=True),
                    sharedResources=None,
                    openshiftServiceAccountTokens=None,
                )

    return list(canonicalized_namespaces.values())
```

may need to adjust tests as result order is undefined.

reconcile/openshift_base.py
chassing: It's not safe to run multiple times for a single namespace. I've copied the behavior of the untyped `aggregate_shared_resources` method.

hemslo: ok, then maybe doc it in docstring as it's not idempotent

reconcile/openshift_serviceaccount_tokens.py
chassing: Unfortunately, this won't work. It will override already existing namespaces with `openshiftServiceAccountTokens: [...]` with  `openshiftServiceAccountTokens: None`

reconcile/openshift_serviceaccount_tokens.py
chassing: Applied in f5c17c5

reconcile/openshift_base.py
chassing: Applied in [7866399](https://github.com/app-sre/qontract-reconcile/pull/4530/commits/786639915f0654194ba807310c49f167da56100b)

reconcile/openshift_serviceaccount_tokens.py
hemslo: then current code may have duplicates, like 2 namespaces `N1, N2`, `N1` has a service account token in `N2`, `N2` will be added when iterate `N1` service accounts, later it's added again when processing `N2`.

If we want to keep original `namespace` object, only append when not exist, then we should do it in 2 iterations like 

```python
    canonicalized_namespaces = {}
    for namespace in namespaces:
        ob.aggregate_shared_resources_typed(namespace)
        if namespace.openshift_service_account_tokens:
            canonicalized_namespaces[f"{namespace.cluster.name}/{namespace.name}"] = (
                namespace
            )
    for namespace in canonicalized_namespaces.values():
        for sat in namespace.openshift_service_account_tokens:
            key = f"{sat.namespace.cluster.name}/{sat.namespace.name}"
            if key not in canonicalized_namespaces:
                canonicalized_namespaces[key] = NamespaceV1(
                    **sat.namespace.dict(by_alias=True),
                    sharedResources=None,
                    openshiftServiceAccountTokens=None,
                )
    return list(canonicalized_namespaces.values())
```

chassing: You are right! 🙇 Applied in 4ac709a

hemslo (approved):
LGTM!

PR #4531, opened 2024-07-08
:sparkles: openshift-serviceaccount-tokens: create service account token secret for OpenShift 4.16+

chassing:
OpenShift [4.16+](https://docs.openshift.com/container-platform/4.16/release_notes/ocp-4-16-release-notes.html#ocp-4-16-notable-technical-changes_release-notes) doesn't automatically create a service account token secret for new service accounts anymore. This PR adds the token creation functionality into `openshift-serviceaccount-tokens`.

Ticket: [APPSRE-10518](https://issues.redhat.com/browse/APPSRE-10518)

Hint for the reviewers: We want to create the token secret and leave it "unmanaged" afterward. So we can leverage `ResourceInventory` instead of manual `oc apply`.

### DIFF
--- a/reconcile/openshift_serviceaccount_tokens.py
+++ b/reconcile/openshift_serviceaccount_tokens.py
@@ -1,4 +1,5 @@
 import logging
+import random
 import sys
 from collections.abc import Callable, Iterable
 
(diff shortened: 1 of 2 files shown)


esron (approved):

reconcile/openshift_serviceaccount_tokens.py
```diff
@@ -35,6 +36,31 @@ def construct_sa_token_oc_resource(name: str, sa_token: str) -> OR:
...
 
+def service_account_token_request(name: str) -> OR:
+    """Create a service account token secret for a given service account."""
+    body = {
+        "apiVersion": "v1",
+        "kind": "Secret",
+        "type": "kubernetes.io/service-account-token",
+        "metadata": {
+            # service-account-name-<random-number>
+            "name": f"{name}-token-{random.randrange(99999):05d}",
```
hemslo, line 47: Should we use a fixed name suffix to have a one-to-one mapping between legacy service account token and account, which means don't use any previously auto generated ones. So whenever we expect a service account token, we know the desired name, don't need to list current state then select.

reconcile/openshift_serviceaccount_tokens.py
```diff
@@ -35,6 +36,31 @@ def construct_sa_token_oc_resource(name: str, sa_token: str) -> OR:
...
+        },
+    }
+    return OR(
+        body,
+        # We are marking this token secret as "unmanaged" because we just want to create it
+        # and not manage it in the future.
+        # Openshift will delete this token secret automatically if the service account is deleted.
+        f"{QONTRACT_INTEGRATION}-unmanaged",
+        QONTRACT_INTEGRATION_VERSION,
+        error_details=name,
```
hemslo, line 60: any case we want to delete the token but keep the service account?

chassing: Not I can think of. 🤷

hemslo: if we don't mark it as `unmanaged`, just treat it like other resources in ri, will it conflict with token auto deletion when service account deleted?

hemslo (approved):
LGTM!

reconcile/openshift_serviceaccount_tokens.py
chassing: I've created a ticket https://issues.redhat.com/browse/APPSRE-10565 for that to discuss.

PR #4532, opened 2024-07-09
terraform-resources module support

maorfr:
depends on https://github.com/app-sre/qontract-schemas/pull/686

with this functionality, we can provision resources based on a module in a git repository:
```
  resources:
  - provider: module
    module:
      url: <repo url>
      path: <path to module>
      ref: <branch or commit sha>
    identifier: my-identifier
    defaults: path/to/file/with/variables.yaml
```


maorfr:
next problem to fix:
```
This module is not yet installed. Run "terraform init" to install all modules
```

PR #4533, opened 2024-07-09
[APPSRE-10253] - Add validation for rds enable blue_green_update

ninabauer:
Adds validation for rds enable blue_green_update

### DIFF
--- a/reconcile/utils/terraform_client.py
+++ b/reconcile/utils/terraform_client.py
@@ -784,7 +784,39 @@ class TerraformClient:  # pylint: disable=too-many-public-methods
                     "allow_major_version_upgrade is not enabled for upgrading RDS instance: "
                     f"{resource_name} to a new major version."
                 )
+            
+            blue_green_update = after.get("blue_green_update", {}).get("enabled", False)
+            if blue_green_update:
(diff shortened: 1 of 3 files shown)


reconcile/utils/terraform_client.py
```diff
@@ -785,6 +785,56 @@ def validate_db_upgrade(
...
+        account_name,
+        engine,
+        version,
+        parameter_group,
+        replica_source,
+        region_name,
+    ):
+        supported_versions = {
+            "mysql": ["5.7", "8.0.15"],
+            "postgres": ["11.21", "12.16", "13.12", "14.9", "15.4", "16.1"],
```
patjlm, line 810: Can we avoid hard coding these lists of versions somehow?

ninabauer: I tried to find an api that provided information specific to blue/green deployment compatibility kind of like how we used DescribeDBEngineVersions for a valid upgrade target but I did not find anything. Do you have any other ideas for how we could avoid hard coding the versions?

BumbleFeng: I think the only way here is hard coding.

BumbleFeng (approved):

PR #4539, opened 2024-07-10
avoid KeyError on promotion without success published

maorfr:
this error will be caught here to give a useful error message: https://github.com/app-sre/qontract-reconcile/blob/921282ce9bafee2e4413ec050d7d61e272c984b6/reconcile/utils/saasherder/saasherder.py#L1828-L1835

### DIFF
--- a/reconcile/utils/promotion_state.py
+++ b/reconcile/utils/promotion_state.py
@@ -95,7 +95,9 @@ class PromotionState:
         if use_cache and path_v2 in self._promotion_data_cache:
             return self._promotion_data_cache[path_v2]
 
-        data = self._state.get(path_v2)
+        data = self._state.get(path_v2, None)
+        if not data:
+            return None
         promotion_data = PromotionData(**data)
         self._promotion_data_cache[path_v2] = promotion_data
         return promotion_data


patjlm (approved):

reconcile/utils/promotion_state.py
```diff
@@ -95,7 +95,9 @@ def get_promotion_data(
         if use_cache and path_v2 in self._promotion_data_cache:
             return self._promotion_data_cache[path_v2]
 
-        data = self._state.get(path_v2)
+        data = self._state.get(path_v2, None)
+        if not data:
+            return None
```
fishi0x01, line 100: Should be caught by https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/test/utils/test_promotion_state.py#L66

fishi0x01: Apologies, that test is for cached commit shas ... We dont have a test for non-existing key without caching yet. Im happy to take a look tomorrow.

fishi0x01 (approved):

jmelis:
This LGTM, but I would like to see if there's a way to not raise an error in this situation, as defining pub/subs in a single MR should be something possible IMO

jmelis:
@maorfr but yes, perhaps that's for another MR :)

PR #4540, opened 2024-07-10
ruff upgrade and enable some additional checks

chassing:
Upgrade to latest ruff version and enable addtional checks:

- **:arrow_up: ruff 0.5.1**
- **:recycle: ruff: enable flake8-simplify checks**
- **:recycle: ruff: enable flake8-bugbear and fix B904**
- **:recycle: B019: Use of `functools.lru_cache` or `functools.cache` on methods can lead to memory leaks**
- **:recycle: B018: Found useless expression. Either assign it to a variable or remove it**
- **:recycle: B015: Pointless comparison. Did you mean to assign a value? Otherwise, prepend assert or remove it.**
- ** ignore B017**
- **fix other minor errors**
- **:recycle: ruff: enable flake8-comprehension**

### DIFF
--- a/reconcile/database_access_manager.py
+++ b/reconcile/database_access_manager.py
@@ -406,13 +406,10 @@ class JobStatus(BaseModel):
     conditions: list[JobStatusCondition]
 
     def is_complete(self) -> bool:
-        return True if self.conditions else False
+        return bool(self.conditions)
 
     def has_errors(self) -> bool:
-        for condition in self.conditions:
-            if condition.type == "Failed":
-                return True
-        return False
+        return any(condition.type == "Failed" for condition in self.conditions)
 
 
 def _populate_resources(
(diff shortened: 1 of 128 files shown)


reconcile/database_access_manager.py
```diff
@@ -406,13 +406,10 @@ class JobStatus(BaseModel):
...
     def is_complete(self) -> bool:
-        return True if self.conditions else False
+        return bool(self.conditions)
 
     def has_errors(self) -> bool:
-        for condition in self.conditions:
-            if condition.type == "Failed":
-                return True
-        return False
+        return any(condition.type == "Failed" for condition in self.conditions)
```
hemslo, line 412: can this be auto fixed now? I was waiting for this similar LINQ simplify transform from C# to port to Python for quite some time.

hemslo (approved):
LGTM! This will simplify future code reviews a lot, no need to check unnecessary wrappings like `if True: return True`, and auto simplify expressions using builtins like `any`. Thank you!

One rule maybe worth doing a knowledge sharing is `raise x from y` to keep trace https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/.

reconcile/database_access_manager.py
chassing: Yes, this comes from `flake8-simplify` and was auto-fixed.

PR #4541, opened 2024-07-10
refactor gitlab-permissions integration

mehfuz:
In an effort to generalize the intergration this the first step i.e refactoring the integrations as per
desired -> current -> reconcile pattern.


reconcile/utils/gitlab_api.py
```diff
@@ -686,6 +664,14 @@ def remove_labels(
         gitlab_request.labels(integration=INTEGRATION_NAME).inc()
         item.save()
 
+    @staticmethod
+    def is_group_project_owner(group_name: str, repo_url: str) -> bool:
+        """Utility function to check if the project is owned by the group by checking if group name part of repo url"""
+        substr = f"/{group_name}/"
+        if substr in repo_url:
+            return True
+        return False
```
hemslo, line 673: is there an official way to check group project owner instead of check url string?

reconcile/gitlab_permissions.py
```diff
@@ -1,21 +1,127 @@
...
+    ) -> None:
+        self.gl = gl
+        self.group_name = group_name
+        self.access_level_string = access
+        self.access_level = self.gl.get_access_level(access)
+        self.dry_run = dry_run
+
+    def run(self, repos: list[str]) -> None:
+        # get repos not owned by app-sre
+        non_app_sre_project_repos = {
```
hemslo, line 40: get rid of `app_sre`

reconcile/gitlab_permissions.py
```diff
@@ -1,21 +1,127 @@
...
+            self.group_name
+        )
+        current_state = {
+            project_repo_url: GroupSpec(
+                shared_projects[project_repo_url]["group_name"],
+                shared_projects[project_repo_url]["group_access_level"],
+            )
+            for project_repo_url in shared_projects
+        }
+        self.reconcile(desired_state, current_state, group_id)
```
hemslo, line 62: why need to pass `group_id` instead of fetch and keep `group` object?

reconcile/gitlab_permissions.py
```diff
@@ -1,21 +1,127 @@
...
+        for repo in diff_data.add.keys():
+            project = self.gl.get_project(repo)
+            if not self.can_share_project(project):
+                logging.error(
+                    "%s is not shared with %s as %s",
+                    repo,
+                    self.gl.user.username,
+                    self.access_level_string,
+                )
+                return None
```
hemslo, line 92: this will terminate loop and no exceptions throw, the integration will always success

reconcile/gitlab_permissions.py
```diff
@@ -1,21 +1,127 @@
...
+        for repo in diff_data.change.keys():
+            project = self.gl.get_project(repo)
+            if not self.can_share_project(project):
+                logging.error(
+                    "%s is not shared with %s as %s",
+                    repo,
+                    self.gl.user.username,
+                    self.group_name,
+                )
+                return None
```
hemslo, line 111: this will terminate loop and no exceptions throw, the integration will always success

reconcile/utils/gitlab_api.py
mehfuz: there should be a way but this is a hack....gitlab follows a standard url for repos i.e https://{instance}/{group_ OR_member}/{reponame}...It will save additional call and processing..

hemslo: see if any group info or project info can be used directly to check, and need to handle multi level urls, like `https://{instance}/{group}/{sub_group}/{repo}`

mehfuz: @hemslo  the `get_all_group_projects`is a single  function will take care of this. It returns all the project inside group (owned+shared+inherited). In case of owned+inherited from parent we cannot share/unshare this project so we keep it as is... https://github.com/app-sre/qontract-reconcile/blob/f977b7143a732ad951456b4051189e4f74b626a5/reconcile/gitlab_permissions.py#L67

hemslo: I don't understand that logic, especially comments inside `extract_group_spec`, this worth doing a review session.

reconcile/gitlab_permissions.py
```diff
@@ -1,21 +1,137 @@
...
+            project = self.gl.get_project(repo)
+            if not self.can_share_project(project):
+                errors.append(
+                    GroupAccessLevelError(
+                        f"{repo} is not shared with {self.gl.user.username} as {self.access_level_string}"
+                    )
+                )
+                continue
+            logging.info([
+                f"reshare_group_{self.group.name}_as_{self.access_level_string}",
```
hemslo, line 121: too many `_` in logging message

reconcile/utils/gitlab_api.py
hemslo: Now we can simplify `extract_group_spec` using `next`, to indicate we always expect a match:

```python
    def extract_group_spec(self, project: GroupProject) -> GroupSpec:
        return next(
            GroupSpec(
                group_name=self.group.name,
                group_access_level=group["group_access_level"],
            )
            for group in project.shared_with_groups
            if group["group_id"] == self.group.id
        )
```

Then construct current state as:

```python
        current_state = {
            project.web_url: self.extract_group_spec(project)
            for project in shared_projects
        }
```

And since `group_name` is always `self.group.name`, it's not required to be in spec for this use case, but can be kept for later support multiple groups.

reconcile/gitlab_permissions.py
hemslo: same in `reshare_with_`

hemslo: I mean other log messages are not using `_` , any reason for this one to be different?

mehfuz: @hemslo now I understand..let me make these logs similar to other integration.

mehfuz:
```
['reshare', '<repo>', '<group_name>', '<access_level>']
```

hemslo (approved):
LGTM!

PR #4555, opened 2024-07-15
APPSRE-10086 sapm cron schedule

fishi0x01:
Adding `schedule` gate to saas auto-promotions.

SAPM will obey to the given cron expression and only schedule auto-promotions if they fall within the schedule.

Note, that we do not add this gate to saasherder, as it could block unrelated promotions. The `schedule` feature is relevant for auto-promotions only.

Schema: https://github.com/app-sre/qontract-schemas/pull/680

### DIFF
--- a/reconcile/saas_auto_promotions_manager/subscriber.py
+++ b/reconcile/saas_auto_promotions_manager/subscriber.py
@@ -123,6 +123,14 @@ class Subscriber:
         return delta >= timedelta(days=self.soak_days)
 
     def _is_valid_deployment_window(self) -> bool:
+        # Ideally we would catch that at schema validation time
+        if not croniter.is_valid(self.schedule):
+            logging.error(
+                "Subscriber at %s has an invalid schedule declaration %s. We will block any promotion for that target until this is fixed.",
+                self.target_file_path,
+                self.schedule,
+            )
+            return False
         return croniter.match(self.schedule, datetime.now(UTC))
 
     def _compute_desired_ref(self) -> None:
--- a/reconcile/saas_auto_promotions_manager/subscriber.py
+++ b/reconcile/saas_auto_promotions_manager/subscriber.py
@@ -45,6 +45,7 @@ class Subscriber:
         uid: str,
         soak_days: int,
         blocked_versions: set[str],
+        schedule: str,
     ):
         self.saas_name = saas_name
         self.template_name = template_name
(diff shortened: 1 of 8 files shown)


reconcile/saas_auto_promotions_manager/utils/saas_files_inventory.py
```diff
@@ -102,6 +104,20 @@ def _assemble_subscribers_with_auto_promotions(self) -> None:
...
+                        else "* * * * *"
+                    )
+                    # Ideally we would catch that at schema validation time
+                    if not croniter.is_valid(schedule):
+                        logging.error(
+                            "Subscriber at %s has an invalid schedule declaration %s. We will block any promotion for that target until this is fixed.",
+                            file_path,
+                            schedule,
+                        )
+                        # Lets set schedule for impossible date to block promotions
```
chassing, line 119: Why not raise an exception here?

fishi0x01: Good point. I moved the validation logic now to the subscriber itself - makes this a lot easier.

chassing (approved):

reconcile/saas_auto_promotions_manager/subscriber.py
```diff
@@ -118,6 +122,17 @@ def _passed_accumulated_soak_days(self) -> bool:
...
+    def _is_valid_deployment_window(self) -> bool:
+        # Ideally we would catch that at schema validation time
+        if not croniter.is_valid(self.schedule):
+            logging.error(
+                "Subscriber at %s has an invalid schedule declaration %s. We will block any promotion for that target until this is fixed.",
+                self.target_file_path,
+                self.schedule,
+            )
+            return False
+        return croniter.match(self.schedule, datetime.now(UTC))
```
hemslo, line 134: @fishi0x01 How do we ensure there is an integration run at the matching time? For example `1 * * * *` to run on min 1 every hour, but `datetime.now(UTC)` may miss that if the integration finish at min 0 then sleep for 10 mins, miss that time window.

fishi0x01: Auto-promotion windows are more tailored towards bigger time windows, e.g., only promote during working hours.

hemslo: then we need to doc or limit that so no unexpected schedule miss.

PR #4556, opened 2024-07-16
Support forced re-runs of Terraform Repo with a new key in the schema file

cubismod:
[APPSRE-8705](https://issues.redhat.com/browse/APPSRE-8705)

Depends on https://github.com/app-sre/qontract-schemas/pull/691, see that MR for more background on why this approach was chosen. Re-runs are implemented in Terraform Repo by just adding a new key that is considered during the diffing process between `existing` and `desired`. If a user updates the timestamp in an MR, then tf-repo will perform a `terraform plan` & `apply` again. No changes are needed to executor since this key isn't used by executor.

### DIFF
--- a/reconcile/terraform_repo.py
+++ b/reconcile/terraform_repo.py
@@ -361,7 +361,7 @@ class TerraformRepoIntegration(
                 )
             if self.params.validate_git:
                 self.check_ref(d.repository, d.ref)
-            if d.force_rerun_timestamp:
+            if c.force_rerun_timestamp != d.force_rerun_timestamp:
                 logging.info("user has forced a re-run of tf-repo execution")
 
         if len(merged) != 0:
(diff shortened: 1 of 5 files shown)


reconcile/terraform_repo.py
```diff
@@ -361,6 +361,8 @@ def calculate_diff(
                 )
             if self.params.validate_git:
                 self.check_ref(d.repository, d.ref)
+            if d.force_rerun_timestamp:
```
hemslo, line 364: how is `force_rerun_timestamp` stored in current state, if not persisted, will this always trigger rerun?

cubismod: The value will be stored in S3 state here: https://github.com/app-sre/qontract-reconcile/blob/828dd8293828dec3e33536358d71620a94e7671b/reconcile/terraform_repo.py#L270 and considered just like any other value in the diffing process so after its written to state once, it should not trigger a subsequent re-run until the value is updated or removed.

esron (approved):
LGTM

PR #4558, opened 2024-07-19
APPSRE-10615 DT token scope updates

fishi0x01:
- adding token scope update functionality to DT client
- restructure test - Callable builder fixture not needed


mehfuz (approved):
LGTM...

PR #4560, opened 2024-07-22
Implement Early-exit in External-Resources integration

lechuk47:
- **Check resource diffs when a resource falls into a PENDING_SERET_SYNC state**
- **Implement Early exit for external-resources integration**

APPSRE-10626

### DIFF
--- a/reconcile/external_resources/integration.py
+++ b/reconcile/external_resources/integration.py
@@ -1,5 +1,6 @@
 import logging
 from collections.abc import Callable
+from typing import Any
 
 from reconcile.external_resources.manager import (
     ExternalResourcesInventory,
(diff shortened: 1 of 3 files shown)


reconcile/external_resources/integration.py
```diff
@@ -141,3 +168,29 @@ def run(
...
+            workers_cluster=kwargs["workers_cluster"],
+            workers_namespace=kwargs["workers_namespace"],
+            thread_pool_size=kwargs["thread_pool_size"],
+            dry_run=True,
+            dry_run_job_suffix=kwargs["dry_run_job_suffix"],
+        )
+
+        return {
+            "desired": er_mgr._get_desired_objects_reconciliations(),
+            "deleted": er_mgr._get_deleted_objects_reconciliations(),
```
hemslo, line 195: should these 2 methods become public or have a new one for this use case?

hemslo (approved):

reconcile/external_resources/integration.py
lechuk47: Yup, it's more correct. PTAL

PR #4561, opened 2024-07-22
[qontract-cli/saas-dev] use target spec to calculate parameters

maorfr:
this functionality is described in https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/continuous-delivery-in-app-interface.md#developer-workflow

this PR reduces duplicate code to use the built-in functionality of TargetSpec.


tools/qontract_cli.py
```diff
@@ -3497,21 +3498,17 @@ def saas_dev(ctx, app_name=None, saas_file_name=None, env_name=None) -> None:
...
-                            continue
-                        if replace_pattern in v:
-                            parameters[k] = v.replace(replace_pattern, replace_value)
+                parameters = TargetSpec(
+                    saas_file=saas_file,
+                    resource_template=rt,
+                    target=target,
+                    # process_template options
+                    image_auth=None,  # type: ignore[arg-type]
+                    hash_length=None,  # type: ignore[arg-type]
```
patjlm, line 3507: are these 2 not needed?

patjlm (approved):
TIL about this CLI command

PR #4563, opened 2024-07-24
ERv2 output resource format tests + ad-hoc reconciliations request cmd

lechuk47:
- **Add OutputSecretsFormatter tests**
- **Add a CLI command to request a resource reconciliation**

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -56,13 +56,21 @@ from reconcile.cli import (
     config_file,
     use_jump_host,
 )
+from reconcile.external_resources.integration import (
+    get_aws_api,
+)
 from reconcile.external_resources.manager import (
     FLAG_RESOURCE_MANAGED_BY_ERV2,
     setup_factories,
 )
 from reconcile.external_resources.model import (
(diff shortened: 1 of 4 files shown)


tools/qontract_cli.py
```diff
@@ -3912,5 +3920,47 @@ def get_input(
...
+@click.argument("provision-provider", required=True)
+@click.argument("provisioner", required=True)
+@click.argument("provider", required=True)
+@click.argument("identifier", required=True)
+@click.pass_context
+def request_reconciliation(
+    ctx, provision_provider: str, provisioner: str, provider: str, identifier: str
+):
+    """Marks a resource as it needs to get reconciled. The itegration will reconcile the resource at
+    its next iteration.
```
hemslo, line 3933: what is the use case of manual status change?

lechuk47: It's useful for testing new module versions, debugging purposes, etc. It's not intended to be used extensively but it helps in determined cases. Specially now with the initial migration.

hemslo (approved):
LGTM!

PR #4569, opened 2024-07-28
[saas-file-validator] validate quay image patterns in quay organizations

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10662

`imagePatterns` -

* we want to allow freedom to self-service (https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/80763)
* we want to restrict usage to registries we trust (https://github.com/app-sre/qontract-schemas/blob/7e64f71e4518b0519c4c2730ffe8f73ad642ecad/schemas/app-sre/saas-file-2.yml#L149-L154)
* we want to prevent usage of just any repository (such as personal, https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/113413)

this PR restricts usage of images from quay to only ones that have a quay org file in app-interface.

this approach will allow a review when users want to use additional images.


reconcile/saas_file_validator.py
```diff
@@ -48,6 +49,16 @@ def run(dry_run: bool, defer: Callable | None = None) -> None:
...
     for r in missing_repos:
         logging.error(f"repo is missing from codeComponents: {r}")
+    app_int_quay_instances = {i.url for i in get_quay_instances()}
+    app_int_quay_orgs = {f"{o.instance.url}/{o.name}" for o in get_quay_orgs()}
+    missing_image_patterns = [
+        p
+        for p in saasherder.image_patterns
+        if any(p.startswith(i) for i in app_int_quay_instances)
+        and not any(p.startswith(o) for o in app_int_quay_orgs)
+    ]
```
hemslo, line 59: using `startswith` may miss some checks like there is an `org` name is `app`, it will match any org name start with `app` like `appsre`, how about parse and strict check like this

```python

    app_int_quay_instances = {i.url for i in get_quay_instances()}
    app_int_quay_orgs = {(o.instance.url, o.name) for o in get_quay_orgs()}

    missing_image_patterns = [
        p
        for p in saasherder.image_patterns
        if (parts := p.split("/"))
        and parts[0] in app_int_quay_instances
        and (parts[0], parts[1]) not in app_int_quay_orgs
    ]
```

maorfr: i am being consistent with https://github.com/app-sre/qontract-reconcile/blob/647017bd71a443613296d7e0f71c5ed3dc1ece85/reconcile/utils/saasherder/saasherder.py#L1106

if we want to change this in a follow up PR that's probably ok.

hemslo: that one is generic, other registries have different url patterns, so parse logic are different, but here we only check quay, the url pattern of base and org are known. And double `startswith` check to find all patterns time complexity is `O(N*I*O)`, parts `in` check is `O(N)`, `N` is number of image patters, `I` is number of quay instances, `O` is number of quay orgs.

hemslo: and to avoid wrong index access, can guard split results first

```python
    missing_image_patterns = [
        p
        for p in saasherder.image_patterns
        if (parts := p.split("/"))
        and len(parts) >= 2
        and parts[0] in app_int_quay_instances
        and (parts[0], parts[1]) not in app_int_quay_orgs
    ]
```

maorfr: https://github.com/app-sre/qontract-reconcile/pull/4569/commits/467b48cf3f8103b85a043522e164de41dde78d3a

hemslo (approved):
LGTM!

PR #4577, opened 2024-07-31
saas re-use check_in time

fishi0x01:
A successful saas re-deployment of the same ref currently overrides the `check_in` time. This can easily block saas files which rely on the soakDays feature. Example https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/113754

In this PR we avoid overrides of `check_in` time if the deployment was successful. On unsuccessful deployments, we update the `check_in` time, i.e., we reset the soakDays counter for this target.


esron (approved):
LGTM

PR #4580, opened 2024-08-01
Introduce Module settings overrides at resource level

lechuk47:
- **Add OutputSecretsFormatter tests**
- **Add a CLI command to request a resource reconciliation**
- **Introduce module settings overrides at resource level**


reconcile/external_resources/model.py
```diff
@@ -60,30 +65,48 @@ def state_path(self) -> str:
...
+                exclude={
+                    FLAG_RESOURCE_MANAGED_BY_ERV2,
+                    FLAG_DELETE_RESOURCE,
+                    MODULE_OVERRIDES,
+                }
+            ),
+            namespace=namespace.dict(),
+        )
+        spec.metadata[FLAG_DELETE_RESOURCE] = resource.delete
+        spec.metadata[MODULE_OVERRIDES] = resource.module_overrides
```
hemslo, line 94: can `metadata` be set in constructor directly like?

```
ExternalResourceSpec(
    ...
    metadata={
        FLAG_DELETE_RESOURCE: resource.delete,
        MODULE_OVERRIDES: module_overrides
    }
)
```

reconcile/external_resources/model.py
```diff
@@ -181,14 +204,21 @@ def image_version(self) -> str:
...
-        # TODO: Modify resource schemas to include this attributes
-        data = {
-            "image": module.image,
-            "version": module.default_version,
-            "reconcile_drift_interval_minutes": module.reconcile_drift_interval_minutes,
-            "reconcile_timeout_minutes": module.reconcile_timeout_minutes,
-        }
-        return ExternalResourceModuleConfiguration.parse_obj(data)
+        if not (module_overrides := spec.metadata.get("module_overrides")):
+            module_overrides = ExternalResourcesModuleOverridesV1(
```
hemslo, line 208: seems we always want to assign `module_overrides`, then can simplify it

```python
module_overrides = spec.metadata.get("module_overrides") or ExternalResourcesModuleOverridesV1(...)
```

hemslo (approved):
LGTM!

reconcile/external_resources/model.py
lechuk47: Nope, metadata is not part of the constructor. It's excluded because `ExternalResourceSpec` is shared with `terraform-resources`.

reconcile/external_resources/model.py
lechuk47: fixed.

PR #4582, opened 2024-08-01
APPSRE-10615 sync token with spec

fishi0x01:
Enable DTP to detect diffs between token spec in app-interface and current existing API token in Dynatrace. If diff is detected, then DTP will make an API call to DT to properly configure the API token.

We use the DT token name to store a config hash (currently scope hash). By doing so, we can detect diffs between the token in Dynatrace and the token spec defined in app-interface. Further, we do not need any additional API calls, as we must query the token name and id with list queries anyways. Note, that a list query on tokens does not return every detail about each token, such as scopes. We do not need to query any additional information about each API token (such as scopes). By comparing the hash in the name, we know if we need to POST an update for the token.

Note, this will rename all existing DTP tokens to the new schema. Renaming a token does not have any further side-effect, so this is fine.

### DIFF
--- a/reconcile/dynatrace_token_provider/integration.py
+++ b/reconcile/dynatrace_token_provider/integration.py
@@ -1,6 +1,7 @@
 import base64
+import hashlib
 import logging
-from collections.abc import Iterable, Mapping
+from collections.abc import Iterable, Mapping, MutableMapping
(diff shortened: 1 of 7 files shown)


esron (approved):
LGTM

PR #4586, opened 2024-08-02
APPSRE-10615 DTP pr-check

fishi0x01:
We cannot catch all spec misconfiguration via json schema. In this PR we add spec validation.

### DIFF
--- a/reconcile/dynatrace_token_provider/validate.py
+++ b/reconcile/dynatrace_token_provider/validate.py
@@ -40,9 +40,9 @@ def validate_token_specs(specs: Iterable[DynatraceTokenProviderTokenSpecV1]) -> 
                     )
                 seen_tokens.add(token.name)
 
-                if token.key_name_in_secret:
-                    if token.key_name_in_secret in seen_key_names:
-                        raise KeyNameNotUniqueInSecretError(
-                            f"A key name must be unique within a secret. Key name '{token.key_name_in_secret}' is used more than once in secret '{secret.name}' in token spec '{spec.name}'."
-                        )
-                    seen_key_names.add(token.key_name_in_secret)
+                secret_key = token.key_name_in_secret or token.name
+                if secret_key in seen_key_names:
+                    raise KeyNameNotUniqueInSecretError(
+                        f"A key name must be unique within a secret. Key name '{secret_key}' is used more than once in secret '{secret.name}' in token spec '{spec.name}'."
+                    )
+                seen_key_names.add(secret_key)
--- a/reconcile/dynatrace_token_provider/validate.py
+++ b/reconcile/dynatrace_token_provider/validate.py
@@ -1,0 +1,60 @@
+from collections.abc import Iterable
+
+from reconcile.gql_definitions.dynatrace_token_provider.token_specs import (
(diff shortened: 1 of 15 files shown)


maorfr:
isn't pr-check == dry-run?

and why wouldn't we do this on non-dry-run as well?

fishi0x01:
@maorfr `validate_token_specs` runs always: `--pr-check`, `--dry-run` and `--no-dry-run`. The idea behind `--pr-check` was to avoid any calls to the OCM API. But I agree that we should be fine with some extra calls and just do `--dry-run`.

reconcile/dynatrace_token_provider/validate.py
```diff
@@ -0,0 +1,60 @@
...
+                )
+            seen_secrets.add(secret_definition_key)
+
+            seen_tokens: set[str] = set()
+            seen_key_names: set[str] = set()
+            for token in secret.tokens:
+                if len(token.name) > MAX_TOKEN_NAME_LENGTH:
+                    raise TokenNameTooLongError(
+                        f"Token name '{token.name}' is too long. The maximum length is {MAX_TOKEN_NAME_LENGTH} characters."
+                    )
```
hemslo, line 47: size limit can be done by schema `maxLength`

reconcile/dynatrace_token_provider/validate.py
```diff
@@ -0,0 +1,60 @@
...
+                        f"A token name must be unique within a secret. Token name '{token.name}' is used more than once in secret '{secret.name}' in token spec '{spec.name}'."
+                    )
+                seen_tokens.add(token.name)
+
+                if token.key_name_in_secret:
+                    if token.key_name_in_secret in seen_key_names:
+                        raise KeyNameNotUniqueInSecretError(
+                            f"A key name must be unique within a secret. Key name '{token.key_name_in_secret}' is used more than once in secret '{secret.name}' in token spec '{spec.name}'."
+                        )
+                    seen_key_names.add(token.key_name_in_secret)
```
hemslo, line 60: `token.name` and `token.key_name_in_secret` used as `secret_key = token.key_name_in_secret or token.name`, should apply combined unique to `secret_key`? https://github.com/app-sre/qontract-reconcile/blob/e41d5c27eec7fc8a181b1d6636d25cfa7c0d891f/reconcile/dynatrace_token_provider/integration.py#L381

reconcile/dynatrace_token_provider/validate.py
fishi0x01: Oh nice! Created https://github.com/app-sre/qontract-schemas/pull/699 and removed token name length validation.

reconcile/dynatrace_token_provider/validate.py
fishi0x01: Awesome catch! Added a new test case for this scenario

hemslo: Is there a use case for `token.name` to be unique? Or just `secret_key` unique is enough?

fishi0x01: Yes `token.name` is used for name in DT API and should be unique.

hemslo (approved):
LGTM!

PR #4599, opened 2024-08-06
APPSRE-10615 more DTP metrics

fishi0x01:
- Fix cluster metric -> previously we also counted filtered clusters
- add token gauge -> `dtp_tokens_managed{ ocm_env_name, dt_tenant_id }`

### DIFF
--- a/reconcile/dynatrace_token_provider/integration.py
+++ b/reconcile/dynatrace_token_provider/integration.py
@@ -3,12 +3,14 @@ import hashlib
 import logging
 from collections.abc import Iterable, Mapping, MutableMapping
 from datetime import timedelta
+from threading import Lock
 from typing import Any
 
 from reconcile.dynatrace_token_provider.dependencies import Dependencies
 from reconcile.dynatrace_token_provider.metrics import (
     DTPClustersManagedGauge,
     DTPOrganizationErrorRate,
+    DTPTokensManagedGauge,
 )
 from reconcile.dynatrace_token_provider.model import DynatraceAPIToken, K8sSecret
 from reconcile.dynatrace_token_provider.ocm import (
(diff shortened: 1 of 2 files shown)


reconcile/dynatrace_token_provider/integration.py
```diff
@@ -69,7 +71,30 @@ def run(self, dry_run: bool) -> None:
...
+                    DTPTokensManagedGauge(
+                        integration=self.name,
+                        ocm_env=ocm_env_name,
+                        dt_tenant_id=dt_tenant_id,
+                    ),
+                    cnt,
+                )
+
     def reconcile(self, dry_run: bool, dependencies: Dependencies) -> None:
+        self._lock = Lock()
```
hemslo, line 95: why use lock, which part is run in threads?

reconcile/dynatrace_token_provider/integration.py
```diff
@@ -69,7 +71,30 @@ def run(self, dry_run: bool) -> None:
         dependencies.populate()
         self.reconcile(dry_run=dry_run, dependencies=dependencies)
 
+    def _token_cnt(self, dt_tenant_id: str, ocm_env_name: str) -> None:
+        with self._lock:
+            if ocm_env_name not in self._managed_tokens_cnt:
+                self._managed_tokens_cnt[ocm_env_name] = {}
+            if dt_tenant_id not in self._managed_tokens_cnt[ocm_env_name]:
+                self._managed_tokens_cnt[ocm_env_name][dt_tenant_id] = 0
+            self._managed_tokens_cnt[ocm_env_name][dt_tenant_id] += 1
```
hemslo, line 80: by using `defaultdict` and `Counter`, all key checks can be saved

first init `_managed_tokens_cnt` as `defaultdict`

```python
    self._managed_tokens_cnt = defaultdict(Counter)
```

then start count directly

```python
    self._managed_tokens_cnt[ocm_env_name][dt_tenant_id] += 1
```

fishi0x01: Good idea! Fixed

reconcile/dynatrace_token_provider/integration.py
fishi0x01: Currently single threaded, but I want to keep this thread-safe as it is quite possible that we might need concurrency.

hemslo: ok, then please move instance vars init (`_lock` and `_managed_tokens_cnt`) to `__init__` so we can clearly see what vars are available.

hemslo (approved):
LGTM!

reconcile/dynatrace_token_provider/integration.py
fishi0x01: Done :+1:

PR #4602, opened 2024-08-06
mypy ci fixes

chassing:
We also need to install the test requirements in your CI `type` test. Otherwise, `mypy` can't validate all our test Python files, and errors may be hidden. 
Especially, `mypy` shows different output/errors in the local developer environment than in the CI checks.


esron (approved):
LGTM

rporres (approved):

hemslo (approved):
LGTM!

PR #4606, opened 2024-08-07
APPSRE-10615 DTP early-exit

fishi0x01:


reconcile/dynatrace_token_provider/integration.py
```diff
@@ -64,6 +67,14 @@ def __init__(self, params: DynatraceTokenProviderIntegrationParams) -> None:
         self._lock = Lock()
         self._managed_tokens_cnt: dict[str, Counter[str]] = defaultdict(Counter)
 
+    def get_early_exit_desired_state(self, *args: Any, **kwargs: Any) -> dict[str, Any]:
+        """Return the desired state for early exit."""
+        return {
+            "specs": [
+                spec.dict() for spec in get_dynatrace_token_provider_token_specs()
+            ]
```
hemslo, line 75: the order of `get_dynatrace_token_provider_token_specs()` maybe different, better to use a dict instead of list. like `(spec.namespace, space.name): spec.dict()`

hemslo (approved):
LGTM!

PR #4608, opened 2024-08-08
APPSRE-10615 DTP default spec and logging

fishi0x01:
Lets name the default spec explicitly to what it is.
Also added more logging which is interesting for dry-runs


hemslo (approved):
LGTM!

PR #4610, opened 2024-08-09
fix qr promote ticks

fishi0x01:
Previously we didnt have `'` here. This breaks our manual `make qr-promote` command.


fishi0x01:
Fixing in promote script instead

PR #4615, opened 2024-08-12
Refactor aws ami cleanup

rporres:
Refactored the ami cleanup integration so that it searches directly in the AWS accounts to find if an AMI is still being used in the latest version of a launch template instead of searching it in app-interface ASG definitions. We're using the AMIs we build in ASGs defined outside app-interface (Jenkins, bastion) so we need a new way to make sure that it is safe to be deleted. In order to know where to look for Launch templates, we go into the `sharing` section of the account where we want to cleanup to get aws accounts in which these AMIs may be used

This patch also refactors a bit the integration:

* to use typed queries to get aws accounts
* to minimise the use of the non-typed AWSApi.

so it may be helpful to review it commit by commit

[APPSRE-10617](https://issues.redhat.com/browse/APPSRE-10617)


reconcile/aws_ami_cleanup/integration.py
```diff
@@ -318,4 +274,4 @@ def run(dry_run: bool, thread_pool_size: int, defer: Callable | None = None) ->
                         else:
                             raise
 
-    sys.exit(exit_code)
+    sys.exit(ExitCodes.SUCCESS)
```
fishi0x01, line 277: Why do we need `sys.exit` here?

fishi0x01 (approved):

reconcile/aws_ami_cleanup/integration.py
rporres: we don't need it 😂 Fixed in 10e0c29c

fishi0x01: Awesome refactor btw! :heart:

PR #4616, opened 2024-08-12
[gql] allow defining current sha

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10734

this is to enable usage of this function while supplying two shas to compare.


patjlm:
should the PR be closed?

PR #4620, opened 2024-08-14
introduce change-management integration

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10755

this integration analyzes bundle diffs which were previously collected.
this PR is mostly for the design doc and is not currently intended to be merged.


reconcile/change_owners/change_log_tracking.py
```diff
@@ -0,0 +1,69 @@
...
+                        context="",
+                        origin="",
+                        context_file=change.fileref,
+                        approvers=[],
+                    )
+                    covered_diffs = change.cover_changes(ctx)
+                    if covered_diffs:
+                        change_log[commit].append({
+                            "change_type": ctp.name,
+                        })
```
hemslo, line 66: one commit has many changes, one change has many change type processors, can there be duplicate `ctp.name` for the same commit?

reconcile/change_owners/change_log_tracking.py
```diff
@@ -0,0 +1,69 @@
...
+            for ctp in fetch_change_type_processors(
+                gql.get_api(), NoOpFileDiffResolver()
+            )
+            if ctp.labels and "change_log_tracking" in ctp.labels
+        ]
+        state = init_state(
+            integration=self.name,
+        )
+        int_state_path = state.state_path
+        state.state_path = "bundle-archive/diff"
```
hemslo, line 38: mutate `state_path` here then reset back is not very clean, can we allow `init_state` to accept `state_path` to override on `State.__init__`, then we can have one for integration state, another for `diff` state.

reconcile/change_owners/change_log_tracking.py
```diff
@@ -0,0 +1,69 @@
...
+        change_log: dict[str, list[dict]] = {}
+        for item in state.ls():
+            key = item.lstrip("/")
+            commit = key.rstrip(".json")
+            logging.info(f"Processing commit {commit}")
+            change_log.setdefault(commit, [])
+            obj = state.get(key, None)
```
hemslo, line 45: This means we always need to process all diff objects in s3 one by one, can we load saved `bundle-diffs.json` first then only process missing commits?
Then `bundle-diffs.json` can keep all diffs even diff objects deleted after expire, we can also include some metadata field like timestamp to help reduce size if needed later, or delete commits when diff no longer exists.

reconcile/change_owners/change_log_tracking.py
```diff
@@ -0,0 +1,69 @@
...
+        change_type_processors = [
+            ctp
+            for ctp in fetch_change_type_processors(
+                gql.get_api(), NoOpFileDiffResolver()
+            )
+            if ctp.labels and "change_log_tracking" in ctp.labels
+        ]
+        state = init_state(
+            integration=self.name,
+        )
```
hemslo, line 36: `init_state` requires `cleanup`, either use `defer` or `with`

reconcile/change_owners/change_log_tracking.py
maorfr: there can be duplicates indeed. i handled it on the qontract-cli side, but you're right - it's confusing.

in https://github.com/app-sre/qontract-reconcile/pull/4620/commits/54804f652c14645769c3daa6bccbdb5c37d3b738 i've switched to proper classes to make this more readable (and used a set).

reconcile/change_owners/change_log_tracking.py
maorfr: i took a different direction in https://github.com/app-sre/qontract-reconcile/pull/4620/commits/1c86071e96c02a496a336dc7fb9ecb8cae52fc60

reconcile/change_owners/change_log_tracking.py
maorfr: added in https://github.com/app-sre/qontract-reconcile/pull/4620/commits/18eca8143b38ef3fa18306b241168d2d5abc5a25

reconcile/change_owners/change_log_tracking.py
maorfr: actually used 2 states: https://github.com/app-sre/qontract-reconcile/pull/4620/commits/d33d772d24be985a0b2928c3cdf3dbf4cc851271

reconcile/change_owners/change_log_tracking.py
maorfr: and some updates in https://github.com/app-sre/qontract-reconcile/pull/4620/commits/9c8bc716ebda6ca25cbe4daebc74afcef3ff85e8

hemslo: Why switch back to `list`? If it's because of json deserialize type issue, can use `BaseModel` instead of `dataclass`, it can handle auto convert list to set.

reconcile/change_owners/change_log_tracking.py
maorfr: this is a great idea. added a click.option implementation in https://github.com/app-sre/qontract-reconcile/pull/4620/commits/3f6215f963f1477834130abed7a0c94fafda0704 in case we need to re-process at some point.

hemslo: nice! a few performance and edge case handling:

* first time `integration.get` may fail if the file not exist
* ChangeLog.items can be a dict, key is commit, value is ChangeLogItem for faster lookup, as the number of commits can be huge
* `if not self.params.process_existing:` I think you mean `if self.params.process_existing:`
* can skip upload to s3 if `change_log` is same as before

maorfr:
we want a first working version to be in production to stitch other connections, like s3 -> configmap.

this version as it is currently is working, but efficiency can be improved (in following iterations?)

PR #4621, opened 2024-08-14
print sqs messages on error

fishi0x01:
When sqs consumer fails it would be nice to see the messages it failed on. Usually this stems from some formatting issue which is much easier to spot if we see the message sample in the logs.

### DIFF
--- a/reconcile/utils/sqs_gateway.py
+++ b/reconcile/utils/sqs_gateway.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import os
 
 from reconcile.utils.aws_api import AWSApi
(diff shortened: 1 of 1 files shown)


reconcile/utils/sqs_gateway.py
```diff
@@ -57,12 +58,16 @@ def receive_messages(
...
+        try:
+            messages = self.sqs.receive_message(
+                QueueUrl=self.queue_url,
+                VisibilityTimeout=visibility_timeout,
+                WaitTimeSeconds=wait_time_seconds,
+            ).get("Messages", [])
+            return [(m["ReceiptHandle"], json.loads(m["Body"])) for m in messages]
+        except Exception as e:
+            logging.error(f"Error receiving messages from SQS: {messages=}")
+            raise e
```
hemslo, line 70: When there is exception raised in `receive_message`, `messages` will be undefined, try to access it will raise another `NameError`.

If we just want to capture `json.loads` exceptions, should only put that in try block, and just `raise` to keep stack.

PR #4626, opened 2024-08-19
APPSRE-7516 saas target re-deploy flag

fishi0x01:
This new flag on saas targets determines whether we want to re-deploy the target on config changes in its publishers. (aka setting target_config_hash)

Corresponding schema https://github.com/app-sre/qontract-schemas/pull/705

### DIFF
--- a/reconcile/saas_auto_promotions_manager/publisher.py
+++ b/reconcile/saas_auto_promotions_manager/publisher.py
@@ -39,7 +39,7 @@ class Publisher:
         target_name: str | None,
         cluster_name: str,
         auth_code: HasSecret | None,
-        publish_job_logs: bool | None,
+        re_deploy_on_config_change: bool | None,
         has_subscriber: bool = True,
     ):
         self._ref = ref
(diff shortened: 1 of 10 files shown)


reconcile/saas_auto_promotions_manager/publisher.py
```diff
@@ -39,7 +39,7 @@ def __init__(
         target_name: str | None,
         cluster_name: str,
         auth_code: HasSecret | None,
-        publish_job_logs: bool | None,
+        re_deploy_on_config_change: bool | None,
```
rporres, line 42: nit: since redeploy is actually a word: https://www.merriam-webster.com/dictionary/redeploy

```suggestion
        redeploy_on_config_change: bool | None,
```

rporres (approved):
apart from the name suggestions, looks fine!

reconcile/saas_auto_promotions_manager/publisher.py
fishi0x01: Good point. In that case I should adjust the naming in schema too. On it :+1:

fishi0x01: Fixed everywhere

rporres (approved):

PR #4627, opened 2024-08-19
 expose metric to record oc._run execution time

mehfuz:
As a part of https://issues.redhat.com/browse/APPSRE-10498 .. need to check the number of time `oc._run` calls are made by each integration in order to minimise api calls for get namespaces and use init_projects to make it more efficient


hemslo (approved):
LGTM!

PR #4632, opened 2024-08-23
APPSRE-10782 support multiple cidr blocks routing to TGW

patjlm:
https://issues.redhat.com/browse/APPSRE-10782
depends on https://github.com/app-sre/qontract-schemas/pull/709

We need to support routing multiple CIDR blocks towards a TGW.

We will name route resources with the target CIDR block and not with a simple list index so we can easily modify the list of CIDR blocks without delete/create actions.

Any CIDR block defined in `cidrBlock` will have its `aws_route` resource renamed to match the new pattern. This is done using the terraform `moved` block which is available since terraform 1.1 ([doc](https://developer.hashicorp.com/terraform/language/modules/develop/refactoring#moved-block-syntax)). This allows to be backward compatible and to move the value from `cidrBlock` to `cidrBlocks` after the resource rename has been performed.

### DIFF
--- a/reconcile/terraform_tgw_attachments.py
+++ b/reconcile/terraform_tgw_attachments.py
@@ -64,11 +64,14 @@ class ValidationError(Exception):
     pass
 
 
-class AccountProviderInfo(BaseModel):
+class TGWAccountProviderInfo(BaseModel):
     name: str
     uid: str
     assume_role: str | None
     assume_region: str
+
+
+class ClusterAccountProviderInfo(TGWAccountProviderInfo):
     assume_cidr: str
 
 
@@ -79,8 +82,9 @@ class Requester(BaseModel):
     routes: list[dict] | None
     rules: list[dict] | None
     hostedzones: list[str] | None
-    cidr_block: str
-    account: AccountProviderInfo
+    cidr_block: str | None
+    cidr_blocks: list[str]
+    account: TGWAccountProviderInfo
 
 
 class Accepter(BaseModel):
@@ -89,7 +93,7 @@ class Accepter(BaseModel):
     vpc_id: str | None
     route_table_ids: list[str] | None
     subnets_id_az: list[dict] | None
-    account: AccountProviderInfo
+    account: ClusterAccountProviderInfo
     api_security_group_id: str | None
 
 
@@ -225,7 +229,7 @@ def _build_account_with_assume_role(
     region: str,
     cidr_block: str,
     ocm: OCM | None,
-) -> AccountProviderInfo:
+) -> ClusterAccountProviderInfo:
     account = peer_connection.account
     # assume_role is the role to assume to provision the
     # peering connection request, through the accepter AWS account.
@@ -235,7 +239,7 @@ def _build_account_with_assume_role(
     # there is no OCM at all.
     if not assume_role:
         if isinstance(cluster.spec, ClusterSpecROSAV1) and cluster.spec.account:
-            return AccountProviderInfo(
+            return ClusterAccountProviderInfo(
                 name=cluster.spec.account.name,
                 uid=cluster.spec.account.uid,
                 assume_role=assume_role,
@@ -247,7 +251,7 @@ def _build_account_with_assume_role(
         assume_role = ocm.get_aws_infrastructure_access_terraform_assume_role(
             cluster.name, account.uid, account.terraform_username
         )
-    return AccountProviderInfo(
+    return ClusterAccountProviderInfo(
         name=account.name,
         uid=account.uid,
         assume_role=assume_role,
@@ -258,7 +262,7 @@ def _build_account_with_assume_role(
 
 def _build_accepter(
     peer_connection: ClusterPeeringConnectionAccountTGWV1,
-    account: AccountProviderInfo,
+    account: ClusterAccountProviderInfo,
     region: str,
     cidr_block: str,
     awsapi: AWSApi,
@@ -290,11 +294,10 @@ def _build_requester(
     peer_connection: ClusterPeeringConnectionAccountTGWV1,
     tgw: Mapping,
 ) -> Requester:
-    tgw_account = AccountProviderInfo(
+    tgw_account = TGWAccountProviderInfo(
         name=peer_connection.account.name,
         uid=peer_connection.account.uid,
         assume_region=tgw["region"],
-        assume_cidr=peer_connection.cidr_block,
     )
     return Requester(
         tgw_id=tgw["tgw_id"],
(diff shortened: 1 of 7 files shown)


reconcile/terraform_tgw_attachments.py
```diff
@@ -290,11 +294,10 @@ def _build_requester(
     peer_connection: ClusterPeeringConnectionAccountTGWV1,
     tgw: Mapping,
 ) -> Requester:
-    tgw_account = AccountProviderInfo(
+    tgw_account = TGWAccountProviderInfo(
         name=peer_connection.account.name,
         uid=peer_connection.account.uid,
         assume_region=tgw["region"],
-        assume_cidr=peer_connection.cidr_block,
```
patjlm, line 297: `assume_cidr` is not used for TGW accounts

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1392,17 +1400,32 @@ def populate_tgw_attachments(self, desired_state):
...
-                        "transit_gateway_id": requester.tgw_id,
-                    }
-                    route_identifier = f"{identifier}-{route_table_id}"
-                    tf_resource = aws_route(route_identifier, **values)
-                    self.add_resource(infra_account_name, tf_resource)
+                    if req_cidr_block:
+                        req_cidr_id = req_cidr_block.replace(".", "-").replace("/", "_")
+                        moved = Moved(
+                            fro=f"aws_route.{identifier}-{route_table_id}",
+                            to=f"aws_route.{identifier}-{route_table_id}-dest-{req_cidr_id}",
```
rporres, line 1412: This is so cool!

nit: I would move this block after the add_resource so that it's clearer why you're moving things, very much as the documentation does.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -299,6 +300,13 @@ class UnapprovedSecretPathError(Exception):
     pass
 
 
+class Moved(Block):
+    """Terraform `moved` block, available since Terraform 1.1"""
+
+    def __init__(self, fro: str, to: str):
+        super().__init__(fro=fro, to=to)
```
rporres, line 307: nit: why not?

```suggestion
    def __init__(self, from: str, to: str):
        super().__init__(from=from, to=to)
```

reconcile/utils/terraform_client.py
```diff
@@ -331,12 +331,27 @@ def log_plan_diff(
...
+                            resource_previous_address,
+                            resource_change["address"],
+                        ])
+
                 if action == "no-op":
                     logging.debug([action, name, resource_type, resource_name])
-                    continue
+                    if resource_previous_address:
+                        # apply resource renaming with no-op
+                        self.increment_apply_count()
```
rporres, line 352: Do moved resources get applied each time or `resource_previous_address` is only going to be defined the first time the `moved` block appears and gets applied?

rporres (commented):
looks great. Some nit picking (accept suggestions if you like them) and a question.

reconcile/utils/terraform_client.py
patjlm: I will double check this

reconcile/utils/terrascript_aws_client.py
patjlm: `from` being a python keyword, this seems unfortunately not possible

reconcile/utils/terrascript_aws_client.py
patjlm: will do

reconcile/utils/terrascript_aws_client.py
rporres: very true :(

reconcile/utils/terraform_client.py
patjlm: confirmed in a test with a `null_resource`:

initial `main.tf`, creating a `null_resource.null-1` resource
```
resource "null_resource" "null-1" {}
```

Rename resource to `null-2`
```
resource "null_resource" "null-2" {}

moved {
  from = null_resource.null-1
  to   = null_resource.null-2
}
```

second terraform run, showing `resource_previous_address` in the terraform plan output :
```
$ terraform plan -out plan.tf
null_resource.null-2: Refreshing state... [id=9133010116952026237]

Terraform will perform the following actions:

  # null_resource.null-1 has moved to null_resource.null-2
    resource "null_resource" "null-2" {
        id = "9133010116952026237"
    }

Plan: 0 to add, 0 to change, 0 to destroy.

$ terraform show -json plan.tfplan | jq '.resource_changes'
[
  {
    "address": "null_resource.null-2",
    "previous_address": "null_resource.null-1",
    "mode": "managed",
    "type": "null_resource",
    "name": "null-2",
    "provider_name": "registry.terraform.io/hashicorp/null",
    "change": {
      "actions": [
        "no-op"
      ],
      "before": {
        "id": "9133010116952026237",
        "triggers": null
      },
      "after": {
        "id": "9133010116952026237",
        "triggers": null
      },
      "after_unknown": {},
      "before_sensitive": {},
      "after_sensitive": {}
    }
  }
]

$ terraform apply plan.tf 

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
```

Once applied & renamed, second plan:
```
$ terraform plan -out plan.tfplan
null_resource.null-2: Refreshing state... [id=9133010116952026237]

No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.

$ terraform show -json plan.tfplan | jq '.resource_changes'
[
  {
    "address": "null_resource.null-2",
    "mode": "managed",
    "type": "null_resource",
    "name": "null-2",
    "provider_name": "registry.terraform.io/hashicorp/null",
    "change": {
      "actions": [
        "no-op"
      ],
      "before": {
        "id": "9133010116952026237",
        "triggers": null
      },
      "after": {
        "id": "9133010116952026237",
        "triggers": null
      },
      "after_unknown": {},
      "before_sensitive": {},
      "after_sensitive": {}
    }
  }
]
```

so the `resource_previous_address` field is not there anymore since no move is needed

reconcile/utils/terrascript_aws_client.py
patjlm: done in c0a166eff75fa0939960659d30830c31b50edd14

reconcile/utils/terraform_client.py
rporres: awesome, thx for the detailed test

rporres (approved):

PR #4641, opened 2024-08-29
[change-log] order commits by creation date

maorfr:
part of part of part of https://issues.redhat.com/browse/APPSRE-10755

this PR orders the commits by their order to create a chronological change log.

### DIFF
--- a/reconcile/change_owners/change_log_tracking.py
+++ b/reconcile/change_owners/change_log_tracking.py
@@ -84,6 +84,8 @@ class ChangeLogIntegration(QontractReconcileIntegration[ChangeLogIntegrationPara
                 for i in existing_change_log.items
             ]
         gl = init_gitlab(self.params.gitlab_project_id)
+        if defer:
+            defer(gl.cleanup)
         change_log = ChangeLog()
         for item in diff_state.ls():
             key = item.lstrip("/")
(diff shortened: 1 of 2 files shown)


reconcile/change_owners/change_log_tracking.py
```diff
@@ -25,6 +28,7 @@
 @dataclass
 class ChangeLogItem:
     commit: str
+    created_at: str
```
fishi0x01, line 31: should we use proper timestamps here? string can lead to unforeseen parsing issues long-term.

fishi0x01: E.g., in the past we had issues with string https://github.com/app-sre/qontract-reconcile/pull/4525

maorfr: this is consistent with every other gitlab usage across the code base

reconcile/change_owners/change_log_tracking.py
```diff
@@ -78,6 +83,7 @@ def run(
                 ChangeLogItem(**i)  # type: ignore[arg-type]
                 for i in existing_change_log.items
             ]
+        gl = init_gitlab(self.params.gitlab_project_id)
```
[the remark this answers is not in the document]
maorfr: using gitlab will probably be beneficial quickly. relying on the s3 object date is not stable, since we don't know if we will update the objects out of band in the future. i'll add a cleanup

hemslo (approved):
LGTM!

reconcile/change_owners/change_log_tracking.py
maorfr: example api response: https://gitlab.cee.redhat.com/api/v4/projects/13582/repository/commits/27fbcf9f76e47a98a17ba146fe0f86e6e608a108

PR #4647, opened 2024-09-03
APPSRE-10844 dedicated dtp.spec label

fishi0x01:
Up until now the DTP spec was set via `sre-capabilities.dtp = spec-name`. However, our integrations treat `sre-capabilites.dtp` as a map.

In this PR we set the DTP spec now via `sre-capabilities.dtp.spec = spec-name` to comply with our schema.
Note, that this is fully backwards compatible. If a spec label is missing, DTP falls back to a default spec (which is what all clusters are doing right now).

In this PR we also start printing host level tooling versions for any potential debugging.

### DIFF
--- a/reconcile/dynatrace_token_provider/ocm.py
+++ b/reconcile/dynatrace_token_provider/ocm.py
@@ -34,7 +34,6 @@ from reconcile.utils.ocm_base_client import (
 Thin abstractions of reconcile.ocm module to reduce coupling.
 """
 
-DTP_LABEL = sre_capability_label_key("dtp", None)
 DTP_TENANT_LABEL = sre_capability_label_key("dtp", "tenant")
 DTP_SPEC_LABEL = sre_capability_label_key("dtp", "token-spec")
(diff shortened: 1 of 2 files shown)


reconcile/dynatrace_token_provider/ocm.py
```diff
@@ -50,7 +51,7 @@ class Cluster(BaseModel):
     @staticmethod
     def from_cluster_details(cluster: ClusterDetails) -> Cluster:
         dt_tenant = cluster.labels.get_label_value(DTP_TENANT_LABEL)
-        token_spec_name = cluster.labels.get_label_value(DTP_LABEL)
+        token_spec_name = cluster.labels.get_label_value(DTP_SPEC_LABEL)
```
hemslo, line 54: Is `DTP_LABEL` used anywhere else?

fishi0x01: Good catch! Yeah it should not be needed any longer.

hemslo (approved):
LGTM!

PR #4648, opened 2024-09-03
[change-log-tracking] replace commit created_at with merge request merged_at

maorfr:
we care less when a commit was created, and more when it became active in production (merged).

this PR changes the date to be merged at, from the associated merge request(s).

### DIFF
--- a/reconcile/change_owners/change_log_tracking.py
+++ b/reconcile/change_owners/change_log_tracking.py
@@ -15,6 +15,7 @@ from reconcile.change_owners.changes import aggregate_file_moves, parse_bundle_c
 from reconcile.typed_queries.apps import get_apps
 from reconcile.utils import gql
 from reconcile.utils.defer import defer
+from reconcile.utils.gitlab_api import MRState
 from reconcile.utils.runtime.integration import (
     PydanticRunParams,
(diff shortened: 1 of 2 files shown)


patjlm (approved):

reconcile/change_owners/change_log_tracking.py
```diff
@@ -28,7 +29,7 @@
 @dataclass
 class ChangeLogItem:
     commit: str
-    created_at: str
+    merged_at: str
```
hemslo, line 32: @maorfr this is a breaking change and init from last saved state is failing

```
ChangeLogItem(**i) # type: ignore[arg-type]
^^^^^^^^^^^^^^^^^^
TypeError: ChangeLogItem.__init__() got an unexpected keyword argument 'created_at'
```

why we need to query `merged_at`? `app-interface` build master commit is the merge commit, it's `created_at` is the commit time on master branch, mr `merged_at` timestamp should have less than 1 second diff with that commit time.

maorfr: did you check this assumption?

maorfr: tl;dr not a merge commit ^

PR #4656, opened 2024-09-10
[openshift-tekton-resources] early-exit

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10864

by using early exit we reduce run time of this integration in case the MR does not affect tekton resources.

for example, a standard promotion.

### DIFF
--- a/reconcile/openshift_tekton_resources.py
+++ b/reconcile/openshift_tekton_resources.py
@@ -464,3 +464,7 @@ def run(
         sys.exit(ExitCodes.ERROR)
 
     sys.exit(0)
+
+
+def early_exit_desired_state(*args: Any, **kwargs: Any) -> dict[str, Any]:
+    return fetch_tkn_providers()


reconcile/openshift_tekton_resources.py
```diff
@@ -464,3 +464,7 @@ def run(
         sys.exit(ExitCodes.ERROR)
 
     sys.exit(0)
+
+
+def early_exit_desired_state(*args: Any, **kwargs: Any) -> dict[str, Any]:
+    return fetch_tkn_providers(saas_file_name=None)
```
hemslo, line 470: the order of saas files in dict values may affect equal check in different run.

maorfr: can you explain more?

hemslo: `fetch_tkn_providers(saas_file_name=None)` returns a dict, key is provider name, value is a list of saas files

```python
saas_files = fetch_saas_files(saas_file_name)
```

the order of saas files returned from gql maybe different, lead to desired state not equal to last one even if no saas file changed, may need to change it to `dict[str, set[str]]` to be stable.

maorfr: have we seen this behavior in the past?

hemslo: yes, changed many early exit state from list to set for diff issues

hemslo: dict keys order doesn't matter, just list values.

hemslo: just need to sort at the end

```python
    return {
        provider_name: sorted(sf, key=lambda e: e["name"])
        for provider_name, sf in tkn_providers.items()
        if is_in_shard(provider_name)
    }
```

maorfr: wdyt https://github.com/app-sre/qontract-reconcile/pull/4656/commits/6fad5edf4e5ae746000a74239c0c22b0e023e192?

hemslo: since here we are not using typed query, `saas_files` is added to gql response dynamicly, it should be safe to use `set` here, unless we want to keep it ordered for other purpose.

hemslo: either

```python
    for provider_name in tkn_providers:
        tkn_providers[provider_name]["saas_files"].sort(key=lambda sf: sf["name"])
```

or 

```python
        if "saas_files" not in tkn_providers[provider_name]:
            tkn_providers[provider_name]["saas_files"] = set()

        tkn_providers[provider_name]["saas_files"].add(sf)
```

no need to `sorted` for dict keys.

rporres:
what's a standard promotion in this case? `make qr-promote` triggers this integration to run because of tekton global defaults being changed: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/hack/update-qr-tag.sh#L7

rporres (approved):

PR #4663, opened 2024-09-11
[gitlab-housekeeping] remove duplicate token expiration metrics

cubismod:
When a PAT becomes inactive, do not publish metrics indicating that token is active since that will result in alerting and pages.

The current behavior is that when a token becomes inactive, two gauges are created with `active=true` and `active=false`.


esron (approved):
LGTM

PR #4664, opened 2024-09-12
APPSRE-8094: Test Vector as a Fluentd Replacement

lisa:
In previous instances, Vector would send messages to Slack (and Gchat,
but hereafter referred to as simply "Slack") in a 1:1 ratio to log event
messages. This causes a problem with Slack's API limits causing event
messages to be dropped or otherwise not sent to Slack, which could cause
a logging blindspot. With this commit, Vector will collect event
messages in bundles of up to 20 before sending them to Slack. There's a
caveat, however, with Vector waiting up to 1000ms for messages to
accumulate.

To explain a bit better: If you imagine a bus stop from which a bus with
20 seats and leaves every 1000ms, passengers (messages) can all pile
into the waiting until all the seats are full, or until it's time for
the bus to leave. Sometimes a bus will leave that only has 2 passengers
because the bus waited 1000ms for seats to fill.

See also #4420 and friends for further discussion.


esron (approved):
LGTM

PR #4669, opened 2024-09-15
[change-log] support for resource analysis

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10755

we iterate over changes. if the change is to a resource file, we want to add each backref to the list of changes, as if that file is changing as well.

for example, if a ConfigMap that is referenced from a namespace is updated, we want to express the change in the context of the namespace, to later match all change types that refer to a namespace.

### DIFF
--- a/reconcile/change_owners/changes.py
+++ b/reconcile/change_owners/changes.py
@@ -10,6 +10,7 @@ from dataclasses import (
 from typing import Any
 
 import anymarkup
+import jsonpath_ng
 
 from reconcile.change_owners.bundle import (
     DATAFILE_PATH_FIELD_NAME,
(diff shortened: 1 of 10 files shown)


reconcile/change_owners/changes.py
```diff
@@ -454,3 +457,35 @@ def parse_bundle_changes(
...
+
+
+def aggregate_resource_changes(
+    bundle_changes: list[BundleFileChange],
+    content_store: list[dict[str, Any]],
+) -> list[BundleFileChange]:
+    new_bundle_changes: list[BundleFileChange] = []
+    resource_changes: list[BundleFileChange] = []
+    for change in bundle_changes:
+        new_bundle_changes.append(change)
```
hemslo, line 469: instead of `append` to `new_bundle_changes` on every `bundle_changes`, can directly return `bundle_changes + resource_changes` as the final result at the end.

reconcile/change_owners/changes.py
```diff
@@ -454,3 +457,35 @@ def parse_bundle_changes(
...
+        for file_ref in change.old_backrefs | change.new_backrefs:
+            file_content = next(c for c in content_store if c["path"] == file_ref.path)
```
hemslo, line 471: to speed up `file_content` lookup, can pre build a lookup dict

```python
file_content_by_path = {c["path"]: c for c in content_store}
```

then the final logic can be like

```python
    file_content_by_path = {c["path"]: c for c in content_store}
    resource_changes = [
        BundleFileChange(
            fileref=change.fileref,
            old=file_content,
            new=file_content,
            old_content_sha="",
            new_content_sha="",
            diffs=[
                Diff(
                    path=jsonpath_ng.parse(file_ref.json_path),
                    diff_type=DiffType.CHANGED,
                    old=file_content,
                    new=file_content,
                )
            ],
        )
        for change in bundle_changes
        for file_ref in change.old_backrefs | change.new_backrefs
        if (file_content := file_content_by_path.get(file_ref.path))
    ]
    return bundle_changes + resource_changes
```

hemslo (approved):
LGTM! Please also update PR description to cover jenkins config support.

reconcile/change_owners/changes.py
maorfr: i was being consistent with above methods

reconcile/change_owners/changes.py
maorfr: i wanted to remain ordered

PR #4670, opened 2024-09-16
External Resources metrics redesign

lechuk47:
- **Ignore mypy errors. This will be enventually solved once all resources have the managed_by_erv2 attr.**
- **Add a method to get the job duration time for a successful run**
- **External resources metrics redesign**
  - reconcile_errors moved out of the state. I t can be handled with prometheus metrics directly
  - add reconcile_time

### DIFF
--- a/reconcile/external_resources/manager.py
+++ b/reconcile/external_resources/manager.py
@@ -13,7 +13,10 @@ from reconcile.external_resources.factories import (
     TerraformModuleProvisionDataFactory,
     setup_aws_resource_factories,
 )
-from reconcile.external_resources.metrics import ExternalResourcesReconcileErrorsGauge
+from reconcile.external_resources.metrics import (
+    ExternalResourcesReconcileErrorsCounter,
+    ExternalResourcesReconcileTimeGauge,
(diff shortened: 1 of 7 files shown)


reconcile/external_resources/manager.py
```diff
@@ -242,12 +249,22 @@ def _update_in_progress_state(
...
+
+                metrics.set_gauge(
+                    ExternalResourcesReconcileTimeGauge(
+                        provision_provider=r.key.provision_provider,
+                        provisioner_name=r.key.provisioner_name,
+                        provider=r.key.provider,
+                        identifier=r.key.identifier,
+                        job_name=job_name,
+                    ),
+                    float(self.reconciler.get_resource_reconcile_duration(r) or 0),
```
hemslo, line 266: is convert to `float` required here? the actual result is `int`.

hemslo (approved):
LGTM!

reconcile/external_resources/manager.py
lechuk47: nope. Fixed.

PR #4676, opened 2024-09-20
[ldap-groups] set request timeout to avoid integration hang

hemslo:
`InternalGroupsApi` sending request without timeout, lead to integrations like `ldap-groups` hang, add `timeout` to `_request` so it can self recover.

[ASIC-549](https://issues.redhat.com/browse/ASIC-549)

### DIFF
--- a/reconcile/utils/internal_groups/client.py
+++ b/reconcile/utils/internal_groups/client.py
@@ -14,6 +14,8 @@ from sretoolbox.utils import retry
 
 from reconcile.utils.internal_groups.models import Group
 
+REQUEST_TIMEOUT = 30
+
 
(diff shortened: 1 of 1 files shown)


mehfuz (approved):

reconcile/utils/internal_groups/client.py
```diff
@@ -14,6 +14,8 @@
 
 from reconcile.utils.internal_groups.models import Group
 
+REQUEST_TIMEOUT = 30
```
rporres, line 17: nit: maybe worth to assign it to an env variable.

rporres (approved):
LGTM

just a nit if you want to consider it :)

PR #4677, opened 2024-09-20
APPSRE-10884 always checkout chart from ref

fishi0x01:
Some tenants would like to checkout the helm chart from the ref in the saas file. I believe this makes sense.

Are there scenarios in which we always want to checkout from main? Do we need an extra saas helm provider flag (`checkoutChartFromRef`?) or can we make this general behavior?


maorfr (approved):
LGTM!

PR #4678, opened 2024-09-20
better error message on missing devtools-bot access

fishi0x01:
Log a message that might solve missing devtools-bot access issues


reconcile/gitlab_permissions.py
```diff
@@ -78,8 +78,14 @@ def extract_group_spec(self, project: GroupProject) -> GroupSpec:
...
-        user = project.members_all.get(id=self.gl.user.id)
-        return user.access_level >= self.access_level
+        try:
+            user = project.members_all.get(id=self.gl.user.id)
+            return user.access_level >= self.access_level
+        except Exception as e:
+            logging.error(
+                f"Could not fetch member from repository {project.attributes.get('web_url', 'WEB_URL_NOT_SET')} - did you forget to make our devtools-bot a maintainer of that repo?"
+            )
+            raise e
```
hemslo, line 88: In which case the request can fail? `self.gl.user.id` not exist in `project.members_all`?

no access case is handled outside as

```python
                    GroupAccessLevelError(
                        f"{repo} is not shared with {self.gl.user.username} as {self.access_level_string}"
                    )
```

so can be treated as no access case here

```python
try: 
    user = project.members_all.get(id=self.gl.user.id)
except gitlab.exceptions.GitlabGetError:
    logging.error()
    return False
return user.access_level >= self.access_level
```

cc @mehfuz

fishi0x01: Good point. Yeah lets rely on `GroupAccessLevelError`. Fixed

hemslo (approved):
LGTM!

rporres (approved):

PR #4679, opened 2024-09-23
[change-log] detect app in status-page component

maorfr:
https://github.com/app-sre/qontract-schemas/blob/d4349be301f929d5bd24f5c60efe52dcbd196017/schemas/dependencies/status-page-component-1.yml#L14


hemslo (approved):
LGTM!

PR #4680, opened 2024-09-23
[change-log] fix docs/hack section indentation

maorfr:
to not depend on `changes`, and to be able to parse a commit that only has changes to docs.


hemslo (approved):
LGTM!

PR #4682, opened 2024-09-23
[change-log] resource backref support for specific schemas

maorfr:
most backrefs are namespaces and jenkins configs. these are "functional" backrefs.

some backrefs are roles, where resources are referenced under the self_service section. these are "non-functional" backrefs. why? because a change to a ConfigMap does not indicate anything about the roles that may own that ConfigMap.

this PR adds a list of supported schemas, so we are future proof.

### DIFF
--- a/reconcile/change_owners/change_log_tracking.py
+++ b/reconcile/change_owners/change_log_tracking.py
@@ -134,8 +134,12 @@ class ChangeLogIntegration(QontractReconcileIntegration[ChangeLogIntegrationPara
                 continue
             diff = QontractServerDiff(**obj)
             changes = aggregate_resource_changes(
-                aggregate_file_moves(parse_bundle_changes(diff)),
-                {c.path: c.dict() for c in namespaces + jenkins_configs},
+                bundle_changes=aggregate_file_moves(parse_bundle_changes(diff)),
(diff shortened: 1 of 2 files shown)


reconcile/change_owners/change_log_tracking.py
```diff
@@ -134,8 +134,12 @@ def run(
...
             diff = QontractServerDiff(**obj)
             changes = aggregate_resource_changes(
-                aggregate_file_moves(parse_bundle_changes(diff)),
-                {c.path: c.dict() for c in namespaces + jenkins_configs},
+                bundle_changes=aggregate_file_moves(parse_bundle_changes(diff)),
+                content_store={c.path: c.dict() for c in namespaces + jenkins_configs},
+                supported_schemas={
+                    "/openshift/namespace-1.yml",
+                    "/dependencies/jenkins-config-1.yml",
+                },
```
hemslo, line 142: can we extract this to a const and doc what schema should be supported and what shouldn't?

hemslo (approved):
LGTM!

reconcile/change_owners/change_log_tracking.py
maorfr: i'll follow up in the future

PR #4685, opened 2024-09-23
APPSRE-10489:reduce the number of api calls for namespace check

mehfuz:
The PR contains code which reduces the check namespace calls made by integrations in dry and non dry modes. This is an effort to improve the execution time of the integration and avoid additional calls to get a resource from openshift api.

The metric `tekton_pipelines_controller_pipelinerun_taskrun_duration_seconds` can be leveraged to check the actual difference in execution time.


hemslo (approved):
LGTM!

PR #4690, opened 2024-09-24
[saasherder/helm] support ref

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10884

another attempt at #4677

this time with `git fetch` as well.

### DIFF
--- a/reconcile/utils/helm.py
+++ b/reconcile/utils/helm.py
@@ -115,7 +115,6 @@ def template_all(
 ) -> Iterable[Mapping[str, Any]]:
     with tempfile.TemporaryDirectory() as wd:
         git.clone(url, wd, depth=1, verify=ssl_verify)
-        git.fetch(ref, wd)
         git.checkout(ref, wd)
         return yaml.safe_load_all(
             do_template(values=values, path=f"{wd}{path}", namespace=namespace)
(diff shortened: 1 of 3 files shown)


reconcile/utils/helm.py
```diff
@@ -108,12 +108,15 @@ def template(
...
     path: str,
+    ref: str,
     namespace: str,
     values: Mapping[str, Any],
     ssl_verify: bool = True,
 ) -> Iterable[Mapping[str, Any]]:
     with tempfile.TemporaryDirectory() as wd:
         git.clone(url, wd, depth=1, verify=ssl_verify)
+        git.fetch(ref, wd)
+        git.checkout(ref, wd)
```
hemslo, line 119: if we always need `fetch & checkout`, then we can skip checkout during clone to save time


```shell
git clone --no-checkout --depth 1 url
git fetch --depth 1 origin sha
git checkout sha
```

In fact, the implicit fetch during clone is wasted since we only care about one commit, the most efficient way to fetch it is:


```shell
git init repo
cd repo
git remote add origin url
git fetch --depth 1 origin sha
git checkout sha
```

maorfr: went with a conditional fetch, and also implemented fetch depth.

hemslo (approved):
LGTM!

hemslo (approved):

PR #4692, opened 2024-09-25
remove git secrets

maorfr:


rporres:
Do you happen to know how was this used and why we don't need it anymore? It would be a great addition to the description of this patch :)

rporres (approved):

PR #4693, opened 2024-09-25
Ensure a status page component is set before creating a maintenance

patjlm:
If there are no components, status page replies with a `500`

Also added a commit to sort current and desired states so we can diff the 2 lists securely

### DIFF
--- a/reconcile/statuspage/integrations/maintenances.py
+++ b/reconcile/statuspage/integrations/maintenances.py
@@ -80,6 +80,7 @@ class StatusPageMaintenancesIntegration(QontractReconcileIntegration[NoParams]):
                     for m in p.maintenances or []
                     if datetime.fromisoformat(m.scheduled_start) > now
                 ]
+                desired_state = sorted(desired_state, key=lambda d: d.name)
                 page_provider = AtlassianStatusPageProvider.init_from_page(
                     page=p,
                     token=self.secret_reader.read_secret(p.credentials),
@@ -90,6 +91,7 @@ class StatusPageMaintenancesIntegration(QontractReconcileIntegration[NoParams]):
                     for m in page_provider.scheduled_maintenances
                     if page_provider.has_component_binding_for(m.name)
                 ]
+                current_state = sorted(current_state, key=lambda c: c.name)
                 self.reconcile(
                     dry_run=dry_run,
                     desired_state=desired_state,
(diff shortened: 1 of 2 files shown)


reconcile/statuspage/integrations/maintenances.py
```diff
@@ -33,8 +33,18 @@ def reconcile(
         diff = diff_iterables(
             current=current_state, desired=desired_state, key=lambda x: x.name
         )
+        errors = False
         for a in diff.add.values():
-            logging.info(f"Create StatusPage Maintenance: {a.name}")
+            if not a.components:
```
hemslo, line 38: no `components` case is checked [here](https://github.com/app-sre/qontract-reconcile/blob/a18009361b26a8b1fe5b07b8a93419f79bc87136/reconcile/statuspage/page.py#L157), can we validate it in the same place?

patjlm: Updated. Now if there are no attached component, we cannot set a status page maintenance (which makes sense IMHO)

reconcile/statuspage/integrations/maintenances.py
```diff
@@ -80,6 +80,7 @@ def run(self, dry_run: bool = False) -> None:
                     for m in p.maintenances or []
                     if datetime.fromisoformat(m.scheduled_start) > now
                 ]
+                desired_state = sorted(desired_state, key=lambda d: d.name)
```
hemslo, line 83: is sorting required in some case?

hemslo (approved):
LGTM!

reconcile/statuspage/integrations/maintenances.py
hemslo: `name` need to be unique, otherwise diff result may miss things (last one wins)

https://github.com/app-sre/qontract-reconcile/blob/60c897ef52b21ae05ec84096e431489fd559a90b/reconcile/statuspage/integrations/maintenances.py#L34

patjlm: > is sorting required in some case?

yes, I saw the case while debugging. desired and current state had reverse orders

> name need to be unique

yes, shoudl be good as per schema: https://github.com/app-sre/qontract-schemas/blob/main/graphql-schemas/schema.yml#L4358

hemslo: if it's unique, then the order doesn't matter, as they will be converted to dict by key during diff

PR #4695, opened 2024-09-27
[git-housekeeping] update token expiration metric to only include active tokens

cubismod:

Updates the `qontract_reconcile_gitlab_token_expiration_days` metric exported by `gitlab-housekeeping` to only factor in active tokens. This also removes the `active` label from the metric which should cut down on unnecessary cardinality. There's no real need to monitor inactive tokens since they cannot be used.

Additionally updates the Grafana dashboard for integrations to remove that `active` label.

Second take at https://github.com/app-sre/qontract-reconcile/pull/4663


esron (approved):
Nice description.
LGTM.

PR #4696, opened 2024-09-30
Erv2: Improve Logs dashboard + add a new metric

lechuk47:
* Add external_resources_resource_status metric to hold resource status. This metric is useful for filtering since it has a value at every run (not only when a resource has been reconciled)
* Use the new metric on the dashboard filtering options


hemslo (approved):
LGTM!

PR #4701, opened 2024-09-30
[change-log] defense against missing app and helpful logging

maorfr:
protect against missing `app` or `app.path` in the query, and be more defensive if that information is missing.

also provide helpful logging.

### DIFF
--- a/reconcile/change_owners/change_log_tracking.py
+++ b/reconcile/change_owners/change_log_tracking.py
@@ -46,6 +46,14 @@ class ChangeLogItem:
 class ChangeLog:
     items: list[ChangeLogItem] = field(default_factory=list)
 
+    @property
+    def apps(self) -> set[str]:
+        return {app for item in self.items for app in item.apps}
+
+    @property
+    def change_types(self) -> set[str]:
+        return {change_type for item in self.items for change_type in item.change_types}
+
 
 class ChangeLogIntegrationParams(PydanticRunParams):
     gitlab_project_id: str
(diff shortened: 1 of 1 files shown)


reconcile/change_owners/change_log_tracking.py
```diff
@@ -160,14 +168,19 @@ def run(
...
-                            and (app_path := app.get("$ref") or app.get("path"))
-                            and (name := app_name_by_path.get(app_path))
-                        }
-                        change_log_item.apps.extend(changed_apps)
+                        for c in change_versions:
+                            c_app: dict[str, str] = c["app"]
+                            app_path = c_app.get("$ref") or c_app.get("path")
+                            if not app_path:
+                                raise KeyError(
+                                    "app path is expected. missing in query?"
```
hemslo, line 176: should include more meta info for debug, like which `app` is this

reconcile/change_owners/change_log_tracking.py
```diff
@@ -160,14 +168,19 @@ def run(
...
+                                )
+                            app_name = app_name_by_path.get(app_path)
+                            if not app_name:
+                                raise KeyError(
+                                    "app name is expected. missing in query?"
```
hemslo, line 181: the reason for this error should be path not found in apps, should also include path value for debug

maorfr: https://github.com/app-sre/qontract-reconcile/pull/4701/commits/a758b2993f948706d4c3df7d95c0c9b084131bfe

reconcile/change_owners/change_log_tracking.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/4701/commits/a758b2993f948706d4c3df7d95c0c9b084131bfe

PR #4702, opened 2024-09-30
[gitlab-housekeeping] remove expired PAT tokens

cubismod:

Follow-up from #4695 

I found out how this `remove()` function is supposed to work (it's not actually documented by the Prometheus client library). This should take care of removing inactive/expired tokens from the exported metrics.


bkez322:
ty python promothesus library

LGTM

bkez322 (approved):

PR #4706, opened 2024-10-01
:bug: endpoints-discovery: avoid duplicated urls

chassing:
Several routes may have the same hostname, e.g., if `path` is used. Remove duplicates and unclutter the `app.yml` files 

Ticket: https://issues.redhat.com/browse/APPSRE-10847

### DIFF
--- a/reconcile/endpoints_discovery/integration.py
+++ b/reconcile/endpoints_discovery/integration.py
@@ -1,4 +1,5 @@
 import logging
+from collections import defaultdict
 from collections.abc import Callable, Iterable
 from typing import TypedDict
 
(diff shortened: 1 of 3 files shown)


reconcile/endpoints_discovery/integration.py
```diff
@@ -137,14 +137,21 @@ def get_routes(self, oc_map: OCMap, namespace: NamespaceV1) -> list[Route]:
...
             return []
 
-        return [
-            Route(
+        routes = {}
+        for item in sorted(
+            oc.get_items(kind="Route", namespace=namespace.name),
+            key=lambda x: x["metadata"]["name"],
+            reverse=True,
+        ):
```
hemslo, line 145: why `sorted` here? How about aggregate all route names so we can trace back source


```python
        routes = defaultdict(list)
        for item in oc.get_items(kind="Route", namespace=namespace.name):
            tls = bool(item["spec"].get("tls"))
            host = item["spec"]["host"]
            routes[(host, tls)].append(item["metadata"]["name"])

        return [
            Route(
                name="|".join(sorted(names)),
                host=host,
                tls=tls,
            )
            for (host, tls), names in routes.items()
        ]
```

another edge case is same host reused in multiple namespaces, deduplication need to be done in app level.

chassing: Your suggestion would be tricky to implement because the namespaces drive everything. Having a host pointing to different namespaces is an edge case we can neglect.

hemslo: ok, then at least keep all route names in final result for visibility, to know the discovered endpoint is from which routes.

chassing: Applied in 60ff049

chassing: Done in 5b91b7a

hemslo (approved):
LGTM!

PR #4707, opened 2024-10-01
:recycle: aws-account-manager: do not raise an error

chassing:
If the AWS account manager template MR has already been merged but the `template-renderer` isn't running yet, the account template collection file already exists, which throws an error message. Just remove this confusing message.

Ticket: [APPSRE-10814](https://issues.redhat.com/browse/APPSRE-10814)

### DIFF
--- a/reconcile/aws_account_manager/merge_request_manager.py
+++ b/reconcile/aws_account_manager/merge_request_manager.py
@@ -89,6 +89,10 @@ class MergeRequestManager:
                 file_path=account_tmpl_file_path
             )
             # File already exists. nothing to do.
+            logging.debug(
+                "The template collection file %s already exists. This may happen if the MR has been merged but template-renderer isn't running yet.",
+                account_tmpl_file_path,
+            )
             return
         except GitlabGetError as e:
             if e.response_code != 404:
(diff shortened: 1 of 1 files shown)


reconcile/aws_account_manager/merge_request_manager.py
```diff
@@ -88,10 +88,8 @@ def create_account_file(
             self._vcs.get_file_content_from_app_interface_master(
                 file_path=account_tmpl_file_path
             )
-            # File already exists
-            raise FileExistsError(
-                f"File {account_tmpl_file_path} already exists in the repository"
-            )
+            # File already exists. nothing to do.
```
rporres, line 91: nit: I think it wouldn't hurt to log an debu/info explanatory message. Empty returns are always hard to analyze for the non-initiated, I've been there a few times

rporres (approved):
just your typical nit-picking :)

reconcile/aws_account_manager/merge_request_manager.py
chassing: You're right. Applied in [22f6945](https://github.com/app-sre/qontract-reconcile/pull/4707/commits/22f69452ce88af1b45dd6aa39a8d67e33372d9f7)

PR #4712, opened 2024-10-03
fix dashdotdb-slo failures when the detected SLO is not a number

patjlm:
https://issues.redhat.com/browse/APPSRE-10974

Fixes
```
requests.exceptions.InvalidJSONError: Out of range float values are not JSON compliant
```

NaN can happen when the SLO query does a divide by zero. For example:
```
sum(fci_successful_ops_total) / sum(fci_processed_ops_total)
```
and there are no `fci_processed_ops_total`...

### DIFF
--- a/reconcile/dashdotdb_slo.py
+++ b/reconcile/dashdotdb_slo.py
@@ -83,20 +83,14 @@ class DashdotdbSLO(DashdotdbBase):
             if self.dry_run:
                 continue
 
-            if isnan(item.value):
-                LOG.info(
-                    f"{self.logmarker} Skipping SLO '{slo_name}' in SLO doc '{item.slo_doc_name}'"
-                    "as the obtained value is not a number"
-                )
-                continue
-
             LOG.info("%s syncing slo %s", self.logmarker, slo_name)
             try:
                 response = self._do_post(endpoint, payload)
                 response.raise_for_status()
-            except requests.exceptions.HTTPError as details:
-                LOG.error("%s error posting %s - %s", self.logmarker, slo_name, details)
-            except requests.exceptions.InvalidJSONError as details:
+            except (
+                requests.exceptions.HTTPError,
+                requests.exceptions.InvalidJSONError,
+            ) as details:
                 LOG.error("%s error posting %s - %s", self.logmarker, slo_name, details)
 
             LOG.info("%s slo %s synced", self.logmarker, slo_name)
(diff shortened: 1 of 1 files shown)


reconcile/dashdotdb_slo.py
```diff
@@ -82,12 +83,21 @@ def _post(self, service_slos: Iterable[ServiceSLO]) -> Response | None:
             if self.dry_run:
                 continue
 
+            if isnan(item.value):
+                LOG.info(
+                    f"{self.logmarker} Skipping SLO '{slo_name}' in SLO doc '{item.slo_doc_name}'"
+                    "as the obtained value is not a number"
+                )
```
hemslo, line 90: we can even capture that in the beginning to detect it in dry run https://github.com/app-sre/qontract-reconcile/blob/3a0cbc38491fa1a9fabb72524fbaf933b37e40a3/reconcile/dashdotdb_slo.py#L142-L146

reconcile/dashdotdb_slo.py
```diff
@@ -82,12 +83,21 @@ def _post(self, service_slos: Iterable[ServiceSLO]) -> Response | None:
...
+
             LOG.info("%s syncing slo %s", self.logmarker, slo_name)
-            response = self._do_post(endpoint, payload)
             try:
+                response = self._do_post(endpoint, payload)
                 response.raise_for_status()
             except requests.exceptions.HTTPError as details:
                 LOG.error("%s error posting %s - %s", self.logmarker, slo_name, details)
+            except requests.exceptions.InvalidJSONError as details:
+                LOG.error("%s error posting %s - %s", self.logmarker, slo_name, details)
```
hemslo, line 100: error handling logic are the same, can merge with previous line

```python
except (requests.exceptions.HTTPError, requests.exceptions.InvalidJSONError) as details:

```

hemslo (approved):
LGTM!

reconcile/dashdotdb_slo.py
patjlm: Ok, but i think we should still not fail

rporres: +1

rporres (approved):

reconcile/dashdotdb_slo.py
patjlm: Fixed in https://github.com/app-sre/qontract-reconcile/commit/dbc08b35e413f2de715929237f2f433bd2cfd884

reconcile/dashdotdb_slo.py
patjlm: Fixed in https://github.com/app-sre/qontract-reconcile/commit/dbc08b35e413f2de715929237f2f433bd2cfd884

PR #4713, opened 2024-10-03
[dashdotdb-slo] query the prometheus instance configured in prometheusAccess, even when SLONamespace is set

patjlm:
https://issues.redhat.com/browse/APPSRE-8513

Depends on https://github.com/app-sre/qontract-schemas/pull/718

This allows to get SLO data into dash.db for SLO queries which are not targeting the local cluster prometheus.

Also do not fail on connection errors to prometheus and allow unauthenticated access (eg local token refresher services)

### DIFF
--- a/reconcile/dashdotdb_slo.py
+++ b/reconcile/dashdotdb_slo.py
@@ -145,13 +145,13 @@ class DashdotdbSLO(DashdotdbBase):
                         password=password,
                     )
                 except requests.exceptions.ConnectionError as error:
-                    # This can happen when prometheus is unreachble, or when running locally
+                    # This can happen when prometheus is unreachable, or when running locally
                     # and some prometheus URL are openshift service names. The trick is to run
                     # with `oc port-forward` and update the local hosts file if we need to query those.
                     LOG.error(
                         f"{self.logmarker} Could not reach prometheus at {promurl}: {error}. Skipping {slo.name}"
                     )
-                    continue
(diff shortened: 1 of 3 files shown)


reconcile/dashdotdb_slo.py
```diff
@@ -132,13 +136,22 @@ def _get_service_slo(self, slo_document: SLODocumentV1) -> list[ServiceSLO]:
...
+                        password=password,
+                    )
+                except requests.exceptions.ConnectionError as error:
+                    # This can happen when prometheus is unreachble, or when running locally
+                    # and some prometheus URL are openshift service names. The trick is to run
+                    # with `oc port-forward` and update the local hosts file if we need to query those.
+                    LOG.error(
+                        f"{self.logmarker} Could not reach prometheus at {promurl}: {error}. Skipping {slo.name}"
+                    )
+                    continue
```
rporres, line 154: I don't like this. We are not going to know when things are breaking since no one is watching logs. And this is not how our integrations work. For better or worse, a break in a single element always make the whole integration fail. We should not introduce new patterns unless we can alert on them.

patjlm: agreed, fixed in https://github.com/app-sre/qontract-reconcile/pull/4713/commits/6b409c8eac0feb11c17b28eab8bbd1ed539a7235

rporres (approved):

PR #4714, opened 2024-10-04
[copy_tfstate] add a check for overwriting statefiles

cubismod:
[APPSRE-10882](https://issues.redhat.com/browse/APPSRE-10882)

Improves the command by adding a check to see if the destination object already exists and quit if it does unless the user passes a `--force` flag.


jfchevrette (approved):
LGTM!

PR #4716, opened 2024-10-07
add typed queries for schemas

maorfr:
to be used for something for sure


fishi0x01:
Looks like a bug in `qenerate`. Seems it doesnt handle the `_` prefix well. 

```
reconcile/gql_definitions/common/schemas.py:50: error: Name "TypeKind" is not defined  [name-defined]
```

PR #4717, opened 2024-10-07
Parameterize ssl policy for alb

esron:
[APPSRE-10769](https://issues.redhat.com/browse/APPSRE-10769)

Schema update https://github.com/app-sre/qontract-schemas/pull/719

Using ELBSecurityPolicy-TLS13-1-0-2021-06 as default since it's the one that supports all TLS versions.

### DIFF
--- a/reconcile/gql_definitions/introspection.json
+++ b/reconcile/gql_definitions/introspection.json
@@ -45358,6 +45358,18 @@
                             "deprecationReason": null
                         },
                         {
+                            "name": "ssl_policy",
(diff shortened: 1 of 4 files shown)


cubismod (approved):
LGTM, reviewing https://docs.aws.amazon.com/elasticloadbalancing/latest/application/describe-ssl-policies.html, looks like the bump in SSL policy for existing load balancers shouldn't cause any backwards incompatibilities and it's also just Quay using this feature right now

PR #4718, opened 2024-10-08
Erv2: publish inventory metrics

lechuk47:
Erv2: publish external resources inventory metrics 

APPSRE-10975

### DIFF
--- a/reconcile/utils/external_resources.py
+++ b/reconcile/utils/external_resources.py
@@ -8,6 +8,7 @@ from typing import Any
 
 import anymarkup
 
+from reconcile.external_resources.model import ExternalResourcesInventory
 from reconcile.utils import (
     gql,
     metrics,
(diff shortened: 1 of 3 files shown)


reconcile/utils/external_resources.py
```diff
@@ -66,14 +65,14 @@ def managed_external_resources(namespace_info: Mapping[str, Any]) -> bool:
 
 
 def get_inventory_count_combinations(
-    inventory: ExternalResourceSpecInventory,
+    inventory: MutableMapping,
```
fishi0x01, line 68: nit: why mutable is needed? It seems we do not expect any mutation here and `Mapping` might suffice for this case?

fishi0x01 (approved):

reconcile/utils/external_resources.py
lechuk47: You are right. I used the parent class of the two classes used with this method, but is not necessary to use Mutable in this case. Thanks for catching it.

cubismod (approved):

reconcile/utils/external_resources.py
hemslo: `Mapping` is too generic as we need to ensure key type match to use `(k.provision_provider, k.provisioner_name, k.provider) for k in inventory`, can we extract a common interface for it?

lechuk47: I Switched it to a inventories static list.

hemslo (approved):
LGTM!

PR #4719, opened 2024-10-09
[saasherder] demo helm capabilities

maorfr:
this PR shows a way to incorporate cluster information into the templating process. in this example, replacing the API versions helm Capability.


reconcile/utils/saasherder/saasherder.py
```diff
@@ -982,6 +983,7 @@ def _process_template(self, spec: TargetSpec) -> tuple[list[Any], Promotion | No
                 ref=ref,
                 namespace=spec.target.namespace.name,
                 values=consolidated_parameters,
+                api_versions=spec.oc.api_resources or [],
```
hemslo, line 986: need to do convert `api_resources` (`dict[str, list[OCCliApiResource]]`) to deduplicated `list[str]`

PR #4720, opened 2024-10-09
qontract-cli: add rds-recommendations command

jfchevrette:
Generate a report for RDS Recommendations

### DIFF
--- a/tools/qontract_cli.py
+++ b/tools/qontract_cli.py
@@ -1749,6 +1749,85 @@ You can view the source of this Markdown to extract the JSON data.
         print_output(ctx.obj["options"], results, columns)
 
 
+@get.command
+@click.pass_context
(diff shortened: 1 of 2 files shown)


tools/qontract_cli.py
```diff
@@ -8,11 +8,7 @@
 import re
 import sys
 from collections import defaultdict
-from datetime import (
```
cubismod, line 11: Can you check your `ruff` config? Looks like a bunch of imports got re-sorted around

jfchevrette: ugh my ruff version indeed got updated. I've reverted the non-relevant changes manually

cubismod: Thanks!

cubismod (approved):

PR #4722, opened 2024-10-11
Allow the replication of single secret policies

rporres:
we may have policies with paths that do not have globs, being just a single secret:
```
path my/nice/secret
```
instead of the more general that catches everything inside the folder
```
path my/nice/*
```
The code didn't behave properly with the former ones.

related to APPSRE-10999

### DIFF
--- a/reconcile/vault_replication.py
+++ b/reconcile/vault_replication.py
@@ -34,6 +34,7 @@ from reconcile.utils.vault import (
 )
 
 QONTRACT_INTEGRATION = "vault-replication"
+SECRET_PATH_PATTERN = re.compile(r"^[\w/-]+?(?P<folder>/\*?)?$")
 
 
 class VaultInvalidPaths(Exception):
@@ -232,14 +233,21 @@ def get_policy_secret_list(
     vault_instance: _VaultClient, policy_paths: Iterable[str]
 ) -> list[str]:
     """Returns a list of secrets to be copied from the given policy"""
-    secret_list = []
+    secrets = set()
     for path in policy_paths:
-        # Remove the * at the end of the path because list method expects
-        # a folder path without any secret or wilcard
-        path = path[:-1] if path.endswith("*") else path
-        secret_list.extend(vault_instance.list_all(path))
+        match = SECRET_PATH_PATTERN.match(path)
+        if not match:
+            logging.error(["get_policy_secret_list", "Invalid path to replicate", path])
+            raise VaultInvalidPaths
 
-    return secret_list
+        if match.group("folder"):
+            # Remove the * at the end of the path because list method expects
+            # a folder path without any secret or wilcard
+            secrets.update(vault_instance.list_all(path.rstrip("*")))
+        else:
+            secrets.add(path)
+
+    return list(secrets)
 
 
 def get_jenkins_secret_list(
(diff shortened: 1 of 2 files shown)


reconcile/vault_replication.py
```diff
@@ -237,7 +237,10 @@ def get_policy_secret_list(
         # Remove the * at the end of the path because list method expects
         # a folder path without any secret or wilcard
         path = path[:-1] if path.endswith("*") else path
-        secret_list.extend(vault_instance.list_all(path))
+        if path.endswith("/"):
+            secret_list.extend(vault_instance.list_all(path))
+        else:
+            secret_list.extend([path])
```
patjlm, line 243: or `.append(path)`?

patjlm (approved):

reconcile/vault_replication.py
hemslo: according comments above about `*`, we have 2 valid cases for `path`, either ends with `/*` or `/` as dir, or a single secret. Invalid case is like `policy/path/3/secret1*`


```python
if path.endswith("/*") or path.endswith("/"):
    secret_list.extend(vault_instance.list_all(path.rstrip("*")))
elif path.endswith("*"):
    raise ...
else:
    secret_list.append(path)
```

rporres: Taking a look into the policies [documentation](https://developer.hashicorp.com/vault/docs/concepts/policies), the situation is a bit worse, since we can only properly support to replicate paths like

* path/to/file
* path/to/folder/*

but there are more combinations with `*` and `+`

I'll rework a bit the whole thing so that we fail properly.

rporres: reworked in f36b21b763eb331df0298aacf6db744cb107279b

reconcile/utils/oc_map.py
```diff
@@ -122,7 +122,8 @@ def _init_oc_client(
                     connection_parameters=connection_parameters
                 )
             try:
-                oc_client: OCCli | OCLogMsg = self._oc_cls(
+                # TODO: fix ignore type
+                oc_client: OCCli | OCLogMsg = self._oc_cls(  # type: ignore
```
rporres, line 126: I honestly don't know why this is failing now... ignoring it for the moment, this should be addressed in a separate PR.

reconcile/vault_replication.py
hemslo: that regex has an issue `/-_` means matches a single character in the range between / (index 47) and _ (index 95) (case sensitive). So this can match `policy/path/3;/secret1` I think what you mean is `^[A-Za-z0-9/\-_]+(?:/\*)?$` or just `^[\w/-]+(?:/\*)?$`, you can debug it at https://regex101.com/r/pp23Tm/1.

Another improvement is to move `re.compile` to a module const to avoid recompile every time the function got called.

And since we are already using regex to match the path, we can use match result directly to avoid recheck ends with `/*` or `/`, note we need non greedy for the first part, otherwise the last `/` will be captured by first part. And named capture for readability. Like this:

```python

SECRET_PATH_PATTERN = re.compile(r"^[\w/-]+?(?P<folder>/\*?)?$")


match = SECRET_PATH_PATTERN.match(path)
if not match:
  raise

if match.group("folder"):
  secrets.update(vault_instance.list_all(path.rstrip("*")))
else:
  secrets.add(path)
```

reconcile/utils/oc_map.py
rporres: related to https://github.com/app-sre/qontract-reconcile/pull/4728

reconcile/vault_replication.py
rporres: nice catch! Fixed in 533875ee0dc3eb459c008d164973231dc9696c0c

PR #4725, opened 2024-10-14
Erv2: Improve metrics for alerting + some refactors

lechuk47:
* Add app + environment to metrics. This will enable an improved alert routing 
* Add ReconciliationStatus class to handle reconciliation metrics and flags 
* Some manager refactorings to make the code cleaner.

### DIFF
--- a/reconcile/external_resources/manager.py
+++ b/reconcile/external_resources/manager.py
@@ -3,6 +3,7 @@ from collections.abc import Iterable
 from datetime import UTC, datetime
 from enum import StrEnum
 
+from pydantic import BaseModel
 from sretoolbox.utils import threaded
 
 from reconcile.external_resources.factories import (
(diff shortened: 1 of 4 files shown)


reconcile/external_resources/manager.py
```diff
@@ -88,6 +88,56 @@ class ReconcileAction(StrEnum):
     DESTROY_ERROR = "Resource status in ERROR state"
 
 
+class ReconciliationStatus:
+    needs_secret_sync: bool = False
+    needs_deletion: bool = False
+    has_errors: bool = False
```
chassing, line 94: `has_errors` could be a `@property` instead of manually setting it

reconcile/external_resources/manager.py
```diff
@@ -88,6 +88,56 @@ class ReconcileAction(StrEnum):
     DESTROY_ERROR = "Resource status in ERROR state"
 
 
+class ReconciliationStatus:
+    needs_secret_sync: bool = False
```
chassing, line 92: `needs_secret_sync` could also be a `@property`

reconcile/external_resources/manager.py
```diff
@@ -88,6 +88,64 @@ class ReconcileAction(StrEnum):
...
+    needs_deletion: bool = False
+
+    reconcile_time: int = 0
+    resource_status: ResourceStatus
```
hemslo, line 96: note these vars are class vars, but usages are instance vars, should either use `@dataclass` or pydantic model, or init them in `__init__`

reconcile/external_resources/manager.py
lechuk47: Added as a property

reconcile/external_resources/manager.py
lechuk47: Added as a property

chassing (approved):

hemslo (approved):
LGTM!

PR #4731, opened 2024-10-16
Fix misleading vault replication test

rporres:
When secrets don't end with a "/", we don't call Vault, hence the side effect was wrong and we were not testing the case of a path with dashes or underscores.


rporres:
related to https://github.com/app-sre/qontract-reconcile/pull/4722

PR #4732, opened 2024-10-16
fix rosa create cluster computation of replicas

patjlm:
https://issues.redhat.com/browse/APPSRE-11013

Each machinepool has its own `replicas`. `length` was only counting the number of machinepools.

### DIFF
--- a/reconcile/templates/rosa-classic-cluster-creation.sh.j2
+++ b/reconcile/templates/rosa-classic-cluster-creation.sh.j2
@@ -49,7 +49,11 @@ rosa create cluster -y --cluster-name={{ cluster_name }} \
     --service-cidr {{ cluster.network.service }} \
     --pod-cidr {{ cluster.network.pod }} \
     --host-prefix 23 \
-    --replicas {{ cluster.machine_pools | length }} \
+    {#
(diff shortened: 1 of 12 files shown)


esron (approved):
LGTM! But this would be fixed by the `ocm-machine-pools` integration.

patjlm:
> LGTM! But this would be fixed by the `ocm-machine-pools` integration.

That's very true! I am trying here to get things a bit more right than today, so we avoid this 2 step scaleup. Using length is pretty wrong. And I think it would even fail on classic, where we define only 1 `worker` machinepool and replicas get split in the background, knowing CS requires a minimum of 4 workers for single AZ clusters and 9 for multi-AZ ones.

At the same time, what is proposed here is only a quick fix and very far from being perfect. It is based on assumptions and does not support autoscaling (which we should pretty much always use..)

So I have a dilemma getting this merged :smile: .

rporres:
my $0.02: If this is going to fix current issues, let's merge it until we got the definitive solution

patjlm:
> If this is going to fix current issues, let's merge it until we got the definitive solution

I don't think we have any critical issue with the current situation (at least on HCP). We only wait a bit longer to get the desired node count.

I'll close. We can always reopen or suggest an improved change later

PR #4737, opened 2024-10-22
:sparkles: ERv2: make outputs-secret image and version configurable

chassing:
Remove hardcoded `er-outputs-secrets` image and version.

The image and the version must be configured in `/external-resources/settings-1.yml` but can be overridden in `/external-resources/module-1.yml` and `/external-resources/module-overrides-1.yml`

Depends on: https://github.com/app-sre/qontract-schemas/pull/724
Ticket: [APPSRE-11049](https://issues.redhat.com/browse/APPSRE-11049)


esron (approved):
LGTM

PR #4738, opened 2024-10-22
Improve error message for ocm machine pools validation

esron:
Related to: [APPSRE-10481](https://issues.redhat.com/browse/APPSRE-10481)

This validates that if the user wants to create machine pools that uses the ROSA HCP default machine pools ids (workers-0, workers-1 and workers-2) and the subnet ids attached to them is not in the correct alphabetical order the `ocm-machine-pools` integrations fails.

This is due to a limitation on the cluster creation side that is well documented on the ticket.

To test this you can use the template from this [MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/114908#969c6a8503d37323d7484aa18c2bfeac0f348c2f).

The templating side was already fixed this is a safe measure for if someone tries to create cluster by using only the cluster schema.

EDIT: 

After much debate and analysis I think It's bet to let things be as they are right now and just improve the error message.

### DIFF
--- a/reconcile/ocm_machine_pools.py
+++ b/reconcile/ocm_machine_pools.py
@@ -465,7 +465,7 @@ def calculate_diff(
         if invalid_diff:
             errors.append(
                 InvalidUpdateError(
-                    f"can not update {invalid_diff} for existing machine pool on cluster {cluster_name}"
+                    f"can not update {invalid_diff} for existing machine pool on cluster {cluster_name}, CURRENT: {diff_pair.current.json()}, DESIRED: {diff_pair.desired.json()}"
                 )
             )
         else:


reconcile/ocm_machine_pools.py
```diff
@@ -453,6 +484,8 @@ def calculate_diff(
     diffs: list[PoolHandler] = []
     errors: list[InvalidUpdateError] = []
 
+    errors += validate_desired_pools(diff_result.add)
```
hemslo, line 487: my understanding is the error mentioned in ticket is this flow:

* create cluster using rosa cli with 3 subnets
* 3 machine pools created
* populate `machinePools` in `$schema: /openshift/cluster-1.yml` with the wrong order
* `ocm-machine-pools` diff detect changes in `subnet` in `diff_result.change` and it's a `invalid_diff`

The validation is already correct, why add another?

And the desired machine pools can be unordered if I do this:

* remove `workers-2` (added some other workers first)
* add `workers-2` back, now the subnet can be any one of the subnets

esron: Upon cluster creation the subnets provided in `machinePools` in `$schema: /openshift/cluster-1.yml` are not used.  And `ocm-machine-pools` cannot detect the error because from the `ocm-machine-pools` perspective those are valid `machinePools`. 

Once the cluster is created by `ocm-cluster`, `ocm-machine-pools` tries to update the machine pools and can fail if the subnets are in the wrong order.

`ocm-cluster` will always create the cluster with 3 machine pools using machine pools ids `workers-0`, `workers-1` and `workers-2`. Assigning subnet ids in alphabetical order respectively. i.e:
```
workers-0 -> subnet-a
workers-1 -> subnet-b
workers-2 -> subnet-c
```

That said, your last point is right.

cubismod: I'm not super familiar with this file so pardon my ignorance, but would it be possible to perform this validation once you have the whole list of `diffs`? I'm thinking of something similar to what happens in `act` already: https://github.com/app-sre/qontract-reconcile/blob/8290077b892118cab06f2bd33623d49df311ecd4/reconcile/ocm_machine_pools.py#L310-L322 except you're just appending/removing/updating from that list and then checking the order of that final list to perform validation

hemslo: It can, but not really needed as current validation can already cover invalid subnet diff, anytime current subnet different than desired subnet will lead to

```
                InvalidUpdateError(
                    f"can not update {invalid_diff} for existing machine pool on cluster {cluster_name}"
                )
```

non alphabetical ordered subnets during HCP init setup is a subset of this validation rule, and non alphabetical ordered subnets in general can be a valid case as long as it matches current state.

for this data:

```text
workers-0 -> subnet-b
workers-1 -> subnet-a
workers-2 -> subnet-c
```

we will have 2 errors

```text
can not update subnet for existing machine pool on cluster xxx
can not update subnet for existing machine pool on cluster xxx
```

So my suggestion is to include more info in error message to make it clear what's the diff, like


```python
                InvalidUpdateError(
                    f"can not update {invalid_diff} for existing machine pool on cluster {cluster_name}, CURRENT: {diff_pair.current.json()}, DESIRED: {diff_pair.desired.json()}"
                )

```

esron: I think @hemslo is right. A real safe check can be too complicated to implement and not give us that much, since the template is already fixed.

esron: I've improved the messaging for this error. Could you please review again?

hemslo (approved):
LGTM!

PR #4741, opened 2024-10-24
Add settings to exclude providers from provisioners in terraform_resources

lechuk47:
This allows external resource providers to be excluded from specific provisioners (accounts) for terraform_resources. For example, once all RDS instances have been migrated to V2, we don't want new RDS resources to use terraform_resources. 

schemas: app-sre/qontract-schemas#725

### DIFF
--- a/reconcile/queries.py
+++ b/reconcile/queries.py
@@ -102,6 +102,12 @@ APP_INTERFACE_SETTINGS_QUERY = """
       readTimeout
       connectTimeout
     }
+    terraformResourcesProviderExclusionsByProvisioner {
+      provisioner {
+          name
+      }
+      excludedProviders
+    }
   }
 }
 """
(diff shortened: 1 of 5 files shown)


reconcile/queries.py
```diff
@@ -2754,3 +2760,24 @@ def get_blackbox_exporter_monitoring_provider() -> dict:
...
+  }
+}
+"""
+
+
+def get_tf_resources_provider_exclusions_by_provisioner():
+    gqlapi = gql.get_api()
+    return gqlapi.query(TF_RESOURCES_PROVIDER_EXCLUSIONS_BY_PROVISIONER)[
+        "tf_provider_exclusions_by_provisioner"
+    ]
```
hemslo, line 2783: can we move this to typed queries?

reconcile/utils/terrascript_aws_client.py
```diff
@@ -1551,9 +1593,19 @@ def init_populate_specs(
...
+
         for namespace_info in namespaces:
-            specs = get_external_resource_specs(
-                namespace_info, provision_provider=PROVIDER_AWS
+            all_specs = get_external_resource_specs(
+                namespace_info,
+                provision_provider=PROVIDER_AWS,
+            )
+            specs = self._filter_specs_managed_by_erv2(
+                all_specs, provider_exclusions_by_provisioner
```
hemslo, line 1608: since `provider_exclusions_by_provisioner` will be used for every namespace, to speed up and simplify filter logic, can pre build a dict


```python
provider_exclusions_by_provisioner_name = {
  p["provisioner"]["name"]: p["excludedProviders"]
  for p in provider_exclusions_by_provisioner
}
```

Then filter logic can use it to speed up from O(NM) to O(M), N is length of `provider_exclusions_by_provisioner`, M is length of `excludedProviders`.

```python
            if spec.provider in provider_exclusions_by_provisioner_name.get(spec.provisioner["name"], []):
                raise ProviderExcludedError(spec)
```

reconcile/queries.py
lechuk47: In this specific case, it's better to stick to queries. This attribute normally comes from app_interface_settings which is obtained using `queries`. This code is only used if the attribute is None to ensure the exclusions are taken into account.  

With a typed query, the code to build the dictionary from your other suggestion will need to handle both cases nad I think it's not necessary in this specific case.

hemslo (approved):
LGTM!

PR #4746, opened 2024-10-29
Improve terraform-resources provider exclusions

lechuk47:
An improvement over how the provider exclusions are managed in AI settings. 
* Allows to exclude all provisioners for a provider: (useful for new modules and for new accounts when created) 
* Now the list is based on providers instead of provisioners. 

Requires: app-sre/qontract-schemas#727

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -9,7 +9,7 @@ import string
 import tempfile
 from collections import Counter
 from collections.abc import Iterable, Mapping, MutableMapping
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from ipaddress import (
     ip_address,
     ip_network,
@@ -377,6 +377,12 @@ class ElasticSearchLogGroupInfo:
     account_id: str
     region: str
     log_group_identifier: str
+
+
+@dataclass
+class Exclusion:
+    all: bool = False
+    provisioners: set[str] = field(default_factory=set)
 
 
 class ProviderExcludedError(Exception):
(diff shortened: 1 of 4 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -1543,38 +1543,48 @@ def populate_resources(self, ocm_map: OCMMap | None = None) -> None:
...
+    ) -> dict[str, list[str]]:
+        exclusions: dict[str, list[str]] = {}
+        for item in provider_exclusions or []:
+            provider = item["provider"]
+            if item.get("excludeAllProvisioners"):
+                exclusions[provider] = ["ALL"]
+            elif item.get("excludeProvisioners"):
+                exclusions[provider] = [
+                    p["name"] for p in item.get("excludeProvisioners") or []
+                ]
```
hemslo, line 1580: now the structure is more complex than before, better to define a class to hold `excludeAllProvisioners` and `excludeProvisioners`, and avoid magic `ALL`, for example a `Exclusion` dataclass

```python
@dataclass
class Exclusion:
    all = False
    provisioners = field(default_factory=set)
```

then can construct exclusions like

```python
    return {
        item["provider"]: Exclusion(
            all = item.get("excludeAllProvisioners") or False,
            provisioners = {p["name"] for p in item.get("excludeProvisioners") or []}
        )
        for item in provider_exclusions or []
    }
```

check exclude also clearer like

```python
return exclusion.all or spec.provisioner_name in exclusion.provisioners
```

lechuk47: yeah, makes sense. Added.

hemslo (approved):
LGTM!

PR #4750, opened 2024-10-31
if secret version is not found default to latest version

jonmosco:
This update modifies the secret retrieval logic to default to the latest version if a specified version does not exist. This ensures that the latest available secret will be used as a fallback in cases where secrets are migrated, consolidated, or versioned differently.


maorfr:
why do you want to change this behavior? i think it is a very important behavior from a reliability perspective.

PR #4752, opened 2024-11-04
add support to remove user ref from schedules when deleting a user

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10359

depends on:
- https://github.com/app-sre/qontract-schemas/pull/734
- https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/123253

this PR adds support to remove user refs from schedules to allow user offboarding.


cubismod (approved):
LGTM

reconcile/utils/mr/user_maintenance.py
```diff
@@ -68,6 +69,21 @@ def process(self, gitlab_cli: GitLabApi) -> None:
                             commit_message=self.title,
                             content=new_content,
                         )
+            elif path_type == PathTypes.SCHEDULE:
+                raw_file = gitlab_cli.project.files.get(file_path=path, ref=self.branch)
+                content = yaml.load(raw_file.decode(), Loader=yaml.RoundTripLoader)
+                for schedule_index, schedule_record in enumerate(content["schedule"]):
+                    for user_index, user in enumerate(schedule_record["users"]):
+                        if self.username in user["$ref"]:
+                            del content["schedule"][schedule_index]["users"][user_index]
```
hemslo, line 78: delete item from list while iterating it can cause issues since index will change after deletion. Can check if comprehension works on this object

```python
                for schedule_record in content["schedule"]:
                    schedule_record["users"] = [
                        user
                        for user in schedule_record["users"]
                        if self.username not in user["$ref"]
                    ]
```

also note username check may wrongly match items, for example, `self.username` is `a`, it will delete all users with `a` in `$ref`.

hemslo: I'm trying to fix another offboard [mr](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/123345), `users` can't be empty, it will fail schema validation as it requires at least one item, the whole schedule item need to be deleted

maorfr: - delete item from list fixed in https://github.com/app-sre/qontract-reconcile/pull/4752/commits/75695856c3b996b551b45f4567fd7a852445bc42
- users can't be empty fixed in https://github.com/app-sre/qontract-schemas/pull/734/commits/39e3f6cb8780add0d7709d15bb3a9ed6d6b4e9bb
- username check may be wrong fixed in https://github.com/app-sre/qontract-reconcile/pull/4752/commits/406844482d2f1e6be6050b89e4a8adf72ffed1e9 according to our conventions

hemslo: that delete by reversed index order is hard to understand, if that yaml object only support deletion in this way, then that's it...

hemslo (approved):
LGTM!

reconcile/utils/mr/user_maintenance.py
hemslo: yes, that's the hidden logic behind reverse, not obvious when first read, mutation is always hard to read, that's why I always prefer comprehension.

PR #4760, opened 2024-11-11
Fix: Erv2 does not update the state when reconciliation ends with error

lechuk47:


reconcile/external_resources/state.py
```diff
@@ -38,6 +38,15 @@ class ResourceStatus(StrEnum):
...
 
+    def is_in_progress(self) -> bool:
+        return self in {ResourceStatus.IN_PROGRESS, ResourceStatus.DELETE_IN_PROGRESS}
+
+    def needs_secret_sync(self) -> bool:
+        return self == ResourceStatus.PENDING_SECRET_SYNC
+
+    def has_errors(self) -> bool:
+        return self == ResourceStatus.ERROR
+
```
chassing, line 49: Please change all three methods to be properties.

lechuk47: done

PR #4761, opened 2024-11-12
Remove wrong logging line

lechuk47:


chassing (commented):
Please put the log info line below, otherwise we don't see anything

chassing (approved):

PR #4763, opened 2024-11-13
:bug: do not run outputs-secrets container for delete action

chassing:
During a delete reconciliation run, the `K8sJob` doesn't have access to the `outputs_secrets` global config because it's not persisted in the DynamoDB state; nevertheless a deletion run don't need to run the `outputs-secrets` container at all.


lechuk47 (approved):
LGTM

PR #4764, opened 2024-11-14
orderby to avoid pagination issue

fishi0x01:
Thanks Gerd for spotting this one :heart: 

~Attempting to set `orderBy: id` as default if no order is given~

### DIFF
--- a/reconcile/utils/ocm_base_client.py
+++ b/reconcile/utils/ocm_base_client.py
@@ -87,10 +87,8 @@ class OCMBaseClient:
         """
         Note, that pagination is currently broken.
         Each call will return a random order, meaning pages are not consistent.
-        ALWAYS by default use "orderBy: id", as id exists for every resource and has an index in the db.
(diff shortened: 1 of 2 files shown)


reconcile/utils/ocm_base_client.py
```diff
@@ -84,6 +84,13 @@ def get_paginated(
         max_page_size: int = 100,
         max_pages: int | None = None,
     ) -> Generator[dict[str, Any], None, None]:
+        """
+        Note, that pagination is currently broken.
+        Each call will return a random order, meaning pages are not consistent.
+        ALWAYS by default use "orderBy: id", as id exists for every resource and has an index in the db.
+        """
+        params = params or {}
+        params.setdefault("orderBy", "id")
```
geoberle, line 93: Are you sure everything has an id to sort?

fishi0x01: Just an assumption. But we can also keep this out just in case :+1:

geoberle (approved):

PR #4768, opened 2024-11-18
[tf-repo] remove restriction on number of repos that can be updated in one MR

cubismod:
Terraform Repo was created initially with a limit at the PR check level: only one repo can be modified in one PR

While this made sense when the project was initially created, it limits AppSRE whenever we need to make an update to [the schema](https://github.com/app-sre/qontract-schemas/blob/main/schemas/aws/terraform-repo-1.yml). If we add a new required property that requires mass-updating each reference to that schema, AppSRE would have to submit a separate MR for each repo file. This removes that limitation allowing for mass updates like other integrations.

The concern around tenants updating multiple repos at once is addressed through change-types within App-Interface limiting the amount of repos a tenant can self-service.

This will enable faster updates of tf-repo going forward.

Related to [APPSRE-11139](https://issues.redhat.com/browse/APPSRE-11139)

### DIFF
--- a/reconcile/terraform_repo.py
+++ b/reconcile/terraform_repo.py
@@ -100,7 +100,6 @@ class TerraformRepoIntegration(
                 desired_state=desired,
                 dry_run=dry_run,
                 state=state,
-                recreate_state=False,
             )
 
             if repo_diff_result:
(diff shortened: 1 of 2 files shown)


esron (approved):
LGTM

PR #4770, opened 2024-11-20
[APPSRE-11112] enable webhook for only MR comments

BumbleFeng:
As part of APPSRE-11112, we need to create webhook for triggering OnDemand test job with only MR comments

### DIFF
--- a/reconcile/jenkins_webhooks.py
+++ b/reconcile/jenkins_webhooks.py
@@ -24,7 +24,13 @@ def get_hooks_to_add(desired_state, gl):
             current_hooks = gl.get_project_hooks(project_url)
             for h in current_hooks:
                 job_url = h.url
-                trigger = "mr" if h.merge_requests_events else "push"
+                trigger = []
+                if h.merge_requests_events:
+                    trigger.append("mr")
+                if h.push_events:
+                    trigger.append("push")
+                if h.note_events:
+                    trigger.append("note")
                 item = {
                     "job_url": job_url.strip("/"),
                     "trigger": trigger,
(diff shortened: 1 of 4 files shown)


reconcile/jenkins_webhooks.py
```diff
@@ -24,7 +24,13 @@ def get_hooks_to_add(desired_state, gl):
...
             for h in current_hooks:
                 job_url = h.url
-                trigger = "mr" if h.merge_requests_events else "push"
+                trigger = (
+                    "mr"
+                    if h.merge_requests_events
+                    else "push"
+                    if h.push_events
+                    else "note"
+                )
```
hemslo, line 33: now the logic is complex, I can't understand the priority in nested if else, better to extract a function and see if can reuse in `jjb_client` too.

reconcile/utils/jjb_client.py
```diff
@@ -399,3 +396,21 @@ def get_trigger_phrases_regex(job: dict) -> str | None:
...
+        # build main/master job should be triggered by push events
+        elif gitlab_triggers.get("trigger-push"):
+            trigger = ["push"]
+        # On-demand test job should be triggered by special comment
+        elif gitlab_triggers.get("trigger-note"):
+            trigger = ["note"]
+        else:
+            trigger = None
+
+        return trigger
```
hemslo, line 416: it's better to keep return value as not `None` to match https://github.com/app-sre/qontract-reconcile/pull/4770/files#diff-f910de274ff6f61d8f3549e1e2a965487e59bf2f7ee8466b6a77659962357dd1R27-R33, then need to update `if trigger is None:` -> `if not trigger:`. This can also make `in` type check easier as `trigger` is always a list.

And can simplify condition by early return

```python
        if gitlab_triggers.get("trigger-merge-request"):
            return ["mr", "note"]
        if gitlab_triggers.get("trigger-push"):
            return ["push"]
        if gitlab_triggers.get("trigger-note"):
            return ["note"]
        return []
```

hemslo (approved):
LGTM!

PR #4771, opened 2024-11-21
:sparkles: oum: replace ocm-org-ids option with query

chassing:
I want to move the `ocm-standalone-user-management` integration to `sre-capabilties`. In `sre-capabilties`, we can use the `/openshift/openshift-cluster-manager-1.yml` definitions instead of relying on a command line option, as we have for `rhidp` or other integrations.

If we want to turn off this integration for a particular org, we can use the `/openshift/openshift-cluster-manager-1.yml.disable` attribute. 

Ticket: [APPSRE-11227](https://issues.redhat.com/browse/APPSRE-11227)


chassing:
@geoberle, anything against this?

reconcile/gql_definitions/oum/organizations.gql
```diff
@@ -0,0 +1,14 @@
+# qenerate: plugin=pydantic_v1
+
+query OUMOrganizations($name: String) {
```
hemslo, line 3: is `$name` used anywhere? There is one client side filter by env name, can we do server side filter?

chassing: It is not needed. Remove in [d590cc8](https://github.com/app-sre/qontract-reconcile/pull/4771/commits/d590cc8291cf1f8b0ce3951bbd99e73d709909dd)

hemslo (approved):
LGTM!

PR #4769, opened 2024-11-20
Container Images used report

rporres:
```
$ qontract-cli --config config.development.toml get  container-images --help
Usage: qontract-cli get container-images [OPTIONS]

  Get all container images in app-interface defined namespaces

Options:
  --cluster-name TEXT             openshift cluster names to act on i.e.:
                                  --cluster-name cluster-1 --cluster-name
                                  cluster-2
  --namespace-name TEXT           namespace name to act on.
  --thread-pool-size INTEGER      number of threads to run in parallel.
  --use-jump-host / --no-use-jump-host
                                  use jump host if defined.
  --exclude-pattern TEXT          Exclude images that match this pattern
  --include-pattern TEXT          Only include images that match this pattern
  --help                          Show this message and exit.
```

APPSRE-11229

PR #4773, opened 2024-11-24
Handle errors in container-images report

rporres:
The code from #4769 did not handle errors properly. Errors have been added in a sort of hackish way (at the end of the image list), but it is better to show them somewhere than hiding them.

Output is now sorted by default for clarity.

APPSRE-11229


hemslo (approved):
LGTM!

PR #4775, opened 2024-11-25
Fix Erv2 drift detection runs

lechuk47:
The state timestamp was updated at every run, causing the drift detection not to trigger.

### DIFF
--- a/reconcile/external_resources/manager.py
+++ b/reconcile/external_resources/manager.py
@@ -1,9 +1,7 @@
 import logging
 from collections.abc import Iterable
 from datetime import UTC, datetime
-from enum import StrEnum
 
-from pydantic import BaseModel
 from sretoolbox.utils import threaded
 
 from reconcile.external_resources.factories import (
@@ -14,11 +12,7 @@ from reconcile.external_resources.factories import (
     TerraformModuleProvisionDataFactory,
     setup_aws_resource_factories,
 )
-from reconcile.external_resources.metrics import (
-    ExternalResourcesReconcileErrorsCounter,
-    ExternalResourcesReconcileTimeGauge,
-    ExternalResourcesResourceStatus,
-)
+from reconcile.external_resources.metrics import publish_metrics
 from reconcile.external_resources.model import (
     Action,
     ExternalResource,
@@ -27,11 +21,12 @@ from reconcile.external_resources.model import (
     ExternalResourcesInventory,
     ExternalResourceValidationError,
     ModuleInventory,
+    ReconcileAction,
     Reconciliation,
+    ReconciliationStatus,
 )
 from reconcile.external_resources.reconciler import (
     ExternalResourcesReconciler,
-    ReconciliationK8sJob,
 )
 from reconcile.external_resources.secrets_sync import InClusterSecretsReconciler
 from reconcile.external_resources.state import (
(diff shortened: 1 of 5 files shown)


reconcile/external_resources/manager.py
```diff
@@ -298,47 +298,50 @@ def _get_reconciliation_status(
...
-            )
+        if not reconciliation_status.needs_state_update:
+            logging.debug("Reonciliation does not need to update the resource state")
             return
-        state.ts = datetime.now(UTC)
+
         if reconciliation_status.resource_status == ResourceStatus.DELETED:
             self.state_mgr.del_external_resource_state(r.key)
         else:
+            state.ts = datetime.now(UTC)
```
hemslo, line 344: When `ts` should change? Should it only change when `state.resource_status` changed?

reconcile/external_resources/manager.py
```diff
@@ -92,6 +92,7 @@ class ReconcileAction(StrEnum):
 class ReconciliationStatus(BaseModel):
     reconcile_time: int = 0
     resource_status: ResourceStatus
+    needs_state_update: bool = False
```
chassing, line 95: Please make it a `property` as we have for [ResourceStatus](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/external_resources/state.py#L46)

lechuk47: Do you mean using logic depending the ResourceStatus values ?

reconcile/external_resources/manager.py
lechuk47: Not exactly. Whenever it changes from in-progress to another state.

reconcile/external_resources/manager.py
chassing: Yes, exactly

reconcile/external_resources/manager.py
hemslo: then you can manage it inside `state`, instead of

```python
            state.ts = datetime.now(UTC)
            state.resource_status = reconciliation_status.resource_status
```

the can be done as a state method


```python
state.update_status(reconciliation_status.resource_status)
```

and handles it internally

```python
def update_status(self, status):
    if self.resource_status == in_progress and status != in_progress:
        self.ts = datetime.now(UTC)
    self.resource_status = status
```

reconcile/external_resources/manager.py
lechuk47: The logic to update this attribute depends on the state's resource_status and the reconciliation resource_status. It can not be set only with its resource_status attribute. 

passing the state's to a method to set this as a property feels a bit weird to me.

chassing: I don't see the dependency on the state's `resource_status`. For me, it looks like

```
@propery
def needs_state_update(self) -> bool:
    return self.resource_state in [ResourceStatus.DELETED, ResourceStatus.ERROR]
```

lechuk47: needs_state_update is False by default, it's only set True in `_get_reconciliation_status`  if the state.resource_status is in progress.

hemslo: https://github.com/app-sre/qontract-reconcile/pull/4775#discussion_r1856614330 can avoid that field or property

reconcile/external_resources/manager.py
lechuk47: OK. I moved some things around to do this and make things a bit clearer (at least for me). PTAL cc: @chassing

chassing (approved):
LGTM

hemslo (approved):
LGTM!

PR #4777, opened 2024-11-26
fix(vpc-peering): match new security group name for HCP cluster

hemslo:
Update security group name match to include `-vpce-private-router` ones as the logic changed in HCP, new clusters will have the new security group name.

[APPSRE-11280](https://issues.redhat.com/browse/APPSRE-11280)


fishi0x01 (approved):
LGTM :rocket:

PR #4782, opened 2024-11-29
fix(change-log-tracking): use app path when name not found

hemslo:
Fix error when app file deleted.

```text
File "/usr/local/lib/python3.11/site-packages/reconcile/change_owners/change_log_tracking.py", line 189, in run
raise KeyError(
KeyError: "app name is expected. missing in query? app information: {'$ref': '/services/xxxx/app.yml'}"
```

This pull request includes changes to handle missing app names more gracefully and adds a new test for this scenario. The most important changes are:

### Error Handling Improvement:
* Modified `reconcile/change_owners/change_log_tracking.py` to log a message and use the app path as the name if the app name is not found, instead of raising a `KeyError`.

### Testing Enhancements:
* Added a comprehensive test in `reconcile/test/change_owners/test_change_log_tracking.py` to verify the behavior when an app is deleted and its name is missing. This includes setting up necessary mocks and verifying the change log.

### DIFF
--- a/reconcile/change_owners/change_log_tracking.py
+++ b/reconcile/change_owners/change_log_tracking.py
@@ -186,9 +186,10 @@ class ChangeLogIntegration(QontractReconcileIntegration[ChangeLogIntegrationPara
                                 )
                             app_name = app_name_by_path.get(app_path)
                             if not app_name:
-                                raise KeyError(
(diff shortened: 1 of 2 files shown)


mehfuz (approved):
LGTM

PR #4785, opened 2024-12-02
Erv2: Implement resource reconciliation dependencies

lechuk47:
This PR implements resource reconciliation dependencies. 

The main use case for now is RDS read replicas. Every time a Primary is reconciled, it will trigger a reconciliation for its read replicas. This is needed to sync the credentials output secrets in case a password reset is requested.

### DIFF
--- a/reconcile/external_resources/aws.py
+++ b/reconcile/external_resources/aws.py
@@ -3,6 +3,7 @@ from typing import Any
 
 from reconcile.external_resources.model import (
     ExternalResource,
+    ExternalResourceKey,
     ExternalResourcesInventory,
 )
 from reconcile.utils.external_resource_spec import (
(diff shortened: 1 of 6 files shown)


reconcile/external_resources/manager.py
```diff
@@ -161,19 +161,20 @@ def _get_desired_objects_reconciliations(self) -> set[Reconciliation]:
...
+                resource = self._build_external_resource(spec)
             except ExternalResourceValidationError as e:
                 self.errors[key] = e
                 continue
 
             reconciliation = Reconciliation(
                 key=key,
                 resource_hash=resource.hash(),
-                input=self._serialize_resource_input(resource),
+                input=resource.json(),
```
chassing, line 172: Looks like `_serialize_resource_input` isn't in use anymore. Please remove it.

reconcile/external_resources/aws.py
```diff
@@ -25,6 +26,11 @@ def resolve(self, spec: ExternalResourceSpec) -> dict[str, Any]: ...
     @abstractmethod
     def validate(self, resource: ExternalResource) -> None: ...
 
+    def find_linked_resources(
+        self, spec: ExternalResourceSpec
+    ) -> set[ExternalResourceKey]:
+        return set()
```
chassing, line 32: Please add a doc string here, mainly because it's the base class.

reconcile/external_resources/manager.py
lechuk47: It's already removed.

reconcile/external_resources/aws.py
lechuk47: Added.

chassing (approved):

PR #4787, opened 2024-12-03
Fail ERv2 DRY-RUN mode when there are orphaned resources

lechuk47:
Prevent Erv2 resources removal without setting the delete flag.

### DIFF
--- a/reconcile/external_resources/model.py
+++ b/reconcile/external_resources/model.py
@@ -34,6 +34,17 @@ from reconcile.utils.external_resource_spec import (
 )
 
 
+class ExternalResourceOrphanedResourcesError(Exception):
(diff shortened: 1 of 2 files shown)


reconcile/external_resources/model.py
```diff
@@ -34,6 +34,17 @@
 )
 
 
+class ExternalResourceOrphanedResourcesError(Exception):
+    def __init__(self, orphans: list["ExternalResourceKey"]) -> None:
```
chassing, line 38: Please use `orphans: Iterable[ExternalResourceKey]`

lechuk47: fixed

chassing (approved):

PR #4796, opened 2024-12-12
Add a dashboard for terraform-repo

cubismod:
[APPSRE-11304](https://issues.redhat.com/browse/APPSRE-11304)

Adds a basic dashboard for Terraform Repo. Screenshot from stage:

image

(AWS account names blurred for privacy). Currently, TF Repo is being utilized in a more limited way which is why logs are missing currently but they will be populated whenever a pipeline is run.


esron (approved):
LGTM

PR #4800, opened 2024-12-17
:zap: slack-usergroups: performance fix

chassing:
Don't download the complete git repo file list.

`slack-usergroups` uses just the root level (`.`) `OWNERS` file but downloads a list of all files from the git repos. [clusterimagesets](https://gitlab.cee.redhat.com/service/clusterimagesets) has more than 31000 files, `get_repository_tree` receives the file catalog in 100 chunks. In total, ~310 HTTP requests, and each takes 1 second.

This fix reduces the runtime of a reconciliation loop by 6 minutes.

### DIFF
--- a/reconcile/utils/repo_owners.py
+++ b/reconcile/utils/repo_owners.py
@@ -133,14 +133,12 @@ class RepoOwners:
                 owners = None
             if owners is None:
                 _LOG.warning(
-                    "Non-parsable OWNERS file "
-                    f"{self._git_cli.project.web_url}:{owner_file.get('path')}"
+                    f"Non-parsable OWNERS file {self._git_cli!s}:{owner_file['path']}"
                 )
                 continue
             if not isinstance(owners, dict):
                 _LOG.warning(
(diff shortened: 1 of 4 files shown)


esron (approved):
LGTM

reconcile/utils/repo_owners.py
```diff
@@ -116,16 +117,13 @@ def _get_owners_map(self):
...
-        for item in repo_tree:
-            if item["path"] == "OWNERS_ALIASES":
-                aliases = self._get_aliases()
-            elif item["name"] == "OWNERS":
-                owner_files.append(item)
+        if self._recursive:
+            repo_tree = self._git_cli.get_repository_tree(ref=self._ref)
+            owner_files = [item for item in repo_tree if item["name"] == "OWNERS"]
+        else:
+            owner_files = [{"path": "OWNERS"}]
```
patjlm, line 126: what happens if there is no such file? Won't we raise an exception later?

PR #4801, opened 2024-12-19
Catch null project

fishi0x01:
We can create new repos via app-interface. Gitlab-permissions will fail though, if a repo doesn't exist yet. This leads to a chicken-egg problem.

We introduce a skip on non-existing app-interface managed projects, whenever a new addition is detected in the diff.

Tested locally with failing tenant MR.


esron (approved):
LGTM

rporres (approved):

PR #4805, opened 2025-01-07
skip target if deleted while assembling channels for promotion

mehfuz:
In case of target marked as deleted and having a publish promotion..all the upcoming subscribe promotion will be blocked..as `_assemble_channels` will still consider the deleted target
fixes : https://issues.redhat.com/browse/APPSRE-11364


patjlm (approved):
nice find!

hemslo (approved):
LGTM!

PR #4810, opened 2025-01-14
APPSRE-11389 annotate DTP version in secret too

fishi0x01:
Follow up on https://github.com/app-sre/qontract-reconcile/pull/4807

Lets also add the DTP version. This helps us to compare if a change actually lead to a reconcile in a secret.

Also adding version to early_exit dict. Not sure if it makes sense, as we always reconcile anyways on version bump w/o early_exit. But at the same time it doesnt harm to have it here.


hemslo (approved):
LGTM!

PR #4812, opened 2025-01-15
Erv2: Output_resource_names duplications validation

lechuk47:
APPSRE-11409

### DIFF
--- a/reconcile/external_resources/manager.py
+++ b/reconcile/external_resources/manager.py
@@ -1,4 +1,5 @@
 import logging
+from collections import Counter
 from collections.abc import Iterable
 from datetime import UTC, datetime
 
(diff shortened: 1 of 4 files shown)


reconcile/external_resources/manager.py
```diff
@@ -73,6 +74,42 @@ def setup_factories(
...
+
+        for spec in self.er_inventory.values():
+            output_key = (
+                spec.cluster_name,
+                spec.namespace_name,
+                spec.output_resource_name,
+            )
+            if output_key in seen:
+                raise ExternalResourceOutputResourceNameDuplication(output_key)
+            seen.add(output_key)
```
hemslo, line 99: if you want to find all duplicates, can either use another set to store duplicates or use `Counter`

```python
specs = Counter(
  (
     spec.cluster_name,
     spec.namespace_name,
     spec.output_resource_name,
  )
  for spec in self.er_inventory.values()
)

duplicates = [key for key, count in specs.items() if count > 1]
```

lechuk47: I prefer to fail fast at pr-checks, there is no need to go through all the objects here as the ones duplicating will be the ones in the related PR.

lechuk47: Implemented this way.

PR #4813, opened 2025-01-16
:sparkles: jira-permissions-validator: sharding support

chassing:
Add sharding support to the `jira-permissions-validator` integration to reduce amount of Jira API calls.

Depends on: https://github.com/app-sre/qontract-schemas/pull/757


rporres:
The code looks fine, what I don't get is the concept. How sharding by board is going to help us?

I'm mainly afraid of how costly is this in resources (we have 50 boards, hence 50 processes running). Have we tried something a bit less costly as static sharding.

Apologies if this has been answered elsewhere...

reconcile/jira_permissions_validator.py
```diff
@@ -256,11 +256,14 @@ def validate_boards(
     return error
 
 
-def get_jira_boards(query_func: Callable) -> list[JiraBoardV1]:
+def get_jira_boards(
```
hemslo, line 259: `get_jira_boards` also called in `early_exit_desired_state`, should it also consider `jira_board_name`?

chassing:
> The code looks fine, what I don't get is the concept. How sharding by board is going to help us?

If the integration hits the rate limit, then just one board is affected. Any retry won't start from the beginning and needs fewer API requests.

> 
> I'm mainly afraid of how costly is this in resources (we have 50 boards, hence 50 processes running). Have we tried something a bit less costly as static sharding.

TBH, I don't understand the static sharing concept.

reconcile/jira_permissions_validator.py
chassing: I don't get it. What do you mean?

rporres (approved):
@chassing and I discussed about this in a call. Since we're in bad position with JIRA guys, let's try this (which may be a bit overkill) and in the future maybe look for other solutions like:

* only run PR checks in affected boards
* introduce a random wait per board in the integration and remove per board sharding

For the moment, let's try the overkill solution

patjlm (approved):

hemslo (approved):
LGTM!

PR #4815, opened 2025-01-20
APPSRE-11428 mitigate DT bug

fishi0x01:
Dynatrace API does not return milliseconds if they are `000`, e.g., `2025-01-18T21:35:03.000Z` gets converted to `2025-01-18T21:35:03Z`

That leads to the following error in the client API:

```
ValueError("time data '2025-01-18T21:35:03Z' does not match format '%Y-%m-%dT%H:%M:%S.%fZ'")
```

We will submit a bug ticket to DT shortly, but until then, we monkey patch a fix to also except the non-ms format, if we hit a `ValueError`.


patjlm (approved):

fishi0x01:
Proposing fix here https://github.com/dynatrace-oss/api-client-python/pull/110

PR #4816, opened 2025-01-20
[APPSRE-11112] add verify_ondemend_tests for gitlab-housekeeping

BumbleFeng:
allow to check if MR has passed all necessary test jobs and add comments to indicate test results.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -49,6 +49,7 @@ from reconcile.utils.mr.labels import (
     prioritized_approval_label,
 )
 from reconcile.utils.sharding import is_in_shard
+from reconcile.utils.state import State, init_state
 
 MERGE_LABELS_PRIORITY = [
     prioritized_approval_label(p.value) for p in ChangeTypePriority
@@ -70,7 +71,6 @@ HOLD_LABELS = [
 QONTRACT_INTEGRATION = "gitlab-housekeeping"
 DATE_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
 EXPIRATION_DATE_FORMAT = "%Y-%m-%d"
-
 
 merged_merge_requests = Counter(
     name="qontract_reconcile_merged_merge_requests",
@@ -156,16 +156,12 @@ def get_timed_out_pipelines(
 
 def clean_pipelines(
     dry_run: bool,
-    gl_instance: str,
-    gl_project_id: int,
-    gl_settings: str,
+    gl: GitLabApi,
+    fork_project_id: int,
     pipelines: list[dict],
 ) -> None:
     if not dry_run:
-        with GitLabApi(
-            gl_instance, project_id=gl_project_id, settings=gl_settings
-        ) as gl:
-            gl_piplelines = gl.project.pipelines
+        gl_piplelines = gl.get_project_by_id(fork_project_id).pipelines
 
     for p in pipelines:
         logging.info(["canceling", p["web_url"]])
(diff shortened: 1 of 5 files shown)


esron (approved):
LGTM

reconcile/gitlab_housekeeping.py
```diff
@@ -179,6 +180,71 @@ def clean_pipelines(
...
+    running_pipelines = [p for p in pipelines if p["status"] == "running"]
+    if running_pipelines:
+        # wait for pipeline complate
+        return False
+
+    gl_fork = GitLabApi(
+        instance=gl_instance,
+        project_id=mr.source_project_id,
+        settings=gl_settings,
+    )
```
hemslo, line 202: any reason to create a new `GitLabApi` instead of reuse current `gl`, `GitLabApi` instance require `cleanup`.

reconcile/gitlab_housekeeping.py
```diff
@@ -179,6 +180,71 @@ def clean_pipelines(
...
+    commit = next(mr.commits())
+    statuses = gl_fork.project.commits.get(commit.id).statuses.list()
+    test_state = {}
+    for s in statuses:
+        test_state[s.name] = s.status
```
hemslo, line 207: simplify with dict comprehension

```python
test_state = {s.name: s.status for s in statuses}

File: reconcile/gitlab_housekeeping.py (hunk 1 above)
Comment (BumbleFeng) [reply reconcile/gitlab_housekeeping.py]: The original gl targets the upstream repo while this gl_fork targets the fork repo. We can only get commit status form the fork repo.

Comment (hemslo) [reply reconcile/gitlab_housekeeping.py]: you need this part https://github.com/app-sre/qontract-reconcile/blob/dde29ad691e1fd073fe437af071ec3657e5b7097/reconcile/utils/gitlab_api.py#L128-L129, right?

maybe worth extract another method `get_project_by_id`, and replace prevous project_url part with current `get_project`. This can save calls used to init `GitLabApi`, and save those paramter passing of `gl_instance` and `gl_settings`.

Also need to make request counter correct as `commits.get(commit.id)` has 1 request, and  `.statuses.list()` has another (if no pagination), need a better way to count mentioned in https://issues.redhat.com/browse/APPSRE-10450

File: reconcile/test/test_gitlab_housekeeping.py [hunk 3]
```diff
@@ -331,3 +333,122 @@ def test_close_item_without_enable_closing(
...
+    return gl
+
+
+@pytest.fixture
+def state() -> Mock:
+    state = create_autospec(State)
+    return state
+
+
+class StatueMock:
```
rporres, line 366:
```suggestion
class StatusMock:
```

uv.lock
```diff
@@ -1699,7 +1699,7 @@ wheels = [
 
 [[package]]
 name = "qontract-reconcile"
-version = "0.10.1.dev1229"
+version = "0.10.1.dev1220"
```
rporres, line 1702: is this needed?

BumbleFeng: no, add accidentally

reconcile/gitlab_housekeeping.py
```diff
@@ -179,6 +173,64 @@ def clean_pipelines(
...
+    pipelines = gl.get_merge_request_pipelines(mr)
+    running_pipelines = [p for p in pipelines if p["status"] == "running"]
+    if running_pipelines:
+        # wait for pipeline complate
+        return False
+
+    commit = next(mr.commits())
+    fork_project = gl.get_project_by_id(mr.source_project_id)
+    statuses = fork_project.commits.get(commit.id).statuses.list()
+    print(statuses)
```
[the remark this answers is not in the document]
BumbleFeng: yes, removed

reconcile/gitlab_housekeeping.py
BumbleFeng: created `get_project_by_id` for both `verify_on_demand_tests` and `clean_pipelines`.
For the request counter, I don't have a good idea of counting it outside of GitLabApi. Can I leave it to you?

hemslo: sure, I can handle that later

reconcile/gitlab_housekeeping.py
```diff
@@ -156,20 +156,14 @@ def get_timed_out_pipelines(
...
-    if not dry_run:
-        with GitLabApi(
-            gl_instance, project_id=gl_project_id, settings=gl_settings
-        ) as gl:
-            gl_piplelines = gl.project.pipelines
-
     for p in pipelines:
         logging.info(["canceling", p["web_url"]])
         if not dry_run:
+            gl_piplelines = gl.get_project_by_id(fork_project_id).pipelines
```
hemslo, line 166: get project outside of `pipelines` for loop to save gitlab calls. btw can we directly pass `pipeline` object in this function and call `p.cancel()`? Or it's unsupported in this lib?

reconcile/gitlab_housekeeping.py
```diff
@@ -306,7 +362,9 @@ def preprocess_merge_requests(
     dry_run: bool,
     gl: GitLabApi,
     project_merge_requests: list[ProjectMergeRequest],
+    state: State,
     users_allowed_to_label: Iterable[str] | None = None,
+    must_pass: Iterable[str] = [],
```
hemslo, line 367: don't use `[]` as default arg value as it's only eval once and will be shared in all function calls, use `None` or no default value at all

reconcile/gitlab_housekeeping.py
BumbleFeng: We use GitLabApi.get_items() to get pipelines info, so it is a dict result. If we change to return typed object as it mentioned in TO DO, we can call p.cancel() directly.

hemslo (approved):
LGTM!

reconcile/gitlab_housekeeping.py
```diff
@@ -179,6 +175,66 @@ def clean_pipelines(
...
+    mr: ProjectMergeRequest,
+    must_pass: Iterable[str],
+    gl: GitLabApi,
+    state: State,
+) -> bool:
+    """
+    Check if MR has passed all necessary test jobs and add comments to indicate test results.
+    """
+    pipelines = gl.get_merge_request_pipelines(mr)
+    running_pipelines = [p for p in pipelines if p["status"] == "running"]
```
patjlm, line 189: this code will run within a pipeline, right? Shouldn´t we exclude self?

patjlm: ah no, part of gitlab housekeeping, sorry

PR #4818, opened 2025-01-23
jira-permissions-validator: use static sharding

chassing:
Remove unused `per-jira-board` sharding strategy because we want to use `static` instead. And fix sharding support for PR/MR checks.


rporres (approved):

reconcile/jira_permissions_validator.py
```diff
@@ -201,12 +206,19 @@ def validate_boards(
...
+    state: State,
     jira_client_class: type[JiraClient] = JiraClient,
 ) -> bool:
     error = False
     jira_clients: dict[str, JiraClient] = {}
     for board in jira_boards:
+        last_successful_run = state.get(board.name, 0)
+        if not dry_run and time.time() <= last_successful_run + board_check_interval:
+            logging.debug(f"[{board.name}] Skipping board")
+            continue
```
hemslo, line 220: can we add a random jitter time here to avoid many boards time match condition at the same time? Can either random on check time or when saving `stage[board.name]`.

chassing: Addressed in #4823

PR #4819, opened 2025-01-23
ERv2: Support for KMS

lechuk47:
Support KMS in Erv2

### DIFF
--- a/reconcile/external_resources/factories.py
+++ b/reconcile/external_resources/factories.py
@@ -41,13 +41,15 @@ AWS_DEFAULT_TAGS = [
 
 
 class ObjectFactory(Generic[T]):
-    def __init__(self) -> None:
-        self._factories: dict[str, T] = {}
-
-    def register_factory(self, id: str, t: T) -> None:
-        self._factories[id] = t
+    def __init__(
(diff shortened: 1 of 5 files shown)


esron (approved):
LGTM

reconcile/external_resources/factories.py
```diff
@@ -48,6 +48,8 @@ def register_factory(self, id: str, t: T) -> None:
         self._factories[id] = t
 
     def get_factory(self, id: str) -> T:
+        if id not in self._factories and "default" in self._factories:
+            return self._factories["default"]
```
hemslo, line 52: Should we make `default` explicit instead of a hard coded string? For example, `ObjectFactory` register all factories in `__init__`

```python
class ObjectFactory(Generic[T]):
    def __init__(self, factories, default_factory = None) -> None:
        self._factories = factories
        self._default_factory = default_factory

    def get_factory(self, id: str) -> T:
        if id not in self._factories and self._default_factory:
            return self._default_factory
        return self._factories[id]
```

init once and immutable

lechuk47: I like it. Added.

hemslo (approved):
LGTM!

PR #4820, opened 2025-01-23
use qr oc base image 0.3.0 for FIPS

jonmosco:
Bump version of qontract-reconcile-oc for the FIPS build layer.  Depended on https://github.com/app-sre/container-images/pull/176


RH-tj:
@jonmosco its weird that the PR build check has no output under Checks tab above.  Going to ci directly to look for build status/output.

PR #4823, opened 2025-01-24
jira-permissions-validator: add jitter and remove extended early exit

chassing:
* Fix default `board-check-interval` to 120 minutes instead of 120 seconds 
* Add random jitter not to check all boards at the same time 
* Remove extended-early-exit because it collides with per board state cache

### DIFF
--- a/reconcile/jira_permissions_validator.py
+++ b/reconcile/jira_permissions_validator.py
@@ -210,15 +210,12 @@ def validate_boards(
     error = False
     jira_clients: dict[str, JiraClient] = {}
     for board in jira_boards:
-        last_successful_run = state.get(board.name, 0)
-        if (
-            not dry_run
-            and time.time() <= last_successful_run + board_check_interval_sec
-        ):
-            logging.debug(f"[{board.name}] Skipping board")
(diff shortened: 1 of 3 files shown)


reconcile/jira_permissions_validator.py
```diff
@@ -243,8 +242,9 @@ def validate_boards(
                     # no errors
                     logging.debug(f"[{board.name}] is valid")
                     if not dry_run:
-                        # remember time of the last successful run
-                        state[board.name] = time.time()
+                        # remember time of the last successful run and add a random offset
+                        # to avoid all boards being checked at the same time
+                        state[board.name] = time.time() + random.randint(0, 3600)
```
hemslo, line 247: how about make the value to be next run time (`current_time + fixed_interval + random_jit)`, so to determine run or not, just compare current time with saved state time.

hemslo (approved):
LGTM!

reconcile/jira_permissions_validator.py
chassing: You're right; it's easier to understand. Done in d2e6a0e.

PR #4826, opened 2025-01-27
APPSRE-11326 support multiple image waits in saas

fishi0x01:
Saas triggers currently allow us to wait for a single image. However, in Konflux we cannot stuff multiple images into a single pipeline, i.e., we must wait for multiple images to wait.

In this PR we extend the schema for `TriggerSpecContainerImage` to allow a list of images. We introduce the `images` field where tenants can specify a list of images to wait for. We keep the old `image` field for backwards compatibility and will remove it once the new feature is in place.

Note, that for now `image` takes precedence over `images`.

Corresponding schema: https://github.com/app-sre/qontract-schemas/pull/761


esron (approved):
LGTM

reconcile/utils/saasherder/saasherder.py
```diff
@@ -1614,25 +1614,42 @@ def get_container_images_diff_saas_file(
...
-                    image_auth = self._initiate_image_auth(saas_file)
-                    error_prefix = f"[{saas_file.name}/{rt.name}] {target.ref}:"
-                    image = self._get_image(
-                        image_uri, saas_file.image_patterns, image_auth, error_prefix
+
+                    image_registries = []
+                    image_uris = []
+                    all_exist = True
+                    wait_for_images = (
+                        [target.image] if target.image else target.images or []
```
chassing, line 1630: Could you please throw an error somewhere if both `image` and `images` are set?

lisa:
This sort of blurs the line into https://github.com/app-sre/qontract-schemas/pull/761 and I'm curious what "exists" means in the context of waiting for images in the list to exist. Does it mean that the the `image:tag` (or `image@sha`) must exist in the registry? If so, how is that determined - by a `pull` action, or other mechanism?

reconcile/utils/saasherder/saasherder.py
lisa: @chassing I reckon a warning might be better? Something like "Hey, we're using `image` and not `images`"  wdyt?

fishi0x01: Initially just wanted to have `image` to silently take precedence over `images` if both exist. But you are right: lets be more transparent about this and raise that something is odd.

Note, that the goal is to have `image` removed completely once everything is migrated to `images`.

fishi0x01:
> This sort of blurs the line into https://github.com/app-sre/qontract-schemas/pull/761 and I'm curious what "exists" means in the context of waiting for images in the list to exist. Does it mean that the the image:tag (or image@sha) must exist in the registry? If so, how is that determined - by a pull action, or other mechanism?

We use the `_get_image` function of saasherder to validate if an image exists. Exists means the given image uri exists in the registry.

reconcile/utils/saasherder/saasherder.py
fishi0x01: The migration will be swift - will try with tenant first who needs this feature. Once it works for them, I will move all other targets in app-interface via migration script.

chassing: Sounds good. Please add me with Glitchtip to the earlier adopters list.

reconcile/utils/saasherder/saasherder.py
```diff
@@ -1616,6 +1620,10 @@ def get_container_images_diff_saas_file(
                 try:
                     if not (target.image or target.images):
                         continue
+                    if target.image and target.images:
+                        raise TriggerSpecContainerImageError(
+                            '"image" and and "images" are mutually exclusive. Do not set both in the same saas target.'
```
chassing, line 1625: There is a typo "... and and ..."

chassing (approved):
Just a small typo; otherwise LGTM. Nice improvement :heart:

reconcile/utils/saasherder/saasherder.py
fishi0x01: Ups good catch. Fixed :+1:

reconcile/utils/saasherder/saasherder.py
hemslo: can we also use `oneOf` in schema to ensure no `image` and `images` set at the same time?

reconcile/utils/saasherder/models.py
```diff
@@ -113,14 +113,15 @@ def state_key(self) -> str:
 
 @dataclass
 class TriggerSpecContainerImage(TriggerSpecBase):
-    image: str
+    images: Sequence[str]
     reason: str | None = None
 
     @property
     def state_key(self) -> str:
+        image_key = ",".join(self.images)
```
hemslo, line 121: should `images` be sorted to avoid order change affect state key?

reconcile/utils/saasherder/saasherder.py
```diff
@@ -1614,25 +1618,46 @@ def get_container_images_diff_saas_file(
...
+                            error_prefix,
+                        )
+                        if not image:
+                            all_exist = False
+                            break
+                        image_registries.append(image_registry)
+                        image_uris.append(image_uri)
+
+                    if not all_exist:
+                        continue
```
hemslo, line 1660: the loop is too deep with nested `break` and `continue`, consider simplify it by reordering statements and put duplicate calculations outside of loop

```python
                    image_auth = self._initiate_image_auth(saas_file) # this can be done once per saas file when target image required
                    image_registries = [
                        f"{image.org.instance.url}/{image.org.name}/{image.name}"
                        for image in target.images or [target.image]
                    ]
                    error_prefix = f"[{saas_file.name}/{rt.name}] {target.ref}:"
                    if not all(
                        self._get_image(
                            image=f"{image}:{desired_image_tag}",
                            image_patterns=saas_file.image_patterns,
                            image_auth=image_auth,
                            error_prefix=error_prefix,
                        )
                        for image in image_registries
                    ):
                        continue
```

```python
                    if self.include_trigger_trace:
                        image_uris = ', '.join(f"{image}:{desired_image_tag}" for image in image_registries)
                        trigger_spec.reason = f"{rt.url}/commit/{commit_sha} build {image_uris}"
```

reconcile/utils/saasherder/models.py
fishi0x01: Good catch! Fixed

reconcile/utils/saasherder/saasherder.py
fishi0x01: Sounds good. Adjusted now :+1:

reconcile/utils/saasherder/saasherder.py
fishi0x01: For now its easier for me to just raise an exception here, as I dont see a smooth way of adding `oneOf` without duplicating the whole target object in the schema :thinking: 
This is also just a temporary co-existence during the migration time - I will remove `image` and the exception once migration is done.

reconcile/utils/saasherder/models.py
hemslo: `images` declared twice, `images: Sequence[str]` and `@property` both set it, can just sort on the fly inside `def state_key`, shouldn't be slow since usually small images, and you changed separator from `,` to `/`?

hemslo: oh, that duplicate already fixed, just confirm about `,` vs `/` then

fishi0x01: Hmm I think you are on an older commit. Most recent shouldnt have property anylonger

reconcile/utils/saasherder/saasherder.py
fishi0x01: Was thinking about it, but I dont see any bigger gain here. Would require an extra null check plus raise in the loop if I put it outside. The image auth doesnt do any calls - its just an object iiu.

reconcile/utils/saasherder/models.py
fishi0x01: Yeah `/` vs `,` is intentional

reconcile/utils/saasherder/saasherder.py
hemslo: it can read secrets from vault https://github.com/app-sre/qontract-reconcile/blob/62f69467cb6d57685d85d59a48dcdad51928f6a9/reconcile/utils/saasherder/saasherder.py#L1213

hemslo: `_image_auth_cache` is being used in multiple threads, not thread safe. better to use [@functools.cache](https://docs.python.org/3/library/functools.html#functools.cache) for thread safe cache, note need to wrap it as instance level cache because this is a method instead of function, [example](https://github.com/app-sre/qontract-reconcile/blob/037953d30fade55984da5491e0aef79927e8dce5/reconcile/dashdotdb_dora.py#L250).

And further look into `_initiate_image_auth`, it only requires `saas_file.authentication.image` as input, so the same auth can be used in multiple saas files if they are using the same secret.

`_initiate_image_auth` is also used in [openshift_saas_deploy](https://github.com/app-sre/qontract-reconcile/blob/037953d30fade55984da5491e0aef79927e8dce5/reconcile/openshift_saas_deploy.py#L344), so the method can be public without `_`.

```python
# __init__()
    self._initiate_image_auth = functools.cache(self._init_image_auth)
```

```python
def initiate_image_auth(self, saas_file):
    return self._initiate_image_auth(saas_file.authentication and saas_file.authentication.image)
```

```python
def _init_image_auth(self, image):
    # old _initiate_image_auth logic
```

note `image` need to be hashable, may need to wrap `saas_file.authentication.image` as immutable dataclass or basemodel.

fishi0x01: Hmm I think Ill just revert and use it in loop as before. This starts to go beyond what I would like to change - I dont want to refactor too much in saasherder now.

fishi0x01: Reverted and using in loop now. This aligns with existing behavior now.

We can do the proposed _image_auth optimization in follow-up MRs outside of this effort.

hemslo: yes, we need metrics data to verify if that change can actually save vault calls

reconcile/utils/saasherder/saasherder.py
```diff
@@ -1615,25 +1619,36 @@ def get_container_images_diff_saas_file(
...
                         github=github,
                     )
-                    desired_image_tag = commit_sha[: rt.hash_length or self.hash_length]
-                    # don't trigger if image doesn't exist
-                    image_registry = f"{target.image.org.instance.url}/{target.image.org.name}/{target.image.name}"
-                    image_uri = f"{image_registry}:{desired_image_tag}"
                     image_auth = self._initiate_image_auth(saas_file)
+                    desired_image_tag = commit_sha[: rt.hash_length or self.hash_length]
+
+                    all_images = [target.image] if target.image else target.images or []
```
hemslo, line 1636: since already verified `image` and `images` can only have one set, can simplify `if else or` with

```python
    all_images = target.images or [target.image]
```

hemslo (approved):
LGTM!

rporres (approved):
This is a great improvement, and swiftly made! <3

PR #4830, opened 2025-01-28
Erv2: Add resources to ERv2 modules

lechuk47:
- **Add Resources to Erv2 modules**: The main reason is to schedule pods as burstable instead of best-effort. 
- **Expose modules RECONCILE_TIMEOUT as an env variable**: Need to handle specific Tf object timeouts like aws_db_instance timeouts. 

Requires:
- https://github.com/app-sre/qontract-schemas/pull/764


reconcile/external_resources/model.py
```diff
@@ -220,13 +223,46 @@ def load_module_inventory(
...
+        DeployResourceFields can not be used directly as it not hashable."""
+        return Resources(
+            requests=ResourcesSpec(
+                cpu=fields.requests.cpu or "", memory=fields.requests.memory or ""
+            ),
+            limits=ResourcesSpec(
+                cpu=fields.limits.cpu or "",
+                memory=fields.limits.memory or "",
+            ),
+        )
```
chassing, line 255: I don't understand the default settings for `ResourcesSpec.cpu`, `ResourcesSpec.memory`, `Resources.requests`, `Resources.limits` and everything in `from_deploy_resources_fields`. Are the defaults `""` or 

```
requests: ResourcesSpec = ResourcesSpec(cpu="100m", memory="128Mi")
limits: ResourcesSpec = ResourcesSpec(memory="4Gi")
```

lechuk47: I changed this code. PTAL.

chassing: Now, it's much easier to understand. Thank you 🙇

chassing (approved):

PR #4832, opened 2025-01-29
:coffin: terraform-resources: remove MSK and ElastiCache

chassing:
MSK and ElastiCache have been moved to ERv2.

Ticket: [APPSRE-11457](https://issues.redhat.com/browse/APPSRE-11457)


rporres (approved):
Most likely out of the scope for this. But since we're removing the old tf-resources support for these resources, can we also remove the `managed_by_erv2` for them?

chassing:
> Most likely out of the scope here. But since we're removing the old tf-resources support for these resources, can we also remove the `managed_by_erv2` for them?

Sure, that's out of scope here. Let's postpone it to a later time.

rporres:
> > Most likely out of the scope here. But since we're removing the old tf-resources support for these resources, can we also remove the `managed_by_erv2` for them?
> 
> Sure, that's out of scope here. Let's postpone it to a later time.

That makes sense. LGTM!

bkez322:
I'm requesting this be reverted, if reasonable.

https://redhat-internal.slack.com/archives/GGC2A0MS8/p1738262418245609?thread_ts=1738257457.026369&cid=GGC2A0MS8

PR #4834, opened 2025-01-29
add state setting to app_interface_merge_queue

BumbleFeng:


patjlm:
Please alwayy state why you're implementing changes, including if possible a link to a ticket. It's not easy for someone who did not follow your train of thoughts to understand why you are pushing this kind of change.

PR #4835, opened 2025-01-30
:bug: honor jira board cache if dry-run mode runs over all boards

chassing:
Typically, the MR checks run only for the new/updated Jira board. Use the Jira board cache in case of running the MR checks for all boards, e.g., during `qontract-reconcile` promotions.

This will avoid hitting the Jira rate limit during our promotions.

### DIFF
--- a/reconcile/jira_permissions_validator.py
+++ b/reconcile/jira_permissions_validator.py
@@ -217,7 +217,7 @@ def validate_boards(
                 continue
             # dry-run mode
             elif len(jira_boards) > 1:
-                # skip for MR checks if there are more than one board to test
+                logging.info(f"[{board.name}] Use cache results. Skipping ...")
                 continue
 
         logging.debug(f"[{board.name}] checking ...")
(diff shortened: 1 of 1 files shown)


reconcile/jira_permissions_validator.py
```diff
@@ -211,9 +211,14 @@ def validate_boards(
...
-            logging.debug(f"[{board.name}] Skipping board")
-            continue
+        if time.time() <= next_run_time:
+            if not dry_run:
+                # always skip for non-dry-run mode
+                continue
+            # dry-run mode
+            elif len(jira_boards) > 1:
+                # skip for MR checks if there are more than one board to test
+                continue
```
hemslo, line 221: do you want to log something here to know the check is skipped?

hemslo (approved):
LGTM!

reconcile/jira_permissions_validator.py
chassing: Good idea. Added in b438c01

PR #4836, opened 2025-01-30
Erv2: Validation improvements

lechuk47:
* Include the module_configuration in the resources resolve and validate methods. 
* Validate RDS timeouts / set default timeouts based on the module spec.

### DIFF
--- a/reconcile/external_resources/aws.py
+++ b/reconcile/external_resources/aws.py
@@ -1,9 +1,11 @@
+import re
 from abc import ABC, abstractmethod
 from typing import Any
 
 from reconcile.external_resources.model import (
     ExternalResource,
     ExternalResourceKey,
+    ExternalResourceModuleConfiguration,
     ExternalResourcesInventory,
 )
 from reconcile.utils.external_resource_spec import (
@@ -21,10 +23,18 @@ class AWSResourceFactory(ABC):
         self.secret_reader = secret_reader
 
     @abstractmethod
-    def resolve(self, spec: ExternalResourceSpec) -> dict[str, Any]: ...
+    def resolve(
+        self,
+        spec: ExternalResourceSpec,
+        module_conf: ExternalResourceModuleConfiguration,
+    ) -> dict[str, Any]: ...
 
     @abstractmethod
-    def validate(self, resource: ExternalResource) -> None: ...
+    def validate(
+        self,
+        resource: ExternalResource,
+        module_conf: ExternalResourceModuleConfiguration,
+    ) -> None: ...
 
     def find_linked_resources(
         self, spec: ExternalResourceSpec
@@ -35,10 +45,18 @@ class AWSResourceFactory(ABC):
 
 
 class AWSDefaultResourceFactory(AWSResourceFactory):
-    def resolve(self, spec: ExternalResourceSpec) -> dict[str, Any]:
+    def resolve(
+        self,
+        spec: ExternalResourceSpec,
+        module_conf: ExternalResourceModuleConfiguration,
+    ) -> dict[str, Any]:
         return ResourceValueResolver(spec=spec, identifier_as_value=True).resolve()
 
-    def validate(self, resource: ExternalResource) -> None: ...
+    def validate(
+        self,
+        resource: ExternalResource,
+        module_conf: ExternalResourceModuleConfiguration,
+    ) -> None: ...
 
 
 class AWSElasticacheFactory(AWSDefaultResourceFactory):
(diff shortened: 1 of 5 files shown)


reconcile/external_resources/aws.py
```diff
@@ -142,9 +175,53 @@ def resolve(self, spec: ExternalResourceSpec) -> dict[str, Any]:
...
+    def _get_timeout_value(
+        self,
+        timeout: str,
+    ) -> int:
+        if not re.match(AWSRdsFactory.TIMEOUT_RE, timeout):
+            raise ValueError(
+                f"Invalid RDS instance timeout format: {timeout}. Specify timeout in minutes(m) or hours(h)."
+            )
+        value, unit = timeout[:-1], timeout[-1]
+        return int(value) * AWSRdsFactory.TIMEOUT_UNIT_CONVERSION[unit]
```
hemslo, line 197: how about we only support `m` to simplify checks, we don't support full golang time duration format anyway. defaults are documented as `m` too https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance#timeouts

and since the return is minutes, better to rename it to `_get_timeout_minutes`

reconcile/external_resources/aws.py
```diff
@@ -142,9 +175,53 @@ def resolve(self, spec: ExternalResourceSpec) -> dict[str, Any]:
...
+        for option in ("create", "update", "delete"):
+            timeout = timeouts.get(option)
+            if (
+                timeout
+                and (value := self._get_timeout_value(timeout))
+                > module_conf.reconcile_timeout_minutes
+            ):
+                raise ValueError(
+                    f"RDS instance {option} timeout value {value} exceeds the module reconcile_timeout_minutes value {module_conf.reconcile_timeout_minutes}."
+                )
```
hemslo, line 217: also need to ensure `timeouts` to be a `dict` with only allowed keys (`create`, `update`, `delete`), either use a new BaseModel or manually verify like

```python
        if not isinstance(timeouts, dict):
            raise ValueError()

        allowed_keys = {"create", "update", "delete"}
        if unknown_keys := timeouts.keys() - allowed_keys:
            raise ValueError()

        for option, timeout in timeouts.items():
            timeout_minutes = self._get_timeout_minutes(timeout)
            if timeout_minutes > module_conf.reconcile_timeout_minutes:
                raise ValueError()
```

reconcile/test/external_resources/test_rds_timeouts.py
```diff
@@ -0,0 +1,39 @@
...
+def test_timeout_values() -> None:
+    f = AWSRdsFactory(er_inventory=Mock(), secret_reader=Mock())
+
+    assert f._get_timeout_value("60m") == 60
+    assert f._get_timeout_value("1h") == 60
+    with pytest.raises(
+        ValueError,
+        match=r".*Invalid RDS instance timeout format: 12mm. Specify timeout in minutes\(m\) or hours\(h\)..*",
+    ):
+        assert f._get_timeout_value("12mm")
```
hemslo, line 22: don't test private method, can make it a public static method or a helper function

reconcile/test/external_resources/test_rds_timeouts.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+    module_conf = ExternalResourceModuleConfiguration(reconcile_timeout_minutes=120)
+    f._validate_timeouts(resource, module_conf)
+
+    module_conf = ExternalResourceModuleConfiguration(reconcile_timeout_minutes=40)
+    with pytest.raises(
+        ValueError,
+        match=r".*RDS instance create timeout value 60 exceeds the module reconcile_timeout_minutes value 40.*",
+    ):
+        f._validate_timeouts(resource, module_conf)
```
hemslo, line 39: don't test private method, make it public or test `f.validate` instead.

reconcile/external_resources/aws.py
lechuk47: Done.

reconcile/external_resources/aws.py
lechuk47: Done.

reconcile/test/external_resources/test_rds_timeouts.py
lechuk47: Changed the test methods to use validate() instead

reconcile/test/external_resources/test_rds_timeouts.py
lechuk47: Changed the test methods to use validate() instead

PR #4841, opened 2025-02-03
:label: more type hints 02

chassing:
Add and enable type hints for `reconcile.utils`


reconcile/utils/aws_api.py
```diff
@@ -637,11 +897,9 @@ def delete_keys(
...
                 if not user_and_user_keys:
                     continue
-                # unpack single item from sequence
-                # since only a single user can have a given key
-                [user_and_user_keys] = user_and_user_keys
-                user = user_and_user_keys[0]
-                user_keys = user_and_user_keys[1]
+
+                user = user_and_user_keys[0][0]
+                user_keys = user_and_user_keys[0][1]
```
hemslo, line 902: there was an implicit check that `user_and_user_keys` has single item

reconcile/utils/vcs.py
```diff
@@ -222,9 +222,11 @@ def get_file_content_from_app_interface_master(self, file_path: str) -> str:
             if not file_path.startswith("data")
             else file_path
         )
-        return self._app_interface_api.project.files.get(
-            file_path=file_path, ref="master"
-        ).decode()
+        return (
+            self._app_interface_api.project.files.get(file_path=file_path, ref="master")
+            .decode()
+            .decode("utf-8")
```
hemslo, line 228: this a breaking change, the first `decode()` returns `byte`, now this method returns `str`, was `byte` expected before?

reconcile/utils/aws_api.py
chassing: I've added an explicit check in [f446ead](https://github.com/app-sre/qontract-reconcile/pull/4841/commits/f446eadf03c2ec694c0f115f5d51db4b51d61812)

hemslo (approved):
LGTM!

PR #4843, opened 2025-02-03
APPSRE-11390 add fleet labeler query

fishi0x01:
Boilerplate for querying new fleet label schema.

Dependency: typo fix https://github.com/app-sre/qontract-schemas/pull/767


reconcile/typed_queries/fleet_labels.py
```diff
@@ -0,0 +1,14 @@
...
+from reconcile.utils import gql
+from reconcile.utils.gql import GqlApi
+
+
+def get_fleet_labels(
+    api: GqlApi | None = None,
+) -> list[FleetLabelsV1]:
+    api = api or gql.get_api()
+    data = query(api.query)
+    return list(data.fleet_labels or [])
```
hemslo, line 14: why need to create another `list` instead of directly `data.fleet_labels or []`?

fishi0x01: Hmm good question. Its a pattern we have in many typed queries - copy/paste. Fixed here now

hemslo (approved):
LGTM!

PR #4846, opened 2025-02-05
APPSRE-11390 initial fleet labeler integration

fishi0x01:
Depends on https://github.com/app-sre/qontract-schemas/pull/769


reconcile/fleet_labeler/dependencies.py
```diff
@@ -0,0 +1,67 @@
...
+        self.secret_reader = secret_reader
+        self.label_specs_by_name: dict[str, FleetLabelsSpecV1] = {}
+        self.ocm_clients_by_label_spec_name: dict[str, OCMClient] = {}
+        self.vcs: VCS | None = None
+        self.dry_run = dry_run
+
+    def populate_all(self) -> None:
+        self.populate_label_specs()
+        self.populate_ocm_clients()
+        self.populate_vcs()
```
hemslo, line 35: it's not clear which fields are used, scan the code, it seems only `label_specs_by_name`, `ocm_clients_by_label_spec_name` and `vcs`. To make it more explicit, can make `Dependencies` fully DI, the `__init__` should be

```python
    def __init__(
        self,
        label_specs_by_name,
        ocm_clients_by_label_spec_name,
        vcs,
    ):
        self.label_specs_by_name = label_specs_by_name
        self.ocm_clients_by_label_spec_name = ocm_clients_by_label_spec_name
        self.vcs = vcs
```

and with a class method or static method to create instance

```python
    @classmethod
    def create(
        cls,
        dry_run: bool = True,
    ) -> Self:
        xxx
        return cls(
            label_specs_by_name=label_specs_by_name,
            ocm_clients_by_label_spec_name=ocm_clients_by_label_spec_name,
            vcs=vcs,
        )
```

Or use `BaseModel`.

The reason is bad smell in current usage, user of `Dependencies` shouldn't need to call `populate_all` to do init.

```python
        dependencies = Dependencies(
            secret_reader=self.secret_reader,
            dry_run=dry_run,
        )
        dependencies.populate_all()
```

reconcile/fleet_labeler/integration.py
```diff
@@ -0,0 +1,151 @@
...
+        all_desired_cluster_ids: list[str] = []
+        clusters_to_add: list[YamlCluster] = []
+        for label_default in spec.label_defaults:
+            discovered_clusters_by_id = {
+                cluster.cluster_id: cluster
+                for cluster in ocm.discover_clusters_by_label_keys(
+                    keys=list(dict(label_default.match_subscription_labels).keys())
+                )
+                if dict(label_default.match_subscription_labels).items()
+                <= cluster.subscription_labels.items()
```
hemslo, line 112: what is comparing here dict items `<=` ?

reconcile/fleet_labeler/integration.py
```diff
@@ -0,0 +1,151 @@
...
+            }
+            diff: DiffResult[str, str, str] = diff_iterables(
+                current=all_current_cluster_ids,
+                desired=discovered_clusters_by_id.keys(),
+            )
+            all_desired_cluster_ids.extend(discovered_clusters_by_id.keys())
```
hemslo, line 118: can there be duplicate id in `all_desired_cluster_ids`?

reconcile/fleet_labeler/integration.py
```diff
@@ -0,0 +1,151 @@
...
+                for cluster in ocm.discover_clusters_by_label_keys(
+                    keys=list(dict(label_default.match_subscription_labels).keys())
+                )
+                if dict(label_default.match_subscription_labels).items()
+                <= cluster.subscription_labels.items()
+            }
+            diff: DiffResult[str, str, str] = diff_iterables(
+                current=all_current_cluster_ids,
+                desired=discovered_clusters_by_id.keys(),
+            )
```
hemslo, line 117: is there a difference if we do diff once in the later stage

```python
        diff = diff_iterables(
            current=all_current_cluster_ids,
            desired=all_desired_cluster_ids,
        )
```

and process both `add` and `delete`?

reconcile/fleet_labeler/validate.py
```diff
@@ -0,0 +1,66 @@
...
+def _validate_match_labels(spec: FleetLabelsSpecV1) -> None:
+    """
+    Match labels should be unique within the same spec.
+    """
+
+    seen: set[str] = set()
+    for label_default in spec.label_defaults:
+        key = ""
+        for k, v in sorted(dict(label_default.match_subscription_labels).items()):
+            key += f"{k}:{v},"
```
hemslo, line 43: avoid `+=` str by `join`

```python
key = ",".join(f"{k}:{v}" for k, v in sorted(dict(label_default.match_subscription_labels).items()))
```

reconcile/fleet_labeler/validate.py
```diff
@@ -0,0 +1,66 @@
...
+        if key in seen:
+            raise MatchLabelsNotUniqueError(
+                f"The 'matchSubscriptionLabels' combinations must be unique within a spec. Found duplicate: {label_default.match_subscription_labels}"
+            )
+        seen.add(key)
```
hemslo, line 48: if you need to gather all duplicates (more than one), can use `Counter`.

reconcile/test/fleet_labeler/test_fleet_labeler_cluster_inventory_lifecycle.py
```diff
@@ -0,0 +1,197 @@
...
+                    cluster_id="cluster_id",
+                    subscription_labels={"sre-capabilities.dtp.managed-labels": "true"},
+                ),
+            ],
+        )
+    }
+
+    integration.reconcile(dependencies=dependencies)
+
+    dependencies.vcs.open_merge_request.assert_called_once_with(  # type: ignore[attr-defined]
```
hemslo, line 49: to avoid type ignore, you can make `build_vcs` return type as `Mock`, then save it as local var, then verify with `vcs.xxxx.asssert`

reconcile/fleet_labeler/integration.py
fishi0x01: This compares whether match_subscription_labels is a subset of the clusters subscription_labels

reconcile/fleet_labeler/integration.py
fishi0x01: I want to create `YamlCluster` which requires a label fetch and some data specific from the current loop.

I could store that in some id dict or just do the diff here

reconcile/fleet_labeler/integration.py
hemslo: can it be done in server side label filter? or it's always `OR` condition match in server and we have to do another client side filter?

reconcile/fleet_labeler/integration.py
fishi0x01: Hmm very good point. It is theoretically possible that multiple label filters target the same cluster - though that would actually be an issue as then label defaults are competing with each other. I.e., such a case is definitely undesired by the user.

I believe in that case we should not process the cluster and instead write an error to log and increase some error counter metric - so we can have an alert on fleet labeler if something like that happened - while not blocking the integration run itself.

reconcile/fleet_labeler/integration.py
hemslo: maybe also related to duplicate cluster id question, can the same cluster returned in different iterations? Do it in each iteration may result duplicate `clusters_to_add`. If we are sure about uniqueness, then diff on either levels are fine.

reconcile/fleet_labeler/integration.py
fishi0x01: Ah subscription_filter actually has an `eq` method too! Yeah will use that!

reconcile/fleet_labeler/integration.py
fishi0x01: Yeah diff is fine. I just didnt want to do more foo with id mappings in dicts xD
For me it was just easier to diff here - my assumption was that the overhead is ok here.

reconcile/fleet_labeler/dependencies.py
fishi0x01: Yes very good point!
Fixed now - this also relieves us of an extra None check :)

reconcile/fleet_labeler/integration.py
fishi0x01: Tried the `eq` filter, but it throws a 400 on OCM side. The or filter works well, so keeping it for now. Will try to figure out how to enhance this with eq filter for server side filtering.

reconcile/fleet_labeler/validate.py
fishi0x01: Fixed :+1:

reconcile/fleet_labeler/validate.py
fishi0x01: Using `Counter` now :+1:

reconcile/fleet_labeler/dependencies.py
hemslo: `dry_run` and `secret_reader` used anywhere after `Dependencies` init? seems only for `vcs` init

reconcile/fleet_labeler/validate.py
hemslo: the new code still only throw for one duplicate, if using `Counter` then you don't have to manually `+= 1` count, get all results like this

```python
    keys = (
        ".".join(
            f"{k}:{v}"
            for k, v in sorted(dict(label_default.match_subscription_labels).items())
        )
        for label_default in spec.label_defaults
    )
    duplicates = [key for key, count in Counter(keys).items() if count > 1]
    if duplicates:
        raise MatchLabelsNotUniqueError(
            f"The 'matchSubscriptionLabels' combinations must be unique within a spec. Found duplicates: {duplicates}"
        )

```

reconcile/fleet_labeler/integration.py
hemslo: on second look at this validation, if you diff 2 simple set, you don't need `diff_iterables`, just do set operations is simpler and enough.

```python
        all_current_cluster_ids = {cluster.cluster_id for cluster in spec.clusters}
        all_desired_cluster_ids = set()
        ...
            if duplicates := discovered_clusters_by_id.keys() & all_desired_cluster_ids:
                raise ValueError(xxx)
            all_desired_cluster_ids.update(discovered_clusters_by_id.keys())
            for cluster_id in discovered_clusters_by_id.keys() - all_current_cluster_ids:
                cluster = discovered_clusters_by_id[cluster_id]
            ...
        cluster_ids_to_delete = all_current_cluster_ids - all_desired_cluster_ids
```

reconcile/fleet_labeler/validate.py
fishi0x01: Yeah I wanted to print the original json version before the `.join` operation. So I decided to against list comprehension here to keep current var context.

reconcile/fleet_labeler/integration.py
fishi0x01: haha yeah good point. Fixed :+1:

reconcile/fleet_labeler/validate.py
fishi0x01: Ah ok you want to print all offending label matchers. Good point!

reconcile/fleet_labeler/integration.py
hemslo: please also change `clusters_to_add` part since it's simple id set compare too

reconcile/fleet_labeler/validate.py
fishi0x01: Fixed

reconcile/fleet_labeler/dependencies.py
hemslo: please leave a comment on these 2 fields so we remember to check usage in next iteration

reconcile/fleet_labeler/integration.py
fishi0x01: Fixed

reconcile/fleet_labeler/dependencies.py
hemslo: I usually only add fields needed, if added too much early on, it's very hard to do cleanup later, leave many unused code

fishi0x01: I get that. But I think given the current code this is very managable.

fishi0x01: Added the comment

reconcile/fleet_labeler/integration.py
hemslo: need type change on `all_current_cluster_ids` and `all_desired_cluster_ids` to be `set`, check example code in https://github.com/app-sre/qontract-reconcile/pull/4846/files#r1957522739, validation part can be added later if not needed now.

fishi0x01: Ok sorry I will focus now on the error metric - will do nits later now

hemslo: after that set change, those vars can be used directly in later check without set convert for `cluster_ids_to_delete`

hemslo: Nice!
With new logic of duplicate handling, if the duplicate cluster already exist in current spec, it will be added to `cluster_ids_to_delete`, is it expected?
Also this deduplicate logic adds some complexity (multiple mutations on different sets and lists), to know details about which `match_subscription_labels` are causing duplicate, we can collect them all and do logging. Flow is like this:

* discover all clusters
* find duplicates and log details
* diff to build `clusters_to_add` and `cluster_ids_to_delete`

With an extra data structure to hold cluster and `label_default`, the whole process can be simplified with


```python
class ClusterData(BaseModel):
    cluster: Cluster
    label_default: FleetLabelDefaultV1
```

```python
        all_current_cluster_ids = {cluster.cluster_id for cluster in spec.clusters}

        clusters = defaultdict(list)
        for label_default in spec.label_defaults:
            match_subscription_labels = dict(label_default.match_subscription_labels)
            for cluster in ocm.discover_clusters_by_labels(
                labels=match_subscription_labels
            ):
                # TODO: ideally we filter on server side - see TODO in ocm.py
                if (
                    match_subscription_labels.items()
                    <= cluster.subscription_labels.items()
                ):
                    clusters[cluster.cluster_id].append(
                        ClusterData(cluster=cluster, label_default=label_default)
                    )

        cluster_with_duplicate_matches = {
            k: v for k, v in clusters.items() if len(v) > 1
        }
        for cluster_id, matches in cluster_with_duplicate_matches.items():
            label_matches = "\n".join(
                str(m.label_default.match_subscription_labels) for m in matches
            )
            logging.error(
                f"Spec '{spec.name}': Cluster ID {cluster_id} is matched multiple times by different label matchers:\n{label_matches}"
            )
        metrics.set_gauge(
            FleetLabelerDuplicateClusterMatchesGauge(
                integration=self.name,
                ocm_name=spec.ocm.name,
            ),
            len(cluster_with_duplicate_matches),
        )

        all_desired_clusters = {k: v[0] for k, v in clusters.items() if len(v) == 1}
        clusters_to_add = [
            YamlCluster(
                cluster_id=cluster_id,
                name=cluster_info.cluster.name,
                server_url=cluster_info.cluster.server_url,
                subscription_labels_content=self._render_default_labels(
                    template=cluster_info.label_default.subscription_label_template,
                    labels=ocm.get_cluster_labels(cluster_id=cluster_id),
                ),
            )
            for cluster_id, cluster_info in all_desired_clusters.items()
            if cluster_id not in all_current_cluster_ids
        ]
        cluster_ids_to_delete = all_current_cluster_ids - all_desired_clusters.keys()
```

fishi0x01: Alright added :+1:

hemslo (approved):
LGTM!

PR #4847, opened 2025-02-05
Add a qontract-cli command for getting logGroup usage from CloudWatch

cubismod:
[APPSRE-11480](https://issues.redhat.com/browse/APPSRE-11480)

I tried fetching log group data via CloudWatch metrics but ran into limitations due to the vast number of log groups we have in our accounts. This is another approach which fetches the data from the AWS API and outputs it with the command.


esron (approved):
LGTM

PR #4854, opened 2025-02-06
:bug: Add policy to s3 bucket for access logs

esron:
This should fix the deployment of access logs for ALB by adding a policy attached to the bucket ans instructed by AWS docs:

https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html#access-log-create-bucket 

I've tested this locally using dry run and printing to file.


uv.lock
BumbleFeng, line 1: This is not needed, right?

BumbleFeng (approved):
LGTM

uv.lock
esron: True.

PR #4857, opened 2025-02-07
External resources logs dashboard: sort log lines

patjlm:
The log lines are sometimes mixed


patjlm:
> It should not happen

I did see it though, for a log that was showing errors. And adding this fixed it.

It's possible this sorting option only considers microseconds, as displayed, but cloudwatch goes to nano, maybe?

hemslo (approved):
LGTM! Only `sort @timestamp asc` can ensure nature ordered, mixed order for `sort @timestamp desc` due to same `@timestamp`.

PR #4860, opened 2025-02-10
:zap: external-resource: specify memory limits for outputs-secret container

chassing:
We want to introduce memory-based quotas to avoid killing the k8s node. If needed, we can replace the hardcoded values with app-interface settings in the future

```
Error creating: pods "er-dry-run-mr-132420-02db23c11d-qhk6p" is forbidden: failed quota: external-resources-quota: must specify limits.memory for: outputs; requests.memory for: outputs
```

Ticket: [APPSRE-11514](https://issues.redhat.com/browse/APPSRE-11514)


hemslo (approved):
LGTM!

PR #4861, opened 2025-02-10
Select the correct policy for access_logs bucket based on region

esron:
The ELB account id map comes from AWS documentation:
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html#access-log-create-bucket

Depending on the region we need to select the `"Principal"` for the policy.

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -270,6 +270,36 @@ DEFAULT_TAGS = {
     },
 }
 
+AWS_ELB_ACCOUNT_IDS = {
+    "us-east-1": "127311923021",
+    "us-east-2": "033677994240",
+    "us-west-1": "027434742980",
+    "us-west-2": "797873946194",
+    "af-south-1": "098369216593",
+    "ap-east-1": "754344448648",
+    "ap-southeast-3": "589379963580",
+    "ap-south-1": "718504428378",
+    "ap-northeast-3": "383597477331",
+    "ap-northeast-2": "600734575887",
+    "ap-southeast-1": "114774131450",
+    "ap-southeast-2": "783225319266",
+    "ap-northeast-1": "582318560864",
+    "ca-central-1": "985666609251",
+    "eu-central-1": "054676820928",
+    "eu-west-1": "156460612806",
+    "eu-west-2": "652711504416",
+    "eu-south-1": "635631232127",
+    "eu-west-3": "009996457667",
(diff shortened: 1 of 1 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -5148,6 +5148,73 @@ def _get_alb_rule_condition_value(condition):
...
+            "eu-south-1",
+            "eu-west-3",
+            "eu-north-1",
+            "me-south-1",
+            "sa-east-1",
+        }:
+            return {"AWS": f"arn:aws:iam::{elb_account_id}:root"}
+        if region in {"us-gov-west-1", "us-gov-east-1"}:
+            return {"AWS": f"arn:aws-us-gov:iam::{elb_account_id}:root"}
+        return {"Service": "logdelivery.elasticloadbalancing.amazonaws.com"}
```
hemslo, line 5216: to avoid duplicate region names in 2 places, can extract mapping to 2 constants, then just get principal by region

```python
AWS_ELB_ACCOUNT_IDS = {
  "us-east-1": "127311923021",
  ...
}

AWS_US_GOV_ELB_ACCOUNT_IDS = {
  "us-gov-west-1": "048591011584",
  "us-gov-east-1": "190560391635",
}

    @staticmethod
    def _get_principal_for_s3_bucket_policy(region: str) -> Mapping[str, str]:
        if region in AWS_ELB_ACCOUNT_IDS:
            return {"AWS": f"arn:aws:iam::{AWS_ELB_ACCOUNT_IDS[region]}:root"}
        if region in AWS_US_GOV_ELB_ACCOUNT_IDS:
            return {"AWS": f"arn:aws-us-gov:iam::{AWS_US_GOV_ELB_ACCOUNT_IDS[region]}:root"}
        return {"Service": "logdelivery.elasticloadbalancing.amazonaws.com"}
```

esron: Done, thanks for the tip.

cubismod (approved):
LGTM despite the AWS API being horrifically designed with hardcoded IDs

reconcile/utils/terrascript_aws_client.py
hemslo: @esron please check again and test it, you are using a non existing method `_get_elb_account_id`, please delete `elb_account_id = self._get_elb_account_id(region)` and do everything in `principal = self._get_principal_for_s3_bucket_policy(region)`. I'll revert this change for now

PR #4867, opened 2025-02-13
Improve error handling for missing registry credentials

jonmosco:
When switching images to another repository, the following error occurred:  

```
****/service/[REDACTED]/blob/[COMMIT_HASH]/[REDACTED].yaml:  
Image : registry.[REDACTED].com/[REDACTED]/vault:[VERSION] does not exist  
```  

The issue is not that the image is missing but that the registry is not included in the pull-secret. This causes authentication to fail, resulting in a misleading "image not found" error.  

This update does not introduce any functional changes—it only improves error messages to better distinguish between a missing image and an inaccessible registry.

### DIFF
--- a/reconcile/utils/saasherder/saasherder.py
+++ b/reconcile/utils/saasherder/saasherder.py
@@ -1113,8 +1113,7 @@ class SaasHerder:  # pylint: disable=too-many-public-methods
                 username, password = (
                     base64.b64decode(auth["auth"]).decode("utf-8").split(":")
                 )
-
-                return SaasHerder._get_and_validate_image(
+                img = SaasHerder._get_and_validate_image(
                     full_image_path=image,
                     username=username,
                     password=password,
@@ -1122,9 +1121,15 @@ class SaasHerder:  # pylint: disable=too-many-public-methods
                     timeout=REQUEST_TIMEOUT,
                     error_prefix=error_prefix,
                 )
+                if img:
+                    return img
+                else:
+                    logging.error(
+                        f"{error_prefix} Failed to authenticate to the repository for image: {image}. "
+                    )
 
         # basic auth fallback for backwards compatibility
-        return SaasHerder._get_and_validate_image(
+        img = SaasHerder._get_and_validate_image(
             full_image_path=image,
             username=image_auth.username,
             password=image_auth.password,
@@ -1132,6 +1137,14 @@ class SaasHerder:  # pylint: disable=too-many-public-methods
             timeout=REQUEST_TIMEOUT,
             error_prefix=error_prefix,
         )
+        # Provide a more appropriate error message if the image is not found due to authentication
+        if img:
+            return img
+        else:
+            logging.error(
+                f"{error_prefix} Failed to authenticate to the repository for image: {image}. "
+            )
+        return None
 
     @staticmethod
     def _get_and_validate_image(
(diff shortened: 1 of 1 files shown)


reconcile/utils/saasherder/saasherder.py
```diff
@@ -1113,25 +1113,38 @@ def _get_image(
...
             auth_server=image_auth.auth_server,
             timeout=REQUEST_TIMEOUT,
             error_prefix=error_prefix,
         )
+        # Provide a more appropriate error message if the image is not found due to authentication
+        if img:
+            return img
+        else:
+            logging.error(
+                f"{error_prefix} Failed to authenticate to the repository for image: {image}. "
```
rporres, line 1145: same as above

reconcile/utils/saasherder/saasherder.py
```diff
@@ -1156,11 +1169,19 @@ def _get_and_validate_image(
...
+            if 'authentication' in str(e).lower():
+                logging.error(
+                    f"{error_prefix} Failed to authenticate to the repository for image: {full_image_path}. "
+                    + f"details: {e!s}"
+                )
+            else:
+                logging.error(
+                    f"{error_prefix} Image is invalid: {full_image_path}. "
+                    + f"details: {e!s}"
+                    )
```
rporres, line 1184: if "authentication" is under the details in `f"details: {e!s}"`, why do we need more info?

reconcile/utils/saasherder/saasherder.py
```diff
@@ -1113,25 +1113,38 @@ def _get_image(
...
                     password=password,
                     auth_server=image_auth.auth_server,
                     timeout=REQUEST_TIMEOUT,
                     error_prefix=error_prefix,
                 )
+                if img:
+                    return img
+                else:
+                    logging.error(
+                        f"{error_prefix} Failed to authenticate to the repository for image: {image}. "
```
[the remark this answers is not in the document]
jonmosco: No.  It can be set in `_get_and_validate_image` since it's looking for the specific string there.

reconcile/utils/saasherder/saasherder.py
jonmosco: The `details` were previously included with the error messages, so I kept it there for consistency.

reconcile/utils/saasherder/saasherder.py
jonmosco: https://github.com/app-sre/qontract-reconcile/pull/4867/commits/246630e25d993a82ac336c48d4c0fe337e2cec5f

reconcile/utils/saasherder/saasherder.py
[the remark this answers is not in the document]
jonmosco: https://github.com/app-sre/qontract-reconcile/pull/4867/commits/246630e25d993a82ac336c48d4c0fe337e2cec5f

reconcile/utils/saasherder/saasherder.py
hemslo: the old message print out is `does not exist`, which means this error handling it not executed, so changes here won't work. If this blocked executed, then we should see `details: authentication xxx`

The root issue is here https://github.com/app-sre/sretoolbox/blob/dacd722f555f0315aab881621f6118b66e39f3bc/sretoolbox/container/image.py#L597-L601

401/403 error already handled and return False, we should check when image not exist, what response should be returned (404?) and only handle that case, re-throw other exceptions

PR #4873, opened 2025-02-17
APPSRE9086 - Error handling on rds upgrades

suzana-nesic:
This is to catch a DB upgrade dependency error that isn't caught until after an upgrade MR is made: `DBUpgradeDependencyFailure: One or more of the DB Instance's read replicas need to be upgraded`

### DIFF
--- a/reconcile/utils/terraform_client.py
+++ b/reconcile/utils/terraform_client.py
@@ -716,6 +716,14 @@ class TerraformClient:  # pylint: disable=too-many-public-methods
                 response = self._aws_api.describe_rds_db_instance(
                     account_name, resource_name, region_name=region_name
                 )
+                if len(response["DBInstances"][0].get("ReadReplicaDBInstanceIdentifiers")) > 0:
+                    replica_db = response["DBInstances"][0].get("ReadReplicaDBInstanceIdentifiers")[0]
+                    replica_db_response = self._aws_api.describe_rds_db_instance(
+                        account_name, replica_db, region_name=region_name
+                    )
+                    replica_db_engine_version = replica_db_response["DBInstances"][0].get("EngineVersion")
+                    if after.get("engine_version") > replica_db_engine_version:
+                        raise RdsUpgradeValidationError("The replica database is running a previous version from the primary and needs to have an upgrade first.")
             except ClientError as e:
                 # The RDS database might have been already removed, or there was
                 # a resource name change, and we no longer use valid references.


reconcile/utils/terraform_client.py
```diff
@@ -716,6 +716,22 @@ def _can_skip_rds_modifications(
                 response = self._aws_api.describe_rds_db_instance(
                     account_name, resource_name, region_name=region_name
                 )
+                replica_identifiers = response["DBInstances"][0].get(
```
rporres, line 719: I don't think validations should be made in `_can_skip_rds_modifications`, they should be made in

rporres (commented):
I think the related ticket is APPSRE-9946

Is this related to https://github.com/hashicorp/terraform-provider-aws/issues/22107 from the very missed @steveteahan?

suzana-nesic:
> The other related hashicorp issue is [hashicorp/terraform-provider-aws#30703](https://github.com/hashicorp/terraform-provider-aws/pull/30703). But it should be fixed in our current version... 🤔

Are you saying our current integrations should already have some kind of check that does what I'm proposing in the PR?

suzana-nesic:
> I think the related ticket is APPSRE-9946
> 

That ticket is related! I'll go ahead and fix it on the Jira side

> Is this related to [hashicorp/terraform-provider-aws#22107](https://github.com/hashicorp/terraform-provider-aws/issues/22107) from the very missed @steveteahan?
> 
> Is there any plans to add this check to ERv2 code?

I can certainly add it to the ERv2 code, I think it would make sense to do so. And it looks like it is related to what Steve posted about

reconcile/utils/terraform_client.py
```diff
@@ -823,6 +823,26 @@ def validate_db_upgrade(
...
+                    replica_db = replica_identifiers[0]
+                    replica_db_response = self._aws_api.describe_rds_db_instance(
+                        account_name, replica_db, region_name=region_name
+                    )
+                    replica_version = replica_db_response["DBInstances"][0].get(
+                        "EngineVersion"
+                    )
+                    if after.get("engine_version") > replica_version:
+                        raise RdsUpgradeValidationError(
+                            "The replica database is running a previous version from the primary and needs to have an upgrade first."
```
rporres, line 843: Re-reading our docummentation, we recommend to remove replicas before upgrade... shouldn't be this message more geared towards that fact? https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/dba/postgresql-rds-instance-major-version-upgrade.md#note-about-upgrading-read-replicas

I'm worried that people actually tries to do what this message says rather than what our documentation says. Just thinking out loud here 😅

suzana-nesic: It's a good point! If we think it's better to error out on recommendations we give to tenants, I'm on board with focusing on that. Is the deletion of the read replica specific to only postgres? Or should we keep the rule general for all engine types?

rporres: Everybody uses postgres... we support MySQL because we use it ourselves in Vault commercial database, but the contract says Postgres :)

suzana-nesic: Okie dokie! I'll put this logic on hold and change it to reflect the replica removal logic check

rporres: I'm not sure what you have in mind. I was just meaning something simple about the messaging. Something like "Detected upgrade of a Postgres RDS database with a read-replica. This is not supported by AWS. Please delete the replica before proceeding"

suzana-nesic: That's what I had in mind, yes

suzana-nesic:
Updated the PR description to include logic handling for managed erv2 instances: https://github.com/app-sre/er-aws-rds/pull/111

reconcile/utils/terraform_client.py
```diff
@@ -823,6 +823,19 @@ def validate_db_upgrade(
...
+            )
+            replica_identifiers = db_response["DBInstances"][0].get(
+                "ReadReplicaDBInstanceIdentifiers"
+            )
+            db_identifier = db_response["DBInstances"][0].get("DBInstanceIdentifier")
+            if replica_identifiers is not None:
+                if len(replica_identifiers) > 0:
+                    raise RdsUpgradeValidationError(
+                        f"The primary database, {db_identifier}, has a read replica associated with it. Please remove the replica before upgrading the primary database"
+                    )
```
hemslo, line 837: this check now is for both major and minor version upgrade, doc only mentions major version upgrade, ask to remove read replica for a prod database is not good.

Since `terraform-resources` for RDS is being deprecated, we only need to handle it in erv2 module, and the better solution to upgrade instance with read replicas is to use new blue/green deployment.

suzana-nesic: Thank you for your input! I'll go ahead and close this PR and will make the appropriate changes within the erv2 module

PR #4875, opened 2025-02-18
APPSRE-11544 remove saas 'image'

fishi0x01:
Lets remove all occurrences of `saas_file.resource_template.target.image`.

This also fixes a bug for moving commits integration, as this is still using the old `image` directive.

Depends on schema change: https://github.com/app-sre/qontract-schemas/pull/772


reconcile/test/test_saasherder.py
```diff
@@ -259,7 +259,7 @@ def test_dangling_target_config_hash(self) -> None:
     def test_check_saas_file_upstream_used_with_image(self) -> None:
         self.saas_file.resource_templates[0].targets[
             0
-        ].image = SaasResourceTemplateTargetImageV1(**{
+        ].images = SaasResourceTemplateTargetImageV1(**{
```
hemslo, line 262: should `images` be a list?

fishi0x01: Good catch! Fixed :+1:

hemslo (approved):
LGTM!

PR #4881, opened 2025-02-19
APPSRE-11390 fleet labeler - handle new spec files

fishi0x01:
We must handle newly created fleet labeler spec files, which are not in main branch yet.

Since these files are not in main branch, we cannot fetch their current content from VCS. If we are in dry-run and hit a 404 on a vcs fetch, then this is very likely because we are adding a new spec file. In that case we can skip the rendering steps, as the desired diff has already been printed before.
We could also skip the rendering altogether in dry-run, however, most MRs will be on existing files. Having the rendering procedure run on those adds extra certainty in MR check, which is why we should keep it for dry-run.

If we are not in dry-run or receive any other non 404 error, then the integration should still fail.

### DIFF
--- a/reconcile/fleet_labeler/integration.py
+++ b/reconcile/fleet_labeler/integration.py
@@ -203,7 +203,7 @@ class FleetLabelerIntegration(QontractReconcileIntegration[NoParams]):
         # If the content exists in main, then it doesnt harm to also run the rendering procedure.
         try:
             current_content = vcs.get_file_content_from_main(path=spec.path)
-        except Gitlab404Error as e:
+        except Gitlab404Error:
             if dry_run:
                 logging.info(
                     f"The file data{spec.path} does not exist in main branch yet. This is likely because it is being created with this MR. We are skipping rendering steps."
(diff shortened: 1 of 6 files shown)


esron (approved):
LGTM

reconcile/fleet_labeler/integration.py
```diff
@@ -189,7 +196,23 @@ class ClusterData(BaseModel):
...
+            current_content = vcs.get_file_content_from_main(path=spec.path)
+        except Gitlab404Error as e:
+            if dry_run:
+                logging.info(
+                    f"The file data{spec.path} does not exist in main branch yet. This is likely because it is being created with this MR. We are skipping rendering steps."
+                )
+                return
+            # 404 must never happen on non-dry-run, as the file must have already passed
+            # MR check and must have been merged to main
+            raise e
```
hemslo, line 214: since no handling logic for `e`, and to keep origin stacktrace, just do `raise` without `e`, https://docs.astral.sh/ruff/rules/verbose-raise/

reconcile/fleet_labeler/vcs.py
```diff
@@ -11,9 +17,16 @@ def __init__(self, vcs: VCSBase):
...
-        return self._vcs.get_file_content_from_app_interface_ref(
-            file_path=path, ref="main"
-        )
+        try:
+            return self._vcs.get_file_content_from_app_interface_ref(
+                file_path=path, ref="main"
+            )
+        except GitlabGetError as e:
+            if e.response_code != 404:
+                raise e
```
hemslo, line 26: similar here, `raise` without `e`

hemslo (approved):
LGTM!

PR #4889, opened 2025-02-25
Add max_session_duration to iam roles

rporres:
depends on
https://github.com/app-sre/qontract-schemas/pull/775

APPSRE-11566


hemslo (approved):
LGTM!

PR #4891, opened 2025-02-26
APPSRE-11390 fleet labeler synchronize subscription labels

fishi0x01:
The Fleet Labeler integration already has the ability to synchronize cluster inventories and declare desired subscription label state.

In this PR, we leverage the existing desired subscription label state and make the actual OCM API calls to create/update subscription labels to have them in synch with the fleet-labeler-spec.

Note, that we intentionally do not remove labels under the managed prefix. This might lead to unwanted removals on prefixes with many sub keys, which might not be managed by fleet labeler. However, if desired, this could easily be added later (maybe with an optional flag in the spec).

### DIFF
--- a/reconcile/fleet_labeler/integration.py
+++ b/reconcile/fleet_labeler/integration.py
@@ -188,8 +188,7 @@ class FleetLabelerIntegration(QontractReconcileIntegration[NoParams]):
         whats written in the rendered spec yet.
         """
         for cluster in spec.clusters:
-            desired_cluster = desired_clusters.get(cluster.cluster_id)
-            if not desired_cluster:
+            if not desired_clusters.get(cluster.cluster_id):
                 # The cluster is not part of the desired inventory (will be updated with MR soon)
                 continue
             # Ensure we only handle labels for our managed prefix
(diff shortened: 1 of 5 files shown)


reconcile/fleet_labeler/integration.py
```diff
@@ -115,71 +172,99 @@ def _render_yaml_file(
...
+                    f"[{spec.name}] Updating label '{key}={value}' for cluster '{cluster.cluster_id}' in subscription '{cluster.subscription_id}'."
+                )
+                if not dry_run:
+                    ocm.update_subscription_label(
+                        subscription_id=cluster.subscription_id,
+                        key=key,
+                        value=value,
+                    )
+            # Note, we dont want to enable removal for now - its too dangerous on a broad managed prefix
+            # However, if it is needed in the future, we could easily add it here.
```
hemslo, line 233: can we log something to know we detected some labels should be deleted?

fishi0x01: Ideally, Id like to add that at a later point, as for the current milestone it is not needed. Logging now would mean we potentially print on every reconcile.

I plan to add a `deleteLabels` flag in the future, with which we can control whether we want to delete or not. I will add the logging aspect then.

hemslo (approved):
LGTM!

PR #4892, opened 2025-02-27
APPSRE-11390 fleet labeler - fix filter

fishi0x01:
The `discover_clusters_by_labels()` function currently only populates labels into the result that are part of the filter. Our desired labels are not part of the default label matchers.

To circumvent this, we for now widen the filter by an `OR {spec.managed_prefix}.%` query. This will lead to all managed labels being populated into the `ClusterDetails` result.

Note, that the API call overhead here is neglectable, since we anyways most of the times have to fetch subscription labels for the whole inventory and our currently managed prefixes are very exclusive. Still, longterm, we should fix this by extending the utils.ocm module to provide this functionality.


lisa:
lgtm

lisa (approved):
lgtm :shipit:

PR #4893, opened 2025-02-27
APPSRE-11584 DTP use v2 labels

fishi0x01:
DTP should favor v2 labels now.

fleet-labeler already migrated integration/staging cluster labeling to the v2 schema.
Production clusters are not yet affected by this yet, which is why we keep the old labeling schema as fallback.

Note, also made adjustments to log `cluster.id` instead `cluster.external_id`, because `cluster.id` is used by fleet labeler (makes debugging easier).


cubismod (approved):
LGTM

PR #4894, opened 2025-02-28
APPSRE-11584 fleet labeler enhance logging

fishi0x01:
Lets expose error metric on label rendering errors. Do not block integration on label rendering errors.


reconcile/fleet_labeler/integration.py
```diff
@@ -261,28 +261,37 @@ def _sync_cluster_inventory(
...
+                        subscription_id=cluster_info.subscription_id,
+                        name=cluster_info.name,
+                        server_url=cluster_info.server_url,
+                        subscription_labels_content=self._render_default_labels(
+                            template=cluster_info.desired_label_default.subscription_label_template,
+                            labels=ocm.get_cluster_labels(cluster_id=cluster_id),
+                        ),
+                    )
+                )
+            except Exception as e:
```
hemslo, line 281: can we limit try except to only code block that we expect to throw exception?
`clusters_to_add.append` is ok.
`YamlCluster` is just a `BaseModel` with no additional validators (fields should already be not `None`).

`self._render_default_labels` and `ocm.get_cluster_labels` looks complex, just capture them?

hemslo (approved):

reconcile/fleet_labeler/integration.py
fishi0x01: Ah actually, let me put this already in this PR with metrics and make this right. The error I was trying to debug is already solved

fishi0x01: Ok I have adjusted the PR to make this properly now.

Catch default_label rendering errors -> skip cluster and expose metric for potential alert. That way, the integration doesnt block, but we are also not blind on such errors.

Also added tests for that metric + tests for the duplicate cluster match metric

fishi0x01: Note, I will add more errors (such as OCM API 5xx) in a different PR

hemslo (approved):

PR #4898, opened 2025-03-06
APPSRE-11584 DTP enforce v2 labels

fishi0x01:
Our fleet labeler capability is now managing DTP subscription labels. Lets enforce the v2 labeling schema and drop backwards compatibility with old schemas. Since fleet labeler is managing the labels, all the labels for the new schema already exist on the subscriptions.

Note, that for now we log a warning if expected labels are missing. During the DTPv3 transition, we will expose this as a metric. If we run into this case, it is likely that fleet-labeler didnt label the cluster yet - however, it should do so soon - thus we shouldnt fail here.


hemslo (approved):
LGTM!

PR #4899, opened 2025-03-07
APPSRE-11583 DTP cleanup dep creation

fishi0x01:
Lets avoid `dependency.populate_all()` and use classmethod creator instead.

### DIFF
--- a/reconcile/dynatrace_token_provider/dependencies.py
+++ b/reconcile/dynatrace_token_provider/dependencies.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from collections.abc import Mapping
 
 from reconcile.dynatrace_token_provider.ocm import OCMClient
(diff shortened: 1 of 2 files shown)


reconcile/dynatrace_token_provider/dependencies.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
hemslo, line 1: where is this used?

fishi0x01: In the creator method

```
def create(cls, secret_reader: SecretReaderBase) -> Dependencies
```

Else Id need to wrap it the type with ticks as forward declaration. But I prefer it without.

hemslo: can use `Self`, `def create(cls, secret_reader: SecretReaderBase) -> Self`

fishi0x01: `Self` doesnt work for me - ruff throwing error. Whats the problem with my current approach?

fishi0x01: Ah forgot the `Self` import. But still, why not my current approach?

hemslo: because that is the new feature in python 3.11 to solve this issue https://peps.python.org/pep-0673/, `__future__. annotations ` is still uncertain (TBD).

fishi0x01: I see. Adjusted :+1:

hemslo (approved):
LGTM!

PR #4904, opened 2025-03-10
(WIP)added slos to saas_files query

mehfuz:


reconcile/utils/slo_gatekeeper.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def is_slo_breached(self) -> bool:
+        for slo in self.slo_details_list:
+            slo_value = slo.get_SLO_value()
+            if slo_value < slo.slo.slo_target:
+                logging.info(
+                    f"SLO {slo.slo.name} from document {slo.slo_document_name} is breached. Expected value:{slo.slo.slo_target} current value:{slo_value}"
+                )
+                return True
+        return False
```
hemslo, line 145: better to aggregate all breached slos

reconcile/utils/slo_gatekeeper.py
```diff
@@ -0,0 +1,145 @@
...
+                is_basic_auth=is_basic_auth,
+                slo=slo,
+            )
+            for slo_document in slo_documents
+            for namespace in slo_document.namespaces
+            if not namespace.slo_namespace or namespace.prometheus_access
+            for prom_url, prom_token, is_basic_auth in [
+                self._get_credentials_from_slo_namespace(namespace)
+            ]
+            for slo in slo_document.slos or []
```
hemslo, line 100: credentials part is too complex, better to extract a class instead of tuple, so it can become

```python
            SLODetails(
                namespace_name=namespace.namespace.name,
                cluster_name=namespace.namespace.cluster.name,
                slo_document_name=slo_document.name,
                slo=slo,
                credentials=self._get_credentials_from_slo_namespace(namespace)
            )
            for slo_document in slo_documents
            for namespace in slo_document.namespaces
            for slo in slo_document.slos or []
```

I don't understand why require filter `if not namespace.slo_namespace or namespace.prometheus_access`

reconcile/utils/slo_gatekeeper.py
```diff
@@ -0,0 +1,145 @@
...
+        response = requests.get(
+            full_prom_url,
+            params={"query": (f"{prom_query}")},
+            headers=headers,
+            timeout=PROM_TIMEOUT,
+        )
+        response.raise_for_status()
+        data = response.json()
+        slo_value = self.parse_prom_response(data)
+        return slo_value
```
hemslo, line 71: please make this a standalone file and reusable by `dashdotdb_slo`

reconcile/utils/slo_details.py
```diff
@@ -0,0 +1,88 @@
...
+                f"prometheus returned empty SLO value for SLO: {self.slo.name} of SLO document: {self.slo_document_name}"
+            )
+        slo_value = float(slo_value[1])
+        if isnan(slo_value):
+            raise InvalidSLOValue(
+                f"invalid format for SLO: {self.slo.name} of SLO document: {self.slo_document_name}"
+            )
+        return slo_value
+
+    def get_SLO_value(self) -> float:
```
chassing, line 69: Please use pep8 naming conventions.

```suggestion
    def get_slo_value(self) -> float:
```

reconcile/utils/slo_details.py
```diff
@@ -0,0 +1,88 @@
...
+        full_prom_url = urljoin((f"{self.prom_credentials.prom_url}"), PROM_QUERY_URL)
```
chassing, line 70: Unnecessary f-string:

```suggestion
        full_prom_url = urljoin(self.prom_credentials.prom_url, PROM_QUERY_URL)
```

reconcile/utils/slo_gatekeeper.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class SLOGateKeeper:
+    def __init__(
+        self, slo_documents: list[SaasSLODocument], secret_reader: SecretReaderBase
+    ):
+        self.secret_reader = secret_reader
+        self.slo_details_list = self._create_SLO_details_list(slo_documents)
+
+    def _create_SLO_details_list(
```
chassing, line 19: Pep8 naming convention:

```suggestion
    def _create_slo_details_list(
```

reconcile/test/utils/test_slo_gatekeeper.py
```diff
@@ -0,0 +1,128 @@
...
+    SLODocumentSLOV1,
+    SLONamespacesV1,
+)
+from reconcile.gql_definitions.fragments.vault_secret import VaultSecret
+from reconcile.utils.secret_reader import SecretReaderBase
+from reconcile.utils.slo_gatekeeper import SLOGateKeeper
+
+
+@pytest.fixture
+def secret_reader(mocker: MockerFixture) -> MockerFixture:
```
chassing, line 19: No need to create this fixture again, just use the default one in `tests/conftest.py`

reconcile/test/utils/test_slo_gatekeeper.py
```diff
@@ -0,0 +1,128 @@
...
+        "reconcile.utils.slo_gatekeeper.SLODetails.get_SLO_value"
+    )
+    get_SLO_value.return_value = 0.96
+    slo_gate_keeper = SLOGateKeeper(
+        secret_reader=secret_reader, slo_documents=slo_documents
+    )
+    assert not slo_gate_keeper.is_slo_breached()
+
+
+def test_slo_gatekeeper_(secret_reader: SecretReaderBase, mocker: MockFixture) -> None:
```
chassing, line 80: Please use a more descriptive method name

reconcile/utils/slo_details.py
chassing, line 1: Please add some unit tests for `SLODetails`

reconcile/utils/slo_gatekeeper.py
```diff
@@ -0,0 +1,80 @@
...
+    def is_slo_breached(self) -> bool:
+        breached_slo_list: list[SLODetails] = []
+        for slo in self.slo_details_list:
+            slo_value = slo.get_SLO_value()
+            if slo_value < slo.slo.slo_target:
+                logging.info(
+                    f"SLO {slo.slo.name} from document {slo.slo_document_name} is breached. Expected value:{slo.slo.slo_target} current value:{slo_value}"
+                )
+                breached_slo_list.append(slo)
+        return bool(breached_slo_list)
```
chassing, line 80: There is no need to have a `list` here.

```suggestion
    def is_slo_breached(self) -> bool:
        error = False
        for slo in self.slo_details_list:
            slo_value = slo.get_SLO_value()
            if slo_value < slo.slo.slo_target:
                logging.info(
                    f"SLO {slo.slo.name} from document {slo.slo_document_name} is breached. Expected value:{slo.slo.slo_target} current value:{slo_value}"
                )
                error = True
        return error
```

reconcile/utils/slo_details.py
```diff
@@ -0,0 +1,88 @@
...
+        headers = {
+            "accept": "application/json",
+        }
+        headers["Authorization"] = (
+            f"{'Basic' if self.prom_credentials.is_basic_auth else 'Bearer'} {self.prom_credentials.prom_token}"
+        )
+        template = jinja2.Template(self.slo.expr)
+        prom_query = template.render({"window": self.slo.slo_parameters.window})
+        response = requests.get(
```
chassing, line 79: You could use the [ApiBase](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/rest_api_base.py#L42) class with dependency injection instead of hardcoding `requests.get` here.

reconcile/utils/slo_gatekeeper.py
hemslo: related to https://github.com/app-sre/qontract-reconcile/pull/4904#discussion_r2006786334, what I mean is `SLOGateKeeper` to return all breached slos instead of just a bool value, let caller handle logging.

include `slo_value` inside `SLODetails`, seperate fetch slo value and return list so fetch can even be threaded.

```python
@property
def breached_slos(self) -> list[SLODetails]:
    return [slo for slo in self.slo_details_list if slo.slo_value < slo.slo.slo_target]
```

reconcile/dashdotdb_slo.py
```diff
@@ -70,19 +72,28 @@ def __init__(
...
-        response = None
-
-        for item in service_slos:
-            slo_name = item.name
-            endpoint = f"{self.dashdotdb_url}/api/v1/serviceslometrics/{slo_name}"
-            payload = item.dashdot_payload()
-            if self.dry_run:
-                continue
+    @staticmethod
+    def get_dash_dot_db_payload(slo: SLODetails) -> dict[str, Any]:
```
hemslo, line 76: is `ServiceSLO` class not needed anymore? this logic is same as `dashdot_payload`

reconcile/dashdotdb_slo.py
```diff
@@ -94,127 +105,53 @@ def _post(self, service_slos: Iterable[ServiceSLO]) -> Response | None:
...
+            slo_documents=slo_documents,
+            secret_reader=self.secret_reader,
             thread_pool_size=self.thread_pool_size,
         )
 
+        slo_details_list = threaded.run(
+            func=self.get_current_slo_details,
+            iterable=slo_document_manager_list,
+            thread_pool_size=self.thread_pool_size,
+        )
```
hemslo, line 140: seems you want to collect `Exception`, then you need `return_exceptions=True,`. A better way is to handle exception inside `get_current_slo_details` so no expected exceptions throw out, and no need to filter valid and exception list later.

reconcile/gql_definitions/fragments/saas_slo_document.gql
```diff
@@ -0,0 +1,45 @@
...
+      name
+      expr
+      SLIType
+      SLOParameters {
+        window
+      }
+      SLOTarget
+      SLOTargetUnit
+    }
+}
```
hemslo, line 45: missing newline

reconcile/test/fixtures/slo_document_manager/api/v1/query/get.json
```diff
@@ -0,0 +1,15 @@
...
+            {
+                "metric": {},
+                "value": [
+                    174.11,
+                    "0.9999"
+                ]
+            }
+        ]
+    }
+}
```
hemslo, line 15: missing newline

reconcile/utils/slo_document_manager.py
```diff
@@ -0,0 +1,204 @@
...
+    username: str
+    password: str
+
+    def get_session(self) -> requests.Session:
+        session = requests.Session()
+        if self.is_basic_auth:
+            session.auth = requests.auth.HTTPBasicAuth(self.username, self.password)
+        else:
+            session.headers.update({"Authorization": f"Bearer {self.prom_token}"})
+        return session
```
hemslo, line 49: this will create a new session, make name more explicit like `build_session`

reconcile/utils/slo_document_manager.py
```diff
@@ -0,0 +1,204 @@
...
+            prom_token = self.secret_reader.read_secret(
+                namespace.namespace.cluster.automation_token
+            )
+        return PromCredentials(
+            prom_url=prom_url,
+            prom_token=prom_token,
+            is_basic_auth=is_basic_auth,
+            username=username,
+            password=password,
+        )
```
hemslo, line 153: avoid mutate fields by make `prom_token`, `username` and `password` optional, and make sure all branches can return valid result.

For example if `namespace.prometheus_access` has value but missing `password`, this function can return a result without throwing exception. We don't expect that to happen.


```python
        if namespace.prometheus_access and namespace.prometheus_access.username and namespace.prometheus_access.password:
            return PromCredentials(
                prom_url=namespace.prometheus_access.url,
                is_basic_auth=True,
                username=self.secret_reader.read_secret(
                    namespace.prometheus_access.username
                ),
                password=self.secret_reader.read_secret(
                    namespace.prometheus_access.password
                ),
            )
        if not namespace.namespace.cluster.automation_token:
            raise Exception(
                    f"cluster {namespace.namespace.cluster.name} does not have automation token set"
                )
        return PromCredentials(
            prom_url=namespace.namespace.cluster.prometheus_url,
            prom_token=elf.secret_reader.read_secret(
                namespace.namespace.cluster.automation_token
            ),
            is_basic_auth=False,
        )
```

reconcile/utils/slo_document_manager.py
```diff
@@ -0,0 +1,204 @@
...
+    def get_slo_details_list(self) -> list[SLODetails]:
+        try:
+            slo_details_list = threaded.run(
+                self._get_slo_details,
+                self.slos,
+                self.thread_pool_size,
+                return_exceptions=True,
+            )
+        finally:
+            self.cleanup()
```
hemslo, line 203: why need cleanup here? then this client can't send any more requests out, this should be done by caller

reconcile/utils/slo_document_manager.py
mehfuz: @hemslo username and password are not always necessary for prometheus_access. There are examples where we have prometheus_access without username and password.

reconcile/dashdotdb_slo.py
mehfuz: will be removing this

reconcile/utils/slo_document_manager.py
hemslo: then what do you expect to return for such case? it's not basic auth, and no token

mehfuz: no, the token from cluster will be returned right. in that case is_basic_auth will be false and token will be automation token from cluster

reconcile/dashdotdb_slo.py
mehfuz: @hemslo we can handle exception  at `get_current_slo_details `,but  the idea is to handle exception part once we have got the current value ot the slo and propagated them to dashdot..and log/fail at the end..so that the valid once are porpagated and do not need to wait due to exveptions

reconcile/utils/slo_document_manager.py
```diff
@@ -0,0 +1,264 @@
...
+    namespace_name: str
+    slo_document_name: str
+    cluster_name: str
+    slo: SLODocumentSLOV1
+    service_name: str
+    current_slo_value: float
+
+
+@dataclass
+class FlatSLODocument:
```
hemslo, line 49: let's make the name more meaningful like `NamespaceSLODocument`

reconcile/utils/slo_document_manager.py
```diff
@@ -0,0 +1,264 @@
...
+class SLODocumentManager:
+    """
+    Manages  SLO document including authentication, querying, and SLO value extraction.
+    """
+
+    def __init__(
+        self,
+        slo_documents: list[SLODocument],
+        secret_reader: SecretReaderBase,
+        thread_pool_size: int = 1,
```
hemslo, line 208: we usually have default pool size as 10, extract as default constant

reconcile/utils/slo_gatekeeper.py
```diff
@@ -0,0 +1,39 @@
...
+        self,
+        slo_documents: list[SLODocument],
+        secret_reader: SecretReaderBase,
+        thread_pool_size: int = 1,
+    ):
+        self.slo_manager = SLODocumentManager(
+            slo_documents=slo_documents,
+            secret_reader=secret_reader,
+            thread_pool_size=thread_pool_size,
+        )
```
hemslo, line 21: is this class still needed or we can let `SLODocumentManager` handles everything including `get_breached_slos`

mehfuz: I think now we do not need this class...let me merge the logic in existng SLODocumentManager

hemslo (approved):
LGTM!

chassing (approved):
LGTM

PR #4906, opened 2025-03-12
[gitlab-owners] support groups

maorfr:
this PR adds support to use a gitlab group in an OWNERS file.

the implementation is to iterate over each name, and if it's a group - add it's members to the list of approvers.

required to support a self-service onboarding, for example: https://gitlab.cee.redhat.com/service/managed-tenants/-/merge_requests/4870

### DIFF
--- a/reconcile/gitlab_owners.py
+++ b/reconcile/gitlab_owners.py
@@ -94,6 +94,13 @@ class MRApproval:
             lgtms.append(comment["username"])
         return lgtms
 
+    def expand_groups(self, owners: list[str]) -> list[str]:
+        members: set[str] = set()
+        for name in owners:
+            if group := self.gitlab.get_group_if_exists(name):
+                members.update(m.username for m in self.gitlab.get_group_members(group))
+        return list(members.union(owners))
+
     def get_approval_status(self):
         approval_status = {"approved": False, "report": None}
 
@@ -116,6 +123,11 @@ class MRApproval:
             for approver in change_owners["approvers"]:
                 if approver in lgtms:
                     change_approved = True
+
+            if not change_approved:
+                for approver in self.expand_groups(change_owners["approvers"]):
+                    if approver in lgtms:
+                        change_approved = True
 
             # Each change that was not yet approved will generate
             # a report message
--- a/reconcile/gitlab_owners.py
+++ b/reconcile/gitlab_owners.py
@@ -94,12 +94,12 @@ class MRApproval:
             lgtms.append(comment["username"])
         return lgtms
 
-    def expand_groups(self, owners: list[str]) -> list[str]:
+    def expand_groups(self, owners: list[str]) -> set[str]:
         members: set[str] = set()
         for name in owners:
             if group := self.gitlab.get_group_if_exists(name):
                 members.update(m.username for m in self.gitlab.get_group_members(group))
-        return list(members.union(owners))
+        return members.union(owners)
 
     def get_approval_status(self):
         approval_status = {"approved": False, "report": None}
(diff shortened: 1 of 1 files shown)


reconcile/gitlab_owners.py
```diff
@@ -64,10 +75,10 @@ def get_change_owners_map(self):
...
 
             change_owners_map[path] = {
-                "approvers": path_approvers,
-                "reviewers": path_reviewers,
-                "closest_approvers": closest_approvers,
-                "closest_reviewers": closest_reviewers,
+                "approvers": self.expand_groups(path_approvers),
+                "reviewers": self.expand_groups(path_reviewers),
+                "closest_approvers": self.expand_groups(closest_approvers),
+                "closest_reviewers": self.expand_groups(closest_reviewers),
```
chassing, line 81: Something is wrong here; `expand_groups` always returns `None`; therefore, you assign `None` to all dictionary fields.
Please change `expand_groups` to return the updated owners list.

reconcile/gitlab_owners.py
```diff
@@ -45,6 +45,17 @@ def __init__(self, gitlab_client, merge_request, owners, dry_run, persistent_lgt
...
+    def expand_groups(self, owners: list[str] | None) -> None:
+        if owners:
+            members = set()
+            for name in owners:
+                if group := self.gitlab.get_group_if_exists(name):
+                    members.update(self.gitlab.get_group_members(group))
+            for member in members:
+                name = member.username
+                if name not in owners:
+                    owners.append(name)
```
hemslo, line 57: since only `username` is required, and `get_group_members` returns a complex object `GroupMember`, better to just grab `username` to set like

```python
    members.update(group_member.username for group_member in self.gitlab.get_group_members(group))
```

`name not in owners` inside `members` loop check is O(N^2), can return set union directly


```python
return list(members.union(owners))
```

reconcile/gitlab_owners.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/4906/commits/929535687bb0a3953c216f4f2748bd7187bf9da2

reconcile/gitlab_owners.py
```diff
@@ -45,6 +45,18 @@ def __init__(self, gitlab_client, merge_request, owners, dry_run, persistent_lgt
             top_commit = next(commits)
             self.top_commit_created_at = dateparser.parse(top_commit.created_at)
 
+    def expand_groups(self, owners: list[str] | None) -> list[str] | None:
+        if owners:
+            members = set()
+            for name in owners:
+                if group := self.gitlab.get_group_if_exists(name):
+                    members.update(self.gitlab.get_group_members(group))
```
hemslo, line 53: even with cache, this will send too many requests to gitlab, should there be a special name pattern for group? Like `group:xxx`

reconcile/gitlab_owners.py
maorfr: https://github.com/app-sre/qontract-reconcile/pull/4906/commits/18d83cef230bda319d0bea3ca5d208823050c438

reconcile/gitlab_owners.py
maorfr: trying to be consistent with CODEOWNERS, which accept a group name. example: https://gitlab.cee.redhat.com/releng/konflux-release-data/-/blob/75f40dbc65dfa543a32f51b0d8b8f294734a0e6b/CODEOWNERS#L258

maorfr: do you have other ideas? i understand this is a lot of calls

maorfr: what about https://github.com/app-sre/qontract-reconcile/pull/4906/commits/26321a876eb07aafca1bac2e1d612953df6f02be?

maorfr: also added a condition that there are lgtms in https://github.com/app-sre/qontract-reconcile/pull/4906/commits/3d4244dc292f88e8a7cf5a102d562a50796c75ca

hemslo (approved):
LGTM!

PR #4908, opened 2025-03-13
promtool 2.54.1

rporres:
APPSRE-11598


patjlm:
I just noticed [asdf supports promtool](https://github.com/looztra/asdf-promtool). Would it make sense to reference it in the `.tool-versions` file?

PR #4909, opened 2025-03-14
ERv2: RuntimeError when dry_run without dry_run_job_suffix

lechuk47:
I observed cases where DRY_RUN  runs are made with `dry_run_job_suffix = ""`.  This parameter contains the MR number to identify the MR on the job pods. I think Jenkins launches these runs under certain circumstances. 

DRY_RUN runs must always run under an MR context. Therefore, these runs will be terminated with a RuntimeException.

### DIFF
--- a/reconcile/external_resources/integration.py
+++ b/reconcile/external_resources/integration.py
@@ -143,6 +143,9 @@ def run(
     workers_cluster: str | None = None,
     workers_namespace: str | None = None,
 ) -> None:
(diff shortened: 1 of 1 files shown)


lechuk47:
> With that change, I can't run `external-resources` in dev mode anymore. How about having a default value for it?

The main purpose is to avoid undesired runs by Jenkins. Why don't you use "DEV-MODE" as a suffix in the dev mode?

chassing (approved):

PR #4910, opened 2025-03-14
APPSRE-11583 more fleet labeler metrics

fishi0x01:
Lets expose more metrics about this capability:
- number unique clusters under management
- number of labels under management across all clusters

### DIFF
--- a/reconcile/fleet_labeler/integration.py
+++ b/reconcile/fleet_labeler/integration.py
@@ -66,11 +66,11 @@ class FleetLabelerIntegration(QontractReconcileIntegration[NoParams]):
             secret_reader=self.secret_reader,
             dry_run=dry_run,
         )
-        self._metrics = dependencies.metrics
         self.reconcile(dependencies=dependencies)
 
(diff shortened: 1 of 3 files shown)


reconcile/fleet_labeler/integration.py
```diff
@@ -99,6 +99,16 @@ def reconcile(self, dependencies: Dependencies) -> None:
                 ocm=ocm,
                 dry_run=synch_labels,
             )
+            num_labels = 0
+            for cluster in spec.clusters:
+                all_cluster_ids.add(cluster.cluster_id)
+                num_labels += sum(1 for _ in cluster.subscription_labels)
```
hemslo, line 105: is `cluster.subscription_labels` a `dict`? why not directly `num_labels += len(cluster.subscription_labels)`

fishi0x01: Absolutely! Fixed

hemslo (approved):
LGTM!

PR #4912, opened 2025-03-18
APPSRE-11551 ALB health checks

fishi0x01:
ALB health checks should be configurable. The defaults are taken from currently provisioned ALB instances.

schema: https://github.com/app-sre/qontract-schemas/pull/783

Verified locally by changing only a few target groups

```
qontract-reconcile-terraform-resources  | [2025-03-18 10:15:17] [INFO] [DRY-RUN] [terraform_client.py:log_plan_diff:374] - ['update', 'quayio-prod', 'aws_lb_target_group', 'quayio-production-alb01-quay-pull', {'health_check'}]
qontract-reconcile-terraform-resources  | [2025-03-18 10:15:17] [INFO] [DRY-RUN] [terraform_client.py:log_plan_diff:374] - ['update', 'quayio-prod', 'aws_lb_target_group', 'quayio-production-alb01-quay-push', {'health_check'}]
qontract-reconcile-terraform-resources  | [2025-03-18 10:15:17] [INFO] [DRY-RUN] [terraform_client.py:log_plan_diff:374] - ['update', 'quayio-prod', 'aws_lb_target_group', 'quayio-production-alb01-quayio-prod-ak', {'health_check'}]
qontract-reconcile-terraform-resources  | [2025-03-18 10:15:17] [INFO] [DRY-RUN] [terraform_client.py:log_plan_diff:374] - ['update', 'quayio-prod', 'aws_lb_target_group', 'quayio-production-alb01-quayio-production-py3', {'health_check'}]
```

Other ALB target groups are not affected and keep their defaults

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -5376,6 +5376,25 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
                 f"{identifier}-{target_name}", **lbt_random_id_values
             )
             tf_resources.append(lbt_random_id)
+            health_check_interval = 10
+            health_check_timeout = 5
+            health_check_unhealthy_threshold = 3
+            health_check_healthy_threshold = 3
+            if health_check := t.get("health_check"):
+                health_check_interval = (
+                    health_check.get("interval") or health_check_interval
+                )
+                health_check_timeout = (
+                    health_check.get("timeout") or health_check_timeout
+                )
+                health_check_unhealthy_threshold = (
+                    health_check.get("unhealthy_threshold")
+                    or health_check_unhealthy_threshold
+                )
+                health_check_healthy_threshold = (
+                    health_check.get("healthy_threshold")
+                    or health_check_healthy_threshold
+                )
 
             # https://www.terraform.io/docs/providers/aws/r/
             # lb_target_group.html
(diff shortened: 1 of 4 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -5376,6 +5376,25 @@ def populate_tf_resource_alb(self, spec, ocm_map=None):
                 f"{identifier}-{target_name}", **lbt_random_id_values
             )
             tf_resources.append(lbt_random_id)
+            health_check_interval = 10
+            health_check_timeout = 5
+            health_check_unhealthy_threshold = 3
+            health_check_healthy_threshold = 3
+            if health_check := t.get("health_check"):
```
patjlm, line 5383: will we ever get `t.get("health_check") == None` ?

patjlm (approved):

reconcile/utils/terrascript_aws_client.py
fishi0x01: Its an optional value in the schema. My understanding is, that this will return `None`. The intention of this check is to only act if `health_check` is configured. But probably Im misinterpreting the question?

patjlm: Since we query
```
 health_check {
    unhealthy_threshold
    timeout
    interval
    healthy_threshold
}
```
I think we will never have `health_check` set to `None` but we can have its fields set to `None` (as tested below).

This is really a nit, but I'd like to make sure I am not wrong :wink:

fishi0x01: Ahh I understand.

Reconfirmed in the debugger now - `health_check` is Ǹone` if its not specified in app-interface.

patjlm: Ack

PR #4918, opened 2025-03-19
Allow users to change project paths & repo URL locations in tf-repo

cubismod:
Git URLs and file paths change over time as users refine their repos. This change removes some guardrails preventing users from updating their `repository` (gitlab URL pointing to their repo) and the `projectPath` (location within that repo). Also updates the error message if a user tries to do this.

Users still cannot move repos between accounts or change the repo name since these are intrinsically linked to the tf-repo statefile.


uv.lock
```diff
@@ -267,7 +267,7 @@ name = "click"
 version = "8.1.8"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "colorama", marker = "sys_platform == 'win32'" },
+    { name = "colorama", marker = "platform_system == 'Windows'" },
```
hemslo, line 270: please update `uv` version locally to avoid this diff

cubismod: ACK thx will fix!

PR #4920, opened 2025-03-20
Handle immutable field changes in Recreate type deployments

lechuk47:
Handle immutable field deployment changes by recreating the deployment objects. 
* This only happens when immutable fields are changed. 
* This rollout strategy has an intrinsic downtime when this behavior happens, this code only automates it.

### DIFF
--- a/reconcile/openshift_base.py
+++ b/reconcile/openshift_base.py
@@ -443,6 +443,11 @@ def apply(
             # spec.selector changes
             current_resource = oc.get(namespace, resource_type, resource.name)
 
+            if current_resource["spec"]["strategy"]["type"] == "Recreate":
+                oc.delete(namespace, resource_type, resource.name)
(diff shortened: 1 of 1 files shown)


fishi0x01 (approved):

reconcile/openshift_base.py
```diff
@@ -443,6 +443,11 @@ def apply(
             # spec.selector changes
             current_resource = oc.get(namespace, resource_type, resource.name)
 
+            if current_resource["spec"]["strategy"]["type"] == "Recreate":
+                oc.delete(namespace, resource_type, resource.name)
+                oc.create(namespace, resource=annotated)
```
hemslo, line 448: @lechuk47 can this be done by `oc replace --force` to save one command process? Since delete and create/apply is used many places in code already, can be a future refactor.

PR #4927, opened 2025-03-21
promtool 2.54.1

rporres:
and an update of the base images that contain a fix for the missing timezone setting, see https://github.com/app-sre/container-images/pull/188

APPSRE-11598


rporres:
third attempt after #4908 and #4915, this times looks certainly better.

PR #4929, opened 2025-03-24
chore(erv2): deprecate old_parameter_group

hemslo:
`old_parameter_group` no longer needed in ERv2 RDS

[APPSRE-11657](https://issues.redhat.com/browse/APPSRE-11657)

### DIFF
--- a/reconcile/external_resources/aws.py
+++ b/reconcile/external_resources/aws.py
@@ -163,9 +163,6 @@ class AWSRdsFactory(AWSDefaultResourceFactory):
         if "parameter_group" in data:
             pg_data = rvr._get_values(data["parameter_group"])
             data["parameter_group"] = pg_data
-        if "old_parameter_group" in data:
-            old_pg_data = rvr._get_values(data["old_parameter_group"])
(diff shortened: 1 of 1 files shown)


mehfuz (approved):

lechuk47:
How does this work on deferred updates on the maintenance window?

hemslo:
> How does this work on deferred updates on the maintenance window?

@lechuk47 I just verified that even explicitly set `apply_immediately` to `false`, parameter group change will be immediately applied, just the status will be `pending-reboot`, same as doing the change manually in AWS Console and select `Apply during the next scheduled maintenance window`. Unused parameter group is safe to delete once no longer reference by RDS, I think internally those parameter values are copied to host until reboot to refresh.

According to [doc](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.Maintenance.html)

>Deferred DB instance modifications that you have chosen not to apply immediately are applied during the maintenance window. For example, you might choose to change the DB instance class or parameter group during the maintenance window. Such modifications that you specify using the pending reboot setting don't show up in the Pending maintenance list.

PR #4931, opened 2025-03-25
APPSRE-11221 - replace gcr_mirror integration with gcp_image_mirror

cubismod:
This merge request updates our GCR Mirror integration to support Artifact Registry endpoints now that [GCR has been replaced with AR](https://cloud.google.com/artifact-registry/docs/transition/prepare-gcr-shutdown).

**Note that I did perform the automatic Google Migration to Artifact Registry. Therefore, the existing GCR integration is still working for now but we currently don't have a solution for adding new repos to sync**

The integration is largely the same logic as before however I went in and added GQL query classes as well as type hints for everything as well as removed some untyped dicts to replace them with Pydantic classes.  The major change is to support loading these new "Artifact Registry" credentials which are separate from the existing GCR ones.

Depends on https://github.com/app-sre/qontract-schemas/pull/792

## Migration Plan
1. Add new `artifactRegistryMirrors` definitions to App-Interface for each existing gcrRepo. These will be pointed to their new `us-docker.pkg.dev` endpoints.
2. Let the integration sync backup images over the weekend while still keeping the `gcr.io` images up to date.
3. Validate that everything has synced and the integration is working, then remove the old `gcrRepos` definitions from App-Interface. 
4. Delete old "GCR" images from GCP once we are certain everything is syncing appropriately.
5. Follow-up MR that will remove old GCR schemas + code paths from this integration and App Interface.


hemslo (approved):
LGTM!

PR #4934, opened 2025-03-27
APPSRE-11690 AUS max parallel upgrades

patjlm:
https://issues.redhat.com/browse/APPSRE-11690

Adding support for limiting the number of parallel upgrades within a sector. The main changes involve adding new fields and methods to handle the `maxParallelUpgrades` attribute for sectors, updating existing functions to consider this new attribute, and adding tests to verify the new functionality.

The new field is added as part of https://github.com/app-sre/qontract-schemas/pull/802

The main change is in commit 4618c40f2c380c460fd4c467bba9e94c8afe8aa8, updating `reconcile/aus/base.py` to introduce the function `verify_max_upgrades_should_skip`

### DIFF
--- a/reconcile/aus/base.py
+++ b/reconcile/aus/base.py
@@ -5,6 +5,7 @@ from abc import (
     ABC,
     abstractmethod,
 )
+from collections import defaultdict
 from collections.abc import Callable, Sequence
 from datetime import (
     datetime,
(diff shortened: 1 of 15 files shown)


reconcile/aus/base.py
```diff
@@ -1042,25 +1086,34 @@ def calculate_diff(
...
+            set_upgrading(spec.cluster.id, spec.effective_mutexes, sector_name)
 
     addon_service = init_addon_service(desired_state.org.environment)
     now = datetime.utcnow()
     gates = get_version_gates(ocm_api)
     for spec in desired_state.specs:
+        sector_name = spec.upgrade_policy.conditions.sector
+        sector = None
+        if sector_name:
+            sector = desired_state.sectors[sector_name]
```
hemslo, line 1115: this part is duplicated inside `verify_max_upgrades_should_skip`, can we do it once here and pass the result to verify?

```python
        sector_name = spec.upgrade_policy.conditions.sector
        sector = desired_state.sectors[sector_name] if sector_name else None # for typing right Sector | None
        ...
        if verify_max_upgrades_should_skip(spec, sector_upgrades, sector):
            continue
```

then for `verify_max_upgrades_should_skip`

```python
def verify_max_upgrades_should_skip(
    desired: ClusterUpgradeSpec,
    sector_upgrades: dict[str, set[str]],
    sector: Sector | None,
) -> bool:
    if sector is None:
        return False
    sector_name = sector.name
    ...
```

reconcile/aus/base.py
```diff
@@ -982,6 +983,49 @@ def verify_lock_should_skip(
...
+        max_parallel_upgrades_percent = int(sector.max_parallel_upgrades[:-1])
+        max_parallel_upgrades = round(
+            sector_cluster_count * max_parallel_upgrades_percent / 100
+        )
+    else:
+        max_parallel_upgrades = int(sector.max_parallel_upgrades)
+
+    # we allow at least one upgrade
+    if max_parallel_upgrades == 0:
+        max_parallel_upgrades = 1
```
hemslo, line 1017: can we limit it at schema level that `max_parallel_upgrades` must be positive int or `1-100%`, like `^(100|[1-9]\d?)%|[1-9]\d*$`. So we don't have special rules about hidden bump it to 1.


```python
        max_parallel_upgrades = math.ceil(
            sector_cluster_count * max_parallel_upgrades_percent / 100
        )
```

patjlm: yes, but if they set 1% on say 50 clusters in the sector, then we will still get `0` here

patjlm: using `ceil` could be invalid in other cases

hemslo: `math.ceil(20 * 1 / 100` = `1`, which case is invalid?

chassing (approved):

PR #4935, opened 2025-03-28
[maintenance] update uv to 0.6.11

cubismod:
Update uv to 0.6.11 in Dockerfile & the `.tool-versions`. 

Reason: the lockfile format changed between recent versions of uv so we should update to the latest version and ensure devs are using the same if they use ASDF for version management.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,6 +1,6 @@
 # this file is used by asdf to automatically select the correct tool versions
 # but it also serves as a reference for non-asdf users
-uv
 terraform 1.6.6
 oc 4.12.46
 helm 3.11.1
+uv 0.6.10
--- a/uv.lock
+++ b/uv.lock
@@ -1,4 +1,5 @@
 version = 1
+revision = 1
 requires-python = "==3.11.*"
 
 [[package]]
@@ -1493,7 +1494,6 @@ wheels = [
 
 [[package]]
 name = "qontract-reconcile"
-version = "0.10.2.dev69"
 source = { editable = "." }
 dependencies = [
     { name = "anymarkup" },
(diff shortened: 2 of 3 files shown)


uv.lock
```diff
@@ -1493,7 +1494,6 @@ wheels = [
 
 [[package]]
 name = "qontract-reconcile"
-version = "0.10.2.dev69"
```
esron, line 1496: In my setup running `make test` locally without version produces an error.

cubismod: Which uv version are you running?

cubismod: also is it an error on the lockfile?

esron: uv version - 0.6.2

Yes.

```
error: Failed to parse `uv.lock`
  Caused by: TOML parse error at line 1495, column 1
     |
1495 | [[package]]
     | ^^^^^^^^^^^
missing field `version`
Error: building at STEP "RUN UV_DYNAMIC_VERSIONING_BYPASS="0.0.0" uv sync --frozen --no-group dev --python /usr/bin/python3": while running runtime: exit status 2
make: *** [Makefile:51: test-app] Error 2
```

esron: I've tried with uv version 0.6.10 and got the same error. Maybe this is something in my local setup. I'll investigate. Meanwhile LGTM

esron (approved):

.tool-versions
```diff
@@ -1,6 +1,6 @@
 # this file is used by asdf to automatically select the correct tool versions
 # but it also serves as a reference for non-asdf users
-uv
 terraform 1.6.6
 oc 4.12.46
 helm 3.11.1
+uv 0.6.10
```
hemslo, line 6: `0.6.11` is released now, you can bump it to match commit message

cubismod: lol this project and their release schedule :shaking_face:

cubismod:
Requesting your review as the UV expert @chassing 

I know you were saying to leave `.tool-versions` without a particular version marker but that's difficult with having a particular version locked in our Dockerfile as `uv` sometimes has backwards-incompatible changes to the lockfile which can break a user's local development environment (that happened to me).

chassing (approved):

PR #4940, opened 2025-04-03
fix rds output

BumbleFeng:
filter out deleted namespace, remove ca_cert_identifier as upgrade finished for all fleet.


hemslo (approved):
LGTM!

patjlm:
is the upgrade finished in fedramp as well?

PR #4941, opened 2025-04-04
leave a comment to avoid resource file deletion

chassing:
Leave a comment for us to prevent any future resource file deletions.

Context: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/138279#note_15164508


hemslo (approved):
LGTM!

PR #4944, opened 2025-04-09
fix(dev): update config to make dev-reconcile-loop works with uv setup

hemslo:
`run-integration` script created by `uv sync` has shebang `#!/work/.venv/bin/python3`, can't simply move `.venv` to `/venv`, have to run `uv sync` under the new location.

Now use `/opt/app-root/src` as code dir in `dev-image` to hold `.venv`.


cubismod (approved):
LGTM, validated on Linux

dwelch0 (approved):
Confirmed on macOS. Thank you ❤️

PR #4949, opened 2025-04-11
APPSRE-11651 - Ensure template-renderer local copy aligns with gql server

dwelch0:
# Summary:

This PR improves data consistency between the GraphQL server and the local cloned Git repository used by the templating integration. Specifically, it ensures that the git repo is checked out to the same commit as the one currently being served by the gql server.

## Why 
This change addresses potential mismatches between:
* the GraphQL data served via the gql API, and
* the file system state checked out during local repository operations.

By synchronizing both to the same git commit, we eliminate race conditions or discrepancies that could arise if the repo and gql server are out of sync.

### DIFF
--- a/reconcile/templating/renderer.py
+++ b/reconcile/templating/renderer.py
@@ -6,7 +6,9 @@ from abc import ABC, abstractmethod
 from collections.abc import Callable
 from pathlib import Path
 from typing import Any, Self
+from urllib.parse import urlparse
 
+import requests
 from ruamel import yaml
 
 from reconcile.gql_definitions.templating.template_collection import (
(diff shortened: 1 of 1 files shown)


reconcile/templating/renderer.py
```diff
@@ -331,9 +345,8 @@ def reconcile_template_collection(
                     persistence_transaction.write(output)
 
     def reconcile(
-        self, persistence: FilePersistence, ruamel_instance: yaml.YAML
+        self, gql_api: GqlApi, persistence: FilePersistence, ruamel_instance: yaml.YAML
     ) -> None:
-        gql_api = init_from_config(validate_schemas=False)
```
dwelch0, line 336: initializing gql_api is moved to run and passed into reconcile in order to take advantage of the url and token attributes retrieved during [gql init](https://github.com/app-sre/qontract-reconcile/blob/f538b381ae60a5bb6ebbb9bdb3f0305146e43fdf/reconcile/utils/gql.py#L392)

hemslo: can you check if we really need to call `init_from_config` here instead of using already inited `gql_api`? And we should be able to get `commit` field directly from that `gql_api` object.

dwelch0: Thank you for the call outs. Code has been simplified.

hemslo (approved):
LGTM!

PR #4951, opened 2025-04-15
Add alb mutual authentication APPSRE-11698

esron:
Related ticket: [APPSRE-11698](https://issues.redhat.com/browse/APPSRE-11698)

This adds the possibility for tenants to enable and [setup mTLS](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/configuring-mtls-with-elb.html) for ALB in AWS.

This uses the most simple implementation of the terraform provider example found here:
https://registry.terraform.io/providers/hashicorp/aws/5.39.1/docs/resources/lb_listener#mutual-tls-authentication

Usage example:
Tenant can add a section like this on their `alb` external resources under a namespace file:
```yaml
# ...
    mutual_authentication:
      mode: passthrough
      ca_cert_bundle_s3_bucket_name: quayio-regproxy-alb-ca-bundle-test-bucket
      ca_cert_bundle_s3_bucket_key: ca-bundle.pem
# ...
```
Related qontract-schemas changes: https://github.com/app-sre/qontract-schemas/pull/818

The resulting terraform configuration was tested in dry-run. We still need to test in a stage environment.


reconcile/utils/terrascript_aws_client.py
```diff
@@ -5493,6 +5497,27 @@ def populate_tf_resource_alb(self, spec, ocm_map=None):
...
+                "ca_certificates_bundle_s3_key": mutual_authentication[
+                    "ca_cert_bundle_s3_bucket_key"
+                ],
+            }
+            trust_store = aws_lb_trust_store(
+                f"{identifier}-trust-store", **trust_store_values
+            )
+            tf_resources.append(trust_store)
+            values["mutual_authentication"] = {
+                "mode": resource["mutual_authentication"]["mode"],
```
hemslo, line 5517: use `mutual_authentication["mode"]` since already extracted

reconcile/utils/terrascript_aws_client.py
```diff
@@ -5493,6 +5497,27 @@ def populate_tf_resource_alb(self, spec, ocm_map=None):
             },
             "depends_on": self.get_dependencies([lb_tf_resource, default_target]),
         }
+
+        # mutual authentication section
+        mutual_authentication = resource.get("mutual_authentication")
+        if mutual_authentication:
```
hemslo, line 5503: inline `mutual_authentication` since only used in if block

```python
if mutual_authentication := resource.get("mutual_authentication"):
```

hemslo (commented):
`alb` is already a complex object, can we add test cases to cover new features?

esron:
> `alb` is already a complex object, can we add test cases to cover new features?

We didn't have any test for alb, I've added a test for the simplest path. In fact we only have tests for s3_bucket resources.

cubismod:
How do we ensure that the ALB has access to the S3 bucket?

PR #4952, opened 2025-04-15
Typed queries for aws-cloudwatch-log-retention integration

a-sinclaire:
Added custom gql definition for `aws-cloudwatch-log-retention` to replace the non-typed `get_aws_accounts` query from `queries.py`. The only option necessary for this integration is `cleanup`, and so that is what the custom gql definition includes.

Modified the the `integration.py` to use the new query, as well as convert its output to a dict in select places to be compatible with other necessary functions.

[APPSRE-11602](https://issues.redhat.com/browse/APPSRE-11602)


chassing:
```
19:43:19 uv run ruff format --check
19:43:19 Would reformat: reconcile/aws_cloudwatch_log_retention/integration.py
19:43:19 1 file would be reformatted, 981 files already formatted
19:43:19 make: *** [Makefile:129: linter-test] Error 1
```

Please rerun `make format`. Another option would be to trigger `ruff` in your IDE automatically.

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -230,17 +239,40 @@ def _reconcile_log_groups(
...
-        if "aws-cloudwatch-log-retention"
-        not in (a.get("disable") or {}).get("integrations", [])
-    ]
+# def get_aws_accounts_test() -> list[dict]:
+#     gql_api_query = gql.get_api().query
+#     return [
+#         account.dict(by_alias=True)
+#         for account in get_aws_accounts(query_func=gql_api_query).accounts
+#         or []
+#     ]
```
suzana-nesic, line 248: I think we can go ahead and clean this section up if not needed anymore

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -230,17 +239,40 @@ def _reconcile_log_groups(
...
+        accounts = [
+            account
+            for account in accounts
+            if account.disable is None
+            or (
+                account.disable.integrations is not None
+                and (
+                    "aws-cloudwatch-log-retention" not in (account.disable.integrations)
+                )
+            )
```
hemslo, line 267: none check can be simplified like this

```python
if not (account.disable and account.disable.integrations and "aws-cloudwatch-log-retention" in account.disable.integrations)
```

and inline return directly instead of overwrite `accounts`, so 

```python
        return [
            account
            for account in accounts
            if not (account.disable and account.disable.integrations and "aws-cloudwatch-log-retention" in account.disable.integrations)
         ]
```

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -230,17 +239,40 @@ def _reconcile_log_groups(
...
+        ]
+    return accounts or []
 
 
 def run(dry_run: bool, thread_pool_size: int) -> None:
-    aws_accounts = get_active_aws_accounts()
+    aws_accounts = [
+        account.dict(by_alias=True) for account in get_active_aws_accounts() or []
+    ]
```
hemslo, line 275: please also pass typed class to other methods defined in this file, like `create_awsapi_client` and `_reconcile_log_groups`, only do to dict when required in usage

inside run

```python
    aws_accounts = get_active_aws_accounts()
    with create_awsapi_client(aws_accounts, thread_pool_size) as awsapi:
        for aws_account in aws_accounts:
            _reconcile_log_groups(dry_run, aws_account, awsapi)
```

do `dict(by_alias=True)` inside `create_awsapi_client` as it's required by `AWS`, but keep the typed class in other methods.


```python
def create_awsapi_client(accounts: list[AWSAccountV1], thread_pool_size: int) -> AWSApi:
    settings = queries.get_secret_reader_settings()
    return AWSApi(
       thread_pool_size,
       accounts=[account.dict(by_alias=True) for account in accounts],
       settings=settings,
       init_users=False,
     )
```

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -230,17 +239,40 @@ def _reconcile_log_groups(
...
+#     gql_api_query = gql.get_api().query
+#     return [
+#         account.dict(by_alias=True)
+#         for account in get_aws_accounts(query_func=gql_api_query).accounts
+#         or []
+#     ]
+
+
+def account_to_dict(account: AWSAccountV1) -> dict[str, Any]:
+    return account.dict(by_alias=True)
```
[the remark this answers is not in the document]
a-sinclaire: It is used in the test file, but I don't explicitly use it here. I can remove this from this file. I could move it to the other file. In actuality I ended up only using it once in that file (I thought it was going to be used more but I didn't end up needing it), so I think I can just remove it.

reconcile/test/fixtures/aws_cloudwatch_log_retention/config.toml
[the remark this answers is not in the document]
a-sinclaire: I thought it was important, but upon deleting it the integration still works and the tests pass, so I think this is safe to remove.

reconcile/aws_cloudwatch_log_retention/integration.py
```diff
@@ -230,17 +239,40 @@ def _reconcile_log_groups(
...
+
+
+def get_active_aws_accounts() -> list[AWSAccountV1]:
+    accounts = get_aws_accounts(gql.get_api())
+    if accounts is not None:
```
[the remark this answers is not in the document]
a-sinclaire: good catch:)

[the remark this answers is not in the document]
hemslo: no need to use `or []` as it's already a list.

reconcile/aws_cloudwatch_log_retention/integration.py
hemslo: similar in https://github.com/app-sre/qontract-reconcile/pull/4952/files#diff-773495110cdde5ed044f7f1bbf0eb68bf7ba5d387c6e3985cedaea64ec613760R52, since this a method defined in this file, now that we have typed class object pass in, we don't need to call `dict` anymore, just access fields.

a-sinclaire: looks good:)
i did my best to replace everything with typed class objects and access fields where I could.
there were some instances where the fields couldn't be found for some reason, so I maintained it as dict access.

suzana-nesic:
I'll give the 🦭 of approval. Feel free to push. this out whenever you are comfortable to do so!

reconcile/aws_cloudwatch_log_retention/integration.py
hemslo: which one is missing? inside `get_desired_cleanup_options_by_region`, you don't need `account_dict = account.dict(by_alias=True)`, can just be

```python
    default_region = account.resources_default_region
    result = defaultdict(list)
    if cleanup := account.cleanup:
```

a-sinclaire: Ah, I was doing `account.resourcesDefaultRegion`, since it is camel case in the actual type definition.

I'm still running into an issue though.
Specifically when accessing attributes of the cleanup options themselves.
ex.
`cleanup_option.region`
`cleanup_option.regex`
`cleanup_option.retention_in_days`
`cleanup_option.delete_empty_log_group`

It seems that not all of the implementations of `AWSAccountCleanupOption_v1` have these.
For example `AWSCleanupOptionAMI_v1` does not have a `delete_empty_log_group` field.
Here are the errors I'm running into for more context:
```bash
reconcile/aws_cloudwatch_log_retention/integration.py:57: error: Item "AWSAccountCleanupOptionV1" of "AWSAccountCleanupOptionCloudWatchV1 | AWSAccountCleanupOptionAMIV1 | AWSAccountCleanupOptionV1" has no attribute "region"  [union-attr]
reconcile/aws_cloudwatch_log_retention/integration.py:60: error: Item "AWSAccountCleanupOptionV1" of "AWSAccountCleanupOptionCloudWatchV1 | AWSAccountCleanupOptionAMIV1 | AWSAccountCleanupOptionV1" has no attribute "regex"  [union-attr]
reconcile/aws_cloudwatch_log_retention/integration.py:61: error: Item "AWSAccountCleanupOptionAMIV1" of "AWSAccountCleanupOptionCloudWatchV1 | AWSAccountCleanupOptionAMIV1 | AWSAccountCleanupOptionV1" has no attribute "retention_in_days"  [union-attr]
reconcile/aws_cloudwatch_log_retention/integration.py:61: error: Item "AWSAccountCleanupOptionV1" of "AWSAccountCleanupOptionCloudWatchV1 | AWSAccountCleanupOptionAMIV1 | AWSAccountCleanupOptionV1" has no attribute "retention_in_days"  [union-attr]
reconcile/aws_cloudwatch_log_retention/integration.py:63: error: Item "AWSAccountCleanupOptionAMIV1" of "AWSAccountCleanupOptionCloudWatchV1 | AWSAccountCleanupOptionAMIV1 | AWSAccountCleanupOptionV1" has no attribute "delete_empty_log_group"  [union-attr]
reconcile/aws_cloudwatch_log_retention/integration.py:63: error: Item "AWSAccountCleanupOptionV1" of "AWSAccountCleanupOptionCloudWatchV1 | AWSAccountCleanupOptionAMIV1 | AWSAccountCleanupOptionV1" has no attribute "delete_empty_log_group"  [union-attr]
```

I'm not sure how to resolve this in a way that is most acceptable for our code base.
My first instinct is to wrap things in try catches, but that feels ugly.
Or to convert this part back to a dict, which also feels bad, since typing is what we are trying to achieve.

hemslo: oh, this is a typing issue, first let's update gql to not have `AWSAccountCleanupOptionAMIV1`

```gql
    cleanup {
      provider
      ... on AWSAccountCleanupOptionCloudWatch_v1 {
        regex
        retention_in_days
        delete_empty_log_group
        region
      }
    }
```

run `make gql-query-classes`

then do a type filter, can also simplify code to save some nesting condition


```python
    default_region = account.resources_default_region
    result = defaultdict(list)
    for cleanup_option in account.cleanup or []:
        if isinstance(cleanup_option, AWSAccountCleanupOptionCloudWatchV1):
```

PR #4957, opened 2025-04-23
[APPSRE-11838] fix openshift-saas-deploy-trigger-configs

BumbleFeng:
openshift-saas-deploy-trigger-configs should not ignore managedResourceNames change


BumbleFeng:
Currently, if the tenant makes a change to managedResourceNames for the saas file e.g., https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/139585  openshift-saas-deploy-trigger-configs would ignore the change. The desired behavior is to trigger a saas deploy job to apply the change.

PR #4958, opened 2025-04-23
APPSRE-11824 - automate updating labels/taints for HCP nodes when NodePool updated

dwelch0:


reconcile/utils/ocm/ocm.py
dwelch0, line 1: Adding the taints logic resolves an existing bug: taints added to a `machinePool` in app-interface cluster file were successfully applied to the machinePool. However, if the taint was later removed from app-interface, this removal was not reconciled on the HCP nodePool.

**NOTE:** this patch assumes that app-interface is single source of truth for machinePool taints. This will overwrite all outstanding taints with only what exists in the cluster's `machinePool` section.

PR #4961, opened 2025-04-24
[elasticache] modify required secrets filtering for elasticache to use set syntax

cubismod:
Instead of mandating that a tenant only has `master_user_name` & `master_user_password` in their Elasticache secret, we instead check if they have the two required keys by checking if they are a subset of all the keys in the secret. This is so we can add additional secret values to be included with the secret (such as Splunk tokens).

Validated this locally against Quay stage elasticache in dry-run.

[APPSRE-11853](https://issues.redhat.com/browse/APPSRE-11853)


chassing:
All Elasticache instances have been migrated to ERv2

cubismod:
> All Elasticache instances have been migrated to ERv2

Ooops :facepalm: ignore me :sweat_smile:

cubismod:
Actually wait sorry this is **Elasticsearch** not **Elasticcache**

chassing:
> Actually wait sorry this is **Elasticsearch** not **Elasticache**

LOL. There is a huge difference between them ;)

chassing (approved):

PR #4962, opened 2025-04-24
[elasticsearch] allow additional keys in admin credentials

cubismod:
Follow-up from #4961

[APPSRE-11853](https://issues.redhat.com/browse/APPSRE-11853)

This PR allows users to add additional secrets to their Elasticsearch `admin_user_credentials` secret which can then be used when setting up a Kinesis lambda (https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/utils/terrascript_aws_client.py#L3994). The additional values are stripped out when provisioning the Elasticache domain to meet the expected schema from AWS.

```yaml
$schema: /aws/elasticsearch-defaults-1.yml
ebs_options:
  ebs_enabled: true
  volume_size: 350

auth:
  admin_user_credentials:
    path: <your_vault_path_here>
```

### DIFF
--- a/reconcile/utils/terrascript_aws_client.py
+++ b/reconcile/utils/terrascript_aws_client.py
@@ -4901,10 +4901,22 @@ class TerrascriptClient:  # pylint: disable=too-many-public-methods
                     )
                 )
         # ++++++++ END: REMOVE ++++++++++
-        # add master user creds to output and secretsmanager if internal_user_database_enabled
         if advanced_security_options:
+            master_user_options_with_optional_keys_values = advanced_security_options[
+                "master_user_options"
+            ]
+            # this secret can include optional kv pairs which are then saved to secrets manager in AWS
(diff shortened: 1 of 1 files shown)


reconcile/utils/terrascript_aws_client.py
```diff
@@ -4900,6 +4901,81 @@ def populate_tf_resource_elasticsearch(self, spec):
                     )
                 )
         # ++++++++ END: REMOVE ++++++++++
+        # add master user creds to output and secretsmanager if internal_user_database_enabled
+        if advanced_security_options:
+            if advanced_security_options.get("internal_user_database_enabled", False):
```
esron, line 4906: Nested ifs can be an and. That saves us one indent level.

cubismod: I actually can't use a nested if here, pushed a commit to move some of the code around and make it clearer

esron (approved):

PR #4967, opened 2025-05-05
APPSRE-11886: Add support to migrate HCL/terraform modules

dwelch0:
## Summary

Currently, `external-resources` integration `migrate` functionality only supports modules of type `cdktf`. `cdktf` has been deprecated in favor of HCL(type `terraform`) modules. This PR adds state migration support for HCL modules.

Depends on ERv2 modules built on base image that includes change in https://github.com/app-sre/er-base-terraform/pull/27


chassing (approved):
LGTM

PR #4969, opened 2025-05-07
APPSRE-11428 bump dt lib

fishi0x01:
DT lib now has a fix for the date time parsing issue https://github.com/dynatrace-oss/api-client-python/pull/110

After bump we do not need the patch fix anylonger.


hemslo (approved):
LGTM!

PR #4976, opened 2025-05-13
fix(quay_mirror_org): update util function usages

hemslo:
`quay_mirror_org` is also using the same function, updates usage after function move from https://github.com/app-sre/qontract-reconcile/pull/4974.

Also sync uv.lock file, not sure why it's missed in https://github.com/app-sre/qontract-reconcile/pull/4969.


mehfuz (approved):
LG!

cubismod:
Thanks for catching this!

PR #4982, opened 2025-05-16
fix(gitlab_fork_compliance): use members_all instead of members

hemslo:
Fix wrong methods replace in https://github.com/app-sre/qontract-reconcile/pull/4981/files#diff-1d5ab1abb6ecb1d8c243a9e8c04d4cf6505e6be2c8932df654ea9a11c6c1b64bR271.

Should use `members_all` instead of just `members`.

[APPSRE-10450](https://issues.redhat.com/browse/APPSRE-10450)


mehfuz (approved):
lgtm

PR #4984, opened 2025-05-16
add slo library to SAPM

mehfuz:
1. Contains code to evaluate SLOs for a target if provider. If the SLOs are breached then the promotion is blocked. In such 
case only if the commit id is present in `hotfixVersions` the promotion are proceed.
2.Contains testcases to validate above logic.
Local testing:-
 Bumped the `SLOTarget` of  github-mirror  `GithubMirrorHighLatency` SLO to 1.0 to simulate breached SLO (its current value is ~ 0.99). Works as expected with the following log
```
(qontract-reconcile) ➜  qontract-reconcile git:(SDE-4778-01) qontract-reconcile  --config ~/dev/config.debug.toml  --dry-run  saas-auto-promotions-manager  --app-name github-mirror
[2025-05-21 11:30:17] [INFO] [DRY-RUN] [subscriber.py:_has_breached_slos:219] - Subscriber at path /services/github-mirror/cicd/deploy.yaml promotion stopped because following breached SLOs
[2025-05-21 11:30:17] [INFO] [DRY-RUN] [subscriber.py:_has_breached_slos:224] - SLO:GithubMirrorHighLatency of document github-mirror is breached. Current value:0.997266 Expected:1.000000 
```


reconcile/saas_auto_promotions_manager/utils/saas_files_inventory.py
```diff
@@ -87,9 +96,14 @@ def _assemble_publishers(self) -> None:
     def _assemble_subscribers_with_auto_promotions(self) -> None:
         for saas_file in self._saas_files:
             blocked_versions: dict[str, set[str]] = {}
+            hotfix_versions: dict[str, set[str]] = {}
```
hemslo, line 99: can use [defaultdict(set)](https://docs.python.org/3/library/collections.html#collections.defaultdict) to simplify add and get

reconcile/saas_auto_promotions_manager/subscriber.py
```diff
@@ -203,6 +212,25 @@ def _compute_desired_ref(self) -> None:
         # Passed all gates -> lets promote desired ref
         self.desired_ref = desired_ref
 
+    def _has_breached_slos(self, desired_ref: str) -> bool:
+        if self.slo_document_manager and desired_ref not in self._hotfix_versions:
+            breached_slos = self.slo_document_manager.get_breached_slos()
+            if breached_slos:
+                logging.info(
```
hemslo, line 219: will this log spam in every integration run?

mehfuz: @hemslo if there are breached slos for that target then we would see this in the integration runs since the promotion will be blocked for that target.

mehfuz: added [here](https://github.com/app-sre/qontract-reconcile/pull/4984/commits/49a95b6a70a0d1589f9558daac0b93f1a09f2a13)

reconcile/saas_auto_promotions_manager/utils/saas_files_inventory.py
mehfuz: added [here](https://github.com/app-sre/qontract-reconcile/pull/4984/commits/49a95b6a70a0d1589f9558daac0b93f1a09f2a13)

hemslo (approved):
LGTM!

PR #4985, opened 2025-05-19
:recycle: jira-boards: common issueFields 

chassing:
Adapt `jira-permissions-validator` and `acs-notifiers` to verify and use the common `jira-board.issueFields` as a replacement for `issueSecurityId`.

This allows any arbitrary issue field definitions for our tenants, such as the new `Work Type` one.

Ticket: [APPSRE-12034](https://issues.redhat.com/browse/APPSRE-12034)
Depends on: https://github.com/app-sre/qontract-schemas/pull/828

### DIFF
--- a/reconcile/jira_permissions_validator.py
+++ b/reconcile/jira_permissions_validator.py
@@ -56,12 +56,15 @@ class ValidationError(IntFlag):
     CANT_TRANSITION_ISSUES = auto()
     INVALID_ISSUE_TYPE = auto()
     INVALID_ISSUE_STATE = auto()
-    INVALID_SECURITY_LEVEL = auto()
+    INVALID_ISSUE_FIELD = auto()
     INVALID_PRIORITY = auto()
     PERMISSION_ERROR = auto()
     PUBLIC_PROJECT_NO_SECURITY_LEVEL = auto()
     INVALID_COMPONENT = auto()
     PROJECT_ARCHIVED = auto()
+
+
+CustomIssueFields = dict[str, dict[str, str]]
 
 
 class RunnerParams(TypedDict):
@@ -81,12 +84,14 @@ def board_is_valid(
     default_reopen_state: str,
     jira_server_priorities: NameToIdMap,
     public_projects: Iterable[str],
-) -> ValidationError:
+) -> tuple[ValidationError, CustomIssueFields]:
     error = ValidationError(0)
+    custom_fields: CustomIssueFields = {}
+
     try:
         if jira.is_archived:
             logging.error(f"[{board.name}] project is archived")
-            return ValidationError.PROJECT_ARCHIVED
+            return ValidationError.PROJECT_ARCHIVED, custom_fields
 
         if not jira.can_create_issues():
             logging.error(f"[{board.name}] can not create issues in project")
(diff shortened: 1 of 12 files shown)


esron (approved):

reconcile/jira_permissions_validator.py
```diff
@@ -109,54 +114,71 @@ def board_is_valid(
...
-                    ])
+                    f"[{board.name}] '{board.issue_resolve_state}' is not a valid state in project. Valid states: {project_issue_type.statuses}"
                 )
-                error |= ValidationError.INVALID_SECURITY_LEVEL
-        elif board.name in public_projects:
+                error |= ValidationError.INVALID_ISSUE_STATE
+
+            for field in board.issue_fields or []:
+                if not (
+                    project_issue_field := jira.project_issue_field(
```
hemslo, line 153: `project_issue_field` is also used outside of `if` section

reconcile/utils/jira_client.py
```diff
@@ -48,6 +43,50 @@ class IssueType(BaseModel):
     statuses: list[str]
 
 
+class FieldOption(BaseModel):
+    """A standard buildin issue field option."""
+
+    name: str
+
+    def __eq__(self, value: Any) -> bool:
```
hemslo, line 51: why override eq for `FieldOption` and `CustomFieldOption`, and only `value` from `JiraBoardIssueFieldV1` used to compare with it?

reconcile/utils/jira_client.py
```diff
@@ -86,6 +125,12 @@ def __init__(
         self.priorities = functools.lru_cache(maxsize=None)(self._priorities)
         self.public_projects = functools.lru_cache(maxsize=None)(self._public_projects)
         self.my_permissions = functools.lru_cache(maxsize=None)(self._my_permissions)
+        self.project_issue_types = functools.lru_cache(maxsize=None)(
+            self._project_issue_types
+        )
+        self.project_issue_fields = functools.lru_cache(maxsize=None)(
+            self._project_issue_fields
+        )
```
hemslo, line 133: can use `functools.cache`

reconcile/jira_permissions_validator.py
```diff
@@ -109,54 +114,71 @@ def board_is_valid(
                 error |= ValidationError.INVALID_COMPONENT
 
         issue_type = board.issue_type or default_issue_type
-        project_issue_types = jira.project_issue_types()
-        project_issue_types_str = [i.name for i in project_issue_types]
-        if issue_type not in project_issue_types_str:
+        if not (project_issue_type := jira.get_issue_type(issue_type)):
```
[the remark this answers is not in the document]
chassing: Why? `project_issue_type` is defined within the `if` block and afterwards.

```
project_issue_type = jira.get_issue_type(issue_type)
if not project_issue_type:
    ...
```

would be the same.

reconcile/jira_permissions_validator.py
chassing: See above

reconcile/utils/jira_client.py
chassing: I want to compare the `JiraBoardIssueFieldV1.value` strings against either `FieldOption` or  `CustomFieldOption`, regardless of the `name` vs `value` attributes. The Jira API is stupid, it's `name` for standard jira fields, but `value` for custom ones :(  

```
{
  "security": {
    "name": "Employee"
  },
  "custom_attr": {
    "value": "whatever"
  }
}

File: reconcile/utils/jira_client.py (hunk 3 above)
Comment (chassing) [reply reconcile/utils/jira_client.py]: Done in [aa589de](https://github.com/app-sre/qontract-reconcile/pull/4985/commits/aa589de057753c09565dc6e1ec435fc584d4a186)

File: reconcile/jira_permissions_validator.py [hunk 5]
```diff
@@ -109,54 +114,71 @@ def board_is_valid(
...
+            project_issue_types_str = ", ".join([
```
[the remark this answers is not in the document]
chassing: Done in [ce1e3fd](https://github.com/app-sre/qontract-reconcile/pull/4985/commits/ce1e3fdcdc05f80f5035d894d259289b780075f0). I should enable more ruff checks to avoid that!

reconcile/jira_permissions_validator.py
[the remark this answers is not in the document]
hemslo: because it confuse people used to scoped var https://stackoverflow.com/questions/67680653/unexpected-python-variable-scope-using-walrus-operator-assignment-expression

reconcile/utils/jira_client.py
hemslo: do we need to differentiate between `FieldOption` and `CustomFieldOption`, or just flat them into `str` to let `options` be `list[str]`?

chassing: Yes, we need to differentiate between them. It's done in the [caching part](https://github.com/app-sre/qontract-reconcile/blob/878c0c86ba88338ee17734a689e72a7203349837/reconcile/jira_permissions_validator.py#L167). This will produce the dictionary mentioned above

hemslo: so in jira alert template, we need to keep the same format as server response, use `name` for standard fields, `value` for custom fields?

chassing: Unfortunately, yes

hemslo (approved):
LGTM!

PR #4986, opened 2025-05-20
Erv2 modules progressive rollout

lechuk47:
- **Fix fypo in deploy_resources fragment**
- **Erv2: Modules gradual rollout**

Requires: https://github.com/app-sre/qontract-schemas/pull/829


reconcile/external_resources/model.py
```diff
@@ -289,15 +300,37 @@ def resolve_configuration(
...
+            None,
+        )
+
+        if channel_data is None:
+            raise ExternalResourceModuleConfigurationError(
+                f"Module {module.provision_provider}/{module.provider} does not have the channel {channel_name}. "
+                f"Resource spec: {spec.id_object()}"
+            )
+
+        channel_image, channel_version = channel_data
```
hemslo, line 329: is pack and unpack needed here as we can just find channel first, then use fields like this

```python
        channel = next(
            (c for c in module.channels if c.name == channel_name),
            None,
        )
        ...
        return ExternalResourceModuleConfiguration(
            image=module_overrides.image or channel.image,
            version=module_overrides.version or channel.version,
```

reconcile/test/external_resources/test_module_configuration.py
```diff
@@ -1,48 +1,132 @@
...
+
+    conf = ExternalResourceModuleConfiguration.resolve_configuration(
+        module=module, spec=spec, settings=settings
     )
+    assert conf.image == expected_image
+    assert conf.version == expected_version
+    if overrides:
+        assert conf.outputs_secret_image == overrides.outputs_secret_image
+        assert conf.outputs_secret_version == overrides.outputs_secret_version
+        assert conf.resources == overrides.resources
```
hemslo, line 114: `if` condition should be avoided in tests, can we include `metadata` inside params and assert on full `conf` object?

reconcile/external_resources/model.py
lechuk47: Fixed.

reconcile/test/external_resources/test_module_configuration.py
lechuk47: Metadata is no part of the constructor. I added the whole configuration object. PTAL.

hemslo (approved):
LGTM!

PR #4988, opened 2025-05-21
:sparkles: automated-actions-config: support action-list arguments

chassing:
Support `action-list` arguments


reconcile/automated_actions/config/integration.py
```diff
@@ -158,8 +158,13 @@ def compile_roles(
                             "name": arg.name_pattern,
                         })
                     case _:
```
patjlm, line 160: shouldn't this be updated to `case AutomatedActionArgumentActionListV1:` and keep the default unsupported `case _:`?

PR #4992, opened 2025-05-21
Deprecate slack-username

biswassri:
Fixes: [APPSRE-11644](https://issues.redhat.com/browse/APPSRE-11644)

Additionally - Fixes Readme changes


reconcile/utils/slack_api.py
```diff
@@ -419,7 +419,7 @@ def get_active_users_by_names(self, user_names: Iterable[str]) -> dict[str, str]
         return {
             k: v["name"]
             for k, v in self._get("users").items()
-            if v["name"] in user_names and not v["deleted"]
+            if v["profile"]["email"].split("@")[0] in user_names and not v["deleted"]
```
hemslo, line 422: this will fail for users without email, such as deleted users, bot users, and special `slackbot` user.

please create a method to extract name from slack user, and remove all usages of "name", including `get_active_users_by_names` and `get_users_by_ids`, so when we compare desired and current state, only valid values are org username, this is recommended by slack api doc https://api.slack.com/types/user

> name | String | Don't use this.

```python
    @staticmethod
    def extract_name_from_user(user: dict[str, Any]) -> str | None:
        if email := user["profile"].get("email"):
            return email.split("@")[0]
        return None
```

Then apply it to `get_active_users_by_names` like this:

```python
        return {
            k: name
            for k, v in self._get("users").items()
            if not v["deleted"]
            and (name := self.extract_name_from_user(v))
            and name in user_names
        }
```

do the same for `get_users_by_ids`:

```python
        return {
            user_id: name
            for user_id in translated_user_ids
            if (user := users.get(user_id))
            and (name := self.extract_name_from_user(user))
        }
```

please also check if these changes are covered in tests, and do a dry run locally.

biswassri: @hemslo updated the PR with the above changes and other necessary test changes. Let me know if there are any other change suggestion :)

hemslo: in `get_active_users_by_names`, missed one name usage change, `k: v["name"]` -> `k: name`

biswassri: @hemslo  Thanks. I'll update that. 

> you can update some test cases to use different email and name to show that we don't rely on name field anymore.

Do you mean add new tests like `def test_extract_name_from_user` ?

hemslo: cover that new util method `extract_name_from_user` is optional, tests should cover existing public methods `get_active_users_by_names` and `get_users_by_ids`.

biswassri: @hemslo  I added some more tests to cover `get_active_users_by_names`. Let me know if this looks good to you.

hemslo: two active slack users with the same email in profile is unlikely to happen, but it's ok in test data to show our code can handle it as well

hemslo:
@biswassri I accidentally rebased this PR, and found konflux build still failed, please sync git tags from upstream to your fork, then rebase & force push again.

BumbleFeng (approved):

PR #4993, opened 2025-05-22
[gitlab] Use archive api to download dir

hemslo:
Use [Get file archive](https://docs.gitlab.com/api/repositories/#get-file-archive) to download tar.gz file for a dir instead of iterate repo tree then download files one by one.

Also refactor determine if a repo is for GitHub or GitLab by extract `GITHUB_BASE_URL` and do strict prefix match to avoid gitlab repo name contains `github` to be classified as GitHub repo.

[APPSRE-12038](https://issues.redhat.com/browse/APPSRE-12038)

### Refactoring for GitHub URL Handling:
* Added `GITHUB_BASE_URL` constant in `reconcile/utils/vcs.py` to centralize the GitHub base URL.
* Replaced hardcoded GitHub URL strings with `GITHUB_BASE_URL` in various files and functions, such as `compare`, `_github_compare_commits`, and `_get_directory_contents`. [[1]](diffhunk://#diff-a1ca374a2cf09b0d017bff4299973c839cdc1380b48f5332350204abf7c43feeL424-R425) [[2]](diffhunk://#diff-a1ca374a2cf09b0d017bff4299973c839cdc1380b48f5332350204abf7c43feeL479-R480) [[3]](diffhunk://#diff-52a815feb4bcd8276212403b5f0ae22e3a3d9508acb5b7a75a5134dcf2657bbcL763-R765) and others)
* Updated logic in multiple modules, including `dashdotdb_dora.py`, `github_repo_permissions_validator.py`, and `saasherder.py`, to use `GITHUB_BASE_URL` for URL manipulation and comparisons. [[1]](diffhunk://#diff-a1ca374a2cf09b0d017bff4299973c839cdc1380b48f5332350204abf7c43feeR37) [[2]](diffhunk://#diff-178b0eea2ebeecaca9755b66defe43f0bd821dd1b266f3d9f433f670c2193759R14) [[3]](diffhunk://#diff-52a815feb4bcd8276212403b5f0ae22e3a3d9508acb5b7a75a5134dcf2657bbcL745-R746)

### Enhancements to GitLab API:
* Added a new static method, `get_directory_contents`, in `reconcile/utils/gitlab_api.py` to retrieve directory contents from a GitLab repository archive.
* Implemented a corresponding test case, `test_get_directory_contents`, in `test_gitlab_api.py` to validate the new functionality.

### Other Changes:
* Updated test data to use HTTPS for GitHub URLs in `test_terrascript_aws_client.py`.
[...]

PR #4994, opened 2025-05-26
refactor(vcs): extract parse_repo_url

hemslo:
follow up on https://github.com/app-sre/qontract-reconcile/pull/4993#discussion_r2102062542

Use one util function `parse_repo_url` to avoid repeated checks.

[APPSRE-12038](https://issues.redhat.com/browse/APPSRE-12038)


chassing (approved):
Nice ❤️

PR #5016, opened 2025-06-10
:label: more type hints

chassing:
Add type hints support to several integrations and tests.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -274,24 +274,20 @@ def handle_stale_items(
 
     now = datetime.utcnow()
     for item in items:
-        item_iid = item.attributes.get("iid")
-        assert item_iid  # make mypy happy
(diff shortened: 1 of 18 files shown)


jfreixa (approved):

reconcile/gitlab_housekeeping.py
```diff
@@ -263,25 +263,28 @@ def close_item(
...
+    enable_closing: bool,
+    items: Iterable[ProjectIssue | ProjectMergeRequest],
+    item_type: str,
+) -> None:
     LABEL = "stale"
 
     now = datetime.utcnow()
     for item in items:
         item_iid = item.attributes.get("iid")
+        assert item_iid  # make mypy happy
```
hemslo, line 278: is this needed if we do `item_iid = item.iid`? same for `updated_at`

hemslo (approved):
LGTM!

PR #5039, opened 2025-06-20
Change gitlab api _is_bot_username documentation

jfreixa:
We considered GitLab's "List all billable members of a group" API (/groups/:id/billable_members), but it didn't meet our needs for two main reasons:

Includes Project Users: It returns users from nested projects, not just direct group members. While we could filter these out, it's an extra step.
Excludes Inactive Users: Crucially, this endpoint doesn't include inactive users, which we need for comprehensive user management.
[Endpoint Info](https://docs.gitlab.com/api/members/#list-all-billable-members-of-a-group)

[APPSRE-8777](https://issues.redhat.com/browse/APPSRE-8777)

### DIFF
--- a/reconcile/utils/gitlab_api.py
+++ b/reconcile/utils/gitlab_api.py
@@ -298,16 +298,15 @@ class GitLabApi:
 
     @staticmethod
     def _is_bot_username(username: str) -> bool:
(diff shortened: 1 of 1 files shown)


hemslo (approved):
LGTM!

PR #5040, opened 2025-06-20
🧹 Remove integration github-users

jfreixa:


jonmosco:
We should probably remove the tests related to this as well.

PR #5041, opened 2025-06-20
feat: Add target_ref to saas deploy trigger

cubismod:
The `target_ref` is the SHA ref of the target and is added as a label to the PipelineRun resource.
We're using pod labels & annotations to build a dashboard for saas deploy logs and a tenant requested the abillity to query by targetted refs, see https://redhat-internal.slack.com/archives/CCRND57FW/p1750433539235259


hemslo (approved):
LGTM!

PR #5048, opened 2025-06-23
fix target ref in trigger reason

ritmun:
target_ref is showing as empty for job built from image updates. 

eg
```
 labels:
    qontract.env_name: osd-integration-hivei01ue1
    qontract.saas_file_name: saas-oeo
    qontract.target_ref: ''
    tekton.dev/pipeline: o-saas-deploy-saas-oeo
```

It might be fixed by using a flexible desired state param available , i.e. state_content


reconcile/openshift_saas_deploy_trigger_base.py
```diff
@@ -269,7 +269,7 @@ def _trigger_tekton(
 
     target_ref = None
     if isinstance(spec, TriggerSpecConfig):
```
cubismod, line 271: You can remove the `isinstance` check since this param is available on all `spec` types.

PR #5063, opened 2025-06-27
🐛: status board exporter id dependency

esron:
Now, if a status board object is in the current map we get the id and put back on the desired state. That is later used on the `create`, `update` or `delete` method to assign the correct parent id.

Related to: [APPSRE-10944](https://issues.redhat.com/browse/APPSRE-10944)


reconcile/status_board.py
```diff
@@ -281,21 +276,25 @@ def get_current_products_applications_services(
...
+                    Service(**s, application=a)
+                    for s in get_application_services(ocm_api, a.id)
                 ]
 
         return products
 
     @staticmethod
     def desired_abstract_status_board_map(
-        desired_product_apps: Mapping[str, set[str]], slodocs: list[SLODocumentV1]
+        current_status_board_map: Mapping[str, AbstractStatusBoard],
```
hemslo, line 295: this will make desired state depends on current state to construct, while these id fields are only needed in diff and reconcile, how about keep these id as None for desired state, but sync them during diff in one go for all classes of `AbstractStatusBoard`, like this:


```python
        for pair in chain(diff_result.identical.values(), diff_result.change.values()):
            pair.desired.id = pair.current.id
```

reconcile/status_board.py
```diff
@@ -305,17 +304,21 @@ def desired_abstract_status_board_map(
...
+                current_app = current_status_board_map.get(key)
+                app_id = current_app.id if current_app else None
                 desired_abstract_status_board_map[key] = Application(
+                    id=app_id,
                     name=a,
                     fullname=key,
                     services=[],
-                    product=desired_abstract_status_board_map[product],
-                    metadata={},
+                    product=desired_abstract_status_board_map.get(product),
```
hemslo, line 321: please rename current local var `product` to `product_name`, and extract a var `product` to hold instance of `Product`, use it to populate map and application, to avoid warning `Expected type 'Product', got 'AbstractStatusBoard | None' instead`:

```python
for product_name, apps in desired_product_apps.items():
    product = Product(...)
    desired_abstract_status_board_map[product_name] = product
    for a in apps:
        ...
        = Application(..., product=product)
```

reconcile/status_board.py
esron: Neat. Done.

reconcile/status_board.py
```diff
@@ -166,12 +163,12 @@ class Service(AbstractStatusBoard):
     # This filed is needed when we are creating a Service on teh OCM API.
     # This field is not used when we are mapping the services that belongs to an
     # application in that case we use the `services` field in Application class.
-    application: Optional["Application"]
+    application: Application
```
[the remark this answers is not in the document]
esron: I removed it.

reconcile/status_board.py
esron: Done.

hemslo (approved):
LGTM!

reconcile/status_board.py
```diff
@@ -461,13 +452,15 @@ def run(self, dry_run: bool) -> None:
...
             )
 
-            current_products = {
-                p.name: p for p in current_products_applications_services
-            }
+            # Desired state
+            desired_product_apps: dict[str, set[str]] = self.get_product_apps(sb)
+            desired_abstract_status_board_map = self.desired_abstract_status_board_map(
+                desired_product_apps, slodocs
+            )
```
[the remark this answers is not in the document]
esron: Done

PR #5092, opened 2025-07-10
APPSRE-11020 pypi push

fishi0x01:
Konflux should also push our package to pypi. Note, for now the push is still commented out, as ci.ext is still doing that job. When we move away from ci.ext, we just have to uncomment that line and it will take over.

### DIFF
--- a/dockerfiles/Dockerfile.konflux
+++ b/dockerfiles/Dockerfile.konflux
@@ -117,17 +117,15 @@ COPY --from=test-image /is_tested /is_tested
 # STAGE 7 - PyPI publish package
 ###############################################################################
 FROM test-image AS pypi
-ARG TWINE_USERNAME
-ARG TWINE_PASSWORD
 
 # Lets make sure we ran previous prod stage before uploading to pypi
 COPY --from=prod-image /is_tested /is_tested
 RUN echo "true" > /is_pypi_pushed
 
 # qontract-reconcile version depends on git tags!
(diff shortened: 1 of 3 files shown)


dockerfiles/Dockerfile.konflux
```diff
@@ -117,17 +117,15 @@ COPY --from=test-image /is_tested /is_tested
...
 COPY --from=prod-image /is_tested /is_tested
 RUN echo "true" > /is_pypi_pushed
 
 # qontract-reconcile version depends on git tags!
-# The .git dir should already be part of the test image.
-# TODO
-# RUN make pypi
+# The .git dir should already be part of the test image
+# TODO: re-enable to take over from ci.ext
+#RUN --mount=type=secret,id=app-sre-pypi-credentials/token export UV_PUBLISH_TOKEN=$(cat /run/secrets/app-sre-pypi-credentials/token); make pypi-konflux
```
hemslo, line 128: can just be
`RUN --mount=type=secret,id=app-sre-pypi-credentials/token,env=UV_PUBLISH_TOKEN make pypi-konflux`

fishi0x01: `env` is not supported bu buildah

chassing (approved):

dockerfiles/Dockerfile.konflux
hemslo: 😿

PR #5093, opened 2025-07-11
:sparkles: profiling support

chassing:
Support Python [cProfile](https://docs.python.org/3/library/profile.html). 

environment variables:
* `ENABLED_PROFILING` for enabling the profiling 
* `PROFILE_DUMP_FILE` configure the path to dump file; default `/tmp/profile.prof`

### DIFF
--- a/reconcile/run_integration.py
+++ b/reconcile/run_integration.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
 
+import contextlib
+import cProfile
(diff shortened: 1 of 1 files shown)


hemslo:
`cProfile` threaded run is not well supported, I use [py-spy](https://github.com/benfred/py-spy) for cpu, [memray](https://github.com/bloomberg/memray) for memory.

PR #5094, opened 2025-07-11
:rotating_light: pep8 pydantic validators

chassing:
Configure ruff pep8 checks for pydantic validators instead of ignoring them in the models files.


hemslo (approved):
LGTM!

PR #5099, opened 2025-07-15
[aws-account-manager] support IT managed accounts

chassing:
We want to onboard and manage AWS accounts created by IT. These accounts have restrictions on changing the alias and the security contact information. The plan is to ignore all AWS API errors as long as the app-interface settings match those in the account; otherwise, throw an exception as usual.

Ticket: [APPSRE-12236](https://issues.redhat.com/browse/APPSRE-12236)

### DIFF
--- a/reconcile/utils/aws_api_typed/iam.py
+++ b/reconcile/utils/aws_api_typed/iam.py
@@ -1,5 +1,6 @@
 from typing import TYPE_CHECKING
 
+import botocore
 from pydantic import BaseModel, Field
 
 if TYPE_CHECKING:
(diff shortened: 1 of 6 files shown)


reconcile/utils/aws_api_typed/iam.py
```diff
@@ -60,8 +63,15 @@ def set_account_alias(self, account_alias: str) -> None:
...
-                raise ValueError(
-                    "Account alias already exists for another AWS account. Choose another one!"
-                ) from None
+        except botocore.exceptions.ClientError as e:
+            if e.response["Error"]["Code"] == "EntityAlreadyExists":
+                if self.get_account_alias() != account_alias:
+                    raise ValueError(
+                        "Account alias already exists for another AWS account. Choose another one!"
+                    ) from e
+            elif e.response["Error"]["Code"] == "AccessDenied":
```
hemslo, line 72: is `botocore.exceptions.ClientError` same as `self.client.exceptions.ClientError`? Is `AccessDenied` missing from exception class? `self.client.exceptions.EntityAlreadyExistsException` looks better in code

chassing: Yes, it looks better, but testing is harder. If I mock `IAMClient`, it also mocks out the `client.exceptions` and raised `TypeError: catching classes that do not inherit from BaseException is not allowed`; therefore using `botocore.exceptions.ClientError` is easier.

`client.exception` doesn't have all "basic" exceptions like `AccessDenied`, you need to use `botocore.exceptions.ClientError` instead. The boto3 exception handling is annoying!

hemslo: yes, I have to manually add that exception for test https://github.com/app-sre/er-aws-rds/blob/69836d1f401039a2c621ebe273def08b55ff8d0f/tests/test_aws_api.py#L214, I verified it in runtime that it's that concrete exception, but not test friendly.

PR #5103, opened 2025-07-16
:rotating_light: remove obsolete pylint exceptions

chassing:
`pylint` isn't in used anymore; therefore we don't need those exception comments anymore.


hemslo (approved):
LGTM!

PR #5111, opened 2025-07-22
update dev deps

chassing:
- **:arrow_up: upgrade dependencies**
- **:rotating_light: fix mypy errors**


jfreixa (approved):
LGTM

reconcile/test/test_version_bump.py
```diff
@@ -1,18 +0,0 @@
-import os
```
rporres, line 1: why this file got deleted?

tools/qontract_cli.py
```diff
@@ -1031,6 +1031,9 @@ def clusters_network(ctx: click.Context, name: str) -> None:
                 ocm_map.get(cluster_name),
                 provided_assume_role=None,
             )
+            assert account is not None, (
+                f"Failed to build assume role for cluster {cluster_name} in account {management_account['name']}"
+            )
```
rporres, line 1036: Where this come from?

chassing: `mypy` complains about `account` can be `None`

reconcile/test/test_version_bump.py
chassing: The test was failing, and I thought instead of fixing it, I would remove it.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -994,7 +994,8 @@ def get_resource_lifecycle(
             if lifecycle.ignore_changes is None:
                 lifecycle.ignore_changes = []
             if "all" in lifecycle.ignore_changes:
-                lifecycle.ignore_changes = "all"
+                # 'all' must be passed as a string! stupid terraform aws provider!
+                lifecycle.ignore_changes = "all"  # type: ignore[assignment]
```
hemslo, line 998: `lifecycle.ignore_changes` type is defined from gql, let's not assign wrong types to it, change return value instead:

```python
            ignore_changes = (
                "all" if "all" in lifecycle.ignore_changes else lifecycle.ignore_changes
            )
            return lifecycle.dict(by_alias=True) | {"ignore_changes": ignore_changes}
```

reconcile/test/test_version_bump.py
chassing: Nevertheless, I won't die on that hill. [a96b11c](https://github.com/app-sre/qontract-reconcile/pull/5111/commits/a96b11c4558714879c7d64558fad95e4d7cdb798) fixes it

hemslo (approved):
LGTM!

rporres (approved):

reconcile/test/test_version_bump.py
chassing: I'll remove this test in another PR.

PR #5112, opened 2025-07-22
:rotating_light: enable flake8-type-checking

chassing:
Enable all ruff `flake8-type-checking` rules.

This avoids code like this:

```python
if TYPE_CHECKING:
    from mypy_boto3_ec2 import EC2Client
    from mypy_boto3_ec2.type_defs import ImageTypeDef
    from mypy_boto3_iam import IAMClient
    from mypy_boto3_route53 import Route53Client
    from mypy_boto3_route53.type_defs import (
        ChangeBatchTypeDef,
        HostedZoneTypeDef,
        ResourceRecordSetTypeDef,
        ResourceRecordTypeDef,
    )

else:
    EC2Client = IAMClient = ImageTypeDef = Route53Client = ResourceRecordTypeDef = (
        HostedZoneTypeDef
    ) = ChangeBatchTypeDef = ResourceRecordSetTypeDef = object
```

Additionally, add `from __future__ import annotations` to the top of the file, if needed. This allows us to use [forward references](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) without needing to use string annotations. Example:

```python

from typing import TYPE_CHECKING

from whateverimport Whatever

if TYPE_CHECKING:
    from ... import MyClass

def func(param: MyClass) -> Whatever:
    return Whatever(param)
```

Instead of:

```python
from typing import TYPE_CHECKING

from whateverimport Whatever

if TYPE_CHECKING:
    from ... import MyClass

def func(param: "MyClass") -> Whatever:
    return Whatever(param)
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -157,6 +157,9 @@ src = ["reconcile", "tools"]
 extend-exclude = ["reconcile/gql_definitions"]
 fix = true
 
+[tool.ruff.format]
+preview = true
+
 [tool.ruff.lint]
 preview = true
 
@@ -215,8 +218,12 @@ ignore = [
     "FURB101", # Replace `open` with `pathlib.Path`
     "FURB103", # Replace `open` with `pathlib.Path`
 ]
-[tool.ruff.format]
-preview = true
+
+[tool.ruff.lint.flake8-annotations]
+allow-star-arg-any = true
+
+[tool.ruff.lint.flake8-type-checking]
+runtime-evaluated-base-classes = ["pydantic.BaseModel"]
 
 [tool.ruff.lint.isort]
 known-first-party = ["reconcile", "tools"]
--- a/reconcile/aus/models.py
+++ b/reconcile/aus/models.py
@@ -1,4 +1,3 @@
-# ruff: noqa: TC001 - pydantic needs those classes during runtime
 from __future__ import annotations
 
 import logging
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -213,11 +213,6 @@ ignore = [
     "B017",    #`pytest.raises(Exception)` should be considered evil
     "RUF005",  # Consider "...."instead of concatenation
     "RUF012",  # Mutable class attributes should be annotated with "typing.ClassVar"
-    "TC001",   # Type-checking block
-    "TC002",   # Type-checking block
-    "TC003",   # Type-checking block
-    "TC004",   # Type-checking block
```
hemslo, line 219: do we want to auto add future annotations? https://docs.astral.sh/ruff/settings/#lint_future-annotations

chassing: I'm afraid to blindly activate that, because Pydantic `BaseModel`s would be affected. Those models require the referenced types during runtime mode, but `ruff` considers them only for type hinting. Activating that would review 2225 changes.

chassing: Nevertheless, I've added [runtime-evaluated-base-classes](https://docs.astral.sh/ruff/settings/#lint_flake8-type-checking_runtime-evaluated-base-classes) to handle it automatically.

PR #5126, opened 2025-07-25
feat(ocm-clusters): set default-ingress-private for private HCP cluster

hemslo:
After changes in rosa cli https://issues.redhat.com/browse/OCM-15992, create private HCP cluster with all private subnets without `--default-ingress-private` will throw error

```text
ERR: The number of public subnets for a public hosted cluster should be at least one
```

we need both `--private` and `--default-ingress-private` to use full private subnets.

[APPSRE-12219](https://issues.redhat.com/browse/APPSRE-12219)


mehfuz (approved):
lgtm

PR #5140, opened 2025-07-29
APPSRE-12043: support role and clusterrole for openshift-rolebindings

mehfuz:
the MR consists of following changes:
1. refactored code for `openshift-rolebindings` to support `Role` and `ClusterRole`
2. use of typed queries for role
3. flag (`--support-role-ref`)setting to support  rolebindings to use `roleref` as `Role` incase of Role and `ClusterRole` in case of clusterRole.(once all app-interface roles are migrated we just need to remove the flag to support `Role` and `ClusterRole`)


reconcile/openshift_rolebindings.py
```diff
@@ -19,40 +31,194 @@
...
+            return None
+        auth_dict = [auth.dict(by_alias=True) for auth in access.namespace.cluster.auth]
+        username_list = RoleBindingSpec.get_usernames_from_role(
+            users,
+            ob.determine_user_keys_for_access(
+                access.namespace.cluster.name,
+                auth_dict,
+                enforced_user_keys,
+            ),
+        )
```
hemslo, line 108: the method name is confusing, it's called `get_usernames_from_role` but params are users and user_keys.

reconcile/openshift_rolebindings.py
```diff
@@ -19,40 +31,194 @@
...
+
+    @staticmethod
+    def get_usernames_from_role(
+        users: list[UserV1] | None = None, user_keys: list[str] | None = None
+    ) -> set[str]:
+        return {
+            getattr(user, user_key)
+            for user in users or []
+            for user_key in user_keys or []
+            if hasattr(user, user_key)
```
hemslo, line 149: Is there a case that user object has that attribute, but the value is None? Can add none check

```python
{
            name
            for user in users or []
            for user_key in user_keys or []
            if (name := getattr(user, user_key, None))
}
```

reconcile/openshift_rolebindings.py
```diff
@@ -19,40 +31,194 @@
...
+        ]
+
+
+class RoleBindingSpec(BaseModel):
+    role_name: str
+    role_kind: str
+    namespace: NamespaceV1
+    cluster: ClusterV1
+    privileged: bool
+    username_list: set[str]
```
hemslo, line 73: type is set, but name is `username_list`, not match, can use `usernames`

reconcile/openshift_rolebindings.py
```diff
@@ -19,40 +31,194 @@
...
+            username_list=username_list,
+            openshift_service_accounts=service_accounts,
+        )
+
+    @staticmethod
+    def create_rb_specs_from_role(
+        role: RoleV1,
+        oc_map: ob.ClusterMap | None = None,
+        enforced_user_keys: list[str] | None = None,
+    ) -> list["RoleBindingSpec"]:
```
hemslo, line 125: can use classmethod to return `list[Self]`

reconcile/openshift_rolebindings.py
```diff
@@ -96,97 +262,37 @@ def fetch_desired_state(
...
-
-            for sa in service_accounts:
-                if ri is None:
-                    continue
-                sa_namespace_name, sa_name = sa.split("/")
-                oc_resource, resource_name = construct_sa_oc_resource(
-                    permission["role"], sa_namespace_name, sa_name
-                )
+            for oc_resource in rolebinding.get_oc_resources():
                 with contextlib.suppress(ResourceKeyExistsError):
```
hemslo, line 276: in which case resource key can exist?

reconcile/openshift_rolebindings.py
```diff
@@ -19,40 +31,194 @@
...
-      }
-      role
-    }
-    expirationDate
-  }
-}
-"""
 
+    def construct_user_oc_resource(
+        self, user: str, force_cluster_role_ref: bool
```
hemslo, line 153: do we need use `force_cluster_role_ref` in so many places, how about just use it once during `RoleBindingSpec` instance creation?

```python
role_kind="ClusterRole" if force_cluster_role_ref or access.cluster_role else "Role",
```

reconcile/openshift_rolebindings.py
mehfuz: key can exis if a user may have a Role assigned to them in case of  multiple app-interface roles present

reconcile/openshift_rolebindings.py
mehfuz: added as part of [f110463](https://github.com/app-sre/qontract-reconcile/pull/5140/commits/f11046342e7fc508ff552c92c21d1b4c74600234)

reconcile/openshift_rolebindings.py
hemslo: to support backward compatibility deployment, let's change `FORCE_CLUSTER_ROLE_REF` to an input param for `run` func, maybe reverse it to `--support-role-ref`, default `False`.

so the role_kind can be

```python
        role_kind = "Role" if access.role and support_role_ref else "ClusterRole"
```

* promote this PR
* migrate data
* specify `--support-role-ref` in deployment
* new PR to remove this param

reconcile/openshift_rolebindings.py
hemslo: this is not very clean, how about use `ri.get_desired` to check exists to conditional add?

reconcile/openshift_rolebindings.py
```diff
@@ -19,40 +31,194 @@
...
+        enforced_user_keys: list[str] | None = None,
+        bots: list[BotV1] | None = None,
+    ) -> Optional["RoleBindingSpec"]:
+        if not access.namespace:
+            return None
+        if not (access.role or access.cluster_role):
+            return None
+        privileged = access.namespace.cluster_admin or False
+        if oc_map and not oc_map.get(access.namespace.cluster.name):
+            return None
```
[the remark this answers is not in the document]
hemslo: when there is an issue init client, will the code crash there? Can we move it out so desired state only come from gql, after fetch current state, exclude missing cluster rolebindings in or before pass to diff?

reconcile/openshift_rolebindings.py
hemslo: somehow missed this change in recent version

reconcile/openshift_rolebindings.py
mehfuz: made use of ri.get_desired

hemslo: why need to check `ri.is_cluster_present`? `ri.get_desired` already tried `self._cluster[cluster]`

mehfuz: Oh yes...fixed at [334ecb9](https://github.com/app-sre/qontract-reconcile/pull/5140/commits/334ecb901938409a648a962a7034418e43884d03)

hemslo: the condition should be the other way around, only add it when it's not in desired state

mehfuz: changed here [a003e34](https://github.com/app-sre/qontract-reconcile/pull/5140/commits/a003e34e120ad02d96ea934acf7454df2de5e8a0)

hemslo (approved):
LGTM!

reconcile/test/test_openshift_rolebindings.py
```diff
@@ -0,0 +1,348 @@
...
+                    "name": "test-role5-test-org-user",
+                },
+            },
+        ),
+    )
+    test_role = get_app_interface_test_roles()
+    role_binding_specs = RoleBindingSpec.create_rb_specs_from_role(test_role[0], None)
+    assert not should_add_desired_state(
+        ri, role_binding_specs[0].get_oc_resources()[0], role_binding_specs[0]
+    )
```
hemslo, line 348: this means we still rely on `ri` to be pre populated to get desired state, what's the business case that we need to skip certain clusters? these can be added to test cases of `fetch_desired_state` instead of testing util function `should_add_desired_state`

reconcile/openshift_rolebindings.py
```diff
@@ -250,19 +250,31 @@ def fetch_desired_state(
     ri: ResourceInventory | None,
     support_role_ref: bool = False,
     enforced_user_keys: list[str] | None = None,
+    allowed_clusters: list[str] | None = None,
```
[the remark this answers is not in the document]
mehfuz: changed here [e5ab569](https://github.com/app-sre/qontract-reconcile/pull/5140/commits/e5ab569c693dcfb0ac6e8eba27053b6d92dc8bd5)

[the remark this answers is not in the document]
hemslo: do we need that `FetchDesiredStateError`? empty clusters can be a valid case, right? it we need that error, it should be thrown by fetch current state instead of fetch desired state

hemslo (approved):
LGTM!

PR #5143, opened 2025-07-30
🏷️ Add types to vpc-peering-validator, utils strics, utils throughtpu…

jfreixa:
…t, sre checkpoints.


chassing (approved):

reconcile/utils/structs.py
```diff
@@ -5,12 +5,12 @@
 class CommandExecutionResult:
     """This class represents a command execution result"""
 
-    def __init__(self, is_ok, message):
+    def __init__(self, is_ok: bool, message: str):
```
rporres, line 8: I don't know why it doesn't complain that type hints here are not complete 🤔 

```suggestion
    def __init__(self, is_ok: bool, message: str) -> None:
```

rporres (approved):
looks good, just a doubt

PR #5147, opened 2025-07-31
🏷️ Add types to terrascript aws client.

jfreixa:


reconcile/utils/terrascript_aws_client.py
```diff
@@ -1446,20 +1465,23 @@ def populate_tgw_attachments(self, desired_state):
...
+            )
             self.add_resource(infra_account_name, tf_resource_association)
 
             # now that the tgw is shared to the cluster's aws account
             # we can create a vpc attachment to the tgw
             subnets_id_az = accepter.subnets_id_az
+            assert subnets_id_az is not None  # make mypy happy
             subnets = self.get_az_unique_subnet_ids(subnets_id_az)
-            values = {
+            values: dict[str, Any] = {
```
hemslo, line 1488: let's rename all `values` in this function to have meaningful names

reconcile/utils/terrascript_aws_client.py
```diff
@@ -2148,39 +2173,38 @@ def _region_from_availability_zone(az):
...
 
     @staticmethod
-    def validate_db_name(name):
+    def validate_db_name(name: str) -> bool:
         """Handle for Error creating DB Instance:
         InvalidParameterValue: DBName must begin with a letter
         and contain only alphanumeric characters."""
         pattern = r"^[a-zA-Z][a-zA-Z0-9_]+$"
-        return re.search(pattern, name) and len(name) < 64
+        return bool(re.search(pattern, name)) and len(name) < 64
```
hemslo, line 2191: explicit None check as return maybe better, and can check len first since it's cheaper

```python
return len(name) < 64 and re.search(pattern, name) is not None
```

same in other `re.search` in this file

reconcile/utils/terrascript_aws_client.py
```diff
@@ -3242,11 +3278,11 @@ def populate_tf_resource_ecr(self, spec):
...
         # iam user for repository
-        values = {}
-        values["name"] = identifier
-        values["tags"] = common_values["tags"]
-        values["depends_on"] = self.get_dependencies([ecr_tf_resource])
-        user_tf_resource = aws_iam_user(identifier, **values)
+        values_iam_user: dict[str, Any] = {}
+        values_iam_user["name"] = identifier
+        values_iam_user["tags"] = common_values["tags"]
+        values_iam_user["depends_on"] = self.get_dependencies([ecr_tf_resource])
```
hemslo, line 3288: can just create it once instead of keep mutating

```python
values_iam_user = {
  "name": identifier,
  "tags": common_values["tags"],
  "depends_on": self.get_dependencies([ecr_tf_resource]),
}
```

can apply to multiple places in this file.

reconcile/utils/terrascript_aws_client.py
```diff
@@ -5213,10 +5278,19 @@ def populate_tf_resource_s3_cloudfront_public_key(self, spec):
...
-        self, identifier, openshift_service, account_name, namespace_info, ocm_map
-    ):
+        self,
+        identifier: str,
+        openshift_service: str,
+        account_name: str,
+        namespace_info: Mapping[str, Any],
+        ocm_map: OCMMap | None,
+    ) -> set[str]:
         account = self.accounts[account_name]
```
hemslo, line 5288: why not mark ocm_map as not None in param?

jfreixa: can be none, I need to check somewhere if this is none. see [this](https://github.com/jfreixa/qontract-reconcile/blob/e8a67af2849e94f2d2354cb855c3f930969dec19/reconcile/terraform_resources.py#L267C14-L267C15)

reconcile/utils/terrascript_aws_client.py
```diff
@@ -997,10 +1010,10 @@ def populate_iam_users(
 
     def populate_users(
         self,
-        roles,
+        roles: list["Role"],
```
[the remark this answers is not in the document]
chassing: ^ This was meant in the whole file

reconcile/utils/terrascript_aws_client.py
hemslo: https://github.com/app-sre/qontract-reconcile/blob/b8eb92e4fec51d7e71badf7ca38a0a3dfa523daf/reconcile/utils/terrascript_aws_client.py#L5423-L5426 this is the only usage, move None check there to make this function type clean.

chassing (approved):

hemslo (approved):
LGTM!

reconcile/aws_iam_keys.py
```diff
@@ -48,7 +48,7 @@ def init_tf_working_dirs(
     accounts: Iterable[dict[str, Any]],
     thread_pool_size: int,
     settings: Mapping[str, Any],
-) -> dict[str, str]:
+) -> Mapping[str, str]:
```
lechuk47, line 51: What is the reason for this change? 
IINW, the general recommendation was using abstract types for parameters and concrete types for return types. 

Just commenting here but there are other ocurrences with the same change (dict -> Mapping/MutableMapping)

hemslo: +1, for return type, use concrete or generic types

jfreixa: I changed the return types of mapping in the last commits but in some functions like dump I was receiving a parameter modifying it and returning it. If I wanted to receive a mapping I was required to return a mapping.

PR #5165, opened 2025-08-08
Refactor vault and add types

jfreixa:


reconcile/utils/vault.py
```diff
@@ -260,8 +289,14 @@ def _read_all_v1(self, path):
 
         return secret["data"]
 
+    class Secret(TypedDict):
```
chassing, line 292: Please move it outside of the class.

reconcile/utils/vault.py
```diff
@@ -66,69 +89,73 @@ def __init__(
...
-                config["vault"]["kube_auth_mount"]
-                if kube_auth_mount is None
-                else kube_auth_mount
-            )
-            self.kube_sa_token_path = os.environ.get(
-                "KUBE_SA_TOKEN_PATH",
-                "/var/run/secrets/kubernetes.io/serviceaccount/token",
-            )
-            self.kube_auth_enabled = True
+        if not hasattr(self, "initialized"):
```
chassing, line 92: Please change the logic to

```python
if hasattr(self, "initialized"):
    return
...
```

This would be easier to read

cubismod (approved):
Nice work!

reconcile/utils/vault.py
```diff
@@ -46,17 +48,45 @@ class VaultConnectionError(Exception):
...
+                is_authenticated = cls._instance._client.is_authenticated()
+            except requests.exceptions.ConnectionError:
+                is_authenticated = False
+
+            if not is_authenticated:
+                cls._instance.close()
+                cls._instance = super().__new__(cls)
+                return cls._instance
+
+            return cls._instance
```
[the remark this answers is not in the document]
jfreixa: Where we need to use VaultClient without singleton?
But this change will need to change every reference in our code

hemslo (approved):
LGTM!

PR #5168, opened 2025-08-08
Update the ALB resource to support query-string conditions

cubismod:
[APPSRE-12347](https://issues.redhat.com/browse/APPSRE-12347)

Updates the ALB provider to support query-string conditionals. Schema MR: https://github.com/app-sre/qontract-schemas/pull/912


reconcile/utils/terrascript_aws_client.py
```diff
@@ -5346,6 +5347,11 @@ def _get_alb_rule_condition_value(
         )
         if condition_type_key is None:
             raise KeyError(f"unknown alb rule condition type {condition_type}")
+
+        # Query string conditions use a different structure than other condition types
+        if condition_type == "query-string":
+            return {condition_type_key: condition[condition_type_key]}
```
hemslo, line 5353: please add a unittest for this special logic

PR #5169, opened 2025-08-11
:label: utils.openshift_resource: type hints

chassing:
More type annotations.

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -1005,8 +1005,9 @@ class CheckClusterScopedResourceNames:
                 for kind in cluster_scoped_types:
                     declared_items = mrn.get(kind, [])
                     desired_items = set(
-                        self.ri.get_desired_by_type(
-                            cluster_name, ns["name"], kind
+                        (
+                            self.ri.get_desired_by_type(cluster_name, ns["name"], kind)
+                            or {}
                         ).keys()
                     )
                     diff = desired_items.difference(declared_items)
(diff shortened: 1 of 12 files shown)


reconcile/utils/openshift_resource.py
```diff
@@ -633,41 +646,54 @@ def add_desired(
...
+        resource_type: str,
+        name: str,
+        value: OpenshiftResource,
+    ) -> None:
         with self._lock:
             current = self._clusters[cluster][namespace][resource_type]["current"]
             current[name] = value
 
-    def __iter__(self):
+    def __iter__(self) -> Generator[tuple[str, str, str, dict[str, Any]], None, None]:
```
hemslo, line 685: here can use `Iterator`

reconcile/utils/three_way_diff_strategy.py
```diff
@@ -61,7 +61,7 @@ def normalize_object(item: OR) -> OR:
         validate_k8s_object=False,
     )
 
-    annotations = n.body.get("metadata").get("annotations") or {}
+    annotations = n.body.get("metadata", {}).get("annotations", {})
```
hemslo, line 64: any edge case that `annotations` key exist but value is `None`?

reconcile/openshift_resources_base.py
```diff
@@ -1212,7 +1213,7 @@ def _early_exit_fetch_resource(spec: Sequence, settings: Mapping) -> dict[str, s
         c = resource["resource"].get("content")
     del resource["resource"]
     resource[IDENTIFIER_FIELD_NAME] = id
-    content_sha = hashlib.md5(c.encode("utf-8")).hexdigest()
+    content_sha = hashlib.md5(str(c).encode("utf-8")).hexdigest()
```
hemslo, line 1216: md5 is used here as well? I need to note this down and sync with validator changes if we want to get rid of deprecate warnings of md5. https://docs.astral.sh/ruff/rules/hashlib-insecure-hash-function/

reconcile/utils/openshift_resource.py
chassing: Good spot. Fixed in [de852f7](https://github.com/app-sre/qontract-reconcile/pull/5169/commits/de852f7e3a716235e1c61a3889ff1905e39f7014)

reconcile/utils/three_way_diff_strategy.py
chassing: TBH, I don't know. I switched it back in 830733e

PR #5171, opened 2025-08-12
Fix jenkins API to trigger job. Endpoint /build no longer works and n…

jfreixa:
…eed to be changed to /buildWithParameters APPSRE-12358


hemslo (approved):
LGTM!

PR #5176, opened 2025-08-19
APPSRE-12383 konflux sqs message

fishi0x01:
Using konflux for SQS auto MRs.

Note, that we currently cannot run after the RPA. I.e., the SQS message will be sent before the RPA finished publishing the prod image. We trigger the SQS message as late as possible in the build pipeline, after all checks have passed. I.e., likely auto-bump MR will fail initially, but should work shortly after with a `/retest`. Note, that we have no guarantee that the RPA passed - so we cannot be sure at that point that we will actually receive a valid image. However, at least we get the MR created and pinged on it - and most of the time that should work out just fine.

Note, that we do not have author-email any longer. We now have the author id.

We could also add a ~30s timeout or some wait-for-image mechanics before opening the MR. However, I'd leave that as a follow-up optimization if needed.

Ideally, we can at some point have a finally task after RPA - that feature exists, however, we havent used it yet and Im not sure if we can transport pipelinerun args to it, such as the github-user-id. Will create a follow-up ticket to investigate post RPA pipeline for this, such as https://issues.redhat.com/browse/APPSRE-11256

Corresponding pipeline change https://github.com/app-sre/shared-pipelines/pull/289

### DIFF
--- a/.tekton/qontract-reconcile-master-push.yaml
+++ b/.tekton/qontract-reconcile-master-push.yaml
@@ -39,6 +39,12 @@ spec:
     value: '100000'
   - name: goss-container-structure-test-file
     value: dockerfiles/goss.yaml
+  - name: github-user-id
+    # https://pipelinesascode.com/docs/guide/authoringprs/
+    value: |-
+      {{sender}}
+  - name: sqs-pr-type
+    value: promote_qontract_reconcile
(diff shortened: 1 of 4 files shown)


chassing (approved):
❤️

PR #5184, opened 2025-08-21
Add robot account management

esron:
JIRA: [APPSRE-11883](https://issues.redhat.com/browse/APPSRE-11883)

This adds a new integration that handles robot account management.

I decided to go with this approach because `quay-permissions` integration already has a complex state management that when mixed with the robot accounts needs of managing teams and repository relationships gets really messy.

This approach also lets robot-accounts be more independent and clear to the user, since it relies on a single file where all relationships are defined.

Also I am getting inspiration from the [original](https://gitlab.cee.redhat.com/exd/quay/quay-access-management) script where the robot accounts for quay are defined.

We still need to decide how we are going to migrate existing robot accounts to `app-interface`.

Also we don't have full read access for robot-account in some of the Quay orgs defined in `app-interface`.

## Changes since last review

- **Rebased onto upstream master**
- **Migrated GQL query to `pydantic_v2`**
- **Fixed `list_robot_accounts` correctness bug** — the endpoint was always returning empty `teams` and `repositories`. Now passes `?permissions=true` for teams, and calls `get_robot_account_permissions` per robot for accurate repo permission levels.
- **Added `delete: true` safety gate** — the integration no longer deletes robot accounts that are absent from app-interface. Deletion now requires an explicit `delete: true` field on the robot account definition. This prevents accidentally removing accounts managed outside app-interface.
- **Fixed recursive loop in `QuayApiStore`** — `get_quay_api_store()` was instantiating `QuayApiStore` which called itself back infinitely.
- **Simplified org_key lookups** — replaced `next()` iteration with direct `OrgKey(...)` dict lookups in `build_current_state` and `apply_action`.
- **Verified with dry-run** — integration runs end-to-end against a local qontract-server without errors.

## How to test this?

Create an integration file in `app-interface`. Here is a snippet, you can copy other integrations file if you get validations errors and edit:

```yaml
$schema: /app-sre/integration-1.yml

labels: {}

name: quay-robot-accounts

description: Manage Quay Robot Accounts

upstream: https://github.com/app-sre/qontract-reconcile

pr_check:
  disabled: true
  cmd: quay-robot-accounts

# Copy this from other integrations
managed:
- ...

schemas:
- /access/quay-robot-1.yml
- /app-interface/app-interface-settings-1.yml
- /app-sre/integration-1.yml
- /aws/account-1.yml
- /dependencies/dependency-1.yml
- /dependencies/quay-instance-1.yml
- /dependencies/quay-org-1.yml
```

Create a new `robot-account-1.yml`:
```yaml
---
$schema: /access/quay-robot-1.yml

labels: {}

name: app-sre-redhat-services-prod-robot-account
quay_username: appsre_pull # redhat-services-prod+appsre_pull

quay_org:
  $ref: /dependencies/quay/redhat-services-prod.yml

teams:
- devs
```

Set your local qontract-schemas to the branch on this [PR](https://github.com/app-sre/qontract-schemas/pull/901).

When you run:
```bash
qontract-reconcile --log-level INFO --config config.toml --dry-run quay-robot-accounts
```

You should see a list of actions. The first one should be to create the robot account defined in the yaml file above.

Drop a comment if you have any questions, issues, or need help testing this.

## Summary by CodeRabbit

* **New Features**
  * Added robot account reconciliation for descriptions, team membership, repository permissions, optional deletion, and dry-run previews.
  * Added a CLI subcommand for robot account synchronization.
  * Added GraphQL support for robot account configuration.

* **Bug Fixes**
  * Improved Quay API store handling across reconciliation workflows.

* **Tests**
  * Added comprehensive robot account and permission API coverage.
  * Updated Quay reconciliation tests and test setup.


reconcile/quay_robot_accounts.py
```diff
@@ -0,0 +1,397 @@
...
+
+def calculate_diff(
+    desired_state: dict[tuple[str, str, str], RobotAccountState],
+    current_state: dict[tuple[str, str, str], RobotAccountState],
+) -> list[RobotAccountAction]:
+    """Calculate the differences between desired and current state"""
+    actions = []
+
+    # Find robots to create
+    for key, desired in desired_state.items():
```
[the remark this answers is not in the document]
esron: Done it.

reconcile/quay_robot_accounts.py
```diff
@@ -0,0 +1,397 @@
...
+    for key, current in current_state.items():
+        if key not in desired_state:
+            actions.append(
+                RobotAccountAction(
+                    action="delete",
+                    robot_name=current.name,
+                    org_name=current.org_name,
+                    instance_name=current.instance_name,
+                )
+            )
```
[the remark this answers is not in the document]
esron: Done it.

reconcile/quay_base.py
```diff
@@ -19,8 +19,8 @@ class OrgInfo(TypedDict):
 
 
 class QuayApiStore(UserDict[OrgKey, OrgInfo]):
-    def __init__(self) -> None:
-        super().__init__(get_quay_api_store())
+    def __init__(self, data: dict[OrgKey, OrgInfo] | None = None) -> None:
+        super().__init__(data or {})
```
hemslo, line 23: **Bug (critical):** This changes `QuayApiStore()` from auto-populating to defaulting to an empty dict, but only `quay_mirror_org.py` was updated. These callers still use `QuayApiStore()` and will silently get an empty store:

- `reconcile/quay_membership.py:222`
- `reconcile/quay_permissions.py:65`
- `reconcile/quay_repos.py:225`
- `reconcile/test/test_quay_repos.py:73`
- `reconcile/test/test_quay_membership.py:91`

Either update all callers to use `get_quay_api_store()`, or make the default backwards-compatible (e.g., call `get_quay_api_store()` when `data is None`).

reconcile/quay_robot_accounts.py
```diff
@@ -0,0 +1,390 @@
...
+
+        case RobotAccountActionType.REMOVE_TEAM:
+            logging.info(
+                f"Removing robot {action.robot_name} from team {action.team} in {action.org_name}"
+            )
+            if not action.team:
+                raise ValueError(f"Team is required for remove_team action: {action}")
+            quay_api.remove_user_from_team(
+                f"{action.org_name}+{action.robot_name}", action.team
```
hemslo: **Bug (critical):** `remove_user_from_team()` ([quay_api.py:79-102](https://github.com/app-sre/qontract-reconcile/blob/feff306161736a2f0428103b629af5314522e2f4/reconcile/utils/quay_api.py#L79-L102)) removes the user from the team **and then also deletes the user from the entire organization** (line 101: `self._delete(url_org)`). For a robot that belongs to multiple teams, removing it from one team would destroy the robot account.

Consider adding a dedicated method that only calls the team-membership endpoint (`DELETE /organization/{org}/team/{team}/members/{user}`) without the org-level deletion.

reconcile/quay_robot_accounts.py
```diff
@@ -0,0 +1,390 @@
...
+    """Apply a single action to Quay"""
+    org_key = OrgKey(action.instance_name, action.org_name)
+    if org_key not in quay_api_store:
+        logging.error(f"No API found for {action.instance_name}/{action.org_name}")
+        return
+
+    quay_api = quay_api_store[org_key]["api"]
+
+    if dry_run:
+        logging.info(f"[DRY RUN] Would perform: {action}")
```
hemslo: **Nit — project convention:** Per AGENTS.md: *"Same log output in both modes. Check `dry_run` only at execution point."* This early return produces a different (less detailed) log in dry-run mode. Move the `dry_run` guard into each `match` arm so the per-action log lines fire regardless of mode, and only the API call is gated.

reconcile/utils/quay_api.py
```diff
@@ -240,3 +256,73 @@ def set_repo_team_permissions(self, repo_name: str, team: str, role: str) -> Non
...
+    def delete_robot_account(self, name: str) -> None:
+        url = f"/api/v1/organization/{self.organization}/robots/{name}"
+        self._delete(url)
+
+    def get_robot_account_permissions(self, name: str) -> list[RobotAccountPermission]:
+        url = f"/api/v1/organization/{self.organization}/robots/{name}/permissions"
+        body = self._get(url)
+        return body["permissions"]
+
+    def set_robot_account_permissions(
```
hemslo: **Cleanup:** `set_robot_account_permissions` (and `delete_robot_account_permissions` below) are defined but never called — not in the integration, not in tests. Remove them to avoid dead code, or add usage + tests if they're planned for a follow-up.

reconcile/utils/quay_api.py
```diff
@@ -240,3 +256,73 @@ def set_repo_team_permissions(self, repo_name: str, team: str, role: str) -> Non
...
+    def create_robot_account(self, name: str, description: str) -> None:
+        url = f"/api/v1/organization/{self.organization}/robots/{name}"
+        body = {"description": description}
+        self._put(url, data=body)
+
+    def delete_robot_account(self, name: str) -> None:
+        url = f"/api/v1/organization/{self.organization}/robots/{name}"
+        self._delete(url)
+
+    def get_robot_account_permissions(self, name: str) -> list[RobotAccountPermission]:
```
hemslo, line 303: **Test gap:** `get_robot_account_permissions` is used in `build_current_state` but has no test in `test_quay_api.py`. All other new `QuayApi` methods have HTTP-level tests — this one should too.

reconcile/gql_definitions/quay_robot_accounts/quay_robot_accounts.gql
```diff
@@ -0,0 +1,26 @@
...
+      }
+    }
+    teams
+    repositories {
+      name
+      permission
+    }
+    delete
+  }
+}
```
hemslo: **Nit:** Missing trailing newline.

hemslo (changes requested):
Thanks for the new integration — the architecture is clean and the `delete: true` safety gate is a smart design choice. Found two critical correctness issues that need fixing before merge, plus a few smaller items. Details inline.

hemslo (commented):
Previous critical bugs are both fixed — nice work. Two minor items remain (one carried over, one new). Details inline.

reconcile/utils/quay_api.py
```diff
@@ -103,6 +119,26 @@ def remove_user_from_team(self, user: str, team: str) -> bool:
 
         return True
 
+    def remove_robot_from_team(self, robot_name: str, team: str) -> None:
```
hemslo, line 122: **Test gap:** `remove_robot_from_team` handles an error case (silencing "does not belong to team" for idempotency, re-raising everything else) but has no HTTP-level test in `test_quay_api.py`. Every other new `QuayApi` method in this PR has success + error tests — this one should too.

Suggested tests: success (204/200), idempotent case (user not in team message → no raise), and error propagation (other HTTP errors → raise).

hemslo (approved):
All feedback addressed. The managedTeams/managedRepos scoping, validation layer, and error handling in run() are solid additions. LGTM.

reconcile/gql_definitions/introspection.json
hemslo, line 1: you may want to regenerate this file with the latest qontract server and schema

esron: Done it.

PR #5186, opened 2025-08-25
Sloth alerts - retain `sloth*` labels in recording rules

dwelch0:


reconcile/utils/sloth.py
```diff
@@ -22,11 +22,15 @@ class PrometheusRuleSpec(TypedDict):
...
-    if rule.get("record") or not rule.get("labels"):
-        return None
-    # sloth adds several sloth_* labels to rules that are not compliant with prometheus-rule-1 schema
-    # see https://sloth.dev/examples/default/getting-started/#__tabbed_1_2
-    labels = {k: v for k, v in rule["labels"].items() if not k.startswith("sloth")}
+    labels = {}
+    if rule.get("record"):
+        # sloth record rules include several sloth_* labels that are later utilized in alert rule expressions
+        # and must be retained
+        labels = {k: v for k, v in rule["labels"].items() if k.startswith("sloth")}
```
hemslo, line 29: do we need to only keep `sloth*` labels, in test fixture

```yaml
  - record: test_metric
    expr: vector(1)
    labels:
      sloth_id: test-id
      sloth_service: test-service
      service: app-interface
```

should `service: app-interface` be kept as well?

reconcile/utils/sloth.py
```diff
@@ -22,11 +22,15 @@ class PrometheusRuleSpec(TypedDict):
...
+    elif rule.get("alert"):
+        # sloth adds several sloth_* labels to alerting rules that are not compliant with prometheus-rule-1 schema
+        # see https://sloth.dev/examples/default/getting-started/#__tabbed_1_2
+        labels = {k: v for k, v in rule["labels"].items() if not k.startswith("sloth")}
     return labels or None
```
hemslo, line 34: `labels or None` is a little confusing, basically we return `None` if it's empty dict, can we always return `dict`, then on usage side, drop `labels` if it's empty instead of none check? Same as `get_cleaned_rule_annotations`

hemslo (approved):
LGTM!

PR #5189, opened 2025-08-27
:sparkles: external-resources: tagging support

chassing:
Add tagging support to all ERv2-managed external resources:

The compliance mandatory ones:
* `app-code`
* `cost-center`
* `service-phase`

And allow custom tags directly on the external resources via the new `tag` attribute.

Ticket: [APPSRE-12423](https://issues.redhat.com/browse/APPSRE-12423)

### DIFF
--- a/reconcile/utils/external_resource_spec.py
+++ b/reconcile/utils/external_resource_spec.py
@@ -144,13 +144,33 @@ class ExternalResourceSpec:
         return {}
 
     def tags(self, integration: str) -> dict[str, str]:
-        return {
+        tags = {
             "managed_by_integration": integration,
             "cluster": self.cluster_name,
             "namespace": self.namespace_name,
             "environment": self.namespace["environment"]["name"],
             "app": self.namespace["app"]["name"],
+            # The attributes below are required by the schema but not all integrations query them yet!
+            "app-code": self.namespace["app"].get("appCode", "unknown"),
+            "cost-center": self.namespace["app"].get("costCenter", "999"),
+            "service-phase": self.namespace["environment"].get("servicePhase", "prod"),
         }
+        resource_tags_str = self.resource.get("tags")
+        if resource_tags_str:
+            resource_tags = json.loads(resource_tags_str)
+            # normalize camelCase keys to kebab-case
+            key_mapping = {
+                "appCode": "app-code",
+                "costCenter": "cost-center",
+                "servicePhase": "service-phase",
+            }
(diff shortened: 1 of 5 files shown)


reconcile/utils/external_resource_spec.py
```diff
@@ -144,13 +144,33 @@ def annotations(self) -> dict[str, str]:
...
+        tags = {
             "managed_by_integration": integration,
             "cluster": self.cluster_name,
             "namespace": self.namespace_name,
             "environment": self.namespace["environment"]["name"],
             "app": self.namespace["app"]["name"],
+            # The attributes below are required by the schema but not all integrations query them yet!
+            "app-code": self.namespace["app"].get("appCode", "unknown"),
+            "cost-center": self.namespace["app"].get("costCenter", "999"),
+            "service-phase": self.namespace["environment"].get("servicePhase", "prod"),
```
hemslo, line 156: why these default values? there is one hard coded default value https://github.com/app-sre/qontract-reconcile/blob/c048032ab42b27eba3cd6b3009dfc7ec3b0bd60f/reconcile/utils/terrascript_aws_client.py#L289-L293, can we replace it with default tags in settings (need new schema), then pass down to terraform provider section to avoid missing tags?

reconcile/utils/external_resource_spec.py
```diff
@@ -144,13 +144,33 @@ def annotations(self) -> dict[str, str]:
...
+            # normalize camelCase keys to kebab-case
+            key_mapping = {
+                "appCode": "app-code",
+                "costCenter": "cost-center",
+                "servicePhase": "service-phase",
+            }
+            normalized_tags = {}
+            for key, value in resource_tags.items():
+                normalized_key = key_mapping.get(key, key)
+                normalized_tags[normalized_key] = value
```
hemslo, line 170: this logic maybe needed in default tags handling as well

reconcile/utils/external_resource_spec.py
```diff
@@ -144,13 +144,33 @@ def annotations(self) -> dict[str, str]:
         return {}
 
     def tags(self, integration: str) -> dict[str, str]:
-        return {
+        tags = {
```
hemslo, line 147: please update test cases in https://github.com/app-sre/qontract-reconcile/blob/c048032ab42b27eba3cd6b3009dfc7ec3b0bd60f/reconcile/test/utils/test_external_resource_spec.py#L201

PR #5195, opened 2025-08-29
Update jenkins-job-builder package

jfreixa:


mehfuz (approved):
question: Do we need additional testcase to verify this?
 i can see the existing testcases are passing so as per that the changes are good

pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
     "filetype~=1.2.0",
     "gql==3.1.0",
     "hvac>=0.7.0,<0.8.0",
-    "jenkins-job-builder~=4.3.0",
+    "jenkins-job-builder~=6.4.2",
```
hemslo, line 29: now that we have renovate to bump version, let's pin exact version in `pyproject.toml`

jfreixa:
> question: Do we need additional testcase to verify this? i can see the existing testcases are passing so as per that the changes are good

Added a test to verify `get_jobs` method in the latest changes

PR #5197, opened 2025-08-29
update 'INTEGRATION_EXTRA_ARGS' qr-fips quay-path in qontract-manager-fedramp.yaml

bkez322:
updates 'INTEGRATION_EXTRA_ARGS' qr-fips quay-path in qontract-manager-fedramp.yaml

Needed for Q-Manager to tell QR pods where to pull their images from.


bkez322:
Reverting and redoing this as I stupidly didn't let the pr-check tests pass.

PR #5211, opened 2025-09-05
external-resources: handle namespace.delete flag

chassing:
Automatically delete all ERv2-managed resources when a namespace is marked for deletion. This enables the deletion of a namespace, including all managed resources, in one MR.

Bugfix: Additionally, external resources in namespaces with `managedExternalResources: false` must be ignored!

Ticket: [APPSRE-11552](https://issues.redhat.com/browse/APPSRE-11552)


hemslo (approved):
LGTM!

PR #5215, opened 2025-09-08
VaultClient - set default value of latest for secret_version

dwelch0:
[APPSRE-12389](https://issues.redhat.com/browse/APPSRE-12389)

Default to reading latest version of kvv2 secret. 

This resolves usages of `read_all` that currently assumes backing secret engine to be kvv1. Example: https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/external_resources/secrets_sync.py#L454


hemslo (approved):
LGTM!

PR #5216, opened 2025-09-09
chore: upgrade and pin ruff version

hemslo:
New ruff version can correctly detect `FURB118` so we don't need to disable it. Pin the exact version, so renovate won't bump ruff version in `lock file maintenance` and failed build. https://github.com/app-sre/qontract-reconcile/pull/5121


mehfuz (approved):
LGTM

PR #5224, opened 2025-09-16
Update sretoolbox to support python version 3.12

jfreixa:


pyproject.toml
```diff
@@ -54,7 +54,7 @@ dependencies = [
     "sendgrid>=6.4.8,<6.5.0",
     "sentry-sdk~=2.0",
     "slack_sdk>=3.10,<4.0",
-    "sretoolbox~=2.6",
+    "sretoolbox~=2.7",
```
hemslo, line 57: let's pin the exact version `==2.7.0`

jfreixa: done in https://github.com/app-sre/qontract-reconcile/pull/5224/commits/bb107bdb5530097311fa27521f2a6e0d158e4ef9

pyproject.toml
```diff
@@ -54,7 +54,7 @@ dependencies = [
     "sendgrid>=6.4.8,<6.5.0",
     "sentry-sdk~=2.0",
     "slack_sdk>=3.10,<4.0",
-    "sretoolbox~=2.6",
+    "sretoolbox==2.7",
```
rporres, line 57: I needed to go to [PEP 440](https://peps.python.org/pep-0440) to check that `X.Y` and `X.Y.0` is the same 😛 

Where did the "explicit is better than implicit" from PEP20 go? 😂

jfreixa: changed in https://github.com/app-sre/qontract-reconcile/pull/5224/commits/a08a3ee42dd536d4b0aa45dea6b9816dded581a8

PR #5235, opened 2025-10-07
Add secret format support to AWS Secrets Manager

rporres:
The idea is to be able to control the format of the secret copied to Secrets Manager from Vault in case that the secret shape in Vault is not what we need in Secrets Manager.

We needed to implement _unflatten_dict to convert flat dictionaries with dotted keys (e.g., "db.password") into nested structures for Jinja2 template processing. Add conflict detection to prevent ambiguous key paths like {"a.b": 1, "a.b.c": 2}.

This work is needed in the context of APPSRE-12414

Assisted-by: Claude <noreply@anthropic.com>


reconcile/utils/terrascript_aws_client.py
```diff
@@ -1,6 +1,7 @@
 # ruff: noqa: N801
 from __future__ import annotations
 
+####
```
chassing, line 4: ?

chassing (approved):
Except the nit. LGTM

reconcile/utils/terrascript_aws_client.py
rporres: a leftover from a comment I wrote 😅

reconcile/utils/terrascript_aws_client.py
```diff
@@ -5832,6 +5840,66 @@ def populate_tf_resource_secrets_manager(self, spec: ExternalResourceSpec) -> No
...
+
+        return result
+
+    @staticmethod
+    def _apply_secret_format(
+        secret_format: str, secret_data: dict[str, str]
+    ) -> dict[str, str]:
+        # Convert flat dict with dotted keys to nested dict for Jinja2
+        nested_secret_data = TerrascriptClient._unflatten_dotted_keys_dict(secret_data)
+        rendered_data = process_jinja2_template(secret_format, nested_secret_data)
```
hemslo, line 5886: @rporres since this is a new feature, how about we put `secret_data` in a new var to avoid structure change?

Like this

```python
rendered_data = process_jinja2_template(secret_format, {"secret": secret_data})
```

so the jinja template is

```
{"password":"{{ secret["db.password"] }}","username":"{{ secret["db.user"] }}"}
```

Btw there is a similar feature [output_format](https://github.com/app-sre/qontract-schemas/blob/main/schemas/openshift/terraform-output-format-1.yml) which enforce result to be `dict` in schema, should we do the same? It will be something like this:

```yaml
    secret_format:
      password: '{{ secret["db.password"] }}'
      username: '{{ secret["db.username"] }}'
```

PR #5236, opened 2025-10-07
Upgrade jsonpath-ng to 1.7.0

esron:
This solves SyntaxWarning on pr check.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
 requires-python = "==3.12.*"
 
 [[package]]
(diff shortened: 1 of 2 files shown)


uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
```
hemslo, line 2: please upgrade you local `uv` version to latest, then run `uv sync` again

reconcile/change_owners/diff.py
```diff
@@ -251,8 +251,6 @@ def build_jsonpath_part(element: str | int) -> jsonpath_ng.JSONPath:
             case int():
                 return jsonpath_ng.Index(element)
             case str():
-                if "." in element:
-                    return jsonpath_ng.Fields(f"'{element}'")
```
hemslo, line 255: is there a test case to cover that element with `.` now auto quoted in new version? https://github.com/h2non/jsonpath-ng/releases/tag/v1.6.0

PR #5315, opened 2025-12-01
APPSRE-13008: added typed queries for clusterrolebinding

mehfuz:
This MR refactors the openshift-clusterrolebindings integration to use type-safe GraphQL queries including testcases.


reconcile/openshift_clusterrolebindings.py
```diff
@@ -45,119 +55,162 @@
...
 QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
+NAMESPACE_CLUSTER_SCOPE = "cluster"
+
+
+class OCResource(BaseModel, arbitrary_types_allowed=True):
+    resource: OR
+    resource_name: str
+
+
+@dataclass
```
chassing: Could you please not mix dataclasses and Pydantic models? Just use a basemodel here too; you don't need to set `arbitrary_types_allowed=True` below.

reconcile/openshift_clusterrolebindings.py
```diff
@@ -45,119 +55,162 @@
...
+class ServiceAccountSpec:
+    sa_namespace_name: str
+    sa_name: str
+
+    @classmethod
+    def create_sa_spec(cls, bots: list[BotV1] | None) -> list[Self]:
```
chassing:
```suggestion
    def create_sa_spec(cls, bots: Iterable[BotV1] | None) -> list[Self]:
```

reconcile/openshift_clusterrolebindings.py
```diff
@@ -45,119 +55,162 @@
...
+    cluster_roles: list[RoleV1] = expiration.filter(get_app_interface_clusterroles())
+    clusters_to_check = allowed_clusters or set()
+    cluster_role_binding_specs = [
+        cluster_role_binding_spec
+        for cluster_role in cluster_roles
+        for cluster_role_binding_spec in ClusterRoleBindingSpec.create_cluster_role_binding_specs(
+            cluster_role
+        )
+        if cluster_role_binding_spec.cluster.name in clusters_to_check
+    ]
```
hemslo: when `allowed_clusters` is `None`, means no filter, `cluster_role_binding_specs` here will be empty, expected?

PR #5489, opened 2026-04-10
fix(vpc-peering): tolerate both HCP security groups on VPC endpoint

TGPSKI:
## Summary

- The HyperShift control plane operator is migrating private-router VPC endpoint security groups from the legacy `-default-sg` naming to `-vpce-private-router`. During the transition, both SGs are attached simultaneously.
- `_get_api_security_group_id` crashes with `ValueError` when it finds 2 matching SGs instead of the expected 1, blocking `terraform-vpc-peerings` reconciliation across multiple shards (`app-sre-stage`, `stonesoup-stage-private-1`, `app-sre`).
- Fix: when both SGs are present, prefer the legacy `-default-sg` because existing peering and TGW ingress rules (21 rules, 38+ CIDR ranges) are configured on that SG. Clusters that have fully migrated (only `-vpce-private-router` present) continue to work as the sole match. Duplicate suffix detection raises `ValueError` to fail fast on unexpected state.

## Context

CloudTrail confirms the HyperShift control plane operator added  `vpce-private-router` sg to VPC endpoint  alongside the existing `default-sg`.

Alerts:
- `QontractReconcileError [FIRING:4]` - terraform-vpc-peerings, terraform-tgw-attachments

## Test plan

- [x] `test_get_cluster_vpc_details_with_hcp_security_group` — single SG (parametrized: `-default-sg`, `-vpce-private-router`)
- [x] `test_get_cluster_vpc_details_with_both_hcp_security_groups` — both SGs present, asserts `-default-sg` is selected
- [x] `test_get_cluster_vpc_details_duplicate_hcp_sg_raises` — duplicate suffix detection (parametrized: both suffixes)

## Release Notes

* **Bug Fixes**
  * Improved security group selection for VPC endpoints to properly handle multiple configuration types
  * Added validation to detect and report duplicate or conflicting security group configurations

* **Tests**
  * Added test coverage for VPC endpoint security group selection scenarios

### DIFF
--- a/reconcile/test/utils/test_aws_api.py
+++ b/reconcile/test/utils/test_aws_api.py
@@ -536,3 +536,112 @@ def test_get_cluster_vpc_details_with_hcp_security_group(
     assert route_table_ids is None
     assert subnets_id_az is None
     assert api_security_group_id == expected_sg_id
+
+
+def test_get_cluster_vpc_details_with_both_hcp_security_groups(
+    mocker: MockerFixture,
+    aws_api: AWSApi,
+) -> None:
+    """When both legacy -default-sg and newer -vpce-private-router SGs are
+    attached to the VPC endpoint, prefer -default-sg because existing
+    peering/TGW ingress rules are configured on that SG."""
+    expected_vpc_id = "id"
+    mocker.patch.object(
+        aws_api,
+        "get_account_vpcs",
+        return_value=[
+            {
+                "CidrBlock": "cidr",
+                "VpcId": expected_vpc_id,
+            }
+        ],
+    )
+    mocker.patch.object(
+        AWSApi,
+        "_get_vpc_endpoints",
+        return_value=[
+            {
+                "VpcEndpointId": "vpce-id",
+                "Groups": [
+                    {
+                        "GroupName": "abc-default-sg",
+                        "GroupId": "sg-legacy",
+                    },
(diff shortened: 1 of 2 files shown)


lisa:
lgtm, notwithstanding coderabbitbot's things.

BumbleFeng:
Since we already have 21 ingress rules, including 38+ CIDR ranges (peering, TGW, local) on default-sg, we should  prefer the older `-default-sg` SG when both are present

TGPSKI:
> Since we already have 21 ingress rules, including 38+ CIDR ranges (peering, TGW, local) on default-sg, we should prefer the older `-default-sg` SG when both are present

**Feng's point**: The `-default-sg` already has 21 ingress rules with 38+ CIDR ranges (peering, TGW, local traffic). The new `-vpce-private-router` SG is freshly created by the HyperShift operator and doesn't have those rules yet. Preferring the new SG would mean terraform-vpc-peerings references an SG without the right ingress rules.

**Changes made**:
- Flipped `selected = vpce_sg or default_sg` to `selected = default_sg or vpce_sg` -- prefer `-default-sg` when both exist
- Updated docstring to explain the rationale
- Updated test assertion from `"sg-new"` to `"sg-legacy"` to match

BumbleFeng (approved):

PR #5508, opened 2026-04-23
APPSRE 13993 pre repo rebase limit

rhur-pixel:
[APPSRE-13993](https://redhat.atlassian.net/jira/software/c/projects/APPSRE/boards/11510?assignee=712020%3Ae95cb166-7599-4e03-aa29-3d40baa2bfce&selectedIssue=APPSRE-13993&useStoredSettings=true): Redefine rebase limit as per-repo pipeline concurrency cap

Problem:
`limit` resets every reconcile run (~1-2 min cycles), causing massive rebase churn — 2,142 "rebase limit reached" events in a 6h 52m window — with most triggered pipelines wasted.

Change:
`limit` now means "at most N MRs with active/successful pipelines at any time." Before rebasing, count MRs that already have running/pending/success pipelines and only rebase up to limit - already_active.

Backward compatible: when 0 MRs have active pipelines, behavior is identical to the old per-run counter.

Files:
reconcile/gitlab_housekeeping.py — split rebase_merge_requests() into a classification pass and a rebase pass
reconcile/test/test_gitlab_housekeeping.py — 12 tests covering budget calculation, edge cases, and backward compatibility

Validation
pytest: 31 passed
ruff lint: all checks passed
ruff format: clean
mypy: no issues

Assisted by Claude

## Summary by CodeRabbit

* **New Features**
  * Selectable rebase strategies (top-k, active-cap, old-burst) via feature-variant; default remains old-burst. Strategy is resolved once per run.
  * Exposed feature-variant helper and Unleash client helper with safe fallback when env/config is missing.

* **Bug Fixes**
  * Rebase concurrency now correctly accounts for active/in-flight pipelines; timed-out pipelines can be canceled and running pipelines block retries. Failed rebases don’t consume budget.

* **Tests**
  * Comprehensive tests for strategies, budgeting, error handling, per-repo limits, dry-run behavior, and feature-variant mapping.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -513,7 +513,7 @@ def rebase_merge_requests(
         RebaseStrategy.TOP_K: _rebase_merge_requests_top_k,
         RebaseStrategy.OLD_BURST: _rebase_merge_requests_old_burst,
     }
-    logging.info(f"rebase strategy: {strategy.value}")
+
     fn = dispatch[strategy]
     fn(
         dry_run=dry_run,
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -482,22 +482,93 @@ def rebase_merge_requests(
     users_allowed_to_label: Iterable[str] | None = None,
     use_active_cap: bool = False,
 ) -> None:
+    if use_active_cap:
+        return _rebase_merge_requests_active_cap(
+            dry_run=dry_run,
+            gl=gl,
+            rebase_limit=rebase_limit,
+            state=state,
+            pipeline_timeout=pipeline_timeout,
+            wait_for_pipeline=wait_for_pipeline,
+            users_allowed_to_label=users_allowed_to_label,
+        )
 
-    all_items = get_merge_requests(
-        dry_run=dry_run,
-        gl=gl,
-        state=state,
-        users_allowed_to_label=users_allowed_to_label,
-    )
-    # top-k: only evaluate the first rebase_limit MRs in priority order
-    if not use_active_cap:
-        all_items = all_items[:rebase_limit]
+    # top-k: only evaluate the first rebase_limit MRs in priority order.
+    # The queue is already sorted by priority, so slicing to K guarantees
+    # at most K MRs are rebased across all runs.
+    merge_requests = [
+        item["mr"]
+        for item in get_merge_requests(
+            dry_run=dry_run,
+            gl=gl,
+            state=state,
+            users_allowed_to_label=users_allowed_to_label,
+        )[:rebase_limit]
+    ]
 
-    merge_requests = [item["mr"] for item in all_items]
+    for mr in merge_requests:
+        if is_rebased(mr, gl):
+            continue
+
+        pipelines = gl.get_merge_request_pipelines(mr)
+
+        # If pipeline_timeout is None no pipeline will be canceled.
+        if pipeline_timeout is not None:
+            timed_out_pipelines = get_timed_out_pipelines(pipelines, pipeline_timeout)
+            if timed_out_pipelines:
+                clean_pipelines(
+                    dry_run=dry_run,
+                    gl=gl,
+                    fork_project_id=mr.source_project_id,
+                    pipelines=timed_out_pipelines,
+                )
+
+        if wait_for_pipeline:
+            if not pipelines:
+                continue
+            # possible statuses:
+            # running, pending, success, failed, canceled, skipped
+            running_pipelines = [
+                p for p in pipelines if p.status == PipelineStatus.RUNNING
+            ]
+            if running_pipelines:
+                continue
+
+        try:
+            logging.info(["rebase", gl.project.name, mr.iid])
+            if not dry_run:
+                mr.rebase()
+                rebased_merge_requests.labels(mr.target_project_id).inc()
+        except gitlab.exceptions.GitlabMRRebaseError as e:
+            logging.error(f"unable to rebase {mr.iid}: {e}")
+
+
+def _rebase_merge_requests_active_cap(
+    dry_run: bool,
+    gl: GitLabApi,
+    rebase_limit: int,
+    state: State,
+    pipeline_timeout: int | None = None,
+    wait_for_pipeline: bool = False,
+    users_allowed_to_label: Iterable[str] | None = None,
+) -> None:
+    """Active-cap strategy: scan the full queue, count MRs with active
+    pipelines, and only rebase up to (rebase_limit - already_active)
+    additional MRs.  This treats rebase_limit as a per-repo concurrency cap
+    on in-flight pipelines rather than a visibility window."""
+    merge_requests = [
+        item["mr"]
+        for item in get_merge_requests(
+            dry_run=dry_run,
+            gl=gl,
+            state=state,
+            users_allowed_to_label=users_allowed_to_label,
+        )
+    ]
 
     # Single pass: classify MRs as already-active or needs-rebase.
-    # active-cap treats rebase_limit as a per-repo concurrency cap --
-    # "at most N MRs with active pipelines" -- not a per-run burst.
+    # rebase_limit is a per-repo concurrency cap -- "at most N MRs with
+    # active pipelines" -- not a per-run burst.
     # Rebased MRs count as active with running/pending/success pipelines
     # (success = green and waiting to merge, still occupying a slot).
     # Non-rebased MRs only count as active with running/pending pipelines
@@ -506,28 +577,18 @@ def rebase_merge_requests(
     for mr in merge_requests:
         pipelines = gl.get_merge_request_pipelines(mr)
         if is_rebased(mr, gl):
-            if (
-                use_active_cap
-                and pipelines
-                and pipelines[0].status
-                in {
-                    PipelineStatus.RUNNING,
-                    PipelineStatus.PENDING,
-                    PipelineStatus.SUCCESS,
-                }
-            ):
+            if pipelines and pipelines[0].status in {
+                PipelineStatus.RUNNING,
+                PipelineStatus.PENDING,
+                PipelineStatus.SUCCESS,
+            }:
                 already_active += 1
             continue
 
-        if (
-            use_active_cap
-            and pipelines
-            and pipelines[0].status
-            in {
-                PipelineStatus.RUNNING,
-                PipelineStatus.PENDING,
-            }
-        ):
+        if pipelines and pipelines[0].status in {
+            PipelineStatus.RUNNING,
+            PipelineStatus.PENDING,
+        }:
             already_active += 1
             continue
 
@@ -555,9 +616,7 @@ def rebase_merge_requests(
 
         needs_rebase.append(mr)
 
-    remaining_budget = (
-        max(rebase_limit - already_active, 0) if use_active_cap else len(needs_rebase)
-    )
+    remaining_budget = max(rebase_limit - already_active, 0)
     rebases = 0
     for mr in needs_rebase:
         if rebases < remaining_budget:
@@ -716,6 +775,9 @@ def run(dry_run: bool, wait_for_pipeline: bool) -> None:
     repos = queries.get_repos_gitlab_housekeeping(server=instance["url"])
     repos = [r for r in repos if is_in_shard(r["url"])]
     app_sre_usernames: Set[str] = set()
+    use_active_cap = get_feature_toggle_state(
+        "gitlab-housekeeping-rebase-active-cap", default=False
+    )
     state = init_state(QONTRACT_INTEGRATION)
 
     for repo in repos:
@@ -759,9 +821,6 @@ def run(dry_run: bool, wait_for_pipeline: bool) -> None:
             ]
             reload_toggle = ReloadToggle(reload=False)
             rebase = hk.get("rebase")
-            use_active_cap = get_feature_toggle_state(
-                "gitlab-housekeeping-rebase-active-cap", default=False
-            )
             must_pass = hk.get("must_pass")
             try:
                 merge_merge_requests(
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -480,7 +480,6 @@ def rebase_merge_requests(
     wait_for_pipeline: bool = False,
     users_allowed_to_label: Iterable[str] | None = None,
 ) -> None:
-    rebases = 0
     merge_requests = [
         item["mr"]
         for item in get_merge_requests(
@@ -490,13 +489,24 @@ def rebase_merge_requests(
             users_allowed_to_label=users_allowed_to_label,
         )
     ]
+
+    # Single pass: classify MRs as already-active (rebased with a running/pending
+    # pipeline from a previous run) or needs-rebase.  rebase_limit is a per-repo
+    # concurrency cap -- "at most N MRs with active pipelines" -- not a per-run burst.
+    already_active = 0
+    needs_rebase: list[ProjectMergeRequest] = []
     for mr in merge_requests:
         if is_rebased(mr, gl):
+            pipelines = gl.get_merge_request_pipelines(mr)
+            if pipelines and pipelines[0].status in {
+                PipelineStatus.RUNNING,
+                PipelineStatus.PENDING,
+            }:
+                already_active += 1
             continue
 
         pipelines = gl.get_merge_request_pipelines(mr)
 
-        # If pipeline_timeout is None no pipeline will be canceled
         if pipeline_timeout is not None:
             timed_out_pipelines = get_timed_out_pipelines(pipelines, pipeline_timeout)
             if timed_out_pipelines:
@@ -510,15 +520,18 @@ def rebase_merge_requests(
         if wait_for_pipeline:
             if not pipelines:
                 continue
-            # possible statuses:
-            # running, pending, success, failed, canceled, skipped
             running_pipelines = [
                 p for p in pipelines if p.status == PipelineStatus.RUNNING
             ]
             if running_pipelines:
                 continue
 
-        if rebases < rebase_limit:
+        needs_rebase.append(mr)
+
+    remaining_budget = max(rebase_limit - already_active, 0)
+    rebases = 0
+    for mr in needs_rebase:
+        if rebases < remaining_budget:
             try:
                 logging.info(["rebase", gl.project.name, mr.iid])
                 if not dry_run:
@@ -532,7 +545,7 @@ def rebase_merge_requests(
                 "rebase",
                 gl.project.name,
                 mr.iid,
-                "rebase limit reached for this reconcile loop. will try next time",
+                f"rebase limit reached ({already_active} active, limit {rebase_limit}). will try next time",
             ])
 
 
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -496,17 +496,20 @@ def rebase_merge_requests(
     already_active = 0
     needs_rebase: list[ProjectMergeRequest] = []
     for mr in merge_requests:
+        pipelines = gl.get_merge_request_pipelines(mr)
         if is_rebased(mr, gl):
-            pipelines = gl.get_merge_request_pipelines(mr)
+            # pipelines = gl.get_merge_request_pipelines(mr)
             if pipelines and pipelines[0].status in {
                 PipelineStatus.RUNNING,
                 PipelineStatus.PENDING,
+                PipelineStatus.SUCCESS,
             }:
                 already_active += 1
             continue
 
-        pipelines = gl.get_merge_request_pipelines(mr)
+        # pipelines = gl.get_merge_request_pipelines(mr)
 
+        # If pipeline_timeout is None no pipeline will be canceled.
         if pipeline_timeout is not None:
             timed_out_pipelines = get_timed_out_pipelines(pipelines, pipeline_timeout)
             if timed_out_pipelines:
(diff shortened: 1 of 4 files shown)


rhur-pixel:
To be reviewed with @TGPSKI later today.

TGPSKI:
## Code Review

**Overall: Good progress, one blocker remaining.**

### What's Solid

- **Two-phase classify-then-rebase design** is clean and correct
- **SUCCESS pipelines count toward `already_active`** — prevents wasted CI from rebased-and-green MRs that haven't merged
- **API calls consolidated** — `get_merge_request_pipelines` called once per MR instead of twice
- **Log message includes `already_active + rebases`** — accurate diagnostics
- **Tests follow existing repo patterns**: module-level functions, `_` prefixed helpers, reuse of `gitlab_api`/`state` fixtures, `@pytest.mark.parametrize` for limit variations
- **Multi-repo independence test** validates the real call pattern from `run()`
- **All four acceptance criteria from APPSRE-13993 are met**

### Blocker

**Non-rebased MRs with active pipelines aren't counted.** `already_active` only increments inside `if is_rebased(mr, gl):`. A non-rebased MR with a running pipeline (e.g., from a previous rebase still in progress) falls into `needs_rebase` and could get rebased again, exceeding the concurrency cap.

### Should-fix

- Classification comment says "running/pending" but code includes SUCCESS — update the comment
- `test_rebase_budget_exhausted` only asserts on one MR; should assert all three are zero
- Failed rebases don't consume budget — intentional or not, it needs a comment explaining the design choice

### Fine as-is

- `_call_rebase` helper follows `_setup_mocks`, `_make_tgw_desired_state_item`, etc. patterns across the repo
- Black-box function-level tests are consistent with every other test in this file
- `_make_rebase_mr` is not duplicating `can_be_merged_merge_request` — different function under test, different input requirements
- Parameter naming is consistent (`gitlab_api` matches the fixture, production params match the production code)

### Housekeeping

- Branch needs rebase against master before merge

TGPSKI (changes requested):
See comment - https://github.com/app-sre/qontract-reconcile/pull/5508#issuecomment-4315530022

TGPSKI (approved):
All previous findings are resolved. 

- Non-rebased MRs with active pipelines are excluded from needs_rebase
- Comment matches the code
- Budget exhaustion test asserts all MRs
- Failed-rebase design decision is documented
- Branch is up to date with master

reconcile/gitlab_housekeeping.py
```diff
@@ -490,13 +489,31 @@ def rebase_merge_requests(
...
+    # regardless of rebase status.  rebase_limit is a per-repo concurrency
+    # cap -- "at most N MRs with active pipelines" -- not a per-run burst.
+    already_active = 0
+    needs_rebase: list[ProjectMergeRequest] = []
     for mr in merge_requests:
+        pipelines = gl.get_merge_request_pipelines(mr)
+        active_pipeline = pipelines and pipelines[0].status in {
+            PipelineStatus.RUNNING,
+            PipelineStatus.PENDING,
+            PipelineStatus.SUCCESS,
```
fishi0x01: If the pipeline succeeded once, but isnt up to date with `master` anylonger, how will it get a rebase? Wont it always be marked as having `active_pipeline` and always be skipped?

reconcile/gitlab_housekeeping.py
```diff
@@ -518,21 +535,28 @@ def rebase_merge_requests(
             if running_pipelines:
                 continue
 
-        if rebases < rebase_limit:
+        needs_rebase.append(mr)
```
fishi0x01, line 656: this is never reached for a `SUCCESS` pipeline. I.e., if a pipeline succeeds once, it will never receive any rebase, even if it requires one.

reconcile/gitlab_housekeeping.py
TGPSKI: @rhur-pixel can you add this as a test case & update logic

reconcile/gitlab_housekeeping.py
TGPSKI: @rhur-pixel can you add this as a test case & update logic

fishi0x01 (commented):
One more question: What would be the effect on priorized MRs? We label MRs with priority labels, so high prio MRs get merged faster. My understanding from this change is, that we will have a fixed number of rebased MRs - new MRs will have to wait, even if they are higher prio. Is that correct?

**Update**:

Maybe we need to track number of MRs per prio level, so low level long running checks dont block high prio MRs. But Im not 100% sure if this properly aligns with the previous design

rhur-pixel:
@fishi0x01 Great question — @TGPSKI and I worked through this today.

**Short answer:** Yes, new MRs wait for a slot, even critical ones. We believe this is correct.

Priority already controls **selection into the budget** — `preprocess_merge_requests` sorts by `(label_priority, approved_at)`, so when a slot frees up, the highest-priority MR always gets it next. Priority governs *queue order*, not what happens inside the budget.

### The scenario

`rebase_limit = 2`, both slots full, a critical MR arrives:

```mermaid
graph LR
    subgraph "Active Budget (limit=2, full)"
        A["MR-A: High\n7/10 min"]
        B["MR-B: Medium\n5/10 min"]
    end
    subgraph "needs_rebase[] (priority sorted)"
        C["MR-C: Critical"]
        D["MR-D: High"]
        E["MR-E: Medium"]
    end
```

**Option A — Preempt (swap MR-B out for MR-C):**

```mermaid
graph LR
    subgraph "Active Budget"
        A1["MR-A: High 7/10 min"]
        C1["MR-C: Critical 0/10 min"]
    end
    subgraph "needs_rebase[]"
        D1["MR-D: High"]
        B1["MR-B: Medium (5 min wasted)"]
        E1["MR-E: Medium"]
    end
```

MR-A merges 3 min later → master advances → MR-C's pipeline is **stale**, must re-rebase. MR-B's 5 min also **wasted**. 

Net: **15 min CI wasted**, critical MR no closer to merging.

**Option B — Wait for next slot (our design):**

```mermaid
graph LR
    subgraph "~3 min later: MR-A merges, slot freed"
        C2["MR-C: Critical 0/10 min"]
        B2["MR-B: Medium 8/10 min"]
    end
    subgraph "needs_rebase[]"
        D2["MR-D: High"]
        E2["MR-E: Medium"]
    end
```

MR-C gets the slot immediately (first in priority queue). MR-B's work preserved. 

Net: **0 min CI wasted**, critical MR waited ~3 min.

### Why Option B

- Preemption wastes CI without meaningfully speeding up the critical MR — it still needs a full ~10 min pipeline regardless.
- Worst-case wait is bounded by the longest running pipeline (minutes, not hours).
- Keeps implementation simple (`remaining_budget` stays an integer).
- Preserves pipeline work for future optimistic multi-merge (Phase 1).

### On per-priority budgets

At `rebase_limit=2` you can't meaningfully split slots. At higher limits, reserved slots sit idle when no critical MRs exist. Current design keeps all slots utilized — we can revisit if critical MRs are materially delayed in practice.

Generated by Claude

reconcile/gitlab_housekeeping.py
```diff
@@ -490,13 +489,34 @@ def rebase_merge_requests(
...
     ]
+
+    # Single pass: classify MRs as already-active or needs-rebase.
+    # rebase_limit is a per-repo concurrency cap -- "at most N MRs with
+    # active pipelines" -- not a per-run burst.
+    # Rebased MRs count as active with running/pending/success pipelines
+    # (success = green and waiting to merge, still occupying a slot).
+    # Non-rebased MRs only count as active with running/pending pipelines
+    already_active = 0
+    needs_rebase: list[ProjectMergeRequest] = []
```
hemslo, line 639: Do we need that complex processing here? Can we say the rebase feature only keep top `rebase_limit` merge requests to be rebased. We already got sorted merge requests in queue in `get_merge_requests` result. Can we simply do

```python
    merge_requests = [
        item["mr"]
        for item in get_merge_requests(
            dry_run=dry_run,
            gl=gl,
            state=state,
            users_allowed_to_label=users_allowed_to_label,
        )[:rebase_limit]
    ]
```

rhur-pixel: This suggestion is similar to the original implementation, where `rebase_limit` defined the amount of MR's to rebase in a given run. The complex processing is implemented to reduce the amount of wasted CI pipelines that run as a result of rebasing MR's that don't need to. 

The new processing logic caps the amount of concurrent pipelines to `rebase_limit` to prevent MR's from rebasing unnecessarily. For example, with your suggestion, if 3 MR's are in flight and rebase_limit = 5, 2 more MR's will have their rebase pipelines triggered, which will ultimately have to be restarted once a single MR gets merged.

fishi0x01:
> Short answer: Yes, new MRs wait for a slot, even critical ones. We believe this is correct.

@rhur-pixel we have MRs that take potentially easily >10m. Im still not really convinced its a good idea to let low prio MRs block high prio ones. But I wont die on that hill :D 
Do we have metrics to compare if the new algo brings any real improvement? I know we have some time-to-merge dashboards - not sure we should maybe extend that first to have a fair comparison of before and after in terms of saved rebases and potential time-to-merge worsening for high prio MRs.

reconcile/gitlab_housekeeping.py
hemslo: No, it's not similar, think of the real workflow business logic, the goal is to rebase top K (at most rebase_limit) candidates so one of them pass CI build can merge.

Given a merge queue with N merge requests
When rebase_merge_requests
Then top K merge requests are checked for rebase

Note the merge requests in queue always sorted, so if we only lookup first K merge requests, we can guarantee at most K rebases across all runs.

The problem for original code is find top K merge requests that can rebase, so multiple runs will eventually rebase all merge requests.

You can paste this comment to your AI Agent to do math prove.

TGPSKI: I think top-K bounds the wrong resource.

For this fix, and especially for future optimistic multi-merge, the useful unit is not “one of the first K MRs was checked.” It is “this MR has a valid active/successful pipeline against the current target SHA/root.”

A strict top-K slice makes that pool smaller and more brittle. Slow, flaky, blocked, or already-active MRs inside the first K consume the whole visibility window, while lower queue entries that could be filling useful CI slots are invisible.

That reduces scan breadth, but it does not model pipeline inventory. The active-cap model is closer to the invariant we actually want: maintain up to N useful in-flight/successful candidates per repo, then refill slots by priority as they free up.

TGPSKI:
I'm adding a task to build out a simulator that models algorithm changes vs impact on merge queue performance. This will back up the proposed changes .

reconcile/gitlab_housekeeping.py
hemslo: our merge queue is a priority queue, for example current rebase limit is 2, the top 2 merge requests are rebased and pipeline is running, now there is new higher merge request come in, we must rebase it and if that pipeline success before the other 2 merge requests, merge the new top priority one.

hemslo: keep the changes to minimal, just consider top K, monitor the metrics after rollout, then check if need more complex handling.

TGPSKI: Preemption makes sense if the value of starting the critical MR a few minutes earlier exceeds the cost of wasting in-flight CI and shrinking the same-root success pool.

Given end goal of muti-merge, my intent is total queue throughput improvement Vs keeping strict priority queue. 

A simulator will help a lot here I think.

TGPSKI: https://github.com/app-sre/qontract-reconcile/blob/master/docs%2Fadr%2FADR-019-merge-queue-acceleration.md#L240

hemslo: top K as default, put new logic behind a feature toggle, do metrics compare in prod

fishi0x01:
should we maybe have some real metrics tracked? I know we have time-to-merge metrics per priority level. Maybe lets also include something that measures what we try to solve here: less rebases. Can we somehow track the number of rebases? Then we can extend our current dashboard and compare algo changes on real data

rhur-pixel:
Will do, thank you all for the suggestions. We're going to start planning out for a feature toggle, added metrics, and a simulation to help guide us throughout the rest of the epic, hopefully making reviews more concrete. Will update soon!

reconcile/gitlab_housekeeping.py
```diff
@@ -715,6 +759,9 @@ def run(dry_run: bool, wait_for_pipeline: bool) -> None:
             ]
             reload_toggle = ReloadToggle(reload=False)
             rebase = hk.get("rebase")
+            use_active_cap = get_feature_toggle_state(
+                "gitlab-housekeeping-rebase-active-cap", default=False
+            )
```
hemslo: since toggle is unrelated to repo, this can move outside of repos loop

reconcile/gitlab_housekeeping.py
```diff
@@ -479,24 +480,58 @@ def rebase_merge_requests(
...
-    ]
+
+    all_items = get_merge_requests(
+        dry_run=dry_run,
+        gl=gl,
+        state=state,
+        users_allowed_to_label=users_allowed_to_label,
+    )
+    # top-k: only evaluate the first rebase_limit MRs in priority order
+    if not use_active_cap:
```
hemslo: to avoid check `use_active_cap` multiple times, and to make later code cleanup easier, better to move all new logic in a new method, like this:

```python
def rebase_merge_requests(
    dry_run: bool,
    gl: GitLabApi,
    rebase_limit: int,
    state: State,
    pipeline_timeout: int | None = None,
    wait_for_pipeline: bool = False,
    use_active_cap: bool = False,
) -> None:
    if use_active_cap:
      return rebase_merge_requests_use_active_cap(...)
    # old logic
    ...
```

reconcile/gitlab_housekeeping.py
hemslo: now that new logic behind toggle, we need to deal with real cases:

Given rebase limit is 2, 3 MRs already rebased and pipeline running
When 1 of 3 pipelines success and MR merged
Then the remaining 2 MR should be rebased immediately even their pipelines are running

Due the the new active counting logic, rebase won't happen until pipeline finish, we are wasting time waiting.

The new active cap should only be paired with later phases change, not current rebase & merge flow.

rhur-pixel: Thank you for the review. I removed the stale counting filter from the active_cap algorithm, hopefully that addresses this.

TGPSKI: @hemslo 

> Due the the new active counting logic, rebase won't happen until pipeline finish, we are wasting time waiting.

Ryan's update removes stale counting.

> The new active cap should only be paired with later phases change, not current rebase & merge flow.

I don't think this logic is valid. active-cap could outperform top-k or old-burst with adjustments to counting logic. Keep the options open until proven otherwise.

hemslo: The new logic fixed waiting for stale pipeline running, but may lead to high priority slow mr stay in queue for longer time.

Given rebase limit is 2, 3 high priority MRs already rebased and pipeline running (average 5 mins)
When 1 of 3 pipelines success and MR merged and 2 new low priority MRs with 10 mins pipeline run time come
Then we will wait for 10 mins to merge one of new low priority mrs (like promote qontract reoncile / schema)

We need to measure SLO of gitlab housekeeping, like expect every merge request can be merged within X mins (or dynamic X based priority), measure P99 of merge time not just merge throughput, as delay slow merge requests can improve throughput but hurt priority (already the case in current logic, but high priority ones still have a chance to merge as it's always rebased).

Also when we promote top K changes, we need to bump `gitlabHousekeeping.limit` from `2` to a higher value like `4` to keep it similar as current rebase logic (multiple runs).

Now that we have feature toggle, we can experiment and compare real time metrics.

TGPSKI:
## Status Update — May 4, 2026

**Alignment meeting with @hemslo completed.** Summary of outcomes:

### Agreed

- **All three policies ship in one MR, one deploy** — active-cap (default), top-k, old-burst selectable via Unleash
- **old-burst is retired as default** — data shows it produces 341 rebases vs 187 (active-cap) and 210 (top-k) with identical throughput. No speed benefit, just CI waste. **edit**: deploy with old-burst as default in this PR re: [additional environment consideration ](https://github.com/app-sre/qontract-reconcile/pull/5508#discussion_r3186354216)cc @rhur-pixel 
- **Limit stays at 2** — sweep data shows raising limit only increases CI churn without throughput gain. Exception: failure-aware exclusion (MRs with failed pipelines consuming a slot) may justify limit=4 temporarily until that's addressed in a follow-up.
- **Limit increase only justified for Phase 1** (optimistic multi-merge) — simulation confirms this is the only scenario where higher limits produce material benefit.
- **Median time-to-merge by priority is identical across all three policies** — CI duration is the physical floor, not policy selection.
- **Run each policy in prod for ~1 week**, measure and compare. I may adjust measurement time period down, we will see.

### Pending

- @rhur-pixel newly pushed unleash client string-variant refactor ( `get_feature_variant` returning strategy name).  If not, fall back to two boolean flags.

### Simulation Data (60-day production calibration)

| Metric | old-burst | top-k | active-cap |
|--------|-----------|-------|------------|
| Total rebases (24h) | 341 | 210 | 187 |
| Wasted rebases | ~240 | ~140 | ~120 |
| Throughput | identical | identical | identical |
| Critical p95 latency | 196 min | 91 min | 91 min |
| Starved MRs | 11 | 14 | 12 |

Full methodology, interactive simulator, and sweep data available — will share in team update.

* [Simulator views from re: Limit increase](https://github.com/app-sre/qontract-reconcile/pull/5508#discussion_r3186070372)
* [Simulator data on median and p95 time to merge per strategy: active-cap, top-k, old-burst](https://github.com/app-sre/qontract-reconcile/pull/5508#discussion_r3186078530)

reconcile/gitlab_housekeeping.py
TGPSKI: Regarding limits: 

I support limit increase to get around high priority failed + retry loop blocking top-k slot, otherwise, limit increase only valuable in ADR-019 scope. top-k keeps the failed MR in the visible window, while active cap won't count the failed pipeline towards already_active.

Details:

Limit increases offer no throughput increase for any merge algorithm (active-cap, top-k, old-burst):

image

 and increase CI churn and load.

image

active-cap + phase 1 (optimistic multi merge non-overlapping labels) keeps CI load consistent across the limit sweep

image

TGPSKI: Regarding high-priority starvation:

Median time-to-merge per priority label is relatively consistent for each priority type. The simulation suggests CI time dominates time in merge-queue for each MR and algorithms are not the limiting factor with our production conditions.

image

TGPSKI (approved):
* pipeline_timeout + wait_for_pipeline + rebase logic is copied to all strategy functions
* looking for local dry runs for new unleash method and glhk fallback behavior to increase confidence

reconcile/gitlab_housekeeping.py
hemslo: `DEFAULT_REBASE_STRATEGY` should be `OLD_BURST` to keep the behaviour same as before. For easier fedramp promotion.

TGPSKI: Yes

reconcile/gitlab_housekeeping.py
```diff
@@ -479,7 +506,183 @@ def rebase_merge_requests(
...
+        )
+
+
+def _should_skip_for_running_pipeline(pipelines: list, wait_for_pipeline: bool) -> bool:
+    """Return True if the MR should be skipped because a pipeline is still running."""
+    if not wait_for_pipeline:
+        return False
+    if not pipelines:
+        return True
+    return any(p.status == PipelineStatus.RUNNING for p in pipelines)
```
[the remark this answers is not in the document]
rhur-pixel: The if not pipelines: continue was intentional in all three strategies. When wait_for_pipeline=True, the code says: "I need to see a pipeline result before rebasing. If there are no pipelines at all, I can't verify anything, so skip this MR."

If you change it to return False (don't skip), you'd rebase MRs that have no pipeline at all — meaning the rebase would fire without any pipeline gate. That defeats the purpose of wait_for_pipeline.

The _should_skip_for_running_pipeline helper faithfully preserves the original logic. The docstring could be more precise about the empty-pipelines case, but the logic itself is correct and matches what's been running in production. Skip this one.

Written by Claude, and also I agree 👍

reconcile/gitlab_housekeeping.py
```diff
@@ -479,7 +506,183 @@ def rebase_merge_requests(
...
+            }:
+                already_active += 1
+            continue
+
+        _cancel_timed_out_pipelines(dry_run, gl, mr, pipelines, pipeline_timeout)
+
+        if _should_skip_for_running_pipeline(pipelines, wait_for_pipeline):
+            continue
+
+        needs_rebase.append(mr)
```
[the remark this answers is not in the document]
TGPSKI: The active-cap logic is correct as written. Cancel-stale-Jenkins-on-rebase is a pure cost optimization with no impact on merge throughput or correctness.

A proper fix for CI waste would be a separate concern — something like _cancel_stale_pipelines_on_rebase that cancels RUNNING/PENDING pipelines at the point of rebase. That's an optimization for runner cost, not for concurrency correctness.

reconcile/gitlab_housekeeping.py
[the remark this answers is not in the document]
TGPSKI: Verdict: Valid finding, but it's a pre-existing gap, not a bug introduced by this PR. The old code also only checked RUNNING. Ryan's refactor preserved the existing behavior correctly.

wait_for_pipeline is false in production

TGPSKI:
**Status: Ready to merge**

1. Locally tested and validated — including Unleash strategy tri-state selector (strategy ordering confirms correct variant dispatch across all three policies).
2. Merge + deployment scheduled for tomorrow morning (May 6, NASA time).

hemslo (approved):
LGTM!

PR #5511, opened 2026-04-24
introduce ERv2 support for vpc-endpoint-service and vpc-endpoint providers

jfchevrette:
Adds qontract-reconcile support for two new ERv2-managed resource types implementing AWS PrivateLink via [er-aws-vpc-endpoint-service](https://github.com/app-sre/er-aws-vpc-endpoint-service) (provider side) and [er-aws-vpc-endpoint](https://github.com/app-sre/er-aws-vpc-endpoint) (consumer side).

## Changes

- Added inline fragments and generated Pydantic models for two new resource types:

  - NamespaceTerraformResourceVpcEndpointService_v1
  - NamespaceTerraformResourceVpcEndpoint_v1 

## Dependencies

- https://github.com/app-sre/qontract-schemas/pull/994

## Related

- Module repos: [er-aws-vpc-endpoint-service](https://github.com/app-sre/er-aws-vpc-endpoint-service), [er-aws-vpc-endpoint](https://github.com/app-sre/er-aws-vpc-endpoint)
- [Design doc](https://github.com/openshift-online/platform-engineering-enhancements/blob/main/app-sre/design-docs/er-aws-privatelink.md)

## Summary by CodeRabbit

* **New Features**
  * Support for AWS VPC Endpoints and VPC Endpoint Services as managed external resources: exposed to inventory/API, selectable VPC/subnets, service consumer allowlisting, and recognized when marked managed_by_erv2; provider/region and output/module settings respected.
* **Tests**
  * Added unit tests covering spec resolution, private-subnet filtering, ARN construction and merging with explicit ARNs, and related edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/app-sre/qontract-reconcile/pull/5511?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/reconcile/external_resources/model.py
+++ b/reconcile/external_resources/model.py
@@ -26,6 +26,7 @@ from reconcile.gql_definitions.external_resources.external_resources_namespaces 
     NamespaceTerraformResourceRDSProxyV1,
     NamespaceTerraformResourceRDSV1,
     NamespaceTerraformResourceVpcEndpointServiceV1,
+    NamespaceTerraformResourceVpcEndpointV1,
     NamespaceV1,
 )
 from reconcile.gql_definitions.external_resources.external_resources_settings import (
@@ -110,6 +111,7 @@ SUPPORTED_RESOURCE_TYPES = (
     | NamespaceTerraformResourceCloudflareZoneV1
     | NamespaceTerraformResourceCloudflareAccountV1
     | NamespaceTerraformResourceVpcEndpointServiceV1
+    | NamespaceTerraformResourceVpcEndpointV1
 )
 
 
--- a/reconcile/external_resources/model.py
+++ b/reconcile/external_resources/model.py
@@ -25,6 +25,7 @@ from reconcile.gql_definitions.external_resources.external_resources_namespaces 
     NamespaceTerraformResourceMskV1,
     NamespaceTerraformResourceRDSProxyV1,
     NamespaceTerraformResourceRDSV1,
+    NamespaceTerraformResourceVpcEndpointServiceV1,
     NamespaceV1,
 )
 from reconcile.gql_definitions.external_resources.external_resources_settings import (
@@ -108,6 +109,7 @@ SUPPORTED_RESOURCE_TYPES = (
     | NamespaceTerraformResourceRDSProxyV1
     | NamespaceTerraformResourceCloudflareZoneV1
     | NamespaceTerraformResourceCloudflareAccountV1
+    | NamespaceTerraformResourceVpcEndpointServiceV1
 )
 
 
--- a/reconcile/gql_definitions/external_resources/external_resources_namespaces.gql
+++ b/reconcile/gql_definitions/external_resources/external_resources_namespaces.gql
@@ -501,6 +501,31 @@ query ExternalResourcesNamespaces {
               ...ExternalResourcesModuleOverrides
             }
           }
+          ... on NamespaceTerraformResourceVpcEndpointService_v1 {
+            region
+            identifier
+            openshift_service_name
+            allowed_consumer_clusters {
+              name
+              spec {
+                ... on ClusterSpecROSA_v1 {
+                  account {
+                    uid
+                    terraformUsername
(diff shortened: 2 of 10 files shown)


reconcile/gql_definitions/external_resources/external_resources_namespaces.gql
chassing, line 1: During my MSK connect development, also a new ERv2 module, I also needed to update `reconcile/gql_definitions/terraform_resources/terraform_resources_namespaces.gql`. Did you run `terraform-resources` against a namespace with your ERv2 module?

reconcile/external_resources/model.py
chassing, line 1: I don't see any changes to `reconcile/external_resources/aws.py` or other file in `reconcile/external_resources`. Can `qontract-reconcile` compile the ERv2 input file?

reconcile/gql_definitions/external_resources/external_resources_namespaces.gql
jfchevrette: Interesting. I only tested with `external-resources`. Why would `terraform-resources` need to run for an ERv2 module?

reconcile/external_resources/model.py
jfchevrette: `qontract-cli external-resources ... get-input` does give me an input file, which I could use in a module container to render and apply resources.

From your PR, I notice the changes to the factories, but TBH I don't understand why they are needed.... I only tested the vpce mdoules manually/locally ... Are the factories needed somewhere that a local dev environment would not?

jfchevrette: Ah I see some modules need it (msk-connect for instance) to do additional processing such as reading a secret, or mapping spec keys to terraform module keys. 

For VPC endpoints, I could probably implement a factory to get info about the VPC endpoint service, and also to generate the ARNs instead of doing it in terraform

reconcile/gql_definitions/external_resources/external_resources_namespaces.gql
chassing: No, you don't need `terraform-resources` to run any ERv2, but because of the GQL query in `terraform-resources`, you may see errors for your resources. The `terraform-resources` query must just know the provider and can ignore it via the `managed_by_erv2` attribute.

reconcile/external_resources/model.py
hemslo: Get info about VPC endpoint service from saved terraform outputs secrets is useful here as it allows to feed one output as input to the other module.

What's ARNs generate case? If you can do it completely in terraform, that's better as it will be easier to migrate to HCP Terraform.

reconcile/gql_definitions/external_resources/external_resources_namespaces.gql
jfchevrette: I added the resources to terraform_resources_namespaces.gql/.py with minimal fields

reconcile/external_resources/model.py
jfchevrette: I implemented the factory for ARN generation. There is two fields now: `allowed_consumer_clusters` gets the cluster AWS account ID and terraformUsername and returns the ARN. There is also `allowed_principal_arns` for the cases we may want to set a specific ARN, for instance we want to allow a non-managed cluster to establish a conenction. Both get merged and a list of ARNs is passed to the module. I like that the business-specific logic is encoded in q-r where the rest of the business logic is, and the erv2 module remains as generic as possible

I'm considering getting the vpc-endpoint-service info from vault, but since we cannot $ref to a resource fields within a data file, I think I will need the user to input cluster name, namespace name, and resource name.... Or I move the VPCES fields to a "defaults" file, which we can link to and get the info there... But I'm not sure there's a lot of benefit over having the user enter the VPC endpoint service name / identifier...

hemslo: it's ok for now, we can adjust it later when required, the user here currently only us, so we can pick any methods suitable for us to use.

hemslo (approved):
LGTM!

jfchevrette:
@coderabbitai full review

PR #5517, opened 2026-04-29
feat(oum): support authenticated LDAP/FreeIPA for group member resolution

chassing:
Extend qontract_utils/ldap_api with get_group_members to query group
memberships via authenticated FreeIPA binds. Redesign LdapGroup model
with typed members and CN extraction. Update OUM providers to use the
new API.

Ticket: [APPSRE-13659](https://redhat.atlassian.net/browse/APPSRE-13659)

## Summary by CodeRabbit

* **New Features**
  * Added group-member retrieval returning groups with canonical name and member user objects.

* **Bug Fixes**
  * Centralized LDAP response validation and standardized LDAP error messages.
  * Ignores unrelated memberships and properly escapes DNs in searches.

* **Chores**
  * STARTTLS enabled by default for LDAP connections; provider wiring updated to use the new API.

* **Tests**
  * Updated and expanded tests for group-member retrieval, DN handling, errors, retries and TLS.

### DIFF
--- a/qontract_utils/qontract_utils/ldap_api/api.py
+++ b/qontract_utils/qontract_utils/ldap_api/api.py
@@ -3,6 +3,7 @@
 import contextvars
 import time
 import types
+from collections import defaultdict
 from collections.abc import Iterable
 from dataclasses import dataclass
 from typing import Self
@@ -20,10 +21,11 @@ from ldap3.core.exceptions import (
     LDAPSocketSendError,
 )
 from ldap3.utils.conv import escape_filter_chars
+from ldap3.utils.dn import parse_dn
 from prometheus_client import Counter, Histogram
 
 from qontract_utils.hooks import Hooks, RetryConfig, invoke_with_hooks, with_hooks
-from qontract_utils.ldap_api.models import LdapUser
+from qontract_utils.ldap_api.models import LdapGroup, LdapUser
 from qontract_utils.metrics import DEFAULT_BUCKETS_EXTERNAL_API
 
 logger = structlog.get_logger(__name__)
@@ -149,7 +151,7 @@ class LdapApi:
         bind_dn: str | None = None,
         bind_password: str | None = None,
         *,
-        start_tls: bool = False,
+        start_tls: bool = True,
         timeout: int = _DEFAULT_TIMEOUT,
         hooks: Hooks | None = None,  # noqa: ARG002 - Handled by @with_hooks decorator
     ) -> None:
@@ -178,6 +180,15 @@ class LdapApi:
     ) -> None:
         self._connection.unbind()
 
+    @staticmethod
+    def _check_ldap_response(status: dict) -> None:
+        """Check LDAP response status and raise LdapApiError on failure."""
+        if (error_code := status.get("result", 99999)) != 0:
+            error_desc = status.get("description", "unknown error")
+            raise LdapApiError(
+                f"LDAP operation failed (error {error_code}: {error_desc})"
+            )
+
     @invoke_with_hooks(
         lambda: LdapApiCallContext(method="get_users"),
         retry_config=_LDAP_RETRY_CONFIG,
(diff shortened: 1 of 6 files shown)


qontract_utils/qontract_utils/ldap_api/api.py
```diff
@@ -149,7 +151,7 @@ def __init__(
         bind_dn: str | None = None,
         bind_password: str | None = None,
         *,
-        start_tls: bool = False,
+        start_tls: bool = True,
```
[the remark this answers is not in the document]
chassing: won't change or fix. it's fine and on purpose

reconcile/test/ocm/oum/test_oum_providers.py
```diff
@@ -1,23 +1,39 @@
...
+            dn="cn=group2,dc=example,dc=com",
+            members=frozenset({
+                LdapUser(username="user3"),
+                LdapUser(username="user4"),
+            }),
+        ),
+    ]
     lc.__enter__ = lc
     lc.__exit__ = mocker.Mock(return_value=None)
     return lc
```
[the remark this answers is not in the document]
chassing: This is incorrect. `lc.__enter__ = lc` sets `__enter__` to the mock object itself. When Python executes `with self.ldap_client as x:`, it calls `lc.__enter__()` which is effectively `lc()` — and calling a Mock returns `lc.return_value`. So `x` is `lc.return_value`, and `lc.return_value.get_group_members.return_value` is the correct configuration path.

The passing tests confirm this — they assert on the concrete return values and would fail if the mock were misconfigured.

qontract_utils/qontract_utils/ldap_api/api.py
```diff
@@ -200,11 +211,43 @@ def get_users(self, usernames: Iterable[str]) -> list[LdapUser]:
...
+            attributes=["uid", "memberOf"],
+        )
+
+        self._check_ldap_response(status)
+
+        groups_and_members: dict[str, set[str]] = defaultdict(set[str])
+        for u in users:
+            uid = u["attributes"]["uid"][0]
+            for group in set(u["attributes"]["memberOf"]).intersection(groups_dns):
+                groups_and_members[group].add(uid)
```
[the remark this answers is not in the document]
chassing: not an issue. won't fix

qontract_utils/qontract_utils/ldap_api/api.py
```diff
@@ -200,11 +211,43 @@ def get_users(self, usernames: Iterable[str]) -> list[LdapUser]:
...
-            raise LdapApiError(f"LDAP search failed (error {error_code}: {error_desc})")
+        self._check_ldap_response(status)
 
         return [LdapUser(username=r["attributes"]["uid"][0]) for r in results]
+
+    @invoke_with_hooks(
+        lambda: LdapApiCallContext(method="get_group_members"),
+        retry_config=_LDAP_RETRY_CONFIG,
+    )
+    def get_group_members(self, groups_dns: Iterable[str]) -> list[LdapGroup]:
```
hemslo: Nit: `Iterable[str]` allows single-use iterators (generators), but this value is consumed twice — once by `sorted()` (line 230) and again by `.intersection()` (line 243). Consider using `Collection[str]` instead, which guarantees `__len__`, `__contains__`, and `__iter__`, matching how the parameter is actually used.

qontract_utils/qontract_utils/ldap_api/api.py
```diff
@@ -178,6 +180,15 @@ def __exit__(
     ) -> None:
         self._connection.unbind()
 
+    @staticmethod
+    def _check_ldap_response(status: dict) -> None:
+        """Check LDAP response status and raise LdapApiError on failure."""
+        if (error_code := status.get("result", 99999)) != 0:
+            error_desc = status.get("description", "unknown error")
```
hemslo, line 198: Nit: The magic number `99999` as a fallback for a missing `result` key is a bit opaque. Consider extracting it to a named constant (e.g., `_UNKNOWN_LDAP_ERROR = 99999`) or just checking `status.get("result") != 0` with a separate branch for a missing key. (Pre-existing from `get_users`, but now that it's in a shared helper it's a good time to clean it up.)

qontract_utils/qontract_utils/ldap_api/api.py
```diff
@@ -200,11 +211,43 @@ def get_users(self, usernames: Iterable[str]) -> list[LdapUser]:
...
+
+        groups_and_members: dict[str, set[str]] = defaultdict(set[str])
+        for u in users:
+            uid = u["attributes"]["uid"][0]
+            for group in set(u["attributes"]["memberOf"]).intersection(groups_dns):
+                groups_and_members[group].add(uid)
+
+        return [
+            LdapGroup(
+                cn=parse_dn(dn)[0][1],
```
hemslo: `parse_dn(dn)[0][1]` assumes the first RDN component is always `cn`. While this is the standard convention for group DNs, a malformed or non-standard DN would silently produce a wrong CN. Consider validating:
```python
rdn = parse_dn(dn)[0]
if rdn[0].lower() != "cn":
    raise LdapApiError(f"Expected CN as first RDN component, got {rdn[0]!r} in {dn!r}")
```

reconcile/oum/providers.py
```diff
@@ -51,12 +54,20 @@ def init_ldap_group_member_provider(group_base_dn: str) -> LdapGroupMemberProvid
     """
 
     settings = get_ldap_settings()
+    vault_settings = get_app_interface_vault_settings()
+    secret_reader = create_secret_reader(use_vault=vault_settings.vault)
+    bind_dn = None
+    bind_password = None
+    if settings.credentials:
+        ldap_credentials = secret_reader.read_all_secret(settings.credentials)
+        bind_dn = ldap_credentials.get("bind_dn")
```
hemslo: If `settings.credentials` is set but the Vault secret is missing the `bind_dn` / `bind_password` keys, `.get()` silently returns `None` and falls back to anonymous bind — which undermines the intent of authenticated binds. Consider using `ldap_credentials["bind_dn"]` (KeyError on missing key) or adding an explicit check, so a misconfigured secret fails loudly rather than silently degrading.

qontract_utils/qontract_utils/ldap_api/api.py
hemslo: or add a function to get `cn`

qontract_utils/qontract_utils/ldap_api/api.py
chassing: Changed in ecd13c7

qontract_utils/qontract_utils/ldap_api/api.py
chassing: Changed in ecd13c7

qontract_utils/qontract_utils/ldap_api/api.py
chassing: Changed in ecd13c7

reconcile/oum/providers.py
chassing: Changed in ecd13c7

hemslo (approved):
LGTM!

PR #5518, opened 2026-04-30
fix(quay-mirror): fail integration when skopeo copy errors occur

jfreixa:
Previously, SkopeoCmdError exceptions were swallowed and logged, causing the integration to exit 0 even when image copies failed. Now QuayMirror.run() tracks errors and returns a bool; the module-level run() exits with ExitCodes.ERROR if any copy failed, while still attempting all remaining tasks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Mirror runs now collect individual image copy failures and fail at the end of a run, reporting all copy errors together instead of silently continuing.

* **Tests**
  * Added unit tests covering successful mirror runs and runs that raise an aggregated copy-error when some copy attempts fail (including multi-task scenarios).

### DIFF
--- a/reconcile/quay_mirror.py
+++ b/reconcile/quay_mirror.py
@@ -126,8 +126,8 @@ class QuayMirror:
     def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
         self.session.close()
 
-    def run(self) -> bool:
-        error = False
+    def run(self) -> None:
+        errors: list[Exception] = []
         sync_tasks = self.process_sync_tasks()
         for org, data in sync_tasks.items():
             for item in data:
@@ -140,12 +140,13 @@ class QuayMirror:
                     )
                 except SkopeoCmdError as details:
                     _LOG.error("skopeo command error message: '%s'", details)
-                    error = True
+                    errors.append(details)
 
         if self.is_compare_tags and not self.dry_run:
             record_timestamp(self.control_file_path)
 
-        return error
+        if errors:
+            raise ExceptionGroup("skopeo copy failures", errors)
 
     @classmethod
     def process_repos_query(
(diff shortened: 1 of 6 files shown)


reconcile/quay_mirror.py
```diff
@@ -126,7 +126,8 @@ def __enter__(self) -> Self:
     def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
         self.session.close()
 
-    def run(self) -> None:
+    def run(self) -> bool:
+        error = False
```
hemslo: instead of bool, we can collect all exceptions and throw a final `ExceptionGroup`

jfreixa: like this?

hemslo: yes, please also check other similar integrations `quay_mirror_org` and `gcr_image_mirror`

reconcile/test/test_gcp_image_mirror.py
```diff
@@ -123,3 +124,54 @@ def test_process_repos_to_sync(
...
+            mocker.patch.object(qm, "process_sync_tasks", return_value=tasks)
+            mocker.patch("reconcile.gcp_image_mirror.gql_gcp_repos")
+            mock_skopeo.copy.side_effect = [SkopeoCmdError("exit code: 1"), None]
+
+            with pytest.raises(ExceptionGroup) as exc_info:
+                qm.run()
+
+            assert mock_skopeo.copy.call_count == 2
+            assert len(exc_info.value.exceptions) == 1
+            assert isinstance(exc_info.value.exceptions[0], SkopeoCmdError)
```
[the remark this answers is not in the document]
hemslo: this is a valid comment

hemslo (approved):
LGTM!

PR #5520, opened 2026-05-04
fix(slack-usergroups): use workspace user IDs for Enterprise Grid users

jfreixa:
SlackUser.id returns the enterprise W... ID when enterprise_user is set,
but usergroups_list returns workspace-level U... IDs in ug.users. This
caused Enterprise Grid users to be silently dropped from the current state
on every reconcile cycle, making them always appear in users_to_add and
causing an infinite loop (observed: bhthakur, jbanerje, faldana).

Fix: build a dual-keyed lookup (both U... and W... → org_username) in
get_slack_usergroups so Enterprise Grid users are correctly resolved from
Slack usergroup membership.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Slack Enterprise Grid user handling in usergroup management by enhancing user identification across workspace and enterprise grid contexts, ensuring users remain properly associated with usergroups

* **Tests**
  * Added test coverage for Enterprise Grid user resolution in usergroup retrieval and configuration
  * Added test coverage validating Enterprise Grid user identification when syncing usergroup membership to Slack

### DIFF
--- a/qontract_api/qontract_api/slack/slack_workspace_client.py
+++ b/qontract_api/qontract_api/slack/slack_workspace_client.py
@@ -291,9 +291,15 @@ class SlackWorkspaceClient:
         channel_name_by_id = {
             channel.id: channel.name for channel in self.get_channels().values()
         }
-        org_username_by_id = {
-            user.id: user.org_username for user in self.get_users().values()
-        }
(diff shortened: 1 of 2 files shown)


hemslo (approved):
LGTM!

chassing (approved):

PR #5522, opened 2026-05-04
qontract-api: remove worker health check

chassing:
Follows #5519

The `qontract-api` server depends on the Redis cache but not on the healthiness of the `qontract-api-workers`. In case of worker issues, all API pods would go in an unhealthy state, and `contract-api` at all isn't working anymore :( 

## Summary by CodeRabbit

* **Changes**
  * Health endpoint no longer reports a separate worker component; overall system health is now based solely on the cache component, so worker failures are no longer reflected.

* **Tests**
  * Readiness tests updated to validate health responses using only the cache component status.


jfreixa (approved):

hemslo (approved):
LGTM!

PR #5527, opened 2026-05-06
perf(jinja2): add thread-safe caching for vault, s3, github, and quer…

jfreixa:
…y template functions

Cache vault reads at path level using read_all, deduplicate github/s3/query fetches within a run, and serialize concurrent fetches of the same key with per-entry locks to avoid thundering-herd duplicate API calls.

## Summary by CodeRabbit

* **Performance Improvements**
  * Added per-run template caching to speed resource templating and reduce redundant external lookups (secrets, files, queries, S3), improving runtime.
* **Behavioral Improvements**
  * More consistent templating across resource types (including Prometheus rules) for more reliable rendered outputs.
* **Tests**
  * Test harness updated to remain compatible with the new caching behavior.

### DIFF
--- a/reconcile/openshift_resources_base.py
+++ b/reconcile/openshift_resources_base.py
@@ -1,4 +1,5 @@
 import base64
+import functools
 import hashlib
 import itertools
 import json
@@ -45,6 +46,7 @@ from reconcile.utils.defer import defer
 from reconcile.utils.exceptions import FetchResourceError
 from reconcile.utils.jinja2.utils import (
     FetchSecretError,
+    Jinja2TemplateCache,
     process_extracurlyjinja2_template,
     process_jinja2_template,
 )
@@ -513,6 +515,7 @@ def fetch_openshift_resource(
     parent: Mapping[str, Any],
     settings: Mapping | None = None,
     skip_validation: bool = False,
+    cache: Jinja2TemplateCache | None = None,
 ) -> OR:
     provider = resource["provider"]
     if provider == "resource":
@@ -553,9 +556,9 @@ def fetch_openshift_resource(
         tt = resource["type"]
         tt = "jinja2" if tt is None else tt
         if tt == "jinja2":
-            tfunc = process_jinja2_template
+            tfunc = functools.partial(process_jinja2_template, cache=cache)
         elif tt == "extracurlyjinja2":
-            tfunc = process_extracurlyjinja2_template
+            tfunc = functools.partial(process_extracurlyjinja2_template, cache=cache)
         else:
             raise UnknownTemplateTypeError(tt)
         try:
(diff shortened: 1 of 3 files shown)


reconcile/utils/jinja2/utils.py
```diff
@@ -111,6 +112,36 @@ def init_github() -> Github:
...
+) -> threading.Lock:
+    with meta_lock:
+        if key not in locks:
+            locks[key] = threading.Lock()
+        return locks[key]
+
+
+_github_cache: dict[tuple[str, str, str], str] = {}
+_github_locks: dict[tuple[str, str, str], threading.Lock] = {}
+_github_locks_lock: threading.Lock = threading.Lock()
```
hemslo: we can't cache it cross multiple runs, as fetch from ref like `master` will change. Same for other caches.

jfreixa: this is a cli everyexecution will have empty cache right?

hemslo: no, global variables like module vars here are kept.

chassing: CLI execution, yes for MR checks, but not for the prod integration. Here it's a `while .. sleep` look

jfreixa: should I empty them after the run of the integration?

chassing (approved):
Let's give it a try.

jfreixa:
Addressing two CodeRabbit comments:

**Lock type annotation** (comment on line 149): Not a real issue — \`mypy\` passes cleanly in strict mode with \`threading.Lock\` as a type annotation. In the typeshed stubs, mypy understands \`threading.Lock\` in annotation position as the lock instance type. Switching to \`_thread.LockType\` would introduce a private/internal import with no practical benefit.

**Vault cache key missing backend discriminator** (comment on line 344): Not actionable in practice. Within a single integration run, \`secret_reader\` is always the same backend (Vault or Config), determined once by \`settings\`. Cross-backend collision within one run cannot happen at runtime. We are also exploring removing \`secret_reader\` as a parameter entirely, which would make this moot.

Fixed the \`json.dumps\` non-serializable kwargs issue (the only valid comment) in the next commit.

reconcile/utils/jinja2/utils.py
jfreixa: added cache in a object that is recreated every run

reconcile/openshift_resources_base.py
```diff
@@ -792,6 +800,7 @@ def fetch_data(
     use_jump_host: bool,
     init_api_resources: bool = False,
     overrides: Iterable[str] | None = None,
+    cache: Jinja2TemplateCache | None = None,
```
hemslo: type check, `cache` always passed in as `Jinja2TemplateCache`, not None

reconcile/openshift_resources_base.py
```diff
@@ -757,6 +763,7 @@ def fetch_states(
     spec: ob.StateSpec,
     ri: ResourceInventory,
     settings: Mapping[str, Any] | None = None,
+    cache: Jinja2TemplateCache | None = None,
```
hemslo: type check, cache always passed in as Jinja2TemplateCache, not None

reconcile/openshift_resources_base.py
```diff
@@ -706,9 +709,12 @@ def fetch_desired_state(
     parent: Mapping[str, Any],
     privileged: bool,
     settings: Mapping[str, Any] | None = None,
+    cache: Jinja2TemplateCache | None = None,
```
hemslo: type check, cache always passed in as Jinja2TemplateCache, not None

hemslo (approved):
LGTM!

PR #5530, opened 2026-05-07
fix(avs): skip ElastiCache MRs during engine upgrades, write engine in overrides

TGPSKI:
## Summary

* Add `elasticache_engine_version_consistent()` guard that skips MR creation when the metric engine+version combination is contradictory (e.g. `engine="redis"` with version >= 7.2)
* Write `overrides["engine"]` alongside `engine_version` in namespace YAML so MRs are self-consistent with ERV2 expectations
* Fix MR dedup logic: use stable resource identity for lookup, compare both engine AND version for no-op check

## Context

* **Jira:** APPSRE-14128
* **Root cause:** During a Redis→Valkey in-place upgrade, the exporter may briefly emit a transitional metric state where engine and engine_version are inconsistent. AVS picked this up and opened MR !186963 with stale data that immediately failed ERV2 validation ("Valkey requires engine_version 7.2 or higher").
* **Key insight (from Wang Di):** This is a timing issue during upgrades. Once the upgrade completes, metrics stabilize and AVS won't regenerate the stale MR.

## Approach

1. **Skip inconsistent metrics** — if `engine="redis"` but version >= 7.2 (or `engine="valkey"` but version < 7.2), skip with a warning. This catches the transitional window.
2. **Write both fields** — when metrics ARE consistent, write both `engine` and `engine_version` in overrides so the MR is self-consistent regardless of what the defaults file says.
3. **Fix dedup** — remove `resource_engine` from MR lookup key so engine transitions match existing MRs by stable identity.

## Changes

| File | Change |
|------|--------|
| `reconcile/aws_version_sync/integration.py` | Add `elasticache_engine_version_consistent()` + `ELASTICACHE_VALKEY_MIN_VERSION`; skip inconsistent metrics in `reconcile()` |
| `reconcile/aws_version_sync/merge_request_manager/merge_request.py` | `render_merge_request_content` accepts optional `resource_engine`; writes `overrides["engine"]` when provided |
| `reconcile/aws_version_sync/merge_request_manager/merge_request_manager.py` | Pass `resource_engine` to renderer; fix dedup key |
| `reconcile/test/...` | Updated tests: consistency check parametrized tests, mid-upgrade skip scenario, engine override in renderer |

## Test plan

* `pytest reconcile/test/aws_version_sync/ -v` — 50 passed, 5 xfailed
* `ruff check` — all checks passed
* `ruff format --check` — all files formatted
* New `ec_mid_upgrade` test fixture verifies inconsistent metrics are skipped
* New `test_elasticache_engine_version_consistent` parametrized tests cover boundary cases

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed database engine upgrade detection logic to properly handle cases where both engine type and version require updates, ensuring accurate synchronization of AWS RDS and ElastiCache resources.

[Review Change Stack](https://app.coderabbit.ai/change-stack/app-sre/qontract-reconcile/pull/5530)

### DIFF
--- a/reconcile/aws_version_sync/integration.py
+++ b/reconcile/aws_version_sync/integration.py
@@ -52,6 +52,8 @@ from reconcile.utils.vcs import VCS
 
 QONTRACT_INTEGRATION = "aws-version-sync"
 
+ELASTICACHE_VALKEY_MIN_VERSION = semver.VersionInfo(major=7, minor=2, patch=0)
+
 
 class AVSIntegrationParams(PydanticRunParams):
     prometheus_timeout: int = 10
@@ -296,6 +298,26 @@ class AVSIntegration(QontractReconcileIntegration[AVSIntegrationParams]):
                 )
         return metrics
 
+    @staticmethod
+    def _resolve_elasticache_engine(
+        reported_engine: str, engine_version: str
+    ) -> str:
+        """Resolve the actual ElastiCache engine from reported metric labels.
+
+        The aws-resources-exporter may still label Valkey-backed replication
+        groups as 'redis' because the metric name predates the Valkey fork.
+        AWS ElastiCache Valkey starts at version 7.2; any version >= 7.2 on an
+        ElastiCache cluster is Valkey regardless of what the metric label says.
+        """
+        try:
+            version = parse_semver(engine_version, optional_minor_and_patch=True)
+        except ValueError:
+            return reported_engine
+
+        if version >= ELASTICACHE_VALKEY_MIN_VERSION:
+            return "valkey"
+        return reported_engine
+
     def _fetch_elasticache_metrics(
         self,
         cluster: ClusterV1,
@@ -321,6 +343,9 @@ class AVSIntegration(QontractReconcileIntegration[AVSIntegrationParams]):
 
         for m in elasticache_metrics:
             try:
+                engine = self._resolve_elasticache_engine(
+                    m["engine"], m["engine_version"]
+                )
                 metrics.append(
                     ExternalResource(
                         provider="aws",
@@ -328,7 +353,6 @@ class AVSIntegration(QontractReconcileIntegration[AVSIntegrationParams]):
                             uid=m["aws_account_id"]
                         ),
                         resource_provider=SupportedResourceProvider.ELASTICACHE,
-                        # replication_group_id != resource_identifier!
                         resource_identifier=elasticache_replication_group_id_to_identifier.get(
                             (
                                 m["aws_account_id"],
@@ -336,7 +360,7 @@ class AVSIntegration(QontractReconcileIntegration[AVSIntegrationParams]):
                             ),
                             m["replication_group_id"],
                         ),
-                        resource_engine=m["engine"],
+                        resource_engine=engine,
                         resource_engine_version=m["engine_version"],
                     )
                 )
--- a/reconcile/aws_version_sync/integration.py
+++ b/reconcile/aws_version_sync/integration.py
@@ -55,6 +55,20 @@ QONTRACT_INTEGRATION = "aws-version-sync"
 ELASTICACHE_VALKEY_MIN_VERSION = semver.VersionInfo(major=7, minor=2, patch=0)
 
 
+def elasticache_engine_version_consistent(
+    engine: str, version: semver.VersionInfo
+) -> bool:
+    """Return True if engine and version are a valid ElastiCache combination.
+
+    Valkey requires >= 7.2; Redis must be < 7.2.  During in-place upgrades the
+    exporter may briefly emit a transitional state where these don't match.
+    """
+    is_valkey_version = version >= ELASTICACHE_VALKEY_MIN_VERSION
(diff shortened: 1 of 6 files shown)


reconcile/aws_version_sync/integration.py
```diff
@@ -296,6 +298,24 @@ def _fetch_rds_metrics(
...
 
+    @staticmethod
+    def _resolve_elasticache_engine(reported_engine: str, engine_version: str) -> str:
+        """Resolve the actual ElastiCache engine from reported metric labels.
+
+        The aws-resources-exporter may still label Valkey-backed replication
+        groups as 'redis' because the metric name predates the Valkey fork.
+        AWS ElastiCache Valkey starts at version 7.2; any version >= 7.2 on an
+        ElastiCache cluster is Valkey regardless of what the metric label says.
+        """
```
hemslo: this is not true, `engine` in metrics is correct.

reconcile/aws_version_sync/integration.py
```diff
@@ -452,6 +467,21 @@ def reconcile(
             ):
                 # do not downgrade the version
                 continue
```
hemslo:
```suggestion
            if (
                aws_resource.resource_engine_version
                <= app_interface_resource.resource_engine_version
                or aws_resource.resource_engine
                != app_interface_resource.resource_engine
            ):
                # do not downgrade the version or change engine
                continue
```

reconcile/aws_version_sync/integration.py
```diff
@@ -447,10 +447,10 @@ def reconcile(
...
                 aws_resource.resource_engine_version
                 <= app_interface_resource.resource_engine_version
-                and aws_resource.resource_engine
-                == app_interface_resource.resource_engine
+                or aws_resource.resource_engine
+                != app_interface_resource.resource_engine
             ):
-                # do not downgrade the version
+                # do not downgrade the version or change engine
                 continue
```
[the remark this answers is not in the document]
TGPSKI: This is desired behavior. AVS is scoped to version bumps, not engine changes.

The trade-off is intentional: after a Redis→Valkey upgrade stabilizes, the engine field in app-interface won't auto-update from redis to valkey. That engine field update would need to happen through a separate process (manual MR or a different integration).

reconcile/test/aws_version_sync/test_avs_integration.py
```diff
@@ -520,7 +520,7 @@ def test_avs_reconcile(mocker: MockerFixture, intg: AVSIntegration) -> None:
         external_resources_aws=external_resources_aws,
         external_resources_app_interface=external_resources_app_interface,
     )
-    assert merge_request_manager_mock.create_merge_request.call_count == 4
+    assert merge_request_manager_mock.create_merge_request.call_count == 2
```
hemslo, line 523: this test case is for aws auto upgraded elasticache from redis to valkey, do we still need to cover this case, as it's conflicting with app-interface upgrade redis to valkey. @chassing

TGPSKI: I asked my agent to reset the test file and add the new tests only and it repeated this change.

TGPSKI: Well, there's no getting around the fact that changing the logic invalidates the call_count as written, and moves it to `call_count = 2`.

I've proposed reducing the call count, or changing the test cases. I'm getting frustrated that this PR is stuck and not moving. What do you want me to do @hemslo ? Close this PR? 


```
The OR logic changes the behavior of 2 existing test cases:

ec_engine_update: engine="valkey", version="13.1" (AWS) vs engine="redis", version="13.1" (AI). Same version, different engine → OR catches the engine mismatch → skipped. Previously created MR.

ec_engine_and_version_update: engine="valkey", version="15.1" (AWS) vs engine="redis", version="13.1" (AI). Higher version, different engine → OR catches the engine mismatch → skipped. Previously created MR.

The OR logic inherently prevents AVS from creating MRs when the engine differs, which is exactly what those 2 test cases exercise. 

So 4 → 2 is the correct expected count with this logic change. The existing test needs == 2 and the 2 engine-change assert_has_calls entries removed.
```

hemslo: We need to verify the real cases:

* AWS force upgrade redis 6.2 to valkey 8.1
  * old logic: AVS open mr to override version from 7 to 8, but this mr will fail due to engine mismatch
  * new logic: AVS skip this due to engine mismatch, the real terraform run may fail due to target redis version not available anymore
* User upgrade redis 6.2 to valkey 8.1
  * old logic: AVS open mr to override version to 6.2 when upgrade not applied yet in AWS
  * new logic: AVS skip this due to engine mismatch, no side effects

Based on this analysis, skip when engine mismatch is better as old logic has issues in both cases. Note down we don't auto sync version when engine changed, add drift caused by AWS force redis -> valkey as a known issues in SOP.

hemslo (approved):
LGTM!

chassing:
Superseded by #5773

PR #5534, opened 2026-05-11
:heavy_plus_sign: use httpxyz instead of httpx

chassing:
Replace unmaintained httpx with [httpxyz](https://httpxyz.org/why-fork/)

Ticket: [APPSRE-14239](https://redhat.atlassian.net/browse/APPSRE-14239)

## Release Notes

* **New Features**
  * Added builder-style HTTP client configuration methods (`with_headers`, `with_cookies`, `with_timeout`)
  * Added context manager support for both synchronous and asynchronous HTTP clients

* **Chores**
  * Updated internal HTTP client dependency
  * Added compatibility mechanisms for seamless integration with existing code

[Review Change Stack](https://app.coderabbit.ai/change-stack/app-sre/qontract-reconcile/pull/5534)


jfreixa (approved):
pipeline failing but LGTM

qontract_api_client/openapi_python_client_templates/client.py.jinja
```diff
@@ -0,0 +1,191 @@
+import ssl
+from typing import Any
+
+from attrs import define, field, evolve
+import httpxyz as httpx
```
fishi0x01, line 5: my understanding is this file is copied from openapi lib (default template) and this is the line we need to fix right? Are there any potential issues when we bump openapi versions? When bumping lib versions, do we need to compare with upstream template and synch any changes here too?

fishi0x01 (approved):

qontract_api_client/openapi_python_client_templates/client.py.jinja
chassing: Potentially, yes, same as for the other template with modifications, but it should be negligible

qontract_utils/qontract_utils/pagerduty_api/client.py
```diff
@@ -10,6 +10,8 @@
 from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
 
+# import httpxyz before pagerduty to ensure httpxyz is overridding the httpx used by pagerduty
```
hemslo: This import-order trick relies on `httpxyz` registering itself as `sys.modules["httpx"]` before `pagerduty` does `import httpx` internally. It works, but it's fragile:

1. **Ruff/isort will want to reorder this.** `httpxyz` sorts after `pagerduty` alphabetically, and `import httpxyz` looks unused (the `allowed-unused-imports` config only suppresses pyflakes, not isort). If someone runs auto-sort, this silently breaks at runtime — pagerduty gets the real (uninstalled) httpx and fails. Consider adding `# noqa: I001` or an isort `forced_separate` config to pin the ordering.

2. **The mechanism is invisible.** A future developer removing this "unused" import won't know it's load-bearing without reading the comment carefully. Consider adding a runtime assertion after the pagerduty import:
```python
import httpxyz
import sys
assert sys.modules.get("httpx") is httpxyz, "httpxyz must be imported before pagerduty to override httpx"
```
This turns a silent runtime mystery into a clear failure.

hemslo (commented):
Two pieces of feedback:

1. **This decision should be documented as an ADR.** Switching from a top-100 PyPI package (`httpx`, 500k+ dependent repos) to a week-old fork is a significant architectural decision with supply-chain implications. An ADR in `docs/adr/` should capture: why httpx is being replaced (disabled issues, no releases since Dec 2024, unresponsive maintainer), why httpxyz was chosen over alternatives (API-compatible fork, drop-in for openapi-python-client templates), the `sys.modules` registration mechanism for transitive deps, and — critically — re-evaluation criteria (what happens if httpxyz goes dormant, or httpx resumes maintenance). NiceGUI's [approach](https://github.com/zauberzeug/nicegui/issues/6024) of explicitly planning to "re-evaluate before committing" is worth emulating.

2. **The pagerduty import-order trick needs protection.** See inline comment.

qontract_utils/qontract_utils/pagerduty_api/client.py
chassing: Di, looks like our LLM doesn't know what **alphabetically** means. Ruff/isort doesn't want to reorder the import!

hemslo: it's an example, if there is another package `import aaa` which also import `httpx` internally, then the order will be changed

chassing:
@hemslo Thanks for the feedback! Both points addressed:

1. **ADR-020** added in `docs/adr/ADR-020-replace-httpx-with-httpxyz.md` — documents why httpx is being replaced, why httpxyz was chosen over alternatives, the `sys.modules` registration mechanism, and re-evaluation criteria (dormant httpxyz, httpx resumes, supply-chain incident).

2. **Centralized compat module** — created `qontract_utils/_httpxyz_compat.py` (inspired by [NiceGUI's approach](https://github.com/evnchn/nicegui/pull/148)). It's imported from `qontract_utils/__init__.py` so httpxyz is registered as `sys.modules["httpx"]` before any transitive dependency can `import httpx`. Includes a `RuntimeError` guard if real httpx was loaded first. The fragile import-order trick in `pagerduty_api/client.py` has been removed — the compat module handles it centrally now. Subprocess-isolated tests cover all scenarios.

hemslo (approved):
Both feedback items addressed thoroughly:

1. **ADR-020** documents the decision, alternatives considered, and re-evaluation criteria — exactly what was needed for a supply-chain change like this.

2. **Centralized compat module** (`_httpxyz_compat.py`) is a clean solution — eliminates the fragile per-file import-order trick, provides a runtime assertion, and has proper subprocess-isolated tests. Much better than the original approach.

LGTM.

evnchn:
@chassing @hemslo I see this PR as it shows up in both https://github.com/zauberzeug/nicegui/issues/6024 (the NiceGUI issue) and https://github.com/evnchn/nicegui/pull/148 (my private fork's PR)

Owing to https://tildeweb.nl/~michiel/httpx2.html I have thus closed that PR. You can read more in the NiceGUI issue but the TL-DR is that **it's a bad idea to lock in to 1 `httpx` sucessor when there's more than 1.** 

You may want to reconsider whether to ship this. Feel free to ping me for my takes but my role otherwise stops at bringing you the message. 

(Also, it may not be the best idea to follow a private fork's PR, when that fork's PRs are to track my crazy ideas not suit for upstream yet, if you look at the vibes of https://github.com/evnchn/nicegui/pulls in general...)

PR #5537, opened 2026-05-11
APPSRE 13994 healthcheck probe pipeline failures

rhur-pixel:
## APPSRE-13994: Healthcheck-probe model for pipeline failures using pipeline history

[APPSRE-13994](https://redhat.atlassian.net/browse/APPSRE-13994): Healthcheck-probe model for pipeline failures using pipeline history
[APPSRE-13992](https://redhat.atlassian.net/browse/APPSRE-13996): Flag MRs that fail to merge repeatedly with merge-error label

### Summary

Adds a label-based DLQ to `gitlab-housekeeping` that flags MRs stuck due to merge-time exceptions or consecutive pipeline failures. Flagged MRs stay visible in the merge queue for IC triage but are skipped during merge and rebase processing.

| Label | Trigger | Auto-recovers? |
|---|---|---|
| `merge-error` | `GitlabOperationError` on `mr.merge()` | No |
| `merge-error/pipeline` | N consecutive pipeline failures (default 3) | Yes — removed when pipeline succeeds |

### What changed

- **Labels**: Added `MERGE_ERROR`, `MERGE_ERROR_PIPELINE` constants and `DLQ_LABELS` set
- **Merge loop**: DLQ'd MRs skipped before any pipeline/rebase API calls
- **Rebase loops**: DLQ skip added to all three strategies (active-cap, top-k, old-burst) to stop wasting rebase cycles
- **Merge error catch**: `GitlabOperationError` handler applies `merge-error` label (after existing `GitlabMRClosedError` handler)
- **Pipeline healthcheck**: `run_pipeline_healthcheck()` runs once per reconcile loop (outside `@retry`), applies/removes `merge-error/pipeline` based on consecutive failures
- **Only `FAILED` counted**: `CANCELED`/`SKIPPED` excluded from failure detection to avoid false-positives from `clean_pipelines` timeout cancellations
- **Prometheus counter**: `dead_lettered_merge_requests_total` with `reason=merge_error|pipeline`
- **Tests**: Merge error labeling, healthcheck detection/recovery, merge-loop skip, queue visibility

### Validation
- **Unit tests**: All new paths covered — merge error labeling (operation error, closed error, dry-run), healthcheck (all-fail, mixed, recovery, configurable N), merge-loop skip, queue visibility
- **Linter/types**: `ruff check`, `ruff format`, and `mypy` all pass
- **Local dry-run**: Ran `qd profile run dev gitlab-housekeeping` against real dev bundle — integration boots and completes without exceptions, healthcheck pre-pass executes, no regressions in queue behavior
- **Post-deploy plan**: Monitor pod logs for `add_label`/`remove_label` with DLQ labels, verify flagged MRs appear on Inscope merge queue, watch for false positives over 24-48h

### Follow-up TODOs

- Slack alert if DLQ count > 0 for > 1 hour
- Cache `get_merge_request_pipelines` to avoid double fetch between healthcheck and merge loop
- Subtract DLQ'd MRs from `merge_requests_waiting` gauge (currently inflated)
- Tune `consecutive_failure_limit` default (3 may be too aggressive for flaky CI — consider 5)
- Add `consecutive_failure_limit` to app-interface `gitlabHousekeeping` schema

Created by Claude

## Summary by CodeRabbit

* **New Features**
  * MRs are auto-labeled for merge or pipeline errors; error-labeled MRs are excluded from rebasing and merging. Merge limit and pipeline-failure window are configurable per repo. Error counts exposed via observability metric and merge queue reports include an error column.

* **Bug Fixes**
  * Merge failures now mark MRs as merge-error (non-dry runs) and healthchecks mark/unmark pipeline-error based on recent pipeline outcomes and human activity.

* **Tests**
  * Expanded coverage for error labeling, healthcheck behavior, dry-run semantics, skipping logic, and end-to-end error cycles.

[Review Change Stack](https://app.coderabbit.ai/change-stack/app-sre/qontract-reconcile/pull/5537)


TGPSKI:
## Code Review: Dead Letter Queue for Stuck Merge Requests

### Overall Assessment

Well-structured feature. The label-based DLQ design is sound: persistent state via GitLab labels, automatic recovery from pipeline failures, and DLQ'd MRs remain visible in the queue for triage. Test coverage is solid for the happy paths. CI passes clean.

A few issues worth addressing before merge — two potential runtime problems and a test coverage gap.

---

### Bugs / Must-Fix

#### 1. `consecutive_failure_limit` is not validated — risk of `TypeError` or silent wrong behavior

**File:** `reconcile/gitlab_housekeeping.py` — the `run()` function

```python
consecutive_failure_limit = hk.get("consecutive_failure_limit") or 3
```

This handles `None` (key absent) correctly, but:
- A **negative value** (e.g., `-1`) is truthy and passes through. `pipelines[:-1]` silently slices off the last element, making `check_pipeline_health` always return `True` (healthy) — defeating the entire feature.
- A **non-integer** from config (e.g., a string `"3"`) passes through and causes a `TypeError` on `len(recent) < consecutive_failure_limit`.
- `0` is coerced to `3` via `or`, which is fine but not obvious.

Since schema validation is a follow-up TODO, the code needs a defensive guard. Suggestion:

```python
raw = hk.get("consecutive_failure_limit")
try:
    consecutive_failure_limit = max(1, int(raw))  # type: ignore[arg-type]
except (TypeError, ValueError):
    consecutive_failure_limit = 3
```

#### 2. `run_pipeline_healthcheck` has no error handling — can skip merge/rebase for entire repo

In `run()`, the healthcheck is called before merge and rebase processing. If `run_pipeline_healthcheck` raises an unexpected exception (e.g., a GitLab API error on `get_merge_request_pipelines`), it will propagate up and **skip all merge and rebase processing** for that repository. The healthcheck is purely advisory — it should not block core operations.

Suggestion:

```python
try:
    run_pipeline_healthcheck(...)
except Exception:
    logging.exception("pipeline healthcheck failed, continuing with merge/rebase")
```

---

### Design Observations (no action required)

#### 3. Exception catch ordering is correct ✅

`GitlabMRClosedError` inherits from `GitlabOperationError` in python-gitlab. The PR correctly catches the more specific exception first (log only) before the broader one (log + label). If reversed, closed-MR errors would incorrectly get the `merge-error` label.

#### 4. DLQ skip in rebase is complete ✅

On master, `rebase_merge_requests` is a dispatcher to the three strategy functions (`top_k`, `active_cap`, `old_burst`), all of which have the DLQ skip. No unprotected path.

#### 5. In-memory label staleness is handled correctly ✅

`gl.add_label_to_merge_request` calls `refresh_labels` internally, updating in-memory `mr.labels` from the server. So when healthcheck labels an MR, the subsequent merge loop sees the updated labels within the same reconcile cycle.

#### 6. Double pipeline fetch (acknowledged follow-up)

`run_pipeline_healthcheck` fetches pipelines for every MR, then `merge_merge_requests` fetches them again. This multiplies API calls. Noted as a follow-up TODO in the PR description.

#### 7. `merge_requests_waiting` gauge inflated by DLQ'd MRs (acknowledged follow-up)

The DLQ skip happens after `preprocess_merge_requests` returns and the gauge is set, so DLQ'd MRs are counted as waiting. Also noted as a follow-up TODO.

---

### Test Coverage Gaps

#### 8. No tests for rebase DLQ skip ⚠️

The DLQ skip was added to all three rebase strategies, but there are zero tests verifying this. The `_make_rebase_mr` helper was updated to accept `labels`, but no test creates an MR with a DLQ label and verifies it's skipped during rebase. Recommend adding at least one parametrized test across both DLQ labels.

#### 9. No tests for `run_pipeline_healthcheck` early-exit branches

`run_pipeline_healthcheck` has three untested early-exit conditions:
- Draft MRs (`if mr.draft: continue`)
- MRs with `CANNOT_BE_MERGED` / `CANNOT_BE_MERGED_RECHECK` status
- MRs with no pipelines (`if not pipelines: continue`)

#### 10. No dry-run test for `run_pipeline_healthcheck`

There's a dry-run test for merge error labeling but none for the healthcheck's own `if not dry_run:` guards.

---

### Minor / Nits

- **`PIPELINE_FAILURE_STATUSES` is a single-element set** — works fine for extensibility, but a comment explaining the design intent would help.
- **`set(mr.labels)` is constructed repeatedly** in the merge loop and all three rebase strategies. Could assign to a local once per iteration.
- **Unrelated whitespace change in `labels.py`** — removes blank lines before the `AVS` comment. Clutters blame history.

---

### Summary

| # | Issue | Severity | Action |
|---|-------|----------|--------|
| 1 | `consecutive_failure_limit` not validated | **Bug** | Fix before merge |
| 2 | `run_pipeline_healthcheck` can block merge/rebase | **Bug** | Wrap in try/except |
| 3 | Exception catch ordering | ✅ OK | Verified correct |
| 4 | Rebase DLQ coverage complete | ✅ OK | Verified |
| 5 | In-memory label staleness | ✅ OK | Verified correct |
| 6 | Double pipeline fetch | Debt | Follow-up TODO |
| 7 | `merge_requests_waiting` gauge inflated | Debt | Follow-up TODO |
| 8 | No rebase DLQ skip tests | **Gap** | Add tests |
| 9 | No healthcheck early-exit tests | Gap | Nice to have |
| 10 | No healthcheck dry-run test | Gap | Nice to have |

**Recommendation:** Address items 1 and 2 (both small fixes), add rebase DLQ skip tests (item 8), then this is good to merge.

reconcile/utils/mr/labels.py
```diff
@@ -7,6 +7,8 @@
 DO_NOT_MERGE_PENDING_REVIEW = "do-not-merge/pending-review"
 HOLD = "bot/hold"
 LGTM = "lgtm"
+MERGE_ERROR = "merge-error"
+MERGE_ERROR_PIPELINE = "merge-error/pipeline"
```
hemslo: Consider renaming to `pipeline-error` instead of `merge-error/pipeline`.

The slash implies a namespace hierarchy (like `do-not-merge/hold`), but `merge-error` and `pipeline-error` are sibling error states. Flat names are cleaner and consistent.

reconcile/gitlab_housekeeping.py
```diff
@@ -73,6 +75,8 @@
     NEEDS_REBASE,
 ]
 
+DLQ_LABELS = {MERGE_ERROR, MERGE_ERROR_PIPELINE}
```
hemslo: Rather than a module-level `DLQ_LABELS` set checked in 4 separate places, consider adding an `"error"` boolean flag to the dict returned by `preprocess_merge_requests`:

```python
item = {
    "mr": mr,
    ...
    "error": bool(ERROR_LABELS & set(mr.labels)),
}
```

Callers check `merge_request["error"]` — one place derives it, all loops consume it. This is consistent with how `label_priority`, `approved_at`, and `approved_by` are already centralised in `preprocess_merge_requests`.

reconcile/gitlab_housekeeping.py
```diff
@@ -108,6 +112,12 @@
     labelnames=["name"],
 )
 
+dead_lettered_merge_requests = Counter(
```
hemslo: Consider replacing this `Counter` with a `Gauge` named `qontract_reconcile_merge_requests_error`, mirroring `qontract_reconcile_merge_requests_waiting`:

```python
merge_requests_error = Gauge(
    name="qontract_reconcile_merge_requests_error",
    documentation="Number of merge requests stuck in an error state.",
    labelnames=["project_id"],
)
```

Set it once per reconcile loop after `preprocess_merge_requests` with the count of MRs where `item["error"]` is true. A gauge reflects the *current* queue state and naturally drops to 0 when the label is removed — a counter never goes down, which makes alerting on recovery impossible.

reconcile/gitlab_housekeeping.py
```diff
@@ -595,6 +623,9 @@ def _rebase_merge_requests_top_k(
     ]
 
     for mr in merge_requests:
+        if DLQ_LABELS & set(mr.labels):
```
hemslo: In the `top-k` strategy the error check runs after slicing to `[:rebase_limit]`, so an error MR consumes one of the K slots and can starve a healthy MR.

If the flag moves to `preprocess_merge_requests` (see comment on `DLQ_LABELS`), filtering before or after the slice becomes an explicit decision rather than an emergent side-effect.

reconcile/gitlab_housekeeping.py
```diff
@@ -753,6 +790,12 @@ def merge_merge_requests(
 
     for merge_request in merge_requests:
         mr: ProjectMergeRequest = merge_request["mr"]
+
+        dlq_labels = DLQ_LABELS & set(mr.labels)
```
hemslo: With an `"error"` flag in the `preprocess_merge_requests` dict this simplifies to:

```python
if merge_request["error"]:
    logging.info(["skip merge", gl.project.name, mr.iid])
    continue
```

No inline set intersection or local variable needed.

reconcile/gitlab_housekeeping.py
```diff
@@ -893,6 +991,22 @@ def run(dry_run: bool, wait_for_pipeline: bool) -> None:
             project_merge_requests = [
                 mr for mr in opened_merge_requests if mr.state == MRState.OPENED
             ]
+            raw_limit = hk.get("consecutive_failure_limit")
+            try:
+                consecutive_failure_limit = max(1, int(raw_limit))
+            except (TypeError, ValueError):
+                consecutive_failure_limit = 3
```
hemslo: Silent fallback hides misconfiguration. Consider a warning:

```python
except (TypeError, ValueError):
    logging.warning(
        f"invalid consecutive_failure_limit {raw_limit!r}, defaulting to 3"
    )
    consecutive_failure_limit = 3
```

reconcile/gitlab_housekeeping.py
hemslo: and we don't need to add `DLQ` as a new concept, as an error flag is enough for this use case, the dead queue in design doc was a thinking process.

reconcile/gitlab_housekeeping.py
rhur-pixel: ~~This makes the logs a little noisy, but that is alright since schema validation is in TODO, thank you!~~

EDIT: I'll include this check but allow `None` configuration values to go through with no logging since None is expected default

reconcile/gitlab_housekeeping.py
```diff
@@ -585,18 +612,22 @@ def _rebase_merge_requests_top_k(
...
             state=state,
             users_allowed_to_label=users_allowed_to_label,
-        )[:rebase_limit]
-    ]
+        )
+        if not item["error"]
+    ][:rebase_limit]
+
+    for item in merge_requests:
+        mr = item["mr"]
```
hemslo: I don't see any other fields in `item` used in this method, why not

```python
    merge_requests = [
        item["mr"]
        for item in get_merge_requests(
            dry_run=dry_run,
            gl=gl,
            state=state,
            users_allowed_to_label=users_allowed_to_label,
        )
        if not item["error"]
    ][:rebase_limit]
```

same for other usages in `_rebase_merge_requests_active_cap` and `_rebase_merge_requests_old_burst`

reconcile/gitlab_housekeeping.py
hemslo: You have to add `consecutive_failure_limit` to schema https://github.com/app-sre/qontract-schemas/blob/main/schemas/app-sre/app-1.yml, limit value to be positive, and update gql query to include the new filed, otherwise the key is always missing in `hk`.

Then you don't need to validate type error as schema validation is done during bundle.


```python
raw_limit = hk.get("consecutive_failure_limit") or 3
```

reconcile/gitlab_housekeeping.py
```diff
@@ -752,9 +789,17 @@ def merge_merge_requests(
...
+    merge_requests_error.labels(gl.project.id).set(
+        sum(1 for item in merge_requests if item["error"])
+    )
 
     for merge_request in merge_requests:
         mr: ProjectMergeRequest = merge_request["mr"]
+
+        if merge_request["error"]:
+            logging.info(["skip merge", gl.project.name, mr.iid])
+            continue
```
hemslo, line 799: If we always skip merge for both pipeline error and merge error, then how to recover? for example can't merge due to gitlab lost changes in branch, then user force pushed again, we should consider this mr again.

Maybe during health check run, if the mr has merge error label but new notes after label note, remove the merge error label.

reconcile/gitlab_housekeeping.py
```diff
@@ -492,6 +518,7 @@ def preprocess_merge_requests(
             "priority": f"{label_priority} - {MERGE_LABELS_PRIORITY[label_priority]}",
             "approved_at": approved_at,
             "approved_by": approved_by,
+            "error": bool(ERROR_LABELS & labels),
```
hemslo: please also process this field in cli command https://github.com/app-sre/qontract-reconcile/blob/a7d039c3ecf1c86936182275d4de03fa10b7b480/tools/qontract_cli.py#L2300 so we can see it  in https://gitlab.cee.redhat.com/service/app-interface-output/-/blob/master/app-interface-merge-queue.md

reconcile/gitlab_housekeeping.py
rhur-pixel: For pipeline_error, `run_pipeline_healthcheck` has self-healing logic. For merge-error, I intended for that to be essentially "un-recoverable" (until IC manually fixes it), however, I think your approach is a lot better, will implement, thank you!

rhur-pixel:
Note: https://github.com/app-sre/qontract-schemas/pull/1010

Added optional merge_limit and consecutive_failure_limit integer fields to the gitlabHousekeeping configuration in both the JSON schema and GraphQL schema.

reconcile/gitlab_housekeeping.py
hemslo: Check new commits is not enough, example mr https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/187719#note_21413447, IC accidentally requested change, bot can't merge it unless approved or bypass. We should clear error label if any new notes posted after add label note, basically any new changes to mr clear the error label to recheck.

reconcile/gitlab_housekeeping.py
```diff
@@ -75,6 +75,8 @@
     NEEDS_REBASE,
 ]
 
+ERROR_LABELS = {MERGE_ERROR, PIPELINE_ERROR}
```
TGPSKI:
```suggestion
ERROR_LABELS = [MERGE_ERROR, PIPELINE_ERROR]
```

reconcile/gitlab_housekeeping.py
TGPSKI, line 976:
```suggestion
    default_limit = 8
    default_merge_limit = 8
    default_consecutive_failure_limit = 3
```

reconcile/gitlab_housekeeping.py
```diff
@@ -862,6 +989,7 @@ def run(dry_run: bool, wait_for_pipeline: bool) -> None:
         days_interval = hk.get("days_interval") or default_days_interval
         enable_closing = hk.get("enable_closing") or default_enable_closing
         limit = hk.get("limit") or default_limit
+        merge_limit = hk.get("merge_limit") or limit
```
TGPSKI:
```suggestion
        merge_limit = hk.get("merge_limit") or default_merge_limit
```

reconcile/gitlab_housekeeping.py
```diff
@@ -862,6 +989,7 @@ def run(dry_run: bool, wait_for_pipeline: bool) -> None:
...
         pipeline_timeout = hk.get("pipeline_timeout")
```
TGPSKI, line 1000:
```suggestion
        consecutive_failure_limit = hk.get("consecutive_failure_limit") or default_consecutive_failure_limit
        pipeline_timeout = hk.get("pipeline_timeout")
```

reconcile/gitlab_housekeeping.py
```diff
@@ -895,6 +1023,18 @@ def run(dry_run: bool, wait_for_pipeline: bool) -> None:
             project_merge_requests = [
                 mr for mr in opened_merge_requests if mr.state == MRState.OPENED
             ]
+            consecutive_failure_limit = hk.get("consecutive_failure_limit") or 3
```
TGPSKI:
```suggestion
```

reconcile/gitlab_housekeeping.py
```diff
@@ -209,6 +217,24 @@ def clean_pipelines(
...
+def check_pipeline_health(
+    pipelines: list[ProjectMergeRequestPipeline],
+    consecutive_failure_limit: int = 3,
+) -> bool:
+    """Return True if MR is healthy (should stay in queue).
+
+    Return False if last `consecutive_failure_limit` pipelines all
+    ended in a non-success terminal state (FAILED).
+    """
+    recent = pipelines[:consecutive_failure_limit]
```
TGPSKI:
```suggestion
    pipeline_failure_window = pipelines[:consecutive_failure_limit]
```

nitpick, but suggest renaming `recent` as it is not describing the variable contents properly

reconcile/gitlab_housekeeping.py
TGPSKI:
```suggestion
            "error": any(l in ERROR_LABELS for l in labels),
```
With L78 change (`ERROR_LABELS = [MERGE_ERROR, PIPELINE_ERROR]`), L521 breaks

reconcile/gitlab_housekeeping.py
```diff
@@ -752,9 +783,17 @@ def merge_merge_requests(
         must_pass=must_pass,
     )
     merge_requests_waiting.labels(gl.project.id).set(len(merge_requests))
+    merge_requests_error.labels(gl.project.id).set(
+        sum(1 for item in merge_requests if item["error"])
+    )
```
TGPSKI, line 792: Adding a note that setting gauges in a retry-decorated loop can be noisy + not optimized. Leaving this as a comment for future potential changes, not blocking this PR on this refactor.

reconcile/gitlab_housekeeping.py
```diff
@@ -895,6 +1023,18 @@ def run(dry_run: bool, wait_for_pipeline: bool) -> None:
             project_merge_requests = [
                 mr for mr in opened_merge_requests if mr.state == MRState.OPENED
             ]
+            consecutive_failure_limit = hk.get("consecutive_failure_limit") or 3
+            try:
+                run_error_healthcheck(
```
TGPSKI, line 1035: Eligibility checks (draft, merge_status, is_good_to_merge) duplicate `preprocess_merge_requests` API calls.

**Future optimization**: extract a shared `is_queue_candidate(mr)` predicate, and consider caching pipeline fetches between healthcheck and merge loop to reduce API calls.

Leaving this as a comment for future potential changes, not blocking this PR on this refactor.

TGPSKI (changes requested):

reconcile/gitlab_housekeeping.py
```diff
@@ -817,8 +856,96 @@ def merge_merge_requests(
...
+            if merge_error_added_at:
+                latest_notes = mr.notes.list(
+                    order_by="created_at", sort="desc", per_page=1
+                )
+                if any(
+                    from_utc_iso_format(note.created_at)
+                    > from_utc_iso_format(merge_error_added_at)
+                    and not (note.system and MERGE_ERROR in note.body)
+                    for note in latest_notes
+                ):
```
TGPSKI, line 941: Bot-authored notes ("Build success", "Change coverage report" from devtools-bot) are `system: false` and pass the current filter. Using MR 187719 as reference, the bot posted ~16 notes over 4 hours while the MR was stuck on a "Request Changes" gate. Under this code, each bot note would clear merge-error on the next healthcheck, the merge loop would retry and fail, and re-apply the label - bouncing every cycle instead of parking the MR.

Fix: exclude notes authored by the bot's own account.


```suggestion
                if any(
                    from_utc_iso_format(note.created_at)
                    > from_utc_iso_format(merge_error_added_at)
                    and not note.system
                    and note.author["username"] != gl.user.username
                    for note in latest_notes
                ):
```

reconcile/test/test_gitlab_housekeeping.py
```diff
@@ -972,3 +1019,749 @@ def test_get_rebase_strategy_toggle_enabled_valid_variant(
...
+        mr.notes.list.return_value = [system_note]
+
+        gl_h.run_error_healthcheck(
+            dry_run=False,
+            gl=gl,
+            project_merge_requests=[mr],
+            consecutive_failure_limit=3,
+        )
+
+        gl.remove_label.assert_not_called()
```
TGPSKI, line 1729:
```suggestion
        gl.remove_label.assert_not_called()

def test_bot_note_does_not_clear_merge_error(project: Project) -> None:
    """Bot-authored notes (Build success, etc.) should not trigger
    merge-error removal."""
    mr = _make_healthcheck_mr(labels=["lgtm", "merge-error"])
    label_event = create_autospec(ProjectMergeRequestResourceLabelEvent)
    label_event.action = "add"
    label_event.label = {"name": "merge-error"}
    label_event.created_at = "2025-06-01T12:00:00.0Z"

    bot_note = Mock()
    bot_note.created_at = "2025-06-01T13:00:00.0Z"
    bot_note.system = False
    bot_note.body = "Build success, build url: https://ci.example.com/123"
    bot_note.author = {"username": "devtools-bot"}
    mr.notes = create_autospec(ProjectMergeRequestNoteManager)
    mr.notes.list.return_value = [bot_note]

    mocked_gl = create_autospec(GitLabApi)
    mocked_gl.project = project
    mocked_gl.user = Mock(username="devtools-bot")
    mocked_gl.get_merge_request_label_events.return_value = [label_event]
    mocked_gl.get_merge_request_pipelines.return_value = _make_pipelines([
        "success", "success", "success",
    ])

    gl_h.run_error_healthcheck(
        dry_run=False,
        gl=mocked_gl,
        project_merge_requests=[mr],
        consecutive_failure_limit=3,
    )

    mocked_gl.remove_label.assert_not_called()
```

TGPSKI: cc https://github.com/app-sre/qontract-reconcile/pull/5537/changes#r3244469712

TGPSKI (changes requested):

reconcile/gitlab_housekeeping.py
TGPSKI: Continued:

- https://github.com/app-sre/qontract-reconcile/pull/5537#discussion_r3244469712
- and the associated test case: https://github.com/app-sre/qontract-reconcile/pull/5537#discussion_r3244473642

TGPSKI (approved):

reconcile/gitlab_housekeeping.py
hemslo: @rhur-pixel @TGPSKI we don't need this additional check, those bot notes are from pipeline run result, when we don't rebase / merge on error, bot won't post more notes, so we just need to clear the error label when the latest note is not `bot added error label`.

PR #5549, opened 2026-05-15
perf: deduplicate promtool runs in prometheus-rules-tester

jfreixa:
Many prometheus rules are identical across clusters (same template, same rendered output). Hash the rendered content and run promtool only once per unique rule, propagating the result to all duplicates.

Reduces promtool invocations by ~38% (932 unique out of 1504 rules), cutting the promtool phase from ~439s to ~356s wall-clock with 10 threads.


hemslo (approved):
LGTM!

PR #5552, opened 2026-05-18
APPSRE 13998 phase1 multi merge logic

rhur-pixel:
## TIckets
[APPSRE-13998](https://redhat.atlassian.net/browse/APPSRE-13998)
[APPSRE-13999](https://redhat.atlassian.net/browse/APPSRE-13999)
## Summary
After the first merge in a loop, merges additional MRs whose `tenant-*` labels don't overlap with already-merged MRs. Gated behind `active-cap-multi-merge` Unleash variant (default off). See [ADR-019](docs/adr/ADR-019-merge-queue-acceleration.md).
## What changed
**Label helpers:** `get_tenant_labels()`, `is_eligible_for_optimistic_merge()`, `has_overlapping_labels()` — pure functions, zero API calls. Labels already present from `gitlab_labeler`.
**Merge loop:**
- First MR: unchanged serial path (`is_rebased` + passing pipeline)
- After first merge: eligible non-overlapping MRs get `skip_ci` rebase then merge
- MRs without `tenant-*` labels or with overlapping labels: skipped
- `InsistOnPipelineError` only raised before first merge — prevents `@retry` from resetting multi-merge state
**Error handling:**
- `GitlabMRRebaseError`: warning + metric, no `merge-error` label, loop continues
- `GitlabMRClosedError`: `merge-error` label applied (recovered by `run_error_healthcheck`), loop continues
**Toggle:** `active-cap-multi-merge` variant on existing `gitlab-housekeeping-rebase-strategy` toggle. When off, breaks after first merge (same as previous `return`).
**Metrics:** `optimistic_merges_total`, `optimistic_merge_rejected_total` (reason: ineligible/overlap/rebase_failed/merge_rejected), `merge_batch_size` histogram
## Validation
- **Tests:** 109 passed — helper unit tests, 17 multi-merge integration tests, existing `test_merge_merge_requests` extended with `rebase=True`
- **Linter/types:** ruff + mypy pass
- **Local dry-run:** `qd profile run dev gitlab-housekeeping` completes without exceptions
## Follow-up TODOs
- Set Unleash variant to `active-cap-multi-merge` after deploy, monitor metrics
- Add `merge_limit` to gitlabHousekeeping schema (currently falls back to default 8)
- Monitor `optimistic_merge_rejected_total{reason="overlap"}` — if >30%, evaluate Phase 2

## Summary by CodeRabbit

* **New Features**
  * Optimistic multi-merge mode: tenant-scoped lead/pending grouping, windowed batch merges, and automatic group formation/clearing
  * CI-aware behavior: cancel timed-out pipelines, eject failed members, and merge only rebased/passing MRs
  * New monitoring: optimistic merge counts, rejection reasons, and merge batch-size histogram

* **Bug Fixes / Behavior**
  * Merge-error labeling on failures; tightened merge-limit and pipeline retry/insist semantics; orphaned group labels are cleared

* **Tests**
  * Expanded unit and integration tests for multi-merge, tenant-label rules, pipeline gating, and edge cases


reconcile/gitlab_housekeeping.py
```diff
@@ -605,6 +641,41 @@ def _try_rebase(
         return False
 
 
+REBASE_POLL_TIMEOUT = 30
+REBASE_POLL_INTERVAL = 1.0
+
+
+def _wait_for_rebase(
```
TGPSKI: **Revised design: async state-driven approach**

Rather than blocking, the revised design:

- Forms a `merge_group` after the first serial merge succeeds
- Persists group membership + status to S3 state
- Each reconcile loop iteration checks rebase status and merges ready members
- Merge group closes after `max_interval` (default 5m, Unleash-configurable) from first merge time
- Zero blocking sleeps. Each loop is a single pass: read state, act, write state, exit.

This eliminates the 30s blocking poll, makes the design multi-repo safe, and lets the HK pod's instant-refire cadence drive progress naturally. 

[Full design proposal](https://github.com/app-sre/qontract-reconcile/pull/5552#issuecomment-4483027451)

TGPSKI (changes requested):

TGPSKI:
## Multi-Merge Phase 1 Proposal: Labels as state with group lifecycle

**Updated** based on @hemslo's feedback (labels instead of S3) and production log replay analysis.

Replace [synchronous rebase-polling](https://github.com/app-sre/qontract-reconcile/pull/5552/changes#diff-36fc33e36c6c75f4740d29326cae1f20567af41c6deb6f17b0a908aaeb83d520R648-R675) with an async label-driven merge group. After the first serial merge, form a group of non-overlapping MRs, kick their rebases, and let successive reconcile loop iterations drive them to merge.

---

### Why the time window is the throughput mechanism

Production log replay (696 MRs, May 6-13, active-cap period) proves that without a multi-iteration window, OMM yields zero throughput gain:

**Inter-merge gap distribution** (695 consecutive pairs):
- Median gap: **6.0m**
- Only **1%** of pairs arrive within 2 minutes of each other
- **40%** arrive within 5 minutes
- **75%** arrive within 10 minutes

**Batch formation at different windows** (service-level non-overlap):

| Window | MRs that could batch | Avg batch size | Max batch |
|--------|---------------------|----------------|-----------|
| ≤2m | **4** | 2.0 | 2 |
| ≤5m | **213** | 2.0 | 3 |
| ≤10m | **394** | 2.3 | 4 |

**Throughput impact**:

| Window | Avg MRs/batch | Merges/hr | vs baseline |
|--------|---------------|-----------|-------------|
| ≤2m | 1.0 | 4.2 | 1.0x (no gain) |
| ≤5m | 1.5 | 6.5 | 1.5x |
| ≤10m | 2.7 | 11.4 | 2.7x |

---

## Implementation

- Remove `_wait_for_rebase()`, `REBASE_POLL_*` constants
- Add label management helpers (`apply_omm_labels`, `clear_omm_group`, `get_omm_pending_mrs`)
- Restructure `merge_merge_requests()` into two paths:
  - **No active group**: serial merge + group formation + kick rebases + apply labels
  - **Active group**: process members (check rebase, merge ready, eject failed, close on expiry)
- `max_interval` as Unleash config (default 5m)
- Adaptive close: exit group early when no candidates have active CI

### State

- `omm-group-lead` on the first merged MR → provides `merged_at` as window start
- `omm-pending` on group member candidates

Group lead's `merged_at` + `max_interval` (default 5m, Unleash-configurable) = window expiry.

### Adaptive Close

Group closes when:
1. No non-overlapping candidates with active CI in the queue, OR
2. Window > `max_interval`

### Cancellation on CI Failure

If post-merge CI fails during the window → remove all `omm-pending` labels immediately → back to serial. Bounds blast radius regardless of window duration.

### Deferred to Phase 2

- speculative pre-rebase: When window is >80% elapsed or partially failed, speculatively rebase 1-2 highest-priority non-overlapping MRs from outside the group.
- max_interval increase towards 10 minutes

### Execution Model

```mermaid
flowchart LR
    A[Loop fires] --> B{Group in state?}
    B -->|No| C[Serial merge 1 MR]
    C -->|OK| D[Form group + trigger rebases]
    D --> E[Write state, exit]
    C -->|Fail| F[Exit]
    B -->|Yes| G{Window open?}
    G -->|No| H[Close group, exit]
    G -->|Yes| I[Process members]
    I --> J{All done?}
    J -->|Yes| H
    J -->|No| E
```

reconcile/gitlab_housekeeping.py
hemslo: we don't need to persist state in s3, can just add another label to mr that rebase skip pipeline, then in each reconcile loop, treat that kind of mr as ready to be merged ones and merge following existing logic

TGPSKI: I agree with using labels instead S3 for state. Good call.

There are two models to consider: `skip` labels vs `merge group` labels. The skip labels  model has gaps.

**Time window**
  - Only 1% of non-overlapping pairs arrive within 2 minutes.
  - If we label and merge on the next loop of housekeeping, production data shows zero throughput gain with multi-merge. 
  - Multi-merge only adds throughput if the merge group stays open for 5-10 minutes.

**Label lifecycle**
- skip-labels only don't have expiration, cancellation, or stale state cleanup
- a time-bounded window and cancellation logic will prevent labeled MRs from infinite retry

hemslo: we can recalculate skip label on every run, if any mr with skip label has overlap tenant labels with another mr, remove skip label to allow it rebase and retest again.

TGPSKI: It's frustrating to engage with your reviews. I put in time and effort to understand and engage with your ideas. It's demoralizing to get single sentence answers in return, especially when your responses completely ignore the substance of my ideas.

hemslo: It's easier to forward my review comments to ai agents directly, if you don't want my review, I can ignore all future changes related to this ticket.

rhur-pixel: To summarize where we landed: the current implementation uses the merge-group model with time-bounded labels (omm-group-lead, omm-pending) and automatic cleanup. This addresses the lifecycle concerns (expiration via max_interval, cancellation on pipeline failure, stale label cleanup on group close).

@hemslo — I think the recalculation idea (remove skip label when overlap changes) is interesting and also way simpler, but without the time-window, its rare for MR's to be both overlapping and merge-ready at the same time (~%1). So I don't believe merge-throughput would increase.

I'm thankful for the reviews so far and hopefully we can continue the process, and now that the implementation is more explicit I'm also happy for any criticisms to be brought up, including previous ones. Maybe I misunderstood the skip-label implementation, but either way hopefully we can come to an agreement and get this merged 💪 Please let me know anything I can do/improve.

hemslo: The bottleneck is the rebase+CI cycle, not the merge itself. After one MR merges, all others need rebasing — and each rebase require gitlab server processing. We can skip that CI for the right candidate.

**Proposed algorithm (one candidate at a time, like SQS FIFO fetch=1):**

Each reconcile loop:
1. Merge the first rebased+passed MR (serial path, unchanged).
2. After that merge: gather all open candidates with passing pipelines + last N recently merged MRs (one `mergerequests.list(state=merged, per_page=20)` call — one page is enough).
3. Find the **highest-priority candidate** whose `tenant-*` labels don't overlap with any other open candidate or any recently merged MR. This is the one safe to rebase without CI.
4. Label it `skip-ci`, call `mr.rebase(skip_ci=True)`. Don't touch anything else.
5. Next loop: if that MR's rebase is done, **re-check overlap against last N merged MRs** before merging. If an overlapping MR was merged in between (race condition), remove the `skip-ci` label and let it go through normal rebase+CI. If still clean, merge it, then go back to step 2.

The key point is the overlap check must happen at merge time too, not just at labeling time, to catch the race between rebase completing and the next loop running.

**Why simpler:**
- 1 label (`skip-ci`) instead of 2 (`omm-group-lead` + `omm-pending`)
- No group lifecycle, no time windows, no cleanup paths — recalculation on every run is the cleanup
- Self-healing: if a skip-ci MR later overlaps, the label is removed automatically
- ~30 lines instead of ~190 lines of group management

hemslo: Regarding merge-throughput, each integration run at most merge 1 mr, the current problem is during one integration run, it can't find ready to be merged mr due to waiting for CI run (new mr or rebased), by skip CI we increase the chance of mr ready during integration run, then merge-throughput will increase over time.

TGPSKI: The throughput impact table [in my May 18 comment](https://github.com/app-sre/qontract-reconcile/pull/5552#:~:text=4-,Throughput%20impact%3A,-Window) already shows that: ≤2m window (single-candidate per loop) = 1.0x, zero throughput gain. The group model with a time window is the speed-up mechanism.

reconcile/gitlab_housekeeping.py
```diff
@@ -759,6 +1015,125 @@ def _rebase_merge_requests_old_burst(
...
+    lead: ProjectMergeRequest,
+    app_sre_usernames: AbstractSet[str],
+    pipeline_timeout: int | None = None,
+) -> int:
+    """Process an active OMM group. Returns number of merges performed.
+
+    Single-pass: check window, verify post-merge CI, process each pending
+    member (merge if ready, eject if failed). Non-blocking.
+    """
+
```
TGPSKI, line 1054:
```suggestion
# Invalidate group if target branch moved unexpectedly
    current_head = gl.project.branches.get(lead.target_branch).commit["id"]
    if current_head != lead.merge_commit_sha:
        logging.warning(["omm-group", "head-moved", gl.project.name, "invalidating"])
        clear_omm_group(dry_run, gl, lead=lead)
        return 0
```

TGPSKI: Fail fast process OMM group if force merge happened while group is open

reconcile/gitlab_housekeeping.py
```diff
@@ -791,6 +1166,40 @@ def merge_merge_requests(
         sum(1 for item in merge_requests if item["error"])
     )
 
+    # --- OMM group cleanup when feature is disabled ---
+    # If the Unleash variant was changed away from active-cap-multi-merge
+    # while a group was active, clean up orphaned labels.
+    if not multi_merge and rebase:
```
TGPSKI, line 1255: I was going to suggest this cleanup, now I see its already here, 👍

reconcile/gitlab_housekeeping.py
```diff
@@ -759,6 +1015,125 @@ def _rebase_merge_requests_old_burst(
...
+            except gitlab.exceptions.GitlabMRClosedError as e:
+                logging.error(f"unable to merge {mr.iid}: {e}")
+                gl.add_label_to_merge_request(mr, MERGE_ERROR)
+                gl.remove_label(mr, OMM_PENDING)
+                optimistic_merge_rejected.labels(
+                    project_id=mr.target_project_id, reason="merge_rejected"
+                ).inc()
+                continue
+        merges += 1
+
```
TGPSKI, line 1212: This is a suggestion for future reference, not intended for implementation in this PR.

**Dynamic Group Expansion (Phase 2)**

The current implementation forms the OMM group once (at first merge) and never adds to it. The group is a static snapshot of whatever's in the queue at formation time. Dynamic expansion means: on each subsequent loop iteration, check for new non-overlapping candidates that have arrived since formation, add them to the group, kick their CI, and merge them when they pass.

Static vs Dynamic Throughput

| Model | What gets batched | Effective gain |
|-------|------------------|----------------|
| Static group (current PR) | Only MRs already queued at formation time | ~1.0-1.1x (queue depth at any moment is low because arrival ≈ merge rate) |
| Dynamic group (Phase 2) | MRs arriving during the window get added | 1.33x at 5m window, 1.7x at 10m |

- Phase 1 (this PR):  Static group. Proves machinery. ~1.0-1.1x.
- Phase 2 (next PR):  Dynamic expansion + head-check. ~1.33x at 5m.
- Phase 3 (future):   Window increase to 10m + speculative pre-rebase. ~1.7x.

```suggestion
# --- Dynamic expansion: add new arrivals to the group ---
current_queue = gl.get_merge_requests(state=MRState.OPENED)
queue_items = preprocess_merge_requests(...)  # or pass in from caller
pending_iids = {mr.iid for mr in pending}

for merge_request in queue_items:
    mr = merge_request["mr"]
    if mr.iid in pending_iids:
        continue
    if merge_request["error"]:
        continue
    if not is_eligible_for_optimistic_merge(mr):
        continue
    mr_labels = get_tenant_labels(mr)
    if has_overlapping_labels(mr_labels, group_labels):
        continue
    # New candidate — add to group, kick rebase with CI
    logging.info(["omm-group", "dynamic-add", gl.project.name, mr.iid])
    if not dry_run:
        gl.add_label_to_merge_request(mr, OMM_PENDING)
        try:
            mr.rebase(skip_ci=False)  # always run CI
        except gitlab.exceptions.GitlabMRRebaseError:
            gl.remove_label(mr, OMM_PENDING)
            optimistic_merge_rejected.labels(
                project_id=mr.target_project_id, reason="rebase_failed"
            ).inc()
    group_labels.update(mr_labels)
```

TGPSKI:
`get_tenant_labels`, `is_eligible_for_optimistic_merge`, `has_overlapping_labels`

Need docstrings

TGPSKI:
## `merge_limit` bypass in OMM group path

The per-repo `merge_limit` (configured in app-interface, default 8) is threaded into `merge_merge_requests` but never passed to `_process_omm_group`. When a group is active, the early return skips the serial path entirely, so the configured limit is never enforced:

```python
if multi_merge and rebase:
    lead = get_omm_group_lead(gl)
    if lead:
        merges = _process_omm_group(...)  # no merge_limit
        merge_batch_size_histogram.labels(project_id=gl.project.id).observe(merges)
        return
```

`_process_omm_group` iterates all pending MRs and merges every one with a green pipeline in a single pass — no cap. This bypasses a per-repo safety contract that operators explicitly configure.

It also means `_check_post_merge_ci` only runs once at the top of the function, not between individual merges. If the 3rd merge in a batch breaks the target branch, merges 4–N still fire before the next loop gets a chance to detect the failure.

### Fix: thread `merge_limit` and reset group on limit

```python
def _process_omm_group(
    dry_run: bool,
    gl: GitLabApi,
    lead: ProjectMergeRequest,
    app_sre_usernames: AbstractSet[str],
    pipeline_timeout: int | None = None,
    merge_limit: int = 8,
) -> int:
    ...
    for mr in pending:
        if merges >= merge_limit:
            logging.info(["omm-group", "merge-limit-reached", gl.project.name, merge_limit])
            clear_omm_group(dry_run, gl, lead=lead, pending=pending)
            return merges
        ...
        merges += 1
```

Resetting the group on limit-hit is preferable to keeping it open because:

1. After `merge_limit` merges land, the target branch HEAD has moved N times. Remaining pending MRs were rebased against the pre-group HEAD — they're stale.
2. Resetting forces fresh re-evaluation: next loop does a serial merge, fresh rebase, fresh group formation with valid candidates.
3. `_check_post_merge_ci` runs at the top of the next loop, catching any breakage from this batch before more merges proceed.
4. Simpler mental model: group = bounded batch, done. Batch size is `min(merge_limit, window_expiry, pending_count)`.

TGPSKI (changes requested):
Fixes needed: 

- [merge_limit bypass](https://github.com/app-sre/qontract-reconcile/pull/5552#issuecomment-4579532110)
- [invalidate group if target sha drift (force merge, etc)](https://github.com/app-sre/qontract-reconcile/pull/5552#discussion_r3326558280)
- [docstrings](https://github.com/app-sre/qontract-reconcile/pull/5552#issuecomment-4579435304)

TGPSKI (approved):

reconcile/gitlab_housekeeping.py
hemslo: You can not exceed single candidate per loop to merge, that's the upper cap due to liner history config in gitlab and async rebase, the best case is always have 1 mr merged in a run when merge queue is not empty, the group model is too complex and no benefits.

TGPSKI: Agreed — at most one MR merges per reconcile loop. OMM does not change that.

The throughput gain comes from later reconcile loops - not from same loop multi-merge. After the lead MR merges, the remaining independent candidates are rebased with `skip_ci`.  Each additional loop then finds an already-ready MR instead of stalling on a fresh CI cycle. The group + time window makes `skip_ci` merges repeatable across loops. OMM is not a way to merge multiple MRs in one loop.

**Without the time window, `skip_ci` alone yields zero gain.** 

The production replay showed this directly:

| Window | Merges/hr | vs baseline |
|--------|-----------|-------------|
| ≤2m (single-candidate, next-loop only) | 4.2 | 1.0x |
| ≤5m | 6.5 | 1.5x |
| ≤10m | 11.4 | 2.7x |

The ≤2m row is the "no group, just `skip_ci` one candidate" model.

It produces no measurable gain because **99% of non-overlapping pairs arrive more than 2 minutes apart**. The group model with a 5-10m window is what converts `skip_ci` from a local latency optimization into sustained queue throughput.

On complexity: A simpler version without the window produces no measurable improvement. Throughput gain depends on group management. The labels, lifecycle, and expiry exist to bound the blast radius and handle stale state cleanly.  The MR has 17 integration tests covering the edge cases.

hemslo: my claude code analysis shows window is irrelevant when use stateless recalculate all mrs (ready to merge + last N merged) in one run, maybe easier if your agent directly talk to my claude code.

PR #5560, opened 2026-05-19
fix(pagerduty): add missing JWT authentication to PagerDuty endpoints

chassing:
## Summary
- Add missing `UserDep` JWT authentication dependency to both PagerDuty external endpoints (`get_schedule_users` and `get_escalation_policy_users`)
- All other external endpoints (Slack, LDAP, VCS) already required JWT auth via `UserDep` — PagerDuty was the only one missing it

## Test plan
- [ ] Verify PagerDuty endpoints now return 401 without a valid JWT token
- [ ] Verify PagerDuty endpoints still work correctly with a valid JWT token
- [ ] Run `cd qontract_api && make test` to ensure no test regressions

## Summary by CodeRabbit

* **Chores**
  * PagerDuty schedule users endpoint now requires bearer-token (HTTP Bearer) authentication.
  * PagerDuty escalation policy users endpoint now requires bearer-token (HTTP Bearer) authentication.
  * Generated client libraries for these PagerDuty endpoints now require an authenticated client to call the APIs.
  * Updated pinned base images in Dockerfiles for the API, client, and utils builds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/app-sre/qontract-reconcile/pull/5560?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/qontract_api/Dockerfile
+++ b/qontract_api/Dockerfile
@@ -1,6 +1,6 @@
 #
 # Base image with defaults for all stages
-FROM registry.access.redhat.com/ubi9-minimal@sha256:fe9e574f04371b333ed4e21d30d984f6b7fcd1046e579f5ddab4816c0c8e231d AS base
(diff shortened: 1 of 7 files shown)


hemslo (approved):
LGTM!

PR #5559, opened 2026-05-19
APPSRE 13230 hold mr owner re-land

rhur-pixel:
Re-land of #5543 (reverted in #5555).

## Summary by CodeRabbit

* **New Features**
  * Merge request authors can now apply hold-type approvals to changes, even when not designated as change owners. Special rules govern hold cancellation operations based on user roles.

* **Tests**
  * Added comprehensive test coverage for merge request author hold approval behavior and interactions with approver holds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/app-sre/qontract-reconcile/pull/5559?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #5543, opened 2026-05-12
APPSRE 13230 hold mr owner

rhur-pixel:
[APPSRE-13230](https://redhat.atlassian.net/browse/APPSRE-13230)

## Summary
The `change-owners` integration ignores `/hold` comments from MR authors who are not part of any change owner group. This means an author cannot hold their own MR from being merged if they are not a listed approver.
This change allows MR authors to `/hold` their own MR even when they are not in any approver group. Only `/hold` is honored from the author — `/hold cancel` from the author is ignored so that only a change owner can release the hold.
When the author is already a change owner, the normal approver path is used (no behavior change).
## Changes
- `reconcile/change_owners/decision.py`: Added `mr_author` parameter to `apply_decisions_to_changes` and `_apply_decision_to_diff`. After processing normal approver decisions, if the author is not in any approver group, their `/hold` is applied via a dedicated `"mr-author"` context. Change owners can `/hold cancel` to release it.
- `reconcile/change_owners/change_owners.py`: Wired `mr_author` through to `apply_decisions_to_changes`. Deduplicated `gl.get_merge_request_author_username()` call into a local variable.
- `reconcile/test/change_owners/test_change_type_decision.py`: Added 6 tests covering author hold, author self-unhold rejected, change owner unhold, hold+approval coexistence, author-as-approver using normal path, and random user hold ignored.

## Validation
pytest
ruff lint: all checks passed
ruff format: clean
mypy: no issues

## Fixes
APPSRE-13230

## Summary by CodeRabbit

* **New Features**
  * Merge request authors can now place holds on changes with restricted cancellation permissions and special handling in approval workflows.

* **Tests**
  * Added comprehensive test coverage for merge request author hold behavior across multiple scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/app-sre/qontract-reconcile/pull/5543?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #5568, opened 2026-05-26
fix(change-owners): apply bot/hold label on non-self-serviceable MRs

rhur-pixel:
Follow up to [APPSRE-13230](https://redhat.atlassian.net/browse/APPSRE-13230)
PR (for context, initial PR was reverted since I forgot to promote in app-interface, re-landed now) https://github.com/app-sre/qontract-reconcile/pull/5543

## Summary

I noticed [this MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/188913) recently and figured that my code changes earlier was the cause of this. 

Fixes two bugs in the `/hold` feature introduced in #5559:

- **`bot/hold` label never applied on non-self-serviceable MRs** — the labeling gate was `self_serviceable and hold`, so non-SS MRs could never get the label even when held. Changed to just `hold`.
- **Stuck holds on mixed MRs** — when the MR author was an approver on some diffs but not others, the author-hold path would create `"mr-author"` context holds on the non-owned diffs. These could never be cancelled because the author's `/hold cancel` is intentionally ignored and change owners only cancel holds on their own contexts.

## Fix

Introduced an MR-wide `author_is_approver_anywhere` flag that checks whether the author is an approver on *any* diff across the entire MR. If they are, the special author-hold path is skipped entirely — the normal approver path handles `/hold` and `/hold cancel` cleanly. The author-hold path only activates for authors with zero approver authority on the MR.

Also removed a redundant per-diff `is_approver` check that became dead code after adding the MR-wide flag.

## Behavior matrix

| | Author is approver | Author is NOT approver |
|---|---|---|
| **SS** | `/hold` + `/hold cancel` via approver path. Blocks auto-merge. | `/hold` via author path. Author `/hold cancel` ignored — only owner can release. |
| **Non-SS** | `/hold` + `/hold cancel` via approver path. Label is informational. | `/hold` via author path. Author `/hold cancel` ignored. Label is informational. |

Box #4 is what I am most unsure of: does this technically break the philosophy of non-self-serviceable? 

Open question: Should author /hold be restricted to self-serviceable MRs only (given mr author is not an approved owner)? On non-SS MRs, the label is purely informational since auto-merge is already blocked. Keeping current behavior (apply /hold on both SS and non-SS) for consistency in this bugfix — happy to discuss as a follow-up.

## Test plan

- [x] Existing author-hold tests pass (hold as non-approver, cancel ignored, coexists with approval, approver uses normal context, owner can unhold)
- [x] New test: mixed MR where author is approver on diff 0 but not diff 1 — verifies no `"mr-author"` hold is created on diff 1, preventing stuck state
- [x] Full test suite: 21/21 pass

Fixes: APPSRE-13230

## Summary by CodeRabbit

* **Bug Fixes**
  * HOLD label now applies whenever any change decision is held.
  * MR-author hold/cancel commands are now honored more broadly, including cases where the author isn’t an approver on some diffs; hold behavior per-diff is preserved.

* **Tests**
  * Updated hold-cancel test and added a test covering mixed diffs where the MR author is approver on one diff but not another.

[Review Change Stack](https://app.coderabbit.ai/change-stack/app-sre/qontract-reconcile/pull/5568?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


reconcile/change_owners/decision.py
```diff
@@ -145,13 +145,21 @@ def apply_decisions_to_changes(
     to generate the coverage report and to reason about the approval
     state of the MR.
     """
+    author_is_approver_anywhere = bool(mr_author) and any(
+        ctx.includes_approver(mr_author)
+        for c in changes
+        for d in c.diff_coverage
+        for ctx in d.coverage
+        if not ctx.disabled
```
hemslo: This MR-wide flag works but adds complexity: a new parameter threaded through recursive calls, and reasoning about cross-diff state in a per-diff function.

**Simpler alternative:** instead of preventing `"mr-author"` holds from being created, let the author cancel their own hold. The stuck state only exists because `/hold cancel` from the author is intentionally blocked. If you add an `is_author_cancel` case:

```python
is_author_cancel = (
    decision.approver_name == mr_author
    and decision.command == DecisionCommand.CANCEL_HOLD
)
# ...
if is_author_hold or is_author_cancel or is_owner_cancel:
    change_decision.apply_context_decision("mr-author", decision.command)
```

No MR-wide flag, no new parameter, no threading through fragment recursion. The tradeoff is the author can now unilaterally cancel their own hold — but since they're the one who placed it, that seems reasonable.

Non-blocking — happy to discuss whether the `author can't self-cancel` safety is worth keeping.

hemslo (commented):
Overall the bugfixes here are correct — both the label gate fix and the stuck-hold prevention. CI passes, test coverage is good.

reconcile/change_owners/decision.py
rhur-pixel: Yeah really good point, I was also considering this but ultimately decided on staying with the requirements of the ticket itself, but when you put it this way it seems a little harder to justify. 

@fishi0x01 What are your thoughts on letting the MR owner be able to /hold cancel their own hold? Is there a specific scenario you had in mind where we would want to prevent that? A change owner’s hold would still be unaffected either way since they both operate on separate contexts. 

EDIT: [Ticket for reference](https://redhat.atlassian.net/browse/APPSRE-13230)

fishi0x01: The initial idea was that only changetype owners can unhold, as they are the resource owners of the change. However, Im also fine with MR owner unholding own MR - they cannot merge it without changeowner lgtm, so it would be fine 👍

hemslo (approved):
LGTM!

PR #5581, opened 2026-06-02
ADR-021: Vault secret deletion guardrails

TGPSKI:
## Summary

- Proposes Vault RBAC policy changes to remove `delete`/`destroy` capabilities from standard team policies on KV v2 mounts, preventing secret deletions that bypass the gitops workflow
- Routes all secret deletions through app-interface (remove reference first, then delete via `vault-delete-secret` integration or privileged role)
- Also adds ADR-020 (httpx → httpxyz) to the README index

## Context

On 2026-06-02, a Vault KV v2 secret was deleted while still referenced in app-interface, blocking CI for ~75 minutes and requiring a force-merge to unblock promotions. This ADR proposes guardrails to prevent recurrence by enforcing the correct order of operations: remove the consumer before removing the resource.

## Test plan

- [ ] Review ADR content for accuracy and completeness
- [ ] Validate proposed Vault HCL policy syntax
- [ ] Confirm ADR-021 numbering doesn't conflict with any in-flight ADRs

## Summary by CodeRabbit

* **Documentation**
  * Added architectural decision record outlining controls and guardrails for Vault secret deletion processes
  * Documented recommended remediation procedures and implementation guidance
  * Updated ADR index with new entries

### DIFF
--- a/docs/adr/ADR-021-vault-secret-deletion-guardrails.md
+++ b/docs/adr/ADR-021-vault-secret-deletion-guardrails.md
@@ -1,0 +1,225 @@
+# ADR-021: Vault Secret Deletion Guardrails
+
+**Status:** Proposed
+**Date:** 2026-06-02
+**Authors:** Tyler Pate
+
+## Context
+
+On 2026-06-02, a Vault KV v2 secret (`hcc/jira-ai-categorizer`, version 2) was deleted from Vault while still referenced by a namespace configuration in app-interface. This caused the `openshift-vault-secrets` integration to fail with:
+
+```
+[hcmaii01ue1/jira-ai-categorizer] error fetching secret: version '2' not found for secret with path 'hcc/jira-ai-categorizer'
+```
+
+The integration correctly treated this as an error — Vault returned HTTP 404 for the destroyed version, `VaultClient.__read_all_v2` raised `SecretVersionNotFoundError`, and the integration exited non-zero. This is the expected behavior when a declared dependency is missing. The existing safety mechanism of disabling cluster deletions on error (`options.enable_deletion = not ri.has_error_registered()`) prevented any damage to running workloads — the existing K8s Secrets remained untouched.
+
+However, the CI failure blocked all subsequent qontract-reconcile promotions that triggered the `openshift-vault-secrets` dry-run, requiring a force-merge of MR !190766 to unblock.
+
+The root cause is that Vault is an external mutable store with no awareness of app-interface's declarative references. Nothing prevents a user from deleting a secret version that is actively depended upon — it is a "rug pull" on a declared dependency in a gitops repository.
+
+### Current behavior
+
+- **KV v2 versioned secrets are treated as immutable** by convention, but Vault RBAC does not enforce this. Any user with `delete` or `destroy` capability on a path can remove versions.
+- **`VaultClient.delete()` already refuses v2 deletions** (raises `ValueError("deleting V2 secrets is not supported yet")`), demonstrating that the codebase already treats v2 versions as immutable. But this only governs qontract-reconcile's own write path, not external Vault users.
+- **Deletions bypass the gitops workflow entirely.** App-interface enforces review and CI checks for all configuration changes, but Vault deletions happen outside this process — there is no MR, no review, and no validation that the secret is still in use.
+
+### Requirements
+
+- Vault secret deletions must go through the gitops workflow: remove the app-interface reference first, then delete from Vault
+- Maintain the ability for teams to migrate secrets between Vault engines (with proper coordination)
+- Keep the solution simple — avoid background integrations when the problem can be prevented at the source
+
+## Decision
+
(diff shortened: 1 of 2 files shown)


hemslo (commented):
with https://github.com/openshift-online/platform-engineering-enhancements/pull/44, secrets in vault and usages in cluster will be fully managed by tenant, not in app-interface anymore.

TGPSKI:
@hemslo Thanks for flagging the ESO design doc — I agree that's the right long-term direction. However, I don't think it supersedes this ADR. A few points:

1. **ESO has the same failure mode.** The ESO design doc itself lists "Secret deleted from Vault" as a failure mode (Section 7), with the mitigation being alerts on sync failures. Moving from `VaultSecret` to `ExternalSecret` shifts the blast radius from "CI pipeline blocked" to "Argo CD degraded + stale/missing secrets on next pod restart" — arguably worse in production.

2. **ESO is months away from full rollout.** Phase 1 hasn't finished validation yet, and the Phase 3 migration is estimated at 6-8 weeks *after* Phases 1 and 2. The current `VaultSecret` model is the production reality today, and this incident can recur tomorrow.

3. **The RBAC fix is complementary, not competing.** The ESO design already scopes Vault policies to read-only for applications and enforces tenant path isolation. ADR-021 formalizes that same principle for the current system. Stripping `delete`/`destroy` from standard policies is a cheap, simple control that protects *any* Vault-backed secret management architecture — current or future.

4. **The problem is human, not architectural.** A platform SRE deleted a secret thinking it was low-impact. Changing the operator that reads secrets doesn't change that behavior. The guardrail (RBAC restriction) addresses the root cause regardless of which secret sync mechanism is in use.

I'd suggest we proceed with the RBAC changes now and add a note to the ADR acknowledging the ESO migration timeline. The two efforts are complementary — ESO reduces the surface area of Vault references in app-interface, and ADR-021 ensures that whatever references *do* exist (now or in the future) can't be rug-pulled.

lisa:
I think the idea is really good. But I'm concerned with how tenants will interact with their secrets. Specifically, this part:

> Routes all secret deletions through app-interface (remove reference first, then delete via vault-delete-secret integration or privileged role)

What does the tenant-facing workflow look like here? I'm guessing there would be a new schema to support this operation?

Additionally, is there a way to unify KV1 into the same worfklow as KV2?

hemslo (changes requested):
## Review: Architectural Concern — Wrong Problem Framing

This ADR treats the incident as an RBAC/process problem ("prevent tenants from deleting secrets"), but the root cause is **platform resilience**: the `openshift-vault-secrets` integration fails hard when a referenced secret is missing, and that hard failure is what blocked CI for 75 minutes. These are actually two separate problems.

---

### Problem 1: Platform Stability *(the fix that should happen now)*

The platform should be resilient to tenant actions on tenant-owned resources. A missing Vault secret version should not block all qontract-reconcile promotions for the whole platform. The fix belongs in code, not RBAC:

- **Handle missing secrets gracefully** — rather than hard-failing and blocking the reconciler, catch the `SecretVersionNotFoundError`, log a warning, expose a metric (e.g. `vault_secret_version_missing`), and continue reconciling other resources in the loop.
- **Add alerting** — alert on that metric so the team owning the secret is notified to take action.
- For reference: ArgoCD/ESO avoids this problem because the ESO CR itself can be deployed regardless of Vault state — the CR spec is valid even when the backend secret is missing. The Vault secret status is tracked per-resource, not as a global blocker.

This approach directly addresses what actually went wrong: the blast radius was too large, not that the deletion happened.

---

### Problem 2: Tenant RBAC *(separate concern, handle separately)*

Tenants should have full self-service on their own Vault secrets. App-interface should not be a mandatory blocker path for tenant secret lifecycle management. Some concerns with the proposed RBAC policy:

- Tenants may manage secrets entirely outside app-interface today (e.g., Terraform-based secret-as-code). Removing `delete`/`destroy` breaks those workflows.
- Forcing all deletions through app-interface MRs adds friction to legitimate self-service operations.
- Offering a `vault-delete-secret` integration as an **optional** gitops path for teams that want it is a reasonable addition, but it should not be the only path.

If the goal is to prevent cross-team accidents (tenant A deleting a secret used by tenant B), that's a namespace/path isolation problem, not a capability restriction problem — and should be addressed at the app-interface schema level rather than as a blanket RBAC restriction.

---

### Suggested rework

1. **Separate the two problems** into distinct sections (or distinct ADRs).
2. **Prioritize the code fix** — make `openshift-vault-secrets` resilient to missing secrets, with metrics and alerting. This directly solves the incident.
3. **Keep tenant self-service intact** — don't remove `delete`/`destroy` from standard team policies. Tenants should own their secrets fully.
4. **Optionally offer** a delete-as-a-service gitops path for teams that want auditability through app-interface — not as a mandate.

The code fix (graceful error handling + metric + alert) solves the actual production impact of this incident without restricting tenant autonomy or making app-interface a bottleneck for secret lifecycle management.

PR #5582, opened 2026-06-03
Replace openapi-python-client with clientele for qontract-api-client

chassing:
> [!IMPORTANT]
> **Requires upstream PR:** https://github.com/phalt/clientele/pull/247

## Summary

- Replace `openapi-python-client` with [`clientele`](https://github.com/phalt/clientele) for generating `qontract_api_client`. The generated client is **dramatically simpler**: ~1,500 lines of clean Pydantic models + async functions vs ~13,500 lines of attrs-based boilerplate across 110+ files.
- **No more `attrs` dependency** — clientele generates standard **Pydantic** models (`pydantic.BaseModel`), aligning the generated client with the rest of the codebase. No more `to_dict()`/`from_dict()` conversion, no more `UNSET` sentinel values, no more `types.py` boilerplate.
- The generated client produces simple decorated async functions (`@client.post(...)`, `@client.get(...)`) instead of deeply nested module hierarchies with sync/async variants and response parsing wrappers.
- Update all imports and call sites across `reconcile/`, `qontract_api/`, and tests.
- Add unit tests for `subscriber/_client.py`.
- Lazy `_setup_client()` initialization + `_build_chat_request` helper to DRY the `ChatRequest`/`Secret` construction.

### Before (openapi-python-client)

```
qontract_api_client/
├── api/external/          # 7 endpoint modules, each ~250 lines
├── api/health/            # 2 modules
├── api/integrations/      # 9 modules
├── models/                # 60+ model files, each 60-200 lines (attrs-based)
├── client.py              # AuthenticatedClient class
├── errors.py
└── types.py               # UNSET sentinel, File wrapper
→ 110+ files, ~13,500 lines
```

### After (clientele)

```
qontract_api_client/
├── client.py    # All endpoints as decorated async functions
├── schemas.py   # All models as Pydantic BaseModel classes
├── config.py    # Client configuration
└── exceptions.py
→ 4 files, ~1,500 lines
```

### Key changes

| Area | What changed |
|---|---|
| `qontract_api_client/` | Regenerated with clientele — flat structure, Pydantic models, async functions |
| `reconcile/utils/runtime/integration.py` | `QontractReconcileApiIntegration` uses `client.configure()` instead of `AuthenticatedClient` |
| `reconcile/*.py` integrations | Import functions directly from `qontract_api_client.client`, models from `qontract_api_client.schemas` |
| `reconcile/test/` | Remove stale `qontract_api_client` property patches, fix `kwargs["body"]` → `args[0]` |
| `qontract_api/subscriber/_client.py` | Lazy init, DRY `_build_chat_request` helper, updated imports to clientele API |
| `qontract_api/tests/subscriber/test_client.py` | **New** — unit tests for `_setup_client`, `post_to_slack`, `send_dm` |
| `qontract_api_client/Dockerfile` | Remove stale `openapi_python_client_templates` COPY |

## Test plan

- [x] All `reconcile/test/ldap_users_api/` tests pass (11 tests)
- [x] All `reconcile/test/glitchtip_api/` tests pass (23 tests)
- [x] All `reconcile/test/test_github_owners_api.py` tests pass (21 tests)
- [x] All `reconcile/test/ldap_users_api/test_mr_builder.py` tests pass (9 tests)
- [x] All `qontract_api/tests/subscriber/` tests pass (26 tests)
- [x] Linter passes (`ruff check`)

## Summary by CodeRabbit

* **Refactor**
  * Reworked client surface to a single global, schema-driven client and consolidated generated schemas into one module.

* **Bug Fixes**
  * Slack usergroup notifications now default to an empty list to avoid null handling.

* **Tests**
  * Added async tests covering Slack posting, direct messages, and client setup/authorization.

* **Chores**
  * Updated client generation and build/tooling configuration.


reconcile/utils/runtime/integration.py
```diff
@@ -266,29 +264,29 @@ async def _raise_on_4xx_5xx(response: httpxyz.Response) -> None:
         if response.status_code != 409:
             response.raise_for_status()
 
-    @property
-    def qontract_api_client(self) -> AuthenticatedClient:
-        """
-        Returns the qontract-api client.
-        """
+    def _setup_qontract_api_client(self) -> None:
+        """Configure the Qontract API client with the server URL and token from the config."""
```
hemslo: **Dropped error handling:** The old `AuthenticatedClient` was configured with `event_hooks={"response": [self._raise_on_4xx_5xx]}`, which automatically raised on 4xx/5xx (except 409, used for conflict/dedup). The new `clientele` `Config` has no equivalent, and `_raise_on_4xx_5xx` is still defined here but **never wired up**.

This is a behavioral regression — HTTP errors from qontract-api that previously raised exceptions will now return silently, potentially causing integrations to proceed with error responses as if they succeeded.

Either:
- Configure clientele's httpx instance with the event hook
- Verify that clientele already raises on 4xx/5xx internally (and handle the 409 exception at call sites)
- Or remove the dead `_raise_on_4xx_5xx` method if it's truly no longer needed

reconcile/utils/runtime/integration.py
```diff
@@ -256,6 +253,7 @@ class QontractReconcileApiIntegration[RunParamsTypeVar: RunParams](ABC):
     def __init__(self, params: RunParamsTypeVar) -> None:
         self.params: RunParamsTypeVar = params
         self._secret_reader: SecretReaderBase | None = None
+        self._setup_qontract_api_client()
```
hemslo: This configures a **module-level singleton** (`qontract_api_client.configure()`) in every constructor call. The old code returned a new `AuthenticatedClient` per property access — stateless and safe.

If multiple integrations are instantiated (even in tests), each reconfigures the shared global client. Consider adding a guard (e.g., configure-once check) or documenting the singleton behavior.

reconcile/utils/runtime/integration.py
```diff
@@ -266,29 +264,29 @@ async def _raise_on_4xx_5xx(response: httpxyz.Response) -> None:
...
-            token=config["qontract-api"].get("token", ""),
-            httpx_args={
-                "event_hooks": {"response": [self._raise_on_4xx_5xx]},
-            },
-            headers=environment_headers,
+        qontract_api_client.configure(
+            config=QontractApiClientConfig(
+                base_url=urlparse(
+                    config.get("qontract-api", {}).get("server", "")
+                ).geturl(),
```
hemslo: Old code used `config["qontract-api"]["server"]` which fails fast with a clear `KeyError` on missing config. This `.get(..., "")` chain silently returns an empty string, leading to confusing downstream errors (requests to `http://` or empty base URL).

Consider keeping `config["qontract-api"]["server"]` for the required server URL — the defensive `.get()` is fine for `token` (optional) but not for `server` (required).

qontract_api/qontract_api/subscriber/_client.py
```diff
@@ -1,54 +1,49 @@
...
 
 logger = get_logger(__name__)
 
 
-def _get_client() -> AuthenticatedClient:
-    """Create authenticated qontract-api client.
-
-    Returns:
-        AuthenticatedClient configured with base URL and auth token
+def _setup_client() -> None:
```
hemslo, line 16: Nit: `_setup_client()` is called at the start of both `post_to_slack()` and `send_dm()`, reconfiguring the global client on every invocation. Harmless but wasteful — consider a module-level flag or `functools.lru_cache` to configure once.

hemslo (commented):
Good direction overall — Pydantic alignment, massive reduction in generated boilerplate, and elimination of the attrs/Unset/from_dict ceremony. A few concerns below, the error handling one being the most important.

**Upstream dependency:** This PR requires https://github.com/phalt/clientele/pull/247 which is still **OPEN** — can't merge until that lands and a release is cut.

reconcile/utils/runtime/integration.py
chassing: Fixed in abdbe9e6. Removed `_raise_on_4xx_5xx` — clientele already calls `raise_for_status()` on all 4xx/5xx by default (when no `response_map` is provided, which is the case for all generated endpoint functions).

reconcile/utils/runtime/integration.py
chassing: Fixed in abdbe9e6. Extracted `_setup_qontract_api_client` to a `@functools.cache` module-level function — lazy (no longer called in `__init__`) and configure-once. The runner calls it before `async_run`.

reconcile/utils/runtime/integration.py
chassing: Fixed in abdbe9e6. Restored `config["qontract-api"]["server"]` for fail-fast `KeyError`. Also made the setup lazy (called from runner, not `__init__`) so tests that don't need the API client aren't affected.

qontract_api/qontract_api/subscriber/_client.py
chassing: Fixed in abdbe9e6. Added a module-level `_client_configured` guard so `_setup_client()` only configures on first call.

reconcile/ldap_users_api/integration.py
```diff
@@ -250,8 +245,7 @@ async def _delete_users_app_interface(
             if not dry_run:
                 sync_calls.append(
                     vcs_file_sync(
```
hemslo, line 247: **Potential 409 regression on dedup:** The old `_raise_on_4xx_5xx` event hook explicitly swallowed 409 responses (used by the API for dedup / idempotent MR creation). With clientele, a 409 from `vcs_file_sync` will now raise an unhandled `HTTPStatusError` instead of passing through silently.

Note that `_get_file_content` correctly catches `HTTPStatusError(404)` — but `vcs_file_sync` calls here have no equivalent 409 guard.

Can you confirm whether the server-side `vcs_file_sync` still returns 409 for duplicate MR titles, or has it been changed to return 200 with a status field? If 409 is still possible, callers need `except HTTPStatusError as e: if e.response.status_code == 409: ...`.

reconcile/utils/runtime/runner.py
```diff
@@ -19,6 +19,7 @@
     QontractReconcileApiIntegration,
     QontractReconcileIntegration,
     RunParams,
+    _setup_qontract_api_client,
```
hemslo: Nit: `_setup_qontract_api_client` starts with `_` (private by convention) but is explicitly imported and called from a separate module. Consider renaming it to `setup_qontract_api_client` (no leading underscore) to signal that it is part of the module's public API.

reconcile/utils/runtime/integration.py
```diff
@@ -241,11 +245,26 @@ def build_integration_instance_for_shard(self, shard: str) -> Self:
         )
 
 
-class FromDict(Protocol):
-    """Protocol for generated API client models with from_dict classmethod."""
+@functools.cache
```
hemslo, line 248: Note for test authors: `@functools.cache` means this function is cached for the lifetime of the process. Tests that need a fresh configuration (e.g., different base URLs) must call `_setup_qontract_api_client.cache_clear()` in teardown. The current tests avoid this by mocking `qontract_api_client` directly (not going through `runner.py`), so no immediate issue — but worth documenting.

hemslo (commented):
Previous issues all addressed — good fixes. One new concern on the 409/dedup behavior, plus two minor notes.

reconcile/ldap_users_api/integration.py
chassing: Not applicable — the qontract-api server never returns 409 anywhere. The `vcs_file_sync` endpoint returns `200 + FileSyncStatus.MR_EXISTS` for dedup cases. The old `_raise_on_4xx_5xx` hook that swallowed 409 was dead code that was never triggered.

reconcile/utils/runtime/runner.py
chassing: Fixed in 1df489bf. Renamed to `setup_qontract_api_client` (no underscore).

reconcile/utils/runtime/integration.py
chassing: Fixed in 1df489bf. Added a docstring note about `cache_clear()` for test teardown.

hemslo (approved):
LGTM!

PR #5583, opened 2026-06-03
Added merge commit sha guard

rhur-pixel:
In gitlab_housekeeping.py, return 0 early if merge_commit_sha is not yet populated, letting the next reconcile loop retry naturally.

Fixes a bug where `lead.merge_commit_sha` can be None

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error resilience with safeguards for missing merge information, allowing automatic retry on subsequent operations instead of proceeding with incomplete data.

* **Tests**
  * Added comprehensive test coverage to verify proper handling of missing merge information scenarios and confirm automatic retry behavior.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -1069,6 +1069,16 @@ def _process_omm_group(
         clear_omm_group(dry_run, gl, lead=lead)
         return 0
 
+    if not lead.merge_commit_sha:
+        logging.warning([
+            "omm-group",
+            "lead-missing-merge-sha",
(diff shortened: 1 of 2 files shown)


TGPSKI (approved):
The guard is placed in exactly the right spot — after the window and post-merge-CI checks pass, but before the current_head != lead.merge_commit_sha comparison:

```
if not lead.merge_commit_sha:
    logging.warning([...])
    return 0
```

Returning 0 without calling clear_omm_group is correct: the group stays intact, and the next reconcile loop retries naturally once GitLab populates the field.LGTM

PR #5585, opened 2026-06-03
fix(vault): add exponential backoff and transport retry to VaultClient auth

TGPSKI:
## Summary

- **Fix broken auth retry**: `_refresh_client_auth` was swallowing all exceptions (caught `Exception`, logged, never re-raised), so the init retry loop never actually retried on transient Vault connectivity failures. The `break` was also unconditional, exiting even when `authenticated=False`. Pods would crash immediately with `VaultConnectionError` on the first transient timeout.
- **Add exponential backoff**: Increased retry attempts from 3→5 with exponential backoff (1s, 2s, 4s, 8s, 16s capped at 30s) instead of flat 1s sleep. Configurable via `VAULT_AUTH_MAX_ATTEMPTS`, `VAULT_AUTH_BACKOFF_BASE`, `VAULT_AUTH_BACKOFF_MAX` env vars.
- **Add urllib3 transport-level retry**: Configure `Retry` on the `HTTPAdapter` with 3 connect retries and 0.5s backoff factor, providing transparent TCP-level retry before exceptions reach the hvac client.
- **Auto-refresh thread resilience**: Wrap `_auto_refresh_client_auth` in try/except so the daemon thread survives transient failures instead of dying silently, retrying on the next interval.

## Context

On 2026-06-03 we experienced intermittent `ConnectTimeoutError` from ~200 integration pods authenticating to `vault.devshift.net` on `appsrep09ue1`. The traffic hairpins through the OpenShift router (pod → ELB → router → vault pod on the same cluster), and the route is fragile under load. Pods were crashing immediately on the first timeout instead of retrying, causing cascading restarts.

Investigation confirmed Vault and node networking were healthy — the issue was purely client-side: stale connection pools from long-running pods + the broken retry logic meant any transient timeout was fatal.

## Test plan

- [x] Existing unit tests pass (`test_vault_utils.py` — 4/4)
- [x] `ruff check` clean
- [x] `mypy` clean
- [ ] Verify in staging that VaultClient retries on transient auth failures instead of crashing
- [ ] Confirm env var overrides work (`VAULT_AUTH_MAX_ATTEMPTS`, `VAULT_AUTH_BACKOFF_BASE`, `VAULT_AUTH_BACKOFF_MAX`)

## Summary by CodeRabbit

* **Improvements**
  * Vault authentication now retries with configurable exponential backoff and more resilient HTTP transport handling for transient failures.
  * Clear connection error is raised if authentication attempts are exhausted after retries.
  * Background token auto-refresh is more resilient: failures are logged and the refresh loop continues.
  * Kubernetes auth failure logs now include server URL, role, and mount for clearer diagnostics.

### DIFF
--- a/reconcile/utils/vault.py
+++ b/reconcile/utils/vault.py
@@ -13,11 +13,21 @@ import requests
 from hvac.exceptions import InvalidPath
 from requests.adapters import HTTPAdapter
 from sretoolbox.utils import retry
+from urllib3.util.retry import Retry as Urllib3Retry
 
 from reconcile.utils.config import get_config
 
 LOG = logging.getLogger(__name__)
 VAULT_AUTO_REFRESH_INTERVAL = int(os.getenv("VAULT_AUTO_REFRESH_INTERVAL") or 600)
+VAULT_AUTH_MAX_ATTEMPTS = int(os.getenv("VAULT_AUTH_MAX_ATTEMPTS") or 5)
+VAULT_AUTH_BACKOFF_BASE = float(os.getenv("VAULT_AUTH_BACKOFF_BASE") or 2.0)
+VAULT_AUTH_BACKOFF_MAX = float(os.getenv("VAULT_AUTH_BACKOFF_MAX") or 30.0)
+
+_VAULT_RETRYABLE_EXCEPTIONS = (
+    requests.exceptions.ConnectionError,
+    requests.exceptions.Timeout,
+    requests.exceptions.TooManyRedirects,
+)
 
 
 class PathAccessForbiddenError(Exception):
@@ -120,27 +130,40 @@ class VaultClient:
         self._read_all_v2 = lru_cache(maxsize=2048)(self.__read_all_v2)
 
         session = requests.Session()
-        # There are at most 10 working threads in reconcile, plus 1 daemon thread for auto refresh
-        adapter = HTTPAdapter(pool_maxsize=11)
+        transport_retry = Urllib3Retry(
+            total=3,
+            connect=3,
+            backoff_factor=0.5,
+        )
+        adapter = HTTPAdapter(pool_maxsize=11, max_retries=transport_retry)
         session.mount("https://", adapter)
         self._client = hvac.Client(url=server, session=session)
         self._close_lock = threading.Lock()
         self._closed = False
 
         authenticated = False
-        for _ in range(3):
+        last_error: Exception | None = None
+        for attempt in range(VAULT_AUTH_MAX_ATTEMPTS):
             try:
                 self._refresh_client_auth()
                 authenticated = self._client.is_authenticated()
-                break
-            except (
-                requests.exceptions.ConnectionError,
-                requests.exceptions.TooManyRedirects,
-            ):
-                time.sleep(1)
+                if authenticated:
+                    break
+            except _VAULT_RETRYABLE_EXCEPTIONS as exc:
+                last_error = exc
+
+            if attempt < VAULT_AUTH_MAX_ATTEMPTS - 1:
+                backoff = min(VAULT_AUTH_BACKOFF_BASE**attempt, VAULT_AUTH_BACKOFF_MAX)
+                LOG.warning(
+                    "Vault auth attempt %d/%d failed, retrying in %.1fs",
+                    attempt + 1,
+                    VAULT_AUTH_MAX_ATTEMPTS,
+                    backoff,
+                )
+                time.sleep(backoff)
 
         if not authenticated:
-            raise VaultConnectionError()
+            raise VaultConnectionError() from last_error
 
         if auto_refresh:
             t = threading.Thread(target=self._auto_refresh_client_auth, daemon=True)
@@ -163,7 +186,9 @@ class VaultClient:
 
     def _auto_refresh_client_auth(self) -> None:
         """
-        Thread that periodically refreshes the vault token
+        Thread that periodically refreshes the vault token.
+        Failures are logged but do not kill the thread; the next
+        interval will retry.
         """
         while True:
             time.sleep(VAULT_AUTO_REFRESH_INTERVAL)
@@ -172,11 +197,17 @@ class VaultClient:
                     LOG.debug("client is closed, exiting auto refresh")
                     break
                 LOG.debug("auto refresh client auth")
-                self._refresh_client_auth()
+                try:
+                    self._refresh_client_auth()
+                except Exception:
+                    LOG.warning(
+                        "auto refresh auth failed, will retry in %ds",
+                        VAULT_AUTO_REFRESH_INTERVAL,
+                        exc_info=True,
+                    )
 
     def _refresh_client_auth(self) -> None:
         if self.kube_auth_enabled:
-            # must read each time to account for sa token refresh
             with open(self.kube_sa_token_path, encoding="locale") as f:
                 try:
                     self._client.auth.kubernetes.login(
@@ -186,9 +217,14 @@ class VaultClient:
                     )
                 except Exception as e:
                     LOG.error(
-                        f"Failed to authenticate to Vault server {self._client.url} "
-                        f"using role '{self.kube_auth_role}' on mount '{self.kube_auth_mount}': {e}"
+                        "Failed to authenticate to Vault server %s "
+                        "using role '%s' on mount '%s': %s",
+                        self._client.url,
+                        self.kube_auth_role,
+                        self.kube_auth_mount,
+                        e,
                     )
+                    raise
         else:
             self._client.auth.approle.login(
                 role_id=self.role_id, secret_id=self.secret_id
--- a/reconcile/utils/vault.py
+++ b/reconcile/utils/vault.py
@@ -149,15 +149,26 @@ class VaultClient:
                 authenticated = self._client.is_authenticated()
                 if authenticated:
                     break
+                LOG.warning(
+                    "Vault auth attempt %d/%d: server responded but "
+                    "is_authenticated() returned False (token may be "
+                    "invalid or session rejected)",
+                    attempt + 1,
+                    VAULT_AUTH_MAX_ATTEMPTS,
+                )
             except _VAULT_RETRYABLE_EXCEPTIONS as exc:
                 last_error = exc
+                LOG.warning(
+                    "Vault auth attempt %d/%d hit retryable error: %s",
+                    attempt + 1,
+                    VAULT_AUTH_MAX_ATTEMPTS,
+                    exc,
+                )
 
             if attempt < VAULT_AUTH_MAX_ATTEMPTS - 1:
                 backoff = min(VAULT_AUTH_BACKOFF_BASE**attempt, VAULT_AUTH_BACKOFF_MAX)
                 LOG.warning(
-                    "Vault auth attempt %d/%d failed, retrying in %.1fs",
-                    attempt + 1,
-                    VAULT_AUTH_MAX_ATTEMPTS,
+                    "retrying in %.1fs",
                     backoff,
                 )
                 time.sleep(backoff)
--- a/reconcile/utils/vault.py
+++ b/reconcile/utils/vault.py
@@ -174,6 +174,17 @@ class VaultClient:
                 time.sleep(backoff)
 
         if not authenticated:
+            failure_reason = (
+                f"retryable transport error: {last_error}"
+                if last_error
+                else "is_authenticated() returned False on every attempt"
+            )
+            LOG.error(
+                "Vault auth failed after %d/%d attempts; last failure: %s",
+                VAULT_AUTH_MAX_ATTEMPTS,
+                VAULT_AUTH_MAX_ATTEMPTS,
+                failure_reason,
+            )
             raise VaultConnectionError() from last_error
 
         if auto_refresh:


reconcile/utils/vault.py
```diff
@@ -120,24 +130,37 @@ def __init__(
...
-            ):
-                time.sleep(1)
+                if authenticated:
+                    break
+            except _VAULT_RETRYABLE_EXCEPTIONS:
+                pass
+
+            if attempt < VAULT_AUTH_MAX_ATTEMPTS - 1:
+                backoff = min(VAULT_AUTH_BACKOFF_BASE**attempt, VAULT_AUTH_BACKOFF_MAX)
+                LOG.warning(
```
jfchevrette, line 170: the warning log fires even on the last failed attempt right before we raise `VaultConnectionError()`. at that point the "retrying in..." message is misleading — we're not retrying, we're about to crash.

maybe guard this with `if attempt < VAULT_AUTH_MAX_ATTEMPTS - 1` (which you already have for the sleep) and move the log inside that block, or log at ERROR level on the final attempt with a different message.

reconcile/utils/vault.py
```diff
@@ -120,24 +130,37 @@ def __init__(
...
         session.mount("https://", adapter)
         self._client = hvac.Client(url=server, session=session)
         self._close_lock = threading.Lock()
         self._closed = False
 
         authenticated = False
-        for _ in range(3):
+        for attempt in range(VAULT_AUTH_MAX_ATTEMPTS):
             try:
                 self._refresh_client_auth()
```
jfchevrette, line 148: this silently swallows the exception details on retryable failures. when all attempts exhaust and we hit `VaultConnectionError()` a few lines down, operators have no idea what the actual error was.

consider at minimum logging the exception in the `except` block (or capturing it and passing it as `__cause__` to `VaultConnectionError`):
```python
except _VAULT_RETRYABLE_EXCEPTIONS as e:
    last_error = e
...
if not authenticated:
    raise VaultConnectionError() from last_error
```
that way the traceback shows the root cause.

reconcile/utils/vault.py
```diff
@@ -172,11 +197,16 @@ def _auto_refresh_client_auth(self) -> None:
                     LOG.debug("client is closed, exiting auto refresh")
                     break
                 LOG.debug("auto refresh client auth")
-                self._refresh_client_auth()
+                try:
+                    self._refresh_client_auth()
+                except Exception:
+                    LOG.warning(
```
jfchevrette, line 225: agree with coderabbit here — `exc_info=True` would be useful. a bare `except Exception` with a generic message makes it hard to tell whether this is a transient network blip or a persistent auth misconfiguration. the difference matters for on-call.

reconcile/utils/vault.py
```diff
@@ -120,24 +130,37 @@ def __init__(
         self._read_all_v2 = lru_cache(maxsize=2048)(self.__read_all_v2)
 
         session = requests.Session()
-        # There are at most 10 working threads in reconcile, plus 1 daemon thread for auto refresh
-        adapter = HTTPAdapter(pool_maxsize=11)
+        transport_retry = Urllib3Retry(
+            total=3,
+            connect=3,
```
jfchevrette, line 135: nit: `status_forcelist` retries on 5xx at the transport level, but hvac likely raises its own exceptions for those status codes before urllib3 gets a chance to retry. have you confirmed that these actually trigger urllib3 retries and don't get intercepted by hvac/requests first?

if they don't fire, the list is dead config — not harmful, but misleading.

reconcile/utils/vault.py
```diff
@@ -186,9 +216,14 @@ def _refresh_client_auth(self) -> None:
...
+                        "using role '%s' on mount '%s': %s",
+                        self._client.url,
+                        self.kube_auth_role,
+                        self.kube_auth_mount,
+                        e,
                     )
+                    raise
         else:
             self._client.auth.approle.login(
                 role_id=self.role_id, secret_id=self.secret_id
```
jfchevrette, line 252: the `raise` is the key fix here and it looks correct. one thing to note: the `open()` call is outside the try/except, so an `OSError` from a missing/unreadable SA token file will propagate unhandled and crash the init loop (it's not in `_VAULT_RETRYABLE_EXCEPTIONS`). that's probably fine for a missing token file (not transient), but worth a quick sanity check that it's the intended behavior.

reconcile/utils/vault.py
```diff
@@ -120,24 +130,37 @@ def __init__(
...
-                break
-            except (
-                requests.exceptions.ConnectionError,
-                requests.exceptions.TooManyRedirects,
-            ):
-                time.sleep(1)
+                if authenticated:
+                    break
+            except _VAULT_RETRYABLE_EXCEPTIONS:
+                pass
```
jfchevrette: The last exception is silently discarded here. When all attempts exhaust, `VaultConnectionError()` is raised with no context about *what* actually went wrong — operators see the crash but not the root cause.

Consider capturing the last exception and chaining it:

```python
last_error: Exception | None = None
for attempt in range(VAULT_AUTH_MAX_ATTEMPTS):
    try:
        ...
    except _VAULT_RETRYABLE_EXCEPTIONS as exc:
        last_error = exc
    ...

if not authenticated:
    raise VaultConnectionError() from last_error
```

This way the traceback carries the actual failure (e.g. `ConnectTimeoutError` to `vault.devshift.net`) instead of just a bare `VaultConnectionError`.

reconcile/utils/vault.py
```diff
@@ -172,11 +197,16 @@ def _auto_refresh_client_auth(self) -> None:
...
                     break
                 LOG.debug("auto refresh client auth")
-                self._refresh_client_auth()
+                try:
+                    self._refresh_client_auth()
+                except Exception:
+                    LOG.warning(
+                        "auto refresh auth failed, will retry in %ds",
+                        VAULT_AUTO_REFRESH_INTERVAL,
+                    )
```
jfchevrette, line 229: nit: Without `exc_info=True`, persistent auth misconfigurations look identical to transient network blips in the logs — the warning message is the same either way. Adding the traceback would help operators tell them apart during an incident.

```suggestion
                except Exception:
                    LOG.warning(
                        "auto refresh auth failed, will retry in %ds",
                        VAULT_AUTO_REFRESH_INTERVAL,
                        exc_info=True,
                    )
```

reconcile/utils/vault.py
TGPSKI: Good catch — but this is actually already guarded. The log + sleep block is inside `if attempt < VAULT_AUTH_MAX_ATTEMPTS - 1`, so it only fires on non-final attempts. On the final attempt, we fall through directly to `raise VaultConnectionError() from last_error`. No misleading message.

reconcile/utils/vault.py
TGPSKI: Agreed. Now capturing `last_error` and chaining it: `raise VaultConnectionError() from last_error`. The traceback will show the actual root cause (e.g. `ConnectTimeoutError` to `vault.devshift.net`).

reconcile/utils/vault.py
TGPSKI: Done — added `exc_info=True`.

reconcile/utils/vault.py
TGPSKI: You're right — hvac/requests intercepts HTTP responses and raises its own exceptions before urllib3's status-based retry can fire. Removed `status_forcelist` entirely. The connect retries (`connect=3`) are the ones that actually matter here and those work at the TCP layer before any HTTP response parsing.

reconcile/utils/vault.py
TGPSKI: Confirmed — this is intentional. A missing/unreadable SA token file is a configuration error, not a transient network issue, so it should propagate immediately and crash rather than retrying for 30+ seconds against a file that won't appear.

reconcile/utils/vault.py
TGPSKI: Fixed — now captures `last_error` and chains: `raise VaultConnectionError() from last_error`. Operators will see the full cause chain in the traceback.

reconcile/utils/vault.py
TGPSKI: Done — added `exc_info=True` so the traceback distinguishes transient network blips from persistent auth misconfigs.

reconcile/utils/vault.py
```diff
@@ -120,27 +130,40 @@ def __init__(
...
         self._close_lock = threading.Lock()
         self._closed = False
 
         authenticated = False
-        for _ in range(3):
+        last_error: Exception | None = None
+        for attempt in range(VAULT_AUTH_MAX_ATTEMPTS):
             try:
                 self._refresh_client_auth()
                 authenticated = self._client.is_authenticated()
```
jonmosco, line 149: Could we add a bit more logging around this auth check/retry path? It would be really helpful to log the concrete exception when a retryable network error happens, and also log explicitly when Vault responds but is_authenticated() is still false. That would make it much clearer whether we’re dealing with connectivity issues or Vault rejecting the auth/session.

TGPSKI: Done — added two specific log messages:

1. **Retryable network error**: `"Vault auth attempt %d/%d hit retryable error: %s"` — logs the concrete exception (e.g. `ConnectTimeoutError`) so operators can see exactly what failed
2. **Auth rejection**: `"Vault auth attempt %d/%d: server responded but is_authenticated() returned False (token may be invalid or session rejected)"` — fires when Vault responds but the session isn't valid

This makes it easy to tell connectivity issues apart from Vault rejecting the auth/session at a glance.

reconcile/utils/vault.py
```diff
@@ -120,27 +130,40 @@ def __init__(
...
+                backoff = min(VAULT_AUTH_BACKOFF_BASE**attempt, VAULT_AUTH_BACKOFF_MAX)
+                LOG.warning(
+                    "Vault auth attempt %d/%d failed, retrying in %.1fs",
+                    attempt + 1,
+                    VAULT_AUTH_MAX_ATTEMPTS,
+                    backoff,
+                )
+                time.sleep(backoff)
 
         if not authenticated:
```
jonmosco, line 176: Could we add a log here before raising as well? Even with the chained exception, I think it would be helpful to have an explicit log saying that all Vault auth retries were exhausted.

Ideally, it would include the attempt count and whether the last failure was a retryable transport error or an unauthenticated response. That would make it a lot easier to spot the actual failure path in the normal logs.

TGPSKI: Done — added an explicit `LOG.error` before raising `VaultConnectionError`:

```
Vault auth failed after 5/5 attempts; last failure: retryable transport error: ConnectTimeoutError(...)
```

or when auth was rejected:

```
Vault auth failed after 5/5 attempts; last failure: is_authenticated() returned False on every attempt
```

Operators can now see at a glance whether retries were exhausted due to network issues or Vault rejecting the session.

reconcile/utils/vault.py
jonmosco: LGTM!

jonmosco (approved):
LGTM - it will be good to get some tracing to get an idea where along the path this is happening.

TGPSKI:
Merging this tomorrow

hemslo (commented):
See inline comments for simplification suggestion using existing `sretoolbox.utils.retry`.

reconcile/utils/vault.py
```diff
@@ -120,27 +130,62 @@ def __init__(
...
+            LOG.error(
+                "Vault auth failed after %d/%d attempts; last failure: %s",
+                VAULT_AUTH_MAX_ATTEMPTS,
+                VAULT_AUTH_MAX_ATTEMPTS,
+                failure_reason,
+            )
+            raise VaultConnectionError() from last_error
 
         if auto_refresh:
             t = threading.Thread(target=self._auto_refresh_client_auth, daemon=True)
```
hemslo, line 191: suggestion: this hand-rolled retry loop reimplements what `sretoolbox.utils.retry` already provides — the same decorator used on `read`, `write`, `delete` in this file.

Extracting auth into a decorated method replaces ~40 lines with ~7:

```python
@retry(
    exceptions=(*_VAULT_RETRYABLE_EXCEPTIONS, VaultConnectionError),
    max_attempts=VAULT_AUTH_MAX_ATTEMPTS,
    hook=lambda exc: LOG.warning("Vault auth attempt failed, retrying: %s", exc),
)
def _initial_auth(self) -> None:
    self._refresh_client_auth()
    if not self._client.is_authenticated():
        raise VaultConnectionError("is_authenticated() returned False")
```

Then in `__init__`:

```python
try:
    self._initial_auth()
except _VAULT_RETRYABLE_EXCEPTIONS as e:
    raise VaultConnectionError() from e
# VaultConnectionError from is_authenticated() failure propagates as-is
```

**Tradeoff:** linear backoff (1s, 2s, 3s, 4s = 10s) vs exponential (1s, 2s, 4s, 8s = 15s). For 5 attempts the difference is modest, and linear is the established pattern in this codebase.

reconcile/utils/vault.py
```diff
@@ -13,11 +13,21 @@
...
 from requests.adapters import HTTPAdapter
 from sretoolbox.utils import retry
+from urllib3.util.retry import Retry as Urllib3Retry
 
 from reconcile.utils.config import get_config
 
 LOG = logging.getLogger(__name__)
 VAULT_AUTO_REFRESH_INTERVAL = int(os.getenv("VAULT_AUTO_REFRESH_INTERVAL") or 600)
+VAULT_AUTH_MAX_ATTEMPTS = int(os.getenv("VAULT_AUTH_MAX_ATTEMPTS") or 5)
+VAULT_AUTH_BACKOFF_BASE = float(os.getenv("VAULT_AUTH_BACKOFF_BASE") or 2.0)
```
hemslo, line 23: These two env vars can be dropped if using `sretoolbox.utils.retry` — its linear backoff (`time.sleep(attempt)`) is sufficient and consistent with the rest of the codebase. `VAULT_AUTH_MAX_ATTEMPTS` is the only knob needed.

reconcile/utils/vault.py
TGPSKI: Appreciate the suggestion — we considered using sretoolbox.utils.retry here but decided against it for a few reasons:

1. **Exponential backoff for thundering herd** — the original incident was ~200 pods all hitting the OpenShift router simultaneously after a mass restart. Exponential backoff (1s, 2s, 4s, 8s) spreads reconnection attempts more effectively than linear (1s, 2s, 3s, 4s) for this failure mode. The env vars let us tune this operationally without a code change.

2. **Granular per-attempt logging** — jonmosco specifically requested logging that distinguishes transport errors (ConnectTimeoutError) from auth rejections (is_authenticated() == False). The hook callback on sretoolbox.retry receives the exception but can't easily differentiate the "server responded but session was rejected" path since that raises a different exception type at a different point in the flow.

3. **Different retry layers** — sretoolbox.retry decorates vault read/write/delete for API-level retries after auth. The auth loop retries the initial authentication itself, and the urllib3.Retry on the HTTPAdapter handles TCP connect retries below both of these. Each layer has different failure modes and appropriate backoff strategies.

The hand-rolled loop is more verbose but gives operators clear visibility into exactly what's failing during an incident, which was the key gap in the original code.

reconcile/utils/vault.py
TGPSKI: Keeping these for operational tuning — during the incident we would have benefited from being able to adjust backoff timing without a code deploy. VAULT_AUTH_MAX_ATTEMPTS controls how many attempts, and the backoff vars control the delay curve shape. With linear backoff the total wait for 5 attempts is ~10s; with exponential it's ~15s, which better absorbs router congestion spikes from mass pod restarts.

reconcile/utils/vault.py
hemslo: Thanks for the detailed response — worth pushing back on each point though:

**1. Exponential backoff for thundering herd**

Exponential vs linear doesn't help here without jitter. When all 200 pods fail at the same moment and sleep for the same duration, they stay synchronized regardless of the backoff curve shape — they just hit Vault in lockstep waves at different intervals. What actually breaks the synchronization is randomizing the sleep time ("full jitter" or "equal jitter"). The PR adds neither — not in the application loop, and `backoff_jitter` is not set on `urllib3.Retry` either. So the thundering herd concern is real, but exponential vs linear doesn't address it.

**2. Granular per-attempt logging**

You can differentiate exception types in the hook via `isinstance`:

```python
def _log_retry(exc: Exception) -> None:
    if isinstance(exc, VaultConnectionError):
        LOG.warning("Vault auth rejected by server, retrying: %s", exc)
    else:
        LOG.warning("Vault auth transport error, retrying: %s", exc)
```

The real limitation is that the hook doesn't receive the attempt counter, so you can't log "attempt 2/5". That's a genuine gap — but the claim that it "can't easily differentiate" the two failure paths isn't accurate.

**3. Different retry layers**

This is a true observation but it doesn't follow that `sretoolbox.retry` can't be used for auth init. The layers stay separate either way — nothing about using the decorator collapses them.

The honest reason to keep the hand-rolled loop is: you want both exponential backoff AND the attempt counter in logs, and `sretoolbox.retry` doesn't provide either. That's a reasonable pragmatic argument, but it's not what was stated.

TGPSKI: @hemslo — fair pushback on all three points, and you're right that my Jun 5 rationale wasn't precise.

**Thundering herd:** Agree — curve shape alone doesn't desync pods; jitter does. That's the gap in this PR as written.

**Logging:** Agree the hook can distinguish exception types via `isinstance`; the real gap is the attempt counter (`2/5`), which jonmosco asked for explicitly.

**Layers:** Agree using `@retry` for init auth doesn't collapse the layers.

Rather than bolting jitter onto a hand-rolled loop here, I've opened [sretoolbox#299](https://github.com/app-sre/sretoolbox/pull/299) to extend `sretoolbox.utils.retry` with optional exponential backoff, full jitter, and attempt-aware hooks — default linear behavior unchanged for existing callers. Tracking in [APPSRE-14592](https://issues.redhat.com/browse/APPSRE-14592).

**Plan for this PR:** hold merge until sretoolbox `4.1.0` lands, then rewrite the auth init path here to use `@retry(backoff="exponential", jitter=True, hook=...)` instead of the hand-rolled loop. urllib3 transport retry and env-var tuning stay in `vault.py`.

The core bugfixes in this PR (re-raise from `_refresh_client_auth`, break only on `authenticated=True`, auto-refresh survival) remain the same either way.

reconcile/utils/vault.py
```diff
@@ -120,27 +130,62 @@ def __init__(
...
+                    exc,
+                )
+
+            if attempt < VAULT_AUTH_MAX_ATTEMPTS - 1:
+                backoff = min(VAULT_AUTH_BACKOFF_BASE**attempt, VAULT_AUTH_BACKOFF_MAX)
+                LOG.warning(
+                    "retrying in %.1fs",
+                    backoff,
+                )
+                time.sleep(backoff)
```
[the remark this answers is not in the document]
TGPSKI: Valid — and aligns with hemslo's feedback. Will address via `jitter=True` on `@retry` after [sretoolbox#299](https://github.com/app-sre/sretoolbox/pull/299) lands and releases as `4.1.0`, plus `backoff_jitter` on the urllib3 `Retry` HTTPAdapter. Not adding ad-hoc jitter to the hand-rolled loop since that loop is being replaced in the rewrite planned for this PR.

reconcile/utils/vault.py
```diff
@@ -13,11 +13,21 @@
...
 from sretoolbox.utils import retry
+from urllib3.util.retry import Retry as Urllib3Retry
 
 from reconcile.utils.config import get_config
 
 LOG = logging.getLogger(__name__)
 VAULT_AUTO_REFRESH_INTERVAL = int(os.getenv("VAULT_AUTO_REFRESH_INTERVAL") or 600)
+VAULT_AUTH_MAX_ATTEMPTS = int(os.getenv("VAULT_AUTH_MAX_ATTEMPTS") or 5)
+VAULT_AUTH_BACKOFF_BASE = float(os.getenv("VAULT_AUTH_BACKOFF_BASE") or 2.0)
+VAULT_AUTH_BACKOFF_MAX = float(os.getenv("VAULT_AUTH_BACKOFF_MAX") or 30.0)
```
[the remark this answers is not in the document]
TGPSKI: Will add — validation before passing values to `@retry` when we rewrite the auth init path after [sretoolbox#299](https://github.com/app-sre/sretoolbox/pull/299) / `4.1.0` lands. Fail fast at import with a clear `ValueError` naming the var and expected bounds (e.g. `VAULT_AUTH_MAX_ATTEMPTS >= 1`, `VAULT_AUTH_BACKOFF_BASE > 0`).

TGPSKI:
Holding for sretoolbox/retry feature

reconcile/utils/vault.py
hemslo: Great plan. Extending sretoolbox upstream is the right move — all callers benefit from exponential backoff + jitter without hand-rolling their own loops. Happy to review sretoolbox#299 when it's ready.

For this PR in the meantime: the core bugfixes (re-raise in `_refresh_client_auth`, conditional break, auto-refresh resilience) are solid and can merge as-is if you don't want to block on sretoolbox 4.1.0. The retry loop rewrite can follow.

PR #5590, opened 2026-06-04
Resolve lead SHA for squash/fast-forward merge projects

rhur-pixel:
## fix(omm): resolve lead SHA for squash/fast-forward merge projects
Follow up [APPSRE-13998](https://redhat.atlassian.net/browse/APPSRE-13998) 
### Problem

The OMM group processing in `_process_omm_group` relied exclusively on `lead.merge_commit_sha` to track the lead MR's resulting commit. On projects that use squash or fast-forward merges (like app-interface), GitLab never creates a merge commit — the resulting SHA only exists in `squash_commit_sha`, leaving `merge_commit_sha` permanently `None`.

This caused two issues in production:
1. **Before the SHA guard** (initial deploy): `repository_compare(current_head, None)` sent a null SHA to the GitLab API, returning 400 errors.
2. **After the SHA guard was added**: The `if not lead.merge_commit_sha` check fired every loop, logging `lead-missing-merge-sha` indefinitely. The retry never resolved because the field is structurally null — not a race condition.

## Fix
Resolve the lead SHA once at the top of the divergence-check block:

lead_sha = lead.merge_commit_sha or lead.squash_commit_sha
This covers both merge strategies:

Regular merge projects: merge_commit_sha is set, squash_commit_sha is null — resolves to merge_commit_sha
Squash/FF projects: merge_commit_sha is null, squash_commit_sha is set — resolves to squash_commit_sha
The guard for neither SHA being populated is retained as a safety net for the (unlikely) case where GitLab hasn't written either field yet.

## Changes

- reconcile/gitlab_housekeeping.py: Replace three bare lead.merge_commit_sha references with lead_sha (guard, head comparison, repository_compare call). Update warning tag to lead-missing-sha.
- reconcile/test/test_gitlab_housekeeping.py:
- Parametrize 6 existing _process_omm_group tests with ids=["merge-commit", "squash-commit"] so each runs against both SHA paths (12 test cases total).
- Update the missing-SHA guard test to set both fields to None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced commit validation to accept alternative SHA formats with automatic retry when neither is available
  * Improved branch divergence detection for more robust merge tracking

* **Tests**
  * Expanded test coverage to validate behavior across both merge and squash commit SHA scenarios

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -1069,10 +1069,11 @@ def _process_omm_group(
         clear_omm_group(dry_run, gl, lead=lead)
         return 0
 
-    if not lead.merge_commit_sha:
+    lead_sha = lead.squash_commit_sha or lead.merge_commit_sha
+    if not lead_sha:
         logging.warning([
             "omm-group",
-            "lead-missing-merge-sha",
+            "lead-missing-sha",
             gl.project.name,
             lead.iid,
             "will retry next loop",
@@ -1080,14 +1081,14 @@ def _process_omm_group(
         return 0
 
     current_head = gl.project.branches.get(lead.target_branch).commit["id"]
-    if current_head != lead.merge_commit_sha:
+    if current_head != lead_sha:
         # Head moved since the lead merged.  This is expected when pending
         # members merge (each advances the target).  Only invalidate if
         # the lead's commit is no longer reachable — meaning the branch
         # diverged (force push / external reset).
         result = cast(
             "dict",
-            gl.project.repository_compare(current_head, lead.merge_commit_sha),
+            gl.project.repository_compare(current_head, lead_sha),
         )
         if len(result["commits"]) > 0:
             logging.warning([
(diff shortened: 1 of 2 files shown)


TGPSKI (commented):
Clean fix, the core logic is correct. Two items to address:

1. **SHA resolution order** — `squash_commit_sha or merge_commit_sha` means squash takes priority. For projects using squash+merge (not FF), GitLab sets **both** fields: `squash_commit_sha` is the intermediate squashed commit and `merge_commit_sha` is the actual commit on the target branch. Would `merge_commit_sha or squash_commit_sha` be safer? It correctly handles all three cases (regular merge, squash+merge, squash+FF). Or does OMM never run against squash+merge projects?

2. **Missing type annotations** on the parametrized test params — will fail strict mypy. Add `merge_sha: str | None, squash_sha: str | None` to all 6 test functions (lines 2120, 2170, 2237, 2288, 2337, 2386).

Tests are well-structured — parametrizing both SHA paths is the right approach.

TGPSKI (approved):
Both items addressed — SHA resolution order corrected to `merge_commit_sha or squash_commit_sha`, type annotations added. LGTM.

TGPSKI (approved):
LGTM

PR #5593, opened 2026-06-09
feat(saasherder): fall back to IMAGE when REGISTRY_IMG is not set for…

jfreixa:
… digest resolution

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fallback logic for image registry configuration to gracefully handle missing values and use alternative sources when needed. Error messages now accurately reflect which configuration options are unavailable, enhancing troubleshooting clarity.

### DIFF
--- a/reconcile/utils/saasherder/saasherder.py
+++ b/reconcile/utils/saasherder/saasherder.py
@@ -955,9 +955,10 @@ class SaasHerder:
             if need_repo_digest or need_image_digest:
                 try:
                     logging.debug("Generating REPO_DIGEST.")
-                    registry_image = consolidated_parameters.get(
-                        "REGISTRY_IMG"
-                    ) or consolidated_parameters["IMAGE"]
+                    registry_image = (
+                        consolidated_parameters.get("REGISTRY_IMG")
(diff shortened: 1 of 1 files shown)


chassing (approved):

reconcile/utils/saasherder/saasherder.py
```diff
@@ -955,11 +955,14 @@ def _process_template(
             if need_repo_digest or need_image_digest:
                 try:
                     logging.debug("Generating REPO_DIGEST.")
-                    registry_image = consolidated_parameters["REGISTRY_IMG"]
+                    registry_image = (
+                        consolidated_parameters.get("REGISTRY_IMG")
+                        or consolidated_parameters["IMAGE"]
```
hemslo, line 960: Where is `IMAGE` defined and used? And for just checking if parameter contain some value or not, should explicitly check if `REGISTRY_IMG in consolidated_parameters` instead of using another try except.

jfreixa: yes... it is specific for some images... closing the pr

PR #5597, opened 2026-06-09
Added per-repo opt in config for multi-merge

rhur-pixel:
## Summary
- [APPSRE-14571](https://redhat.atlassian.net/browse/APPSRE-14571)
- Moves the multi_merge flag from a global (strategy-only) toggle to a per-repo opt-in. Repos must now explicitly set multi_merge: true in their gitlabHousekeeping config to participate in optimistic multi-merge, in addition to the Unleash strategy being active-cap-multi-merge.
- Adds multi_merge to the APPS_QUERY GraphQL query so the field is fetched from app-interface.
## Changes
- reconcile/gitlab_housekeeping.py: Removed the global multi_merge assignment before the repo loop. Moved it inside the per-repo config block, gating on both the rebase strategy and hk.get("multi_merge", False).
- reconcile/queries.py: Added multi_merge field to the housekeeping section of APPS_QUERY.
## Why
Currently, enabling the active-cap-multi-merge Unleash strategy turns on multi-merge for all repos at once. This makes rollout risky. With this change, we can enable the strategy globally but opt in repos individually via their app-interface config, allowing incremental rollout.


esron (approved):
LGTM

PR #5600, opened 2026-06-09
feat: support old_parameter_group resolution in RDS external resources

BumbleFeng:
## Summary
- Resolve `old_parameter_group` field references (via `ResourceValueResolver`) alongside `parameter_group` in the RDS factory
- Ensures the old parameter group data is properly fetched from app-interface before being passed to the er-aws-rds module

## Problem
During RDS major version upgrades, the old parameter group must be kept in Terraform state to avoid `InvalidDBParameterGroupState` errors. The er-aws-rds module now accepts `old_parameter_group`, but qontract-reconcile needs to resolve its references before passing the input.

## Test plan
- [x] `test_resolve_old_parameter_group` — verifies both parameter groups are resolved correctly

### DIFF
--- a/reconcile/external_resources/aws.py
+++ b/reconcile/external_resources/aws.py
@@ -163,6 +163,9 @@ class AWSRdsFactory(AWSDefaultResourceFactory):
         if "parameter_group" in data:
             pg_data = rvr._get_values(data["parameter_group"])
             data["parameter_group"] = pg_data
+        if data.get("old_parameter_group"):
+            old_pg_data = rvr._get_values(data["old_parameter_group"])
+            data["old_parameter_group"] = old_pg_data
         if (
             (blue_green_deployment := data.get("blue_green_deployment"))
             and (target := blue_green_deployment.get("target"))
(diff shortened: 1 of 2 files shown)


esron (approved):
LGTM

TGPSKI:
## PR #5600 (qontract-reconcile) -- Reconcile side

**Changes:**
- Adds 3 lines in `AWSRdsFactory.resolve` to resolve `old_parameter_group` references via `ResourceValueResolver`
- Adds one test

**Looks good:**
- Uses `data.get("old_parameter_group")` (truthy check) rather than `"old_parameter_group" in data` (key-only check), so it correctly skips when the value is `None` or absent. This addresses the CodeRabbit comment that was already resolved.
- The pattern mirrors the existing `parameter_group` resolution exactly.
- Test properly mocks `_get_values` with `side_effect` to return different results for each call and asserts call count.

**No issues found.** This is a straightforward 3-line change that follows the established pattern.

TGPSKI (approved):

hemslo (commented):
This is not needed, see https://github.com/app-sre/qontract-reconcile/pull/4929

PR #5602, opened 2026-06-09
fix(sapm): add soak buffer and align state fetch with saasherder

TGPSKI:
## Summary

- Adds a 6-hour buffer to the soak days requirement (only when `soak_days > 0`) so SAPM doesn't open promotion MRs that are barely past the threshold. This prevents the TOCTOU race where S3 state changes (check_in resets from redeployments) between SAPM's evaluation and the pipeline validation, causing immediate `pipeline-error` failures.
- Uses `pre_check_sha_exists=False` in publisher state fetch to always read fresh S3 state, aligning with saasherder's validation behavior.
- Adds per-publisher INFO-level logging with check_in timestamps and accumulated deltas when soak blocks a promotion, enabling fast diagnosis of future soak-related failures.

## Root Cause (APPSRE-14585)

SAPM was opening promotion MRs that immediately failed saasherder's pipeline validation because:
1. SAPM calculated accumulated soak at time T and the result was marginally above the threshold
2. Between T and the pipeline run (minutes to hours later), publishers were redeployed (e.g., rolling wave promotions) which reset their `check_in` timestamps in S3
3. Saasherder read fresh S3 state and computed less accumulated soak, failing the validation

This caused a recurring pattern of `pipeline-error`-labeled MRs clogging the merge queue.

## Test plan

- [x] All existing SAPM tests pass (80/80)
- [x] New test for buffer behavior (`test_single_publisher_soak_days_passed_without_buffer`)
- [x] Updated existing test to validate buffer threshold (`test_single_publisher_soak_days_passed_with_buffer`)
- [x] ruff lint + format pass
- [x] mypy type check passes

Closes APPSRE-14585

## Summary by CodeRabbit

* **Improvements**
  * Enhanced promotion validation by introducing a stricter soak-days gate with an added buffer requirement (6 hours), ensuring promotions are only approved after meeting both the soak period and buffer threshold.
  * Improved diagnostics with enhanced logging when deployment information is unavailable.

* **Tests**
  * Updated test coverage to validate both soak-days scenarios with and without buffer requirements.

### DIFF
--- a/reconcile/saas_auto_promotions_manager/publisher.py
+++ b/reconcile/saas_auto_promotions_manager/publisher.py
@@ -73,6 +73,7 @@ class Publisher:
                 sha=self.commit_sha,
                 channel=channel,
                 target_uid=self.uid,
+                pre_check_sha_exists=False,
             )
             if not (
                 promotion_data
--- a/reconcile/saas_auto_promotions_manager/subscriber.py
+++ b/reconcile/saas_auto_promotions_manager/subscriber.py
@@ -17,6 +17,7 @@ from reconcile.utils.datetime_util import utc_now
 from reconcile.utils.slo_document_manager import SLODocumentManager
 
 CONTENT_HASH_LENGTH = 32
+SOAK_DAYS_BUFFER = timedelta(hours=6)
 
 
 @dataclass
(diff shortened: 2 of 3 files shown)


fishi0x01:
We also see issues with soakDays gate in saasherder - outside of SAPM. I think there is a deeper issue in how we deal with the gate in both integrations.

Additionally: 

**The buffer doesn't fix the root cause**

  The race condition exists because publish_promotions resets check_in to now on failure. The 6h buffer shrinks the window, but if publishers fail more than 6h before CI validates, saasherder still sees < soakDays accumulated time. 

**The buffer silently breaks user-configured semantics**

  soakDays: 1 now effectively means 1d6h. The renamed test makes this concrete — test_single_publisher_soak_days_passed previously asserted that meeting soakDays triggers a promotion; the same setup now asserts it doesn't.

PR #5604, opened 2026-06-10
:arrow_up: upgrade clientele to 2.2.2 and revert async client workaround

chassing:
## Summary

- Upgrade clientele from 2.2.1 to 2.2.2
- Revert the `_async_client = None` workaround from #5599

## Context

The clientele 2.2.1 migration (#5582) introduced a `RuntimeError: Event loop is closed` crash in all API-based integrations (`github-owners-api`, `glitchtip-api`, `glitchtip-project-alerts-api`, `slack-usergroups-api`).

**Root cause:** clientele's httpx2 backend uses a singleton `AsyncClient`. Integrations run in a `while True` reconcile loop (`run_integration.py:238`), where each iteration calls `asyncio.run()` — creating and closing an event loop. After the first iteration, the singleton still held TCP connections bound to the dead event loop. On the next iteration, httpcore2 tried to clean up those stale connections during a new request and hit `RuntimeError: Event loop is closed`.

**Workaround attempts:**
- #5598 added `aclose()` in a `finally` block — but on the next loop iteration the closed-but-non-None client was reused → `RuntimeError: Cannot send a request, as the client has been closed.`
- #5599 added `_async_client = None` reset after close — worked, but required reaching into clientele's private internals

**Upstream fix:** clientele 2.2.2 fixes this properly by resetting the internal `_async_client` reference in `aclose()`, so `build_async_client()` creates a fresh client for the next event loop. This makes the private attribute hack from #5599 unnecessary while keeping the `aclose()` call from #5598 (still needed for clean connection teardown).

## Test plan

- [x] mypy passes
- [x] Runner tests pass
- [ ] Deploy and verify API integrations survive multiple reconcile loop iterations without `RuntimeError`

## Summary by CodeRabbit

* **Chores**
  * Updated clientele dependencies to version 2.2.2

* **Bug Fixes**
  * Improved shutdown/cleanup of the internal API client to ensure integrations close cleanly and avoid lingering background resources.

### DIFF
--- a/qontract_api_client/pyproject.toml
+++ b/qontract_api_client/pyproject.toml
@@ -5,11 +5,11 @@ description = "Auto-generated client for qontract-api"
 authors = [{ name = "Red Hat App-SRE Team", email = "sd-app-sre@redhat.com" }]
 license = { text = "Apache 2.0" }
 requires-python = ">=3.12"
-dependencies = ["clientele[core]==2.2.1"]
+dependencies = ["clientele[core]==2.2.2"]
 
 [dependency-groups]
 dev = [
-    "clientele[cli]==2.2.1",
+    "clientele[cli]==2.2.2",
     # development and testing tools
     "mypy==2.1.0",
     "pytest==9.0.3",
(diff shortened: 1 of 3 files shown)


hemslo (approved):
LGTM!

PR #5606, opened 2026-06-10
:bug: handle empty KV v2 engines in vault replication (APPSRE-14594)

chassing:
## Summary

- Catch `hvac.exceptions.InvalidPath` in `VaultClient._list_kv2()` and `_list()` and return an empty dict
- Prevents vault-replication from crashing when a newly created KV v2 engine has no secrets yet
- Triggered by empty `releng-stage/` engine created via app-interface MR !191856

## Test plan

- [x] Parametrized tests for `_list_kv2`, `list`, `list_all` with empty KV v2 engine
- [x] Integration test for `get_policy_secret_list` with empty engine
- [x] All existing vault tests pass
- [x] Linter + mypy clean

Closes: [APPSRE-14594](https://redhat.atlassian.net/browse/APPSRE-14594)

## Summary by CodeRabbit

* **Bug Fixes**
  * Listing from empty or invalid Vault KV v2 and non-KV-v2 paths now returns empty results instead of raising errors, improving robustness of secret discovery.
* **Tests**
  * Added tests covering empty/invalid Vault engine/listing scenarios to verify listing functions return empty results when paths are absent.

### DIFF
--- a/reconcile/test/test_vault_replication.py
+++ b/reconcile/test/test_vault_replication.py
@@ -567,6 +567,13 @@ def test_get_policy_secret_list(mocker: MockerFixture) -> None:
     }
 
 
+def test_get_policy_secret_list_empty_engine(mocker: MockerFixture) -> None:
+    vault_client = mocker.patch("reconcile.utils.vault.VaultClient", autospec=True)
+    vault_client.list_all.return_value = []
+
+    assert integ.get_policy_secret_list(vault_client, ["policy/path/1/*"]) == []
+
+
 @pytest.mark.parametrize(
     "paths",
     [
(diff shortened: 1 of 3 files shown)


fishi0x01 (approved):
❤️ 

Would it make sense to also use the except `InvalidPath` approach in `_list()` to be consistent in both?

chassing:
Thanks for the reviews!

@fishi0x01 — Good call, added `except InvalidPath` to `_list()` as well for consistency. Fixed in 7b98bafa.

@coderabbitai — Refactored the three tests into a shared `@pytest.fixture` + `@pytest.mark.parametrize`. Fixed in 7b98bafa.

PR #5608, opened 2026-06-10
Adaptive close fix

rhur-pixel:
## Problem

When `_process_omm_group` rebases an MR with `skip_ci=True`, GitLab creates a synthetic pipeline with `status: skipped`. On the next reconcile loop, this `skipped` pipeline appeared as the latest pipeline, and no existing branch handled it — so `any_active` was never set, causing premature `adaptive-close` of the OMM group.

This was the root cause of OMM being reverted to `active-cap` after testing.

## Changes

**Primary fix** — filter out `skipped` pipelines before any status logic, so the real pre-rebase `SUCCESS` pipeline drives merge decisions.

**Defensive empty-pipeline handling** — after filtering, if `pipelines` is now empty but the MR is rebased, mark it as active and wait for the next loop. Previously an empty pipeline list silently skipped the MR without setting
`any_active`.

**Catch-all for rebased MRs with unexpected statuses** — if an MR is rebased but has a pipeline status that isn't `SUCCESS`, `RUNNING`, or `PENDING` (e.g. `canceled`, `manual`), keep the group active rather than letting it fall through to `adaptive-close`. Logs `omm-group unhandled-status-rebased` for observability.

**Diagnostic logging** — added `omm-group unhandled-status-not-rebased` for MRs that hit none of the known status branches and are not rebased, making it easier to trace future `adaptive-close` events.

## Tests

3 new test cases covering:

- `skipped` pipeline filtered, pre-rebase `SUCCESS` pipeline drives merge
- All pipelines filtered out + MR is rebased → group stays active
- Unexpected status (`canceled`) on rebased MR → group stays active

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pipeline state evaluation to filter skipped pipelines and better handle edge cases.
  * Enhanced detection and logging of pipeline status conditions in pending member workflows.

* **Tests**
  * Added test coverage for skipped pipeline filtering scenarios.
  * Added test coverage for unhandled pipeline status cases.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -1130,7 +1130,7 @@ def _process_omm_group(
 
         # skip_ci=True rebase causes GitLab to create a "skipped" pipeline;
         # filter these out so the real (pre-rebase) pipeline drives decisions.
-        pipelines = [p for p in pipelines if p.status != "skipped"]
+        pipelines = [p for p in pipelines if p.status != PipelineStatus.SKIPPED]
 
         if not pipelines:
             if is_rebased(mr, gl):
@@ -1248,7 +1248,13 @@ def _process_omm_group(
             any_active = True
             continue
 
-        logging.info(["omm-group", "unhandled-status-not-rebased", gl.project.name, mr.iid, latest_status])
+        logging.info([
+            "omm-group",
+            "unhandled-status-not-rebased",
+            gl.project.name,
+            mr.iid,
+            latest_status,
+        ])
 
     if not any_active:
         logging.info(["omm-group", "adaptive-close", gl.project.name])
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -1128,7 +1128,19 @@ def _process_omm_group(
                     pipelines=timed_out,
                 )
 
+        # skip_ci=True rebase causes GitLab to create a "skipped" pipeline;
+        # filter these out so the real (pre-rebase) pipeline drives decisions.
+        pipelines = [p for p in pipelines if p.status != "skipped"]
+
         if not pipelines:
+            if is_rebased(mr, gl):
+                logging.info([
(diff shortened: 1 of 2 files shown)


TGPSKI:
LGTM — three changes, all clean.

**OMM adaptive-close fix**

Filtering `skipped` pipelines before status logic is the right fix. After filtering, `pipelines[0]` is the pre-rebase `SUCCESS` pipeline, and the existing `is_rebased + SUCCESS → merge` path fires correctly. The defensive empty-pipelines-after-filter path and the catch-all for unexpected statuses on rebased MRs both do the right thing (keep group alive, log for observability).

**Review queue: surface approved MRs with error labels**

Previously `is_good_to_merge` filtered these out silently. Now approved+error MRs show up in the review queue. Good add.

**One thing to verify**: the `runner.py` change removes the `HttpxHTTPBackend._async_client = None` workaround. Confirm clientele 2.2.2 handles the event-loop cleanup upstream — otherwise the `RuntimeError('Event loop is closed')` regression would come back on the next reconcile loop iteration.

reconcile/gitlab_housekeeping.py
```diff
@@ -1150,6 +1162,15 @@ def _process_omm_group(
...
                     any_active = True
+                else:
+                    logging.info([
+                        "omm-group",
+                        "unhandled-status-rebased",
+                        gl.project.name,
+                        mr.iid,
+                        latest_status,
+                    ])
+                    any_active = True
```
TGPSKI:
```suggestion
if is_rebased(mr, gl):
    if latest_status != PipelineStatus.SUCCESS:
        if latest_status not in {PipelineStatus.RUNNING, PipelineStatus.PENDING}:
            logging.info([...])
        any_active = True
        continue
```

TGPSKI: line selector not working properly for me but i hope you get the idea

TGPSKI (approved):

PR #5569, opened 2026-05-27
fix(gitlab-housekeeping): apply and clear merge-error on rebase state changes

rhur-pixel:
## Fix merge-error label lifecycle: apply on failed rebase, clear on successful rebase
Fixes [APPSRE-13994](https://redhat.atlassian.net/browse/APPSRE-13994)
### Problem
Two gaps in the `merge-error` label lifecycle were identified:
1. **merge-error not applied after failed rebase** — When an async rebase fails, the MR enters `cannot_be_merged` state but no error label was applied. The MR sat in the queue invisible to operators.
   - Example: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/188613
2. **merge-error not cleared after author rebase + passing pipeline** — After an author manually rebased and the pipeline passed, the label remained. The only removal path was a human comment, requiring manual intervention to unblock the MR.
   - Example: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/188950
### Changes
- `run_error_healthcheck` now **applies** `merge-error` to queue-eligible MRs stuck in `cannot_be_merged` / `cannot_be_merged_recheck` state (e.g. after a failed async rebase or unresolved conflicts).
- `run_error_healthcheck` now **removes** `merge-error` when the latest pipeline has succeeded AND the MR is rebased on top of the target branch — without requiring any human comment.
- The existing notes-based removal path (human comments after the label) is preserved as a fallback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined handling of merge-error labels for rebased merge requests with successful pipelines.
  * Improved label removal logic when new notes are posted to merge requests.

* **Tests**
  * Expanded test coverage for merge-error label application and removal based on pipeline status and merge request states.

[Review Change Stack](https://app.coderabbit.ai/change-stack/app-sre/qontract-reconcile/pull/5569?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #5612, opened 2026-06-11
Widen merge-error clearing to include system notes

rhur-pixel:
## Widen merge-error clearing to include system notes

Follow up to https://github.com/app-sre/qontract-reconcile/pull/5569 (reverted in #5577) — implements Di's [feedback](https://github.com/app-sre/qontract-reconcile/pull/5569#discussion_r3254432410) to simplify merge-error label clearing.

Fixes APPSRE-14372

### Problem

When `merge-error` is applied, the only way to clear it was a human comment on the MR. System notes from human actions (pushes, rebases, approval changes) were ignored because the filter required `not note.system`. This meant authors who pushed fixes had to also leave a comment to unblock their MR.

### Changes

Two changes in `run_error_healthcheck`:

1. **Remove `not note.system`** — human-authored system notes (e.g. "added 4 commits" from a push) now trigger label removal. The bot-author check (`!= gl.user.username`) is kept to filter out bot activity.

2. **Increase `per_page=1` to `per_page=5`** — bot "Build success" notes interleaved after a human push would mask the human's activity with only 1 note in the scan window. Scanning 5 notes ensures we find human activity underneath.

### Validated against production data

Verified on [MR 192296](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/192296) (pulp onboarding):
- Label added at `18:10:11` — no corresponding system note in `mr.notes.list` (label events are only in `resource_label_events`)
- Author pushed at `18:44:08` — system note with `author: "fehuang"`, `system: true`
- Bot rebased at `19:15+` — system note with `author: "devtools-bot"`, `system: true`
- Human comment at `19:03:09` triggered removal under old code

With this change, the author's push at `18:44:08` would have cleared the label automatically — 34 minutes earlier than the manual workaround.

## TODO
[Apply merge-error label to queue-eligible MRs stuck in cannot_be_merged state](https://redhat.atlassian.net/browse/APPSRE-14604)

## Release Notes

* **Bug Fixes**
  * Improved merge-error label cleanup to review the most recent merge request notes and only remove the label when newer, user-authored activity indicates it’s safe.

* **Tests**
  * Updated and expanded health check coverage to better exercise note-authorship and note-type scenarios, ensuring merge-error behavior matches expectations across new-note cycles.


rhur-pixel:
@hemslo Revisiting this, sorry for the delay. Would love to get your feedback on this 👍

reconcile/gitlab_housekeeping.py
```diff
@@ -1477,12 +1477,11 @@ def run_error_healthcheck(
...
             if merge_error_added_at:
                 latest_notes = mr.notes.list(
-                    order_by="created_at", sort="desc", per_page=1
+                    order_by="created_at", sort="desc", per_page=5
                 )
                 if any(
                     from_utc_iso_format(note.created_at)
                     > from_utc_iso_format(merge_error_added_at)
-                    and not note.system
                     and note.author["username"] != gl.user.username
```
hemslo: Suggest removing this line too. Any note newer than the `merge-error` label event is sufficient to clear the label — no need to filter by author.

Validated against MR 192296:
- `18:10:11` — `merge-error` added (no note in `mr.notes.list` — label events only appear in `resource_label_events`)
- `18:44:08` — Author push note → clears merge-error immediately
- No flapping risk: healthcheck skips `CANNOT_BE_MERGED` MRs, and `merge-error` is only re-applied on actual `GitlabMRClosedError`

The resulting condition becomes just:
```python
if any(
    from_utc_iso_format(note.created_at)
    > from_utc_iso_format(merge_error_added_at)
    for note in latest_notes
):
```

reconcile/gitlab_housekeeping.py
```diff
@@ -1477,12 +1477,11 @@ def run_error_healthcheck(
 
             if merge_error_added_at:
                 latest_notes = mr.notes.list(
-                    order_by="created_at", sort="desc", per_page=1
+                    order_by="created_at", sort="desc", per_page=5
```
hemslo: With the simpler "any note" approach, `per_page=1` is sufficient — we no longer need to look past bot notes to find human ones. The most recent note sorted `desc` by `created_at` is the only one that matters.

reconcile/test/test_gitlab_housekeeping.py
```diff
@@ -1717,6 +1717,7 @@ def test_system_note_does_not_trigger_recheck(
         system_note.created_at = "2025-06-01T11:00:01.0Z"
         system_note.system = True
         system_note.body = "added ~merge-error label"
+        system_note.author = {"username": gl.user.username}
```
hemslo: This test should be removed rather than patched. It tests a label-add system note, but in production label events don't create notes in `mr.notes.list` (confirmed by MR 192296 data). The negative case (note older than label → no removal) is already covered by `test_merge_error_label_not_removed_without_new_notes`.

reconcile/test/test_gitlab_housekeeping.py
```diff
@@ -1763,6 +1764,44 @@ def test_bot_note_does_not_clear_merge_error(project: Project) -> None:
     mocked_gl.add_label_to_merge_request.assert_not_called()
 
 
+def test_human_system_note_clears_merge_error(project: Project) -> None:
```
hemslo: With the "any note" approach, this test is redundant with the existing `test_merge_error_label_removed_on_new_notes` (line 1232) which already covers "note newer than label → removal". Suggest removing this and instead simplifying that existing test to drop the `.system` and `.author` setup from its mock note.

hemslo (commented):
Suggestion: simplify to "any note after merge-error clears the label" — drop the author check too.

Validated against MR 192296 production data. Three structural guards make this safe:
1. Label-add events don't create notes in `mr.notes.list` — no self-triggering
2. Healthcheck skips `CANNOT_BE_MERGED` MRs (line 1429-1433) — persistent issues never reach the note check
3. `merge-error` is only applied on actual `GitlabMRClosedError` (lines 1212, 1395) — not re-applied by healthcheck

Also: `test_bot_note_does_not_clear_merge_error` (line 1732) should be removed — bot notes should now clear the label too.

rhur-pixel:
Thank you for the review! I made changes to reflect all of your changes 👍

hemslo (approved):
LGTM!

PR #5614, opened 2026-06-12
Speed up CI tests with pytest-xdist and fix UnleashClient leak

chassing:
## Summary

- Add `pytest-xdist` and run tests with `-n auto --dist worksteal` to parallelize across all available CPU cores. Local benchmark: **130s → 68s (~48% faster)**. CI improvement expected from **277s → ~140s**.
- Fix `UnleashClient` background thread leak: the `reset_client` test fixture set the global reference to `None` without calling `destroy()`, leaving background threads that blocked process exit with 3 rounds of connection retry tracebacks after every test run.

## Test plan

- [x] Verified locally: same test results with and without xdist (31 pre-existing failures from missing binaries like `amtool`, `skopeo` — unchanged)
- [x] Unleash tests pass (15/15) and no more leaked tracebacks after pytest exits
- [x] Verify CI pipeline passes with parallel test execution

## Summary by CodeRabbit

* **Tests**
  * Improved test execution performance by enabling parallel test running.
  * Added development dependency to support parallel test infrastructure.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ helm-test: generate
 	git diff --exit-code helm openshift
 
 unittest: ## Run unit tests
-	uv run pytest --cov=reconcile --cov-report=term-missing --cov-report xml reconcile tools
+	uv run pytest -n auto --dist worksteal --cov=reconcile --cov-report=term-missing --cov-report xml reconcile tools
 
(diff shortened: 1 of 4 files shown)


jfreixa (approved):
Nice!

PR #5625, opened 2026-06-17
Fix S3 gateway VPC endpoint route table association for cluster VPCs

esron:
## Summary
- Fix `service_type` typo so S3 endpoints are created as Gateway endpoints instead of defaulting to Interface.
- Move `route_table_ids` out of `tags` and reference `module.<vpc>.private_route_table_ids` so private route tables get S3 prefix-list routes.
- Add a unit test covering the expected vpc-endpoints module configuration.

## Test plan
- [x] `uv run pytest reconcile/test/utils/test_terrascript_aws_client.py::test_populate_vpc_requests_s3_gateway_endpoint_route_tables -xvs`
- [ ] Re-run `terraform-vpc-resources` on an affected VPC and confirm Terraform plans route table associations for the S3 gateway endpoint

Made with [Cursor](https://cursor.com)

## Bug Fixes
* Fixed S3 VPC endpoint configuration parameter naming
* Updated route table references to use the correct source for private route tables
* Ensured proper tag assignment for S3 endpoints

## Tests
* Added test coverage for S3 VPC endpoint configuration, including service type validation, route table assignment, and endpoint naming

### DIFF
--- a/reconcile/test/utils/test_terrascript_aws_client.py
+++ b/reconcile/test/utils/test_terrascript_aws_client.py
@@ -13,6 +13,7 @@ from terrascript.resource import (
     aws_s3_bucket_policy,
 )
 
+from reconcile.gql_definitions.fragments.aws_vpc_request import VPCRequest
 from reconcile.terraform_tgw_attachments import (
     Accepter,
     ClusterAccountProviderInfo,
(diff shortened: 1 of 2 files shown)


RH-tj (approved):
thx for quick turn-around @esron

PR #5628, opened 2026-06-18
chore(lint): move type-only imports into TYPE_CHECKING blocks (TC001/TC002/TC003)

jfreixa:
## Summary

- Moves third-party (TC002), first-party (TC001), and stdlib (TC003) imports that are used exclusively in type annotations into `if TYPE_CHECKING:` blocks across ~300 files
- Adds `from __future__ import annotations` where needed so annotations remain deferred at runtime on Python 3.12
- Fixes three importers that were accessing symbols via module re-exports that were moved to TYPE_CHECKING: `Approver` (change_owners), `ExternalResourcesSettingsV1` (external_resources), and `Organization` (github_org test)

## Test plan

- [x] `uv run ruff check --no-fix` passes with no errors
- [x] `uv run mypy` passes with no issues
- [x] `uv run pytest reconcile/test/` — 3613 passed, 104 xfailed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -258,6 +258,11 @@ allow-star-arg-any = true
 
 [tool.ruff.lint.flake8-type-checking]
 runtime-evaluated-base-classes = ["pydantic.BaseModel"]
+runtime-evaluated-decorators = [
+    "dataclasses.dataclass",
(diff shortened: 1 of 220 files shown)


chassing (approved):
Nice!

PR #5633, opened 2026-06-19
fix: fail in case tekton timoeut invalid

mehfuz:
When a saas file has timeout set below 60 minutes (e.g. 30m0s), trigger-configs previously raised TektonTimeoutBadValueError as an uncaught exception inside threaded.run, producing a noisy traceback and making it hard to see which saas file/env failed.

This change catches that error in _trigger_tekton, logs a clear message including saas_file_name and env_name, and returns error=True so the integration exits with a non-zero status through the normal failure path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened Tekton pipeline timeout validation with stricter minimum duration requirements.
  * Improved error handling and logging for timeout configuration issues.

* **Refactor**
  * Optimized timeout validation logic for better maintainability.

### DIFF
--- a/reconcile/openshift_saas_deploy_trigger_base.py
+++ b/reconcile/openshift_saas_deploy_trigger_base.py
@@ -50,6 +50,18 @@ class TektonTimeoutBadValueError(Exception):
     pass
 
 
+def _validate_tekton_timeout(
+    saas_file_name: str, env_name: str, timeout: str | None
+) -> None:
+    if not timeout:
+        return
+    seconds = dhms_to_seconds(timeout)
+    if seconds < 3600:  # 1 hour
+        raise TektonTimeoutBadValueError(
+            f"[{saas_file_name}/{env_name}] timeout {timeout} is smaller than 60 minutes"
+        )
+
+
 @defer
 def run(
     dry_run: bool,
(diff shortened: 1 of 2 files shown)


chassing (approved):

reconcile/openshift_saas_deploy_trigger_base.py
```diff
@@ -267,19 +279,23 @@ def _trigger_tekton(
...
+            tkn_namespace_name,
+            integration,
+            integration_version,
+            saasherder.include_trigger_trace,
+            spec.reason,
+            spec.target_ref,
+        )
+    except TektonTimeoutBadValueError as e:
+        logging.error(str(e))
+        return True
```
chassing, line 298: Return `true`? This won't raise an error, or am I wrong?

mehfuz:
```
if spec.pipelines_provider.provider == Providers.TEKTON.value:
        error = _trigger_tekton(
            spec,
            dry_run,
            saasherder,
            oc_map,
            already_triggered,
            integration,
            integration_version,
        )
    else:
        error = True
        logging.error(
            f"[{saas_file_name}] unsupported provider: "
            + f"{spec.pipelines_provider.provider}"
        )

    return error
```

return true here is trigger failed not success

chassing: ok

PR #5638, opened 2026-06-22
Enforce minimum OMM group size of 3

rhur-pixel:
## Summary
Enforce minimum OMM group size of 3
Adds a minimum group size requirement for optimistic multi-merge (OMM) groups. A group is now only formed when there are at least 2 pending candidates in addition to the lead MR (total size >= 3).

This avoids creating an OMM group for a single pending MR where the overhead of group management (labeling, window tracking, CI checks) isn't justified over regular serial merging.

## Summary by CodeRabbit

* **Chores**
  * Enforced minimum group size requirement for merge request coordination workflows. The setup process now requires a lead plus at least two additional participants with pending merge requests before the workflow can begin.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -82,6 +82,7 @@ HOLD_LABELS = [
 ERROR_LABELS = [MERGE_ERROR, PIPELINE_ERROR]
 
 TENANT_LABEL_PREFIX = "tenant-"
+MIN_OMM_GROUP_SIZE = 3  # 1 lead + 2 pending MRs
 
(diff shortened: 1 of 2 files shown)


esron (approved):

TGPSKI:
What is the cost of "group management [overhead]"?

rhur-pixel:
@TGPSKI An OMM group with a single MR (MR A) would save some time for MR A, but the entire merge queue will have to wait for that singular MR to pass/fail. Also, MR A would be "lower priority" compared to other MR's in the merge queue most likely, having the minimum group size ensures we get meaningful speedup that justifies the switching-of-order for MR's in the merge queue. 

The minimum group size might not be necessary after dynamic grouping, however.

PR #5642, opened 2026-06-24
fix: filter post-rebase fork pipelines in OMM group to avoid CI wait

rhur-pixel:
### Problem

When `_process_omm_group` performs a `skip_ci=True` rebase on a fork-based MR, GitLab suppresses the MR-event pipeline on the target project but the fork repo's own CI fires a `push`-source pipeline on the new commit. Since `get_merge_request_pipelines` returns all pipelines (including fork ones), the code sees the fork's running pipeline and waits for it to complete (~8+ minutes per rebase), defeating the purpose of `skip_ci=True`.

Observed in production with MR 193109 (`jharting/app-interface` → `service/app-interface`): the skip-ci rebase at 14:15:42 triggered fork pipeline 16120963, causing the OMM group to wait for it instead of merging immediately on the pre-rebase SUCCESS result.

### Fix

Filter out fork pipelines whose SHA matches `mr.sha` (the current MR HEAD after rebase), **gated on the rebased state**. The filter only activates when `is_rebased` is True, so non-rebased MRs retain their current-HEAD fork pipeline for correct CI decisions.

```python
mr_is_rebased = is_rebased(mr, gl)

if mr_is_rebased:
    pipelines = [
        p for p in pipelines
        if not (p.project_id != gl.project.id and p.sha == mr.sha)
    ]
```

Without the gate, the filter would unconditionally remove the fork's current-SHA pipeline on the first OMM visit (before any rebase). This could permanently stall single-pipeline fork MRs (empty list, silent `continue`) or cause decisions to be driven by stale older pipelines instead of the current CI result.

The `is_rebased` result is computed once and reused in the later branching logic, replacing the two separate `is_rebased(mr, gl)` calls.

Non-fork MRs are completely unaffected since all their pipelines have `project_id == gl.project.id`.

### Testing

- `test_omm_group_fork_pipeline_post_rebase_filtered_merges` -- post-rebase fork pipeline is ignored, MR merges on pre-rebase SUCCESS
- `test_omm_group_fork_pipeline_running_old_sha_waits` -- fork pipeline with old SHA still running is kept, code correctly waits
- `test_omm_group_fork_pipeline_not_rebased_triggers_skip_ci` -- non-rebased fork MR's current-SHA pipeline is preserved and triggers skip-ci rebase

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -1099,18 +1099,23 @@ def _process_omm_group(
 
         # skip_ci=True rebase causes GitLab to create a "skipped" pipeline;
         # filter these out so the real (pre-rebase) pipeline drives decisions.
-        # Also filter post-rebase fork pipelines: a skip-ci rebase pushes a
-        # new commit to the fork, which can trigger CI there.  We keep fork
-        # pipelines for the *old* SHA (pre-rebase CI results) but discard
-        # those matching the current MR HEAD (post-rebase noise).
-        pipelines = [
-            p for p in pipelines
-            if p.status != PipelineStatus.SKIPPED
-            and not (p.project_id != gl.project.id and p.sha == mr.sha)
-        ]
+        pipelines = [p for p in pipelines if p.status != PipelineStatus.SKIPPED]
+
+        mr_is_rebased = is_rebased(mr, gl)
+
+        # After a skip-ci rebase the fork may start its own CI for the new
+        # commit.  Filter those post-rebase fork pipelines only when the MR
+        # is already rebased — for non-rebased MRs the current-HEAD fork
+        # pipeline is the real CI result we need.
+        if mr_is_rebased:
+            pipelines = [
+                p
+                for p in pipelines
+                if not (p.project_id != gl.project.id and p.sha == mr.sha)
+            ]
 
         if not pipelines:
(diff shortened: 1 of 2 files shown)


reconcile/gitlab_housekeeping.py
```diff
@@ -1140,6 +1153,14 @@ def _process_omm_group(
                         mr.iid,
                         latest_status,
                     ])
+                else:
+                    logging.info([
```
esron, line 1157: Those messages will go to `#sd-app-sre-reconcile` channel. If they are only used for debug I would make them a `logging.debug` message.

Otherwise, LGTM.

esron (approved):

reconcile/gitlab_housekeeping.py
rhur-pixel: All existing log messages in _process_omm_group (merge, skip-ci-rebase, eject-failed, etc.) are logging.info. These new messages are the same level and frequency. Can you clarify how info logs reach #sd-app-sre-reconcile? I checked #sd-app-sre-reconcile and don't see any gitlab-housekeeping logs there.

esron: Oh, sorry. You're right. I was thinking of `gitlab-members` and `gitlab-projects`.

PR #5644, opened 2026-06-25
fix(acs-rbac): strip extra fields from AccessScope namespaces to stop reconcile loop

hemslo:
## Summary

- ACS API started returning a `clusterId` field in `includedNamespaces` items; the reconciler built desired state without it, so current vs desired always differed → perpetual update loop in `appsrep09ue1`
- Replace `list[dict[str, str]]` with a typed `AccessScopeNamespace` Pydantic model using `Field(alias=...)` for `clusterName`/`namespaceName` — the parser only captures known fields and silently drops any extras the API adds
- `model_dump(by_alias=True)` handles serialization back to camelCase for the ACS API payload

## Test plan

- [ ] `uv run pytest reconcile/test/test_acs_rbac.py -x -q` — all 10 pass
- [ ] `test_access_scope_strips_cluster_id_from_namespaces` regression test covers the exact scenario: API response with `clusterId` field is parsed without it
- [ ] `api_response_access_scopes` fixture now includes `clusterId` in namespace items to exercise stripping through the full current-state path

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of access-scope namespaces so namespace entries are stored and sent in the expected format.
  * Cluster identifiers are now removed from namespace entries when displaying or processing them, reducing inconsistent data.
  * Access-scope updates and creation now preserve namespace details more reliably.

### DIFF
--- a/reconcile/acs_rbac.py
+++ b/reconcile/acs_rbac.py
@@ -15,7 +15,7 @@ from qontract_utils.differ import (
 from reconcile.gql_definitions.acs.acs_rbac import OidcPermissionAcsV1
 from reconcile.gql_definitions.acs.acs_rbac import query as acs_rbac_query
 from reconcile.utils import gql
-from reconcile.utils.acs.rbac import AcsRbacApi, Group, RbacResources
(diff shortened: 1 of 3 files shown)


jfreixa (approved):
Seems good!

PR #5647, opened 2026-06-30
APPSRE-14695 fix(saasherder): add target identity and per-UID soak breakdown to gate-check logs

fishi0x01:
## Summary

- All log messages in `_validate_promotion()` are now prefixed with `[saas_file=.../target_uid=...]` so the failing target is immediately identifiable without manual S3 cross-referencing
- The publisher UID is now included in the success-check error message
- When the soak gate fails, a per-channel per-publisher-UID breakdown is logged showing each UID's `check_in` timestamp and individual soak contribution

## Motivation

Investigated a CI failure on app-interface MR !194721 where the log only showed:

```
SoakDays in publishers did not pass. So far accumulated soakDays is 3:58:39.252499,
but we have a soakDays setting of 1. We cannot proceed with this promotion.
```

With no indication of which target or channel was failing, root-cause analysis required manual S3 inspection across 57+ publisher UIDs. The new output will look like:

```
[saas_file=rhobs-metric-collection-prod/target_uid=abc123...] SoakDays gate failed for commit f071f4b7.... Accumulated=3:58:39, required=1 day(s).
  channel=rhobs-metrics-cardinality-check-passed-prod-wave1 uid=8866fd... check_in=2026-06-29 18:37:56+00:00 contribution=3:58:38
```

Relates to APPSRE-14695.

## Test plan

- [ ] Existing saasherder unit tests pass
- [ ] Manually verify log output format in a dry-run promotion

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved promotion validation messaging to clearly reference the affected target when commits are blocked, parent configuration is missing, or newer parent configuration is detected.
  * Enhanced soak-gate failure errors to report accumulated versus required soak days, including a detailed per-check breakdown when applicable.
  * Updated publication validation to produce more complete, actionable diagnostics when publication status can’t be confirmed.

### DIFF
--- a/reconcile/utils/saasherder/saasherder.py
+++ b/reconcile/utils/saasherder/saasherder.py
@@ -2117,8 +2117,12 @@ class SaasHerder:
         if not promotion.subscribe:
             return True
 
+        target_id = (
+            f"[saas_file={promotion.saas_file}/target_uid={promotion.saas_target_uid}]"
+        )
+
         if promotion.commit_sha in self.blocked_versions.get(promotion.url, set()):
-            logging.error(f"Commit {promotion.commit_sha} is blocked!")
+            logging.error(f"{target_id} Commit {promotion.commit_sha} is blocked!")
             return False
 
         # hotfix must run before further gates are evaluated to override them
@@ -2127,6 +2131,9 @@ class SaasHerder:
 
         now = utc_now()
         passed_soak_days = timedelta(days=0)
+        # Track (channel_name, publisher_uid, check_in, soak_contribution) for
+        # detailed logging when the soak gate fails.
+        soak_log: list[tuple[str, str, str | None, timedelta | None]] = []
 
         for channel in promotion.subscribe:
             config_hashes: set[str] = set()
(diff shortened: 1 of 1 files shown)


hemslo (approved):
LGTM — clean, focused logging improvement. One minor note left as an inline comment.

reconcile/utils/saasherder/saasherder.py
```diff
@@ -2141,20 +2148,29 @@ def _validate_promotion(self, promotion: Promotion) -> bool:
                     deployment and (deployment.success or deployment.has_succeeded_once)
                 ):
                     logging.error(
-                        f"Commit {promotion.commit_sha} was not "
-                        + f"published with success to channel {channel.name}"
+                        f"{target_id} Commit {promotion.commit_sha} was not "
+                        f"published with success to channel {channel.name} "
+                        f"(publisher uid: {target_uid})"
                     )
```
hemslo, line 2154: nit: Embedding `\n` inside a single `logging.error()` call can cause issues with log aggregation tools (Splunk, CloudWatch, etc.) that treat each line as a separate entry. If that's a concern, you could join with `" | "` instead, or split into a `logging.error` for the summary + `logging.info` for the per-UID breakdown. Fine to leave as-is if your log infra handles multiline messages well.

fishi0x01: Good point. Fixed

chassing (approved):

PR #5648, opened 2026-06-30
Added rebase-error label and detection

rhur-pixel:
## Summary

[APPSRE-14604](https://redhat.atlassian.net/browse/APPSRE-14604) 

Adds `rebase-error` label detection to the gitlab-housekeeping error healthcheck. When GitLab's async rebase operation fails (e.g., triggered by OMM skip-ci rebases), the only signal is the `merge_error` field on the MR API response. Without detection, these MRs sit in the queue indefinitely — they appear healthy (`merge_status=can_be_merged`) but can never actually merge.

## Slack threads
https://redhat-internal.slack.com/archives/C098G63A9JQ/p1779352336285799?thread_ts=1779351458.851689&cid=C098G63A9JQ
https://redhat-internal.slack.com/archives/C098G63A9JQ/p1782433873580279?thread_ts=1782399601.804349&cid=C098G63A9JQ

## What it does

- **Healthcheck detection**: Applies `rebase-error` when `merge_error` contains "Rebase failed" AND `detailed_merge_status == "need_rebase"`. The dual-check prevents false positives from stale `merge_error` fields (a known GitLab quirk where the field persists after the author fixes the issue — [gitlab#33231](https://gitlab.com/gitlab-org/gitlab/-/issues/33231)).
- **Healthcheck removal**: Removes `rebase-error` when either `merge_error` clears or `detailed_merge_status` moves to a non-rebase state.
- **OMM ejection**: MRs with any error label (`merge-error`, `pipeline-error`, `rebase-error`) are ejected from OMM groups immediately, preventing the group from being stuck on an un-mergeable MR.
- **Rebase skip**: Error-labeled MRs are already filtered from rebase strategies via `preprocess_merge_requests`.

## Context

Discovered during IC triage of MR !193960 (`service/app-interface`). The MR had:
- `merge_status`: `can_be_merged`
- `detailed_merge_status`: `need_rebase`
- `merge_error`: `"Rebase failed: Rebase locally, resolve all conflicts, then push the branch."`

The OMM group was repeatedly calling `mr.rebase(skip_ci=True)` on it every cycle with no effect. The `rebase-error` label makes this state visible and prevents queue pollution.

## Design decisions

- **`detailed_merge_status` cross-check**: `merge_error` is set server-side by async rebase failures and is only reliably cleared by a subsequent rebase/merge API call — NOT by git push events. The `detailed_merge_status` field IS updated on push, so checking both fields prevents the label from sticking on MRs that have been manually rebased.
- **`continue` after applying label**: Skips pipeline-error check since it's irrelevant for an MR blocked on rebase. Pipeline-error will be evaluated on the next cycle when rebase-error is already present and the code falls through.

### `reconcile/utils/mr/labels.py`
- Added `REBASE_ERROR = "rebase-error"` constant.

### `reconcile/gitlab_housekeeping.py`
- Added `REBASE_ERROR` to `ERROR_LABELS`.
- `_process_omm_group`: New guard ejects pending MRs carrying any error label (removes `omm-pending`, increments rejection counter, skips pipeline/merge logic).
- `run_error_healthcheck`: New rebase-error block before pipeline-error check — applies label when `merge_error` contains "Rebase failed" AND `detailed_merge_status == "need_rebase"`, removes it when either condition clears.

### `reconcile/test/test_gitlab_housekeeping.py`
- Extended `_make_healthcheck_mr` factory with `merge_error` and `detailed_merge_status` params.
- 6 new tests: apply rebase-error, remove on clear, skip if already labeled, ignore non-rebase errors, remove when detailed status resolves, OMM group ejection (parametrized over all error labels).
- Expanded 3 existing parametrized tests to include `"rebase-error"`.

## TODO

- [ ] Update IC docs to include `rebase-error` label handling/triage procedures
- [ ] Refactor `_process_omm_group` — at 236 lines with 4+ nesting levels, it's accumulating too much responsibility and should be decomposed

Generated by Claude

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of merge requests with rebase-related failures during queue health checks.
  * Reconcile now adds or clears the **rebase-error** label automatically based on the MR’s rebase state.
  * Rebase-error (along with other error labels) is now rejected early in optimistic merge processing, avoiding unnecessary pipeline/merge attempts.
  * Updated test coverage to ensure rebase-error behavior and skip rules work consistently.


esron (approved):
LGTM

reconcile/gitlab_housekeeping.py
```diff
@@ -84,7 +85,7 @@
     NEEDS_REBASE,
 ]
```
hemslo, line 87: nit: Consider making this a `frozenset`:
```python
ERROR_LABELS = frozenset({MERGE_ERROR, PIPELINE_ERROR, REBASE_ERROR})
```
Membership tests become O(1), and it enables `.intersection()` below (see next comment).

reconcile/gitlab_housekeeping.py
```diff
@@ -1085,6 +1086,22 @@ def _process_omm_group(
...
+        if any(label in ERROR_LABELS for label in mr.labels):
+            logging.info([
+                "omm-group",
+                "eject-error-label",
+                gl.project.name,
+                mr.iid,
+                [l for l in mr.labels if l in ERROR_LABELS],
+            ])
+            if not dry_run:
+                gl.remove_label(mr, OMM_PENDING)
```
hemslo, line 1099: suggestion: If `ERROR_LABELS` is a `frozenset`, this block can use a single set intersection instead of iterating twice (`any()` + list comprehension):
```python
error_labels = ERROR_LABELS.intersection(mr.labels)
if error_labels:
    logging.info([
        "omm-group",
        "eject-error-label",
        gl.project.name,
        mr.iid,
        sorted(error_labels),
    ])
```

reconcile/gitlab_housekeeping.py
```diff
@@ -1085,6 +1086,22 @@ def _process_omm_group(
...
+            optimistic_merge_rejected.labels(
+                project_id=mr.target_project_id,
+                reason="error_label",
+            ).inc()
+            continue
```
hemslo, line 1104: question: When an error-labeled MR is ejected, `any_active` is not set to `True`. If *all* pending MRs in the group carry error labels, `any_active` stays `False` and `clear_omm_group` dissolves the entire group (lead included) at line 1257.

That seems reasonable — no point keeping a group alive with only broken MRs — but if a single MR has a transient `rebase-error` that clears next cycle, the group will have already been torn down and must reform from scratch. Is that intentional?

hemslo (commented):
Good work overall — clean, well-tested change. A couple of suggestions inline.

reconcile/gitlab_housekeeping.py
rhur-pixel: Hm, yes this was intentional but I believe you bring up a good point in that group dissolving is unnecessary if the issue is transient. 

I think a single-retry mechanism would be useful for these transient failures (no need to label transient failures): when the healthcheck sees an MR that already has `rebase-error `AND `rebase_failed` is still true, it attempts one `mr.rebase()` before keeping the label. If the retry succeeds, `merge_error` gets cleared by GitLab on the new rebase call, the label gets removed next cycle, and the group stays intact. If it fails again, the label sticks.

rhur-pixel:
@hemslo I addressed the comments, thank you for the feedback again!

reconcile/gitlab_housekeeping.py
hemslo: The retry-at-detection-time approach won't work here because `mr.rebase()` is async — GitLab accepts the request and returns success immediately. The actual failure only surfaces later in the `merge_error` field. So the try/except would almost always "succeed" and skip the label.

The current implementation also doesn't achieve "single retry" — since `mr.rebase()` rarely throws, the `elif rebase_failed and has_rebase_error` branch fires every cycle:

1. Cycle N: detect failure → apply `rebase-error`
2. Cycle N+1: label present + still failing → `mr.rebase()` → returns success (async) → `continue`
3. Cycle N+2: async rebase failed again → `merge_error` still set → retry again
4. ...repeats forever

For a permanently conflicting MR, this queues a new server-side async rebase every cycle indefinitely.

Suggestion: remove the retry branch entirely. The label makes the problem visible. When the underlying issue resolves (author pushes a fix, conflicting MR merges), `detailed_merge_status` changes, `rebase_failed` becomes false, and the existing `elif not rebase_failed and has_rebase_error` branch removes the label. The normal rebase flow picks the MR back up from there — no need for the healthcheck to actively retry.

rhur-pixel: Ahh you're right thank you! Will remove the retry branch.

hemslo (approved):
LGTM!

PR #5650, opened 2026-07-01
APPSRE-14363 chore(test): migrate testslide tests to pytest-mock

jfreixa:
## Summary

- Rewrites `test_terraform_vpc_peerings.py` (the `TestRun` class) and `test_terraform_vpc_peerings_build_desired_state.py` (4 testslide test classes) to use pytest-mock and `MagicMock` instead of testslide
- Removes `testslide==2.7.1` from the dev dependency group and updates `uv.lock`

## Why

testslide is incompatible with Python 3.14 (PEP 649 deferred annotation evaluation causes `RuntimeError: dictionary changed size during iteration` in `testslide/strict_mock.py:416` when iterating `__annotations__`). The library is effectively unmaintained (Red Hat Bugzilla #2343971). This is a prerequisite for the Python 3.14 upgrade (APPSRE-14363).

## Changes

- `testslide.TestCase` subclasses → standalone pytest functions with `@pytest.fixture` for shared setup
- `testslide.StrictMock(X)` → `MagicMock(spec=X)`
- `self.mock_constructor(module, "Class").to_return_value(mock)` → `mocker.patch.object(module, "Class", return_value=mock)`
- `self.mock_callable(obj, "method").to_return_value(val).and_assert_called_once()` → set `return_value`, then `assert_called_once()` after the call
- `.to_raise(exc)` → `.side_effect = exc`

## Test plan

- [ ] All 42 tests in the two files pass with Python 3.12 (`uv run pytest reconcile/test/test_terraform_vpc_peerings.py reconcile/test/test_terraform_vpc_peerings_build_desired_state.py`)
- [ ] `ruff check` passes on both files

## Summary by CodeRabbit

* **Chores**
  * Removed an unused development dependency from the project setup.
* **Tests**
  * Refactored Terraform VPC peering integration tests from class-based patterns to pytest module-level tests.
  * Refactored Terraform VPC peering desired-state tests to use pytest fixtures and dataclass-based shared mock state.
  * Strengthened assertions across success, dry-run, and failure paths, including recoverable vs unexpected error handling.

### DIFF
--- a/reconcile/test/test_terraform_vpc_peerings.py
+++ b/reconcile/test/test_terraform_vpc_peerings.py
@@ -385,7 +385,8 @@ def test_empty_run(mocker: MockerFixture) -> None:
     integ.run(True)
 
     mocked_logging.warning.assert_called_once_with(
-        "No participating AWS accounts found, consider disabling this integration, account name: None"
+        "No participating AWS accounts found, consider disabling this integration, account name: %s",
+        None,
     )
 
 
@@ -524,6 +525,10 @@ def test_run_all_fine(run_mocks: RunMocks) -> None:
     run_mocks.terraform.apply.assert_called_once()
     run_mocks.clusters.assert_called_once()
     run_mocks.settings.assert_called_once()
+    run_mocks.terrascript.populate_vpc_peerings.assert_called_once()
+    run_mocks.terrascript.populate_configs.assert_called_once()
+    run_mocks.terrascript.dump.assert_called_once()
+    run_mocks.terrascript.terraform_configurations.assert_called_once()
 
 
 def test_run_fail_state(run_mocks: RunMocks) -> None:
@@ -569,6 +574,7 @@ def test_run_dry_run_with_failures(run_mocks: RunMocks) -> None:
         integ.run(True, print_to_file=None, enable_deletion=False)
 
     run_mocks.terraform.plan.assert_not_called()
+    run_mocks.terraform.cleanup.assert_not_called()
     run_mocks.terraform.apply.assert_not_called()
     run_mocks.exit.assert_called_once_with(1)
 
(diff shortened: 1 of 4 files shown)


chassing (approved):

reconcile/test/test_terraform_vpc_peerings.py
```diff
@@ -388,194 +389,205 @@ def test_empty_run(mocker: MockerFixture) -> None:
...
+                    "account": {"name": "mesh_account"},
+                },
+                "accepter": {
+                    "account": {"name": "mesh_account"},
+                },
+            }
+        ],
+        error_code,
+    )
+    mocks.terrascript.populate_additional_providers.return_value = None
```
hemslo, line 511: The `populate_additional_providers` assertion was dropped in this migration.

The old `setUp` used:
```python
self.mock_callable(self.terrascript, "populate_additional_providers").for_call(
    "desired_account",
    [
        {"name": "mesh_account"},
        {"name": "all_clusters_account"},
        {"name": "mesh_account"},
        {"name": "all_clusters_account"},
    ],
).to_return_value(None).and_assert_called_once()
```

The new code only stubs the return value but never verifies the call happens with the right arguments. A regression in which account list is wired up, or a refactor that skips the call entirely, would not be caught.

In `test_run_all_fine` (and `test_run_dry_run`), add:
```python
run_mocks.terrascript.populate_additional_providers.assert_called_once_with(
    "desired_account",
    [
        {"name": "mesh_account"},
        {"name": "all_clusters_account"},
        {"name": "mesh_account"},
        {"name": "all_clusters_account"},
    ],
)
```
And in the failure/dry-run-with-failures tests, add `assert_not_called()` if the production code skips the call in those paths.

hemslo (approved):
The `populate_additional_providers` assertion has been fully restored — all five `test_run_*` functions now verify it is called exactly once with the correct arguments. Migration looks good.

PR #5653, opened 2026-07-02
APPSRE-14722 fix(openshift-rhcs-certs): trigger regeneration on certificate_format change

chassing:
## Summary

The `openshift-rhcs-certs` integration did not trigger certificate regeneration when
`certificate_format` changed (e.g. from PEM to PKCS12). It only regenerated when the
cert was missing in Vault or approaching the `auto_renew_threshold_days` expiry threshold.

This meant that changing `certificate_format: PKCS12` in a namespace config had no effect
until the next renewal cycle — which could be months away. The K8s Secret would be created
with the correct `type: Opaque` for PKCS12, but populated with old PEM data keys (`tls.crt`,
`tls.key`, `ca.crt`) instead of the expected PKCS12 keys (`keystore.pkcs12.b64`,
`truststore.pkcs12.b64`).

**Workaround before this fix:** manually deleting the Vault secret to force regeneration.

### Changes

- Added `cert_format_mismatch()` that detects when the Vault secret's keys don't match the
  desired `certificate_format` (PEM expects `tls.crt`, PKCS12 expects `keystore.pkcs12.b64`)
- Added the mismatch check to the regeneration condition in
  `fetch_openshift_resource_for_cert_resource()`

### Context

Discovered during IC shift on 2026-07-02 while investigating a tenant request in
[#sd-app-sre](https://redhat-internal.slack.com/archives/CCRND57FW/p1782988263405989).

JIRA: [APPSRE-14722](https://redhat.atlassian.net/browse/APPSRE-14722)

## Test plan

- [x] Parametrized unit test for `cert_format_mismatch()` covering all 4 format combinations
- [x] Integration test verifying only the format-mismatched cert triggers regeneration
- [x] All 13 existing + new tests pass
- [x] ruff lint clean
- [x] mypy strict clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved certificate reconciliation so it now detects when stored certificate data doesn’t match the expected format and regenerates it automatically.
  * Reduced cases where certificates could remain in the wrong key format, helping keep managed secrets consistent.

### DIFF
--- a/reconcile/openshift_rhcs_certs.py
+++ b/reconcile/openshift_rhcs_certs.py
@@ -156,6 +156,25 @@ def cert_expires_within_threshold(
     return False
 
 
+def cert_format_mismatch(
+    ns: NamespaceV1,
+    cert_resource: OpenshiftResourceRhcsCert,
+    vault_cert_secret: Mapping[str, Any],
+) -> bool:
+    """Check if the Vault secret's keys don't match the desired certificate_format."""
+    desired_format = get_certificate_format(cert_resource)
+    match desired_format:
+        case CertificateFormat.PEM:
+            is_mismatch = "tls.crt" not in vault_cert_secret
+        case CertificateFormat.PKCS12:
+            is_mismatch = "keystore.pkcs12.b64" not in vault_cert_secret
+    if is_mismatch:
+        logging.info(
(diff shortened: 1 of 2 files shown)


fishi0x01 (approved):

hemslo (approved):
LGTM!

PR #5655, opened 2026-07-06
feat(APPSRE-14727: add Y-stream channel switching)

devppratik:
## Summary

Add Y-stream channel switching to AUS for OCM's migration from flattened channel groups (`stable`, `fast`) to Y-stream versioned channels (`stable-4.Y`, `eus-4.Y`) starting with OpenShift 4.22 (ROSA-377).

## Problem

OCM is migrating from flattened channel groups (e.g. `stable`) to Y-stream versioned channels (e.g. `stable-4.20`). After migration, `available_upgrades` on a cluster is scoped to the current channel's Y-stream only. A cluster on `stable-4.20` will never see `4.21.x` versions in `available_upgrades` until its channel is switched to `stable-4.21`. Without this change, AUS cannot perform Y+1 minor version upgrades for any cluster on versioned channels.

## Approach

AUS owns the channel switch as part of its existing upgrade automation — no app-interface config changes needed for Y+1 upgrades.

The reconcile is two-phase: phase 1 switches the channel, phase 2 (next reconcile) creates the upgrade policy. This keeps reconcile loops fast and idempotent.

Y-stream channels act as an upper boundary, not a strict filter. Switching from `stable-4.20` to `stable-4.21` preserves all `4.20.z` patch upgrades. Soak days and `blocked_versions` continue to gate the actual upgrade policy creation in the next reconcile cycle.

Channel switches are non-disruptive metadata changes — no rolling restart, no workload impact — so they do not consume mutex/parallelism slots.

### available_channels from Versions API

The cluster list API intentionally omits `available_channels` (OCM-24287) — it is only populated on individual cluster GET. To avoid N+1 per-cluster API calls, we fetch `available_channels` from the **Versions API** (`/api/clusters_mgmt/v1/versions/{version_id}`) and cache results by version ID. Since `available_channels` is a property of the version (from CIS annotations), multiple clusters on the same version share one API call. 200 clusters on ~5 versions = 5 API calls instead of 200.

Channel switches are forward-only. `blocked_versions` serves as the opt-out mechanism. EUS channels skip odd minors (`eus-4.20` → `eus-4.22`).

**Decision flow:**
1. `is_y_stream_channel?` → No: skip
2. `target = next Y-stream` (e.g. stable-4.20 → stable-4.21)
3. `already on target?` → skip
4. `target in available_channels` (from versions API, cached)? → No: skip
5. Switch channel, emit gauge metric

## Changes

- `OCMClusterVersion` — new `available_channels` field from cluster API
- `OCMCluster` — `channel` field (now required, confirmed always present since May 2026) + helpers (`is_y_stream_channel`, `channel_type`, `target_y_stream_channel`)
- `upgrades.py` — `update_cluster_channel` (PATCH `{"channel": channel}` — new top-level field per ROSA-377, distinct from legacy `version.channel_group`)
- `metrics.py` — `AUSClusterChannelSwitchGauge`
- `base.py` — `ChannelSwitchAction`, `check_y_stream_channel_switch_needed`, `_get_available_channels` (versions API + cache), `CalculateDiffResult`, `act_channel_switches` (with `ExceptionGroup` error collection)
- `ocm_upgrade_scheduler.py` — wired channel switches into reconcile loop
- `ocm_addons_upgrade_scheduler_org.py` — updated for new `CalculateDiffResult` return type

## Breaking change note

`AUSClusterUpgradePolicyInfoMetric.channel` label now emits the versioned channel (e.g. `stable-4.20`) instead of the legacy channel group (`stable`). Dashboards or alerts filtering on `channel="stable"` will need updating.

## OCM API context

- **channel field**: Always present on all prod clusters since May 2026 (OCM team confirmed). Hard requirement — OCM defaults it if not specified. `channel_group` is deprecated.
- **available_channels in list API**: Intentionally set to nil in cluster list response (OCM-24287, May 2026). Filtering requires per-cluster AMS capability checks too expensive for list. Versions API returns it efficiently.
- **PATCH payload**: Uses `{"channel": channel}` (new top-level field), not `{"version": {"channel_group": ...}}` (legacy). Both are valid OCM API fields for their respective codepaths.

## Test plan

- [x] 26 channel switch tests: helpers, eligibility, versions API fallback, cache hit, HTTP error, act_channel_switches (dry-run, success gauge, error collection)
- [x] test_already_on_target_channel — verifies no-op when cluster already switched
- [x] Existing calculate_diff tests updated and passing
- [x] Full AUS test suite: 249 passed
- [x] mypy: no issues
- [x] Integration tested locally against OCM production API — channel switch candidate-4.21 to candidate-4.22 detected via versions API

JIRA: https://redhat.atlassian.net/browse/APPSRE-14727

## Summary by CodeRabbit

* **New Features**
  * Added automatic Y-stream channel switching for eligible upgrades, with validation against available channels (including a versions API fallback and caching).
  * Upgrade planning now returns both upgrade-policy changes and pending channel-switch actions, and applies channel transitions with per-switch error handling.
  * Added a new metric gauge to track successful channel transitions.
* **Bug Fixes**
  * Updated AUS metrics to report the cluster’s actual channel.
  * Improved default value handling when schema “default” entries are missing.
* **Tests**
  * Added and updated unit tests for Y-stream classification/decisioning, available-channel fallback/caching, switch execution (including dry-run and failure paths).


hemslo (commented):
## Review

Well-structured PR — clear two-phase reconcile design, proper mutex integration, and good test coverage. A few items:

### Missed update site

`reconcile/aus/ocm_addons_upgrade_scheduler_org.py:228` — `expose_org_upgrade_spec_metrics` still uses `cluster_upgrade_spec.cluster.version.channel_group` for the `channel` metric label. The identical line was updated in `reconcile/aus/base.py:329` to `cluster_upgrade_spec.cluster.effective_channel()`. For clusters with a Y-stream channel, the addon metrics will report the bare group (`stable`) while cluster metrics report the versioned channel (`stable-4.20`).

### Duplicated channel switch block

The 8-line channel-switch-check-then-append block is copy-pasted verbatim in two places within `calculate_diff()`: once in the `if not next_schedule:` early-continue path and once at the end as `elif not addon_id:`. Both do identical `check_y_stream_channel_switch_needed` → `verify_max_upgrades_should_skip` → append → `set_upgrading`. Extracting a small helper like `_try_channel_switch(...)` would eliminate the copy-paste risk.

### Suggestion: simplify channel switch guards

Consider removing the `version_blocked` and `any_y1_version_soaked` pre-checks from `check_y_stream_channel_switch_needed`. The channel switch is just a preparatory step to make Y+1 versions visible in `available_upgrades` — the existing upgrade path already gates on soak days and blocked versions in `upgradeable_version()` / `version_conditions_met()` before creating an upgrade policy. Double-gating at channel-switch time adds complexity without additional safety.

Simplified `check_y_stream_channel_switch_needed` would just check: `is_y_stream_channel()` → `target_y_stream_channel()` → same-channel guard → target in `available_channels` → return action. This also lets you delete `any_y1_version_soaked` entirely and drop the `version_data` parameter from the function.

hemslo (commented):
## Review (v2)

All previous feedback addressed — this version is clean. `available_channels` comes from the cluster response (no extra API call), channel switches are just a metadata change (no mutex/parallelism), and `_try_channel_switch` is now a simple three-line helper. See inline comments for v4 feedback.

hemslo (commented):
*(consolidated into v2 review above)*

devppratik:
Thanks @hemslo!
The review comments have been addressed

reconcile/utils/ocm/base.py
```diff
@@ -287,6 +290,38 @@ def minor_version(self) -> str:
...
+    def is_y_stream_channel(self) -> bool:
+        if not self.channel:
+            return False
+        parts = self.channel.rsplit("-", 1)
+        if len(parts) != 2:
+            return False
+        version_part = parts[1]
+        segments = version_part.split(".")
+        return len(segments) == 2 and all(s.isdigit() for s in segments)
+
```
hemslo, line 301: Per the [enhancement doc](https://github.com/openshift-online/rosa-enhancements/blob/main/enhancements/managed_openshift_supports_y%2B1.md), `channel` is populated on all clusters — new clusters default to `stable-{versionY}`, existing clusters get it via DB migration. `channel_group` is deprecated (removal within 1 year).

If the OCM migration is complete, `channel` is always present and `effective_channel()` is unnecessary — just use `cluster.channel` directly. Similarly `channel` could be `str` instead of `str | None`, and `is_y_stream_channel()` could be simplified since all clusters are on Y-stream channels post-migration.

Worth confirming with OCM team whether the migration is fully rolled out or if there's still a transition period where `channel` could be absent.

devppratik: Ack. Will confirm and update

devppratik: Confirmed here https://redhat-internal.slack.com/archives/C0A8S4L99C2/p1783514422931169 and updated the logic accordingly

reconcile/utils/ocm/upgrades.py
```diff
@@ -175,3 +175,17 @@ def get_version_gates(ocm_api: OCMBaseClient) -> list[OCMVersionGate]:
...
+#
+
+
+def update_cluster_channel(
+    ocm_api: OCMBaseClient, cluster_id: str, channel: str
+) -> None:
+    ocm_api.patch(
+        build_cluster_url(cluster_id),
+        {"version": {"channel_group": channel}},
+    )
```
hemslo, line 191: Per the JIRA and [enhancement doc](https://github.com/openshift-online/rosa-enhancements/blob/main/enhancements/managed_openshift_supports_y%2B1.md), the PATCH should use the top-level `channel` field:

```python
{"channel": channel}
```

not `{"version": {"channel_group": channel}}`. OCM stores `channel` directly, but interprets `channel_group` as a group name and tries to derive a channel from it — sending `channel_group: "stable-4.21"` would confuse that logic.

Please verify this locally against the OCM API before merging.

reconcile/aus/base.py
```diff
@@ -1151,14 +1153,68 @@ def _calculate_node_pool_diffs(
...
+        return None
+
+    target_channel = cluster.target_y_stream_channel()
+    if not target_channel:
+        return None
+
+    if cluster.channel == target_channel:
+        return None
+
+    if target_channel not in cluster.version.available_channels:
```
hemslo: This debug log will fire on every reconcile run for every Y-stream cluster in steady state (e.g. cluster on `stable-4.21`, `stable-4.22` not available yet). That's the normal case, not worth logging — consider removing it.

reconcile/aus/base.py
```diff
@@ -1315,6 +1378,43 @@ def act(
...
+        try:
+            update_cluster_channel(
+                ocm_api, switch.cluster.id, switch.to_channel
+            )
+        except HTTPError as e:
+            detail = e.response.text if e.response is not None else ""
+            logging.error(
+                f"{switch.cluster.name}: channel switch failed: {e}: {detail}"
+            )
+            continue
```
hemslo, line 1447: HTTPError is caught and swallowed — the integration run will report success even if channel switches failed. Use the established `ExceptionGroup` pattern (see `ocm_machine_pools.py:557`, `openshift_namespaces.py:193`): collect errors in a list, then `raise ExceptionGroup(...)` at the end.

Also please verify the full integration locally (dry-run against a real OCM env) before removing the hold label.

reconcile/aus/base.py
```diff
@@ -1151,14 +1153,89 @@ def _calculate_node_pool_diffs(
...
+    try:
+        data = ocm_api.get(
+            api_path=f"/api/clusters_mgmt/v1/clusters/{cluster.id}"
+        )
+        return data.get("version", {}).get("available_channels") or []
+    except Exception:
+        logging.warning(
+            f"failed to fetch available_channels for {cluster.name}"
+        )
+        return []
```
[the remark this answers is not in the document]
devppratik: Fixed. Now catches HTTPError specifically and includes response detail in the warning. Auth and unexpected failures will propagate.

reconcile/aus/base.py
devppratik: Removed.

reconcile/aus/base.py
devppratik: Fixed. act_channel_switches now collects errors in a list and raises ExceptionGroup("Channel switch errors", errors) at the end. Also verified the full integration locally (dry-run against OCM prod with appsret01ue1)

reconcile/utils/ocm/upgrades.py
devppratik: Fixed. Verified locally `update_cluster_channel` now uses {"channel": channel} at the top level. Also confirmed via OCM CLI that the channel field is top-level on the cluster object, not under version.

reconcile/aus/base.py
```diff
@@ -1151,14 +1153,90 @@ def _calculate_node_pool_diffs(
...
+
+    The cluster list API does not populate available_channels, so we
+    need a per-cluster GET when checking for Y-stream channel switches.
+    """
+    try:
+        data = ocm_api.get(
+            api_path=f"/api/clusters_mgmt/v1/clusters/{cluster.id}"
+        )
+        return data.get("version", {}).get("available_channels") or []
+    except HTTPError as e:
```
hemslo, line 1180: This fallback shouldn't be needed. `available_channels` is already on `OCMClusterVersion` and gets populated automatically when the cluster JSON is deserialized — the JIRA explicitly says "no extra API call." If the list API doesn't return it, verify with the OCM team rather than working around it here.

Removing this also simplifies `check_y_stream_channel_switch_needed` and `_try_channel_switch` — the `ocm_api` parameter can be dropped from both.

devppratik: I did confirm with OCM team that channel is always present on all prod clusters (migrated since May). 
The list API `(GET /api/clusters_mgmt/v1/clusters?search=...)` returns `"available_channels": []` in the JSON payload itself. So Pydantic correctly deserialises it to an empty list. The individual GET returns the actual values.
```
  # List API response (raw JSON)
  "version": { "available_channels": [], ... }

  # Individual GET response (raw JSON)
  "version": { "available_channels": ["candidate-4.21", "candidate-4.22", "fast-4.21", "fast-4.22"], ... }
```
`_fetch_available_channels` works around this API-level gap. Will follow up with OCM team on whether list API can include available_channels — if fixed upstream, will remove the fallback.

hemslo: if it's not available in list endpoint, then we may need to bring back that get `available_channels` from current version endpoint like `/api/clusters_mgmt/v1/versions/openshift-v4.16.50` and cache result during the integration run.

devppratik: Updated and pushed the changes
Verified locally in dry run mode that it is working
```
  [DEBUG] skipping cluster with no upcoming upgrade
  GET /api/clusters_mgmt/v1/versions/openshift-v4.21.24-candidate  → 200
  [INFO]  switching channel candidate-4.21 -> candidate-4.22
```

reconcile/aus/base.py
```diff
@@ -1151,14 +1153,101 @@ def _calculate_node_pool_diffs(
...
+def check_y_stream_channel_switch_needed(
+    spec: ClusterUpgradeSpec,
+    ocm_api: OCMBaseClient | None = None,
+    available_channels_cache: dict[str, list[str]] | None = None,
+) -> ChannelSwitchAction | None:
+    cluster = spec.cluster
+    if not cluster.is_y_stream_channel():
+        return None
+
+    target_channel = cluster.target_y_stream_channel()
```
hemslo, line 1198: Consider pulling the fallback out of this function so it stays pure — just `spec` + `available_channels` in, action out. Resolve channels in `_try_channel_switch` instead:

```python
def check_y_stream_channel_switch_needed(
    spec: ClusterUpgradeSpec,
    available_channels: list[str],
) -> ChannelSwitchAction | None:
    # no ocm_api, no cache — just check and return
    ...

def _try_channel_switch(spec, addon_id, channel_switches, ocm_api, available_channels_cache):
    if addon_id:
        return
    available_channels = (
        spec.cluster.version.available_channels
        or _get_available_channels(ocm_api, spec.cluster.version.id, available_channels_cache)
    )
    channel_switch = check_y_stream_channel_switch_needed(spec, available_channels)
    if channel_switch:
        channel_switches.append(channel_switch)
```

Removes `ocm_api` and `available_channels_cache` from `check_y_stream_channel_switch_needed`, making it easier to test (no mocks — just pass a list).

devppratik: Updated and Refactored

hemslo (approved):
LGTM!

PR #5665, opened 2026-07-09
:recycle: vault-replication: refactor

chassing:
## Summary

- Remove separate `vault_policies` GQL query — policy `rules` are now fetched inline via the `vault_instances` query (added `rules` field to the GQL fragment)
- `get_policy_paths()` takes a policy object directly instead of `(name, instance_name, query_data)` — eliminates the filtering loop and the extra GQL call
- `get_vault_credentials()` accepts a `SecretReaderBase` instead of internally calling `VaultClient.get_instance()` — makes it testable and removes the singleton dependency
- `run()` uses `create_secret_reader()` and passes it through, consistent with other integrations
- Deleted the now-unused `vault_policies` GQL definition module
- Adapted all tests to the new signatures

## Test plan

- [x] All 39 existing tests pass (`uv run pytest reconcile/test/test_vault_replication.py`)
- [x] `ruff check` clean
- [x] `mypy` clean

## Summary by CodeRabbit

* **New Features**
  * Vault replication now includes policy rules when displaying or syncing vault instances.
  * Replication flows now use the selected policy information directly, reducing extra lookups.

* **Bug Fixes**
  * Fixed missing address data in Vault replication policy records.
  * Improved credential handling for Vault replication to better support authentication-based access.


fishi0x01 (approved):
LGTM

PR #5666, opened 2026-07-10
refactor: remove dead jump host functionality

jfreixa:
## Summary

The jump host (SSH bastion tunnel) feature is confirmed dead code: the `--use-jump-host` CLI flag defaults to `False`, the Helm chart never sets it, and no integration's `extraArgs` in app-interface passes it. While 32 clusters in app-interface still have `jumpHost` data configured (pointing to `bastion.ci.int.devshift.net`), the code path that would use it is never activated in production.

This removes the feature across 4 commits:

- **CLI & integration modules**: removes the `--use-jump-host` Click option and its plumbing from `reconcile/cli.py` and ~35 integration `run()` functions, plus the now-unreachable `sshuttle-command` in `qontract-cli`
- **OC client / OCMap layer**: removes the `Jumphost` protocol, SSH tunnel creation, and `jumphost_*` fields from `oc.py`, `oc_map.py`, `oc_connection_parameters.py`
- **GraphQL queries, core module, dependencies**: removes the `CommonJumphostFields` fragment, the `jumphosts` query, `jumpHost` selections from all consuming queries (regenerated via `make gql-query-classes`), `reconcile/utils/jump_host.py`, and the `sshtunnel`/`paramiko` dependencies
- **Tests & fixtures**: removes `test_jump_host.py` and cleans up fixture data

Note: this only removes the qontract-reconcile code. The `jumpHost` data on clusters in app-interface (and the `ClusterJumpHost_v1` GraphQL schema type) are untouched — that's a separate app-interface cleanup if desired.

## Test plan

- [x] `ruff check` / `ruff format --check` pass across `reconcile/` and `tools/`
- [x] `mypy` passes with zero errors across 1191 source files
- [x] `make gql-query-classes` regenerates cleanly with no residual diff
- [x] Full `pytest` suite: 3686 passed (the only 11 failures are pre-existing environment issues — missing `promtool`/`terraform`/`sloth` binaries locally — reproduced identically on the base commit)

## Release Notes

* **Refactor**
  * Removed SSH jump-host tunneling support across cluster connectivity and reconciliation workflows.
  * Eliminated the `--use-jump-host` CLI option and related integration parameters; jump-host data is no longer requested or used.
* **Chores**
  * Updated project dependencies and type-check configuration by removing jump-host related packages.
* **Tests**
  * Refreshed fixtures and test expectations to match the updated connection and API shapes.


jfreixa:
Re https://github.com/app-sre/qontract-reconcile/pull/5666#discussion_r3557711035 — good catch, confirmed. `ssh` was only ever needed by the jump-host SSH tunnel implementation (`JumpHostSSH.get_ssh_base_cmd()`, now removed), which shelled out to the system `ssh` binary. No other code path invokes it. This actually applied to all 26 `@binary([..., "ssh", ...])` decorators in `cli.py`, not just `openshift_users` — fixed in 25a32de0.

chassing (approved):
RIP ❤️

fishi0x01 (approved):
🧹 ❤️

reconcile/utils/oc.py
```diff
@@ -330,7 +324,6 @@ def _init_old_without_types(
         cluster_name: str,
         server: str | None,
         token: str | None,
-        jh: Mapping[Any, Any] | None = None,
         settings: Mapping[Any, Any] | None = None,
```
hemslo: nit: `settings` is now dead in this method. Its only consumer was `SecretReader(settings=settings)` inside the removed `jh is not None` block. After this PR, the parameter is accepted and passed through by callers but never read in the method body.

`settings` is still legitimately used in `OC_Map` itself (line ~1748) for reading cluster tokens, so only *this* parameter is vestigial — not the broader plumbing. Worth removing while you're already in this code.

hemslo (commented):
Clean removal overall — one minor leftover noted inline.

jfreixa:
Re https://github.com/app-sre/qontract-reconcile/pull/5666#discussion_r3568164061 — confirmed and fixed in f38a8797. `settings` was dead through the whole `OCCli` → `OCNative` → `OC.__new__` chain (not just `_init_old_without_types`), so I removed it end-to-end along with its only caller (`OC_Map.init_oc_client`). `OC_Map.settings` itself is untouched since it's still used for reading cluster automation tokens.

PR #5668, opened 2026-07-10
fix(glhk): fix rebase-error detection in healthcheck and OMM formation

rhur-pixel:
## Summary

Fix two gaps in `rebase-error` label detection that allowed MRs with persistent rebase failures to escape labeling and poison OMM groups.

**Root cause**: The `run_error_healthcheck` relied on `detailed_merge_status` from the `.list()` endpoint, which [GitLab does not proactively update](https://docs.gitlab.com/ee/api/merge_requests/#merge-requests-list-response-notes) without `with_merge_status_recheck=true`. The field was stale, so the condition `detailed_merge_status == "need_rebase"` was always False — even when `merge_error` was set.

**Evidence**: During the July 8 OMM test, MR 196062 had `merge_error: "Rebase failed: Rebase locally, resolve all conflicts, then push the branch."` confirmed via `.get()`, yet the healthcheck never applied `rebase-error`. The MR poisoned 3 consecutive OMM groups for 29 minutes.

## Changes

1. **Healthcheck fix** (`run_error_healthcheck`): When `merge_error` contains "Rebase failed", do a targeted `.get(mr.iid)` to fetch fresh `detailed_merge_status` before deciding. This adds at most 1 API call per MR with a rebase failure (typically 0-2 MRs at any time). Wrapped in try/except so a transient API failure doesn't crash the healthcheck loop.

2. **Formation label** (`apply_omm_pending`): Apply `rebase-error` label when `GitlabMRRebaseError` is raised at group formation. Previously the handler removed `omm-pending` but didn't label, allowing the MR to be re-selected into the next group.

## Context

This is PR 1 of 3 in the OMM inline rebase-poll-merge plan:

- **PR 1 (this)**: Fix rebase-error detection
- **PR 2 (next)**: Rewrite `_process_omm_group` with inline rebase-poll-merge — eliminates the cascade, re-rebase race, and push-pipeline blockage
- **PR 3 (after PR 2 validated)**: Remove `MIN_OMM_GROUP_SIZE` to increase group formation frequency

## Test plan

- [x] `test_healthcheck_applies_rebase_error_on_merge_error_field` — stale list data, fresh `.get()` confirms `need_rebase` → label applied
- [x] `test_healthcheck_no_rebase_error_when_resolved_despite_stale_merge_error` — `merge_error` persists but `.get()` shows `mergeable` → no false positive
- [x] `test_healthcheck_skips_rebase_error_if_already_labeled` — no duplicate label
- [x] `test_healthcheck_removes_rebase_error_when_detailed_status_resolved` — label removed when condition clears
- [x] `test_apply_omm_pending_rebase_error_applies_label` — formation exception → `rebase-error` applied
- [x] Full suite: 140 passed

Generated by Claude

## Summary by CodeRabbit

* **Bug Fixes**
  * Merge requests that fail rebasing during merge-group/OMM pending processing are now consistently flagged with the rebase-error label.
  * If a rebase error is detected, the omm-pending label is cleared as part of recovery.
  * Rebase-error health checks now use up-to-date merge-request details to avoid incorrect labeling when earlier data is stale.
  * The rebase-error label is applied only when confirmed, preserved if refresh fails, and removed once the issue is resolved.


TGPSKI (approved):
## Review

**Verdict: Approve** — targeted fix for a real production gap, well-tested.

### What looks good

- Root cause is well-diagnosed: stale `detailed_merge_status` from the list endpoint meant the healthcheck never applied `rebase-error` even when `merge_error` was set (July 8 MR 196062 evidence is convincing).
- Healthcheck fix is surgical: only calls `.get(mr.iid)` when `merge_error` contains `"Rebase failed"` (low API cost, typically 0–2 MRs).
- `GitlabGetError` + `continue` correctly preserves existing label state on transient failures — good fix for the CodeRabbit finding in `c90ec3c`.
- Formation fix (`apply_omm_pending` → `rebase-error` on `GitlabMRRebaseError`) closes the re-selection gap immediately; defense-in-depth with the healthcheck path.
- Test coverage is solid: stale list → fresh `.get()`, false-positive guard, already-labeled skip, label removal, and formation exception path.

### Minor nits (non-blocking)

1. **Exception-path test** — still worth adding a test where `project.mergerequests.get` raises `GitlabGetError` and asserting no label add/remove (locks in the `continue` behavior).
2. **Wrapper consistency** — `gl.get_merge_request(mr.iid)` exists in `gitlab_api.py` and would be slightly more consistent than `gl.project.mergerequests.get(mr.iid)`.
3. **Optional hardening** — `with_merge_status_recheck=True` on the `.get()` call if you want belt-and-suspenders; plain `.get()` worked in the July 8 incident so not required.

CI is green. Good to merge.

reconcile/gitlab_housekeeping.py
```diff
@@ -1452,12 +1453,22 @@ def run_error_healthcheck(
...
+            try:
+                fresh = gl.project.mergerequests.get(mr.iid)
+            except gitlab.exceptions.GitlabGetError as e:
+                logging.warning([
+                    "error-healthcheck",
+                    "rebase-status-refresh-failed",
+                    gl.project.name,
+                    mr.iid,
+                    str(e),
+                ])
```
hemslo, line 1466: `continue` on `GitlabGetError` skips unrelated pipeline-error and merge-error healthchecks for that MR.

When the fresh `.get()` call fails with a transient API error (e.g. 500), `continue` skips the entire rest of the loop body — including pipeline-error checks (lines 1483-1507) and merge-error label cleanup (lines 1509-1538) that are completely independent of rebase status. A stale `pipeline-error` or `merge-error` label would persist indefinitely, blocking the MR from the merge queue.

Suggested fix: replace `continue` with `rebase_failed = False` and let the loop fall through.

reconcile/gitlab_housekeeping.py
```diff
@@ -1452,12 +1453,22 @@ def run_error_healthcheck(
...
         has_rebase_error = REBASE_ERROR in labels
         mr_merge_error = getattr(mr, "merge_error", None)
-        mr_detailed = getattr(mr, "detailed_merge_status", None)
-        rebase_failed = bool(
-            mr_merge_error
-            and "Rebase failed" in mr_merge_error
-            and mr_detailed == "need_rebase"
-        )
+        if mr_merge_error and "Rebase failed" in mr_merge_error:
+            try:
```
hemslo, line 1457: Nit: consider using `gl.get_merge_request(mr.iid)` instead of `gl.project.mergerequests.get(mr.iid)`. The wrapper is used elsewhere in this file (line 436) for the same operation. It's a trivial pass-through today, but staying consistent means this call site benefits if the wrapper gains caching/retry logic later.

reconcile/gitlab_housekeeping.py
rhur-pixel: Good catch — continue was too broad and skipped unrelated pipeline-error/merge-error checks unnecessarily.

However, `rebase_failed = False` isn't safe either: when the MR already has rebase-error, the elif not rebase_failed and has_rebase_error branch would incorrectly remove the label based on a failed API call rather than confirmed resolution.

Alternative: replaced `continue` with rebase_failed = has_rebase_error so that the labels is preserved and `pipeline-error` and `merge-error` checks still occur.

rhur-pixel:
**Updated: inverted healthcheck detection logic based on live API finding**

While validating the implementation, a live API test against `gitlab.cee.redhat.com` (2026-07-13) revealed that `merge_error` is **not returned** by GitLab's `.list()` endpoint — only by `.get()`. The field simply isn't in the response (53 keys returned, `merge_error` not among them; `detailed_merge_status` is present).

This means the previous implementation (triggering on `merge_error` from the list-sourced MR object) was still dead code — `getattr(mr, "merge_error", None)` always returned `None`, so the `.get()` confirmation call was never reached.

**Fix**: Inverted the detection path:
- **Trigger**: `detailed_merge_status == "need_rebase"` from `.list()` (confirmed present)
- **Confirm**: `.get(mr.iid)` → check `merge_error` contains `"Rebase failed"` (confirmed present on `.get()`)

This matches how the fields are actually distributed across endpoints:
| Field | `.list()` | `.get()` |
|-------|-----------|----------|
| `detailed_merge_status` | Present (possibly stale) | Present (fresh) |
| `merge_error` | **Absent** | Present |

The double-gate prevents false positives: a stale `detailed_merge_status` triggers the check, but the label is only applied if `.get()` confirms `merge_error` contains a rebase failure message.

Error handling (hemslo's feedback) is preserved: on `GitlabGetError`, `rebase_failed = has_rebase_error` keeps existing label state and doesn't skip pipeline-error/merge-error checks.

Tests updated to reflect real API behavior — `merge_error` is only set on the `.get()` mock, never on the list-sourced MR object. Full suite: 140 passed.

Docs reference: GitLab's [List project merge requests](https://docs.gitlab.com/api/merge_requests/#list-project-merge-requests) response does not include merge_error (only [Retrieve a merge request](https://docs.gitlab.com/api/merge_requests/#retrieve-a-merge-request) does). Both endpoints return detailed_merge_status. Confirmed via live API test against gitlab.cee.redhat.com (2026-07-13).

Written by Claude

hemslo (approved):
LGTM — both original findings addressed. One minor test gap noted inline.

reconcile/test/test_gitlab_housekeeping.py
```diff
@@ -1449,8 +1516,8 @@ def test_healthcheck_removes_rebase_error_when_detailed_status_resolved(
         project_merge_requests=[mr],
```
hemslo, line 1516: Nit: this test would also pass if the `except` handler used `continue` (the original bug from the first revision), since it only asserts that `add_label` and `remove_label` are not called.

Consider adding `mocked_gl.get_merge_request_pipelines.assert_called_once()` to verify that pipeline/merge-error checks still run after the `GitlabGetError` fallthrough.

rhur-pixel: Good catch — added `mocked_gl.get_merge_request_pipelines.assert_called_once()` in 29dc501. This distinguishes the correct fallthrough from a `continue` that would skip the pipeline/merge-error checks.

PR #5669, opened 2026-07-13
feat(signalfx_endpoint_monitoring): emit COO Probe alongside legacy Probe

fishi0x01:
## What

`signalfx_endpoint_monitoring` only ever built a single `Probe` body with `apiVersion: monitoring.coreos.com/v1`, unlike `blackbox_exporter_endpoint_monitoring`, which was already extended to also emit `monitoring.rhobs/v1` into `app-sre-observability-per-cluster` as part of the COO migration.

This meant `signalfx`-provider probes (e.g. `catchpoint-result-importer`) could never get a COO counterpart no matter what changed in app-interface, since there was no code path that creates one — confirmed via `coo-parity-check` (a new CronJob comparing scrape/rule parity between the legacy and COO Prometheus stacks, APPSRE-14333) reporting `catchpoint-result-importer` as permanently missing in COO on `appsrep11ue1`.

## How

Mirrors `blackbox_exporter_endpoint_monitoring`'s dual-emission pattern exactly:
- Build both the legacy `coreos_body` and a new `rhobs_body` targeting `app-sre-observability-per-cluster`
- Add `Probe.monitoring.rhobs` to `managed_types`
- Return both resources from `build_probe`

## Test plan

- [x] Updated `test_signalfx_probe_building` to assert 2 probes are built, with a new assertion block verifying the COO `rhobs` probe's `apiVersion`, target namespace, and that its `spec` matches the legacy probe's `spec`
- [x] Updated `test_signalfx_filling_desired_state` to parametrize over both probes (mirroring `test_blackbox_exporter_filling_desired_state`), verifying both the legacy and COO resource get added to the desired state with the correct namespace/resource_type
- [x] `uv run pytest reconcile/test/test_closedbox_endpoint_monitoring.py -v` — 11 passed
- [x] `uv run ruff check` / `uv run ruff format --check` — clean
- [x] `uv run mypy reconcile/signalfx_endpoint_monitoring.py` — no issues

## Summary by CodeRabbit

* **New Features**
  * Endpoint monitoring now provisions probes for both the existing SignalFx monitoring setup and the consolidated OpenShift observability system.
  * Probes are created in their appropriate namespaces while keeping existing scraping intervals and timeout behavior consistent.

* **Bug Fixes**
  * Improved reconciliation so all supported probe destinations are correctly tracked and managed during updates.

### DIFF
--- a/reconcile/signalfx_endpoint_monitoring.py
+++ b/reconcile/signalfx_endpoint_monitoring.py
@@ -17,6 +17,8 @@ QONTRACT_INTEGRATION_VERSION = make_semver(0, 1, 0)
 LOG = logging.getLogger(__name__)
 
 PROVIDER = "signalfx"
+COO_NAMESPACE_NAME = "app-sre-observability-per-cluster"
+MANAGED_TYPES = ["Probe.monitoring.coreos.com", "Probe.monitoring.rhobs"]
 
 
 def run(dry_run: bool, thread_pool_size: int, internal: bool) -> None:
@@ -29,6 +31,7 @@ def run(dry_run: bool, thread_pool_size: int, internal: bool) -> None:
             dry_run=dry_run,
             thread_pool_size=thread_pool_size,
             internal=internal,
+            managed_types=MANAGED_TYPES,
         )
     except Exception as e:
         LOG.error(e)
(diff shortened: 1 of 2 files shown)


chassing (approved):

hemslo (approved):
LGTM!

PR #5670, opened 2026-07-13
fix(ldap-users-api): remove deleted user from GABI signoffManagers too

fltataru:
## Summary
  - `remove_user_from_gabi` in `reconcile/ldap_users_api/mr_builder.py` only removed the offboarded user from `users[]`, never from `signoffManagers[]`
  - This left dangling `$ref`s pointing at deleted user files whenever the offboarded user was also a GABI signoff manager, which fails app-interface's reference
  validation and blocks the bot's automerge delete-user MR
  - Found while unblocking [!196496](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/196496) (delete user ahrabovs) in app-interface, where this hit exactly that case in 2 gabi-instance files

  ## Changes
  - Strip matching `$ref` entries from both `users` and `signoffManagers` when present
  - Guard on field presence (some fixtures/older files may omit `signoffManagers`) to avoid a KeyError
  - If a user was the sole signoff manager, the list is written out as `signoffManagers: []` rather than crashing or leaving it untouched — schema allows an empty 
  list (no minItems), though this may still warrant a manual follow-up to designate a replacement signoff manager

  ## Test plan
  - [x] Added `test_remove_user_from_gabi_sole_signoff_manager` covering the empty-list case
  - [x] Updated `GABI_YAML`/`GABI_YAML_ALICE_REMOVED` fixtures to include `signoffManagers`
  - [x] `pytest reconcile/test/ldap_users_api/` — 37 passed
  - [x] `ruff check` and `mypy` clean on changed files

## Summary by CodeRabbit

- **Bug Fixes**
  - Removing a user now also removes them from sign-off manager lists.
  - Sign-off manager fields are handled correctly, including when the removed user is the only entry or when the field is missing.

- **Tests**
  - Extended YAML fixtures and assertions to cover removal from both user and sign-off manager lists.
  - Added test cases for the sole sign-off manager scenario and for missing sign-off manager fields.

### DIFF
--- a/reconcile/ldap_users_api/mr_builder.py
+++ b/reconcile/ldap_users_api/mr_builder.py
@@ -24,19 +24,26 @@ FileSyncFileOp = FileSyncCreate | FileSyncDelete | FileSyncUpdate
 
 
 def remove_user_from_gabi(raw_yaml: str, username: str) -> str:
-    """Remove user from GABI users list.
+    """Remove user from GABI users and signoffManagers lists.
 
-    Removes entries from content["users"] where $ref contains the username.
+    Removes entries from content["users"] and content["signoffManagers"]
+    where $ref contains the username.
     """
     yaml = create_ruamel_instance(explicit_start=True)
     content = yaml.load(raw_yaml)
 
-    original_length = len(content["users"])
-    content["users"] = [
-        user for user in content["users"] if Path(user["$ref"]).stem != username
-    ]
+    changed = False
+    for field in ("users", "signoffManagers"):
+        if field not in content:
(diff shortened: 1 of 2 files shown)


fishi0x01 (approved):
LGTM. Solid, minimal fix that matches the file's existing conventions. Recommend adding one test for the missing-`signoffManagers`-key guard path before merging, since that's the specific scenario the guard was written for.

chassing (approved):

hemslo (approved):
LGTM!

PR #5677, opened 2026-07-15
feat(qontract-api): split celery queues and autoscale worker via KEDA

chassing:
## Summary
- Split the `qontract-api-worker` Celery queue into `qontract-api-prod` (dry_run=False) and `qontract-api-mr-check` (dry_run=True), each with its own worker Deployment, so a stuck task on one can't block the other (this is what happened during the openshift-namespaces-api prod rollout when an OpenShift cluster was unreachable).
- Added a KEDA `ScaledObject` + `TriggerAuthentication` per worker Deployment, scaling on Redis queue backlog (`listLength`) instead of a fixed replica count (min 3, max 15).
- Dry-run tasks now skip the distributed lock entirely in `_deduplication.py` - they never write anything, so there's no correctness reason to serialize them, and locking them would have defeated the point of autoscaling the mr-check pool.
- Kept the production lock key format stable (`task_lock:{name}:dry_run=false:{suffix}`) so old and new worker pods compute the identical key for the same resource during a rolling deploy.
- Companion app-interface MR wires up the SaaS file, secret template, and Prometheus alerts.

## Test plan
- [x] `uv run pytest` in `qontract_api/` - 505 passed
- [x] `uv run ruff check` / `ruff format --check` - clean
- [x] `uv run mypy` - clean
- [x] Verified queue name constants (`qontract_api/tasks/__init__.py`) match the `listName` values in `openshift/qontract-api.yaml`'s ScaledObjects

Jira: APPSRE-14860

## Summary by CodeRabbit

* **New Features**
  * Reconciliation tasks are now routed to dedicated production or merge-request-check queues based on the request’s `dry_run` flag.
  * Split worker processing into separate deployments with dedicated KEDA autoscaling for production vs merge-request checks.
* **Bug Fixes**
  * Dry-run executions now bypass distributed locking (and associated deduplication behavior), preventing interference with production runs.
* **Tests**
  * Expanded integration and unit tests to verify queue routing and dry-run deduplication/locking behavior.

### DIFF
--- a/qontract_api/qontract_api/tasks/_deduplication.py
+++ b/qontract_api/qontract_api/tasks/_deduplication.py
@@ -81,8 +81,14 @@ def deduplicated_task(
             if kwargs.get("dry_run", True):
                 return func(*args, **kwargs)
 
+            # Keep the pre-existing key format (with the now-constant
+            # dry_run=false segment) so that during a rolling deploy, old
+            # and new worker pods compute the identical lock key for the
+            # same production task - a differing format would let an old
+            # and a new pod both believe they hold the lock for the same
+            # resource.
             lock_key_suffix = lock_key_fn(*args, **kwargs)
-            lock_key = f"task_lock:{func.__name__}:{lock_key_suffix}"
+            lock_key = f"task_lock:{func.__name__}:dry_run=false:{lock_key_suffix}"
 
             logger.debug(
                 f"Attempting to acquire lock for task {func.__name__}",
--- a/qontract_api/qontract_api/tasks/_deduplication.py
+++ b/qontract_api/qontract_api/tasks/_deduplication.py
@@ -31,6 +31,11 @@ def deduplicated_task(
     - Redis/Valkey: Native lock with Lua scripts + watch-dog
     - Other backends: Backend-specific distributed locking
 
+    Dry-run tasks (kwarg `dry_run=True`) are read-only and never write
+    anything, so they bypass locking entirely and always execute. Locking
+    only applies to production (`dry_run=False`) runs, where concurrent
+    writes to the same resource must be prevented.
+
     Args:
         lock_key_fn: Function to generate lock key from task arguments.
                     Example: `lambda workspace, **kw: workspace`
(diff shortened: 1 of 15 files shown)


hemslo (commented):
Minor deploy concern — see inline comment.

qontract_api/qontract_api/tasks/_deduplication.py
```diff
@@ -70,12 +75,14 @@ def decorator(func: Callable[P, R]) -> Callable[P, R]:
...
         @wraps(func)
         def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
-            # Generate lock key from task arguments.
-            # Always include dry_run in the key so that dry-run and production
-            # tasks for the same resources do not block each other.
+            # Dry-run tasks never write anything, so there is no correctness
+            # reason to serialize them - skip locking entirely so MR-check
+            # traffic isn't throttled by unrelated in-flight dry runs.
+            if kwargs.get("dry_run", True):
+                return func(*args, **kwargs)
```
hemslo, line 82: **Deploy strategy note:** This changes the lock key format from `task_lock:{name}:dry_run={val}:{suffix}` to `task_lock:{name}:{suffix}`. During a rolling deploy where old and new workers coexist, an old worker could hold the old-format key while a new worker acquires the new-format key for the same resource, briefly allowing two concurrent production writes. The window is small (deploy duration) and the blast radius is one duplicate reconciliation, but a scale-down-then-up deploy (rather than rolling) would eliminate the risk entirely.

jfreixa (approved):
Nice!

qontract_api/qontract_api/tasks/_deduplication.py
chassing: Good catch — fixed in 5a545315. Kept the pre-existing `dry_run=false:` segment in the key so old and new pods compute the identical lock key for the same production task during a rolling deploy.

chassing:
@coderabbitai Addressed the nitpick in dcc46ffb — extended the "must match" comment to `QAPI_WORKER_PROD_CELERY_OPTS`/`QAPI_WORKER_MR_CHECK_CELERY_OPTS` as well.

hemslo (approved):
LGTM!

PR #5679, opened 2026-07-15
APPSRE-12007: Unify container registry mirroring

lisa:
The quay-mirror and gcp-image-mirror integrations duplicated the same tag sync algorithm (enumerate tags, filter, check existence, compare manifests, copy via skopeo, aggregate errors). The only differences were how mirror definitions were discovered (different GraphQL queries) and how push credentials were resolved (plain Vault read vs Vault read + base64 decode for GCP service account keys). Skopeo itself is completely destination-agnostic.

This change extracts the shared algorithm into a new container_registry_mirror module with a Protocol-based design: each destination type implements four methods (discover mirrors, resolve source credentials, resolve destination credentials, determine skip conditions) and the shared engine handles everything else.

The deep sync timing logic (control file timestamp, interval comparison, CLI override) was also duplicated across integrations and has been extracted into a standalone DeepSyncTimer class.

The gcp-image-mirror integration now uses InstrumentedSkopeo, closing a four-year instrumentation gap where Docker Hub pulls from GCP mirroring were invisible to Prometheus monitoring. The original gcr_mirror.py (June 2020) predated instrumentation (July 2021), and neither the 2021 instrumentation PR nor the 2025 gcp_image_mirror.py rewrite added it.

quay_mirror_org.py (out of scope for the full refactoring) has been updated to use DeepSyncTimer from the new module, removing its dependency on QuayMirror.check_compare_tags_elapsed_time.

## Summary by CodeRabbit

* **New Features**
  * Added a shared container image mirroring framework with a sync engine and pluggable GCP and Quay integrations.
  * Introduced periodic deep-sync throttling for mutable tags to reduce expensive manifest comparisons.
* **Documentation**
  * Added design and usage documentation for the mirroring framework.
* **Bug Fixes**
  * Prevents mirroring Docker Hub images into public Quay destinations.
  * Updated CLI wiring to use the new destination integrations (no CLI interface changes).
* **Refactor**
  * Removed legacy GCP/Quay mirroring modules in favor of the new framework.
* **Tests**
  * Expanded coverage for the sync engine, deep-sync timer, registry factory, and GCP/Quay discovery/credential handling.


hemslo:
## Suggestion: skopeo sync for v2

The Protocol + Engine design here is clean, but looking ahead there's an opportunity to simplify the engine significantly by leveraging [`skopeo sync`](https://developers.redhat.com/articles/2025/09/24/skopeo-unsung-hero-linux-container-tools) directly.

### The idea

Instead of the Python-side tag enumeration, manifest comparison, and per-tag `skopeo copy` loop, the engine could:

1. `skopeo list-tags` on both source and destination (one subprocess each)
2. Apply include/exclude filtering in Python (existing logic, unchanged)
3. Generate a YAML config — contents depend on sync mode:
   - **Fast sync** (default): YAML contains tags present at source but missing at destination (set diff), plus `latest` (which is expected to change frequently). `skopeo sync` copies missing tags directly and compares digests only for `latest`.
   - **Deep sync** (periodic, controlled by timer): YAML contains all filtered source tags. `skopeo sync` compares digests via `OptimizeDestinationImageAlreadyExists` and catches mutable tag drift.
4. `skopeo sync --src yaml --dest docker --scoped --keep-going` (one subprocess per credential group)

The `DeepSyncTimer` becomes a one-liner: it controls whether the YAML gets the diff or the full list. No Python manifest comparison code either way.

### What this eliminates

- `sretoolbox.container.Image` dependency entirely (tag iteration, `__contains__`, `__eq__`, `is_part_of`)
- The manifest comparison code path in the engine — `skopeo sync` compares digests natively via `OptimizeDestinationImageAlreadyExists` ([source](https://github.com/containers/skopeo/blob/main/cmd/skopeo/sync.go))
- Per-tag `skopeo copy` subprocess overhead — one `skopeo sync` handles the entire batch
- The response cache / session / `InstrumentedImage` concerns from the old Quay mirror

### The `--scoped` flag solves the naming problem

Currently 11 mirrors have renamed destinations (e.g., `docker.io/prom/pushgateway` → `prom-pushgateway`). `skopeo sync` derives destination paths from source paths, so arbitrary renames don't work.

`--scoped` fixes this by using the full source path as the destination: `quay.io/app-sre/docker.io/prom/pushgateway`. This is valid per OCI spec and gives collision-free, deterministic naming.

The 11 renamed mirrors that would need migration:
- `docker.io/caarlos0/domain_exporter` (currently `domain-exporter`)
- `docker.io/cloudability/metrics-agent` (currently `cloudability-metrics-agent`)
- `docker.io/dynatrace/dynatrace-synthetic-vuc` (currently `synthetic-vuc`)
- `docker.io/dynatrace/dynatrace-synthetic-vuc-worker` (currently `synthetic-vuc-worker`)
- `docker.io/lablabs/cloudflare_exporter` (currently `cloudflare-exporter`)
- `docker.io/prom/cloudwatch-exporter` (currently `prom-cloudwatch-exporter`)
- `docker.io/prom/pushgateway` (currently `prom-pushgateway`)
- `quay.io/prometheus/blackbox-exporter` (currently `prom-blackbox-exporter`)
- `ghcr.io/cert-manager/cert-manager-openshift-routes` (currently `openshift-cert-manager-routes`)
- `ghcr.io/zapier/prom-aggregation-gateway` (currently `aggregation-pushgateway`)
- `registry.access.redhat.com/ubi8/ubi-minimal` (currently `ubi8-ubi-minimal`)

### Also: 12 mirrors can be deleted now

Cross-referencing mirror definitions against [container-images.md](https://gitlab.cee.redhat.com/service/app-interface-output/-/blob/master/container-images.md), these mirrors have no matching image in any cluster deployment:

| Source | App |
|--------|-----|
| `docker.io/golangci/golangci-lint` | ocm |
| `docker.io/securego/gosec` | ocm/ocm-addons-service |
| `docker.io/grafana/k6` | unleash |
| `docker.io/docker/dockerfile` | clair |
| `docker.io/moby/buildkit` | clair |
| `docker.io/envoyproxy/envoy` | app-sre |
| `docker.io/adorsys/keycloak-config-cli` | app-interface/vault-manager |
| `public.ecr.aws/dynatrace/dynatrace-activegate` | fedramp-tools |
| `public.ecr.aws/dynatrace/dynatrace-oneagent` | fedramp-tools |
| `docker.io/minio/minio` | insights/cicd |
| `docker.io/unleashorg/unleash-server` | insights/cicd |
| `docker.io/library/eclipse-mosquitto` | insights/cloud-connector |

Deleting these + migrating the 11 renames to `--scoped` naming would allow a v2 engine that's just: discover mirrors → list-tags both sides → filter & diff → generate YAML → `skopeo sync`. No custom Python sync logic needed.

reconcile/container_registry_mirror/gcp.py
```diff
@@ -0,0 +1,201 @@
...
+QONTRACT_INTEGRATION = "gcp-image-mirror"
+
+
+def run(dry_run: bool) -> None:
+    """Module-level entry point called by the integration framework."""
+    impl = GcpMirror()
+    specs = impl.discover_mirrors()
+    engine = MirrorEngine(
+        skopeo=Skopeo(dry_run),
+        dry_run=dry_run,
```
esron, line 209: I think we are skipping the `deep_sync` check. In this module `is_deep_sync` would never be True. If this is the intended behavior we should  remove the code that checks for `deep_sync`. Let me know if I'm wrong.

lisa: Good spot. Should be in 0ad4f568 now.

esron (approved):
LGTM.
I would wait for Di's approval before merging.

reconcile/container_registry_mirror/engine.py
```diff
@@ -0,0 +1,185 @@
...
+            self.is_deep_sync = deep_sync_timer.should_run
+        else:
+            self.is_deep_sync = is_deep_sync
+
+    def _build_images(self, spec: MirrorSpec) -> tuple[Any, Any]:
+        """Build source and destination Image objects for a spec.
+        Exists as a separate method so tests can patch it without
+        needing to mock the Image constructor import path."""
+        from sretoolbox.container import Image  # noqa: PLC0415
+
```
hemslo, line 66: The old Quay mirror used `InstrumentedImage` (counting `_get_manifest` calls via the `registry_reachouts` Prometheus counter) and a shared `response_cache` dict + `requests.Session` (connection pooling + manifest caching across repos within the same org).

Using plain `Image` here drops:
- `registry_reachouts` metric — dashboards tracking Docker Hub API call volume go blind
- `cache_hits`, `cache_misses`, `cache_size` metrics
- Connection reuse across specs (each `Image()` opens its own connections)

This becomes moot if v2 moves to `skopeo sync`, but worth noting for this iteration.

reconcile/container_registry_mirror/protocol.py
```diff
@@ -0,0 +1,52 @@
...
+    ) -> str | None:
+        """Fetch read credentials for the source registry and return
+        them in skopeo's "user:password" format. Returns None for
+        public sources requiring no authentication. Handles any
+        encoding specific to the credential storage (e.g., base64
+        for GCP service account keys)."""
+        ...
+
+    def resolve_destination_credentials(self, key: str) -> str:
+        """Fetch write credentials for the destination registry and
```
hemslo, line 26: Protocol declares `key: str` but `QuayMirror.resolve_destination_credentials` implements it as `key: Any` and actually receives an `OrgKey` namedtuple, not a string. GCP passes a URL string (matches). The protocol signature doesn't accurately describe the Quay implementation's contract.

Consider using a broader type (e.g., `key: Any` or `key: Hashable`) or making the Quay implementation extract a string key from the `OrgKey` before calling this method.

reconcile/container_registry_mirror/quay.py
```diff
@@ -0,0 +1,249 @@
...
+                    destination_url = f"{server_url}/{org}/{item['name']}"
+                    source_image = Image(mirror_url)
+                    if self.should_skip_mirror(
+                        source_registry=source_image.registry,
+                        source_url=mirror_url,
+                        destination_url=destination_url,
+                        destination_public=item.get("public"),
+                    ):
+                        _LOG.error(
+                            "Image %s cannot be mirrored to a public repository at %s",
```
hemslo: Behavioral change: the old code called `sys.exit(ExitCodes.ERROR)` here, aborting the entire integration run when a docker.io image was configured to mirror to a public Quay repo. The new code logs and skips.

This is arguably better (one bad config doesn't block all mirrors), but it's a policy change — operators who relied on the hard-fail to catch misconfigurations in CI/staging will now get a log line instead of a crash.

reconcile/container_registry_mirror/engine.py
hemslo: please cover newly added `image_class` usage in tests

reconcile/container_registry_mirror/quay.py
lisa: This one is tricky because, yes, the current behaviour is to `exit`, but that will cause build failures for unrelated changes. I don't know that we have recently had a need to abort like we did back in the day, maybe we don't have as many people trying to mirror Dockerhub to a public quay.io registry any longer; a cursory look at git history didn't show any obvious recent examples. I also figured that if someone wasn't getting their expected mirror that they would have a chatbot parse through the logs for them, and the updated message is more likely to be captured.

hemslo (approved):
All prior review findings have been addressed:

- `MirrorEngine` now supports `image_class`, `response_cache`, `session` injection, and `quay.py`'s `run()` wires in `InstrumentedImage` + shared cache/session (restoring the observability/performance parity with the old Quay mirror), closed properly in a `finally` block.
- `protocol.py`'s `resolve_destination_credentials` now uses `key: Any`, matching both the Quay (`OrgKey`) and GCP (`str`) implementations.
- The docker.io→public-repo skip now logs a clear `[MIRROR_BLOCKED]` message instead of silently changing behavior.
- Solid new test coverage: `TestImageClassParameter`, `TestResponseCacheParameter`, `TestSessionParameter`, `TestDeepSyncTimerIntegration`, `TestTimerRecordOnSuccess`, `TestDiscoverMirrorsSkipsPublicDockerHub`.

One request: could you file a Jira ticket to track the `skopeo sync`-native v2 approach discussed in the comments above, to be picked up when these integrations migrate to qontract-api? Summary of that direction: `skopeo list-tags` on both source/dest → diff for fast sync (plus `latest` always included) or full list for periodic deep sync → generate YAML → `skopeo sync --scoped --keep-going`. This would eliminate the `sretoolbox.container.Image` dependency, the manifest comparison code path, and per-tag `skopeo copy` subprocess overhead — but requires migrating ~11 renamed mirror destinations to `--scoped` naming first (listed in the earlier comment), plus deleting the ~12 mirrors with no matching deployed image.

Approving — nice refactor, and thanks for iterating through the review rounds.

lisa:
@hemslo Thanks. I'll make a card for it.

PR #5681, opened 2026-07-16
revert: restore sshuttle-command and jump host cluster lookup in qont…

jfreixa:
…ract-cli

The jump host cleanup in #5666 removed sshuttle_command and the jumpHost/sshuttle column from `qontract-cli get clusters`. Restore them along with the supporting get_jumphosts() query and gql_definitions/jumphosts module they depend on.

## Summary by CodeRabbit

* **New Features**
  * Added jump host and cluster network information to cluster data.
  * Added `get sshuttle-command` to generate SSHuttle commands for matching jump hosts and clusters.
  * Cluster listings now include an SSHuttle command when a jump host is available.


fishi0x01 (approved):
Not a big fan of the queries.py, but since its restore of previous running code LGTM. Thank you for fast fix 🙇

PR #5682, opened 2026-07-16
fix(glhk): gate OMM group formation on lead tenant-label eligibility

rhur-pixel:
## Problem

When the first serially-merged MR (the OMM group lead) has no tenant labels, `merged_labels` is empty and `_form_omm_group` cannot detect overlap between the lead and candidates. Candidates get skip-ci rebased and merged without CI validating against the lead's changes, breaking the non-overlap safety guarantee.

## Why this matters

The skip-ci rebase at group formation means candidates' pipelines run against pre-lead code. When the pipeline succeeds and the candidate is rebased, `_process_omm_member` merges it — but no CI ever validated the candidate against the lead's changes. The tenant-label overlap check is the only mechanism compensating for the skipped CI, and it's ineffective when the lead contributes nothing to `merged_labels`.

## Fix

Gate OMM group formation on `is_eligible_for_optimistic_merge(mr)` so that a lead without tenant labels falls back to serial merge instead of forming a group with an empty overlap set. This is consistent with how candidates are already handled — `is_eligible_for_optimistic_merge` in `_form_omm_group` already excludes candidates without tenant labels with the rationale "conservatively serialised."

## Testing

The existing `test_multi_merge_no_tenant_labels_falls_back_serial` test was rewritten to cover the lead case (lead without labels, candidate with labels) instead of the candidate case, which is already covered by the overlap and partial-overlap tests exercising `_form_omm_group`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Optimistic multi-merge grouping is now applied only to merge requests that meet optimistic-merge eligibility, avoiding attempts on ineligible merges.
  * Improved fallback behavior when tenant labels are missing, preventing unnecessary rebase actions and ensuring proper candidate filtering.
* **Tests**
  * Updated and added coverage to confirm ineligible merge requests are excluded from optimistic multi-merge candidate handling.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -1417,7 +1417,7 @@ def merge_merge_requests(
         merges += 1
 
         if rebase and merges == 1:
-            if multi_merge:
+            if multi_merge and is_eligible_for_optimistic_merge(mr):
                 candidates = _form_omm_group(
                     gl=gl,
                     merge_requests=merge_requests,
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -1426,6 +1426,9 @@ def merge_merge_requests(
                 if candidates:
                     apply_omm_group_lead(dry_run, gl, mr)
                     apply_omm_pending(dry_run, gl, candidates)
+            elif multi_merge:
+                # lead has no tenant labels — fall back to serial merge
+                logging.info(["omm-group", "lead-ineligible", gl.project.name, mr.iid])
             break
 
     merge_batch_size_histogram.labels(project_id=gl.project.id).observe(merges)
--- a/reconcile/test/test_gitlab_housekeeping.py
+++ b/reconcile/test/test_gitlab_housekeeping.py
@@ -2096,6 +2096,8 @@ def test_multi_merge_overlapping_tenants_serialized(
 def test_multi_merge_no_tenant_labels_falls_back_serial(
     mocker: MockerFixture,
 ) -> None:
+    # Lead has no tenant labels → ineligible for OMM, falls back to serial merge.
+    # Candidate has a tenant label but must not be rebased or merged.
     mr1 = _make_merge_mr(10, ["approved"])
     mr2 = _make_merge_mr(11, ["approved", "tenant-bar"])
     items = [_make_merge_item(mr1), _make_merge_item(mr2)]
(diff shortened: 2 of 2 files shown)


reconcile/gitlab_housekeeping.py
```diff
@@ -1417,7 +1417,7 @@ def merge_merge_requests(
         merges += 1
 
         if rebase and merges == 1:
-            if multi_merge:
+            if multi_merge and is_eligible_for_optimistic_merge(mr):
```
TGPSKI, line 1420: Clean fix — this makes the lead eligibility check symmetric with the candidate-side check already inside `_form_omm_group` (line 848).

Minor suggestion: consider adding a log line when the lead is ineligible (no tenant labels) so operators can see *why* OMM didn't form a group during a reconcile loop. Something like:

```python
if multi_merge and not is_eligible_for_optimistic_merge(mr):
    logging.info(["omm-group", "lead-ineligible", gl.project.name, mr.iid])
```

Currently, when the lead has no tenant labels, the code silently falls through to `break` without any signal in the logs. This could make debugging harder when an operator expects OMM to kick in but it doesn't.

reconcile/test/test_gitlab_housekeeping.py
```diff
@@ -2113,6 +2113,7 @@ def test_multi_merge_no_tenant_labels_falls_back_serial(
 
     mr1.merge.assert_called_once()
     mr2.merge.assert_not_called()
+    mr2.rebase.assert_not_called()
```
TGPSKI, line 2118: Good addition — this assertion verifies the critical safety property: when OMM group formation is skipped, no candidate gets a skip-ci rebase.

One concern: by rewriting this test to exercise the *lead*-without-labels path, we lose direct coverage of the *candidate*-without-labels path that was previously tested here (old mr2 had `["approved"]` with no tenant label). The PR description notes this is "already covered by the overlap and partial-overlap tests exercising `_form_omm_group`" — but those tests exercise overlapping tenant labels, not the absence of tenant labels on a candidate. `_form_omm_group`'s `is_eligible_for_optimistic_merge` filter for candidates isn't directly exercised by any remaining test.

Consider keeping a small dedicated test for the candidate case (lead has labels, one candidate doesn't) to ensure `_form_omm_group`'s eligibility filter is directly covered. This would also guard against a future regression if someone removes the `is_eligible_for_optimistic_merge` check from `_form_omm_group`.

reconcile/test/test_gitlab_housekeeping.py
```diff
@@ -2096,8 +2096,8 @@ def test_multi_merge_overlapping_tenants_serialized(
 def test_multi_merge_no_tenant_labels_falls_back_serial(
     mocker: MockerFixture,
 ) -> None:
-    mr1 = _make_merge_mr(10, ["approved", "tenant-foo"])
-    mr2 = _make_merge_mr(11, ["approved"])
+    mr1 = _make_merge_mr(10, ["approved"])
```
TGPSKI, line 2101: nit: The scenario is now inverted from the original — the *lead* (mr1) lacks tenant labels while the *candidate* (mr2) has them. A brief inline comment would help future readers understand the intent at a glance, e.g.:

```python
# Lead has no tenant labels → ineligible for OMM, falls back to serial merge.
# Candidate has a tenant label but must not be rebased or merged.
mr1 = _make_merge_mr(10, ["approved"])
mr2 = _make_merge_mr(11, ["approved", "tenant-bar"])
```

TGPSKI (commented):
**Overall: solid, well-scoped fix.** The one-line production change is correct and consistent with how candidates are already handled.

**What's good:**
- The fix correctly addresses the root cause — a lead MR without tenant labels makes `merged_labels` empty, bypassing the overlap safety net.
- The `is_eligible_for_optimistic_merge` reuse is clean; no new abstractions needed.
- The `mr2.rebase.assert_not_called()` test assertion is a valuable addition that directly verifies the safety property.

**Suggestions (non-blocking):**
1. **Observability**: Add a log line when the lead is deemed ineligible so operators can diagnose why OMM didn't activate.
2. **Test coverage**: The rewrite of `test_multi_merge_no_tenant_labels_falls_back_serial` trades candidate-without-labels coverage for lead-without-labels coverage. Consider adding a small separate test for the candidate case to avoid a coverage gap in `_form_omm_group`'s eligibility filter.
3. **Test readability**: A brief comment in the rewritten test clarifying the inverted scenario (lead without labels, candidate with labels) would help.

TGPSKI (approved):
All prior feedback addressed cleanly. The updates look great:

- **Observability**: The `elif multi_merge:` log branch is well-placed — it only fires when OMM is enabled but the lead lacks tenant labels, giving operators clear signal without noise.
- **Test coverage**: `test_multi_merge_candidate_no_tenant_labels_excluded` is a strong addition. Testing with 3 MRs (eligible lead, ineligible candidate, eligible candidate) validates both exclusion *and* inclusion in a single test, which also exercises `_form_omm_group`'s eligibility filter end-to-end.
- **Readability**: Inline comments in both tests make the intent immediately clear.

LGTM.

PR #5687, opened 2026-07-20
add optional gov slack username field q-r logic

suzana-nesic:
As the current code stands, users within the FedRAMP environment are not able to utilize the qontract-api slack-usergroups-api qontract-reconcile integration due to how GovSlack user handles are assigned compared to commercial Slack. The comparison differentiation comes from `qontract_api/qontract_api/slack/slack_workspace_client.py`:

```
        valid_org_names = {
            user.org_username for user in self.get_users().values() if not user.deleted
        }
```

As it stands, because the code is currently looking for and validating only `org_username`, a new field will be needed in-boundary to ensure users are seen as valid users within gov slack.

This pull request allows either the `org_username` field value or the `gov_slack_email_local_part` field value to be used in order to assign users to various slack user groups. If the optional `gov_slack_email_local_part` field is present within the user file, that value will be used instead of `org_username`.

Note the field `gov_slack_email_local_part` will only utilize the management of Slack user groups and not utilize services like PagerDuty, as it is not approved to use in-boundary.

qontract-schemas logic: https://github.com/app-sre/qontract-schemas/pull/1039

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving users’ government Slack usernames.
  * Slack identity matching now prefers the government Slack username when available, with fallback to the existing username.
  * User reconciliation across schedules, roles, ownership, and access now uses improved identity resolution.
* **Bug Fixes**
  * Ensured government Slack usernames are included in reconciliation data when available.
* **Chores**
  * Updated supporting data and test coverage for the new optional username field.

### DIFF
--- a/reconcile/slack_usergroups_api.py
+++ b/reconcile/slack_usergroups_api.py
@@ -89,16 +89,16 @@ DATE_FORMAT = "%Y-%m-%d %H:%M"
 
 class _UserWithSlackIdentity(Protocol):
     org_username: str
-    gov_slack_username: str | None
+    gov_slack_email_local_part: str | None
 
 
 def slack_username(user: _UserWithSlackIdentity) -> str:
     """Return the Slack identity for an app-interface user.
 
-    Prefer gov_slack_username when set (gov Slack email local-part), otherwise
+    Prefer gov_slack_email_local_part when set (gov Slack email local-part), otherwise
     fall back to org_username (commercial Slack / LDAP).
     """
-    return user.gov_slack_username or user.org_username
+    return user.gov_slack_email_local_part or user.org_username
 
 
 class SlackWorkspace(BaseModel, arbitrary_types_allowed=True):
--- a/reconcile/slack_usergroups_api.py
+++ b/reconcile/slack_usergroups_api.py
@@ -18,7 +18,7 @@ import logging
 import sys
 from collections import defaultdict
 from datetime import datetime
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Protocol
 
 from pydantic import BaseModel
 from qontract_api_client.client import (
@@ -85,6 +85,20 @@ if TYPE_CHECKING:
 QONTRACT_INTEGRATION = "slack-usergroups-api"
 INTEGRATION_VERSION = "0.1.0"
 DATE_FORMAT = "%Y-%m-%d %H:%M"
+
+
+class _UserWithSlackIdentity(Protocol):
+    org_username: str
+    gov_slack_username: str | None
+
+
+def slack_username(user: _UserWithSlackIdentity) -> str:
+    """Return the Slack identity for an app-interface user.
+
+    Prefer gov_slack_username when set (gov Slack email local-part), otherwise
+    fall back to org_username (commercial Slack / LDAP).
+    """
+    return user.gov_slack_username or user.org_username
 
 
 class SlackWorkspace(BaseModel, arbitrary_types_allowed=True):
@@ -210,7 +224,7 @@ class SlackUsergroupsIntegration(
             start = ensure_utc(datetime.strptime(entry.start, DATE_FORMAT))  # ruff: ignore[call-datetime-strptime-without-zone]
             end = ensure_utc(datetime.strptime(entry.end, DATE_FORMAT))  # ruff: ignore[call-datetime-strptime-without-zone]
             if start <= now <= end:
-                all_usernames.extend(u.org_username for u in entry.users)
+                all_usernames.extend(slack_username(u) for u in entry.users)
         return all_usernames
 
     @staticmethod
@@ -224,7 +238,9 @@ class SlackUsergroupsIntegration(
             List of usernames from roles
         """
 
-        return [user.org_username for role in roles or [] for user in role.users or []]
+        return [
+            slack_username(user) for role in roles or [] for user in role.users or []
+        ]
 
     async def fetch_owners(
         self,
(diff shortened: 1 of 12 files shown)


esron:
Does that exposes gov_slack_username for people outside boundary?

esron (approved):

reconcile/slack_usergroups_api.py
```diff
@@ -87,6 +87,20 @@
...
+    gov_slack_username: str | None
+
+
+def slack_username(user: _UserWithSlackIdentity) -> str:
+    """Return the Slack identity for an app-interface user.
+
+    Prefer gov_slack_username when set (gov Slack email local-part), otherwise
+    fall back to org_username (commercial Slack / LDAP).
+    """
+    return user.gov_slack_username or user.org_username
```
hemslo: Question on intent: today `org_username` is matched against Slack purely via email-prefix comparison on the server side — it fetches each Slack user's profile email and computes `org_username = email.split("@")[0]` (see `qontract_utils/qontract_utils/slack_api/models.py`), then does a plain string match against whatever the client sends in `SlackUsergroupConfig.users`. There's no separate identity table; the match is entirely email-local-part vs the string sent.

Is `gov_slack_username` intended to work the same way — i.e. should it equal the local-part of the user's *GovSlack* email address, so this same email-prefix match resolves it correctly server-side? Want to confirm that's the intent, since if GovSlack profile emails don't line up 1:1 with this field's expected value, it won't resolve to a valid Slack user ID.

hemslo:
Nit while reviewing: `reconcile/slack_usergroups_api.py` line 367 (in `compile_users_from_pagerduty_schedules`) still builds its map with `user.org_username` directly:

```python
users_map: dict[str, str] = {
    user.pagerduty_username: user.org_username
    for user in app_interface_users
    if user.pagerduty_username
}
```

Every other user-collecting method in this file (`compile_users_from_schedule`, `compile_users_from_roles`, `fetch_owners`, `compile_desired_state_cluster_usergroups`) was switched to the new `slack_username(user)` helper, but this one wasn't. PagerDuty-sourced usergroup members in GovSlack would fall back to `org_username` and fail the email-prefix match server-side. This line isn't in the diff (untouched), so I couldn't attach this as an inline comment on it directly — flagging as a general comment instead.

reconcile/slack_usergroups_api.py
chassing: ^same question from my side

suzana-nesic:
> Does that exposes gov_slack_username for people outside boundary?

I wouldn't think so since those values would be defined through in-boundary app-interface.

reconcile/slack_usergroups_api.py
suzana-nesic: Yes, `gov_slack_username` should work the same way. I was only able to have lines like `action.usergroup='<some-group>' action.users_to_add=[<gov_slack-alias>]` populate with changes like this one. I'll look more into the debug today to find where exactly it would leave `action.users_to_add` empty with the current code configuration.

suzana-nesic:
> Nit while reviewing: `reconcile/slack_usergroups_api.py` line 367 (in `compile_users_from_pagerduty_schedules`) still builds its map with `user.org_username` directly:
> 
> ```python
> users_map: dict[str, str] = {
>     user.pagerduty_username: user.org_username
>     for user in app_interface_users
>     if user.pagerduty_username
> }
> ```
> 
> Every other user-collecting method in this file (`compile_users_from_schedule`, `compile_users_from_roles`, `fetch_owners`, `compile_desired_state_cluster_usergroups`) was switched to the new `slack_username(user)` helper, but this one wasn't. PagerDuty-sourced usergroup members in GovSlack would fall back to `org_username` and fail the email-prefix match server-side. This line isn't in the diff (untouched), so I couldn't attach this as an inline comment on it directly — flagging as a general comment instead.

Pagerduty is not approved to use in-boundary so I left that part of the code untouched. Is there still issues I would need to consider if we don't use Pagerduty?

reconcile/slack_usergroups_api.py
suzana-nesic: `org_username` is used for other functionalities, right? Right now our username does work in that sense as we define it in `org_username`, like `gitlab_members`, for example, but we will need to have another property to recognize a different username when it comes to anything specifically Slack related.

hemslo: So in gov slack, you have a different email like `<gov_slack_username>@<xxx>`? To avoid confuse with username in slack, better to give it a different name, like `gov_slack_email_local_part`

hemslo: and since you extracted a new method to wrap `user.gov_slack_username or user.org_username`, then `org_username` in this integration should be only used in this method, means pager duty related call should also invoke new method.

suzana-nesic: Our username in FedRAMP is a different value from `<gov-identity>@<in-boundary>`. I can rename the variable name to avoid confusion

hemslo: Just clarify that in this integration, we don't use slack username from slack api response at all, as slack user can change username then it breaks the integration, instead we use email field to extract local part to match org username, as that's immutable field.

suzana-nesic: Okay, I'll step through the code again and see why it's not able to extract from the email field without any code changes. Thank you!

reconcile/slack_usergroups_api.py
```diff
@@ -87,6 +87,20 @@
...
+    gov_slack_email_local_part: str | None
+
+
+def slack_username(user: _UserWithSlackIdentity) -> str:
+    """Return the Slack identity for an app-interface user.
+
+    Prefer gov_slack_email_local_part when set (gov Slack email local-part), otherwise
+    fall back to org_username (commercial Slack / LDAP).
+    """
+    return user.gov_slack_email_local_part or user.org_username
```
hemslo: to avoid confuse with `name` field in slack (slack username), better to rename func name to `slack_identity`

suzana-nesic: Resolved

hemslo (approved):
LGTM!

bkez322:
late LGTM

PR #5691, opened 2026-07-21
fix(aus): fetch available_channels from cluster GET instead of versions API

devppratik:
## Summary

- The versions API (`/api/clusters_mgmt/v1/versions/{id}`) returns 403 for org-less service accounts on non-stable channel groups (candidate, nightly), blocking Y-stream channel switching for AUS clusters
- Switch `_get_available_channels` to use individual cluster GET (`/api/clusters_mgmt/v1/clusters/{id}`) which populates `version.available_channels` without the 403 restriction
- Results remain cached by `version_id` — only the first cluster per version pays an API call

Ref: ROSAENG-62163

## Test plan

- [x] All 28 channel switch tests pass
- [x] All 252 AUS tests pass
- [x] mypy clean
- [ ] Verify in integration/staging with org-less SA that cluster GET returns `available_channels` for candidate versions

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic channel switching by retrieving available channels from the cluster’s current version information.
  * Updated fallback behavior to use the correct cluster details when determining eligible channels.
  * Enhanced handling of channel-discovery errors, including more informative diagnostics.

* **Tests**
  * Updated coverage to validate cluster-based channel discovery, caching, error handling, and fallback behavior.

### DIFF
--- a/reconcile/aus/base.py
+++ b/reconcile/aus/base.py
@@ -1180,9 +1180,7 @@ def _get_available_channels(
         return cache[version_id]
     try:
         data = ocm_api.get(api_path=build_cluster_url(cluster_id))
-        channels = set(
-            (data.get("version") or {}).get("available_channels") or []
-        )
+        channels = set((data.get("version") or {}).get("available_channels") or [])
     except HTTPError as e:
         detail = e.response.text if e.response is not None else ""
         logging.warning(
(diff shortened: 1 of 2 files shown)


esron (approved):
LGTM, just need to run `make format`.

hemslo (approved):
LGTM!

PR #5704, opened 2026-07-27
fix(helm): quote extra_args/extraArgs env values to prevent broken YAML

chassing:
## Summary
- `INTEGRATION_EXTRA_ARGS` was interpolated in the Helm chart as `value: "{{ ... }}"` without escaping. Any `extra_args`/`extraArgs` value containing a literal double quote (e.g. a JSON payload passed to `--keycloak-instances` for `rhidp-sso-client-api`) produced invalid YAML, causing `helm template` to fail with `did not find expected key`.
- Fixed both occurrences (`$shard.extra_args` for regular Deployments, `$integration.extraArgs` for CronJob-based integrations) to use the `quote` template function, which properly escapes embedded quotes.

## Test plan
- Added `test_template_extra_args_with_embedded_quotes` in `reconcile/test/utils/test_helm.py`, reproducing an `extra_args` value with embedded JSON/double quotes.
- Verified the test fails with the old (unfixed) template (`HelmTemplateError: ... did not find expected key`) and passes with the fix.
- `uv run pytest reconcile/test/utils/test_helm.py` — 29 passed.
- `uv run ruff check` / `ruff format --check` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Helm rendering of `INTEGRATION_EXTRA_ARGS` to handle values containing embedded double quotes, avoiding incorrect YAML escaping in generated Deployments/StatefulSets and CronJobs.
* **Tests**
  * Added a new Helm fixture covering JSON-like `extra_args` with embedded quotes.
  * Added a pytest that asserts rendered manifests match the fixture output.

### DIFF
--- a/helm/qontract-reconcile/templates/template.yaml
+++ b/helm/qontract-reconcile/templates/template.yaml
@@ -368,7 +368,7 @@ objects:
             value: {{ $integration.name }}
           {{- with $shard.extra_args }}
           - name: INTEGRATION_EXTRA_ARGS
-            value: "{{ $shard.extra_args }}"
+            value: {{ $shard.extra_args | quote }}
           {{- end }}
           - name: SLEEP_DURATION_SECS
             {{- if $integration.sleepDurationSecs }}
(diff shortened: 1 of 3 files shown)


hemslo (approved):
LGTM!

PR #5706, opened 2026-07-27
fix(glhk) Filter push pipelines

rhur-pixel:
## Summary

Simplify the OMM pipeline filter to remove ALL push-source pipelines instead of only those matching the current MR sha. Push pipelines in app-interface are exclusively artifacts of `rebase(skip_ci=True)` and carry no meaningful CI signal.

## Problem

The current filter (`p.sha == mr.sha and p.source == "push"`) has two failure modes that cause wasted reconcile loops:

1. **Stale sha from list API**: GitLab's list endpoint caches MR sha for ~60s after a rebase. When `mr.sha` is stale, the filter fails to match push pipelines at the real (new) sha, causing them to leak through as RUNNING → `waiting-pipeline` for one extra loop.

2. **Cascading push from prior rebases**: When a member needs multiple skip-ci-rebases (e.g., target moves because another member merged), push pipelines from earlier rebases have a different sha than the current one. The filter only removes push at the current sha, letting older push pipelines leak through as RUNNING.

Both issues cause unnecessary `waiting-pipeline` loops (~60-80s each) without affecting correctness — the leaked push pipeline always shows as RUNNING/PENDING, never causing a false merge.

### Push pipelines are exclusively skip-ci-rebase artifacts

Verified via GitLab API (2026-07-27):

- **27 same-project MRs** checked: 26 have zero push pipelines. The 1 exception (MR !198782) was an OMM member that went through skip-ci-rebase.
- **MR !198599** (rebased 20+ times via normal rebase, never OMM'd): zero push pipelines across 20 pipeline entries.
- **No `.gitlab-ci.yml`** exists in app-interface or its forks. 

### Mechanism

- Normal `mr.rebase()`: creates `merge_request_event` pipeline on fork → external CI posts status as `source=external`. No push.
- `mr.rebase(skip_ci=True)`: MRE pipeline is SKIPPED → fork/project gets an empty `push` pipeline → external CI status attaches to it.

Push pipelines have only been observed as artifacts of `skip_ci=True` rebase. Without `.gitlab-ci.yml` in the project or forks, no mechanism exists for creating push pipelines with meaningful CI jobs.

### Test run observations (2026-07-27)

- MR !198703 (Group 4): stale sha caused `cf69fcb7 push` to leak through filter on first loop → unnecessary `waiting-pipeline`. With the new filter, it would have immediately seen `a8f06f13 success external` and proceeded.
- MR !198774 (Group 5): stuck on `waiting-pipeline running` for 5 consecutive loops. Without probe data we cannot confirm whether this was a leaked push pipeline or genuinely slow external CI, but the new filter eliminates the former possibility.

## Risk Assessment

| Risk | Status | Evidence |
|------|--------|----------|
| Same-project MRs with push-only CI | Eliminated | 27 same-project MRs checked, none have push without OMM |
| Formation admits member that processing can't evaluate | Eliminated | Push only appears after formation (from skip-ci-rebase) |
| Empty pipeline list causing infinite wait | Eliminated | External pipelines always exist alongside push |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pipeline filtering for OMM member merge handling.
  * Push pipelines from earlier rebases are now excluded reliably, even when their commit differs from the merge request’s current commit.
* **Tests**
  * Added coverage to verify correct merge processing when prior rebase pipelines are present.


hemslo (approved):
LGTM — well-evidenced fix with clear risk analysis. The broadened filter is safe given the audit data.

reconcile/test/test_gitlab_housekeeping.py
```diff
@@ -3027,14 +3027,14 @@ def test_omm_group_unhandled_status_rebased_stays_active(
     [("abc123", None), (None, "abc123")],
     ids=["merge-commit", "squash-commit"],
 )
-def test_omm_group_fork_pipeline_post_rebase_filtered_merges(
+def test_omm_group_push_pipeline_filtered_even_at_different_sha(
```
hemslo, line 3030: Nit: the section header just above (line 3022: `# --- Fork pipeline SHA filtering in OMM group ---`) is slightly stale now — the test is about push-source filtering, no longer SHA filtering. Maybe `# --- Push pipeline filtering in OMM group ---`?

PR #5707, opened 2026-07-28
:loud_sound: improve qontract api error logging

chassing:
$TITLE

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized unexpected-exception failure details across reconciliation tasks by formatting error messages as `Unexpected {err=}` while keeping failure status, empty actions, and zero applied count.
  * Centralized API exception logging in reconciliation flows, including capturing APIException reason and truncating large response text, while preserving existing control flow and error propagation.
* **New Features**
  * Exposed `APIException` through the client exceptions interface for consistent API error handling.

### DIFF
--- a/qontract_api/qontract_api/integrations/github_owners/tasks.py
+++ b/qontract_api/qontract_api/integrations/github_owners/tasks.py
@@ -134,11 +134,11 @@ def reconcile_github_owners_task(
 
         return result
 
-    except Exception as e:
+    except Exception as err:
         logger.exception(f"Task {request_id} failed with error")
         return GithubOwnersTaskResult(
             status=TaskStatus.FAILED,
             actions=[],
             applied_count=0,
-            errors=[str(e)],
+            errors=[f"Unexpected {err=}"],
         )
--- a/qontract_api_client/qontract_api_client/exceptions.py
+++ b/qontract_api_client/qontract_api_client/exceptions.py
@@ -1,3 +1,3 @@
-from clientele.api.exceptions import HTTPStatusError
+from clientele.api.exceptions import APIException, HTTPStatusError
 
-__all__ = ["HTTPStatusError"]
+__all__ = ["APIException", "HTTPStatusError"]
--- a/qontract_api_client/qontract_api_client/exceptions.py
+++ b/qontract_api_client/qontract_api_client/exceptions.py
@@ -1,3 +1,6 @@
+# This file isn't automatically generated, but it is a re-export of the exceptions from the `clientele` package.
+# This allows users of `qontract-api-client` to import exceptions directly from this package
+# without needing to know about the underlying `clientele` package.
 from clientele.api.exceptions import APIException, HTTPStatusError
 
 __all__ = ["APIException", "HTTPStatusError"]
--- a/reconcile/github_owners_api.py
+++ b/reconcile/github_owners_api.py
@@ -25,7 +25,6 @@ from collections import defaultdict
 from typing import TYPE_CHECKING
 
 from qontract_api_client.client import github_owners as reconcile_github_owners
-from qontract_api_client.exceptions import APIException
 from qontract_api_client.schemas import (
     GithubOrgDesiredState,
     GithubOwnersReconcileRequest,
@@ -170,11 +169,8 @@ class GithubOwnersIntegration(
             organizations=organizations,
             dry_run=dry_run,
         )
-        try:
+        with self.log_api_exceptions():
             response = await reconcile_github_owners(request)
-        except APIException as e:
-            logging.error(f"Error occurred: {e.reason}; Response details: {e.response}")
-            raise
         logging.info(f"request_id: {response.id}")
         return response
 
--- a/reconcile/github_owners_api.py
+++ b/reconcile/github_owners_api.py
@@ -25,6 +25,7 @@ from collections import defaultdict
 from typing import TYPE_CHECKING
 
 from qontract_api_client.client import github_owners as reconcile_github_owners
+from qontract_api_client.exceptions import APIException
 from qontract_api_client.schemas import (
     GithubOrgDesiredState,
     GithubOwnersReconcileRequest,
@@ -169,7 +170,11 @@ class GithubOwnersIntegration(
             organizations=organizations,
             dry_run=dry_run,
         )
-        response = await reconcile_github_owners(request)
+        try:
+            response = await reconcile_github_owners(request)
+        except APIException as e:
+            logging.error(f"Error occurred: {e.reason}; Response details: {e.response}")
+            raise
         logging.info(f"request_id: {response.id}")
         return response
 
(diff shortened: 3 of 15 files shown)


reconcile/github_owners_api.py
```diff
@@ -169,7 +170,11 @@ async def reconcile(
             organizations=organizations,
             dry_run=dry_run,
         )
-        response = await reconcile_github_owners(request)
+        try:
+            response = await reconcile_github_owners(request)
+        except APIException as e:
```
lisa: Why are some of the changes in the PR (like this instance) `except ... as err` and `except ... as e`? I feel like there should be greater consistency, especially since this PR has many changing _from_ `e` _to_ `err`

qontract_api_client/qontract_api_client/exceptions.py
```diff
@@ -1,3 +1,3 @@
-from clientele.api.exceptions import HTTPStatusError
+from clientele.api.exceptions import APIException, HTTPStatusError
 
-__all__ = ["HTTPStatusError"]
+__all__ = ["APIException", "HTTPStatusError"]
```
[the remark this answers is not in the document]
chassing: I've added a note in [93ddc39](https://github.com/app-sre/qontract-reconcile/pull/5707/commits/93ddc3935d266428a4864df57b9bb8c81e83a2db) that this file isn't auto generated

lisa (commented):
nits to pick.

Consider also a more specific message other than "Error occurred," such as "Error occurred while <doing this thing>."

reconcile/github_owners_api.py
chassing: I've changed it to `err` because `f"Unexpected {err=}"` will generate logs like
`Unexpected err=KeyError("foobar")` and I thought `err=` is more clear than `e=`

lisa: Ah, I see what you're saying.

lisa (approved):
lgtm

reconcile/github_owners_api.py
```diff
@@ -169,7 +170,11 @@ async def reconcile(
...
+            logging.error(f"Error occurred: {e.reason}; Response details: {e.response}")
```
hemslo: This exact 3-line `try/except APIException` block is repeated 9 times across this PR with identical logic. Consider centralizing:

- Override `APIException.__str__` to include `reason` and `response` — then default exception propagation surfaces this info without any try/except at call sites
- Or a small decorator/context manager in `reconcile/utils/runtime/`

Also: `e.response` will render via `repr()`, which may include the full `content: bytes` field (e.g., an HTML error page). Consider logging `e.response.status_code` and truncated `e.response.text` instead of the entire object.

qontract_api/qontract_api/integrations/github_owners/tasks.py
```diff
@@ -134,11 +134,11 @@ def reconcile_github_owners_task(
 
         return result
 
-    except Exception as e:
+    except Exception as err:
         logger.exception(f"Task {request_id} failed with error")
```
hemslo, line 138: `logger.exception` here already logs the full traceback including the exception message. With the client-side `logging.error` added in this PR, API failures will produce two log entries for the same error — one from the explicit `logging.error` before re-raise, and this one from the catch-all handler. Consider whether the client-side logging is redundant given this already captures everything.

reconcile/utils/runtime/integration.py
```diff
@@ -318,12 +320,16 @@ async def poll_task_status[T: BaseModel](
...
-        )
+        try:
+            return await qontract_api_client.arequest(
+                method="get",
+                path=urlparse(status_url).path,
+                response_map={200: result_type},
+                query={"timeout": timeout},
+            )
+        except APIException as e:
+            logging.error(f"Error occurred: {e.reason}; Response details: {e.response}")
```
hemslo: Since `poll_task_status` is the common chokepoint for task-based integrations, logging here may already be sufficient — the identical catch blocks at each individual call site in the reconcile integrations could be redundant if the exception propagates through here.

hemslo (commented):
No tests added for the new `APIException` catch-and-reraise behavior. At minimum, the existing integration tests should verify that `APIException` propagates correctly and produces the expected log output.

reconcile/github_owners_api.py
chassing: Good catch. Centralized this in 9c7a0aa9: added `QontractReconcileApiIntegration.log_api_exceptions()` (a context manager in `reconcile/utils/runtime/integration.py`) and replaced all ~9 duplicated try/except blocks with `with self.log_api_exceptions():`. Went with the base-class helper rather than overriding `APIException.__str__`, since that class comes from the pinned third-party `clientele==2.2.2` package — not something we can patch here without an upstream fork.

On the repr() concern: I checked `clientele.http.response.Response.__repr__` — it's already overridden to only print `<Clientele HTTP Response [status_code]>`, not the body/content, so the original code wasn't actually leaking full HTML/JSON error pages. Still, to be explicit and not depend on that upstream implementation detail, the new helper logs `status_code` and a truncated (500 char) `response.text` directly instead of relying on repr().

qontract_api/qontract_api/integrations/github_owners/tasks.py
chassing: I don't think these two log statements can actually collide. This file (`qontract_api/qontract_api/integrations/github_owners/tasks.py`) is Celery worker/server-side code — the `except Exception as err` here wraps a call into `GithubOwnersService.reconcile()`, a plain in-process Python call with no HTTP involved (confirmed no `APIException`/`logging.error` usage anywhere in that service module). The new client-side `logging.error` on `APIException` added in this PR lives in `reconcile/github_owners_api.py`, which is a completely separate process (the CLI-side integration that calls qontract-api over HTTP). They're different services with different log streams, so an API failure client-side and a task failure server-side aren't "the same error" logged twice — they're two different failure classes that happen to both get logged, each in its own place. Let me know if I'm missing a path where these actually overlap.

reconcile/utils/runtime/integration.py
chassing: Checked this — the per-call-site `except APIException` blocks (e.g. `github_owners_api.py:175`) wrap the *initial task-submission* calls (`reconcile_github_owners(request)`, `sso_client(request)`, etc.), not `poll_task_status`. None of the reconcile integrations actually wrap their `poll_task_status(...)` calls in try/except — only the shared base-class method itself does, so polling is already centralized with no duplication there.

That said, you're right that the submission-call sites repeat the same boilerplate as each other (and as `poll_task_status` did before this PR) — folded that into the same fix as your other comment: `QontractReconcileApiIntegration.log_api_exceptions()` in 9c7a0aa9, used by `poll_task_status` and all the individual call sites alike.

hemslo (approved):
LGTM. The centralized `log_api_exceptions()` context manager is clean, all prior feedback addressed. Verified the author's replies — architecture claims (separate processes, no double-logging, no redundant wrapping of `poll_task_status`) all check out.

PR #5711, opened 2026-07-29
Revert fix(aus): skip 400 errors on channel switch for org-less SAs

devppratik:
## Summary
- Reverts #5695 — the 400 error workaround for org-less service accounts on channel switch
- OCM fixed the underlying issue (ROSAENG-62340) server-side, so the client-side workaround is no longer needed
- The workaround was masking a channel flip-flop loop between AUS and OSD Fleet Manager

## Test plan
- [ ] CI passes (lint, types, unit tests)
- [ ] Verify channel switches return 200 without the workaround (OCM fix deployed)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during channel switching so all HTTP errors are consistently reported.
  * Failed channel-switch operations now contribute to the final error result instead of being silently ignored.
  * Processing continues for other channel switches, while affected operations are clearly surfaced.

### DIFF
--- a/reconcile/aus/base.py
+++ b/reconcile/aus/base.py
@@ -1445,16 +1445,6 @@ def act_channel_switches(
             update_cluster_channel(ocm_api, switch.cluster.id, switch.to_channel)
         except HTTPError as e:
             detail = e.response.text if e.response is not None else ""
-            if (
-                e.response is not None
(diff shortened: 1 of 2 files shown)


chassing (approved):

hemslo (approved):
LGTM!

PR #5712, opened 2026-07-29
fix(gitlab-housekeeping): re-fetch MR before is_rebased check in OMM

rhur-pixel:
## Summary

OMM (Optimistic Merge Mode) groups form when the lead MR merges — members get skip-ci-rebased onto the new target HEAD and should merge on the next loop once their pipeline succeeds. In practice, members are getting skip-ci-rebased 2-3 times before finally merging, wasting ~90s per unnecessary cycle.

**Root cause:** `_process_omm_member` receives MR objects from `get_omm_pending_mrs()`, which uses the list API. GitLab's list API caches the MR `sha` for 60-120s after a rebase. `is_rebased()` compares `mr.sha` against the target branch HEAD — a stale sha makes it return `False`, triggering a redundant skip-ci-rebase. Worse, each unnecessary rebase creates a new SHA, resetting the cache timer, so the loop repeats.

**Evidence from today's live run (2026-07-29):**
- MR !199226: skip-ci-rebased at formation (14:21:57), then hit a *second* skip-ci-rebase (14:23:34) before finally merging (14:25:07) — the extra cycle added ~90s of delay with no target HEAD change in between
- Across the full OMM session: 28 skip-ci-rebases for only 13 merges (2.15x ratio; ideal is ~1:1)
- Debug probes from 7/27 confirmed the mechanism: list API returned stale sha while the detail API (`GET /merge_requests/:iid`) returned the current sha at the same instant

GitLab issue [#26654](https://gitlab.com/gitlab-org/gitlab/-/issues/26654) documents the async rebase race — it resolves within ~30s, well within GLHK's ~90s loop cycle, so the detail API is reliable by the time we check.

## Changes

- **`reconcile/gitlab_housekeeping.py`**: Re-fetch MR from detail API before calling
  `is_rebased()` in `_process_omm_member`
- **`reconcile/test/test_gitlab_housekeeping.py`**: New test
  `test_omm_member_refreshes_mr_before_rebased_check` that verifies a stale-SHA MR
  merges (instead of rebasing) after the detail API returns the fresh SHA

## Test plan

- [ ] New test passes locally (parametrized for merge-commit and squash-commit)
- [ ] Existing OMM tests still pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved merge request handling by refreshing merge request details before determining whether a rebase is needed.
  * Prevented valid, already-rebased merge requests from being incorrectly skipped or sent through a rebase workflow.

* **Tests**
  * Added coverage for scenarios where stale merge request information could lead to incorrect processing.


hemslo (approved):
LGTM — clean, well-motivated fix. The detail-API refresh pattern is already established in this file (lines 435, 1462), so this is consistent. Left a minor suggestion on the test.

reconcile/test/test_gitlab_housekeeping.py
```diff
@@ -2551,6 +2551,65 @@ def test_omm_group_skip_ci_rebase_on_success_not_rebased(
...
+    mocked_gl.get_merge_request.return_value = fresh_mr
+
+    head_commit = Mock()
+    head_commit.id = head_sha
+    mocked_gl.project.commits.list.return_value = [head_commit]
+    mocked_gl.project.repository_compare.return_value = {"commits": []}
+
+    merges = gl_h._process_omm_group(
+        dry_run=False,
+        gl=mocked_gl,
```
hemslo, line 2609: Nit: `repository_compare` is mocked to return `{"commits": []}` unconditionally, so even the stale SHA would make `is_rebased` return `True`. The test proves the refresh *happens* but not that it *changes the outcome*.

A tighter mock could use `side_effect` to return `{"commits": ["x"]}` for `"stale-sha-old"` and `{"commits": []}` only for `"fresh-sha-rebased"` — that way the test would fail if the refresh line were reverted.

Not blocking, just a suggestion for a stronger assertion.

PR #5717, opened 2026-07-31
fix(change-owners): render Diff.path without jsonpath_ng 1.8.0 parens

chassing:
## Summary
- jsonpath_ng 1.8.0 wraps every `Child` node in parentheses when stringified (e.g. `(((a.b).c).d)`) to disambiguate precedence for re-parsing. Since the change-owners MR check puts raw `Diff.path` objects into report tables that stringify them implicitly, this made the change coverage report unreadable, e.g. `(((((resourceTemplates.[0]).targets).[0]).parameters).A)` instead of `resourceTemplates.[0].targets.[0].parameters.A`.
  - Example MR comment **with** the parens (jsonpath-ng 1.8.0): https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/199516
  - Example MR comment **without** the parens (jsonpath-ng 1.7.0, before the bump): https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/199413
- `Diff` now normalizes its `path` once in `__post_init__`, rebuilding it with a private `Child` subclass that renders as a plain dotted string. This also covers paths built internally by jsonpath_ng's own `find()` traversal (used when resolving a change-type's self-service path against real file content for partial/split diff coverage), so no display call site needs to change.

## Test plan
- [x] `uv run pytest reconcile/test/change_owners/ reconcile/test/utils/test_jsonpath.py reconcile/test/runtime/test_utils_desired_state_diff.py` — 218 passed
- [x] Added tests reproducing the bug (verified they fail without the fix): rendering of nested paths, paths built via jsonpath_ng's `find()` traversal, and the actual `format_table`/tabulate rendering path used by the MR report
- [x] `mypy reconcile/change_owners/` — clean
- [x] `ruff check` / `ruff format --check` — clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved change report paths with clearer dotted notation.
  * Removed confusing nested parentheses from displayed JSON paths.
  * Ensured paths are formatted consistently across generated reports and merge-request tables.
  * Made nested fields easier to read and interpret in change summaries.

* **Tests**
  * Added coverage for parsed paths, traversed paths, nested field formatting, and report table output.

### DIFF
--- a/reconcile/change_owners/diff.py
+++ b/reconcile/change_owners/diff.py
@@ -12,7 +12,7 @@ from deepdiff.helper import CannotCompare
 from deepdiff.path import parse_path
 
 from reconcile.utils.json import json_dumps
-from reconcile.utils.jsonpath import remove_prefix_from_path
+from reconcile.utils.jsonpath import jsonpath_parts, remove_prefix_from_path
 
 if TYPE_CHECKING:
     from deepdiff.model import DiffLevel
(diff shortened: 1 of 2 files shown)


reconcile/change_owners/diff.py
```diff
@@ -24,6 +24,46 @@ class DiffType(Enum):
...
 
 
+class _DottedChild(jsonpath_ng.Child):
+    """
+    Same as `jsonpath_ng.Child`, but renders as a plain dotted path, e.g.
+    `resourceTemplates.[0].targets.[0].parameters.A`.
+
+    jsonpath_ng 1.8.0 wraps every `Child` in parentheses when stringified
+    (e.g. `(((a.b).c).d)`) to disambiguate operator precedence for
+    re-parsing. `Diff.path` is only ever rendered for display (MR check
```
hemslo, line 34: nit / question (non-blocking): `_join_for_display` treats `Root` the same as `This` on the left side — it drops it and returns just `right`. That means a path like `Child(Root(), Fields('a'))` (i.e. `$.a`) normalizes to `Fields('a')` and renders as `"a"` instead of `"$.a"`.

I can see this is fine in practice because `deepdiff_path_to_jsonpath` never produces a Root-prefixed path, and the `extract_diffs` Root-only case (`Diff(path=Root(), ...)`) survives since `reduce` on a single-element list returns `Root()` directly.

But is the Root-dropping intentional? If a `find()` traversal ever returned a Root-prefixed path, this would silently strip the `$`. Would it be worth either:
- adding a `("$.field", "field")` test case to document that Root-stripping is expected, or
- handling Root separately from This (keep it, build `_DottedChild(Root(), right)`) so `$` is preserved?

Not blocking — just want to make sure the behavior is deliberate.

hemslo (approved):
Clean fix — normalizing in `__post_init__` covers all construction paths (deepdiff, `find()` traversal, `create_subdiff`) without touching any call site. `_DottedChild` only overrides `__str__`, so structural behavior (`find()`, `child()`, `__eq__`) is inherited unchanged. Verified `Child.__eq__` uses `isinstance(other, Child)`, so mixed comparisons work symmetrically. Tests are thorough — left one non-blocking question inline.

reconcile/change_owners/diff.py
chassing: Good catch, and yes it's deliberate. I checked: `jsonpath_ng.parse("$.a")` does build a raw `Child(Root(), Fields('a'))` (the parser bypasses jsonpath_ng's own `.child()` canonicalization that normally drops a leading `This`/`Root`), so this is reachable, not just theoretical.

I'm keeping the Root-drop behavior rather than preserving `$`: every other displayed path in this codebase already renders without a leading `$` (e.g. `"field"`, `"resourceTemplates.[0]"`), so showing `$.a` only for paths that happened to be authored with an explicit `$.` prefix would be inconsistent rather than more informative — readers would have no way to know why some paths have it and others don't.

Added the `("$.field", "field")` case you suggested in 1b40249e to document this explicitly.

PR #5718, opened 2026-07-31
fix(rhidp): correct grafana dashboard queries after qontract-api migration

chassing:
## Summary
- The `rhidp-sso-client` integration now runs as multiple `qontract-api-worker` replicas instead of a single `reconcile` pod, so `rhidp_managed_clusters` and `rhidp_sso_client_number_of_clients` were being summed once per replica, inflating the numbers on the dashboard. Queries now dedupe with `max by (...)` before summing, so the panels are correct regardless of replica count.
- Updated the `rhidp_sso_client_inital_access_token_expiration` panels to match the new Vault `path` values for the prod/stage Keycloak IAT secrets (the old `app-sre/creds/rhidp/auth*.redhat.com` paths no longer exist).
- Bumped the dashboard `version`.

## Test plan
- [x] Validated the ConfigMap YAML/embedded dashboard JSON parses correctly
- [x] Manually verified new PromQL expressions against a live Prometheus query dump from both the old (`reconcile`) and new (`qontract-api`) integration

## Summary by CodeRabbit

* **Dashboard Updates**
  * Improved Grafana metrics for managed clusters and SSO clients by environment and organization.
  * Renamed IAT expiration panels for clearer identification of production and staging data.
  * Updated metric paths used by the IAT expiration panels.
  * Incremented the dashboard version.

### DIFF
--- a/grafana-dashboards/sre-capability-idp.configmap.yaml
+++ b/grafana-dashboards/sre-capability-idp.configmap.yaml
@@ -137,7 +137,7 @@ data:
                 "uid": "P7B77307D2CE073BC"
               },
               "editorMode": "code",
-              "expr": "sum(rhidp_managed_clusters{integration=~\"rhidp-sso-client.*\"})",
+              "expr": "sum(max by (org_id, ocm_environment) (rhidp_managed_clusters{integration=~\"rhidp-sso-client.*\"}))",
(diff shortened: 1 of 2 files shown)


hemslo (approved):
LGTM!

PR #5722, opened 2026-08-03
fix(glhk) Re-fetch MR before is_rebased check in serial rebase paths

rhur-pixel:
## Problem

In `_rebase_merge_requests_active_cap` (line 896), the MR objects are fetched once via `get_merge_requests()` at the top of the function (line 909-918). When `is_rebased(mr, gl)` is called at line 930, it uses `mr.sha` from this batch-fetched object. If the previous cycle's async `mr.rebase()` hasn't been reflected yet in the list endpoint, `mr.sha` is stale, `is_rebased` returns False, and the MR gets rebased again unnecessarily.

This is the same bug fixed in #5712 for the OMM path, where `gl.get_merge_request(mr.iid)` is called to get a fresh object before checking `is_rebased` (line 1072-1073).

## Evidence from logs (8/3 production run)

MRs 199925 and 199917 are rebased in cycle 1, then rebased again in cycle 2 with no merge in between:

```
# Cycle 1 (13:37:24) - rebases issued
[2026-08-03 13:37:24] [INFO] [gitlab_housekeeping.py:_try_rebase:690] - ['rebase', 'app-interface', 199925]
[2026-08-03 13:37:24] [INFO] [gitlab_housekeeping.py:_try_rebase:690] - ['rebase', 'app-interface', 199917]
[2026-08-03 13:37:24] [INFO] [gitlab_housekeeping.py:_try_rebase:690] - ['rebase', 'app-interface', 199928]

# No merge log between cycles — only a skip:
[2026-08-03 13:38:32] [INFO] [gitlab_housekeeping.py:merge_merge_requests:1351] - ['skip merge', 'app-interface', 199729]

# Cycle 2 (13:38:49) - same MRs rebased again
[2026-08-03 13:38:49] [INFO] [gitlab_housekeeping.py:_try_rebase:690] - ['rebase', 'app-interface', 199947]
[2026-08-03 13:38:50] [INFO] [gitlab_housekeeping.py:_try_rebase:690] - ['rebase', 'app-interface', 199925]
[2026-08-03 13:38:50] [INFO] [gitlab_housekeeping.py:_try_rebase:690] - ['rebase', 'app-interface', 199917]
```

The GQL bundle SHA changes between cycles (`ce95c7a6...` -> `d105c7e4...`), but this reflects a data refresh, not necessarily a target-branch merge. Nothing in the logs shows a merge that would invalidate the prior rebases. The re-rebases are redundant — caused by `is_rebased` reading stale `mr.sha` from the batch-fetched MR list rather than re-fetching the individual MR object after the async rebase completed.

In `_process_omm_member`, a pipeline with status `canceled` falls into the catch-all `unhandled-status` branch, which returns `active=True` when the MR is rebased. This holds the OMM group hostage — the member can't merge (not SUCCESS) but prevents adaptive-close (active=True). The group spins until the 5-minute window expires.

## Fix
In [`reconcile/gitlab_housekeeping.py`](qontract-reconcile/reconcile/gitlab_housekeeping.py), change the `is_rebased` call at line 930 to re-fetch the MR first:

```python
# Before (line 930):
if is_rebased(mr, gl):

# After:
fresh_mr = gl.get_merge_request(mr.iid)
if is_rebased(fresh_mr, gl):
```

This adds one API call per MR in the loop, but it's the same pattern already used in the OMM path and is necessary to get an accurate `sha` value.

## Same fix for old-burst path

The `_rebase_merge_requests_old_burst` function (line 968) has the same pattern at line 992:

```python
if is_rebased(mr, gl):
```

Apply the same re-fetch there for consistency.

Treat `canceled` the same as `failed` — eject the member from the group so it returns to the serial queue for a fresh rebase + CI run.

## Test

- Verify existing tests pass (the `is_rebased` mock may need updating if tests supply MR objects directly)
- The log output should show fewer redundant rebase entries across consecutive cycles

Written by Claude

## Summary by CodeRabbit

* **Bug Fixes**
  * Merge request rebase checks now use the latest information from GitLab, improving accuracy when statuses change after initial retrieval.
  * Canceled pipelines are now handled like failed pipelines, removing affected members and performing the necessary cleanup.
  * Unrecognized pipeline statuses, such as manual, no longer incorrectly deactivate active groups.
* **Tests**
  * Expanded coverage for refreshed merge request details and pipeline status handling.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -1098,7 +1098,8 @@ def _process_omm_member(
         if not dry_run:
             gl.remove_label(mr, OMM_PENDING)
         optimistic_merge_rejected.labels(
-            project_id=mr.target_project_id, reason="pipeline_failed"
+            project_id=mr.target_project_id,
+            reason=f"pipeline_{latest_status.value}",
         ).inc()
         return _MemberResult()
 
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -927,7 +927,8 @@ def _rebase_merge_requests_active_cap(
     for mr in merge_requests:
         pipelines = gl.get_merge_request_pipelines(mr)
         pipelines = [p for p in pipelines if p.status != PipelineStatus.SKIPPED]
-        if is_rebased(mr, gl):
+        fresh_mr = gl.get_merge_request(mr.iid)
+        if is_rebased(fresh_mr, gl):
             if pipelines and pipelines[0].status in {
                 PipelineStatus.RUNNING,
                 PipelineStatus.PENDING,
@@ -989,7 +990,8 @@ def _rebase_merge_requests_old_burst(
         if not item["error"]
     ]
     for mr in merge_requests:
-        if is_rebased(mr, gl):
+        fresh_mr = gl.get_merge_request(mr.iid)
+        if is_rebased(fresh_mr, gl):
             continue
 
         pipelines = gl.get_merge_request_pipelines(mr)
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -1087,12 +1087,13 @@ def _process_omm_member(
 
     latest_status = pipelines[0].status
 
-    if latest_status == PipelineStatus.FAILED:
+    if latest_status in {PipelineStatus.FAILED, PipelineStatus.CANCELED}:
         logging.info([
             "omm-group",
             "eject-failed",
             gl.project.name,
             mr.iid,
+            latest_status,
         ])
         if not dry_run:
             gl.remove_label(mr, OMM_PENDING)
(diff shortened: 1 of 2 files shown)


TGPSKI:
LGTM — both fixes look correct and match the production evidence.

The serial-path re-fetch mirrors #5712 and addresses the stale list-API `sha` causing redundant rebases across cycles. Treating `CANCELED` like `FAILED` in OMM is the right call so a canceled member doesn’t hold the group active via the unhandled-status branch.

One non-blocking ask: the serial rebase tests still return the same mock from `get_merge_request` as from the batch list, and `is_rebased` only checks `iid`, so removing the re-fetch would still pass. A stale-vs-fresh SHA regression test like #5712 (for both strategies) would lock this in.

Nits: assert `remove_label` wasn’t called in the manual-status test; consider a distinct metric reason for canceled vs failed.

TGPSKI (approved):

PR #5725, opened 2026-08-04
chore: update stale uv.lock

jfchevrette:
## Summary
- Add an `uv-lock-test` Makefile target that runs `uv lock --check` to catch out-of-sync lockfiles in CI, and add missing targets to `.PHONY`
- Update the stale `uv.lock` so it passes the new check

## Test plan
- `make uv-lock-test`
- N/A for the lockfile change itself — lockfile-only diff

## Summary by CodeRabbit

* **Tests**
  * Expanded the complete test suite to verify that dependency lockfiles are up to date.
  * Added a dedicated validation check for lockfile consistency.
  * Updated build targets so the new validation runs automatically with all tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help build push rc build-test test-app test clean
+.PHONY: help build push rc build-test test-app test clean all-tests uv-lock-test linter-test types-test qenerate-test helm-test unittest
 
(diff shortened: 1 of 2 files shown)


jfchevrette:
uv lock --check fails in CI because it runs inside the container build, which intentionally omits qontract_api/ (one of the workspace members declared in pyproject.toml). With that directory missing, uv sees the
  workspace as incomplete and reports the lockfile as stale — even though it's actually correct. Locally you have the full checkout, so it passes.

PR #5727, opened 2026-08-04
chore: set desired uv version in .tool-versions

jfchevrette:
Without a version this does not work when using `mise`

## Summary by CodeRabbit

* **Chores**
  * Pinned the project’s `uv` tool version to 0.12.1 for more consistent development environments.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -4,4 +4,4 @@ terraform 1.6.6
 oc 4.19.0
 helm 3.21.3
 sloth 0.12.0
-uv latest
+uv 0.12.1
--- a/.tool-versions
+++ b/.tool-versions
@@ -4,4 +4,4 @@ terraform 1.6.6
 oc 4.19.0
 helm 3.21.3
 sloth 0.12.0
-uv
+uv latest


.tool-versions
```diff
@@ -4,4 +4,4 @@ terraform 1.6.6
 oc 4.19.0
 helm 3.21.3
 sloth 0.12.0
-uv
+uv latest
```
esron: is `latest` the best thing here?

jfchevrette: Indeed it is probably best to set it to what we have in pyproject.toml, which right now is `0.12.1`

My intent was to keep the same behavior but make it work with `mise`

jfchevrette:
```suggestion
uv 0.12.1
```
How about this?

esron: That looks better to me.

esron (approved):

.tool-versions
```diff
@@ -4,4 +4,4 @@ terraform 1.6.6
 oc 4.19.0
 helm 3.21.3
 sloth 0.12.0
-uv
+uv 0.12.1
```
hemslo, line 7: @jfchevrette this will cause renovate to generate seperate PR to bump version like this https://github.com/app-sre/qontract-reconcile/pull/5736, must be grouped with other version bumps. See https://github.com/app-sre/shared-pipelines/pull/384

jfchevrette: According to my and claude's understanding, that should do it: https://github.com/app-sre/qontract-reconcile/pull/5744

PR #5749, opened 2026-08-11
fix(qontract-utils): use safe YAML loader for untrusted OWNERS file content

chassing:
## Summary
- `OwnersParser` parsed OWNERS/OWNERS_ALIASES file content fetched from an arbitrary attacker-controlled repository using `create_ruamel_instance()`'s default round-trip loader ([APPSRE-14304](https://redhat.atlassian.net/browse/APPSRE-14304))
- Round-trip mode doesn't construct arbitrary Python objects from tags like PyYAML's default loader does, but it also doesn't reject them - non-standard YAML tags (e.g. `!!python/object/apply:...`) are silently loaded as plain data instead of raising, letting attacker-controlled tag content flow into the parsed approvers/reviewers lists undetected
- Add a `typ` parameter to `create_ruamel_instance()` (defaulting to the existing `"rt"` behavior, so no other caller is affected - verified against every call site in the codebase) and pass `typ="safe"` in `OwnersParser`, which only ever reads OWNERS content one-way and never needs round-trip formatting preservation
- The safe loader rejects unknown tags outright, which `get_owners()` already catches and treats as an unparsable file

## Test plan
- Added `test_get_owners_rejects_unsafe_yaml_tags`; confirmed it failed against the previous code (approver list silently populated from the tagged content) before the fix
- Added `test_create_ruamel_instance_defaults_to_round_trip_typ` and `test_create_ruamel_instance_safe_typ_rejects_python_object_tags`
- `make test` in `qontract_utils/` passes (ruff, mypy strict, pytest - 705 passed)

## Summary by CodeRabbit

- **New Features**
  - Added configurable YAML parsing modes, including round-trip and safe parsing.

- **Bug Fixes**
  - Improved handling of ownership files by rejecting unsafe YAML object tags.
  - Prevented potentially executable YAML content from being interpreted during parsing.

- **Tests**
  - Added coverage for default round-trip behavior and safe parsing protections.
  - Added regression coverage for unsafe ownership-file content.

### DIFF
--- a/qontract_utils/qontract_utils/ruamel.py
+++ b/qontract_utils/qontract_utils/ruamel.py
@@ -1,5 +1,5 @@
 from io import StringIO
-from typing import Any
+from typing import Any, Literal
 
 from ruamel.yaml.scalarstring import PreservedScalarString
 
@@ -19,7 +19,7 @@ def create_ruamel_instance(
     explicit_start: bool = False,
     width: int = 4096,
     pure: bool = False,
-    typ: str = "rt",
+    typ: Literal["rt", "safe"] = "rt",
 ) -> yaml.YAML:
     """Create a configured ruamel.yaml YAML instance.
 
--- a/qontract_utils/qontract_utils/ruamel.py
+++ b/qontract_utils/qontract_utils/ruamel.py
@@ -19,8 +19,17 @@ def create_ruamel_instance(
     explicit_start: bool = False,
     width: int = 4096,
     pure: bool = False,
+    typ: str = "rt",
 ) -> yaml.YAML:
-    ruamel_instance = yaml.YAML(pure=pure)
+    """Create a configured ruamel.yaml YAML instance.
+
+    typ defaults to "rt" (round-trip), needed by callers that read-modify-write
+    YAML and must preserve comments/formatting. Pass typ="safe" when parsing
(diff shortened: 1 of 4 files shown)


jfreixa (approved):

hemslo (approved):
Clean, well-scoped security fix. Backward-compatible default, correct use of safe loader for untrusted input, good test coverage at both unit and integration level.

Two optional nits (non-blocking):
- `typ: str` could be `typ: Literal["rt", "safe"]` for mypy-level typo protection
- Legacy `reconcile/utils/repo_owners.py` (lines 123, 183) has the same exposure — worth a follow-up if that code path is still exercised

LGTM 👍

chassing:
> Clean, well-scoped security fix. Backward-compatible default, correct use of safe loader for untrusted input, good test coverage at both unit and integration level.
> 
> Two optional nits (non-blocking):
> 
> * `typ: str` could be `typ: Literal["rt", "safe"]` for mypy-level typo protection

Done in 71fa62d

> * Legacy `reconcile/utils/repo_owners.py` (lines 123, 183) has the same exposure — worth a follow-up if that code path is still exercised
> 

Next on my list is https://redhat.atlassian.net/browse/APPSRE-13625 and `reconcile/utils/repo_owners.py` is obsolete.

PR #5751, opened 2026-08-11
fix(glhk) Break OMM group when force merge occurs

rhur-pixel:
## Summary

When an external MR is force-merged to the target branch during an active OMM group, the group now correctly dissolves instead of continuing to merge members onto an untested base.

## Approach

**SHA-matching**: After detecting HEAD moved forward, query all merged MRs with `omm-pending` (our own members), refresh each via the detail API, and check if `current_head` matches any of their `merge_commit_sha` / `squash_commit_sha`. If not, something external merged — dissolve.

This requires keeping `omm-pending` on successfully merged members (removed two `gl.remove_label` calls). `clear_omm_group` already cleans up stale labels on MERGED/CLOSED MRs when a group dissolves, so no label leakage.

## Changes

- **Removed** `gl.remove_label(mr, OMM_PENDING)` from merge-success and `GitlabMRClosedError` paths in `_process_omm_member`
- **Added** `_check_target_branch_integrity()` — extracted + extended the inline reachability check with SHA-matching detection
- **Replaced** the inline divergence check in `_process_omm_group` with a single call to the new function

## Test plan

- [x] External merge (HEAD not in OMM SHAs) → dissolves group
- [x] OMM member merge (HEAD matches member SHA) → group continues
- [x] Multiple merged members (HEAD matches any) → group continues
- [x] Force-push divergence (lead unreachable) → dissolves group (existing test)
- [x] All 161 existing tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Pending merge items now remain visible after successful merges or merge-closed errors for improved follow-up.
  * Added protection against unexpected target-branch changes and branch divergence.
  * Groups are cleared when unrecognized branch changes are detected.
  * Valid merge activity continues processing as expected.

* **Tests**
  * Expanded coverage for pending-item behavior and target-branch integrity checks.


TGPSKI (commented):
## Review

The scenario this targets is real: an external merge (GitLab UI / maintainer bypass) can FF onto the target branch during an active OMM window, and the existing `repository_compare` reachability check will pass because the lead is still an ancestor. Pending members then `skip_ci` rebase onto the new HEAD and merge on a stale SUCCESS pipeline that never validated against the external change.

That said, the effective risk is narrower than the PR description implies. For this to actually break master:

1. External merge must land during the 5-min window
2. It must touch the **same tenant** as a pending member (different tenant = safe by ADR-019's isolation argument)
3. It must touch **different files** (same files = rebase conflict = member ejected)
4. There must be semantic coupling (crossrefs) between the changed files

All four conditions must hold simultaneously. The rebase itself is already a partial guard — textual conflicts are caught.

### Implementation concerns

**SHA selection bug:**
```python
sha = fresh.merge_commit_sha or fresh.squash_commit_sha
```
GitLab populates both fields on squash-merge-with-merge-commit. If HEAD happens to equal `squash_commit_sha`, it won't be in `omm_shas` and the group will false-positive dissolve. Should add both when populated.

**Single-point check:** The integrity validation runs once at the top of `_process_omm_group`. An external merge landing *between* member processing (same reconcile loop) still slips through — same class of bug, smaller window.

**Label semantics change:** Keeping `omm-pending` on successfully merged MRs changes the label's meaning from "waiting to merge" to "participated in group." This could cause issues if two groups form back-to-back and stale labels from the prior group's merged MRs pollute the SHA lookup (the query filters by `target_branch` but not by group identity).

**API cost:** Every loop where HEAD != lead_sha now does a list call + N detail-API fetches for merged members. Acceptable for small groups but worth noting.

### Summary

The concept is sound — dissolving on unrecognized HEAD advance is the safe default when you can't verify tenant overlap with the external change. But the implementation needs the fixes above before it's merge-ready. The SHA selection bug in particular could cause false dissolves in production (liveness regression).

rhur-pixel:
@TGPSKI Think I can get your re-review on this? Thank you 🙏

TGPSKI (approved):
## Re-review (post c323e08)

Previous concerns addressed:

**SHA selection bug** — Fixed. Both `merge_commit_sha` and `squash_commit_sha` are now added independently to `omm_shas`, with a parametrized test covering the `both-shas-head-matches-squash` case. No more false-positive dissolves.

**Single-point check** — Adequately addressed by the inline comment. `is_rebased()` inside `_process_omm_member` re-fetches live HEAD on every call, so a mid-loop external merge prevents the *next* member from merging (fails `is_rebased` → takes skip_ci rebase → returns `active=True` → dissolves on the next reconcile iteration when the integrity check sees the unrecognized HEAD). The window-within-a-window is bounded to a single member merge, which is acceptable.

**Label semantics** — Still a minor concern (`omm-pending` on merged MRs until group close), but `clear_omm_group` already scans `state=MERGED` for stale labels, and `get_omm_pending_mrs` only queries `state=OPENED`, so operational impact is nil. Back-to-back group confusion is mitigated by the `target_branch` filter on the merged-member query.

CI is green. Logic is sound. LGTM.

PR #5754, opened 2026-08-12
[APPSRE-13941] Add automationTokens list-based rotation support to openshift-cluster-bots

fltataru:
## Context

This is **Phase 2 of 5** in the `automationTokens` rotation migration.

> **Design doc:** [Vault Secret Rotation](https://github.com/openshift-online/platform-engineering-enhancements/blob/main/app-sre/design-docs/vault-secret-rotation.md)

### Migration plan

| Phase | Repo | Description | Status |
|-------|------|-------------|--------|
| 1 | qontract-schemas | Add `automationTokens`/`clusterAdminAutomationTokens` fields to cluster schema | [#1050](https://github.com/app-sre/qontract-schemas/pull/1050) |
| 2 | qontract-reconcile | Prefer list tokens in `oc_connection_parameters`, implement list-based SA/Vault/MR workflow in `openshift-cluster-bots` | **This MR** |
| 3 | app-interface | Migrate clusters one-by-one: add `automationTokens` entries, let integration populate secrets, activate | Pending |
| 4 | qontract-schemas + qontract-reconcile | Remove singular fields from schema and all legacy code paths — must happen together, as ~30 GQL queries across qontract-reconcile still reference the singular `automationToken`/`clusterAdminAutomationToken` fields | Pending |
| 5 | app-interface | Remove singular `automationToken`/`clusterAdminAutomationToken` from all clusters | Pending |

> **Note:** Phase 4 combines the former steps 4 and 5. The singular schema fields cannot be removed until all qontract-reconcile GQL queries referencing them are updated in the same change.

### Design

The goal is to support **zero-downtime token rotation** for `openshift-cluster-bots`-managed clusters. Today each cluster has a single `automationToken` secret; rotating it requires a cutover window. The new model uses a list where multiple entries can coexist — the consumer picks the first `active: true, delete: false` entry with a populated secret — making rotations seamless.

Operators declare entries with `name`/`namespace`. The integration creates the ServiceAccount and Kubernetes secret, writes the token to Vault, and back-fills the `secret` reference via MR. The `active` flag signals readiness; `delete` flags entries for cleanup.

### Changes

**`reconcile/utils/oc_connection_parameters.py`**
- Add `_find_active_list_token()`: returns the secret from the first active, non-deleted list entry that has a secret populated
- `OCConnectionParameters.from_cluster()` now prefers list tokens via `getattr` (protocol-safe for GQL types that predate this field), falling back to singular `automationToken`/`clusterAdminAutomationToken`
- All legacy fallback paths marked with `TODO(APPSRE-13941)` for cleanup after full migration

**`reconcile/openshift_cluster_bots.py`**
- `filter_clusters()` splits clusters into `legacy` (singular field) and `list_based` (automationTokens entries present)
- `process_cluster_entries()` / `process_all_list_entries()`: handle create, sync, and delete lifecycle for each list entry — creates SA+secret on cluster, writes token to Vault, submits MR to back-fill `secret` reference in app-interface
- `submit_list_mr()`: generates targeted MR updating only the `automationTokens`/`clusterAdminAutomationTokens` fields
- `_entry_to_dict()`: uses `model_dump(by_alias=True, exclude_none=True)` to avoid empty `version:`/`format:` fields in auto-generated MRs
- `cluster_misses_bot_tokens()`: checks both list and singular sources for backward compatibility
- Legacy `create_all_bots()` path preserved for clusters not yet migrated, marked with `TODO(APPSRE-13941)`

**`reconcile/gql_definitions/fragments/oc_connection_cluster.gql` + `.py`**
- Added `automationTokens` and `clusterAdminAutomationTokens` to the shared `OcConnectionCluster` fragment so the preference logic in `oc_connection_parameters` fires for all ~32 integrations that use it
- `AutomationTokenEntryV1` class added to the generated Python module

**`reconcile/gql_definitions/openshift_cluster_bots/clusters.gql` + `.py`**
- Added `automationTokens` and `clusterAdminAutomationTokens` query fields

**Tests**
- 9 new unit tests for `_find_active_list_token` and `from_cluster` preference/fallback behavior in `test_oc_connection_parameters.py`
- `test_cluster_misses_bot_tokens_list_based` covering all list-based scenarios in `test_openshift_cluster_bots.py`

### Backward compatibility

All existing clusters using singular `automationToken`/`clusterAdminAutomationToken` continue to work without changes. The list path is only activated when `automationTokens` entries are present in the cluster definition.

## Test plan

- [ ] `uv run pytest reconcile/test/oc/test_oc_connection_parameters.py reconcile/test/test_openshift_cluster_bots.py` — 40 tests pass
- [ ] Dry-run against a cluster with `automationTokens` entries populates Vault and generates MR with correct `secret` fields (no empty `version:`/`format:`)
- [ ] Clusters without `automationTokens` continue through the legacy path unchanged

## Summary by CodeRabbit

* **New Features**
  * Added support for managing multiple automation tokens per cluster, including cluster-admin tokens.
  * Tokens can be created, synchronized with Vault, annotated, and deleted based on their status.
  * List-based tokens take precedence while retaining compatibility with existing single-token configurations.
  * Unmanaged secrets are preserved during synchronization.

* **Bug Fixes**
  * Improved token selection to ignore inactive, deleted, or incomplete entries.

* **Tests**
  * Added coverage for token selection, lifecycle processing, mixed configurations, and dry-run behavior.


reconcile/openshift_cluster_bots.py
```diff
@@ -279,6 +373,294 @@ def submit_mr(clusters: list[ClusterV1], config: Config) -> None:
...
+
+    if entry.secret is not None:
+        logging.info(f"[{cluster.name}] deleting vault secret {entry.secret.path}")
+        if not config.dry_run:
+            VaultClient.get_instance().delete(entry.secret.path)
+
+    return EntryResult(
+        entry=entry, cluster_admin=cluster_admin, vault_secret=None, action="deleted"
+    )
+
```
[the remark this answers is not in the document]
fltataru: This is intentional and follows the existing app-interface convention for deletions: set `delete: true`, let the integration clean up the actual resources (Kubernetes secret and Vault secret), then manually remove the entry from the YAML.

Auto-generating a removal MR was considered but rejected — operators explicitly removing the entry themselves keeps the pattern consistent with how deletions are handled elsewhere in app-interface, and gives a natural review point before the entry disappears from the source of truth.

The repeat Vault deletion on subsequent runs is accepted as a known trade-off: `delete: true` entries are rare and the operation is effectively a no-op once the secret is already gone.

fishi0x01:
Please update the `introspection.json` -> start local qontract-server with your schema changes, then `make gql-introspection`. Then the tests should pass.

chassing:
Please fix the failing `reconcile/test/test_typed_queries/test_saas_files.py::test_export_model` test

reconcile/openshift_cluster_bots.py
```diff
@@ -279,6 +394,294 @@ def submit_mr(clusters: list[ClusterV1], config: Config) -> None:
...
+    if entry.active is not None:
+        entry_dict["active"] = entry.active
+    if entry.delete is not None:
+        entry_dict["delete"] = entry.delete
+    if entry.secret is not None:
+        entry_dict["secret"] = entry.secret.model_dump(by_alias=True, exclude_none=True)
+    return entry_dict
+
+
+def _merge_entry_updates(
```
hemslo, line 604: **Bug: delete-flagged entries cause perpetual reconciliation.**

`_process_delete_entry` returns `EntryResult(vault_secret=None, action="deleted")` for every delete — that's correct, there's nothing to write back. But `_merge_entry_updates` filters results with `if r.vault_secret is not None`, so delete results are always excluded. If a cluster's only work is deletes, `result_by_key` ends up empty, this returns `None`, and `submit_list_mr` never submits an MR to remove the `delete: true` entry from app-interface.

Since the entry is never removed, `_entry_needs_processing` (`bool(entry.delete) or entry.secret is None`) returns `True` again next run — the integration reprocesses the same delete indefinitely. Worse, `entry.secret` is still populated in app-interface, so `_process_delete_entry` re-issues a Vault delete call on the (already-deleted) secret path every single reconcile cycle.

Suggest: have `_merge_entry_updates` (or its caller) drop the entry entirely from the merged list when the result is a completed delete, so the MR removes it from app-interface.

reconcile/openshift_cluster_bots.py
```diff
@@ -293,15 +696,49 @@ def create_all_bots(
...
+
+
+def cluster_needs_list_processing(cluster: ClusterV1) -> bool:
+    entries = list(cluster.automation_tokens or []) + list(
+        cluster.cluster_admin_automation_tokens or []
+    )
+    return any(_entry_needs_processing(entry) for entry in entries)
+
+
+def filter_clusters(
```
hemslo, line 711: **Bug: clusters with only inactive list entries and no singular token fall through both processing paths.**

`has_list_entries` is `True` whenever `automation_tokens` is non-empty, regardless of whether any entry is actually active. If a cluster has `automationToken=None` and an `automationTokens` list where every entry has `active: false` (secret already populated, not flagged for delete), then:

- `cluster_needs_list_processing` → `_entry_needs_processing` is `False` for that entry (`delete` is false, `secret` is populated), so the cluster isn't added to `list_based`.
- The `elif cluster_misses_bot_tokens(...)` legacy branch is unreachable because `has_list_entries` was `True`.

The cluster ends up in neither `legacy` nor `list_based` — nothing flags it as broken. Meanwhile `_find_active_list_token` in `oc_connection_parameters.py` skips the inactive entry and falls back to the now-null `automation_token`, leaving the cluster with no usable connection token. The qontract-schemas PR #1050 schema doesn't prevent this state (no constraint ties `active` entries to the presence of a singular token), and it's a plausible mid-migration state per the documented Phase 3 sequence (add entries → integration populates secret → activate) if steps happen out of order.

Suggest at minimum: log a warning when a cluster has list entries but none are active and no singular token exists, so it's visible instead of silently dropped.

reconcile/openshift_cluster_bots.py
```diff
@@ -279,6 +394,294 @@ def submit_mr(clusters: list[ClusterV1], config: Config) -> None:
...
+                    process_entry(kc.name, cluster, config, entry, cluster_admin)
+                )
+    except subprocess.CalledProcessError as e:
+        logging.error(e.stderr)
+        raise
+
+    return results
+
+
+def _entry_to_dict(entry: AnyTokenEntry) -> dict[str, object]:
```
hemslo, line 600: **Cleanup: `_entry_to_dict` reimplements `model_dump`.**

This manually reconstructs a dict from `entry.name`, `entry.namespace`, `entry.active`, `entry.delete`, `entry.secret` — which is what `entry.model_dump(by_alias=True, exclude_none=True)` already does (per the PR description, that's exactly the technique used elsewhere to avoid empty `version:`/`format:` fields). If a new field is added to `AutomationTokenEntry_v1` later, this function won't know about it and will silently drop it from generated MRs, corrupting the data file on merge.

Consider using `entry.model_dump(by_alias=True, exclude_none=True)` directly (with a nested dump for `secret` if needed) instead of hand-listing fields.

reconcile/openshift_cluster_bots.py
```diff
@@ -112,6 +175,50 @@ def oc_apply(kubeconfig: str, namespace: str, items: list[dict]) -> None:
...
+def is_managed_secret(secret: dict, sa_name: str) -> bool:
+    labels = secret.get("metadata", {}).get("labels") or {}
+    if labels.get(MANAGED_LABEL_KEY) == sa_name:
+        return True
+    # Adopt legacy secrets that predate the managed label but belong to the SA.
+    annotations = secret.get("metadata", {}).get("annotations") or {}
+    return annotations.get("kubernetes.io/service-account.name") == sa_name
+
+
+def secret_has_vault_annotation(secret: dict) -> bool:
```
hemslo: **Possible gap: `secret_has_vault_annotation` only checks the annotation key, not its value.**

```python
def secret_has_vault_annotation(secret: dict) -> bool:
    annotations = secret.get("metadata", {}).get("annotations") or {}
    return VAULT_PATH_ANNOTATION_KEY in annotations
```

If `vault_creds_path` ever changes (e.g. a Vault mount migration) while an existing K8s secret still carries the annotation with the *old* path, `_process_create_entry` treats it as already synced and skips re-writing to Vault at the new path — even though the referenced secret may no longer exist there. Consider comparing the annotation's value against the expected `vault_path`, not just presence of the key.

reconcile/openshift_cluster_bots.py
```diff
@@ -45,10 +61,29 @@ class Config(BaseModel):
     dry_run: bool
 
 
+def _has_active_token_with_secret(entries: Sequence[AnyTokenEntry] | None) -> bool:
```
hemslo, line 65: **Cleanup: duplicated "active token" predicate.**

```python
return any(e.active and not e.delete and e.secret is not None for e in entries)
```

This is the same predicate as `_find_active_list_token` in `reconcile/utils/oc_connection_parameters.py` (line 27): active, not deleted, secret populated. If the definition of a usable entry changes (e.g. adding an expiry check), it's easy to update one and miss the other, causing `cluster_misses_bot_tokens` and `OCConnectionParameters.from_cluster` to disagree on which entries are valid. Worth factoring into one shared helper.

reconcile/utils/oc_connection_parameters.py
```diff
@@ -24,6 +24,20 @@ class OCConnectionError(Exception):
     pass
 
 
+def _find_active_list_token(entries: list | None) -> HasSecret | None:
```
hemslo: **Type-check gap: bare `list` type bypasses mypy strict mode.**

```python
def _find_active_list_token(entries: list | None) -> HasSecret | None:
```

`list` has no type parameter here, so mypy strict (per AGENTS.md, `disallow_untyped_defs`) can't check the element type. The body accesses `.active`/`.delete`/`.secret` via `getattr(..., default)`, which is duck-typed and won't fail type-checking even if a caller passes the wrong element type. If a schema change ever removes `.active` from the model, this silently treats every entry as inactive instead of failing at type-check time. Consider `list[AnyTokenEntry] | None` or a `Protocol`/`TypeVar` bound, consistent with the `Cluster` Protocol already used in this file.

hemslo (commented):
Automated code review — found 6 issues across correctness and cleanup in the automationTokens list-based rotation logic.

reconcile/gql_definitions/openshift_serviceaccount_tokens/tokens.py
```diff
@@ -30,9 +30,27 @@
...
   }
+  clusterAdminAutomationTokens {
+    name
+    namespace
+    active
+    delete
+    secret {
+      ...VaultSecret
+    }
+  }
```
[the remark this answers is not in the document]
fltataru: Done — regenerated from a running qontract-server with the feature schemas rebased on top of upstream/main. Diff vs master adds only `AutomationTokenEntry_v1`, the two new `Cluster_v1` fields (`automationTokens`, `clusterAdminAutomationTokens`), and `managedRobotAccounts` (an upstream schema change not yet picked up by master's introspection.json).

reconcile/openshift_cluster_bots.py
```diff
@@ -279,6 +395,294 @@ def submit_mr(clusters: list[ClusterV1], config: Config) -> None:
...
+            },
+            decode_base64=False,
+        )
+        oc_annotate_secret(kubeconfig, entry.namespace, entry.name, vault_path)
+        return EntryResult(
+            entry=entry,
+            cluster_admin=cluster_admin,
+            vault_secret=vault_ref,
+            action="created",
+        )
```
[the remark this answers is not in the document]
fltataru: Pre-existing behavior, not introduced by this PR. The dedicated-admin namespace is a prerequisite for any cluster being onboarded — if it doesn't exist, that's a misconfiguration in app-interface that should surface as a failure rather than be silently skipped. The `create_namespace=cluster_admin` flag is intentional: only cluster-admin entries provision their own namespace.

reconcile/test/test_openshift_cluster_bots.py
```diff
@@ -159,33 +241,272 @@ def test_retrieve_token(mocker: MockerFixture) -> None:
...
+
+    assert legacy_result == [legacy_needs_work]
+    assert list_result == [list_needs_work]
+
+
+def test_process_entry_create_new(
+    mocker: MockerFixture, config: ocb.Config, cluster: Callable
+) -> None:
+    c = cluster(automation_tokens=[automation_token_entry(name="tok", namespace="ns")])
+    entry = c.automation_tokens[0]
```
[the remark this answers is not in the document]
fltataru: Fixed — added `assert c.automation_tokens` before each of the six indexing sites.

reconcile/test/test_openshift_cluster_bots.py
```diff
@@ -159,33 +241,272 @@ def test_retrieve_token(mocker: MockerFixture) -> None:
...
+    )
+
+    result = ocb.process_entry("kubeconfig", c, config, entry, cluster_admin=False)
+
+    assert result.action == "skipped"
+    assert oc_mock.call_count == 1
+    vault_mock.return_value.delete.assert_not_called()
 
 
 class Mocks:  # ruff: ignore[class-as-data-structure]
```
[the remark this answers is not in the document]
fltataru: Fixed — converted `Mocks` to a `@dataclass`, which resolves the `class-as-data-structure` finding properly and removes the need for the suppression entirely.

reconcile/openshift_cluster_bots.py
fltataru: Warning log added in filter_clusters when a cluster has list entries but none are active with a secret and no singular token fallback exists.

reconcile/openshift_cluster_bots.py
fltataru: eplaced the manual field-by-field reconstruction with entry.model_dump(by_alias=True, exclude_none=True). New schema fields will no longer be silently dropped from generated MRs.

reconcile/openshift_cluster_bots.py
fltataru: secret_has_vault_annotation now takes vault_path: str and compares the annotation value against it. Both call sites in _process_create_entry updated. Test fixture updated to use the correct path.

reconcile/openshift_cluster_bots.py
fltataru: Extracted is_active_token_entry(entry: TokenEntry) -> bool into oc_connection_parameters.py. Both _has_active_token_with_secret (in openshift_cluster_bots.py) and _find_active_list_token (in oc_connection_parameters.py) now call it — one definition, no drift risk.

reconcile/utils/oc_connection_parameters.py
fltataru: Added TokenEntry Protocol (read-only properties for active, delete, secret) to oc_connection_parameters.py. _find_active_list_token is now typed list[TokenEntry] | None and uses direct attribute access instead of getattr.

reconcile/openshift_cluster_bots.py
fltataru: Deleting the entry via MR would still make the integration attempt the delete operations on every reconcile cycle until the MR is merged. The only way to prevent that is checking the content of the open MR, but the added complexity isn't worth it (in my opinion) compared to the current behavior. The K8s delete is already idempotent (we check existence first) and the Vault delete is a no-op on an already-deleted path.

Token rotation will be handled by us and won't be frequent. I'll make sure to document the full flow in the SOP that follows this MR.

hemslo (approved):
LGTM!

PR #5756, opened 2026-08-13
fix(qontract-utils): set User-Agent header on all OCM/HTTP API clients

chassing:
## Summary

On 2026-07-21, CS pods crashed under request queue buildup. Alongside the primary
CS-side causes (N+1 queries, rate limiter bug), CS traced significant `list-clusters`
API traffic to the shared `sd-app-sre-ocm-sa` service account, but couldn't attribute
it to a specific caller/version because no custom `User-Agent` was set — all traffic
showed up as generic `python-requests`/`httpx` traffic.

- Set a custom `User-Agent` (`qontract-reconcile/<version>`) on `OCMBaseClient`
  (`reconcile/utils/ocm_base_client.py`), resolved via `importlib.metadata`, sent on
  every request including the OAuth token exchange. It also accepts an optional
  `user_agent` override now, so a CLI tool or `qontract-api` task wrapping it can
  identify itself distinctly from the default reconcile loop.
- Added `qontract_utils/qontract_utils/user_agent.py`, a single shared default
  (`qontract-utils/<version>`) for every HTTP-based Layer 1 client in
  `qontract_utils`: `OcmApi`, `GlitchtipApi`, `KeycloakApi`, `GithubOrgApi`,
  `GitHubRepoApi`, `GitLabRepoApi`. Each client also accepts an optional
  `user_agent` override, but nothing in `qontract_api` needs to configure or pass
  one — the default is enough for OCM/CS (or any other backend) to distinguish this
  traffic from `qontract-reconcile`'s.
- The `importlib.metadata.version` → `PackageNotFoundError` → `"unknown"` fallback
  pattern is now a single shared `resolve_version()` helper in
  `qontract_utils/user_agent.py`, used by both `qontract_utils` and
  `reconcile/utils/ocm_base_client.py`, instead of being duplicated in each module.
- The Kubernetes client (`qontract_utils/kubernetes/client.py`, lightkube-based) is
  intentionally excluded — lightkube has no headers/user_agent hook, and that
  traffic goes to per-tenant clusters rather than a shared rate-limited backend
  like OCM/CS.
- Per review feedback: the User-Agent is now applied before the OAuth2
  client-credentials token request fires in both `OCMBaseClient` and `OcmApi`, so
  the token exchange itself is attributable too, not just the subsequent API calls.
- Per review feedback: `user_agent` sits after `hooks` in every constructor
  (`OcmApi`, `KeycloakApi`, `GlitchtipApi`, `GithubOrgApi`), consistently, closing a
  latent positional-argument risk in `GlitchtipApi`/`GithubOrgApi`.
- Per review feedback: `GitHubRepoApi` and `GitLabRepoApi` (the vcs provider clients
  used for MR/PR operations) now also send a configurable User-Agent — they were the
  last HTTP-based Layer 1 clients still sending the third-party library default
  (PyGithub/python-gitlab). Verified both libraries actually apply the `user_agent`
  kwarg to the real request header rather than ignoring it.

Related: [APPSRE-14889](https://redhat.atlassian.net/browse/APPSRE-14889)

## Test plan

- [x] Added `test_user_agent_header_is_set`,
      `test_user_agent_header_is_sent_on_token_request`, and
      `test_user_agent_header_can_be_overridden` in
      `reconcile/test/ocm/test_utils_ocm_base.py` (verified to fail without the fix)
- [x] Added User-Agent default/override/token-request tests for `OcmApi`,
      `GlitchtipApi`, `KeycloakApi`, `GithubOrgApi`, `GitHubRepoApi`, and
      `GitLabRepoApi` in `qontract_utils/tests/`
- [x] Added `qontract_utils/tests/test_user_agent.py` covering `resolve_version()`'s
      installed-version and fallback paths
- [x] Added an autouse fixture in `qontract_utils/tests/ocm_api/test_client.py` to
      close every `OcmApi` created in the file after each test (review feedback)
- [x] `uv run pytest reconcile/test/ocm/` — 391 passed
- [x] `uv run pytest` (qontract_utils, full suite) — 732 passed
- [x] `uv run pytest -k "ocm or glitchtip or github or keycloak or sso_client"` (qontract_api) — 235+ passed
- [x] `uv run mypy` clean on all changed files
- [x] `uv run ruff check` / `ruff format --check` clean on all changed files

## Summary by CodeRabbit

- **New Features**
  - API requests now include a descriptive `User-Agent` identifying the client and its version.
  - Added consistent default identification across supported OCM, GitHub, GlitchTip, Keycloak, GitLab, and reconciliation requests.
  - Applications can provide a custom `User-Agent` for supported API clients.

- **Documentation**
  - API client options now describe default and customizable request-identification behavior.

- **Tests**
  - Added coverage confirming default and custom `User-Agent` values are sent correctly.

### DIFF
--- a/qontract_utils/qontract_utils/github_org/api.py
+++ b/qontract_utils/qontract_utils/github_org/api.py
@@ -116,18 +116,18 @@ class GithubOrgApi:
         self,
         token: str,
         base_url: str = "https://api.github.com",
-        user_agent: str = DEFAULT_USER_AGENT,
         hooks: Hooks | None = None,  # ruff: ignore[unused-method-argument] - handled by @with_hooks
+        user_agent: str = DEFAULT_USER_AGENT,
     ) -> None:
         """Initialize GithubOrgApi.
 
         Args:
             token: GitHub API token
             base_url: GitHub API base URL (override for GHE)
+            hooks: Optional custom hooks merged with built-in hooks
             user_agent: User-Agent header sent with every request. Defaults to
                 identifying qontract-utils itself; callers embedded in a larger
                 service should pass their own app name/version instead.
-            hooks: Optional custom hooks merged with built-in hooks
         """
         self._token = token
         self._base_url = base_url
--- a/qontract_utils/qontract_utils/github_org/api.py
+++ b/qontract_utils/qontract_utils/github_org/api.py
@@ -23,6 +23,7 @@ from prometheus_client import Counter, Histogram
 
 from qontract_utils.hooks import Hooks, invoke_with_hooks, with_hooks
 from qontract_utils.metrics import DEFAULT_BUCKETS_EXTERNAL_API
+from qontract_utils.user_agent import DEFAULT_USER_AGENT
 
 logger = structlog.get_logger(__name__)
 
@@ -115,6 +116,7 @@ class GithubOrgApi:
         self,
         token: str,
         base_url: str = "https://api.github.com",
+        user_agent: str = DEFAULT_USER_AGENT,
         hooks: Hooks | None = None,  # ruff: ignore[unused-method-argument] - handled by @with_hooks
     ) -> None:
         """Initialize GithubOrgApi.
@@ -122,11 +124,15 @@ class GithubOrgApi:
         Args:
             token: GitHub API token
             base_url: GitHub API base URL (override for GHE)
+            user_agent: User-Agent header sent with every request. Defaults to
+                identifying qontract-utils itself; callers embedded in a larger
+                service should pass their own app name/version instead.
             hooks: Optional custom hooks merged with built-in hooks
         """
         self._token = token
         self._base_url = base_url
-        self._gh = Github(token, base_url=base_url)
+        self._user_agent = user_agent
+        self._gh = Github(token, base_url=base_url, user_agent=user_agent)
 
     def _paginated_get(self, path: str) -> list[dict[str, Any]]:
         """Perform a paginated GET against the GitHub REST API.
@@ -141,6 +147,7 @@ class GithubOrgApi:
         headers = {
             "Authorization": f"token {self._token}",
             "Accept": "application/vnd.github.v3+json",
+            "User-Agent": self._user_agent,
         }
         items: list[dict[str, Any]] = []
 
--- a/qontract_utils/qontract_utils/ocm_api/client.py
+++ b/qontract_utils/qontract_utils/ocm_api/client.py
@@ -10,7 +10,6 @@ from __future__ import annotations
 import contextvars
 import time
 from dataclasses import dataclass
-from importlib.metadata import PackageNotFoundError, version
 from typing import TYPE_CHECKING, Self
 
 import httpx2
@@ -36,6 +35,7 @@ from qontract_utils.ocm_api.models import (
     OcmSubscription,
     OcmSubscriptionLabel,
 )
+from qontract_utils.user_agent import DEFAULT_USER_AGENT
 
 if TYPE_CHECKING:
     from qontract_utils.ocm_api._raw_client import RawCluster, RawSubscription
@@ -46,13 +46,6 @@ logger = structlog.get_logger(__name__)
 TIMEOUT = 60.0
 MAX_RETRIES = 3
 CHUNK_SIZE = 100
-
-try:
-    _QONTRACT_UTILS_VERSION = version("qontract-utils")
-except PackageNotFoundError:
-    _QONTRACT_UTILS_VERSION = "unknown"
-
-DEFAULT_USER_AGENT = f"qontract-utils/{_QONTRACT_UTILS_VERSION}"
 
 # Prometheus metrics
 ocm_request = Counter(
--- a/qontract_utils/qontract_utils/ocm_api/client.py
+++ b/qontract_utils/qontract_utils/ocm_api/client.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import contextvars
 import time
 from dataclasses import dataclass
+from importlib.metadata import PackageNotFoundError, version
 from typing import TYPE_CHECKING, Self
 
 import httpx2
@@ -45,6 +46,13 @@ logger = structlog.get_logger(__name__)
 TIMEOUT = 60.0
 MAX_RETRIES = 3
 CHUNK_SIZE = 100
+
+try:
+    _QONTRACT_UTILS_VERSION = version("qontract-utils")
+except PackageNotFoundError:
+    _QONTRACT_UTILS_VERSION = "unknown"
+
+DEFAULT_USER_AGENT = f"qontract-utils/{_QONTRACT_UTILS_VERSION}"
 
 # Prometheus metrics
 ocm_request = Counter(
@@ -171,6 +179,7 @@ class OcmApi:
         hooks: Hooks | None = None,
         timeout: float = TIMEOUT,
         max_retries: int = MAX_RETRIES,
+        user_agent: str = DEFAULT_USER_AGENT,
     ) -> None:
         """Initialize the OCM API client.
 
@@ -181,6 +190,10 @@ class OcmApi:
             access_token_client_secret: OAuth2 client secret (already resolved plaintext)
             timeout: API request timeout in seconds (default: 60)
             max_retries: number of transport-level retries for failed requests (default: 3)
+            user_agent: User-Agent header sent with every request. Defaults to
+                identifying qontract-utils itself; callers embedded in a larger
+                service (e.g. qontract-api) should pass their own app name/version
+                so OCM can attribute traffic to the actual caller.
             hooks: Optional custom hooks to merge with built-in hooks. Not read here -
                 @with_hooks intercepts and merges it into self._hooks before this body runs.
         """
@@ -195,7 +208,10 @@ class OcmApi:
         )
         self._client = httpx2.Client(
             base_url=url,
-            headers={"Authorization": f"Bearer {access_token}"},
+            headers={
+                "Authorization": f"Bearer {access_token}",
+                "User-Agent": user_agent,
+            },
             timeout=timeout,
             transport=httpx2.HTTPTransport(retries=max_retries),
         )
--- a/qontract_utils/qontract_utils/ocm_api/client.py
+++ b/qontract_utils/qontract_utils/ocm_api/client.py
@@ -118,6 +118,7 @@ def _fetch_access_token(
     access_token_client_id: str,
     access_token_client_secret: str,
     timeout: float,
+    user_agent: str,
 ) -> str:
     response = httpx2.post(
         access_token_url,
@@ -126,6 +127,7 @@ def _fetch_access_token(
             "client_id": access_token_client_id,
             "client_secret": access_token_client_secret,
         },
+        headers={"User-Agent": user_agent},
         timeout=timeout,
     )
     response.raise_for_status()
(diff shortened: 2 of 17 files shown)


qontract_utils/qontract_utils/ocm_api/client.py
```diff
@@ -181,6 +190,10 @@ def __init__(
             access_token_client_secret: OAuth2 client secret (already resolved plaintext)
             timeout: API request timeout in seconds (default: 60)
             max_retries: number of transport-level retries for failed requests (default: 3)
+            user_agent: User-Agent header sent with every request. Defaults to
+                identifying qontract-utils itself; callers embedded in a larger
+                service (e.g. qontract-api) should pass their own app name/version
+                so OCM can attribute traffic to the actual caller.
```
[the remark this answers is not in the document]
chassing: Good catch. Fixed in 4c55931c — the User-Agent is now set on the session/client headers before the OAuth token request fires in both `OCMBaseClient` (reconcile) and `OcmApi` (qontract_utils), so the token exchange is attributable too, not just the subsequent OCM API calls. Added tests for both, verified they fail without the fix.

qontract_utils/tests/ocm_api/test_client.py
```diff
@@ -74,6 +74,40 @@ def test_bearer_token_sent_on_subsequent_requests(httpserver: HTTPServer) -> Non
...
+    api = _make_ocm_api(httpserver, token="test-token")
+    httpserver.expect_request(LABELS_PATH, method="GET").respond_with_json(
+        _empty_page()
+    )
+
+    api.get_labels(Filter().eq("key", "x"))
+
+    label_requests = [req for req, _ in httpserver.log if req.path == LABELS_PATH]
+    assert label_requests[0].headers["User-Agent"].startswith("qontract-utils/")
+
```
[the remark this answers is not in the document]
chassing: Fixed in 4c55931c — added an autouse fixture in this test file that closes every `OcmApi` created via `_make_ocm_api` after each test. Applied it to the whole file (not just the 2 new tests) since the same gap existed throughout.

qontract_utils/qontract_utils/github_org/api.py
```diff
@@ -115,18 +116,23 @@ def __init__(
         self,
         token: str,
         base_url: str = "https://api.github.com",
+        user_agent: str = DEFAULT_USER_AGENT,
         hooks: Hooks | None = None,  # ruff: ignore[unused-method-argument] - handled by @with_hooks
```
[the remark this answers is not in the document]
chassing: Good catch. Fixed in 77e89df7 — moved user_agent after hooks in both GlitchtipApi and GithubOrgApi, matching the OcmApi/KeycloakApi convention. No current caller was affected (nothing passes hooks positionally today), but this closes the latent risk.

qontract_utils/qontract_utils/user_agent.py
```diff
@@ -0,0 +1,18 @@
+"""Default User-Agent for qontract_utils HTTP clients.
+
+Every Layer 1 API client sends this by default so external services can
+attribute traffic to qontract-utils. Callers embedded in a larger service
```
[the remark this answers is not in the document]
chassing: Fixed in 77e89df7 — updated the docstring to say "Every HTTP-based Layer 1 API client".

qontract_utils/tests/ocm_api/test_client.py
```diff
@@ -21,25 +23,42 @@
 CLUSTERS_PATH = "/api/clusters_mgmt/v1/clusters"
 IDPS_PATH = "/api/clusters_mgmt/v1/clusters/cluster-1/identity_providers"
 
+_created_apis: list[OcmApi] = []
+
+
+@pytest.fixture(autouse=True)
+def _close_ocm_apis() -> Generator[None]:
```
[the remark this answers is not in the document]
chassing: Checked this against our actual project config rather than a standalone --strict run: qontract_utils/pyproject.toml has `select = ["ALL"]` for ruff, which includes UP043 (unnecessary-default-type-args) — it auto-rewrites `Generator[None, None, None]` back to bare `Generator[None]` (this project targets a Python version where Generator has default type params). I tried the fully-parameterized form locally and `ruff check` immediately reverted it. Our actual `uv run mypy` (no --strict override) already passes with the bare form, and there's a pre-existing identical pattern in qontract_utils/tests/conftest.py::enable_retry. Keeping it as-is since the suggested fix conflicts with our own enforced lint rule and would just get undone by `make format`/CI.

reconcile/utils/ocm_base_client.py
```diff
@@ -52,6 +60,7 @@ def __init__(
         self._access_token_url = access_token_url
         self._url = url
         self._session = session or Session()
```
hemslo, line 58: **simplification**: `OCMBaseClient` hardcodes `USER_AGENT` with no override parameter.

Unlike all the `qontract_utils` clients (`OcmApi`, `KeycloakApi`, etc.) which accept an optional `user_agent` kwarg, `OCMBaseClient` hardcodes `qontract-reconcile/<version>` with no way for a caller to override it. If a CLI tool or a `qontract-api` task wraps `OCMBaseClient`, the traffic is indistinguishable from the default reconcile loop.

Consider adding an optional `user_agent: str = USER_AGENT` parameter to `__init__` for consistency with the Layer 1 clients.

qontract_utils/qontract_utils/github_org/api.py
```diff
@@ -116,17 +117,22 @@ def __init__(
...
         base_url: str = "https://api.github.com",
         hooks: Hooks | None = None,  # ruff: ignore[unused-method-argument] - handled by @with_hooks
+        user_agent: str = DEFAULT_USER_AGENT,
     ) -> None:
         """Initialize GithubOrgApi.
 
         Args:
             token: GitHub API token
             base_url: GitHub API base URL (override for GHE)
             hooks: Optional custom hooks merged with built-in hooks
```
hemslo, line 127: **correctness**: `GithubOrgApi` sets `user_agent` on both PyGithub and raw `requests.Session`, but these are independent HTTP paths.

PyGithub manages its own `User-Agent` header internally. The `user_agent` parameter passed to `Github(...)` may be silently ignored or overwritten by PyGithub's own logic on its HTTP path, while the `_paginated_get` path correctly uses `self._user_agent` in the headers dict.

Worth verifying that `Github(token, base_url=..., user_agent=...)` actually propagates the custom UA on all PyGithub-routed requests (e.g. `get_organization`, `get_repo`), or document the limitation.

qontract_utils/qontract_utils/user_agent.py
```diff
@@ -0,0 +1,18 @@
...
+can attribute traffic to qontract-utils. Callers embedded in a larger service
+(e.g. qontract-api) should pass their own app name/version instead so the
+embedding service is what shows up in the traffic, not the shared library.
+"""
+
+from __future__ import annotations
+
+from importlib.metadata import PackageNotFoundError, version
+
+try:
```
hemslo: **simplification**: The version-detection pattern (`importlib.metadata.version` → `PackageNotFoundError` → fallback) is duplicated here and in `reconcile/utils/ocm_base_client.py`.

If the fallback logic ever needs to change (e.g. adding a git hash for dev builds), it must be updated in two places. Consider extracting a shared `_resolve_version(package_name, fallback)` helper into this module and having `ocm_base_client.py` call it, or at minimum importing the pattern from one canonical location.

qontract_utils/qontract_utils/github_org/api.py
```diff
@@ -116,17 +117,22 @@ def __init__(
         token: str,
         base_url: str = "https://api.github.com",
         hooks: Hooks | None = None,  # ruff: ignore[unused-method-argument] - handled by @with_hooks
+        user_agent: str = DEFAULT_USER_AGENT,
```
hemslo, line 120: **completeness**: `GitHubRepoApi` (`qontract_utils/qontract_utils/vcs/providers/github_client.py`) was not updated with a `user_agent` parameter.

This sibling client (used for MR/PR operations via PyGithub) still sends the default PyGithub/requests `User-Agent`. Since this PR aims to make all HTTP traffic attributable, the per-repo client has the same observability gap. Might be intentional if it's a follow-up, but worth noting.

hemslo (commented):
Code review — 4 findings (automated)

reconcile/utils/ocm_base_client.py
chassing: Fixed in d8ff7d52 — added an optional `user_agent: str = USER_AGENT` parameter to `OCMBaseClient.__init__`, matching the qontract_utils Layer 1 clients. Added a test (`test_user_agent_header_can_be_overridden`) that overrides it and asserts the header on both the token request and a subsequent API call.

qontract_utils/qontract_utils/github_org/api.py
chassing: Checked PyGithub's `Requester.__requestEncode` directly: it unconditionally sets `requestHeaders["User-Agent"] = self.__userAgent` on every single request, where `self.__userAgent` is exactly the value passed to `Github(..., user_agent=...)`. So the `user_agent` kwarg is not silently ignored or overwritten — both the PyGithub HTTP path and the raw `requests.Session` path in `_paginated_get` end up sending the same custom User-Agent. No code change needed here.

qontract_utils/qontract_utils/user_agent.py
chassing: Fixed in d8ff7d52 — extracted `resolve_version(package, fallback="unknown")` into `qontract_utils/user_agent.py`. `DEFAULT_USER_AGENT` now uses it, and `reconcile/utils/ocm_base_client.py`'s `USER_AGENT` imports and calls the same helper instead of duplicating the try/except. (`reconcile/` is allowed to import from `qontract_utils` per AGENTS.md's import rules.)

qontract_utils/qontract_utils/github_org/api.py
chassing: Fixed in 8f610c8c — added the same `user_agent: str = DEFAULT_USER_AGENT` parameter to `GitHubRepoApi` (and to `GitLabRepoApi` too, since it had the identical gap), passed through to the underlying `Github(...)`/`gitlab.Gitlab(...)` clients. Verified both PyGithub and python-gitlab actually apply the `user_agent` kwarg to the real request header, not just accept-and-ignore it.

reconcile/utils/ocm_base_client.py
hemslo: Resolved in d8ff7d52 — `OCMBaseClient.__init__` now accepts `user_agent: str = USER_AGENT` with an override test.

qontract_utils/qontract_utils/user_agent.py
hemslo: Resolved in d8ff7d52 — extracted `resolve_version()` helper in `user_agent.py`, now used by both `user_agent.py` and `ocm_base_client.py`.

qontract_utils/qontract_utils/github_org/api.py
hemslo: Resolved in 8f610c8c — `GitHubRepoApi` and `GitLabRepoApi` both now accept `user_agent` with tests.

hemslo (approved):
All review findings addressed. Nice work extracting `resolve_version()` and extending coverage to `GitHubRepoApi`/`GitLabRepoApi` as well.

PR #5759, opened 2026-08-13
APPSRE-15090: Use a context manager for DashdotDB tokens

lisa:
Use a context manager for DashdotDB tokens

## Problem

[dashdotdb_base.py](reconcile/dashdotdb_base.py) provided a `_get_token()` method that acquired authentication tokens by setting an instance attribute as a side effect and returning None. Every caller had to remember to inspect the attribute after calling the method. Two of three callers (DVO and SLO) did not perform this check, allowing unauthenticated `POST`s when token acquisition failed.

## Solution

A `_token()` context manager on `DashdotdbBase` replaces the manual `_get_token()` / `_close_token()` pair. Token acquisition, error handling, and cleanup are encapsulated:

```python
try:
    with self._token():
        self.post(data)
except DashdotdbTokenError:
    LOG.error(...)
```

### `dry_run=True`

Yields None. No HTTP calls are made and no cleanup runs. Individual `_post()` methods on subclasses already guard their HTTP calls on not `self.dry_run`.

### Acquisition succeeds

Sets `self.dashdotdb_token`, yields the token string, and calls `_close_token()` on exit, even if the body raises an exception.

### Acquisition fails

Raises `DashdotdbTokenError` before yield. The body never executes, so no `POST` goes out without authentication. No token-close is attempted.

## Additional changes

`DashdotdbDORA.post()` gains a `dry_run` guard, matching the pattern already used by DVO and SLO.

Co-Authored-By: Claude Opus 4.6

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved credential handling to reliably release tokens, including when errors occur.
  - Prevented deployment, SLO, and DVO updates when token acquisition fails or returns an empty token.
  - Dry-run operations now skip network submissions as expected.
  - Improved error reporting for token acquisition failures.

- **Tests**
  - Added coverage for successful, failed, and interrupted token workflows.
  - Verified cleanup failures do not hide processing errors.
  - Confirmed dry runs make no HTTP requests.


esron (approved):
LGTM

hemslo (commented):
Code review — 4 findings (automated, adversarially verified)

reconcile/dashdotdb_dora.py
```diff
@@ -480,20 +483,19 @@ def _github_compare_commits(self, rc: RepoChanges, repo: str) -> list[Commit]:
...
-            return
-        try:
-            self.post({"deployments": deployments})
-        finally:
-            self._close_token()
 
     def post(self, data: Mapping[str, Any]) -> None:
+        if self.dry_run:
+            return
         endpoint = f"{self.dashdotdb_url}/api/v1/dora"
```
hemslo, line 506: **dry_run observability gap** — `post()` returns immediately with no logging in dry_run mode. DVO's `_post()` logs `Processing N metrics for cluster` and `DVO data synced` even in dry_run; SLO's `_post()` logs `About to POST SLO JSON item` at DEBUG. DORA is completely silent — operators running dry_run to verify wiring get no visibility into what _would_ be posted or how many deployments were collected.

Suggestion: add a `LOG.info` before the return, e.g. `LOG.info("%s dry_run: would post %d deployments", self.logmarker, len(data.get("deployments", [])))` to restore observability.

reconcile/dashdotdb_dvo.py
```diff
@@ -238,24 +239,32 @@ def run(self, cname: str | None = None) -> None:
...
+                        func=self._post,
+                        iterable=validations,
+                        thread_pool_size=self.thread_pool_size,
+                    )
+        except DashdotdbTokenError:
+            LOG.error(
+                "%s error acquiring token for %s, skipping",
+                self.logmarker,
+                self.scope,
             )
```
hemslo, line 267: **nit: triplicated boilerplate** — This `try/except DashdotdbTokenError` block is near-identical across all three callers (DORA `_post_deployments`:486, DVO `run`:242, SLO `_post`:104). The only textual difference is `skipping POST` vs `skipping`. If the error-handling convention ever changes (e.g. retry logic, metrics), all 3 must be updated in lockstep.

A base-class helper like `_with_token(action: Callable)` would centralize the policy. Not blocking, but worth considering.

reconcile/dashdotdb_base.py
```diff
@@ -56,43 +61,57 @@ def __init__(
...
+        token = response.text.replace('"', "").strip()
+        if not token:
+            raise DashdotdbTokenError(
+                f"error retrieving token for {self.scope}: empty token"
             )
-            return
-        self.dashdotdb_token = response.text.replace('"', "").strip()
+        self.dashdotdb_token = token
+        try:
+            yield self.dashdotdb_token
```
hemslo: **nit: unused yield value** — `_token()` yields `self.dashdotdb_token`, but all 3 callers use `with self._token():` (no `as`), ignoring the yielded value. They rely on `_do_post()` reading `self.dashdotdb_token` (the instance attribute set at line 95). The `Generator[str | None]` return type suggests callers should capture the yielded value, but nobody does.

Simplifying to `yield` (or `yield None`) would clarify that the context manager is purely for lifecycle management, not for passing a token to the caller.

reconcile/test/test_dashdotdb_dora.py
```diff
@@ -254,22 +254,46 @@ def test_get_repo_changes(mocker: MockerFixture) -> None:
...
     mocker.patch("reconcile.dashdotdb_dora.DashdotdbDORA.__init__").return_value = None
     d = DashdotdbDORA(False, "1", 1)
     d.dry_run = False
-    d.dashdotdb_token = None
     d.logmarker = "DORA"
     d.scope = "dora"
 
-    d._get_token = MagicMock()  # type: ignore[method-assign]
+    @contextmanager
+    def _failing_token():  # type: ignore[no-untyped-def]
```
hemslo: **nit: missing type annotation** — `_failing_token` has a `# type: ignore[no-untyped-def]` suppression — the only one in `reconcile/test/`. Fix:
```python
def _failing_token() -> Iterator[None]:
```
(with `from collections.abc import Iterator`)

hemslo (commented):
Re-reviewed after `a2c6936` ("PR Feedback"). All 4 findings from the previous review are addressed:

1. **DORA dry_run logging** — Fixed. `post()` now logs deployment count in dry_run, with a new test (`test_post_logs_deployment_count_in_dry_run`).
2. **Triplicated boilerplate** — Acknowledged with a documented consolidation path in `dashdotdb_base.py` and deferred rationale (DVO's multi-statement body needs extracting first). Reasonable.
3. **Unused yield value** — Fixed. `_token()` now returns `Iterator[None]` and yields nothing. Tests updated to assert `token is None`.
4. **Missing type annotation** — Fixed. `_failing_token() -> Iterator[None]` with proper imports.

New `test_dashdotdb_base.py` (199 lines) has solid coverage: dry_run skip, successful acquire/release, failed acquisition, transport errors, empty token rejection, body-exception-still-releases, and release-failure-doesn't-mask-body-exception.

No new issues found. LGTM.

hemslo (approved):
All findings addressed. LGTM.

PR #5769, opened 2026-08-17
feat(qontract-cli): add RDS EOL and next-version columns

RH-tj:
## Summary

- [APPSRE-12579](https://redhat.atlassian.net/browse/APPSRE-12579): improve `qontract-cli get rds` output
- Remove the `storage_type` column
- Add `auto_minor_version_upgrade`, `eol_date`, and `next_version`
- `eol_date` comes from the AWS RDS calendar dataset in [app-sre/aws-generated-data](https://github.com/app-sre/aws-generated-data/blob/main/output/rds_eol.yaml)
- `next_version` is the next non-EOL version in the **same major** line
- Lookup is an exact `(engine, version)` match; unmatched versions stay blank
- Fetch failures leave EOL columns blank instead of failing the command

## Test plan

- [x] Unit tests for EOL lookup, same-major next-version, exact-match blanks, and HTTP fetch fallback
- [x] CLI column test (`storage_type` gone; new columns present)
- [x] `rds_value` preserves `auto_minor_version_upgrade: false`
- [x] `make unittest` on Python 3.14 (3983 passed)

## Summary by CodeRabbit

* **New Features**
  * RDS reports now include engine version, end-of-life date, and recommended next version information.
  * Recommendations remain within the same major version and exclude releases that are already end-of-life.
  * EOL details appear only for exact date matches.
  * RDS end-of-life data can be loaded from a configurable source.

* **Bug Fixes**
  * Explicit configuration values, including `False`, are now preserved in reports.

* **Changes**
  * Removed the previous storage type field from RDS output.

### DIFF
--- a/tools/cli_commands/rds_eol.py
+++ b/tools/cli_commands/rds_eol.py
@@ -3,10 +3,11 @@ from __future__ import annotations
 import logging
 from collections import defaultdict
 from operator import itemgetter
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
 import requests
 import yaml
+from pydantic import BaseModel
 
 from reconcile.utils.datetime_util import utc_now
 from reconcile.utils.semver_helper import parse_semver
@@ -14,25 +15,47 @@ from reconcile.utils.semver_helper import parse_semver
 if TYPE_CHECKING:
     from semver import VersionInfo
 
-RDS_EOL_URL = (
+DEFAULT_RDS_EOL_URL = (
     "https://raw.githubusercontent.com/app-sre/aws-generated-data"
     "/main/output/rds_eol.yaml"
 )
 
 
-def load_rds_eol_data() -> list[dict[str, Any]]:
+class RdsEolEntry(BaseModel):
+    """A single entry from the AWS RDS end-of-life calendar dataset."""
+
+    engine: str
+    version: str
+    eol: str
+
+
+def load_rds_eol_data(url: str = DEFAULT_RDS_EOL_URL) -> list[RdsEolEntry]:
     """Fetch the AWS RDS calendar dataset. Returns an empty list on failure."""
     try:
-        resp = requests.get(RDS_EOL_URL, timeout=10)
+        resp = requests.get(url, timeout=10)
         resp.raise_for_status()
         data = yaml.safe_load(resp.text) or []
     except (requests.RequestException, yaml.YAMLError) as exc:
-        logging.warning("Could not fetch RDS EOL data from %s: %s", RDS_EOL_URL, exc)
+        logging.warning("Could not fetch RDS EOL data from %s: %s", url, exc)
         return []
     if not isinstance(data, list):
-        logging.warning("RDS EOL data from %s is not a list", RDS_EOL_URL)
+        logging.warning("RDS EOL data from %s is not a list", url)
         return []
-    return [entry for entry in data if isinstance(entry, dict)]
+    entries: list[RdsEolEntry] = []
+    for raw in data:
+        if not isinstance(raw, dict):
+            continue
+        try:
+            entries.append(
+                RdsEolEntry(
+                    engine=str(raw["engine"]),
+                    version=str(raw["version"]),
+                    eol=str(raw["eol"]),
+                )
+            )
+        except KeyError, TypeError:
+            continue
+    return entries
 
 
 def _parse_rds_version(version: str) -> VersionInfo | None:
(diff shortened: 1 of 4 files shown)


jfchevrette:
Some errors to fix, otherwise LGTM. Happy to approve when the tests pass

RH-tj:
> Some errors to fix, otherwise LGTM. Happy to approve when the tests pass

fixed thanks!

jfchevrette (approved):

tools/qontract_cli.py
```diff
@@ -1873,31 +1893,45 @@ def rds(ctx: click.Context) -> None:
...
                 "account": spec.provisioner_name,
                 "account_uid": accounts[spec.provisioner_name]["uid"],
                 "region": rds_region(spec, overrides, defaults, accounts),
-                "engine": rds_attr("engine", overrides, defaults),
-                "engine_version": rds_attr("engine_version", overrides, defaults),
+                "engine": engine,
+                "engine_version": engine_version,
                 "instance_class": rds_attr("instance_class", overrides, defaults),
-                "storage_type": rds_attr("storage_type", overrides, defaults),
+                "auto_minor_version_upgrade": rds_value(
```
hemslo: `_format_cell` in `reconcile/utils/output.py:55` uses `raw_data.get(token) or {}` which converts any falsy value (`False`, `0`) to `{}`, then returns `""`. This means `auto_minor_version_upgrade=False` will render as **blank** in table output, while `True` renders correctly — asymmetric and misleading.

Quick fix: convert to string before inserting into the results dict:
```python
"auto_minor_version_upgrade": str(rds_value(
    "auto_minor_version_upgrade", overrides, defaults
)) if rds_value("auto_minor_version_upgrade", overrides, defaults) is not None else "",
```
Or more cleanly, assign the value first and stringify it:
```python
amvu = rds_value("auto_minor_version_upgrade", overrides, defaults)
...
"auto_minor_version_upgrade": "" if amvu is None else str(amvu),
```

tools/cli_commands/rds_eol.py
```diff
@@ -0,0 +1,124 @@
...
+            continue
+        for candidate, candidate_str in by_engine_major[engine_str, current.major]:
+            if candidate <= current:
+                continue
+            candidate_eol = eol_lookup.get((engine_str, candidate_str), "")
+            if candidate_eol >= today:
+                result[engine_str, version_str] = candidate_str
+                break
+    return result
+
```
hemslo, line 114: nit: `engine` and `engine_version` are typed `object | None` but callers always pass `str | None` (from `rds_attr`). Using `str | None` would be more precise and let mypy catch misuse at call sites.

tools/qontract_cli.py
```diff
@@ -1831,6 +1837,15 @@ def rds_attr(
     return overrides.get(attr) or defaults.get(attr)
 
 
+def rds_value(
```
hemslo: nit: `rds_value` duplicates `rds_attr` with different falsy-value handling. Consider fixing `rds_attr` to use `if attr in overrides: return overrides[attr]` instead — that would handle falsy values correctly for all callers and eliminate the need for two near-identical functions.

hemslo (commented):
Review comments from code review.

tools/cli_commands/rds_eol.py
chassing, line 1: Could you please use a proper Pydantic model class instead of a plain dictionary?

tools/cli_commands/rds_eol.py
```diff
@@ -0,0 +1,124 @@
...
+import requests
+import yaml
+
+from reconcile.utils.datetime_util import utc_now
+from reconcile.utils.semver_helper import parse_semver
+
+if TYPE_CHECKING:
+    from semver import VersionInfo
+
+RDS_EOL_URL = (
```
chassing: Don't hardcode the URL. Can be a default value in `qontract-cli`

tools/qontract_cli.py
RH-tj: fixed

tools/qontract_cli.py
RH-tj: fixed

tools/cli_commands/rds_eol.py
RH-tj: fixed

RH-tj:
> Review comments from code review.

@hemslo Indeed! The biggest takeaway for me is to stick to Opus, this PR was made with assistance of Grok and despite same harness, same rules, same guidance, same workflow it produced vastly inferior code output.

chassing (approved):

PR #5773, opened 2026-08-18
fix(aws-version-sync): do not downgrade ElastiCache engine on stale AWS metrics

chassing:
## Summary
- Fixes a bug in `aws-version-sync` where a stale AWS ElastiCache metric (still reporting `redis`) could cause the integration to open an MR reverting app-interface's already-applied `redis` -> `valkey` engine upgrade.
- The existing anti-downgrade check only compared version numbers when the AWS and app-interface engines already matched, so a mismatched engine always went through as a "change", regardless of direction.
- Adds an `ENGINE_UPGRADE_PATHS` map (`redis` -> `valkey`) and skips reconciliation when AWS still reports the pre-upgrade engine while app-interface already reflects the upgrade — treating it as stale AWS data rather than a desired change.
- The reverse case (AWS enforces the upgrade: metrics show `valkey`, app-interface still shows `redis`) is unaffected and continues to open an MR as before.
- Replaces/Supersedes #5530

## Test plan
- [x] Added a TDD regression test (`ec_stale_aws_metrics_no_engine_downgrade_*`) to `test_avs_reconcile` reproducing the stale-metrics scenario; confirmed it failed before the fix (5 MRs instead of 4) and passes after.
- [x] `uv run pytest reconcile/test/aws_version_sync/` — 40 passed, 5 xfailed.
- [x] `uv run ruff check` / `uv run ruff format --check` — clean.
- [x] `uv run mypy reconcile/aws_version_sync/integration.py` — no issues.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented reconciliation from proposing engine downgrades when AWS reports a predecessor engine after an upgrade.
  - Preserved existing protection against version downgrades.
  - Added coverage for scenarios where Redis resources have been upgraded to Valkey.

### DIFF
--- a/reconcile/aws_version_sync/integration.py
+++ b/reconcile/aws_version_sync/integration.py
@@ -171,7 +171,7 @@ EXTENDED_SUPPORT_VERSION_INDICATOR = "-rds."
 # supports upgrading a Redis instance to Valkey, but never the other way round.
 # used to detect stale AWS metrics that still report the pre-upgrade engine
 # while app-interface already reflects the (already applied) upgrade.
-ENGINE_UPGRADE_PATHS = {"redis": "valkey"}
+ENGINE_UPGRADE_PATHS: dict[str, str] = {"redis": "valkey"}
 
 
 class AVSIntegration(QontractReconcileIntegration[AVSIntegrationParams]):
--- a/reconcile/aws_version_sync/integration.py
+++ b/reconcile/aws_version_sync/integration.py
@@ -167,6 +167,11 @@ AppInterfaceExternalResources = list[ExternalResource]
 UidAndReplicationGroupId = tuple[str, str]
 ReplicationGroupIdToIdentifier = dict[UidAndReplicationGroupId, str]
 EXTENDED_SUPPORT_VERSION_INDICATOR = "-rds."
+# maps an engine to the engine it can be upgraded to, e.g. AWS ElastiCache
+# supports upgrading a Redis instance to Valkey, but never the other way round.
+# used to detect stale AWS metrics that still report the pre-upgrade engine
+# while app-interface already reflects the (already applied) upgrade.
+ENGINE_UPGRADE_PATHS = {"redis": "valkey"}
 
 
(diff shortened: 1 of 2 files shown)


reconcile/aws_version_sync/integration.py
```diff
@@ -167,6 +167,11 @@ def resource_engine_version_string(self) -> str:
 UidAndReplicationGroupId = tuple[str, str]
 ReplicationGroupIdToIdentifier = dict[UidAndReplicationGroupId, str]
 EXTENDED_SUPPORT_VERSION_INDICATOR = "-rds."
+# maps an engine to the engine it can be upgraded to, e.g. AWS ElastiCache
+# supports upgrading a Redis instance to Valkey, but never the other way round.
+# used to detect stale AWS metrics that still report the pre-upgrade engine
+# while app-interface already reflects the (already applied) upgrade.
+ENGINE_UPGRADE_PATHS = {"redis": "valkey"}
```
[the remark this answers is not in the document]
chassing: Fixed in 2a3d3786.

reconcile/aws_version_sync/integration.py
```diff
@@ -449,13 +454,20 @@ def reconcile(
...
             ):
                 # do not downgrade the version
                 continue
+            if (
+                ENGINE_UPGRADE_PATHS.get(aws_resource.resource_engine)
+                == app_interface_resource.resource_engine
+            ):
+                # the AWS metrics still report the pre-upgrade engine (e.g. "redis")
+                # while app-interface already reflects the applied engine upgrade
+                # (e.g. "valkey"); the AWS metrics are stale, do not downgrade the engine
```
fishi0x01, line 469: my understanding is this will eventually get consistent right? Just needs some time on AWS side?

fishi0x01 (approved):

reconcile/aws_version_sync/integration.py
chassing: Exactly, just a timing "issue"

PR #5778, opened 2026-08-19
fix(ocm-clusters): replace deprecated OCM version.channel_group with channel

devppratik:
## Problem

[APPSRE-15073](https://redhat.atlassian.net/browse/APPSRE-15073)

The `ocm-clusters` integration sends the deprecated OCM API field `version.channel_group`. OCM warns and instructs callers to use the top-level `channel` field instead for Y-stream version selection. Observed during `scilwp01ue1` cluster creation (APPSRE-14985):

```
[ocm-clusters] WARNING: You are using OCM API fields that have been deprecated -
version.channel_group: The 'version.channel_group' field will be deprecated in the
future. Please use 'channel' instead for Y-stream version selection.
```

## Investigation

Verified against `uhc-clusters-service`:
- The warning fires **only on writes** that send `version.channel_group`. Read responses still return it.
- The replacement `channel` field must be **Y-stream format** (regex-validated, e.g. `stable-4.16`), and is mutually exclusive with `version.channel_group`.
- app-interface only stores the plain channel group (e.g. `stable`), so the Y-stream value is built from the group + the cluster version's major.minor.

The actual ROSA create path is the **rosa CLI** invocation (Jinja templates), not the JSON builder — the `rosa create cluster` command passed `--channel-group`. The rosa CLI exposes a `--channel` flag (Y-stream) as the replacement.

## Changes

**Write paths** (source of the warning):
- `rosa-classic` + `rosa-hcp` creation templates: `--channel-group <group>` → `--channel <group>-<major.minor>` (real ROSA create path).
- OSD create (JSON POST) and OSD/ROSA update (JSON PATCH): send top-level `channel`; current cluster version threaded through `update_cluster` to build the Y-stream value on updates.
- New `build_ystream_channel()` helper (single source of truth for the format).

**Read paths**: intentionally left on `version.channel_group` — OCM still returns it, it holds the plain group that `OCMClusterSpec.channel` expects, and the deprecation only affects writes.

## Testing

- `reconcile/test/test_ocm_clusters.py` + `reconcile/test/utils/rosa/` — 51 pass.
- POST/PATCH JSON fixtures and all rosa creation-script fixtures updated to the Y-stream `channel` output.
- `mypy` (strict) + `ruff` clean.

## Out of scope

`rosa upgrade account-roles --channel-group` (`reconcile/utils/rosa/session.py`) is a **different** command flag (account-role upgrade), unrelated to the cluster `version.channel_group` field. Newer rosa CLI no longer exposes `--channel-group` there — flagging as a potential latent issue, not addressed here.

## Summary by CodeRabbit

* **New Features**
  * Cluster creation and updates now use version-specific Y-stream channels, such as `stable-4.8` or `candidate-4.8`.
  * Channel values are automatically derived from the configured channel and cluster version.

* **Bug Fixes**
  * Improved channel handling for ROSA classic, ROSA HCP, and OSD cluster operations.
  * Updated dry-run and cluster configuration outputs to reflect the correct channel format.

* **Tests**
  * Added coverage to verify channel formatting across supported cluster types and update scenarios.

### DIFF
--- a/reconcile/templates/rosa-hcp-cluster-creation.sh.j2
+++ b/reconcile/templates/rosa-hcp-cluster-creation.sh.j2
@@ -62,4 +62,4 @@ rosa create cluster --cluster-name={{ cluster_name }} \
     {% if cluster.spec.fips -%}
     --fips \
     {% endif -%}
-    --channel {{ cluster.spec.channel }}-{{ (cluster.spec.initial_version | split("."))[:2] | join(".") }}
+    --channel {{ channel }}
--- a/reconcile/templates/rosa-hcp-cluster-creation.sh.j2
+++ b/reconcile/templates/rosa-hcp-cluster-creation.sh.j2
@@ -62,4 +62,4 @@ rosa create cluster --cluster-name={{ cluster_name }} \
     {% if cluster.spec.fips -%}
     --fips \
     {% endif -%}
-    --channel-group {{ cluster.spec.channel }}
+    --channel {{ cluster.spec.channel }}-{{ (cluster.spec.initial_version | split("."))[:2] | join(".") }}
--- a/reconcile/utils/ocm/products.py
+++ b/reconcile/utils/ocm/products.py
@@ -756,6 +756,10 @@ class OCMProductHypershift(OCMProduct):
     ) -> dict[str, Any]:
         ocm_spec: dict[str, Any] = {}
 
+        channel = update_spec.get(SPEC_ATTR_CHANNEL)
+        if channel is not None:
+            ocm_spec["channel"] = build_ystream_channel(channel, version)
+
         disable_uwm = update_spec.get(SPEC_ATTR_DISABLE_UWM)
         if disable_uwm is not None:
             ocm_spec["disable_user_workload_monitoring"] = disable_uwm
--- a/reconcile/utils/ocm/products.py
+++ b/reconcile/utils/ocm/products.py
@@ -72,6 +72,7 @@ def build_ystream_channel(channel_group: str, version: str) -> str:
     major_minor = ".".join(version.split(".")[:2])
     return f"{channel_group}-{major_minor}"
 
+
 OCM_PRODUCT_OSD = "osd"
 OCM_PRODUCT_ROSA = "rosa"
 OCM_PRODUCT_HYPERSHIFT = "hypershift"
@@ -259,7 +260,8 @@ class OCMProductOsd(OCMProduct):
                 "id": f"openshift-v{cluster.spec.initial_version}",
             },
             "channel": build_ystream_channel(
-                cluster.spec.channel, cluster.spec.initial_version or cluster.spec.version
+                cluster.spec.channel,
+                cluster.spec.initial_version or cluster.spec.version,
             ),
             "multi_az": cluster.spec.multi_az,
             "nodes": self._get_nodes_spec(cluster),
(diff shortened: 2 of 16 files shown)


reconcile/utils/ocm/products.py
```diff
@@ -728,7 +752,7 @@ def get_ocm_spec(
         return ocm_spec
```
hemslo, line 752: Bug: This adds the `version` parameter to the Hypershift `_get_update_cluster_spec` signature, but the body still only handles `disable_uwm`. OSD and ROSA both got channel handling added:

```python
channel = update_spec.get(SPEC_ATTR_CHANNEL)
if channel is not None:
    ocm_spec["channel"] = build_ystream_channel(channel, version)
```

Since `SPEC_ATTR_CHANNEL` is in Hypershift's `ALLOWED_SPEC_UPDATE_FIELDS`, channel changes will pass validation but get silently dropped in the PATCH to OCM. The cluster stays on the old channel and the reconciler retries the no-op update every loop.

reconcile/utils/ocm/products.py
```diff
@@ -242,8 +258,11 @@ def _get_create_cluster_spec(
             "region": {"id": cluster.spec.region},
             "version": {
                 "id": f"openshift-v{cluster.spec.initial_version}",
-                "channel_group": cluster.spec.channel,
             },
+            "channel": build_ystream_channel(
```
hemslo, line 262: Nit: The `or cluster.spec.version` fallback is inconsistent with the Jinja2 templates (`rosa-classic-cluster-creation.sh.j2:65`, `rosa-hcp-cluster-creation.sh.j2:65`) which hardcode `initial_version` with no fallback. At cluster creation time, `initial_version` should always be present — either both paths need the fallback or neither does.

reconcile/templates/rosa-hcp-cluster-creation.sh.j2
```diff
@@ -62,4 +62,4 @@ rosa create cluster --cluster-name={{ cluster_name }} \
     {% if cluster.spec.fips -%}
     --fips \
     {% endif -%}
-    --channel-group {{ cluster.spec.channel }}
+    --channel {{ cluster.spec.channel }}-{{ (cluster.spec.initial_version | split("."))[:2] | join(".") }}
```
hemslo: The Y-stream channel format (`channel-major.minor`) is now duplicated between this inline Jinja2 expression and the Python `build_ystream_channel()` function. If the format changes, these two must be updated independently. Consider passing a pre-computed channel value from `generate_rosa_creation_script` instead.

devppratik:
Thanks @hemslo — all three addressed in ccb4bbd0:

1. **Hypershift channel silently dropped** — `OCMProductHypershift._get_update_cluster_spec` now emits the Y-stream `channel` like OSD/ROSA, so a channel change in `ALLOWED_SPEC_UPDATE_FIELDS` is actually PATCHed instead of looping as a no-op. Added a regression test.
2. **Format duplication (Jinja vs Python)** — the Y-stream channel is now computed once via `build_ystream_channel()` in `generate_rosa_creation_script` and passed to the templates as `{{ channel }}`. Single source of truth.
3. **Fallback consistency** — with the channel precomputed in Python for the rosa templates too, both the JSON create path and the rosa CLI path now use the same `initial_version or version` fallback.

reconcile/templates/rosa-hcp-cluster-creation.sh.j2
devppratik: fixed

reconcile/utils/ocm/products.py
devppratik: fixed

reconcile/utils/ocm/products.py
devppratik: fixed

hemslo (approved):
LGTM!

PR #5781, opened 2026-08-20
feat(aws_account_manager): ensure ELB SLR before ROSA HCP enablement

devppratik:
## What

Adds an idempotent pre-flight step to `enable_rosa_marketplace` that ensures `AWSServiceRoleForElasticLoadBalancing` exists before proceeding with ROSA HCP marketplace enablement.

## Why

Brand-new AWS accounts lack `AWSServiceRoleForElasticLoadBalancing`, causing downstream cluster creation to fail with:

```
CLUSTERS-MGMT-400: 'AWSServiceRoleForElasticLoadBalancing' role doesn't exist
```

The AWS Console "Enable ROSA with HCP" flow creates this role as a side effect. The programmatic path (`aws_account_manager`) does not — this PR closes that gap.

Ref - https://redhat-internal.slack.com/archives/C0A8S4L99C2/p1787148122532539?thread_ts=1787147955.579829&cid=C0A8S4L99C2

## Changes

- `qontract_utils/aws_api_typed/iam.py`: add `has_service_linked_role` and `create_service_linked_role` methods
- `reconcile/aws_account_manager/reconciler.py`: add `_ensure_elb_service_linked_role` (state-tracked, dry-run aware, handles concurrent creation via `InvalidInput` catch); call it from `enable_rosa_marketplace` before `_enable_rosa_marketplace`
- Tests for all new code paths

Refs APPSRE-15113

## Summary by CodeRabbit

* **New Features**
  * Automatically ensures the required Elastic Load Balancing service-linked IAM role exists before enabling the ROSA Marketplace subscription.
  * Supports safe handling of existing roles and dry-run operations.
  * Records completion state to avoid repeating role setup.

* **Bug Fixes**
  * Improved handling of missing roles and expected AWS responses while surfacing unexpected errors.

* **Tests**
  * Added coverage for role creation, existing roles, dry-run behavior, and error handling.


chassing (approved):

reconcile/aws_account_manager/reconciler.py
```diff
@@ -387,6 +391,27 @@ def _set_supported_regions(
...
+    def _ensure_elb_service_linked_role(self, aws_api: AWSApi, name: str) -> None:
+        with self.state.transaction(
+            state_key(name, TASK_ENSURE_ELB_SERVICE_LINKED_ROLE)
+        ) as state:
+            if state.exists:
+                return
+
+            if aws_api.iam.has_service_linked_role(ELB_SERVICE_LINKED_ROLE_NAME):
+                logging.info(f"{name}: ELB service-linked role already exists")
+                if self.dry_run:
```
hemslo: Nit: inconsistent `dry_run` handling in the "already exists" path.

The sibling method `_enable_rosa_marketplace` (line ~397-399) saves state **without** a `dry_run` guard when the subscription already exists — recording that something is already present is an observation, not a mutation, so it's safe regardless of mode:

```python
# _enable_rosa_marketplace pattern:
if aws_api.marketplace.has_rosa_subscription():
    logging.info(f"{name}: ROSA HCP marketplace already subscribed")
    state.value = "already-subscribed"   # no dry_run check
    return
```

Here, the `dry_run` check causes every dry-run reconcile loop to re-call `has_service_linked_role` and re-log instead of short-circuiting on `state.exists`. Removing lines 403-404 would match the established pattern and avoid the repeated API call.

devppratik: fixed in the latest commit

hemslo (approved):
LGTM!

.claude/settings.json
```diff
@@ -0,0 +1,17 @@
+{
+  "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "Bash",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "cmd=$(jq -r '.tool_input.command'); if echo \"$cmd\" | grep -qE '^git (commit|push)'; then make linter-test && make types-test && make unittest; fi",
```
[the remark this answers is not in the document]
devppratik: Wrong Push. File Removed

PR #5789, opened 2026-08-24
APPSRE-14629: Add env label to SLO alerts

lisa:
Why

Sloth-generated SLO alerts carried only service, severity, and slo labels, so the shared Alertmanager config could not route them by environment. Hand-written PrometheusRules add an env label (prod or stage) that Alertmanager keys on; without it, generated alerts fall through to catch-all receivers and stage and prod alerts are indistinguishable.

What

generate_sloth_rules now reads an optional env value from the SLO document and, when set, emits it as the env label on both the page and ticket alert rules. Record rules are unaffected, which is what Alertmanager routing needs. When env is absent, output is unchanged, preserving backward compatibility.

The env value is a dedicated field on the SLO document rather than derived from a namespace reference, because a document may span multiple namespaces across environments and stage and prod are already modelled as separate SLO documents.

This is part 2 of 3, and depends on https://github.com/app-sre/qontract-schemas/pull/1057.

## Summary by CodeRabbit

* **New Features**
  * Sloth-generated alert rules now include the configured environment label when available.
  * Existing alert labels remain unchanged when no environment is specified.

* **Bug Fixes**
  * Improved consistency so environment labels are applied across generated page and ticket alerts.
  * Empty or whitespace-only environment values are now rejected with a clear validation error.


esron (approved):

lisa:
Turns out, this is not needed.

PR #5790, opened 2026-08-24
feat: migrate quay-robot-accounts to qontract-api

esron:
## Summary
- Add a qontract-api integration (`quay-robot-accounts-api`) that reconciles Quay robot accounts, team membership, and repo permissions from App-Interface desired state.
- Introduce a shared Layer 1 `quay_api` client and Layer 2 Quay workspace client/factory; the client compiles GraphQL desired state and never talks to Quay.
- Leave the legacy `quay-robot-accounts` CLI integration in place for a side-by-side rollout.

## Test plan
- [ ] `uv run pytest qontract_utils/tests/quay_api -q`
- [ ] `cd qontract_api && uv run pytest tests/quay tests/integrations/test_quay_robot_accounts_service.py tests/integrations/test_quay_robot_accounts_router.py tests/tasks/test_quay_robot_accounts.py -q`
- [ ] `uv run pytest reconcile/test/test_quay_robot_accounts_api.py -q`
- [ ] Confirm `qontract-reconcile quay-robot-accounts` still works (legacy untouched)
- [ ] Dry-run `qontract-reconcile quay-robot-accounts-api --dry-run` against a test environment

## Summary by CodeRabbit

- **New Features**
  - Added Quay robot-account reconciliation with team memberships and repository permissions.
  - Added asynchronous reconciliation APIs, task-status tracking, dry-run support, and CLI access.
  - Added repository management operations, including creation, deletion, description updates, and visibility changes.
- **Bug Fixes**
  - Improved Vault secret version handling and enforced HTTPS for Quay connections.
  - Added validation for duplicate robots and repositories and improved error reporting.
- **Documentation**
  - Added comprehensive Quay Robot Accounts documentation and updated the integrations index.

PR #5791, opened 2026-08-25
Base qontract-api test stage on builder image

lisa:
Why

The qontract-api image build fails in its `test` stage with `failed to remove directory /opt/app-root: Permission denied`. The stage was based on `base` (`ubi10-minimal` plus a system-installed python3.14) while the virtualenv was built on `python-314-minimal`. uv sees a different interpreter, tries to recreate `UV_PROJECT_ENVIRONMENT` (`/opt/app-root`), and cannot remove it as UID 1001.

What

Base the `test` stage on `builder` so uv reuses the existing venv instead of recreating it, and re-add the make/diffutils install and the RUFF_CACHE_DIR/TESTFLAG env that `base` previously provided.

The change mirrors the qontract-api fix in
[#5790](https://github.com/app-sre/qontract-reconcile/pull/5790) so the two branches do not conflict on
[qontract_api/Dockerfile](qontract_api/Dockerfile), whichever merges first. The same qontract-api build failure is also observed on [#5789](https://github.com/app-sre/qontract-reconcile/pull/5789); landing this change independently unblocks the qontract-api check on both pull requests.

Scope

This addresses only the qontract-api image build. The separate qontract-reconcile-master-on-pull-request failure seen on [#5789](https://github.com/app-sre/qontract-reconcile/pull/5789) is not addressed here; its root cause is unconfirmed.

Attribution

The Dockerfile change originates from
[#5790](https://github.com/app-sre/qontract-reconcile/pull/5790). The branch and this commit message were prepared by the model named below at the user's direction.

## Summary by CodeRabbit

* **Chores**
  * Updated the test environment to use the complete build environment.
  * Improved test-stage setup and tooling availability.
  * Simplified environment configuration for more consistent test execution.

### DIFF
--- a/qontract_api/Dockerfile
+++ b/qontract_api/Dockerfile
@@ -61,15 +61,20 @@ RUN cd qontract_api && uv sync --frozen --no-group dev
 #
 # Test image
 #
-FROM base AS test
-COPY --from=ghcr.io/astral-sh/uv:0.12.5@sha256:e85be844203885286c60ffad8a858d48afb6c5a5c237ca0e67f12e74b8f174b1 /uv /bin/uv
+# Must use the builder image (python-314-minimal) rather than `base`
+# (ubi10-minimal + /usr/bin/python3.14). UV_PROJECT_ENVIRONMENT is /opt/app-root;
+# a different interpreter makes uv recreate that directory, which fails as
+# UID 1001 with: failed to remove directory `/opt/app-root`: Permission denied.
+FROM builder AS test
+
+ENV \
+    RUFF_CACHE_DIR=/tmp/.ruff_cache \
+    TESTFLAG=/tmp/.tests_passed
 
 USER root
-# install dependencies
-RUN microdnf install -y diffutils && microdnf clean all
+RUN microdnf install -y make diffutils && microdnf clean all
 USER 1001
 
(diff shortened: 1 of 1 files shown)


esron (approved):

qontract_api/Dockerfile
```diff
@@ -61,15 +61,20 @@ RUN cd qontract_api && uv sync --frozen --no-group dev
...
-COPY --from=ghcr.io/astral-sh/uv:0.12.5@sha256:e85be844203885286c60ffad8a858d48afb6c5a5c237ca0e67f12e74b8f174b1 /uv /bin/uv
+# Must use the builder image (python-314-minimal) rather than `base`
+# (ubi10-minimal + /usr/bin/python3.14). UV_PROJECT_ENVIRONMENT is /opt/app-root;
+# a different interpreter makes uv recreate that directory, which fails as
+# UID 1001 with: failed to remove directory `/opt/app-root`: Permission denied.
+FROM builder AS test
+
+ENV \
+    RUFF_CACHE_DIR=/tmp/.ruff_cache \
+    TESTFLAG=/tmp/.tests_passed
```
hemslo, line 72: why we didn't use `ubi10/python-314-minimal` in base? @chassing

PR #5724, opened 2026-08-04
feat(aws-account-manager): automate ROSA HCP marketplace enablement

devppratik:
## Summary

Automates the manual SOP step of enabling ROSA HCP on AWS accounts during onboarding ([APPSRE-14982](https://redhat.atlassian.net/browse/APPSRE-14982)).

Currently, enabling ROSA HCP requires logging into the AWS console, searching for ROSA, and clicking "Enable ROSA with HCP" (per `openshift-rosa-clusters-prerequisites.md`). This PR replaces that manual step with programmatic marketplace subscription using the `marketplace-agreement` and `marketplace-discovery` boto3 APIs.

### Changes

- **New `qontract_utils/aws_api_typed/marketplace.py`** — Layer 1 client wrapping both marketplace services (pinned to `us-east-1`):
  - `has_rosa_subscription()` — idempotency check via `search_agreements` (filters: `PartyType=Acceptor`, `AgreementType=PurchaseAgreement`, `ResourceIdentifier`)
  - `discover_rosa_offer()` — discovers offer ID, proposal ID, and term IDs via `list_purchase_options` → `get_offer` → `get_offer_terms`
  - `subscribe_rosa()` — creates and accepts agreement request
- **Updated `aws_api_typed/api.py`** — Added `marketplace` cached property creating both clients
- **Updated `reconciler.py`** — Added `TASK_ENABLE_ROSA_MARKETPLACE` with state-tracked `_enable_rosa_marketplace()` method
- **Updated `integration.py`** — Wired into `create_accounts()` after IAM user creation, gated by `additionalFeatures.rosa: true`

### Design

- Uses hardcoded ROSA HCP product ID (`bfdca560-2c78-4e64-8193-794c159e6d30`) — stable AWS Marketplace identifier
- Dynamic offer/term discovery via `list_purchase_options` → `get_offer` → `get_offer_terms`
- Idempotent: checks existing agreements before subscribing, state-tracked to skip on re-runs
- No new dependencies — uses existing boto3 (no type stubs available for marketplace services, clients typed as `Any`)

### IAM permissions required on target account role

- `aws-marketplace-agreement:SearchAgreements`
- `aws-marketplace-agreement:CreateAgreementRequest`
- `aws-marketplace-agreement:AcceptAgreementRequest`
- `aws-marketplace-discovery:ListPurchaseOptions`
- `aws-marketplace-discovery:GetOffer`
- `aws-marketplace-discovery:GetOfferTerms`

## Test plan

[...]

PR #5794, opened 2026-08-26
fix(aws-account-manager): select active ROSA marketplace offer and rate card

devppratik:
## Summary

Fixes the runtime failure `ValidationException: Provided agreement proposal is inactive` seen when `aws_account_manager` enables ROSA HCP marketplace subscription (introduced in #5724).

### Root cause

`discover_rosa_offer` picked `purchaseOptions[0]` blindly. `list_purchase_options` returns **all** purchase options for the product — including expired / rotated-out offers, in no guaranteed order. The first one could be a stale offer whose `agreementProposalId` is already inactive, so `create_agreement_request` rejected it.

### Changes

- **Active offer selection** — new `_select_active_purchase_option` picks the currently-active purchase option by its availability window (`availableFromTime` .. `expirationTime`, boto3 timezone-aware datetimes), preferring the most recently available. Raises a clear `RuntimeError` when none are active instead of forwarding a stale proposal to AWS.
- **Deterministic rate-card selection** — a `configurableUpfrontPricingTerm` can offer several rate cards (one per contract-duration selector, e.g. `P12M`/`P24M`/`P36M`) in arbitrary order. `_select_rate_card` now picks the shortest valid duration (via `_duration_months`) rather than the positional first. Zero-commit (`dimensionValue: 0`) keeps this cost-neutral regardless of duration.

### Why enum-based filtering was not used

`list_purchase_options` supports `AVAILABILITY_STATUS` / `PURCHASE_OPTION_TYPE` filter types, but the exact `filterValues` enum strings could not be verified against the live service model (docs are JS-rendered, botocore carries no enum). Sending a wrong value would produce another `ValidationException`. Client-side filtering on `expirationTime`/`availableFromTime` — fields confirmed present in the botocore output model — is safe and needs no unverified strings.

## Test plan

- [x] 5 new unit tests: skip-expired option, pick most-recently-available, all-expired raises, shortest-duration rate card, skip rate card without selector
- [x] Existing marketplace tests unchanged and passing (18 total)
- [x] `ruff check` clean
- [x] `mypy` clean

## Follow-up (not in this PR)

Residual race: if AWS invalidates the proposal between `discover_rosa_offer` and `subscribe_rosa` (back-to-back calls, ms window), the same error could reappear. Not the observed cause; would need a re-discover+retry in the reconciler. Can add if desired.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ROSA HCP marketplace offer selection by excluding unavailable or expired options.
  * The newest active offer is now selected automatically.
  * Upfront pricing uses the shortest valid contract duration.
  * Clear errors are raised when no active offers or valid rate cards are available.

### DIFF
--- a/qontract_utils/qontract_utils/aws_api_typed/marketplace.py
+++ b/qontract_utils/qontract_utils/aws_api_typed/marketplace.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import logging
+import re
+from datetime import UTC, datetime
 from typing import Any
 
 from pydantic import BaseModel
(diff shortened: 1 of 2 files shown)


fltataru (approved):
LGTM!

PR #5803, opened 2026-09-04
fix(openshift-upgrade-watcher): log cluster name on per-cluster error

fishi0x01:
## Summary
- `notify_upgrades_start` now logs the cluster name (via `logging.exception`) before re-raising when fetching a cluster's upgrade start info fails, instead of crashing the whole run with no indication of which cluster was responsible.
- Prompted by a live incident where `openshift-upgrade-watcher` (shard 0-1) crashed on every run for over an hour with `kubernetes.dynamic.exceptions.ForbiddenError: 403` after a newly onboarded ROSA cluster got its automation token before its dedicated-admin RBAC for `upgradeconfigs` had finished provisioning — identifying the offending cluster required correlating app-interface commit history with CloudWatch timestamps, since the cluster name wasn't logged anywhere.
- Behavior is unchanged otherwise: the exception still propagates and the integration still exits non-zero.

## Test plan
- [x] `uv run pytest reconcile/test/test_openshift_upgrade_watcher.py` — 11 passed, including new `test_new_upgrade_error_logs_cluster_name`
- [x] `uv run ruff check` / `ruff format --check`
- [x] `uv run mypy`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when checking upgrade status for clusters.
  * Failures now include the affected cluster name in logs while still being reported to the caller.

* **Tests**
  * Added coverage to verify upgrade lookup errors are logged with the relevant cluster name and propagated correctly.

### DIFF
--- a/reconcile/openshift_upgrade_watcher.py
+++ b/reconcile/openshift_upgrade_watcher.py
@@ -109,14 +109,18 @@ def notify_upgrades_start(
 ) -> None:
     now = utc_now()
     for cluster in clusters:
-        if cluster.spec and not cluster.spec.hypershift:
-            upgrade_at, version = _get_start_osd(oc_map, cluster.name)
-        elif cluster.spec and cluster.spec.q_id:
-            upgrade_at, version = _get_start_hypershift(
-                ocm_map.get(cluster.name)._ocm_client, cluster.spec.q_id
-            )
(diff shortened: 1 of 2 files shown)


hemslo (approved):
LGTM!

PR #5805, opened 2026-09-07
fix(github-owners): stop rate-limit alert storm, add retry (APPSRE-15206)

chassing:
## Summary

`github-owners.reconcile` (qontract-api Celery task) shares a GitHub API credential's rate limit with two legacy integrations (`saas-metrics-exporter`, `openshift-saas-deploy-trigger-moving-commits`). When the shared quota is exhausted, the task retried on its normal ~60s schedule and reported `FAILURE` every cycle, firing the `qontract-api-task-failure` alert repeatedly and paging the IC for a condition that is not actionable and self-recovers once GitHub resets the quota at the top of the hour.

Scoped to an in-repo code fix only (no changes to the alert definition, token provisioning, or the legacy integrations — see [APPSRE-15206](https://redhat.atlassian.net/browse/APPSRE-15206) for the full analysis and out-of-scope items).

### Hooks enhancement (`qontract_utils/qontract_utils/hooks.py`)

Before this PR, `error_hooks` were only ever called with the pre/post-hook context (`hook(*hook_args)`) — a hook that wanted to know *what* exception triggered it had to reach for `sys.exc_info()` from inside the hook body, which is implicit and easy to get wrong (e.g. it silently returns `None` if called outside an active exception context). `retry_hooks` already had precedent for a trailing argument (`retry_hook(*hook_args, attempt.num)`), so `error_hooks` now follow the same convention: `hook(*hook_args, exc)` — i.e. `hook(exc)` with no context, or `hook(context, exc)` with one. This is what makes the GitHub rate-limit detection below possible as a single shared hook instead of duplicated `try/except` blocks in every API method.

### GitHub rate-limit handling

- **`qontract_utils/qontract_utils/github_org/api.py`**: detects GitHub's primary (403 + `X-RateLimit-Remaining: 0`) and secondary (429, or 403 + `Retry-After` on a separate quota bucket) rate limits via a shared `error_hook`, raising a typed `GithubRateLimitExceededError` carrying the reset time. Adds a predicate-based `retry_config` that retries transient network/server errors (connection errors, timeouts, 5xx) but never retries a rate limit — retrying before the reset window is guaranteed to fail and only burns more of the shared quota.
- **`qontract_api/qontract_api/github/github_org_workspace_client.py`**: caches a rate-limit marker per org (on both the read path and the write path) so subsequent reconcile cycles short-circuit instead of repeating a doomed GitHub call.
- **`qontract_api/qontract_api/integrations/github_owners/service.py`**: treats a rate-limited org as a skip (logged, not recorded as an error) and stops trying further actions for that org once its shared credential is exhausted, so the task reports `SUCCESS` and the alert stops firing. Documented as a sanctioned carve-out to AGENTS.md's "silently-skipped units must FAIL" convention, since the reset time is known upfront and no operator action is possible before it.

### Review fixes (coderabbitai, fishi0x01, hemslo)

- Non-numeric rate-limit header values (e.g. an HTTP-date `Retry-After`, which HTTP allows) no longer raise `ValueError` out of the error hook — they fall back to the conservative default instead of masquerading as a generic task failure.
- An oversized-but-numeric `Retry-After` (e.g. `10**20`) no longer raises `OverflowError` from `timedelta(seconds=...)` — same fallback applies.
- A rate limit hit while applying an action now stops the remaining actions for that org (shared credential, guaranteed to fail the same way) instead of burning more quota.
- The rate-limit marker is now also cached when `add_member_as_admin` (the write path) hits a rate limit, not just `get_current_members` (the read path) — previously the stale members cache would regenerate the same doomed action every cycle.
- Secondary-limit 403 responses without `X-RateLimit-Remaining: 0` (a separate quota bucket per GitHub's docs) are now correctly classified as rate limits instead of falling through as generic errors.
- `get_current_members`'s locked path now checks the rate-limit marker before the members cache (matching the outer, pre-lock check order), closing a race where a marker written while a caller waited on the lock could be shadowed by a concurrently-populated members cache and return stale data.

## Test plan

- [x] `cd qontract_utils && make test` (789 passed, ruff + mypy clean)
- [x] `cd qontract_api && make test` (809 passed, ruff + mypy clean)
- [x] Tests confirm: rate limits fail fast with zero retries (including the secondary-limit, malformed-header, and oversized-header shapes); connection errors, 503s, and PyGithub 5xx errors are retried and recover; persistent versions of each exhaust retries and propagate correctly; the apply loop stops after the first rate limit in an org; the marker is cached on both read and write rate-limit hits
- [x] Regression: existing `github-owners` service/task tests (generic error handling, dry-run/non-dry-run, multi-org reconciliation) still pass unchanged

## Summary by CodeRabbit

* **New Features**
  * Added GitHub rate-limit detection with reset-time reporting and temporary caching.
  * Rate-limited organizations and actions are skipped until access is expected to recover.
  * Processing continues for unaffected organizations.
  * Error hooks now receive the original exception for improved diagnostics.

* **Bug Fixes**
  * Prevented rate-limit errors from being reported as reconciliation failures.
  * Improved handling of invalid or oversized retry timing headers.

* **Tests**
  * Expanded coverage for rate limits, caching, and error-hook behavior.

### DIFF
--- a/qontract_api/qontract_api/github/github_org_workspace_client.py
+++ b/qontract_api/qontract_api/github/github_org_workspace_client.py
@@ -123,11 +123,11 @@ class GithubOrgWorkspaceClient:
             return cached.members
 
         with self._cache.lock(cache_key):
-            if cached := self._cache.get_obj(cache_key, CachedOrgMembers):
-                return cached.members
-
             if marker := self._cache.get_obj(rate_limit_key, RateLimitMarker):
                 raise GithubRateLimitExceededError(marker.reset_at)
+
+            if cached := self._cache.get_obj(cache_key, CachedOrgMembers):
+                return cached.members
 
             try:
                 admin_members = self._api.get_admin_members(org_name)
--- a/qontract_api/qontract_api/github/github_org_workspace_client.py
+++ b/qontract_api/qontract_api/github/github_org_workspace_client.py
@@ -84,6 +84,23 @@ class GithubOrgWorkspaceClient:
         except RuntimeError as e:
             logger.warning(f"Could not acquire lock to clear cache for {org_name}: {e}")
 
+    def _cache_rate_limit_marker(
+        self, org_name: str, exc: GithubRateLimitExceededError
+    ) -> None:
+        """Cache a rate-limit marker so subsequent calls short-circuit until reset."""
+        ttl = max(
+            1,
+            min(
+                int((exc.reset_at - datetime.now(UTC)).total_seconds()),
+                self._settings.github_org.members_cache_ttl,
+            ),
+        )
+        self._cache.set_obj(
+            self._rate_limit_key(org_name),
+            RateLimitMarker(reset_at=exc.reset_at),
+            ttl,
+        )
+
     def get_current_members(self, org_name: str) -> list[str]:
         """Get the combined set of admin members + pending invitations (cached).
 
@@ -113,18 +130,7 @@ class GithubOrgWorkspaceClient:
                 admin_members = self._api.get_admin_members(org_name)
                 pending_invitations = self._api.get_pending_invitations(org_name)
             except GithubRateLimitExceededError as e:
-                ttl = max(
-                    1,
-                    min(
-                        int((e.reset_at - datetime.now(UTC)).total_seconds()),
-                        self._settings.github_org.members_cache_ttl,
-                    ),
-                )
-                self._cache.set_obj(
-                    rate_limit_key,
-                    RateLimitMarker(reset_at=e.reset_at),
-                    ttl,
-                )
+                self._cache_rate_limit_marker(org_name, e)
                 raise
 
             combined = sorted(set(admin_members) | set(pending_invitations))
@@ -143,5 +149,9 @@ class GithubOrgWorkspaceClient:
             org_name: GitHub organization name
             username: GitHub username to add as admin
         """
-        self._api.add_member_as_admin(org_name, username)
+        try:
+            self._api.add_member_as_admin(org_name, username)
+        except GithubRateLimitExceededError as e:
+            self._cache_rate_limit_marker(org_name, e)
+            raise
         self._clear_cache(org_name)
--- a/qontract_api/qontract_api/github/github_org_workspace_client.py
+++ b/qontract_api/qontract_api/github/github_org_workspace_client.py
@@ -114,17 +114,20 @@ class GithubOrgWorkspaceClient:
             Sorted list of lowercase GitHub usernames (admins + pending invitees)
         """
         cache_key = self._cache_key(org_name)
+        rate_limit_key = self._rate_limit_key(org_name)
+
+        if marker := self._cache.get_obj(rate_limit_key, RateLimitMarker):
+            raise GithubRateLimitExceededError(marker.reset_at)
 
         if cached := self._cache.get_obj(cache_key, CachedOrgMembers):
             return cached.members
 
-        rate_limit_key = self._rate_limit_key(org_name)
-        if marker := self._cache.get_obj(rate_limit_key, RateLimitMarker):
-            raise GithubRateLimitExceededError(marker.reset_at)
-
         with self._cache.lock(cache_key):
             if cached := self._cache.get_obj(cache_key, CachedOrgMembers):
                 return cached.members
+
+            if marker := self._cache.get_obj(rate_limit_key, RateLimitMarker):
+                raise GithubRateLimitExceededError(marker.reset_at)
 
             try:
                 admin_members = self._api.get_admin_members(org_name)
--- a/qontract_api/qontract_api/github/github_org_workspace_client.py
+++ b/qontract_api/qontract_api/github/github_org_workspace_client.py
@@ -8,9 +8,11 @@ This layer sits between the stateless GithubOrgApi and business logic, providing
 
 from __future__ import annotations
 
+from datetime import UTC, datetime
 from typing import TYPE_CHECKING
 
 from pydantic import BaseModel, Field
+from qontract_utils.github_org import GithubRateLimitExceededError
 
 from qontract_api.logger import get_logger
 
@@ -27,6 +29,16 @@ class CachedOrgMembers(BaseModel, frozen=True):
     """Cached combined list of admin members + pending invitations."""
 
     members: list[str] = Field(default_factory=list)
+
+
+class RateLimitMarker(BaseModel, frozen=True):
+    """Marks that GitHub's API rate limit was exhausted for an org's credential.
+
+    Cached with a TTL derived from the rate limit reset time so subsequent
+    reconcile cycles short-circuit instead of repeating a doomed GitHub call.
+    """
+
+    reset_at: datetime
 
 
 class GithubOrgWorkspaceClient:
@@ -59,6 +71,10 @@ class GithubOrgWorkspaceClient:
     def _cache_key(org_name: str) -> str:
         return f"github-org:{org_name}:members"
 
+    @staticmethod
+    def _rate_limit_key(org_name: str) -> str:
+        return f"github-org:{org_name}:rate-limited"
+
     def _clear_cache(self, org_name: str) -> None:
         """Clear cached members for the given org."""
         cache_key = self._cache_key(org_name)
@@ -85,12 +101,31 @@ class GithubOrgWorkspaceClient:
         if cached := self._cache.get_obj(cache_key, CachedOrgMembers):
             return cached.members
 
+        rate_limit_key = self._rate_limit_key(org_name)
+        if marker := self._cache.get_obj(rate_limit_key, RateLimitMarker):
+            raise GithubRateLimitExceededError(marker.reset_at)
+
         with self._cache.lock(cache_key):
             if cached := self._cache.get_obj(cache_key, CachedOrgMembers):
                 return cached.members
 
-            admin_members = self._api.get_admin_members(org_name)
-            pending_invitations = self._api.get_pending_invitations(org_name)
+            try:
+                admin_members = self._api.get_admin_members(org_name)
+                pending_invitations = self._api.get_pending_invitations(org_name)
+            except GithubRateLimitExceededError as e:
+                ttl = max(
+                    1,
+                    min(
+                        int((e.reset_at - datetime.now(UTC)).total_seconds()),
+                        self._settings.github_org.members_cache_ttl,
+                    ),
+                )
+                self._cache.set_obj(
+                    rate_limit_key,
+                    RateLimitMarker(reset_at=e.reset_at),
+                    ttl,
+                )
+                raise
 
             combined = sorted(set(admin_members) | set(pending_invitations))
             cached_obj = CachedOrgMembers(members=combined)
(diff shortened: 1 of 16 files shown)


qontract_api/qontract_api/integrations/github_owners/service.py
```diff
@@ -172,6 +179,12 @@ def reconcile(
                     try:
                         self._execute_action(github_client, action)
                         applied_actions.append(action)
+                    except GithubRateLimitExceededError as e:
```
fishi0x01, line 182: This `except` only logs and swallows the rate-limit error — the enclosing `for action in org_actions` loop keeps iterating. If an org has multiple pending actions and the shared credential's quota is exhausted on the first one, every remaining action for that org still calls `_execute_action` and independently fails the same way, burning several more requests against the same quota. Given the PR description's own rationale ("guaranteed to fail and only burns more quota"), this loop should probably `break` (or skip the rest of `org_actions`) once a rate limit is hit, the same way the outer loop's `continue` at line 168 skips to the next org.

qontract_api/qontract_api/integrations/github_owners/service.py
```diff
@@ -159,6 +160,12 @@ def reconcile(
                 github_client = self._create_github_org_client(org)
                 org_actions = self._calculate_actions(org, github_client)
                 all_actions.extend(org_actions)
+            except GithubRateLimitExceededError as e:
```
fishi0x01, line 163: This `continue` skips the org without appending anything to `errors`, so `GithubOwnersTaskResult.status` will report `SUCCESS` even though this org's owners weren't reconciled this cycle. Worth confirming this is intentional — AGENTS.md's Service Layer Patterns note that silently-skipped units should surface as FAILURE rather than SUCCESS. If a warn-and-skip is the deliberate design here (transient, self-healing next cycle), it'd be worth reflecting that carve-out back into the docs so it doesn't read as a violation of the convention.

fishi0x01 (commented):
Automated review found 4 issues, two of which undercut this PR's own stated goal of protecting the shared GitHub rate-limit quota. Inline comments are on the two lines that are part of this diff; the other two target pre-existing code that this PR's new behavior now depends on (GitHub's review API won't let me anchor a comment to a line outside the diff hunks), so I'm calling them out here:

- **`qontract_api/qontract_api/github/github_org_workspace_client.py:139`** (`add_member_as_admin`) — this write path never catches `GithubRateLimitExceededError` or sets the `RateLimitMarker`, so it's invisible to the short-circuit cache `get_current_members` relies on. If the write call is what first exhausts the quota, the very next reconcile cycle's `get_current_members` finds no marker and makes a fresh live call that's very likely to fail again — defeating the stated purpose of short-circuiting instead of repeating a doomed call.
- **`qontract_utils/qontract_utils/github_org/api.py:304`** (`get_pending_invitations` → `self._paginated_get(...)`) — enabling `_RETRY_CONFIG` on `_DEFAULT_HOOKS` means a transient error on a later page of this multi-page fetch now restarts the *entire* fetch from page 1 (`_paginated_get` accumulates `items` in-memory across `while url:` iterations, and the retry wraps the whole call). A 503 on page 3 re-fetches pages 1-2 again, burning extra requests against the same quota this PR is trying to protect. This is newly exposed here since previously `NO_RETRY_CONFIG` meant zero retries.

See inline comments for the other two findings.

qontract_api/qontract_api/integrations/github_owners/service.py
hemslo, line 182: [P1] Persist the rate-limit marker when owner application is throttled. This handler only logs the exception, while GithubOrgWorkspaceClient.add_member_as_admin does not write the new rate-limited cache entry. Since get_current_members has already populated the members cache before this loop, the next scheduled run reuses that stale list and generates the same action again, so the rate-limit mitigation is bypassed for failures during the PUT. Record the marker at the workspace mutation boundary and stop the remaining actions for this org.

qontract_utils/qontract_utils/github_org/api.py
```diff
@@ -12,21 +12,120 @@
...
+    converts the exception) and the retry predicate (which must never retry
+    it), so the two can never disagree on what counts as a rate limit.
+    """
+    if isinstance(exc, RateLimitExceededException):
+        return _reset_at_from_headers(exc.headers or {})
+    if isinstance(exc, GithubException) and exc.status == HTTPStatus.TOO_MANY_REQUESTS:
+        return _reset_at_from_headers(exc.headers or {})
+    if isinstance(exc, requests.HTTPError) and exc.response is not None:
+        response = exc.response
+        if response.status_code == HTTPStatus.TOO_MANY_REQUESTS or (
```
hemslo: [P1] Classify secondary-limit 403 responses from the direct invitations request. A 403 with Retry-After but X-RateLimit-Remaining not equal to 0, or with that header absent, falls through here; the error hook leaves requests.HTTPError unchanged, so the workspace and service record a failure and no marker is cached. GitHub documents that secondary limits may return 403 or 429 and may use Retry-After: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api. Treat this response as a rate limit and derive reset_at before generic error handling.

qontract_utils/qontract_utils/github_org/api.py
```diff
@@ -12,21 +12,120 @@
...
+    Prefers `Retry-After` (used for GitHub's secondary rate limit, seconds
+    from now) over `X-RateLimit-Reset` (used for the primary rate limit, an
+    absolute epoch timestamp). Falls back to a conservative default when
+    neither header is present.
+    """
+    if retry_after := headers.get("retry-after"):
+        return datetime.now(UTC) + timedelta(seconds=int(retry_after))
+    if reset := headers.get("x-ratelimit-reset"):
+        return datetime.fromtimestamp(int(reset), tz=UTC)
+    return datetime.now(UTC) + timedelta(seconds=_DEFAULT_RATE_LIMIT_FALLBACK_SECONDS)
```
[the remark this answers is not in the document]
chassing: Fixed in 68727b3f — `_reset_at_from_headers` now wraps both `int()` conversions in `contextlib.suppress` and falls back to the conservative default when a header value doesn't parse.

qontract_api/qontract_api/integrations/github_owners/service.py
chassing: Fixed in 7a045ff0 — the apply loop now breaks for that org once a rate limit is hit, instead of trying (and failing) the remaining actions against the same exhausted credential.

qontract_api/qontract_api/integrations/github_owners/service.py
chassing: Yes, intentional — documented in 5c7b122d (AGENTS.md Service Layer Patterns) as a sanctioned carve-out: unlike a generic silently-skipped unit, the reset time is known upfront, no operator action is possible before it, and the FAILURE alert it would otherwise trigger is exactly what this PR exists to remove. Appreciate the push to make that explicit rather than leaving it as an implicit exception to the convention.

qontract_api/qontract_api/integrations/github_owners/service.py
chassing: Fixed in 7a045ff0 — GithubOrgWorkspaceClient.add_member_as_admin now catches GithubRateLimitExceededError, caches the marker (same as get_current_members), and re-raises. Combined with 7a045ff0's apply-loop break, this closes both the stale-cache bypass and the quota burn for remaining actions.

qontract_utils/qontract_utils/github_org/api.py
chassing: Fixed in 68727b3f — _rate_limit_reset_at now treats any 403 carrying a Retry-After header as a rate limit (in addition to X-RateLimit-Remaining=="0"), matching GitHub's documented secondary-limit behavior. Applied consistently to both the GithubException and requests.HTTPError branches, not just the direct-request path.

qontract_utils/qontract_utils/github_org/api.py
```diff
@@ -12,21 +12,140 @@
...
+
+    Prefers `Retry-After` (used for GitHub's secondary rate limit, seconds
+    from now) over `X-RateLimit-Reset` (used for the primary rate limit, an
+    absolute epoch timestamp). Falls back to a conservative default when
+    neither header is present or a value fails to parse - `Retry-After` may
+    also carry an HTTP-date per RFC 9110, which `int()` cannot parse.
+    """
+    if retry_after := headers.get("retry-after"):
+        with suppress(ValueError):
+            return datetime.now(UTC) + timedelta(seconds=int(retry_after))
```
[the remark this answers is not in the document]
chassing: Good catch. Fixed in 2a935214 — added OverflowError to the suppress() on the retry-after branch, matching what was already done on the x-ratelimit-reset branch.

fishi0x01 (approved):
LGTM

qontract_api/qontract_api/github/github_org_workspace_client.py
```diff
@@ -85,12 +118,20 @@ def get_current_members(self, org_name: str) -> list[str]:
         if cached := self._cache.get_obj(cache_key, CachedOrgMembers):
             return cached.members
 
+        rate_limit_key = self._rate_limit_key(org_name)
+        if marker := self._cache.get_obj(rate_limit_key, RateLimitMarker):
+            raise GithubRateLimitExceededError(marker.reset_at)
```
hemslo: [P1] Check the rate-limit marker before the member-cache hit.

The marker is written after a rate-limited PUT while the member cache is retained. This order still returns stale members on the next cycle, so the same PUT is retried. Move this lookup above the cache hit; also recheck it after acquiring the lock to cover a marker written while waiting.

```suggestion
        rate_limit_key = self._rate_limit_key(org_name)
        if marker := self._cache.get_obj(rate_limit_key, RateLimitMarker):
            raise GithubRateLimitExceededError(marker.reset_at)

        if cached := self._cache.get_obj(cache_key, CachedOrgMembers):
            return cached.members
```

chassing: Good catch. Fixed in 857b7d99 — `get_current_members` now checks the rate-limit marker before the member-cache hit, and rechecks it after acquiring the per-org lock to close the race with a marker written while waiting.

qontract_api/qontract_api/github/github_org_workspace_client.py
```diff
@@ -89,8 +126,15 @@ def get_current_members(self, org_name: str) -> list[str]:
             if cached := self._cache.get_obj(cache_key, CachedOrgMembers):
                 return cached.members
 
-            admin_members = self._api.get_admin_members(org_name)
-            pending_invitations = self._api.get_pending_invitations(org_name)
+            if marker := self._cache.get_obj(rate_limit_key, RateLimitMarker):
+                raise GithubRateLimitExceededError(marker.reset_at)
```
hemslo: [P2] Check the marker before the locked cache hit.

The initial marker check handles normal calls, but this second check is after returning `CachedOrgMembers`. A caller that saw no cache before waiting for the lock can still observe a newly-written marker and a populated members cache here, then return stale data. Swap these checks.

```suggestion
            if marker := self._cache.get_obj(rate_limit_key, RateLimitMarker):
                raise GithubRateLimitExceededError(marker.reset_at)

            if cached := self._cache.get_obj(cache_key, CachedOrgMembers):
                return cached.members
```

chassing: Fixed in 1987876e — `get_current_members` now checks the rate-limit marker before the members-cache hit inside the lock too, matching the outer check order and closing the race where a marker written while waiting for the lock could be shadowed by a concurrently-populated members cache.

hemslo (approved):
LGTM!

PR #5809, opened 2026-09-08
feat(slack-usergroups): resolve github org members into usergroups

chassing:
## Summary

Add **GitHub organization membership** as a source of truth for Slack usergroup members ([APPSRE-15221](https://redhat.atlassian.net/browse/APPSRE-15221)).

Use case: broadcasting to all members of a GitHub org (e.g. security announcements) without maintaining a parallel member list in app-interface.

GitHub org members are resolved to Red Hat `org_username`s in two stages:

1. **app-interface `github_username`** — the primary mapping.
2. **LDAP `rhatSocialURL` fallback** — for members not present in app-interface, the GitHub login is resolved to a Red Hat uid via the `rhatSocialURL` attribute (`Github->https://github.com/<login>`), scoped to `cn=users,cn=accounts`.

Members that cannot be resolved by either stage are **logged and skipped**. The client-side warning lists every unmapped member's GitHub profile URL and a hint to add it to their Rover page, so the message is actionable in app-interface MR check output (which only surfaces client integration logs).

### Change set (bottom-up across the three layers)

- **`reconcile/` GraphQL query** — new optional `github` field (org name + Vault token) on the slack-usergroups permission query.
- **`qontract_utils/` (Layer 1)** — stateless API clients: `github_org` (fetch org members) and `ldap_api` (resolve GitHub logins to uids via `rhatSocialURL`, accepting only bare `github.com/<login>` profile URLs and preserving original login casing; the lookup is paged via the RFC 2696 paged results control).
- **`qontract_api/` (Layer 2 + routers)** — workspace clients + external endpoints: `GET /api/v1/external/github-org/members` and `GET /api/v1/external/ldap/github-usernames` (cached with a dedicated TTL, guarded by the double-check locking pattern).
- **`qontract_api_client/`** — regenerated client for the new endpoints.
- **`reconcile/slack_usergroups_api.py`** — resolve org members client-side, honoring an app-interface user's Slack identity override (`gov_slack_email_local_part`) when present and otherwise using the LDAP-resolved `org_username` directly.
- **Fluentd config** — exclude the aggregated "could not map N GitHub member(s)" warning from the aggregated log stream (same treatment as the `request_id` exclusion), across `template.yaml`, the rendered `openshift/` manifests, and the golden helm fixtures.
- **Docs** — document the new source of truth and mapping behavior.

### Precise LDAP fallback

- **Bare-profile-only parsing + scoped ambiguity** (`ldap_api` + `qontract_api`, decbe09f) — associates link many kinds of `github.com` URLs in `rhatSocialURL`: their own profile, their own repos, organizations they belong to, and other people's repos. The parser previously took the first path segment of *any* github.com URL, which conflated all four — attributing org names (stackrox, openshift, …) or another user's repo owner to the linking associate, and logging an ambiguity warning for every ambiguous directory key at map-build time regardless of the org being reconciled. Now:
  - Layer 1 `_parse_github_login` accepts **only** a bare `github.com/<login>` URL (exactly one path segment). A repo owner also links their bare profile, so no genuine login is lost, but org / other-repo names are no longer misattributed.
  - Layer 1 `get_github_usernames` returns `dict[str, list[str]]`, grouping every uid that claims a (lowercased) login — pure data extraction, no dropping/logging.
  - Layer 2 `resolve_github_usernames` applies the ambiguity policy **scoped to the requested logins**: a requested login mapping to >1 uid is skipped and logged once; ambiguity on unrelated directory entries never reaches the logs.

### Review follow-ups (CodeRabbit)

- **Ambiguous GitHub logins** (`ldap_api`, 1c8044ab) — a GitHub login that maps case-insensitively to more than one distinct LDAP uid is now omitted from the mapping and logged, instead of letting arbitrary LDAP result order pick a winner. Added regression tests. (Superseded by the scoped-ambiguity handling in decbe09f above.)
- **Response contract** (`qontract_api`, ae4dc4e6) — `GithubOrgMembersResponse.members` and `LdapGithubUsernamesResponse.users` are now `Field(...)` (required) rather than `default_factory=list`, so the schema, `openapi.json`, and generated client all agree they are always present. Client regenerated (no diff).
- **Typed test fixtures** (`qontract_api`, 3c9f42d7) — the github-org and ldap github-usernames router tests build their request fixtures from `GithubOrgMembersParams` / `LdapGithubUsernamesRequest` / `LdapDirectSecret` and serialize via `model_dump`, so a schema change fails the fixture at construction instead of drifting silently.
- **Raw ldap3 entry contract** (`ldap_api`) — declined: `_paged_search` returns the generic ldap3 search-result entries consumed in-module (identical to the shape `get_users`/`get_group_members` already consume inline); the typed Pydantic contracts live at the boundary (`LdapUser`/`LdapGroup`/`LdapGithubUser`).

### Review follow-ups (hemslo)

- **Paginate LDAP lookup** (`ldap_api`, 6a4a0b01) — 389 Directory Server caps a single search response at ~2000 entries, so the unpaged `rhatSocialURL` lookup silently dropped associates past that limit. The lookup now uses the RFC 2696 paged results control (page size 1000), looping on the cookie until exhausted. Added a regression test asserting results merge across pages.
- **Invalidate both member caches** (`qontract_api`, d9097547) — `add_member_as_admin` cleared only the `:members` key; the `:all-members` key consumed by slack-usergroups was left stale until TTL. `_clear_cache` now invalidates both keys (each under a key-scoped lock).
- **Aggregate the unmapped-member warning** (`slack-usergroups`, 12bd1b17) — instead of one warning line per unmapped member, emit a single warning per org that still lists every unmapped login with its github.com profile URL, so MR-check output keeps full visibility for the MR author. The fluentd exclusion filter was updated to match the aggregated message (`could not map \d+ GitHub member`) and the openshift manifests regenerated.
- **Docs in sync with aggregated warning** (`docs`, 8fb87da0) — the GitHub-org-members doc section now describes the single aggregated per-org warning (listing every unresolved login) instead of one warning per login.
- **Cache partition by credential** — declined: in app-interface each GitHub org maps 1:1 to a credential, so the org-scoped cache key cannot collide across credentials.
- **Conditional LDAP settings load** — declined: the integration always needs LDAP settings to resolve org members, so an unconditional load is simplest.

## Test plan

- `make linter-test` — ruff check + format, clean.
- `make types-test` — mypy strict, clean.
- `make helm-test` — helm lint + `git diff --exit-code helm openshift`, clean (rendered manifests + fixtures regenerated).
- `make unittest` — reconcile + qontract_utils suites pass, including a comprehensive `reconcile/test/test_slack_usergroups_api.py` covering all user sources, permission processing, workspace assembly, and `async_run` orchestration (dry-run / apply / polling / errors), plus the github-org resolution paths.
- `qontract_api` and `qontract_utils` package `make test` pass (mypy + ruff + openapi-diff + pytest), including the new LDAP pagination, both-cache-invalidation, and scoped-ambiguity regression tests.

Verified against live LDAP that an org member present only in LDAP (valid `rhatSocialURL`, absent from app-interface) is now resolved rather than skipped.

## Summary by CodeRabbit

* **New Features**
  * Added GitHub organization membership support for Slack user groups.
  * Added authenticated APIs for retrieving organization members and resolving GitHub usernames to LDAP users.
  * Added case-insensitive matching, caching, and LDAP fallback for membership resolution.
  * Added GitHub organization and token fields to Slack user group permissions.
* **Bug Fixes**
  * Unresolved or ambiguous members are skipped without interrupting processing, with consolidated warnings.
  * Reduced forwarding of expected member-mapping messages in integration logs.
* **Documentation**
  * Updated integration documentation with membership sources, workflows, validation, and API details.
* **Tests**
  * Added coverage for membership, LDAP resolution, caching, authentication, and fallback behavior.

### DIFF
--- a/docs/integrations/slack-usergroups.md
+++ b/docs/integrations/slack-usergroups.md
@@ -1,10 +1,10 @@
 # Slack Usergroups Integration
 
-**Last Updated:** 2025-11-26
+**Last Updated:** 2026-09-08
 
 ## Description
 
-This integration manages Slack usergroups across multiple workspaces (Slack instances) by reconciling desired state from app-interface with current state in Slack. It supports Slack Usergroup management based on roles, schedules, git repository ownership, and PagerDuty on-call rotations. The integration uses a client-server architecture where GraphQL queries happen client-side and reconciliation logic runs server-side via qontract-api.
+This integration manages Slack usergroups across multiple workspaces (Slack instances) by reconciling desired state from app-interface with current state in Slack. It supports Slack Usergroup management based on roles, schedules, git repository ownership, PagerDuty on-call rotations, and GitHub organization membership. The integration uses a client-server architecture where GraphQL queries happen client-side and reconciliation logic runs server-side via qontract-api.
 
 ## Features
 
@@ -14,6 +14,7 @@ This integration manages Slack usergroups across multiple workspaces (Slack inst
   - Time-based on-call schedules (with active time window filtering)
   - Git repository OWNERS files (GitHub/GitLab)
   - PagerDuty schedules and escalation policies
+  - GitHub organization members (e.g. to broadcast to all members of an org)
   - Cluster access roles (automatic cluster-specific usergroups)
 - Update usergroup metadata (description, default channels)
 - Whitelist-based management (only reconciles explicitly managed usergroups)
@@ -22,7 +23,7 @@ This integration manages Slack usergroups across multiple workspaces (Slack inst
 
 ## Desired State Details
 
-The desired state for Slack usergroup membership is compiled from **five different sources**, each with specific inclusion conditions. All sources are combined using set union logic (no duplicates), and the final user list is sorted alphabetically.
+The desired state for Slack usergroup membership is compiled from **six different sources**, each with specific inclusion conditions. All sources are combined using set union logic (no duplicates), and the final user list is sorted alphabetically.
 
 ### User Sources
 
@@ -66,7 +67,35 @@ Users are included from PagerDuty if:
 - `/api/v1/external/pagerduty-schedule-users`
 - `/api/v1/external/pagerduty-escalation-policy-users`
 
-#### 5. Cluster Access (Automatic Cluster Usergroups)
+#### 5. GitHub Organization Members (`/access/permission-1.yml.github`)
+
+All members of a GitHub organization are included. This is intended for
+broadcasting to an entire organization (e.g. security announcements) without
+maintaining an explicit member list in app-interface.
+
+Configuration provides the organization `name` and a Vault `token` reference
+with read access to the org's membership.
+
+**User Mapping (two stages, GitHub login → org_username):**
+
+1. **app-interface `github_username`** — a member is mapped to the app-interface
+   user whose `github_username` matches the GitHub login (compared
+   case-insensitively).
+2. **LDAP `rhatSocialURL` fallback** — remaining unmapped logins are resolved
+   via the FreeIPA `rhatSocialURL` attribute (`Github->https://github.com/<login>`),
+   which maps the GitHub login to the user's LDAP `uid` (the app-interface
+   `org_username`). The full map is cached server-side.
+
+Members that resolve to neither an app-interface `github_username` nor an LDAP
+`rhatSocialURL` entry are **logged and skipped** (one warning per unresolved
+login); they do not block reconciliation of the rest of the usergroup.
+
+**API Endpoints:**
+
+- `/api/v1/external/github-org/members`
+- `/api/v1/external/ldap/github-usernames`
+
+#### 6. Cluster Access (Automatic Cluster Usergroups)
 
 **Automatic Generation:** No explicit permission needed - generated from cluster access roles
 
@@ -107,6 +136,8 @@ After compiling users from all sources, the following validation and filtering o
 - Only users with `org_username` field are included
 - For PagerDuty, RedHat users no longer require an app-interface user file (`/access/user-*.yml`)
 - GitHub username mapping for OWNERS files requires `user.github_username` field
+- GitHub organization members are mapped via `user.github_username`, falling back
+  to the LDAP `rhatSocialURL` attribute; unresolved members are logged and skipped
 - `tag_on_merge_requests` filtering for git OWNERS (default: true)
 - `tag_on_cluster_updates` filtering for cluster usergroups (default: true)
 
@@ -121,13 +152,14 @@ After compiling users from all sources, the following validation and filtering o
 **Client-Side ([reconcile/slack_usergroups_api.py](../../reconcile/slack_usergroups_api.py)):**
 
 - Fetches desired state:
-  - from app-interface (GraphQL) permissions, users, clusters, and roles
-  - from qontract-api external endpoints (VCS repo owners, PagerDuty)
+  - from app-interface (GraphQL) permissions, users, clusters, roles, and LDAP settings
+  - from qontract-api external endpoints (VCS repo owners, PagerDuty, GitHub org members)
 - Compiles usergroup membership from multiple sources:
   - Roles (with expiration filtering)
   - Time-based schedules (filters by active time windows)
   - Git OWNERS files (via qontract-api external endpoints)
   - PagerDuty schedules and escalation policies (via qontract-api)
+  - GitHub organization members (via qontract-api, mapped via github_username/LDAP)
   - Cluster access roles (generates cluster-specific usergroups)
 - Transforms data to API request format
 - Calls qontract-api reconciliation endpoint
@@ -485,6 +517,8 @@ The integration uses GraphQL queries to fetch desired state from app-interface. 
    - Time-based schedules (filtered by active time window)
    - Git OWNERS files (via qontract-api `/api/v1/external/vcs-repo-owners`)
    - PagerDuty schedules/policies (via qontract-api `/api/v1/external/pagerduty-*`)
+   - GitHub org members (via qontract-api `/api/v1/external/github-org/members`,
+     mapped via `github_username` then `/api/v1/external/ldap/github-usernames`)
    - Cluster access roles
 3. Build `SlackWorkspace` objects with usergroups
 4. POST to `/reconcile` endpoint (queue task)
--- a/docs/integrations/slack-usergroups.md
+++ b/docs/integrations/slack-usergroups.md
@@ -87,8 +87,10 @@ with read access to the org's membership.
    `org_username`). The full map is cached server-side.
 
 Members that resolve to neither an app-interface `github_username` nor an LDAP
-`rhatSocialURL` entry are **logged and skipped** (one warning per unresolved
-login); they do not block reconciliation of the rest of the usergroup.
+`rhatSocialURL` entry are **logged and skipped**: a single aggregated warning is
+emitted per organization, listing every unresolved login (with its
+`github.com` profile URL) so the MR author sees them in the app-interface MR
+check output. They do not block reconciliation of the rest of the usergroup.
 
 **API Endpoints:**
 
--- a/qontract_api/openapi.json
+++ b/qontract_api/openapi.json
@@ -662,6 +662,21 @@
         "title": "GithubOrgDesiredState",
         "type": "object"
       },
+      "GithubOrgMembersResponse": {
+        "description": "Response with all members of a GitHub organization.",
+        "properties": {
+          "members": {
+            "description": "GitHub usernames (original case) of all organization members",
+            "items": {
+              "type": "string"
+            },
+            "title": "Members",
+            "type": "array"
+          }
+        },
+        "title": "GithubOrgMembersResponse",
+        "type": "object"
+      },
       "GithubOwnerActionAddOwner": {
         "description": "Action: Add a user as an org admin (owner).\n\nNote: Owner removal is intentionally NOT supported. The integration only\nadds owners and never removes them, preserving the behavior of the original\ngithub-owners reconcile integration. This is a deliberate safety decision \u2014\nremoving org admins is a high-impact operation that requires explicit manual\nreview rather than automated removal.",
         "properties": {
@@ -2195,6 +2210,65 @@
           "base_dn"
         ],
         "title": "LdapDirectSecret",
+        "type": "object"
+      },
+      "LdapGithubUser": {
+        "description": "Resolved mapping of a single GitHub username to an LDAP uid.",
+        "properties": {
+          "github_username": {
+            "description": "GitHub username (as requested)",
+            "title": "Github Username",
+            "type": "string"
+          },
+          "org_username": {
+            "description": "LDAP uid (app-interface org_username)",
+            "title": "Org Username",
+            "type": "string"
+          }
+        },
+        "required": [
+          "github_username",
+          "org_username"
+        ],
+        "title": "LdapGithubUser",
+        "type": "object"
+      },
+      "LdapGithubUsernamesRequest": {
+        "description": "Request to resolve GitHub usernames to LDAP uids via rhatSocialURL.",
+        "properties": {
+          "logins": {
+            "description": "GitHub usernames to resolve to LDAP uids",
+            "items": {
+              "type": "string"
+            },
+            "title": "Logins",
+            "type": "array"
+          },
+          "secret": {
+            "$ref": "#/components/schemas/LdapDirectSecret",
+            "description": "Vault secret reference for LDAP credentials"
+          }
+        },
+        "required": [
+          "logins",
+          "secret"
+        ],
+        "title": "LdapGithubUsernamesRequest",
+        "type": "object"
+      },
+      "LdapGithubUsernamesResponse": {
+        "description": "Response with resolved GitHub-username -> uid mappings.\n\nOnly contains entries for requested logins that were found in LDAP;\nunresolved logins are omitted.",
+        "properties": {
+          "users": {
+            "description": "Resolved GitHub username to org_username mappings",
+            "items": {
+              "$ref": "#/components/schemas/LdapGithubUser"
+            },
+            "title": "Users",
+            "type": "array"
+          }
+        },
+        "title": "LdapGithubUsernamesResponse",
         "type": "object"
       },
       "LdapUserStatus": {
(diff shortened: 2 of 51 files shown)


maorfr (approved):

jfreixa (approved):

qontract_api/openapi.json
```diff
@@ -662,6 +662,21 @@
...
+            "items": {
+              "type": "string"
+            },
+            "title": "Members",
+            "type": "array"
+          }
+        },
+        "title": "GithubOrgMembersResponse",
+        "type": "object"
+      },
```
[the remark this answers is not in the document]
chassing: Fixed in ae4dc4e6. `GithubOrgMembersResponse.members` and `LdapGithubUsernamesResponse.users` now use `Field(...)` (no `default_factory`), so the schema, `openapi.json`, and the generated client all agree these fields are required — the server always populates them.

qontract_api/qontract_api/external/ldap/ldap_workspace_client.py
```diff
@@ -27,6 +27,12 @@ class CachedUserCheck(BaseModel, frozen=True):
     result: list[LdapUserStatus]
 
 
+class CachedGithubUsernames(BaseModel, frozen=True):
+    """Cached GitHub-username -> uid map (for two-tier cache serialization)."""
+
+    mapping: dict[str, str]
```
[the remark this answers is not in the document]
chassing: Keeping the `dict[str, str]` mapping here. This is an internal Layer 2 compute/cache type; the typed Pydantic contract already lives at the API boundary (`LdapGithubUser` in `schemas.py`). Introducing a record type for the internal path adds indirection without changing the wire contract, so we're leaving it as-is for this PR.

qontract_api/qontract_api/external/ldap/router.py
```diff
@@ -66,3 +69,51 @@ def check_users_exist(
...
+    Args:
+        request: GitHub usernames to resolve and the Vault secret reference
+        cache: Cache dependency
+        secret_manager: Secret manager dependency
+
+    Returns:
+        LdapGithubUsernamesResponse with resolved username -> org_username pairs
+    """
+    client = create_ldap_workspace_client(
+        secret=request.secret,
```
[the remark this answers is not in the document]
chassing: Not applying this one. The SSRF concern is mitigated by the existing authorization layers: every request carries a JWT that is authenticated and authorized, and OPA enforces authN/authZ on top. The `server_url`/`base_dn` come from a trusted, app-interface-managed Vault secret reference — not arbitrary user input — so there is no untrusted-destination path here.

qontract_api/qontract_api/github/github_org_workspace_client.py
```diff
@@ -71,6 +71,10 @@ def __init__(
     def _cache_key(org_name: str) -> str:
         return f"github-org:{org_name}:members"
 
+    @staticmethod
+    def _all_members_cache_key(org_name: str) -> str:
+        return f"github-org:{org_name}:all-members"
```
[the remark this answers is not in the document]
chassing: Keeping the current cache key. `github_org_workspace_client` is a pre-existing shared client with a 1:1 org→credential convention (also used by `github-owners`), and the members of a given org are the same regardless of which configured token lists them. Partitioning by credential identity would fragment the cache without a correctness benefit in this model; changing that established convention is out of scope for this PR.

qontract_api/tests/external/github_org/test_github_org_members_router.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+MEMBERS_ENDPOINT = "/api/v1/external/github-org/members"
+
+QUERY_PARAMS = {
+    "org_name": "my-org",
+    "secret_manager_url": "https://vault.example.com",
+    "path": "secret/github/my-org",
+    "field": "token",
+}
```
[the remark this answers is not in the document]
chassing: Fixed in 3c9f42d7. The github-org and ldap github-usernames router tests now build their request fixtures from `GithubOrgMembersParams`, `LdapGithubUsernamesRequest`, and `LdapDirectSecret` and serialize them via `model_dump`, so a schema change fails the fixture at construction instead of drifting silently.

qontract_utils/qontract_utils/ldap_api/api.py
```diff
@@ -271,3 +308,39 @@ def get_group_members(self, groups_dns: Collection[str]) -> list[LdapGroup]:
             )
             for dn, members in groups_and_members.items()
         ]
+
+    @invoke_with_hooks(
+        lambda: LdapApiCallContext(method="get_github_usernames"),
+        retry_config=_LDAP_RETRY_CONFIG,
+    )
+    def get_github_usernames(self) -> dict[str, str]:
```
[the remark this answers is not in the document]
chassing: Keeping the `dict[str, str]` return type in this Layer 1 client. The typed Pydantic contract lives at the API boundary (`LdapGithubUser`); the Layer 1 method returns a plain login→uid map by design, mirroring `get_group_members`'s internal style. We did harden the semantics of this method though — see 1c8044ab, which now rejects ambiguous logins instead of overwriting.

qontract_utils/qontract_utils/ldap_api/api.py
```diff
@@ -271,3 +308,39 @@ def get_group_members(self, groups_dns: Collection[str]) -> list[LdapGroup]:
...
+        )
+        self._check_ldap_response(status)
+
+        mapping: dict[str, str] = {}
+        for r in results:
+            attributes = r["attributes"]
+            uid = attributes["uid"][0]
+            for value in attributes.get("rhatSocialURL", []):
+                if login := _parse_github_login(value):
+                    mapping[login] = uid
```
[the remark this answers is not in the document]
chassing: Fixed in 1c8044ab. Ambiguous GitHub logins (a login mapping case-insensitively to more than one distinct LDAP uid) are now omitted from the mapping and logged with a warning, instead of letting arbitrary LDAP result order pick a winner. Added regression tests for the ambiguous case and the duplicate-same-uid case.

reconcile/slack_usergroups_api.py
```diff
@@ -377,11 +385,125 @@ async def compile_users_from_pagerduty_schedules(
...
+        for login in members:
+            if org_username := gh_to_org_username.get(login.lower()):
+                resolved[login] = org_username
+            else:
+                unresolved.append(login)
+
+        # Fall back to LDAP rhatSocialURL for members without a github_username
+        resolved.update(
+            await self._resolve_github_logins_via_ldap(unresolved, ldap_settings)
+        )
```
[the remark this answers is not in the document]
chassing: This is already handled, so no change needed. `_resolve_github_logins_via_ldap` returns the mapping keyed by the login as requested, and the server side (`LdapWorkspaceClient.resolve_github_usernames`) matches case-insensitively (`lookup[login.lower()]`) while echoing back the originally-requested login. The `resolved.get(login)` lookup in `compile_users_from_github_org` therefore uses the exact key that was sent — traced the full client→server→client path to confirm no mixed-case miss occurs.

qontract_api/qontract_api/github/github_org_workspace_client.py
```diff
@@ -71,6 +71,10 @@ def __init__(
     def _cache_key(org_name: str) -> str:
         return f"github-org:{org_name}:members"
 
+    @staticmethod
+    def _all_members_cache_key(org_name: str) -> str:
```
hemslo, line 75: [P1] Partition GitHub caches by credential identity. The membership and rate-limit caches are keyed only by organization, while this endpoint can build clients from different Vault credentials. Different credentials can therefore share membership data and throttling state. Include a stable non-secret credential identity in the relevant keys and add a cross-credential regression test.

reconcile/slack_usergroups_api.py
```diff
@@ -649,11 +781,13 @@ async def async_run(self, dry_run: bool) -> None:
         clusters = self.get_clusters(query_func=gqlapi.query)
         roles = self.get_roles(query_func=gqlapi.query)
         vcs_instances = get_vcs_instances(query_func=gqlapi.query)
+        ldap_settings = get_ldap_settings(query_func=gqlapi.query)
```
hemslo, line 794: [P2] Load LDAP settings only when required. `async_run()` now calls `get_ldap_settings()` for every Slack usergroups run, even when no permission uses GitHub membership or there is no desired state. This makes existing non-GitHub configurations depend on LDAP being configured. Load LDAP settings conditionally and add a no-GitHub/no-LDAP regression test.

qontract_api/qontract_api/github/github_org_workspace_client.py
```diff
@@ -145,6 +149,49 @@ def get_current_members(self, org_name: str) -> list[str]:
...
+        original case exactly as GitHub reports them; callers that need
+        case-insensitive comparison must lowercase on their side.
+
+        Args:
+            org_name: GitHub organization name
+
+        Returns:
+            Sorted list of GitHub usernames (original case) of all org members
+        """
+        cache_key = self._all_members_cache_key(org_name)
```
hemslo, line 176: [P2] Invalidate the all-members cache after owner mutations. `get_all_members()` writes the `:all-members` key, while `add_member_as_admin()` still clears only the legacy `:members` key. After an owner operation, Slack membership can remain stale until TTL expiry. Clear both keys with matching lock protection and add regression coverage.

qontract_utils/qontract_utils/ldap_api/api.py
```diff
@@ -271,3 +308,59 @@ def get_group_members(self, groups_dns: Collection[str]) -> list[LdapGroup]:
...
+        identity-selection rule - so it is omitted from the mapping and logged,
+        rather than letting an arbitrary entry win.
+
+        Returns:
+            Mapping of GitHub username to LDAP uid (org_username)
+
+        Raises:
+            LdapApiError: If the LDAP search fails
+        """
+        _, status, results, _ = self._connection.search(
```
hemslo: [P1] Paginate the LDAP GitHub mapping search. `get_github_usernames()` calls `connection.search()` once without `paged_size`, and the broad filter retrieves every matching GitHub-linked LDAP user. The documented 389 Directory Server default search limit is 2,000 entries; when exceeded, the server returns partial entries plus `LDAP_SIZELIMIT_EXCEEDED`, which `_check_ldap_response()` raises. Use `ldap3` paged-search/cookie iteration and add multi-page coverage.

reconcile/slack_usergroups_api.py
```diff
@@ -377,11 +385,125 @@ async def compile_users_from_pagerduty_schedules(
...
+
+        # Fall back to LDAP rhatSocialURL for members without a github_username
+        resolved.update(
+            await self._resolve_github_logins_via_ldap(unresolved, ldap_settings)
+        )
+
+        slack_identities: set[str] = set()
+        for login in members:
+            if not (org_username := resolved.get(login)):
+                logging.warning(
```
hemslo: [P2] Avoid warning for expected external members. Organizations can legitimately contain external or non-Red Hat users who have no app-interface or LDAP mapping, and this branch logs one warning per such member on every reconciliation. Would a debug-level message or one aggregated/rate-limited warning per organization be more appropriate? Keep warning-level logging for unexpected mapping failures.

qontract_utils/qontract_utils/ldap_api/api.py
chassing: Fixed in 6a4a0b01. The lookup now uses the RFC 2696 paged results control (page size 1000) via a new `_paged_search` helper that loops on the cookie until exhausted, so associates past 389 DS's ~2000-entry single-response cap are no longer silently dropped.

qontract_api/qontract_api/github/github_org_workspace_client.py
chassing: Fixed in d9097547. `_clear_cache` now invalidates both the `:members` and `:all-members` keys (each under a lock scoped to that key), so a freshly added member shows up in the slack-usergroups resolution immediately instead of waiting for the `:all-members` TTL.

reconcile/slack_usergroups_api.py
chassing: Fixed in 12bd1b17. The warning is now aggregated into a single line per org, but it deliberately still lists every unmapped GitHub login (with its github.com profile URL) so app-interface MR checks surface the full set to the MR author. The `#reconcile` fluentd exclusion filter was updated accordingly (`could not map \d+ GitHub member`), and the openshift manifests were regenerated via `make generate`.

qontract_api/qontract_api/github/github_org_workspace_client.py
chassing: Keeping the cache key partitioned by org name only. In app-interface each GitHub org maps to exactly one credential (1:1 org→credential convention), so `github-org:{org}:members` cannot collide across credentials — adding the credential to the key would only bloat it without preventing any real collision. Happy to revisit if we ever introduce multiple credentials per org.

reconcile/slack_usergroups_api.py
chassing: Leaving the LDAP settings loaded unconditionally. The integration always needs them to resolve GitHub org members via LDAP, so a conditional load would add a branch without saving any meaningful work in the common path. If a config surfaces where the settings are genuinely optional we can make it conditional then.

docs/integrations/slack-usergroups.md
```diff
@@ -66,7 +67,35 @@ Users are included from PagerDuty if:
...
+
+1. **app-interface `github_username`** — a member is mapped to the app-interface
+   user whose `github_username` matches the GitHub login (compared
+   case-insensitively).
+2. **LDAP `rhatSocialURL` fallback** — remaining unmapped logins are resolved
+   via the FreeIPA `rhatSocialURL` attribute (`Github->https://github.com/<login>`),
+   which maps the GitHub login to the user's LDAP `uid` (the app-interface
+   `org_username`). The full map is cached server-side.
+
+Members that resolve to neither an app-interface `github_username` nor an LDAP
```
hemslo, line 89: [P3] Keep the warning documentation in sync. This section still says unresolved members produce one warning per unresolved login, but the current implementation emits one aggregated warning per organization. Please update the wording to describe the aggregate behavior so operators do not expect one log line per member.

chassing: Fixed in 8fb87da0. The docs now describe the actual behavior: a single aggregated warning per organization that lists every unresolved login (with its `github.com` profile URL), instead of one warning per login.

qontract_utils/qontract_utils/ldap_api/api.py
```diff
@@ -200,6 +244,37 @@ def _check_ldap_response(status: dict) -> None:
                 f"LDAP operation failed (error {error_code}: {error_desc})"
             )
 
+    def _paged_search(
+        self, search_base: str, search_filter: str, attributes: list[str]
+    ) -> list[dict]:
```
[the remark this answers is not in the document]
chassing: Keeping `_paged_search` returning the raw ldap3 entries (`list[dict]`). These are the generic ldap3 search-result entries consumed immediately within this Layer 1 module; `get_users` and `get_group_members` already consume the identical raw-entry shape from inline `self._connection.search(...)` calls — `_paged_search` only factors out the RFC 2696 paging loop they would otherwise duplicate. The typed Pydantic contracts live at the boundary (`LdapUser`, `LdapGroup`, `LdapGithubUser`), and the raw entry is just a transient container whose `attributes` vary per query. Wrapping it in a model here would be inconsistent with the rest of the client for no added safety — same rationale as the `dict[str, str]` mapping you accepted as a Layer 1 exception.

qontract_api/tests/external/ldap/test_ldap_workspace_client.py
```diff
@@ -156,3 +157,170 @@ def test_check_users_exist_empty_input(
...
+) -> None:
+    """Ambiguity outside the requested logins produces no warning.
+
+    The directory holds many ambiguous logins (orgs/repos claimed by several
+    associates) that are irrelevant to the org being reconciled. Only requested
+    logins are considered, so those never spam the logs.
+    """
+    mock_cache.get_obj.return_value = CachedGithubUsernames(
+        mapping={
+            "stackrox": ["chsheth", "jvmartin"],  # ambiguous, but not requested
```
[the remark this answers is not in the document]
chassing: Fixed in 947b1fd9. The unrequested-ambiguity fixture now uses synthetic identifiers (`example-org`: `carol`/`dave`) instead of real GitHub org/user names, same test behavior.

qontract_api/qontract_api/external/ldap/ldap_workspace_client.py
```diff
@@ -27,6 +27,17 @@ class CachedUserCheck(BaseModel, frozen=True):
...
 
+class CachedGithubUsernames(BaseModel, frozen=True):
+    """Cached GitHub-login -> uid(s) map (for two-tier cache serialization).
+
+    Each lowercased login maps to the sorted list of uids that claim it; a login
+    with more than one uid is ambiguous and resolved (skipped + logged) per
+    request, not at build time.
+    """
+
+    mapping: dict[str, list[str]]
```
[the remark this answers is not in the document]
chassing: Not bumping the cache key for this. Nothing from this PR has been deployed to production yet, so there is no live payload in the old `dict[str, str]` shape to be incompatible with — the cache key change only matters for a rolling deploy/rollback against an already-running old version, which doesn't apply here. If this key's value shape changes again after this feature ships, a version bump would be the right call then.

hemslo (approved):
LGTM!

PR #5816, opened 2026-09-10
fix(vault-secret): adapt VaultSecret usages for new `url` field

chassing:
## Summary
- [app-sre/qontract-schemas#1073](https://github.com/app-sre/qontract-schemas/pull/1073) adds a new optional `url` field to `VaultSecret_v1`, so a secret reference can point at a Vault server other than the default AppSRE instance. This is needed for [APPSRE-15221](https://redhat.atlassian.net/browse/APPSRE-15221): the Red Hat SSO initial access token is hosted on the IT Vault instance, not the AppSRE one.
- Regenerated GraphQL models (`make gql-query-classes`) to pick up the new field.
- Adapted every place that constructs a `VaultSecret` directly (not via live GraphQL data) to pass `url=None`/`url: null`, since the generated model requires the field to be passed explicitly even though it's nullable.
- No behavioral change here — the `url` field is not consumed anywhere yet (that's follow-up work).

## Test plan
- [x] `uv run mypy` — clean (only pre-existing, unrelated `reconcile/utils/vault.py` hvac-stub errors remain)
- [x] `uv run ruff check` / `uv run ruff format --check` — clean
- [x] Full `reconcile`/`tools` test suite — identical failure set to a pre-change baseline run (pre-existing env-dependent failures only, e.g. missing `amtool`/`promtool`/`skopeo` binaries); zero new failures introduced

## Summary by CodeRabbit

- **New Features**
  - Vault-backed secrets now support an optional URL field.
  - Secret data retrieval includes the URL wherever Vault secrets are used.
  - Exported secret configurations can include the URL alongside existing credential details.

- **Compatibility**
  - Existing secrets without a URL remain supported; the field defaults to empty or unset.


fishi0x01 (approved):

reconcile/gql_definitions/fragments/vault_secret.py
```diff
@@ -25,6 +25,7 @@ class ConfiguredBaseModel(BaseModel):
 
 
 class VaultSecret(ConfiguredBaseModel):
+    url: Optional[str] = Field(..., alias="url")
```
hemslo, line 28: [P1] Preserve legacy OCM query results with this required field.\n\nThe generated VaultSecret field is nullable in type syntax but required at runtime because it uses Field(...). The legacy CLUSTERS_QUERY, CLUSTER_PEERING_QUERY, and OCM_QUERY in reconcile/queries.py still select only path, field, format, and version. Those raw dictionaries flow into OCMMap.init_ocm_client() and VaultSecret(**access_token_client_secret), so legacy OCM integrations now raise a Pydantic validation error before making the OCM call. Add url to those query selections or make this generated field default to None, and add a regression test for the legacy path.

chassing: Confirmed — reproduced this exactly (`pydantic.ValidationError: url Field required` via `OCMMap.init_ocm_client()`), and you're right that it wasn't caught because existing tests patch out `init_ocm_client_from_cluster`/`init_ocm_client` entirely.

Fixed in #5817: instead of adding `url` to the legacy query selections, I removed the `reconcile.gql_definitions` dependency from `reconcile/utils` for this construction (a `VaultSecretRef` dataclass implementing `HasSecret`), since `reconcile/utils` shouldn't depend on generated GQL types. Added a regression test that exercises the real code path end-to-end. Thanks for catching this!

PR #5817, opened 2026-09-10
fix(ocm): decouple VaultSecret construction from generated GQL model

chassing:
## Summary

Follow-up to #5816. [@hemslo's review comment](https://github.com/app-sre/qontract-reconcile/pull/5816#discussion_r3977520548) (posted after the PR merged) is correct:

- The generated `VaultSecret` model's new `url` field is nullable in type syntax but required at construction time (`Field(...)`).
- The legacy dict-based `CLUSTERS_QUERY`, `CLUSTER_PEERING_QUERY` and `OCM_QUERY` in `reconcile/queries.py` only ever select `path`/`field`/`format`/`version` for `accessTokenClientSecret` — never `url`.
- Those raw dicts flow into `OCMMap.init_ocm_client()`, which builds `VaultSecret(**access_token_client_secret)` directly — this now raises a `pydantic.ValidationError` for every legacy OCM integration path (e.g. `terraform_vpc_peerings.py` via `queries.get_clusters_with_peering_settings()`, or any `OCMMap(clusters=...)`/`OCMMap(ocms=...)` caller).
- This wasn't caught by existing tests because they all patch out `init_ocm_client_from_cluster`/`init_ocm_client` entirely (see `test_ocm_clusters.py`, `test_ocm_additional_routers.py`), so the `VaultSecret(...)` construction never actually ran in CI.

## Fix

Rather than adding `url` to the legacy query selections (it isn't consumed anywhere yet per #5816) or relaxing the generated model, this removes the `reconcile.gql_definitions` dependency from `reconcile/utils` entirely for this construction:

- Add a plain `VaultSecretRef` dataclass implementing the existing `HasSecret` protocol in `reconcile/utils/secret_reader.py`.
- Use it in `reconcile/utils/ocm/ocm.py` (legacy dict path) and `reconcile/utils/state.py` (env-var path) instead of the generated `VaultSecret` model — `reconcile/utils` should not depend on generated GraphQL query types.

## Test plan

- [x] Added `test_ocmmap_init_ocm_client_with_legacy_dict_missing_url` in `reconcile/test/utils/test_ocm.py`, exercising `OCMMap.init_ocm_client()` end-to-end (not mocked away) with a legacy-shaped dict lacking `url`.
- [x] Verified the test fails with `pydantic.ValidationError: url Field required` against the pre-fix code, and passes after the fix (TDD).
- [x] `uv run mypy` clean on all changed files.
- [x] `uv run ruff check` / `ruff format --check` clean.
- [x] Existing `reconcile/test/utils/test_state.py`, `reconcile/test/test_secret_reader.py`, `reconcile/test/test_ocm_clusters.py`, `reconcile/test/test_ocm_additional_routers.py`, `reconcile/test/ocm/` all still pass (466 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with legacy Vault secret configurations.
  * Preserved optional secret version and format settings when connecting to OCM.
  * Maintained existing Vault state credential version handling.

### DIFF
--- a/reconcile/test/utils/test_ocm.py
+++ b/reconcile/test/utils/test_ocm.py
@@ -4,8 +4,9 @@ from typing import TYPE_CHECKING, Any
 
 import pytest
 
-from reconcile.utils.ocm import OCM
+from reconcile.utils.ocm import OCM, OCMMap
 from reconcile.utils.ocm_base_client import OCMBaseClient
+from reconcile.utils.secret_reader import VaultSecretRef
 
 if TYPE_CHECKING:
     from pytest_mock import MockerFixture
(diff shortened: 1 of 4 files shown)


hemslo (approved):
LGTM!

PR #5819, opened 2026-09-11
perf(glhk): cache pipeline lists in one loop

rhur-pixel:
## Summary
- Cache GitLab pipeline lists for one reconcile loop so `run_error_healthcheck` and the serial merge path do not both call `get_merge_request_pipelines` for the same MRs ([APPSRE-14268](https://redhat.atlassian.net/browse/APPSRE-14268)).
- Lookup uses `iid in cache` so an empty `[]` is a hit, not a refetch. The dict is per-repo; `reload_toggle.reload` clears it in place; the 10-retry fallback gets a fresh `{}`.
- OMM and rebase still fetch live. Out of scope for this ticket.

With `--no-wait-for-pipeline`, a pipeline that flips `running` → `success` during healthcheck can wait until the next loop. `--wait-for-pipeline` + insist still retries and refetches.

## Test plan
- [ ] `uv run pytest reconcile/test/test_gitlab_housekeeping.py`
- [ ] Cache hit: success list → merge, no second pipeline API call
- [ ] Empty `[]` is a hit: no refetch, no merge
- [ ] `reload_toggle.reload` clears the cache and fetches again

## Performance
- Reduced repeated pipeline checks during merge request processing for faster operation.
- Reuses refreshed pipeline status to support current merge decisions after reloads.

## Bug Fixes
- Improved merge safety by detecting newer pipeline failures before completing a merge.
- Correctly handles empty pipeline results and resets cached results after reloads.

## Tests
- Added coverage for cached results, pipeline refreshes, empty results, and cache invalidation.

### DIFF
--- a/reconcile/gitlab_housekeeping.py
+++ b/reconcile/gitlab_housekeeping.py
@@ -1438,9 +1438,14 @@ def merge_merge_requests(
     users_allowed_to_label: Iterable[str] | None = None,
     must_pass: Iterable[str] | None = None,
     multi_merge: bool = False,
+    pipeline_cache: dict[int, list[ProjectMergeRequestPipeline]] | None = None,
 ) -> None:
     if reload_toggle.reload:
         project_merge_requests = gl.get_merge_requests(state=MRState.OPENED)
+        # Same dict as run() created. clear() is in-place so the insist
+        # retry does not reuse pipeline lists from before the wait.
+        if pipeline_cache is not None:
+            pipeline_cache.clear()
     merge_requests = preprocess_merge_requests(
         dry_run=dry_run,
         gl=gl,
@@ -1500,7 +1505,10 @@ def merge_merge_requests(
         if rebase and not is_rebased(mr, gl):
             continue
 
-        pipelines = gl.get_merge_request_pipelines(mr)
+        if pipeline_cache is not None and mr.iid in pipeline_cache:
+            pipelines = pipeline_cache[mr.iid]
+        else:
+            pipelines = gl.get_merge_request_pipelines(mr)
         if not pipelines:
             continue
 
@@ -1584,6 +1592,7 @@ def run_error_healthcheck(
     gl: GitLabApi,
     project_merge_requests: list[ProjectMergeRequest],
     consecutive_failure_limit: int = 3,
+    pipeline_cache: dict[int, list[ProjectMergeRequestPipeline]] | None = None,
 ) -> None:
     """Check error labels for queue-eligible MRs. Apply/remove
     rebase-error based on merge_error field from .get(),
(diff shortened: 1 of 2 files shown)


TGPSKI (commented):
Correction: withdrawing my previously posted P2 finding.

I rechecked PR head `8e39736f` and reproduced the relevant cases: cached success is refetched before merging, and cached running with `--wait-for-pipeline` plus insist triggers a retry, clears the cache, and fetches fresh state. This preserves the running-to-success behavior described in the PR.

A cached failure followed by a newly started pipeline can defer processing until a later reconcile loop. That timing change is real, but I overstated it as a defect: I did not establish a requirement to discover newly started pipelines within the same loop, and the behavior is consistent with the per-loop cache freshness tradeoff. It does not permit an unsafe merge or permanently prevent retries. The earlier request to refetch all results whenever waiting is enabled is withdrawn.

Validation: all 175 tests in `reconcile/test/test_gitlab_housekeeping.py` passed on this PR head; the additional mocked checks confirmed the cases above. No actionable defect remains from this finding.

TGPSKI (commented):
Reviewed commit `8e39736f`. No actionable findings.

Cached successful pipeline results are refreshed before merging. With `--wait-for-pipeline` and insist enabled, cached running pipelines trigger a retry that clears the cache and fetches fresh state. Cached non-success results can defer newly started pipelines until a later reconcile loop, consistent with the per-loop cache freshness tradeoff.

Validation: `uv run pytest reconcile/test/test_gitlab_housekeeping.py` — 175 passed. Additional mocked checks confirmed success-to-failure refresh, running-to-success retry, and failure-to-running deferral.

TGPSKI (approved):

reconcile/gitlab_housekeeping.py
```diff
@@ -1438,9 +1438,12 @@ def merge_merge_requests(
     users_allowed_to_label: Iterable[str] | None = None,
     must_pass: Iterable[str] | None = None,
     multi_merge: bool = False,
+    pipeline_cache: dict[int, list[ProjectMergeRequestPipeline]] | None = None,
```
hemslo: Could we make this a required keyword-only dict instead of a nullable optional? run() already creates the per-repo cache and the fallback passes a fresh empty dict, so this helper can always clear and read pipeline_cache without the None branches. Please avoid a mutable empty-dict default; update direct test callers with fresh dictionaries instead.

reconcile/gitlab_housekeeping.py
```diff
@@ -1584,6 +1597,7 @@ def run_error_healthcheck(
     gl: GitLabApi,
     project_merge_requests: list[ProjectMergeRequest],
     consecutive_failure_limit: int = 3,
+    pipeline_cache: dict[int, list[ProjectMergeRequestPipeline]] | None = None,
```
hemslo: Could we apply the same required keyword-only dict contract here? This helper is called from run() with the shared per-repo cache, so making pipeline_cache non-nullable removes the None guard on the assignment and makes the cache invariant explicit. Direct tests can pass fresh dictionaries; please do not use a mutable empty-dict default.

hemslo (approved):
LGTM!

PR #1077, opened 2020-09-18
[WIP] fix tf per account

BumbleFeng:
1. Run disable_keys per account
2. Skip populate_oc_resources for dry-run but keep oc_map for rds determine_db_password
3. Filter tf_namespaces with aws account
4. Filter ResourceInventory with aws account

Merge after https://github.com/app-sre/qontract-reconcile/pull/1074

Signed-off-by: Feng Huang <fehuang@redhat.com>

PR #5821, opened 2026-09-15
feat(managed-sso-client): deliver tenant credentials to namespaces as OpenShift Secrets

chassing:
## Summary
- New `openshift-managed-sso-client-secrets` integration: a namespace's `openshiftResources` entry can reference a `managed-sso-client` object directly (via the new `managed-sso-client` provider, see the paired [qontract-schemas PR #1077](https://github.com/app-sre/qontract-schemas/pull/1077)), and the reconciler resolves its tenant-facing Vault credential path and materializes it as an OpenShift Secret - implementing Milestone 2 ("secret delivery") of the [managed-sso-client design doc](https://github.com/openshift-online/platform-engineering-enhancements/pull/83).
- Class-based `QontractReconcileIntegration` with a `--vault-path-prefix` CLI option (must stay in sync with qontract-api's `ManagedSsoClientSettings.default_output_vault_path_prefix`).
- Shared `derive_managed_sso_client_id()` helper extracted to `reconcile/utils/managed_sso_client.py`, used by both this integration and `reconcile/managed_sso_client/integration.py`.
- Secret bodies use `stringData` (not manually base64-encoded `data`) - the apply/diff pipeline already normalizes `stringData` -> `data`.
- `orb.QONTRACT_INTEGRATION`/`QONTRACT_INTEGRATION_VERSION` are overridden before calling `orb.fetch_current_state()` directly, so this integration can recognize and garbage-collect its own previously-applied secrets.
- `qontract_utils.keycloak_api.KeycloakApi` now raises `KeycloakBadRequestError` with an actionable hint (likely clientId collision on create, or a rejected field value on update) instead of a bare `httpx2.HTTPStatusError`, when Keycloak's client-registration endpoint returns 400 Bad Request.
- Fixed two sharding/early-exit bugs found by CodeRabbit review: `get_early_exit_desired_state()`'s shape now actually matches the `state.*.shard` selector (dict+hash, matching `terraform_resources.py`/`openshift_resources_base.py`'s own pattern), and `shard_arg_is_collection=True` is now set so a sharded run doesn't degrade `cluster_name` into a bare string.

Requires the paired [qontract-schemas PR #1077](https://github.com/app-sre/qontract-schemas/pull/1077) adding the `managed-sso-client` openshift-resource provider schema and graphql type.

Jira: https://redhat.atlassian.net/browse/APPSRE-15234

## Test plan
- [x] `uv run pytest reconcile/test/test_openshift_managed_sso_client_secrets.py reconcile/test/test_managed_sso_client.py` - 29 passed
- [x] `cd qontract_api && uv run pytest tests/integrations/managed_sso_client/` - 73 passed
- [x] `cd qontract_utils && uv run pytest tests/keycloak_api/` - 39 passed
- [x] `make linter-test` / `make types-test` clean on all touched files
- [x] Verified `--vault-path-prefix` CLI option and `openshift-managed-sso-client-secrets` command registration manually via `qontract-reconcile --help`
- [x] Verified the fixed early-exit state shape against the real `jsonpath_ng` selector evaluation (test asserts the selector actually matches, not just that the code runs)

## Summary by CodeRabbit

- **New Features**
  - Added support for managing tenant-declared SSO clients.
  - Added an integration that creates OpenShift Secrets containing managed SSO client credentials.
  - Added cluster and namespace filtering, threaded execution, and configurable Vault path prefixes.
  - Added GraphQL support for retrieving managed SSO client and namespace resource data.

- **Bug Fixes**
  - Improved Keycloak registration and update error handling with clearer client-specific messages.

- **Tests**
  - Added coverage for reconciliation, Secret creation, filtering, Vault paths, and error handling.


reconcile/openshift_managed_sso_client_secrets.py
```diff
@@ -0,0 +1,216 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Callable, Mapping
+
+    from reconcile.gql_definitions.openshift_managed_sso_client_secret.openshift_resource_managed_sso_client import (
+        ManagedSsoClientV1,
+        OpenshiftResourceManagedSsoClient,
+    )
+    from reconcile.utils.secret_reader import SecretReaderBase
+
+QONTRACT_INTEGRATION = "openshift-managed-sso-client-secrets"
```
hemslo, line 45: Given the planned migration to ESO, could we reuse the existing `openshift-vault-secrets` integration with `provider: vault-secret` as the temporary bridge? That would keep managed-sso-client responsible for writing credentials to Vault and avoid introducing a second Vault-to-OpenShift Secret reconciler that must later be migrated and removed. Please identify the concrete behavior the existing provider cannot support before adding this integration.

reconcile/openshift_managed_sso_client_secrets.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def get_early_exit_desired_state(self) -> dict[str, Any] | None:
+        return {"namespace": self.get_namespaces(gql.get_api().query)}
+
+    def get_desired_state_shard_config(self) -> DesiredStateShardConfig:
+        return DesiredStateShardConfig(
+            shard_arg_name="cluster_name",
+            shard_path_selectors={
+                "state.*.shard",
+            },
```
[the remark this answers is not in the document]
chassing: Confirmed and fixed in 0b21ae7b, though not quite via the suggested diff: I verified directly with `jsonpath_ng` that `"state.*.shard"` (dot-star) only descends into dict values, not list elements - so the suggested `{"state": [...]}` list shape wouldn't actually have satisfied the existing selector either (0 matches, tested empirically). Went with the dict+hash shape instead, matching the established pattern in `terraform_resources.py`/`openshift_resources_base.py`'s own `early_exit_desired_state()`:

```python
"state": {cluster: {"shard": cluster, "hash": DeepHash(state).get(state)} for cluster, state in ...}
```

Also confirmed you're right that this was a silent no-op, not a crash - `jsonpath_ng` swallows the `AttributeError` from trying `.get()` on a pydantic model internally, so it just always returned zero affected shards rather than raising. Added a test that runs the real jsonpath selector against the real return value to lock this in.

README.md
```diff
@@ -159,6 +161,9 @@ OpenShift templates can be found [here](/openshift/qontract-reconcile.yaml). In
                                   clusters.
   openshift-groups                Manages OpenShift Groups.
   openshift-limitranges           Manages OpenShift LimitRange objects.
```
hemslo, line 163: Please document the temporary `openshift-vault-secrets` configuration and the ESO cutover: Vault path/version, credential keys, destination Secret contract, authentication prerequisites, rotation verification, and rollback order. Ensure the old writer is disabled before ESO takes ownership so both controllers never manage the same Secret concurrently.

reconcile/openshift_managed_sso_client_secrets.py
```diff
@@ -0,0 +1,216 @@
...
+            sharded_run_review=lambda proposal: len(proposal.proposed_shards) <= 2,
```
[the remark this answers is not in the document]
chassing: Confirmed and fixed in 0b21ae7b. I verified this mechanically: `PydanticRunParams.copy_and_update()` (`model_copy(update=...)`) doesn't validate/coerce, so a sharded run really was going to assign a bare string to `cluster_name`, and `ns.cluster.name in self.params.cluster_name` really does become substring containment on a str. It was dormant only because of the shape bug in the sibling comment (no shard was ever detected as affected, so `build_integration_instance_for_shard` was never reached) - fixing that one without this one would have made this bug immediately live, so fixed together in the same commit.

chassing:
@coderabbitai Both sharding/early-exit issues flagged in the "Merge Risk" summary are confirmed and fixed in 0b21ae7b (see individual replies on each inline comment for details - the shape fix ended up different from the suggested diff since the suggested list shape doesn't actually satisfy the existing `state.*.shard` selector; went with a dict+hash shape matching this codebase's own `terraform_resources.py`/`openshift_resources_base.py` pattern instead).

Re: the docstring coverage warning - not planning to address that one. This project's convention (per its own contributor guidelines) is to only comment/docstring where the *why* isn't obvious from the code, not to document every function - the low percentage here reflects that convention, not missing documentation for anything genuinely non-obvious.

reconcile/openshift_managed_sso_client_secrets.py
chassing: I chose a dedicated integration instead of `openshift-vault-secrets` because I don't want to expose the Vault path `app-sre/integrations-throughput/managed-sso-client/output` to tenants. This has already caused a lot of trouble in the past; I mean everything around `app-sre/integrations-throughput|output` and cluster changes. It's also a clear interface for our tenants with a dedicated integration.

hemslo: We can support a custom vault path in sso client request, which tenant controls, then external secrets operator sync it to tenant namespace, they now use vault-secrets with that path for now. A new integration handling secrets sync is a new debt.

chassing: We already have an optional `output` attribute in the `/dependencies/managed-sso-client-1.yml` schema to support a custom Vault path, but that requires the tenant to allow read/write access via a custom Vault policy, not very convenient.

I don't see new tech debt with this integration. Also, in the future, our tenants want to configure the secrets deployment in app-interface (`openshiftResources`), regardless of the technical implementation (ESO vs qontract-reconcile). I think a small layer in qontract-reconcile (app-interface - qr - ESO - k8s) would still be beneficial.

hemslo: `openshiftResources` will be gone from app-interface for tenant namespace, tenant manage k8s workload in their own gitops repo

chassing: This needs to be discussed!

hemslo: There won't be tenant namespace workload or even namespace file in app-interface anymore, all tenant delivery will be managed in https://github.com/openshift-online/hp-gitops-tenants by delivery team, we no longer provide CD for tenants, only infra support for delivery team.

A few PR related to this and with similar pattern:

* https://github.com/openshift-online/platform-engineering-enhancements/pull/44
* https://github.com/openshift-online/platform-engineering-enhancements/pull/61
* https://github.com/openshift-online/platform-engineering-enhancements/pull/76
* https://github.com/openshift-online/platform-engineering-enhancements/pull/67
* https://github.com/openshift-online/architecture/pull/84

I have a similar concern about too many places for tenants to touch, discussion here https://redhat-internal.slack.com/archives/CMK13BP4J/p1788298049386669?thread_ts=1787837833.025849&cid=CMK13BP4J

chassing: Given the uncertainties around HCP Vault and the like, I'd prefer to use a dedicated integration rather than `openshift-vault-secrets` because of the UX for our tenants.

hemslo: Thanks. The intended platform direction is narrower than this integration assumes: AppSRE will provide Vault access to tenants, while ownership of tenant namespaces and workloads is moving out of app-interface to the delivery team and tenant GitOps. Security also requires the credential path to be Vault-to-workload without a third-party copy or access. ESO syncing into a Kubernetes Secret is already a compromise; the preferred target is Secrets Store CSI Driver so workloads consume Vault-backed data directly.

Given that direction, this integration should be explicitly transitional if it is merged. Please mark `openshift-managed-sso-client-secrets` and the corresponding schema/provider contract as temporary/deprecated, and add a migration plan with an owner, target delivery repository, migration/rollback steps, and exit criteria. The plan should cover migrating each consumer from the qontract-reconcile-created Secret to Secrets Store CSI Driver, preventing competing Secret owners during cutover, and deleting the integration, provider, CLI/GraphQL wiring, disable entries, documentation, and tests after migration.

If this is intended to be a permanent tenant-facing Secret delivery layer, that needs a separate architecture decision because it conflicts with the Vault-direct/CSI Driver direction.

chassing: I'll keep an eye on it and address that in future design docs for https://github.com/openshift-online/platform-engineering-enhancements/pull/90. Especially because the migration will be much easier than for `openshift-vault-secrets`.

reconcile/openshift_managed_sso_client_secrets.py
```diff
@@ -0,0 +1,229 @@
...
+        ):
+            ob.aggregate_shared_resources_typed(ns)
+            if (
+                integration_is_enabled(self.name, ns.cluster)
+                and not bool(ns.delete)
+                and (
+                    not self.params.cluster_name
+                    or ns.cluster.name in self.params.cluster_name
+                )
+                and ns.openshift_resources
```
hemslo: [P1] Preserve cleanup coverage when the final resource is removed. Filtering on ns.openshift_resources means a namespace disappears from this integration as soon as its final managed-sso-client resource is deleted. run() then returns before init_specs_to_fetch and fetch_current_state, so the Secret previously created by this integration is never observed or deleted and can retain stale client credentials in the namespace. Please include empty namespaces in the current-state/cleanup scan while keeping them out of desired-state generation, and add a regression test for removing the final resource.

reconcile/utils/managed_sso_client.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import annotations
+
+# Must stay in sync with qontract-api's
+# ManagedSsoClientSettings.default_output_vault_path_prefix default
+# (qontract_api/qontract_api/config.py) - this is the only way this
+# reconcile-side default can be resolved, since qontract-api's config value
+# is never exposed back through app-interface data.
+DEFAULT_OUTPUT_VAULT_PATH_PREFIX = (
```
hemslo: [P2] Avoid a second, independently configurable Vault path default. The API resolves an omitted output using ManagedSsoClientSettings.default_output_vault_path_prefix, which is configurable through qontract-api settings, while this integration resolves the same case from this hard-coded constant and the CLI only warns that it must stay in sync. If the API prefix changes, the API writes tenant credentials to one path while this integration reads another, registers SecretNotFound, and never creates the Kubernetes Secret. Please use a shared or explicitly propagated resolved path instead of manual synchronization, and document the required Vault path/version contract if this default remains.

reconcile/openshift_managed_sso_client_secrets.py
chassing: Confirmed and fixed in 5432ac1c. Dropped the `and ns.openshift_resources` clause from `get_namespaces()` - `fetch_desired_state` already safely iterates `ns.openshift_resources or []`, so a namespace with zero remaining resources still gets scanned for current-state/cleanup but simply contributes no desired secret. Added a regression test (`test_get_namespaces_keeps_namespaces_without_resources_for_cleanup`) that reproduces the orphaned-Secret bug against the old code before the fix.

reconcile/utils/managed_sso_client.py
chassing: Addressed in 80d6dbcc, via the partial-fix option: moved `DEFAULT_OUTPUT_VAULT_PATH_PREFIX` (and `derive_managed_sso_client_id`) into `qontract_utils/managed_sso_client.py`, which both `reconcile/` and `qontract_api/` can import per ADR-007 (neither can import the other directly). `qontract_api`'s `ManagedSsoClientSettings.default_output_vault_path_prefix` now defaults to this shared constant instead of an independent string literal, so the *defaults* can no longer drift. This doesn't close the gap you flagged for a runtime-configured override of the qontract-api setting - reconcile still has no way to observe that at runtime - so I made that explicit in the `--vault-path-prefix` CLI help text instead of leaving it implied by a comment.

README.md
chassing: This ask was contingent on adopting the openshift-vault-secrets bridge suggested in the sibling thread (the one starting at this same line) - that suggestion wasn't adopted, this PR ships a dedicated integration instead, so there's no 'temporary openshift-vault-secrets configuration' in this PR to document as worded. The transitional/deprecation documentation for this dedicated integration itself is being tracked in that sibling thread (see my reply there: I'll address it in a future design doc for openshift-online/platform-engineering-enhancements#90).

reconcile/openshift_managed_sso_client_secrets.py
hemslo: That is a reasonable fallback if making the cleanup scan targeted is too difficult. Since other integrations have the same lifecycle, please make the behavior an explicit contract rather than relying on an implementation detail: either add or support an explicit `delete: true` path in the provider schema for removing the Secret, or document that deleting the final desired resource can leave the Secret orphaned and requires separate cleanup or rotation. The documentation should state the security impact of retaining stale credentials and give the operator or migration procedure; otherwise this fix trades a known orphan case for a broad read of every tenant Secret without defining which behavior is supported.

chassing: You're right, and I got this wrong - reverted in 049d2405. Tracing it further: dropping that filter removed the only thing bounding get_namespaces()'s scope. integration_is_enabled() defaults every cluster to in-scope, and init_specs_to_fetch() then builds an unbounded (resource_names=None) "list every Secret" CurrentStateSpec for every namespace on every one of those clusters - exactly the broad read you flagged.

Checked how openshift_resources_base.py (the reference this integration follows) handles the same scenario: canonicalize_namespaces has the identical `if ors and providers:` filter, so every other openshiftResources-based integration in this codebase - including openshift-vault-secrets - has this same last-resource-removal orphan gap and accepts it as a known trade-off rather than scanning unconditionally. Restored that same behavior here (with a comment documenting why) instead of introducing a fleet-wide broad Secret read to close one edge case. Not pursuing the explicit delete-marker schema idea for this PR - it's a legitimate way to close the gap properly, but it's new schema surface beyond this PR's scope.

hemslo (approved):
LGTM!

reconcile/gql_definitions/openshift_managed_sso_client_secret/namespaces.gql
```diff
@@ -0,0 +1,32 @@
...
+    delete
+    clusterAdmin
+    openshiftResources {
+      ...OpenshiftResourceManagedSsoClient
+    }
+    sharedResources {
+      openshiftResources {
+        ...OpenshiftResourceManagedSsoClient
+      }
+    }
```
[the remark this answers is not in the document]
chassing: Confirmed and fixed in 79b3de4f - verified empirically before fixing: parsing a raw dict shaped like a real GraphQL response for a non-matching (e.g. vault-secret) openshiftResources item against the old generated model raised exactly the ValidationError you describe.

Went with a variant of your suggested fix: selecting `provider` plus an inline `... on NamespaceOpenshiftResourceManagedSsoClient_v1 { ... }` type condition, but with the fragment's fields inlined directly at both occurrence sites (top-level and sharedResources-nested) rather than spread via `...OpenshiftResourceManagedSsoClient`. Your suggested diff (keeping the named fragment spread inside the inline type condition) turned out to produce an empty class body for the matching subtype with this qenerate version - confirmed empirically, so any use of NamespaceOpenshiftResourceManagedSsoClient_v1's fields would have failed extra_forbidden validation on a real managed-sso-client resource. Filed the inline-fields version instead, which does produce a properly composed discriminated union at each site. get_namespaces()/fetch_desired_state() now filter via isinstance against both resulting matching classes (query-path-based codegen naming means the two occurrence sites get distinct classes instead of one shared type). Added a regression test mixing a vault-secret and a managed-sso-client resource on one namespace to lock in the fix.

reconcile/openshift_managed_sso_client_secrets.py
```diff
@@ -0,0 +1,255 @@
...
+            if not dry_run:
+                ri.register_error(cluster=ns.cluster.name)
+            return
+        ri.add_desired_resource(
+            cluster=ns.cluster.name,
+            namespace=ns.name,
+            resource=self.construct_managed_sso_client_secret(
+                resource, client, secret_data
+            ),
+        )
```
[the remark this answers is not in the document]
chassing: Confirmed and fixed in b21725c6, matching openshift_resources_base.fetch_desired_state's identical handling for this exact scenario. Registered the error scoped per-cluster (`register_error(cluster=ns.cluster.name)`) rather than the reference's unscoped call, to match this integration's own existing per-cluster SecretNotFoundError handling and the per-cluster error isolation convention. Wrote the regression test first and confirmed it crashed with an unhandled ResourceKeyExistsError against the old code before adding the fix.

PR #5823, opened 2026-09-16
chore(glhk) remove old-burst rebase strategy

rhur-pixel:
## Summary

Remove the `old-burst` rebase strategy from gitlab-housekeeping. Default is now `active-cap`. Drop `_rebase_merge_requests_old_burst` and the dual-strategy test parametrization.

`active-cap` already caps rebases by in-flight pipelines; `old-burst` only counted rebases per loop and ignored that. Toggle value `old-burst` is no longer valid.

## Summary by CodeRabbit

* **Changes**
  * Rebase operations now use the Active Cap strategy by default.
  * Removed support for the obsolete Old Burst strategy.
  * Updated fallback behavior and accepted strategy variants to use active-cap modes.


esron (approved):
LGTM

PR #5825, opened 2026-09-16
(FedRAMP) fix(openshift-resources): FIPS-safe sha256 for early-exit content hash

bkez322:
## tl;dr
Fixes noisy md5/fips stacktrace errors during FedRAMP Jenkins builds.

## What
Replace the `hashlib.md5()` used to compute the early-exit `content_sha` in `_early_exit_fetch_resource` (`reconcile/openshift_resources_base.py`) with `hashlib.sha256()`.

## Why
`hashlib.md5()` raises `_hashlib.UnsupportedDigestmodError: [digital envelope routines] unsupported` on FIPS-enforcing hosts (observed on a RHEL9 / Python 3.14 executor). That makes `runner.get_desired_state_diff` fail — it's caught, so the run still passes, but early-exit is silently disabled and a full traceback is printed on affected dry-runs. This was the only **unguarded** `md5` in the repo. (The one other md5 — `reconcile/utils/sharding.py` — already passes `usedforsecurity=False` and is intentionally kept as md5 to preserve shard-key stability; everything else uses `sha256`.)

## Why this is safe (no behavioral change)
- **Ephemeral & in-process.** `content_sha` is used only to diff early-exit *desired state*. `get_desired_state_diff` hashes BOTH the comparison-bundle and main-bundle desired states in the same process (same code/algorithm) and diffs them in memory — so swapping the algorithm is self-consistent and the diff's meaning is unchanged.
- **Never persisted.** The digest is not written to any durable store, so there is no cross-run / cross-version comparison and no migration required (unlike the DynamoDB-persisted `resource_hash` in `external_resources`).
- **Cluster drift-detection is unaffected.** What actually gets applied/compared on live objects uses the separate `qontract.sha256sum` annotation (`openshift_resource.py`), which is untouched.
- The only side effect is a one-time, self-healing repopulation of the opt-in extended early-exit cache (its cache-source digest changes) — equivalent to a normal version bump.

---

Ref: [HCMSEC-3667](https://redhat.atlassian.net/browse/HCMSEC-3667)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved desired-state comparison reliability during resource reconciliation by using stronger content verification.
  * Reduced the risk of incorrectly treating resource content as unchanged when differences are present.


suzana-nesic (approved):

hemslo (approved):
LGTM!

PR #5827, opened 2026-09-17
fix(aws-account-manager): self-heal stale FAILED CreateAccount cache on email change

chassing:
## Summary

- When an AWS `CreateAccount` request fails (e.g. `EMAIL_ALREADY_EXISTS`), the integration cached the request ID keyed only by account name. Once a tenant updated the owner email in app-interface and reran the MR, the integration kept polling the same stale, failed request and erroring forever — the only workaround was an AppSRE IC manually deleting the cached S3 state key.
- `_create_account` now stores the request email alongside the request ID (`{"request_id": ..., "email": ...}`). On the next run, if the desired email differs from the cached one, it polls the old request:
  - `FAILED` → logs and automatically retries with the new email (self-heals, no manual cleanup needed)
  - `IN_PROGRESS` → aborts safely, waits for the in-flight request to resolve
  - `SUCCEEDED` → warns that the email change has no effect and returns the existing result
  - Legacy plain-string state values are still supported for backwards compatibility
- `_org_account_exists` now gives an actionable error message when `EMAIL_ALREADY_EXISTS` is the failure reason, telling the tenant to update the owner email in app-interface.

Note: AWS account emails are globally unique across all of AWS (not just within one Organization), and there is no AWS API to check email uniqueness upfront without calling `CreateAccount` — so self-healing after failure is the only reliable fix, rather than upfront validation.

Ref: [APPSRE-15293](https://redhat.atlassian.net/browse/APPSRE-15293)
Originating Slack thread: https://redhat-internal.slack.com/archives/CCRND57FW/p1789640059739279

## Test plan

- [x] 7 new tests added to `reconcile/test/aws_account_manager/test_aws_account_manager_reconciler.py`, written first (TDD) and confirmed failing before the fix
- [x] Full `aws_account_manager` test suite passes: 92/92
- [x] `ruff check` / `ruff format --check` clean on changed files
- [x] `mypy` strict clean on changed files

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved AWS account creation recovery when the request email changes.
  - Failed requests can now be retried automatically, while in-progress requests are safely preserved and completed requests are retained.
  - Existing account-creation state remains compatible with older saved formats.
  - Added clearer guidance when an email address is already associated with an account, including steps to update the owner email for automatic retry.


reconcile/aws_account_manager/reconciler.py
```diff
@@ -72,16 +72,48 @@ def _create_account(
         """Create the organization account and return the creation status ID."""
         with self.state.transaction(state_key(name, TASK_CREATE_ACCOUNT)) as state:
             if state.exists:
-                # account already exists, nothing to do
-                return state.value
+                cached = state.value
+                # legacy state format: a plain request id string
+                if isinstance(cached, str):
+                    return cached
```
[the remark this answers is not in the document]
chassing: No account creations in flight; therefore, not needed.

reconcile/aws_account_manager/reconciler.py
```diff
@@ -72,16 +72,48 @@ def _create_account(
...
 
             logging.info(f"{name}: Creating account")
             if self.dry_run:
                 raise AbortStateTransactionError("Dry run")
 
             status = aws_api.organizations.create_account(email=email, name=name)
-            # store the status id for future reference
-            state.value = status.id
+            # store the status id and email for future reference
+            state.value = {"request_id": status.id, "email": email}
```
hemslo, line 116: [P2] Keep persisted state compatible with older releases

TASK_CREATE_ACCOUNT is durable S3 state. The previous release's _create_account returns state.value directly; if this dictionary is present during a rolling update or rollback, _org_account_exists forwards it as CreateAccountRequestId, and botocore rejects it before AWS. Please keep this key's body as the request-ID string and store the email in a companion key or S3 object metadata. For a companion key, treat missing metadata as legacy, write the ID before the metadata, and add a rollback/old-reader test.

reconcile/aws_account_manager/reconciler.py
```diff
@@ -72,16 +72,48 @@ def _create_account(
...
+                if isinstance(cached, str):
+                    return cached
+
+                if cached["email"] == email:
+                    # account creation already requested for this email, nothing to do
+                    return cached["request_id"]
+
+                # the desired email changed since the last request - check if the
+                # stale request can be retried with the new email
+                old_status = aws_api.organizations.describe_create_account_status(
```
hemslo, line 86: [P3] Avoid rechecking completed requests on every loop

When the email differs and the old request is SUCCEEDED, this branch performs DescribeCreateAccountStatus and returns the request ID. create_organization_account then calls _org_account_exists; if TASK_DESCRIBE_ACCOUNT has not been written yet, it performs the same lookup again. After the UID is stored, the cached email still differs, so every later loop repeats the first lookup before _org_account_exists can short-circuit. Please check terminal TASK_DESCRIBE_ACCOUNT state before querying, and reuse the SUCCEEDED status/UID or persist it so the first successful transition makes only one AWS call. Add tests for existing terminal state, the first SUCCEEDED transition, and repeated runs.

reconcile/aws_account_manager/reconciler.py
chassing: Acknowledged — accepting this risk. In the rare event of a rollback while a dict-shaped `create-account` entry exists, I'll manually fix up the S3 state for the affected account(s) rather than adding a companion-key abstraction for it.

reconcile/aws_account_manager/reconciler.py
chassing: Good catch in theory, but it doesn't apply in practice here: the integration removes the account request from app-interface once account creation succeeds, so a tenant can no longer change the email after the request has resolved to SUCCEEDED. That branch is effectively unreachable from a real desired-state change, so I'll leave it as-is.

hemslo (approved):
LGTM!

PR #5828, opened 2026-09-18
[APPSRE-13624] migrate gitlab-projects to qontract-api

gdragonetti:
## Summary by CodeRabbit

* **New Features**
  * Added GitLab project reconciliation across configured instances and groups.
  * Added dry-run support, project creation, and SaaS bundle repository initialization.
  * Added asynchronous task submission and status tracking APIs.
  * Added CLI filtering by GitLab instance and project names.
  * Added validation for duplicate or undeclared project requests.
  * Added caching to improve GitLab group and project lookups.
* **Bug Fixes**
  * Improved reconciliation resilience by reporting individual errors without stopping other projects or instances.
* **Documentation**
  * Added documentation for configuring and using the GitLab Projects integration.


gdragonetti:
# Change summary

**qontract_api/**
├── **openapi.json** — regenerated OpenAPI spec, adds the gitlab-projects endpoints/schemas
├── **qontract_api/**
│   ├── **config.py** — new GitlabProjectsSettings (api_timeout, group_projects_cache_ttl)
│   ├── **gitlab/**
│   │   ├── **__init__.py**
│   │   ├── **gitlab_client_factory.py** — Layer 2 factory: resolves secret, builds GitlabWorkspaceClient
│   │   └── **gitlab_workspace_client.py** — Layer 2: cached + distributed-locked group/project access (ADR-014)
│   ├── **integrations/gitlab_projects/**
│   │   ├── **__init__.py** — exposes the router
│   │   ├── **router.py** — FastAPI endpoints: POST /reconcile, GET /reconcile/{task_id}
│   │   ├── **schemas.py** — request/response/action Pydantic models
│   │   ├── **service.py** — Layer 3: diff + apply reconciliation, atomic create+SaaS-bundle actions
│   │   └── **tasks.py** — Celery task entrypoint + instance-set dedup lock key
│   ├── **routers/integrations.py** — wires gitlab_projects_router into the /integrations router
│   └── **tasks/__init__.py** — registers gitlab_projects.tasks in the Celery worker's include list
└── **tests/**
    ├── **gitlab/**
    │   ├── **conftest.py** — shared fixtures: mock_gitlab_api, mock_cache, mock_settings
    │   ├── **test_gitlab_client_factory.py** — Layer 2 factory unit tests
    │   └── **test_gitlab_workspace_client.py** — Layer 2 cache/lock/mutation unit tests
    └── **integrations/gitlab_projects/**
        ├── **test_gitlab_projects_service.py** — Layer 3 diff/apply/error-isolation unit tests
        ├── **test_router.py** — FastAPI endpoint unit tests
        └── **test_tasks.py** — lock-key generation unit tests

**qontract_api_client/**
└── **qontract_api_client/**
    ├── **client.py** — regenerated: adds gitlab_projects()/gitlab_projects_task_status() calls
    └── **schemas.py** — regenerated: adds GitlabGroupConfig/GitlabInstanceConfig/Action/TaskResult schemas

**qontract_utils/**
├── **qontract_utils/gitlab_api/**
│   ├── **__init__.py**
│   ├── **client.py** — Layer 1: stateless GitlabApi (group/project/file/branch ops, hooks, dedicated metrics)
│   └── **models.py** — Layer 1 Pydantic models: GitlabGroup, GitlabProject
└── **tests/test_gitlab_api.py** — Layer 1 unit tests

**reconcile/**
├── **cli.py** — registers the `gitlab-projects-api` CLI command
├── **gitlab_projects_api.py** — client-side integration: GraphQL fetch, codeComponent fail-fast validation, calls qontract-api
├── **gql_definitions/gitlab_projects/**
│   ├── **__init__.py**
│   ├── **app_code_components.gql** / **app_code_components.py** — GraphQL query + generated model for app codeComponents
│   └── **gitlab_instances.gql** / **gitlab_instances.py** — GraphQL query + generated model for GitLab instances/projectRequests
└── **test/test_gitlab_projects_api.py** — client-side integration unit tests

# Functional behavior changes
- No longer hard-coded to support a single gitlab instance (vs. using instances[0])
- Any project without associated code components will cause client side failure (vs. checking condition only on new projects)

qontract_utils/qontract_utils/gitlab_api/client.py
```diff
@@ -0,0 +1,196 @@
...
+            id=project.id,
+            name=project.name,
+            path_with_namespace=project.path_with_namespace,
+            web_url=project.web_url,
+        )
+
+    @invoke_with_hooks(
+        lambda self: GitlabApiCallContext(method="create_file", url=self.url)
+    )
+    def create_file(
```
chassing, line 146: I would like to see no VCS actions in the `gitlab_api` layer. Just GitLab specific operations. Please use the `qontract_utils/vcs` instead

qontract_utils/qontract_utils/gitlab_api/client.py
```diff
@@ -0,0 +1,196 @@
...
+        lambda self: GitlabApiCallContext(method="create_branch", url=self.url)
+    )
+    def create_branch(
+        self, project_id: int, new_branch: str, source_branch: str
+    ) -> None:
+        """Create a branch from an existing source branch."""
+        project = self._gitlab.projects.get(project_id, lazy=True)
+        project.branches.create({"branch": new_branch, "ref": source_branch})
+
+    def initiate_saas_bundle_repo(self, project_id: int) -> None:
```
chassing, line 176: I'm not familiar with this feature. Could you please elaborate on that? Is it still required?

gdragonetti: I am not sure whether it is still used but it is still in place in the gitlab-projects current integration and I see some code components carry this resource type ([ex](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/osd-operators/app.yml?ref_type=heads#L388))

If the code component has `resource: bundle` it adds a README and staging and production branches to the created project

gdragonetti: As pointed by code rabbit this part is not really re-entering, if some of these additional steps fail but the project was created then we will skip these steps on all subsequent runs.

Didn't realize this before and copied original behavior but I am not sure whether it is worth it to do several more gitlab API calls to verify on all created bundle projects that they satisfy all conditions. What do you recommend?

qontract_api/qontract_api/integrations/gitlab_projects/tasks.py
```diff
@@ -0,0 +1,104 @@
...
+                        data=action.model_dump(mode="json"),
+                        datacontenttype="application/json",
+                    )
+                )
+            for error in result.errors or []:
+                event_manager.publish_event(
+                    Event(
+                        source=__name__,
+                        type="qontract-api.gitlab-projects.error",
+                        data={"error": error},
```
chassing: `data={"error": error}` is a raw dict for structured event data, which AGENTS.md explicitly prohibits ("no arbitrary dictionaries ... use proper data classes"). The sibling `managed_sso_client` integration already solves this with a typed `ManagedSsoClientErrorEvent(BaseModel, frozen=True)` (`error: str`) in its schemas.py, used as `data=ManagedSsoClientErrorEvent(error=error).model_dump(mode="json")`. Please add an equivalent typed event model here instead of the inline dict.

qontract_api/tests/integrations/gitlab_projects/test_router.py
```diff
@@ -0,0 +1,345 @@
...
+        "/api/v1/integrations/gitlab-projects/reconcile",
+        json=sample_reconcile_request,
+        headers=auth_headers,
+    )
+    assert response.status_code == HTTPStatus.UNPROCESSABLE_ENTITY
+    assert "duplicate project names" in response.text
+
+
+# ---------------------------------------------------------------------------
+# GET /reconcile/{task_id}  # ruff: ignore[commented-out-code]
```
chassing, line 205: This section-divider comment has a stray trailing fragment (`# ruff: ignore[commented-out-code]` isn't valid ruff suppression syntax, and there's no commented-out code here to suppress). Looks like a leftover artifact — harmless, but worth cleaning up.

qontract_utils/qontract_utils/gitlab_api/client.py
chassing, line 176: ADR-014's Layer 1 checklist requires "no business logic" — pure communication primitives only. This method encodes a specific domain policy (which branches a "SaaS bundle repo" needs, README content/message, ordering) rather than being a primitive itself. `create_file`/`create_branch` are fine as Layer 1 primitives, but this composition reads as Layer 2/3 orchestration and might fit better in the workspace client, built from these primitives.

qontract_utils/qontract_utils/gitlab_api/client.py
```diff
@@ -0,0 +1,196 @@
...
+        self,
+        url: str,
+        token: str,
+        *,
+        ssl_verify: bool = True,
+        timeout: int = 30,
+        hooks: Hooks | None = None,  # ruff: ignore[unused-method-argument] - Handled by @with_hooks decorator
+    ) -> None:
+        self.url = url
+        self._gitlab = gitlab.Gitlab(
```
chassing, line 105: Other Layer 1 clients in this repo pass `user_agent=DEFAULT_USER_AGENT` (from `qontract_utils.user_agent`) so external services can attribute traffic to qontract-utils — including the sibling `qontract_utils/vcs/providers/gitlab_client.py:GitLabRepoApi`, which uses the same python-gitlab library. This client omits it. Please add `user_agent=DEFAULT_USER_AGENT` (or an injectable param, matching `GitLabRepoApi`'s pattern) to the `gitlab.Gitlab(...)` call.

reconcile/gitlab_projects_api.py
```diff
@@ -0,0 +1,240 @@
...
+
+    @staticmethod
+    def filter_requested_projects(
+        gl_instances: list[GitlabInstanceV1],
+        allowed_instance_name: str | None,
+        allowed_project_names: frozenset[str] | None,
+    ) -> list[GitlabInstanceV1]:
+        """Narrow instances/projects to the given filters, for targeted debug runs."""
+        if allowed_instance_name:
+            gl_instances[:] = [
```
chassing: This mutates the input list (`gl_instances[:] = ...`) and the GQL-result model objects themselves (`instance.project_requests = ...`, `project_request.projects = ...` below) instead of building new filtered structures. The sibling `reconcile/openshift_namespaces_api.py:178-183` does equivalent filtering functionally (no mutation). Not a live bug today (the one caller reassigns to the same variable), but a method named "filter" silently mutating its input is a footgun for future callers, and it's inconsistent with the sibling pattern. Recommend rebuilding new lists instead.

reconcile/gitlab_projects_api.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+class GitlabProjectsIntegrationParams(PydanticRunParams):
+    """Parameters for gitlab-projects-api integration."""
+
+    instance_name: str | None = None
+    project_names: frozenset[str] | None = None
+
+
+class GitLabProjectsIntegration(
```
chassing, line 66: Every other `_api` integration class (`QuayReposIntegration`, `OpenShiftNamespacesIntegration`, `GithubOwnersIntegration`, `SlackUsergroupsIntegration`) has a docstring summarizing what it does. Minor consistency nit — consider adding one here too.

qontract_utils/qontract_utils/gitlab_api/client.py
```diff
@@ -0,0 +1,196 @@
...
+
+@dataclass(frozen=True)
+class GitlabApiCallContext:
+    """Context information passed to API call hooks."""
+
+    method: str
+    url: str
+
+
+def _metrics_hook(context: GitlabApiCallContext) -> None:
```
chassing, line 55: This client duplicates a fair amount of hook/metrics/latency-tracking boilerplate already present in the sibling `qontract_utils/vcs/providers/gitlab_client.py:GitLabRepoApi` (both wrap python-gitlab with near-identical `_metrics_hook`/`_latency_start_hook`/`_latency_end_hook`/`_request_log_hook` scaffolding and a `_latency_tracker` ContextVar). They do serve different scopes (instance-level group/project admin vs. single-project content/MR management), so this isn't a strict violation — just worth a second look for whether the connection/hook scaffolding could be shared.

chassing (commented):
Automated review pass covering the qontract-api server layer (service/tasks/router/schemas, ADR-014 three-layer compliance, event publishing, caching) and the client-side layer (Layer 1 GitLab client, client-side integration, GQL definitions). Overall this is a solid, well-tested migration that closely follows established sibling patterns (managed_sso_client, openshift_namespaces_api). A handful of findings below, none blocking in my view but worth a look.

chassing:
**Missing integration documentation**

This PR adds a new qontract-api integration (`gitlab-projects`) but there's no `docs/integrations/gitlab-projects.md`. Every other migrated integration has one (e.g. `managed-sso-client.md`, `openshift-namespaces.md`), and `docs/integrations/template.md` exists specifically to make this easy to produce. Per the project's migration checklist, documentation for new integrations is required, not optional. Could you add one before merging?

qontract_utils/qontract_utils/gitlab_api/client.py
gdragonetti: Unless I am looking in the wrong place ([vcs/provider_protocol.py](https://github.com/app-sre/qontract-reconcile/blob/master/qontract_utils/qontract_utils/vcs/provider_protocol.py)), it currently does not support committing a file directly or branch generation, do you want me to extend the interface?

qontract_api/qontract_api/gitlab/gitlab_workspace_client.py
```diff
@@ -0,0 +1,107 @@
...
+    # ------------------------------------------------------------------
+
+    def create_project(
+        self, group_name: str, group_id: int, name: str
+    ) -> GitlabProject:
+        """Create a project under the group and invalidate its cached project list."""
+        cache_key = self._cache_key_group(group_name)
+        with self.cache.lock(cache_key):
+            project = self.gitlab_api.create_project(group_id, name)
+            self.cache.delete(cache_key)
```
fishi0x01: `create_project()` calls `gitlab_api.create_project()` then `self.cache.delete(cache_key)` with no `try/finally`. If the create succeeds server-side but the call raises afterward (e.g. a client-side read timeout), the exception propagates and cache invalidation is skipped — the group's cached project list keeps missing this project for up to `group_projects_cache_ttl`. Meanwhile `_calculate_actions` keeps re-issuing a create action for a project that already exists, and every retry fails with GitLab's "already exists" error until the cache expires.

Suggest wrapping the delete in a `finally` (or moving it before the early return path) so invalidation always happens once the create request has been sent.

qontract_api/qontract_api/gitlab/gitlab_workspace_client.py
```diff
@@ -0,0 +1,107 @@
...
+    # ------------------------------------------------------------------
+    # Mutations
+    # ------------------------------------------------------------------
+
+    def create_project(
+        self, group_name: str, group_id: int, name: str
+    ) -> GitlabProject:
+        """Create a project under the group and invalidate its cached project list."""
+        cache_key = self._cache_key_group(group_name)
+        with self.cache.lock(cache_key):
```
fishi0x01, line 91: This acquires the distributed cache lock and issues `cache.delete()` on every single `create_project()` call. When bootstrapping N missing projects under a group (the common case for a newly-declared group), `_apply_actions` calls this once per project — producing N sequential lock+delete round trips to the cache backend where only the last one has any effect.

Worth invalidating the group's cache once after all creates for that group are applied, rather than per-project.

qontract_api/qontract_api/integrations/gitlab_projects/service.py
```diff
@@ -0,0 +1,239 @@
...
+    @staticmethod
+    def _calculate_actions(
+        instance_name: str,
+        group_name: str,
+        desired_projects: list[GitlabProjectConfig],
+        existing_project_names: list[str],
+    ) -> list[GitlabProjectAction]:
+        # Actions are desired projects whose name isn't already present in the
+        # group. Deletions are deliberately never actioned.
+        existing = set(existing_project_names)
```
fishi0x01: This hand-rolls the desired-vs-current diff with `set()` + comprehensions. AGENTS.md's Service Layer Patterns section says to use `qontract_utils.differ.diff_iterables` for state comparison and explicitly "never hand-roll set operations" — every sibling integration (`github_owners`, `ocm_groups`, `ocm_oidc_idp`, `glitchtip_project_alerts`, `slack_usergroups`) follows that convention. Using `diff_iterables` here would keep this consistent and let this integration inherit any future fixes made centrally.

qontract_api/qontract_api/integrations/gitlab_projects/schemas.py
```diff
@@ -0,0 +1,119 @@
...
+
+from collections import Counter
+from typing import Literal
+
+from pydantic import BaseModel, Field, field_validator
+
+from qontract_api.models import Secret, TaskResult, TaskStatus
+
+
+class GitlabProjectConfig(BaseModel, frozen=True):
```
fishi0x01: AGENTS.md's "qontract-api Integration File Layout" calls for pure domain models (internal business-logic types) to live in a separate `domain.py`, with `schemas.py` reserved for API request/response types. `GitlabProjectConfig`, `GitlabGroupConfig`, and `GitlabInstanceConfig` here are desired-state domain models consumed by `service.py`'s reconciliation logic, not wire types — they'd fit the documented layout better in a new `domain.py`, alongside the actual API schemas (`GitlabProjectsReconcileRequest`, `GitlabProjectsTaskResult`, etc.) staying in `schemas.py`. Most existing integrations (8/10: `github_owners`, `glitchtip`, `glitchtip_project_alerts`, `managed_sso_client`, `ocm_groups`, `ocm_oidc_idp`, `openshift_namespaces`, `sso_client`) follow this split.

qontract_api/qontract_api/integrations/gitlab_projects/schemas.py
```diff
@@ -0,0 +1,119 @@
...
+    )
+
+    @field_validator("projects")
+    @classmethod
+    def project_names_unique(
+        cls, projects: list[GitlabProjectConfig]
+    ) -> list[GitlabProjectConfig]:
+        duplicate_counter = Counter(project.name for project in projects)
+        duplicates = {name for name, count in duplicate_counter.items() if count > 1}
+        if duplicates:
```
fishi0x01: AGENTS.md's Coding Conventions: "Prefer walrus operator (`if value := expression:`) over `value = expression; if value:`". Lines 38-39 are exactly that pattern — could be:

```python
if duplicates := {name for name, count in duplicate_counter.items() if count > 1}:
```

Minor, but worth flagging since the correct idiom is used a file away in `gitlab_workspace_client.py` (`if cached := self.cache.get_obj(...)`).

fishi0x01 (commented):
A few additional findings from a follow-up pass (correctness risks + convention gaps), on top of the existing review threads.

qontract_api/tests/integrations/gitlab_projects/test_router.py
gdragonetti: It is indeed complaining if I don't put the ignore, it happens to be "valid" python:

GET / reconcile / {task_id} reads to the tokenizer as "divide GET by reconcile, divide that by a set containing task_id" — a syntactically valid Python expression.

qontract_api/qontract_api/gitlab/gitlab_workspace_client.py
gdragonetti: I am not sure what the best way to go about this is.
Although not the best performance wise I don't think it's a good choice to introduce this behavior for a couple of reasons:
1. Crash of the worker after delaying the cache invalidation will lead to chain-failures until cache reaches TTL (ex. 1 of 3 projects is created then worker crashes -> cache still sees 0 of 3 projects created -> next run tries to recreate the 1st project, failing -> cache is left untouched -> repeat ... and so on until cache record expires). If instead we do an invalidation at the start with no repetition we might get off easier but that's only if we assume that only one workspace client exists at a time, leading to point 2.
2. To implement this behavior we need to either manage the invalidation in the business logic (layer 3) which is a nono or implement a stateful layer 2, which seems not desirable, especially to handle cache invalidation optimizations, since this model fails if more than one process is doing operations and risking accessing an invalid cache record.

A possible non-solution would be to move from instance-group -> projects list cache mappings to an instance-group-project -> True (to signal project existence), that way we would not invalidate the same record multiple times, but we are still invalidating N records, so yes, a non-solution :sweat_smile:

qontract_utils/qontract_utils/gitlab_api/client.py
chassing: Please take a look at https://github.com/app-sre/qontract-reconcile/blob/master/qontract_utils/qontract_utils/vcs/provider_protocol.py#L73
and for its usage https://github.com/app-sre/qontract-reconcile/blob/master/qontract_api/qontract_api/external/vcs/vcs_workspace_client.py#L119

qontract_utils/qontract_utils/gitlab_api/client.py
chassing: Please move the logic up one layer.

qontract_api/tests/integrations/gitlab_projects/test_router.py
chassing: wth!!!

qontract_api/qontract_api/gitlab/gitlab_workspace_client.py
fishi0x01: @gdragonetti you are right - lets keep as is :+1:

fishi0x01 (approved):
Great job! LGTM!

qontract_utils/qontract_utils/gitlab_api/client.py
hemslo: [P3] Validate and remove the `resource: bundle` bootstrap if no live consumer requires it.

The App Interface consumers I checked use bundle content from `main`; they do not establish the hard-coded `master`/`staging`/`production` contract implemented here. This post-create README/branch bootstrap also leaves a partially initialized project if it fails, and its separate `GitLabRepoApi` path cannot carry the configured custom `ssl_verify` or CA settings. Please confirm whether this feature is still required; if not, remove `is_saas_bundle`/`initiate_saas_bundle_repo` and keep project creation plain, which also removes the retry and bootstrap TLS complexity.

